PR #621, opened 2018-02-28
[schema] Change slug uniqueness function name to `isUnique`

rexxars:
This changes the name of the uniqueness check from `checkUnique` to `isUnique`. This makes it easier to reason about what the function should return.

I also snuck in a small change where the default uniqueness check is passed down to custom `isUnique` functions so that you can compose it more easily.

### DIFF
--- a/packages/@sanity/schema/src/legacy/types/slug.js
+++ b/packages/@sanity/schema/src/legacy/types/slug.js
@@ -19,7 +19,7 @@ function getDocumentIds(id) {
   }
 }
 
-const defaultChecker = (slug, options) => {
+const defaultIsUnique = (slug, options) => {
   const {document, path} = options
   const {published, draft} = getDocumentIds(document._id)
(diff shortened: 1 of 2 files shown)


packages/@sanity/schema/src/legacy/types/slug.js
```diff
@@ -46,9 +46,9 @@ const SLUG_CORE = {
       }
 
       const errorMessage = 'Slug is already in use'
-      const checkUnique = get(options, 'type.options.checkUnique', defaultChecker)
-      return Promise.resolve(checkUnique(value.current, options)).then(
-        hasDupe => (hasDupe ? errorMessage : true)
+      const isUnique = get(options, 'type.options.isUnique', defaultUniqueCheck)
```
bjoerge: Hm... maybe it should have been named `defaultIsUnique` instead of `defaultUniqueCheck`?

PR #626, opened 2018-03-01
[client] Emit error on unexpected disconnect

rexxars:
This reverts the change introduced in 292ed3bc6 and changes the behavior to emitting an error instead. Because this error usually happens on long offline periods, simply reconnecting usually won't be enough since we can't give any history back.

This is the first step needed for @bjoerge to ensure re-fetching and re-subscribing happens when the client is disconnected.


rexxars:
We decided not to go through with this.

PR #628, opened 2018-03-01
[cli] Allow creating new projects in unattended mode

rexxars:
This restructures the CLI init process a bit in order to allow creation of new projects in unattended mode (through the `--create-project <name>` flag), as well as skipping prompts when CLI-flags has been given to specify a project ID or dataset.


packages/@sanity/cli/src/actions/init-project/initProject.js
```diff
@@ -322,9 +289,123 @@ export default async function initSanity(args, context) {
     return prompt.single({
       type: 'confirm',
       message: message || 'This template includes a sample dataset, would you like to import it?',
-      default: true
+      defaawdult: true
```
bjoerge: Is this intentional? 😆

rexxars: uuuuh...

PR #648, opened 2018-03-06
[schema] Fix validation of slugs in arrays

rexxars:
Fix a bug where the uniqueness check of slugs fails because it is placed inside the path of an array, as well as the validation failing when placed directly inside of an array and not having a current value.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/Slug/SlugInput.js
+++ b/packages/@sanity/form-builder/src/inputs/Slug/SlugInput.js
@@ -32,6 +32,7 @@ export default withDocument(
       }),
       document: PropTypes.shape({_id: PropTypes.string}).isRequired,
       onChange: PropTypes.func,
+      onFocus: PropTypes.func,
       markers: PropTypes.arrayOf(
         PropTypes.shape({
           type: PropTypes.string.isRequired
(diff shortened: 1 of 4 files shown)


bjoerge:
Also added default preview config for slug type, and fixed the issue with editing we saw earlier. It was caused by an invalid focus path being emitted (not entirely sure how 😅)

packages/@sanity/form-builder/src/inputs/Slug/SlugInput.js
```diff
@@ -55,14 +56,14 @@ export default withDocument(
       this._isMounted = false
     }
 
-    updateValue(value) {
-      if (!value) {
-        this.props.onChange(PatchEvent.from(unset()))
+    updateCurrent(current) {
+      const {onChange, type} = this.props
+      if (!current) {
+        onChange(PatchEvent.from(unset(['current'])))
```
rexxars: Won't this cause the object to remain? Isn't it generally a recommended pattern to unset the whole object when it is empty?

bjoerge: The problem appears when editing slugs contained in an array and the user removes the last character in the input. Then the editing dialog will suddenly (and unexpectedly, IMO) be closed cause the item gets removed.
The pattern should rather be to check for emptiness when the dialog closes, but that is a concern for the array input I'd say.

bjoerge: Oh, btw, I don't mean to say this is how it _should_ work, I'm rather describing how it works now. I'm all for a discussion about collapsing empty values, but there are quite a few considerations we need to do before supporting it.

rexxars: Interesting problem. When it is *not* inside of an array, you'll be left with a whole bunch of empty `{_type: "slug"}` objects, which isn't great either?

rexxars: Also, take the unicorn slider from the custom input widgets example. It generates an unset patch on empty values for the current path. Won't that also cause the same issue?

rexxars: This PR will fix a bunch of issues regardless of this conversation, so let's continue somewhere else.

bjoerge: 👍 agreed, let's talk tomorrow.

PR #649, opened 2018-03-07
Remove typo in form-builder example in README.MD

jaaberg:


bjoerge:
Ah, thanks!

This reminds me that this README is embarrassingly outdated and whatever it says should be taken with a huge grain of salt.

PR #656, opened 2018-03-13
[components] Make arrow ignore pointer-event and clicks the select

kristofferj:


bjoerge:
What does this fix?

kristofferj:
This fixes that the arrow on DefaultSelect does not open the <select>. This fix sets pointer-events to none. That means that clicking the arrow is the same as clicking the <select>, and will open the <select>

bjoerge:
Thanks! LGTM.

PR #670, opened 2018-03-15
[cli] Use React 16 by default in new studios

bjoerge:
This will use v16 as default React version for new studios (prop-types is still on 15.6).

### DIFF
--- a/packages/@sanity/cli/src/versionRanges.js
+++ b/packages/@sanity/cli/src/versionRanges.js
@@ -7,8 +7,9 @@ export default {
     '@sanity/default-layout': 'latest',
     '@sanity/default-login': 'latest',
     '@sanity/desk-tool': 'latest',
-    react: '^15.6',
-    'react-dom': '^15.6'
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/versionRanges.js
```diff
@@ -7,8 +7,9 @@ export default {
     '@sanity/default-layout': 'latest',
     '@sanity/default-login': 'latest',
     '@sanity/desk-tool': 'latest',
-    react: '^15.6',
-    'react-dom': '^15.6'
+    react: '^16.2',
+    'react-dom': '^6.2',
```
rexxars: I think this should be `^16.2`? :)

bjoerge: Ouch 🤦‍♂️  Of course, good catch 😅

bjoerge: (fixed)

PR #701, opened 2018-03-25
add postcss-url dependency to webpack-integration

dan-dr:
v1 and v2 need it and it's not listed as a dependency


rexxars:
Thanks again! :)

Issue #527, opened 2018-01-25
RadioSelect should have a .focus-method


bjoerge:
RadioSelect is currently not focusable.

Not exactly sure whether `.focus()` should trigger focus on the first radio input in it, or if we should add tabIndex=0 and set focus on a wrapper element.

kristofferj:
I think it should focus the first item by default.

kristofferj:
https://github.com/sanity-io/sanity/pull/711

kristofferj:
This is now in a PR, and it focuses the first item  if no selcted, or the currently selected one.

PR #711, opened 2018-04-04
Fix/focus radio

kristofferj:
related to this RadioSelect should have a .focus-method #527


packages/@sanity/form-builder/src/inputs/SelectInput.js
```diff
@@ -27,6 +28,10 @@ export default class StringSelect extends React.Component<Props> {
     value: ''
   }
 
+  componentDidMount() {
+    this.name = uniqueId('RadioName')
```
bjoerge: It would be better if this is set either in the constructor or as a class field.

packages/@sanity/form-builder/src/inputs/SelectInput.js
```diff
@@ -45,19 +50,19 @@ export default class StringSelect extends React.Component<Props> {
...
     const isRadio = type.options && type.options.layout === 'radio'
+
     return (
       <FormField markers={markers} level={level} label={type.title} description={type.description}>
         {isRadio ? (
           // todo: make separate inputs
           <RadioSelect
             {...rest}
-            name={type.name}
+            name={this.name || type.name}
```
bjoerge: So `this.name` will never be false, right? Can just remove `|| type.name` from here I think.

kristofferj:
NOT READY to merge: note to self before merging. selectContainer is composed by statelessSearchableSelect. Must be moved there, or taken back.

PR #714, opened 2018-04-04
[components] Fixing styling bugs and removing focus completly from va…

kristofferj:
related to Validation warnings have display bug on fields / inconsistent display #180


kristofferj:
@rexxars Could you look at this?

PR #720, opened 2018-04-05
[server] Allow specifying a basePath for a studio

### LINKED ISSUE #692: The studio can't run if not on the root domain
If I try to put the studio inside a path in my site,  for example `https://www.example.com/studio` it doesn't work. I have to put it directly on `https://www.mysite.com`.
I understand some work must be done around the router. Not sure if it's a bug or a feature suggestion, but it's a pretty big deal for me as I'm trying to put this on my static web site (gatsby.js).

rexxars:
When building a studio for use outside of development and `sanity.studio`-context, you might want not want to host it at the root of a domain name. In order to allow for this, we need to:

- Make the CLI tool build an index file containing paths to static files that is not always fixed to `/static`
- Make the studio aware of the base path, prefixing the routes with it
- Make it possible to customize the base path through configuration

This PR implements the above, and also (for consistency) makes the development server use the same base path. I felt this would make it easier to debug routing issues, among other things.

Using it is as simple as adding a `basePath` property under the `project` object in `sanity.json`:
```js
{
  // ...
  "project": {
    "name": "Movies studio",
    "basePath": "/studio"
  },
  // ...
}
```

When you access `/` in a development environment and a base path is set, we need to redirect to the correct location. This is currently handled inside of `@sanity/default-layout` since it contains the root router, but we could consider moving this to the `SanityRoot` component in `@sanity/base` in order to ensure this is "always" applied regardless of layout. Thoughts?

Fixes #692

### DIFF
--- a/packages/@sanity/core/src/actions/build/buildStaticAssets.js
+++ b/packages/@sanity/core/src/actions/build/buildStaticAssets.js
@@ -31,7 +31,8 @@ export default async (args, context) => {
     outputPath: path.join(outputDir, 'static'),
     sourceMaps: flags['source-maps'],
     skipMinify: !flags.minify,
-    profile: flags.profile
+    profile: flags.profile,
+    project: config.get('project')
   }
 
   await tryInitializePluginConfigs({workDir, output})
@@ -96,7 +97,7 @@ export default async (args, context) => {
     const indexStart = Date.now()
     spin = output.spinner('Building index document').start()
     const doc = await getDocumentElement(
-      {...compilationConfig, hashes: chunkMap, project: config.get('project')},
+      {...compilationConfig, hashes: chunkMap},
       {
         scripts: ['vendor.bundle.js', 'app.bundle.js'].map(asset => {
           const assetPath = absoluteMatch.test(asset) ? asset : `js/${asset}`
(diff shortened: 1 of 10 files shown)


dan-dr:
Hey I found out it'd be very useful if I could specify the basePath as an option in the build command, because it'll make it much easier to use in CI and build systems (my basePath is different in production/staging).

I found how to do it, will probably PR soon.

packages/@sanity/core/src/actions/build/buildStaticAssets.js
```diff
@@ -96,7 +97,7 @@ export default async (args, context) => {
     const indexStart = Date.now()
     spin = output.spinner('Building index document').start()
     const doc = await getDocumentElement(
-      {...compilationConfig, hashes: chunkMap, project: config.get('project')},
```
dan-dr: Regarding adding the ability to specify basePath as an option in the cli, I was thinking of using it like this:

```javascript
    const doc = await getDocumentElement(
      {...compilationConfig, hashes: chunkMap},
      {
        basePath: flags.basePath || (project && project.basePath) || '',
        scripts: ['vendor.bundle.js', 'app.bundle.js'].map(asset => {
          const assetPath = absoluteMatch.test(asset) ? asset : `js/${asset}`
          return {
            path: assetPath,
            hash: chunkMap[assetPath] || chunkMap[asset]
          }
        })
      }
    )
`
where somewhere before you do `const project = config.get('project')`

PR #728, opened 2018-04-08
[image-url] Fix parsing image source

enigosi:
According to documentation `image(source)` "Accepts either a Sanity image record, an asset record, or just the asset id as a string."

`@sanity/image-url` from master works only when whole Sanity image record is passed. For an asset record it is returning `null` and for asset id as string (`_ref`) is trowing an exception.

### DIFF
--- a/packages/@sanity/image-url/src/urlForImage.js
+++ b/packages/@sanity/image-url/src/urlForImage.js
@@ -65,19 +65,19 @@ export function parseSource(source) {
   let image
 
   // Did we just get an asset id?
-  if (typeof source == 'string') {
+  if (typeof source === 'string') {
     image = {
       asset: {_ref: source}
     }
   } else if (
-    source._type == 'sanity.imageAsset' ||
-    (typeof source == 'object' && typeof source._ref == 'string')
+    source._type === 'sanity.imageAsset' ||
+    (typeof source === 'object' && typeof source._ref === 'string')
   ) {
     // We just got passed an asset directly
     image = {
       asset: source
     }
-  } else if (typeof source == 'object' && typeof source.asset == 'object') {
+  } else if (typeof source === 'object' && typeof source.asset === 'object') {
     image = source
   } else {
     // We got something that does not look like an image, or it is an image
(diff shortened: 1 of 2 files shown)


packages/@sanity/image-url/test/parseSource.test.js
```diff
@@ -0,0 +1,28 @@
...
+import {imageWithNoCropSpecified, croppedImage} from './fixtures'
+
+function compareParsedSource(outputSource, exptectedSource) {
+  should(outputSource).be.an.Object()
+  should(outputSource.asset).be.an.Object()
+  should(outputSource.asset._ref).be.eql(exptectedSource.asset._ref)
+  should(outputSource).have.keys('crop', 'hotspot')
+}
+
+describe.only('[image-url] parseSource', () => {
```
rexxars: Could you remove the `only` flag?

packages/@sanity/image-url/src/urlForImage.js
```diff
@@ -61,15 +61,18 @@ export default function urlForImage(options) {
...
   // Did we just get an asset id?
   if (typeof source == 'string') {
     image = {
-      asset: {_ref: image}
+      asset: {_ref: source}
     }
-  } else if (source._type == 'sanity.imageAsset') {
+  } else if (
+    source._type == 'sanity.imageAsset' ||
+    (typeof source == 'object' && typeof source._ref == 'string')
```
rexxars: Could you use triple equals (`===`) for these checks?

packages/@sanity/image-url/test/parseSource.test.js
enigosi: 😳Oops

packages/@sanity/image-url/src/urlForImage.js
enigosi: Sure, I didn't want to change convention.

rexxars: We should probably fix the other occurrences as well, but leave that up to us. Seems like an oversight on our part

rexxars:
Thanks a bunch!

enigosi:
Great! Cheers

Issue #708, opened 2018-04-03
List dragging outside of field starts text selection (Safari)


atombender:
Steps:

1. Click and drag a list item.
2. Drag it ouside the field.
3. Text will be selected.

![sanity](https://user-images.githubusercontent.com/50314/38262850-5e8e02b8-373c-11e8-8e42-5be32f5d4ff1.gif)

Safari 11. Does not happen in Chrome.

kristofferj:
Confirmed problem in safari. Not in chrome

kristofferj:
This should be fixed in #708

PR #737, opened 2018-04-16
[components] User-select none on fieldsets and array #708

kristofferj:
Fixes bug #708


kristofferj:
Yes. But if we enable user-select on e.x description, it will be selected when drag outside. 
Anyway, I don't believe that UI-text as in an app needs to be selectable for the end user.

bjoerge:
Hmm... That's a bummer. Being able to select text from a field description can be quite useful sometimes. Any other way we could solve this?

kristofferj:
Maybe there are some stuff you can do with the js-event, but afaik in css user-select is the only option

PR #741, opened 2018-04-18
[desk-tool] Add support for live edit

thomax:
- New flag for schema types: `liveEdit: true|false` (will of course evaluate to falsy if flag is missing)
- EditorWrapper call patch and commit correctly when liveEdit is true
- Don't render publish button when in liveEdit mode
- Don't render the "Has changes not yet published" pencil in the documents pane when in liveEdit mode
- Remove "Published about 17 hours ago" when in liveEdit mode
- Don't render Discard Changes and Unpublish choices from the dropdown when in liveEdit mode


thomax:
Thanks @bjoerge and @rexxars! Merging now.

PR #745, opened 2018-04-20
[example-studio] Add an example of how to rename a document field

bjoerge:


packages/example-studio/migrations/renameField.js
```diff
@@ -0,0 +1,43 @@
+import baseClient from 'part:@sanity/base/client'
```
rexxars: Perhaps start by explaining what the script does? What is the source and target fields?

packages/example-studio/migrations/renameField.js
```diff
@@ -0,0 +1,43 @@
...
+// Fetching documents that matches the precondition for the migration
+const fetchDocuments = () =>
+  client.fetch(`*[_type == 'author' && defined(fullname)] {_id, _rev, fullname}`)
+
+const buildPatches = docs =>
+  docs.map(doc => ({
+    id: doc._id,
+    patch: {
+      set: {name: doc.fullname},
+      unset: doc.name,
```
rexxars: Unset takes an array of fields to unset, and this seems to be using a property value instead of a field name as well

packages/example-studio/migrations/renameField.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Remove when we get support for sanity exec --with-user-token
+import ConfigStore from 'configstore'
+const authToken = new ConfigStore('sanity', {}, {globalConfigPath: true}).get('authToken')
+const client = baseClient.config({token: authToken})
+// ---
+
+// Fetching documents that matches the precondition for the migration
+const fetchDocuments = () =>
+  client.fetch(`*[_type == 'author' && defined(fullname)] {_id, _rev, fullname}`)
```
rexxars: Should probably make it explicit how many items we are fetching (at most). _Might_ also be worth mentioning that `defined()` will return `false` for an empty array, which might not be obvious.

packages/example-studio/migrations/renameField.js
```diff
@@ -0,0 +1,43 @@
...
+  )
+  return patches
+}
+
+const createTransaction = patches =>
+  patches.reduce((tx, patch) => tx.patch(patch.id, patch.patch), client.transaction())
+
+const commitTransaction = tx => tx.commit()
+
+fetchDocuments()
```
rexxars: Perhaps we could use async/await and a do/while loop to keep fetching until there are no results left?

packages/example-studio/migrations/renameField.js
bjoerge: Done 👇

PR #747, opened 2018-04-21
[form-builder] Fix bug causing form-builder to fail on patches w/ifRevisionID

bjoerge:
This fixes a bug causing an error when form builder code expected the recieved patch to include a path property.

Fixed it by filtering out the `ifRevisionID`-field from received patches and not treating `ifRevisionID` as a kind of patch.


bjoerge:
So this function converts from a gradient patch to a patch that can be passed to formbuilder inputs (that for some reason needs access to the patch that targeted it's value, and not the value itself). This always happens "after the fact", that is, after the patch already has been successfully applied. So the `ifRevisionID` doesn't really makes sense at this point.

rexxars:
I see, LGTM then!

PR #754, opened 2018-04-24
[core] Remove production server option from start command

rexxars:
We used to have an option for running the "production build" of a built Sanity studio. This is seldom if ever used, and is not documented anywhere. This PR removes the functionality since it has been out of order for a good number of versions.


bjoerge:
btw, do we still need `prodServer.js` in @sanity/server, or can it be removed too?

PR #756, opened 2018-04-24
[form-builder] Refactor DateTimeInput and add DateInput

bjoerge:
This is a refactor of the DateTimeInput, and also adds a default DateInput for `type: 'date'` (this stores dates on the `YYYY-MM-DD` format, ignoring the time part).

I've basically moved the current DateTimeInput to a `BaseDateTimeInput` and made it possible to disable the time part of the datetime string with the `onlyDate` prop.

The `DateTimeInput` and `DateInput` wraps the `BaseDateTimeInput`, and passes down the relevant props.

This also includes a few consistency fixes and code cleanup (naming, unused code, etc).


bjoerge:
> Would it be worthwhile to let the schema disable the datepicker widget? Use case: You dislike the datepicker and prefer a plain keyboard input.

Good point. That could probably be solved by replacing it with the default string input component, or?

thomax:
> Good point. That could probably be solved by replacing it with the default string input component, or?

Absolutely!

PR #760, opened 2018-04-26
[preview] Tolerate invalid values returned from preview config and provide helpful errors

bjoerge:
This will prevent the studio from crashing if preview config targets values that cannot be handled by the preview component (e.g. `{title: {foo: 'bar'}}`). It will show a warning in stead of the placeholder preview, and provide a detailed explanation in the developer console.

image

prepare-guard

### DIFF
--- a/packages/@sanity/preview/src/constants.js
+++ b/packages/@sanity/preview/src/constants.js
@@ -1,4 +1,4 @@
 export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
 export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
 
-export const INVALID_PREVIEW_CONFIG = {}
+export const INVALID_PREVIEW_CONFIG = Symbol('invalid preview config')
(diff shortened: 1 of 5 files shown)


packages/@sanity/preview/src/constants.js
```diff
@@ -1,2 +1,4 @@
 export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
 export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
+
+export const INVALID_PREVIEW_CONFIG = {}
```
rexxars: Should'nt this use `Symbol.for()` instead?

packages/@sanity/preview/src/prepareForPreview.js
```diff
@@ -1,74 +1,243 @@
...
+  select: {[string]: string},
+  prepare?: (SelectedValue, ViewOptions) => PreparedValue
+}
+
+type Type = {
+  name: string,
+  preview: PreviewConfig
+}
+
+type PrepareInvocationResult = {|
```
rexxars: Not familiar with this pipe syntax, what does it mean/do?

packages/@sanity/preview/src/constants.js
bjoerge: That's probably better, yeah, thanks for pointing out :)

packages/@sanity/preview/src/prepareForPreview.js
bjoerge: By default in Flow, object types are open ended, meaning you may keep additional props on it, the pipe is the syntax for "exact type", meaning you'll get a type error if the value contains additional properties.

PR #761, opened 2018-04-26
[client] Don't retry if SSE connection fails before open

bjoerge:
I'm not entirely sure this fix is correct, but it prevents retrying if the SSE connection fails before it's established. This prevents the client from retrying the eventsource connection if the server responds with 404. I wasn't able to find any other way to detect details around _why_ the SSE connection errored Don't seem like EventSource specifies any response status code (or headers) to be exposed.

Let me know if you know of better ways to solve this, @rexxars

### DIFF
--- a/packages/@sanity/client/src/data/listen.js
+++ b/packages/@sanity/client/src/data/listen.js
@@ -40,9 +40,11 @@ module.exports = function listen(query, params, opts = {}) {
     let es = getEventSource()
     let reconnectTimer
     let stopped = false
+    let didOpen = false
 
     function onError() {
-      if (stopped) {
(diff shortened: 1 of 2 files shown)


rexxars:
I've changed the server to respond with a channel error and a disconnect event instead of a 404 JSON response. This results in the following:

image

On a lower level, the `channelError` event triggers an error on the observable, while the `disconnect` event completes the observable and closes the SSE connection. I feel like this is a more correct approach for this error.

I think this PR might cause unwanted side effects, so I think we should revisit this if it becomes an issue again

bjoerge:
Awesome, that is a way way way better solution!

PR #762, opened 2018-04-27
[cli] Run yarn with --ignore-engines when resolving dependencies

bjoerge:
Currently `sanity init` fails on Node 10 due to a thirdparty dependency that has `node < 10` set in its engines field.

This fixes it by running yarn with the `--ignore-engines` flag the engines field when resolving dependencies in new projects.


rexxars:
Looks like this needs to be rebased, though?

bjoerge:
Yup, changed the base to master so we can release this in a patch bump immediately.

PR #764, opened 2018-04-27
[reflector-client] Reuse sanity client instead of using get-it directly

rexxars:


rexxars:
Going to take a risk and merge this since it's not actually in use anywhere at the moment.

PR #765, opened 2018-04-27
[presence-store] Fix path to entry file

rexxars:
Couple of code style fixes as well


rexxars:
Going to take a risk and merge this since it's not actually in use anywhere at the moment.

PR #766, opened 2018-04-29
[default-layout] Provide error boundary at tool level

rexxars:
This PR provides an error boundary at the tool level, meaning that if a crash happens during the render phase of a React component inside of a tool, the error is caught in default layout when it tries to render the tool, and instead of displaying a blank page, provides a more user-friendly error message.

I took a stab at some design work since this wasn't really part of any sprint, but rather the result of some personal frustration. It seems to be working on desktop and mobile devices.

When in production, it just displays an "oops, tool x crashed" with buttons to display error details and to retry mounting of the tool, whereas in development it will automatically display the error details to prevent unnecessary clicks.

screen shot 2018-04-29 at 22 23 20

screen shot 2018-04-29 at 22 23 38

screen shot 2018-04-29 at 22 39 57


packages/@sanity/default-layout/src/components/RenderTool.js
```diff
@@ -1,28 +1,101 @@
+import React, {Component} from 'react'
 import PropTypes from 'prop-types'
-import React from 'react'
 import tools from 'all:part:@sanity/base/tool'
+import Button from 'part:@sanity/components/buttons/default'
+import ErrorBorkImage from './ErrorBorkImage'
+import styles from './RenderTool.css'
+
+class RenderTool extends Component {
```
bjoerge: This is a matter of preference, but IMO, exporting immediately (`export default class RenderTool`) is preferable since the reader don't have to scroll all the way to the bottom of the file to figure out if there's any wrappers or HOCs involved.

packages/@sanity/default-layout/src/components/RenderTool.js
```diff
@@ -1,28 +1,101 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {error: null, showErrorDetails: __DEV__}
```
bjoerge: Unless we don't need anything from props to set initial state, I think it's better to set initial state using class field declarations, e.g. like we do here: https://github.com/sanity-io/sanity/blob/9fdd9ed6c2c318bbae8bbbf097e3530afebc2478/packages/%40sanity/default-layout/src/components/DefaultLayout.js#L40-L45

PR #768, opened 2018-04-30
[chore] Replace @sanity/observable with RxJS

bjoerge:
This is an initial stab at replacing `@sanity/observable` with rxjs@6, and not yet ready to merge.

A few things to consider:

### Bundle size
With RxJS 6 you only import the creation functions and operators needed, e.g `import {map, filter} from 'rxjs/operators`. With tree shaking this can be optimized, but there's currently no easy way to tree shake this AFAIK. Also, in RxJS6 there's only two publicly exposed export sites (`rxjs` and `rxjs/operators`), so any form of tree shaking involves remapping the individual imported functions to internal (private) RxJS modules). Unfortunately this will not work server side with CommonJS `require()`

A possible way to fix this is to keep `@sanity/observable` and maintain a list of mappings from the internal `rxjs` modules there, so that you import e.g. `import flatMap from '@sanity/observable/operators/flatMap` (optionally use something like [transform-named-imports](https://www.npmjs.com/package/babel-plugin-transform-named-imports) to rewrite named imports).

### Importing map, filter, reduce, etc. feels a bit tedious
Especially since these methods are already implemented on standard tracking observable implementations. OTOH, it probably reads better if operators are consistently applied using pipe, instead of a mix.
E.g. consistently using pipe:
```js
client.listen('*[_id == "somedoc"')
  .pipe(
    filter(event => event.type === 'mutation'),
    map(event => event.resultRev),
    scan((revisions, revision) => revisions.concat(revision), []),
    flatMap(fetchAdditionalRevisionMetadata)
  )
  .subscribe(revisions => {
    console.log('Revisions seen so far: ', revisions)
  })
```
vs a mix:
```js
client.listen('*[_id == "somedoc"')
  .filter(event => event.type === 'mutation')
  .map(event => event.resultRev)
  .pipe(
    scan((revisions, revision) => revisions.concat(revision), []),
    flatMap(fetchAdditionalRevisionMetadata)
  )
  .subscribe(revisions => {
    console.log('Revisions seen so far: ', revisions)
  })
```
IMO the first example reads better, and as an added bonus, it is easier to determine whether we're working with an observable or a plain array. If we want the basic operators, we will need to keep a SanityObservable that implement these methods, but I'm not sure it's worth the extra maintenance burden.


rexxars:
+1 for always using pipe

bjoerge:
I realized this accidentally doubled the bundled size of `@sanity/client`, so I've kept the `@sanity/observable/minimal` entry for now (and also proxied to internal rxjs `require()`-s which feels a bit hacky, but is the only option we have apart from proper tree shaking afaik).

I also added deprecation warnings that recommends switching to `pipe(...)` when any of the member operators on `SanityObservableMinimal` is used (removing them now would be a breaking change).

In other words: `client.observable.fetch(query).map(mapFn).filter(filterFn)` will still work, but using them will issue a deprecation warning, saying that `.pipe(map(mapFn))` is now preferred.

I think this is good to merge now, but we should probably keep it for a few days for testing in `next`, and releasing a canary for testing in Vega and other studios.

PR #773, opened 2018-05-03
[cli] Make upgrade message less confusing

kmelve:
The previous copy assumed contextual knowledge of the CLI, this text makes it a bit clearer.


bjoerge:
I took the liberty of rebasing this against current next and making the commit message align with current conventions.

Issue #716, opened 2018-04-05
Leave package manager choice up to the user


bjoerge:
Supporting two different package managers with their own quirks and subtle differences is time consuming and error prone. Personally, because I'm more used to typing `npm` over `yarn`, I regularily run into issues caused by mixed use. I've seen that other users of sanity run into these kinds of issues too.

I also got bitten by the subtle differences in how npm and yarn hoist packages yesterday when attempting to fix #709. The initial fix worked when using npm, but not yarn, and in order to verify that the fix worked I had to test with both `npm install` and `yarn install`. This took several hours to figure out.

npm has improved greatly since yarn was introduced, and overall I think the issues that yarn solved when is now mostly fixed by npm.

Also, due to the dominance npm has in the Node.js ecosystem, I don't think its feasible for us to stop supporting npm.

Another point is the number of technologies you need to know about when maintaining a Sanity Studio project. Today we assume you use `npm` for installing the cli initially, but then stick to using `yarn` later on (e.g. when adding new dependencies to a studio). This is also not very clearly communicated from our docs (but that's another issue I guess).

I know that we have built a major part of the cli on yarn, and I assume it will take a considerable effort to replace yarn with npm, but I suspect only supporting one package manager will pay off in the long run, both for us and our users.

rexxars:
I'm curious as to what you mean by _support_ in this case. Since we can't _block_ yarn from being used (unless we fail on the presence of a `yarn.lock` or similar), wouldn't the same issues occur to users anyway? Is saying "You need to use npm" good enough?

bjoerge:
I was about to clarify that, actually :) I'm not proposing explicitly _not_ supporting yarn, but rather to change our official stance to _we use npm internally, if you use yarn you are on your own_ rather than _we use yarn internally, if you use npm you are on your own_

Another option would of course be to support both officially, e.g. by making `sanity upgrade` conditionally use yarn or npm based on config (or precense of yarn.lock vs package-lock.json).

rexxars:
I personally feel we should stick with one of them, and I'm open to switching to npm since it is after all the default package manager. Actually making use of both internally seems like something that would lead to a much larger surface area to maintain.

rexxars:
As a side note; maybe we should warn on startup in the case where both a `yarn.lock` and a `package-lock.json` file is found in a studio?

bjoerge:
>Actually making use of both internally seems like something that would lead to a much larger surface area to maintain.

[...]

PR #774, opened 2018-05-03
[cli] Delete Sanity-modules that needs upgrading to work around yarn bug

rexxars:
Two fixes:

- When running `sanity upgrade`, it previously included `@sanity/cli` in the output, which was very misleading.
- When running `sanity upgrade`, we've encountered an inexplicable bug where yarn will upgrade the lockfile with the resolved tarball URL and version, but will somehow not extract and replace the installed module. With `dce1716`, we will _delete_ the modules that need to be upgraded before actually running `yarn` to work around this bug.

This once again puts some further pressure on prioritizing #716


bjoerge:
Should go into master and be released immediately, maybe?

PR #775, opened 2018-05-03
Responsive dialogs width

kristofferj:
- Dialog now has a min-width
- Array input sets its min-width of the content in dialog
- Composable CSS to use when components wants to make the dialog bigger
- Max-width on editor in desk-tool to prevent too wide inputs


kristofferj:
Rebased 👍

PR #776, opened 2018-05-03
[components] Make tags textfield display readOnly correct

kristofferj:


packages/@sanity/components/src/tags/styles/TextField.css
```diff
@@ -53,14 +59,14 @@
...
 
 .content {
-  composes: input from 'part:@sanity/components/textinputs/default-style';
+  composes: root from 'part:@sanity/base/theme/forms/text-input-style';
   display: block;
   overflow: hidden;
   padding-left: 0.3em;
 
-  @nest :focus-within & {
+  /* @nest :focus-within & {
```
bjoerge: Could you either delete this block or write a comment about why it needs to stay?

kristofferj: Oh. this should be there. thx!

PR #777, opened 2018-05-03
Styleselect and official react popper

kristofferj:
- Fixing style select (the one in block editor)
- Using official react-popper

### DIFF
--- a/packages/@sanity/components/src/selects/StyleSelect.js
+++ b/packages/@sanity/components/src/selects/StyleSelect.js
@@ -1,7 +1,6 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import styles from 'part:@sanity/components/selects/style-style'
-import {includes} from 'lodash'
 import FaAngleDown from 'part:@sanity/base/angle-down-icon'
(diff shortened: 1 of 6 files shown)


packages/@sanity/components/src/selects/StyleSelect.js
```diff
@@ -81,58 +85,72 @@ class StyleSelect extends React.Component {
...
-          </div>
+          </Target>
+          <Portal className={styles.portal}>
+            <Escapable onEscape={this.handleCloseList} />
+            {showList && (
+              <Popper placement="bottom-start">
+                <div ref={this.setPopperElement}>
+                  <List className={styles.list}>
+                    {items.map((item, index) => {
+                      const isSemiSelected = value && value.length > 1 && includes(value, item)
```
bjoerge: Could you use `value.includes(item)` here instead?

Edit: I see now that this was not added in this PR. Bonus point if you would like to change it to `value.includes(item)` still

kristofferj: yep

PR #778, opened 2018-05-03
[test-studio] Add custom reference input example

rexxars:
This PR adds an example of how to create a custom input for the reference type. Kind of simple and naive, but should explain the basics clear enough.


rexxars:
An obvious next task for this would be to listen for changes to the authors, remove deleted authors, add newly created authors etc, but I think we need some simpler abstractions for those things before I want to add that amount of complexity to this example.

PR #783, opened 2018-05-03
[form-builder] Fix tab/focus on confirm delete button

bjoerge:
Now that popovers are rendered synchronously, we can set focus on the _confirm delete button_ for array items. This also fixes an issue that made hitting the tab key on the _confirm delete button_ to move focus out of the document.


bjoerge:
Yep, don't like it either. Since this applies to all focusable elements inserted into a portal, it would probably be better to fix it in the portal component instead. What do you think?

PR #784, opened 2018-05-03
[schema] Add missing fields for reference and slug types

rexxars:
The reference and slug types were missing their fields. If the default input component was not in play, the form builder would not render any input/field. The validation infrastructure also traverses the schema definition, and missing fields will prevent it from functioning as intended.

An interesting thought would be whether or not we should add fields to the document type as well, for `_id`, `_type`, `_createdAt` etc. They could be marked as hidden and read only, but it would make schema traversal more complete/correct. Thoughts?

### DIFF
--- a/packages/@sanity/schema/src/legacy/types/reference.js
+++ b/packages/@sanity/schema/src/legacy/types/reference.js
@@ -1,6 +1,20 @@
-import {pick} from 'lodash'
 import arrify from 'arrify'
+import {pick} from 'lodash'
 import {lazyGetter} from './utils'
+
+export const REF_FIELD = {
+  name: '_ref',
+  title: 'Referenced document ID',
+  type: 'string'
+}
+
+export const WEAK_FIELD = {
+  name: '_weak',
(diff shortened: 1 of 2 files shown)


bjoerge:
> An interesting thought would be whether or not we should add fields to the document type as well, for _id, _type, _createdAt etc. They could be marked as hidden and read only, but it would make schema traversal more complete/correct.

This makes sense to me and I can't see any reason to not add them.

rexxars:
What is perhaps a bit confusing is that you can use a document as an object (embedded somewhere) in which case it does not receive these fields. But maybe that doesn't matter given the fields are not marked as required or shown in any way.

bjoerge:
Yeah, that's a very good point. It doesn't make much sense to have `_id`, `_rev`, etc. defined on the type when a document is used as an embedded object. Even though they are not required, they should never be present on embedded objects, so it doesn't really makes sense to have them defined as fields (even if hidden and readOnly). Maybe not the biggest deal, though.

PR #786, opened 2018-05-04
Fix array grid view

kristofferj:
- adding progress to grid view (thumbnails)
- Cleaning up circle preview
- Styling fallback icon

Think it is safe include `options: {layout: 'grid'}` in the documentation now.

### DIFF
--- a/packages/@sanity/components/src/previews/MediaPreview.js
+++ b/packages/@sanity/components/src/previews/MediaPreview.js
@@ -1,7 +1,10 @@
+/* eslint-disable complexity */
 import PropTypes from 'prop-types'
 import React from 'react'
 import styles from 'part:@sanity/components/previews/media-style'
+import ProgressCircle from 'part:@sanity/components/progress/circle'
 import SvgPlaceholder from './common/SvgPlaceholder'
+
 const fieldProp = PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.func])
 
 export default class MediaPreview extends React.PureComponent {
@@ -10,6 +13,7 @@ export default class MediaPreview extends React.PureComponent {
     subtitle: fieldProp,
     description: fieldProp,
     media: fieldProp,
+    progress: PropTypes.number,
     mediaDimensions: PropTypes.shape({
       width: PropTypes.number,
       height: PropTypes.number,
@@ -25,6 +29,7 @@ export default class MediaPreview extends React.PureComponent {
     subtitle: undefined,
     description: undefined,
     media: undefined,
+    progress: undefined,
     isPlaceholder: false,
     children: undefined,
     mediaDimensions: {width: 160, height: 160, aspect: 1, fit: 'crop'}
@@ -38,7 +43,8 @@ export default class MediaPreview extends React.PureComponent {
       media,
       mediaDimensions,
       children,
-      isPlaceholder
+      isPlaceholder,
+      progress
     } = this.props
 
     if (isPlaceholder) {
@@ -48,9 +54,6 @@ export default class MediaPreview extends React.PureComponent {
             className={styles.padder}
             style={{paddingTop: `${100 / mediaDimensions.aspect || 100}%`}}
           />
-          <div className={styles.mediaContainer}>
-            <SvgPlaceholder styles={styles} />
-          </div>
         </div>
       )
     }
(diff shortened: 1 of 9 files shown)


kristofferj:
Node 10 check does not complete the CI check. It says it's a path in `ArrayInput.css`.
Can't figure out if it is something related to this branch.

kristofferj:
Seems that all CI checks have passed now.

packages/@sanity/components/src/previews/MediaPreview.js
```diff
@@ -25,6 +29,7 @@ export default class MediaPreview extends React.PureComponent {
     subtitle: undefined,
     description: undefined,
     media: undefined,
+    progress: undefined,
```
bjoerge: No need to define a default prop as `undefined` (as that is already the default)

packages/@sanity/components/src/previews/MediaPreview.js
```diff
@@ -1,7 +1,10 @@
+/* eslint-disable complexity */
 import PropTypes from 'prop-types'
 import React from 'react'
 import styles from 'part:@sanity/components/previews/media-style'
+import ProgressCircle from 'part:@sanity/components/progress/circle'
 import SvgPlaceholder from './common/SvgPlaceholder'
```
bjoerge: `SvgPlaceholder` is no longer in use in this file, so this import should be removed.

packages/@sanity/components/src/progress/ProgressCircle.js
```diff
@@ -6,86 +6,77 @@ const radiusFactor = 1.3
...
+    showPercent: PropTypes.bool
   }
 
   static defaultProps = {
-    radius: 50,
-    percentage: 0,
-    strokeWidth: 10
+    percent: 0,
+    style: {},
+    text: undefined,
```
bjoerge: `text` should be removed from defaultProps here as it is redundant.

packages/@sanity/preview/src/components/SanityDefaultPreview.js
```diff
@@ -75,7 +75,7 @@ export default class SanityDefaultPreview extends React.PureComponent {
   renderIcon = options => {
     const {type} = this.props
     const Icon = type.icon
-    return Icon && <Icon />
+    return Icon && <Icon className="sanity-studio__preview-fallback-icon" />
```
bjoerge: Why is this a hard coded string instead of an imported css modules class?

packages/@sanity/components/src/previews/MediaPreview.js
kristofferj: Could we remove `react/require-default-props` in the linter then?

packages/@sanity/components/src/previews/MediaPreview.js
kristofferj: 👍

packages/@sanity/preview/src/components/SanityDefaultPreview.js
kristofferj: Mainly because its need to be referred in the preview styling. Making it a css module would generate unreadable code and nested composed parts. 

This is easier to read
```
/* Mainly for styling fallback icons */
  @nest & :global(.sanity-studio__preview-fallback-icon) {
    width: 2em;
    height: 2em;
    opacity: 0.5;
  }
```

packages/@sanity/components/src/progress/ProgressCircle.js
kristofferj: The linter says I should have it. Think it is the samme issue as above.

packages/@sanity/preview/src/components/SanityDefaultPreview.js
bjoerge: I see 👍

packages/@sanity/components/src/progress/ProgressCircle.js
bjoerge: Yeah, I think we should switch that rule off entirely.

bjoerge: On second thought, I think the issue is rather that the default type is `undefined` and not an empty string here.

Specifying defaults for non-required props is a good thing, but they should always be of the same type as the prop, so when the prop is an array the default should be `[]`, if string, `''`, etc. This way you can blindly call properties and methods defined on the prop value (e.g. `text.substring(0, 10)` without having to first do a undefined check.

kristofferj: Ah. That makes sense

PR #793, opened 2018-05-07
[default-login] Part-ify rendering of login providers

thomax:
Put the rendering of login providers in a overrideable component.


thomax:
Not sure if https://github.com/sanity-io/sanity/blob/customize-login-wrapper/packages/@sanity/default-login/sanity.json should also contain a `{name: 'part:@sanity/base/login-dialog-content', description: '...'}` for `part:@sanity/base/login-dialog-content`?

thomax:
> I think it would be good form to add a defined part with description

I agree. Fixed.

PR #796, opened 2018-05-07
Upgrade storybook

kristofferj:
Don't know if I did the middleware thing correctly. It works, but it generates a warning:
```Storybook: (node:8943) DeprecationWarning: importing default webpack config generator from '@storybook/react/dist/server/config/defaults/webpack.config.js' is deprecated. Use third argument of your exported function instead. See https://storybook.js.org/configurations/custom-webpack-config/#full-control-mode--default```

### DIFF
--- a/packages/@sanity/components/src/previews/styles/DefaultPreview.css
+++ b/packages/@sanity/components/src/previews/styles/DefaultPreview.css
@@ -62,10 +62,10 @@
   margin: 0;
   padding: 0;
   padding-bottom: 0.25em;
-  height: 1em;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by rexxars** [comment]
Getting this error when running, so I think this needs some 🔎 
```
Accessing nonexistent addons channel, see https://storybook.js.org/basics/faq/#why-is-there-no-addons-channel
```
Mayhaps a conflict in module versions.

kristofferj:
This worked when i tested. Anyway, seems that it is the knobs plugins and sanity plugin that creates this error.

PR #815, opened 2018-05-16
DialogContent component for sizing dialogs

kristofferj:
- DateTime input dialog as now working
- DialogContent component is now a part and can be used to size content
- minor cleanup

### DIFF
--- a/packages/@sanity/components/src/dialogs/styles/PopOver.css
+++ b/packages/@sanity/components/src/dialogs/styles/PopOver.css
@@ -43,7 +43,7 @@
   z-index: var(--zindex-portal);
   position: relative;
 
-  @media (--max-screen-medium) {
+  /* @media (--max-screen-medium) {
     position: fixed;
     box-sizing: border-box;
     top: 0;
@@ -52,7 +52,7 @@
     height: 100vh;
     width: 100vw;
     overflow: auto;
-  }
+  } */
 }
 
 .arrow {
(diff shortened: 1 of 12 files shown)


packages/@sanity/components/src/dialogs/styles/PopOver.css
```diff
@@ -234,11 +230,11 @@
   font-weight: 400;
   border-radius: var(--border-radius-base) var(--border-radius-base) 0 0;
 
-  @media (--max-screen-medium) {
+  /* @media (--max-screen-medium) {
```
bjoerge: Should these be commented out?

packages/@sanity/form-builder/src/inputs/ArrayInput/ItemValue.js
```diff
@@ -210,6 +211,8 @@ export default class RenderItemValue extends React.Component<Props> {
       )
     }
 
+    const dialogSize = 'medium'
```
bjoerge: This is supposed to be hard coded?

packages/@sanity/components/src/dialogs/styles/PopOver.css
kristofferj: No. Pushing a fix now.

packages/@sanity/form-builder/src/inputs/ArrayInput/ItemValue.js
kristofferj: Yes. Until we have a resolver for size. This sets the size for both popover and defaultdialog

kristofferj:
I will test this a little bit before i merge it.

kristofferj:
Rebased against next

Issue #817, opened 2018-05-17
allow basePath to be overridden in environment variable or cli option


dan-dr:
Related: https://github.com/sanity-io/sanity/pull/720

Hey guys. The basePath feature is excellent, but even since it's added , I found that I need to have a different basePath for different environments.
My basePath in the staging env (and `sanity deploy`) is '/' but in my production it's '/studio'.

I was thinking it should be able to change it during build time, which means either adding it as an option to the `sanity build` command or maybe have the cli use an environment variable like STUDIO_BASE_PATH

I've already dabbled a bit with the code, and an environment variable will be the cleanest solution I think. What do you think?

rexxars:
Forgot to reply to this, but I did follow up and implement this in #819 - the change is part of the newly released 0.132.0

PR #819, opened 2018-05-17
[core] Add ability to override project basePath option with an environment variable

dan-dr:
Solves #817 

Better than specifying it in the cli as I did in #818 , as basePath is used in files that aren't referring to cli flags (throw that PR away)


rexxars:
Thanks for the PR!
There were a couple of issues with it, mainly that the default layout router doesn't have access to process.env since it's running in the browser. I took the liberty of merging the base path into the sanity config if available so that it will be part of the bundled javascript blob.

There was also an issue where it would always fall back to the default (`/static`) if no base path was configured in `sanity.json`.

I'm requesting a review from some other team members to see if they agree that an environment variable is an acceptable way to go.

PR #826, opened 2018-05-22
[form-builder] Use FocusManager with inline items

bjoerge:
This fixes a bug causing editing of inline objects to crash


skogsmaskin:
LGTM! 👍

PR #827, opened 2018-05-23
Toggle sort direction on desk-tool documents-pane

kristofferj:
Toggling between asc/desc is not possible, since the sort order are an array and can have multiple ordering. Just picking the first in array is not a good solution. I tried to solve this by setting a toggle on the ordering. This also solves other scenarios that could appear, like if it is needed to explain the scenario when toggling. By also providing a title (fallback to asc/desc) it is possible to change the title when it toggles (not good UI, but possible)

### Example
```
title: 'Newest',
  name: 'updatedAt',
  by: [{field: '_updatedAt', direction: 'desc'}],
  toggle: {
    title: 'Oldest',
    by: [{field: '_updatedAt', direction: 'asc'}]
  }
```
### Icons
If there is a toggle option, the icon will show an arrow up or down when it is toggled. If there is no toggle option the arrow goes both ways. It is also possible to provide your own icon to the ordering if that is needed.

If we go for this solution, we need to include this in the documentation.

### DIFF
--- a/packages/@sanity/base/sanity.json
+++ b/packages/@sanity/base/sanity.json
@@ -647,6 +647,14 @@
       "path": "components/icons/Sort.js"
     },
     {
+      "implements": "part:@sanity/base/sort-asc-icon",
+      "path": "components/icons/SortAsc.js"
+    },
+    {
+      "implements": "part:@sanity/base/sort-desc-icon",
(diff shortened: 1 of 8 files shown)


kristofferj:
The infrastrucure this is based on is changed with "structure". Think it is best to do this from skratch based on the next branch with structure.

PR #847, opened 2018-05-30
[components] Using stacked and captureclickoutside

kristofferj:
instead of enhanceWithClickOutside

This fixes a bug where styleselect in an dialog closes when a style is selected.

This still have an issue where the dialog seems to re-render when onChange is called.
Maybe @skogsmaskin have a solution to this?


skogsmaskin:
Fixed the re-render issue (removed refresh CSS hack, as it is now fixed in Webkit)

PR #848, opened 2018-05-31
[core] Pass non-exec flags/args onto target script

rexxars:
Allows using `sanity exec --with-user-token somefile.js arbitrary args --here`


rexxars:
Good call, I'll update to use subargs instead.

PR #865, opened 2018-06-11
[components] Test button focus

kristofferj:
This is a test to solve the problem in #834
Seems to work fine in safari and chrome.


kristofferj:
This is cherry-picked into the delightful branch

PR #866, opened 2018-06-12
add missing asset type check

kmelve:
Not sure if I hit the mark on this. I thought about validating the import strings in the import script but was not sure where to write that in. The goal is to relay a more helpful error message if the user has forgotten to specify the type-prefix.

### DIFF
--- a/packages/@sanity/client/src/validators.js
+++ b/packages/@sanity/client/src/validators.js
@@ -16,8 +16,17 @@ exports.projectId = id => {
 }
 
 exports.validateAssetType = type => {
+  const VALID_ASSET_TYPES_LIST = VALID_ASSET_TYPES.join(', ')
+  if (!type) {
+    const errorMessage = `
+      Asset type is not specified.
+      Have you remebered prefixing your import string with <type>@?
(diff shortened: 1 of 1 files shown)


rexxars:
Don't think this is the right place to pick this up - the client can be used outside of import context, and the error message doesn't make sense if `assets.upload()` is called with an invalid type

kmelve:
Yes, that's what I suspected. Thought about putting a check here, would that make more sense? https://github.com/sanity-io/sanity/blob/720c909f4795823a1f2e0507140891eb3d31e0e2/packages/%40sanity/import/src/assetRefs.js#L36

rexxars:
Seems like a good choice - you have access to both the document with a possible ID and the path to the asset at that point.

kmelve:
Moved it to the import package. Still not sure if I found the right entry point though.

rexxars:
Could you rebase and squash the changes? I _think_ this looks good, but easier to see if rebased.

rexxars:
Also; please prefix the commit message with `[import]`

kmelve:
Cool! By the way, we should probably lift the hardcoded `image` and `file` strings out to the SUPPORTED_TYPES variable?

PR #874, opened 2018-06-17
[cli] Add option for initiating sanity with a clean template

kmelve:
Yep. Here it is. Since I'm keeping forgetting to set the clean template flag, and since we want to communicate that Sanity can be used for anything.

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.js
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.js
@@ -511,4 +511,3 @@ async function promptForAclMode(prompt, output) {
 
   return mode
 }
- 32e8383cdea0e851b1376c627075a3bc6155e0ea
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.js
```diff
@@ -507,3 +511,4 @@ async function promptForAclMode(prompt, output) {
 
   return mode
 }
+ 32e8383cdea0e851b1376c627075a3bc6155e0ea
```
rexxars: What's this?

rexxars:
@evenwestvang made me remove this last time it was present, so would like him to chime in.

evenwestvang:
Now that we have 3 other options and the number of 0 document installs is down I’m all :thumbsup: about this.

PR #875, opened 2018-06-17
[core] More helpful CLI messages

kmelve:
Suggestion for some alternative wording at startup. We want to be clearer about what you're running locally, i.e. the Content Studio. Ideally, we could add wording about `the studio being connected to the project ${projectId} and the ${dataset} dataset through the Sanity API`, but I was not sure how to get to the configuration in startAction.js

### DIFF
--- a/packages/@sanity/core/src/actions/start/startAction.js
+++ b/packages/@sanity/core/src/actions/start/startAction.js
@@ -65,7 +65,7 @@ export default async (args, context) => {
 
     if (!hasErrors && !hasWarnings) {
       output.print(
-        chalk.green(`Sanity Content Studio compiled successfully! Open it locally at http://${httpHost}:${httpPort}`) // eslint-disable-line max-len
+        chalk.green(`Content Studio successfully compiled! Go to http://${httpHost}:${httpPort}`) // eslint-disable-line max-len
       )
       return
     }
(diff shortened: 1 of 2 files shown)


packages/@sanity/core/src/actions/start/startAction.js
```diff
@@ -65,7 +65,8 @@ export default async (args, context) => {
 
     if (!hasErrors && !hasWarnings) {
       output.print(
-        chalk.green(`Compiled successfully! Server listening on http://${httpHost}:${httpPort}`)
+        chalk.green(`Sanity Content Studio compiled successfully!
+        Open it locally at http://${httpHost}:${httpPort}`)
```
rexxars: Is the line wrapping/indentation here intentional?

kmelve: The wrapping is (the linter complained), but I guess the indentation’s not.

rexxars: RIght, cause since this is targetted at a terminal, whitespace is actually respected. Probably better to split this up in a different way, or disable the max-length rule for the lines in question.

rexxars:
Technically OK - I'll let @evenwestvang decide on the wording

PR #880, opened 2018-06-18
[form-builder] Allow viewing block editor nodes when input is read only

bjoerge:


packages/@sanity/form-builder/src/inputs/BlockEditor/Editor.js
```diff
@@ -150,16 +150,18 @@ export default class Editor extends React.Component<Props> {
...
   scrollIntoView(change, node) {
-    const {onChange} = this.props
-    change.collapseToEndOf(node)
+    const {onChange, readOnly} = this.props
+    if (!readOnly) {
+      change.collapseToEndOf(node)
+    }
     const element = findDOMNode(node)
     element.scrollIntoView({behavior: 'instant', block: 'center', inline: 'nearest'})
     onChange(change)
```
skogsmaskin: I think this should not be called when readOnly. Though there probably aren't any changes, better be on the safe side.

PR #902, opened 2018-07-03
[base] Add browser support + javascript check

rexxars:
This PR does three things:

1. Fix an eyesore typo in the app loading screen class names (`contetStudio` => `contentStudio`)
2. Adds a noscript "javascript disabled, please enable"-dialog
3. Adds a "your browser is not supported, please upgrade"-dialog

The two dialogs mentioned here take the simple and easy "Login successful"-message we use in the CLI tool as a template, just a vertically centered dialog like this:

image

The check to see if the browser is supported is slightly hackish - it checks for the presence of  `window.fetch`, which isn't really something we actually use, but it was first introduced in Edge 14, which is the lowest version currently supported by the studio. All other browsers support this feature, so I figured it was the easiest way to achieving a basic check.


evenwestvang:
LGTM

PR #934, opened 2018-08-29
[structure] assign document list item titles

rdunk:
Not sure if this is intentional behaviour, but it seems that the DocumentListItemBuilder title method isn't actually assigning the passed string, so using `StructureBuilder.documentListItem().title('Some Title')...` is not currently possible.


rexxars:
Did you actually run into a case where you were using `documentListItem()`? I'm wondering if it actually makes any sense to expose it, as document lists should render items based on the filter instead of explicitly taking list items...

rexxars:
Thanks for looking into this. We've just merged the structure feature into `next`, meaning it's scheduled for release real soon!

PR #937, opened 2018-08-31
Improve cli onboarding

kmelve:
Based on user tests, here's some changes to make things clearer in the on-boarding context. Feedback appreciated!

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.js
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.js
@@ -31,9 +31,9 @@ export default async function initSanity(args, context) {
   const unattended = cliFlags.y || cliFlags.yes
   const print = unattended ? noop : output.print
 
-  print(`This utility walks you through creating a Sanity project.
-    It installs a local instance of the Sanity studio,
-    which you can configure connects to the real-time hosted API.`)
+  print('This utility walks you through creating a Sanity project.')
+  print('It installs a configurable local instance of the Sanity studio,')
+  print('which connects to your real-time hosted API.')
   print('Press ctrl + C at any time to quit.\n')
 
   // If the user isn't already authenticated, make it so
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.js
```diff
@@ -31,8 +31,10 @@ export default async function initSanity(args, context) {
   const unattended = cliFlags.y || cliFlags.yes
   const print = unattended ? noop : output.print
 
-  print('This utility walks you through creating a Sanity installation.')
-  print('Press ^C at any time to quit.\n')
+  print(`This utility walks you through creating a Sanity project.
+    It installs a local instance of the Sanity studio,
+    which you can configure connects to the real-time hosted API.`)
```
rdunk: > which you can configure connects to

Missing a word here? Could also be:

> It installs a local instance of the Sanity studio, which you can configure to connect to the real-time hosted API.

or

> It installs a configurable local instance of the Sanity studio, which connects to the real-time hosted API.

packages/@sanity/cli/src/actions/init-project/initProject.js
```diff
@@ -31,8 +31,10 @@ export default async function initSanity(args, context) {
   const unattended = cliFlags.y || cliFlags.yes
   const print = unattended ? noop : output.print
 
-  print('This utility walks you through creating a Sanity installation.')
-  print('Press ^C at any time to quit.\n')
+  print(`This utility walks you through creating a Sanity project.
```
rexxars: If you actually want these to be on separate lines, it's better to split this up into separate print statements, to maintain proper indentation when printed.

packages/@sanity/cli/src/actions/init-project/initProject.js
kmelve: Thanks for the input @rdunk! Fixed.

PR #938, opened 2018-08-31
[components] Using poppable and set correct modifiers

kristofferj:


packages/@sanity/components/src/selects/StyleSelect.js
```diff
@@ -80,85 +99,67 @@ class StyleSelect extends React.PureComponent {
...
-              <Stacked>
-                {isActive => {
-                  return (
-                    <div className={styles.portal}>
-                      <Popper placement="bottom-start">
-                        <Escapable onEscape={isActive ? this.handleCloseList : undefined} />
-                        <CaptureOutsideClicks
-                          onClickOutside={isActive ? this.handleCloseList : undefined}
+            <div className={styles.popper}>
+              <CaptureOutsideClicks onClickOutside={showList ? this.handleCloseList : undefined}>
```
bjoerge: Why is this needed when Poppable has `onClickOutside`?

PR #946, opened 2018-09-04
[chore] Upgrade to babel v7

bjoerge:
This upgrades to babel v7, which was finally released last week 🎉

I've tested pretty extensively, and everything seems to work fine (I also verified that init with all templates results in runnable studios).

I've also removed the flow and react presets from the root level babel config (keeping only the presets/plugins we use across all repos), and explicitly added react/flow presets where we use them.


packages/@sanity/server/src/configs/webpack.config.js
```diff
@@ -87,11 +87,10 @@ export default (config = {}) => {
           use: {
             loader: resolve('babel-loader'),
             options: babelConfig || {
-              presets: [resolve('babel-preset-react'), resolve('babel-preset-es2015')],
+              presets: [resolve('@babel/preset-react'), resolve('@babel/preset-env')],
```
rexxars: Where are the env targets for the studio specified?

bjoerge: They are not specified, which makes it behave like preset-es2015 like before. It may be a good idea to specify, tough. What should it be?

rexxars: What we're using generally is:
```
{
  "node": "6",
  "chrome": "59",
  "safari": "10",
  "firefox": "56",
  "edge": "14"
}
```

bjoerge:
Added the babel env configuration (with targets) to the @sanity/server package and included it from the root babel.config.js.

Also removed the `proposal-object-rest-spread` since it is officially approved and included in `@babel/preset-env` now.

SimenB:
There isn't any timeline on this by any chance? 🙏 We use sanity in a monorepo, and due to hoisting and other fun stuff (peer dependencies, resolving stuff from user code instead of through `require` etc), it's hard to make it work correctly when migrating to babel 7 in our own stuff

bjoerge:
Hi @SimenB, we just released [v0.133.0](https://github.com/sanity-io/sanity/releases/tag/v0.133.0) which includes babel 7 support. Please let us know if you run into any issues when upgrading 🙏

PR #955, opened 2018-09-07
[core] Honor --dataset option on get/query/create/delete CLI commands

thomax:


bjoerge:
Looks like it could need a rebase against current `next`

rexxars:
Perhaps also squash the three relevant commits?

thomax:
@rexxars and @bjoerge Sound suggestions!

rexxars:
I see that most of these commands are missing a help text. If you have time, it would be great if you could add that as well. The create command has an example, and should have the `--dataset` flag added.

rexxars:
The dataset option is still missing in most commands help text

PR #960, opened 2018-09-12
[form-builder] Make sure we pass down onFocus to wrapped input components

bjoerge:
I noticed that #942 accidentally broke focus handling in certain cases by not passing `onFocus` to the actual input component.

It's not super critical from what I can tell, so lets merge it into next for now (and instead cherry-pick it  into a hotfix if it turns out otherwise).


bjoerge:
Can't be 100% certain, but I know for sure that `onFocus` was missing :)

PR #986, opened 2018-09-28
[chore] Fix test setup across all packages

bjoerge:
This provides a few tweaks to our babel setup so that it's possible to run tests from within package directories, using `npm run test-all` or with `npx lerna run test --scope=@sanity/<some-package>`

Also did some cleanup, deps normalization and fixed a few broken tests along the way, and noticed a few packages that already had test failures:

- @sanity/block-tools
- @sanity/desk-tool
- @sanity/document-window
- @sanity/import
- @sanity/plugin-loader
- @sanity/preview
- @sanity/resolver

I've disabled the tests for these packages in a [separate commit](dd7b3bb) that I'll keep when merging. Also made travis run all tests for us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ before_script:
   - travis_wait npm run bootstrap
   - npm run build
 script:
+  # Run tests in all packages
+  - npm run test-all
+
   # Allow running the Sanity CLI tool without specifying absolute path
   - export PATH="${TRAVIS_BUILD_DIR}/packages/@sanity/cli/bin:$PATH"
 


babel.config.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable import/no-commonjs */
 module.exports = {
-  presets: [['@babel/env', require('./packages/@sanity/server/src/configs/babel-env-config')]],
-  plugins: ['lodash', '@babel/plugin-proposal-class-properties'],
+  ...require('./.babelrc'),
```
rexxars: This needs to use `Object.assign` to be compatible with node 6

babel.config.js
```diff
@@ -1,6 +1,5 @@
...
   babelrcRoots: ['.', ...require('./lerna.json').packages]
```
rexxars: ... and this should probably be using `.concat()`

babel.config.js
bjoerge: Ah, good point, thanks!

PR #999, opened 2018-10-09
[import] Add `--allow-failing-assets` flag

### LINKED ISSUE #54: [cli] Dataset import should not fail on failed asset download
Say your image assets are sourced from Amazon S3. Currently, if the importer gets a failed download (say 403) the whole import will die immediately. This might be the desired default (not sure about that), but we should have a "keep going on missing asset"-flag which just saves the current document without the desired asset. Finally, when the import is done, produce a report which lists:

- document(s) which encountered a problem
- which asset URLs which failed
- response status code (403 or whatever)

Same behaviour should occur if assets are local files.

Maybe name the flag `keep-going-on-missing-asset`,  `dont-stop-on-missing-asset` or `tolerate-missing-assets`?

rexxars:
Fixes #54

With this flag enabled, we'll just continue as if the asset wasn't there, then print a list of failing URLs at the end.

In the future, we'll probably want to write the generated warnings to a file and add a flag for retrying the failed operations.

### DIFF
--- a/packages/@sanity/import/src/assetRefs.js
+++ b/packages/@sanity/import/src/assetRefs.js
@@ -9,7 +9,16 @@ const assetMatcher = /^(file|image)@([a-z]+:\/\/.*)/
 // Note: mutates in-place
 function unsetAssetRefs(doc) {
   findAssetRefs(doc).forEach(path => {
-    unset(doc, path)
+    const parentPath = path.slice(0, -1)
+    const parent = get(doc, parentPath)
+
+    // If the only key in the object is `_sanityAsset`, unset the whole thing,
+    // as we will be using a `setIfMissing({[path]: {}})` patch to enforce it.
+    // Prevents empty objects from appearing while import is running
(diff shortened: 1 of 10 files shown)


packages/@sanity/import/src/assetRefs.js
```diff
@@ -9,7 +9,16 @@ const assetMatcher = /^(file|image)@([a-z]+:\/\/.*)/
 // Note: mutates in-place
 function unsetAssetRefs(doc) {
   findAssetRefs(doc).forEach(path => {
-    unset(doc, path)
+    const parentPath = path.slice(0, -1)
+    const parent = get(doc, parentPath)
+
+    // If the only key in the object is `_sanityAsset`, unset the whole thing,
+    // as we will be using a `setIfMissing({[path]: {}})` patch to enforce it.
+    // Prevents empty objects from appearing while import is running
```
thomax: Nice

PR #1013, opened 2018-10-12
[structure] Allow disabling icons for lists and list items

rexxars:
This PR enables the user to disable the icons for a given list/document list pane, as well as for individual list items.

To disable icons for a pane:
```js
S.list()
  .title('Random stuff')
  .showIcons(false)
  .items([...])
```

To disable icon for a list item:
```js
S.listItem()
  .title('Site config')
  .icon(false)
```

We also do some best-effort disabling of icons by default in the following cases:
- If the default structure is used and no content types have a defined icon
- If the schema type for a document list pane does not define a preview `prepare` function or a selection containing `media`

Other notes: Fixed structure tests and an object not being cloned in editor spec, possibly leading to an unexpected mutation


thomax:
Structure is the new Schema 👍

thomax:
If the the `list` function for toggling icons is named `showIcons`, maybe the equivalent function on `listItem` should have a (more) similar name?

Instead of `S.listItem().icons(Boolean)`, make it `S.listItem().showIcon(Boolean)`?

rexxars:
Thought about it, but listItem already has an icon setter, so I figured it would be confusing being able to both set an icon *and* set whether or not it should be displayed.

Let me know what you think.

thomax:
Hmm, I see. To my mind, predictability is worth a lot, but I agree this is a tricky one. Functions with overlapping areas of responsibility increase code complexity.

From the users perspective, we're looking at either:

```js
S.list()
  .showIcons(true)
  .items([...])

S.listItem()
  .icon(today() === 'banana day' ? 'banana.png' : false)
```
or:

```js
S.list()
  .showIcons(true)
  .items([...])

S.listItem()
  .icon('banana.png')
  .showIcon(today() === 'banana day')
```

Based on API predictability, I _think_ I'd prefer the latter alternative, but I'm not sure I'm sure anymore :)

Maybe make `listItem.showIcon()` an alias for `listItem.icon(true|false)`?

thomax:
Come to think of it, separate functions for setting and toggling makes it easier to infer user/consumer intent. Such that `listItem.showIcon('false')` could be safely assumed to mean "no icon please" whereas `listItem.icon('false')` would force us to look for the image named "false".

rexxars:
OK, I've restructured this now based on your feedback:

To disable icons for a pane:
```js
S.list()
  .title('Random stuff')
  .showIcons(false)
  .items([...])
```

To disable icon for a list item:
```js
S.listItem()
  .title('Site config')
  .showIcon(false)
```

The item setting always has priority over the list setting - a list with `showIcons: false` can still hold an item with a specific setting of `showIcon: true`. The default is undefined, which means "do whatever the list thinks is best".

PR #1017, opened 2018-10-15
Updating .NET client - changing URL and author

huayra:


rexxars:
I would think we could link to both clients and let the user decide which one they want to use?

huayra:
Better now?

PR #1025, opened 2018-10-18
[client] Warn if client is configured with a token in the browser

bjoerge:
This issues a console warning if you configure the sanity client to use a token in the browser. The warning will only be visible on local hosts.

Changed the order of appearances of warnings so that you get them in the following order (and only one at a time):

1) Token in browser (fixing this may invalidate the warning below)
2) Token together with `useCdn: true` (fixing this may invalidate the warning below)
3) Missing `useCdn`

### DIFF
--- a/packages/@sanity/client/src/config.js
+++ b/packages/@sanity/client/src/config.js
@@ -1,6 +1,7 @@
 const generateHelpUrl = require('@sanity/generate-help-url')
 const assign = require('object-assign')
 const validate = require('./validators')
+const once = require('./util/once')
 
 const defaultCdnHost = 'apicdn.sanity.io'
 const defaultConfig = {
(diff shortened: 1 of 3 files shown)


packages/@sanity/client/src/config.js
```diff
@@ -10,29 +11,31 @@ const defaultConfig = {
   isPromiseAPI: true
 }
 
-const cdnWarning = [
+const LOCALHOSTS = ['localhost', '127.0.0.1', '0.0.0.0']
+const isLocal = host => LOCALHOSTS.indexOf(host) !== -1
+
+// eslint-disable-next-line no-console
+const printWarning = message => once(() => console.warn(message.join(' ')))
```
rexxars: Shouldn't this be `getWarningPrinter` or something? The `printWarning` calls below confused me quite a bit since there were no conditional around them

packages/@sanity/client/warnings.js
```diff
@@ -0,0 +1,7 @@
+const SanityClient = require('./src/sanityClient')
+
+global.window = {
+  location: {hostname: 'prod.com'}
+}
+// should warn
+new SanityClient({projectId: 'skajfie', useCdn: false, token: 'foo'})
```
rexxars: Is this file here intentionally, or left over from testing?

PR #1027, opened 2018-10-19
[vision] Add support for string and non-object outputs

kmelve:


packages/@sanity/vision/src/components/JsonDump.js
```diff
@@ -3,12 +3,22 @@ import React, {Fragment} from 'react'
 import PropTypes from 'prop-types'
 import ReactJson from 'react-json-view'
 
-class JsonBlock extends React.Component {
+function typeCheck(data) {
```
bjoerge: For the sake of clarity maybe rename this function to `isJSONValue` (or something), and invert the check to be instead `return data !== null && typeof data === 'object'`?

If you think that's a good idea, the ternary in render below needs to be flipped too.

packages/@sanity/vision/src/css/jsonDump.css
```diff
@@ -7,7 +7,7 @@
 }
 
 .string {
```
bjoerge: Just to be extra nitpicky, maybe also rename this to `primitive` since it also gets applied to `null`, `number` and `boolean` values?

kmelve:
```suggestion
.primitive {
```

Like so?

PR #1030, opened 2018-10-24
[validation] Allow field validation overrides for top-level objects

rexxars:
## Background

You have defined a `localeString` type at the root level, and you have certain validation rules you generally apply ("english is required, because it is the default").

In certain locations, you want to override the validation with more rules. For instance, for a `title` field, you may want to set all the languages as required, or set the max length of all fields as 80 characters.

This is currently not possible, since validation is applied at definition time. 

## Implementation

This PR implements a `.fields()` validation rule on objects, which takes an object where the keys represent the name of the fields to provide rules for, which will override the previously declared rule. Should be used sparingly, but in certain situations (like the above), I think it makes sense.

## Usage

```js
{
  name: 'localeTitle',
  title: 'Localized title',
  type: 'localeString',
  description: 'English required by localeString, swedish from field',
  validation: Rule =>
    Rule.fields({
      se: fieldRule => fieldRule.required().max(80),
      no: fieldRule => fieldRule.optional().max(80)
    })
}
```

Feedback welcome!

### DIFF
--- a/packages/@sanity/schema/src/legacy/types/object.js
+++ b/packages/@sanity/schema/src/legacy/types/object.js
@@ -15,7 +15,8 @@ const OVERRIDABLE_FIELDS = [
   'description',
   '__unstable_searchFields',
   'options',
-  'inputComponent'
+  'inputComponent',
+  'validation'
 ]
 
 export const ObjectType = {
--- a/packages/@sanity/validation/src/Rule.js
+++ b/packages/@sanity/validation/src/Rule.js
@@ -56,6 +56,7 @@ class Rule {
     this._message = null
     this._required = undefined
     this._level = 'error'
+    this._fieldRules = undefined
     return this
   }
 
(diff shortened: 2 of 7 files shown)


superfunkminister:
I'm not sure if this is intentional or a bug, but if I have multiple `localeString` fields in the same document, this rule is applied to all of them, even if there's no validation set on the others...
```
// Rule is applied here as wanted
{
  name: 'localeTitle',
  title: 'Localized title',
  type: 'localeString',
  description: 'English required by localeString, swedish from field',
  validation: Rule =>
    Rule.fields({
      se: fieldRule => fieldRule.required().max(80),
      no: fieldRule => fieldRule.optional().max(80)
    })
},
// The same rule is applied here but unwanted
{
  name: 'localeTitle',
  title: 'Localized sentence',
  type: 'localeString'
}
```

PR #613, opened 2018-02-25
[preview] Apply incoming mutations to previewed documents

bjoerge:
Note: this is experimental and based on the refactoring done in #612.

This applies mutations to cached representations of previewed documents as they arrive. It will significantly reduce requests from the studio and will result in more instant UI updates of content that is currently being edited.

Consider this a proof of concept, we need to test this extensively before merging.

PR #1045, opened 2018-11-08
[form-builder] Guard against missing metadata on image asset

bjoerge:
This fixes an issue that could under some circumstances cause an error due to `assetDocument.metadata` not being set.

Even though this particular document type (`sanity.imageAsset`) never will have an undefined value for `metadata` in the data store, it can still be missing in the frontend as a consequence of how we optimize data fetching for previews. This will most likely not be an issue after #613 is in place.


bjoerge:
> maybe call the variable srcAspectRatio to avoid confusion with aspectRatio prop.

Good point, fixed.

PR #1050, opened 2018-11-15
[vision] Fix bug in vision plugin not remembering last used dataset.

### LINKED ISSUE #1042: [vision] Dataset selection isn't persistent
It seems that the dataset selection in Vision either isn't recovered from localStorage, always defaulting back to the first dataset in the list.

dan-dr:
fixes #1042 
also make default dataset the one defined in sanity.json (or the first dataset received if for some reason it doesn't exist)

For some reason there were extra changes for import orders and some formatting though prettier shouldn't have done these many changes. I think it's my VS Code doing it. Anyways it's very late and I just wanna PR. feel free to revert those changes :)

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.js
+++ b/packages/@sanity/vision/src/components/VisionGui.js
@@ -1,17 +1,17 @@
-import React from 'react'
-import PropTypes from 'prop-types'
-import queryString from 'query-string'
-import {storeState, getState} from '../util/localState'
-import parseApiQueryString from '../util/parseApiQueryString'
-import tryParseParams from '../util/tryParseParams'
-import DelayedSpinner from './DelayedSpinner'
-import QueryEditor from './QueryEditor'
-import ParamsEditor from './ParamsEditor'
-import ResultView from './ResultView'
-import NoResultsDialog from './NoResultsDialog'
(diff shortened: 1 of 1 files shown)


kmelve:
Thanks @dan-dr! We'll take a look at it soon 🙌

packages/@sanity/vision/src/components/VisionGui.js
```diff
@@ -273,7 +275,7 @@ class VisionGui extends React.PureComponent {
             <label className={styles.datasetSelectorContainer}>
               <span className={styles.datasetLabel}>Dataset</span>
               <Select
-                value={this.state.dataset || client.config().dataset}
+                value={this.state.dataset || client.config().dataset || this.props}
```
bjoerge: Is passing `this.props` as value to Select intentional?

dan-dr: :open_mouth: well of course not. this is what happens when you PR at 3AM... it's leftover code, must have glanced and thought there was only a formatting change there. can disregard this line.

bjoerge:
Aight, fixed formatting and pulled in the changes in 595b2de. Will be included in the next release.

Thanks again, @dan-dr, this is a time-saver! 👏

PR #1060, opened 2018-11-23
[form-builder] Block editor closes in dialog fix

kristofferj:
Fixes the problem where an blockeditor inside a dialog closes.


kristofferj:
This is fixed in the merged real-time-block-editor branch

PR #1072, opened 2018-11-30
Autocollapse hover & click area

kristofferj:
Larger click area on panes to collapse/expand and slightly gray (text-color alpha 1%) background color on hovered collapsed panes


mariuslundgard:
It works well!

Not sure sure about the grey active color on the collapsed pane. It's a little confusing, and also doesn't look that well. I think simply white works and looks better.

Also, would be great if we could lose the resizer border directly to the right of a collapsed pane. It makes the shadow look blurry. See screenshot:

image

PR #1075, opened 2018-12-01
[create-sanity] Add create-sanity package

bjoerge:
This adds the `create-sanity` package which enables `npm init sanity` scaffolding introduced in npm [5.10.0](https://github.com/npm/npm/releases/tag/v5.10.0).

More info here: https://docs.npmjs.com/cli/init#description

We could also consider adding `@sanity/create`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -10,7 +10,8 @@
     "packages/example-studio",
     "packages/movies-studio",
     "packages/storybook",
-    "packages/test-studio"
+    "packages/test-studio",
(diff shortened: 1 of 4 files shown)


bjoerge:
This also enables `yarn create sanity` btw

bjoerge:
> Is the use of var over const intentional? Not important, just curious.

LOL no, think my brain just went into legacy JavaScript mode given no transpiling here 😆

PR #1079, opened 2018-12-04
[components] Sync styleselect with realtime block editor branch

kristofferj:


kristofferj:
I have tested it. Seems to work nice both with arrow navigation and scroll on overflow.

PR #1083, opened 2018-12-05
Deprecate button collection

kristofferj:
Some plugins seems to use button-collection. Lets give a warning that it is deprecated


skogsmaskin:
:+1:

kristofferj:
ButtonGroup is to bind buttons together. ButtonGrid is a wrapper that places buttons in a grid.
We changed the name of buttonCollection so that it is not confused with buttonGroup.

PR #1091, opened 2018-12-07
Poc/button group in toolbar

kristofferj:
needs proper testing. Seems that we don't need `ToolbarClickAction` anymore, so i removed it.
It needs to be removed since it create an element that is not a button. `ButtonGroup` needs to have buttons as childs to work properly.


skogsmaskin:
Ok if I close this @kristofferj?

skogsmaskin:
Bump @kristofferj

PR #1093, opened 2018-12-07
[core] Add support for wildcard + credential-less CORS-origins

rexxars:
This PR adds support for adding CORS-origins which can contain wildcards, as well as being able to disallow credentials from certain CORS-origins.  The messages might need some work, they are difficult to get the right tone of voice on. Feedback welcome.


evenwestvang:
I think it's clear and does a good job in explaining what the dangers are. Made a microscopic change.

PR #1095, opened 2018-12-08
Search score by preview fields

kristofferj:
Not finished. This is a WIP branch, but we want some opinions if this will work


packages/@sanity/default-layout/src/components/SearchContainer.js
```diff
@@ -34,8 +34,34 @@ function removeDupes(documents) {
...
+function isGroq(query) {
+  if (!query) {
+    return false
+  }
+  return query.startsWith('*[') && query.endsWith(']')
+}
+
 const combineFields = flow([flatten, union, compact])
 
+const UNSAFE_GROQ_FIELDS = ['match', 'in', 'asc', 'desc', 'true', 'false', 'null']
```
mariuslundgard: We should probably call this variable `RESERVED_GROQ_KEYWORDS` instead.

mariuslundgard:
It needs more discussion, but I think it's better if the GROQ support works in the same in the global search as in the reference search.

Global seach input: `1984 [__type == 'book']` means `search for fields containing "1984" in documents of type "_book".

BTW. I think the GROQ filter should be wrapped in square brackets `[]` also in the reference search (where I believe we currently support `()`?).

I think it complicates things to support full-on GROQ queries in the global search (because we want to control which fields are returned).

mariuslundgard:
I made it so you can search for `text (filter)`, `text [filter]`, `[filter]`, `[filter] text [filter] text` and so on.

mariuslundgard:
Styled the search errors:

image

rexxars:
The tests aren't working anymore because of importing the schema part. We should either mock the schema or come up with a different solution for that.

kristofferj:
This is divided into smaller PR

PR #1098, opened 2018-12-11
[code-input] Implement schema part, better readme, use lazy requires

rexxars:
This PR does three things:

- Implement the schema type part so you don't have to explicitly require it
- Use lazy require statements in order to fix the issue where ace depend on window globals at require-time, which prevents schema introspection in Node.js
- Brush up the readme to reflect the latest changes, the name of the plugin, the correct usage, the available options as well as the data model

### DIFF
--- a/packages/@sanity/code-input/README.md
+++ b/packages/@sanity/code-input/README.md
@@ -1,18 +1,64 @@
-# code-editor
+# @sanity/code-input
 
-Ace editor for editing code.
+Code input for [Sanity](https://sanity.io/).
 
-Use `type: text` and `options.editor = 'code'`
+Currently only a subset of languages and features are exposed, over time we will implement a richer set of options.
 
-## In schema
+## Installation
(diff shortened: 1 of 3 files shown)


rexxars:
It would be nice if there was a way for schema types to hook into the schema error machinery in order to provide helpful error messages for cases like this.

If null/undefined values in the types array is filtered, we could simply implement the part as being a `null` (or rather, a self-executing function that  logs a warning and returns null)

An alternative is to implement the old schema part as some dummy type (`thatOldTypePreviouslyKnownAsCode`) and log a warning using a Immediately Invoked Function Expression ™️ 

Open to other suggestions.

rexxars:
I implemented the last one, since its the only one that doesn't require any changes to other Sanity packages. I still think it would be awesome to be able to provide errors to the schema machinery so they are surfaced in a better way than just the developer console

PR #1104, opened 2018-12-12
[preview] Use shallow equals on value when comparing props

bjoerge:
Did a small goof up in #1094 and assumed the preview component would always get passed a value with `_id`, but that's not the case for e.g. array items. This fixes it by instead doing a shallow compare on `props.value` in order to detect changes.

### DIFF
--- a/packages/@sanity/preview/src/components/ObserveForPreview.js
+++ b/packages/@sanity/preview/src/components/ObserveForPreview.js
@@ -2,6 +2,7 @@ import React from 'react'
 import WarningIcon from 'part:@sanity/base/warning-icon'
 import observeForPreview from '../observeForPreview'
 import {withPropsStream} from 'react-props-stream'
+import shallowEquals from 'shallow-equals'
 import {
   distinctUntilChanged,
   filter,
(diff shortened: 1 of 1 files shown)


packages/@sanity/preview/src/components/ObserveForPreview.js
```diff
@@ -44,7 +43,7 @@ const connect = props$ => {
   const isActive$ = sharedProps$.pipe(map(props => props.isActive !== false))
 
   return sharedProps$.pipe(
-    distinctUntilChanged((props, nextProps) => getId(props.value) === getId(nextProps.value)),
+    distinctUntilChanged((props, nextProps) => shallowEquals(props.value, nextProps.value)),
```
skogsmaskin: How often does this compare? Only when directly targeted via patch, or more than that? I'm haven't seen the context.

Anyway, ``getId`` is broken so approved upfront.

skogsmaskin:
:+1:

skogsmaskin:
Sorry, I meant to close the COMMENT window

packages/@sanity/preview/src/components/ObserveForPreview.js
bjoerge: It does a shallow compare whenever the component gets passed new props. Buffered document can be one of many possible sources of this - why?

PR #1109, opened 2018-12-13
[groq] Add groq module for use with tagged template literals

rexxars:
## Description

This module exports a single function that can be called with an ES2015 [template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals) to signal that it represents a GROQ-query.

The result will be the exact same string as the input, - this is currently helpful for getting syntax highlighting in editors, but in the future it might also parse and validate queries, strip unncessary whitespace and similar.

Pairs well with [vscode-sanity](https://github.com/sanity-io/vscode-sanity)!

## Usage

```js
import groq from 'groq'

const query = groq`*[_type == 'products'][0...10]`
```


rexxars:
As of now it doesn't really... do anything. Except help syntax highlighters.
We _could_ make it a UMD bundle, but I don't really see the big use case.

PR #1115, opened 2018-12-17
[desk-tool] Render pane content only when expanded

bjoerge:
This skips rendering of collapsed panes, which fixes an issue that made only the first item of collapsed panes visible.

As a bonus this means we now skip unnecessary DOM updates of contents that's not visible to the user. Also, we no longer refresh contents in collapsed panes, which should save a significant amount of API calls when editing contents on mobile devices

This also uncovered an issue with the global event listener that due to a race condition could cause the initial welcome event to be "consumed" before it got memoized with publishReplay.


bjoerge:
Merged in b3991cf

PR #1116, opened 2018-12-18
v0.138.0

bjoerge:


bjoerge:
Ah, good point, changed to 0.138.0

PR #1117, opened 2018-12-18
[form-builder] Block editor: fix scrollcontainer width issues

skogsmaskin:
Mystery bug, scrollbar eats of scroll-container width differently on same Chrome version and MacOS. Also different in various Chrome on Linux versions.

I added ``overflow-y: 'overlay';`` to make it calculate the same width everywhere.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/BlockEditor/styles/BlockEditor.css
+++ b/packages/@sanity/form-builder/src/inputs/BlockEditor/styles/BlockEditor.css
@@ -23,6 +23,7 @@
   border: 1px solid var(--input-border-color);
   border-radius: var(--input-border-radius);
   box-sizing: border-box;
+  overflow-y: overlay;
 
   @nest &.hasBlockExtras::after {
     content: '';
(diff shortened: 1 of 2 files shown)


kristofferj:
Overflow overlay is not recommended. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow
We had a big issue with this earlier. However, it's added to popper, but should not be used since it can cause major issues when someone removes this support from a library. https://github.com/FezVrasta/popper.js/pull/540

skogsmaskin:
It's obviously supported where it is an issue. It will fallback to 'auto' where not supported.

I recommend we use this now, and see what we can do about it at a later time.

kristofferj:
The problem is not the fallback, but other libs that looks for the supported overflows to calculate position of popovers etc.
It should be possible to calculate everything from the left side? However, we can do use overlay for a small time, but we have to make an issue to fix it before it is forgotten.

skogsmaskin:
Ah, good point!

The problem seems to be things like `calc(100% - something)` that doesn't account for the scrollbar width in some Chrome's. I actually got this varying on MacOS as well with the same Chrome version (but different OS version). Really strange.

skogsmaskin:
I've merged this for now. Let's have a closer look at it soon.

PR #1135, opened 2018-12-21
[core] Add GraphQL CLI commands for deploy / delete actions

rexxars:
## Note: this is still in beta.

This PR adds commands for deploying and undeploying GraphQL APIs to the CLI.

More information can be found [here](https://www.sanity.io/help/graphql-beta)

Moving this into next in order to make it easier to try out for our users, as well as keeping it in sync with the other improvements over time.


packages/@sanity/cli/src/util/noSuchCommandText.js
```diff
@@ -7,6 +7,7 @@ const coreCommands = [
   'build',
   'check',
   'configcheck',
+  'cors',
```
bjoerge: Is this intentional? Doesn't seem to be related to cors?

packages/@sanity/core/src/actions/graphql/deleteApiAction.js
```diff
@@ -0,0 +1,30 @@
+module.exports = async function deployApiActions(args, context) {
```
bjoerge: Should the function be named `deleteApiAction` here?

packages/@sanity/core/src/commands/graphql/deleteGraphQLAPICommand.js
```diff
@@ -0,0 +1,9 @@
+const lazyRequire = require('@sanity/util/lib/lazyRequire')
+
+export default {
+  name: 'delete',
```
bjoerge: Did we consider `sanity graphql undeploy`? Since we went for `sanity undeploy` for content studio, I feel it's a bit more natural to name it undeploy here as well.

packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
```diff
@@ -0,0 +1,410 @@
...
+        fieldName: '_rev',
+        type: 'String',
+        isNullable: false,
+        description: 'Current document revision'
+      }
+    ]
+  }
+
+  function arrayify(thing) {
+    return Array.isArray(thing) ? thing : [thing]
```
bjoerge: Depending on how `arrayify` is called, this could potentially yield `[undefined]` if `thing` is `undefined`. Maybe do an additional check and return an empty array in that case.

packages/@sanity/cli/src/util/noSuchCommandText.js
rexxars: I realized it was missing from the array while editing the file - didn't think to open a separate PR for it. 🤷‍♂️

packages/@sanity/core/src/actions/graphql/deleteApiAction.js
rexxars: Whoops, thanks!

packages/@sanity/core/src/commands/graphql/deleteGraphQLAPICommand.js
rexxars: True.

packages/@sanity/cli/src/util/noSuchCommandText.js
bjoerge: Ah, no need for a separate PR I think - but should'nt `graphql` be in that list too?

rexxars: Not yet, we don't want to explicitly hint at it before it's "public"

PR #1157, opened 2019-01-10
[ci] Visual regression testing using backstopjs

kristofferj:
- Dedicated studio
- Chrome headless is running in docker for exact render across platforms

### For designers:
`npm backstop-test-studio` to view studio and edit data
`npm test:backstop` (while studio is running) to test local changes. (docker required)
`npm test:backstop:approve` to approve visual changes


kristofferj:
Did som changes before we merge
- Moved the backstop folder to test 
- No need to run backstop-test-studio before running test (it starts and quits on npm run test:backstop )

mariuslundgard:
Hm, when I run `npm run test:backstop` I get

```
@sanity/core not installed in current project
Project-specific commands not available until you run `sanity install`
"start" is not available outside of a Sanity project context.
Run the command again within a Sanity project directory, where "@sanity/core"
is installed as a dependency.
```

mariuslundgard:
Yes, it works!

package.json
```diff
@@ -20,13 +20,17 @@
     "start": "npm run test-studio",
     "ecommerce-studio": "gulp ecommerce-studio",
     "test-studio": "gulp test-studio",
+    "backstop-test-studio": "gulp backstop-test-studio",
```
mariuslundgard: Is this script still used or can it be removed?

mariuslundgard: Ah, it's used for updating content.

PS. Would be nice to document how to use these tests in the root README.

PR #1178, opened 2019-01-23
[schema] Allow tel: and mailto: as URI schemes for block link annotation

rexxars:
We previously did not do any validation on annotations. After the new block editor, we do.

This PR will allow the mailto and tel schemes in the default link annotation.


skogsmaskin:
It will not allow for relative URL's I guess. Maybe it should?

rexxars:
Good question. In a sense, the only "correct" thing for backwards compatibility is to _disable_ validation, but that seems like a bad default for _new_ projects. Thoughts?

skogsmaskin:
I'm in favour of validating, but allowing any kind of URI.

PR #1185, opened 2019-01-29
[reflector-client] Add basic WebSocket support

rexxars:
As the title says - basic.

In the next iteration, I would want to try websockets first, then check if we get a close/error event from the socket before an open event is triggered. In this case, we would switch to EventSource mode instead.  We also need to add reconnect handling to the websocket implementation at some point. Probably a simple task for RxJS extraordinaire @bjoerge :)


rexxars:
Closing this as we are moving away from reflector

PR #1186, opened 2019-01-29
[test-studio] Add missing name

kristofferj:
Fixes a bug that causes this error in test-studio.
screenshot 2019-01-29 at 12 35 15


skogsmaskin:
👍

PR #1195, opened 2019-02-01
feat: add API versioning support to client and studio

rexxars:
This PR adds support for the upcoming API versioning feature.

Warns if using "tomorrow", in terms of UTC, throws if using a future date.


packages/@sanity/client/README.md
```diff
@@ -23,14 +23,25 @@ const sanityClient = require('@sanity/client')
...
+
+### Specifying API version
+
+Sanity uses ISO dates (YYYY-MM-DD) in UTC timezone for versioning. The explanation for this can be found [in the documentation](http://sanity.io/help/api-versioning)
+
+In general, unless you know what API version you want to use, you'll want to set it to todays UTC date. By doing this, you'll get all the latest bugfixes and features, while preventing any timezone confusion and locking the API to prevent breaking changes.
+
+**Note**: Do not be tempted to use a dynamic value for the `apiVersion`. The whole reason for setting a static value is to prevent unexpected, breaking changes.
+
+In future versions, specifying an API version will be required. For now (to maintain backwards compatiblity) not specifying a version will trigger a deprecation warning and fall back to using `v1`.
```
erikgrinaker: Is it required or not? First paragraph says `apiVersion` is required, here we say it's optional.

packages/@sanity/client/src/config.js
```diff
@@ -89,13 +76,60 @@ exports.initConfig = (config, prevConfig) => {
...
+  }
+
+  const today = new Date(Date.now() + nowOffset)
+  const tomorrow = new Date(Date.now() + nowOffset)
+  tomorrow.setDate(today.getDate() + 1)
+
+  const todayUtc = getUtcDate(today)
+  const tomorrowUtc = getUtcDate(tomorrow)
+
+  if (apiVersion === tomorrowUtc) {
```
erikgrinaker: I wonder whether we should have these sorts of checks in the client or not. We may want to change it (for whatever reason), and then the server- and client-side checks will no longer be in sync.

packages/@sanity/client/README.md
rexxars: "Almost required"? :)

packages/@sanity/client/src/config.js
erikgrinaker: Maybe the client should warn on dates in the future (because this is never a good idea in any case), but let the server-side API handle rejection of versions past tomorrow?

rexxars: This is the same check as we did in the PHP client (which you approved), but fair point.
It is kind of convoluted code that is prone to local clock drifts and whatnot, so if we all agree, I'd rather not include it.

erikgrinaker: Yeah, didn't think of it until now. I'd say let's warn on future UTC dates (to help the poor Aussies who are bound to set their local date), but skip the rest - or do you think we should just remove it altogether?

rexxars: Not sure I get the distinction:
"Client should warn on dates in the future"
"Server-side API should reject on versions past tomorrow"

Do you mean that the client should warn if the date is beyond "tomorrow" UTC-wise, but leave the "tomorrow" check up to the server? If this is the case, I'd rather just remove the logic from the client at all, as date checks feel kind of brittle 🤷‍♂️

rexxars:
This is now ready to be reviewed again.

- Client has more basic versioning checks, but prints `X-Sanity-Warning` headers
- Studio client now has `withConfig()` method which requires you set an `apiVersion`. Backwards-compatible with a warning.
- All studio code (I think) updated to specify an API version (currently `v1`)

packages/@sanity/client/README.md
```diff
@@ -23,14 +23,25 @@ const sanityClient = require('@sanity/client')
...
 `const client = sanityClient(options)`
 
-Initializes a new Sanity Client. Required options are `projectId` and `dataset`.
+Initializes a new Sanity Client. Required options are `projectId`, `dataset` and `apiVersion`. Setting a value for `useCdn` is encouraged.
+
+### Specifying API version
+
+Sanity uses ISO dates (YYYY-MM-DD) in UTC timezone for versioning. The explanation for this can be found [in the documentation](http://sanity.io/help/api-versioning)
+
+In general, unless you know what API version you want to use, you'll want to set it to todays UTC date. By doing this, you'll get all the latest bugfixes and features, while preventing any timezone confusion and locking the API to prevent breaking changes.
```
erikgrinaker: Nitpick: maybe put "locking the API" before "timezone confusion", since it's a much greater benefit than the other.

packages/@sanity/client/src/config.js
```diff
@@ -89,13 +76,27 @@ exports.initConfig = (config, prevConfig) => {
...
-      newConfig.url = `${newConfig.apiHost}/v1`
+      newConfig.url = `${newConfig.apiHost}/v${newConfig.apiVersion}`
       newConfig.cdnUrl = newConfig.url
     }
   }
 
   return newConfig
 }
+
+exports.validateApiVersion = function validateApiVersion(apiVersion, nowOffset = 0) {
```
erikgrinaker: We don't strictly need to validate this, since the API gateway will also validate it and generate an appropriate error message - but obviously feel free to include it if you feel like it's beneficial.

rexxars:
> LGMT! 👍 Tests are failing, in case you're not aware already.

Fixed the tests, thanks for noticing :)

skogsmaskin:
Should we merge this, or what's the status?

rexxars:
Given that it's been two years, maybe you can give this another look, @bjoerge ? 😅 
I'll squash/tidy up the commits a bit after approval.

bjoerge (approved):
LGTM! (still 😄 ) And kudos for Vision => Sanity UI work!

PR #1196, opened 2019-02-01
[default-layout] Reset tool when selecting active tool

kristofferj:
When clicking on the current active tool it is expected that the tool resets to its initial state.
Let me know if this is an ok way of doing it.


bjoerge:
Looks good! I did a tiny rewrite in order to make it more future proof by having less assumptions about which router state keys might or might not be present.

PR #1197, opened 2019-02-03
[core] Fix breaking change in upstream terser dependency

rexxars:
Terser decided to switch to bundle their releases with rollup in a minor change. This broke our minifying;

```
TypeError: Cannot read property 'minify' of undefined
    at Object.<anonymous> (/Users/espenh/webdev/tersertest/node_modules/@sanity/core/lib/actions/build/compressJavascript.js:12:29)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Module._compile (/Users/espenh/webdev/tersertest/node_modules/pirates/lib/index.js:83:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.newLoader [as .js] (/Users/espenh/webdev/tersertest/node_modules/pirates/lib/index.js:88:7)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
```

This PR imports both the default import and the named `minify` import, checks which one exists and uses that. Ugly, but probably a good choice given they may revert the change.

### DIFF
--- a/packages/@sanity/core/src/actions/build/compressJavascript.js
+++ b/packages/@sanity/core/src/actions/build/compressJavascript.js
@@ -1,5 +1,5 @@
 import fse from 'fs-extra'
-import Terser from 'terser'
+import Terser, {minify as minifyJs} from 'terser'
 
 export default async inputFile => {
   const content = await fse.readFile(inputFile, 'utf8')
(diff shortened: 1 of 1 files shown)


rexxars:
Ironically, the CI fails because the webpack plugin used to build the CLI depends on terser.

PR #1202, opened 2019-02-06
[cli] Show stack traces on errors by default

rexxars:
At some point we (I?) decided that not showing stack traces on errors would be "cleaner" for regular users. After trying to debug problems with our users for a while, I feel perhaps it is saner to display a stacktrace so we can more easily track down the source of problems. This is especially true on intermittent errors that are not reproducible when re-running with the `-d` flag.


kmelve:
Would it be possible to have “both”, i.e. make the error message easily identified and readable, in addition to having the stack trace available? I agree that would make support a bit easier, which is good for all parties.

rexxars:
> Would it be possible to have “both”, i.e. make the error message easily identified and readable, in addition to having the stack trace available? I agree that would make support a bit easier, which is good for all parties.

I'm not sure how you mean.
```js
Error: Some descriptive error here
  at someSourceLocation.js:123
  at someSourceLocation.js:321
  at someSourceLocation.js:19
  at someSourceLocation.js:25
```

How do you want this to be printed, @kmelve ?

kmelve:
In my experience, people tend to miss the actual error in stack traces because they're confronted with a bunch of text. Some simple formatting/whitespace between the error message and the trace would alleviate that. 

```
Error: Some descriptive error here

  at someSourceLocation.js:123
  at someSourceLocation.js:321
  at someSourceLocation.js:19
  at someSourceLocation.js:25
```

rexxars:
How about using something like [this](https://github.com/shinnn/neat-stack)?

PR #1203, opened 2019-02-06
[schema] Fix block annotation validation not handling certain objects

rexxars:
In certain cases, it seems `jsonType` is not returned when validating the schema types. For instance, when using the color schema type as a block annotation, `type` is returned as `object`, but `jsonType` is not present. Maybe @bjoerge has some insight on whether this approach is wrong in the  first place.

Example:
```js
export default {
  name: 'pageBody',
  title: 'Page body',
  type: 'array',
  of: [
    {
      type: 'block',
      marks: {
        annotations: [
          {
            name: 'link',
            title: 'URL',
            type: 'object',
            fields: [
              {
                title: 'URL',
                name: 'href',
                type: 'url'
              }
            ]
          },
          {name: 'color', title: 'Color', type: 'color'}
        ]
      }
    },
    {type: 'image'},
    {type: 'callToAction'},
    {type: 'placeholder'}
  ]
}
```

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/types/block.js
+++ b/packages/@sanity/schema/src/sanity/validation/types/block.js
@@ -183,7 +183,8 @@ function validateAnnotations(annotations, visitorContext, problems) {
 
     const {_problems} = visitorContext.visit(annotation, visitorContext)
     const targetType = annotation.type && visitorContext.getType(annotation.type)
-    if (targetType && targetType.jsonType !== 'object') {
+    const jsonType = targetType && (targetType.jsonType || targetType.type)
+    if (jsonType && jsonType !== 'object') {
       _problems.push(
         error(
           `Annotation cannot have type "${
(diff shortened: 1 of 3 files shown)


bjoerge:
The fact that there's no `jsonType` here sounds to me like the block annotation types aren't getting compiled properly, so what you get back is probably the raw schema definitions, is that right?

skogsmaskin:
It should just ignore `jsonType` here all together. Just check on type. This is run before the schema is proplery compiled. It will get `jsonType` when validated and then compiled.

skogsmaskin:
Forget what I said, inline defined objects will have `type: 'type'` here.

bjoerge:
I totally missed the context here - so since this in the schema validation step (I somehow managed to perceive the context as form builder _validation_)

Pushed a fix for this that figures out jsontype through the parent type chain (the validation type traversal works somewhat different than the regular schema traversal/parsing, and the jsonType is not "inherited" here, but only exists on the root type).

PR #1205, opened 2019-02-07
[example-studio] Add duration w/derived seconds input example

bjoerge:
This adds an example of how to create a custom input that preserves a user typed value (duration on format `hh:mm:ss`), while at the same time storing a derived value (the duration converted to seconds).

### DIFF
--- a/packages/example-studio/components/DurationWithSecondsInput.js
+++ b/packages/example-studio/components/DurationWithSecondsInput.js
@@ -1,0 +1,80 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import {PatchEvent, set, setIfMissing, unset} from 'part:@sanity/form-builder/patch-event'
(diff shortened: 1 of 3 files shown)


kmelve:
Very cool – I missed a simple example like this! 👏 

But wouldn't it be better as an example for `sanity init plugin`? We have the chess input example there now, which is pretty cool, but also a bit complex and elaborate when you just want to make a simple input. Also, seems like it conceptually fits better there, than in the blog example?

rexxars (approved):
👍

skogsmaskin (approved):

camjc:
Would be cool to follow iso 8601 like for dates (P1YT2S is one year 2s)

rexxars:
Any reason not to merge this, @bjoerge ?

PR #1218, opened 2019-02-21
[form-builder] Better handling of unsetting field when block editor is empty

skogsmaskin:
The last PR on this issue broke the tests because there were no longer a 1-1 relationship between the editor state and the actual data.

This is now handled in a way that the editors content gets in a placeholder state (a single block  with no text in the document gets a data attribute `placeholder` toggled by an own plugin on `keyUp`).

The function that converts editor operations to patches looks for this state and will send an unset patch if that is the case. Any other operations on a placeholder block is just ignored until there is actual content.

Also included are some minor fixes (setifMissing patches in additional required contexts) plus some minor other fixes.


mariuslundgard:
👏

PR #1222, opened 2019-02-28
[base] Split string on space and comma instead of using lodash.words for tokenization

bjoerge:
In #1168 we used `lodash.words` for tokenization of the search string. Can't really remember the rationale behind it (think it had something to do with unicode, but in retrospect it feels misguided). Using `lodash.words` had some unintended side effects, like splitting words like `2pac` into `['2', 'pac']`.

This PR replaces `lodash.words` with a simple regex that splits the search string on space+comma.


mariuslundgard:
A few more test cases could be:

```
"foo,bar,baz"
"foo,,bar,,,baz"
"foo ,, bar ,,, baz"
```

PR #1226, opened 2019-03-04
Block editor dialog title

kristofferj:
- Use DialogContent to set the size of the content
- Set title on dialogs on block editor to the type title

Before:
screenshot 2019-03-04 at 13 25 16

After:
screenshot 2019-03-04 at 13 20 38


packages/@sanity/form-builder/src/inputs/BlockEditor/EditNode.js
```diff
@@ -105,17 +105,19 @@ export default class EditNode extends React.Component<Props> {
     const {type, node} = this.props
     const nodeRef = findDOMNode(node)
     const editModalLayout = get(type.options, 'editModal')
+    const title = `${type.title}`
```
kmelve:
```suggestion
    const {title} = type
```

PR #1234, opened 2019-03-12
[code-input] Add a filename field to code input blocks

nervetattoo:
This adds a filename input field for the code-input blocks:

Screenshot 2019-03-12 at 12 06 23

The filename will be used for the previewing instead of the language as the language is both represented as the file ending as well as with a logo

Screenshot 2019-03-12 at 12 06 30

### DIFF
--- a/packages/@sanity/code-input/src/CodeInput.js
+++ b/packages/@sanity/code-input/src/CodeInput.js
@@ -7,6 +7,7 @@ import styles from './CodeInput.css'
 import FormField from 'part:@sanity/components/formfields/default'
 import Fieldset from 'part:@sanity/components/fieldsets/default'
 import DefaultSelect from 'part:@sanity/components/selects/default'
+import TextInput from 'part:@sanity/components/textinputs/default'
 import createHighlightMarkers from './createHighlightMarkers'
 import {
   ACE_EDITOR_PROPS,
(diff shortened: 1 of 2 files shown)


rexxars:
Sweet! I was wondering if perhaps it would make sense to add an `options.withFilename` (or similar, naming is hard) to the options for a field. We're using this in a lot of our documentation for snippets, where a filename doesn't necessarily make any difference/sense.

nervetattoo:
I were thinking of that as well @rexxars , as you mention it doesn't make sense for small partials.

What I think would be a better approach, but I'm not sure if it goes against any best practices or philosophies you have, is to allow to set `fields` on the schema object and have it merged with the built-in fields. It's another ballpark of complexity implementation wise though as it will have to disallow conflicts with the existing fields with smart merging.

I can simply add the `withFilename` option as I personally don't want to spend time on the mergable fields 🙈

rexxars:
We're moving away from having the shape of a schema type differ based on the options, since it makes strict schemas (for GraphQL, for instance) hard to deal with. For instance, we previously included the `hotspot` and `crop` fields in the image type conditionally based on whether or not they were enabled, whereas we now simply hide the fields in the UI. This makes the content model more predictable, but still allows a good user interface.

nervetattoo:
`withFilename` option added, defaulted to `false` (force pushed to keep at 1 commit)

rexxars:
Thanks! We'll include this in the next release

PR #1240, opened 2019-03-15
[form-builer] Block editor - fix focus bug when toggling fullscreen mode

skogsmaskin:
There was a bug when toggling to fullscreen on an editor which not yet has received focus that the `requestAnimationFrame` was placed in the wrong location. `requestAnimationFrame` must be run before trying to reference the editor React node.


skogsmaskin:
@kristofferj - it will wait for the fullscreen to complete the render, and then there will we a populated React reference again to do `.focus` on.

PR #1245, opened 2019-03-19
Custom block type support

skogsmaskin:
This makes support for defining and naming your own `block` types and using them with editors for portable text just like the standard `block` type:

```
[
  {
    type: "array",
    name: "body",
    title: "Body Text",
    of: [{ type: "myCustomBlockTypeWithASpecialFeatureSet" }]
  }
]
```


jemmima:
@skogsmaskin @simen are we ready to do this now do you think?

skogsmaskin:
We have this functionality now.

PR #1252, opened 2019-03-23
[code-input] Add support for .sass and .scss

mmgj:
Because life is too short to be using vanilla CSS.


KacperZi:
If I understand this correctly, SASS is now possible to use within sanity studio right? Could someone explain please, how to be able to import SCSS in my `./web/pages/_app.js`? 

I've run `sanity install @sanity/code-input` to install "code-input" but i can't find the package after installing.

If i simply add `import ../styles/custom.scss` inside `./web/pages/_app.js` im getting the following error:
"Module parse failed: Unexpected token (2:0)
You may need an appropriate loader to handle this file type."

So i tried to run `npm install node-sass --save` in my root as well as in ./web which is not fixing this problem. 

I've also checked the documentation of Sanity IO but I seem to missunderstand something, please help.

PR #1258, opened 2019-03-29
[cli] Prompt to send community invitation on first project creation

rexxars:
When a user creates their first project, this PR will ask if they want to be invited to join the community.  I'm not attached to the prompt text, feel free to suggest changes (and decide whether or not this is a good idea ;) )

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.js
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.js
@@ -250,7 +250,7 @@ export default async function initSanity(args, context) {
     (await prompt.single({
       type: 'confirm',
       message:
-        'We have an awesome developer community, would you like us to send you an invitation to join?',
+        'We have an excellent developer community, would you like us to send you an invitation to join?',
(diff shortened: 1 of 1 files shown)


kmelve:
It's a good idea.

evenwestvang:
If this PR was a teddy bear I would be hugging it. Only suggestion would be to change _awesome_ to _excellent_.

packages/@sanity/cli/src/actions/init-project/initProject.js
```diff
@@ -245,6 +245,24 @@ export default async function initSanity(args, context) {
     print(`\n${successMessage}`)
   }
 
+  const sendInvite =
+    isFirstProject &&
+    (await prompt.single({
+      type: 'confirm',
+      message:
+        'We have an awesome developer community, would you like us to send you an invitation to join?',
```
kmelve:
```suggestion
        'We have an excellent developer community, would you like us to send you an invitation to join?',
```

PR #1262, opened 2019-04-05
[core] Allow specifying non-HTTP protocols for CORS-origins

rexxars:
Needs backend change to be merged first, but this will allow the CLI to add any protocol as CORS-origin.


skogsmaskin:
Manage should probably be updated as well?

rexxars:
> Manage should probably be updated as well?

Yeah. That is a separate task.

PR #1264, opened 2019-04-06
[preset-default] Introduce default preset

rexxars:
I'm not sure whether or not this is a good idea, so just think of this as a proof of concept.

I wanted to cut down on the amount of plugins we need to list in `package.json` and `sanity.json`. This PR introduces a "default" preset which includes the packages we basically include in all studios at this point: base, components, default-layout, default-login, desk-tool.

Thoughts? Opinions?


mariuslundgard:
I think this would simplify, but perhaps make more sense when/if there are many more dependencies to track?

kristofferj:
This looks simpler, but I think it does not communicate that the studio actually are building blocks. I think that is a nice think about splitting it.

PR #1266, opened 2019-04-10
[form-builder] Correct function name and remove asset from state

kristofferj:


packages/@sanity/form-builder/src/inputs/ImageInput/SelectAsset.js
```diff
@@ -52,8 +52,16 @@ export default class SelectAsset extends React.Component<Props, State> {
     })
   }
 
-  handleDeleteFinished = () => {
-    this.fetchPage(this.pageNo)
+  handleDeleteFinished = id => {
+    this.setState(prevState => {
```
bjoerge: Rather than mutating `prevState`, this should be simplified to

```js
this.setState(prevState => ({
  assets: prevState.assets.filter(asset => asset._id !== id)
}))
```

PR #1270, opened 2019-04-14
[form-builder] Support additional filter to reference input

rexxars:
Sometimes, you may want to be more specific about which documents to reference than just the document type. This is a basic implementation where you can pass a `filter` and an optional `filterParams` in the `options` object of reference fields.

Let the discussion commence!

### DIFF
--- a/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
+++ b/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
@@ -1,9 +1,10 @@
+/* eslint-disable import/prefer-default-export */
+import {compact, toLower, flatten, uniq, flow, sortBy, union} from 'lodash'
 import {map} from 'rxjs/operators'
 import {joinPath} from '../../util/searchUtils'
-import {compact, toLower, flatten, uniq, flow, sortBy, union} from 'lodash'
 import {removeDupes} from '../../util/draftUtils'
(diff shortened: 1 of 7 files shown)


kmelve:
Cool! Should it be `params` like in the structure builder API?

rexxars:
> Cool! Should it be `params` like in the structure builder API?

Not sure. There is a chance we will add other options to the reference fields at some point, which might make it hard to know what "params" applies to

kmelve:
I'm not sure either. Does it make sense to use `params` here at all btw? Why wouldn't you just have it all in the `filter` expression?

E.g. we don't have `params` for `filter` in the access control (at least not documented).

rexxars:
Not having params in filters seems like an oversight more than a feature. Parameters are more readable and less error-prone, so I definitely prefer it to string concatenation.

rexxars:
After some discussion on Slack, I've changed the implementation to be able to take a function as `filter`, which can return `filter` and `params`. The function receives the document, the parent node and the parent path. With this, you should be able to create a constraint based on the surrounding document. I added an example in the test studios reference schema, where a nested object has a "decade" field and controls a sibling reference.

Also added validation of the properties. Will add help pages if PR is accepted.

kmelve:
[this is super cool!]

rexxars:
Help articles written. Will add documentation before release.

Issue #1272, opened 2019-04-18
Custom wrappers/providers from plugins


good-idea:
When developing plugins, it would be great to be able to supply custom providers that are rendered at the top of the tree. In my case: I'm building a Shopify syncing plugin. User settings for the plugin are stored in a document. These settings are required in both a "Settings" tab (rendered as a Tool), as well as in each instance of a Shopify product or collection within the Desk tool.

Without a provider, each of these components requires fetching these settings when they mount. This adds up as each component mounts and unmounts - and, there isn't any guarantee that they are able to keep in sync.

With a custom provider, I can fetch these settings at the top-level, one time, and pass them down anywhere I need them using context.

I've made a PR that sets this up. Here is how it looks:

![image](https://user-images.githubusercontent.com/11514928/56391763-70896100-61e4-11e9-9a95-06ee826aac83.png)

![2019-04-18 14 17 59](https://user-images.githubusercontent.com/11514928/56391980-091fe100-61e5-11e9-8ba3-755a91d1e7e1.gif)

It's not perfect - I'll outline the issue in the PR.

simen:
This is an interesting pattern, and we will consider it. What we do ourselves is implementing singleton data provider instances (datastores) for stuff like users and document state etc. Every component that need those pieces of information will call on these datastores and the datastores will take care of caching and reusing objects or otherwise minimizing actual data traffic. This is very flexible and is the way most of the studio is implemented. I can absolutely see the benefit of your proposed approach, and we will take it into consideration as we iterate on the studio customization APIs.

PR #1273, opened 2019-04-18
[desk-tool] Plugin providers

good-idea:
Addresses #1272

Allows plugins to include custom wrappers. (See demo screengrabs in issue)

One problem: these components re-render whenever the user switches from tool to tool. This is implemented in `<DefaultLayoutContainer>` - I wasn't able to identify a better place for it. But, it would be simple to move elsewhere.

- [ ] - write tests
- [ ] - fix multiple renders

If you're into this, it would be super helpful when developing plugins!!

### DIFF
--- a/packages/@sanity/default-layout/src/components/ComposedProvider.js
+++ b/packages/@sanity/default-layout/src/components/ComposedProvider.js
@@ -1,0 +1,36 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const Composed = ({wrappers, children}) => {
+  console.log(wrappers)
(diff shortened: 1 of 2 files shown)


good-idea:
Hey all, any thoughts on if you'd be into supporting something like this? (or, any suggestions on how to achieve this with a plugin?)

kmelve:
Hi @good-idea, and thanks for the PR! We're currently evaluating this. Stay tuned!

rexxars:
We're refactoring the studio internals and will keep this use case in mind - thanks for the PR and inspiration!

PR #1276, opened 2019-04-23
[all] Adding standard sanity favicons for all templates

kristofferj:
Don't know if there is any better way of doing this. Feels wrong to put the same files many places.


rexxars:
We decided we'd solve this differently, so closing this for now.

PR #1280, opened 2019-04-26
[dashboard] Update design

mariuslundgard:
Updates the tutorials stream to use flex box (to allow for cut-off of the last item, for discoverability).

image

Changes the max-height logic to fix weird overflow scrolling.

Adds play icon to tutorial items.

### DIFF
--- a/packages/@sanity/dashboard/src/widget.css
+++ b/packages/@sanity/dashboard/src/widget.css
@@ -21,7 +21,7 @@
 }
 
 .title {
-  composes: heading4 from "part:@sanity/base/theme/typography/headings-style";
+  composes: heading4 from 'part:@sanity/base/theme/typography/headings-style';
   padding: var(--small-padding) var(--medium-padding) !important;
 }
 
@@ -34,7 +34,6 @@
 
   @media (--screen-medium) {
     height: stretch;
-    max-height: 30rem;
     overflow-y: auto;
   }
 }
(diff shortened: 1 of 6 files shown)


packages/@sanity/dashboard/src/widget.css
```diff
@@ -21,7 +21,7 @@
 }
 
 .title {
-  composes: heading4 from "part:@sanity/base/theme/typography/headings-style";
+  composes: heading4 from 'part:@sanity/base/theme/typography/headings-style';
```
kristofferj: Is there a reason for using single quotes?

mariuslundgard: It’s prettier that’s doing it!

kristofferj: Ah. You should disable prettier for CSS make it follow til stylelint rules. A lot of the changes on this PR was probably just prettier

mariuslundgard: I’ve enabled that stylelint integration for Prettier in VSCode, though.

mariuslundgard: image

kristofferj: Maybe we should put a line in `.editforconfig`

mariuslundgard: I think the `stylelint-config-standard` which is used, prefers single quotes.

PR #1285, opened 2019-04-29
[validation] Add new `assetRequired`-flag

rexxars:
When you have an image or a file field with custom metadata fields attached, `required()` will still yield true if you have set one of those fields. This PR introduces `assetRequired()`, which requires that the actual `asset` field is set and has a `_ref`.


nickgriffiths:
I just came across this and found it useful. I couldn't find it in the [documentation](https://www.sanity.io/docs/image-type). Do you think it's worth adding it, or was I looking in the wrong place?

PR #1286, opened 2019-05-02
[client] Raise default timeout, allow overriding per-request

rexxars:
This PR ensures that the default timeout is above the technical limits (5 minutes, vs 3 minute mutation and 1 minute query timeout). This will lead to more understandable error messages on timeouts, and being able to more easily differentiate between _client_ and _server_ timeouts.

Also added the ability to override timeout and token on a per-request basis, which can be quite useful in a lot of situations.


erikgrinaker:
SGTM. It might be worth differentiating between request timeouts and network timeouts, if possible. The initial TCP connection handshake should have a much lower timeout (e.g. 10 seconds or something reasonable), but the response timeout should be above our operation timeouts.

PR #1288, opened 2019-05-03
[import] Don't override _type for imported assets

rexxars:
If you have the following in an exported dataset dump:
```js
{
  _id: 'foo',
  _type: 'someDocument',
  coverImage: {
    _type: 'myCustomImage',
    _sanityAsset: 'image@some-path.jpg'
  }
}
```

The `_type` property will be overwritten:
```js
{
  _id: 'foo',
  _type: 'someDocument',
  coverImage: {
    _type: 'image',
    asset: {_ref: 'some-ref'}
  }
}
```

This PR fixes this by using a `setIfMissing` patch instead.


evenwestvang:
tres grand vitesse @rexxars

PR #1290, opened 2019-05-07
[dashboard] Fixes

kristofferj:


packages/@sanity/dashboard/src/widget.css
```diff
@@ -45,6 +44,7 @@
   min-height: 4em;
   height: 4em;
   margin-top: auto;
+  overflow:hidden;
```
thomax: Is there no linter in operation which auto-adds a whitespace after `:`?

kristofferj: oh. i will add that space manually

kristofferj: There is no auto-fix enabled. But stylelint gave it a red line that i missed

kristofferj:
Yes, this fix "wrong number of feed items".

PR #1304, opened 2019-05-16
[base] Include arrays in search

runeb:
This code allows searching in things like tag arrays. The search term must be a complete match to get a hit.

### DIFF
--- a/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
+++ b/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
@@ -7,9 +7,16 @@ import {tokenize} from '../common/tokenize'
 
 const combinePaths = flow([flatten, union, compact])
 
+const isArrayPath = path => path.endsWith('[]')
+
+// "some[].thing[]" => "some[].thing"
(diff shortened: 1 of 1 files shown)


packages/@sanity/base/src/search/weighted/createWeightedSearch.js
```diff
@@ -7,9 +7,16 @@ import {tokenize} from '../common/tokenize'
...
 const combinePaths = flow([flatten, union, compact])
 
+const isArrayPath = path => path.endsWith('[]')
+
+// "some[].thing[]" => "some[].thing"
+const joinedPathRemoveArray = path => path.substr(0, path.length - 2)
+
 const toGroqParams = terms =>
   terms.reduce((acc, term, i) => {
     acc[`t${i}`] = `${term}*` // "t" is short for term
```
runeb: Couldn't we just append the `*` later? Why does it need to be included in the terms like this?

coreyward:
Would be great to get this merged in—I'm trying to search documents by a field that holds an array of tags and right now it's not possible.

rexxars:
We need to reimplement this in a larger studio search refactoring as we're currently hitting the limits of the current implementation.

PR #1310, opened 2019-05-20
[schema] Provide fallback for preview select if none is given

rexxars:
This does not yield `Static title`:
```js
export default {
  name: 'noTitleField',
  type: 'document',
  title: 'No title field',
  preview: {
    prepare: () => ({title: 'Static title'})
  },
  fields: [
    {
      name: 'isChecked',
      type: 'boolean',
      title: 'Is checked?'
    }
  ]
}
```

This works:
```js
export default {
  name: 'noTitleField',
  type: 'document',
  title: 'No title field',
  preview: {
    select: {},
    prepare: () => ({title: 'Static title'})
  },
  fields: [
    {
      name: 'isChecked',
      type: 'boolean',
      title: 'Is checked?'
    }
  ]
}
```

With this PR, the first example also works :)

### DIFF
--- a/packages/@sanity/schema/src/legacy/preview/createPreviewGetter.ts
+++ b/packages/@sanity/schema/src/legacy/preview/createPreviewGetter.ts
@@ -13,7 +13,7 @@ function parsePreview(preview) {
   if (!preview) {
     return preview
   }
-  const select = preview.select || preview.fields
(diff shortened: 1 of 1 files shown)


runeb:
This just bit me as well. Any reason why this is not merged?

PR #1312, opened 2019-05-20
[desk-tool] Moving overlay out of scroll container

kristofferj:
Moved the overlay out of the scroll container to maintain relative height.


kristofferj:
I agree. 
Do you think it should prevent scroll/pointerEvents?

rexxars:
Yeah, that seems like a good idea to me. Just hard to see what's going in currently.

kristofferj:
Fixed and rebased

PR #1313, opened 2019-05-20
Throw error when documentTypeListItems is configured with non-existing type

thomax:


thomax:
Fixes https://github.com/sanity-io/sanity-priv/issues/1058

PR #1336, opened 2019-05-24
[components] Checkbox font size fix

kristofferj:


packages/@sanity/components/src/toggles/styles/Checkbox.css
```diff
@@ -17,7 +17,7 @@
   z-index: 1;
   display: inline-block;
   box-sizing: border-box;
-  height: var(--checkbox-label-height);
+  /* height: var(--checkbox-label-height); */
```
bjoerge: Can this be removed? Or if it needs to stay - add a comment about why?

kristofferj: It shoudl go

PR #1355, opened 2019-06-07
[server] Prevent build error on symlinked modules

rexxars:
Fixes react-hot-loader not resolving in symlinked modules


rexxars:
Meh, looks like the new hot-loader is significantly slower than the old one. I'm going to reduce this down to just the resolving issue for now.

PR #1379, opened 2019-06-19
[desk-tool] Omit _createdAt, _updatedAt when creating duplicate document

rexxars:
When you duplicate a document using the desk tool, the timestamps (`_createdAt`, `_updatedAt`) are kept as-is. Whether or not this is a bug or a feature is up for discussion, I guess, but it feels odd to me. I was expecting to see my new duplicate document at the top of the list when ordering by updated date.

This PR omits these timestamps from the cloned document.

### DIFF
--- a/packages/@sanity/desk-tool/src/pane/Editor.js
+++ b/packages/@sanity/desk-tool/src/pane/Editor.js
@@ -305,10 +305,11 @@ export default withRouterHOC(
     handleCreateCopy = () => {
       const {router, draft, published, paneIndex} = this.props
       const prevId = getPublishedId((draft || published)._id)
+      const omit = ['_createdAt', '_updatedAt']
 
       const duplicatedDocument = this.isLiveEditEnabled()
-        ? copyDocument(published)
-        : newDraftFrom(copyDocument(draft || published))
+        ? copyDocument(published, {omit})
+        : newDraftFrom(copyDocument(draft || published, {omit}))
 
       this.duplicate$ = documentStore.create(duplicatedDocument).subscribe(copied => {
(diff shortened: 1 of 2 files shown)


packages/@sanity/desk-tool/src/pane/Editor.js
```diff
@@ -305,10 +305,11 @@ export default withRouterHOC(
     handleCreateCopy = () => {
       const {router, draft, published, paneIndex} = this.props
       const prevId = getPublishedId((draft || published)._id)
+      const omit = ['_createdAt', '_updatedAt']
 
       const duplicatedDocument = this.isLiveEditEnabled()
-        ? copyDocument(published)
-        : newDraftFrom(copyDocument(draft || published))
+        ? copyDocument(published, {omit})
+        : newDraftFrom(copyDocument(draft || published, {omit}))
```
thomax: Should we also omit `_createdAt` and `_updatedAt` when moving from doc to draft?

thomax: Not saying we should/shouldn't, but unsure if I can imagine all use cases from where I'm sitting right now

rexxars: This is only when explicitly using the "duplicate document" action - the check highlighted here just checks whether or not we're in "live edit" mode or not.

When editing a document, I think it makes sense to keep the timestamps. When _duplicating_ a document, I feel like that should be treated as a new document but with the same... "user-level" field values. My two cents anyway :)

thomax: Gotcha. Then I'm all for this! Good job Espen 👍

PR #1386, opened 2019-06-25
[base] History support

skogsmaskin:


packages/@sanity/base/src/datastores/history/createHistoryStore.js
```diff
@@ -0,0 +1,111 @@
...
+    scan((prev, next) => {
+      return {...prev, ...next}
+    }, {}),
+    map(transactions =>
+      transactionsToEvents(pairs, Object.keys(transactions).map(key => transactions[key])).reverse()
+    )
+  )
+}
+
+export default function createHistoryStore() {
```
rexxars: Why don't we just export these directly as named exports instead of exporting a default factory function? If we're going for the factory pattern, we should at least  make the cache local to the instance returned?

packages/@sanity/base/src/datastores/user/createUserStore.js
```diff
@@ -75,10 +75,16 @@ const getUser = id => {
   return userCache[id]
 }
 
+// TODO Optimize for getting all users in one query
+const getUsers = ids => {
+  return Promise.all(ids.map(id => getUser(id))).then(users => users)
```
rexxars: This last `then` is a noop

packages/@sanity/base/sanity.json
```diff
@@ -368,6 +358,10 @@
       "name": "part:@sanity/base/datastore/document",
       "path": "datastores/document"
     },
+    {
```
rexxars: We should potentially start to think about reducing the number of parts we expose - for instance by exposing a `@sanity/base/datastores` directly so you could import from that directly? This would allow for better autocomplete in editors, for instance. Not _critical_ right now, but worth thinking about

packages/@sanity/base/src/datastores/history/createHistoryStore.js
skogsmaskin: I think this is actually out of the scope for this PR, this is the pattern used on every other store we have, so it should be refactored in it's own PR I think.

skogsmaskin: (but I totally agree!)

packages/@sanity/base/sanity.json
skogsmaskin: Good idea, should probably be done in regard to the refactoring of datastores as I mentioned in your first comment.

PR #1391, opened 2019-06-25
[chore] Upgrade tap/jest/babel/gulp setup to latest versions

rexxars:
We're having tests fail on travis. After looking into it, it seems we're running a range of different versions of jest. The babel setup didn't work as intended with the latest version of babel, so had to upgrade that. This broke tap, so it had to be updated. I then added node 12 to the test matrix, and it turns out that gulp 3 doesn't work on node 12, so had to upgrade that and rewrite the entire gulpfile.

Long story short, this makes sure that we can run tests on travis again.


skogsmaskin:
When I ran `test-all` the two first times it failed on the `state-router` and `observable` pacakge. After running the tests inside their folders, they passed, and then `test-all` passed. Not sure what is going on there.

thomax:
Hm. Done a full clean/install/bootstrap on this branch. Running `npm run test-all` and keep getting
```
@sanity/client: Error: Cannot find module './lib/SanityObservableMinimal'
@sanity/client:     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
```

rexxars:
@thomax Can you run `npx gulp` and re-run the tests, to see if that changes anything?

thomax:
`npx gulp` did the trick. All lights are green!

PR #1402, opened 2019-06-27
[all] History UI fixes

kristofferj:
- Disabled hover on touch-devices
- Smooth overflow scroll on touch devices
- Removed colored border around avatars and make the border transparent with calculated clip-path to show the background color
- Enables as many context menu items as possible in history mode
- Send `_rev` to document preview
- 2px gap between user avatars


skogsmaskin:
Rebased against next, and also fixed some PropTypes and lint errors.

PR #1411, opened 2019-07-04
[preview] Don't toggle visiblity for 'inline' and 'block' previews

skogsmaskin:
There is currently a regression in the way visibility toggling for previews are keeping their height when hidden. This is a quick fix to disable visibility toggling for previews that are either `inline` or `block`. They should only apply to previews inside the PT editor, where the bug is currently really annoying.


skogsmaskin:
Tnx. Added a comment.

PR #1415, opened 2019-07-07
[core] Allow specifying multiple document IDs to delete

rexxars:
This PR adds the ability to delete multiple documents in the same transaction:
`sanity documents delete id1 id2 id3` etc.


rexxars:
Yes - but it's sort of dangerous. I think we should give it a good think-through in terms of how to communicate it. "Are you sure you want to delete 1337 documents from the dataset 'production'"?

kmelve:
Yeah, some failsafe or dry-run should probably be in place.

PR #1416, opened 2019-07-08
Icons in blockeditor

kristofferj:
- Support adding `icon` directly on `blockTypes`, `annotations` and  `decorators` in the same way we add icons on other types
- Resolve annotation icons from types/references in schema


kristofferj:
Agreed. In case of ditching the `blockEditor` key we should throw a warning that it will be deprecated.

kristofferj:
https://github.com/sanity-io/sanity-priv/issues/1578

PR #1419, opened 2019-08-01
[default-layout] Prettier styling of broken avatar image

kristofferj:
In some cases the avatar image is broken, and the `alt` text is no very pretty. Uses a CSS-trick to show the initials of the user in the avatar circle instead.

### DIFF
--- a/packages/@sanity/default-layout/src/components/LoginStatus.js
+++ b/packages/@sanity/default-layout/src/components/LoginStatus.js
@@ -64,7 +64,7 @@ class LoginStatus extends React.PureComponent {
                 src={user.profileImage}
                 className={styles.userImage}
                 alt={`${user.name}'s profile image`}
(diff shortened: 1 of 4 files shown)


packages/@sanity/default-layout/src/components/LoginStatus.js
```diff
@@ -64,6 +64,7 @@ class LoginStatus extends React.PureComponent {
                 src={user.profileImage}
                 className={styles.userImage}
                 alt={`${user.name}'s profile image`}
+                data-initials={user.name ? user.name.charAt(0) : user.email.charAt(0)}
```
bjoerge: Is it possible that `user.email` is missing here?

To be extra safe and future proof this can be written
`(user.name || user.email || '?').charAt(0)`

PR #1427, opened 2019-08-02
Show incoming references

bjoerge:
This needs more work on the visual part, but provides us with building blocks + an example for displaying the documents that have references to the document being edited.

### DIFF
--- a/packages/@sanity/base/sanity.json
+++ b/packages/@sanity/base/sanity.json
@@ -191,6 +191,10 @@
       "description": "A utility component that takes a document id and calls its child render prop with a list of documents referring to it"
     },
     {
+      "name": "part:@sanity/base/listen-query",
+      "description": "A utility function that can be used to listen for a query"
+    },
+    {
       "name": "part:@sanity/base/theme/layout/resets-style",
       "description": "Provides resets for resetting browser styling"
     },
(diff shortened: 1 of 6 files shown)


kmelve:
Just a thought. Should it be possible to configure the title/description for the element? 

For example, let’s say you use references to make a hierarchal content structure. Then it would be useful to be able to contextualize the incoming references as “parent documents”.

readeral:
This is looking rad

readeral:
Hey @bjoerge - anything I can do to help test this/push it along?

I'm going to have a play with your changes in a branch of a project I'm currently working on, but don't really know what's next for this to know how to contribute.

kmelve:
@readeral We're considering creating some new better primitives for stuff like this. I believe the best thing, for now, is to stay tuned :)

EivindGelato:
Would love to have this feature... Any ETA on merge?

readeral:
Just dropping by to say I'm still tuned ;-P

pawelngei:
This could be a really useful feature!

@bjoerge what do you think about creating a Part for the visuals of the IncomingLinks instead of adding it straight to FormView? This way it'd be easier pluggable / configurable.

rexxars:
There's a plugin now that provides a pretty good and customizable implementation of this feature: https://www.sanity.io/plugins/documents-pane

We _might_ revisit this officially later, but for now we'd recommend people use that!

PR #1439, opened 2019-08-13
[chore] Run rename-unsafe-lifecycles codemod on all source files

bjoerge:
This is the result of running the [rename-unsafe-lifecycles codemod](https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles) on all source files. Should remove a bunch of deprecation warnings in studios using React ^16.9.

There's probably still some more places (e.g. libraries) that will trigger this warning, but let's do them in a separate pass.


rexxars:
For anyone who's looking; yes, we will rewrite these to use "safe" lifecycle methods - this is just a temporary stage to reduce the noise ;)

PR #1443, opened 2019-08-13
[chore] Remove visual regression testing for now

bjoerge:
This removes the visual regression testing setup for now. We should prioritize getting it back in at a later point, but as it stands today it doesn't provide much value as the tests are currently failing and seems largely outdated, which also indicates that nobody is really running them on a regular basis. It also creates a lot of delay when initializing a clean checkout of this repo, and when reinstalling node_modules since it triggers a chromium build download which is sometimes incredibly slow.

This PR can be reverted when we figure out a better and more sustainable approach (e.g. a CI environment that can run the tests for us).


bjoerge:
We can either revert commit 9e14760, or go back to the commit/branch where it was originally introduced (in #1157). There's a risk of it diverging from `next`, though. Do you want to keep it in a branch that can regularily be rebased against next?

kristofferj:
Yeah. that would be nice :)

bjoerge:
Restored the branch [https://github.com/sanity-io/sanity/tree/visual-regression-testing](visual-regression-testing)

PR #1450, opened 2019-08-16
[form-builder] Add error boundary around portable text editor

rexxars:
Currently, if something goes wrong during the rendering of the portable text editor, the whole desk tool crashes. Because this happens in the case outlined in #1449, we need something to prevent it from fully crashing so the user can still restore other versions of the document, for instance.

What I ended up with was a very minimal but possibly reusable error boundary that is _currently_ only wrapped around the PT editor, but could potentially be included by default on the FormBuilderInput level (unless it has drawbacks in terms of performance and the like). 

## Screenshot 

image


skogsmaskin:
In the future we might want to build in some reset field button here too.

PR #1465, opened 2019-08-29
Various Portable Text Editor fixes

skogsmaskin:
First off, sorry this PR got a little bigger than first intended. There were so many Flow lint errors, which I just couldn't resist to not to fix. So a lot of changes are that (but marked as individual commits). So if you review, I would suggest you look at the commits and not the total files changed.

Mainly, this PR fixes the following:

Everywhere we rely on selecting children of a text block from a focus or patch path, we must first check that the root block is non-void (as in not a custom object block). This is because a void block in theory could use the same keys as a text block would have . Typical like this if we try to select a child of a contentBlock's `.children`:

```
focusPath[1] === 'children' && focusPath[2] && focusPath[2]._key
```

It needs to be:
```
!isVoidRootBlock && focusPath[1] === 'children' && focusPath[2] && focusPath[2]._key
```
to ensure that it is a text-block's child we are dealing with and not some void-block using the same keys.

Also there was a problem with some unset patches that targeted something inside a custom block or inline's data, which resulted in the containing node being unset, and not just data values inside (https://github.com/sanity-io/sanity/pull/1465/commits/3906e7285d5a25e37cbe84c894051dfb3f46a6b5)

This PR will fix the following issues:

https://github.com/sanity-io/sanity-priv/issues/1582
https://github.com/sanity-io/sanity-priv/issues/1616


skogsmaskin:
Thanks @kristofferj. Yes, if you could just test that the Toolbar is still working as intended (I did some small refactorings in order to make Flow happy). I've tested it myself, and it seems to be working allright, but you have a better overview of the new toolbar.

PR #1472, opened 2019-09-01
[default-layout] improve studio module update indicator

vicbergquist:
Improvements to the way we indicate that the Studio has available updates:
- Font size is now relative instead of fixed to 14px
- Package icon uses a colored circle to indicate severity level, and is labeled for accessibility
- A more accessible approach regardless of severity level and navigation bar background color (like when using custom Studio themes)

Further possible improvements to these changes:
- Position and size of indicator on package icon
- Colors indicating severity level

Examples, current vs. proposed changes:

update indicator old vs proposal

### DIFF
--- a/packages/@sanity/default-layout/src/components/SanityStatus.js
+++ b/packages/@sanity/default-layout/src/components/SanityStatus.js
@@ -45,7 +45,7 @@ export default function SanityStatus(props) {
               <span className={styles.updateIcon}>
                 <div
                   className={styles.updateIndicator}
-                  data-severity={"notice"}
+                  data-severity={severity}
                   aria-label={`${severity} severity level.`} />
                 <PackageIcon />
               </span>
--- a/packages/@sanity/default-layout/src/components/SanityStatus.js
+++ b/packages/@sanity/default-layout/src/components/SanityStatus.js
@@ -24,11 +24,10 @@ export default function SanityStatus(props) {
   } = props
   const currentLevel = outdated.length ? level : 'notice'
   const severity = isSupported ? currentLevel : 'high'
-  const className = `${styles.root} ${styles[severity]}`
   const Dialog = isUpToDate ? CurrentVersionsDialog : UpdateNotifierDialog
 
   return (
-    <div className={className}>
+    <div className={styles.root}>
       {showDialog && (
         <Dialog
           severity={severity}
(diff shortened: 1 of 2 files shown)


vicbergquist:
> Have we tested it on mobile?
@kristofferj Yeah, it doesn't show up. But don't think the current update indicator even shows up on mobile:
Screenshot 2019-09-03 at 12 14 31
!

kristofferj:
👍 Yeah. That comes to mind now :) Good job!

packages/@sanity/default-layout/src/components/styles/SanityStatus.css
```diff
@@ -48,18 +48,42 @@
...
+  border: 1px solid var(--main-navigation-color--inverted);
+  box-sizing: border-box;
+}
+
+.updateIndicator[data-severity="notice"] {
+  background: var(--main-navigation-color--inverted);
+}
+
+.updateIndicator[data-severity="low"] {
+  background: #6cb8ff;
```
mariuslundgard: It would be great if we should get rid of this hard-coded color in this update. And use `background: currentColor` instead.

vicbergquist: What color will current color be? The color of low severity update now is a tone of blue, is there a variable for a blue?

packages/@sanity/default-layout/src/components/SanityStatus.js
```diff
@@ -42,9 +41,16 @@ export default function SanityStatus(props) {
...
             <span>Up to date</span>
           ) : (
-            <span>
-              <PackageIcon /> {formatUpdateLabel(outdated.length)}
-            </span>
+            <div className={styles.hasUpdates}>
+              <span className={styles.updateIcon}>
+                <div
+                  className={styles.updateIndicator}
+                  data-severity={"notice"}
```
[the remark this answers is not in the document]
vicbergquist: oops! Yes!

packages/@sanity/default-layout/src/components/styles/SanityStatus.css
mariuslundgard: I don't think we have a variable for the blue. So we can use `currentColor` which is the same as the text color (https://css-tricks.com/currentcolor/).

vicbergquist: Yes, but then the low status is no different from the notice status?

mariuslundgard: True, but think that's fine – especially since the "notice" status is only visible when it says "Up to date".

packages/@sanity/default-layout/src/components/styles/SanityStatus.css
```diff
@@ -5,7 +5,7 @@
   align-items: center;
 
   @nest & span {
-    font-size: 14px;
+    font-size: var(--font-size-small);
     line-height: 25px;
```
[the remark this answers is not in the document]
vicbergquist: but the small font-size isn't always 14?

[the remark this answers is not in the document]
mariuslundgard: Yeah. That's true.

If the font size has been set to `0.5em` in the overrides, the line height would be interpreted to `21.428px` which is totally fine.

Can't think of a more elegant solution right now, but open to suggestions!

PR #1474, opened 2019-09-04
[core] Deprecate React < 16.9

bjoerge:
This adds a deprecation warning for studios running on a React version < 16.9. Also did a refactor of the checkReactVersions, renaming it to checkStudioVersions, opening for the possibility of adding non-react packages. Also improved the error message/help article a bit.


bjoerge:
Good point! Renamed to `checkStudioDependencyVersions.js`.

bjoerge:
Giving this a bit more though, I changed it from < 16.9 to < 16.8. This will make the deprecation warning affect less users (everyone that has created a new studio after 16.8 was released should already have it installed). (Also, 16.9 hardly introduced any new feature, mostly only deprecation warnings)

PR #1477, opened 2019-09-05
[cli] inform about /create when init new project

vicbergquist:
Inform about /create when running `sanity init` and creating a new project.

```
print('Prefer web interfaces to terminals?')
print('You can also set up best practice Sanity projects with')
print('your favorite frontends on https://sanity.io/create\n')
```

image


mariuslundgard:
Looks good! Could an idea be to make the headline more prominent – perhaps make it white and bold like the "Select project to use" text?

E.g.:

> **Prefer web interfaces to terminals?**
> You can also set up best practice Sanity projects with
> your favorite frontends on https://sanity.io/create

PR #1480, opened 2019-09-06
[form-builder] Block editor: avoid toolbar making drag and scroll up not function

skogsmaskin:
In fullscreen mode.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/BlockEditor/BlockEditor.js
+++ b/packages/@sanity/form-builder/src/inputs/BlockEditor/BlockEditor.js
@@ -1,6 +1,7 @@
 // @flow
 import type {Element as ReactElement, ElementRef} from 'react'
 import React from 'react'
+import {debounce} from 'lodash'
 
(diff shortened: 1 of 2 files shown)


packages/@sanity/form-builder/src/inputs/BlockEditor/BlockEditor.js
```diff
@@ -226,6 +229,14 @@ export default class BlockEditor extends React.PureComponent<Props, State> {
     return null
   }
 
+  handleOnDragEnter = (event: SyntheticDragEvent<>) => {
+    this.setState({isDragging: true})
+  }
+
+  handleOnDragLeave = debounce((event: SyntheticDragEvent<>) => {
```
bjoerge: It would be good to cancel the debounce on componentWillUnmount here. I think it should be a matter of adding this
```js
componentWillUnmount() {
  this.handleOnDragEnter.cancel()
}
```

skogsmaskin: Good catch, fixed.

PR #1489, opened 2019-09-10
[form-builder] Block editor: send setIfMissing patch before trying to do a insert

skogsmaskin:
We must send a setIfMissing here, because the block array could be unset in some circumstances.


bjoerge:
This was cherry-picked into master (176913c) and released yesterday

Issue #1493, opened 2019-09-13
htmlToBlocks loses whitespace between two hyperlinks


felthy:
When using `blockTools.htmlToBlocks()` to parse HTML markup like this:

```html
<p>I should have a space between <a href="#lala">these</a> <a href="#lala">two</a> links.</p>
```

the space between the links is lost, resulting in the words running together:

`I should have a space between thesetwo links.`

**To Reproduce**

Pass the above html through `blockTools.htmlToBlocks()`.

**Expected behavior**

There should be a span with text `" "` between the two links:

```json
      { "_type": "span", "marks": ["randomKey2"], "text": "these" },
      { "_type": "span", "marks": [], "text": " " },
      { "_type": "span", "marks": ["randomKey3"], "text": "two" },
```

I think the easiest way to reproduce might be for me to submit a PR that adds the relevant test to `packages/@sanity/block-tools/test/tests/HtmlDeserializer/whitespace2`.

skogsmaskin:
This is fixed and staged for release.

PR #1494, opened 2019-09-13
Test for #1493 htmlToBlocks loses whitespace between two hyperlinks

felthy:
A test case seemed like the simplest and clearest way to communicate the issue. See other notes in #1493. Test currently fails - on purpose!


skogsmaskin:
Thank you! We'll look into it!

skogsmaskin:
Fix is staged for release. Closing this.

PR #1496, opened 2019-09-13
Bug 1478: [graphql] fix deploy where type derived from string used as reference

relefant:
I think this should fix this bug. If the type derived from a scalar was being used as a field, it was already correctly handled. However, if the type derived from a scalar was being used as a reference or in an array, then it would not be correctly dereferenced.

I've tested as well as I can, but you probably want to check this closely.

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
+++ b/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
@@ -9,6 +9,7 @@ const Schema = schemaCompiler.default || schemaCompiler
 const skipTypes = ['document', 'reference']
 const allowedJsonTypes = ['object', 'array']
 const disallowedCustomizedMembers = ['object', 'array', 'image', 'file', 'block']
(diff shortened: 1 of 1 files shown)


packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
```diff
@@ -9,6 +9,7 @@ const Schema = schemaCompiler.default || schemaCompiler
 const skipTypes = ['document', 'reference']
 const allowedJsonTypes = ['object', 'array']
 const disallowedCustomizedMembers = ['object', 'array', 'image', 'file', 'block']
+const nonScalarTypes = ['document', 'object', 'block', 'reference', 'array']
```
rexxars: I think it's probably better to list the scalar types, as those will be more static?

relefant: Sure, if you give me a list of those types I can update the change.

Issue #1482, opened 2019-09-06
[graphql] API generation fails if schema has self-referencing types


relefant:
API generation fails is schema has self-referencing types.

**To Reproduce**

1. Use a schema with a self referencing type:

e.g.

```js
export default {
  name: 'pageSection',
  title: 'Page Section',
  type: 'document',
  fields: [
    {
      name: 'content',
      title: 'Content',
      type: 'array',
      of: [{ type: 'reference', to: [{ type: 'pageSection' }, { type: 'simpleContent' }] }],
    },
  ],
};
```
2. Run `sanity graphql deploy`

Schema generation goes into an infinite loop in `extractFromSanitySchema.js` 
Fails with error: `RangeError: Maximum call stack size exceeded`

**Expected behavior**

(Maybe) GraphQL schema should be generated successfully. I'm not a GraphQL expert and am not 100% sure a schema with such recursive references is possible.

**Which versions of Sanity are you using?**
```
@sanity/cli                0.142.2 (up to date)
@sanity/base               0.142.2 (up to date)
@sanity/components         0.142.4 (up to date)
@sanity/core               0.142.4 (up to date)
@sanity/default-layout     0.142.4 (up to date)
@sanity/default-login      0.142.3 (up to date)
@sanity/desk-tool          0.142.4 (up to date)
@sanity/google-maps-input  0.142.4 (up to date)
```
**What operating system are you using?**

MacOS 10.14.6

**Which versions of Node.js / npm are you running?**

npm 6.10.1
node v10.16.0

rexxars:
This has now been fixed 🎉

tbaustin:
I'm getting this same error when trying to deploy.

```
{
	name: `navigationLink`,
	title: `Navigation Link`,
	type: `object`,
	fields: [
		{ name: `title`, type: `string`, title: `Title` },
		{ name: `subTitle`, type: `string`, title: `Sub Title` },
		{ name: `url`, type: `string`, title: `Url` },
		{ 
			name: `images`, 
			type: `array`, 
			title: `Image/Icon(s)`,
			of: [{ type: `navIcon` }],
		},
		{ 
			title: `Sub Links`, 
			name: `subLinks`, 
			type: `array`,
			of: [{ type: `navigationLink` }], 
		},
	],
}
```

devinhalladay:
Hate to bump a closed issue, but I'm still getting this error years later :(

rneatherway:
[...]

PR #1497, opened 2019-09-13
[graphql] fix API generation fails if schema self-referencing types #1482

relefant:
Bug occurs when a schema has a parent type containing an array of references which reference a union containing the parent type.

In generating the union type, the parent type is recursed into - and this ends in an infinite loop.

My fix is to put a recursion guard in `getUnionDefinition()` which checks for the same parent type occuring during recursion, and if this happens then bail out of the recursion. This should not affect the generation of the union type itself, and the recursion path is calculating child types which are not required for the union type generation in any case.

You probably want to test this code.


kmelve:
Thanks for this! We will review it after the weekend 👍

rexxars:
Thanks a bunch for this, very appreciated!

I discovered that the fix only worked for one level of depth - if you have a Post which can reference a Post/Category, and Category can reference a Post/Author, you would still get the recursion error.

Added a small change which uses an array as the guard and pushes/splices from the guard array when it completes one level. Does that look right to you?

Sorry about the force-pushing to your branch, we had some major changes done to our repo and I needed to rebase your fix.

relefant:
Looks good to me. I neglected the case of more than one union in the hierarchy ;)
How do you test the schema generation?

rexxars:
Unfortunately we don't yet have tests for the generation of these schemas. They are however verified on the server side. I assume you got the CLI running in development mode and was able to deploy?

PR #1505, opened 2019-09-18
[test-studio] Make test with custom markers

skogsmaskin:
Just a test editor in the test-studio which shows custom markers.


kmelve:
I don't think we have gotten around to documenting those btw.

PR #1507, opened 2019-09-20
[client] Add client.getDocuments

bjoerge:
This adds a client method for fetching multiple _entire_ documents in one go, using the `/doc` endpoint.

Usage: `client.getDocuments(['foo', 'bar']) => [{_id: 'foo', ...}, {_id: 'bar', ...}]`

Note: This adds a workaround for an issue with the `/doc` HTTP endpoint not yielding `null` values for nonexsting documents in the resulting array. This can lead to nasty bugs, e.g.:

`const [draft, published] = await client.getDocuments([draftId, publishedId])`

Without the workaround, this will assign the published document to the `draft` variable if the draft does not exists.

### DIFF
--- a/packages/@sanity/client/README.md
+++ b/packages/@sanity/client/README.md
@@ -32,10 +32,18 @@ const client = sanityClient({
 
 Initializes a new Sanity Client. Required options are `projectId` and `dataset`.
 
-### Fetch a single document
+### Fetch an entire single document
 
 ```js
 client.getDocument('bike-123').then(bike => {
+  console.log(`${bike.name} (${bike.seats} seats)`)
+})
+```
+
+### Fetch multiple entire documents in one go
+
+```js
+client.getDocuments(['bike123', 'bike345']).then(([bike123, bike345]) => {
   console.log(`${bike.name} (${bike.seats} seats)`)
 })
 ```
--- a/packages/@sanity/client/README.md
+++ b/packages/@sanity/client/README.md
@@ -32,22 +32,6 @@ const client = sanityClient({
 
 Initializes a new Sanity Client. Required options are `projectId` and `dataset`.
 
-### Fetch an entire single document
-
-```js
-client.getDocument('bike-123').then(bike => {
-  console.log(`${bike.name} (${bike.seats} seats)`)
-})
-```
-
-### Fetch multiple entire documents in one go
-
-```js
-client.getDocuments(['bike123', 'bike345']).then(([bike123, bike345]) => {
-  console.log(`${bike.name} (${bike.seats} seats)`)
-})
-```
-
 ### Performing queries
 
 ```js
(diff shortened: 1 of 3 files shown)


packages/@sanity/client/README.md
```diff
@@ -32,14 +32,22 @@ const client = sanityClient({
...
 client.getDocument('bike-123').then(bike => {
   console.log(`${bike.name} (${bike.seats} seats)`)
 })
 ```
 
+### Fetch multiple entire documents in one go
+
+```js
+client.getDocuments(['bike123', 'bike345']).then(([bike123, bike345]) => {
+  console.log(`${bike.name} (${bike.seats} seats)`)
```
rexxars: The `bike` variable is not defined

packages/@sanity/client/README.md
```diff
@@ -32,14 +32,22 @@ const client = sanityClient({
...
+})
+```
```
rexxars: We should probably document that missing documents are returned as `null`-values, and that they are ordered in the same order as the input IDs (I'm used to thinking in SQL `SELECT * FROM documents WHERE id IN (a, b, c)`)

packages/@sanity/client/test/client.test.js
```diff
@@ -363,6 +363,43 @@ test('can query for single document', t => {
...
+test('preserves the position of requested documents', t => {
+  nock(projectHost())
+    .get('/v1/data/doc/foo/abc123,abc321')
+    .reply(200, {
+      ms: 123,
+      documents: [{_id: 'abc321', mood: 'tense'}]
+    })
+
+  getClient()
+    .getDocuments(['abc123', 'abc321'])
```
rexxars: Perhaps add a third document (which also exist, but is returned in the wrong order), so as to make sure we treat both missing and present documents correctly?

packages/@sanity/client/README.md
```diff
@@ -32,14 +32,22 @@ const client = sanityClient({
 
 Initializes a new Sanity Client. Required options are `projectId` and `dataset`.
 
-### Fetch a single document
+### Fetch an entire single document
```
rexxars: Why "entire"? What would be the alternative?

packages/@sanity/client/README.md
bjoerge: Good point! Guess I assumed that the order was preserved apart from the omitted null values.

packages/@sanity/client/README.md
bjoerge: The alternative would be a _partial_ document. Maybe "complete", "full" is better? Or just leave it as is?

packages/@sanity/client/README.md
bjoerge: Good catch, thanks!

packages/@sanity/client/README.md
bjoerge: Noted this in the README now, linking to the official doc endpoint docs. Also moved the `getDocument(s)` sections from the top of the README to below the _queries_ section to make it less prominent/significant in the README.

PR #1511, opened 2019-09-23
add check for prior __block

ajonp:
This prevents program from crashing when an image is not contained within a block, but parent is __block

image


skogsmaskin:
Thank you for this @ajonp! I would like to add a test here, do you have some piece of html that makes this crash?

ajonp:
@skogsmaskin I have markdown that I run through https://github.com/ajonp/markdown-to-sanity if that would work?

skogsmaskin:
@ajonp - do you convert that to HTML first, if so, could you just output the HTML it produces and give it to me?

skogsmaskin:
@ajonp - I've managed to write a test that failed and passed with your patch. Thank you!

PR #1517, opened 2019-09-27
[form-builder] Pass accepted mime types to input

rexxars:
For file and image inputs, we seem to somewhere in the stack reject files that do not match the configured `accept` value, but we don't actually pass the accept value to the underlying file input.

This is confusing, as you'll be able to select files of any type, but they are silently ignored if they do not match. It also creates a less than ideal selection of files, especially on mobile devices.

This PR passes the configured accept value down to the file input, and also defaults to `image/*` for the image input - this will make mobile devices (at least Android) use a different mode.


bjoerge:
Oh, wow, great catch! I was certain it was getting forwarded already. Thanks!

PR #1524, opened 2019-10-02
[components] Add dropdown menu for initial value templates

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Screenshot 2019-10-02 at 12 11 49

**Description**

Adds a dropdown menu to the "+" (new) button in a pane if initial value templates exist and styled according to the sketches.

Screenshot 2019-10-01 at 21 32 03

**Note for release**

- Added dropdown menu in panes that lets you create new documents from existing templates

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/components/src/panes/DefaultPane.js
+++ b/packages/@sanity/components/src/panes/DefaultPane.js
@@ -1,15 +1,17 @@
 import PropTypes from 'prop-types'
 import React from 'react'
-import {negate} from 'lodash'
+import {negate, get} from 'lodash'
 import shallowEquals from 'shallow-equals'
 import classNames from 'classnames'
-import S from '@sanity/base/structure-builder'
 import Menu from 'part:@sanity/components/menus/default'
 import IconMoreVert from 'part:@sanity/base/more-vert-icon'
 import {IntentLink} from 'part:@sanity/base/router'
 import ScrollContainer from 'part:@sanity/components/utilities/scroll-container'
+import DropDownButton from 'part:@sanity/components/buttons/dropdown'
+import DefaultPreview from 'part:@sanity/components/previews/default'
 import Styleable from '../utilities/Styleable'
 import defaultStyles from './styles/DefaultPane.css'
+import S from '@sanity/base/structure-builder'
 
 function getActionKey(action, index) {
   return (typeof action.action === 'string' ? action.action + action.title : action.title) || index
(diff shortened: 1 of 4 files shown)


packages/@sanity/desk-tool/src/pane/DocumentsListPane.js
```diff
@@ -267,72 +267,6 @@ export default withRouterHOC(
       return `*[${filter}] | order(${toOrderClause(sort)}) [0...50000] {${finalProjection}}`
     }
 
-    renderNewDocumentButton() {
```
rexxars: Hmm, `renderNewDocumentButton` is still used in `render()`?

packages/@sanity/desk-tool/src/pane/DocumentsListPane.js
```diff
@@ -16,12 +14,14 @@ import {collate, getPublishedId} from 'part:@sanity/base/util/draft-utils'
...
 import {combineLatest} from 'rxjs'
+import {map, tap} from 'rxjs/operators'
 import settings from '../settings'
 import styles from './styles/DocumentsListPane.css'
 import listStyles from './styles/ListView.css'
 import InfiniteList from './InfiniteList'
 import PaneItem from './PaneItem'
-import {map, tap} from 'rxjs/operators'
+import S from '@sanity/base/structure-builder'
+import {getTemplateById} from '@sanity/base/initial-value-templates'
```
rexxars: Are these still in use?

packages/@sanity/components/src/panes/DefaultPane.js
```diff
@@ -295,41 +297,78 @@ class Pane extends React.Component {
...
-          className={styles.actionButton}
-          data-menu-button-id={this.templateMenuId}
-          type="button"
-          title={action.title}
-          // eslint-disable-next-line react/jsx-no-bind
-          onClick={this.handleMenuAction.bind(this, action)}
-        >
-          <div className={styles.actionButtonInner} tabIndex={-1}>
-            <Icon />
+        {action.action != 'toggleTemplateSelectionMenu' && (
```
rexxars: Coding standard says to use triple equals :)

packages/@sanity/components/src/panes/DefaultPane.js
```diff
@@ -295,41 +297,78 @@ class Pane extends React.Component {
...
   }
 
+  renderActionMenuItem = item => {
+    const {styles} = this.props
+    if (!item) {
+      return null
+    }
+    const params = item.intent.params
+    const schemaType = S.documentTypeListItem(params.type).spec.schemaType
+    const icon = schemaType.icon
```
rexxars: This should preferably use the icon declared on the template (which falls back to the schema types icon, if not explicitly provided). 

I'm thinking it's perhaps best to do the resolving of this icon in the structure builders `S.menuItemsFromInitialValueTemplateItems()` method - replacing the `add` icon for the template icon, and falling back to the add icon?

```js
return new MenuItemBuilder()
  .title(title)
  .icon((tpl && tpl.icon) || getPlusIcon())
```

packages/@sanity/components/src/panes/DefaultPane.js
```diff
@@ -295,41 +297,78 @@ class Pane extends React.Component {
...
+      <IntentLink
+        className={styles.initialValueTemplateDropDownItem}
+        intent="create"
+        params={params}
+      >
+        <div>
+          <div>{item.title}</div>
+          <div className={styles.initialValueTemplateSubtitle}>{params.type}</div>
+        </div>
+        <div className={styles.initialValueTemplateDropDownItemIcon}>{icon()}</div>
```
rexxars: Since icons can be classes as well as stateless function components, it's best to use the `<Icon />` form.

packages/@sanity/components/src/panes/DefaultPane.js
vicbergquist: OK, thank you! That makes a lot more sense. 

I did this, but now the icon of the second template is always undefined. How come? Shouldn't the fallback always be the icon of the schemaType the template is from, which is author? Or is there some logic somewhere that looks for an icon based on the `author-developer` template name?

image

This seems to give the right icon for both, but I'm guessing the problem is somewhere higher up in the chain and I can't find it.

```
const schema = getDefaultSchema().get(tpl.schemaType)
 return new MenuItemBuilder()
      .title(title)
      .icon((tpl && tpl.icon) || schema && schema.icon || getPlusIcon())
```

rexxars: Hmm, good question. I think we might have chosen not to fall back on the schema type icon so as not to repeat the same icon many times in a row.

I'm a bit unsure whether or not we should change that, or leave it up to the implementation to use the schema type icon 🤔

vicbergquist: OK. Where would it be repeated?
Should I get the icon like in the code in my previous comment?

rexxars: In the global "create new document" modal, I think we made a choice not to repeat the icon. 

Yeah, I think for now you can just use the code you outlined, and we can consider changing/cleaning it up when we have more time.

rexxars:
I'm going to merge this in and we can do any changes that might be necessary there, going to see if I can get a tagged release out for testing.

PR #1525, opened 2019-10-02
[desk-tool] Style broken references errors and warnings

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

64277456-3acce680-ceff-11e9-854a-91ec0734f92b

**Description**

Styled error dialogs that appear when document template references unpublished or missing documents.

65834536-36b49e80-e2dc-11e9-9ad3-e8225dda5a0c
65834537-36b49e80-e2dc-11e9-87e7-fdf49550bba5

**Comments**

Some color changes may still be coming, depending on what @mariuslundgard says.

**Note for release**

- Styled error dialogs that appear when document template references unpublished or missing documents

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/components/src/dialogs/PanePopover.js
+++ b/packages/@sanity/components/src/dialogs/PanePopover.js
@@ -1,0 +1,75 @@
+/* eslint-disable complexity */
+import React from 'react'
+import PropTypes from 'prop-types'
+import CheckCircleIcon from 'part:@sanity/base/circle-check-icon'
+import WarningIcon from 'part:@sanity/base/warning-icon'
+import ErrorIcon from 'part:@sanity/base/error-icon'
+import InfoIcon from 'part:@sanity/base/info-icon'
+import classNames from 'classnames'
+import styles from './styles/PanePopover.css'
+
+export default class PanePopover extends React.PureComponent {
+  static propTypes = {
+    children: PropTypes.node,
+    icon: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.bool]),
+    kind: PropTypes.oneOf(['info', 'warning', 'error', 'success', 'neutral']),
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
+  }
+
+  static defaultProps = {
+    children: null,
+    icon: null,
+    kind: 'info',
+    title: 'Pane Popover',
+    subtitle: 'Popover content'
+  }
+
+  DEFAULT_ICONS = {
+    info: <InfoIcon />,
+    success: <CheckCircleIcon />,
+    warning: <WarningIcon />,
+    error: <WarningIcon />
+  }
+
+  snackIcon = () => {
+    const {icon, kind} = this.props
+    if (typeof icon === 'boolean' && icon) return this.DEFAULT_ICONS[kind]
+    if (typeof icon === 'object' || typeof icon === 'string') return icon
+    return undefined
+  }
+
+  render() {
+    const {children, icon, id, kind, title, subtitle} = this.props
+
+    const role = () => {
+      if (kind === 'success') return 'status'
+      if (kind === 'info') return 'log'
+      return 'alert'
+    }
+    return (
+      <div
+        aria-label={kind}
+        aria-describedby={`snackbarTitle-${kind}-${id}`}
+        className={classNames([styles.root, styles.dialog])}
+        data-kind={kind}
+      >
+        <div className={styles.inner}>
+          <div className={styles.content}>
+            <div id={`snackbarTitle-${kind}-${id}`} className={styles.title}>
+              {icon && (
+                <div role="img" aria-hidden className={styles.icon}>
+                  {this.snackIcon()}
+                </div>
+              )}
+              {title}
+            </div>
+            {subtitle && <div className={styles.subtitle}>{subtitle}</div>}
+            {children && <div className={styles.children}>{children}</div>}
+          </div>
+        </div>
+      </div>
+    )
+  }
+}
--- a/packages/@sanity/components/src/dialogs/PanePopover.js
+++ b/packages/@sanity/components/src/dialogs/PanePopover.js
@@ -1,9 +1,7 @@
-/* eslint-disable complexity */
 import React from 'react'
 import PropTypes from 'prop-types'
 import CheckCircleIcon from 'part:@sanity/base/circle-check-icon'
 import WarningIcon from 'part:@sanity/base/warning-icon'
-import ErrorIcon from 'part:@sanity/base/error-icon'
 import InfoIcon from 'part:@sanity/base/info-icon'
 import classNames from 'classnames'
 import styles from './styles/PanePopover.css'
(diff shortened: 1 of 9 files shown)


vicbergquist:
I see that there are problems with the `ReferringDocumentsList` component. I mistook it for being specific to `BrokenReferences`. Working on a fix 👩‍🔧

packages/@sanity/desk-tool/src/components/BrokenReferences.js
```diff
@@ -7,12 +7,15 @@ import {uniq, uniqBy} from 'lodash'
 import DefaultPane from 'part:@sanity/components/panes/default'
 import {observePaths} from 'part:@sanity/base/preview'
 import {getDraftId, getPublishedId} from 'part:@sanity/base/util/draft-utils'
+import FormBuilder from 'part:@sanity/form-builder'
+import PanePopover from 'part:@sanity/components/dialogs/pane-popover'
+import styles from './styles/BrokenReferences.css'
 import ReferringDocumentsList from './ReferringDocumentsList'
 
 // @todo get some designers to pretty this up (and reword it)
```
rexxars: This can obviously be removed now 😅

packages/@sanity/desk-tool/src/components/BrokenReferences.js
```diff
@@ -22,50 +25,62 @@ function BrokenRefs(props) {
...
+            subtitle={`A document can only refer to published documents. Publish the following ${
+              unpublished.length === 1 ? 'draft' : 'drafts'
+            } before creating
+            a new document.`}
+          >
+            <ReferringDocumentsList
+              documents={unpublished.map(({id, type, hasDraft}) => ({
+                _id: `drafts.${id}`,
+                _type: type,
+                _hasDraft: hasDraft
```
rexxars: A bit of a nitpick, but I feel like the `_hasDraft` property here is a little dirty - given it's not actually a part of the document. Not sure what the best approach is. Maybe send an array of `drafts` (ids) along with `documents`? 
Or just unprefix the properties (`_id` => `id`) and map them back in the `ReferringDocumentsList` component?
Maybe @bjoerge has some better ideas.

packages/@sanity/desk-tool/src/components/BrokenReferences.js
```diff
@@ -22,50 +25,62 @@ function BrokenRefs(props) {
...
+              }))}
+            />
+          </PanePopover>
+        )}
+      </div>
+      <form className={styles.editor}>
+        <FormBuilder readOnly type={schemaType} schema={schema} />
+      </form>
```
rexxars: I need @bjoerge to approve this, I'm not sure whether there are side effects of doing something like this that we should consider.

packages/@sanity/desk-tool/src/components/BrokenReferences.js
```diff
@@ -90,7 +105,13 @@ const BrokenReferences = streamingComponent(props$ =>
         filter(docs => docs.length === ids.length),
         map(docs => docs.filter(isMissingPublished)),
         map(broken =>
-          broken.length > 0 ? <BrokenRefs documents={broken}>eek</BrokenRefs> : props.children
+          broken.length > 0 ? (
+            <BrokenRefs documents={broken} type={type} schema={schema}>
+              eek
```
rexxars: Should remove the children here (`eek`) since it's not actually used. Just leftovers from my testing - sorry about that.

packages/@sanity/components/src/dialogs/PanePopover.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const role = () => {
+      if (kind === 'success') return 'status'
+      if (kind === 'info') return 'log'
+      return 'alert'
+    }
+    return (
+      <div
+        aria-label={kind}
+        aria-describedby={`snackbarTitle-${kind}-${id}`}
```
rexxars: Is it still a snackbar?

packages/@sanity/components/src/dialogs/PanePopover.js
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable complexity */
```
rexxars: It doesn't appear to me that this component is over the complexity limit, so this can probably be removed?

packages/@sanity/components/src/dialogs/PanePopover.js
```diff
@@ -0,0 +1,75 @@
...
+    subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
+  }
+
+  static defaultProps = {
+    children: null,
+    icon: null,
+    kind: 'info',
+    title: 'Pane Popover',
+    subtitle: 'Popover content'
```
rexxars: This will make subtitle always be present - I think it should either not have a default subtitle or the check for the subtitle on render can be removed

packages/@sanity/components/src/dialogs/PanePopover.js
```diff
@@ -0,0 +1,75 @@
...
+    info: <InfoIcon />,
+    success: <CheckCircleIcon />,
+    warning: <WarningIcon />,
+    error: <WarningIcon />
+  }
+
+  snackIcon = () => {
+    const {icon, kind} = this.props
+    if (typeof icon === 'boolean' && icon) return this.DEFAULT_ICONS[kind]
+    if (typeof icon === 'object' || typeof icon === 'string') return icon
```
rexxars: What is this case intended for? Strings are icons? Object?
I would personally just make the `DEFAULT_ICONS` be the non-rendered versions, eg `info: InfoIcon`.

packages/@sanity/components/src/dialogs/PanePopover.js
```diff
@@ -0,0 +1,75 @@
...
+        data-kind={kind}
+      >
+        <div className={styles.inner}>
+          <div className={styles.content}>
+            <div id={`snackbarTitle-${kind}-${id}`} className={styles.title}>
+              {icon && (
+                <div role="img" aria-hidden className={styles.icon}>
+                  {this.snackIcon()}
+                </div>
+              )}
```
rexxars: I feel like the `snackIcon` function (maybe not the most suitable name?) and this `icon` check kind of collides. Perhaps it's better to assign the return method of  `snackIcon` to a variable and then render that if it returned anything?
Eg:
```js
const Icon = this.snackIcon()
// ...
{
  icon && (<div><Icon /></div>)
}
```

packages/@sanity/desk-tool/src/components/BrokenReferences.js
vicbergquist: I'm guessing you mean the comments 😄

packages/@sanity/components/src/dialogs/PanePopover.js
vicbergquist: Oops, you're right

packages/@sanity/desk-tool/src/components/BrokenReferences.js
vicbergquist: Ok! Tried to follow what was already in there, but would love some more input. I think `ReferringDocumentsList` would become more flexible if whether to show a status badge or not could be baked in. Was unsure how to determine if a document was a draft by looking at other preview components and their badges.

packages/@sanity/components/src/dialogs/PanePopover.js
vicbergquist: Yes, you're right!

packages/@sanity/components/src/dialogs/PanePopover.js
vicbergquist: The initial idea was for the snackbar, that strings and icons were both valid icon types depending on what someone would want to use in their snackbar. I guess this doesn't apply here!

packages/@sanity/components/src/dialogs/PanePopover.js
vicbergquist: Not entirely sure what you mean with this and the other comment, but will make some changes!

packages/@sanity/desk-tool/src/components/BrokenReferences.js
bjoerge: The only thing I can think of is focus handling and validation, etc. It will probably not be able to navigate through this form (e.g. open array items), but not sure if that's a problem here?

vicbergquist: I don't think any interaction with the form should be allowed here, seeing as the errors need to be resolved first before one can continue with the editing

bjoerge: This should be fine, then 👍

packages/@sanity/desk-tool/src/components/BrokenReferences.js
rexxars: I don't think is critical - we're not exposing it to other people, so we can always clean this up later.

PR #1530, opened 2019-10-04
[cli] Add a default dataset config

vicbergquist:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

Currently the CLI always prompts for setting a dataset name and choosing a dataset visibility mode

**Description**

This adds a simple setup mode that skips dataset name and visibility mode prompt and instead use the default values of `production` and `public`. This mode can also be used with a `--simple` flag.

cli-simple-mode

With `--simple` flag:

cli-simple-flag

**Note for release**

- Added a simple configuration mode to `sanity init` that doesn't prompt for dataset name and visibility mode, but instead uses default name and mode (production/public).

**Checklist** 

- [x]  The code is linted

**Notes/Comments**

I wonder if I did this right 😅

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -71,10 +71,17 @@ if (path.win32 === path) {
   libFragment = '$1/lib/'
 }
 
-const logCompile = from => (file, enc, cb) => {
-  gutil.log(`[${from}] Compiling`, `'${chalk.cyan(file._path || file.path)}'...`)
-  cb(null, file)
-}
+const pkgPath = (cwd, sourcePath) => path.relative(path.join(cwd, 'packages'), sourcePath)
+
+const log = prefix =>
+  through.obj((file, enc, cb) => {
+    if (process.env.VERBOSE) {
+      gutil.log(prefix, `'${chalk.green(pkgPath(file.cwd, file._path || file.path))}'...`)
+    }
+    cb(null, file)
+  })
+
+const logCompile = from => log(`[${from}] Compiling `)
 
 const mapToDefinitionPath = baseDir => {
   const packagesRoot = path.join(__dirname, 'packages')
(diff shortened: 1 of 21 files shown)


vicbergquist:
> I think `--simple` is underspecified. It's not obvious to me what "simple" means - what will I miss out on when choosing it? What is the name of my dataset? Are all my data available to the public?
> 
> I guess it does save us from (trying/failing to) explain what a dataset _is_ during setup, but I'm not sure hiding a core concept is the right choice, either.
> 
> These questions are really something @evenwestvang should answer and not me, though.
> 
> From a technical perspective, the `-s` alias doesn't exist and should be reserved for `--silent` if we ever add that. This also breaks the unattended mode, we need to wrap the simple prompt check in `if (!unattended)`.

Yes, that's true. I've gone ahead and renamed it to `default` and provided an info comment.

cli-mode-default-config

evenwestvang:
Would be cool if we could tie the question unequivocally to the proceeding explanation – does this help:

> Dataset default configuration is a world readable (public) with the name `production`
Use the default config? [y]

vicbergquist:
@rexxars I've made some bigger changes as I noticed some flaws, and missing spinners from (#1531), and hopefully improved the logic of prompting for a default dataset configuration, taking into the dataset and visibility flag. Still unsure about the unattended mode...

I also reordered the prompts, because I think it makes more sense to get the project name first. I gathered the dataset info in the default config prompt. If someone is using the dataset and/or the visibility flag, should we assume they don't need extra info about datasets? That's how it's like now and I think it works.

@evenwestvang ok! How about this wording?

cli-fixed-new
cli-fixed-new-no

rexxars:
I'm tempted to say the flag should be plural? `--defaults` or `--use-defaults`? Thoughts?

vicbergquist:
> I'm tempted to say the flag should be plural? `--defaults` or `--use-defaults`? Thoughts?

Hm... well, defaults feel more like it’s more defaults than just the dataset config.

But maybe it should be more specific than just default? `—default-dataset’ maybe

vicbergquist:
> Oh, I was actually picturing the flag selecting many defaults. If it's just the defaults for the dataset, is it really needed? If so, I'd say `--default-dataset`

Ok! I've renamed it to `--dataset-default` to align it with the existing dataset flag for consistency

PR #1532, opened 2019-10-07
[cli] install vision in CLI templates by default

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No 

Maybe?

**Current behavior**

When creating a new project from a template in the CLI, the vision plugin is not installed by default.

**Description**

Makes the vision plugin be installed in templates by default

**Note for release**

- When creating a project from the CLI using a template, the vision plugin is now installed by default.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/createManifest.js
+++ b/packages/@sanity/cli/src/actions/init-project/createManifest.js
@@ -120,7 +120,8 @@ export function createSanityManifest(data, opts) {
         '@sanity/components',
         '@sanity/default-layout',
         '@sanity/default-login',
-        '@sanity/desk-tool'
+        '@sanity/desk-tool',
+        '@sanity/vision'
       ],
 
       parts: [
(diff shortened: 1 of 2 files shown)


vicbergquist:
> It also needs to be enabled in sanity.json: https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/cli/src/actions/init-project/createManifest.js#L118-L124

done!

> Maybe we should also consider only enabling it in development? Or do we feel it should also be shipped to editors? Thoughts?
Yeah, that's a good idea! No need to clutter it for editors, I guess.

rexxars:
If we want it to just be in development, I think this should do it (in the `sanity.json` manifest creation step):
```
"env": {
  "development": {
    "plugins": [
      "@sanity/vision"
    ]
  }
}
```

evenwestvang:
👍 and good call @rexxars on making it dev only.

vicbergquist:
@rexxars ok, thanks!

PR #1543, opened 2019-10-11
[components] add dev server status snackbar

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When the dev server disconnects, there is no way of knowing this happened by looking at the Studio in the browser. Can be confusing, if changes are being made and you don't understand why the Studio isn't updating.

**Description**

This adds a snackbar to show that the dev server was disconnected and needs to be restarted to see any changes made to the Studio.

**Note for release**

- If your development server is disconnected/stops working, the Studio will now show you a notification to let you know that you should restart with `sanity start` to see your latest changes.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/base/src/components/DevServerStatus.js
+++ b/packages/@sanity/base/src/components/DevServerStatus.js
@@ -7,15 +7,16 @@ const EventSource = isWindowEventSource
   ? window.EventSource // Native browser EventSource
   : polyfilledEventSource // Node.js, IE, Edge etc
 
-const STATE_CLOSED = 0
+const STATE_CONNECTING = 0
 const STATE_OPEN = 1
+const STATE_CLOSED = 2
 
 class DevServerStatus extends PureComponent {
   constructor(...args) {
     super(...args)
     this.enabled = __DEV__ && EventSource
     this.state = {
-      connectionState: STATE_CLOSED,
+      connectionState: STATE_OPEN,
       hasHadConnection: false
     }
   }
@@ -65,8 +66,11 @@ class DevServerStatus extends PureComponent {
       return null
     }
 
-    // We are disconnected, show snackbar
-    if (this.state.connectionState === STATE_CLOSED) {
+    // We are disconnected
+    if (
+      this.state.connectionState === STATE_CLOSED ||
+      this.state.connectionState === STATE_CONNECTING
+    ) {
       return (
         <Snackbar
           id="__dev-server-status"
@@ -76,8 +80,10 @@ class DevServerStatus extends PureComponent {
           title={<strong>Disconnected from the dev server!</strong>}
           subtitle={
             <div>
-              To see your latest changes, restart the Studio with <code>sanity start</code> from
-              your project folder.
+              {this.state.connectionState === STATE_CLOSED
+                ? 'To see your latest changes,'
+                : 'We are trying to reconnect you, but you can also'}{' '}
+              restart the Studio with <code>sanity start</code> in your project folder.
             </div>
           }
         />
--- a/packages/@sanity/base/src/components/DevServerStatus.js
+++ b/packages/@sanity/base/src/components/DevServerStatus.js
@@ -16,7 +16,7 @@ class DevServerStatus extends PureComponent {
     super(...args)
     this.enabled = __DEV__ && EventSource
     this.state = {
-      connectionState: STATE_OPEN,
+      connectionState: STATE_CONNECTING,
       hasHadConnection: false
     }
   }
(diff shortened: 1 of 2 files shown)


packages/@sanity/base/src/components/DevServerStatus.js
```diff
@@ -0,0 +1,90 @@
...
+import polyfilledEventSource from '@sanity/eventsource'
+import Snackbar from 'part:@sanity/components/snackbar/default'
+
+const isWindowEventSource = Boolean(typeof window !== 'undefined' && window.EventSource)
+const EventSource = isWindowEventSource
+  ? window.EventSource // Native browser EventSource
+  : polyfilledEventSource // Node.js, IE, Edge etc
+
+const STATE_CLOSED = 0
+const STATE_OPEN = 1
```
rexxars: `0` actually means it's connecting (EventSource will automatically try to reconnect when disconnected), so the constant should reflect this. `2` means it's closed, which can happen in some cases (browser gives up reconnecting after closing laptop, or similar).

packages/@sanity/base/src/components/DevServerStatus.js
```diff
@@ -0,0 +1,90 @@
...
+      return null
+    }
+
+    // We're connected, don't show anything
+    if (this.state.connectionState === STATE_OPEN) {
+      return null
+    }
+
+    // We are disconnected, show snackbar
+    if (this.state.connectionState === STATE_CLOSED) {
```
rexxars: Should probably check for STATE_CLOSED || STATE_CONNECTING once that is added

packages/@sanity/base/src/components/DevServerStatus.js
vicbergquist: OK! So snackbar should be displayed when it's reconnecting and closed.

vicbergquist:
> Nice! That's some very clean and readable code @vicbergquist!
> 
> @bjoerge proposed creating a separate dev server endpoint for this purpose, so as not to depend on any webpack logic at all. But since @rexxars suggested this, let's go! It certainly results in less code (no additional serverside code) and that's plus.

Kudos to Espen, he wrote it 😄

packages/@sanity/base/src/components/DevServerStatus.js
vicbergquist: I had this and removed it, because the snackbar was never called...

packages/@sanity/base/src/components/DevServerStatus.js
rexxars: Correct

packages/@sanity/base/src/components/DevServerStatus.js
rexxars: It's hard to reproduce, I'm not 100% sure when the browser eventually gives up. But both reconnecting and closed are states when it should be displayed.

rexxars:
> @bjoerge proposed creating a separate dev server endpoint for this purpose, so as not to depend on any webpack logic at all. But since @rexxars suggested this, let's go! It certainly results in less code (no additional serverside code) and that's plus.

I sort of agree, but I think it's slightly too much work to factor this in. We're not using any webpack logic here apart from the hot endpoint existing. If it changes, we'll update both `@sanity/core` and `@sanity/base` at the same time and we should be fine.

packages/@sanity/base/src/components/DevServerStatus.js
vicbergquist: OK. I've tried leaving for a while after disconnecting, but it still doesn't seem to give up. Results in over 100 errors in the console 🙃

packages/@sanity/base/src/components/DevServerStatus.js
```diff
@@ -7,15 +7,16 @@ const EventSource = isWindowEventSource
...
 const STATE_OPEN = 1
+const STATE_CLOSED = 2
 
 class DevServerStatus extends PureComponent {
   constructor(...args) {
     super(...args)
     this.enabled = __DEV__ && EventSource
     this.state = {
-      connectionState: STATE_CLOSED,
+      connectionState: STATE_OPEN,
```
rexxars: Was this intentional? Until the eventsource is actually open, it should probably stay as closed? :)
If this is to prevent an initial "we are disconnected", I think we should use the `hasHadConnection` flag instead:

Only show the disconnected message if we have previously been connected and we are now in closed or reconnecting state.

vicbergquist: It's to prevent the snackbar showing for a split second when the Studio starts! :/

rexxars: I think my hasHadConnection proposal would do the trick, then?

vicbergquist: Alright! Will test tomorrow, thank you! 😃

PR #1544, opened 2019-10-13
[default-layout] fix brand logo sizing

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When using a custom brand logo component that returns an image or SVG, the sizing is botched.

**Description**

This makes images and SVGs display correctly in terms of size and how they take up space. Works when returning an img and svgs  in a custom brand logo component, as well as anything else.

**Note for release**

- Improved styling and sizing when using a custom logo

**Checklist** 

- [x]  The code is linted

**Other comments**

There seem to be a chrome glitch when it comes to SVGs and the height property. The SVG won't fill the space as it should if the logo is long (compared to round or squared), but when you toggle on and off the height in dev tools, it works as expected. Not sure how this will be in production, but the glitchy behaviour is the same as the current behaviour so it shouldn't have a big impact.

### DIFF
--- a/packages/@sanity/base/src/styles/variables/globals.css
+++ b/packages/@sanity/base/src/styles/variables/globals.css
@@ -62,5 +62,6 @@
   --block-extras-background-color: #eee;
 
   /* Others */
-  --pane-header-height: 3.4375rem; /* 55px */
+  --pane-header-height: 3.4375rem;
+  --nav-bar-box: 3.0625em; /* 55px */
 }
--- a/packages/@sanity/default-layout/src/components/styles/Branding.css
+++ b/packages/@sanity/default-layout/src/components/styles/Branding.css
@@ -1,50 +1,52 @@
 @import "part:@sanity/base/theme/variables-style";
 
 .root {
+  display: flex;
   text-decoration: none;
   color: inherit;
   font-weight: 400;
-  width: 100%;
+  height: 100%;
   outline: none;
-  padding: 1em;
+  padding: var(--extra-small-padding) var(--medium-padding);
+  padding-right: var(--extra-small-padding);
   box-sizing: border-box;
   position: relative;
+  max-height: var(--nav-bar-box);
+}
 
-  @media (--screen-medium) {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-  }
+.projectNameContainer {
+  display: flex;
+  align-items: center;
 }
 
 .projectName {
-  display: block;
   color: inherit;
   font-size: inherit;
   margin: 0;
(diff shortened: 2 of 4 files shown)


kristofferj:
Cool! Is this tested against different SVG's, text-only and different studios?

vicbergquist:
> Cool! Is this tested against different SVG's, text-only and different studios?

Tested against SVG of various sizes, using project name only. Tested only in the test-studio.
Still baffled by that SVG height issue tho.

packages/@sanity/default-layout/src/components/styles/Branding.css
```diff
@@ -1,50 +1,52 @@
...
-  display: block;
+  display: flex;
+  align-items: center;
   cursor: pointer;
   color: inherit;
+  flex-grow: 1;
 
-  @nest & svg {
-    display: block;
-    fill: currentColor;
```
kristofferj: Was it intended to remove the direct styling on the SVG element?

vicbergquist: Yes, it messes up the styling of some SVGs. SVG styling should be left to the actual SVG or person implementing the logo, IMO.

kristofferj: Agreed 👍 This could however change behaviour for some. Let's add a comment in the release.

vicbergquist:
> If tested properly and we add a comment in the release this is good AFAICT. Nice work! 👍

I've tested it some more today. Seems like if an SVG is returned, a width need to be set in order to grow properly (and not be tiny). Don't know why it suddenly works without it when toggling height in dev tools, but probably because the SVG is then aware of the width of its parent.

Width of the SVG can be set with `width` on the SVG or wrapping it in a container with a width, but tested it with large values, but it doesn't overflow and stays the max width 👌 So yeah, this is ready. Let me know if you think otherwise.

packages/@sanity/base/src/styles/variables/globals.css
```diff
@@ -62,5 +62,6 @@
   --block-extras-background-color: #eee;
 
   /* Others */
-  --pane-header-height: 3.4375rem; /* 55px */
+  --pane-header-height: 3.4375rem;
+  --nav-bar-box: 3.0625em; /* 55px */
```
mariuslundgard: The comment `/* 55px */` should go on line above. This value equal `49px`.

skogsmaskin:
Restored the branch, we will revert this because it needs to be investigated.

PR #1550, opened 2019-10-15
[groq] Shorten groq to its bare minimum

boyum:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

`groq` does nothing to the input string, which is exactly the same as `String.raw` does.

**Description**

Not many bytes are saved, and the discussion in #1109 might suggest that there are plans to extend `groq`'s capabilities, however until then, perhaps shortening it as much as possible is a viable idea.

**Note for release**

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/groq/src/groq.js
+++ b/packages/groq/src/groq.js
@@ -1,8 +1,2 @@
 /* eslint-disable import/no-commonjs */
-module.exports = function groq(strings, ...keys) {
-  const lastIndex = strings.length - 1
-  return (
-    strings.slice(0, lastIndex).reduce((acc, str, i) => acc + str + keys[i], '') +
-    strings[lastIndex]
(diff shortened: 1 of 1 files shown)


rexxars:
Thanks for this, but `String.raw` doesn't process newline characters, which could potentially be used in some cases. For instance:

```js
function groq(strings, ...keys) {
  const lastIndex = strings.length - 1
  return (
    strings.slice(0, lastIndex).reduce((acc, str, i) => acc + str + keys[i], '') +
    strings[lastIndex]
  )
}

console.log(groq`*\n[0]`)
// Result:
// *
// [0]

console.log(String.raw`*\n[0]`)
// Result:
// *\n[0]
```

PR #1553, opened 2019-10-16
Normalize decorators

skogsmaskin:
The block normalization function should normalize span marks and remove those who aren't allowed by the schema. This includes both decorator marks and annotation marks.

In order to differentiate between decorator marks and annotation marks, we need to know which decorator names are allowed in the schema.

Fixes https://github.com/sanity-io/sanity-priv/issues/1632

### DIFF
--- a/packages/@sanity/block-tools/README.md
+++ b/packages/@sanity/block-tools/README.md
@@ -164,7 +164,7 @@ const partialBlock = {
     }
   ]
 }
-normalizeBlock(partialBlock, {decorators: ['strong']})
+normalizeBlock(partialBlock, {alowedDecorators: ['strong']})
 ```
 Will produce
 ```
--- a/packages/@sanity/block-tools/README.md
+++ b/packages/@sanity/block-tools/README.md
@@ -149,7 +149,7 @@ blockTools.htmlToBlocks(
 
 ```
 
-### ``normalizeBlock(block, [decoratorMarks])``
+### ``normalizeBlock(block, [options={}])``
 Normalize a block object structure to make sure it has what it needs.
 
 ```js
@@ -164,7 +164,7 @@ const partialBlock = {
     }
   ]
 }
-normalizeBlock(partialBlock, ['strong'])
+normalizeBlock(partialBlock, {decorators: ['strong']})
 ```
 Will produce
 ```
--- a/packages/@sanity/block-tools/README.md
+++ b/packages/@sanity/block-tools/README.md
@@ -149,7 +149,7 @@ blockTools.htmlToBlocks(
 
 ```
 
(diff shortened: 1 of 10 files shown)


packages/@sanity/block-tools/README.md
```diff
@@ -159,11 +159,12 @@ const partialBlock = {
...
     {
       _type: 'span',
-      text: 'Foobar'
+      text: 'Foobar',
+      marks: ['strong', 'df324e2qwe']
     }
   ]
 }
-normalizeBlock(partialBlock)
+normalizeBlock(partialBlock, {decorators: ['strong']})
```
rexxars: If I understand this correctly, this option should actually be named `allowedDecorators` - right now it's a bit hard to know what this option does.

packages/@sanity/block-tools/src/util/normalizeBlock.ts
```diff
@@ -30,6 +30,11 @@ export default function normalizeBlock(block) {
     return block
   }
   let usedMarkDefs = []
+  const allowedDecorators =
+    options.decorators &&
+    Array.isArray(options.decorators) &&
+    options.decorators.length > 0 &&
```
rexxars: If you have a block where no decorators are allowed, doesn't this check prevent you from stripping disallowed decorators?

packages/@sanity/block-tools/src/util/normalizeBlock.ts
```diff
@@ -53,6 +58,12 @@ export default function normalizeBlock(block) {
       if (child._type === 'span' && !child.marks) {
         child.marks = []
       }
+      if (allowedDecorators && child._type === 'span') {
+        child.marks = child.marks.filter(
+          mark =>
+            allowedDecorators.includes(mark) || block.markDefs.map(def => def._key).includes(mark)
```
rexxars: The `map().includes()` check could be written slightly more performant by using  `.some()` or `.find()` - exiting once a match is encountered

packages/@sanity/block-tools/src/util/normalizeBlock.ts
skogsmaskin: Very good point!

packages/@sanity/block-tools/README.md
skogsmaskin: Yeah, I agree, that's a more sound name.

packages/@sanity/block-tools/src/util/normalizeBlock.ts
skogsmaskin: True!

skogsmaskin:
Thanks for the review @rexxars, addressed your (valid) points!

PR #1560, opened 2019-10-22
[block-tools] Deserialize HTML text nodes directly on the body better

skogsmaskin:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**
When using `htmlToBlocks` in script context (i.e. with JSDOM), text nodes which are a child of the body, was ignored.

This fix will make sure text is wrapped in container tags (spans), before they are processed further. 

In a browser context, the browser makes sure this is done before putting the HTML on the clipboard.

**Note for release**
Fixed a bug in the HTML deserializer where text nodes directly on the body was not handled correctly.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/block-tools/src/HtmlDeserializer/preprocessors/html.ts
+++ b/packages/@sanity/block-tools/src/HtmlDeserializer/preprocessors/html.ts
@@ -1,20 +1,40 @@
 import {_XPathResult} from './index'
-
+// Remove this cruft from the document
+const unwantedWordDocumentPaths = [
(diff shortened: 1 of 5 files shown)


skogsmaskin:
> Looks good to me!
> 
> Is the span node considered the "correct" way of patching up html like this? As opposed to div or p?

Yes, because you could have: `<body>Hey <span>there!</span></body>`, then you would not like to have "Hey" in it's own block.

PR #1562, opened 2019-10-23
[form-builder] prevent slate-react built in scrolling

skogsmaskin:
There is a very annoying bug in the version of `slate-react` we are using that tries to scroll to focus in invalid situations. Typically where focus is set programatically. By aborting the `onFocus` plugin chain the built in onFocus functions in `slate-react` doesn't kick in.

This was recently fixed in `slate-react`: https://github.com/ianstormtaylor/slate/pull/3019 and we should upgrade the Slate version when time.

I'm still not 100% certain where we are in the plugin chain by setting onFocus as prop on the component, as opposed to the `onFocus` plugins we are running ourselves  *This needs to be checked up upon before merging.*

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

**Description**

**Note for release**

Fix portable text editor scroll bug where the editor scrolled to top after acquiring focus.

**Checklist** 

- [x ]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [x]  The code is linted

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/BlockEditor/plugins/OnFocusPlugin.ts
+++ b/packages/@sanity/form-builder/src/inputs/BlockEditor/plugins/OnFocusPlugin.ts
@@ -1,13 +1,22 @@
 // This plugin runs when the editor receives focus
 // Ensures that an empty block is inserted when the editor gets focus and it's empty.
 
+// It also prevents default onFocus actions in slate-react being triggered as we
+// abort the plugin chain. The version of slate-react we are currently using (0.21.16)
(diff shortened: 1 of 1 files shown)


skogsmaskin:
I've checked up on things. Moved the canceling of the plugin chain in our own `onFocus` plugin, which needs to run first.

The built in after chain `onFocus` plugin does this: https://github.com/ianstormtaylor/slate/pull/2396/files#diff-6982ba4f0afdc505d31b674b30e6d6f1R387

As far as I can tell, this is not relevant for us and I can't see any side effects by not doing the `.deselect()`.

Issue #1520, opened 2019-10-01
Feature request / discussion: Added extensibility for browsing and selecting images


robinpyon:
**Is your feature request related to a problem? Please describe.**
Sanity's image browser could use a lot of love! This is for a number of reasons, including but not limited to:
- no view or filtering options (e.g. tabular vs grid views, ability to filter by format, size etc)
- no options for batch editing / deletion
- no batch uploading (or even single uploading - you need to exit the dialog first!)
- inability to browse or organise images in a way that suits the needs / workflow of the organisation

I think the last point is the most salient of all – everyone has a different view of what makes an ideal image browser. Some orgs may want a higher density of content in a tabular / paginated form whilst some others may want to see larger images with more visible metadata, or something else entirely different.

I believe the flexibility of the structure builder is one of Sanity's key features and it'd be great to see that philosophy carried over to how assets are managed too.

In my specific use case, editors need to edit old entries containing an array of any number of objects (containing images) and add an image from that object array to another field. Copying and pasting those images also isn't an option as we're dealing with more complex objects (and not just an array of images), and it's not something that can be automated as it requires manual curation.

More often than not, editors are just re-uploading files (running the risk of adding duplicates) as it's faster than having to browse for them in a dataset containing thousands of images. So in this organisation's case, the ability to quickly see linked assets in the *current* document is important - though this isn't something that may be relevant to many others.

**Describe the solution you'd like**
One proposal:
- The ability to specify a custom `AssetSelect` (or equivalent) component that can be specified in the options object of an `image` field, similar to how custom input components currently work.
- A generic `Asset` *part* which can be easily imported in a custom `AssetSelect` implementation if users don't want to create their own by hand

I don't think this is necessarily the most thought-out solution, but it was one that allowed me to quickly get the following (super rough) POC up:

![Kapture 2019-10-01 at 1 50 16](https://user-images.githubusercontent.com/209129/65926609-02310780-e3ee-11e9-8bed-9489e03218ff.gif)

The above uses a fork of Sanity that allows a custom `AssetSelect` component to be specified in an image field's `options` object. This custom component has a few buttons which toggle between all images, unused images – and specific to my use case – images referenced by the current document. 

A nice feature here could be the ability to search for a specific post and then display all images linked therein. That particular workflow feels intuitive to me ("find me all images from XYZ post") rather than searching for file names or other metadata criteria - which going by experience, only works if editors are maniacally consistent in file naming and tagging, which isn't always the case!

[...]

PR #1580, opened 2019-10-30
[form-builder] Discussion: decouple SelectAsset and create a separate asset browser package (#1520)

robinpyon:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

---

### Current behavior

See #1520

### Description

This is a *draft proposal* of one way extensibility could be added to Sanity's image browser. 

### Example video: 

https://streamable.com/oiq99

### Example studio with setup instructions:

https://github.com/robinpyon/sanity-asset-browser-studio
 
### Summary

- All previous `SelectAsset` logic has been moved out of `form-builder/inputs/ImageInput` and into it's own `@sanity/asset-browser` package
- This package uses redux to manage state and rxjs to manage requests and batch / buffer multiple snackbar displays.
- Two new parts are exposed, providing access to asset browser state and actions:
	- `part:@sanity/asset-browser/context/dispatch`
	- `part:@sanity/asset-browser/context/state`
- This package is opinionated in that:
	- it wraps both dialogs and snackbar notifications, assuming you don't want to bother re-inventing the wheel in custom implementations
	- it takes care of all Sanity's `client` calls, namely fetching and deleting
	- it allows you to send custom GROQ queries, but you're limited to simpler projections (e.g. you can't use named outer return fields / combine multiple unrelated queries). 

#### Features

- Keeps current `SelectAsset` unchanged (except you now get snackbar notifications when assets are deleted)
- Opt-in for custom components, in a similar way to how custom input components work
- Exposes asset browser state + actions via contexts and avoids prop-drilling
- Multiple asset selection / deletion
- Manages common state and actions that you *probably* don't want to rewrite in your custom components (fetching & deleting via the client, keeping track of selected / updating items, showing snackbars and dialogs etc)

#### Defining a custom asset browser

In your document schema:

```
import MyComponent from './MyComponent'
...
{
  title: 'Image',
  name: 'image',
  type: 'image',
  assetSelectComponent: MyComponent,
},
```

#### Customising the browser

`asset-browser` exposes two contexts (as parts) which can be imported into any custom browser component:

1. `part:@sanity/asset-browser/context/state`
Exposes asset browser state:
- whether assets are being fetched (`fetching`)
- whether there's been an error fetching (`fetchingError`)
- an array of all items (`items`) - each item also keeping track of it's picked / updating / error status.
- the *total* number of assets  (`totalCount`) - and not just those returned with the current query.

2. `part:@sanity/asset-browser/context/dispatch`
Exposes dispatchable actions:
- delete an asset (`onDelete`)
- delete all picked assets (`onDeletePicked`)
- show an error dialog displaying conflicts / references (`onDialogShowConflicts`)
- show a dialog displaying references (`onDialogShowRefs`)
- fetch assets (`onFetch`), with the option to replace all existing items (paginated) or to append (infinite scroll)
- pick an asset (`onPick`)
- pick all assets (`onPickAll`)
- clear all picked assets (`onPickClear`)
- select / insert an asset (`onSelect`)

_(The action 'pick' is used instead of 'select', as that word is reserved for the action of inserting an image into an entry)._

#### Future goals
- Allow for the asset browser to be promoted to a *tool*, always accessible in the header
- Support for file uploads
- Filtering by entry search
- Be able to read and respect custom `accepts` options defined on image fields (https://www.sanity.io/docs/image-type#accept-785f76a21c7c). Side note: how would you turn something like `accept="image/*,.pdf"` into a GROQ query?
- (Long term) Being able to select individual / multiple assets and assign custom tags / metadata / organise them into folders

#### Other notes
- Deleting in batches currently send out multiple requests (E.g. deleting 100 items calls `client.delete` 100 times) which is far from ideal. However, currently batching deletes in a transaction only yields the first encountered error, so not sure of a way around this.
- The `Preview` component used in dialogs to render entry previews was promoted to a part (`part:@sanity/form-builder/preview`) so it could be used across packages. However, as it depends on the `formBuilder` context which isn't available when `asset-browser` is instantiated outside a form input (e.g. as a tool), perhaps another approach is in order.
- The 'Complex' example in the example studio is wrapped with the `withDocument` HOC in order to get the current document passed as a prop (so we can use it to do things like search for images linked in the current document). `withDocument` seems to make assumptions about usage however (only composing React classes), so I'm sure this isn't the most idiomatic way to go..
- Everything remains wrapped in a `<FullscreenDialog>` though I'd argue that such usage could fall outside the boundaries of what a 'dialog' is, especially in the 'Complex' example which relies on sticky positioning and doesn't utilise dialog headers / actions.

#### Final notes
Despite all this, I'm on the fence as to whether similar customisation should be offered, as once given it's very difficult to take back - I also question what sort of usage it'd receive.

At the very least, I'd be stoked if this led to an improved *base* asset browsing experience in Sanity, directly or otherwise.

---

**Checklist** 

- [x]  The code is linted


evenwestvang:
Thank you @robinpyon – judging from the video this is more or less totally awesome.

Grsmto:
What happened to this? Looks like a promising addition to the current simplistic media library!
Is that it? https://github.com/robinpyon/sanity-plugin-media

Thanks!

PR #1583, opened 2019-10-31
[base, schema] Added support for exact number searching

tomsseisums:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

If numeric type is exposed to search (via `preview` or `__experimental_search`), there is no possibility to search for it, because by default, the search uses `match` operator, which works on strings only.

**Description**

🔢 This PR adds partial support for exact number searches.

It is done by adding extra `type` parameter to `__experimental_search` specification which is then consumed by `createWeightedSearch` to change the search operator and provide the search term value accordingly.

Partial, because the type is also added to default search configuration (base and previews and strings), but it's currently hardcoded to string in all scenarios, to be backwards compatible with previous solution. Because otherwise, type inferring has to be setup.

⛲️ The design of the PR is such that there is already a foundation to support other types too.

📈 Currently the operator is inferred from the provided type, but the feature could be extended to also provide user-land configurable operator to use for the field.

🔬 It would also be possible to infer type from schema, but that's too much effort for now to support properly - so I have skipped it for now, adding a TODO pointer instead. This should be also built in for search setup that's being inferred from preview setup.

🧹 Search conditions and terms have been heavily sanitized so that they support string searching together with number. Otherwise GROQ error got thrown: `No function match() defined for arguments (float, string)`. That happened in cases when term for number would resolve to `null` in case a non-numeric term got provided.

I need this (and actually a lot more, but that appears to be impossible for now due to GROQ) for a customer project, that have LOADS of imported/generated documents, that are mostly referenced by numbers. Browsing for them is helluva slow, when one could simply search for the necessary entity by number/code. The number type is necessary for other filtering, so transforming the field to a string was not an option. This implementation, though, acts as a good enough MVP.

**Note for release**

- Added exact search support for numeric fields via `type: number` specification
  - Document with:  
    ```
    __experimental_search: [{ weight: 5, path: 'numberField', type: 'number' }]
    ```
  - Will allow searches for:  
    Exact number now matches
  - INEXACT NUMBER, THOUGH, WON'T BE FOUND:  
    Inexact number still don't match

**Checklist** 

- [ ]  The code is linted

### DIFF
--- a/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
+++ b/packages/@sanity/base/src/search/weighted/createWeightedSearch.js
@@ -1,32 +1,72 @@
 import {map} from 'rxjs/operators'
-import {joinPath} from '../../util/searchUtils'
+import {joinPath, getMatchingOperatorForType, getTypedTermValue, getUsedTermTypes} from '../../util/searchUtils'
 import {compact, toLower, flatten, uniq, flow, sortBy, union} from 'lodash'
 import {removeDupes} from '../../util/draftUtils'
 import {applyWeights} from './applyWeights'
 import {tokenize} from '../common/tokenize'
 
 const combinePaths = flow([flatten, union, compact])
+const combineTermTypes = flow([flatten, uniq])
 
-const toGroqParams = terms =>
+// "t" is short for term
+const typedTerm = (term, type) => `t${term}${type.charAt(0)}`
+
+const toGroqParams = (terms, termTypes) =>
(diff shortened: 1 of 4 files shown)


rexxars:
We're hitting the limits of the current studio search implementation, so it needs to be refactored a bit. Will make sure we address this once we do the refactoring - sorry for keeping this open for so long.

tomsseisums:
Lovely, this was a pure hack implementation anyways, and I've since found better workarounds.

PR #1588, opened 2019-11-05
[core] Deploy: use passed SOURCE_DIR even when building

rexxars:
This fixes two issues:

1. `sanity deploy graphql` vs `sanity graphql deploy`. I can't see someone actually wanting to deploy a `graphql` directory, so I decided to just error out when specifying that.
2. `sanity deploy somedir` wouldn't actually use `somedir` unless `--no-build` was passed. Very confusing.


rexxars:
That's already in the PR, Knut ;)

PR #1591, opened 2019-11-06
[form-builder] Asset Source Plugin support

skogsmaskin:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

**Description**
## Introducing Asset Source Plugins

This PR will make support for various asset sources and DAM functionality through plugins when using an `image` type in Sanity.

### Adding sources (plugins)
You can add more asset-sources by installing plugins that implements `part:@sanity/form-builder/input/image/asset-source`:

```
    {
      "name": "part:@sanity/form-builder/input/image/asset-source-my-dam",
      "implements": "part:@sanity/form-builder/input/image/asset-source",
      "path": "./parts/myDAM.js"
    }
```

This will make the "Select" button for the image field a dropdown (when there are more than one source):

image

### Defining a set of sources globally
This will override every image field to only support a specified set of sources. In this case I only want `MyDAM`:
```
    {
      "implements": "part:@sanity/form-builder/input/image/asset-sources",
      "path": "./parts/assetSources.js"
    }
```
`./parts/assetSources.js`:
```
import MyDAM from 'part:@sanity/form-builder/input/image/asset-source-my-dam'
export default [MyDAM]
```

### Overriding a specific field

You can customize sources a single field in the schema via the `options.sources` property:
```
{
  name: 'bannerImage',
  title: 'Banner image',
  type: 'image',
  options: {sources: [MyDAM]}
}
```
Now the select button will be a normal button again, as we only have one source.

### The part

A source part is something that exports an object like this:
```
export default {
  name: 'my-dam',
  title: 'My Digital Asset Manager',
  component: MyDAM,
  icon: Icon
}
```
Where the component must implement:
```
    onSelect: (..args) => {this.props.onSelect([Asset, ...])},
    onClose: () => {this.props.onClose()}
```

Props sent to the component:
```
    onSelect: PropTypes.func.isRequired,
    onClose: PropTypes.func.isRequired,
    selectedAssets: PropTypes.arrayOf(PropTypes.shape({originalFilename: PropTypes.string, ...})), // Array of Asset documents
    selectionType: PropTypes.string.isRequired // 'single' or 'multiple'
```

An `Asset` returned for `this.props.onSelect` has the following shape:

```
 [
  {
    kind: 'assetDocumentId' | 'file' | 'base64' | 'url'
    value: string | File
    assetDocumentProps?: {originalFilename?: string, label?: string, source?: string, sourceId?: string} 
  }
]
```
We make sure to upload anything that is returned via our normal image pipeline (returning an URL will actually upload it!). Except kind `assetDocumentId` - which signals that the image is already uploaded. A single image input field would use the first entry of the returned asset array.

### Disabling direct uploads
Direct uploads can be disabled with a new default Sanity config file for the form builder (`form-builder.json`) which defaults to true, but can be disabled like this:
```
{
  "images": {
    "directUploads": false
  }
}
```

A little demo for the Cloudinary plugin I made while making this: https://imgur.com/a/lwBU9sX

**Note for release**

**Checklist** 

- [x]  The code is linted


skogsmaskin:
I think this is good to go now, seems like there was a but where `label` for the asset was not respected when uploading it with the @sanity/client (@rexxars , could you have a look at this change: https://github.com/sanity-io/sanity/pull/1591/commits/290319fbb032d8e488ca90c39a575cd786fd70cf)

skogsmaskin:
We can use the `label` for passing through info to connect the files externally.

skogsmaskin:
Removed the client patch, as it is working as is (thanks @rexxars )

skogsmaskin:
@bjoerge - what do you think of the introduced config file for the form builder to disable direct uploads? It currently looks like this:

```
{
  "images": {
    "directUploads": true
  }
}
```
For context we need some way of disabling that the user can upload images on their own, if the studio should only allow assets from some other source.

PR #1593, opened 2019-11-09
[components] Spinner: Allow passing children instead of message

rexxars:
The `Spinner` component currently blindly renders `message`, but it does not accept anything but a string. I have a case where I want to render a message with `<strong>Some message</strong>`. Let me know if this is "bad behavior".

### DIFF
--- a/packages/@sanity/components/src/loading/Spinner.js
+++ b/packages/@sanity/components/src/loading/Spinner.js
@@ -10,15 +10,17 @@ export default class Spinner extends React.PureComponent {
     message: PropTypes.string,
     fullscreen: PropTypes.bool,
     center: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


kristofferj:
I think that should be fine. We could maybe add support for children and keep the message as string?  It is always nice to have a pure string for accessibility/aria purposes if we need that it in the future.

rexxars:
> I think that should be fine. We could maybe add support for children and keep the message as string? It is always nice to have a pure string for accessibility/aria purposes if we need that it in the future.

Good point. Would `children` take presedence over `message`, or be in addition?

kristofferj:
Yes, that was my thought. Children could be optional, like  `children || message`

PR #1595, opened 2019-11-09
[server] Implement custom HMR client for in-studio build status

rexxars:
I liked the "you are disconnected" dev server status message so much that I wanted to also allow it to show  messages based on the current build status. When you save a file,  there is currently no indication that the server is building a new bundle until it's completed. Also, sometimes it cannot hot-reload the changes and you need to open up the dev console to actually see what's going on.

Unfortunately, the events I needed to subscribe to were not exposed by the webpack hot middleware client, so I've reimplemented the client and exposed a thin "eventbus" on the window when in development mode, which you can use to subscribe to events.

With this change, the dev server status will show a "Building..."-spinner in a snackbar item while it's working, which is removed when the build completes. If the patch cannot be applied and the user needs a full reload of the studio, we also show a message letting the user know about this (potentially we could auto-reload, but that may not always be what you want).

### DIFF
--- a/packages/@sanity/base/package.json
+++ b/packages/@sanity/base/package.json
@@ -29,7 +29,6 @@
   "dependencies": {
     "@sanity/client": "0.144.0",
     "@sanity/document-store": "0.144.0",
-    "@sanity/eventsource": "0.144.0",
     "@sanity/image-url": "0.140.15",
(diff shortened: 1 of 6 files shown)


evenwestvang:
Great! Prompting reloads is especially <3.

PR #1597, opened 2019-11-11
[core] Include GraphQL commands in CLI help output

rexxars:
Since the GraphQL API is in open beta, it doesn't make sense to hide the commands from the help output anymore.


saasen:
👍

jorngeorg:
👍

PR #1598, opened 2019-11-12
[groq] Don't use arrow functions - fixes IE11 compatibility

rexxars:
We should probably add a specific  babel config here soon, but for now this should do the trick.


saasen:
👍

PR #1599, opened 2019-11-13
[initial-value-templates] Allow resolving initial value using observable

rexxars:
We currently allow promises to be used when resolving initial values, but using observables allows us to do things like provide a stream of updates about what the asynchronous process is up to. 

This PR implements support for observable for just that reason.

In theory it could deliver a stream of updated values, but I'm not sure it's a good idea - it would at the very least have to report back to the resolving machinery that the initial value is no longer needed once it has been persisted to a draft.


rexxars:
This had a few issues I didn't account for, and we have not heard any requests for it yet, so closing this for now.

Issue #859, opened 2018-06-11
Inject configuration through environment variables


saasen:
Hi.

We are hosting the Sanity studio ourselves in Kubernetes. We have different environments we want to support. However, when docker images are built, `sanity.json` is locked and can't be changed. We are currently using the `dataset` as a form for different environments. To solve this we are using [Kubernetes ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/). The problem with this is that you would constantly need to think about duplicating the `sanity.json` somewhere whenever it changes (it rarely does though).

My proposal is that when self-hosting it should be able to override the `sanity.json` / use environment variables to define the `dataset`. This way we wouldn't need to use the `ConfigMap` feature. I can't think of any other configuration that one would want to change based on environment, but right now it's `dataset` that we would want to be injected. Maybe someone would want the `projectId` as well, I don't know.

I don't have any good suggestions for how this would be implemented, but one solution could be to make environment variables take precedence over configuration in `sanity.json`.

rexxars:
Currently we are building the `sanity.json` configuration into the JS bundle, so it's the environment variables available whenever you do `sanity deploy` or `sanity build` that will be in use.

Judging by what you write, you've already built a studio and want to configure it through environment variables when _serving_ it? That would require a change where we move the API configuration into runtime, which might not be the worst idea, to be honest. Have I understood your requirements correctly?

saasen:
I think so!

Given this `Dockerfile`:
```
FROM ubuntu:latest
RUN \
    apt-get -y update &&\
    apt-get -y install curl nginx gnupg &&\
    curl -sL https://deb.nodesource.com/setup_9.x -o nodesource_setup.sh &&\
    chmod +x nodesource_setup.sh &&\
    ./nodesource_setup.sh &&\
    apt-get -y install nodejs &&\
    rm -rf /var/lib/apt/lists/*

RUN \
    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - &&\
    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list &&\
    apt-get -y update &&\
    apt-get -y install yarn &&\
    rm -rf /var/lib/apt/lists/*

RUN npm install -g @sanity/cli

COPY . /app
WORKDIR /app

RUN \
    cp nginx/site.conf /etc/nginx/conf.d/site.conf &&\
[...]

PR #1610, opened 2019-11-21
[all] Allow setting studio environment variables

rexxars:
# Environment variable support

## Overriding project ID / dataset

There are many cases where you might want to override the Sanity project ID and dataset which is defined in `sanity.json`. Common use cases is having a different dataset for development/staging/production, building the same studio for many different project IDs in a CI-workflow etc. The common way people want this to work is to defined environment variables.

With this PR, this is now possible by defining `SANITY_STUDIO_API_DATASET` and `SANITY_STUDIO_API_PROJECT_ID`. 

## Passing variables to the JS-environment

We currently have no way of passing environment variables to the javascript bundle. For instance, an agency might want to provide the same base schema and studio for many different clients, but alter both the project ID / dataset as well as things inside the bundle. 

With this PR, all environment variables starting with `SANITY_STUDIO_` is exposed to the Webpack bundle environment, so you can do things like `SANITY_STUDIO_CLIENT_NAME="Acme Inc"` and use that inside your own plugins/schemas/studio code as `process.env.SANITY_STUDIO_CLIENT_NAME`.

## Environment-specific configurations

In many cases you might want to define "static" environment variables for different environments in a `.env`-file alongside the studio. For instance, you could use a different dataset name for development and production by setting  `SANITY_STUDIO_API_DATASET=development` in a `.env.development` file.

With this PR, these environment files are now automatically loaded on studio startup (and on `sanity build`/`sanity deploy`). The environment name is fetched from the `NODE_ENV` environment variable - if it has not been set, it defaults to `development` when running `sanity start` or `sanity exec`, and `production` when using `sanity build`/`sanity deploy`.

I'm a little unsure as to whether we should automatically use `production` for `build`/`deploy` or if we should just fall back to development. Thoughts welcome.

## Related issues

_Partially_ solves #859 (this PR doesn't allow injecting configuration _post-build_).

### DIFF
--- a/packages/@sanity/cli/src/CommandRunner.js
+++ b/packages/@sanity/cli/src/CommandRunner.js
@@ -65,7 +65,9 @@ export default class CommandRunner {
     debug(`Reading "${manifestPath}"`)
 
     const baseManifest = await loadJson(manifestPath)
-    const manifest = reduceConfig(baseManifest || {}, environment)
+    const manifest = reduceConfig(baseManifest || {}, environment, {
+      studioRootPath: options.workDir
+    })
     const apiClient = clientWrapper(manifest, manifestPath)
 
     const context = {
--- a/packages/@sanity/cli/src/CommandRunner.js
+++ b/packages/@sanity/cli/src/CommandRunner.js
@@ -16,7 +16,7 @@ import {
 } from './util/generateCommandsDocumentation'
 
 /* eslint-disable no-process-env */
-const sanityEnv = process.env.SANITY_ENV
+const sanityEnv = process.env.SANITY_INTERNAL_ENV
 const environment = sanityEnv ? sanityEnv : process.env.NODE_ENV
 /* eslint-enable no-process-env */
 
--- a/packages/@sanity/cli/src/actions/init-project/initProject.js
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.js
@@ -692,7 +692,9 @@ async function doDatasetImport(options) {
   const {outputPath, coreCommands, template, datasetName, context} = options
   const manifestPath = path.join(outputPath, 'sanity.json')
   const baseManifest = await loadJson(manifestPath)
-  const manifest = reduceConfig(baseManifest || {}, environment)
+  const manifest = reduceConfig(baseManifest || {}, environment, {
+    studioRootPath: outputPath
+  })
 
   const importCmd = coreCommands.find(cmd => cmd.name === 'import' && cmd.group === 'dataset')
   return importCmd.action(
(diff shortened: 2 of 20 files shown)


skogsmaskin:
Could you please rebase against next @rexxars?

rexxars:
> Could you please rebase against next @rexxars? Had to to some force-pushing again this morning to get the release out.

Sure. Rebased.

jorngeorg:
This is brilliant! I just want to remind ourselves that we should make a guide for how to use this, and that we will need to update at least the following docs:
https://www.sanity.io/docs/sanity-json
https://www.sanity.io/docs/deployment

Also wondering if it's worth running this past any of the larger client that have been asking about this and ask if this is the thing they need / and or if they have any other suggestions.

bjoerge:
> I'm a little unsure as to whether we should automatically use production for build/deploy or if we should just fall back to development. Thoughts welcome.

Does it make sense to use `development` as default for build, and `production` as default for deploy?

rexxars:
> This is great! I only have a bit of a reservation about the `SANITY_STUDIO_`-prefix and the potential for unintentionally exposing sensitive data.
> 
> If you've got environment variables named `SANITY_STUDIO_BASE_PATH` or `SANITY_STUDIO_DEPLOY_KEY`, these would now end up in the bundle, right?

Correct. Gatsby and create react app both use this approach, but I suppose they might have done so since launch and thus it's less of a _change_ per say.

> Would it be better to either explicitly whitelist the environment variables you want to go into the bundle somehow? (e.g. though a separate dotfile or similar). Or perhaps change the prefix to something with less potential for accidents? `PUBLIC_BUNDLE_*`, `JS_BUNDLE_ENV_*`, …?.

I  suppose we could make an even clearer prefix if people agree with you on this - but I'd rather not introduce even more complexity in the form of additional files, if I can help it. 

> Does it make sense to use development as default for build, and production as default for deploy?

`sanity build` is basically 1:1 with `sanity deploy` except for the actual deployment - I don't see why `build` would not use the same as  `deploy`.

bjoerge:
> Correct. Gatsby and create react app both use this approach, but I suppose they might have done so since launch and thus it's less of a change per say.

Wasn't aware that this is an established practice, that changes the equation a bit.

I think we should still make an effort to make Sanity developers aware of what get's passed to the JS-bundle. Could we somehow log to terminal which environment variables will be included, possibly also with a link to a guide about what kind of variables you should be careful about including?

bjoerge:
> sanity build is basically 1:1 with sanity deploy except for the actual deployment - I don't see why build would not use the same as deploy.

I guess in my mind _build_ is not the same thing as _build for production_, whereas _deploy_ to me means _build for production and deploy_, but I can totally see that there's no "right" answer here. One case for having `build` default to development is in situations where you serve the studio statically from disk (e.g. serving it using something else than `@sanity/server`), and don't want minification, etc. (this is something I've seen with `now dev`, for example), but as long as we document what the default is and how to get around it, I don't think it matters all that much what it is.

PR #1611, opened 2019-11-21
[mutator] Increase test timeout to 120 seconds

rexxars:
The mutator uses quite a long time to "boot up" the tests, especially in a CI-environment, which leads to flakey tests when it sometimes goes over the 30s default timeout for tap.

~This PR adds the `-T` flag ("no timeout") to the mutator tests and should hopefully remove the flakey CI results.~
This PR increases the timeout to 2 minutes (120 seconds) through the `-t` flag ("timeout") for the mutator tests and should hopefully remove the flakey CI results.

### DIFF
--- a/packages/@sanity/mutator/package.json
+++ b/packages/@sanity/mutator/package.json
@@ -5,7 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "clean": "rimraf lib",
-    "test": "tap test/*.test.ts",
+    "test": "tap -t120 test/*.test.ts",
     "perf": "node ./perf/run.js"
   },
   "keywords": [


saasen:
Just wanted to pitch in here. Wouldn't it be better to set a higher timeout? If we for some reason introduce a test that never finish, we will at least get an error.

bjoerge:
Agree with @saasen and think it should be set to a higher value. Otherwise LGTM!

rexxars:
Always hard to know what a good timeout value _is_, but fair point. 120 seconds ought to be enough?

saasen:
> Always hard to know what a good timeout value _is_, but fair point. 120 seconds ought to be enough?

Aye. Should be enough.

The only thing I can think of is investigating why the boot up takes that long, but there's probably a good reason for it being that slow?

rexxars:
> > Always hard to know what a good timeout value _is_, but fair point. 120 seconds ought to be enough?
> 
> Aye. Should be enough.
> 
> The only thing I can think of is investigating why the boot up takes that long, but there's probably a good reason for it being that slow?

Lerna is running a bunch of tests from packages in parallel and Travis it has limited hardware to do so. I do think it seems excessively slow, but this is mostly about preventing ~12 minute builds from failing due to a single package sometimes taking 40 seconds instead of 30 to complete tests and stopping a PR from being merged.

saasen:
> > > Always hard to know what a good timeout value _is_, but fair point. 120 seconds ought to be enough?
> > 
> > 
> > Aye. Should be enough.
> > The only thing I can think of is investigating why the boot up takes that long, but there's probably a good reason for it being that slow?
> 
> Lerna is running a bunch of tests from packages in parallel and Travis it has limited hardware to do so. I do think it seems excessively slow, but this is mostly about preventing ~12 minute builds from failing due to a single package sometimes taking 40 seconds instead of 30 to complete tests and stopping a PR from being merged.

Nice. Thanks for clarifying. Yes, that sounds stupidly long. 🤔

PR #1621, opened 2019-11-22
[form-builder] Cleanup isEmpty function

camjc:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**
Same, but with a bunch of redundant stuff.

**Description**
`null` values were going down the isEmptyObject code path. There was a check that type !== null (but it never will be, I think the intent was to check the value not the type).
Just exit early for null cases and stop worrying about them later on in the function.

isEmptyArray was checking if the value was undefined, which we already know its not.

Same with isEmptyPrimative; we already know its not undefined, why check again.
**Note for release**

Tidy up isEmpty function.

**Checklist** 

- [x]  The code is linted


bjoerge:
Thanks @camjc!

PR #1633, opened 2019-11-27
[desk-tool] Resolve document type from template if available

rexxars:
Given the following structure:

```js
S.documentList()
  .title('Unspecified books list')
  .menuItems(S.documentTypeList('book').getMenuItems())
  .filter('_type == $type')
  .params({type: 'book'})
```

You may end up in a situation where the automatically generated editor has the schema type value `*`, which should be resolved by the desk tool to the correct type (structure doesn't figure out that this pane holds a specific schema type, since it's not set using `schemaType(someTypeName)`/`documentTypeList(someTypeName)`.

With this PR, we use the template parameter to resolve the schema type to use if none is explicitly defined.

### DIFF
--- a/packages/@sanity/desk-tool/src/utils/withInitialValue.js
+++ b/packages/@sanity/desk-tool/src/utils/withInitialValue.js
@@ -46,10 +46,8 @@ const withInitialValue = Pane => {
           switchMap(document => {
             const {templateName, parameters} = getInitialValueProps(document, props, paneContext)
             const shouldResolve = Boolean(templateName)
-            const documentType = options.type || (document && document._type)
-
-            // If we were not passed a schema type, use the resolved value if available
-            const paneOptions = options.type ? options : {...options, type: documentType}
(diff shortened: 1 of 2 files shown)


packages/@sanity/desk-tool/src/utils/withInitialValue.js
```diff
@@ -160,4 +158,18 @@ function resolveInitialValueWithParameters(templateName, parameters) {
   )
 }
 
+function resolvePaneOptions(props, templateName) {
+  const {options} = props
+  const hasDefinedType = options.type && options.type !== '*'
+  const typeFromOptions = hasDefinedType ? options.type : undefined
+  let documentType = typeFromOptions || (document && document._type)
```
thomax: I must be blind or something, but I can't see how `document` is available in this function?

rexxars: Well spotted!

PR #1637, opened 2019-11-28
[structure] Allow specifying a default document node resolver

rexxars:
## Background

There are several node types in the desk tool structure which automatically return a document (previously "editor") node. For instance, a `documentTypeList('book')` will make sure that clicking on a book in the generated pane will end up opening a document node configured with the correct schema type and document ID.

Now that we're adding options to customize this node type in a deeper way, most users will want an easy way to set the "defaults" without having to reimplement the child resolver in every step.

## Implementation

This PR adds the ability to define a "resolver"/"getter" function which should return a document node. A user can do this by exporting a named function in their structure definition with the name `getDefaultDocumentNode`, in other words:

 ```js
import S from '@sanity/desk-tool/structure-builder'

export const getDefaultDocumentNode = ({schemaType}) => {
  // Conditionally return a different configuration based on the schema type?
  return S.document().views([
    S.view.form(),
    S.view.component(JsonView).title('JSON')
  ])
}

export default S.defaults()
```

A user can choose between the `S.defaultDocument({schemaType: 'book'})` to use the resolver, or `S.document()` as before to get an editor without any configuration. 

## Questions/concerns

-  All current `S.editor()`/`S.document()` instances will _not_ have the defaults applied. This is backwards-compatible and in line with the implementation, but might be a little bit of a chore to go through and replace.

### CODE REVIEW

File: packages/@sanity/desk-tool/src/utils/resolvePanes.js [hunk 1]
```diff
@@ -297,3 +302,28 @@ function isStructure(structure) {
...
+    .forEach(key => {
+      const {closest} = known.reduce(
+        (acc, current) => {
+          const distance = leven(current, key)
+          return distance < 3 && distance < acc.distance ? {closest: current, distance} : acc
+        },
+        {closest: null, distance: +Infinity}
+      )
+
+      const hint = closest ? ` - did you mean "${closest}"` : ''
```
thomax: Sometimes, when I read nice code, it makes me happy. This code makes me happy :)

kmelve:
Yes, please! This makes a lot of sense from an end-dev perspective, and I dig that you can do stuff conditionally based on the `schemaType` in there.

mariuslundgard:
I'm not a fan of the "fragment" naming (that is, if I understand what it is correctly). "Fragment" sounds like a slice/chunk/partial of a document's data object.

Other suggestions:

```js
getDefaultStructureDocumentNode()
getDefaultDocumentNode()
getDefaultDocumentDefinition()

// We could go further and drop `default` in there.
// And `resolve` is probably more accurate than `get`?
resolveStructureDocumentNode()
resolveDocumentNode()
resolveDocumentDefinition()
```

kmelve:
I like

`resolveDocumentNode` 

out of these. Don't think we need to repeat `structure` within the structure API. Node is a good name for whatever it is.

simen:
I don't think we should remove default from the nomenclature, since this is always overridable in the desk structure. These will only be used when nothing specific if specified.

The word "fragment" just came to me when drafting this feature with Espen without giving it proper thought. My thinking is that it resolves not to a node, but a whole sub structure describing how that document type will appear.

Now I'm thinking there is nothing special about sub-structures—they are just structures, so what about `getDefaultStructureForDocument`?

simen:
Ah, I have to retract a lot of my stuff from earlier today. I was still pretty much controlled by the fever. I agree with knut, but I still think we need the "default" concept, since these are overridden in any situation where another node is specified in the structure.

I'm proposing for `getDefaultDocumentNode`.

PR #1646, opened 2019-12-03
[form-builder] Pass parent object to slug source function

rexxars:
Quite often you'll want to reference a property on the parent of the slug, especially when it is nested. Right now, we only pass the parent path and the document, which requires a user to look up the parent using something like lodash's `_.get()`. The issue is that we also pass keyed path segments (`{_key: 'abc123'}`) which makes it needlessly complicated to do inside of arrays.

With this PR, we pass both `parent` and `parentPath` to the `source` function.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/Slug/SlugInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/Slug/SlugInput.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import speakingurl from 'speakingurl'
-import {get} from 'lodash'
+import {get} from '@sanity/util/paths'
 import Button from 'part:@sanity/components/buttons/default'
 import FormField from 'part:@sanity/components/formfields/default'
 import TextInput from 'part:@sanity/components/textinputs/default'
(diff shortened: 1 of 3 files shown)


kristofferj:
I think @bjoerge or @skogsmaskin is a better reviewer for this than me.

coreyward:
@rexxars Did this make it into today's release?

rexxars:
> @rexxars Did this make it into today's release?

Yes it did! I've added some [basic documentation](https://www.sanity.io/docs/slug-type#custom-source-function-ccba6e80529c) for it as well.

blimpmason:
@rexxars I'm actually having trouble implementing this — I'm trying to use a `slug` field within an object which sources from the title field in the object itself, but the `parent` is coming back undefined. 

Here's what my schema looks like: 

```
export default {
	name: "buildingUnitCategory",
	title: "Unit Category",
	type: "object",
	liveEdit: false,
	fields: [
		{
			name: "title",
			title: "Title",
			type: "string",
		},
		{
			name: "slug",
			title: "Slug",
			type: "slug",
			options: {
				source: (doc, {parent}) => parent && parent.title
			},
		},
		{
			name: "bedroomCount",
			title: "Bedroom Count",
			type: "number",
		},
	],
};
```

Let me know if I'm missing something!

PR #1649, opened 2019-12-06
[document-store] Cleanup and move to @sanity/base

bjoerge:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

**Description**

What we currently exposed as the _document-store_ was a pretty messy and complex machine spread across two packages. In addition, hardly any other method than `checkoutPair` has actually been in use, (some of them didn't really make any sense, or was duplicated elsewhere like in `listenQuery`). There's still complexity (esp. in everything surrounding the buffered document stuff), but this PR reorganises it so it should be a lot easier to follow at least.

In addition:
- Added TypeScript to the @sanity/base package and annotated the document store (best effort)
- Deprecated the methods that's not in use / should not be relied upon (with warning), see `_deprecatedAPI.ts`
- Removed the @sanity/document-store package as it only represented an extra indirection and never used standalone (we can always add it back later if we want)
- Instead of defining the exported API of the document store in a file named `index.js`, it's now named `document-store.ts` (for easier file lookup)

**Note for release**
N/A

**Checklist** 

- [x]  The code is linted


bjoerge:
These changes have been merged through other commits. Thanks for feedback!

PR #1651, opened 2019-12-09
[desk-tool] Handle permission errors from database

skogsmaskin:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**
If you have set up custom access control and get a permission error trying to do something with the document, the studio will break with an ugly technical error message.

**Description**

The error handling in `DocumentPane` has been streamlined and is always written to state.error by the various subscriptions. By a common error handler function, we now can test if it is a permission error, and if it is, output an own Snackbar error message for it.

What complicated this was that permission errors that happened when trying to create a document, happened another place (formBuilderValueStore) which gets a rebase event with a null document. In order to fix this, I added an `onError` callback function for  `BufferedDocument`. Then document store's `createObservableBufferedDocument` can register it's `onError` callback, and if it is called and the error is a permission error, it can error the stream (`updates$.error(error)`). Then we are able to catch the original error inside `DocumentPane`.

**Note for release**

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/desk-tool/src/pane/DocumentPane.js
+++ b/packages/@sanity/desk-tool/src/pane/DocumentPane.js
@@ -41,6 +41,7 @@ import {historyIsEnabled} from './Editor/history'
 import menuItemStyles from './styles/documentPaneMenuItems.css'
 import {getDocumentPaneFooterActions} from './documentPaneFooterActions'
 import {getProductionPreviewItem, getMenuItems} from './documentPaneMenuItems'
+import {ClientError} from '@sanity/client'
 import {validateDocument} from '@sanity/validation'
 
 const DEBUG_HISTORY_TRANSITION = false
@@ -70,6 +71,11 @@ function isPreviewHotkey(event) {
 }
 
 const isValidationError = marker => marker.type === 'validation' && marker.level === 'error'
+
+const isPermissionError = error =>
+  !!(error && error instanceof ClientError && error.message.match(/permissions/gi))
+
+const isClientError = error => !!(error && error instanceof ClientError)
 
 const INITIAL_DOCUMENT_STATE = {
   isLoading: true,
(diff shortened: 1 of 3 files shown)


packages/@sanity/desk-tool/src/pane/DocumentPane.js
```diff
@@ -41,6 +41,7 @@ import {historyIsEnabled} from './Editor/history'
 import menuItemStyles from './styles/documentPaneMenuItems.css'
 import {getDocumentPaneFooterActions} from './documentPaneFooterActions'
 import {getProductionPreviewItem, getMenuItems} from './documentPaneMenuItems'
+import {ClientError} from '@sanity/client'
```
rexxars: `@sanity/client` is not a (direct) dependency of `@sanity/desk-tool` - see below

packages/@sanity/desk-tool/src/pane/DocumentPane.js
```diff
@@ -71,6 +72,11 @@ function isPreviewHotkey(event) {
 
 const isValidationError = marker => marker.type === 'validation' && marker.level === 'error'
 
+const isPermissionError = error =>
+  !!(error && error instanceof ClientError && error.message.match(/permissions/gi))
+
+const isClientError = error => !!(error && error instanceof ClientError)
```
rexxars: I think we should "duck type" this instead of checking `instanceof` - since the client used is the one from `@sanity/base`, it's not guaranteed to be the same. If someone installs `@sanity/client` of a different version in their studio, the check won't match, since it's not a direct dependency. 

I'm thinking it would probably be good enough to check for the correct value in `error.statusCode`, unless there is something I'm overlooking?

skogsmaskin:
We now have better support for this, rendering this PR obsolete.

PR #1663, opened 2019-12-16
[desk-tool] Fix deprecated part crashing on unpublished documents

rexxars:
Fixes a bug where using the `part:@sanity/desk-tool/after-editor-component` part will crash the desk tool if the currently viewed document is not yet published.


jorngeorg:
Thank you!

PR #1665, opened 2019-12-17
[base] Maintain order of documents when collating drafts/published

rexxars:
#1422 introduced a change of behavior in `collate()`, where the order of the passed documents would not be maintained anymore.

I'm not sure if this was intentional, but there is code that depends on the order currently - the desk tool, when showing a list of documents ordered by a specific field, for instance.

This PR switches from using an object (which does not guarantee order) + Object.values, to using a Map - which is guaranteed to maintain order.


rexxars:
> Oh, great catch! Was this causing problems with numeric indices perhaps?

Numeric strings, indeed.

PR #1666, opened 2019-12-17
[mutator] Fixes the problem where diff-match-patch and inc-dec would y…

simen:
Fixes a problem where diff-match-patch and inc/dec patches could yield exceptions when targetting array indicies or attributes that were missing. This is almost impossible to experience except under very laggy network conditions when more than one user is battle-editing the same fields.


simen:
I'm guessing noone else can review this? Let's work on that problem in the new year!

PR #1671, opened 2019-12-18
[desk-tool] Optimize loading of large lists

rexxars:
Document lists with a large number of documents are currently pretty slow to load. At some point we want to provide a scrolling window implementation that we've talked about for a while.

This is a temporary workaround until we can get around to that implementation, however. In essence, it:

- Loads only the 100 first documents matching the query
- Waits for the user to scroll half way down in that list
- Starts loading the "full" list (without clearing the list while loading)
- Re-renders the list with the "full" document set, which should usually be an add-only operation

I'm putting "full" in quotes because we were previously loading up to 50 000 document IDs, while we're now stopping at 2000. Even this seems excessive, as I/we can't see a legitimate reason for using such an enormous list to find anything - you're much better off using the search or providing a more fine-grained navigation using the structure builder if you've got that amount of documents. 

If you _do_ scroll to the bottom of the 2000 documents, we'll show a message saying that more documents _do_ exist, but are not loaded.

### DIFF
--- a/packages/@sanity/base/src/components/QueryContainer.js
+++ b/packages/@sanity/base/src/components/QueryContainer.js
@@ -36,8 +36,7 @@ const createErrorChildProps = error => ({
   error
 })
 
-// todo: split into separate standalone parts so that behavior can be re-used
-export default streamingComponent(receivedProps$ => {
+export const getQueryResults = receivedProps$ => {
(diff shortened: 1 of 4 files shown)


jemmima:
I've asked Thomas to look at this urgently

evenwestvang:
Sound completely reasonable to me

PR #1680, opened 2019-12-24
[@sanity/core] Allow startup with env without api settings

chrispecoraro:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

`ensureProjectConfig()` in startAction.js fails when there are no values in `sanity.json`'s `api.project_id` and `api.dataset`, while with the recent addtion of dotenv (`.env`) functionality, `.env.[development|productions|etc.]` can be used to set these values. There should be no values in `sanity.json` in this case for these since it would confuse someone who reads through the configuration. Also, it is easy to configure these values through `.env` in cases where `gatsby-source-sanity` is used and the same values need to be set there as well.

**Description**

This code change checks `context.apiClient().clientConfig.projectId` and `context.apiClient().clientConfig.dataset`, since the values for `projectId` and `dataset` are present both when there are values set in either the `sanity.json` file OR the `.env` file.

`SANITY_STUDIO_API_PROJECT_ID=adsfasdf`
`SANITY_STUDIO_API_DATASET=123123123`

- A message "Using project id and dataset from .env file instead of sanity.json file" is displayed to show that the values are being overridden by the `.env` file. (Even better would be detecting when both values are present in both the sanity.json file AND the `.env` file, since the `.env` values are used in this case.
- Instead of using an or to check for values not present and then a return if not present (confusing), an `and` is used to make this clearer (if config values present, then return).
 

**Note for release**

This change allows for Sanity project id and dataset values to be set through the dotenv files and not through the sanity.json configuration file. 

**Checklist** 

- [ ]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [x]  The code is linted

### DIFF
--- a/packages/@sanity/cli/src/util/clientWrapper.js
+++ b/packages/@sanity/cli/src/util/clientWrapper.js
@@ -39,15 +39,12 @@ export default function clientWrapper(manifest, configPath) {
   return function(opts = {}) {
     const {requireUser, requireProject, api} = {...defaults, ...opts}
     const userConfig = getUserConfig()
-    const userApiConf = userConfig.get('api')
     const token = envAuthToken || userConfig.get('authToken')
     const apiHost = apiHosts[sanityEnv]
-    const apiConfig = Object.assign(
-      {},
-      userApiConf || {},
-      (manifest && manifest.api) || {},
-      api || {}
-    )
+    const apiConfig = {
+      ...((manifest && manifest.api) || {}),
+      ...(api || {})
+    }
 
     if (apiHost) {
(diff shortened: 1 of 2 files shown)


kmelve:
Hi! Thanks for this PR. We'll review it and get back to you.

rexxars:
Sorry, this totally fell under my radar.

I did some changes to this;
- Notify on non-default project ID and dataset individually (and include their new values)
- Prevented the `apiClient` call from throwing when no project ID was set (allows reconfiguring through prompts)

Will do a check-off with my colleagues before merging - thanks for contributing!

PR #1687, opened 2020-01-06
[export/import] Improve asset performance

rexxars:
Exporting and importing assets is currently quite slow. The assumption was that we would be able to saturate a clients connection and should thus keep the number of parallel downloads/uploads to a minimum.

This isn't the case, however, so by increasing the number of concurrent downloads and uploads, we greatly increase the speed of the process.

Exporting a large number of assets is improved by a factor of ~300%
Imports greatly vary based on the size of the assets, but should see benefits too.

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/exportDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/exportDatasetCommand.js
@@ -10,12 +10,13 @@ const noop = () => null
 
 const helpText = `
 Options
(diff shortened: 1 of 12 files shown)


thomax:
I see that we're moving from 3 to 8 as default concurrency. Are we sure we shouldn't keep the default concurrency low for a while, and let the power users who need a high concurrency number find and report any side ill effects? Then, if all's good after a while, bump up the default?

PR #1708, opened 2020-01-21
[form-builder] Use shallow equality check on path property

rexxars:
While doing some shallow profiling work, I discovered that the `path` property passed to the `FormBuilderInput` component was causing it to be re-rendered more often than necessary.

By comparing the path array using the path utils and the rest of the properties with a shallow equals function, we can significantly reduce the amount of re-renders.

### DIFF
--- a/packages/@sanity/form-builder/src/FormBuilderInput.tsx
+++ b/packages/@sanity/form-builder/src/FormBuilderInput.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable complexity */
 import React from 'react'
+import shallowEquals from 'shallow-equals'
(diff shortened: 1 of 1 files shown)


packages/@sanity/form-builder/src/FormBuilderInput.tsx
```diff
@@ -65,6 +66,16 @@ export class FormBuilderInput extends React.PureComponent<Props> {
     }
   }
 
+  shouldComponentUpdate(nextProps) {
+    const {path: oldPath, ...oldProps} = this.props
+    const {path: newPath, ...newProps} = nextProps
```
bjoerge: Neat pattern!

PR #1710, opened 2020-01-22
[client] Add typescript definitions

rexxars:
This PR adds hand-written typescript definitions to the client.

It doesn't try to be clever in reusing methods/types for the promise/observable versions - I tried a few approaches but couldn't come up with something that easily worked, so I decided to just do it the simple way for now.

I initially started converting the client to typescript, but quickly realized it was quite a lot of work trying to figure out the right way to model everything in a backwards-compatible way. We've been talking about slimming down the client to a more "importable"/"tree-shakable" implementation anyway, so I feel like the typescript rewrite is better done as part of that process.

Tried to have some community members give the definitions a try, but have not heard of any 👍 / 👎 , so I assume the easiest way to put it to the test is to simply merge it and see if we encounter any issues?

### DIFF
--- a/packages/@sanity/client/sanityClient.d.ts
+++ b/packages/@sanity/client/sanityClient.d.ts
@@ -169,10 +169,10 @@ export type PatchBuilder = (patch: Patch) => Patch
 
 export type PatchMutationOperation = PatchOperations & MutationSelection
 
-export type Mutation =
-  | {create: SanityDocumentStub}
-  | {createOrReplace: IdentifiedSanityDocumentStub}
-  | {createIfNotExists: IdentifiedSanityDocumentStub}
+export type Mutation<R = any> =
+  | {create: SanityDocumentStub<R>}
+  | {createOrReplace: IdentifiedSanityDocumentStub<R>}
+  | {createIfNotExists: IdentifiedSanityDocumentStub<R>}
   | {delete: MutationSelection}
   | {patch: PatchMutationOperation}
 
@@ -305,14 +305,14 @@ export class Patch {
    *
    * @param options Options for the mutation operation
    */
-  commit(options: FirstDocumentMutationOptions): Promise<SanityDocument>
+  commit<R = any>(options: FirstDocumentMutationOptions): Promise<SanityDocument<R>>
 
   /**
    * Commit the patch, returning a promise that resolves to an array of the mutated documents
    *
    * @param options Options for the mutation operation
    */
-  commit(options: AllDocumentsMutationOptions): Promise<SanityDocument[]>
+  commit<R = any>(options: AllDocumentsMutationOptions): Promise<SanityDocument<R>[]>
 
   /**
    * Commit the patch, returning a promise that resolves to the document ID of the first mutated document
@@ -333,7 +333,7 @@ export class Patch {
    *
    * @param options Options for the mutation operation
    */
-  commit(options?: BaseMutationOptions): Promise<SanityDocument>
+  commit<R = any>(options?: BaseMutationOptions): Promise<SanityDocument<R>>
 
   /**
    * Clears the patch of all operations
@@ -351,7 +351,7 @@ export class Transaction {
    *
    * @param doc Document to create. Requires a `_type` property.
    */
-  create(doc: SanityDocumentStub): this
+  create<R = any>(doc: SanityDocumentStub<R>): this
 
   /**
    * Creates a new Sanity document. If a document with the same `_id` already exists, the create operation will be ignored.
@@ -359,7 +359,7 @@ export class Transaction {
    *
    * @param doc Document to create if it does not already exist. Requires `_id` and `_type` properties.
    */
-  createIfNotExists(doc: IdentifiedSanityDocumentStub): this
+  createIfNotExists<R = any>(doc: IdentifiedSanityDocumentStub<R>): this
 
   /**
    * Creates a new Sanity document, or replaces an existing one if the same `_id` is already used.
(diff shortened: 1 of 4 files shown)


joltmode:
I'm in the process now to switch over to these typings from my own homebrewn. Will give report later or tomorrow.

packages/@sanity/client/sanityClient.d.ts
```diff
@@ -0,0 +1,1784 @@
...
+}
+
+type ListenEvent =
+  | MutationEvent
+  | ChannelErrorEvent
+  | DisconnectEvent
+  | ReconnectEvent
+  | WelcomeEvent
+
+interface ListenOptions {
```
joltmode: I have no idea of the usage scope for all of the interfaces, but the ones that play a role as inputs, those should definitely be exported.

Use cases: decorators.

In this case, I have custom React Hooks that wrap around Sanity.  
Before trying out these - official TS defs, I had my own that exposed those that I found worthy.  
Since I am making them as generics (looking forward to contribute to OSS community), they expose full API surface, in `useSubscription` case, allowing to set custom `ListenOptions` per hook-usage.

packages/@sanity/client/sanityClient.d.ts
```diff
@@ -0,0 +1,1784 @@
...
+type MutationEvent = {
+  type: 'mutation'
+  documentId: string
+  eventId: string
+  identity: string
+  mutations: Mutation[]
+  previousRev?: string
+  resultRev?: string
+  result?: SanityDocument
+  previous?: SanityDocument | null
```
joltmode: This currently is an issue with all of these typings, it's not possible to override response types.

For these, my typings had: `ListenEvent<T>` that nested into `result?: T`. I didn't have a `SanityDocument` type, but here it would simply nest into `result?: SanityDocument<T>`.

packages/@sanity/client/sanityClient.d.ts
```diff
@@ -0,0 +1,1784 @@
...
+  | AllDocumentsMutationOptions
+  | AllDocumentIdsMutationOptions
+
+interface RawQueryResponse {
+  q: string
+  ms: number
+  result: any
+}
+
+interface SanityDocument {
```
joltmode: `interface SanityDocument<T = {}>` + `[P in keyof T]: T[P]` alongside `[key: string]: any`.

(I may have gotten the syntax wrong, on a quick pass to set down all notes)

packages/@sanity/client/sanityClient.d.ts
```diff
@@ -0,0 +1,1784 @@
...
+    params?: QueryParams,
+    options?: T
+  ): Promise<T extends UnfilteredReponseQueryOptions ? RawQueryResponse : any>
+
+  /**
+   * Fetch a single document with the given ID.
+   *
+   * @param id Document ID to fetch
+   */
+  getDocument(id: string): Promise<SanityDocument | undefined>
```
joltmode: `getDocument<R = any>(id: string): Promise<SanityDocument<R> | undefined>`

Again, not sure, just making a point, that users should have means to provide the typings upfront instead of casting on response.

joltmode: This should have to cascade to all responses.

packages/@sanity/client/sanityClient.d.ts
joltmode: Yeah, the syntax would be:

```
type SanityDocument<T extends object = {}> = {
  [P in keyof T]: T[P]
} & {
  _id: string
  _rev: string
  _type: string
  _createdAt: string
  _updatedAt: string
  [key: string]: any
}
```

packages/@sanity/client/sanityClient.d.ts
rexxars: Agreed, and I see your point. I've exported a bunch of interfaces/types now.

packages/@sanity/client/sanityClient.d.ts
rexxars: Thanks, should be able to set type on these now!

packages/@sanity/client/sanityClient.d.ts
rexxars: Thanks so much for this - I'm still fairly new to typescript, and this was very useful. I've now implemented this throughout.

packages/@sanity/client/sanityClient.d.ts
rexxars: I _think_ I got all the methods covered by the new signature, allowing you to set the return type.

rexxars:
Hey @joltmode - thanks so much for your review!
I've updated the PR to reflect your proposed changes - would you be able to give the new changes a try?

joltmode:
Besides the fact that my custom typings exposed `SanityClient` instance as `Client` and had to rename a few imports, everything else went seamless!

Thanks!

Got some stats for you! While not the biggest of projects for now, approximately 60% of the codebase is actually work with Sanity client.

```
cloc --vcs git --exclude-dir=sanity-information-central --exclude-lang JSON,Markdown,HTML
     108 text files.
     100 unique files.
      48 files ignored.

github.com/AlDanial/cloc v 1.84  T=0.29 s (244.3 files/s, 18735.6 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
TypeScript                      62            736            214           4331
JavaScript                       9             16              4            143
-------------------------------------------------------------------------------
SUM:                            71            752            218           4474
-------------------------------------------------------------------------------
```

rexxars:
Thanks so much for testing!

PR #1735, opened 2020-02-14
[core] Generate GraphQL schemas as generation 2

rexxars:
This PR introduces the necessary changes for the new generation of GraphQL schemas.

Roughly summarized:
- Tagged APIs
- Breaking changes checking on deploy
- New (deep) filters
- Sorting
- No pluralization of root queries

This does require a server change which should be deployed soonish.


rexxars:
Rebased :)

packages/@sanity/core/src/commands/graphql/deployGraphQLAPICommand.js
```diff
@@ -3,13 +3,19 @@ import lazyRequire from '@sanity/util/lib/lazyRequire'
 const helpText = `
 Options
   --dataset <dataset> Deploy API for the given dataset
+  --tag <tag> Deploy API to given tag (defaults to 'default')
+  --generation <generation> API generation to deploy
```
saasen: Since we now default to the latest generation, it should also be noted in the parameter here that is the default.

kmelve: Fixed!

Spy-Seth:
That's an amazing PR! We are block to use GraphQL with Sanity because of the lacks of filters in the current Schema. 

We could help if you need some help to have this merged :smile:

saasen:
@Spy-Seth: Thank you!  We are really looking forward to this, and it's so nice to hear you really need this. This will be merged very soon.

PR #1762, opened 2020-02-28
[base] Make sure restoring from history takes liveEdit into account

bjoerge:
This patch ensures we restore to the published document instead of creating a new draft when restoring a document from history that has `liveEdit` enabled.


thomax:
Nice 👍

Issue #1809, opened 2020-03-28
DraftStatus (used in DocumentPaneItemPreview) should comply with the custom badges API


israelroldan:
**Is your feature request related to a problem? Please describe.**
In the editorial workflow one of my projects is using the badges for Drafts / Published are replaced with Prospect / Active respectively. This works great on the document pane status bar (as per the custom badges API). 

However, the item list still shows DRAFT as the badge:
<img width="1096" alt="Screenshot 2020-03-28 at 21 19 03" src="https://user-images.githubusercontent.com/159962/77832950-f07cb280-7139-11ea-96ec-b29639970220.png">

**Describe the solution you'd like**
I checked the code that is generating this DRAFT badge in `@sanity/desk-tool/src/components/DocumentPaneItemPreview`, it comes from `@sanity/desk-tool/src/components/DraftStatus` component:

```js
import DraftStatus from './DraftStatus'
//...
const getStatusIndicator = (draft, published) => {
  if (draft) {
    return DraftStatus
  }
  return published ? null : NotPublishedStatus
}
//...
```

IMHO, `DocumentPaneItemPreview` should be made aware of the custom badges, but if this is out the table, then it would be extremely useful if `DraftStatus` could be reimplemented by making it a `part:`, since currently it is imported directly with `require('./DraftStatus')` so there is no way to extend this bit if needed.

Something like this:
```js
import DraftStatus from 'part:@sanity/desk-tool/components/draft-status'
```
would then allow us to reimplement this behaviour using `sanity.json`.

israelroldan:
Gave it a try with #1810

vicbergquist:
> IMHO, `DocumentPaneItemPreview` should be made aware of the custom badges, but if this is out the table, then it would be extremely useful if `DraftStatus` could be reimplemented by making it a `part:`, since currently it is imported directly with `require('./DraftStatus')` so there is no way to extend this bit if needed.

This makes total sense 👍  Thank you for submitting!

kmelve:
Hi! 

We are currently working on improving our workflows and follow-up on our open GitHub repository. In that work, we have decided to close most issues older than [the release of Sanity Studio v3](https://www.sanity.io/blog/sanity-studio-v3-simplified-yet-powerful-customization).

[...]

PR #1810, opened 2020-03-28
[desk-tool] Make desk-tool/components/DraftStatus an implementable part (#1809)

israelroldan:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

https://github.com/sanity-io/sanity/issues/1809

**Description**

Before this PR, DraftStatus is directly required from local source.
This PR makes it an implementable part so it can also be overwritten, which will introduce consistency with the custom badges API in case you want to override the draft badge entirely.

**Note for release**

Allow "DRAFT" badge in DocumentPaneItemPreview to be re-implemented as `part:@sanity/desk-tool/components/draft-status`

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/desk-tool/sanity.json
+++ b/packages/@sanity/desk-tool/sanity.json
@@ -42,6 +42,14 @@
       "description": "An observable that emits a function to filter fields in form builder"
     },
     {
+      "name": "part:@sanity/desk-tool/components/draft-status",
+      "description": "A component that is displayed as a badge in the document list pane"
+    },
+    {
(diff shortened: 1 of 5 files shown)


israelroldan:
For my use case this is enough, but I see that it could greatly benefit from getting the draft / published version of the document (for example using `useEditState`). What do you think about it?

rexxars:
We're refactoring these fundamentals right now, so we'll revisit this with a different API

PR #1826, opened 2020-04-04
feat(client): add support for patch/delete with query and params

KarlGe:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

The javascript client getSelection (used in patch and delete) accepts an object with a query, but not params for that query

**Description**

I've added the option to pass params in adition to the query that was already accepted. I added a test for this as well, but I'm not 100% sure if the test is helpful?
Running the tests works fine, and testing this feature in my own project also worked fine 
(This example isn't extremely useful since I could have just used the document id, but the intention is to use the query to check the author for example)
```
sanity
    .delete({
      query: `*[_type == 'submission' && _id == $submissionId]`,
      params: { submissionId: id },
    })
    .then((sanityRes) => res.sendStatus(200))
    .catch((err) => {
      return res.json(err);
    });
```
Not sure if this counts as a feature or a bug (or maybe it's even intended? 😮) - also I don't think passing a query to the delete function is documented at all.

**Note for release**
Allow passing params when using a query instead of just a document id in the delete and patch functions of the javascript client.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/client/src/util/getSelection.js
+++ b/packages/@sanity/client/src/util/getSelection.js
@@ -4,7 +4,7 @@ module.exports = function getSelection(sel) {
   }
 
   if (sel && sel.query) {
-    return {query: sel.query}
+    return 'params' in sel ? {query: sel.query, params: sel.params} : {query: sel.query}
   }
 
   const selectionOpts = [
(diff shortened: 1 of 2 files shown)


rexxars:
Thanks, and sorry for the long wait!

PR #1829, opened 2020-04-06
[core] GraphQL: Only use field description if type is string

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

If a field has a description set to a React element (`description: <span>Foo</span>`), it gets serialized and sent to the server as the object description of the element, and is rejected by the validation.

**Description**

This PR silently ignores any non-string description fields. I don't see any reason for adding a flag or a warning as I think this is an edge-case (in an upcoming "editor declaration" implementation, we'll have better ways of achieving rich "descriptions" for fields while keeping the schema definition "pure".

**Note for release**

- Fixed bug where non-string field descriptions would prevent GraphQL APIs from being deployed

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
+++ b/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
@@ -421,7 +421,8 @@ function extractFromSanitySchema(sanitySchema) {
   }
 
   function getDescription(type) {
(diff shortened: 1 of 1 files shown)


rexxars:
Released in [v1.149.11](https://github.com/sanity-io/sanity/releases/tag/v1.149.11)

PR #1831, opened 2020-04-07
[core] Add `require.context` hook for schema generation, script exec

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

If you're using `require.context` in the studio (for instance, to load all schema type files without having to specify them all), this works fine until you try to deploy a GraphQL API or running a script with `sanity exec`, at which point it'll crash because `require.context` does not exist in the node APIs. 

**Description**

This PR injects a "polyfill" for `require.context` into the node environment when running `sanity graphql deploy` and `sanity exec`. It's a bit of a hairy beast, so it comes with a set of pessimistic try/catch clauses, warning should it fail to register itself. 

To try it out, here's a sample `schema.js` file which requires all `.js` files in the same folder (recursively):

```js
import createSchema from 'part:@sanity/base/schema-creator'
import schemaTypes from 'all:part:@sanity/base/schema-type'

const context = require.context('./', true, /^(?!.*schema\.js$).*\.js$/)
const types = context
  .keys()
  .map((filename) => context(filename))
  .map((mod) => (mod.__esModule && mod.default) || mod)

export default createSchema({
  name: 'default',
  types: schemaTypes.concat(types),
})
```

**Note for release**

- Implemented polyfill for `require.context` in node.js environment, preventing `sanity graphql deploy` and `sanity exec` from crashing if used.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/package.json
+++ b/packages/@sanity/core/package.json
@@ -52,6 +52,7 @@
     "jsdom-global": "^3.0.2",
     "json-lexer": "^1.1.1",
     "json5": "^1.0.1",
+    "klaw-sync": "^4.0.0",
     "lodash": "^4.17.15",
     "log-symbols": "^2.2.0",
     "oneline": "^1.0.3",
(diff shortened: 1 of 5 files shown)


rexxars:
Released in [v1.149.11](https://github.com/sanity-io/sanity/releases/tag/v1.149.11)

PR #1834, opened 2020-04-09
[@sanity/schema] Allow __ in field names for denoting secondary "system" fields

LiamMartens:

- [ ]  Maintenance
- [x]  Other, please describe:
Suggestion and food for thought

- [x]  Yes
- [ ]  No

**Current behavior**

**Description**
It would allow field names to start with 2 underscores (single underscore is still not allowed to prevent collisions). These could be fields that are ie injected by code or plugins but are not relevant to an end user. It would be good to have a specific notation for these that relate to the system field convention but does not collide with it.

**Note for release**

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/types/object.ts
+++ b/packages/@sanity/schema/src/sanity/validation/types/object.ts
@@ -24,10 +24,10 @@ function validateFieldName(name): Array<any> {
       )
     ]
   }
-  if (name.startsWith('_')) {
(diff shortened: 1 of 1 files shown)


rexxars:
After some discussion we've decided we want to reserve underscore-fields for internals, and we feel even a double underscore approach might be confusing to some users. Apologies for leaving this open for so long.

PR #1846, opened 2020-04-21
[desk-tool] Refactor file organization (chore)

mariuslundgard:
#### Type of change (check at least one)
- [ ]  Maintenance
- [x]  Other, please describe: **Code/file reorganizationo**

#### Does this change require a documentation update? (Check one)
- [ ]  Yes
- [x]  No

#### Current behavior

_n/a_

#### Description

These changes include moving around React components and CSS files to fit into the following structure:

```
@sanity/
  desk-tool/
    src/
      components/
        # All React components that are used in multiple places (panes)
      panes/
        # The various distinct panes that can be rendered by the Desk tool
      tool/
        # The root React component for the Desk tool
```

#### Note for release

_n/a_

#### Checklist
- [ ]  The code is linted


mariuslundgard:
> isn't it nicer to keep CSS files in a styles folder so that the components folder doesn't get too cluttered/long?

@vicbergquist I changed that because I find it quicker to jump between the CSS file and the JS file if they are in the same directory. I consider both CSS and JS to be the component, so think it should be in one place. If components list gets too long, it can be divided into subfolders?

vicbergquist:
> > isn't it nicer to keep CSS files in a styles folder so that the components folder doesn't get too cluttered/long?
> 
> @vicbergquist I changed that because I find it quicker to jump between the CSS file and the JS file if they are in the same directory. I consider both CSS and JS to be the component, so think it should be in one place. If components list gets too long, it can be divided into subfolders?

@mariuslundgard ok makes sense! 👍

mariuslundgard:
Good! If we later find that using `styles` directories was better, we can always go back. Or better yet – we are on `styled-components` by then :)

PR #1860, opened 2020-04-29
Custom inline objects

saasen:


saasen:
> Great work! Could it be that we should name this stuff `ObjectHmhm` instead of `BlockHmhm`? Like `ObjectRendering`? InlineBlock doesn't make much sense, as it's not a block. We also need to edit the annotations in the same fashion, and those are def. not blocks.

It does make sense! https://www.w3schools.com/css/css_inline-block.asp
Partial joke aside, what do you think about `InlineObject`?

I can rename the other stuff as well.

saasen:
Done!

PR #1862, opened 2020-04-30
[form-builder] Made styling work for objects and inline objects

saasen:


saasen:
Not sure if we can remove some of the CSS in the files.

skogsmaskin:
I can have a look afterwards regarding if there is anything  to be removed. I need to run through lot's of CSS anyway for the toolbar.

PR #1869, opened 2020-05-05
[desk-tool] Prevent duplicate publish Snackbars when using split panes

saasen:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When publishing a document when using split panes, you get _n_ amount of Snackbar messages based on how many panes are open.

**Description**

This PR uses the `preventDuplicate` functionality already available in the `SnackbarProvider` to prevent multiple instances of the same Snackbar type to appear.

Not sure if this was the intended solution for this, as it still generates multiple Snackbars when multiples panes are open. It kind of just swallows if it's of the same type in this PR. I saw other Snackbars which would also appear multiple times from what I can see, as they are rendered in each pane. A more robust way to fix this would be to move the connectionstate up in the hierarchy and have a common place for handling connectionstate Snackbars and document actions Snackbars. What do you think?

**Note for release**

Fix a problem where publishing a document when multiple panes are open would generate multiple Snackbars

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/base/src/components/DevServerStatus.js
+++ b/packages/@sanity/base/src/components/DevServerStatus.js
@@ -86,6 +86,7 @@ class DevServerStatus extends PureComponent {
           title={<strong>Reload required!</strong>}
           subtitle={<div>To see your latest changes, you need to reload the browser window.</div>}
           action={{title: 'Reload', callback: () => window.location.reload()}}
+          allowDuplicateSnackbarType
         />
       )
     }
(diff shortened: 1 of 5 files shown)


mariuslundgard:
As @rexxars is saying, not displaying duplicates should be the default behavior!

saasen:
I can change that in this PR if that is the wanted behaviour? I'll also have a check to see if it could have some unintended behaviour of the existing snackbars!

saasen:
I changed the default behaviour. It looks like this should be fine with the existing snackbars that are in use.

PR #1886, opened 2020-05-18
[chore] Fix eslint so it uses a common config for packages

mariuslundgard:
- Resolves the issue of eslint config for TS files "bleeding" over into JS files.
- Configures the monorepo to use a common `.eslintrc.js` for all packages, no matter if they include JS or TS or both (no need to add a `.eslintrc` in the packages that extends the root config).
- Installs and configures `eslint-plugin-react-hooks`.


.eslintrc.js
```diff
@@ -0,0 +1,75 @@
...
+  plugins: ['import', 'prettier', 'react', 'react-hooks'],
+
+  // TypeScript files
+  overrides: [
+    {
+      ...common,
+      files: ['**/*.ts', '**/*.tsx'],
+      parser: '@typescript-eslint/parser',
+      extends: [
+        // './packages/eslint-config-sanity/index.js',
```
rexxars, line 59: Isn't this saying not to use any of our own rules?

mariuslundgard: That config is left out, because I copied the setup from the current TS config:
https://github.com/sanity-io/sanity/blob/next/.eslintrc-ts.js#L19-L25

Maybe @bjoerge knows the reason for not including it?

mariuslundgard:
*Update*: did some changes because the `import` plugin seemed to still get loaded for TS files, even though it was overridden. (So it seems the "overrides" items doesn't replace, but the common configuration.)

.eslintrc.js
mariuslundgard: I just tried enabling `./packages/eslint-config-sanity/index.js` and running `npm run lint`. Some of the rules in there seem to somehow be incompatible with `@typescript-eslint/parser` (anyways, that's by best guess – as it throws a "TypeError: Cannot read property 'body' of null" exception originating in the eslint source code).

bjoerge: Fixed in d6298925 by replacing the `no-useless-constructor` rule with `@typescript-eslint/no-useless-constructor`:
https://github.com/sanity-io/sanity/blob/d629892523030786bef879b3e258b2f0305be74f/.eslintrc.js#L24-L27

PR #1898, opened 2020-05-27
[validation] format dates in validation messages and support validations on date type

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When you have validation on date fields, the date isn't formatted when showing in the validation message, which makes it hard to read.

Screenshot 2020-05-26 at 16 46 13

Also, validation hasn't been possible on the `date` type.

**Description**

This gets the options from the field to be validated, and uses it in the validator to format the dates using either the default format, or the format specified in options.

This adds the same validation methods to the `date` type. 

**Note for release**

- Fixed an issue where `date` wouldn't get the same built-in validation methods as `datetime`
- Fixed date formatting in validation messages

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/validation/src/validateDocument.js
+++ b/packages/@sanity/validation/src/validateDocument.js
@@ -131,7 +131,12 @@ function validatePrimitive(item, type, path, options) {
 
   const results = type.validation.map(rule =>
     rule
-      .validate(item, {parent: options.parent, document: options.document, path})
+      .validate(item, {
+        parent: options.parent,
+        document: options.document,
+        path,
+        type: {name: options.type.name, options: options.type.options}
+      })
       .then(currRuleResults => applyPath(currRuleResults, path))
   )
 
--- a/packages/@sanity/validation/package.json
+++ b/packages/@sanity/validation/package.json
@@ -35,7 +35,8 @@
   "dependencies": {
     "clone-deep": "^2.0.1",
     "es6-error": "^4.0.2",
-    "type-of-is": "^3.5.1"
+    "type-of-is": "^3.5.1",
+    "date-fns": "^1.29.0"
   },
   "jest": {
     "verbose": true,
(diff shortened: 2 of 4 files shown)


packages/@sanity/validation/src/validateDocument.js
```diff
@@ -131,7 +131,13 @@ function validatePrimitive(item, type, path, options) {
 
   const results = type.validation.map(rule =>
     rule
-      .validate(item, {parent: options.parent, document: options.document, path})
+      .validate(item, {
+        parent: options.parent,
+        document: options.document,
+        path,
+        fieldOptions: options.type && options.type.options,
+        typeName: options.type && options.type.name
```
saasen, line 139: Since it's now two fields from the `type` we are sending in here, would it make more sense to send in the whole `type`? So we don't have to change this and add a new field if we happen to need another field from the `type`?

vicbergquist: I guess so 🤔

vicbergquist:
Also pushed a change to enable validations on the `date` type

packages/@sanity/validation/src/validateDocument.js
saasen: It would also align with the other fields passed in already (`document` and `parent`). 👍 I'll approve given these changes (if you agree, of course).

vicbergquist: ok! Will change it

vicbergquist:
I guess I also have to now add `moment.js` to the `package.json`? @rexxars

rexxars:
> I guess I also have to now add `moment.js` to the `package.json`? @rexxars

I'd prefer if we could use [date-fns](https://date-fns.org/v1.29.0/docs/format)@^1.29.0 - we're trying to migrate away from moment, and date-fns is already used by the form builder etc. And yes, it needs to be added to package.json

vicbergquist:
> > I guess I also have to now add `moment.js` to the `package.json`? @rexxars
> 
> I'd prefer if we could use [date-fns](https://date-fns.org/v1.29.0/docs/format)@^1.29.0 - we're trying to migrate away from moment, and date-fns is already used by the form builder etc. And yes, it needs to be added to package.json

@rexxars Ok! I've replaced it with `date-fns` and added it to the `package.json` 👍

PR #1916, opened 2020-06-08
[form-builder] Add octet-stream asset source uploader (#2266)

runeb:

- [X]  Bug fix (non-breaking change which fixes an issue)
- [ ]  Maintenance

- [ ]  Yes
- [X]  No

**Current behavior**
Asset source upload silently fails (error in console) when asset is served as `application/octet-stream`.

**Description**

This PR introduces an uploader which will resolve for `application/octet-stream` and also provides a fallback for filenames when mime type is not of `image/*`

**Note for release**
Support `application/octet-stream` in custom asset sources

**Checklist** 

- [X]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [X]  The PR title includes a link to the relevant issue
- [X]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [X]  The code is linted
- [X]  The test suite is passing

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/ImageInput/utils/image.ts
+++ b/packages/@sanity/form-builder/src/inputs/ImageInput/utils/image.ts
@@ -1,17 +1,40 @@
 import UUID from '@sanity/uuid'
 
+function extFromData(string: string): string | undefined {
+  let ext: string
+
+  if (string.match(/^data:application\/octet-stream/)) {
+    ext = 'bin'
+  }
+
+  if (string.match(/^data:image/)) {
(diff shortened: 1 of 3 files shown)


runeb:
A bonus in this changeset is that we no longer call `toString()` on the FileReader result all the time, just when we don't have a filename and need to determine an extension for a constructed filename

runeb:
This is on hold as introducing `application/octet-stream` as an image asset source needs to be better probed in the cases where the data isnt an image

PR #1937, opened 2020-07-09
Show warning when an array contains values with types not defined in the schema

vicbergquist:

**Description**

When an array has values with types not defined in the schema, they're still rendered as items in the array with a broken preview, and when you try to open it, the desk tool crashes.

This adds the same time of warning with details that objects have when an object has unknown fields, and makes it possible to see the value and unset if needed, without crashing the desk tool.

**Before**
Screenshot 2020-07-07 at 10 08 20
Screenshot 2020-07-09 at 10 13 30

**After**
Screenshot 2020-07-09 at 10 09 33

Screenshot 2020-07-09 at 09 44 58


vicbergquist:
Closed. Improved and expanded a solution and pushed to `presence` 👍

PR #1943, opened 2020-07-17
[studio-hints] Fix dependencies

vicbergquist:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

**Description**

**Note for release**


packages/@sanity/studio-hints/package.json
```diff
@@ -37,12 +34,13 @@
   "devDependencies": {
     "@sanity/base": "1.149.18",
     "@sanity/check": "1.149.16",
-    "prop-types": "^15.6.0",
-    "rimraf": "^2.7.1"
+    "rimraf": "^2.7.1",
+    "lodash": "^4.17.15"
```
rexxars, line 38: lodash seems to be used in `src/components/HintsPackage.js`, so should remain a dependency (or, better yet, replace the usage of `isEmpty()` in that file with something like `if (!hintsPackage || !hintsPackage.hintsTitle)`, then remove the dependency altogheter

packages/@sanity/studio-hints/package.json
```diff
@@ -25,10 +25,7 @@
   ],
   "dependencies": {
     "@sanity/block-content-to-react": "^2.0.7",
-    "lodash": "^4.17.15",
-    "next": "^9.1.3",
     "react-copy-to-clipboard": "^5.0.2",
-    "react-icons": "^2.2.7",
```
rexxars, line 31: Two files in this package is using react-icons:
```js
// src/components/ToggleSidecarButton.js
import HelpOutline from 'react-icons/lib/io/ios-help-outline'
import Exit from 'react-icons/lib/io/ios-close-outline'

// src/components/serializers/internalLinkSerializer.js
import DocsIcon from 'react-icons/lib/fa/book'
```

We should either keep `react-icons` as a dependency, or move these icons to be parts and import them from there. Moving this to a peer dependency means each studio using the package should declare it as a dependency, which we don't want.

packages/@sanity/studio-hints/package.json
```diff
@@ -25,10 +25,7 @@
   ],
   "dependencies": {
     "@sanity/block-content-to-react": "^2.0.7",
-    "lodash": "^4.17.15",
-    "next": "^9.1.3",
```
rexxars, line 29: I don't know what the deal is here, but `next` seems to be in use:
```js
// src/components/HintPage.js
import dynamic from 'next/dynamic'

// src/components/serializers/internalLinkSerializer.js
import Link from 'next/link'

// src/components/serializers/Video.js
import dynamic from 'next/dynamic'
```

Not sure why that is. We shouldn't be using next within studio code, obviously.
`dynamic` can be replaced with `React.lazy`, the `Link` seems to not actually be in use, so can be removed.

vicbergquist: Hm. Will take a look

PR #1947, opened 2020-07-20
[form-builder] Hide insert menu item when hidden true

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

**Not sure if intended,** but if a field in a block array has hidden: true, it's still possible to insert it from the insert menu in the PT editor.

```
  {
      name: 'defaults',
      title: 'Content',
      description: 'Profound description of what belongs here',
      type: 'array',
      of: [
        {type: 'image', name: 'image1', title: 'Image, hidden true', hidden: true}, // should be hidden in the menu list
        {type: 'image', name: 'image2', title: 'Image, hidden false'},
        {
          type: 'block',
          of: [
            {
              type: 'color',
              title: 'Color'
            }
          ]
        }
      ]
    },
```

Screenshot 2020-07-20 at 11 03 53

**Description**

Return `null` when `hidden: true` to prevent that item from showing up in the menu.

**Note for release**

**Checklist** 

- [ ]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [x]  The code is linted

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/helpers.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/helpers.tsx
@@ -274,5 +274,8 @@ export function getInsertMenuItems(
     })
   )
 
-  return blockItems.concat(inlineItems)
+  // Do not include items that are supposed to be hidden
+  const filteredBlockItems = blockItems.concat(inlineItems).filter(item => !item.type?.hidden)
+
+  return filteredBlockItems
 }


skogsmaskin:
@vicbergquist - I think the right place to filter those would be here: https://github.com/sanity-io/sanity/blob/presence/packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/helpers.tsx#L241

That's the main source of the insert menu.

vicbergquist:
> @vicbergquist - I think the right place to filter those would be here: https://github.com/sanity-io/sanity/blob/presence/packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/helpers.tsx#L241
> 
> That's the main source of the insert menu.

Fixed 👍

PR #1949, opened 2020-07-22
[POC] Option to have PTE insert menu as dialog and description on items

vicbergquist:

**Current behavior**

https://sanity-io-land.slack.com/archives/CB2F69QBU/p1595335709087000?thread_ts=1595335709.087000&cid=CB2F69QBU

> When a Portable Text editor starts having too many types other than the block, it becomes extremely hard to find which type you want, even if you memorize their icons (see first screenshot). It's a constant, painful look-up based on the tool-tip names...

**Description**

While the feature requests is:

> I'd love to have the possibility to customize the toolbar with React components to create, for example, a rich type-selection dialog with screenshots of what each type could look like on different situations.

Here's a POC (and WIP) for opening the insert menu in the PTE editor in a dialog rather than a dropdown, as this is easier to navigate when the list is very long. Also made it possible to add descriptions to the items.

I added a `insertMenu` option (_edit:_ arrays have a `editModal` prop, maybe this one should be called `insertModal` instead of `insertMenu` to align the wording) to the block type that accepts a string, `dropdown` (default) or `dialog` to determine how the insert menu is rendered. We could also easily add a `fullscreen` option as well.

Screenshot 2020-07-22 at 11 16 27

_Edit1:_
- Maybe `insertModal` makes more sense to use as the option name rather than `insertMenu`. Array items already have a `editModal` option, so `insertModal` here in this case would be more intuitive.

To test it:
1. Go to a `Blocks test` in the test studio 
2. Scroll to the `Customized with block types` field
3. Click `+ Insert`


skogsmaskin:
Sorry @vicbergquist for this has been left unresponded for so long. I think we should take this into account when we have another look at the Toolbar (which we have to do soon).

PR #1951, opened 2020-07-23
[cli] Add a CLI command to copy a dataset

saasen:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

N/A

**Description**

This PR adds a CLI command to start the process of copying a dataset. It also exposes a `--wait` flag to wait for the process to finish.

**Note for release**

Add CLI command to copy a dataset

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -1,0 +1,139 @@
+import EventSource from 'eventsource'
+import Observable from '@sanity/observable/minimal'
+import chalk from 'chalk'
+import promptForDatasetName from '../../actions/dataset/datasetNamePrompt'
+import validateDatasetName from '../../actions/dataset/validateDatasetName'
+import debug from '../../debug'
+
+const helpText = `
+Options
+  --wait Waits for the operation to finish
+
+Examples
+  sanity dataset copy
+  sanity dataset copy <source-dataset>
+  sanity dataset copy <source-dataset> <target-dataset>
+`
+
+const progress = url => {
+  return new Observable(observer => {
+    const progressSource = new EventSource(url)
+
+    function onError(error) {
+      progressSource.close()
+      observer.error(error)
+    }
+
+    function onMessage(event) {
+      const data = JSON.parse(event.data)
+      if (data.state === 'failed') {
+        debug(`Job failed. Data: ${event}`)
+        observer.error(event)
+      } else if (data.state === 'completed') {
+        debug(`Job succeeded. Data: ${event}`)
+        onComplete()
+      } else {
+        debug(`Job progressed. Data: ${event}`)
+        observer.next(data)
+      }
+    }
+
+    function onComplete() {
+      progressSource.removeEventListener('error', onError)
+      progressSource.removeEventListener('channelError', onError)
+      progressSource.removeEventListener('job', onMessage)
+      progressSource.removeEventListener('done', onComplete)
+      progressSource.close()
+      observer.complete()
+    }
+
+    progressSource.addEventListener('error', onError)
+    progressSource.addEventListener('channelError', onError)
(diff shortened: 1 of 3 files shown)


packages/@sanity/client/src/datasets/datasetsClient.js
```diff
@@ -22,6 +22,12 @@ assign(DatasetsClient.prototype, {
     return this.request({uri: '/datasets'})
   },
 
+  copy(name, targetDataset) {
+    validate.dataset(name)
+    validate.dataset(targetDataset)
+    return this.request({method: 'PUT', uri: `/datasets/${name}/copy`, body: {targetDataset}})
+  },
+
```
rexxars, line 30: I'm actually prefer if we could just call `client.request()` in the CLI for this. If we do expose a client method, I think it should listen for progress events and either emit those as an observable, or simply wait for completed task/error and return a promise that resolves/rejects.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -0,0 +1,114 @@
...
+    const client = apiClient()
+
+    const nameError = sourceDataset && validateDatasetName(sourceDataset)
+    if (nameError) {
+      throw new Error(nameError)
+    }
+
+    const [datasets] = await Promise.all([
+      client.datasets.list().then(sets => sets.map(ds => ds.name))
+    ])
```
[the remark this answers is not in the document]
saasen: Yup! Will fix.

packages/@sanity/client/src/datasets/datasetsClient.js
saasen: Ah, what you mean is that we don't create a dedicated `copy` command for this?

saasen: I doubt it would be preferable to clone the `validate` code, as that lives in the `client` package, and I would need this to be done in `core`. Do we have a smart way to use this code from the client, or do I have to copy the functionality (although it isn't very big).

rexxars: > Ah, what you mean is that we don't create a dedicated `copy` command for this?

Yep.

rexxars: > I doubt it would be preferable to clone the `validate` code, as that lives in the `client` package, and I would need this to be done in `core`. Do we have a smart way to use this code from the client, or do I have to copy the functionality (although it isn't very big).

You could either copy it or just leave it up to the API to respond with an error on invalid dataset names - I assume it's doing validation on the server side anyway? :)

saasen: Yes, of course. I'll think about duplicating it and adding it to the client side.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -0,0 +1,135 @@
...
+
+      const spinner = output
+        .spinner({
+          text: `~0% done.`
+        })
+        .start()
+
+      const listenUrl = `${client.config().apiHost}/v1/jobs/${response.jobId}/listen`
+      debug(`Listening on ${listenUrl}`)
+      progress(`${client.config().apiHost}/v1/jobs/${response.jobId}/listen`).subscribe({
```
rexxars, line 116: Maybe use `client.getUrl()` here? And reuse the `listenUrl?` in the `progress` call?

rexxars (approved):
👍   Small suggestion, no showstopper

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
saasen: Haha, very good point!

PR #1957, opened 2020-07-29
[preview] fix type icons not passed down to DefaultPreview for references

vicbergquist:

**Current behavior**

If you have a reference to a type, it doesn't seem to be passed down to the DefaultPreview component.

Reported by the community:
> If I have a reference type field in a block array, how can I customize the icon to match the selected reference? For example, in this case I’d like the icon to use the configured pie chart icon if a pie chart is selected as the reference.

**Description**

This checks if it's a reference and tries to find the referenced type icon. Test it in `Array of named references` here: 
**This branch:** https://test-studio-git-type-icons-for-references-in-preview.sanity-io.now.sh/test/desk/arraysTest;4d7e9f30-20eb-4894-a4f1-c6a4eae2646a
**Current:** https://test-studio-git-current.sanity-io.now.sh/test/desk/arraysTest;4d7e9f30-20eb-4894-a4f1-c6a4eae2646a

Before and after
Screenshot 2020-10-12 at 10 51 53:

Screenshot 2020-10-12 at 10 52 58

**Note for release**

- Fixed type icons not passed down in previews for references in the default preview component (and affecting previews in arrays)

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
+++ b/packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
@@ -1,8 +1,8 @@
 import React from 'react'
 import customResolver from 'part:@sanity/base/preview-resolver?'
 import {get} from 'lodash'
-import SanityDefaultPreview from './SanityDefaultPreview'
 import {Type} from '../types'
+import SanityDefaultPreview from './SanityDefaultPreview'
 
 // Set this to true for debugging preview subscriptions
 const DEBUG = false
@@ -17,7 +17,7 @@ function resolvePreview(type) {
 }
 
 type Props = {
-  snapshot: object
+  snapshot: any
   type: Type
   isLive: boolean
   layout: string
@@ -28,13 +28,14 @@ export default function RenderPreviewSnapshot(props: Props) {
 
   // TODO: Bjoerge: Check for image type with "is()"
   const renderAsBlockImage = layout === 'block' && type && type.name === 'image'
-
+  const typeName = snapshot?._type
+  const icon = (type.to && type.to.find(t => t.name === typeName)?.icon) || type.icon
   const preview = (
     <PreviewComponent // Render media always until we have schema functionality for determining if there is media
       media={() => undefined}
       {...rest}
       value={snapshot}
-      icon={type && type.icon}
+      icon={icon}
       layout={layout}
       isPlaceholder={!snapshot}
       _renderAsBlockImage={renderAsBlockImage}


packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
```diff
@@ -28,13 +28,14 @@ export default function RenderPreviewSnapshot(props: Props) {
 
   // TODO: Bjoerge: Check for image type with "is()"
   const renderAsBlockImage = layout === 'block' && type && type.name === 'image'
-
+  const typeName = snapshot?._type
+  const icon = (type.to && (type.to || []).find(t => t.name === typeName)?.icon) || (type && type.icon)
```
kmelve, line 32: Can `type.to` be something else than an array?

packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
```diff
@@ -17,7 +17,7 @@ function resolvePreview(type) {
 }
 
 type Props = {
-  snapshot: object
+  snapshot: any
```
kmelve, line 20: Doesn't L31 imply that this is an object?

kmelve (changes requested):
I don't exactly feel like the most competent person to review this, but here are my 2 cents at least – I'll let @bjoerge weigh in :)

packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
vicbergquist: It can also be an object I believe?

vicbergquist: When it's just a reference to a single type: `to: {type: 'myType'}`

kmelve: That works? Well, TIL!

vicbergquist: There's a lot of it in the test-studio schemas 😅 It was also news to me

kmelve (commented):
LGTM!

vicbergquist:
@bjoerge 👆

packages/@sanity/preview/src/components/RenderPreviewSnapshot.tsx
bjoerge: Although the schema accepts an object as value for `to`, I'm 99% sure it's normalized to an array after it's been processed/parsed.

bjoerge, line 32: This could be simplified a bit
```suggestion
  const icon = (type.to && type.to.find(t => t.name === typeName)?.icon) || type.icon
```

bjoerge (commented):
Just a small improvement suggestion, otherwise LGTM!

Ideally we should look into fixing the fact that this generic preview component needs to know what type it deals with, but let's save that for later :)

bjoerge (approved):

PR #1965, opened 2020-08-11
[clean-studio] Add missing dependencies

vnys:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When using [PNPM](https://pnpm.js.org/) to install dependencies it can’t find `@sanity/data-aspects` or `@sanity/form-builder` which are being referenced in `config/.checksums` because they are missing from `package.json`.

**Description**

Using packages that are not in `package.json` is possible when using Yarn or NPM because they create a flat `node_modules` structure. PNPM does things differently, by installing dependencies in a global store, and then using hard links. This strategy makes it faster than both Yarn and NPM, and also preserves disk space since a dependency is only downloaded once even though you may have several projects that uses the same package. The downside is that all dependencies for a project must be in package.json – which can be hard to discover when using a flat `node_modules` structure.

**Note for release**

No changes necessary, other than including the updated package.json in the next release.

**Checklist** 

- [x]  The code is linted


rexxars:
These are actually (intentionally) dependencies of `@sanity/desk-tool`, so they shouldn't need to be first-level dependencies of the studio itself

PR #1991, opened 2020-08-31
[base] Fix setState after unmount in hook collection state

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When developing on the studio, the "Hook collection state" utility sometimes re-renders, and some logic is calling a `setTick` method, which attempts to set state on a now unmounted component. This results in the following error in the console:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in GetHookCollectionState (created by RenderActionCollectionState)
    in RenderActionCollectionState
    in Unknown (created by DocumentPane)
    in Popover (created by DocumentPane)
    in DocumentPane
    in DocumentHistoryProvider
```

**Description**

This PR does a hack to work around the bug. I attempted to find the actual source of the issue, but with the amount of throttling, idle callbacks and re-renders, I couldn't figure out what exactly in the dependency chain was causing issues.

By setting a "mounted" flag to true in a mutable ref (and setting to false on unmount), we can check that value before calling the state setter, which seems to fix the issue.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/base/src/actions/utils/GetHookCollectionState.tsx
+++ b/packages/@sanity/base/src/actions/utils/GetHookCollectionState.tsx
@@ -40,6 +40,13 @@ export function GetHookCollectionState<T, K>(props: Props<T, K>) {
   const [, setTick] = React.useState(0)
 
   const [keys, setKeys] = React.useState({})
+  const mountedRef = React.useRef(true)
+
+  React.useEffect(() => {
+    return () => {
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
LGTM

PR #1996, opened 2020-09-03
[@sanity/vision] Use default dataset if loading dataset names from API fails

runeb:

- [X]  Bug fix (non-breaking change which fixes an issue)
- [ ]  Maintenance

- [ ]  Yes
- [X]  No

**Current behavior**
When attempting to use the Vision tool when you are logged in via SSO, a call to the backend for a list of datasets will fail, since the user doesnt have that access level. This causes the Vision tool to end up in an Error message state and does not present its main functionality.

**Description**

This PR simply falls back to the dataset name configured in `sanity.json` in the event that any error occurs while loading the list of datasets. This should be fine, as any other errors would be handled by its child component when attempting to run queries etc.

**Note for release**

Vision tool now works with SSO

**Checklist** 

- [X]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [X]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [X]  The code is linted
- [X]  The test suite is passing

### DIFF
--- a/packages/@sanity/vision/src/containers/VisionContainer.js
+++ b/packages/@sanity/vision/src/containers/VisionContainer.js
@@ -8,25 +8,20 @@ import LoadingContainer from './LoadingContainer'
 class VisionContainer extends LoadingContainer {
   getSubscriptions() {
     return {
-      datasets: {uri: '/datasets'}
+      datasets: { uri: '/datasets' }
     }
   }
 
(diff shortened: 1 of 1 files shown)


rexxars (approved):
👍

PR #1998, opened 2020-09-04
[form-builder] Add setIfMissing patch to createInitialUploadEvent

skogsmaskin:
If you have a nested object inside the PTE which you upload images to, there will be an error because the _upload key is not set yet. This ensures that key is always set.


bjoerge (approved):
LGTM

PR #2008, opened 2020-09-16
[components] Fix use of undeclared react-icons dependency in story

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Breaks if `react-icons` is installed in project at a non-2x version range.

**Description**

A file inside of `@sanity/components` is requiring something from `react-icons` without `@sanity/components/package.json` having declared react-icons as a dependency.

This usually works because of hoisting - `@react/base` has a dependency on `react-icons`, and usually lifts this dependency up to the root, which means `@sanity/components` can find it and use it.

However, when the user installs `react-icons` in their project and the version range does not match ours (2.x), `react-icons` will be installed two places: inside of `@sanity/base/node_modules` (2x) and inside of the studio `node_modules` (3x/4x/whatever range they specified)

Given that this is just a use inside of a story for storybook, the easy option was to use a different icon in this case.

**Note for release**

- Fixed incorrect path for icon in a storybook story

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/components/src/tabs/stories/withIcons.js
+++ b/packages/@sanity/components/src/tabs/stories/withIcons.js
@@ -8,7 +8,7 @@ import React from 'react'
 // Import icons
 import EyeIcon from 'part:@sanity/base/eye-icon'
 import EditIcon from 'part:@sanity/base/edit-icon'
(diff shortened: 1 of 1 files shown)


mariuslundgard (approved):
👍

PR #2010, opened 2020-09-16
[eslint] Tweak ESLint rules for commonly ignored cases

rexxars:
Rationale for rule changes:

- `complexity` - Seems very low, very common to have large React components where breaking them into separate modules only leads to prop-drilling and unnecessarily deep trees. Also quite common to have readable but long validation/switch statements where adjusting to standard only makes code less readable rather than more. Depend on code reviews to catch hard-to-read code. If 30 is not enough for actual cases, ignore in-place.
- `id-length` - `itemA`/`itemB` and similar is not necessarily more readable than `a`, `b`. Lots of exceptions already in place. Commonly disabled. Catch in code reviews if variables are difficultly named.
- `max-depth` - Seems arbitrarily low. Quite common to have valid, deeper blocks than this. Again, catch exceptions in code-reviews rather than linting.
- `max-len` - Tab size seems to be set wrong. Have not seen this play out anywhere, but tweaked for consistency.
- `no-await-in-loop` - Lots of cases where this is useful and "correct" (retrying, for instance). Catch in code reviews if using it in cases where `Promise.all` or a concurrency limiter should be in place.
- `@typescript-eslint/explicit-module-boundary-types` - disable for tsx because defining `React.ReactElement` as return value is annoying. Still a good rule to ensure return values are what you expect in most other cases, however.
- `react/prop-types` - Disable for typescript because we want types, not runtime prop checks
- `react/require-default-props` - Disable for typescript because we want types, not runtime prop checks

Also exposed an `eslint-config-sanity/typescript` preset for easy use in other projects

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,24 +18,15 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier/@typescript-eslint',
     'prettier',
-    'prettier/react'
+    'prettier/react',
+    './packages/eslint-config-sanity/typescript.js'
   ],
   rules: {
-    // --- causing parse errors with d.ts files, see https://github.com/typescript-eslint/typescript-eslint/issues/420
-    'no-useless-constructor': 'off',
-    '@typescript-eslint/no-useless-constructor': 'error',
-    // ---
-    '@typescript-eslint/no-use-before-define': 'off',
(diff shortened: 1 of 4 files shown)


bjoerge (approved):

mariuslundgard (approved):
✨ Thank you ✨

saasen:
Nice work!

PR #2016, opened 2020-09-21
[field] add layout fixed/fluid and enable layout direction

vicbergquist:

- [x]  Yes
- [ ]  No

**Description**

Took a stab at updating the layout to use fluid vs. fixed, as well as adding layout direction horizontal (default) and vertical 👍


vicbergquist:
Should I update this PR to make it work or just close it? @mariuslundgard

PR #2020, opened 2020-09-21
[form-builder] Show warning for array items without `_type`

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

If an array member type has a `name`, but array items exist that do not have a `_type`, they are silently hidden.

Example: https://test-studio-git-next.sanity-io.now.sh/test/desk/arraysTest;untyped-objects

- "Array of multiple types" actually has one array item, but is not displayed

**Description**

This PR checks for the case when a member type cannot be found.

- If there is only a single member type that is valid, it renders an option to set the type name
- If there are multiple member types that are valid, it renders the possible types and an option to remove the item

Example: https://test-studio-git-fix-array-missing-type.sanity-io.now.sh/test/desk/arraysTest;untyped-objects

**Note for release**

- Fixed bug where objects in arrays without a `_type` property would not be displayed

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/ArrayInput/InvalidItem.tsx
+++ b/packages/@sanity/form-builder/src/inputs/ArrayInput/InvalidItem.tsx
@@ -1,33 +1,24 @@
 import React from 'react'
-import {resolveTypeName} from '../../utils/resolveTypeName'
-import InvalidValue from '../InvalidValueInput'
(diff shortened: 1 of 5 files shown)


bjoerge (approved):
Looks great, thank you!

PR #2023, opened 2020-09-24
[cli] Add codemod command with react-icons migration

rexxars:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Description**

Every once in a while we find ourselves in a situation where doing changes to a studio is either required or is encouraged. For instance, we've for a while been encountering issues with people inadvertently depending on `react-icons` as a side-effect of `@sanity/base` using it, as well as our documentation having examples encouraging the usage of it without perhaps clearly stating that you should be adding it as a dependency of the studio.

This PR introduces a new CLI command, "codemod", which allows us to run certain code modifications using  [jscodeshift](https://github.com/facebook/jscodeshift). For now, the only included codemod is one that fixes the import/require statements people might have in their studio to use syntax and paths compatible with react-icons v3. 

Ideally we should be bundling jscodeshift with the CLI and using its programatic API, but because of how it's built (and certain time constraints on our part), I took the cheap route and instead spawn `jscodeshift` through `npx`. We can always clean this up in the future, given that the CLI command remains the same.

**Note for release**

- Add `sanity codemod` CLI command

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/cli/src/commands/codemod/codemodCommand.js
+++ b/packages/@sanity/cli/src/commands/codemod/codemodCommand.js
@@ -1,0 +1,116 @@
+/* eslint-disable no-process-exit, no-sync */
+const childProcess = require('child_process')
+const path = require('path')
+const fs = require('fs')
+const mods = require('../../actions/codemod/mods')
+
+const helpText = `
+Runs a given code modification script on the current studio folder.
+Running the command without a specified codemod name will list available transformations.
+
+Options
+  --dry Dry run (no changes are made to files)
+  --extensions=EXT Transform files with these file extensions (comma separated list)
+                   (default: js,ts,tsx)
+
+Examples
+  # Show available code mods
+  sanity codemod
+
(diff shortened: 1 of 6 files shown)


packages/@sanity/cli/src/commands/codemod/codemodCommand.js
```diff
@@ -0,0 +1,116 @@
...
+const path = require('path')
+const fs = require('fs')
+const mods = require('../../actions/codemod/mods')
+
+const helpText = `
+Runs a given code modification script on the current studio folder.
+Running the command without a specified codemod name will list available transformations.
+
+Options
+  --dry Dry run (no changes are made to files)
```
saasen, line 12: Really nice option, but I've seen many other CLIs call it `--dry-run`. Longer, but maybe consistent? No biggie.

saasen (approved):
LGTM!

PR #2032, opened 2020-10-01
[chore] Upgrade date-fns to v2

vicbergquist:

- [x]  Maintenance

**Description**

Upgrade date-fns to v2.16.1, as part of the discussion at https://sanity-io.slack.com/archives/CUDPSB8AG/p1601502874326400.

What should we do with this? Made a draft PR to RC specifically because of a new useTimeAgo hook.


packages/@sanity/desk-tool/src/components/TimeAgo.js
```diff
@@ -72,7 +74,7 @@ export default class TimeAgo extends React.PureComponent {
   }
 
   render() {
```
rexxars, line 76: Could we refactor this component to a functional component that uses `useTimeAgo` from `@sanity/base/hooks`?

packages/@sanity/base/src/time/useTimeAgo.ts
```diff
@@ -48,33 +48,34 @@ export function useTimeAgo(time: Date | string, {minimal}: TimeAgoOpts = {}): st
 // eslint-disable-next-line complexity
 function formatRelativeTime(date: Date | string, opts: TimeAgoOpts = {}): TimeSpec {
   const now = Date.now()
+  const parsedDate = new Date(date)
```
rexxars, line 51:
```suggestion
  const parsedDate = date instanceof Date ? date : new Date(date)
```

packages/@sanity/base/src/time/useTimeAgo.ts
```diff
@@ -48,33 +48,34 @@ export function useTimeAgo(time: Date | string, {minimal}: TimeAgoOpts = {}): st
 // eslint-disable-next-line complexity
```
rexxars, line 48: I think the complexity here is now allowed by our updated eslint config, could you remove this while you're in here? :)

packages/@sanity/form-builder/src/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -19,7 +19,7 @@ type SchemaOptions = {
   timeStep?: number
   calendarTodayLabel?: string
 }
-const DEFAULT_DATE_FORMAT = 'YYYY-MM-DD'
```
rexxars, line 22: I don't think this is correct - this is a moment.js format string

packages/@sanity/rich-date-input/README.md
```diff
@@ -36,7 +37,7 @@ Typical data output:
 This component accepts the following options via the Sanity schema:
 
 ```
-options.dateFormat || 'YYYY-MM-DD'
```
rexxars, line 39: This module is also using moment, so this isn't the right format here

packages/@sanity/vision/src/util/calendarDate.js
```diff
@@ -2,7 +2,7 @@ import moment from 'moment'
 
 function calendarDate(date) {
   return moment(date).calendar(null, {
-    sameElse: 'YYYY-MM-DD HH:mm:ss'
```
rexxars, line 5: This is also using moment, is this format correct here?

rexxars (changes requested):
There are some locations where the format has been changed where we're actually using moment.js, which I don't think will give the right result. Also, `date-fns` needs to be declared as a dependency in `@sanity/base`

vicbergquist:
> There are some locations where the format has been changed where we're actually using moment.js, which I don't think will give the right result. Also, `date-fns` needs to be declared as a dependency in `@sanity/base`

oh yeah true, we didn't check for that 👍

packages/@sanity/vision/src/util/calendarDate.js
vicbergquist: will leave all formats as they are where moment is being used 👍

rexxars (approved):
👍 Great work :)

PR #2033, opened 2020-10-01
[core] Add "Document" query field, returning any document with given ID

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Currently you have to know which type a document is before querying for it with a `DocumentTypeName` query.

**Description**

This PR introduces a `Document` field (matches the interface name, is a reserved type for Sanity Studio, and matches the other `DocumentTypeName` queries). It has the same, single argument of a document ID, and has the return type of `Document`, which means you'll have to use conditional fragments to query on non-document properties.

In time we should also see if we can make some kind of psuedo-filter for all types, which would allow `allDocument(...)`, but that's a much larger task.

**Note for release**

- Introduce new `Document` field on GraphQL APIs for fetching any document by ID.

**Checklist** 

- [x]  The code is linted


saasen (approved):
LGTM! 👍

PR #2038, opened 2020-10-08
[core] Add support in GraphQL endpoint for `allDocument`

andyrichardson:

- [x]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

A top-level `Document` field [now exists in sanity graphql](https://github.com/sanity-io/sanity/pull/2033) but there is no `allDocument` field.

**Description**

I've added support for `allDocument` as well as the top level `Document` query - with filtering available on all fields of the `Document` type.

You can check out a [working GraphQL deployment here](https://y0wl8twk.api.sanity.io/v1/graphql/production/default)

Screenshot 2020-10-08 at 13 23 08

**Note for release**

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/gen2/filters/documentFilters.js
+++ b/packages/@sanity/core/src/actions/graphql/gen2/filters/documentFilters.js
@@ -1,6 +1,6 @@
 function createDocumentFilters() {
   return {
-    name: 'DocumentFilter',
+    name: 'GlobalDocumentFilter',
     kind: 'InputObject',
     isConstraintFilter: true,
     fields: [
--- a/packages/@sanity/core/src/actions/graphql/gen2/filters/documentFilters.js
+++ b/packages/@sanity/core/src/actions/graphql/gen2/filters/documentFilters.js
@@ -1,6 +1,6 @@
 function createDocumentFilters() {
   return {
-    name: 'GlobalDocumentFilter',
+    name: 'DocumentFilter',
     kind: 'InputObject',
     isConstraintFilter: true,
     fields: [
--- a/packages/@sanity/core/src/actions/graphql/deployApiAction.js
+++ b/packages/@sanity/core/src/actions/graphql/deployApiAction.js
@@ -8,11 +8,13 @@ const SchemaError = require('./SchemaError')
 
 const gen1 = require('./gen1')
 const gen2 = require('./gen2')
+const gen3 = require('./gen3')
 
-const latestGeneration = 'gen2'
+const latestGeneration = 'gen3'
 const generations = {
   gen1,
   gen2,
+  gen3,
 }
 
 module.exports = async function deployApiActions(args, context) {
--- a/packages/@sanity/core/src/actions/graphql/gen2/generateTypeFilters.js
+++ b/packages/@sanity/core/src/actions/graphql/gen2/generateTypeFilters.js
@@ -37,7 +37,9 @@ function generateTypeFilters(types) {
 
   const unionTypes = types.filter((type) => type.kind === 'Union').map((type) => type.name)
   const documentTypes = types.filter(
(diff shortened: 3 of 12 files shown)


packages/@sanity/core/src/actions/graphql/gen2/filters/documentFilters.js
```diff
@@ -1,6 +1,6 @@
 function createDocumentFilters() {
   return {
-    name: 'DocumentFilter',
+    name: 'GlobalDocumentFilter',
```
andyrichardson, line 3: This would be a breaking change in the (likely rare?) case that this is being done:

```gql
query GetAllUser($docFilter: DocumentFilter) {
  allUser(_: $docFilter) {
    _id
  }
}
```

andyrichardson:
One edge case not covered here

Screenshot 2020-10-08 at 13 32 32

Not 100% sure why this is failing as we [explicitly filter out items which don't have a `_type` attribute](https://github.com/sanity-io/sanity/pull/2038/files#diff-684b86eb67ed3df3d68de5851f184ea2R42)

andyrichardson:
@rexxars do you or anyone else know whether this is something we can get added in?

Only thing blocking that I'm aware of is [this issue](https://github.com/sanity-io/sanity/pull/2038#issuecomment-705538733) which I wouldn't be able to fix (the target API isn't open source)

rexxars:
@andyrichardson Sorry for not getting back to you sooner and thanks so much for contributing! I've taken a deep dive into this, and have a couple of thoughts:

- The document filter rename is a potential problem. And _theoretically_ users could have a document type named `GlobalDocument`, which would create a conflict. I'll bring this up internally to see whether or not we think this is a big problem. Maybe we should go with a prefix if we can't guarantee a conflict-free filter name.

- The error you were seeing was actually an incorrect error message (thanks for reporting that, it's now been fixed!) - the _actual_ problem was that it was returning a document type which was not known to the schema (the same could have happened if it didn't implement the document interface). This can happen when you have changed the schema and it finds an old document which isn't in the schema anymore. I've changed the filter from `defined(_type)` to `_type in [$knownDocumentTypes]`

- There was also a small bug in that the document interface was sent to the server as a graphql _type_, which I also fixed.

I've rebased the changes against the latest changes in the repo and force-pushed my changes onto your branch - but we'll still have to figure out if the document filter name change is ok or if we need to introduce a new schema generation so as not to introduce breaking changes.

andyrichardson:
@rexxars thanks so much for the in depth update!

Let me know if you can think of a way around the breaking name change!

saasen:
@andyrichardson: Could you rebase on `next`? When generating the schema using this code, I can see an old bug re-appearing. How old is this branch?

rexxars:
> @andyrichardson: Could you rebase on `next`? 

Rebased and up to date now :)

saasen:
Hey @andyrichardson! 👋 

I did some changes here. We decided to introduce a new generation to avoid friction around the breaking change (the type name change). I reverted back the changes to `gen2` and introduced a new generation `gen3` with your changes. It should now be possible to deploy this with `gen3`, and it would also be more obvious that we have introduced some breaking changes.

saasen:
I'll do some testing to verify that the schema generation still works as intended for both generations.

saasen:
I tested and it works like intended. Fixed a problem where the paths in `gen3` was wrong. Should be fine now!

saasen:
We changed the name of the type `GlobalDocumentFilter` to be `Sanity_DocumentFilter` for a more future-proof approach. 👍

rexxars (approved):
Looks good to me 👍

saasen:
Thanks, @andyrichardson!

PR #2039, opened 2020-10-08
[components, field] fix diff tooltip placement when close to top or bottom of panel

vicbergquist:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When a diff is hovered and it's close to the top or bottom of the change panel, the tooltip will cover the content instead of positioning itself correctly.

**Description**

This will make the tooltip place itself where there's enough space, with placement auto and fallback to top and bottom if needed.

**Note for release**

- Fixed an issue where tooltips in diff components had the wrong placement when close to top or bottom of the changes panel 

**Checklist** 

- [x]  The PR title is appropriately formatted: `[some-package] PR title`
- [x]  The code is linted

### DIFF
--- a/packages/@sanity/components/src/tooltip/tooltip.tsx
+++ b/packages/@sanity/components/src/tooltip/tooltip.tsx
@@ -16,6 +16,8 @@ export interface TooltipProps {
   placement?: TooltipPlacement
   portal?: boolean
   tone?: 'navbar'
+  allowedAutoPlacements?: TooltipPlacement[]
+  fallbackPlacements?: TooltipPlacement[]
 }
 
 export function Tooltip(
@@ -56,6 +58,13 @@ export function Tooltip(
       {
         name: 'offset',
         options: {offset: [0, 3]}
+      },
+      {
+        name: 'flip',
+        options: {
+          allowedAutoPlacements: props.allowedAutoPlacements,
+          fallbackPlacements: props.fallbackPlacements
+        }
       }
     ]
   })
--- a/packages/@sanity/field/src/diff/components/DiffTooltip.tsx
+++ b/packages/@sanity/field/src/diff/components/DiffTooltip.tsx
@@ -36,7 +36,7 @@ export function DiffTooltip(props: DiffTooltipProps | DiffTooltipWithAnnotations
 }
 
 function DiffTooltipWithAnnotation(props: DiffTooltipWithAnnotationsProps) {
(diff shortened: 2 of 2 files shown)


mariuslundgard:
I think we should make both the `allowedAutoPlacements` and `fallbackPlacements` configurable aspects of the tooltip – because we need different settings in other contexts.

My proposal:

```ts
// Add these proptypes to TooltipProps interface:
allowedAutoPlacements: Placement[]
fallbackPlacements: Placement[]
```

And then in the `flip` modifier:
```ts
    {
      name: 'flip',
      options: {
        allowedAutoPlacements: props.allowedAutoPlacements,
        fallbackPlacements: props.fallbackPlacements
      },
    },
```

Now it can be configured specifically for diff tooltips in the `DiffTooltipWithAnnotation`:

```tsx
    <Tooltip content={content} placement='top' fallbackPlacements={['top', 'bottom']} portal>
      {children}
    </Tooltip>
```

vicbergquist:
@mariuslundgard ok, pushed a fix to make that work. Is it ok?

vicbergquist:
Actually I guess we don't need the `fallbackPlacement` in `DiffAnnotationTooltip` then

mariuslundgard:
I think `fallbackPlacement` is better to use. But either way works 👍

packages/@sanity/components/src/tooltip/tooltip.tsx
```diff
@@ -56,6 +58,13 @@ export function Tooltip(
       {
         name: 'offset',
         options: {offset: [0, 3]}
+      },
+      {
+        name: 'flip',
+        options: {
+          allowedAutoPlacements: props.allowedAutoPlacements || ['top', 'bottom'],
```
mariuslundgard, line 65: Here we should remove the `|| ['top', 'bottom']` part. Otherwise it’s not possible to get the default setting without defining all possible placements?

packages/@sanity/components/src/tooltip/tooltip.tsx
```diff
@@ -56,6 +58,13 @@ export function Tooltip(
...
+          fallbackPlacements: props.fallbackPlacements || ['top', 'bottom']
```
mariuslundgard, line 66: Same here.

mariuslundgard (approved):
Thank you for fixing this, Vic! 💯

PR #2055, opened 2020-10-13
[base] Allow connector match to shorter paths

benedicteb:
For portable text fields that aren't full size you want the keyed
objects to resolve to the text field itself. Otherwise we won't display
connectors for non-fullsize portable text editors.

Can do this by making sure that the target and path only differ by
being:

* 1 path segment shorter
* The last path segment is a keyed object

Then only perform the length-check where this doesn't apply.

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

recording_2020-10-13-13:35:22

**Description**

After the fix the connectors will render correctly.

recording_2020-10-13-13:36:22

**Note for release**

n/a

**Checklist** 

- [n/a]  The PR title includes a link to the relevant issue
- [x]  The code is linted
- [n/a]  Corresponding changes to the documentation have been made

### DIFF
--- a/packages/@sanity/base/src/change-indicators/helpers/findMostSpecificTarget.ts
+++ b/packages/@sanity/base/src/change-indicators/helpers/findMostSpecificTarget.ts
@@ -1,4 +1,5 @@
 import * as PathUtils from '@sanity/util/paths'
+import {isKeyedObject} from '@sanity/types'
 import {TrackedChange, TrackedArea} from '../'
 
 export function findMostSpecificTarget(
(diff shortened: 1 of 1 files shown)


bjoerge:
Looks good! Only thing I can think of is whether this can potentially affect items in array of objects where we actually want to draw a line from a changed (keyed) item?

benedicteb:
> Looks good! Only thing I can think of is whether this can potentially affect items in array of objects where we actually want to draw a line from a changed (keyed) item?

Good point! I'll add some kind of mechanism to make sure we always use the most precise path possible :)

benedicteb:
> Looks good! Only thing I can think of is whether this can potentially affect items in array of objects where we actually want to draw a line from a changed (keyed) item?

So I did some testing on this now - it looks like for arrays you *usually* use this part:

```typescript
const path = PathUtils.fromString(id.slice(id.indexOf('-') + 1))
const exactId = `${targetType}-${PathUtils.toString(path)}`

if (values.has(exactId)) {
  return values.get(exactId) as TrackedChange
}
```

Which isn't affected by this change. I tested for arrays of:

* strings
* numbers
* slugs

Should I still make a mechanism for making sure the `mostSpecific` you end up with actually is the `mostSpecific`?

bjoerge (approved):

PR #2056, opened 2020-10-13
[base, desk-tool] Update useTimeAgo hook to make trailing `ago` optional, and update sparkline to use minimal timestamp

vicbergquist:

**Description**

Noticed that the sparkline doesn't currently use the minimal timestamp format, which sometimes makes it hard to read + causing layout issues when document was only published or updated a while ago. Also didn't match the Figma sketch.

This uses the minimal format for the timestamp in the sparkline, but also adds an option called `useAgo` to the `useTimeAgo` hook to make it possible to use both the default and minimal format with and without a trailing `ago`. 

The components using the `useTimeAgo` hook with the minimal format have been updated to use the appropriate `ago` as well.

**Note for release**

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/base/src/time/useTimeAgo.ts
+++ b/packages/@sanity/base/src/time/useTimeAgo.ts
@@ -22,25 +22,26 @@ const ONE_HOUR = ONE_MINUTE * 60
 
 interface TimeAgoOpts {
   minimal?: boolean
+  agoSuffix?: boolean
 }
 
-export function useTimeAgo(time: Date | string, {minimal}: TimeAgoOpts = {}): string {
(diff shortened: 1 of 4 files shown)


mariuslundgard:
Looks good! The only issue I have is with the option name. By having `use` in the name, it can be confused with a React hook name. How about renaming to e.g. `agoSuffix`?

vicbergquist:
> Looks good! The only issue I have is with the option name. By having `use` in the name, it can be confused with a React hook name. How about renaming to e.g. `agoSuffix`?

that's true! `agoSuffix` is good!

vicbergquist:
Updated 👍

mariuslundgard (approved):
LGTM

PR #2059, opened 2020-10-14
[base] Switch to a warning instead of hard failure on missing tracker context

bjoerge:
Do not delete the instructional comments. -->

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**
When using the `useReporter()` or `useReportedValues()` hook outside of a Tracker element context we currently throw an error which makes the studio crash. Since this is currently used for non-critical features like the change connectors and presence we should hard fail when it happens.

**Description**
 This PR issues a console warning instead of crashing the studio. A future improvement would be to make the failure mode configurable on a per-tracker basis, since there might be situations where it's used for more critical features.

**Note for release**
Added a more graceful handling of cases where no tracker context could be found by issuing a console warning instead of throwing an error, causing studio crash.

### DIFF
--- a/packages/@sanity/base/src/components/react-track-elements/createScope.tsx
+++ b/packages/@sanity/base/src/components/react-track-elements/createScope.tsx
@@ -2,24 +2,54 @@ import * as React from 'react'
 import {TrackerContext} from './types'
 import {createUseReporter, IsEqualFunction} from './createUseReporter'
 import {createStore} from './createStore'
+import {Reported} from './index'
 
-const useReporterGuard = (id: string) => {
-  throw new Error(
-    `No context provided for reporter. Make sure that the component calling "useReporter(${id}, ...)", is wrapped in a <Tracker> element`
-  )
+// Todo: consider memozing individual functions or move the context assertion/guard to a separate step.
+let didWarn = false
+const useReporterGuard = (id: string): void => {
+  if (!didWarn) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      new Error(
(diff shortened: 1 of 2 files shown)


packages/@sanity/base/src/components/react-track-elements/createScope.tsx
```diff
@@ -2,24 +2,53 @@ import * as React from 'react'
 import {TrackerContext} from './types'
 import {createUseReporter, IsEqualFunction} from './createUseReporter'
 import {createStore} from './createStore'
+import {Reported} from './index'
 
-const useReporterGuard = (id: string) => {
-  throw new Error(
-    `No context provided for reporter. Make sure that the component calling "useReporter(${id}, ...)", is wrapped in a <Tracker> element`
-  )
+let didWarn = false
```
benedicteb, line 7: I'm sure this isn't a problem at all, just noting some thoughts:

Are we sure that the value isn't initialized more than once in some weird way since it's put there in the outer scope? If more than one thing imports this file?

We could move the initialization into the function we're memoizing to have complete control over when it's initialized :)

benedicteb (approved):
Looks good to me! :) 

Just wish we could avoid those eslint-exceptions, but that'll probably be a bit tricky!

packages/@sanity/base/src/components/react-track-elements/createScope.tsx
bjoerge: The script is only executed once even if it's imported/require'd several times, so can't really see how that can happen. That said, it's now a bit random what error you'll see, so memoizing the individual functions and show errors for all of them could be a worthwhile improvement. Let's make a note of that and revisit later?

PR #2061, opened 2020-10-14
[base] Various visual and stability improvements to the connectors overlay

bjoerge:
- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Description**

This provides numerous visual and stability improvements to the connectors overlay.
- Using @mariuslundgard's new connectors drawing algorithm, fixing a bunch of small rendering glitches
- Increased the hit box/click area around the connector line and adds a hover effect so it's more apparent
- Improves how we detect scrollable containers and calculate bounding areas for connectors.
- Lot's of smaller fixes


packages/@sanity/base/src/change-indicators/overlay/Connector.css
```diff
@@ -0,0 +1,59 @@
...
+  stroke-linejoin: round;
+  opacity: 0;
+  @nest &:hover {
+    opacity: 0.1;
+  }
+}
+
+.rightBar {
+  stroke: none;
+  fill: var(--brand-primary);
```
mariuslundgard, line 45: This variable should be `var(--focus-color)`

packages/@sanity/base/src/change-indicators/overlay/Connector.css
```diff
@@ -0,0 +1,59 @@
...
+  pointer-events: none;
+}
+
+.rightBarFocused {
+  fill: var(--brand-primary);
```
mariuslundgard: Same as above.

packages/@sanity/base/src/change-indicators/overlay/Connector.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.hovered {
+  stroke: var(--state-warning-color);
+}
+
+.interactivePath {
+  fill: none;
+  pointer-events: stroke;
+  stroke: var(--brand-primary);
```
mariuslundgard: This variable should be var(--focus-color)

packages/@sanity/base/src/change-indicators/overlay/Connector.css
```diff
@@ -0,0 +1,59 @@
...
+.path {
+  fill: none;
+  stroke: #ccc;
+  pointer-events: none;
+  stroke-linecap: round;
+  stroke-linejoin: round;
+}
+
+.focused {
+  stroke: var(--brand-primary);
```
mariuslundgard: This variable should be var(--focus-color)

mariuslundgard (changes requested):

packages/@sanity/base/src/change-indicators/overlay/Connector.css
bjoerge: Thanks! I've fixed all of these now.

packages/@sanity/base/src/change-indicators/helpers/getElementGeometry.ts
```diff
@@ -0,0 +1,10 @@
+import getOffsetsTo from './getOffsetsTo'
+import {Rect} from '../overlay/types'
+
+export function getElementGeometry(element, parent): {rect: Rect; bounds: Rect} {
```
benedicteb, line 4: Are the arguments missing types or is typescript okay with it? :)

```suggestion
export function getElementGeometry(element: HTMLElement, parent: HTMLElement): {rect: Rect; bounds: Rect} {
```

packages/@sanity/base/src/change-indicators/overlay/ConnectorsOverlay.tsx
```diff
@@ -176,122 +90,60 @@ export const ConnectorsOverlay = React.memo(function ConnectorsOverlay(props: Pr
...
+                      from={{rect: field.rect, bounds: field.bounds}}
+                      to={{rect: change.rect, bounds: change.bounds}}
+                      focused={hasFocus}
+                      hovered={hasHover || isHoverConnector}
+                      revertHovered={hasRevertHover}
                     />
+                  </g>
+                  {DEBUG_LAYER_BOUNDS && (
+                    <g style={{pointerEvents: 'none'}}>
+                      <rect
```
benedicteb, line 109: Perhaps this could be made into a sub-component as well to make the main component a bit more expressive? :) 

```typescript
const DebugRects = (field, change) => (
  <g style={{pointerEvents: 'none'}}>
    <Rect points={field.bounds} />
    <Rect points={field.rect} />
    ...
  </g>
)

const SvgRect = (points: Rect) => (
  <rect x={points.left} .... />
)
```

packages/@sanity/base/src/change-indicators/overlay/connectorPath.ts
```diff
@@ -0,0 +1,153 @@
+import {
+  ARROW_MARGIN_X,
+  ARROW_MARGIN_Y,
+  ARROW_SIZE,
+  ARROW_THRESHOLD,
+  CORNER_RADIUS
+} from '../constants'
+import {ConnectorLine} from './types'
+
+export function arrowPath(x: number, y: number, dir: number) {
```
benedicteb, line 10: It makes me so happy to see all these things get to live in their own files :heart_eyes: :heart_eyes:

packages/@sanity/base/src/change-indicators/overlay/connectorPath.ts
```diff
@@ -0,0 +1,153 @@
...
+    )
+  } else {
+    cmds.push(moveTo(fromX, fromY))
+  }
+
+  // TO
+  if (to.isAbove) {
+    if (fromY < to.bounds.top) {
+      cmds.push(
+        lineTo(to.bounds.left - 8 - r1, fromY),
```
benedicteb, line 65: Something we could maybe do to make this piece of code even more easier to read is perhaps to divide up into "generate"-style functions so that we can have a specific name for each *segment* of the path. That way this could be something like:

```typescript
if (to.isAbove) {
  if (fromY < to.bounds.top) {
    lineSegments.push(
      generateCorner(from, to, ...)
    )
```

Basically "grouping" curves and lines into segments so we can reuse more and have more expressive names :)

packages/@sanity/base/src/change-indicators/overlay/types.ts
```diff
@@ -0,0 +1,28 @@
+export interface ConnectorLinePoint {
```
benedicteb, line 1: Can't we put types in `.d.ts` files and that way get to use them in the project without having to import them? :)

benedicteb (approved):
Looks very good to me! It'll be so nice to get the connector code so much more neat :D 

Made some comments about some thoughts I had, but if we want some of those changes it might be wise to do it in separate later PRs :)

packages/@sanity/base/src/change-indicators/helpers/getElementGeometry.ts
bjoerge: Great point! Eslint actually complains here, so will fix :)

The reason typescript is happy is because strict mode is off in this project (due to legacy reasons). I'll check how much work it is to enable it only for this subfolder.

packages/@sanity/base/src/change-indicators/overlay/ConnectorsOverlay.tsx
bjoerge: That's a very good idea. Let me see if I can sneak in a quick refactor.

packages/@sanity/base/src/change-indicators/overlay/connectorPath.ts
bjoerge: Yes, totally agree. Let's keep it as a future improvement?

packages/@sanity/base/src/change-indicators/overlay/types.ts
bjoerge: Hmm... I think the same reasons for preferring explicit imports over implicit globals applies to types as much as it does for values?

mariuslundgard (approved):
LGTM! Thanks for doing this!

packages/@sanity/base/src/change-indicators/overlay/types.ts
benedicteb: I just didn't know preferring explicit imports of types was a thing - but it makes sense :) No worries!

packages/@sanity/base/src/change-indicators/helpers/getElementGeometry.ts
benedicteb: Oh I see! I didn't know that :)

PR #2064, opened 2020-10-15
Connectors redraw on resize

bjoerge:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Currently the connector lines are not redrawn/recalculated on resize.

**Description**
This makes sure the connector lines are being redrawn properly on window resize.

There are three separate commits that can be reviewed individually:
- a9718b8 Exports a single resizeObserver instance that can be shared across different studio components (there are performance gains to be had from using a single instance instead of creating multiple). It's also exported with a bit nicer API that fits React hooks a bit better than the native ResizeObserver API.
- 4825c6c Makes desk-tool import the shared resizeObserver instance from base instead of creating its own.
- 55e0e5f Makes use of the same resizeObserver instance to force redraw the connector lines on resize.

**Note for release**
Fixes an issue with lines connecting a change with it's corresponding input not being redrawn on window/pane resize.

**Checklist** 

- [ ]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [ ]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [ ]  The code is linted

### DIFF
--- a/packages/@sanity/base/package.json
+++ b/packages/@sanity/base/package.json
@@ -65,6 +65,7 @@
     "shallow-equals": "^1.0.0"
   },
   "devDependencies": {
+    "@types/resize-observer-browser": "^0.1.4",
     "jest": "^26.4.2",
     "prop-types": "^15.6.0",
     "react": "16.9.0",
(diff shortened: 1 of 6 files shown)


benedicteb (approved):
Looks good to me :) 

Only thing I'm thinking of is perhaps moving the resize observer stuff somewhere more general where more things can use it?

bjoerge:
> Only thing I'm thinking of is perhaps moving the resize observer stuff somewhere more general where more things can use it?

Yep, we can consider that. Not sure yet if this should be considered public API, so I think we can live with importing it from `@sanity/base/lib/util` as we do in desk-tool for now.

Thanks for reviewing!

PR #2104, opened 2020-11-03
[chore] Remove unnecessary internal dependency

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

These three modules are explicitly listing `@sanity/base` as a dependency, when they're actually only using the parts system to import. That means the additional dependency is a potential for version mismatches, which has been happening now and then.

**Description**

This PR removes the direct dependency on `@sanity/base`, hopefully solving the problem where the modules would not get a bumped version of `@sanity/base` when publishing new versions, and reducing cross-dependency chatter.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/dashboard/package.json
+++ b/packages/@sanity/dashboard/package.json
@@ -6,9 +6,7 @@
   "author": "Sanity.io <hello@sanity.io>",
   "license": "MIT",
   "scripts": {
-    "clean": "rimraf lib",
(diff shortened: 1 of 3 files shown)


skogsmaskin (approved):
Good riddance

bjoerge (approved):

PR #2106, opened 2020-11-04
[core] Support React 17

rexxars:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

The studio fails to start when using React 17, with an unhelpful message to "upgrade" to React 16.9.

**Description**

This PR changes the peer dependency ranges for packages to allow for React 17, upgrades the dev dependencies to use React 17, as well as explicitly adding support for it in the CLI startup check. The checks were very generalized, but given that it only runs checks for react and react-dom, I also changed from blocking the studio from starting to instead giving a warning on startup, if the version found is newer than what is explicitly defined as "supported".

Also found a couple of places where we had peer dependency requirements on outdated versions of `@sanity/base`, which I fixed.

Hard to give the entire studio a thorough test, but from the testing I did things seems to be working as expected.

**Note for release**

- Add support for React 17

**Checklist** 

- [x]  The code is linted


bjoerge (approved):
Great addition to allow for untested with a warning 💯

PR #2116, opened 2020-11-10
[chore] Improve performance of webpack development server

mariuslundgard:

- [ ]  Maintenance
- [x]  Other, please describe: **Improved tooling**

- [ ]  Yes
- [x]  No

**Current behavior**

The gulp script for running studios is slow and consumes a lot of CPU and memory because it watches and rebuilds JS/TS, assets and type definitions in a many packages.

**Description**

1. These changes eliminate the need for watching and rebuilding each package by configuring webpack to read directly from source files (in `src` directories). This way, e.g. the `npm run design-studio` script only starts a single webpack process.

2. When working in an IDE, TS is now also configured to read from directly from `src` directories, so that watching/rebuilding TS definitions in no longer necessary.

**Note for release**

**n/a**

**Checklist** 

- [x]  The code is linted


packages/@sanity/base/components.d.ts
```diff
@@ -1 +1,3 @@
-export * from './src/components'
+import * as components from './src/components'
+
+export = components
```
bjoerge, line 3: Why is this needed? Won't this require the consuming end to have `esModuleInterop` enabled?

packages/@sanity/base/src/presence/FieldPresence.tsx
```diff
@@ -31,7 +31,7 @@ function FieldPresenceWithOverlay(props: FieldPresenceProps) {
   const {presence = contextPresence, maxAvatars = DEFAULT_MAX_AVATARS_FIELDS} = props
   const ref = React.useRef(null)
 
-  useReporter(useId(), () => ({presence, element: ref.current!, maxAvatars: maxAvatars}))
+  useReporter(useId() || '', () => ({presence, element: ref.current!, maxAvatars: maxAvatars}))
```
bjoerge, line 34: Why do we need the fallback to empty string here? If useId() can return a non-string I think it's better if we throw a descriptive error instead. The id's here needs to be unique, and this change is likely to yield an error further down the line that becomes hard to trace.

packages/@sanity/base/src/presence/PresenceTooltip.tsx
```diff
@@ -34,7 +32,7 @@ export function PresenceTooltip(props: PresenceTooltipProps) {
 
   return (
     <Tooltip content={content} placement={placement}>
-      {children}
+      {children as any}
```
bjoerge, line 35: Can we remove the `any` cast here if we change the type of `children` to React.ReactNode?

packages/@sanity/base/tsconfig.dist.json
```diff
@@ -0,0 +1,24 @@
...
+        "options": {
+          "customMatcher": "\\.css$"
+        }
+      }
+    ],
+
+    "baseUrl": "./src",
+    "paths": {
+      "part:@sanity/base/datastore/history": ["./datastores/history"],
+      "part:@sanity/base/user": ["./datastores/user"]
```
bjoerge, line 21: Should more of the datastore parts be aliased here? Or, what is the reason we only alias these?

packages/@sanity/base/user-color.d.ts
```diff
@@ -1 +1,3 @@
-export * from './src/user-color'
+import * as userColor from './src/user-color'
+
+export = userColor
```
bjoerge, line 3: Same as comment about `esModuleInterop` above.

packages/@sanity/google-maps-input/src/loader/LoadError.tsx
```diff
@@ -17,7 +17,7 @@ export function LoadError(props: Props) {
           <>
             <h3>Error details:</h3>
             <pre>
-              <code>{props.error.message}</code>
+              <code>{props.error?.message}</code>
```
bjoerge, line 20: This is just TypeScript not being clever enough to know that `props.error` must be defined in this branch, so I believe you can safely use `props.error!.message` here instead.

packages/@sanity/util/test/PathUtils.test.ts
```diff
@@ -24,7 +24,7 @@ test('fromString: throws if not a string', (t) => {
   t.throws(() => fromString(), 'Path is not a string')
   // @ts-ignore
   t.throws(() => fromString(13), 'Path is not a string')
-  t.throws(() => fromString(null), 'Path is not a string')
+  t.throws(() => fromString(null as any), 'Path is not a string')
```
bjoerge, line 27: Better to use `// @ts-expect-error` instead of casting to any here

packages/@sanity/util/test/PathUtils.test.ts
```diff
@@ -84,7 +84,7 @@ test('toString: throws if not an array', (t) => {
   t.throws(() => toString(), 'Path is not an array')
   // @ts-ignore
   t.throws(() => toString(13), 'Path is not an array')
-  t.throws(() => toString(null), 'Path is not an array')
+  t.throws(() => toString(null as any), 'Path is not an array')
```
bjoerge, line 87: Same as above, `// @ts-expect-error` FTW

bjoerge (commented):
Overall this is looks really good. Great work! Looking forward to have it merged.

Left a few comments while reading through the diffs. On a general note, it's a bit unclear to me why we need the tsconfig.dist.json's in the individual packages. This brings a lot of duplication between the "non-dist" and the "dist"-configs which makes me worry a bit about the added maintenance burden. Could the `dist` variant extend from the non-dist and override only the options it needs to, or are there  alternative ways of setting this up that we could consider?

packages/@sanity/base/components.d.ts
mariuslundgard: It's using the `export =` because that's how to define CommonJS export (afaik). See the `components.js` that this def is referring to:

```js
module.exports = require('./lib/components')
```

packages/@sanity/base/src/presence/FieldPresence.tsx
mariuslundgard: Did this to make the TS type check to pass. Will try your suggestion 👍

packages/@sanity/util/test/PathUtils.test.ts
```diff
@@ -24,7 +24,8 @@ test('fromString: throws if not a string', (t) => {
   t.throws(() => fromString(), 'Path is not a string')
   // @ts-ignore
   t.throws(() => fromString(13), 'Path is not a string')
-  t.throws(() => fromString(null as any), 'Path is not a string')
+  // @ts-ignore
```
bjoerge, line 27: Why not `// @ts-expect-error` instead of `// @ts-ignore`?

mariuslundgard: I copied what was used elsewhere in the file, for consistency.

packages/@sanity/base/tsconfig.dist.json
mariuslundgard: These are the parts which `@sanity/base` itself depends on. This could also be defined in `src/@types/parts.d.ts`, but it's less verbose when defined in tsconfig.json.

bjoerge (approved):

PR #2118, opened 2020-11-10
[chore] Build modules before running lint

rexxars:
The `import/no-unresolved` rule will fail if you depend on another module in the monorepo which hasn't yet been built. Thus; we need to actually build all modules before running the linter.

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

eslints `import/no-unresolved` fails when depending on sibling module, since it points to a compiled version which is not yet available.

**Description**

This PR makes sure we build the modules prior to running the linter. Not ideal, but can't see a good workaround right now.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,6 +24,12 @@ jobs:
 
       - name: Install npm dependencies
         run: npm install
+      
+      # Required for import lint rules to work - packages pointing to ex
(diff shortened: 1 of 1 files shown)


benedicteb (approved):
Great! Thanks for this :smile:

PR #2145, opened 2020-11-19
[client] Add dataset alias support in client

saasen:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

The Sanity JS client currently crashes when using an alias in the `dataset` field.

**Description**

This PR let's the alias format through the dataset validator. Aliases starts with the `~` character, and isn't currently allowed.

**Note for release**

Add support for using aliases in the Sanity JS client

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/client/src/validators.js
+++ b/packages/@sanity/client/src/validators.js
@@ -2,7 +2,7 @@ const VALID_ASSET_TYPES = ['image', 'file']
 const VALID_INSERT_LOCATIONS = ['before', 'after', 'replace']
 
 exports.dataset = name => {
-  if (!/^[~][a-z0-9][-\w]{2,20}$|^[a-z0-9][-\w]{1,20}$/.test(name)) {
+  if (!/^(~[a-z0-9]{1}[-\w]{0,25}|[a-z0-9]{1}[-\w]{0,19})$/.test(name)) {
     throw new Error(
       'Datasets can only contain lowercase characters, numbers, underscores and dashes, and start with tilde, and be maximum 20 characters'
     )
--- a/packages/@sanity/client/src/validators.js
+++ b/packages/@sanity/client/src/validators.js
@@ -2,9 +2,9 @@ const VALID_ASSET_TYPES = ['image', 'file']
 const VALID_INSERT_LOCATIONS = ['before', 'after', 'replace']
 
 exports.dataset = name => {
-  if (!/^[-\w]{1,128}$/.test(name)) {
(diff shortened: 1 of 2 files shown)


packages/@sanity/client/src/validators.js
```diff
@@ -2,9 +2,9 @@ const VALID_ASSET_TYPES = ['image', 'file']
 const VALID_INSERT_LOCATIONS = ['before', 'after', 'replace']
 
 exports.dataset = name => {
-  if (!/^[-\w]{1,128}$/.test(name)) {
+  if (!/^[~][a-z0-9][-\w]{2,20}$|^[a-z0-9][-\w]{1,20}$/.test(name)) {
```
benedicteb, line 5: We're reducing the maximum allowed length of a dataset name by quite a lot. Are we sure that's safe to do without investigating whether users depend on long dataset names or something? :smiley:

saasen: I asked Espen about this, since I've seen various different validations for dataset. He said max 20 chars, but please confirm again, @rexxars!

rexxars: The API has max 20 characters as the limit, so should be safe!

rexxars: The regex is a bit off though - aliases (should) need between 1 and 26 characters, where the leading character cannot be a dash, eg `^~[a-z0-9]{1}[-\w]{0,25}$`. Datasets are between 1 and 20 characters, eg `^[a-z0-9][-\w]{0,19}`. I believe this catch both cases:

`^(~[a-z0-9]{1}[-\w]{0,25}|[a-z0-9]{1}[-\w]{0,19})$`

rexxars (changes requested):
Apart from the regex this looks good!

packages/@sanity/client/src/validators.js
rexxars, line 5: The regex is a bit off though - aliases (should) need between 1 and 26 characters, where the leading character cannot be a dash, eg `^~[a-z0-9]{1}[-\w]{0,25}$`. Datasets are between 1 and 20 characters, eg `^[a-z0-9][-\w]{0,19}`. I believe this catch both cases:

```suggestion
  if (!/^(~[a-z0-9]{1}[-\w]{0,25}|[a-z0-9]{1}[-\w]{0,19})$/.test(name)) {
```

rexxars (changes requested):

packages/@sanity/client/src/validators.js
saasen: Thanks for pointing that out. I'll fix that.

rexxars (approved):
👍

PR #2149, opened 2020-11-25
[chore] Switch to project references instead of path aliases for TypeScript compilation

bjoerge:
This PR builds on #2150 (sorry about the misleading branch name).

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**
Currently, TypeScript in this repository are compiled during development from each package's source files using path aliases. This has a couple of suboptimal side effects, the most important one being that source files from one package is compiled with the compiler options from the _consuming_ package. This leads to really difficult error messages, and also issues where you get compile errors during build that you won't see in your editor, because the the TypeScript language service running in the IDE compiles it with the source file's "own" tsconfig.

**Description**
This PR removes the path aliases and instead use project references (based on the example here https://github.com/RyanCavanaugh/learn-a) which has a number of advantages over the current approach:

- All packages is compiled in a single (incremental) process, internal dependencies are taken care of and compiled in the correct order (no more manual dependency topology managing needed from us)
- Every package is compiled in isolation, making it easier for us to progressively adopt strict mode, have individual compiler options for each package (e.g. `jsx: "react"`, `allowJs: true`, etc).

More about project references here: https://www.typescriptlang.org/docs/handbook/project-references.html

**Other implications**
- Every package's tsconfig.json must now have `composite: true` in it's compilerOptions. The packages that depend on other TypeScript packages must declare their dependencies in the `references`-array (see this example from [`@sanity/base`](https://github.com/sanity-io/sanity/blob/5ab2b4ddaa573043d40660dbf7ec7ed105179ada/packages/@sanity/base/tsconfig.json#L23-L45) )
- The `type-check` scripts are removed from the individual `package.json`s, instead to type check a single package, use `npx tsc -b <path to package>` , e.g `npx tsc -b packages/@sanity/base`.
- Project references doesn't work well with circular dependencies, so we have to avoid these (which is a good thing in any case)

**Known issues**
- I noticed during development that sometimes the incremental build info needs to be reset. (mainly seems to happen after modifying `tsconfig.json`s). I've used `find . -name 'tsconfig.tsbuildinfo' -exec rm {} \;` to clear the build info cache. If this becomes a problem I'll look into better ways of mitigating this.


mariuslundgard:
Wow, did you write all those part defs by hand? Massive amount of work! 🙌

Firstly – I hope we can get this approach to work. Would of course be better if we could use TS's provided method for references between packages.

That said, I currently get a long list of errors when running:

```sh
npm run clean
rm -rf node_modules
npm install
npm start
```

When running `npm start`, this warning message is showing:
```
WARNING: failed to load paths from tsconfig.json in monorepo undefined
```

And later webpack complains of many packages that are not found, e.g.:
```
Error in sanity/packages/@sanity/base/src/schema/createSchema.js (part:@sanity/base/schema-creator)
Module not found: Error: Can't resolve '@sanity/validation' in 'sanity/packages/@sanity/base/src/schema'
 @ sanity/packages/@sanity/base/src/schema/createSchema.js (part:@sanity/base/schema-creator 11:18-47)
 @ ./schemas/index.ts (part:@sanity/base/schema)
 @ sanity/packages/@sanity/default-layout/src/schemaErrors/SchemaErrorReporter.tsx
 @ sanity/packages/@sanity/default-layout/src/DefaultLayout.tsx
 @ sanity/packages/@sanity/default-layout/src/Root.tsx (part:@sanity/base/root)
 @ sanity/packages/@sanity/base/src/components/SanityRoot.tsx (part:@sanity/base/sanity-root)
 @ sanity/packages/@sanity/server/lib/browser/entry-dev.js
 @ multi sanity/node_modules/normalize.css/normalize.css sanity/packages/@sanity/server/lib/browser/entry-dev.js
```

---

Some questions:

- How will webpack know which files to watch for rebuilding? (which prior to this PR is directly from the `src` directories)
- Can we get IDE linking to work, so that when you click a path (e.g. `@sanity/base/components`) it will open the source file?

bjoerge:
> Did you write all those part defs by hand? Kudos 🙌

Wrote a script to extract the complete list from the source code and consolidated the ones we already had by hand 😅 

Good point about webpack, missed that bit. I've restored the path aliases for internal modules (except parts) in the root tsconfig. This should make webpack happy.

> Can we get IDE linking to work, so that when you cmd-click a path (e.g. @sanity/base/components) it will open the source file?

This should also work again now. Good catch!

Looking into the lint errors.

packages/@sanity/default-layout/package.json
```diff
@@ -6,9 +6,7 @@
   "scripts": {
     "coverage": "nyc --report=lcov --report=text _mocha -- test/**/*.test.js",
     "clean": "rimraf lib .nyc_output coverage",
-    "posttest": "npm run type-check",
-    "type-check": "tsc --noEmit -p tsconfig.dist.json",
-    "type-check:src": "tsc --noEmit"
+    "posttest": "npm run '"
```
mariuslundgard, line 9: Is `npm run '` (apostrophe) a thing?

bjoerge: Definitely not! Not sure what happened there 😳. Fixing.

rexxars (approved):
There's a stray `packages/part-list/package.json`, but aside from that 👍

PR #2152, opened 2020-11-26
[core] Specify target/source dataset in copy prompt

benedicteb:
Without this users have no way of knowing which dataset is prompted for
in interactive mode.

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

There is no difference between the interactive prompts for source and target dataset when initiating a Cloud Copy.

screenshot_2020-11-26-08:43:39

**Description**

This fix customises the prompt so they're different.

screenshot_2020-11-26-08:33:52

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -75,14 +75,16 @@ export default {
       .list()
       .then((datasets) => datasets.map((ds) => ds.name))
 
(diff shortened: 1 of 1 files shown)


saasen (approved):
LGTM! Thanks for fixing this.

saasen:
Maybe we should look into why the tests fail on 14.x. I don't know if this is a common issue?

benedicteb:
Thanks for review!

> Maybe we should look into why the tests fail on 14.x. I don't know if this is a common issue?

Yes! It should be fixed in #2153 🙂 Very weird issue.

PR #2153, opened 2020-11-26
[base] Fix bug where type-guard didn't return bool

benedicteb:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

We get a Typescript error on build.

screenshot_2020-11-26-09:08:37

This is a weird error though, since it appears to be caused by unrelated changes and only for node 10.

**Description**

Fixes the build error by making sure the type guard always returns a boolean.

**Checklist** 

- [x]  The code is linted


benedicteb:
The same bug happened with node 14.x actually. See #2152 for example.

packages/@sanity/base/src/datastores/user/createUserStore.ts
```diff
@@ -115,7 +115,7 @@ function arrayify(users: User | User[]): User[] {
 }
 
 function isUser(thing: unknown): thing is User {
-  return thing && thing !== null && typeof (thing as User).id === 'string'
+  return thing !== null && typeof (thing as User).id === 'string'
```
rexxars, line 118: If `thing` is undefined, this will crash. Maybe we can cast the expression to boolean instead?
```suggestion
  return Boolean(thing && thing !== null && typeof (thing as User).id === 'string')
```

rexxars (changes requested):

packages/@sanity/base/src/datastores/user/createUserStore.ts
benedicteb: Good point! Added in d8dab71da 🙂

rexxars (approved):
👍

bjoerge:
I suspect this happened as a side effect of typescript 4.1, which happened 7 days ago. Since we don't use lockfiles to pin dependencies in this repo we're susceptible to run into this kind of issues. Good catch though, @benedicteb - I should have noticed the build failures on my previous commits. Sorry about that!

PR #2154, opened 2020-11-26
[core] Dataset copy wait by default

benedicteb:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

The `sanity dataset copy` command will not wait for copy completion by default. In addition the CLI doesn't give you an obvious way of checking in with the copy progress after initating the copy unless you add the `--wait` flag.

**Description**

This changes the `sanity dataset copy` command to `--wait` by default.

It also adds a new flag `--background` which runs the copy in the background. If you do that the CLI will give you a job ID and a command you can run to check in with the progress.

**Note for release**

This changes the behaviour of the `sanity dataset copy` command.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -7,12 +7,15 @@ import debug from '../../debug'
 
 const helpText = `
 Options
-  --wait Waits for the operation to finish
+  --detach Start the copy without waiting for it to finish
+  --attach <job-id> Attach to the running copy process to show progress
 
 Examples
   sanity dataset copy
   sanity dataset copy <source-dataset>
   sanity dataset copy <source-dataset> <target-dataset>
+  sanity dataset copy --detach <source-dataset> <target-dataset>
+  sanity dataset copy --attach <job-id>
 `
 
 const progress = (url) => {
@@ -54,6 +57,32 @@ const progress = (url) => {
   })
 }
 
+const followProgress = (jobId, client, output) => {
+  const spinner = output
+    .spinner({
+      text: `Copy in progress: 0%`,
+    })
+    .start()
+
+  const listenUrl = client.getUrl(`jobs/${jobId}/listen`)
+
+  debug(`Listening to ${listenUrl}`)
+
+  progress(listenUrl).subscribe({
+    next: (event) => {
+      const eventProgress = event.progress ? event.progress : 0
+
+      spinner.text = `Copy in progress: ${eventProgress}%`
+    },
+    error: () => {
+      spinner.fail('There was an error copying the dataset.')
+    },
+    complete: () => {
+      spinner.succeed(`Copy finished.`)
+    },
+  })
+}
+
 export default {
   name: 'copy',
   group: 'dataset',
@@ -62,9 +91,22 @@ export default {
   description: 'Copies a dataset including its assets to a new dataset',
   action: async (args, context) => {
     const {apiClient, output, prompt} = context
-    const [sourceDataset, targetDataset] = args.argsWithoutOptions
     const flags = args.extOptions
     const client = apiClient()
+
+    if (flags.attach) {
+      const jobId = flags.attach
+
+      if (!jobId) {
+        throw new Error('Please supply a jobId')
+      }
+
+      followProgress(jobId, client, output)
+
+      return
+    }
+
+    const [sourceDataset, targetDataset] = args.argsWithoutOptions
 
     const nameError = sourceDataset && validateDatasetName(sourceDataset)
     if (nameError) {
@@ -103,33 +145,16 @@ export default {
         `Copying dataset ${chalk.green(sourceDatasetName)} to ${chalk.green(targetDatasetName)}...`
       )
 
-      if (!flags.wait) {
+      if (flags.detach) {
+        output.print(`Copy initiated.`)
+        output.print(
+          `\nRun:\n\n    sanity dataset copy --attach ${response.jobId}\n\nto watch attach`
+        )
+
         return
       }
 
-      const spinner = output
-        .spinner({
-          text: `~0% done.`,
-        })
-        .start()
-
-      const listenUrl = client.getUrl(`jobs/${response.jobId}/listen`)
-      debug(`Listening on ${listenUrl}`)
-      progress(listenUrl).subscribe({
-        next: (event) => {
-          spinner.text = `~${event.progress}% done.`
-        },
-        error: () => {
-          spinner.fail('There was an error copying the dataset.')
-        },
-        complete: () => {
-          spinner.succeed(
-            `Finished copying dataset ${chalk.green(sourceDatasetName)} to ${chalk.green(
-              targetDatasetName
-            )}.`
-          )
-        },
-      })
+      followProgress(response.jobId, client, output)
     } catch (error) {
       if (error.statusCode) {
         output.print(`${chalk.red(`Dataset copying failed:\n${error.response.body.message}`)}\n`)


packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -7,12 +7,15 @@ import debug from '../../debug'
 
 const helpText = `
 Options
-  --wait Waits for the operation to finish
+  --background Start the copy and return without waiting for it to finish
```
saasen, line 10: Wouldn't a better wording be: _Start the copy without waiting for it to finish_

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -7,12 +7,15 @@ import debug from '../../debug'
...
+  --progress <job-id> Show progress for a running copy
 
 Examples
   sanity dataset copy
   sanity dataset copy <source-dataset>
   sanity dataset copy <source-dataset> <target-dataset>
+  sanity dataset copy --background <source-dataset> <target-dataset>
+  sanity dataset copy --progress <job-id>
```
saasen, line 18: Not sure if we should add this as an option, as `copy` sounds like a command, but when adding this flag, it doesn't do that anymore, only listen on an existing copy operation. Maybe this progress thing should be a part of a top-level API in the CLI?

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -103,33 +143,19 @@ export default {
         `Copying dataset ${chalk.green(sourceDatasetName)} to ${chalk.green(targetDatasetName)}...`
       )
 
-      if (!flags.wait) {
+      if (flags.background) {
+        /*
+         * Use doesn't want to wait for copy to finish.
```
saasen, line 148: Typo: Should be `User`.

saasen: Also, not sure to have a three-line comment when the explanation of what this flag does is in the top of this file.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -103,33 +143,19 @@ export default {
...
+         */
+        output.print(`Copy initiated. Received job ID: ${chalk.blue(response.jobId)}`)
```
saasen, line 150: I added this initially, but was removed due it being very technical. Thoughts, @rexxars?

saasen (changes requested):

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Good point! 🙂 Removed the comment in ae19b5957

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Yeah that's another good point 🙂 We could remove the job ID from this printout considering the whole `--progress` command is printed just below it.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: I like to think of `dataset` as the top-level "category" and then "copy" as the category below it, and that it therefore makes okay sense to have it here. I feel that moving it to a different top-level thing could make it more difficult to find doc-wise.

However I also recognise that this is not a good "UNIX"-way of doing it. So if we think it's worth spending time on moving into a top-level category then I'd love to do it 🙂 Just let me know!

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Sure! 🙂 Fixed in 7841ced59

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
rexxars: I figure we might as well remove it - until we have a stronger concept of the job system, this can possibly be a bit confusing right now.

rexxars (approved):
Looks good to me. No strong opinion on whether or not to print the job ID explicitly, but leaning towards removing it.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
j33ty: I think having a seperate subcommand to track progress of "any job" may also be an option here: `sanity job listen <job-id>`. This is not the only job which CLI will ever listen to, we may have an export to bucket job tomorrow which is triggered with `sanity dataset export ...` and can be tracked with `sanity job listen ...`. I would say this is worth discussing in the team.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -54,6 +57,30 @@ const progress = (url) => {
   })
 }
 
+const followProgress = (jobId, client, output) => {
+  const spinner = output
+    .spinner({
+      text: `~0% done.`,
```
j33ty, line 63: Nitpick: This looks like a false promise to me, using the word `done`. It's common english expression of mentioning progress but an explicit word like `progress: 0%` may be the other option. (Job may fail for various reaons.)

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -54,6 +57,30 @@ const progress = (url) => {
...
+    })
+    .start()
+
+  const listenUrl = client.getUrl(`jobs/${jobId}/listen`)
+
+  debug(`Listening on ${listenUrl}`)
```
j33ty, line 69: Tiny Nitpick: 'Listening on' is used by host and not the clients. Alternative is 'Listening to'.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -54,6 +57,30 @@ const progress = (url) => {
...
+
+  progress(listenUrl).subscribe({
+    next: (event) => {
+      spinner.text = `~${event.progress}% done.`
```
j33ty, line 73: Use `{event.progress ? event.progress : 0}` ? Safer to add a field check than saying `~undefined% done.`, or just cry out loud saying can not process event.

j33ty (approved):
LGTM. I have added some nitpicks.

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
saasen: Good point!

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: > I think having a seperate subcommand to track progress of "any job" may also be an option here: `sanity job listen <job-id>`. This is not the only job which CLI will ever listen to, we may have an export to bucket job tomorrow which is triggered with `sanity dataset export ...` and can be tracked with `sanity job listen ...`. I would say this is worth discussing in the team.

Thanks for the input @j33ty! 🙂 We've discussed this within the team and found that this is a good enough solution for now because the only thing we have which uses it at this point is the copy command and it doesn't state verbosely that this is a job, just that you can attach/detach to a copy.

But when we get more opportunities to use listen we'll move it to be a top level command 🙂

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: That makes sense! 🙂 Removed in 27d1a6373

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Changed it a bit in 8c8b25e88. Hope that's better 🙂

recording_2020-11-27-10:08:44

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Makes sense! 🙂 Added in 065f86eb8

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Fixed in 77c78deda 🙂

saasen (approved):
LGTM!

PR #2157, opened 2020-11-27
[chore] Make linting in CI handle changes with many files better

benedicteb:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

For cases where there are many hundred changed files in a PR the shell will break because of too long a command.

**Description**

This change alters the behaviour of the lint CI action to lint each file individually. This will take longer, however in most cases the slowdown won't be noticed.

If it's a problem over time we can look into chunking.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,7 +31,7 @@ jobs:
         id: buildModules
         run: npm run build
 
-      - name: List which files changed
+      - name: Lint changed files
         id: listChangedFiles
         run: |
           # Is this a pull request?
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -52,10 +52,10 @@ jobs:
           # Don't include files like LICENSE and css files
           CHANGED_JS_FILES=$(echo "$CHANGED_FILES" | sed -e '/^.\+\(mjs\|tsx\|jsx\|js\|ts\)$/!d')
 
-          # Replace \n characters with a space
-          CHANGED_FILES_SEPARATED_BY_SPACES=$(echo "$CHANGED_JS_FILES" | sed ':a;N;$!ba;s/\n/ /g')
+          # Don't include :parts files
+          CHANGED_JS_FILES=$(echo "$CHANGED_JS_FILES" | sed -e '/^.\+\:.*$/d')
```
bjoerge, line 56: Maybe just drop this since it's a bit coincidental that it's needed right now and it may be hard for a future maintainer to assess whether it can be safely removed?

benedicteb: Yes that makes sense! I have removed it :smile:

bjoerge (approved):

PR #2158, opened 2020-11-27
[core] Add --skip-history flag to dataset copy

benedicteb:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

No option for skipping history when copying datasets.

**Description**

This change adds an option to the dataset copy command to skip history during copy.

**Note for release**

Should update docs.

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -9,11 +9,13 @@ const helpText = `
 Options
   --detach Start the copy without waiting for it to finish
   --attach <job-id> Attach to the running copy process to show progress
+  --skip-history Don't preserve document history on copy
 
 Examples
   sanity dataset copy
   sanity dataset copy <source-dataset>
   sanity dataset copy <source-dataset> <target-dataset>
+  sanity dataset copy --skip-history <source-dataset> <target-dataset>
   sanity dataset copy --detach <source-dataset> <target-dataset>
   sanity dataset copy --attach <job-id>
 `
@@ -107,6 +109,7 @@ export default {
     }
 
     const [sourceDataset, targetDataset] = args.argsWithoutOptions
+    const shouldSkipHistory = Boolean(flags['skip-history'])
 
     const nameError = sourceDataset && validateDatasetName(sourceDataset)
     if (nameError) {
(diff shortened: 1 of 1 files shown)


packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -9,11 +9,13 @@ const helpText = `
 Options
   --detach Start the copy without waiting for it to finish
   --attach <job-id> Attach to the running copy process to show progress
+  --skip-history Don't preserve document history on copy
```
saasen, line 12: Isn't it the _dataset_ history that is skipped?

saasen (approved):
Other than my comment, looks good! Thanks!

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: That might be! I'm unsure as to how to phrase this in the best way 🙂 My thinking is that what we're "actually" preserving is the history of all the documents within the dataset right? Like creation, versions and so on. It's not the history of the dataset *itself* we're preserving. But if it makes more sense to say "dataset history" then we can change this 🙂

rexxars: No, I think document history is correct as it is. History is per document, and there's no history outside of that.

saasen: My mistake!

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -107,6 +109,7 @@ export default {
     }
 
     const [sourceDataset, targetDataset] = args.argsWithoutOptions
+    const shouldSkipHistory = !!flags['skip-history']
```
rexxars, line 112: I personally prefer
```suggestion
    const shouldSkipHistory = Boolean(flags['skip-history'])
```

but just a nitpick

rexxars (approved):
👍

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
benedicteb: Thanks! 🙂

PR #2160, opened 2020-11-30
Make sure all environments use the same version of prettier

benedicteb:

- [x]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Since we clamp the prettier version in `package.json` with `^2.1.2` we will automatically pull in all upgrades as long as the major version is 2. However this might produce different formatting conventions depending on "when" you ran `npm install`. CI always being the one running the newest since they always run install before running format.

**Description**

This change clamps the version on `2.1.2`. Producing consistent formatting. However we should make sure to update `prettier` now and then by upping the version in the string. Although that might mean changes to the whole code base so it's nice to have control over.

This could also be solved by having a more expressive `.prettierrc` which conserves our options regardless of `prettier` defaults. However as we've discussed before that might also mean we diverge from the "norm".

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "minimist": "^1.2.0",
     "path-exists-cli": "^1.0.0",
     "pre-commit": "^1.2.2",
-    "prettier": "^2.1.2",
+    "prettier": "2.1.2",
     "rimraf": "^2.7.1",
     "semver": "^6.1.2",
     "stylelint": "^13.4.1",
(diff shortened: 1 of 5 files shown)


bjoerge (approved):
Good call! Agree it's better to bump prettier in a controlled way rather than hard wiring the current default formatting options 👍

rexxars (approved):
Yeah, pinning the version makes sense to me here. LGTM.

PR #2166, opened 2020-12-02
[core] Add dataset alias support

juicycleff:

- [x] New feature (non-breaking change which adds functionality)
- [ ]  Maintenance

- [x] Yes
- [ ]  No

**Current behavior**

**Description**
This PR adds support for `dataset` `alias` in the sanity CLI. It provides the interface to interact and manage `dataset` `alias`

**Note for release**

Added support for the following sub commands to dataset
```bash
  $ sanity dataset alias create <alias-name> <target-dataset>

  $ sanity dataset alias delete <alias-name>

  $ sanity dataset alias link <alias-name> <target-dataset>

  $ sanity dataset alias unlink <alias-name>
```

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/actions/dataset/alias/validateDatasetAliasName.js
+++ b/packages/@sanity/core/src/actions/dataset/alias/validateDatasetAliasName.js
@@ -1,0 +1,33 @@
+module.exports = (datasetName) => {
+  if (!datasetName) {
+    return 'Alias name is missing'
+  }
+
+  const name = `${datasetName}`
+
+  if (name.toLowerCase() !== name) {
+    return 'Alias name must be all lowercase characters'
+  }
+
+  if (name.length < 2) {
+    return 'Alias name must be at least two characters long'
+  }
+
+  if (name.length > 20) {
+    return 'Alias name must be at most 20 characters'
+  }
+
+  if (!/^[a-z0-9~]/.test(name)) {
+    return 'Alias name must start with a letter or a number'
+  }
+
+  if (!/^[a-z0-9~][-_a-z0-9]+$/.test(name)) {
+    return 'Alias name must only contain letters, numbers, dashes and underscores'
+  }
+
+  if (/[-_]$/.test(name)) {
+    return 'Alias name must not end with a dash or an underscore'
+  }
+
+  return false
+}
--- a/packages/@sanity/core/src/commands/dataset/alias/createAliasHandler.js
+++ b/packages/@sanity/core/src/commands/dataset/alias/createAliasHandler.js
@@ -1,0 +1,64 @@
+import promptForDatasetName from '../../../actions/dataset/datasetNamePrompt'
+import promptForDatasetAliasName from '../../../actions/dataset/alias/datasetAliasNamePrompt'
+import validateDatasetAliasName from '../../../actions/dataset/alias/validateDatasetAliasName'
+import validateDatasetName from '../../../actions/dataset/validateDatasetName'
+import * as aliasClient from './datasetAliasesClient'
+import {ALIAS_PREFIX} from './datasetAliasesClient'
+
+export default async (args, context) => {
+  const {apiClient, output, prompt} = context
+  const [, alias, targetDataset] = args.argsWithoutOptions
+  const client = apiClient()
+
+  const nameError = alias && validateDatasetAliasName(alias)
+  if (nameError) {
+    throw new Error(nameError)
+  }
+
+  const [datasets, aliases, projectFeatures] = await Promise.all([
+    client.datasets.list().then((sets) => sets.map((ds) => ds.name)),
+    aliasClient.listAliases(client).then((sets) => sets.map((ds) => ds.name)),
+    client.request({uri: '/features'}),
+  ])
+
+  let aliasName = await (alias || promptForDatasetAliasName(prompt))
+  let aliasOutputName = aliasName
+
+  if (aliasName.startsWith(ALIAS_PREFIX)) {
+    aliasName = aliasName.substring(1)
+  } else {
+    aliasOutputName = `${ALIAS_PREFIX}${aliasName}`
+  }
+
+  if (aliases.includes(aliasName)) {
+    throw new Error(`Dataset alias "${aliasOutputName}" already exists`)
+  }
+
+  if (targetDataset) {
+    const datasetErr = validateDatasetName(targetDataset)
+    if (datasetErr) {
+      throw new Error(datasetErr)
+    }
+  }
+
+  const datasetName = await (targetDataset || promptForDatasetName(prompt))
+  if (datasetName && !datasets.includes(datasetName)) {
+    throw new Error(`Dataset "${datasetName}" does not exist `)
+  }
+
+  const canCreateAlias = projectFeatures.includes('advancedDatasetManagement')
+  if (!canCreateAlias) {
+    throw new Error(`This project cannot create a dataset alias`)
+  }
+
+  try {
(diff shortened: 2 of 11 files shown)


packages/@sanity/core/src/actions/dataset/alias/validateDatasetAliasName.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!/^[a-z0-9~][-_a-z0-9]+$/.test(name)) {
+    return 'Alias name must only contain letters, numbers, dashes and underscores'
+  }
+
+  if (/[-_]$/.test(name)) {
+    return 'Alias name must not end with a dash or an underscore'
+  }
+
+  return false
+}
```
saasen, line 33: This may be waaaaay late, but should we try to re-use existing validations?

We got one we recently tweaked: `packages/@sanity/client/src/validators.js`.

saasen: The code already provides better error messages though. Not sure if we should change this now.

packages/@sanity/core/src/commands/dataset/alias/createAliasHandler.js
```diff
@@ -0,0 +1,64 @@
...
+  const client = apiClient()
+
+  const nameError = alias && validateDatasetAliasName(alias)
+  if (nameError) {
+    throw new Error(nameError)
+  }
+
+  const [datasets, aliases, projectFeatures] = await Promise.all([
+    client.datasets.list().then((sets) => sets.map((ds) => ds.name)),
+    aliasClient.listAliases(client).then((sets) => sets.map((ds) => ds.name)),
```
saasen, line 20: Shouldn't we try to re-use the `client` and not importing and calling this directly?

saasen (changes requested):
I think this looks very good! I left a few comments for what I think needs to change.

packages/@sanity/core/src/commands/dataset/alias/deleteAliasHandler.js
```diff
@@ -0,0 +1,42 @@
+import validateDatasetAliasName from '../../../actions/dataset/alias/validateDatasetAliasName'
+import * as aliasClient from './datasetAliasesClient'
+import {ALIAS_PREFIX} from './datasetAliasesClient'
+
+export default async (args, context) => {
+  const {apiClient, prompt, output} = context
+  const [, ds] = args.argsWithoutOptions
```
benedicteb, line 7: Perhaps we could use a more descriptive name for `ds`? 🙂

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  if (!aliases.includes(aliasName)) {
+    throw new Error(`Dataset alias "${aliasOutputName}" does not exist `)
+  }
+
+  // get the current alias from the remote alias list
+  const linkedAlias = fetchedAliases.find((elem) => elem.name === aliasName)
+
+  if (linkedAlias) {
```
benedicteb, line 35: What happens if `.find` doesn't find anything? 🙂

benedicteb (approved):
Nice work! :grinning: Don't think I have any other obvious things other than the ones Anders have already pointed out 🙂

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
saasen: The intention is probably to throw the error below, but I think the code needs to be moved.

packages/@sanity/core/src/commands/dataset/alias/deleteAliasHandler.js
juicycleff: Ds is used almost everywhere in the cli just followed the maiming.

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
juicycleff: Why do you think the intention is to throw an error. Please go through the logic again and understand what its doing :)

packages/@sanity/core/src/commands/dataset/alias/createAliasHandler.js
juicycleff: The handlers are just fine the way it is when the cli supports grouped commands.

packages/@sanity/core/src/actions/dataset/alias/validateDatasetAliasName.js
rexxars: We shouldn't import from the client internals, so I feel like this is fine.

juicycleff: Well I do have objections depending on client package for validations, but either way, you will have to update even datasets too.

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
saasen: It `.find` doesn't find anything given the predicate, it returns undefined.

So shouldn't the `if (!linkedAlias.datasetName)` be moved to before this `if` and then throw the error?

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
```diff
@@ -0,0 +1,59 @@
...
+  const {apiClient, output, prompt} = context
+  const [, alias] = args.argsWithoutOptions
+  const client = apiClient()
+
+  const nameError = alias && validateDatasetAliasName(alias)
+  if (nameError) {
+    throw new Error(nameError)
+  }
+
+  const [fetchedAliases] = await Promise.all([aliasClient.listAliases(client)])
```
rexxars, line 16: This is only doing a single fetch, so it reads a little better if it's simplified to not use `Promise.all`:
```suggestion
  const fetchedAliases = await aliasClient.listAliases(client)
```

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (aliasName.startsWith(ALIAS_PREFIX)) {
+    aliasName = aliasName.substring(1)
+  } else {
+    aliasOutputName = `${ALIAS_PREFIX}${aliasName}`
+  }
+
+  if (!aliases.includes(aliasName)) {
+    throw new Error(`Dataset alias "${aliasOutputName}" does not exist `)
+  }
```
rexxars, line 30: Maybe instead of mapping the aliases here we could reuse the `linkedAlias` check below? It wasn't obvious to me that the `linkedAlias` would always find the target, given it's in an `if` branch

packages/@sanity/core/src/commands/dataset/alias/unlinkAliasHandler.js
rexxars: After re-reading this, as far as I can tell `linkedAlias` will actually always find a value, given the check on line 28, which means this `if` branch isn't actually necessary? I think it would read better if refactored so the check on line 28 uses `linkedAlias` here instead.

rexxars (changes requested):

rexxars (approved):

PR #2199, opened 2020-12-22
Fix a few TS issues

bjoerge:
This adds three commits that can be reviewed individually:

### 8de684a [base] Fix typing issues with setTimeout
There was suddenly build errors on setTimeout return values which I suspect was due to a new minor/patch release of TypeScript. This fixes the issue by using `ReturnType<typeof setTimeout>` as type for timer id where applicable.

### 973f89b [google-maps-input] Upgrade @types/googlemaps
Looks like there has been some changes to the type definitions in @types/googlemaps recently that broke the google-maps-input package

### bfea836a5 [chore] Upgrade typescript and remove as dependency from individual packages

This upgrades typescript in the monorepo to latest (4.1.3) and at the same time removes typescript as dev dependency from some individual packages that still had it around)

- [x]  Maintenance

- [ ]  Yes
- [x]  No

### DIFF
--- a/examples/design-studio/package.json
+++ b/examples/design-studio/package.json
@@ -30,7 +30,6 @@
     "sanity-plugin-asset-source-unsplash": "^0.1.3"
   },
   "devDependencies": {
-    "typescript": "^4.0.2",
     "typescript-plugin-css-modules": "^2.4.0"
   }
 }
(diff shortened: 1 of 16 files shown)


rexxars (approved):
👍 Nice

PR #2200, opened 2020-12-23
[desk-tool] Scroll to nearest block if on focus path change

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Scrolls to _top_ of the element containing the first segment of the focus path.

**Description**

When you have a portable text field at the bottom of object and have not yet focused any element in the form, then click the "activate" handler on it, it scrolls to the top of the form. This is caused by the editor attempting to scroll to the top of the first segment in the focus path (the object). Previously we scrolled to the _center_, which meant we sometime scrolled _past_ the field (such as with a large form and clicking the first field). This PR scrolls to the _nearest_ point, which seems more correct at least.

I'm still not sure I feel this is the right solution, but I can't quite remember why we wanted to scroll in the first place. 

**Note for release**

- Fix bug where form would scroll to top when first focusing a child of a nested field

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/documentPane/documentPanel/views/editForm.tsx
+++ b/packages/@sanity/desk-tool/src/panes/documentPane/documentPanel/views/editForm.tsx
@@ -53,7 +53,7 @@ export const EditForm = memo((props: Props) => {
   useEffect(() => {
     const el = document.querySelector(`[data-focus-path="${startSegment}"]`)
     if (el) {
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
> I can't quite remember why we wanted to scroll in the first place.

IIRC this was added after we got the ability to deep link to content that might be edited in dialogs, to make sure we had the enclosing value visible in the background.

This seems like a good quickfix, but let's see if we can address this properly (along with other issues related to focus handling) in the coming time.

PR #2221, opened 2021-01-15
[base,desk-tool,form-builder] Move ScrollContainer/ScrollMonitor from legacy to base/components

benedicteb:
* Move `ScrollConatiner`, `ScrollMonitor` and relevant methods from
  `__legacy` to `components/` in `@base`.
* Rewrite usages to import from `@sanity/base` instead of parts
* Fix some lint errors complaining of unused props

Let me know if something doesn't make sense in this move!


packages/@sanity/types/parts/part:@sanity/components/components.d.ts
```diff
@@ -192,15 +192,6 @@ declare module 'part:@sanity/components/menu-button' {
 declare module 'part:@sanity/components/radiobutton/default' {
   export {default} from './build-snapshot/__legacy/@sanity/components/radiobutton/RadioButtonDefault'
 }
-declare module 'part:@sanity/components/scroll' {
```
bjoerge, line 195: This is only used internally, so should be safe to remove 👍

bjoerge (approved):
Looks good to me!

PR #2222, opened 2021-01-15
[form-builder] Migrate Details component to Sanity UI

mariuslundgard:
The `Details` component is used by:

- `inputs/Warning.tsx`
- `inputs/arrays/ArrayOfObjectsInput.tsx`
- `inputs/ObjectInput/UnknownFields.tsx`

### DIFF
--- a/packages/@sanity/form-builder/src/components/Details.tsx
+++ b/packages/@sanity/form-builder/src/components/Details.tsx
@@ -1,6 +1,6 @@
 import {ToggleArrowRightIcon} from '@sanity/icons'
 import {Box, Flex, Text} from '@sanity/ui'
-import React, {useCallback, useState} from 'react'
+import React, {useCallback, useEffect, useState} from 'react'
 import styled from 'styled-components'
 
 interface DetailsProps {
@@ -32,7 +32,6 @@ const HeaderButton = styled.button`
 
 const Header = styled(Flex)`
   cursor: default;
-  user-select: none;
 `
 
 const IconBox = styled(Box)`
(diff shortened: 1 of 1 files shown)


packages/@sanity/form-builder/src/components/Details.tsx
```diff
@@ -1,50 +1,71 @@
...
+`
 
-export class Details extends React.Component<DetailsProps, DetailsState> {
-  static defaultProps = {
-    title: 'Details',
-    isOpen: false,
-  }
+const Header = styled(Flex)`
+  cursor: default;
+  user-select: none;
```
bjoerge, line 27: What's the reason for `user-select: none` here? 

It seems to be different from how [native `<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) elements work.

packages/@sanity/form-builder/src/components/Details.tsx
```diff
@@ -1,50 +1,71 @@
...
-          </div>
-        </button>
-        <div className={styles.content}>{children}</div>
-      </div>
-    )
-  }
+export function Details(props: DetailsProps) {
+  const {children, open: openProp, title = 'Details'} = props
+
+  const [open, setOpen] = useState(openProp || false)
```
bjoerge, line 44: Can we assume it's controlled by the consumer If we receive a non-undefined `open` prop and not maintain internal state here? As it is now a consumer can't fully control its open/closed state.

bjoerge (commented):
Looks good, left a couple of comments.

benedicteb (approved):

packages/@sanity/form-builder/src/components/Details.tsx
mariuslundgard: Fixed

PR #2224, opened 2021-01-15
[form-builder] Make uploadTarget a HOC that builds on fileTarget and move into ArrayInput

bjoerge:
This refactors UploadTargetFieldset so that it builds on the `fileTarget` HOC previously introduced. I've placed it with the ArrayOfObjectsInput for now, since it's the only place where it is currently in use. (We're likely to revisit how this works in a future iteration).


benedicteb (approved):
Nice work :grinning:

PR #2228, opened 2021-01-18
Move default asset source out of the image input

bjoerge:
This moves the default (sanity cdn backed) asset source out of the image input itself to get better separation of concerns between an asset source and the image input. The ImageInput is now only dependent on the asset source type/interface, and not any particular implementation of it.


packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -350,14 +343,14 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
   }
 
   handleSelectFile = (files: FileList) => {
-    this.uploadFirstAccepted(files)
+    this.uploadFirstAccepted(Array.from(files))
   }
 
-  handleUpload = ({file, uploader}) => {
+  handleUpload = ({file, uploader}: TODO) => {
```
benedicteb, line 349: Maybe it's more descriptive to call this `PploadReturnType` or something then set it as any with a todo above it?

```javascript
// TODO(@bjoerge, 2021-01-18) Populate this return type
type UploadReturnType = any;
``` 

This would both make it clear what the return type is used for *and* that it's not finished yet.

benedicteb (approved):

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
bjoerge: Excellent point! Will address this in my next task.

PR #2230, opened 2021-01-19
[form-builder] Remove storybook and stories

benedicteb:
I also removed `stories/objectField.js` but I am unsure whether it should also be deleted since we didn't specify it in the story.


bjoerge (approved):
LGTM! 🙌 

fyi, I took the liberty of rebasing against `ui/main`

PR #2236, opened 2021-01-20
Migrate primitive value inputs to Sanity UI

bjoerge:

- [x]  Maintenance

- [x]  No

**Description**
This migrates the default primitive value inputs to use Sanity UI instead of css modules.


packages/@sanity/form-builder/src/inputs/BooleanInput.tsx
```diff
@@ -1,21 +1,20 @@
...
-import ValidationStatus from 'part:@sanity/components/validation/status'
-import FieldStatus from '@sanity/base/lib/__legacy/@sanity/components/fieldsets/FieldStatus'
-import {ChangeIndicator} from '@sanity/base/lib/change-indicators'
+import React, {useCallback} from 'react'
+import {useId} from '@reach/auto-id'
+import {Box, Card, Checkbox, Flex, Switch} from '@sanity/ui'
 import {BooleanSchemaType} from '@sanity/types'
+import {FormFieldHeaderText} from '@sanity/base/components'
+import {ChangeIndicator} from '@sanity/base/lib/change-indicators'
+import FieldStatus from '@sanity/base/lib/__legacy/@sanity/components/fieldsets/FieldStatus'
```
mariuslundgard, line 7: Can we make this importable from `@sanity/base/components`?

packages/@sanity/form-builder/src/inputs/BooleanInput.tsx
```diff
@@ -24,43 +23,43 @@ const BooleanInput = React.forwardRef(
...
+                ref={ref}
                 label={type.title}
-                readOnly={readOnly}
+                readOnly={Boolean(readOnly)}
                 onChange={handleChange}
                 onFocus={onFocus}
-                checked={value}
+                indeterminate={indeterminate}
+                checked={checked}
+                content={type.description}
```
mariuslundgard, line 45: I'm unfamiliar with using the `content` attribute on input elements. What does it do?

packages/@sanity/form-builder/src/inputs/BooleanInput.tsx
```diff
@@ -24,43 +23,43 @@ const BooleanInput = React.forwardRef(
       [onChange]
     )
 
+    const indeterminate = typeof value !== 'boolean'
+    const checked = value || false
+
+    const Input = layout === 'checkbox' ? Checkbox : Switch
```
benedicteb, line 29: Super minor this but maybe we should call this something that signifies more clearly that it can vary between different components? 🙂 Like `CheckboxOrSwitch` or `LayoutSpecificInput` or something like that just to increase readability? Could perhaps even wrap it into a separate mini-component to make it more clearly that it branches into two paths.

packages/@sanity/form-builder/src/inputs/NumberInput.tsx
```diff
@@ -1,56 +1,53 @@
...
-}
+const NumberInput = React.forwardRef(function NumberInput(
+  props: Props<number, NumberSchemaType>,
+  forwardedRef: React.ForwardedRef<HTMLInputElement>
+) {
+  const {value = '', readOnly, markers, type, level, onFocus, onChange, presence} = props
+  const validation = markers.filter((marker) => marker.type === 'validation')
+  const errors = validation.filter((marker) => marker.level === 'error')
+  const id = useId()
+  // Show numpad on mobile if only positive numbers is preferred
```
mariuslundgard, line 18: Is this comment still accurate?

benedicteb (approved):

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
```diff
@@ -1,86 +1,111 @@
...
+          items={items}
+          value={currentItem}
+          onChange={handleChange}
+          direction={type.options.direction || 'vertical'}
+          ref={forwardedRef as React.ForwardedRef<HTMLInputElement>}
+          readOnly={readOnly}
+          customValidity={errors?.[0]?.item.message}
+        />
+      ) : (
+        <Select
```
mariuslundgard, line 58: The `Select` is missing the `disabled` property when in read only mode.

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
```diff
@@ -1,86 +1,111 @@
...
+    direction: 'horizontal' | 'vertical'
+    readOnly: boolean
+    onChange: (value: TitledListValue<string | number> | null) => void
+    customValidity?: string
+  },
+  forwardedRef: React.ForwardedRef<HTMLInputElement>
+) {
+  const {items, value, onChange, readOnly, customValidity, direction} = props
+
+  const Flow = direction === 'horizontal' ? Inline : Stack
```
mariuslundgard, line 90: Rename `Flow` to `Layout` here?

packages/@sanity/form-builder/src/inputs/Slug/SlugInput.tsx
```diff
@@ -42,169 +28,119 @@ type Props = {
...
+      >
+        <Stack space={3}>
+          <Flex>
+            <Box flex={1}>
+              <TextInput
+                id={inputId}
+                ref={forwardedRef}
+                customValidity={errors.length > 0 ? errors[0].item.message : ''}
+                disabled={isUpdating}
+                onChange={React.useCallback((event) => updateSlug(event.currentTarget.value), [
```
mariuslundgard, line 118: Assigning a variable for this (e.g. `handleChange`) would de-nest and make this a bit easier to read.

packages/@sanity/form-builder/src/inputs/Slug/utils/useAsync.tsx
```diff
@@ -0,0 +1,49 @@
...
+  status: 'pending'
+}
+export type AsyncErrorState = {
+  status: 'error'
+  error: Error
+}
+
+export type AsyncState<T> = AsyncPendingState | AsyncCompleteState<T> | AsyncErrorState
+
+export function useAsyncCallback<T, U>(
```
mariuslundgard, line 17: We should add some TSDoc to describe what this is and what it does, and whether there are any known gotchas.

packages/@sanity/form-builder/src/inputs/UrlInput.tsx
```diff
@@ -1,54 +1,49 @@
...
+    >
+      <TextInput
+        type="url"
+        inputMode="url"
+        id={inputId}
+        customValidity={errors.length > 0 ? errors[0].item.message : ''}
+        value={value || ''}
+        readOnly={Boolean(readOnly)}
+        placeholder={type.placeholder}
+        onChange={React.useCallback(
```
mariuslundgard, line 34: Same as above.

packages/@sanity/form-builder/src/inputs/TextInput.tsx
```diff
@@ -1,52 +1,52 @@
...
+      __unstable_presence={presence}
+      inputId={inputId}
+    >
+      <TextArea
+        id={inputId}
+        customValidity={errors && errors.length > 0 ? errors[0].item.message : ''}
+        value={value || ''}
+        readOnly={Boolean(readOnly)}
+        placeholder={type.placeholder}
+        onChange={React.useCallback(
```
mariuslundgard, line 36: Same as above.

packages/@sanity/form-builder/src/inputs/Warning.tsx
```diff
@@ -23,7 +23,7 @@ export default function Warning({heading, message, values = [], children}: Props
           </>
         )}
       </h2>
-      <div className={styles.content}>
+      <div>
```
mariuslundgard, line 26: Should this be part of this commit?

packages/@sanity/form-builder/src/inputs/BooleanInput.tsx
mariuslundgard: (Should be done in a separate PR so we can merge this!)

packages/@sanity/form-builder/src/inputs/BooleanInput.tsx
bjoerge: No idea what happened there. Removed.

packages/@sanity/form-builder/src/inputs/NumberInput.tsx
bjoerge: Looks like it? It was introduced in db3d22875ce9f8d2fdb983ed8a741ec84fa86cb3 and I'm thinking it's out of scope to change this as part of this PR?

bjoerge: Ah, you mean after adding `inputMode="numeric"`. Good point. I'm changing to `text` in the case of `onlyPositiveNumber`

packages/@sanity/form-builder/src/inputs/Slug/utils/useAsync.tsx
bjoerge: Added some tsdocs to it (and realized it was named wrong). Fixed & pushed. Let's consider if we should lift it over to a generic tool library at some point.

mariuslundgard (approved):
Great work!

PR #2237, opened 2021-01-20
Rewrite date inputs to use date picker based on UI

bjoerge:

- [x]  Maintenance

- [x]  No

**Description**
This is a complete rewrite of our default date+datetime input to a new one that's based on Sanity UI.

**Note for release**

Migrates default date and datetime inputs to use Sanity UI


rexxars (approved):
Looks great! Found a few edge-cases, though:

rexxars:
Found a couple of edge-cases:

On short browser windows, the date picker sometimes pushes the footer... up? Kind of weird
image

rexxars:
On mobile there's also some odd similar behavior:
image

rexxars (changes requested):
I think the mobile bug needs to be looked at, at least.

bjoerge:
> I think the mobile bug needs to be looked at, at least.

Added a workaround for this in 770e16c.

PR #2240, opened 2021-01-21
[base] Use legacy theming for Sanity UI inputs

mariuslundgard:
This change adds mapping of color values from the legacy theming to the Sanity UI theme object, so that input elements get the correct colors.

### In order to test:

1. Open `examples/design-studio/themes/dev.css`
2. Change the `@import` to use `dark.css`
3. Start the Design Studio using `npm run design-studio`
4. Make sure all inputs have the expected color (defined in `dark.css`)


bjoerge (approved):
LGTM!

PR #2244, opened 2021-01-22
Migrate form-builder to Sanity UI

bjoerge:
This PR reflects the current state of our ongoing effort to migrate packages to use Sanity UI. It's mainly oriented around the `form-builder`-package, but also includes necessary changes from other packages.


bjoerge:
This work has been continued in other branches

PR #2246, opened 2021-01-22
[core] GraphQL: Add --non-null-document-fields flag

rexxars:

- [ ]  Maintenance

- [x]  Yes
- [ ]  No

**Current behavior**

In a previous release, we set the document interface fields (`_id`, `_type`, `_createdAt` and such) to be nullable, because you can use a document type as an (unreferenced) object in a document field or array. In these situations, `_createdAt` would not be present - so flagging it as non-nullable was incorrect.

**Description**

In many cases this is not something that you actually use, however - you usually create objects for things you want to inline and documents for things you want to reference. Having these fields be non-null can greatly enhance the development experience.

This PR adds a flag you can use to opt-in for non-nullable document fields. This way the technically correct behavior is the default, while the user (who knows their schema best) can _choose_ to opt in.

I've added yargs and explicit argv parsing within the action because these boolean flags gets unwieldy if not parsed correctly. We should find the time to improve the general flags parsing in the CLI soon, but leaving that for another time.

**Note for release**

- Add `--non-null-document-fields` flag to the `sanity graphql deploy` command. Use it to force fields like `_id`, `_type` and `_createdAt` to be marked as non-nullable fields in GraphQL APIs.

**Checklist** 

- [x]  The code is linted
- [x]  Corresponding changes to the documentation have been made _(staged)_

### DIFF
--- a/packages/@sanity/cli/src/util/parseArguments.js
+++ b/packages/@sanity/cli/src/util/parseArguments.js
@@ -27,6 +27,7 @@ module.exports = function parseArguments(argv = process.argv) {
       v, version,
     },
 
+    argv, // forwarded to commands that want to be more explicit about parsing
     extOptions, // forwarded to commands
     argsWithoutOptions, // remaining arguments
     extraArguments, // arguments after the ended argument list (--)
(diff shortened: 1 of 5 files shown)


saasen:
It looks like this must be rebased right now. Getting some conflicts when I tried. Do you want to do it, @rexxars?

rexxars:
> It looks like this must be rebased right now. Getting some conflicts when I tried. Do you want to do it, @rexxars?

Yep, thanks - rebased.

saasen (approved):
Sweet! This is very useful. LGTM! 👍

PR #2247, opened 2021-01-24
[validation] Fix bug on object type with "value" field + "list" option

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Note: This is quite an edge-case, but be patient with me:

If you...
1. Have an object type that defines the field `value`
2. Add the (non-standard) `options.list` configuration (generally seen in `string` and `number` types) on the field/type
3. Set a (valid) value on the field 

... when the validation inference runs on the schema types, it will incorrectly grab the value of the `value` field, instead of the object itself. When comparing the value of the object field (`{value: 'foo', other: 'props'}`) against the "allowed" values, it will never find a match, since it's comparing the nested value of the `value` prop instead of the root object.

This results in a "Value does not match allowed values" error.

This particular case was uncovered by the [color-list](https://github.com/KimPaow/sanity-color-list) plugin.

**Description**

This PR changes the logic of the `extractValueFromListOption` method in the validation module to look for a defined `value` field on the object. If defined, it will use the root-level object as the value - otherwise it will assume you wanted the nested `value`.

While this technically is an edge case, I figured it's not too far-fetched that we will support an object select in the same way we support string/number selects, so this will come in handy in that case. 

I also added an example input component mirroring the behavior of the color list plugin (but without the fancy UI logic), which lets us reproduce the issue.

**Note for release**

- Fix bug when object type has a `value` field and is paired with the (unsupported) `options.list` schema type option

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/examples/test-studio/schemas/validation.js
+++ b/examples/test-studio/schemas/validation.js
@@ -1,6 +1,7 @@
 import client from 'part:@sanity/base/client'
 import {points, featureCollection} from '@turf/helpers'
 import pointsWithinPolygon from '@turf/points-within-polygon'
+import {CustomObjectSelectInput} from '../src/components/CustomObjectSelectInput'
 import norway from '../data/norway'
 
 export const validationArraySuperType = {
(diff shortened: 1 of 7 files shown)


benedicteb (approved):
Nice work! 😀

Since I'm thinking a lot of about automatic tests these days - can you picture a way this could be tested automatically? Like a "test script" or something with the browser?

rexxars:
> Nice work! 😀
> 
> Since I'm thinking a lot of about automatic tests these days - can you picture a way this could be tested automatically? Like a "list of actions" you can do with the test-studio to make sure this does what it should?

Thanks! 

This is a pretty deep edge-case, but you _should_ be able to test it with the new example I added to the test-studio;

- Open a document of type `validationTest`, scroll down to the `objectList` field
- Select a value from the custom input component (from the dropdown (`<select>`))
- Ensure that no validation errors show up

Having said that, I added some test cases in unit-test land which I ensured failed before the bugfix and passes now. 

Will merge this based on previous review given it's even more thorough now - thanks for pushing us to do more tests ❤️

PR #2252, opened 2021-01-25
[form-builder] Fix bug where SelectInput sent the wrong update patch

benedicteb:
* Fixed a typo in the value where we should have used `-` instead of `+`. Other than that the idea is the same - use index to map between value and item
* Added `value` to the select so that it'll have the right value selected on for example undo from review changes

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
@@ -33,7 +33,7 @@ const SelectInput = React.forwardRef(function SelectInput(
   }
 
   const optionValueFromItem = (item) => {
-    return items.indexOf(item)
+    return String(items.indexOf(item))
   }
 
   const inputId = useId()
--- a/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
@@ -49,9 +49,12 @@ const SelectInput = React.forwardRef(function SelectInput(
   const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     const nextItem = itemFromOptionValue(event.currentTarget.value)
 
-    if (nextItem) {
-      handleChange(nextItem)
+    if (!nextItem) {
+      handleChange(EMPTY_ITEM)
+      return
     }
+
+    handleChange(nextItem)
   }
 
   return (
--- a/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
@@ -26,15 +26,34 @@ const SelectInput = React.forwardRef(function SelectInput(
   const validation = markers.filter((marker) => marker.type === 'validation')
   const errors = validation.filter((marker) => marker.level === 'error')
 
+  const itemFromOptionValue = (optionValue) => {
+    const index = Number(optionValue)
+
+    return items[index]
+  }
+
+  const optionValueFromItem = (item) => {
+    return items.indexOf(item)
+  }
+
   const inputId = useId()
   const handleChange = React.useCallback(
-    (nextItem) => {
+    (nextItem: TitledListValue<string | number>) => {
       onChange(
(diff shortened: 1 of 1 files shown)


packages/@sanity/form-builder/src/inputs/SelectInput.tsx
```diff
@@ -26,15 +26,34 @@ const SelectInput = React.forwardRef(function SelectInput(
...
     },
     [onChange]
   )
+
+  const handleSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const nextItem = itemFromOptionValue(event.currentTarget.value)
+
+    if (nextItem) {
+      handleChange(nextItem)
+    }
```
mariuslundgard, line 54: We need an `else` condition to set the value back to `undefined`, something like:

```js
} else {
  handleChange(EMPTY_ITEM)
}
```

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
```diff
@@ -26,15 +26,34 @@ const SelectInput = React.forwardRef(function SelectInput(
...
   const errors = validation.filter((marker) => marker.level === 'error')
 
+  const itemFromOptionValue = (optionValue) => {
+    const index = Number(optionValue)
+
+    return items[index]
+  }
+
+  const optionValueFromItem = (item) => {
+    return items.indexOf(item)
```
mariuslundgard, line 36: I believe this should be modified to a string, and not return `-1`, e.g. `return String(items.indexOf(item) || 0)`

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
```diff
@@ -26,15 +26,34 @@ const SelectInput = React.forwardRef(function SelectInput(
   const validation = markers.filter((marker) => marker.type === 'validation')
   const errors = validation.filter((marker) => marker.level === 'error')
 
+  const itemFromOptionValue = (optionValue) => {
+    const index = Number(optionValue)
+
+    return items[index]
```
mariuslundgard, line 32: I might be wrong, but I believe it should be `items[index - 1]` to find the correct item?

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
benedicteb: We are using `-1` as the first value for the undefined element though since `0` is the first index.

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
mariuslundgard: (Ah, never mind)

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
benedicteb: Made it return a string in 0c9c47f03

mariuslundgard: Ah, clever! Yes, that makes sense 👍

bjoerge:
I think we could simplify quite a bit if we added the empty item up-front here, e.g. replace this line:
https://github.com/sanity-io/sanity/blob/0c9c47f032204eb23fbe710bef0b3a5574ff321b/packages/%40sanity/form-builder/src/inputs/SelectInput.tsx#L23

with something like
```ts
const items = [EMPTY_ITEM, ...((type.options?.list || []) as unknown[]).map(toSelectItem)]
```

That way we can treat the empty case the same way as everything else, and we avoid the slightly confusing `index - 1` handling.

packages/@sanity/form-builder/src/inputs/SelectInput.tsx
benedicteb: Good point! Added in 32f22ed19

benedicteb:
> I think we could simplify quite a bit if we added the empty item up-front here [...]

Great observation! :grinning: 

Maybe we should make an issue for it so we can do that after we've done the release? Just to avoid more changes than necessary in this release.

mariuslundgard (approved):
💯

PR #2259, opened 2021-01-26
[form-builder] TextInput: Only apply default when type.rows is not a number

### LINKED ISSUE #2258: `rows: 0` is broke with the new sanity/ui primitives
**Describe the bug**

Although #2255 fixes the use of the rows property, when you use `rows: 0` that is completely ignored and the input falls back to the default 10 rows.

With most of my studios I set `rows: 0`, so this will break all of them, and I suspect others do the same, too 🤔 

**To Reproduce**

1. Add a `text` field and set its `rows` to `0`
2. Open the field in the studio and notice it's much higher than expected

**Expected behavior**

Apply `rows={0}` to the `<StyledTextArea>` element.

**Screenshots**

```
@sanity/cli                 2.2.2 (up to 
date)
@sanity/base                2.2.2 (up to 
date)
@sanity/cli                 2.2.2 (up to 
date)
@sanity/components          2.1.4 (up to 
date)
@sanity/core                2.2.2 (up to 
date)
@sanity/dashboard           2.1.4 (up to 
date)
@sanity/default-layout      2.2.2 (up to 
date)
@sanity/default-login       2.1.4 (up to 
date)
@sanity/desk-tool           2.2.2 (up to date)
@sanity/production-preview  2.1.4 (up to date)
@sanity/vision              2.1.4 (up to date)
```

**Additional context**

The problem can be traced to line 52 in `/inputs/TextInput.tsx`:
https://github.com/sanity-io/sanity/blob/de414bf98dd0181173a7dd2a749f60f3fbff9469/packages/%40sanity/form-builder/src/inputs/TextInput.tsx#L52

Changing that line to `typeof type.rows === 'number' ? type.rows : 10` would fix the problem. I can do the PR myself, **I'd love to push code to Sanity** 😬

bjoerge:
### Description

Fixes #2258

### Notes for release
Fixed an issue with text areas getting default (10) rows when using `rows: 0` on the schema type

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/TextInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/TextInput.tsx
@@ -49,7 +49,7 @@ const TextInput = React.forwardRef(function TextInput(
         onChange={handleChange}
         onFocus={onFocus}
         onBlur={onBlur}
-        rows={type.rows || 10}
+        rows={typeof type.rows === 'number' ? type.rows : 10}
         ref={forwardedRef}
       />
     </FormField>


mariuslundgard (approved):
👍

PR #2261, opened 2021-01-27
[chore] Add end-to-end testing with Cypress

benedicteb:
Can be used to write automated tests which actually utilize a browser.

Will produce videos and screenshots as artifacts.

Can also be run locally.

Uses an environment variable for logging in.

Includes an example test which checks the heigh of a boolean input.


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
mariuslundgard, line 5: There's a newline at the end of this file.

.gitignore
```diff
@@ -48,3 +48,7 @@ jspm_packages
 # Lockfiles
 yarn.lock
 package-lock.json
+
+# Cypress
+cypress/videos/
+cypress/screenshots/
```
mariuslundgard, line 54: Might be cleaner if this is moved to a file in `cypress/.gitignore`:

```
/videos
/screenshots
```

cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:3333"
+}
```
mariuslundgard, line 3: Can this file live in `cypress/` somehow? Grouping the cypress-related things in the cypress directory would make it more self-contained and nice IMO :)

cypress/integration/booleanInput.js
```diff
@@ -0,0 +1,7 @@
+describe('BooleanInput', () => {
+  it('should have a height of 35px if no description', () => {
+    cy.visit('/test/desk/booleansTest;bd99e58a-845f-4d52-b54a-56a9b7af3be1')
```
mariuslundgard, line 3: I believe we should avoid using IDs in the URLs – just in case this document is somehow deleted by someone. How about making it so that this test first creates a new document?

cypress/support/commands.js
```diff
@@ -0,0 +1,29 @@
...
+        httpOnly: true,
+        sameSite: 'None',
+        domain: `.${domain}`,
+      })
+    })
+  })
+})
+
+Cypress.Commands.add('getField', (fieldName) => {
+  // TODO(@benedicteb, 2021-01-26) Add <data-qa=..." /> or something to html to make select super robust
```
mariuslundgard, line 21: (Side note: I think we should consider `data-test="*"` since that is preferred by Cypress. `data-qa` is a qawolf thing.)

.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     './packages/eslint-config-sanity/import.js',
     'plugin:@typescript-eslint/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:cypress/recommended',
```
mariuslundgard, line 20: This should be moved to a file in `cypress/.eslintrc`, so that the cypress-specific eslint rules don't affect all source files in the monorepo:

```
{
  "plugins": ["cypress"],
  "extends": ["plugin:cypress/recommended"]
}
```

cypress/integration/booleanInput.js
mariuslundgard: Also, maybe we should suffix test files with `.test.js` or `_spec.js` or something?

cypress/fixtures/example.json
benedicteb: Thanks! :grinning: Added the missing newline in ba336589e

.gitignore
benedicteb: Great point! :grinning: Moved in 349c0ed2d

cypress/integration/booleanInput.js
benedicteb: Oh definitely! This is just an example to have a test that makes sense make it into `next`. I'll submit a fix for it quite soon after merging this 🙂

cypress/support/commands.js
benedicteb: Yes great point :grinning: I just added `data-qa` to the comment since we talked about it. It doesn't matter what we end up with just so long as it makes sense for us is what I'm thinking. For the moment `data-focus-path` should do the job!

.eslintrc.js
benedicteb: Very good point! :grinning: Moved in f1e7fe7f3

cypress.json
benedicteb: Yes no worries! :grinning: Moved in f85c9eaef. Cypress does this annoying thing where `--config-file` has to be at the *end* of the command. So instead of making it nice and general by having it in the `cypress` script in `package.json` we have to add it wherever it's run. Added it to `test-e2e` and then we can add it to all newer commands we add later.

bjoerge (approved):
🎉 !

PR #2262, opened 2021-01-27
[base] Fix placement of SearchableSelect's result box

mariuslundgard:
### Description

Code changes:
- The `SearchableSelect` now calls `popper.forceUpdate` whenever the height of the result items change.
- Replaces the [`altAxis`](https://popper.js.org/docs/v2/modifiers/prevent-overflow/#altaxis) and [`tether`](https://popper.js.org/docs/v2/modifiers/prevent-overflow/#tether) options in the `preventOverflow` modifier with using the `flip` modifier to fix situations in which the popper has insufficient room to expand below the reference element.

These changes solves the issue when a reference input is placed at the bottom of an array item, and the results box was cut off by the pane's scrollable element.

image

### What to review

- Make sure the reference input's search result box is placed correctly in all situations.

### Notes for release

- Fixes an issue with reference search results being cut off by the editor’s scrollable element.

### DIFF
--- a/examples/test-studio/schemas/poppers.js
+++ b/examples/test-studio/schemas/poppers.js
@@ -1,0 +1,29 @@
+export default {
+  type: 'document',
+  name: 'poppers',
+  title: 'Poppers',
+  fields: [
+    {
+      type: 'string',
+      name: 'title',
+      title: 'Title',
+    },
+    {
+      type: 'array',
+      name: 'objectsWithReference',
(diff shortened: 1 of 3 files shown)


vicbergquist (approved):
works like a charm as far as I've tested it 👍

mariuslundgard:
Thanks you @vicbergquist 🙏

PR #2266, opened 2021-01-28
[base] Improve theme errors originating in @sanity/ui

mariuslundgard:
### Description

- Test Studio: Adds a tool called `Test UI error` (http://localhost:3333/test/test-ui-error)
- Removes the `^` from the version string written by `packages/@sanity/base/scripts/writeRequiredUIVersion.js`
- Ignore errors containing `useRootTheme():` in the global error handler `packages/@sanity/base/src/util/uncaughtErrorHandler.js`
- Introduces a `onUIError` callback in `ErrorHandler`, so that the theme error is caught and stored in `SanityRoot`’s react state.
- Now renders the theme error message when the error was uncaught.
- Renders the install command as `npm install @sanity/ui@0.32.8` instead of `npm install @sanity/ui@"^0.32.8"` (which was causing issues, and would actually set `"@sanity/ui": "^0.32.9"` in the user’s package.json)

### What to review

- Make sure uncaught exceptions containing `useRootTheme():` makes the Studio render a helpful error message.

### Notes for release

- Uncaught exceptions containing `useRootTheme():` now makes the Studio render a helpful error message.

### DIFF
--- a/examples/test-studio/sanity.json
+++ b/examples/test-studio/sanity.json
@@ -102,6 +102,10 @@
     {
       "implements": "part:@sanity/base/tool",
       "path": "src/tools/css-variables/tool"
+    },
+    {
+      "implements": "part:@sanity/base/tool",
+      "path": "src/tools/test-ui-error/tool"
     }
   ],
   "env": {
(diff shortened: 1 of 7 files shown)


bjoerge (approved):
LGTM!

PR #2268, opened 2021-01-29
Add basic tooling for tracking Studio performance metrics

bjoerge:
### Description

This provides the basic tooling for tracking various performance metrics in the studio. 

As a humble start, this PR only includes a single perf test that's using puppeteer to open the studio, measures how long it takes to type a-z into a text field in the form builder and post the result (along with additional metadata about package versions, commit hash, and the environment the test ran in) to a sanity dataset.

It also runs the  perf tests in a github action on push, so we can now track how each commit/release affects this number, making it possible for us to catch performance regressions early and see the effect of various measures.

I also introduced local environment variable config through `.env` files (and unified the setup between perf and e2e tests)

I've kept it minimal for now, but there's a bunch of possible improvements, in addition to adding more test cases:
- Post a comment on the commit/pr that summarises how performance is affected by it
- Historical stats in the performance studio
- We currently use the same studio both for running the performance tests and for tracking the stats. I can't think of any practical issues this can cause right now, but we should consider splitting it out so that the perf tests are run separately.
- Make it easier to add new tests.

### Notes for release
Added basic tooling for tracking various performance related metrics in the Sanity Studio.


.gitignore
```diff
@@ -48,3 +48,5 @@ jspm_packages
 # Lockfiles
 yarn.lock
 package-lock.json
+
+.env
```
mariuslundgard, line 52: Would be good if we would ignore `.env.local` instead, I think, so that `.env` can be checked in and used for defaults?

examples/perf-studio/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 - 2020 Sanity.io
```
mariuslundgard, line 3: 2021 :-)

bjoerge: Copy/paste - but yeah. Let's instead update all licenses in a separate commit?

mariuslundgard (approved):
Very cool 💯

examples/perf-studio/LICENSE
mariuslundgard: Sure :-)

PR #2279, opened 2021-02-02
[core] Close JSDOM window after schema is resolved

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

This is an odd one, so hang tight:

- In order to evaluate and get hold of the Sanity schema within a Node.js context, we have to provide a fake browser environment. This is because of input components and such that are imported and assume a browser environment, which tends to crash. We use JSDom for this.
- That's not always enough. JSDom is missing a few globals, such as `requestAnimationFrame`, `InputEvent` and a few other oddities that some plugins feature detect or call. So we explicitly put these things onto the global, if not already there, and remove them again once  the schema is resolved. Pain.
- Some components add _events_ or event set _timers_ - `DOMReady` being one such. In certain cases, there is a race condition where a timer or event is fired _after_ the schema has been resolved. By then, we've done our best to "clean up" the JSDOM context. If that event or timer tries to access, say... `document`: boom 💥.

**Description**

This PR adds a missing piece of the puzzle! When we do the cleanup, by calling `window.close()`, JSDOM will remove all registered timers and events. Hurray! 

**Note for release**

- Fix bug where a `document is not defined` error might be thrown while attempting to deploy a GraphQL API

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/getSanitySchema.js
+++ b/packages/@sanity/core/src/actions/graphql/getSanitySchema.js
@@ -30,6 +30,7 @@ function provideFakeGlobals() {
 
 function getSanitySchema(basePath) {
   const domCleanup = jsdomGlobal()
+  const windowCleanup = () => global.window.close()
   const globalCleanup = provideFakeGlobals()
   const contextCleanup = requireContext.register()
   const cleanupFileLoader = pirates.addHook(
@@ -49,6 +50,7 @@ function getSanitySchema(basePath) {
   cleanupFileLoader()
   contextCleanup()
   globalCleanup()
+  windowCleanup()
   domCleanup()
 
   return schema


saasen:
That sounds like it was a hard nut to crack. Great job!

packages/@sanity/core/src/actions/graphql/getSanitySchema.js
```diff
@@ -30,6 +30,7 @@ function provideFakeGlobals() {
 
 function getSanitySchema(basePath) {
   const domCleanup = jsdomGlobal()
+  const windowCleanup = () => global.window.close()
   const globalCleanup = provideFakeGlobals()
```
saasen, line 34: The only thing I could think of is that if this cleanup fits in the `globalCleanup`?

saasen (approved):
Just had a small comment, otherwise LGTM!

PR #2283, opened 2021-02-02
[chore] Tie perf run results to hardware profile to get more comparable results

bjoerge:
### Description
When running the perf tests in a github action, the hardware of the VM you get has some variations in cpu speed. This PR creates a hardware profile for each variation and ties the perf test result to the hardware profile it was ran in. This gives us more comparable results for similar hardware profiles.

In addition this PR adds a perf summary document that ​​we use to track results of perf runs for a given hardware profile (as of this PR only the most recent 200 results are kept)

### What to review
​
See it in action here:
https://perf-studio-f6m14jgdv.sanity.build/desk

### Notes for release
​
N/A


benedicteb (approved):
Cool! :grinning:

PR #2284, opened 2021-02-02
[chore] Add `e2e-test-studio`

mariuslundgard:
This change introduces a new example Studio with which to run e2e tests (Cypress).


package.json
```diff
@@ -20,10 +20,11 @@
     "deploy-test": "npm run build && cd examples/test-studio && sanity deploy",
     "deploy-movies": "npm run build && cd examples/movies-studio && sanity deploy",
     "lint": "npm run eslint -- .",
-    "eslint": "node_modules/.bin/eslint --ext=.js,.jsx,.mjs,.ts,.tsx --quiet",
+    "eslint": "eslint --ext=.js,.jsx,.mjs,.ts,.tsx --quiet",
```
benedicteb, line 23: Without something like npx won't this break if you haven't added `node_modules/.bin` to your path locally? 🙂

package.json
```diff
@@ -36,22 +37,22 @@
...
+    "prettier": "prettier '**/*.{ts,js,tsx,jsx}'",
     "lint-next": "npm run eslint -- `git diff --name-only next --diff-filter ACMRTUXB`",
     "verify-format": "lerna run clean && npm run prettier -- -c",
     "postinstall": "npm run bootstrap",
     "vercel-build": "npm run build && cd examples/test-studio && npm run build -- -y",
     "type-check": "tsc -b ./packages --force",
-    "cypress": "cypress",
-    "cypress:open": "cypress open --config-file cypress/cypress.json"
+    "cypress:open": "cypress open --config-file cypress/cypress.json",
+    "cypress:run": "cypress run --config-file cypress/cypress.json"
```
benedicteb, line 55: Maybe we should use `-` instead of `:` to be consistent with the rest of the file? 🙂

package.json
mariuslundgard: It won’t break without `npx`, since `npm` will look for the executable in `node_modules/.bin` by default.

mariuslundgard: https://stackoverflow.com/questions/33270880/run-npm-scripts-using-local-deps

package.json
mariuslundgard: Sure :)

mariuslundgard: On second though, we do it a few other scripts already.

Isn't it a good way to give scripts some modifiers? E.g.:

- `npm run lint-examples`
- `npm run lint-examples:fix`

PR #2289, opened 2021-02-03
[base] Map legacy theme values from CSS variables to Sanity UI theme object

mariuslundgard:
This change makes `@sanity/base` generate a complete `@sanity/ui` theme object based on the custom CSS properties in a given Sanity Studio. This means that the CSS properties defined by a 3rd party developer in an `override.css` will be used to generate the theme object (TS-typed by `@sanity/ui`).

#### Test default theme

- Start the server: `npm run design-studio`
- Check that both non-`@sanity/ui` and `@sanity/ui` components use similar colors, based on the CSS properties.

#### Test custom theme

- Stop the server. Change `@import './sanity.css';` to `@import './dark.css';` in `examples/design-studio/themes/dev.css`.
- Start the server: `npm run design-studio`
- Check that both non-`@sanity/ui` and `@sanity/ui` components use similar colors, based on the CSS properties.


packages/@sanity/base/src/components/ZIndexProvider.tsx
```diff
@@ -1,26 +1,26 @@
...
 
 const defaults = {
   /*
     used by
     - Navbar
   */
-  navbar: parseInt(vars['--zindex-navbar'], 10) || 200,
-  navbarPopover: parseInt(vars['--zindex-navbar-popover'], 10) || 500000,
-  navbarDialog: parseInt(vars['--zindex-navbar-dialog'], 10) || 500001,
+  navbar: parseInt(cssCustomProperties['--zindex-navbar'], 10) || 200,
```
bjoerge, line 9: Using parseInt here may obscure an error if the property doesn't exist on cssCustomProperties: parseInt will return NaN and the default value will be assigned instead. Perhaps make a separate function that ensures that the property exists and is a number or else throws?

bjoerge (approved):

rexxars (approved):

benedicteb (approved):

PR #2291, opened 2021-02-04
[form-builder] Fix wrong invalid state in UrlInput

mariuslundgard:
### Description

- Fixes a bug where the URLInput was erroneously showing as `:invalid` when `allowRelative: true` was set in the URI validation.

Report:
> I have a standard URL field that is rather angry on the new version, but the same field with the same data is fine on the previous version. There are no validation errors on it in either version.

### How to test

- Go to the URLs test: http://localhost:3333/test/desk/urlsTest
- Check that plain URL fields and URL fields with `allowRelative` behave correctly.

### DIFF
--- a/examples/test-studio/schemas/urls.js
+++ b/examples/test-studio/schemas/urls.js
@@ -16,6 +16,21 @@ export default {
       type: 'url',
       title: 'Plain url',
       description: 'A plain URL field',
+      // validation: (Rule) =>
+      //   Rule.uri({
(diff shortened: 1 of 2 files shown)


benedicteb (approved):

bjoerge (approved):

sargsyan:
Hello, seems that the same issue happens when`relativeOnly : true` is used. Could you please check and fix if my statement is true.

PR #2293, opened 2021-02-04
[form-builder] Fix bug where datetime popper would be hidden behind dialogs

benedicteb:
### Description
​
We use `react-datepicker`, it let's us specify the component that should wrap the datepicker. Previously we used `React.usePortal` directly to do this. But that doesn't work optimally with the new `Layer`-system of Sanity UI.

Change it to be wrapped inside a `Layer` so it's put on top.

This also adds an e2e-test which makes sure the datepicker is visible both on top of a dialog and without a dialog.
​
### What to review

Can we use `Layer` "naively" in this way or should we specify the z-index directly?

I also believe this changes where the datepicker is put into the DOM. Does that create some consequences for other inputs that I haven't noticed while testing?
​
### Notes for release
​
* Fixed a bug where datepicker was hidden behind dialogs

### DIFF
--- a/cypress/integration/datetimeInput.js
+++ b/cypress/integration/datetimeInput.js
@@ -1,0 +1,28 @@
+describe('DatetimeInput react-datepicker popper', () => {
+  beforeEach(() => {
+    cy.visit(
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
Nice 👍

mariuslundgard (approved):
💯

PR #2298, opened 2021-02-10
[base] Prevent printing warnings when importing ZIndexProvider

mariuslundgard:
This change prevents console warnings from being printed when importing `ZIndexProvider`, by wrapping the conversion in a function.


bjoerge (approved):
LGTM!

PR #2302, opened 2021-02-11
[desk-tool] Fix layering of the document pane header

mariuslundgard:
This fixes the issue with the document version dropdowns being rendered behind nested array item dialogs.


bjoerge (approved):
LGTM!

PR #2303, opened 2021-02-12
[initial-value-templates] Initial values for any field

juicycleff:
This PR adds support for deep initial values fields of all `schema` types except `arrays`.

It supports

1. Object Schema,
2. String Schema
3. Boolean Schema and
4. Number Schema.

However, if an array type has its initial value set, it will not resolve fields that are missing, but only fields that were specified


bjoerge:
Superseded by #2444

PR #2308, opened 2021-02-17
perf(form-builder): wrap the asset source plugin components in withDocument instead of the image input

bjoerge:
Previously the ImageInput was wrapped by the `withDocument` HOC that provided it with the current document as a prop. This was then passed to the asset source plugin for convenience. This meant that whenever the document changed the image input would then receive a new `document` prop. Since it only forwarded the document to the asset source plugin, this patch instead wraps the asset source plugin in `withDocument`.

This should improve the render performance of the ImageInput since it no longer receives the enclosing `document` as a prop, which changes on every keystroke.

### Note for release
- Skip uneccessary re-rendring of image input when the edited document is changed


mariuslundgard (approved):

rexxars (approved):
💯 
I think changes to the `options.sources` won't be reflected since it's set in the constructor, but I guess that's not new behavior so let's go with this for now :)

bjoerge:
> I think changes to the options.sources won't be reflected since it's set in the constructor, but I guess that's not new behavior so let's go with this for now :)

Yeah, noticed this too. Been fixed in the ui-refactor branch, so will be improved upon in an upcoming PR :)

PR #2309, opened 2021-02-17
perf(form-builder): include path and focusPath in FormBuilderInput's shouldComponentUpdate check

bjoerge:
This includes a shallow equality check of `path` and `focusPath` in the FormBuilderInput. These props rarely change, and doing this check should be fairly cheap (especially when using the pathFor utility introduced in #2307 for the passed `path`/`focusPath` props)

### Note for release
- Various performance optimizations


mariuslundgard (approved):

rexxars (approved):
Unfortunately some old lint errors that follows this one, but 👍  for the change itself!

PR #2311, opened 2021-02-17
[chore] Add optimizations to prevent rerendering

mariuslundgard:
This change adds various performance optimizations to prevent rendering.

### DIFF
--- a/packages/@sanity/base/src/presence/overlay/StickyOverlay.tsx
+++ b/packages/@sanity/base/src/presence/overlay/StickyOverlay.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
 
-import React from 'react'
+import React, {memo, useMemo} from 'react'
 import {flatten, groupBy, orderBy, sortBy} from 'lodash'
 import {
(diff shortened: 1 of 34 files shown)


packages/@sanity/base/src/presence/overlay/StickyOverlay.tsx
```diff
@@ -173,11 +172,23 @@ export function StickyOverlay(props: Props) {
       return (
         <>
           {[
-            renderDock('top', margins, grouped.top, counts.nearTop),
+            <PresenceDock
```
bjoerge, line 175: Perhaps we could avoid the array around here now?

bjoerge (approved):
Ran some numbers for a deeply nested field locally and it looks very good.


### Current release
```
Samples: 6, avg: 2472ms
```

### This PR
```
Samples: 6, avg: 1739ms
```

Great work!

packages/@sanity/base/src/presence/overlay/StickyOverlay.tsx
mariuslundgard: Sure!

PR #2317, opened 2021-02-19
Fix type error on explicit return type

andyrichardson:
## About

Use implicit function return type


andyrichardson:
@bjoerge ☝️

bjoerge:
Squashed all of these into a single commit for the async filter support: https://github.com/sanity-io/sanity/commit/36345a5a64aa722d2218bd5032b6b42810bc9afc

Thanks again!

PR #2320, opened 2021-02-21
feat(cli): load `.env` files before running any command

### LINKED ISSUE #2318: Allow initial configuration of projectId and dataset from environment variables
Unless I'm misunderstanding what I'm seeing in development, it seems as though `projectId` and `dataset` config variables can be **overridden** by environment variables(`SANITY_STUDIO_API_PROJECT_ID` and `SANITY_STUDIO_API_DATASET`), but must be **initially defined** in `sanity.json`. I have been trying to set these configs solely from environment variables after removing the `api` section from `sanity.json`, but it seems that the environment variables are not inspected until *after* `sanity.json` is validated.

I would like to be able to omit the `api` section from `sanity.json` and define those two config items using environment variables, so that as much config as possible exists in env vars rather than being split between `sanity.json` and env vars.

For now I have my default config in `sanity.json` and my overrides in an environment-specific env file (e.g. `.env.production`). This works but I would prefer to be able to define everything in environment variables for consistency. I have also tried setting `projectId` and `dataset` to empty strings in `sanity.json`, but this appears to mess up which dotenv file is read by Sanity.

Apologies if I've misunderstood what I'm observing locally – pointers appreciated if this is in fact already possible. Thanks for your ongoing work on this fantastic product!

rexxars:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

When running CLI commands, any environment variables set in `.env`-files are not respected. For instance, running `sanity graphql deploy`, `sanity deploy` and `sanity users list` uses the values defined in `sanity.json` instead of respecting any `.env` file overrides.

**Description**

This PR ensures any `.env`-file is read and used before running any CLI commands.

Fixes #2318

**Note for release**

- `.env` files will now be loaded before running any Sanity CLI command

**Checklist** 

- [x]  The code is linted

### DIFF
--- a/packages/@sanity/cli/src/cli.js
+++ b/packages/@sanity/cli/src/cli.js
@@ -2,6 +2,7 @@
 // eslint-disable-next-line import/no-unassigned-import
 import path from 'path'
 import chalk from 'chalk'
+import dotenv from 'dotenv'
 import fse from 'fs-extra'
 import neatStack from 'neat-stack'
 import resolveFrom from 'resolve-from'
@@ -23,6 +24,11 @@ module.exports = async function runCli(cliRoot) {
   const isInit = args.groupOrCommand === 'init' && args.argsWithoutOptions[0] !== 'plugin'
   const cwd = checkCwdPresence()
   const workDir = isInit ? process.cwd() : resolveRootDir(cwd)
+
+  // Try to load .env files from the sanity studio directory
+  // eslint-disable-next-line no-process-env
+  const env = process.env.SANITY_ACTIVE_ENV || process.env.NODE_ENV || 'development'
+  dotenv.config({path: path.join(workDir, `.env.${env}`)})
 
   await updateNotifier({pkg, cwd, workDir}).notify()
 
--- a/packages/@sanity/cli/package.json
+++ b/packages/@sanity/cli/package.json
@@ -48,6 +48,7 @@
     "debug": "^3.1.0",
     "decompress": "^4.2.0",
     "deep-sort-object": "^1.0.1",
+    "dotenv": "^8.2.0",
     "es6-promisify": "^6.0.0",
     "eventsource": "^1.0.6",
     "execa": "^1.0.0",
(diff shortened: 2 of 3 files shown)


packages/@sanity/cli/src/cli.js
```diff
@@ -24,6 +25,11 @@ module.exports = async function runCli(cliRoot) {
   const cwd = checkCwdPresence()
   const workDir = isInit ? process.cwd() : resolveRootDir(cwd)
 
+  // Try to load .env files from the sanity studio directory
+  // eslint-disable-next-line no-process-env
+  const env = process.env.SANITY_ACTIVE_ENV || process.env.NODE_ENV || 'development'
+  dotenv.config({path: path.join(workDir, `.env.${env}`)})
```
bjoerge, line 31: What is the use case for having one env file per environment? I'm a bit worried this will encourage keeping `env`-files under source control which is sort of defeating the purpose of env files. It's also advised against as per the official documentation https://github.com/motdotla/dotenv#should-i-have-multiple-env-files. Perhaps we could support it through an explicit option instead?

bjoerge (approved):
Added a comment/question about having multiple .env files as the default. LGTM beyond that 💯

packages/@sanity/cli/src/cli.js
rexxars: > What is the use case for having one env file per environment? 

Different project IDs/datasets per environment, usually, without having to "remember" the different IDs.
```
// .env.staging
SANITY_STUDIO_API_PROJECT_ID=stagingId
```

> I'm a bit worried this will encourage keeping env-files under source control which is sort of defeating the purpose of env files.

Generally agree, but this concern was not raised back in #1610 - so it's a bit late to change this now, as this is the current and documented behavior: https://www.sanity.io/docs/studio-environment-variables#dot-env-files-env-f5e9e3158896

I'd rather not break this right now - but might be worth changing if we have to do a batch of breaking changes later?

bjoerge: Makes a lot of sense. Thanks for the clarification!

PR #2328, opened 2021-02-22
chore(form-builder): avoid destroying dom tree when enabling/disabling change indicator for a field

bjoerge:
Currently a different DOM tree is returned when toggling the change indicator for a field: https://github.com/sanity-io/sanity/blob/9f38c5a26652a2ad7f1aa69405c724e23a1d60c6/packages/%40sanity/base/src/change-indicators/ChangeIndicator.tsx#L240-L242

As a consequence the react reconciler will destroy the child nodes entirely and recreate when it is toggled. This is especially noticable when opening array items for editing in which the preview will be destroyed and recreated when opening/closing for edit, often triggering a reload of images etc.

### Before
before

### After
after

This patch makes sure we maintain the same dom structure when toggling change indicators for a field.

**Note for release**

- Prevent reloading/re-mounting of array items after open/close for edit

### DIFF
--- a/packages/@sanity/base/src/change-indicators/ChangeBar.tsx
+++ b/packages/@sanity/base/src/change-indicators/ChangeBar.tsx
@@ -40,8 +40,9 @@ export function ChangeBar(props: {
   children: React.ReactNode
   hasFocus: boolean
   isChanged: boolean
+  disabled?: boolean
 }) {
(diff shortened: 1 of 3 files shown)


mariuslundgard (approved):
Great stuff 👍

Issue #2312, opened 2021-02-18
Sanity v.2.3.8 scroll bug


myronov:
**Describe the bug**

Page is scrolled after delete button is clicked

**To Reproduce**

1. Go to 'any block with delete button
2. Click on 'delete'
3. Page scrolls up by itself

**Expected behavior**

There should be no scroll jumps.

**Screenshots**

https://user-images.githubusercontent.com/55454191/108393489-9f6f2480-720b-11eb-9f64-980f2b8d3a9a.mp4

**Which versions of Sanity are you using?**

@sanity/cli                    2.1.4 (latest: 2.3.8)
@sanity/base                   2.3.8 (up to date)
@sanity/components             2.2.6 (up to date)
@sanity/core                   2.3.8 (up to date)
@sanity/default-layout         2.3.8 (up to date)
@sanity/default-login          2.2.6 (up to date)
@sanity/desk-tool              2.3.8 (up to date)
@sanity/google-maps-input      2.3.8 (up to date)
@sanity/image-url           0.140.22 (up to date)
@sanity/production-preview     2.2.6 (up to date)
@sanity/vision                 2.2.6 (up to date)

**What operating system are you using?**

Windows 10

**Which versions of Node.js / npm are you running?**

6.14.6
v14.15.4

**Additional context**

**Security issue?**

Any security issues should be submitted directly to [security@sanity.io](mailto:security@sanity.io). In order to determine whether you are dealing with a security issue, ask yourself these two questions:

* Can I access something that's not mine, or something I shouldn't have access to?
* Can I disable something for other people? If the answer to either of those two questions are "yes", then you're probably dealing with a security issue. Note that even if you answer "no" to both questions, you may still be dealing with a security issue, so if you're unsure, just email us at [security@sanity.io](mailto:security@sanity.io.

Issue #2280, opened 2021-02-02
Clicking an item further down on a desk page causes disruptive scroll jump


fusepilot:
**Describe the bug**

Scrolling down on a long page, clicking an element, or simply within the page causes the page to scroll jump. This forces the user to relocate what they were initially trying to click on and click it a second time to finally open it.

**To Reproduce**

1. Scroll down on a long desk page.
2. Click somewhere on the page.
3. See the page scroll jump

**Expected behavior**

The page should not unexpectedly scroll when the user clicks on something.

**Screenshots**

Here's a video of the problem occurring:

https://user-images.githubusercontent.com/187164/106563879-5320a680-64e1-11eb-8a06-e9c135a2fb33.mp4

**Which versions of Sanity are you using?**

```
$ sanity versions
@sanity/cli                 2.2.6 (up to date)
@sanity/base                2.3.0 (up to date)
@sanity/cli                 2.2.6 (up to date)
@sanity/components          2.2.6 (up to date)
@sanity/core                2.2.6 (up to date)
@sanity/dashboard           2.2.6 (up to date)
@sanity/default-layout      2.3.0 (up to date)
@sanity/default-login       2.2.6 (up to date)
@sanity/desk-tool           2.3.1 (up to date)
@sanity/production-preview  2.2.6 (up to date)
@sanity/vision              2.2.6 (up to date)
```

**What operating system are you using?**

macOS 10.15.7

**Which versions of Node.js / npm are you running?**

```
$ npm -v && node -v
6.14.9
v14.15.3
```

**Additional context**

- I noticed this behavior starting with the 2.0.0 release.
- Happens in the latest Chrome and Safari. 
- Happens in both development and production.
- No custom input components are being used.

bjoerge:
Thank you for a great bug report, @fusepilot.

I believe this issue has been fixed in a more recent release, If you're still experiencing it, feel free to open a new issue with a link to this one.

bjoerge:
Did some more digging and turns out this is still an issue, so reopening. Working on a fix.

PR #2329, opened 2021-02-23
fix(form-builder): move handling of ensuring parent item visibility inputs

bjoerge:
When a document node value is beeing edited in a dialog we want to make sure the input/preview of the parent/enclosing is scrolled into view in the background. Originally we solved this by responding to changes in the first element of the focus path (e.g. the top most document field name). Whenever the first element in the focus path changed, we'd then scroll to the dom element that had this as the value for `data-focus-path`. As a consequence, opening an item for edit would scroll the editor to the path of the _parent input_ into view, and this would sometimes trigger a scroll jump actually scrolling the field out of view because scrollIntoView would ensure the top edge of the parent input was in view. Instead, what we really wanted was to have the edited _item_ in view.

Fixes: #2280
Fixes: #2312

### Note for release
- Fixed various issues causing unexpected scroll jumps in the editor

As a bonus, this PR also adds a subtle focus ring on the background element that is currently being edited.

image

### DIFF
--- a/packages/@sanity/form-builder/src/hooks/useDidUpdate.ts
+++ b/packages/@sanity/form-builder/src/hooks/useDidUpdate.ts
@@ -1,0 +1,32 @@
+/* eslint-disable no-shadow */
+// eslint is giving false positives no-shadow for typed arguments here
+
+import {useEffect} from 'react'
+import {usePrevious} from './usePrevious'
+
+/**
+ * A hook for doing side effects as a response to a change in a hook value between renders
+ * Usage:
+ * ```js
+ * useDidUpdate(hasFocus, (hadFocus, hasFocus) => {
+ *  if (hasFocus) {
+ *    scrollIntoView(elementRef.current)
+ *   }
+ * })
+ * ```
+ * @param current The value you want to respond to changes in
+ * @param didUpdate Callback to run when the value changes
+ */
+export function useDidUpdate<T>(current: T, didUpdate: (previous: T, current: T) => void): void
+export function useDidUpdate<T>(
+  current: T,
+  didUpdate: (previous: T, current: T | undefined) => void
+): void {
+  const previous = usePrevious<T>(current)
+
+  useEffect(() => {
+    if (previous !== current) {
+      didUpdate(previous, current)
+    }
+  }, [didUpdate, current, previous])
+}
--- a/packages/@sanity/form-builder/src/hooks/usePrevious.ts
+++ b/packages/@sanity/form-builder/src/hooks/usePrevious.ts
@@ -1,0 +1,14 @@
+import {useEffect, useRef} from 'react'
+
+/**
+ * A hook that returns the previous value of a component variable
+ * This might be provided by React in the future (https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state)
+ * @param value The value to track. Will return undefined for first render
+ */
+export function usePrevious<T>(value: T): T {
+  const ref = useRef<T>()
+  useEffect(() => {
+    ref.current = value
+  }, [value])
+  return ref.current
+}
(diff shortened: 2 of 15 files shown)


packages/@sanity/form-builder/src/hooks/useDidUpdate.ts
```diff
@@ -0,0 +1,17 @@
+import {useEffect} from 'react'
+import {usePrevious} from './usePrevious'
+
+export function useDidUpdate<T>(value: T, didUpdate: (current: T, previous: T) => void): void
```
mariuslundgard, line 4: This is a hard thing to name right. Maybe a more explicit naming could be `useOnChange`?

And the signature of the callback be similar to `shouldComponentUpdate`? e.g.
```
type OnChangeCallback<T> = (prev: T, next: T) => void
```

packages/@sanity/form-builder/src/hooks/usePrevious.ts
```diff
@@ -0,0 +1,10 @@
+import {useEffect, useRef} from 'react'
+
+export function usePrevious<T>(value: T, initial: T): T
+export function usePrevious<T>(value: T, initial?: T): T | undefined {
```
mariuslundgard, line 4: I don’t understand what this hook is doing. Can you add a description?

mariuslundgard: And since it’s only being used by `useDidUpdate`, maybe this is not needed but should instead be part of `useDidUpdate`?

packages/@sanity/form-builder/src/hooks/useDidUpdate.ts
bjoerge: I originally considered onChange but landed on useDidUpdate because onChange is usually about events coming from the UI whereas this one is "the value updated since previous render", so it more closely aligns with  componentDidUpdate semantics.

Agree on the order of callback params - your suggestion also aligns better with setState, reduce, etc. which expects the previous first.

packages/@sanity/form-builder/src/hooks/usePrevious.ts
bjoerge: Yes, good point. Thanks!

mariuslundgard: I still think it’s a little unnecessary to make a new hook for this, since it’s used by another hook and nothing else yet. But it’s not critical of course.

packages/@sanity/form-builder/src/inputs/ArrayInput/item/ArrayInputListItem.css
```diff
@@ -13,12 +13,16 @@
   background: var(--component-bg);
   border-radius: var(--border-radius-small);
   padding: calc(var(--extra-small-padding) - 1px);
-  transition: box-shadow 100ms;
+  transition: box-shadow 500ms;
```
mariuslundgard, line 16: This is a nice touch! I think a duration of `250ms` would be a tad more subtle (in the long run).

packages/@sanity/form-builder/src/inputs/ArrayInput/item/ArrayInputGridItem.css
```diff
@@ -11,12 +11,15 @@
   background: var(--component-bg);
   border-radius: var(--border-radius-small);
   padding: calc(var(--extra-small-padding) - 1px);
-  transition: box-shadow 100ms;
+  transition: box-shadow 500ms;
```
mariuslundgard, line 14: Same as below.

mariuslundgard (approved):

packages/@sanity/form-builder/src/hooks/usePrevious.ts
bjoerge: I generally agree that we shouldn't prematurely make things generic and divide up functions/hooks etc that isn't reused, but since this is such a common hook (I've needed it several times myself) and I don't see the harm in having it separate and readily available for the next person that needs it. It's also better separation of concerns and IMO makes the intent behind the resulting code more clear.

PR #2331, opened 2021-02-24
Add schema type defintions

andyrichardson:
# About

This would require a fair amount of more work but I thought this could be good as an initial example.

This would add type safety and improve DX.

> Note: There are limitations in TS which prevent us from getting more precise types (e.g. `Exclude<string, 'array'> == string` )

## Screenshots

Screenshot 2021-02-24 at 16 21 37

Screenshot 2021-02-24 at 16 41 12


rexxars:
Thanks Andy! We've already started some of this work over in the [@sanity/types](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/types/src/schema/types.ts) module, but from a "compiled schema type" perspective (which differs slightly from the raw schema definition). It would be ideal to keep the definitions there and find a way to share the common properties/interfaces for both the compiled and uncompiled versions

andyrichardson:
Thanks for the response @rexxars!

I'll close this for now as it was more of a proposal than anything else. It's not something I have a lot of time to work on atm :+1:

PR #2356, opened 2021-03-04
Fix changed default for graphql deployment (non null)

### LINKED ISSUE #2355: --non-null-document-fields flag defaults to false
**Describe the bug**

Despite being in a [non-major release](https://github.com/sanity-io/sanity/releases/tag/v2.2.2), the `--non-null-document-fields` flag is now `false` by default.

Users of pre v2.2.2 releases will have noticed that updating to v2.2.2 changed all fields in the schema to now be optional (unless explicitly adding the `--non-null-document-fields` flag which previously was default functionality).

andyrichardson:
## About

Fix #2355 

Related to [this change](https://github.com/sanity-io/sanity/commit/511f9a8ffd167f4d464e3a14ed83d1175673b546)

CC: @rexxars

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/deployApiAction.js
+++ b/packages/@sanity/core/src/actions/graphql/deployApiAction.js
@@ -157,7 +157,7 @@ function parseCliFlags(args) {
     .option('dataset', {type: 'string'})
     .option('tag', {type: 'string', default: 'default'})
     .option('generation', {type: 'string'})
-    .option('non-null-document-fields', {type: 'boolean', default: false})
+    .option('non-null-document-fields', {type: 'boolean', default: true})
     .option('playground', {type: 'boolean'})
     .option('force', {type: 'boolean'}).argv
 }
--- a/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
+++ b/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
@@ -375,7 +375,7 @@ function extractFromSanitySchema(sanitySchema, extractOptions = {}) {
   }
 
   function getDocumentInterfaceFields() {
-    const isNullable = typeof nonNullDocumentFields === 'boolean' ? !nonNullDocumentFields : true
+    const isNullable = !nonNullDocumentFields;
     return [
       {
         fieldName: '_id',


packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
```diff
@@ -375,7 +375,7 @@ function extractFromSanitySchema(sanitySchema, extractOptions = {}) {
   }
 
   function getDocumentInterfaceFields() {
-    const isNullable = typeof nonNullDocumentFields === 'boolean' ? !nonNullDocumentFields : true
```
andyrichardson, line 378: I'm guessing this is where the default to true condition was expected to occur.

This doesn't happen because the condition `typeof nonNullDocumentFields === 'boolean'` will always be true.

packages/@sanity/core/src/actions/graphql/deployApiAction.js
```diff
@@ -157,7 +157,7 @@ function parseCliFlags(args) {
     .option('dataset', {type: 'string'})
     .option('tag', {type: 'string', default: 'default'})
     .option('generation', {type: 'string'})
-    .option('non-null-document-fields', {type: 'boolean', default: false})
+    .option('non-null-document-fields', {type: 'boolean', default: true})
```
andyrichardson, line 160: This defaulted to `true` for all releases prior to `v2.5.0`

rexxars:
Nullable document fields has actually been the default since [v1.149.12](https://github.com/sanity-io/sanity/releases/tag/v1.149.12), so this flag is actually correct. The flag was introduced to reverse this behavior - in other words, to make the fields non-nullable.

packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
rexxars: Yeah this check is unnecessary given that the flag is explicitly a boolean now - good point.

andyrichardson:
> Nullable document fields has actually been the default since [v1.149.12](https://github.com/sanity-io/sanity/releases/tag/v1.149.12), so this flag is actually correct. The flag was introduced to reverse this behavior - in other words, to make the fields non-nullable.

Hi @rexxars thanks for clarifying - it sounds like something has gone wrong on our end then (going from having `--non-nullable` to not).

PR #2360, opened 2021-03-05
fix(desk-tool): make sure there's a valid selection state before opening the review changes panel

bjoerge:

- [ ]  Yes
- [x]  No

**Description**

This fixes an issue that would sometimes make the studio crash with an error saying "start required". This would in some rare cases happen when history controller's selectionState was in a loading state. This patch fixes this by instead checking against `selectionState=="range"`, which is the only state where we are guaranteed to have the start document.

**Note for release**
- Fixes an issue that could cause an error saying "start required" during review changes

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,7 +4,7 @@
       "@babel/env",
       {
         "targets": {
-          "node": "8",
+          "node": "10",
           "chrome": "59",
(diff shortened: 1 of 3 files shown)


judofyr (approved):
LGTM!

PR #2368, opened 2021-03-08
test(form-builder): give all inputs a data-testid with their given path

bjoerge:
This adds a `data-test-id` attribute around every input component in the sanity studio in order to provide a stable way of selecting inputs in e2e tests.

In addition, this PR also removes the `data-focus-path` attribute that we no longer use after merging #2329.

Based on #2367

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -22,8 +22,7 @@ Cypress.Commands.add('login', (sanitySessionToken) => {
 })
 
 Cypress.Commands.add('getField', (fieldName) => {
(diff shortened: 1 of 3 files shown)


mariuslundgard (approved):
LGTM

bjoerge:
> What do we mean by "local path"? The same as relative path?

I realize "local path" wasn't the best choice of wording here. It's the path passed to FormBuilderInput, which is usually `['someFieldName']` in the case of object inputs and `[{_key: 'someKey'}]` for array items, so guess you can call it "relative" path yeah. Ideally it should be the global path, but we should look into the performance implications of resolving that on every render, so think this good enough:tm: for now.

PR #2369, opened 2021-03-08
fix(desk-tool): fix position of pane header and presence avatars

mariuslundgard:
- Fixes an issue that made the document pane header not have `position: sticky`. 
- Fixes the positioning of the presence avatars on narrow screens.


bjoerge (approved):
LGTM!

PR #2373, opened 2021-03-11
fix(google-maps-input): make search results visible by raising z-index

rexxars:
The search results for the google maps input falls behind the modal because of z-index changes. This PR adds a high number to the z-index in order to fix it - the rationale behind simply increasing the z-index is mentioned in the CSS file.

Also replaces the API key used in examples to allow it to work on `*.sanity.build`


bjoerge (approved):
LGTM!

PR #2374, opened 2021-03-11
fix(form-builder): hide slug generate button if field has no source

rexxars:
If a slug field has no `source` option, it shouldn't show a "Generate" button. Currently it does, and shows an error.


bjoerge (approved):
Good catch. LGTM!

Issue #2376, opened 2021-03-12
Expose default implementation of UI elements for use with Custom Input components


andyrichardson:
## About

It would be great if users could make use of the Form Builder components.

### Example

```tsx
import ArrayInput from 'part:@sanity/form-builder/input/array';

// ...
inputComponent: (props) => {
  // Example render condition
  if (props.value.length < 5) {
    return null;
  }
  
  return <ArrayInput {...props} />
},
```

### Additional info

Currently, an implementation of the form builder components [doesn't exist](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/form-builder/sanity.json#L23).

riccardogiorato:
Was this issue closed by your PR? https://github.com/sanity-io/sanity/pull/2377
Do you have a complete example on how to use this?

Where do I define the inputComponent with inside a custom ArrayInput?

andyrichardson:
Hey @riccardogiorato, my PR added the _ArrayInput_ component  but the remainder are still missing.

> Where do I define the inputComponent with inside a custom ArrayInput?

Inside of a schema [(docs)](https://www.sanity.io/docs/custom-input-widgets)

kmelve:
This is largely solved with the new customization framework in Sanity Studio v3.

PR #2377, opened 2021-03-12
Add default ArrayInput implementation

andyrichardson:
## About

Related to https://github.com/sanity-io/sanity/issues/2376

Uses approach documented in the [parts docs](https://www.sanity.io/docs/parts#de2ed8f04eba).

> This allows other plugins to compose from the original implementation, extending it as need be.

## Usage

```tsx
import ArrayInput from 'part:@sanity/form-builder/input/array-default';
```

### DIFF
--- a/packages/@sanity/form-builder/sanity.json
+++ b/packages/@sanity/form-builder/sanity.json
@@ -24,6 +24,11 @@
       "description": "Input component for the core array type"
     },
     {
+      "name": "part:@sanity/form-builder/input/array-default",
+      "implements": "part:@sanity/form-builder/input/array",
+      "path": "inputs/ArrayInput/ArrayInput"
+    },
+    {
       "name": "part:@sanity/form-builder/input/array/functions",
       "description": "React component defining array functions such as add item"
     },


pawelngei:
We were looking for exactly this in our project! Great idea!

Does your implementation allow to edit specific items easily, so that they do something inline instead of opening modals? Does it allow them access to `document`?

bjoerge:
LGTM, thank you @andyrichardson!

fabien:
@andyrichardson @bjoerge is there a reason not to export all the other input components as well?

As far as I can see, only `part:@sanity/form-builder/input/array-default` is exported.

andyrichardson:
@fabien - not at all. Ideally all of these components would be exported

GLips:
@pawelngei do you have example code you used to create a custom `ArrayInput`? I'm working on a very similar use case and can't currently see a way to do it without rewriting large chunks of `ArrayInput`, `ArrayItem`, `ItemRow`, etc.

pawelngei:
@GLips in the end I rewrote only ArrayItem, left the rest alone. Sadly, the code is copyrighted and I can't share it :disappointed:

GLips:
@pawelngei Thanks! That pointer is still super helpful. One question if you can indulge me—how did you inject the rewritten `ArrayItem`?

pawelngei:
Ah, sorry for the confusion, this was some time ago. I just checked, we ended up not even modifying ArrayItem itself.

What we needed was a better list display + a way to handle custom actions when an array item is opened. Both of those could be addressed with the custom preview components ( https://www.sanity.io/docs/previews-list-views#d82fd035385f ) and input components ( https://www.sanity.io/docs/custom-input-widgets ). The second is what opens within a modal once you click on the ArrayItem.

Far from perfect, but workable.

PR #2388, opened 2021-03-17
fix: strip special characters from studio search

bjoerge:
Previously we'd only strip whitespace, comma and dash. This PR makes sure we ignore more special characters from the search string when doing a search from the Studio.

*Note for release*
Studio search will now strip special characters from the search string before searching


simen (approved):
This works great!

rexxars (changes requested):
It looks like "Will: match" now works as expected, but "Sanity.io" no longer produces a match

rexxars (approved):
Looks good now!

PR #2398, opened 2021-03-22
fix: date input emit value on right format

bjoerge:
This fixes an accidental breaking change we shipped with the new date inputs released in [v2.6.3](https://github.com/sanity-io/sanity/releases/tag/v2.6.3) that made the input for `date` fields emit the full ISO string instead of an `YYYY-MM-DD`-formatted string. We should ship this as soon as possible.

- [ ]  Yes
- [x]  No

**Note for release**
Fix a regression introduced in v2.6.3 that caused values coming from `date` inputs to be stored on the same format as `datetime` fields.


rexxars (approved):
👍

PR #2399, opened 2021-03-22
feat(base): add `altText` field to asset types

rexxars:
The [media plugin](https://github.com/robinpyon/sanity-plugin-media) defines a new "altText" field on the image and file assets, which we agree should be an option, but wasn't currently included in the schema. With this change, GraphQL schemas and such will have access to the field.


bjoerge (approved):
👍

PR #2401, opened 2021-03-24
fix(base): prevent cancellation of list fetching while receiving mutation events

bjoerge:
Currently, if a list fetch takes > 1 second and receive mutation events in the meantime, the request would be cancelled and a new one re-initated. In cases where you'd have many ongoing mutations happening while loading a list this could make the list load forever (or until the mutations stops from arriving).

This patch fixes the issue by instead using an rxjs operator that will ignore received mutation events during ongoing fetces, and issue a new one at the trailing end for any mutation event received during the preceding fetch.

**Note for release**
- Fixes an issue that could in some cases cause loading of document lists in the studio to take a long time.


rexxars (approved):
👍

PR #2404, opened 2021-03-24
fix(base): strip symbols from user initials

rexxars:
When trying to find a user's initials and the name contains symbols (eg `Espen "Hooverdam"`), we can end up with initials like  `E"`. This PR removes symbols from the name before trying to find the first character of the last word, which would change the previous examples initials into `EH`. 

Couldn't find any regex class or similar that would catch _all_ cases, for instance we can't use `\w` as we want to Make `Albert Åberg` into `AÅ`, not `AB`. Better than nothing?

### DIFF
--- a/packages/@sanity/base/src/components/UserAvatar.tsx
+++ b/packages/@sanity/base/src/components/UserAvatar.tsx
@@ -25,11 +25,14 @@ interface UnloadedUserProps extends BaseProps {
 
 type UserProps = LoadedUserProps | UnloadedUserProps
 
+const symbols = /[^\p{Alpha}\p{White_Space}]/gu
+const whitespace = /\p{White_Space}+/u
+
 function nameToInitials(fullName: string) {
-  const namesArray = fullName.split(' ')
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
LGTM!

Think it's [safe](https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes) for us now to use unicode property escapes given the browsers we support, so it could _potentially_ be improved with something like this:
```js
const namesArray = fullName
  .replace(/[^\p{Alpha}\p{White_Space}]/ug, '')
  .split(/\p{White_Space}+/u)
```

Do we also want to make sure it's uppercased?

rexxars:
> ```js
> const namesArray = fullName
>   .replace(/[^\p{Alpha}\p{White_Space}]/ug, '')
>   .split(/\p{White_Space}+/u)
> ```
> 
> Do we also want to make sure it's uppercased?

TIL about the unicode regexp classes! Let's go with that! And yes, let's uppercase the initials, too :)

PR #2405, opened 2021-03-25
feat(base): remove unused intl parts

rexxars:
As far as I can tell, we don't actually use or render these wrappers and providers, and I couldn't find anything using this on a global github search either. Probably remnants of an ancient past? 

We'll want to reintroduce i18n more properly in the future, so it would be nice to get rid of these for now.


bjoerge (approved):
Totally leftovers from ancient times 💯

PR #2410, opened 2021-03-26
[@sanity/core] Fix nullable array children in GraphQL schema

### LINKED ISSUE #2384: Remove Nullable array types from GraphQL schema
## About

When deploying a Sanity GraphQL endpoint, all array types are nullable.

Generated schemas have the following format for array types.

```gql
type MyDocument {
  field: [T]
}
```

Sanity will never return null within an array of documents so the type should instead be

```gql
type MyDocument {
  field: [T!]
}
```

> Note: This is different to #1430

### Nullable field

It's fair to assume this is possible if a given field does not have a value/relationship.

```js
{
  field: null
}
```

### Array of type T

Given a field exists, this is a valid response

```js
{
  field: [item1, item2, item3]
}
```

### Array of nullable type T

I'm not aware of any condition where sanity would return an array of documents including null values

```js
{
  field: [item1, null, item3]
}
```

andyrichardson:

- [ ]  Maintenance

- [ ]  Yes
- [x]  No

**Current behavior**

Fix #2384

**Description**

When an array is returned from Sanity, only values that exist in the array are returned.

There is no use case (to my knowledge) where sanity would return an array where some of the items in the array have a value of `null`.

```js
// Type implies this is possible
[someValue, null, someValue]

// In reality, sanity doesn't return null values in arrays
[someValue, someValue]
```

**Note for release**

**Checklist** 

- [ ]  The code is linted

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
+++ b/packages/@sanity/core/src/actions/graphql/extractFromSanitySchema.js
@@ -256,15 +256,21 @@ function extractFromSanitySchema(sanitySchema, extractOptions = {}) {
     }
 
     if (isReference(child)) {
-      return getReferenceDefinition(child, arrayDef)
+      return {
+        ...getReferenceDefinition(child, arrayDef),
+        isNullable: false,
+      };
     }
 
     // In the case of nested scalars, recurse (markdown -> longText -> text -> string)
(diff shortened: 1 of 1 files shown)


andyrichardson:
Not needed - see https://github.com/sanity-io/sanity/issues/2384#issuecomment-810963584

PR #2444, opened 2021-04-23
feat(initial-value-templates): add support for defining initial values for all schema types

bjoerge:
### Description
This adds support for initial values for any schema type, including resolving initial values when adding new array items both for "regular" arrays and portable text arrays.

This PR also includes a few smaller refactorings needed to support the feature.

#### Limitations / gotchas
- There's a max recursion limit currently set to 10 for recursive types that has initial values. Beyond that the fields will be left blank.
- If an async initial value fails to resolve when adding a new array item, an error will be displayed and you will still be able to edit the item. 
- There's currently no timeout when resolving an initial value for array items, so if the async operation never finishes it can effectively block adding new array items entirely.

### What to review
- Initial value templates
- Initial value definitions on schema types
- Verify that adding new array items where the item type has an initial value definition works as expected
- Verify that error handling works as expected

There's a schema type that captures different cases around initial values in the deployed test studio for this branch here: http://test-studio-git-feat-ch2841initial-values-for-any-field.sanity.build/test/desk/initialValuesTest

### Notes for release
- Adds support for initial values on any schema type


packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/Toolbar.tsx
```diff
@@ -32,18 +38,95 @@ function PTEToolbar(props: Props) {
...
+      return resolveInitialValueForType((type as any) as SchemaType)
+        .then((value) => {
+          if (isSlow) {
+            toast.push({
+              closable: true,
+              id: 'resolving-initial-value',
+              status: 'info',
+              duration: 1000,
+              title: 'Initial value resolved',
+            })
```
rexxars, line 65: Not sure we need to show this, as the value appears?

rexxars (approved):
This is really impressive work 👏 
Looks solid, and can't find any bugs.

LGTM!

packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/Toolbar.tsx
bjoerge: I found no way to close an existing toast, so this will replace the message in the "Resolving initial value…"-toast and then make sure it's closed.

bjoerge: Decreased the duration to 500, so it disappears a bit faster.

PR #2449, opened 2021-04-27
fix(field): fix z-index issues of tooltip and revert dialog

mariuslundgard:
### Description

- The poppers (dialog and tooltip) in `@sanity/field` now use the `LegacyLayerProvider` to set the z-offset to a higher number than the poppers in the `@sanity/form-builder`.
- This fixes the issues by making sure that the revert changes dialog is always on top of the form builder, even when viewing a deeply nested array item.

### What to review

- Open a document with deeply nested array items (2 level or deeper).
- Edit a field in a deeply nested array item.
- Click the `Revert changes` button below the changed field, and make sure the dialog is available (on both wide and narrow screens).

### Notes for release

- Fixes a z-index issue with revert changes dialogs.

### DIFF
--- a/packages/@sanity/base/package.json
+++ b/packages/@sanity/base/package.json
@@ -45,7 +45,7 @@
     "@sanity/structure": "2.9.1",
     "@sanity/transaction-collator": "2.7.4",
     "@sanity/types": "2.9.1",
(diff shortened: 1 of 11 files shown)


bjoerge (approved):
LGTM!

PR #2350, opened 2021-03-03
Migrate array inputs to use Sanity UI

bjoerge:
### Description
​This migrates the different variations of array inputs (except Portable Text Editor) to use Sanity UI under the hood.

### What to review
Everything related to different variations of array inputs

Note: this PR includes some commits at the beginning that's not *strictly* related to the migrated array inputs. The array inputs migration work can be reviewed individually in this range 6fa09acc761fd110409ec70c2858c4f1ff299910...bb7cb54e0f6d2478aca7e123d402e60af8ae0653
​
### Notes for release

- Migrates default array inputs in the studio to use [Sanity UI](https://sanity.io/ui) under the hood​

PR #2450, opened 2021-04-27
perf(form-builder): improve array input performance

bjoerge:
### Description

Some low hanging performance optimizations after seeing regressions from #2350.

Comparison of typing A-Z in a text field inside an object in an array of 100 items:
*base branch*: Average of 4 samples: **5799ms**
*this branch*: Average of 4 samples: **1588ms**

### What to review

Edit an array of objects and verify that presence/markers works as before, but faster.

### Notes for release
- Improved performance of editing array items


packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
```diff
@@ -22,6 +22,7 @@ interface ArrayInputListItemProps {
   index: number
   compareValue?: any[]
   markers: Marker[]
+  itemKey: string | undefined
```
mariuslundgard, line 25: Nitpick: nice with `itemKey?: string` here?

mariuslundgard (approved):

packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
bjoerge: It's not equivalent. I want `itemKey` to be explicitly given, but allow its value to be `string | undefined`, but `itemKey?: string` will allow omitting the prop at the usage site.

PR #2452, opened 2021-04-28
build: update `@sanity/color` dependency

mariuslundgard:
### Description

- This change updates `@sanity/color` from 1.0.0 to 2.0.15 in `@sanity/base` (required updating code in the user color manager + form validation component).
- This change adds `@sanity/color` as dependency where it was missing.

### What to review

- The studio builds and runs using the new version of `@sanity/color`.
- Look especially at presence avatars (if the studio doesn’t crash when rendering user avatars, it works)


bjoerge (approved):
🎉💯

PR #2453, opened 2021-04-28
perf(mutator): remove duplicated createIfNotExists

judofyr:
### Description

Detect when there are multiple createIfNotExists in the same transaction and then only keep the first one. This works nicely
with the way @sanity/base uses createIfNotExists to ensure that a draft version of a document exists.

### Notes for release

- Reduce the size of mutations when editing text fields.


judofyr:
Btw: The lint errors are not related to changes in this PR.

bjoerge (approved):
LGTM!

PR #2477, opened 2021-05-05
chore(schema): warn if inputComponent is set on type but doesn't look like a React component

bjoerge:
### Description

When setting `inputComponent` on a type or field it's easy to import the wrong symbol or forget to export the input component. This PR adds a schema warning if we see the existence of a `inputComponent`-property on the type or field  definition, but it doesn't seem to be a React component. This warning will be displayed in the console along with other schema warnings. Example:

image

Note: didn't find a reliable way of checking whether a value is a React Component that did not import React itself, so wrote a basic function to detect it. It's limited to checking whether the value is a function or has a symbol property added by React for some of their built in higher order components like forwardRef, memo, etc, and is only meant to catch the most obvious cases. Worst case we might get some false positives here, but it should be pretty straightforward to add new cases as we discover them.

### What to review

Add `inputComponent: undefined` to a type in your schema, and verify that the schema warning appears in the developer console.

### Notes for release

- Warns if the `inputComponent` for a type or field doesn't look like a React component

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/utils/isReactComponentIsh.ts
+++ b/packages/@sanity/schema/src/sanity/validation/utils/isReactComponentIsh.ts
@@ -1,0 +1,9 @@
+const REACT_SYM_RE = /^Symbol\(react\..+\)$/
+
+export function isReactComponentIsh(value: any) {
+  const type = typeof value
+  return (
+    type === 'function' ||
+    (typeof value?.$$typeof === 'symbol' && REACT_SYM_RE.test(String(value?.$$typeof)))
+  )
+}
--- a/examples/test-studio/schemas/customInputs.js
+++ b/examples/test-studio/schemas/customInputs.js
@@ -9,6 +9,8 @@ export default {
   title: 'Custom input tests',
   type: 'document',
   icon,
+  // this should make a validation warning to appear in the console
+  inputComponent: 'NOT A REACT COMPONENT',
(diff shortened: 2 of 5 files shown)


packages/@sanity/schema/src/sanity/validation/utils/isReactComponentIsh.ts
```diff
@@ -0,0 +1,9 @@
+const REACT_SYM_RE = /^Symbol\(react\..+\)$/
+
+export function isReactComponentIsh(value: any) {
+  const type = typeof value
+  return (
+    type === 'function' ||
+    (typeof value?.$$typeof === 'symbol' && REACT_SYM_RE.test(String(value?.$$typeof)))
+  )
+}
```
rexxars, line 9: Could we use [react-is](https://www.npmjs.com/package/react-is) instead?

rexxars (approved):

packages/@sanity/schema/src/sanity/validation/utils/isReactComponentIsh.ts
bjoerge: ah, realized now. It allows too much (e.g. strings), which doesn't make all that much sense for an input-component.

bjoerge: (added a code comment about this)

PR #2481, opened 2021-05-06
fix: moment/date-fns parse error in diff component

bjoerge:
### Description

We currently only support moment date format tokens when defining a custom date format for the `date` and `datetime` schema types. However, the Diff component for date and datetime values were using `date-fns` which operates with a different set of formatting tokens. This again could crash the diff component when using a custom format with tokens not supported by date-fns.

### What to review

Changing date and datetime values and make sure the diff looks right and doesn't cause an error.

### Notes for release

- Fixes an issue that could crash date and datetime diffs when using a custom format


rexxars (approved):
Thanks for spotting and fixing this!

PR #2491, opened 2021-05-11
fix(default-layout): protect against undefined state tool

dcilke:
### Description
Fix a bug where the desk tool crashes when opened from a link in the studio.

image

### What to review
This was difficult to reproduce on dev versions of the studio. A production build is required.

### Notes for release
Fixed issue where opening internal links in a new tab causes the studio to crash

### DIFF
--- a/packages/@sanity/default-layout/src/DefaultLayout.tsx
+++ b/packages/@sanity/default-layout/src/DefaultLayout.tsx
@@ -138,6 +138,7 @@ class DefaultLayout extends React.PureComponent<Props, State> {
     const {tools, router} = this.props
     const {createMenuIsOpen, menuIsOpen, searchIsOpen} = this.state
     const isOverlayVisible = menuIsOpen || searchIsOpen
+    const tool = router.state?.tool || ''
 
     return (
       <div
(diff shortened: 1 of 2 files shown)


packages/@sanity/default-layout/src/DefaultLayout.tsx
```diff
@@ -138,6 +138,7 @@ class DefaultLayout extends React.PureComponent<Props, State> {
     const {tools, router} = this.props
     const {createMenuIsOpen, menuIsOpen, searchIsOpen} = this.state
     const isOverlayVisible = menuIsOpen || searchIsOpen
+    const tool = router.state && router.state.tool ? router.state.tool : ''
```
mariuslundgard, line 141: Nitpick: this could be simplified to `const tool = router.state?.tool`?

Is the idea of using the empty string (and not e.g. `undefined`) to not change the typing?

packages/@sanity/default-layout/src/navbar/Navbar.tsx
```diff
@@ -62,6 +62,7 @@ export default function Navbar(props: Props) {
   const rootState = HAS_SPACES && router.state.space ? {space: router.state.space} : {}
   const className = classNames(styles.root, showToolMenu && styles.withToolMenu)
   const searchClassName = classNames(styles.search, searchIsOpen && styles.searchIsOpen)
+  const tool = router.state && router.state.tool ? router.state.tool : ''
```
mariuslundgard, line 65: (Same comment as above)

mariuslundgard (approved):
This looks good to me! Just have some non-blocking style comments :-)

packages/@sanity/default-layout/src/DefaultLayout.tsx
dcilke: 👍
Also, since the majority of components which took the tool as input required a string, I felt it was a safer to let the default value be an empty string.

PR #2493, opened 2021-05-11
Adds a plain custom string input component to the CLI plugin init

brob:
### Description
This adds a new "starter" plugin to the `sanity init plugin` command.

This is in the hopes that a clear, clean custom input component plugin can help folks get started a bit quicker at creating their own.

### What to review
It'll be good to test the CLI for it, as well as take a look at the plugin it will install which you can find here: https://github.com/sanity-io/sanity-plugin-custom-string

### Notes for release

`sanity init plugin` now has an option to create a starting point for a custom string input to help create your next custom input plugin.

### DIFF
--- a/packages/@sanity/cli/src/actions/init-plugin/pluginTemplates.js
+++ b/packages/@sanity/cli/src/actions/init-plugin/pluginTemplates.js
@@ -15,14 +15,14 @@ module.exports = [
     url: 'https://github.com/sanity-io/plugin-template-tool-with-routing/archive/master.zip',
   },
   {
-    value: 'chessInput',
-    name: 'Chess board input component w/ block preview',
-    url: 'https://github.com/sanity-io/plugin-template-chess-input/archive/master.zip',
-  },
-  {
     value: 'customInput',
     name: 'A custom string input starting point',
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/actions/init-plugin/pluginTemplates.js
```diff
@@ -19,6 +19,11 @@ module.exports = [
     name: 'Chess board input component w/ block preview',
     url: 'https://github.com/sanity-io/plugin-template-chess-input/archive/master.zip',
   },
+  {
+    value: 'customInput',
+    name: 'A custom string input starting point',
+    url: 'https://github.com/sanity-io/sanity-plugin-custom-string/archive/refs/heads/main.zip',
+  },
```
kmelve, line 26: I would move this above the Chess Board input

```suggestion
  {
    value: 'customInput',
    name: 'A custom string input starting point',
    url: 'https://github.com/sanity-io/sanity-plugin-custom-string/archive/refs/heads/main.zip',
  },
```

kmelve (changes requested):
Looks good! I think we should move it above the chess board input since it's what more people wants

rexxars:
Closing this as we're getting closer to releasing v3 with a new plugin API.

PR #2501, opened 2021-05-19
feat(client): add support for tagged requests

rexxars:
### Description

This PR introduces client and studio support for "tagged requests": an optional `tag` query parameter can be sent to our APIs. Down the line this will allow for easier grouping of requests in statistics and similar.

The studio will now use a base prefix of `sanity.studio` - all requests going through the studio client will in other words have _at least_ `sanity.studio` as `tag` - and specifying an explicit tag will add to it, eg `sanity.studio.my.thing`. It's not critical that every feature specifies a tag, so we can improve the granularity here over time - but I did an initial pass and added tags for the most common interactions.

### What to review

Click around the studio and see that anything triggering requests still work:

- Desk tool panes
- Mutations
- List previews
- Global search
- References
- Review changes
- ...other things I've forgotten about...

Also taking feedback on the actual client API.

### Notes for release

- Added client support for tagged requests


judofyr (commented):
LGTM!

bjoerge (approved):
LGTM. Great work!

PR #2504, opened 2021-05-21
chore: stricter dependency checking

bjoerge:
### Description

This adds strict dependency checking to the monorepo packages using the [depcheck](https://github.com/depcheck/depcheck) package.

From now on, CI will run a depcheck script in every package folder and will fail if:
  - There are dependencies in package.json that is not in use
  - A package is used but not declared in the project's package.json

### What to review

This adds and removes a lot dependencies in almost all the packages, so should be thoroughly tested before releasing. There's a canary released that can be installed with `sanity upgrade --tag=depcheck-canary --save-exact`

The script that does the actual dependency checking is found here: https://github.com/sanity-io/sanity/blob/e42e60bbf6c0977334f496c8b4de424d79de611e/scripts/depcheck.js

### Notes for release

- Adds strict dependency checking to packages


jaska120:
I hope this PR would get more attention to enforce strict dep checks on Sanity. I have noticed while using Sanity on Yarn workspaces setup, I sometimes need to hoist deps to overcome issues on Sanity deps. This will also speed up the deps installation when correct deps are defined in all packages.

rexxars (approved):
Looks good to me! The PR is marked as a draft, are you considering additional changes to this before it's ready?

jaska120:
@bjoerge Any updates on this PR?

bjoerge:
@jaska120 It's merged and will be included in next release (likely early next week)

PR #2514, opened 2021-05-25
fix(client): make projectId optional; combine config types

### LINKED ISSUE #2508: @sanity/client fails type checking when building a NextJS project
**Describe the bug**

```
> next build

info  - Loaded env from /Users/mircea/Documents/github/radu-website-nextjs/web/.env.local
info  - Using webpack 5. Reason: no custom webpack configuration in next.config.js https://nextjs.org/docs/messages/webpack5
Failed to compile.

./src/pages/profile.tsx:11:23
Type error: Argument of type '{ projectId: string | undefined; dataset: string | undefined; apiVersion: string; token: string | undefined; useCdn: false; }' is not assignable to parameter of type 'ClientConfig'.
  Type '{ projectId: string | undefined; dataset: string | undefined; apiVersion: string; token: string | undefined; useCdn: false; }' is not assignable to type 'ProjectlessClientConfig'.
    Property 'useProjectHostname' is missing in type '{ projectId: string | undefined; dataset: string | undefined; apiVersion: string; token: string | undefined; useCdn: false; }' but required in type '{ useProjectHostname: false; projectId?: string | undefined; dataset?: string | undefined; }'.

> 11 | const client = sanity({
     |                       ^
  12 |   projectId: process.env.SANITY_IO_PROJECT_ID,
  13 |   dataset: process.env.SANITY_IO_DATASET,
  14 |   apiVersion: 'v1',
```

**To Reproduce**

1. Install the `@sanity/client@2.8.0` in a NextJS project
2. Instantiate the client
```
const client = sanity({
  projectId: process.env.SANITY_IO_PROJECT_ID,
  dataset: process.env.SANITY_IO_DATASET,
  apiVersion: 'v1',
  token: process.env.SANITY_IO_TOKEN,
  useCdn: false,
});
```
3. Build the NextJS application

It seems that the `useProjectHostname` is deprecated in the `BaseClientConfig` but still being used in the `ProjectlessClientConfig` and `ProjectClientConfig`

If I do set the `useProjectHostname` to `false` when instantiating the client, the type checking works but the client fails to fetch.
If I do set the `useProjectHostname` to `true` when instantiating the client,  the type checking fails with the following error ( but the client is able to fetch data )

```
> next build

info  - Loaded env from /Users/mircea/Documents/github/radu-website-nextjs/web/.env.local
info  - Using webpack 5. Reason: no custom webpack configuration in next.config.js https://nextjs.org/docs/messages/webpack5
Failed to compile.

./src/pages/index.tsx:13:23
Type error: Argument of type '{ useProjectHostname: true; projectId: string | undefined; dataset: string | undefined; apiVersion: string; token: string | undefined; useCdn: false; }' is not assignable to parameter of type 'ClientConfig'.
  Types of property 'projectId' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

ricokahler:
#### What changes are introduced?

Updates the types for the config object in the client to work with environment vars (e.g. `process.env.EXAMPLE_SANITY_PROJECT_ID`). This is common in tools like next.js.

#### Why are these changes introduced?

See here: https://github.com/sanity-io/sanity/issues/2508#issuecomment-848283281

#### What issue(s) does this solve? (with link, if possible)

Closes #2508

### What to review

Take a look at the types and check to see if there are any breaking changes.

### Notes for release

Allow `projectId` to be optional in all cases to prevent typescript DX issues. See #2508.

### DIFF
--- a/packages/@sanity/client/sanityClient.d.ts
+++ b/packages/@sanity/client/sanityClient.d.ts
@@ -492,7 +492,9 @@ export class Transaction {
   reset(): this
 }
 
-export interface BaseClientConfig {
+export interface ClientConfig {
+  projectId?: string
+  dataset?: string
   useCdn?: boolean
   token?: string
   apiHost?: string
@@ -522,19 +524,18 @@ export interface BaseClientConfig {
   requester?: GetItRequester
 }
 
-export type ProjectlessClientConfig = BaseClientConfig & {
-  useProjectHostname: false
-  projectId?: string
-  dataset?: string
-}
+/**
+ * @deprecated Don't use
+ */
+export type ProjectlessClientConfig = ClientConfig & {useProjectHostname: false}
 
-export type ProjectClientConfig = BaseClientConfig & {
+/**
+ * @deprecated Don't use
+ */
+export type ProjectClientConfig = ClientConfig & {
   useProjectHostname?: true
   projectId: string
-  dataset?: string
 }
-
-export type ClientConfig = ProjectClientConfig | ProjectlessClientConfig
 
 export interface RequestOptions {
   timeout?: number


rexxars (approved):
👍

PR #2516, opened 2021-05-26
feat: improve read only states

bjoerge:
### Description
This PR includes a bunch of visual improvements regarding readOnly states for input components and the Studio UI
It also includes a commit that enables presence for readOnly fields. It seemed to have been deliberately disabled, so I'm not 100% sure if this is something we want, but it seemed a bit strange at least to not display presence on a readOnly field. Thoughts?

### What to review
- Schema types and fields with `readOnly: true`

### Notes for release

- Various visual improvements for `readOnly` input components


mariuslundgard (approved):
LGTM!

PR #2528, opened 2021-05-31
fix(form-builder): render array item menu in a portal

mariuslundgard:
### Description

- Added test cases in Test Studio for showing an array menu with more than 1 object type
- Use the `constrainSize` and `portal` properties on Sanity UI’s `Popover` to make sure the popover renders in a portal element, and that the popover has a `maxHeight`.

### What to review

- Open https://test-studio-8pv11zpc4.sanity.build/test/desk/poppers;7ab62dbe-d336-4b05-9245-b61c608449f9
- Add an item under "Arrays in arrays"
- Click the "Add item…" in the nested "Arrays in arrays"
- Make sure the popover menu works as expected – even on small screens.

### Notes for release

- Fixes overflow issues with the array item menu.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/arrays/common/ArrayFunctions.tsx
+++ b/packages/@sanity/form-builder/src/inputs/arrays/common/ArrayFunctions.tsx
@@ -1,6 +1,6 @@
 import {ArraySchemaType, isReferenceSchemaType} from '@sanity/types'
 import {AddIcon} from '@sanity/icons'
-import React, {ReactNode} from 'react'
+import React, {ReactNode, useMemo} from 'react'
 import {Button, Grid, Menu, MenuButton, MenuItem} from '@sanity/ui'
 import {useId} from '@reach/auto-id'
 import PatchEvent from '../../../PatchEvent'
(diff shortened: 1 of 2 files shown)


packages/@sanity/form-builder/src/inputs/arrays/common/ArrayFunctions.tsx
```diff
@@ -38,6 +38,8 @@ export default function ArrayFunctions<MemberType>(
     insertItem(type.of[0])
   }, [type, insertItem])
 
+  const popoverProps = useMemo(() => ({constrainSize: true, portal: true}), [])
```
bjoerge, line 41: Any reason not to make this a constant?

bjoerge (approved):
Just a small comment, otherwise LGTM!

packages/@sanity/form-builder/src/inputs/arrays/common/ArrayFunctions.tsx
mariuslundgard: That feels like moving it too far away from the context. And if for some reason the value should be dynamic at a later stage, it’s a shorter path to that now, e.g.:

```tsx
function ArrayFunctions (props: ArrayFunctionProps) {
  const {portal: portalProp = true} = props
  
  const popoverProps = useMemo(() => ({constrainSize: true, portal: portalProp}), [portalProp])
```

PR #2532, opened 2021-06-02
fix: add star matcher to cypress intercept

ricokahler:
### Description

Just trying to fix the e2e tests. This adds a `*` matcher to the cypress intercept. This is probably what we want but I wonder whether or not to use the `vX` should stay in the code.

### What to review

If it passes, we should be good.


ricokahler:
superceded by f655482959b82933eea385fff3784648f10735b3

PR #2542, opened 2021-06-08
feat(form-builder): Add asset source support and file select to file input

### LINKED ISSUE #1377: The file field does not a have browse button
This is no way to select a previously uploaded file in the file field in studio. 

The file field should behave like the image field. There should be a browse button for seeing the uploaded files. 

Using the image field for files. It didn't work.

sjelfull:
### Description

This change aligns the `File` field with the `Image` field by adding asset source support, allowing users to select existing assets in addition to uploading directly. You can also disable direct uploads to the `File` field by setting `formBuilderConfig?.files.directUploads = false`.

Adding [custom asset sources](https://www.sanity.io/docs/custom-asset-sources#8a7448076430) for the `File` field follows the same pattern as the `Image` field, by adding a new part:

```js
{
  "implements": "part:@sanity/form-builder/input/file/asset-sources",
  "path": "./parts/assetSources.js"
}
```

Fixes #1377.

### What to review

- Verify that the built-in default source works
- Verify that the asset preview in the file select handles both image thumbs and non-image files.
- If you have a custom asset source available, verify that it works with said asset source.
- Check if disabling direct upload to File field works

### Notes for release

- Added asset select support to the File field
- Changed the default asset source to support non-image previews
- Added support for disabling direct upload to File field
- Fixed alt text on thumbnails in default asset source


packages/@sanity/form-builder/src/sanity/DefaultAssetSource/Asset.tsx
```diff
@@ -124,31 +132,75 @@ export default class Asset extends React.PureComponent<AssetProps, State> {
...
+                  align="center"
+                  justify="center"
+                  padding={3}
+                  style={{
+                    position: 'absolute',
+                    top: 0,
+                    left: 0,
+                    right: 0,
+                    bottom: 0,
+                  }}
```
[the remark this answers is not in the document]
sjelfull: I'll wrap it with a styled component to make it consistent with the others.

sjelfull:
For reference this is how the default source looks like:

image

And this is how the asset preview looks like, regardless if its a image that is selected:

image

I also added a tooltip that shows the complete filename if it is clipped.

image

FYI I already got some design input from @mariuslundgard but happy for more.

packages/@sanity/form-builder/src/inputs/files/FileInput/FileInput.tsx
```diff
@@ -259,17 +382,80 @@ export default class FileInput extends React.PureComponent<Props, FileInputState
...
-    // Single asset source (just a normal button)
-    // @todo add select handling here
-    return <Button mode="bleed" text="Select" />
-  }
-
   renderFields(fields: Field[]) {
     return fields.map((field) => this.renderField(field))
   }
 
+  handleSelectAssetFromSource = (assetFromSource: AssetFromSource) => {
```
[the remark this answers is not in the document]
sjelfull: Done.

packages/@sanity/form-builder/src/inputs/files/FileInput/FileInput.tsx
```diff
@@ -259,17 +399,22 @@ export default class FileInput extends React.PureComponent<Props, FileInputState
...
+    const {onChange, type, resolveUploader} = this.props
+    handleSelectAssetFromSource({
+      assetFromSource,
+      onChange,
+      type,
+      resolveUploader,
+      uploadWith: this.uploadWith.bind(this),
```
bjoerge, line 413: nit: instead of using .bind here you can instead change the `uploadWith` method to be a bound instance method, e.g. change it from:
```
  uploadWith(uploader: Uploader, file: File, assetDocumentProps: UploadOptions = {}) {
    //...
```
to:
```
  uploadWith = (uploader: Uploader, file: File, assetDocumentProps: UploadOptions = {}) => {
    //...
```

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -320,71 +321,14 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
...
-      default: {
-        throw new Error('Invalid value returned from asset source plugin')
-      }
-    }
+    handleSelectAssetFromSource({
+      assetFromSource,
+      onChange,
+      type,
+      resolveUploader,
+      uploadWith: this.uploadWith.bind(this),
```
bjoerge, line 329: Same as comment about .bind above :)

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/Asset.tsx
```diff
@@ -124,32 +140,69 @@ export default class Asset extends React.PureComponent<AssetProps, State> {
   }
 
   render() {
-    const {asset, onClick, onKeyPress, isSelected} = this.props
+    const {asset, assetType = 'image', onClick, onKeyPress, isSelected} = this.props
+    const {originalFilename, _id, mimeType, extension, url} = asset
     const {isDeleting, showUsageDialog, deleteError} = this.state
     const imgH = 200 * Math.max(1, DPI)
+    const isImageType = assetType === 'image'
+    const isImageMimetype = mimeType.includes('image')
```
bjoerge, line 148: edge case for sure, but this should probably check the first part of the mime type? Might give false positives on things like disk images? You could also consider using the `attr-accept` package that we use elsewhere for checking mime type: `accepts('image/*', mimeType)`

Side note: browsers typically only supports a [subset of all possible image types](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types) too, so might be an idea to a) check up front if it can be displayed, or b) only display types of a known format.

packages/@sanity/types/src/assets/types.ts
```diff
@@ -113,6 +113,8 @@ export type AssetFromSource = {
 }
 
 export interface AssetSourceComponentProps {
+  assetType?: 'file' | 'image'
+  dialogHeaderTitle?: string
```
bjoerge, line 117: Assume we need to document these?

bjoerge (commented):
Great work @sjelfull! Left some questions/comments/suggestions :)

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/Asset.tsx
sjelfull: How would you approach it? Better to just whitelist the ones we know are safe?

bjoerge: Maybe if we verify first that the mime type is `image/*`, then we can try to display it with a fallback in case of an error?

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/Asset.tsx
```diff
@@ -124,32 +140,69 @@ export default class Asset extends React.PureComponent<AssetProps, State> {
...
+              />
+              {isDeleting && <FullscreenSpinner />}
+            </Container>
+          )}
+          {!isImageType && (
+            <CardContainer radius={2} tone="transparent">
+              {isPreviewable && (
+                <Image
+                  alt={originalFilename}
+                  src={`${url}?h=${imgH}&fit=max`}
```
bjoerge, line 179: Since we're in the context of a file here, the image will be served from the file cdn, so these parameters will not have any effect.

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/Asset.tsx
bjoerge: Thinking some more about this I'm not really sure we want to display images uploaded as files. Since we can't get a thumbnail version for it (they are served from the file cdn, which doesn't support image url parameters), we'd in many cases end up wasting bandwidth on large image files not intended for being viewed. I think it's fair to assume that if an image is uploaded as a file asset it's not intended to be displayed as a regular image - what do you think?

rexxars: In hindsight, I agree with Bjørge on the last comment here. Not showing any previews for images makes it more obvious that you are using the wrong asset type.

sjelfull:
Added table view for file select:
image

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/useMatchesMedia.ts
```diff
@@ -0,0 +1,25 @@
+import {useState, useCallback, useEffect} from 'react'
+import {useTheme} from '@sanity/ui'
+
+export function useMatchesMedia(
```
rexxars, line 4: Given that this is only used in one place, and resides within the `DefaultAssetSource` utils, is this a bit too generic/premature? There's something about the API here that I don't find too intuitive:

```ts
const isMobile = useMatchesMedia(1)
```
I had to go digging for the implementation to find out why `1` was used. I'd prefer to wait with this generalization and see if we can bake this into `@sanity/ui` or `@sanity/base`, instead inlining the check for now.

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
...
+    title: 'Icon',
+  },
+  'image/png': {
+    title: 'PNG',
+  },
+  'image/svg+xml': {
+    title: 'SVG',
+  },
+  'image/webp': {
+    title: 'WEBP',
```
rexxars, line 26:
```suggestion
    title: 'WebP',
```

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
...
+    title: 'MP3 Audio',
+  },
+  'audio/ogg': {
+    title: 'OGG Audio',
+  },
+  'audio/wav': {
+    title: 'WAW Audio',
+  },
+  'audio/webm': {
+    title: 'WEBM Audio',
```
rexxars, line 51:
```suggestion
    title: 'WebM Audio',
```

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
+interface MimeTypeInfo {
+  title: string
+}
+
+const MIME_TYPES: Record<string, MimeTypeInfo> = {
```
rexxars, line 5: Some of these have suffixes ("Archive", "Audio") while others just have the file extension - would it make sense to hint at the type for all of them? `JPEG Image`, `AVI Video` etc?

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
...
+  },
+  'video/quicktime': {
+    title: 'QuickTime',
+  },
+  // Documents
+  'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': {
+    title: 'Microsoft Excel',
+  },
+  'application/vnd.ms-excel': {
+    title: 'Microsoft Excel',
```
rexxars, line 80:
```suggestion
    title: 'Excel Spreadsheet',
  },
  'application/vnd.ms-excel': {
    title: 'Excel Spreadsheet',
```

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
...
+    title: 'Text',
+  },
+  'text/javascript': {
+    title: 'JavaScript',
+  },
+  'text/markdown': {
+    title: 'Markdown',
+  },
+  'text/csv': {
+    title: 'text/csv',
```
rexxars, line 92:
```suggestion
    title: 'CSV File',
```

rexxars (changes requested):
Looks good for the most part! There are some undeclared dependencies that we need to add, and left some other suggestions for general code style

packages/@sanity/form-builder/src/sanity/uploads/types.ts
```diff
@@ -32,4 +32,11 @@ export type Uploader = {
   priority: number
 }
 
-export type UploaderResolver = (type: SchemaType, file: File) => Uploader | null
+export interface FileIsh {
```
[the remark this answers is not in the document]
rexxars: Realized that this code is probably based on the commits from #2547 so duplicating my comment over there

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
sjelfull: We don't use the File suffix in any of the other files, and I think CSV/Comma Separated Values reads just fine, so will skip this.

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
sjelfull: Discussed this with @jorngeorg, and we landed on it being smart to include it for all the files, so non-technical users easily can distinguish the type.

packages/@sanity/form-builder/src/inputs/files/ImageInput/types.ts
```diff
@@ -1,6 +1,11 @@
 export interface AssetRecord {
```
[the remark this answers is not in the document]
sjelfull: That is a good idea, will replace all instances with that type.

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/useMatchesMedia.ts
sjelfull: I just found out that we have a util in Sanity UI called `useMediaIndex()` for this. It is kind of similar to the utility in this PR and gives us an index that maps to the breakpoints in Sanity UI.

packages/@sanity/types/src/assets/types.ts
sjelfull: Where?

packages/@sanity/form-builder/src/sanity/DefaultAssetSource/utils/mimeType.ts
```diff
@@ -0,0 +1,132 @@
...
+    title: 'MIDI Audio',
+  },
+  'audio/mpeg': {
+    title: 'MP3 Audio',
+  },
+  'audio/ogg': {
+    title: 'OGG Audio',
+  },
+  'audio/wav': {
+    title: 'WAW Audio',
```
coreyward, line 48: I think there's a typo here. Should be “WAV Audio”.

sjelfull: Thanks, you are correct. Fixed in `next`.

PR #2544, opened 2021-06-09
feat(base): migrate `components/Document` to TypeScript

mariuslundgard:
### Description

- Rename `components/Document.js` to `components/Document.tsx`
- Refactor from `PropTypes` to TS props interface
- Use `useMemo` to prevent mapped arrays from causing value changes

### What to review

- That the HTML source of the Studio is exactly the same as before.

### Notes for release

- n/a


packages/@sanity/base/src/components/Document.tsx
```diff
@@ -0,0 +1,118 @@
...
+    staticPath: staticPathProp = '/static',
+    favicons: faviconsProp = DEFAULT_FAVICONS,
+    stylesheets: stylesheetsProp = [],
+    scripts: scriptsProp = [],
+  } = props
+
+  const basePath = basePathProp.replace(/\/+$/, '')
+  const staticPath = `${basePath}${staticPathProp}`
+
+  const stylesheets = useMemo(
```
bjoerge, line 42: Curious about the impact of using `useMemo` here. Is this component ever re-rendered? Could it have been a memo-component instead?

packages/@sanity/base/src/components/Document.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const {
+    basePath: basePathProp = '',
+    charset = 'utf-8',
+    title = 'Sanity',
+    viewport = 'width=device-width, initial-scale=1, viewport-fit=cover',
+    loading = 'Connecting to Sanity.io',
+    staticPath: staticPathProp = '/static',
+    favicons: faviconsProp = DEFAULT_FAVICONS,
+    stylesheets: stylesheetsProp = [],
+    scripts: scriptsProp = [],
```
bjoerge, line 36: Won't this invalidate the useMemos that has these props as one of their dependencies when these props aren't given? Could perhaps consider re-using the same empty array instance.

bjoerge (approved):
Left a few comments/questions, otherwise LGTM!

PR #2547, opened 2021-06-10
Use check for direct uploads support in resolveUploader + improve feedback on dragging over drop targets

bjoerge:
### Description
This refactors the resolveUploader interface so it takes a "file-like" argument (an object that has a mime type and a  name) instead of an actual `File` instance. This allows us to check the data transfer items we get from the drag event when files are dragged over the drop target, and allows us to show instant feedback about what files (if any) will be uploaded when dropped. See screenshots below.

This PR also includes a commit that factors in the `directUploads` studio configuration into the `resolveUploader` function instead of having a separate code path to handle this.

image
image
(note: the wrong pluralization in the above screenshot has been fixed)

### What to review
- Drag some files of various types from disk over to the drop area in a file/image input and see that it works as expected

### Notes for release

- Give better feedback about accepted files when dragging over file and image inputs

### DIFF
--- a/packages/@sanity/form-builder/src/sanity/uploads/types.ts
+++ b/packages/@sanity/form-builder/src/sanity/uploads/types.ts
@@ -39,4 +39,4 @@ export interface FileLike {
   name?: string
 }
 
-export type UploaderResolver = (type: SchemaType, file: File) => Uploader | null
+export type UploaderResolver = (type: SchemaType, file: FileLike) => Uploader | null
--- a/packages/@sanity/form-builder/src/sanity/uploads/types.ts
+++ b/packages/@sanity/form-builder/src/sanity/uploads/types.ts
@@ -32,4 +32,11 @@ export type Uploader = {
   priority: number
 }
 
+export interface FileLike {
+  // mime type
+  type: string
+  // file name (e.g. somefile.jpg)
+  name?: string
+}
+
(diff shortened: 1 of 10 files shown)


mariuslundgard:
Did you look into and fix the `z-index` issue already?

bjoerge:
> Did you look into and fix the z-index issue already @bjoerge ?

No, didn't get time for it. Can take a look on monday 🤓

packages/@sanity/form-builder/src/sanity/uploads/types.ts
```diff
@@ -32,4 +32,11 @@ export type Uploader = {
   priority: number
 }
 
-export type UploaderResolver = (type: SchemaType, file: File) => Uploader | null
+export interface FileIsh {
+  // mime type
+  type: string
+  // file name (e.g. somefile.jpg)
+  name?: string
+}
```
rexxars, line 40: I feel like the common pattern for these "duck-types" is to use the "Like" suffix, eg PromiseLike - could we name this FileLike instead?

rexxars (approved):
Looks good, but I would prefer `FileLike` over `FileIsh` if possible

packages/@sanity/form-builder/src/sanity/uploads/types.ts
bjoerge: Yeah, good call!

PR #2558, opened 2021-06-17
fix(code-input): ensure change indicators get correct focus path

bjoerge:
This passes on the `focusPath` to the ChangeIndicatorProvider (which provides data needed by the FormField/Fieldset components to properly render change indicators). Also wraps the "field-less" render mode with a `ChangeIndicatorProvider`.


hermanwikner:
🙏

hermanwikner (approved):

PR #2559, opened 2021-06-17
fix(block-tools) global/class var issue

skogsmaskin:
### Description

This will fix an issue when combining several HTML serializers on the same time.

### What to review

Using this global temp var here to store the serialized markDefs while we deserialize, will not support running multiple HTML-deserializers at the same time, which the test will show (fails before the fix commit).

I have a vague memory that it was a reason I used this a var like that and not put it on the class itself, but I can't seem to remember it now. :facepalm: 

However all the tests pass. And I can't really see why it needs to be a global var. There are no other matches on that used by other files.

### Notes for release

Fix issue with combining multiple block-tools HTML deserializers.

### DIFF
--- a/packages/@sanity/block-tools/package.json
+++ b/packages/@sanity/block-tools/package.json
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@sanity/schema": "2.10.0",
+    "@types/jsdom": "^12.2.4",
     "assert": "^1.4.1",
     "jest": "^26.6.3",
     "jsdom": "^12.0.0",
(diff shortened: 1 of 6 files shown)


bjoerge (approved):
Good catch, LGTM!

packages/@sanity/block-tools/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@sanity/schema": "2.10.0",
+    "@types/jsdom": "^16.2.11",
```
bjoerge, line 32: This should probably match the version of jsdom actually used?

skogsmaskin: It should. I trusted VS code's suggestion of installing it for me.

PR #2560, opened 2021-06-17
fix(form-builder): Fix overflowing PT toolbar menu inside of dialog

sjelfull:
### Description

If you have a rich text editor inside of an object alongside an image, and the editor has a ton of items in the toolbar, this causes an overflow issue where the modal content is clipped.

We have a smart `OverflowMenu` component that normally is supposed to prevent overflows in the toolbar.

The overflow in this case happens because of the content, in this case the toolbar, defines the width of the parent wrapper.
Each column inside the CSS grid wrapper by default has a width of `auto` instead of `0`.

More details here: https://css-tricks.com/preventing-a-grid-blowout/

### More details

Screenshot of the problem:
image

### Notes for release

- Fixed a problem where a rich text editor with too many toolbar items inside of a object inside of a modal caused the content to overflow and be clipped


mariuslundgard:
Great work finding the particular spot, @sjelfull  👍

bjoerge (approved):

PR #2562, opened 2021-06-18
proposal: data agnostic FormField and FormFieldset components

bjoerge:
### Description

This takes data concerns away from the `FormField`, `FormFieldSet` and their support components in `@sanity/base`, so that they can be used purely as view components (tl;dr it replaces `__unstable`-props that currently expects data, with props that expects React nodes). The form-builder package now provides their own internal `FormField`/`FormFieldSet` components that's optimized for it's own use case, that still support receiving the data props, convert them to react nodes and delegates to the `FormField`/`FormFieldSet` components imported from `@sanity/base`.

### What to review
Naming in particular. It might be confusing operating with two set of components with same names (FormField/FormFieldSet) that isn't actually the same component. We could e.g. call the ones in form-builder for `FormBuilderFormField`/`FormBuilderFormFieldSet`, but I'm all ears for suggestions.

### Notes for release
n/a


rexxars:
We're rethought how this works in v3, so I will close this.

Issue #2257, opened 2021-01-26
Field validations are broken


VWSCoronaDashboard1:
Let's say I have two document schemas. One has a title field like this:
```
    {
      name: 'title',
      type: 'localeString',
      title: 'Title',
      validation: (Rule) =>
        Rule.fields({
          nl: (fieldRule) => fieldRule.required(),
          en: (fieldRule) => fieldRule.required(),
        }),
    }
```

The other has a title field like this:
```
    {
      name: 'title',
      type: 'localeString',
      title: 'Title',
    }
```

In Sanity studio in BOTH cases the required rule is being applied. :(

I have tried working around this by adding this to the second schema: `fieldRule.reset()`. But alas, this doesn't make a difference.
It seems as if the Rule instance is being re-used across schema's. It's being reset in between, but probably the reset logic is buggy. Besides, it would probably make more sense to create a new Rule instance for each schema, IMHO.

**To Reproduce**

1. Create a schema with a localized field called title and add a required validation rule.
2. Create a schema with a localized field called title without any validations.
3. Open sanity studio
4. See that in both cases the title field is required in the UI.

**Expected behavior**

The first title field ought to be required, the second should not.

**Which versions of Sanity are you using?**

@sanity/cli                  2.2.2 (up to date)
@sanity/base                 2.2.2 (up to date)
@sanity/cli                  2.2.2 (up to date)
@sanity/components           2.1.4 (up to date)
@sanity/core                 2.2.2 (up to date)
@sanity/default-layout       2.2.2 (up to date)
@sanity/default-login        2.1.4 (up to date)
@sanity/desk-tool            2.2.2 (up to date)
@sanity/production-preview   2.1.4 (up to date)
@sanity/ui                  0.32.9 (up to date)
@sanity/vision               2.1.4 (up to date)

**Which versions of Node.js / npm are you running?**

yarn v1.22.5
npm 6.14.8
node v14.13.0

andyrichardson:
I'm also seeing this issue, it seems to be caused by reuse of "name" attributes (which imho should be fine).

I found that changing the name attribute fixed the problem. Obviously this isn't a fix and would be a breaking change.

In my case, here's the stack trace that led to the phantom "required" rule being validated

```
Error
    at module.exports (validate.js:23)
    at Rule.validate (Rule.js:111)
    at validateDocument.js:179
    at Array.map (<anonymous>)
    at validatePrimitive (validateDocument.js:176)
[...]

PR #2569, opened 2021-06-22
fix(validation): fix issue causing field validations on object types to be unintentionally re-used across contexts

bjoerge:
### Description

This fixes the issue documented in https://github.com/sanity-io/sanity/issues/2257 by not assigning field rules to the individual fields. Instead of constructing the field validation at the time it gets inferred from the schema, this PR will call the rule creator at _validation_ time, which may be suboptimal in some cases. I don't know this code well enough to find a better way, but this solution may be good enough for now. Thoughts?

### What to review
- All things related to validation, especially field validations on object types.

### Notes for release

- Fixes an issue that caused field validations on object types to be written on individual field types, causing unintentional re-use of validation rules.

### DIFF
--- a/examples/test-studio/schemas/fieldValidationInferRepro.js
+++ b/examples/test-studio/schemas/fieldValidationInferRepro.js
@@ -1,0 +1,31 @@
+export const fieldValidationInferReproSharedObject = {
+  type: 'object',
+  name: 'someObjectType',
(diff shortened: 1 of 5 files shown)


rexxars (approved):
This seems like a decent workaround. I doubt the runtime penalty will be very high.

PR #2570, opened 2021-06-22
fix(form-builder): make sure hidden/filtered fields are actually hidden

bjoerge:
### Description

The refactor of Object Input in v2.12.0 accidentally removed support for hidden object fields/field filtering. This adds it back, along with a unit test to catch future regressions.

### What to review
Verify that it looks ok implementation wise, and that hidden fields actually gets hidden

### Notes for release
- Fixes a bug introduced in v2.12.0 that made hidden fields appear visible


rexxars:
Looks good to me - the tests seem to be failing because it doesn't find the new `@sanity/types` type declarations, not sure what the right solution is here.

rexxars (approved):

PR #2574, opened 2021-06-23
Revert "feat(base): add basic permission check to default pane menu"

bjoerge:
### Description
This reverts commit 8680eea9b03867b55386f75dd9be23b08c786f38, which added a (too) simplistic permission checking to the DefaultPane's create menu. In some cases we ended up disabling it when the current user actually had access to create the document. Reverting this will always make the create button enabled an clickable, but if you try to create a document that you don't have access to, you'll see the "lack of permission"-message above the form instead.

### Notes for release
- Fixes a bug which caused the create menu to be disabled in cases where the current user actually had access to create the type of document


rexxars (approved):
👍  

/cc @ricokahler

PR #2575, opened 2021-06-23
fix(base): guard against missing user roles in InsufficientPermissionsMessage

bjoerge:
### Description
This adds a safeguard against missing roles property on the user object.

### Notes for release

N/A


vicbergquist (approved):
Great, thank you!

Issue #2576, opened 2021-06-23
Cannot read property 'from' of undefined


dnlmzw:
**Describe the bug**

I have multiple plugins that is now throwing a TypeError after recently updating Sanity.

```TypeError: Cannot read property 'from' of undefined```

**To Reproduce**

Install either the [table](https://www.sanity.io/plugins/sanity-plugin-table) or [datatable](https://www.sanity.io/plugins/sanity-plugin-datatable) plugin and try creating a new table, or deleting some rows/columns.

**Expected behavior**

Since this issue came after upgrading Sanity, in both plugins, I suspect it might be a more general issue, that could affect other plugins as well.

**Screenshots**

![Skærmbillede 2021-06-23 kl  12 20 49](https://user-images.githubusercontent.com/960055/123080654-74951700-d41d-11eb-95a1-4b917fdd4b73.png)

**Which versions of Sanity are you using?**

@sanity/cli             2.10.5 (latest: 2.12.0)
@sanity/base            2.12.0 (up to date)
@sanity/cli             2.12.0 (up to date)
@sanity/color-input     2.11.0 (up to date)
@sanity/components       2.2.6 (up to date)
@sanity/core            2.12.0 (up to date)
@sanity/dashboard       2.10.5 (up to date)
@sanity/default-layout  2.12.0 (up to date)
@sanity/default-login   2.11.1 (up to date)
@sanity/desk-tool       2.12.0 (latest: 2.12.1)
@sanity/vision          2.11.2 (up to date)

**What operating system are you using?**

MacOS Big Sur

**Which versions of Node.js / npm are you running?**

6.14.4
v12.16.3

bjoerge:
Thanks for reporting! Fixed by #2577 which will be released soon.

bjoerge:
Fix shipped in [v2.12.2](https://github.com/sanity-io/sanity/releases/tag/v2.12.2)

PR #2577, opened 2021-06-23
fix(form-builder): reexport PatchEvent default export

bjoerge:
### Description
So apparently `export * from …` does not reexport the `default` export (TIL), which is likely to cause the issue described in #2576.

### Notes for release
- Fixes an issue that caused some plugins to error with the message "Cannot read property 'from' of undefined".


rexxars (approved):
👍 LGTM

PR #2578, opened 2021-06-23
fix(field): prevent crash when unable to resolve reference type

rexxars:
### Description

The preview component seems to handle reference types just fine without us having to look up the individual type ahead of time. In fact, it might technically be incorrect to pass the target schema type along with a reference value. I'm not sure why I implemented it this way in the first place, but it seems to have started crashing when showing a diff of a schema type that can be of multiple different types.

### What to review

Diffs for references - both those of a single type and of those targeting multiple types.

### Notes for release

- Fixed bug where review changes might crash when showing change in a reference that could target more than one schema type


bjoerge (approved):
LGTM!

vicbergquist (approved):
yey, thank you! 👍

PR #2579, opened 2021-06-23
fix(cli): add missing undeclared dependency

rexxars:
### Description

`@sanity/cli` was implicitly depending on `@sanity/generate-help-url` in `clientWrapper`, but did not declare it as a dependency. This only works because other modules depend on it and the CLI bundles dependencies to a single file on publish. This PR adds the missing dependency.


bjoerge (approved):
👍🏼

PR #2581, opened 2021-06-23
feat(form-builder): add option to hide empty value in fields of type string with predefined list option

vicbergquist:
### Description

We've received feedback that some wants to be able to not be able to select an empty value when using the predefined list option for the string input (which renders a select dropdown). This PR adds an option to hide the undefined/empty value, while keeping the default behaviour of it being rendered. 

I've added an example field in the `stringsTest` document type called `Select string, without empty value`.

### What to review

- Open the test studio and test the string input with predefined lists (the `Select string` and `Select string, without empty value` fields in the string test documents (`/test/desk/stringsTest`).

### Notes for release

- add option to hide empty value in fields of type string with predefined list option

### DIFF
--- a/examples/test-studio/schemas/strings.js
+++ b/examples/test-studio/schemas/strings.js
@@ -52,6 +52,30 @@ export default {
       },
     },
     {
+      name: 'selectWithoutEmptyValue',
+      type: 'string',
+      title: 'Select string, without empty value',
+      description:
+        'Select a single string value from a set of predefined options. It should NOT be possible to unset a selected value.',
+      options: {
+        hideEmptyValue: true,
+        list: [
+          {
+            title: 'One (1)',
+            value: 'one',
+          },
+          {
+            title: 'Two (2)',
+            value: 'two',
+          },
+          {
+            title: 'Three (3)',
+            value: 'three',
+          },
+        ],
+      },
+    },
+    {
       name: 'radioSelectHorizontal',
       title: 'Select (layout: radio, direction: horizontal)',
       type: 'string',
--- a/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/SelectInput.tsx
@@ -92,7 +92,8 @@ const SelectInput = React.forwardRef(function SelectInput(
         />
       )
     }
-
+    const listItems =
+      value !== undefined && type?.options?.hideEmptyValue ? items : [EMPTY_ITEM, ...items]
     return (
       <Select
         onChange={handleSelectChange}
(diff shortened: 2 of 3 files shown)


rexxars:
This needs a bit of design/UX considerations - there are a few possible scenarios here:

- Field does not currently have any value (`undefined`)
- Field already has a value, but the value is not allowed according to the schema
- Field already has a value, and is one of the allowed values

In particular, I'm interested in the user experience aspect when a field does _not_ currently have a value. In my mind, the select has to reflect that the field doesn't have a value - so the question becomes what happens when you choose one of the values. Should it now remove the empty (undefined) value as an allowed option? Should it keep the empty value as an option for this "session" (would require us to keep a "has had undefined"-state in the component)?

Would like to hear some thoughts on this from @jorngeorg / @mikolajdobrucki

mariuslundgard:
Here’s [a sketch in Sanity UI Arcade](https://www.sanity.io/ui/arcade?mode=jsx&width=2&jsx=eJx9kbFugzAQhvc8xYkFkKKiAFUXzBC2Do3UVF2qDk58UKvURsZUqYjfvQYTkixMp7O%2Fu8%2F6nRVUMWgoY1xUpP9I1pCu4fHT5CuAbI81HjVIUXxRUSHpbWE1us7AL607ezgWOJ%2FB98cxOygbzaUAxlt6qJFNpOe5a4BASA0t6tDhkePvh6eZjZfv3MFmkYtnLl7kkplLFrl05tJbLotcKPlqaMb0fqiquHiTDemfzDXLxICWwm7Sioq2oQqFngLICskw75%2F3u5eHViuL8%2FIvuOY4x%2BOHJotGeFQPOifeytOtN74Ev%2B20lU4p23%2Br%2BfGb9EEIJB8Wvg%2BKoBMMSy6QhWYiNZ408V7RIlBKBQwPXWWXV94IRE5vpfnq8op%2FPqiw3w%3D%3D&hook=eJxtjrEKwzAQQ3d%2FxY02BP9BuuQTWrqEDFf76pYeNtjnLEn%2FvU4CpUMnCfSQ5FIsAuOMXKmDQnLd3AQ91EJnQSFdo6f7M5I3Srkdf2D0TEOTQAc5IPMN3UtrmimKgf4EiwLY%2BMRkOYUjsa7m3PSCOZC0RviO%2FgPsfgzWFX5evDsYJ6M%2BSCxBiA%3D%3D) that addresses the scenario you describe @rexxars.

bjoerge:
Not sure what do to here, so removing myself as reviewer for now. Feel free to add me back when it's ready for review :)

PR #2582, opened 2021-06-23
build(ci): skip cypress binary install on non-cypress workflows

rexxars:
### Description

Like with puppeteer, let's skip downloading the cypress binary on non-e2e github workflows


bjoerge (approved):
💯

PR #2552, opened 2021-06-11
refactor(field): migrate layout components to sanity ui - phase 3

vicbergquist:
**NOTE: depends on #2546 and #2550, so changes from those are included here. Those PR should be merged first, and then this branch should be rebased**

### Description

The smaller components ValueError, RevertChangesButton, and MetaInfo are migrated to sanity ui.

### What to review

- Test the review changes panel to make sure these changes don't break layouts or functionality

### Notes for release

- Migrated components in @sanity/field to Sanity UI and styled-components

PR #2550, opened 2021-06-11
refactor(field): migrate layout components to sanity ui - phase 2

vicbergquist:
**NOTE: depends on #2546, so changes from #2546 are included here. That PR should be merged first, and then this branch should be rebased against `next`**

### Description

Migrated more components in @sanity/field to Sanity UI and styled components. Added relevant screenshots of the before and after.

#### DiffString
Before
old-diffstring
After
new-diffstring

#### ChangeTitleSegment
Before
old-changetitlesegment
After
new-changetitlesegment

#### DiffInspectWrapper
Before
old-diffinspector
After
new-diffinspector

#### FromTo, FromToArrow, DiffCard
Before
old-diffcard-fromto-arrow
After
new-diffcard-fromto-arrow

### What to review

Make sure that things still work as they should: tooltips, change segments, layouts

### Notes for release

- Migrated components in @sanity/field to Sanity UI and styled-components

PR #2584, opened 2021-06-24
refactor(field): migrate layout components to sanity ui - final phase

vicbergquist:
**NOTE: depends on #2550 and #2552, so changes from those are included here. Those PRs should be merged first, and then this branch should be rebased**

### Description

This migrates the `FieldChange` and `GroupChange` components to sanity ui. In order to get this to work properly with the revert hover styles, I also had to make some minor changes in components that these components work with.

A change that this refactor introduces is how the hover styles are applied when hovering the revert all changes button. Previously, the individual fields inside changes in objects and array would be highlighted and not the style of the group as a whole. Now the group will have the correct hover style. See screenshots of the before and after.

**Before**
Screenshot 2021-06-24 at 11 11 19

**After**
Screenshot 2021-06-24 at 11 11 08

### What to review

- The border styles next to field changes in the review changes panel
- The hover styles when hovering the revert changes buttons

### Notes for release

- Migrated components in @sanity/field to Sanity UI and styled-components


bjoerge (approved):
LGTM! 👏🏼

PR #2585, opened 2021-06-28
feat(base): add preview override for references permissions

ricokahler:
### Description

This implements a notice for insufficient permissions for references. This approach augments the `resolveRefType` implementation that goes downstream to the `ObserveForPreview` component. This is a particularly nice spot since it's already attempting to query the ref and it has access to the type resolving the ref. If a ref could not be resolved while the related type says the reference is not `weak` then we can assume the user has insufficient permissions and return a Symbol.

Similar to the `INVALID_PREVIEW_FALLBACK`, we can catch this symbol in `ObserveForPreview` and return a similar fallback for insufficient permissions.

Questions/Notes on the approach:

~1. As a result of handling this at such a low level, the snapshot value can sometimes be a Symbol. I'm not sure if this will affect anything downstream. I believe the invalid preview config also does this that's more of an edge case. Is this okay?
2. Does it make sense to expose this symbol? It would be nice to import it and check if the snapshot matches (especially in the ReferenceInput component)?
3. If reference is marked as weak, there is no way to distinguish between non-existence and not allowed.
4. How does it look? cc @mikolajdobrucki~

Update: I remove the possibility of a symbol being sent downstream in both the invalid preview case and the insufficient permissions case. This one should be good to go now.

CleanShot 2021-06-28 at 17 34 01@2x
CleanShot 2021-06-28 at 17 33 53@2x

### What to review

Take a look at the approach and think of how this could affect things downstream. I also left a few comments in the code in this draft state I could use some help with.

Update: I also added a `meta` property to the prepared result. This is used in the ReferenceInput component to check if the snapshot is an insufficient permissions notice.

### Notes for release

Adds a notice for insufficient permissions for references

### DIFF
--- a/packages/@sanity/base/src/preview/constants.tsx
+++ b/packages/@sanity/base/src/preview/constants.tsx
@@ -7,18 +7,16 @@ import {PreparedValue} from './prepareForPreview'
 export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
 export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
 
-// The `<small>` element is used for more compatibility
-// with the different downstream preview components
-const SmallText = styled.small`
-  color: ${({theme}) => theme.sanity.color.muted.default.enabled.fg};
+// NOTE: have to use color inherit to make it work correctly with the
+// `isSelected` state in the list pane
+const IconSizer = styled(Text)`
+  color: inherit;
 `
 
 function IconWrapper({children}: {children: React.ReactNode}) {
   return (
     <Flex>
-      <Text muted size={3}>
-        {children}
-      </Text>
+      <IconSizer size={3}>{children}</IconSizer>
     </Flex>
   )
 }
@@ -26,8 +24,10 @@ function IconWrapper({children}: {children: React.ReactNode}) {
 export class InsufficientPermissionsError extends Error {}
 
 export const INVALID_PREVIEW_FALLBACK: PreparedValue = {
-  title: <SmallText>Invalid preview config</SmallText>,
-  subtitle: <SmallText>Check the error log in the console</SmallText>,
+  // The `<small>` element is used for more compatibility
+  // with the different downstream preview components.
+  title: <small>Invalid preview config</small>,
+  subtitle: <small>Check the error log in the console</small>,
   media: (
     <IconWrapper>
       <WarningOutlineIcon />
@@ -37,7 +37,9 @@ export const INVALID_PREVIEW_FALLBACK: PreparedValue = {
 }
 
 export const INSUFFICIENT_PERMISSIONS_FALLBACK: PreparedValue = {
-  title: <SmallText>Insufficient permissions to access this reference</SmallText>,
+  // The `<small>` element is used for more compatibility
+  // with the different downstream preview components.
+  title: <small>Insufficient permissions to access this reference</small>,
   media: (
     <IconWrapper>
       <AccessDeniedIcon />
--- a/packages/@sanity/base/src/preview/constants.tsx
+++ b/packages/@sanity/base/src/preview/constants.tsx
@@ -1,0 +1,47 @@
+import React from 'react'
+import styled from 'styled-components'
+import {Flex, Text} from '@sanity/ui'
+import {WarningOutlineIcon, AccessDeniedIcon} from '@sanity/icons'
+import {PreparedValue} from './prepareForPreview'
+
+export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
+export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
+
+// The `<small>` element is used for more compatibility
+// with the different downstream preview components
+const SmallText = styled.small`
+  color: ${({theme}) => theme.sanity.color.muted.default.enabled.fg};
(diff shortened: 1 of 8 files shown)


packages/@sanity/base/src/preview/prepareForPreview.ts
```diff
@@ -10,10 +10,19 @@ const EMPTY = []
...
+  title?: React.ReactNode
+  subtitle?: React.ReactNode
+  description?: React.ReactNode
+  media?: React.ReactNode
+  /**
+   * optional object used to attach meta data to the prepared result.
+   * currently used to add a flag for the invalid preview error fallback and
+   * insufficient permissions fallback
+   */
+  meta?: {type?: string}
```
ricokahler, line 25: TODO: (from call with @rexxars), add `_` prefix to `_meta`

packages/@sanity/base/src/preview/createPreviewObserver.ts
```diff
@@ -1,17 +1,20 @@
...
+
+export interface PreparedSnapshot {
   type?: SchemaType
-  snapshot: null | PrepareInvocationResult
+  snapshot: null | PreparedValue
 }
 
 // Takes a value and its type and prepares a snapshot for it that can be passed to a preview component
 export function createPreviewObserver(
-  observePaths: (value: any, paths: Path[]) => any,
```
rexxars, line 14: Why was this changed?

packages/@sanity/form-builder/src/inputs/ReferenceInput/ReferenceInput.tsx
```diff
@@ -217,6 +222,15 @@ export const ReferenceInput = forwardRef(function ReferenceInput(
       description={type.description}
     >
       <Stack space={3}>
+        {hasInsufficientPermissions && (
+          <Alert title="Insufficient permissions to access this reference" status="warning">
+            <Text as="p" muted size={1}>
```
rexxars, line 227: Are there any particular reason for using `as="p"` here? Curious

rexxars (approved):
Looks alright in my book - I'd like to have @bjoerge take a second closer look when he gets back, as he knows this code better than I do.

packages/@sanity/base/src/preview/constants.tsx
```diff
@@ -0,0 +1,49 @@
...
+  // The `<small>` element is used for more compatibility
+  // with the different downstream preview components.
+  title: <small>Invalid preview config</small>,
+  subtitle: <small>Check the error log in the console</small>,
+  media: (
+    <IconWrapper>
+      <WarningOutlineIcon />
+    </IconWrapper>
+  ),
+  _meta: {type: 'invalid_preview'},
```
bjoerge, line 36: Will this be exposed to custom preview components? If so, we might consider naming it `_internal` to make it clear that it's not meant to be relied upon?

Would it also be possible to use this to defer the rendering of the error states to the actual preview component? (having jsx in here has always felt a bit backwards to me)

packages/@sanity/base/src/preview/createPreviewObserver.ts
ricokahler: I think this was unintentional. I will revert.

packages/@sanity/base/src/preview/constants.tsx
ricokahler: > Will this be exposed to custom preview components?

Yes. This one is kind of interesting so I think I'll give the full story for more context.

Previously the invalid preview fallback returned a unique symbol that we checked and used to fall back to this invalid preview value. Custom preview components most likely didn't cover this case but that was okay because this is an exception case.

With the addition of the insufficient permissions case, we now have a case that's not an exception. In order to stay compatible with custom previous components, we don't really want to return a symbol so these fallbacks were introduced and passed down at this level.

> Would it also be possible to use this to defer the rendering of the error states to the actual preview component?

I'm also not super happy with the JSX here but it seemed like the only way to get some sort of special styling at this level without having to touch every preview component. This is why I used the `<small>` element, for semantic meaning that this is a note + a semi-hack to get it to look visually different in all contexts.

ricokahler: As far as the naming, I can change it to `_internal` if you prefer, though I'm also okay with leaving this convention as is. I would say it's valid for downstream preview components to use but if we're unsure we can name it `_internal`.

Let me know what you think!

packages/@sanity/form-builder/src/inputs/ReferenceInput/ReferenceInput.tsx
ricokahler: I was following a previous convention I saw in this document. Unless `as="p"` is the default for Text, I think this has some nice semantics.

packages/@sanity/base/src/preview/constants.tsx
bjoerge: I think it makes sense to name it `_internal_meta` (or something) for now to buy us some time to figure out a better way of passing this  type of information to preview component implementors. Easier to rename it to something official if/when we decide there's a real user need for it than going the opposite way :)

ricokahler: I tried `_internal_meta` but that was breaking our eslint camelcase rules, which is an error level rule so i went with `_internalMeta`

PR #2587, opened 2021-06-29
refactor(field): migrate basic diff components to @sanity/ui

vicbergquist:
### Description

Migrated the simpler diff components to @sanity/ui (boolean, string, url, date time, slug).

### What to review

There are no major visual changes, but some styling have been moved into the `PreviewComponent`s, which means that instead of the `DiffCard` and `FromTo` components handling the styling of the wrapper around the diff, the preview component takes care of it, which actually feels like a cleaner solution. 

### Notes for release

- Migrated url, string, datetime, slug, and number diff components to @sanity/ui


vicbergquist:
Assigned you as reviewers @hermanwikner and @ricokahler in case you want to take a look!

rexxars (approved):
Looks good to me! There are some minor spacing differences in the boolean diff, but I personally think it looks better in the new style.

PR #2588, opened 2021-06-29
fix(base): fix [object Object] for title value

ricokahler:
### Description

I noticed that sending a `title` node that's a react node (vs a string)` will make the browser tooltip say `[object object]` instead of the correct title.

This PR fixes that by adding a ref callback + `el.innerText`

### What to review

Ensure I didn't change the API of one and it works as intended. If the one looks good to you, feel free to merge it. 😄

### DIFF
--- a/packages/@sanity/base/src/__legacy/@sanity/components/previews/MediaPreview.tsx
+++ b/packages/@sanity/base/src/__legacy/@sanity/components/previews/MediaPreview.tsx
@@ -46,7 +46,7 @@ export default class MediaPreview extends React.PureComponent<MediaPreviewProps>
     }
 
     return (
-      <div className={styles.root} title={title}>
+      <div className={styles.root} title={typeof title === 'string' ? title : undefined}>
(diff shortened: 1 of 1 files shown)


bjoerge:
For the sake of simplicity I think it would be sufficient to just check whether the type of the `title`-prop is a string here. I get that this fix displays a tooltip with whatever the passed react node renders (initially), but not sure if it's worth the added complexity. Since it will not be updated after initial mount it is likely to get out of sync with the actual value too.

ricokahler:
> For the sake of simplicity I think it would be sufficient to just check whether the type of the title-prop is a string here. I get that this fix displays a tooltip with whatever the passed react node renders (initially), but not sure if it's worth the added complexity. Since it will not be updated after initial mount it is likely to get out of sync with the actual value too.

@bjoerge makes sense. I showed this to @vicbergquist a while back and she had similar thoughts so I think just checking for a string is the best path forward for this small bug. This is now a 1 line PR 😇

vicbergquist (approved):
LGTM 👍

bjoerge (approved):

PR #2590, opened 2021-06-30
refactor(dashboard): migrate to `@sanity/ui`

hermanwikner:
### Description

- Migrate all the dashboard components to `@sanity/ui` and `styled-components` 
- Add `DashboardWidget` component 

The `widget.css` is kept because current custom widgets might rely on it. However, the new pattern should be to use the new `DashboardWidget` component instead of using the `widget.css`.

### What to review

- Make sure that the dashboard works as before. Modify the dashboard configuration to see that e.g. the grid layout behaves correctly.
- Create a custom widget with the `DashboardWidget` component

### Notes for release

Migrate `@sanity/dashboard` to `@sanity/ui` and `styled-components`


packages/@sanity/dashboard/package.json
```diff
@@ -20,6 +20,8 @@
   ],
   "dependencies": {
     "@sanity/image-url": "^0.140.19",
+    "@sanity/ui":"0.34.4",
+    "@sanity/icons":"1.1.2",
```
rexxars, line 24: This may seem a little nitpicky, but tooling (npm, lerna) will sort these dependencies alphabetically and add spaces, so might as well do it consistently right away :)

```suggestion
    "@sanity/icons": "1.1.2",
    "@sanity/image-url": "^0.140.19",
    "@sanity/ui": "0.34.4",
```

packages/@sanity/dashboard/src/components/NotFoundWidget.js
```diff
@@ -1,16 +1,28 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+import {Card, Stack, Heading, Box} from '@sanity/ui'
+import styled from 'styled-components'
```
rexxars, line 4: `styled-components` needs to be added as a dependency

packages/@sanity/dashboard/src/widgets/projectInfo/ProjectInfo.js
```diff
@@ -147,54 +141,69 @@ class ProjectInfo extends React.PureComponent {
...
+                  </Stack>
+                </Stack>
               )
             })}
-          </table>
-
-          <div className={styles.footer}>
-            <AnchorButton href={getManageUrl()} bleed color="primary" kind="simple">
-              Manage project
-            </AnchorButton>
```
rexxars, line 194: Hmm, why did we remove this link?

packages/@sanity/dashboard/src/widgets/projectUsers/ProjectUsers.js
```diff
@@ -85,64 +98,74 @@ class ProjectUsers extends React.PureComponent {
...
+                <Text size={3}>
+                  <RobotIcon />
+                </Text>
               ) : (
-                <div className={styles.avatar}>
-                  {user.imageUrl && <img src={user.imageUrl} alt={user.displayName} />}
-                </div>
+                <AvatarWrapper tone="transparent">
+                  {user?.imageUrl && <img src={user.imageUrl} alt={user?.displayName} />}
+                </AvatarWrapper>
```
rexxars, line 154: Would it be possible to use the [UserAvatar](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/base/src/components/UserAvatar.tsx) component here instead?

```tsx
import {UserAvatar} from '@sanity/base/components'

<UserAvatar user={user} />
// or
<UserAvatar userId={userId} />
```

packages/@sanity/dashboard/src/components/dashboardWidget.tsx
```diff
@@ -0,0 +1,48 @@
...
+  position: relative;
+  z-index: 1;
+  overflow: auto;
+  outline: none;
+`
+
+interface DashboardWidgetProps {
+  header?: string
+  children: React.ReactNode
+  footer?: React.ReactNode
```
rexxars, line 31: Pondering if we should extend this widget component to allow for "actions" or some such - eg a way to put consistent-looking buttons/links in the footer. Let's leave it for now and revisit later, though

rexxars (changes requested):
Can't speak for the styling aspects (not my forte), but looks good code-wise.
Added a few comments inline.

packages/@sanity/dashboard/src/widgets/projectInfo/ProjectInfo.js
hermanwikner: Fixed, thank you 👍

packages/@sanity/dashboard/src/widgets/projectUsers/ProjectUsers.js
hermanwikner: I tried that, but the styling of the UserAvatar was breaking for some reason. I guess it's some styling inherited from the parent component of `DefaultPreview` (the soon outdated one) that is causing this.

Screenshot 2021-07-08 at 15 12 45

packages/@sanity/dashboard/package.json
hermanwikner: Fixed :)

packages/@sanity/dashboard/src/components/dashboardWidget.tsx
hermanwikner: Yes, I thought about it too. However I like that the component provides 100% flexibility for what can be rendered in the footer. But I agree with you, let's revisit later!

packages/@sanity/dashboard/src/widgets/projectUsers/ProjectUsers.js
rexxars: Interesting. Lets make a note to revisit this later, would be good to cut down on the number of different avatars throughout the studio

packages/@sanity/dashboard/src/widgets/projectInfo/ProjectInfo.js
rexxars: The URL doesn't actually work anymore, I realize. We should either grab the `organizationId` from the project info response above and build the correct URL (`/:orgId/project/:projectId`) or (as I suggested on Slack) build a route on manage that does the resolving and redirects

rexxars (changes requested):
Looks good now, apart from the manage URL not working.

vicbergquist (approved):
lgtm! the PR to manage to fix the project link is also deployed so this can now be merged

PR #2586, opened 2021-06-29
refactor(desk-tool): migrate document status bar to `@sanity/ui`

vicbergquist:
### Description

Migrate the document status bar components to @sanity/ui, except the `actionStateDialog` component and different dialogs used.

### What to review

- Review changes button
- Document badges
- Status badges
- publish button
- document action menu with and without custom actions

### Notes for release

- N/A

PR #2600, opened 2021-07-02
feat(desk-tool): add document action dialog size and padding options

SimeonGriggs:
### Description

[Document Actions with a Modal type](https://www.sanity.io/docs/document-actions-api#037f877ad3f1) currently don't pass `size` or `padding` options to the `<Dialog />` component.

These changes would allow you to pass those options in – with the fallback the original defaults.

A client requested this to be able to have full-window sizes modals in a Document Action.

### What to review

Create a document action with these settings:

```
    dialog: dialogOpen && {
      type: 'modal',
      size: 'auto',
      padding: 'large',
      content: <div>Hello</div>,
      onClose: () => onComplete(),
    },
```

### Notes for release

Added `size` and `padding` options to `modal` type Document Actions.

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/documentPane/statusBar/actionStateDialog.tsx
+++ b/packages/@sanity/desk-tool/src/panes/documentPane/statusBar/actionStateDialog.tsx
@@ -19,6 +19,8 @@ interface ModalDialogProps {
   content: React.ReactNode
   onClose: () => void
   showCloseButton: true
+  size?: 'small' | 'medium' | 'large' | 'auto'
+  padding?: 'none' | 'small' | 'medium' | 'large'
 }
 
 // Todo: move these to action spec/core types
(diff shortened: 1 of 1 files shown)


vicbergquist (approved):
LGTM 👍🥳

vicbergquist:
I guess the options will change again when we change to the @sanity/ui dialogs, but not sure when exactly that will be/how long it will take. Probably worth adding this first?

runeb:
@vicbergquist Perhaps we can make these options match the @sanity/ui ones now and map them until its updated so any implementer does not have to change their actions when we update?

vicbergquist:
> @vicbergquist Perhaps we can make these options match the @sanity/ui ones now and map them until its updated so any implementer does not have to change their actions when we update?

good idea 👍

vicbergquist:
Also this is related to work in #2586, which might include migrating the dialog and popover components used in document actions to @sanity/ui.

mariuslundgard:
This PR is superseeded by a complete migration of Document Actions to use `@sanity/ui` and more in this PR: https://github.com/sanity-io/sanity/pull/2636

PR #2607, opened 2021-07-07
fix(desk-tool): use the selected document revision as value for inspect document

skogsmaskin:
### Description

If you change the document to another revision, the "Inspect document" dialog would still be displaying the latest version. This is very confusing.

This is fixed by using the `displayed` value from the history provider as prop for `InspectDialog` and not the `value` which will always be the latest (current version).

### What to review

Is there is anything we didn't think of doing this change? We do have support to fallback to `initialValue` if there currently isn't anything in the`displayed` value. For instance if you select the "create" (first) revision. 

### Notes for release

Fix issue where the inspected document version is always the latest version and not the current selected version.


saasen:
This looks good to me (because we did this together).

rexxars (approved):
👍

PR #2608, opened 2021-07-07
refactor(default-login): migrate to `@sanity/ui`

hermanwikner:
### Description

Migrate default-login to `@sanity/ui`.

### What to review

Make sure that the default-login works as before. 

Note: there is a currently a workaround to remove the close button from the `Dialog`. I will create a PR to update the `Dialog` in `@sanity/ui` so that this workaround can be removed.

### Notes for release

Migrate default-login to `@sanity/ui`


packages/@sanity/default-login/src/LoginDialog.js
```diff
@@ -3,12 +3,23 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import authenticationFetcher from 'part:@sanity/base/authentication-fetcher'
 import pluginConfig from 'config:@sanity/default-login'
-import FullscreenDialog from 'part:@sanity/components/dialogs/fullscreen'
 import LoginDialogContent from 'part:@sanity/base/login-dialog-content'
```
rexxars, line 6: Can we move all of these parts to the `legacyParts` file, too?

packages/@sanity/default-login/src/LoginDialogContent.js
```diff
@@ -1,70 +1,32 @@
+/* eslint-disable react/jsx-no-bind */
 /* eslint-disable react/no-multi-comp */
 import React from 'react'
 import PropTypes from 'prop-types'
 import config from 'config:sanity'
```
rexxars, line 5: Can we move this part to `legacyParts` as `sanityConfig` or some such?

packages/@sanity/default-login/src/LoginWrapper.js
```diff
@@ -1,10 +1,9 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import userStore from 'part:@sanity/base/user'
-
 import LoginDialog from 'part:@sanity/base/login-dialog'
```
rexxars, line 4: These can be moved to `legacyParts` too

packages/@sanity/default-login/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "../../../tsconfig.settings",
+  "include": ["src"],
+  "compilerOptions": {
+    "composite": true,
+    "outDir": "./lib",
+    "rootDir": "./src",
+    "jsx": "react",
+    "noImplicitAny": false,
+  }
+}
```
rexxars, line 11: Nitpick: add an empty newline at the end :)

rexxars (approved):
Good work!
Would be good to move even more of the part imports into the `legacyParts` file, otherwise this looks good!

rexxars (changes requested):
Actually, I found an issue with the error messages being cut off:

image

I haven't tried all of the different error states, but might be the same for multiple errors?

Can be reproduced by changing `apiVersion` to `1` in `versionedClient.js` (submitted an upstream bugfix to make the CORS settings work on any api version) and changing the project ID in the test studio to some bogus non-existing project ID.

packages/@sanity/default-login/src/LoginDialog.js
hermanwikner: Yes! 👍

hermanwikner:
Thanks! 👍 

I fixed the issue with error messages being cut off. And I added more parts to `legacyParts` file!

rexxars (approved):
👍

PR #2616, opened 2021-07-12
feature(grants): add support for validating built-in $identity param

sgulseth:
### Description

To support more advanced ACL where users have access to edit their own
documents we need to inject the user id when evaluating the grant
filters

### What to review

* Is this the idiomatic way of getting the current user?

### Notes for release

add support for validating built-in $identity param


vicbergquist (approved):
Afaik it looks good! 😬

PR #2618, opened 2021-07-14
fix(base): use title, not value in prepare preview when field has predefined string options

### LINKED ISSUE #769: Use the list option when previewing strings
Currently `string` types can be configured with a `list` option, that defines a set of possible values to select from, and the title to show in the dropdown, e.g.:
```js
{
  title: 'Country',
  name: 'country',
  type: 'string',
  options: {
    list: [{value: 'NO', title: 'Norway'}, {value: 'SE', title: 'Sweden'}, /*...*/],
  },
}
```
However, when previewing values of this type, the title defined from the corresponding list option is ignored, and the string value is used as preview title instead:
image

When previewing string values that has a `list` option, we should display the `title` for the corresponding value instead of the raw string value.

cc @dan-dr

vicbergquist:
### Description

When a string field with predefined options is the first field in an object in an array, the preview renders using the value of the field, instead of the title defined in the list option (fixes #769).

I've added some logic to make sure the `title` and not the `value` is used when preparing the preview in these special cases, however I'm not sure if this is done in the right place and that it's too far up and should be done someplace else. However, with this approach the previews in document lists also use the `title` instead of the `value` if the first field in the document is a string field with predefined list options, which is nicer.

### What to review

I've added some test cases in the `arrays.js` schema in the test studio, please test it there and if possible with other combinations you can think of. Please let me know if it's suboptimal that the logic lives in `prepareForPreview.ts`.

To see this fix work in the document lists previews, add a string field with predefined list options as the first field in a document, and you'll see that it uses the `title` as the title instead of the value for that field.

### Notes for release

- Fixed an issue where the preview of a string field with predefined list options used the value of the field when selecting it for previewing instead of the title specified in the list options.

### DIFF
--- a/examples/test-studio/schemas/arrays.js
+++ b/examples/test-studio/schemas/arrays.js
@@ -43,6 +43,48 @@ export default {
       type: 'string',
     },
     {
+      name: 'predefinedStringArray',
+      title: 'Array of strings',
+      description: 'First field in object is string with list options',
+      type: 'array',
+      of: [{type: 'string'}],
+      options: {
+        list: [
+          {title: 'Cats', value: 'cats4ever'},
+          {title: 'Dogs', value: 'dogs4ever'},
+        ],
+      },
+    },
+    {
+      name: 'objectArrayWithPrefinedStringField',
+      title: 'Array of objects',
+      description: 'First field in object is string with list options',
+      type: 'array',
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
LGTM! 👏🏼

PR #2624, opened 2021-07-20
fix(base): fallback to published version if no draft exists

bjoerge:
### Description

Currently we check for update permission by stubbing the drafts document (a stub only includes _id and _type). In cases where a document is published and you have a grants check that requires certain conditions to be met on the document itself, we then end up running the grants check on the stubbed version, which may miss the fact that the published version meet the conditions specified in the grants rule. This PR changes it so we run the check against the published version (if exists) instead of the stub.

### Notes for release
- Fixes an issue that could prevent drafts from being edited even in cases where a grant rule would permit it


rexxars (approved):
👍

Issue #2631, opened 2021-07-22
Regression in base@2.13.1 with list options in object previews


geoffreyjball:
**Describe the bug**

Starting in @sanity/base@2.13.1, list options (on a `type: string`) are not implemented in a preview when using the `prepare()` function. The bug does not occur if the `prepare()` function is omitted. It occurs on both object and document previews.

**To Reproduce**

1. Create a string field that is configured with `options.list` values:
```javascript
{
  title: "Size",
  name: "size",
  type: "string",
  options: {
    list: [
      { title: "Small", value: "small" },
      { title: "Medium", value: "medium" },
      { title: "Large", value: "large" },
      { title: "X-Large", value: "xlarge" },
      { title: "Full", value: "full" },
    ],
  },
},
```
2. Create a `preview` that includes a `prepare` function:
```javascript
preview: {
  select: {
    size: "size",
  },
  prepare(selection) {
    const { size } = selection;
    return {
      title: size ? `Size: ${size}` : null;
    };
  },
},
```
3. Run `sanity start` and create a document/object.
4. Preview will return a blank title. In base@2.13.0, the title is returned correctly. Commenting out `options` in the schema will also correctly return the values in the preview.

**Expected behavior**

It's expected that the title will be returned correctly (i.e., displaying `Size: <size>`) in the preview.

**Which versions of Sanity are you using?**

base@2.13.1

**What operating system are you using?**

macOS 11.3

**Which versions of Node.js / npm are you running?**

7.18.1
v16.4.1

**Additional context**

The regression appears related to the code that was added [here](https://github.com/sanity-io/sanity/commit/c1d9d8fee3ddbc7b96b85fea98a645996e270720#diff-8b30eb59a556413ef80f0f6061d599f22016e644008caff7d715732a12edb42c), as commenting out the resulting code in `node_modules/@sanity/base/lib/preview/prepareForPreview.js` causes the preview to work as expected. That code is:
```javascript
    // Find the field the value belongs to
    var valueField = type === null || type === void 0 ? void 0 : (_type$fields = type.fields) === null || _type$fields === void 0 ? void 0 : _type$fields.find(f => f.name === selection[key]); // Check if predefined options exist

[...]

PR #2632, opened 2021-07-22
fix(base): render list items in object previews (#2631)

geoffreyjball:
Resolves issue #2631.

### Description

Modifies the return from `targetKeys.reduce()` in base/src/preview/prepareForPreview.ts. This solves the empty return encountered when using options.list and a prepare function (issue #2631 and https://sanity-io-land.slack.com/archives/C9Z7RC3V1/p1626976557089000).

### What to review

This code was compiled and tested against the [test studio](https://github.com/sanity-io/sanity/commit/d23ec5d0b3bb311b4b92653e206f15e3097ee3c1).

### Notes for release

Resolves issue #2631.

### DIFF
--- a/examples/test-studio/schemas/author.js
+++ b/examples/test-studio/schemas/author.js
@@ -1,5 +1,11 @@
 import icon from 'part:@sanity/base/user-icon'
 
+const AUTHOR_ROLES = [
+  {value: 'developer', title: 'Developer'},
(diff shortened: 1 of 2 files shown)


vicbergquist (approved):
Thanks, looks great! 🙏

bjoerge (approved):

PR #2633, opened 2021-07-23
feat(desk-tool): add timeline event for live edits

### LINKED ISSUE #2599: "Review Changes"-panel doesn't handle direct changes to the published document.
**Describe the bug**

"Review Changes"-panel doesn't handle direct changes to the published document. This is something which shouldn't be too hard to support.

**To Reproduce**

1. Create a new document.
2. Publish it.
3. Modify the published document directly through the API.
4. Open the document in Studio again. There it says that there are "changes", but there is no 
5. Also open "Review Changes". There it says that the latest version has been "Edited" (which we use to refer to the _draft_ being edited — which hasn't happened here).

**Expected behavior**

There shouldn't be any "changes" in the document.

Also, inside "review changes" there should be *two* "published" events

**Screenshots**
Incorrect document "changes" state:
Screenshot 2021-07-02 at 14 01 05

Incorrect history:
Screenshot 2021-07-02 at 14 03 07

sgulseth:
### Description

This will differentiate between when when a document is edited through
the api and edited as a draft. Fixes a bug where the studio shows that
there are changes when there are none

### What to review

Are we comparing chunk types anywhere else in the codebase that I have missed?

### Notes for release

fixes #2599

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/documentPane/documentHistory/history/chunker.ts
+++ b/packages/@sanity/desk-tool/src/panes/documentPane/documentHistory/history/chunker.ts
@@ -76,6 +76,8 @@ export function chunkFromTransaction(transaction: Transaction): Chunk {
     // We don't really know anything more at this point since the actual
     // behavior depends on the earlier state.
     type = 'delete'
+  } else if (modifiedPublished) {
+    type = 'editLive'
   }
 
   return {
--- a/packages/@sanity/desk-tool/src/panes/documentPane/documentHistory/history/chunker.ts
+++ b/packages/@sanity/desk-tool/src/panes/documentPane/documentHistory/history/chunker.ts
@@ -1,5 +1,5 @@
 /* eslint-disable no-nested-ternary */
-import {Transaction, MendozaPatch, ChunkType, Chunk} from './types'
+import {Transaction, MendozaEffectPair, MendozaPatch, ChunkType, Chunk} from './types'
 
 function canMergeEdit(type: ChunkType) {
   return type === 'create' || type === 'editDraft'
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
LGTM

judofyr:
Just jotting down some thoughts here:

Right now there's four variables: `modifiedDraft ` + `draftDeleted` and `modifiedPublished` + `publishedDeleted`. These are a bit unfortunate since if `draftDeleted` is true, then `modifiedDraft` will _also_ be true. This makes the conditions a bit hard to reason about.

If we instead change it to `draftState: 'unedited' | 'deleted' | 'upsert'` (where `upsert` means that the document has been created or updated) then we have three distinct states which completely covers the cases we need. Going over the current `if-else-if` that you've written then these is the table that we end up with:

| published | draft unedited | draft deleted | draft upsert |
| -- | -- | -- | -- |
| **unedited** | X | delete | editDraft |
| **deleted** | delete | delete | delete |
| **upsert** | liveEdit | publish | liveEdit |

(Note that both the draft and the published document can't be unedited [since then it wouldn't be part of the transaction], and that `delete` is actually later modified into `discardDraft` or `unpublish` depending on the document state.)

The only thing that looks potentially problematic here is that if you modify both the draft and the published document in the same transaction, then it will also mark that one as "liveEdit". That is probably more correct than today's "editDraft". I'm not quite sure.

judofyr (approved):
Looks sensible to me. See the previous comment for where I wrote up a better explanation for the different states. We might consider changing the code to match that (and be table-based instead of if-based).

sgulseth:
@judofyr Added the simpler logic for determining chunk type. However I'm falling back to 'liveEdit', should we also add an `unknown` type?

judofyr (approved):
💯

PR #2637, opened 2021-07-26
fix(form-builder): prevent studio crash when component passed to `withDocument()` is missing a focus() method

bjoerge:
### Description

If the component passed to `withDocument()` doesn't expose a `.focus()` instance method or forwards it's ref to a component that implements a .focus() method, the studio will crash with an error saying "Cannot read property 'focus' of undefined". This PR adds a safeguard against it and also issues a console warning that should make it easier to pinpoint the issue.

### What to review
Review and see if the code/message makes sense

### Notes for release
- Adds safeguard against missing `.focus()` method on components used with the `withDocument()` higher order component.

### DIFF
--- a/packages/@sanity/form-builder/src/utils/withDocument.tsx
+++ b/packages/@sanity/form-builder/src/utils/withDocument.tsx
@@ -1,8 +1,23 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+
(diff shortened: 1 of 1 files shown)


mariuslundgard (approved):
Nice work 👍

PR #2638, opened 2021-07-29
fix(base): fixes a bug where isMissing was triggered when the reference exists

sgulseth:
### Description

The reference was marked as missing when the document existed because of a missing conditional clause.

For cross reference: https://github.com/sanity-io/sanity/pull/2585/files#diff-0419dde3aa4b7ffa09d19c960d248addc8fe46dd8955042a28018f8df203d444L114


bjoerge (approved):
Great catch, LGTM!

ricokahler:
oops 😅 

great catch indeed!

PR #2639, opened 2021-07-29
feat(form-builder) support a callback as `hidden` option on field type

bjoerge:
### Description

This adds support for specifying a callback on the `hidden` option for field types. The callback will be passed a context object that includes the current field value, the parent (enclosing object) value and the document value.

This "solves" conditional fields by allowing to conditionally hide/show fields depending on conditions in the current document/object/field

#### Limitations
It's currently not possible to do the hidden check async. This is something we can add support for in the future if there are use cases that justifies it. Since the hidden checks are currently done very frequently, async support will have to consider being run debounced, and/or with affordances to abort any pending promise from the previous check.

### What to review
Code + schema API for conditionally hiding fields

### Notes for release

- Adds support for specifying a hidden callback to conditionally hide fields on object types


packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const ConditionalField = (props: Props) => {
+  if (!props.hidden || props.hidden === true) {
+    return <>{props.hidden === true ? null : props.children}</>
+  }
+  return <ConditionalFieldWithDocument {...props} />
+}
+
+const ConditionalFieldWithDocument = withDocument(
+  rxComponent(function ConditionalFieldWithDocument(props$: Observable<Props & {document: any}>) {
```
mariuslundgard, line 45: Wondering is this could be solved using `useMemo` somehow. Would probably be simpler to read and maintain for more people.

packages/@sanity/form-builder/src/utils/withDocument.tsx
```diff
@@ -14,19 +14,25 @@ function warnMissingFocusMethod(ComposedComponent) {
   )
 }
 
-export default function withDocument(ComposedComponent: any) {
-  return class WithDocument extends React.PureComponent {
+interface WithDocumentProps<Doc extends Record<string, unknown> = Record<string, unknown>> {
```
mariuslundgard, line 17: Can this be exported so that it can be used outside of this package?

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
mariuslundgard: Could `{document: any}` here instead be the interface imported from `withDocument`?

```tsx
import {WithDocumentProps} from '../../utils/withDocument'

// ...

const ConditionalFieldWithDocument = withDocument(
  rxComponent(function ConditionalFieldWithDocument(props$: Observable<Props & WithDocumentProps>) {
```

packages/@sanity/form-builder/src/utils/withDocument.tsx
mariuslundgard: Also, would be good if this used `SanityDocument` from @sanity/types

```
interface WithDocumentProps<Doc extends SanityDocument = SanityDocument> {
```

packages/@sanity/form-builder/src/utils/withDocument.tsx
```diff
@@ -14,19 +14,25 @@ function warnMissingFocusMethod(ComposedComponent) {
...
+  return class WithDocument extends React.PureComponent<Omit<T, 'document'>> {
     _input: any
     _didShowFocusWarning = false
     static displayName = `withDocument(${ComposedComponent.displayName || ComposedComponent.name})`
     static contextTypes = {
       formBuilder: PropTypes.any,
     }
     state: {
-      document: Record<string, any>
+      document: Record<string, unknown>
```
mariuslundgard, line 32: Use `SanityDocument` here?

packages/@sanity/types/src/schema/types.ts
```diff
@@ -3,6 +3,8 @@ import React from 'react'
 import {ReferenceOptions} from '../reference'
 import {AssetSource} from '../assets'
 import {SlugOptions} from '../slug'
+import {Observable} from 'rxjs'
+import {SanityDocument} from '../documents'
```
mariuslundgard, line 7: Import from `@sanity/types` instead?

packages/@sanity/types/src/schema/types.ts
```diff
@@ -31,6 +33,15 @@ export type SortOrdering = {
   }
 }
 
+export interface HiddenPredicateContext {
```
mariuslundgard, line 36: Wondering if `HiddenCallbackContext` is more idiomatic to JS 🤔 See e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

packages/@sanity/types/src/schema/types.ts
```diff
@@ -31,6 +33,15 @@ export type SortOrdering = {
...
+  parent?: Record<string, unknown>
```
ricokahler, line 37: I think parent can be an array as well

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
```diff
@@ -99,13 +99,14 @@ export const ObjectInput = memo(
...
-        if (!filterField(type, field) || field.type.hidden) {
+        const fieldValue = value?.[field.name]
+        if (!filterField(type, field)) {
           return null
         }
 
-        const fieldValue = value && value[field.name]
         return (
           <ObjectInputField
+            parent={value}
```
ricokahler, line 109: I think it would be better if you wrapped the parent value in a ref and synced the ref via `useLayoutEffect`. this will create a stable value that won't cause re-renders and remove the need for the useMemo introduced in `ObjectInputField`.

it should be fine to have parent in a ref since the condition callback function has the ability to pull values

```js
function ObjectInput({value, /* ... */}) {
  const parentRef = useRef(value)

  // sync the parentRef
  useLayoutEffect(() => {
    parentRef.current = value
  }, [value])

  // ...
      <ObjectInputField parentRef={parentRef} {/* ... */} />
  // ...
}
```

ricokahler: oooo new idea!: could we utilize the path to grab the parent value? no need to pass the parent at all then

```js
import {get} from 'lodash'

// in the child
const parent = get(document, path.slice(0, path.length - 1))
```

packages/@sanity/types/src/schema/types.ts
ricokahler: I think we should include `path` here as well since the context in Rule.custom also includes path

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
ricokahler: > Wondering is this could be solved using useMemo somehow. Would probably be simpler to read and maintain for more people.

I agree. unless i'm mistaken, i think it would look like this

```js
const ConditionalFieldWithDocument = withDocument(({hidden, document, parent, value, children}) => {
  const shouldHide = useMemo(() => {
    return hidden({document, parent, value, path})
  }, [hidden, document, parent, value, path])

  return shouldHide ? null : children
})

ConditionalFieldWithDocument.displayName = 'ConditionalFieldWithDocument'
```

ricokahler (commented):
i think we should include the `path` in the context and then grab the `parent` utilizing that. i think that could potentially remove some churn and simplify things a bit. also check out my suggestion with the `ConditionalFieldWithDocument` component. is it to naive? lol

packages/@sanity/types/src/schema/types.ts
```diff
@@ -3,6 +3,8 @@ import React from 'react'
 import {ReferenceOptions} from '../reference'
 import {AssetSource} from '../assets'
 import {SlugOptions} from '../slug'
+import {Observable} from 'rxjs'
```
mariuslundgard, line 6: `Observable` is not in use.

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
mariuslundgard: Yeah, that’s much simpler IMO @ricokahler 👍 

Later, the `useMemo` can be refactored to a hook and support promises and observables, something like:
```tsx
function getHidden$(hidden, context) {
  return of(context).pipe(
    switchMap(c => {
      const val = hidden(c)

      if (!val || val === tue) return of(val)

      return from(val)
    })
  )
}

function useHiddenAsync(hidden, context) {
  const hidden$ = useMemo(() => getHidden$(hidden, context), [hidden, context])

  return useObservable(hidden$, false)
}

const ConditionalFieldWithDocument = withDocument(({hidden, document, parent, value, children}) => {
  const context = useMemo(
    () => ({document, parent, value, children}),
    [document, parent, value, children]
  )

  const shouldHide = useHiddenAsync(hidden, context)

  return shouldHide ? null : children
})

ConditionalFieldWithDocument.displayName = 'ConditionalFieldWithDocument'
```

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
bjoerge: I like the path idea, but getting the parent this way is arguably less convenient than having it available on the `parent` context key? I think also we _do_ want to re-render each field's hidden check when the parent changes, and making it a ref here would potentially prevent fields from re-checking when the parent value changes, and therefore not update correctly? In practice, they will all update anyway since the document changes all of the time. As it is now, the re-rendering the ConditionalField component is cheap, so not sure if it's worth the added complexity.

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
```diff
@@ -0,0 +1,72 @@
...
+      map((props) => props.children),
+      distinctUntilChanged()
+    )
+
+    // A stream of arguments passed to the hidden callback
+    const contextArg$ = props$.pipe(map(({document, parent, value}) => ({document, parent, value})))
+
+    const isHidden$ = hiddenProp$.pipe(
+      switchMap((hiddenProp) =>
+        !hiddenProp || typeof hiddenProp === 'boolean'
```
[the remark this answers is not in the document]
bjoerge: Great point!

packages/@sanity/types/src/schema/types.ts
bjoerge: Currently only objects are supported. Conditionally hiding array items calls for a different set of considerations and is out of scope for now.

packages/@sanity/types/src/schema/types.ts
bjoerge: >Wondering if HiddenCallbackContext would be a more idiomatic naming

Yeah, lets follow the MDN terminology here. Thanks!

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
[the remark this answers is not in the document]
bjoerge: Is there any benefit in allowing "truthy"? The idea is to enforce it to be a boolean to prevent objects from being treated as `truthy` now, so that we can support a config object in the future without breaking existing code. It's an edge case perhaps, but if you have `hidden: {}` with today's behavior it will still show the field, and if we introduced a config object here later on, we'd have to keep supporting this as truthy to avoid it being a breaking change, so I think we make things easier to reason about if we require it to be a boolean and not truthy

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
bjoerge: > Later, the useMemo can be refactored to a hook and support promises and observables, something like:

Agree in general, although this isn't quite what we want from the observable case though. If you provide a reactive/observable callback you should be able to build on a stream of context args by providing a callback that is called once initially with the context stream:

```ts
{
  //...
  hidden: {
    stream: (context$) =>
      context$.pipe(
        debounceTime(200),
        switchMap((context) => fetch(context.parent.productId)),
        map((product) => product.expired)
      )
  }
}
```

bjoerge: I've refactored to functional component and it became a lot simpler, thanks for pointing out!

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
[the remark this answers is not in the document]
ricokahler: ah that makes sense. yeah let's enforce `true`

packages/@sanity/types/src/schema/types.ts
ricokahler: oh yeah you're right. i don't really think it makes sense to hide an item of an array ever 😄

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
```diff
@@ -0,0 +1,61 @@
...
+  schemaHiddenOption: HiddenOptionCallback | undefined | boolean,
+  context: HiddenOptionCallbackContext
+): boolean {
+  const result =
+    typeof schemaHiddenOption === 'function' ? schemaHiddenOption(context) : schemaHiddenOption
+
+  if (isThenable(result)) {
+    console.warn(
+      '[Warning]: The hidden option is either a promise or a promise returning function. Async callbacks for `hidden` option is not currently supported.'
+    )
```
ricokahler, line 27: we could store some state to prevent this warning from over logging

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
```diff
@@ -0,0 +1,61 @@
...
+function isThenable(value: any) {
+  return typeof value?.then === 'function'
+}
+
+function checkCondition(
+  schemaHiddenOption: HiddenOptionCallback | undefined | boolean,
+  context: HiddenOptionCallbackContext
+): boolean {
+  const result =
+    typeof schemaHiddenOption === 'function' ? schemaHiddenOption(context) : schemaHiddenOption
```
ricokahler, line 22: it might be a good idea to try catch this result and warn. a common case for this would be users not handling the undefined case on the initial undefined document and in that case i think we can save the editor from crashing without many unwanted effects

ricokahler (approved):
it's looking really good. i had some other small things i caught but i think the only thing left is double checking naming

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
bjoerge: Great point! Fixed in b2b4797 (I didn't add any state for preventing overlogging for errors thrown in the callback  as the error may differ from each call, but we can consider improving on this later if needed).

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
bjoerge: Fixed in b2b4797

packages/@sanity/form-builder/src/inputs/ObjectInput/ConditionalField.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {
+  BaseSchemaType,
+  HiddenOptionCallbackContext,
+  HiddenOptionCallback,
+  SanityDocument,
+} from '@sanity/types'
+
+import withDocument from '../../utils/withDocument'
+
+type HiddenOption = BaseSchemaType['hidden']
```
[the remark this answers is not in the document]
bjoerge: Fixed, thanks!

packages/@sanity/form-builder/src/utils/withDocument.tsx
mariuslundgard: Looks like this was missed? (use `Doc` now?)

PR #2642, opened 2021-07-30
Remove duplicate parts definition

judofyr:
Fixes #2558

### Description

I identified the duplicate definitions with the following command:

    cat packages/@sanity/base/sanity.json | groq '(parts[].implements)[defined(@)]' -o ndjson | sort | uniq -c | sort -n

### What to review

* Double check that the removed lines actually are present another place in the file.


rexxars (approved):
Thanks!

PR #2648, opened 2021-08-04
Export types correctly in built @sanity/base package

### LINKED ISSUE #2519: @sanity/base presence type is incorrect
Greetings!

It looks like because your types are not generated at compile time, the mapping of the presence type does not actually work when `@sanity/base` is packaged because the packaged code is bundled into a `lib` directory and not a `src` directory.

https://github.com/sanity-io/sanity/blob/dd995f06d8f5086b4b1c0b76c8991a55f3fa297d/packages/%40sanity/base/presence.d.ts#L1

The solution would be to either have typescript generate these types on the fly or provide a `paths` mapping that uses lib as a fallback.

judofyr:
### Description

These files are compiled by anything during packaging which means that they end up referring to `src`-directories which is not included in the NPM package.

This changes the imports to refer to `lib` which should work in the
built NPM package.

Fixes #2519.

### What to review

- Check that we didn't break "Go to definition" in an IDE.

### Notes for release

-  `@sanity/base` now exports TypeScript definitions for the `presence`, `user-color`, `components` and `hooks` files.

### DIFF
--- a/packages/@sanity/base/components.d.ts
+++ b/packages/@sanity/base/components.d.ts
@@ -1,3 +1,3 @@
-import * as components from './src/components'
+import * as components from './lib/components'
 
 export = components
--- a/packages/@sanity/base/hooks.d.ts
+++ b/packages/@sanity/base/hooks.d.ts
@@ -1,3 +1,3 @@
-import * as hooks from './src/hooks'
+import * as hooks from './lib/hooks'
 
 export = hooks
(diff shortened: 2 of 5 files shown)


bjoerge:
This will be fixed properly by #2662

(the reason we exported `src` here was to have IDE jump to source work during development, but that's now taken care of by a proper [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) setup)

PR #2655, opened 2021-08-06
feat(client): allow api version to be overridden for requests

dcilke:
### Description

When calling `client.request()` allow the api version to be overridden. This is to support only using the experimental api for certain requests.

### What to review

* The logic for generating the base url from the config. This was broken out to allow the same logic to be used when overriding the api version.

### Notes for release

This patch allows the api version to be set on a per request basis.


skogsmaskin (approved):
Looks good to me!

PR #2659, opened 2021-08-09
fix(default-login): remove temporary solution to hide close button in dialogs

hermanwikner:
### Description

This PR removes the temporary solution to hide the close button in dialogs.


bjoerge (approved):
LGTM

mariuslundgard (approved):

PR #2661, opened 2021-08-10
feat(client): support `proxy` in config + options

### LINKED ISSUE #2652: Need sanity client config for corporate proxy
I need to set proxy at environment level to make sanity client work. I would like it to be part of sanity client config so its applicable only for all http calls originating from sanity client

a new config - proxy for sanoity client config

Updated the packages/@sanity/client/src/http/requestOptions.js directly in node_module
Included proxy in sanity client config ; Added a new variable in requestOptions,js to get config.proxy.Included the proxy variable in returned  object

**Additional context**

judofyr:
### Description

This adds support for configuring a single client and/or request to use a proxy.

Fixes #2652.

### What to review

I was not able to write a test for this since it doesn't seem like Nock supports retrieving the proxy of a request:

```js
test('can be configured with proxy', (t) => {
  // TODO: How to check that example.com is being used?
  nock(projectHost()).get('/v1/ping').reply(200, {pong: true})

  getClient({proxy: 'http://example.com'})
    .request({uri: '/ping'})
    .then((res) => t.equal(res.pong, true))
    .catch(t.ifError)
    .then(t.end)
})
```

Would like feedback on suggestions to verify that this works as expected.

### Notes for release

* Support `proxy`-option in `@sanity/client`.

### DIFF
--- a/packages/@sanity/client/sanityClient.d.ts
+++ b/packages/@sanity/client/sanityClient.d.ts
@@ -42,6 +42,7 @@ interface RawRequestOptions {
   query?: {[key: string]: string | string[]}
   headers?: {[key: string]: string}
   timeout?: number
+  proxy?: string
   body?: any
 }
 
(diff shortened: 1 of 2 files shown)


rexxars (approved):
LGTM! 👍

Issue #2591, opened 2021-06-30
Unable to install Sanity dependencies on a Windows machine


alexbchr:
**Describe the bug**

It seems impossible to develop with Sanity using a Windows machine. I already reached out to Sanity on Slack, but I feel this issue deserves more visibility as I had 3 different devs with the issue on different Windows machines (all our other devs use Mac or Linux and don't have any issues). So basically when installing dependencies using Yarn on a Windows machine (running `yarn install` on an existing project of our company), we get the following error:

```
Error: While persisting /D:/Projects/projectName/.yarn/cache/@sanity-types-npm-2.6.0-96b9f7d8b0-77d01c3bcd.zip/node_modules/@sanity/types/ -> /D:/Projects/projectName/node_modules/@sanity/portable-text-editor/node_modules/@sanity/types ENOENT: no such file or directory, mkdir 'D:\Projects\projectName\node_modules\@sanity\portable-text-editor\node_modules\@sanity\types\parts\part:@sanity'
```

I have validated that the issue is the `:` character in the D:\Projects\projectName\node_modules\@sanity\portable-text-editor\node_modules\@sanity\types\parts\part:@sanity file path. Windows does not allow colons in filenames.

**To Reproduce**

1. Create a Sanity Studio project on Windows (we use Typescript, maybe this has an impact as well)
2. Install dependencies using `yarn install`
3. It crashes 💥 

**Expected behavior**

We should be able to install sanity dependencies on Windows without any workaround like enabling Windows Subsystem for Linux (WSL).

**Which versions of Sanity are you using?**

Run `sanity versions` in the terminal and copy-paste the result here.

```
@sanity/cli             2.10.5 (latest: 2.12.2)
@sanity/base            2.10.5 (latest: 2.12.2)
@sanity/cli             2.10.5 (latest: 2.12.2)
@sanity/components       2.2.6 (up to date)
@sanity/core            2.10.5 (latest: 2.12.2)
@sanity/default-layout  2.10.5 (latest: 2.12.2)
@sanity/default-login   2.10.5 (latest: 2.11.1)
@sanity/desk-tool       2.10.5 (latest: 2.12.2)
@sanity/icons            1.1.2 (up to date)
@sanity/vision          2.10.5 (latest: 2.11.2)
```

**What operating system are you using?**
Windows

**Which versions of Node.js / npm are you running?**
Node 12, Yarn 2.4.2

Run `yarn -v && node -v` in the terminal and copy-paste the result here.
```
[...]

Issue #2188, opened 2020-12-11
Unable to clone repo


andre-brdoch:
**Describe the bug**

Attempting to clone the repo leads to the following error:
```
git clone https://github.com/sanity-io/sanity.git
Cloning into 'sanity'...
remote: Enumerating objects: 474, done.
remote: Counting objects: 100% (474/474), done.
remote: Compressing objects: 100% (269/269), done.
remote: Total 159850 (delta 338), reused 295 (delta 202), pack-reused 159376
Receiving objects: 100% (159850/159850), 33.77 MiB | 5.91 MiB/s, done.
Resolving deltas: 100% (121761/121761), done.
error: invalid path 'packages/@sanity/types/parts/all:part:@sanity/index.d.ts'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'

and retry with 'git restore --source=HEAD :/'
```

**What operating system are you using?**
win 10

bjoerge:
This is a current known issue on Windows I'm afraid.

You should be able to work around it for now by doing a "sparse" checkout.

- Clone with `git clone https://github.com/sanity-io/sanity.git --no-checkout`
- CD into the local clone (e.g. `cd sanity`)
- Run `git config core.sparsecheckout true`
- Open `.git\info\sparse-checkout` in a text editor and add the following
  ```
  !packages/@sanity/types/parts
  /*
  ```
- Run `git checkout --`

andre-brdoch:
Hi @bjoerge, thanks for helping out! 😄 I followed the steps, however no `.git\info\sparse-checkout` is being created. I tried creating it manually and adding the contents, but it was still giving errors.

Do you know what is the issue? My git knowledge is unfortunately not yet there to see what is going on!

aweber1:
I also am running into this same annoying issue and wasn't able to get sparse checkout working on Windows either. FWIW, I found a couple references with suggestions that _seemed_ like they would work, but ultimately didn't for me:
https://stackoverflow.com/a/63742773/9324
https://github.com/git-for-windows/git/issues/2777

I ended up just downloading the repo as a .zip and extracting it. So much for being able to contribute anything though ;)

PR #2662, opened 2021-08-10
Various package/typescript setup improvements

bjoerge:
### Description
Sorry for a huge PR, but since these commits are more or less tied together it makes sense to merge them together.

This includes several changes and improvements to our TypeScript configuration/dev environment. Here's a summary:

* TypeScript definitions for all packages are now compiled to `./dist/ts` (commonjs is still compiled to `./lib`)
* TypeScript project references now *work properly* with the language service without having to compile anything up front. This means:
    * we can now clone the repo, run `npm install`, open the editor and start coding with no compile errors in the IDE and have jump to source etc. work properly across package boundaries without having to build any type definitions first. If a type changes in one package it's instantly reflected across the consuming monorepo packages.
    * we get the exact same type errors during development in the IDE and when compiling type definition files build / CI time
* Fixes an issue that would occasionally cause a compile error in monorepo after doing changes to `@sanity/types` APIs (the compile error would usually go away after releasing a new version of @sanity/types, but cause CI builds failures for branches, etc. in the meantime)
* Removed the path aliases in root tsconfig, and replaced them with a dedicated file (`.webpack-aliases.json5` ) with the paths webpack needs to map package imports to `src`-dirs
* Replaced internal use of part typings through tsconfig/`typeRoots` in tsconfigs with `///<reference types="@sanity/types/parts" />` directives in ts files that still have part imports. (also added a codemod to @sanity/cli that insert this directive to typescript files with part imports). This made it possible to fix an issue people were having that made it impossible to clone the repo on a windows platform (#2591, #2188).
* removed the `typescript-plugin-postcss` from all packages since it gave us very little value, typescript has gotten better native support for css imports by now, and we're moving away from it anyways

Edit: we decided to remove esm compilation for now since it's not likely to work with part imports anyways. Let's reconsider these ones for later:
* ~Added compilation of [ESM](https://2ality.com/2021/06/typescript-esm-nodejs.html) versions of all packages to `./dist/esm` (still compiling commonjs to `./lib`). This enables tree-shaking and paves the way better integration with with modern bundlers like Vite, Snowpack, Parcel, etc.~
* ~Added `exports` fields to packages for consumers that supports it (those who don't still have the package root js files that reexports from `./lib`)~ 

This simplifies our current setup a lot and hopefully it provides us with a more predictable, future proof and flexible dev environment going forward.

### How to review
- This branch is released under the tag `ts-improvments` and can be installed using `sanity upgrade --tag=ts-improvements --save-exact`

### Notes for release

- Compile ESM targets for @sanity/packages
- Add codemod for adding reference types directives to TypeScript files with part imports


bjoerge:
I'm going to leave the lint check erroring, fyi, as it's too much work to get all the changed files to pass the linter. I touched a bunch of files by adding a header comment and not too keen on fixing existing lint errors in all of them (the check will be green again once merged, fwiw).

PR #2663, opened 2021-08-10
Add defaultLocale to @sanity/google-maps-input config

PetterRuud:
### Description
What changes are introduced?
ADD new config variable than can be used to set locale
Why are these changes introduced?
ADD possibilty to set your own locale in the Google Maps input plugin
What issue(s) does this solve? (with link, if possible)
None that i found.

### What to review
What steps should the reviewer take in order to review?
Just see if the variable names are ok
What parts/flows of the application/packages/tooling is affected?
@sanity/google-maps-input

### Notes for release

You can now add a default locale for Google Maps, it will make the search prioritize places/addresses in that locale.

### DIFF
--- a/packages/@sanity/google-maps-input/config.dist.json
+++ b/packages/@sanity/google-maps-input/config.dist.json
@@ -1,6 +1,7 @@
 {
   "apiKey": null,
   "defaultZoom": 11,
+  "defaultLocale": "en-US",
   "defaultLocation": {
     "lat": 40.7058254,
     "lng": -74.1180863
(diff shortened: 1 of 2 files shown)


rexxars (changes requested):
Feature wise this looks good to me, but including it in the dist config means it will always default to `en-US` instead of `window.navigator.language` which is the current default.

Maybe we could add a note in the readme about it, or leave it as `null` in the dist config so it is discoverable but not set by default? What do you think?

PetterRuud:
> 
> 
> Feature wise this looks good to me, but including it in the dist config means it will always default to `en-US` instead of `window.navigator.language` which is the current default.
> 
> Maybe we could add a note in the readme about it, or leave it as `null` in the dist config so it is discoverable but not set by default? What do you think?

Done, it is now null in config.dist

rexxars (approved):
Thanks!

PR #2664, opened 2021-08-10
refactor(color-input): migrate to `@sanity/ui`

hermanwikner:
### Description

Migrate `color-input` to `@sanity/ui`

### What to review

Make sure that the plugin works as before. There are some minor visual differences after migrating to `@sanity/ui`. 

Screenshot 2021-08-10 at 16 06 17

### Notes for release

Migrate `color-input` to `@sanity/ui`


packages/@sanity/color-input/src/ColorPicker.js
```diff
@@ -1,10 +1,25 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+`
+
+const ReadOnlyContainer = styled(Flex)`
+  margin-top: 6rem;
+  background-color: ${({theme}) => theme.sanity.color.base.bg};
```
mariuslundgard, line 19: I believe you can use `background-color: var(--card-bg-color)` to achieve the same thing?

hermanwikner: Yes, that works as well! Is it preferable to use css variables instead of accessing the theme as I did?

mariuslundgard:
This looks really good to me!

I know it’s not strictly in scope, but think it would be a quick win to modify the UI slightly to achieve this:
image

This would fix the current issues with the change connector UI. Let me know what you think @hermanwikner!

hermanwikner:
> This looks really good to me!
> 
> I know it’s not strictly in scope, but think it would be a quick win to modify the UI slightly to achieve this:
> image
> 
> This would fix the current issues with the change connector UI. Let me know what you think @hermanwikner!

Thanks! I agree, I'll update the PR with that fix 👍

packages/@sanity/color-input/src/ColorInput.js
```diff
@@ -1,11 +1,12 @@
 /* eslint-disable id-length */
 import React, {PureComponent} from 'react'
-import Button from 'part:@sanity/components/buttons/default'
 import PropTypes from 'prop-types'
 import {PatchEvent, patches} from 'part:@sanity/form-builder'
 import Fieldset from 'part:@sanity/components/fieldsets/default'
```
bjoerge, line 5: Should these be replaced with part-less imports? (cc @mariuslundgard)

packages/@sanity/color-input/src/ColorInput.js
```diff
@@ -84,7 +85,7 @@ export default class ColorInput extends PureComponent {
   render() {
     const {type, readOnly, value} = this.props
     return (
-      <Fieldset legend={type.title} description={type.description}>
+      <Fieldset legend={type.title} description={type.description} level={0}>
```
bjoerge, line 88: I think we should rather forward `props.level` here instead of hard coding it to `0`

bjoerge (commented):
Left a few comments/questions, and would be great to get rid of all part imports if possible :) 

Otherwise LGTM 👏🏼

packages/@sanity/color-input/src/ColorPicker.js
RitaDias: I had the same question on my pr as well! ☝️

packages/@sanity/color-input/src/ColorInput.js
hermanwikner: Thanks, fixing! 👍

ricokahler (approved):
looks great!

I had just one note on dependencies

mariuslundgard (approved):

PR #2675, opened 2021-08-13
refactor(base): move icons to __legacy directory

mariuslundgard:
### Description

- The icons in `@sanity/base` have been moved to `packages/@sanity/base/src/__legacy/components/icons` as they are superseded by `@sanity/icons` now.
- The `sanity.json` file has been updated so e.g. `part:@sanity/base/file-icon` points to `__legacy/components/icons/File` instead of `components/icons/File`.

### What to review

- That the part-imports of icons still work as expected.

### Notes for release

- n/a


bjoerge (approved):
LGTM!

rexxars:
I'm worried that people have imported from `lib/components/icons` - I seem to already recall seeing this in some third party code. Should we add deprecation proxies, or leave them as-is?

rexxars (changes requested):
It would be nice to have a page listing all the available icons in @sanity/icons so you know how to migrate, and then add deprecation proxies telling the user to migrate and return the icon from the new location.

mariuslundgard:
@rexxars If we need to leave deprecation proxies, then this change is not needed. The whole point is to make it easier to see what is legacy / soon to be deprecated, and what is being actively maintained. Let’s just close this PR?

rexxars:
Agreed!

PR #1563, opened 2019-10-23
code-input Adding new languages

ajonp:
**Type of change (check at least one)**

- [ ]  Bug fix (non-breaking change which fixes an issue)
- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  Documentation (fix or update to documentation)
- [ ]  Maintenance
- [ ]  Other, please describe:

**Does this change require a documentation update? (Check one)**

- [ ]  Yes
- [x]  No

**Current behavior**
[code-input](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/code-input) Adding new languages
Missing often used languages for webdev in the `code-input` plugin.

**Description**

Added these additional languages to the `code-input` plugin. Tested locally by building the repo and then moving the code-input package to my sanity studio locally. Also ran `npm run test-all`. In addition I alphabetized the list to make it easier in the future when languages are needed.

**Note for release**

Ace offers several languages https://ace.c9.io/. I use a lot more in my every day code samples and thought some of the more common ones should be added. 

**Checklist** 

- [x]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [x]  The PR title includes a link to the relevant issue
- [x]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [x]  The code is linted
- [x]  The test suite is passing
`lerna success run Ran npm script 'test' in 25 packages in 60.9s:`
- [x]  Corresponding changes to the documentation have been made

PR #2676, opened 2021-08-13
feat(code-input): add java, go, xml, c#, tsx, ruby, yaml support to code input

skogsmaskin:
This is a follow up on this PR: https://github.com/sanity-io/sanity/pull/1563

Typescript support was already added in the new refactored Code Editor (typescript / sanity-ui).

Added support for java, go, xml, c#, tsx, ruby, yaml.

Also refactored the code so that we don't have to keep two separate language setups for the previews and editor.

Not sure about the increased bundle size from these changes. Maybe we should check that before merging.


packages/@sanity/code-input/src/config.ts
```diff
@@ -2,20 +2,27 @@ import {CodeInputLanguage} from './types'
 
 export const SUPPORTED_LANGUAGES: CodeInputLanguage[] = [
   {title: 'Batch file', value: 'batchfile'},
+  {title: 'C#', value: 'csharp'},
   {title: 'CSS', value: 'css'},
-  {title: 'SASS', value: 'sass'},
-  {title: 'SCSS', value: 'scss'},
+  {title: 'GoLang', value: 'golang'},
```
rexxars, line 7:
```suggestion
  {title: 'Go', value: 'golang'},
```
I think Go is usually just referred to as "Go" unless you need to differentiate between the language and the.. game? word?

rexxars:
This PR adds about 126 kB pre-minifying. Thoughts?

rexxars (changes requested):

packages/@sanity/code-input/src/config.ts
skogsmaskin: Fixed.

skogsmaskin:
I added Yarn also by popular demand.

skogsmaskin:
> This PR adds about 126 kB pre-minifying. Thoughts?

I think that's ok? I much rather have that than loading it from somewhere async.

packages/@sanity/code-input/src/editorSupport.ts
```diff
@@ -0,0 +1,32 @@
...
+import 'brace/mode/python'
+import 'brace/mode/ruby'
+import 'brace/mode/sass'
+import 'brace/mode/scss'
+import 'brace/mode/sh'
+import 'brace/mode/text'
+import 'brace/mode/tsx'
+import 'brace/mode/typescript'
+import 'brace/mode/xml'
+import 'brace/mode/yaml'
```
llibdude, line 25: I had requested `yaml`  [here](https://github.com/sanity-io/sanity/issues/2682)  so thanks for including it!  I noticed `yaml` isnt referenced in the `config.ts` file above. I don't know much about the codebase but I assume it should be?

packages/@sanity/code-input/src/config.ts
```diff
@@ -1,21 +1,30 @@
...
+  {title: 'Ruby', value: 'ruby'},
+  {title: 'SASS', value: 'sass'},
+  {title: 'SCSS', value: 'scss'},
   {title: 'sh', value: 'sh'},
-  {title: 'Plain text', value: 'text'},
-  {title: 'GROQ', value: 'groq'},
+  {title: 'TSX', value: 'tsx'},
+  {title: 'TypeScript', value: 'typescript'},
+  {title: 'XML', value: 'xml'},
+  {title: 'Yarn', value: 'yarn'},
```
llibdude, line 27: oh I see, should this be `yaml` not `yarn`?

skogsmaskin: Most def should!

packages/@sanity/code-input/src/editorSupport.ts
skogsmaskin: Yep! As you pointed out and now fixed. Thank you.

rexxars (approved):
Given we don't have time to refactor this with upgraded dependencies and make dynamic loading of languages work, I think this looks good.

PR #2689, opened 2021-08-19
fix(desk-tool): use `editState` as source for publish status

mariuslundgard:
### Description

- With this change, the document sparkline uses the `editState` to find out if a document is published or not (instead of looking at the history events and finding the last `publish` or `unpublish` event).

### What to review

- Test both live document and regular published/draft documents, and make sure the publish state is accurate in the document status bar.

### Notes for release

n/a

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/documentPane/statusBar/sparkline/documentSparkline.tsx
+++ b/packages/@sanity/desk-tool/src/panes/documentPane/statusBar/sparkline/documentSparkline.tsx
@@ -59,6 +59,10 @@ export function DocumentSparkline(props: DocumentSparklineProps) {
   const [transition, setTransition] = useState(false)
   useEffect(() => {
     if (!transition && loaded) {
+      // NOTE: the reason for double RAF here is a common "bug" in browsers.
+      // See: https://stackoverflow.com/questions/44145740/how-does-double-requestanimationframe-work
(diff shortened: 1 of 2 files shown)


packages/@sanity/desk-tool/src/panes/documentPane/statusBar/sparkline/documentSparkline.tsx
```diff
@@ -75,6 +55,14 @@ export function DocumentSparkline(props: DocumentSparklineProps) {
     [reviewChangesButtonWidth]
   )
 
+  // Only transition between subsequent state, not the initial
+  const [transition, setTransition] = useState(false)
+  useEffect(() => {
+    if (!transition && loaded) {
+      requestAnimationFrame(() => requestAnimationFrame(() => setTransition(true)))
+    }
```
bjoerge, line 63: a bit unclear to me why we have two nested rafs here - is it intentional? if so, why is it needed? A short comment to clarify would be good.

also, should it be a return callback with cancelAnimationFrame here? If not, add a comment to explain why it's not needed?

bjoerge (commented):
Have a couple of questions regarding the transition effect/callbacks that would be good to have answered by code comments - otherwise LGTM!

bjoerge (approved):

PR #2693, opened 2021-08-20
Refactor ActivateOnFocus

RitaDias:
### Description

While discussing the component ActivateOnFocus it was decided that instead of using the component that exists today to move it and create a new one that has as a goal to replace it in other places. 

This new component has been moved to the components folder and does not live in the same space as the original one.

https://user-images.githubusercontent.com/6951139/130198889-4694a463-33f2-4a0d-a18c-d89500953fd7.mov

https://user-images.githubusercontent.com/6951139/130198902-71fabf4a-b06a-4246-8da8-5e5ca4183197.mov

The behaviour and look is the same, however it now uses `@sanity/ui` components and has been simplified.
The only place that uses the new version is the `input` component and the goal is to eventually replace all of the current uses of the old ActivateOnFocus. We can replace it in more places but I wanted to keep the scope for this pr small and separate from more larger changes across the product :) 

I renamed the components props to be more consistent to what makes sense in the scope of the component `ActivateOnFocus`
- `IsActive` -> `IsOverlayActive`, with this change it makes it that we don't use negatives in the code
- `html` -> `message` , just found the name to be clearer (though not really explicit immediately about what the type is expected)

These will make it that when we replace with this new component we'll have to make some more manual changes when changing the names, but I thought them worth it because they feel clearer. 

I've also removed a few props `inputId` and `overlayClassName` which I didn't see a use for.

### What to review

The behaviour and quirks remain the same as well as the old one:
- One the overlay is disabled you cannot get the overlay to return. The way to get it working is to `isOverlayActive` from `false` to `true`

Make sure that you can tab into the content 👇 

https://user-images.githubusercontent.com/6951139/130198698-3d9dfee3-4c03-4a40-abe6-e57761abd14e.mov

The only section that should be affected is wherever the input component is being used, you can see it in action in the `blocks test`


skogsmaskin (approved):
Phew, that's a lot simpler!

Thanks @RitaDias !

PR #2708, opened 2021-08-26
refactor(form-builder): remove old ActivateOnFocus usage

RitaDias:
### Description

Removed mentions of the old `ActivateOnFocus` and removed it from components that we are planning on changing in the future.

- Removed old and dead files no longer used

### What to review

It would be good to go through the examples where they were before and see if the ux has improved / remained the same given the current goals.

- Protein input no longer has the ActivateOnFocus the example-studio
- Unknown input no longer has ActivateOnFocus due to update in use case (even if there is a scroll content, the activate on focus will never actually be useful)


bjoerge (approved):
LGTM!

PR #2717, opened 2021-08-27
chore: upgrade lerna

ricokahler:
### Description

The following upgrades lerna. 

### What to review

Try out all the lerna commands we use to make sure they still work. It looks like lerna 4 was a mjor primarily because of breaking support for older node versions.

See here: https://github.com/lerna/lerna/releases/tag/v4.0.0


skogsmaskin:
Sorry I forgot to approve this.

> Nothing was newly deprecated, nothing previously deprecated was removed. Migrating to v4 should be straightforward.

PR #2718, opened 2021-08-28
fix(validation): run nested validation in nested fields if required

### LINKED ISSUE #1713: Object fields don't correctly run validation until dirtied
**Describe the bug**

Nested object fields (of any depth) don't appear to run validation until they've been dirtied – or edited.  

This means that it's possible to publish documents with invalid object fields, unless an editor has entered values into those fields first.

**To Reproduce**

1. Have a document schema with an object with nested fields requiring validation
2. Create a new document and enter a value into any other field

**Expected behavior**

All fields should run validation rules on any field change, regardless of whether they're in an encapsulating `object` or not.

**Screen recording**

In the below recording, `age` and `meta.description` both have custom rules which spits out a `console.log` message. 

Note that:
- only `age`'s validation rule initially fires on document change, but once `meta.description` is dirtied, it then too fires validation as expected.
- it's possible to publish despite `meta.description` not being correctly validated

The same issue occur with non-custom validation rules too.

Kapture 2020-01-25 at 1 00 08

```
@sanity/cli             0.147.8 (up to date)
@sanity/base            0.147.8 (up to date)
@sanity/components      0.147.9 (up to date)
@sanity/core            0.147.9 (up to date)
@sanity/default-layout  0.147.9 (up to date)
@sanity/default-login   0.147.0 (up to date)
@sanity/desk-tool       0.147.9 (up to date)
@sanity/vision          0.147.9 (up to date)
```

macOS 10.15

```
6.13.4
v12.14.1
```

**Additional context**

Schema for the above:

```
    {
      title: 'Article',
      name: 'article',
      type: 'document',
      fields: [
        {
          title: 'Title',
          name: 'title',
          type: 'text',
        },  {
          title: 'Age',
          name: 'age',
          type: 'number',
          validation: Rule => Rule.custom(val => {
            console.log('age: Rule.custom()');
            if (val < 50) {
              return true
            }
            return 'Too old'
          })
        },  {
          title: 'Location',
          name: 'location',
          type: 'string',
        }, {
          title: 'Meta',
          name: 'meta',
          type: 'object',
          fields: [
            {
              title: 'Description',
              name: 'description',
              type: 'text',
              validation: Rule => Rule.custom(() => {
                console.log('*** meta.description: Rule.custom()');
                return 'Nope'
              })
            },
          ]
        }
      ]
    }
```

### LINKED ISSUE #2630: Improve support for optional object fields with required subfields (nested objects, validation)
- In a schema with an optional object-type field that has one or more required fields on it, validations are not executed on the nested object fields until the object is created (as expected)
- Once an editor has provided content for one of the fields, the object is created and the `_type` field is set. Validations on the object fields are now evaluated
- At this point it is not possible for an editor in the Studio to remove the created object. While clearing simple fields is possible, the object with the `_type` property remains
- The result is validation errors on required fields that prevent the document from being published

<details>
  <summary>Example schema (expand to view)</summary>

```js
const document = {
  title: "Example Document",
  name: "example",
  type: "document",
  fields: [
    {
      title: "Optional Quote",
      name: "quote",
      type: "quote",
    },
  ]
}

const quote = {
  title: "Quote",
  name: "quote",
  type: "object",
  fields: [
    {
      title: "Quoted Text",
      name: "text",
      type: "text",
    },
    {
      title: "Author",
      name: "authorName",
      type: "string",
    },
  ]
}
```
</details>

There are a number of potential solutions. A few I’ve considered include…

- A UI control enabling an editor to unset an object in the Studio (this has the bonus of working for other scenarios as well where it might just be more convenient than clearing multiple nested fields)
- Automatically removing objects that only have internal, non-content properties like `_type` and `_key` set (potentially a bit dangerous)
- Disabling validation on fields on an object if the object only has internal, non-content properties set

<details>
  <summary>
As an interim workaround, I use a document action that implements the second solution above (expand for full code).
<br /><br />

```js
import defaultResolve from "part:@sanity/base/document-actions"
import RemoveEmptyObjects from "../actions/RemoveEmptyObjects"

export default (props) => [...defaultResolve(props), RemoveEmptyObjects].filter(Boolean)

// expand for summary for full document action code
```
</summary>

```js
import { useDocumentOperation } from "@sanity/react-hooks"

const RemoveEmptyObjects = (props) => {
  const { patch } = useDocumentOperation(props.id, props.type)

  // Only operates on drafts because it shouldn't be possible
  // to get into an invalid state on a published document
  if (!props.draft) {
    return null
  }

  return {
    label: "Remove empty objects",
    onHandle: () => {
      const emptyObjects = findEmptyObjectPaths(props.draft)

      if (emptyObjects.length > 0) {
        patch.execute([{ unset: emptyObjects }])
      }
    },
  }
}

export default RemoveEmptyObjects

const findEmptyObjectPaths = (input, startPath = "") =>
  typeof input === "object"
    ? Object.entries(input).reduce((emptyObjPaths, [key, value]) => {
        const currentPath = [startPath, key].filter(Boolean).join(".")

        if (Array.isArray(value)) {
          // Recurse for each entry, expanding results with previous results
          return [
            ...emptyObjPaths,
            ...value.flatMap((v) =>
              typeof v === "object"
                ? findEmptyObjectPaths(v, `${currentPath}[_key=="${v._key}"]`)
                : []
            ),
          ]
        } else if (typeof value === "object") {
          if (isEmptyObject(value)) {
            emptyObjPaths.push(currentPath)
            return emptyObjPaths
          } else {
            // Recurse for each item in the object
            return [
              ...emptyObjPaths,
              ...findEmptyObjectPaths(value, currentPath),
            ]
          }
        }

        return emptyObjPaths
      }, [])
    : []

// Ignore internal fields that do not indicate content (e.g. keep if _ref present)
const isEmptyObject = ({ _type, _key, ...fields }) =>
  Object.keys(fields).length === 0
```
</details>

ricokahler:
### Description

> currently if the value is undefined or null, it won't run any nested validation at all. even if required.
> I think we should run nested validation if the field is required — even if that value is undefined.

See here: https://github.com/sanity-io/sanity/issues/1713#issuecomment-907546288

Closes #1713
Closes #2630

CleanShot 2021-08-27 at 21 45 05@2x

See the changes to the example validation schema for more info.

### What to review

- Ensure this change doesn't break any of the current expectations in validation
- Try it out in the studio and see if it behaves the way you'd expect it to or if it's a regression.

### Notes for release

- Validation markers will appear on nested fields if the containing field is marked as `Rule.required()`.<br />Fixes #1713
- Objects and arrays are now unset if they are empty.<br />Fixes #2630

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
@@ -13,6 +13,7 @@ import {FormFieldSet} from '@sanity/base/components'
 
 import {Grid} from '@sanity/ui'
 import PatchEvent, {set, setIfMissing, unset} from '../../PatchEvent'
+import {applyAll} from '../../patch/applyPatch'
 import {EMPTY_ARRAY} from '../../utils/empty'
 import {ObjectInputField} from './ObjectInputField'
 import {UnknownFields} from './UnknownFields'
@@ -25,6 +26,10 @@ const EMPTY_PATH: Path = EMPTY_ARRAY
 
 function isSingleFieldset(fieldset: Fieldset): fieldset is SingleFieldSet {
   return Boolean(fieldset.single)
+}
+
+function isRecord(obj: unknown): obj is Record<string, unknown> {
+  return typeof obj === 'object' && obj !== null && !Array.isArray(obj)
 }
 
 export interface Props {
@@ -74,8 +79,31 @@ export const ObjectInput = memo(
     } = props
 
     const handleFieldChange = React.useCallback(
-      (fieldEvent: PatchEvent, field) => {
+      (fieldEvent: PatchEvent, field: ObjectField) => {
         let event = fieldEvent.prefixAll(field.name)
+
+        const patchesIncludesUnset = event.patches.some((patch) => patch.type === 'unset')
+
+        // check if the result of the incoming patches will result in an empty object.
+        // first apply all the patches to the current value to get the result.
+        const result =
+          // check if the patch includes an unset and bail out early if not
+          patchesIncludesUnset &&
+          applyAll(
+            value || {},
+            // unset the `_type` when computing the result since its auto-populated.
+            // see note below for `_key`
+            event.append(unset(['_type'])).patches
+          )
+
+        // if the result has no keys left in it.
+        // note: for arrays we retain empty objects so `_key` is not considered
+        if (isRecord(result) && Object.keys(result).length === 0) {
+          // then unset the whole object
+          onChange?.(PatchEvent.from(unset()))
+          return
+        }
+
         if (!isRoot) {
           event = event.prepend(setIfMissing(type.name === 'object' ? {} : {_type: type.name}))
           if (value) {
--- a/examples/test-studio/schemas/validation.js
+++ b/examples/test-studio/schemas/validation.js
@@ -533,5 +533,32 @@ export default {
       },
       inputComponent: CustomObjectSelectInput,
     },
+    {
+      name: 'requiredObjectWithFields',
+      description:
+        'Because this object is required, the sub-field validation will run even if the object is undefined.',
+      type: 'object',
+      validation: (Rule) => Rule.required().error('Please fill out this object'),
+      fields: [
+        {
+          name: 'title',
+          type: 'string',
+          validation: (Rule) => Rule.required().error('Title is required'),
+        },
+      ],
+    },
+    {
+      name: 'optionalObjectWithFields',
+      description:
+        'Because this object is not required, the sub-field validation will NOT run if the object is undefined. You must touch it first.',
+      type: 'object',
+      fields: [
+        {
+          name: 'title',
+          type: 'string',
+          validation: (Rule) => Rule.required().error('Title is required.'),
+        },
+      ],
+    },
   ],
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -7,9 +7,9 @@ const changed = require('gulp-changed')
 const filter = require('gulp-filter')
 const chalk = require('chalk')
 const babel = require('gulp-babel')
-const {runTsc} = require('./scripts/runTsc')
 const log = require('fancy-log')
 const through = require('through2')
+const {runTsc} = require('./scripts/runTsc')
 
 const {getPackagePaths} = require('./scripts/utils/getPackagePaths')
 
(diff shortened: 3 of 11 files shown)


mariuslundgard:
Here‘s a comment on the UX of this only:

I find it problematic that these 2 fields behave exactly the same:
image

The second field claims to be optional, but that’s not the case if you have added a character to the field, and then removed that character (then it’s not optional anymore).

Almost makes me wonder if we should `unset` the object value when it doesn’t have any properties. What do you think? Will that have unwanted consequences?

coreyward:
@ricokahler Would be great to see this address #2630, which I believe is the same as what @mariuslundgard describes above.

skogsmaskin:
> It makes me think we should unset the object value when it doesn’t have any properties. What do you think, @rexxars? Will that have unwanted consequences?

I agree. The object container should check if object text fields are empty and other field types are undefined. If so, remove the object.

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
```diff
@@ -74,8 +79,24 @@ export const ObjectInput = memo(
...
+        // check if the result of the incoming patches will result in an empty object.
+        // first apply all the patches to the current value to get the result.
+        const result = applyAll(
+          value || {},
+          // unset the `_type` when computing the result since its auto-populated
+          event.append(unset(['_type'])).patches
+        )
+
+        // if the result has no keys left in it
+        if (isRecord(result) && Object.keys(result).length === 0) {
```
bjoerge, line 94: Can we add a note about how `_key` for array objects fit into this? Currently, objects in arrays are retained because they have a `_key` property. This is how it has to work, but it seems rather coincidental from reading this code.

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
```diff
@@ -74,8 +79,24 @@ export const ObjectInput = memo(
     } = props
 
     const handleFieldChange = React.useCallback(
-      (fieldEvent: PatchEvent, field) => {
+      (fieldEvent: PatchEvent, field: ObjectField) => {
         let event = fieldEvent.prefixAll(field.name)
+
+        // check if the result of the incoming patches will result in an empty object.
+        // first apply all the patches to the current value to get the result.
+        const result = applyAll(
```
bjoerge, line 87: I'm a bit concerned about potential performance penalties with doing this, especially for patches that targets deeply nested objects. That said, I did some quick testing without noticing any significant difference, so I guess it's fine. Something to keep an eye out for nevertheless.

bjoerge (approved):
LGTM, would be good to add a comment about `_key` for array items as mentioned inline.

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
ricokahler: i think that's a good point. i could probably check if there is an `unset` in the chain and bail out if that's not found. i think that works out for the most part

PR #2723, opened 2021-08-30
refactor(language-filter): migrate to sanity ui and typescript

sjelfull:
### Description

This refactor migrates the Language Filter plugin to Sanity UI and TypeScript.

The UI has changed to this:

image

### What to review

- Run example studio and check if language filter works for `Blogpost`
- Review UI for language filter 
- Check that selected languages is persisted
- Check that config options like `documentTypes` still works

### Notes for release

Migrated `@sanity/language-filter` to Sanity UI


packages/@sanity/language-filter/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../../tsconfig",
+  "include": ["src"],
+  "compilerOptions": {
+    "composite": true,
+    // todo: we need to disable noImplicitAny in a transition phase and should be removed eventually
+    "noImplicitAny": false,
```
bjoerge, line 7: since we're converting this package to typescript now, would it be much extra work to not include this? (it's a flag we've traditionally only been setting on older packages that would be too much work to get to strict mode)

bjoerge (approved):
LGTM! Would be great to have it pass strict mode too, if possible

packages/@sanity/language-filter/tsconfig.json
sjelfull: Can do, yes.

PR #2725, opened 2021-08-31
feat(core): GraphQL: treat added optional input fields as non-dangerous

rexxars:
### Description

When adding fields to your schema, a side effect is that the input types for field filters and sortings will also get new fields (for instance, adding a `name` field on a `BlogPost` type will yield the follow "dangerous changes":

```
An optional field `name` on input type BlogPostFilter was added.
An optional field `name` on input type BlogPostSorting was added.
```

It's hard to reason about why this should be considered a dangerous change, and it adds quite a bit of noise when trying to automate deployments of schemas. 

This PR explicitly ignores this specific warning. 

### What to review

- That the code makes sense
- That the _change_ makes sense

### Notes for release

- Added schema fields will no longer be treated as a dangerous change when deploying GraphQL schemas


skogsmaskin (approved):
LGTM! :+1:

PR #2727, opened 2021-08-31
chore: consolidate jest config across monorepo packages

bjoerge:
### Description
This consolidates all the different variations of jest config spread across the different packages into a common, extendable config file. This enables, among other things, running all jest tests in one go, which offers significant speed improvements both locally and during CI.

### What to review
Review configuration and see that it makes sense. The significant commiit is 0f6b8c6cf1

### Notes for release
n/a


rexxars (approved):
👏  Great work! LGTM

PR #2729, opened 2021-09-01
chore: upgrade image-url library

rexxars:
### Description

Upgrades the image library to the latest version, which solves a (small) annoyance where passing a dpr property of `1` would result in `dpr=1` to be added to the URL (which is unnecessary, given that is the default). Also - a more logical semver range - yay!

### What to review

That things still work


bjoerge (approved):
lgtm!

PR #2731, opened 2021-09-02
chore: upgrade terser

ricokahler:
### Description

There were some issue around the minification step in `sanity build` around `resolve-bin` so I removed it and opted for the Terser JS API.

### What to review

See if `sanity build` still works with no regression. Also affects the `@sanity/client` build since that was also using an older version of terser.

### Notes for release

N/A


rexxars (approved):
👍

PR #2734, opened 2021-09-03
fix(portable-text-editor): fix re-render regression from v2.16.0

skogsmaskin:
### Description

When moving the PTE into the monorepo some refactoring was done to please typescript. It went wrong here obviously making block re-render themselves even though they weren't changed. This is a regression since v2.16.0.

I have fixed the regression.

It should also be added a test for this in the test branch we are working on.

### Notes for release
Fix regression in portable text editor since 2.16.0 where blocks are re-rendered when they should not be.


rexxars (approved):
Thanks for finding the issue and fixing this!

PR #2737, opened 2021-09-03
build: run build before publish-tag

ricokahler:
### Description

I had to run the build prior to running publish tag and that might be a missed step in the future so this may help.

notice that `npm run publish` runs the build prior too

### What to review

Check the logic; try the command


rexxars (approved):
Might have to do a rebase on this one, but the change seems logical - I assume this is an oversight.

PR #2739, opened 2021-09-03
chore: prettier updates; remove husky

ricokahler:
### Description

Note: this PR is waiting on a few larger code changes before merging.

- Adds `trailingComma: 'all'` to `.prettierrc`
- Adds a `.prettierignore` file
- Prettiers all the things
- Removes `validateCommitMessage.js`
- Adds `sort-package-json` and its prettier plugin
- Updates the eslint config to simplify the ignore file. Adds the `--cache` option
- Runs stylelint in CI following the same conventions as eslint
- Updates style-lint and adds uses the prettier style-lint plugin (it was installed but not configured)

### What to review

Review all the new configuration. Make sure different parts are working together (e.g. running prettier should run sort-package-json, eslint should run prettier, verify-format should never mismatch, etc).


rexxars:
@ricokahler Would you mind rebasing this? I tried, but it has a few too many conflicts for me to reason about.

Glancing at the changes you described, the only potential issue I have is the pre-commit lint hook. I'm wondering if it would be possible to move it to pre-push? At least for my workflow, I tend to do small in-progress commits which might not lint, which I later fix and squash - but open to hearing counter arguments if people find this useful on pre-commit.

rexxars (approved):
Looks good - but I would prefer to hold this until after merging the default layout PR at the very least, since I can see this causing a whole slew of conflicts due to prettierifying. Maybe also post-panes/desk-tool migration. Also, there's a tiny lint error that should be fixed :)

.github/workflows/lint.yml
```diff
@@ -52,12 +52,15 @@ jobs:
...
-          # Don't include files like LICENSE and css files
-          CHANGED_JS_FILES=$(echo "$CHANGED_FILES" | sed -e '/^.\+\(mjs\|tsx\|jsx\|js\|ts\)$/!d')
+          # Don't include files like LICENSE, md, etc
+          FILES_TO_LINT=$(echo "$CHANGED_FILES" | sed -e '/^.\+\(js\|ts\tsx\|css\)$/!d')
 
-          for FILE_NAME in $CHANGED_JS_FILES
+          for FILE_NAME in $FILES_TO_LINT
           do
-            npm run eslint -- $FILE_NAME
+            # if you pass a direct filename to eslint nd syt, you don't need to specify extensions
```
bjoerge, line 60: this sentence reads a bit weird to me…

.eslintignore
```diff
@@ -1,13 +1,8 @@
...
-/packages/@sanity/*/node_modules
-/packages/*/dist
-/packages/@sanity/*/dist
-/packages/*/lib
-/packages/*/coverage
-/packages/@sanity/*/lib
-/packages/@sanity/*/coverage
+**/dist
+**/node_modules
+**/lib
```
ricokahler, line 3: note to self: check for `lib` folders in `src` (or reconsider)

mariuslundgard: I sometimes use `lib` directory in source code. We should switch to `dist` for compiled files imo. 😇

rexxars: I know `lib` means something different in other languages/ecosystems, but in the JS world I feel it's so established as being "compiled source" (similar to dist) that I tend to read it as an error when things import from say `../../lib/some/thing`. So would personally steer away from using it.

PR #2741, opened 2021-09-06
Test environment and tests for collaborative editing

skogsmaskin:
### Description

This will add a collaborative editing test suite to the `@sanity/portable-text-editor`.

It's implemented as a Jest environment, using Puppeteer to control real life browsers.

Finally we are able to test things properly like selection adjustment on incoming patches.

The collaborative jest environment will provide each test two new global functions:
```
declare global {
  function getEditors(): Promise<Editor[]>
  function setDocumentValue(value: Value): void
}
```

And a editor can be controlled with this interface:

```
type Editor = {
  editorId: string
  focus: () => Promise<void>
  getSelection: () => Promise<EditorSelection | null>
  getValue: () => Promise<Value>
  insertText: (text: string) => Promise<void>
  pressKey: (keyName: string, times?: number) => Promise<void>
  setSelection: (selection: EditorSelection | null) => Promise<void>
  testId: string
}
```

One can then produce tests like this:
```
    await setDocumentValue(initialValue)
    const [editorA, editorB] = await getEditors()
    await editorA.setSelection({
      anchor: {path: [{_key: 'randomKey0'}, 'children', {_key: 'randomKey1'}], offset: 11},
      focus: {path: [{_key: 'randomKey0'}, 'children', {_key: 'randomKey1'}], offset: 11},
    })
    await editorA.insertText(' world')
    await editorB.pressKey('Enter')
    await editorB.insertText('Starting a new line!')
    const valA = await editorA.getValue()
    const valB = await editorB.getValue()
    expect(valA).toEqual(valB)
```

See https://github.com/sanity-io/sanity/pull/2741/files#r702683234 for a full example.

### What to review

There is a lot to review here, but the most important would be the changes made to the editor itself.
But as they are made to make the tests pass I'm pretty confident about them :)

I would love some feedback on the test environment. I'ts now adding about 10 seconds to the tests on Github (still under a minute). We could decide to run these as an own thing and not part of the standard test suite.

### Notes for release

Test environment and tests for collaborative editing with the Portable Text Editor base package.
Fixed selection adjustment bugs in the editor.


.github/workflows/test.yml
```diff
@@ -29,7 +29,6 @@ jobs:
       - name: Install npm dependencies
         env:
           CYPRESS_INSTALL_BINARY: '0'
-          PUPPETEER_SKIP_DOWNLOAD: 'true'
```
skogsmaskin, line 32: We need to download here now if we want to run the collaboration tests as part of the standard test suite.

packages/@sanity/portable-text-editor/test/__tests__/writingTogether.collaborative.test.ts
```diff
@@ -0,0 +1,421 @@
+/**
+ * @jest-environment ./test/setup/collaborative.jest.env.ts
+ */
+
```
skogsmaskin, line 4: This is how we make a collaborative test by setting `@jest-environment ./test/setup/collaborative.jest.env.ts`

This exposes global test functions like `setDocumentValue` `getEditors` to be used inside each test:
```
type Editor = {
  editorId: string
  focus: () => Promise<void>
  getSelection: () => Promise<EditorSelection | null>
  getValue: () => Promise<Value>
  insertNewLine: () => Promise<void>
  insertText: (text: string) => Promise<void>
  pressKey: (keyName: string, times?: number) => Promise<void>
  setSelection: (selection: EditorSelection | null) => Promise<void>
  testId: string
}
declare global {
  function getEditors(): Promise<Editor[]>
  function setDocumentValue(value: Value): void
}
```

bjoerge (approved):
Solid work. LGTM!

skogsmaskin:
Improved testing selections and key presses. The test now runs 100% reliable on every machine I've tested.

PR #2742, opened 2021-09-06
feat: move styled-components to explicit studio dependency

rexxars:
### Description

`@sanity/ui` and all our UI-related packages now rely on `styled-components`. Until now, we've chosen to keep `styled-components` as an explicit dependency of the individual packages, with a liberal range which should ensure it gets hoisted to the top level and reused.

However, we've seen many scenarios now where for some reason yarn/npm chooses to not do this. A very reproducible example is to bootstrap a new studio and run `yarn add react-twitter-embed`. What happens is that you'll get two copies of `styled-components`, both of the same version (`5.3.1` at the time of writing) - one in `@sanity/base/node_modules` and one in the studio `node_modules`. Theoretically there shouldn't be a copy in `@sanity/base`, since it has the same version as the other dependencies, and should thus be hoisted. Running `yarn why` yields this output:

```
➜  ~/studio yarn why styled-components
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "styled-components"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "styled-components@5.3.1"
info Reasons this module exists
   - "@sanity#base" depends on it
   - Hoisted from "@sanity#base#styled-components"
   - Hoisted from "@sanity#default-layout#styled-components"
   - Hoisted from "@sanity#default-login#styled-components"
   - Hoisted from "@sanity#desk-tool#styled-components"
   - Hoisted from "@sanity#desk-tool#@sanity#field#styled-components"
   - Hoisted from "@sanity#desk-tool#@sanity#form-builder#styled-components"
   - Hoisted from "@sanity#desk-tool#@sanity#form-builder#@sanity#portable-text-editor#styled-components"
```

This seems to indicate that yarn _thinks_ it's hoisted (or should be), but it's actually not.

Long story short, we're seeing too many issues with the current approach, and can't rely on npm/yarn to get the behavior we want.

The change proposed in this PR is to move the dependency from an explicit, direct dependency in each module, to a peer dependency. For this to work, we need the studio to depend on the module instead, so to help people migrate to this new pattern, this PR introduces a step to the CLI `start`, `deploy` and `build` commands which checks whether or not `styled-components` is declared as a dependency, that it's the right version, and that it's actually installed. If it is not declared, we will tell the user that it needs to be, and prompt the user which package manager they want to use to install the dependencies.

This should hopefully solve the case where people are seeing `sanity is not a property of undefined` or similar, which appears because the theme context in `styled-components` does not contain the properties we set on it, because it has been rendered with two different versions.

#### PR notes

This PR also adds some typescript configuration for `@sanity/core` and some workable but placeholder typings for the CLI context and such (after not having looked at the CLI code for a while, I felt like this provided me with some more needed safety). To make this work typescript, I had to install some `@types` dev-dependencies, but some were not available for the version we were depending on, so I had to upgrade a few dependencies (and adjust certain consumers of those dependencies to account for breaking changes, like the release notes script).

Apologies for not making a separate PR, but the fix seemed important enough to warrant a bit of a shortcut here.

### What to review

- Does the change in dependency configuration make sense, and are there any better alternatives?
- Does the code make sense?

To test the change, you can bootstrap a new studio and upgrade to a tagged release: `sanity upgrade --save-exact --tag=styled-peer-dep`, then run `sanity start` or `sanity build` etc. It should prompt for the missing dependency.

### Notes for release

- Move `styled-components` from a dependency to a peer dependency. Studios will now have to install `styled-components` as a dependency.


bjoerge (approved):
LGTM!

mariuslundgard (approved):
LGTM ✨

packages/@sanity/core/src/util/checkRequiredDependencies.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  // The declared version should be semver-compatible with the version specified as a
+  // peer dependency in `@sanity/base`. If not, we should tell the user to change it.
+  //
+  // Exception: Ranges are hard to compare. `>=5.0.0 && <=5.3.2 || ^6`... Comparing this
+  // to anything is going to be challenging, so only compare "simple" ranges/versions
+  // (^x.x.x / ~x.x.x / x.x.x)
+  if (
+    isComparableRange(declaredStyledComponentsVersion) &&
```
ricokahler, line 56: would `semver.valid` work here?

packages/@sanity/core/src/util/checkRequiredDependencies.ts
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @param dependencies Object of dependencies (package name => version)
+ * @param context CLI context
+ */
+async function installDependenciesWithPrompt(
+  dependencies: Record<string, string>,
+  context: CliCommandContext
+): Promise<unknown> {
+  const {output, prompt, workDir, yarn} = context
+  const yarnLockExists = await fileExists(path.join(workDir, 'yarn.lock'))
```
ricokahler, line 158: i typically reach for `fs.existsSync` for this kind of check. i don't know if there are any downsides to synchronous   file existence checks but node says it's not deprecated 🤷. no action requested for this one, just curious on your thoughts of if i shouldn't do that lol

packages/@sanity/core/src/util/checkRequiredDependencies.ts
```diff
@@ -0,0 +1,238 @@
...
+  // The declared version should be semver-compatible with the version specified as a
+  // peer dependency in `@sanity/base`. If not, we should tell the user to change it.
+  //
+  // Exception: Ranges are hard to compare. `>=5.0.0 && <=5.3.2 || ^6`... Comparing this
+  // to anything is going to be challenging, so only compare "simple" ranges/versions
+  // (^x.x.x / ~x.x.x / x.x.x)
+  if (
+    isComparableRange(declaredStyledComponentsVersion) &&
+    !semver.satisfies(
+      normalizeVersion(declaredStyledComponentsVersion),
```
ricokahler, line 58: i think `normalizeVersion` could be replaced with `semver.minVersion`

packages/@sanity/core/src/util/checkRequiredDependencies.ts
```diff
@@ -0,0 +1,238 @@
...
+  if (!installedStyledComponentsVersion) {
+    throw new Error(oneline`
+      Declared dependency \`styled-components\` is not installed - run
+      \`npm install\` or \`yarn\` to install it before re-running this command.
+    `)
+  }
+
+  // The studio should have an _installed_ version of `styled-components`, and it should
+  // be semver compatible with the version specified in `@sanity/base` peer dependencies.
+  if (!semver.satisfies(installedStyledComponentsVersion, wantedStyledComponentsVersionRange)) {
```
ricokahler, line 79: you may want to wrap `installedStyledComponentsVersion` in `semver.minVersion`

```suggestion
  if (!semver.satisfies(semver.minVersion(installedStyledComponentsVersion), wantedStyledComponentsVersionRange)) {
```

packages/@sanity/core/src/actions/start/startAction.js
```diff
@@ -32,7 +33,8 @@ export default async (args, context) => {
     project: sanityConfig.get('project'),
   }
 
-  checkStudioDependencyVersions(workDir)
+  await checkStudioDependencyVersions(workDir)
```
ricokahler, line 36: i think `checkStudioDependencyVersions` runs synchronously

packages/@sanity/core/src/util/checkRequiredDependencies.ts
```diff
@@ -0,0 +1,238 @@
...
+ * Read the `package.json` file at the given path and return an object that guarantees
+ * the presence of name, version, dependencies, dev dependencies and peer dependencies
+ *
+ * @param packageJsonPath Path to package.json to read
+ * @returns Reduced package.json with guarantees for name, version and dependency fields
+ */
+async function readPackageManifest(packageJsonPath: string): Promise<PackageManifest> {
+  let manifest: unknown
+  try {
+    manifest = await readJSON(packageJsonPath)
```
ricokahler, line 132: i wonder if `require(packageJsonPath)` would be better. might be more compatible with node e.g. if someone had registered a require hook in there or something. in theory resolveFrom should catch that though

ricokahler (commented):
this change makes a lot of sense (and reminds me of [this](https://reactjs.org/warnings/invalid-hook-call-warning.html#duplicate-react) a lot lol)

~this comment in particular: https://github.com/sanity-io/sanity/pull/2742#discussion_r703558936~ eh looks good!

i would double check my suggestions around using `semver.minVersion`. besides that, the rest of the comments are minor suggestions and curiosities

packages/@sanity/core/src/util/checkRequiredDependencies.ts
ricokahler: ah nevermind. the version comes from the package itself so there is no carrot or anything. this isn't needed

ricokahler (approved):
LGTM. the one comment i had was fine actually

packages/@sanity/core/src/types.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Note: These typings actually belong in CLI, but for now we only use TS in
+ * _parts_ of `@sanity/core` and in none of `@sanity/cli`. At some point we'll
+ * probably want to rewrite the CLI, with typescript and a more plain sucommand
+ * approach - but until then this serves as a partial typing helper
+ */
```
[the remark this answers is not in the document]
rexxars: This divide between `@sanity/cli` and `@sanity/core` is a bit of a legacy thing which helped us install the CLI tool really quickly (by bundling all required dependencies to a single file) in order to get started with the init process, and have the `@sanity/core` be just a regular dependency which we could upgrade more often, and have regular dependencies declared.

This isn't really necessary anymore, due to faster npm and yarn, and other approaches (such as `npx create-sanity-app` or similar).

Given this is due for a rewrite at some point, I don't want to spend _too_ much time on this, but we could also just move the typings to `@sanity/cli` in some upcoming PR and add `@sanity/cli` as a dev dependency from `@sanity/core`, only importing the types.

Anyway. I feel ok with this for now, and will revisit if/when we do more typescript changes in the CLI.

package.json
```diff
@@ -97,7 +97,7 @@
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-react": "^7.20.6",
     "eslint-plugin-react-hooks": "^4.1.2",
-    "execa": "^1.0.0",
+    "execa": "^2.0.0",
```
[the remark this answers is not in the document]
rexxars: I don't have time to go through all the breaking changes from 1.x to 7.x - will revisit this later

packages/@sanity/core/src/util/checkRequiredDependencies.ts
rexxars: Nothing wrong with that except it blocks the event loop, so spinners and such will pause. Not a huge deal, but also don't see much of a downside in doing it async (except for the potential race condition of the file disappearing before the next operation, but that isn't really relevant in this case)

packages/@sanity/core/src/util/checkRequiredDependencies.ts
rexxars: Ah, beautiful. Thanks.

PR #2744, opened 2021-09-06
feat(default-login): add support for sso login providers

dcilke:
### Description

With the introduction of SSO we need a way to determine if we should allow sso prompts on studio login. The `/auth/providers` endpoint has been updated to include an `sso` object which contains the keys of enabled `sso` types. This PR checks if any of those flags are set and if so, allow custom entities in `default-login.json`.

### Notes for release

This will be a silent release until after the beta period.

### DIFF
--- a/packages/@sanity/default-login/src/util/getProviders.js
+++ b/packages/@sanity/default-login/src/util/getProviders.js
@@ -9,7 +9,7 @@ export default function getProviders() {
       withCredentials: true,
     })
     .then((res) => {
-      const {providers, thirdPartyLogin} = res
+      const {providers, thirdPartyLogin, sso} = res
       const customProviders = (config.providers && config.providers.entries) || []
       if (customProviders.length === 0) {
         return providers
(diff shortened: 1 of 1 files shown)


packages/@sanity/default-login/src/util/getProviders.js
```diff
@@ -20,7 +20,8 @@ export default function getProviders() {
         // login options through the config. These shouldn't be treated as custom login providers
         // which require the third-party login feature, but as the official provider
         const isOfficial = providers.some((official) => official.url === provider.url)
-        return {...provider, custom: !isOfficial, supported: isOfficial || thirdPartyLogin}
+        const isSupported = isOfficial || thirdPartyLogin || (sso && sso[provider.type])
```
sgulseth, line 23: Would it be better to do an explicit check? Or if we want to reduce the configuration surface we could do something like `sso && Object.values(sso).some(Boolean)`?

dcilke: 👍

rexxars (approved):
Looks good to me 👍

PR #2745, opened 2021-09-06
Move base/styles to __legacy folder

RitaDias:
### Description

Move styled to __legacy folder within base.

### What to review

Since the potentially affected areas are rather large and silent (since all of them are css), it's hard to narrow down to a single one.

For example, I followed examples like the `media-preview` that imports parts from these styles (using my branch vs the build).

### Notes for release

Nothing to add!


rexxars (approved):
👍 Thanks! LGTM!

PR #2752, opened 2021-09-07
fix(server): alias styled-components in webpack config

rexxars:
### Description

This PR aliases `styled-components` to the version installed in the studio (or a hoisted copy, potentially) within the webpack config. This is done to prevent the error where multiple versions of `styled-components` are used in the same app unintentionally. Ideally we wouldn't need this, but it seems npm/yarn hoisting isn't always perfect, so this at least provides some way to alleviate that pain. 

### What to review

- Does it make sense?


bjoerge (approved):
LGTM!

ricokahler (approved):

PR #2754, opened 2021-09-08
fix(default-layout): handle render errors with no Error instance

rexxars:
### Description

Fixes an issue where (for some reason) components can crash during rendering
without emitting an Error instance. This leads to the error screen crashing
when attempting to extract the stack trace from it.

With this change, we render a generic unknown error if we don't receive any
error instance, and check for a stack property before attempting to use it.

For additional safety, we also check for the info property and the component
stack being present before attempting to use it.

### What to review

Reproducing this is a bit hard - the example I have found was using `react-player` in a preview component and passing it a valid youtube URL. I think something about the lazy loading approach it takes makes the component crash in some unexpected way.

### Notes for release

- Fixed an error where components causing errors during rendering might crash the error boundary because of a missing error message

### DIFF
--- a/packages/@sanity/default-layout/src/main/ErrorScreen.tsx
+++ b/packages/@sanity/default-layout/src/main/ErrorScreen.tsx
@@ -72,15 +72,24 @@ function ErrorScreen(props: Props) {
 
       {showErrorDetails && (
         <>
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
LGTM

PR #2755, opened 2021-09-08
chore(import-cli): upgrade meow dependency

rexxars:
### Description

Upgrades the meow dependency to the latest version that still supports node.js 10. This fixes a security issue that Snyk reported: https://github.com/sanity-io/sanity/pull/2735

Closes #2735

### What to review

- Run the CLI tool and make sure it parses the flags correctly (it should, from my testing)

```
packages/@sanity/import-cli$ node src/sanity-import.js -p abc123 -d dataset --token=foo --replace --allow-failing-assets some-source.ndjson
```

### Notes for release

None (invisible for end-user)

### DIFF
--- a/packages/@sanity/import-cli/package.json
+++ b/packages/@sanity/import-cli/package.json
@@ -27,7 +27,7 @@
   "dependencies": {
     "@sanity/client": "2.18.0",
     "@sanity/import": "2.18.0",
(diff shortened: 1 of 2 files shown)


skogsmaskin (approved):
LGTM! As you have tested it I trust that it works :)

PR #2760, opened 2021-09-09
chore(components): remove deprecated @sanity/components module

rexxars:
### Description

- Removes the `@sanity/components` module from our `packages` folder - it'll stay at 2.14 forever with a deprecation warning. If we need to do changes to this module later, it's not hard to initialize an empty npm package and publish it outside of the monorepo context. This creates less clutter in the repo and prevents us from publishing a new major of a deprecated package later.
- Prevents the `@sanity/components` dependency from being installed in new studios, and from being added to `sanity.json`
- Removed all references to `@sanity/components` in test stubs (mostly to simplify grepping for `@sanity/components` and seeing where it's still being used/mentioned, this isn't exactly an important change)

### What to review

_Theoretically_ this can cause issues if plugins/customizations are importing from `@sanity/components`, but as this only applies to _new_ studios I don't feel too worried about it. Plugins that depend on this module being present will have had a deprecation warning printed to their console for a while now too, and would only have worked with root-level imports. If you can think of any other cases where this could break, let me know.

### Notes for release

- Deprecated `@sanity/components` - no longer installed for new studios


bjoerge (approved):
LGTM!

PR #2761, opened 2021-09-09
refactor(base): drop usage of parts from default preview

rexxars:
### Description

This PR:
- Moves all imports of the preview resolver part to a central location so it's at least contained. People could theoretically have implemented this so I'm a bit wary of removing it.
- Migrates the asset-url-builder to typescript and exports it as `assetUrlBuilder` (since we now prefer named exports). Didn't do any changes beyond this - this utility should be reworked and removed later.
- Replaces the file icon import and asset url builder import with static imports. I can't picture anyone having overwritten these, and the asset URL builder only kicks in as a fallback.

### What to review

- That the code runs and functions the same as before
- That the changes makes sense

### Notes for release

None. Invisible to the end user.


sjelfull (approved):
LGTM

PR #2762, opened 2021-09-09
refactor(base): migrate search utils to typescript

rexxars:
### Description

- Migrates the search utils to typescript
- Uses regular import of search utils instead of part for search utils (I can't picture anyone overriding these?)

**Note**: The `joinPath` function was rewritten to use a regular for loop instead of a reduce because typescript was being difficult for no good reason, hope it's not too much of an eyesore. Tests are passing so assuming it does what it says on the tin ;)

### What to review

- That the code still makes sense
- That you agree with dropping the overridable search utils part
 
### Notes for release

None. Invisible for end-user.


packages/@sanity/base/test/searchUtils.test.ts
```diff
@@ -22,7 +22,7 @@ test('escapeField', () => {
   expect(escapeField('foobar')).toBe('["foobar"]')
 
   // Keywords
-  ;[('match', 'in', 'asc', 'desc', 'true', 'false', 'null')].forEach((kw) => {
+  ;['match', 'in', 'asc', 'desc', 'true', 'false', 'null'].forEach((kw) => {
```
bjoerge, line 25: eek 🙈

bjoerge (approved):
LGTM! (and well done spotting the unintentional comma expression 😳)

PR #2764, opened 2021-09-10
Refactor/replace router part with regular import

bjoerge:
### Description

This adds the `@sanity/base/router` export and replaces internal usage of `part:@sanity/base/router` with it. This lets us not having to maintain a separate dependency to @sanity/state-router for studio packages that needs routing, and minimises the risk of consuming duplicate versions.

### What to review
Check that the code looks sane, that studio routing still works

### Notes for release

N/A. This is an internal refactoring.


rexxars (approved):
👍

PR #2768, opened 2021-09-13
feat(datasets): increase max dataset name length

dcilke:
### Description

Increase the dataset name character limit to 64

### Notes for release

Dataset name limit increased to 64 characters

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/promptForDatasetName.js
+++ b/packages/@sanity/cli/src/actions/init-project/promptForDatasetName.js
@@ -1,9 +1,11 @@
 import chalk from 'chalk'
 
+const MAX_DATASET_NAME_LENGTH = 64
+
 const datasetNameError =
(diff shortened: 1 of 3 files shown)


rexxars (approved):

skogsmaskin (commented):
Would it make sense to have `const MAX_DATASET_NAME_LENGTH = 64` be defined from one single place from somewhere? Like in Sanity core.

rexxars:
> Would it make sense to have `const MAX_DATASET_NAME_LENGTH = 64` be defined from one single place from somewhere? Like in Sanity core.

Theoretically, but it's a bit clunky since `@sanity/cli` doesn't depend on `@sanity/core`, and `@sanity/core` can't _import_ from `@sanity/cli`, so it would have to be passed down as some sort of context argument. Good suggestion though.

skogsmaskin:
Guess it helps anyway to have something constant to search for throughout the code.

PR #2769, opened 2021-09-13
fix(base): fix usage of inset property

sjelfull:
### Description

This reverts the usage of `inset` css property instead of `left/right/top/bottom: 0`. The property is only [supported in Safari 14.1](https://caniuse.com/?search=inset) and up, which was released this year, so we should hold off on using it until support improves.


rexxars (approved):
👍

PR #2772, opened 2021-09-14
refactor(base): migrate Presence to `@sanity/ui` and styled components

RitaDias:
### Description

Migrate presence from /base to use `@sanity/ui` and styled components

### What to review

- It's likely easier to do it if you have two page instances open in order to see your user on the presence of a document
  - I used `presence-test` -> and any of the files within

- You should also find another user in the studio and try to be in the same document
  - hover over the users to be sure that the tooltips work as expected
  - make sure that the top and bottom sticky work as expected

- Make sure to test out when the Presence is on debug 
  - go to `presence` -> `constants.ts`
  - Change `debug` constant to `true`

### Notes for release

Migrate Presence to use `@sanity/ui` and styled components


rexxars (commented):
Are these tooltips supposed to be square, or rounded in the same way as the validation? 

image

RitaDias:
> Are these tooltips supposed to be square, or rounded in the same way as the validation?
> 
> image

Good catch! After double checking I think they are meant to be rounded! Not sure why they seem to be snipped like that but I'll have a look 👀

packages/@sanity/base/src/presence/overlay/WithIntersection.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import {tap} from 'rxjs/operators'
+import {ObservableIntersectionObserver} from './intersectionObserver'
```
bjoerge, line 3: this would preferably be a type import: `import type {ObservableIntersectionObserver} from './intersectionObserver'`

packages/@sanity/base/src/presence/FieldPresence.tsx
```diff
@@ -41,7 +35,7 @@ function FieldPresenceWithOverlay(props: FieldPresenceProps) {
   const minWidth = -AVATAR_DISTANCE + (AVATAR_SIZE + AVATAR_DISTANCE) * props.maxAvatars
 
   return (
-    <div ref={ref} className={styles.root} style={{minWidth: minWidth, minHeight: AVATAR_SIZE}} />
+    <FlexBox justify="flex-end" ref={ref} style={{minWidth: minWidth, minHeight: AVATAR_SIZE}} />
```
bjoerge, line 38: I find the naming here a bit confusing. Is `FlexBox` any different from `<Flex><Box>(...)</Box></Flex>`?

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
```diff
@@ -0,0 +1,65 @@
...
+    top: ${margins[0] - 1}px;
+    background-color: ${debug ? 'red' : 'none'};
+  `
+})
+
+export const MiddleRegionWrapper = styled(WithIntersection)(({debug}: styleProps) => {
+  return css`
+    ${RegionWrapper}
+
+    visibility: 'none';
```
bjoerge, line 45: don't think there should be quotes around `none` here

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    ${debug &&
+    css`
+      background: rgba(255, 0, 0, 0.25);
+      outline: 1px solid #00b;
+      visibility: 'visible';
```
bjoerge, line 51: same as above re: quotes

bjoerge (changes requested):
Left a couple comments inline. I think the quoted css properties (+ the aforementioned tooltip/padding issue) should be fixed before merge, looks great otherwise!

packages/@sanity/base/src/presence/FieldPresence.tsx
RitaDias: Ah, this is just a misunderstanding 😄 
I was following one of the suggestions of the naming guidelines which suggested "[name]box" and I understand that in this situation it would be very confusing. It's unrelated. 

I'm going to be changing it to `FlexWrapper` :)

packages/@sanity/base/src/presence/overlay/WithIntersection.tsx
RitaDias: I will change it! :D 
Can you go into more detail as to what's the difference between one or the other so that I can tell when to use one vs the other next time? 🙏

bjoerge: the type import will be removed entirely in transpiled code/at runtime. When importing types we should always use type imports, but we haven't gotten around to enforce it with tooling yet :)

RitaDias: Gotcha 🙌 thanks!

RitaDias:
> Are these tooltips supposed to be square, or rounded in the same way as the validation?
> 
> image

This is now fixed 🙌

packages/@sanity/base/src/presence/PresenceTooltip.tsx
```diff
@@ -1,40 +1,37 @@
...
 
   const content = (
-    <div className={styles.root}>
+    <Card radius={2} padding={2}>
       {items.map((item) => (
-        <div className={styles.item} key={item.user.id}>
-          <div className={styles.avatarContainer}>
-            <UserAvatar user={item.user} status="online" />
-          </div>
+        <Card key={item.user.id} padding={1}>
```
mariuslundgard, line 20: Same here – this should be `<Box key={...`

packages/@sanity/base/src/presence/PresenceTooltip.tsx
```diff
@@ -1,40 +1,37 @@
...
+          </Flex>
+        </Card>
       ))}
-    </div>
+    </Card>
   )
 
   return (
-    <Tooltip content={content} placement={placement}>
+    <Tooltip content={content} open placement={placement}>
```
mariuslundgard, line 34: I believe `open` is not needed here?

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
```diff
@@ -0,0 +1,65 @@
...
+    `}
+  `
+})
+
+export const BottomRegionWrapper = styled(WithIntersection)(({debug}: styleProps) => {
+  return css`
+    ${RegionWrapper}
+
+    position: sticky;
+    bottom: -1;
```
mariuslundgard, line 61: This should probably be `-1px`?

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
```diff
@@ -0,0 +1,65 @@
...
+`
+
+export const OverlayWrapper = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  pointer-events: none;
+  z-index: 5;
```
mariuslundgard, line 20: This `z-index` actually needs to be higher than 10, since `ChangeBar` makes each form field have `z-index: 10` on :hover.

PS. This is an old bug, so strictly not in scope in this PR. But maybe worth doing anyway?

image

mariuslundgard (approved):
Super-solid work. Have some minor comments here and there, tho :-)

packages/@sanity/base/src/presence/PresenceTooltip.tsx
```diff
@@ -1,40 +1,37 @@
...
-  placement?: TooltipPlacement
+  placement?: Placement
 }
 
 export function PresenceTooltip(props: PresenceTooltipProps) {
   const {children, items, placement} = props
 
   const content = (
-    <div className={styles.root}>
+    <Card radius={2} padding={2}>
```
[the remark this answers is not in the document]
RitaDias: Thank you for the tip! Done 👍

packages/@sanity/base/src/presence/PresenceTooltip.tsx
RitaDias: Done!

packages/@sanity/base/src/presence/PresenceTooltip.tsx
RitaDias: Good catch, this is a remnant from when I was trying the popovers out to debug a visual bug :) 
Will remove!

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
RitaDias: Yep! Also changing it!

packages/@sanity/base/src/presence/overlay/RegionsWithIntersections.styled.tsx
RitaDias: Updated the z-index to be 13. I was doing some more tests and noticed that this also happened when the `ActivateOnFocus` overlay was active. That overlay (`ActivateOnFocus`) is at z-index 12, so this value I picked is the lowest that works.

mariuslundgard (approved):
LGTM ✨

PR #2777, opened 2021-09-15
refactor(state-router): move hooks to separate files and simplify

mariuslundgard:
### Description

- Moves `useRouter` and `useRouterState` to separate files.
- Removes opaque `deps` logic from `useRouterState` (it was not in use).

### DIFF
--- a/packages/@sanity/state-router/src/useRouterState.ts
+++ b/packages/@sanity/state-router/src/useRouterState.ts
@@ -1,0 +1,17 @@
+import {useContext, useEffect, useState} from 'react'
+import {RouterState} from './components/types'
+import {RouterContext} from './RouterContext'
+
+export const useRouterState = (): RouterState => {
+  const router = useContext(RouterContext)
+  const [routerState, setState] = useState<RouterState>(router.getState())
+
+  useEffect(() => {
+    // subscribe() returns an unsubscribe function, so this'll handle unmounting
+    return router.channel.subscribe(() => {
+      setState(router.getState())
+    })
+  }, [router])
+
+  return routerState
+}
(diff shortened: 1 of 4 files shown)


packages/@sanity/state-router/src/useRouterState.ts
```diff
@@ -0,0 +1,17 @@
...
+export const useRouterState = (): RouterState => {
+  const router = useContext(RouterContext)
+  const [routerState, setState] = useState<RouterState>(router.getState())
+
+  useEffect(() => {
+    // subscribe() returns an unsubscribe function, so this'll handle unmounting
+    return router.channel.subscribe(() => {
+      setState(router.getState())
+    })
+  }, [router])
```
bjoerge, line 14: This should probably be `[router.channel]` to avoid excessive resubscribes (afaik `router` changes more frequently)?

bjoerge (commented):
is there any real benefit of moving useRouter + useRouterState to separate files? The advantage of having them in the same file as where we declare the context is that we don't need to export the context, and therefore can avoid exposing it for direct consumption.

packages/@sanity/state-router/src/useRouterState.ts
mariuslundgard: Good point, will fix 👍

mariuslundgard: Ah, it was because it’s using `router.getState()` below there.

bjoerge: Ah, I see! Probably better to push down the new state at some point, but let's leave that for later 👍🏼

mariuslundgard:
> is there any real benefit of moving useRouter + useRouterState to separate files? The advantage of having them in the same file as where we declare the context is that we don't need to export the context, and therefore can avoid exposing it for direct consumption.

Just clean and easier to locate in my opinion. Also, the `RouterContext` has been explicitly exported for a while. Fwiw, it’s possible to not export it even with separate files.

bjoerge (approved):
> However the RouterContext has been explicitly exported for a while (not introduced by this change).

Ah, I missed that. Approved!

PR #2781, opened 2021-09-15
refactor(base): preview improvements

mariuslundgard:
### Description

Notable changes:
- `DefaultPreview`
  - Hides the skeleton for `media`, if the preview is configured to not have icon/image.
  - Adjusts padding depending on existence of `media`
- `DetailPreview`
  - Adds rounded corner to media images, similar to `DefaultPreview`.
  - Adds inner shadow to media images, similar to `DefaultPreview`
  - Adjusts the sizing of the icons
  - Remove CSS that is provided by `@sanity/ui`
- `SanityDefaultPreview`
  - Migrate to using icons from `@sanity/icons`

### DIFF
--- a/packages/@sanity/base/src/components/previews/detailPreview.tsx
+++ b/packages/@sanity/base/src/components/previews/detailPreview.tsx
@@ -2,15 +2,7 @@ import React from 'react'
 import {Box, Stack, Text, Skeleton, TextSkeleton, useTheme} from '@sanity/ui'
 import {getDevicePixelRatio} from 'use-device-pixel-ratio'
 import {MediaDimensions, PreviewProps} from './types'
-import {
-  Root,
-  Top,
-  Content,
-  Header,
-  StatusWrapper,
-  MediaWrapper,
-  MediaString,
-} from './detailPreview.styled'
+import {Root, Top, Content, Header, StatusWrapper, MediaWrapper} from './detailPreview.styled'
 
 const DEFAULT_MEDIA_DIMENSIONS: MediaDimensions = {
   width: 80,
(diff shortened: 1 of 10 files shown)


packages/@sanity/base/src/preview/components/SanityPreview.tsx
```diff
@@ -1,15 +1,15 @@
...
-import {SortOrdering, Type} from '../types'
 
 interface Props {
   type: Type
   fields: string[]
   value: any
   ordering?: SortOrdering
   children: (props: any) => React.ComponentType
-  layout: string
+  layout?: 'inline' | 'block' | 'default' | 'card' | 'media'
```
sjelfull, line 12: We should remove `card` from this type everywhere since its deprecated

packages/@sanity/base/src/preview/components/PreviewSubscriber.tsx
```diff
@@ -11,7 +11,7 @@ interface Props {
   value: any
   ordering?: SortOrdering
   children: (props: any) => React.ReactElement
-  layout?: string
+  layout?: 'inline' | 'block' | 'default' | 'card' | 'media'
```
sjelfull, line 14: Same as the others

packages/@sanity/base/src/preview/components/PreviewFields.tsx
```diff
@@ -1,24 +1,33 @@
...
+
   return typeof val === undefined ? [] : [val]
 }
 
-type Props = {
-  document: Document
+interface PreviewFieldsProps {
+  document: SanityDocument
   fields: string | string[]
+  layout?: 'inline' | 'block' | 'default' | 'card' | 'media'
```
sjelfull, line 17: Same as the others

packages/@sanity/base/src/components/previews/detailPreview.tsx
```diff
@@ -58,9 +58,11 @@ export const DetailPreview: React.FunctionComponent<PreviewProps<'detail'>> = (p
             })}
           {typeof media === 'string' && <MediaString>{media}</MediaString>}
           {React.isValidElement(media) && media}
+          <span />
```
sjelfull, line 61: This is to render the shadow right? Could you use a pseudo element?

sjelfull (approved):
Looks good for the most part, left some suggestions!

packages/@sanity/base/src/components/previews/detailPreview.tsx
mariuslundgard: Yes, it’s to render the shadow.

The thing is we only want the shadow when then `<MediaWrapper>` contains something else `<svg>` – icons should not have the shadow.

So the idea is the use the `*:not(svg) + span` selector target only that situation to render the shadow.

Will add a comment about this in the code 👍

PR #2782, opened 2021-09-15
ci(github): run tests on windows

rexxars:
### Description

This PR is a (work in progress) attempt at ensuring we have full windows support in our monorepo modules.

### Notes for release

None. Transparent to the end user.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -66,6 +66,6 @@ jobs:
           SLACK_BOT_TOKEN: ${{ secrets.ACTIONS_SLACK_BOT_TOKEN }}
         uses: voxmedia/github-action-slack-notify-build@v1
         with:
-          channel: prj-components
+          channel: studio-internal
           status: FAILED
           color: danger
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ on: [push]
 
 jobs:
   test:
+    name: Test (${{ matrix.os }} / node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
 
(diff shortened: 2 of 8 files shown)


hdoro:
I fixed the errors in `jest` - all of which happened in `resolvePlugins.test.js`.

Now we need to fix the tests in `lerna run --scope groq --scope '@sanity/{client,generate-help-url,plugin-loader}' test`. The first error to tackle is on the `@sanity/client` package, which starts its `test` command with `NODE_ENV=test`. We'll probably need to reach for [cross-env](https://www.npmjs.com/package/cross-env) or similar to do that. Suggestions, @rexxars ?

I also need to fix some linting errors that I couldn't test locally. We'll get to those before merging :)

rexxars:
> I fixed the errors in `jest` - all of which happened in `resolvePlugins.test.js`.

Obrigado! I agree with the comments in your commit message, these paths do seem a little random. Given that we're migrating away from parts, hopefully we can live with this being a bit random until we can get rid of it ;)

> The first error to tackle is on the `@sanity/client` package, which starts its `test` command with `NODE_ENV=test`. We'll probably need to reach for [cross-env](https://www.npmjs.com/package/cross-env) or similar to do that. Suggestions, @rexxars ?

I investigated briefly and there doesn't seem to be anything in the client that actually depends on this environment variable, so I simply dropped it from the test script. We should/could probably use cross-env if there are other modules that require some of these env vars to be set, good call.

> I also need to fix some linting errors that I couldn't test locally. We'll get to those before merging :)

Fair! I took the liberty of prettifying and squashing, hope that's not too intrusive :)

hdoro:
Thank you for putting in the work, Espen!

I lack context on all changes, but didn't see anything suspicious. Just managed to pass all tests on my machine 🎉

bjoerge (approved):
LGTM! Great stuff!

PR #2784, opened 2021-09-16
fix(form-builder): don't unset empty objects if on root level

bjoerge:
### Description

This fixes an issue introduced in v2.18.0 that would in some rare cases trigger an unset patch on the document itself, which again would trigger studio error. The fix is to not do the empty check if we're on root level (e.g. the object input is used for the document itself).

### What to review
The proposed code change to see if it makes sense.

### Notes for release

Fixes a bug that could in some rare cases trigger an error saying "Cannot read properties of null (reading 'expr')"


rexxars (approved):
👍  Makes sense to me!

PR #2787, opened 2021-09-16
chore: global eslint jest config

ricokahler:
### Description

Since #2727, we now have the ability to run jest at the top-level of the repo. The following is a complementary change that unifies eslint errors/warnings for any files that match `*.test.{js,ts,tsx}`

### What to review

See if the configuration is correct.


bjoerge (approved):
Nice!

PR #2788, opened 2021-09-16
feat(structure): allow specifying API version for document lists

rexxars:
### Description

This PR introduces two new methods on `DocumentList` and friends:
- `apiVersion(someApiVersion)` - sets the API version to use for the given filter
- `getApiVersion()` - retrieves the set API version

This allows the user to use new GROQ functions, as well as getting faster queries in some cases.

To allow for this, we also expand the `listenQuery` utility to allow specifying API version, and try to re-use studio clients where possible.

The default API version is still `v1` for most lists, since we can't guarantee that the previous filter will work with newer API versions. Simply document type lists however will default to `v2021-06-07`, since we can guarantee that those work.

### What to review 

- Does the API version get passed on correctly?
- Does the API align with current APIs?

### Notes for release

**Highlight**: Added ability to specify API version for document list structures by using the new `apiVersion()` method.

### DIFF
--- a/packages/@sanity/base/src/client/versionedClient.ts
+++ b/packages/@sanity/base/src/client/versionedClient.ts
@@ -7,6 +7,7 @@
  * Preconfigured, versioned client used for all the
  * internal tooling within @sanity/base.
  */
+import type {SanityClient} from '@sanity/client'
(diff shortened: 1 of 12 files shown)


bjoerge (approved):
LGTM!

PR #2793, opened 2021-09-20
refactor(base): move logos to __legacy directory

sjelfull:
### Description

- The logos in `@sanity/base` have been moved to `packages/@sanity/base/src/__legacy/components/logos` as they are superseded by `@sanity/logos` now.
- The `sanity.json` file has been updated so e.g. `part:@sanity/base/sanity-logo` points to `__legacy/@sanity/components/logos/SanityLogo` instead of `components/logos/SanityLogo`.

### What to review

- That the part-imports of logos still work as expected.

### Notes for release

- n/a


RitaDias:
I've looked through the app and everything seems to be running smoothly, though I wasn't able to find any specific place where this would be breaking if the paths were wrong (that's why I'm not giving a green for now).

sjelfull:
@RitaDias 
- `part:@sanity/base/sanity-logo-icon` is used in the [portable text toolbar](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/form-builder/src/inputs/PortableText/Toolbar/helpers.tsx#L49) as a fallback icon

The others I don't think is used anywhere in the monorepo (the loading screen got the logo inlined).

RitaDias:
> * portable text toolbar

Thank you 👍 
I just double checked (went to studio test, PT Documents) and it looks all good 👍

RitaDias (approved):

PR #2795, opened 2021-09-20
test(validation): add infer schema mailto test

ricokahler:
### Description

There was an open question around some of the mailto behavior so here is a quick test to describe and preserve that behavior

### What to review

The test itself


ricokahler:
i think there's another bug actually 😅

PR #2797, opened 2021-09-21
fix(base): make sure the old `styles` properly for previews still work

sjelfull:
### Description

This adds back support for passing a `styles` prop to the `DefaultPreview` component. These styles isn't actually passed anywhere in the monorepo, but for someone using the component directly we still want to support it.

Note: Some styles values like `inner` doesn't make sense anymore since we removed the corresponding element.

### What to review

That the `DefaultPreview` component renders with the passed class names.

### Notes for release

n/a


mariuslundgard (approved):
LGTM!

PR #2802, opened 2021-09-23
Clean up hook dependencies and props

skogsmaskin:


skogsmaskin:
There is an issue with block extras not refreshing when toggling fullscreen.

mariuslundgard (approved):
Nice work! We should hold on merging this until we have gotten the `default-layout` and `desk-tool` release out, tho!

PR #2803, opened 2021-09-24
test: unify test config

ricokahler:
### Description

Unifies more of the testing config to make it easier to write tests! (Shoutout to @bjoerge for the initial work unifying this 🙌)

Specially, this PR includes a naive-but-still-useful parts aliasing mechanism.

### What to review

See the logic in the `jest.config.base.js`


.module-aliases.js
```diff
@@ -3,10 +3,12 @@
 //
 // this file is currently read by:
 // - webpack when running the dev server
+// - vite config for the workshop
 // - jest when running test suite
-// note: order matters. The most specific paths should appear first
```
bjoerge, line 7: order doesn't matter anymore?

jest.config.base.js
```diff
@@ -1,51 +1,147 @@
 const path = require('path')
```
bjoerge, line 1: maybe the file name of this one should reflect that it's no longer exporting a jest config, but a config creator? Rename to `createJestConfig.ts` or something?

jest-setup.js
```diff
@@ -0,0 +1,21 @@
+const path = require('path')
```
bjoerge, line 1: maybe this could live in the `test` directory as well?

bjoerge (approved):
LGTM! Left some minor comments inline.

.module-aliases.js
ricokahler: yeah i guess it doesn't. nice catch!

PR #2808, opened 2021-09-28
test: desk tool coverage

ricokahler:
### Description

The following PR ups the coverage of desk-tool + adds some type script improvements for documentation.

### What to review

Review the new test files:
- `DeskTool.test.tsx` which tests and documents the DeskTool behavior concerns pane emissions over time
- `helpers.test.ts` which tests and documents the current serialization/deserialization of params. Note that I tested a few scenarios that did not meet what I was expecting so i `.skip`ed them (cc @rexxars)

Review the new type changes, ensure all the terminology makes sense etc.


packages/@sanity/state-router/src/useRouterState.ts
```diff
@@ -1,15 +1,26 @@
-import {useContext, useEffect, useState} from 'react'
+import {useContext, useEffect, useState, useRef} from 'react'
 import {RouterState} from './components/types'
 import {RouterContext} from './RouterContext'
 
 export const useRouterState = (): RouterState | null => {
   const {channel, getState} = useContext(RouterContext)
   const [routerState, setState] = useState<RouterState>(getState())
 
+  // prevents "Can't perform a React state update on an unmounted component."
```
bjoerge, line 9: Do you know why this happened? Can't see any obvious reason this should not work with the current code...

packages/@sanity/desk-tool/src/DeskTool.test.tsx
```diff
@@ -0,0 +1,549 @@
+/* eslint-disable max-nested-callbacks */
+/// <reference types="@sanity/types/parts" />
+
+import {render, act} from '@testing-library/react'
+import React, {useEffect, useMemo, useState} from 'react'
+import {Subject, of} from 'rxjs'
+import {useRouter} from '@sanity/base/router'
+import * as ops from 'rxjs/operators'
```
bjoerge, line 8: wasn't clear to me when reading the code in this file that `ops` was all the rxjs operators. By established convention elsewhere we import the individual operators as we need them. Any reason to not do that here? (imo it would also be more readable if it was named `operators` instead of `ops`)

bjoerge (approved):
Great work! LGTM - left a couple of minor suggesitions/comments/questions

packages/@sanity/desk-tool/src/DeskTool.test.tsx
ricokahler: In the test file i found it pretty annoying to write imports for the different operators while deving and these test files won't ever be shipped to users so we don't have worry about tree shaking etc. maybe i just need to get better tooling for auto imports haha

packages/@sanity/state-router/src/useRouterState.ts
ricokahler: it's possible that this could happen because there is a `setState` in the clean up portion of the original `useEffect`. I'm unsure it's possible that this could occur in app but in the tests definitely spin up and tear down multiple times

packages/@sanity/desk-tool/src/DeskTool.test.tsx
```diff
@@ -0,0 +1,551 @@
+/* eslint-disable max-nested-callbacks */
+/// <reference types="@sanity/types/parts" />
+
+import * as Rx from 'rxjs'
+import * as operators from 'rxjs/operators'
+import {render, act} from '@testing-library/react'
+import React, {useEffect, useMemo, useState} from 'react'
+import {useRouter} from '@sanity/base/router'
+import {RouterProvider} from '@sanity/state-router/components'
```
mariuslundgard, line 9: Should `RouterProvider` be imported from `@sanity/base` here? Then we don’t have to depend on state-router in desk-tool (and have more parity, I believe).

```suggestion
import {RouterProvider, useRouter} from '@sanity/base/router'
```

packages/@sanity/desk-tool/package.json
```diff
@@ -57,7 +57,9 @@
   },
   "devDependencies": {
     "@sanity/base": "2.20.0",
+    "@sanity/state-router": "2.20.0",
```
mariuslundgard, line 60: I believe we can get rid of this import 🤔

PR #2809, opened 2021-09-28
fix(desk-tool): fix line when withinSelection

RitaDias:
### Description

Fix issue where the line between timeline items looked wrong when within two separate selections

| before | after |  
|--------|-------|
| image | image

### What to review

Make sure that the colours of the lines work with the different cases.
You can test it out on the test-studio on the document-actions

### Notes for release

Fix visual issue with the connecting line between timeline items.


vicbergquist (approved):
lgtm! 👍

PR #2812, opened 2021-09-28
refactor(structure): replace icons with icons from `@sanity/icons`

sjelfull:
### Description

This migrates the icons used in @sanity/structure to use `@sanity/icons`.

This also caused a series of linting/type issues in structure because the types currently use legacy types like `Function` that isn't allowed in newer lint rules. The test snapshots does not match either since the icons is wrapped in `React.forwardRef`.

I have worked around/fixed this by:
- mapped a custom type `FixMe` to `Function` with a comment explaining why we lint ignore it and replaced all instances with the new `FixMe` type
- updated types where I think it made sense, like replacing `{}` with `Record<string, unknown>`, replacing `any` with `unknown` etc.
- updating the test snapshots.

[sc-11466] [sc-11467]

### What to review

- That all the type fixes looks fine.
- That the test snapshot change is okay

### Notes for release

n/a

### DIFF
--- a/packages/@sanity/structure/src/ChildResolver.ts
+++ b/packages/@sanity/structure/src/ChildResolver.ts
@@ -1,4 +1,5 @@
 import {CollectionBuilder, Collection, SerializePath, SerializeOptions} from './StructureNodes'
+import {FixMe} from './types'
 
 export interface ChildResolverOptions {
   index: number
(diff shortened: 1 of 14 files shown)


packages/@sanity/structure/src/ChildResolver.ts
```diff
@@ -8,10 +9,11 @@ export interface ChildResolverOptions {
   serializeOptions?: SerializeOptions
 }
 
-export type ItemChild = CollectionBuilder | Collection | Function | undefined
+// eslint-disable-next-line @typescript-eslint/ban-types
```
bjoerge, line 12: is this still needed?

bjoerge (approved):
Think there might be one eslint disabled-comment that's no longer needed. Otherwise LGTM!

packages/@sanity/structure/src/ChildResolver.ts
sjelfull: Nope

PR #2815, opened 2021-09-29
fix(state-router): fix disabled intent links

sjelfull:
### Description

This fixes a problem with IntentLinks that is passed the `disabled` prop or in other ways isn't "ready", as with the link in the reference input.

image

Before this fix you could click the intent link, and it just sent you to a non-existing url `/intent/edit//`, which in turn gave you an error screen.

### What to review

- Compare behaviour of the intent link on the reference input before and after the fix is applied
- Does a `span` for a "disabled" link make sense?

### Notes for release

- Fixed JS error when clicking intent links in empty reference inputs 

[sc-11533]


RitaDias (approved):
Someone else should probably get some eyeballs on this as well but 👍

sjelfull:
Closing in favour of another approach to this by @RitaDias

PR #2820, opened 2021-09-30
fix(form-builder): add fix for reference input when it's disabled

RitaDias:
### Description

This fixes a problem with IntentLinks that is passed the disabled prop or in other ways isn't "ready", as with the link in the reference input.

image

Before this fix you could click the intent link, and it just sent you to a non-existing url `/intent/edit//`, which in turn gave you an error screen.

### What to review

- Make sure that you cannot click on the link (no navigation) when the reference input is empty
- Make sure that the title anchor (even when empty) shows the right text (should be empty when the anchor is empty)

### Notes for release

Fixed JS error when clicking intent links in empty reference inputs

### DIFF
--- a/packages/@sanity/base/src/components/IntentButton.tsx
+++ b/packages/@sanity/base/src/components/IntentButton.tsx
@@ -11,5 +11,5 @@ export function IntentButton(
     params: Record<string, string>
   }
 ) {
-  return props.disabled ? <Button {...props} as={'a'} /> : <Button {...props} as={IntentLink} />
+  return props.disabled ? <Button {...props} as="a" /> : <Button {...props} as={IntentLink} />
 }
--- a/packages/@sanity/base/src/components/IntentButton.tsx
+++ b/packages/@sanity/base/src/components/IntentButton.tsx
@@ -11,5 +11,5 @@ export function IntentButton(
     params: Record<string, string>
   }
 ) {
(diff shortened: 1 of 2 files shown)


packages/@sanity/base/src/components/IntentButton.tsx
```diff
@@ -11,5 +11,5 @@ export function IntentButton(
     params: Record<string, string>
   }
 ) {
-  return <Button {...props} as={IntentLink} />
+  return props.disabled ? <Button {...props} as="a" /> : <Button {...props} as={IntentLink} />
```
sjelfull, line 14: A link without `href` doesn't have an implicit `link` role, so I suggest we add these attributes:

```suggestion
  return props.disabled ? <Button {...props} as="a" role="link" aria-disabled="true" /> : <Button {...props} as={IntentLink} />
```

[Source](https://www.scottohara.me/blog/2021/05/28/disabled-links.html)

sjelfull (changes requested):

packages/@sanity/base/src/components/IntentButton.tsx
RitaDias: Big brain! 🧠

sjelfull (approved):

bjoerge (approved):

PR #2826, opened 2021-10-04
Fix issue with readOnly state in the PTE

skogsmaskin:
### Description

After #2814 it became an issue in the editor with the readOnly state of things. We need to make sure the editor is re-rendered whenever readOnly changes.

This will be dealt with more gracefully in another branch in a PR at a later time.

Also I have discovered a prop that is not memoized and triggers unnecessary re-renders. This is now fixed in it's own commit.

### What to review

That you can edit Portable Text after reloading the page with an PT editor on that page.

### Notes for release
Not needed as the problem is solely in `next`.


mariuslundgard (approved):
Yes, so good to figure this out!

PR #2831, opened 2021-10-05
feat(core): add support for groq-powered webhooks

rexxars:
### Description

This PR adds support for the `sanity hook` commands to work with GROQ-powered webhooks. This is primarily done by upgrading the API version for the list commands, and replacing the creation logic with opening a browser on the management interface (in a later revision we'll put in the work to make the creation flow work within the CLI too, but there's a bit too many variables to fit within scope right now)

### What to review

- Does the code make sense?
- Do the commands work as expected? (`npm run build && cd examples/test-studio && ../../packages/@sanity/cli/bin/sanity hook create` et al)

### Notes for release

- Added support for GROQ-powered webhooks in the `sanity hook` CLI commands


kmelve (approved):
Awesome work! 

I found some usability issues, but I wonder if they're outside the scope of this PR, since they were there before as well.

rexxars:
> I found some usability issues, but I wonder if they're outside the scope of this PR, since they were there before as well.

Let's revisit them when we take another pass at this with creation flow etc.

PR #2867, opened 2021-10-15
feat(base): improve change indicators

mikolajdobrucki:
### Description

A small redesign of change indicators (yellow lines suggesting that input has changed since last published)

- indicators are now moved outside of the inputs, 2px to the right
- the hover effect is much simpler now, icon and tooltip are removed
- indicators (and therefore connectors) are always yellow, don't turn blue when a field is focused
- indicators are not recognised when navigating a form with a keyboard (when you tab from a field, the focus jumps directly to the next input, ignoring the indicator. it makes it inaccessible for keyboard users but this is acceptable as we have other ways of opening review changes that keyboard users can exploit. whereas this behaviour is annoying and breaks patterns regarding what is usually expected from web forms)
- spacing between array items in grid layout is increased to accommodate new indicators

### What to review

only review changes related elements should be affected.
it's a plainly visual change, the behaviour of change indicators, connectors and review changes should not change.

### Notes for release

* updates change indicators and connectors in Review Changes

### DIFF
--- a/packages/@sanity/base/src/change-indicators/ChangeBar.styled.tsx
+++ b/packages/@sanity/base/src/change-indicators/ChangeBar.styled.tsx
@@ -1,7 +1,5 @@
-import React from 'react'
 import styled, {css} from 'styled-components'
 import {Theme} from '@sanity/ui'
-import {EditIcon} from '@sanity/icons'
 
 interface ThemeContext {
   theme: Theme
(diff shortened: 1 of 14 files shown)


mikolajdobrucki:
There is an issue that appeared after the PT migration we need to fix before merging this

image

mikolajdobrucki:
> There is an issue that appeared after the PT migration we need to fix before merging this

We agreed with Marius that solving this properly might require larger changes to the layout of PT Editor. As this issue is blocking the release of References in Place I prepared a solution that is not ideal but solves the bug without introducing any design changes to PT. 

Solution: the indicators are now moved inside the white container of the document so the button that opens Review Changes doesn't cause overflow issues.

@mariuslundgard @bjoerge would you mind reviewing again

bjoerge (approved):

mariuslundgard (approved):
Sorry for the delay – this looks good to go! 👍

PR #2895, opened 2021-10-29
feat: object `modal.type` and `modal.width` option in PTE

mariuslundgard:
### Description

With this change, the `object` schema type takes a new option called `modal`:

```ts
interface ModalOption {
  type?: 'popover' | 'dialog'
  width?: 'small' | 'medium' | 'large' | 'full'
}
```

This means you can set a custom modal width for e.g. link annotation objects in PTE like so:

```ts
{
  type: 'block',
  marks: {
    annotations: [
      {
        type: 'object',
        name: 'link',
        title: 'Link',
        fields: [
          // ...
        ],
        options: {
          modal: {
            width: 'medium' // defaults to 'small' for annotation objects
          }
        }
      }
    ],
  },
}
```

With this change, the `editModal` option should be considered legacy (it was problematic because it did not allow defining a custom width) – and using it will print a warning in the console.

```
Warning: The "editModal" option will be deprecated. Please use `options.modal.type` instead.
```

### What to review

Try out the new `option.modal` in the Test Studio and make sure it behaves in the expected manner.

The `simpleBlock` document type in the Test Studio is configured to use it.

### Notes for release

Introducing a new `modal` option for objects in PTE, which allows for defining a custom modal `width`.


dev/test-studio/schema/standard/portableText/simpleBlock.js
```diff
@@ -69,7 +75,15 @@ export default {
             spellCheck: true,
           },
         },
-        {type: 'image', name: 'image'},
+        {
+          type: 'image',
+          name: 'image',
+          options: {
+            modal: {
+              // width: 'small',
```
bjoerge, line 83: remove or add a note about why it's still here / when it can be removed?

packages/@sanity/form-builder/src/inputs/PortableText/__workshop__/customSchema/schema.ts
```diff
@@ -18,11 +19,26 @@ const linkType = {
...
+  },
+  fields: [{type: 'string', name: 'title'}],
+}
+
 const someObject = {
   type: 'object',
   name: 'someObject',
   options: {
-    editModal: 'fullscreen',
+    // editModal: 'fullscreen',
```
bjoerge, line 37: remove or add a note about why it's still here / when it can be removed?

bjoerge (approved):
Nice! Left a couple comments about comments, otherwise good to merge.

dev/test-studio/schema/standard/portableText/simpleBlock.js
mariuslundgard: This is Test Studio only, thought it would be good to leave it as documentation. But will add comments 👍

PR #2900, opened 2021-11-01
fix(desk-tool/base): update permission calls

RitaDias:
### Description

As part of a list of bugs regarding roles, it became clear that we would need to refactor our permission calls to not only having to rely on the roles of the user but also on the initialValues from the documents and schema. This was detected when a user with permissions only based on document properties (`requiresApproval`) were not returning the right permissions. 

To reduce the scope of this task, only a few methods were refactored (where the initialValues are not resolved in the `documentPair` (new) / `highlevel` (old) methods). The plan is to eventually move all of the methods to the new `documentPair` file. But for now, to keep this scope small and the potential impact reduced, only the needed methods were replaced: 

- `create` call in the `desk-tool`
- `update` call in the `desk-tool`
- `publish` call in the `PublishAction`
- `duplicate` call in the `DuplicateAction`

All other calls are using the old calls.
The `create` call outside of the `desk-tool` (in the NavBar) will be updated in a different pr.

### What to review

Check the use cases in the tests.

#### Update
- [x] as an administrator user, the permission banner will not be visible (has permissions)
- [x] as a restricted user, the permission banner will be visible (does not have permission)
- [x] as a requiresApproval user, the permission banner will be not visible (does have permission) when updating approved published documents
- [x] as a requiresApproval user, the permission banner will be visible (does not have permission) on non-approved published documents
- [x] as a requiresApproval user, the permission banner will be not visible (does have permission) when updating draft documents

#### Create
- [x] as an administrator user, the permission banner will not be visible (has permissions)
- [x] as a requiresApproval user, the permission banner will be not visible (does have permission) on specific document types

#### Publish
- [x] as an administrator user, the publish button will be active
- [x] as an restricted user, the publish button will be disabled
- [x] as a requiresApproval user on an approved published document, the publish button will be active
- [x] as a requiresApproval user on a non-approved document, the publish button will be disabled

#### Duplicate
- [x] as an administrator user, the duplicate button will be active
- [x] as an restricted user, the duplicate button will be disabled
- [x] as a requiresApproval user on an approved published document, the duplicate button will be active
- [x] as a requiresApproval user on a non-approved document, the duplicate button will be disabled

Also added tests for the permissions that haven't been changed (as a safety net for next time)
- `Delete`
- `Unpublish`
- `DiscardDraft`

### Notes for release

Fix roles permissions on document based on document permissions

### DIFF
--- a/cypress/integration/desk-tool/actionPermission.test.js
+++ b/cypress/integration/desk-tool/actionPermission.test.js
@@ -86,3 +86,41 @@ describe('@sanity/desk-tool: duplicate permission', () => {
     cy.get('[data-testid=action-Duplicate]').should('have.attr', 'disabled')
   })
 })
+
+describe('@sanity/desk-tool: delete permission', () => {
+  it('as an administrator user, the duplicate button will be active', () => {
+    cy.visit(
+      '/test/desk/input-standard;booleansTest;1053af2f-84af-49a1-b42b-e2156470bb77%2Ctemplate%3DbooleansTest#_debug_roles=administrator'
+    )
+
+    cy.get('[data-testid=action-menu-button').click()
+    cy.get('[data-testid=action-Delete]').should('not.have.attr', 'disabled')
+  })
+
+  it('as an restricted user, the duplicate button will be disabled', () => {
+    cy.visit(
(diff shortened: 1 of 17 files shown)


RitaDias:
Adding this: There are a couple of files/additions that I've added that will be reviewed in the PR #2872. So that one should be reviewed before this (so I can rebase)

RitaDias:
Closing because this is included in #2939

PR #2902, opened 2021-11-01
fix(vision): fix min and max pane sizes so controls or results is never cut off

sjelfull:
### Description

This improves the pane resizing by setting better min/max boundaries for all panes (current version of library doesn't support percentages).

It also fixes a padding issue that prevented users from getting to the bottom of results.

### Video that shows the new behaviour

https://user-images.githubusercontent.com/10508/150348812-7b88d6b5-0745-46aa-9b17-a879bedf1c0a.mp4

**Results**
image

### What to review

- That panes resizes well
- That min and max size for panes looks good
- That results bottom is not cut off
- That disabling resize when container is lower than 550px behaves fine

### Notes for release

- Improves the panes in Vision so controls or result never is cut off when there isn't enough room

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.js
+++ b/packages/@sanity/vision/src/components/VisionGui.js
@@ -509,9 +509,15 @@ class VisionGui extends React.PureComponent {
           </Grid>
         </Header>
         <SplitpaneContainer flex={1}>
-          <SplitPane split="vertical" minSize={150} defaultSize={400}>
+          <SplitPane split="vertical" minSize={280} defaultSize={400} maxSize={-400}>
             <Box height="stretch" flex={1}>
-              <SplitPane split="horizontal" defaultSize={'50%'}>
+              <SplitPane
+                split="horizontal"
+                defaultSize={'50%'}
+                minSize={160}
+                maxSize={-300}
+                primary="second"
+              >
                 <InputContainer display="flex" ref={this._queryEditorContainer}>
                   <Card flex={1}>
                     <InputBackgroundContainerLeft>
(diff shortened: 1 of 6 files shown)


packages/@sanity/vision/src/components/VisionGui.js
```diff
@@ -509,14 +498,20 @@ class VisionGui extends React.PureComponent {
           </Grid>
         </Header>
         <SplitpaneContainer flex={1}>
-          <SplitPane split="vertical" minSize={150} defaultSize={400}>
+          <SplitPane split="vertical" minSize={280} defaultSize={400} maxSize={-400}>
```
mariuslundgard, line 501: I don’t understand how a negative number would work here. Maybe add a comment?

packages/@sanity/vision/src/components/VisionGui.js
```diff
@@ -509,14 +498,20 @@ class VisionGui extends React.PureComponent {
...
             <Box height="stretch" flex={1}>
-              <SplitPane split="horizontal" defaultSize={'50%'}>
+              <SplitPane
+                split="horizontal"
+                defaultSize={'50%'}
+                minSize={160}
+                maxSize={-300}
```
mariuslundgard, line 507: Same here with the negative number `-300`.

mariuslundgard (approved):
LGTM! Only have a question about `maxSize` :-)

packages/@sanity/vision/src/components/VisionGui.js
sjelfull: I don't either, the syntax is kind of weird 🤡  I just tweaked it till it looked alright. Will add a comment though

sjelfull:
@mariuslundgard I updated the PR to do a bunch of calculations to set all the `*size*` props in pixels instead of using the funky `-n` syntax. It also will disable resizing when the container height is below 550, since its becomes pretty crowded then.

PR #2905, opened 2021-11-03
chore(deps): upgrade dependencies

rexxars:
### Description

- Upgrades ora and strip-ansi to latest non-esm versions to fix security vulnerabilities

Supersedes #2766 and #2767

### What to review

That the code runs and works as expected.

`ora` is used in the CLI for all spinners, and `strip-ansi` is used for hot module reloading in development mode of the server.

### Notes for release

None, transparent to the end-user.


ricokahler (approved):
LGTM! i double checked the notes tagged releases for these too and didn't see anything relevant

PR #2906, opened 2021-11-03
fix(core): resolve `createRequireContext` relative to core

ricokahler:
### Description

Internal GraphQL deploys broke with #2903 because trying to `require('@sanity/core/_internal')` would not resolve in the local packages symlinked with lerna.

To fix that, this PR:
1. runs `require.resolve` to get an absolute path and then injects that path into the script so it's always imported relative to `requireContext.ts`
2. moves the `createRequireContext` helper into the `@sanity/util` package, which is already a dependency of `@sanity/core`. this makes is so that it can be imported inside of internal repos symlinked with lerna.

### What to review

- Try deploying GraphQL of an internal example repo
- Try packing `@sanity/core` and `@sanity/util` and running the GraphQL repo

### Notes for release

N/A


rexxars (approved):
Looks good. Let's schedule `require.context` for removal with v3, if not sooner.

PR #2907, opened 2021-11-04
wip

vicbergquist:


packages/@sanity/form-builder/src/FormBuilderInput.tsx
```diff
@@ -220,11 +227,10 @@ export class FormBuilderInput extends React.Component<FormBuilderInputProps> {
 
     if (typeof readOnly === 'function' || typeof type.readOnly === 'function') {
       return (
-        <PropertyCallbackField
+        <ConditionalReadOnlyField
           parent={parent}
           value={value}
-          checkProperty={readOnly ?? type.readOnly}
-          checkPropertyKey="readOnly"
+          readOnly={readOnly || type.readOnly}
```
sjelfull, line 233: I think this needs to use `??` instead of `||` to work correctly?

sjelfull (approved):
Good to merge if the comment doesn't need to be changed

PR #2908, opened 2021-11-04
[structure] Allow a component as title

EECOLOR:
### Description

Allow a component as a title of list item.

Instead of writing this:

```js
S.list().title('Testies').items([
  {
    type: 'listItem',
    id: ...,
    title: <MyComponent ... />,
    child: ...,
  }
])
```

I want to write this:

```js
S.list().title('Testies').items([
  S.listItem().id(...).title(<MyComponent ... />).child(...)
])
```

### What to review

This change reflects the actual situation more precisely. Instead of stating that a `title` is required when it is not a `string`, it will state that an `id` is requied if the `title` is not a string.

I have added tests to reflect the behavior.

### Notes for release

`S.listItem().title(...)` now allows components and React nodes.


rexxars:
We'll revisit this feature request as part of a slightly bigger scoped project - thanks for raising this!

PR #2909, opened 2021-11-04
docs(core): fix typo in help text and error message

kmelve:
### Description

`--api-version` didn't work because the code looks for `--apiVersion`. I also wonder if we shouldn't support setting it as an environment variable or something? Having the warning appear just because you didn't care to specify it, is pretty annoying.

### What to review

Check if
```sh
sanity documents query --apiVersion 2021-11-03 '*[]' 
```
works.

### Notes for release

Fixes help text and warning for `sanity documents query` when the API version isn't set

### DIFF
--- a/packages/@sanity/core/src/commands/documents/queryDocumentsCommand.ts
+++ b/packages/@sanity/core/src/commands/documents/queryDocumentsCommand.ts
@@ -7,7 +7,7 @@ Run a query against the projects configured dataset
 Options
   --pretty colorized JSON output
   --dataset NAME to override dataset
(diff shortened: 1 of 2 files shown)


rexxars (approved):
Flags should generally be dash-cased (or support both), so I've updated the PR with reparsing of the CLI arguments, enabling it to work with both `--api-version` and `--apiVersion`. I also added an `SANITY_CLI_QUERY_API_VERSION` environment variable that can be set and used as the default when querying. Please let me know what you think.

(I also found a bug where if your query returned just a string, the `--pretty` flag would crash when attempting to colorize the output)

PR #2914, opened 2021-11-07
fix: intent resolver follow up

ricokahler:
### Description

This is a small follow up to the new (unreleased as of now) intent resolver. This PR:

- limits default resolution to the default filter (e.g. `_type == $type`). This is to prevent false positive matches where the user may not want an intent match.
- refactors the breadth depth search into a concurrent search using `Promise.all` for structure that use promises

### What to review

See the changes to `resolveIntent.ts` and that all intents resolve as expected.

### Notes for release

N/A

### DIFF
--- a/packages/@sanity/desk-tool/src/getIntentState.ts
+++ b/packages/@sanity/desk-tool/src/getIntentState.ts
@@ -40,8 +40,12 @@ export function getIntentState(
         index: i,
       }) ||
       // OR
-      // the pane is a documentList with a matching schemaTypeName
-      (pane.type === 'documentList' && pane.schemaTypeName === params.type)
+      // if the resolved pane is a document list
+      (pane.type === 'documentList' &&
+        // and the schema type matches
+        pane.schemaTypeName === params.type &&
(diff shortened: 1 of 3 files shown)


mariuslundgard (approved):
Looks sane to me 👍

packages/@sanity/desk-tool/src/getIntentState.ts
```diff
@@ -40,8 +40,12 @@ export function getIntentState(
         index: i,
       }) ||
       // OR
-      // the pane is a documentList with a matching schemaTypeName
-      (pane.type === 'documentList' && pane.schemaTypeName === params.type)
+      // if the resolved pane is a document list
+      (pane.type === 'documentList' &&
+        // and the schema type matches
+        pane.schemaTypeName === params.type &&
+        // and the filter is the default filter
```
ricokahler, line 47: note from espen review:

add comment for rationale: want to have decent default matches but don't want to match too heavily. aka false positives etc

if the user wants the match, they can implement canHandleIntent themselves

PR #2917, opened 2021-11-08
fix(desk-tool): use `getDefaultDocumentNode` in fallback editor

ricokahler:
### Description

This PR makes the fallback editor utilize `getDefaultDocumentNode` so that configured views in there will still show up when the intent can't be resolved.

### What to review

Implement getDefaultDocumentNode and then use an intent link that resolves to the fallback editor. Views configured there should be visible in the fallback editor now.

### Notes for release

- Fixes an issue where document views would not appear in certain cases when implementing `getDefaultDocumentNode`

### DIFF
--- a/packages/@sanity/desk-tool/src/DeskTool.test.tsx
+++ b/packages/@sanity/desk-tool/src/DeskTool.test.tsx
@@ -64,11 +64,15 @@ jest.mock('part:@sanity/desk-tool/structure?', () => {
     child: mockChild,
   }))
 
+  const {StructureBuilder} = jest.requireActual('@sanity/structure')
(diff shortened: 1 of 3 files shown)


packages/@sanity/desk-tool/src/DeskTool.test.tsx
```diff
@@ -436,91 +445,11 @@ describe('DeskTool', () => {
           },
           type: 'document',
         })
-      })
-    })
-
-    it('passes the correct arguments to the structure child/items', async () => {
```
ricokahler, line 442: note: i deleted this test because it already has some weird outputs and i've fix those in #2916.

ricokahler:
I reviewed with one with espen in a call

PR #2926, opened 2021-11-11
fix(conditional-fields): resolve readOnly in changes panel

vicbergquist:
- Fixes a problem where the form view would not know that a whole document is read-only and it also accepts a callback, so now the resolved readOnly value is passed down through the document
- Updates the context the changes panel uses to include the document value so that FieldChange and GroupChange can resolve their readOnly callbacks if any.
- Updates the change components to resolve the hidden property if it's a function

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
+++ b/packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
@@ -1,7 +1,7 @@
 // @todo: remove the following line when part imports has been removed from this file
 ///<reference types="@sanity/types/parts" />
 
-import {useDocumentPresence} from '@sanity/base/hooks'
+import {useDocumentPresence, useResolveConditionalProperty} from '@sanity/base/hooks'
 import {PresenceOverlay} from '@sanity/base/presence'
 import {Box, Container, Flex, Spinner, Text} from '@sanity/ui'
 import afterEditorComponents from 'all:part:@sanity/desk-tool/after-editor-component'
@@ -12,6 +12,7 @@ import filterFieldFn$ from 'part:@sanity/desk-tool/filter-fields-fn?'
 import {FormBuilder} from 'part:@sanity/form-builder'
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import {tap} from 'rxjs/operators'
+import {SanityDocument} from '@sanity/client'
 import {useDocumentPane} from '../../useDocumentPane'
 import {Delay} from '../../../../components/Delay'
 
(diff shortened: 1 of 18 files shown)


packages/@sanity/types/src/schema/types.ts
```diff
@@ -84,7 +84,7 @@ export interface BaseSchemaType {
   description?: string
   type?: SchemaType
   liveEdit?: boolean
-  readOnly?: ConditionalProperty
+  readOnly?: ConditionalProperty | undefined
```
sjelfull, line 87: This is typed to be optional, does it need the extra `undefined` at the end?

packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
```diff
@@ -60,15 +61,23 @@ export function FormView(props: FormViewProps) {
...
+
   const readOnly = useMemo(() => {
     return (
       !ready ||
       rev !== null ||
       !permission.granted ||
       !isActionEnabled(documentSchema, 'update') ||
-      (isNonExistent && !isActionEnabled(documentSchema, 'create'))
+      (isNonExistent && !isActionEnabled(documentSchema, 'create')) ||
+      resolvedReadOnly
```
sjelfull, line 78: I don't see why we shouldn't, but are we sure we want to add `readOnly` support on the document level since it wasn't supported before?

sjelfull (approved):
Looks good to me. Left a few comments.

packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
vicbergquist: It was supported, it just didn't show visually that they become readOnly, apparently

packages/@sanity/types/src/schema/types.ts
vicbergquist: true

packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
sjelfull: Ah, you're right! We haven't documented it here either https://www.sanity.io/docs/document-type

vicbergquist: yeah! will let martin know

packages/@sanity/field/src/diff/components/FieldChange.tsx
```diff
@@ -114,4 +116,44 @@ export function FieldChange({
       </FieldWrapper>
     </Stack>
   )
+
+  const hidden = change.schemaType.hidden
+
```
sjelfull, line 121: Should we add the `checkProperty` below here: `const readOnly = readOnly || change?.parentSchema?.readOnly || change.schemaType.readOnly`

..and then use it below when doing the type check + passing it to `ConditionalPropertyResolver`?

vicbergquist: yeah, already on it 👍

PR #2927, opened 2021-11-11
fix(form-builder): update PTE tooltips

hermanwikner:

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObject.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObject.tsx
@@ -250,47 +250,44 @@ export function BlockObject(props: BlockObjectProps) {
   const hasErrors = errorMarkers.length > 0
   const hasWarnings = warningMarkers.length > 0
 
-  const markersToolTip = useMemo(
-    () =>
-      hasErrors || hasWarnings || (hasMarkers && renderCustomMarkers) ? (
-        <Tooltip
-          placement="top"
-          portal="editor"
-          content={
-            <TooltipBox padding={2}>
-              <Markers markers={markers} renderCustomMarkers={renderCustomMarkers} />
-            </TooltipBox>
-          }
-        >
-          <div>{blockPreview}</div>
-        </Tooltip>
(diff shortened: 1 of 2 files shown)


packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObject.tsx
```diff
@@ -250,47 +250,47 @@ export function BlockObject(props: BlockObjectProps) {
...
-  )
-
   const isImagePreview = type?.type?.name === 'image'
 
   const blockPath = useMemo(() => [{_key: block._key}], [block._key])
 
+  const tooltipEnabled = useMemo(
+    () => hasErrors || hasWarnings || (hasMarkers && renderCustomMarkers),
+    [hasErrors, hasMarkers, hasWarnings, renderCustomMarkers]
+  )
```
mariuslundgard, line 260: Since [booleans are immutable](https://developer.mozilla.org/en-US/docs/Glossary/Primitive), the `useMemo` is not necessary.
```suggestion
  const tooltipEnabled = hasErrors || hasWarnings || Boolean(hasMarkers && renderCustomMarkers)
```

packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -349,45 +349,45 @@ export function TextBlock(props: TextBlockProps): React.ReactElement {
...
-          }
-        >
-          {text}
-        </Tooltip>
-      ) : undefined,
-    [blockMarkers, blockRef, hasErrors, hasWarning, hasMarkers, renderCustomMarkers, text]
+  const tooltipEnabled = useMemo(
+    () => hasErrors || hasWarning || (hasMarkers && renderCustomMarkers),
+    [hasErrors, hasMarkers, hasWarning, renderCustomMarkers]
   )
```
mariuslundgard, line 355: (Same as above)

mariuslundgard (approved):
Looks good to me! I’ve commented on unnecessary use of `useMemo`, but it’s not critical to change now.

PR #2929, opened 2021-11-12
refactor(form-builder): simplify block components

mariuslundgard:
### Description

This refactors `TextBlock` and `Decorator`:

- Removes unused `data-*` attributes.
- Uses `data-*` attributes instead of conditional CSS.
- Moves styles to a separate file – `TextBlock.styles.ts`.
- Moves constants to a separate file – `constants.ts`.
- Moves helpers to a separate file – `helpers.ts`.
- Replaces `switch` statements with constant config objects.
- Uses CSS variables to colorize marker backgrounds.

And `BlockObject`:

- Add missing bottom padding.

### What to review

- That markers work as they have been.
- That blocks render and function as they have been.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
@@ -119,11 +119,26 @@ export const BlockActionsOuter = styled(Box)`
   }
 `
 
-export const BlockActionsInner = styled(Flex)`
-  position: absolute;
-  right: 0;
-  top: -7px;
-`
+export const BlockActionsInner = styled(Flex)(({theme}: {theme: Theme}) => {
+  const {fonts, space} = theme.sanity
+  const textSize1 = fonts.text.sizes[1]
+  const textSize2 = fonts.text.sizes[2]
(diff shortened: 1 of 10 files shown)


packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
```diff
@@ -0,0 +1,161 @@
...
+  &:before {
+    content: ' ';
+    font-size: 0;
+  }
+`
+
+export const BlockActionsInner = styled(Flex)`
+  position: absolute;
+  right: 0;
+  top: -7px;
```
hermanwikner, line 125: Should this value be calculated with theme values?

hermanwikner:
Looks really good! 👍

hermanwikner:
Looks really good! 👍

packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
mariuslundgard: True, I’ve added a commit to fix this 👍

hermanwikner (approved):

PR #2930, opened 2021-11-13
chore: upgrade puppeteer

ricokahler:
### Description

Upgrades puppeteer for better compat with Node 16 on apple silicon

### What to review

Do the tests still pass?

Also, check the changelog [here](https://github.com/puppeteer/puppeteer/blob/23ee295f348d114617f2a86d0bb792936f413ac5/CHANGELOG.md) for anything that could potentially cause a break. It looks like the breaking changes are very minor.


rexxars:
Could you upgrade the dependency inside of the portable text editor too and make sure it works?

ricokahler:
@rexxars yup will give it a go. I was attempting to de-dupe the puppeteer dep in PTE by deleting it but it looks like the dep checker didn't like that. Testing locally it passed fine though. I'll add it back end and we'll see what the CI says.

rexxars (approved):
👍

PR #2934, opened 2021-11-15
fix(validation): fix nested portable text markDefs paths

ricokahler:
### Description

The following refactors and fixes how the markDefs in portable text are validated. This work continues off of @skogsmaskin's work in #2931.

Previously `markDefs` were validated inside of the `blockValidator` function. This PR moves that logic into the recursive`validateItem` function. This was done in order to fix two pre-existing bugs:

1. incorrect or repeat paths due to the incorrect path context
2. nested validation for markDefs (e.g. a mark def in a mark def)

This PR also:
1. adds the compiled block schema types to `@sanity/types` + type predicates
2. re-routes `Rule.block` calls to `Rule.custom` and deprecates `Rule.block`

### What to review

Try this out in the portable text editor. Does everything work as expected?

Also, review the tests to see if the assertions are correct.

### Notes for release

Fixes nested validation inside of `markDefs` in the portable text editor.


packages/@sanity/types/src/portableText/types.ts
```diff
@@ -1,8 +1,8 @@
-export interface Block<O = Span> {
+export interface Block<TChild = Span> {
   _type: 'block'
```
skogsmaskin, line 2: This `_type: 'block'` should be string as well as we now supports custom block type names.

skogsmaskin (commented):
Great work! I just noticed one thing which I commented on that we should look into.

packages/@sanity/types/src/portableText/types.ts
skogsmaskin: I can confirm that it's not validating properly on hoisted block types.

skogsmaskin: I changed it.

skogsmaskin (approved):
I added some fixes that makes validation work on hoisted (named) block types as well. Please have a look @ricokahler.

ricokahler:
@skogsmaskin ah nice catch! looks good to me. feel free to take over merging etc

PR #2911, opened 2021-11-05
feat(base): internal new-document options API

ricokahler:
### Description

The following PR refactors and lifts the `getNewDocumentModalActions` API into `@sanity/base/_internal` so it can be used outside of the `default-layout`.

### What to review

- the code
- the API updates
- the tests
- see if everything works in the studio

### Notes for release

N/A

PR #2939, opened 2021-11-18
fix: update permissions w/ template + precondition support (#2939)

ricokahler:
### Description

This PR:

1. Introduces `useDocumentPairPermissions` — this is similar to the previous `useCheckDocumentPermissions` except it excludes the `create` action and replaces that with `duplicate`. This API is aware of the draft model and checks preconditions of the grants via a live query.
2. Introduces `useDocumentValuePermissions` — this is similar to the approach @RitaDias had in #2900. This API is _not_ drafts-model aware and can't check pre-conditions like before, however if you have a document value on hand, you pass it to this API and have it check against atomic grants (e.g. `'read' | 'create' | 'update'`)
3. Introduces `useTemplatePermissions` — this API takes in a list of initial template value items and returns a record keyed by the IDs of each initial template value where the values are permissions results alongside other template metadata
4. Introduces `getNewDocumentOptions` — this API lifts the previous `getNewDocumentModalActions` function in `default-layout` and moves it into `base`. This API pulls from the [new-document structure](https://www.sanity.io/docs/initial-value-templates#56b4073ca73a) and returns a list of initial value template items (that could be paired with the above API)
5. Refactors the default-layout's `NavBar` create button and `CreateDocumentDialog`  to use the above APIs. Also deprecates unused components in `base`.
6. Refactors the fields permission aware components to use the new permissions APIs
6. Refactors the desk-tool's actions to use the new permissions APIs
7. Refactors the structure menu items to be less concerned with document creation actions so the downstream desk-tool can pick this up instead. This was done because the create actions need to be permissions aware which is not possible inside of the structure builder.
8. Along with the above, came the refactor to the pane headers. The component `PaneHeaderActions` was introduced to unify all of this logic and handle permissions and intents better. 

This PR is reviewable now but there is still some clean up needed:
- [x] fix the e2e tests (or migrate them to jest)
- [x] improve some test coverage around the new APIs
- [x] fix an issue with `createLoadableHook` (will explain this later)
- [x] double check error-handling

Supersedes #2911 #2872 #2900 

### What to review

- Review the final API for the permissions APIs + new document options
- See that all permissions related affordances work. This includes the navbar new-document menu and all the different types of list panes we have in the test studio (use the `#_debug_roles=restricted` debug scheme for now)
- Try out the new `PaneHeaderActions` implementation by adding menu items with intents and/or different actions + try the `showAsAction()` flag too

### Notes for release

- Improves permissions support. Permissions now respect document preconditions and initial value templates.
- Fixes an issue that caused the document list create buttons to not respect permissions.


packages/@sanity/desk-tool/src/panes/list/ListPane.tsx
```diff
@@ -1,9 +1,9 @@
-import React, {useCallback, useMemo} from 'react'
+import React, {useCallback} from 'react'
```
mariuslundgard, line 1: The `useCallback` is unused here

mariuslundgard:
The `paneHeaderActions` look solid to me!

#### Would prefer to move it to `src/components/paneHeaderActions`

The reason being I want us to think of the `pane/*` components as being agnostic UI components that could be used in pane layout outside of the Desk Tool (at some point). The moment we introduce awareness of _initial value templates_ in the pane components, they are being tied to the desk tool only. I’ve worked quite hard to separate these concerns (ui vs desk-tool specific types/logic) 😅

packages/@sanity/desk-tool/src/components/pane/paneHeaderActions/CreateButton.tsx
```diff
@@ -0,0 +1,127 @@
+import {getTemplateById} from '@sanity/base/initial-value-templates'
+import {ComposeIcon} from '@sanity/icons'
+import {InitialValueTemplateItem} from '@sanity/structure/src'
```
ricokahler, line 3: OPE

packages/@sanity/desk-tool/src/components/pane/paneHeaderActions/PaneHeaderActions.tsx
```diff
@@ -0,0 +1,183 @@
...
+          handler(item.params as Record<string, string>)
+          return true
+        }
+
+        return false
+      },
+      [actionHandlers]
+    )
+
+    const [actionMenuItems, contextMenuItem] = useMemo(() => {
```
ricokahler, line 70: that should be plural lol

mariuslundgard (approved):
✅ on the `PaneHeaderActions` – lots of great improvements here 👍

RitaDias (commented):
Fantastic work 👏 
I left some comments because I was either curious or wanted to understand better, have a look when you have the time!

packages/@sanity/base/src/datastores/grants/templatePermissions.ts
```diff
@@ -0,0 +1,121 @@
...
+            )
+          })
+        )
+      )
+  ).pipe(
+    map((results) =>
+      results.reduce<Record<string, TemplatePermissionsResult>>((acc, next) => {
+        acc[next.id] = next
+        return acc
+      }, {})
```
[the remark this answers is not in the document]
bjoerge: I had the same question as @RitaDias here. Wondering if there are any compelling benefits of it being recod/map as opposed to an array of results? I found myself having to use `Object.keys(result).map(id => (...))` in the consuming end, which seems less convenient than having it as an array in the first place.

packages/@sanity/base/src/_exports/hooks.ts
```diff
@@ -1,12 +1,16 @@
...
-  unstable_useCheckDocumentPermission,
-  // eslint-disable-next-line camelcase
-  unstable_useCanCreateAnyOf,
-} from '../datastores/grants/hooks'
+  unstable_getDocumentPairPermissions,
+  unstable_getDocumentValuePermissions,
+  unstable_getTemplatePermissions,
+  unstable_useDocumentPairPermissions,
+  unstable_useDocumentValuePermissions,
+  unstable_useTemplatePermissions,
```
bjoerge, line 11: could we also export the `TemplatePermissionsResult` type from either here or the `_internal` exports?

bjoerge (approved):
Great work! Left a couple of comments, LGTM otherwise

PR #2944, opened 2021-11-21
chore(portable-text-editor): remove src import

ricokahler:
### Description

I had made this same mistake in my own branch and found this while grepping.

Happens to the best of us ❤️

**Edit:** huh it looks like this change makes the CI fail. was the source import intentional?


skogsmaskin:
Hmm. That's weird. Something funny with the `@types` package's exports?

skogsmaskin (approved):

PR #2946, opened 2021-11-21
fix(desk-tool): refine intent resolver

ricokahler:
### Description

The following PR:
1. adds timeouts and guards to the `resolveIntent` function to prevent it from never resolving (by returning the fallback editor in many cases).
2. moves the extra documentList case (i.e. `pane.type === 'documentList' && pane.schemaTypeName === params.type &&  pane.options.filter === '_type == $type'`) into the default intent checker in structure. This should give users more control so they could opt-out of that behavior.
3. adds the method `disableNestedIntentResolution()` to the structure builder to stop the intent resolver from traversing the list. This will help in the cases where lists are very long or infinitely recursive and we want users to opt out of that behavior.

### What to review

See the new APIs and whether or not they make sense. Try out the new options in structure too.

Since these add to the existing structure, I think makes sense to update docs for this as well. I'll get the process going internally soon.

### Notes for release

- Adds timeouts and fallbacks to the intent resolver
- Adds `S.list.disableNestedIntentResolution()` to disable the intent resolver from searching in that list for intent handlers.


rexxars (approved):
Solid work!

binoy14:
@rexxars is this still needed or it can be closed given it's ~3 years old and it's for v2

PR #2950, opened 2021-11-23
fix(state-router): prevent ID flashing when resolving intents

ricokahler:
### Description

Fixes a small bug that causes URLs to flash while resolving intent IDs.

### What to review

Try creating a new document via intent links and ensure there is no URL flashing.

### Notes for release

Fixes a bug that caused the URL to flash different IDs while resolving create intents.


bjoerge (approved):
Nice one!

PR #2951, opened 2021-11-23
reference updates with permissions

bjoerge:
Adds permission considerations to the recent reference input updates. Also adds a new flag to disable in-place creation

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
+++ b/packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
@@ -1,4 +1,3 @@
-// eslint-disable-next-line camelcase
 import {unstable_useTemplatePermissions, useDocumentType} from '@sanity/base/hooks'
 import {LegacyLayerProvider, useZIndex} from '@sanity/base/components'
 import {ChangeConnectorRoot} from '@sanity/base/change-indicators'
@@ -17,7 +16,6 @@ import {
 import React, {memo, useMemo, useState} from 'react'
 import styled from 'styled-components'
 import {fromString as pathFromString} from '@sanity/util/paths'
-// eslint-disable-next-line camelcase
 import {Unstable_ReferenceInputOptionsProvider} from '@sanity/form-builder/_internal'
 import {Path} from '@sanity/types'
 import {getNewDocumentOptions} from '@sanity/base/_internal'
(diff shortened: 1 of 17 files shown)


packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
```diff
@@ -1,4 +1,5 @@
-import {useDocumentType} from '@sanity/base/hooks'
+// eslint-disable-next-line camelcase
+import {unstable_useTemplatePermissions, useDocumentType} from '@sanity/base/hooks'
```
ricokahler, line 2: i like renaming the import without the unstable so the linter doesn't go off at each new line

packages/@sanity/form-builder/src/inputs/ReferenceInput/types.ts
```diff
@@ -22,6 +22,31 @@ export interface ReferenceInfo {
...
+  type: string
+  template: ReferenceTemplate
+}
+
+export type ReferenceParams = Record<string, string | number | boolean>
+
+export interface CreateOption {
+  id: string
+  title: string
+  icon?: React.ReactNode
```
ricokahler, line 41: I think this can also be a component type too? e.g. `icon: RocketIcon` and `icon: <RocketIcon />` would both allowed

ricokahler (approved):
looks great! nothing blocking feel free to merge

packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
bjoerge: I'm a bit torn on removing the `unstable_` prefix on imports like these. I think it's nice to be able to see clearly that it's unstable at the usage site as well. Perhaps we could modify the eslint rule to dirsegard `unstable_` instead?

bjoerge: pushed a commit that allows `unstable_`-prefixes here: 95449799af

packages/@sanity/form-builder/src/inputs/ReferenceInput/types.ts
bjoerge: fixed, thanks!

PR #2952, opened 2021-11-24
fix(form-builder): fix clearing invalid date values

sjelfull:
### Description

This fixes an issue where you weren't able to clear an invalid date in a date field if the field `value` was empty/`undefined`.

Normally this would be handled by the `useEffect` triggering on `value` changing, but in this case `value` is always `undefined` since the invalid date is saved in local state.

### What to review

- That you are able to enter a invalid date like `this is invalid`, then able to clear it again

### Notes for release

- Fixes an issue that prevents clearing a invalid date value when the field is empty

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/DateInputs/base/LazyTextInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/DateInputs/base/LazyTextInput.tsx
@@ -66,6 +66,7 @@ export const LazyTextInput = React.forwardRef(function LazyTextInput(
   return (
     <TextInput
       {...rest}
+      data-testid="date-input"
       ref={forwardedRef}
       value={inputValue === undefined ? value : inputValue}
       onChange={handleChange}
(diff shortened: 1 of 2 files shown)


SimeonGriggs:
Reported issue appears resolved in the Test Studio, thanks!

packages/@sanity/form-builder/src/inputs/DateInputs/base/LazyTextInput.tsx
```diff
@@ -66,6 +66,7 @@ export const LazyTextInput = React.forwardRef(function LazyTextInput(
   return (
     <TextInput
       {...rest}
+      data-testid="date-input"
```
bjoerge, line 69: Did we actually add any tests as part of this change? I think this would be better to pass at usage site instead of hard coding here. Depending on context you might want to give it different test-ids, but this change makes that not possible.

sjelfull: Me and Rico tried, but wasn't able to figure out how to test this since the behaviour was actually correct in tests.

PR #2953, opened 2021-11-24
refactor(desk-tool): improve document pane min width

mariuslundgard:
### Description

- Introduces a new `currentMinWidth` property on the `Pane` component, which is used to render e.g. the `DocumentPane` at a larger width initially, while still allowing the user to resize to a smaller width if desired.
- Refactors the pane layout logic responsible for resizing, collapsing and expanding panes to a controller.

### What to review

- [Code] The code changes to `packages/@sanity/desk-tool/src/components/pane/*`
- [Design] Testing the Desk Tool in a running Studio to make sure that panes behave in the desired manner.

### Notes for release

- Introducing a larger initial `min-width` for the `DocumentPane`.


bjoerge (approved):
LGTM! Great work!

mikolajdobrucki:
We still have this weird case when a seemingly random list pane expands when there is enough space for it:
image
It's not crucial but if we can't change it now, I think it's worth creating a story for it to fix it in the future. 
Otherwise, great work! Good to go! 🔥

mikolajdobrucki (approved):

PR #2954, opened 2021-11-24
fix(client): deprecate defer and add deferred visibility

sgulseth:
### Description

Transaction visibility doesn't support the defer keyword, so deprecate the old in case someone is using it, and add deferred.
See https://www.sanity.io/docs/transactions#80cd6d6bd47b

Note that sending `visibility=defer` will cause:

```
{"error":{"description":"Could not decode query string parameter \"visibility\" value \"defer\": Invalid visibility value","type":"httpBadRequest"}}
```

### What to review

* Implementation 
* 
### Notes for release

Fixes `deferred` mutations

### DIFF
--- a/packages/@sanity/client/sanityClient.d.ts
+++ b/packages/@sanity/client/sanityClient.d.ts
@@ -553,7 +553,7 @@ export interface RequestOptions {
 }
 
 type BaseMutationOptions = RequestOptions & {
-  visibility?: 'sync' | 'async' | 'defer'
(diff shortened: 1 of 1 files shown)


rexxars (commented):
I don't think we need to deprecate it given it doesn't actually work - I think we should just remove it (replace it with deferred). Also tempted to say we should just live without constants - with the constants defined I'd expect them to exist in runtime.

rexxars (approved):

PR #2958, opened 2021-11-25
feat(client): respect useCdn when queries are sent with POST

judofyr:
### Description

`useCdn: true` only applied to GET/HEAD requests sent to the /data-endpoint. However, in the case of a huge query then we send the query as a POST request and this is still supported and cached by the API CDN. This PR makes sure that these requests also gets sent to the API CDN.

This implements it by introducing an internal `canUseCdn`-option to the `_requestObservable` method. (A potential improvement in the future would be to also use this explicitly in the code base instead of having a magical `if (uri.startsWith("/data"))` deep inside a core method, but I decided against to keep the diff small).

### What to review

- Check that adding this functionality into the `_requestObservable` method doesn't cause any problems elsewhere.

### Notes for release

- Calling `fetch` with huge queries will now respect `useCdn: true`.

### DIFF
--- a/packages/@sanity/client/src/data/dataMethods.js
+++ b/packages/@sanity/client/src/data/dataMethods.js
@@ -120,6 +120,7 @@ module.exports = {
 
   _dataRequest(endpoint, body, options = {}) {
     const isMutation = endpoint === 'mutate'
+    const isQuery = endpoint === 'query'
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM!

PR #2961, opened 2021-11-29
Configure PT spellchecking via the schema

skogsmaskin:
### Description

Latest Chromium has some serious performance issues when it comes to spellchecking via the browsers default spellchecking.

This is really noticable on larger documents, and it seems that almost everyone doing text editing online is being hit by this regardless of editor.

https://bugs.chromium.org/p/chromium/issues/detail?id=1271918

To work around this let's finally make spellchecking configurable via the schema for the `block` type:

```
        {
          type: 'block',
          options: {
            spellCheck: true,
          },
        },
```

Spellchecking is off by default with this change which should probably be the default anyway.

### What to review
That this option does the default and is configurable.

### Notes for release
Disabled default-on spellchecking for the Portable Text Input, and introduced schema option to configure this via `options.spellCheck`

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
@@ -203,7 +203,6 @@ export function Editor(props: EditorProps) {
                   renderDecorator={renderDecorator}
                   scrollSelectionIntoView={handleScrollSelectionIntoView}
                   selection={initialSelection}
-                  spellCheck
                 />
               </BoundaryElementProvider>
             </EditableWrapper>
--- a/dev/test-studio/schema/standard/portableText/simpleBlock.js
+++ b/dev/test-studio/schema/standard/portableText/simpleBlock.js
@@ -65,6 +65,9 @@ export default {
               const text = extractTextFromBlocks([block])
               return text.length === 1 ? 'Please write a longer paragraph.' : true
             }),
+          options: {
+            spellCheck: true,
+          },
         },
         {type: 'image', name: 'image'},
         myStringType,
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Input.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Input.tsx
@@ -86,6 +86,16 @@ export function Input(props: InputProps) {
   const [scrollElement, setScrollElement] = useState<HTMLElement | null>(null)
   const handledFocusPath = useRef(null)
 
+  const textBlockSpellCheck = useMemo(() => {
+    // Chrome 96. has serious perf. issues with spellchecking
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=1271918
+    // TODO: check up on the status of this.
+    const spellCheckOption = editor.portableTextFeatures.types.block.options?.spellCheck
+    const isChrome96 =
+      typeof navigator === 'undefined' ? false : /Chrome\/96/.test(navigator.userAgent)
+    return spellCheckOption === undefined && isChrome96 === true ? false : spellCheckOption
+  }, [editor])
+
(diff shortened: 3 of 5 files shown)


packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
```diff
@@ -175,6 +175,14 @@ export function Editor(props: EditorProps) {
     [scrollElement]
   )
 
+  const spellCheck = useMemo(() => {
+    // Only enable from schema
+    if (editor.portableTextFeatures.types.block.options?.spellCheck) {
+      return true
+    }
+    return false
```
rexxars, line 183: Do we not want to enable/disable this based on user agent, or is that too inconsistent?

```suggestion
  const spellCheck = editor.portableTextFeatures.types.block.options?.spellCheck

  // Allow the user to explicitly opt-in/opt-out of spellchecking
  if (typeof spellCheck !== 'undefined') {
    return Boolean(spellCheck)
  }

  // Chrome 96+ has a severe performance regression when using spell checking for contentEditable fields
  return typeof navigator === 'undefined' ? true : !/Chrome\/96/.test(navigator.userAgent)
```

Maybe the regex should be `(9\d|1\d\d)` or something until the issue is fixed, but you get the gist. Thoughts?

skogsmaskin: I was thinking to draw to birds with one stone here. We have been planning to make spellCheck configurable via the schema, and when this is the case, the default should probably be "off". If you want spellCheck then enable it.

spellChecking works with various success both because of language mismatches, and it can also be a performance hog so I'm in favour of putting it off by default.

If this is not what we want, and we want default on, then we def. should do the browser version check.

mariuslundgard: I think this should be allowed to return `undefined` so that it will use the browser default by default.

packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
```diff
@@ -203,7 +211,7 @@ export function Editor(props: EditorProps) {
                   renderDecorator={renderDecorator}
                   scrollSelectionIntoView={handleScrollSelectionIntoView}
                   selection={initialSelection}
-                  spellCheck
+                  spellCheck={spellCheck}
```
mariuslundgard, line 214: Can we set this property on each `TextBlock` instead?

packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
```diff
@@ -175,6 +175,14 @@ export function Editor(props: EditorProps) {
     [scrollElement]
   )
 
+  const spellCheck = useMemo(() => {
+    // Only enable from schema
+    if (editor.portableTextFeatures.types.block.options?.spellCheck) {
+      return true
+    }
+    return false
+  }, [editor])
```
mariuslundgard, line 184:
```suggestion
  // The `spellCheck` can be either `true | false | undefined`
  // When `undefined` the editor will use the browser default (which is the PTE’s default behavior)
  const spellCheck = editor.portableTextFeatures.types.block.options?.spellCheck
```

skogsmaskin:
Thank you for the feedback @rexxars  and @mariuslundgard !

I've force-pushed some changes to this PR.

1) Spellchecking is now set specifically for the PT block type. So it's the only one that is toggled spellcheck for as opposed to the whole PT editor as a whole. This means we might set some types to have spell-checking and other not in the PT-editor.

2) The config supports `undefined` which will let the user/browser decide on spell-checking.

3) If the option is `undefined` and the browser is Chrome 0.96, then disable spell-checking for now.

rexxars (approved):

PR #2965, opened 2021-11-30
test(desk-tool): proxy module instead of spread

ricokahler:
### Description

This PR improves the reliability of the desk-tool tests by using a proxy instead of an object spread.

@sjelfull showed me [a run](https://github.com/sanity-io/sanity/runs/4369598916?check_suite_focus=true) where this test broke to a very strange error `TypeError: Cannot read property 'unstable_useConditionalProperty' of undefined`. This PR fixes that issue by preserving the module and intercepting property calls instead of creating a new object.

### Notes for release

N/A


sjelfull (approved):
Great work, thank you thank you!

vicbergquist (approved):
Aha! Great! Thank you!!

PR #2968, opened 2021-12-02
fix(desk-tool): fix flashing reference changed banner; update copy

ricokahler:
### Description

This PR re-works the reference changed banner a bit to fix some bugs + updates the copy. I noticed that if you quickly switch templates, you'll get the reference changed banner flash.

### What to review

Try replacing an existing reference with a new document created in-place. In the current branch, you should see the reference banner flash but in this version, it's fixed.


bjoerge (approved):
Looks great!

bjoerge:
Oh, but there's a depcheck failure

PR #2969, opened 2021-12-02
fix(default-layout): fix permissions bug in global "create document" menu

bjoerge:
### Description
Fixes a bug that disabled all options in the global 'create document'-menu due to insufficient permissions

### What to review
Verify that the global create document menu display correct disabled state based on current user's permission

### Notes for release
- Fixes a bug that disabled all options in the global 'create document'-menu due to insufficient permissions


RitaDias (approved):
Looks gooood 👌

PR #2972, opened 2021-12-02
test: update document creation and navigation

ricokahler:
### Description

This PR fixes the failing e2e tests for conditional fields by hard-coding a specific test document ID and moving some async logic around. I also removed the clean up phase so concurrent runs shouldn't fail

### What to review

Do the e2e test pass now? Do the changes break any existing assumptions?


RitaDias (approved):
The idea at the time was that we could re-use the creating document logic on future tests. However, if this is the reason that was making the tests flaky (as it seems to be the case), I am fully on board with this approach 👍

PR #2975, opened 2021-12-06
fix(desk-tool): wrap with publishReplay(1)

ricokahler:
### Description

Fix a pane resolution bug where new subscribers of an observable would not get an emission causing a stuck loading pane.

### What to review

I'll have a loom internally with more info.

### Notes for release

Fixes an issue where desk structures with observables would be stuck loading.

### DIFF
--- a/packages/@sanity/desk-tool/src/utils/createPaneResolver.ts
+++ b/packages/@sanity/desk-tool/src/utils/createPaneResolver.ts
@@ -1,5 +1,5 @@
 import {Observable, of as observableOf, from as observableFrom, Subscribable} from 'rxjs'
-import {switchMap} from 'rxjs/operators'
+import {publishReplay, refCount, switchMap} from 'rxjs/operators'
 import {PaneNode, RouterPaneSiblingContext, UnresolvedPaneNode} from '../types'
 import {isRecord} from './isRecord'
 import {PaneResolutionError} from './PaneResolutionError'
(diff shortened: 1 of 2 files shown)


packages/@sanity/desk-tool/src/utils/createPaneResolver.ts
```diff
@@ -53,37 +53,49 @@ const rethrowWithPaneResolutionErrors: PaneResolverMiddleware = (next) => (
   }
 }
 
+const wrapWithPublishReplay: PaneResolverMiddleware = (next) => (...args) => {
+  return next(...args).pipe(
+    // need to add publishReplay + refCount to ensure new subscribers always
+    // get an emission. without this, memoized observables may get stuck
+    // waiting for their first emissions resulting in a loading pane
+    publishReplay(1),
+    refCount()
```
bjoerge, line 62: Not 100% how it fits into the broader context, but did you consider using `shareReplay` here? If you use `shareReplay({bufferSize: 1, refCount: true})` instead of `publishReplay(1)` & `refCount()` the stream will be retry-able in case of an error. (`publishReplay` will replay any error).

From the [shareReplay](https://rxjs.dev/api/operators/shareReplay) docs:
> A successfully completed source will stay cached in the shareReplayed observable forever, but an errored source can be retried.

ricokahler: ah that's a good thought. i did not consider it.

thinking about it now, i'd say the desired behavior is to replay the error. e.g. if the source errors once then switching panes should replay the error i'd think

bjoerge (approved):

PR #2983, opened 2021-12-07
test(cypress, default-layout): add smoke test for initial value templates

RitaDias:
### Description

Add smoke tests for bug with initial value templates when creating new document from the navbar

### Notes for release

Add smoke tests for bug with initial value templates when creating new document from the navbar


rexxars (approved):
👍

PR #2986, opened 2021-12-07
refactor(form-builder): make reference input responsive based on available width

bjoerge:
### Description

Make the reference input responsive based on available/container size rather than screen size

### What to review
- Test studio with references in different context/dialog widths etc.


ricokahler (approved):
💯

PR #2988, opened 2021-12-07
fix(desk-tool): add responsiveness to document footer (collapse publish and changed text)

RitaDias:
### Description

Add responsive way to not have the buttons overlap when the panel (or window) is too small to show all the buttons with the right length.

https://user-images.githubusercontent.com/6951139/145195174-575bf7dd-0595-4a25-b4c4-02aeebb99530.mov

### What to review

- Drag panel and check if the labels work as expected. 
- Check on mobile devices

### Notes for release
Published and Changed Text labels will now collapse once there isn't enough space


packages/@sanity/base/src/components/ElementContainerQuery/ElementContainerQuery.tsx
```diff
@@ -0,0 +1,28 @@
+import {useElementRect} from '@sanity/ui'
```
RitaDias, line 1: Adding this here, this component was a creation of @hermanwikner and I'm so thankful 🙌

PR #2996, opened 2021-12-09
test: add form validation test

hermanwikner:
### Description

This test focuses on the visibility of the validation icons displayed in each field, and whether the publish button is enabled or disabled depending on the validation of a document.

Screenshot 2021-12-09 at 15 10 28

### What to review

- The test in general and how it is authored
- Should more scenarios be tested?

### Notes for release

Add form validation test

### DIFF
--- a/cypress/integration/form-builder/validation.test.js
+++ b/cypress/integration/form-builder/validation.test.js
@@ -1,0 +1,102 @@
+import {nanoid} from 'nanoid'
+import client from '../../helpers/sanityClientSetUp'
+
+const docId = `drafts.${nanoid()}`
+
+const doc = {
+  _id: docId,
+  _type: 'validationCI',
+  errorString: '',
+  warningString: '',
+  infoString: '',
+}
+
+/** Each fields validation rule is configured with min(5) */
+const VALID_VALUE = 'String longer than 5'
+const INVALID_VALUE = 'abc'
+
+async function patchDocument(payload) {
+  return client.patch(docId).set(payload).commit({visibility: 'async'})
+}
+
+describe('@sanity/form-builder: Validation', () => {
+  before(async () => {
+    cy.viewport(1000, 1500)
+    await client.create(doc, {visibility: 'async'})
+  })
+
+  beforeEach(() => {
+    cy.visit(`/test/desk/input-ci;validationCI;${docId}%2Ctemplate%3DvalidationCI`)
+  })
+
+  /** Error */
+  it('when there are validation errors, the publish button is disabled', () => {
+    cy.wrap(patchDocument({errorString: INVALID_VALUE})).then(() => {
+      cy.get('[data-testid="action-Publish"]').should('have.attr', 'disabled')
+    })
+  })
+
+  it('when there are validation errors, the error icon is visible', () => {
+    cy.get('[data-testid="input-errorString"]')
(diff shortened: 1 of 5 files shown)


ricokahler (commented):
i don't think your assertions are running. i'll my findings async

cypress/integration/form-builder/validation.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+  /** Error */
+  it('when there are validation errors, the publish button is disabled', () => {
+    client
+      .patch(docId)
+      .set({errorString: INVALID_VALUE})
+      .commit()
+      .then(() => {
+        cy.get('[data-testid="action-Publish"').should('have.attr', 'disabled')
+      })
```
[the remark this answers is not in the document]
hermanwikner: Thanks! I have updated the test, so this should be solved! :)

cypress/integration/form-builder/validation.test.js
```diff
@@ -0,0 +1,102 @@
...
+  warningString: '',
+  infoString: '',
+}
+
+/** Each fields validation rule is configured with min(5) */
+const VALID_VALUE = 'String longer than 5'
+const INVALID_VALUE = 'abc'
+
+async function patchDocument(payload) {
+  return client.patch(docId).set(payload).commit()
```
rexxars, line 19: Cool trick: If you use async visibility here it'll return a lot quicker :)
This tells the backend to not wait for the document to be "indexed" before returning, but you'll still be able to navigate to it in the studio and such!

```suggestion
  return client.patch(docId).set(payload).commit({visibility: 'async'})
```

cypress/integration/form-builder/validation.test.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+describe('@sanity/form-builder: Validation', () => {
+  before(async () => {
+    await client.create(doc)
```
rexxars, line 24:
```suggestion
    await client.create(doc, {visibility: 'async'})
```

cypress/integration/form-builder/validation.test.js
```diff
@@ -0,0 +1,102 @@
...
+  it('when there are no validation infos, the info icon is hidden', () => {
+    cy.wrap(patchDocument({infoString: VALID_VALUE})).then(() => {
+      cy.get('[data-testid="input-infoString"]')
+        .find('[data-testid="input-validation-icon-info"]')
+        .should('not.exist')
+    })
+  })
+
+  after(async () => {
+    await client.delete(docId)
```
rexxars, line 100:
```suggestion
    await client.delete(docId, {visibility: 'async'})
```

rexxars (approved):
Looks good! Couple of suggestions on how to speed up the tests slightly - make sure to verify that the tests still works with these changes, there might be race conditions due to it being a bit faster, but hopefully not :)

cypress/integration/form-builder/validation.test.js
hermanwikner: Thanks! Will add 👍

hermanwikner:
> 

Added the `{visibility: 'async'}` and the test still works. Also moved the `cy.viewport(1000, 1500)` from `beforeEach` to `before`! @rexxars

PR #2998, opened 2021-12-09
fix(core): fix file asset urls not working with ndjson target

rexxars:
### Description

If you try to run `sanity dataset import <someNdJsonFile>` or `sanity-import <someNdJsonFile>`, any relative file URLs within that file will cause the upload to fail. 

If you point to a folder _containing_ an ndjson file, it will work - as will pointing to a tarball.

This PR fixes the issue by passing the directory root path to use for relative assets when importing from a stream (which will happen with ndjson files).

### Notes for release

- Fixes `sanity dataset import` not working when targeting an ndjson file with relative file URLs


bjoerge (approved):
Nice! LGTM

PR #2999, opened 2021-12-09
chore: prefer get-it for http requests

rexxars:
### Description

We have multiple places in our codebase that uses simple request libraries like simple-get and get-uri for doing HTTP requests - in particular when we want to use streams. Apparently this is causing issues for some of our customers who are using proxies, which these libraries do not work with.

I recently added support for returning streams from [get-it](https://github.com/sanity-io/get-it/commit/fa85fe952baa859c35d3da6eeec8359d94e86aae), and reusing the same library that we use for our Sanity client will make things more consistent while enabling features like redirect handling and decompressing responses.

### What to review

- That CLI commands still work
- That the code changes makes sense

### Notes for release

- Added support for HTTP proxies when using dataset `import`/`export` commands


bjoerge (approved):
LGTM 👏

PR #3000, opened 2021-12-10
feat(cli): use non-legacy yarn v1.22.17 under the hood

rexxars:
### Description

- Upgrades yarn to the latest 1.x version, and uses the non-legacy version (eg the one that only works on node 8 and up) for less transpiled code
- Bumps the network timeout when running yarn commands to 1 minute instead of the default (30 seconds). This _might_ help the issues we're seeing on CI where windows builds fail to install dependencies during the init step


skogsmaskin (approved):
LGTM!

PR #3001, opened 2021-12-10
fix(cli): clear auth token on 401 on logout

rexxars:
### Description

When attempting to log out of the CLI, one might encounter situations where the session is either timed out, has been cleared or does not exist. In these cases it is perfectly acceptable to just clear the token from the CLIs point of view, instead of throwing an error which is not actionable.

Also removes a bunch of ancient provisional session handling that is no longer used.

### Notes for release

- Fixes a bug preventing `sanity logout` from working in the case of expired sessions


skogsmaskin (approved):
:+1:

PR #3002, opened 2021-12-10
fix(cli): use authless requests when logging in

rexxars:
### Description

When attempting to log in, if the CLI already has an auth token it will be used for requests to endpoints such as `/providers`. These requests might be rejected if the session is timed out or invalid, which is unfortunate - it might be the reason why the user is attempting to log in, after all.

This PR:
- Uses an unauthenticated client for requesting these routes, to prevent them from being rejected
- Attempts to log out any existing session upon successful login
- (Unrelated) Sets a stable API version for the SAML providers endpoint instead of using `vX`

### Notes for release

- Fixes a bug where the CLI would not be able to log a user in an existing invalid/expired session was present

### DIFF
--- a/packages/@sanity/cli/src/actions/login/login.js
+++ b/packages/@sanity/cli/src/actions/login/login.js
@@ -166,7 +166,7 @@ async function getProvider({output, client, sso, experimental, prompt}) {
 }
 
 async function getSSOProvider({client, prompt, slug}) {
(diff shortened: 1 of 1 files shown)


skogsmaskin (approved):
LGTM!

sgulseth (approved):

PR #3006, opened 2021-12-13
feat(form-builder): add field groups

vicbergquist:
### Description

This PR adds the concept of Field Groups to schemas, along with a way to filter fields in document view based on said field groups.

Field Groups is presentational only, and should replace any uses of existing tab plugins like `sanity-plugin-tabs`.

A field can belong to multiple field groups.

Example schema:

```js
{
  name: 'article',
  title: 'Article',
  type: 'document',
  groups: [
    {
      name: 'seo',
      title: 'SEO',
    },
  ],
  fields: [
    {name: 'title', title: 'Title', type: 'string'},
    {name: 'icon', title: 'Icon', type: 'image'},
    {
      name: 'related',
      title: 'Related',
      type: 'array',
      of: [{type: 'reference', to: [{type: 'article'}]}],
    },
    {name: 'seoTitle', title: 'SEO title', type: 'string', group: 'seo'},
    {name: 'seoKeywords', title: 'Keywords', type: 'string', group: 'seo'},
    {name: 'seoSlug', title: 'Slug', type: 'slug', group: 'seo'},
    {name: 'seoImage', title: 'Image', type: 'image', group: 'seo'},
  ],
}
```

|Screenshots||
|--|--|
|image|image|

### What to review

[You can test field groups in this deployed test studio](https://test-studio-git-feature-sc-12720implement-fieldgroups.sanity.build/test/desk/input-debug;fieldsetsTest;ba7e4d38-0b7e-4f01-8278-f3cdd117c85e)

- Performance when typing in simple and more complex schemas
- That field groups works with validation markers and presence
- That the `hidden` property on groups work both with booleans and callbacks
- That the document view switches the All fields group when Review Changes is open
- That the field groups tab is hidden if the only group remaining is "All fields" - like when you have one group named "SEO" and you set `hidden: true`
- That nested field groups works
- That the first field group with `default: true` is selected by default

### Notes for release

- Added Field Groups to schemas and document view


packages/@sanity/form-builder/src/inputs/ObjectInput/fieldGroups/Group.tsx
```diff
@@ -0,0 +1,88 @@
+import React, {forwardRef, useEffect} from 'react'
```
bjoerge, line 1: This file should probably be renamed to GroupTab.tsx so that it matches the component name

packages/@sanity/form-builder/src/inputs/ObjectInput/fieldGroups/Group.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const {autoFocus, selected, hidden, parent, onStateChange, ...group} = props
+  const isHidden = useConditionalProperty({
+    checkProperty: hidden,
+    checkPropertyKey: 'hidden',
+    value: group,
+    parent,
+  })
+
+  useEffect(() => {
+    onStateChange(name, Boolean(isHidden))
```
bjoerge, line 70: Same concern as above re: triggering an update in the parent component as a side effect of a state change in this. Would hope we can lift this calculation up to a common ancestor so the required data can be shared between child components.

packages/@sanity/form-builder/src/inputs/ObjectInput/fieldGroups/Group.tsx
```diff
@@ -0,0 +1,88 @@
...
+    parent,
+  })
+
+  // Here goes the content of our component
+  const handleClick = React.useCallback(() => {
+    onClick(name)
+  }, [name, onClick])
+
+  useEffect(() => {
+    onStateChange(name, Boolean(isHidden))
```
bjoerge, line 37: Any way we can avoid triggering a state update in the parent component as a side effect of a state update in this one? Maybe we could somehow lift the isHidden calculation up one level?

packages/@sanity/form-builder/src/inputs/ObjectInput/fieldGroups/Group.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const {onClick, onStateChange, parent, hidden, ...group} = props
+
+  const isHidden = useConditionalProperty({
+    checkProperty: hidden,
+    checkPropertyKey: 'hidden',
+    value: group,
+    parent,
+  })
+
+  // Here goes the content of our component
```
bjoerge, line 31: Not sure if this comment makes all that much sense to me… :)

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
```diff
@@ -80,6 +84,56 @@ export const ObjectInput = memo(
...
+    }, [filterGroups])
+    const [selectedFieldGroupName, setSelectedFieldGroupName] = useState(defaultFieldGroupName)
+    const [isAllGroupsHidden, setIsAllGroupsHidden] = useState(false)
+    const fieldGroupRootFocusPaths = React.useMemo(() => {
+      if (filterGroups.length === 0) {
+        return type.fields.map((field) => field.name)
+      }
+
+      return (
+        find(filterGroups, (fieldGroup) => fieldGroup.name === selectedFieldGroupName)?.fields || []
```
bjoerge, line 122: any reason to use `lodash.find` here? I think it's sensible to prefer native array methods unless there's a clear benefit with using lodash.

packages/@sanity/form-builder/src/inputs/ObjectInput/ObjectInput.tsx
```diff
@@ -181,12 +235,48 @@ export const ObjectInput = memo(
       ]
     )
 
+    const fieldGroupPredictive = useCallback(
```
bjoerge, line 238: Looks like a typo - should it be `fieldGroupPredicate`?

packages/@sanity/form-builder/src/FormBuilderInput.tsx
```diff
@@ -214,6 +214,12 @@ export class FormBuilderInput extends React.Component<FormBuilderInputProps> {
     return PathUtils.trimChildPath(path, focusPath)
   }
 
+  handleTabChange = (newFocusPath) => {
```
bjoerge, line 217: This doesn't seem to be in use?

packages/@sanity/form-builder/src/FormBuilderInput.tsx
```diff
@@ -230,22 +236,24 @@ export class FormBuilderInput extends React.Component<FormBuilderInputProps> {
...
-            childFocusPath={this.getChildFocusPath()}
-            context={this.context}
-            component={InputComponent}
-            onBlur={this.handleBlur}
-            onChange={this.handleChange}
-            onFocus={this.handleFocus}
-            setInput={this.setInput}
-          />
-        </ConditionalReadOnlyField>
+        <>
```
bjoerge, line 239: Unnecessary fragment

packages/@sanity/language-filter/src/datastore.ts
```diff
@@ -63,7 +63,8 @@ const filterField = config.filterField || defaultFilterField
 
 export const filterFn$ = selectedLanguages$.pipe(
   map((langs) => {
-    return (enclosingType: SchemaType, field: ObjectField) =>
-      filterField(enclosingType, field, langs)
+    return (enclosingType: SchemaType, field: ObjectField) => {
```
bjoerge, line 66: this diff doesn't seem relevant? can it be (re)moved?

bjoerge (commented):
Amazing work!!!

Highlighted a few issues that would be great if we could take a look at together. Some of them should be straight forward but I suspect some of the issues related to state updates/data flow are rather tricky to fix right now, but I think it's worth considering if there's something we can do to improve anyway (or at least make a plan for how to fix properly later).

packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
```diff
@@ -50,10 +51,11 @@ export function FormView(props: FormViewProps) {
     historyController,
     markers,
     ready,
+    changesOpen,
   } = useDocumentPane()
   const presence = useDocumentPresence(documentId)
   const {revTime: rev} = historyController
-  const [{filterField}, setState] = useState<FormViewState>(INITIAL_STATE)
+  const [{filterField}, setFilterField] = useState<FormViewState>(INITIAL_STATE)
```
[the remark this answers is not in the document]
sjelfull: I think this is some leftover from when we re-factored away from using the filterField function. I'll see about cleaning this up.

packages/@sanity/language-filter/src/datastore.ts
sjelfull: Same as above re. `filterField`

packages/@sanity/desk-tool/src/panes/document/documentPanel/documentViews/FormView.tsx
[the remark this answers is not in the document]
sjelfull: Looks like this is how it was originally written. Unless there is a reason not no, i'll clean this up regardless.

packages/@sanity/form-builder/src/inputs/ObjectInput/fieldGroups/Group.tsx
sjelfull: The tricky part here is that our conditional logic is a hook and component that processes the group, and this means it can't easily be used in the parent component. We could possibly create a smarter hook that processes the array of groups instead.

bjoerge (approved):

PR #3009, opened 2021-12-14
feat(cli): respect specified ranges when upgrading

rexxars:
### Description

When running `sanity upgrade`, we would previously resolve the `latest` tag for
each installed `@sanity` module and blindly upgrade them to this version.

This is confusing, as it breaks with how npm/yarn handles upgrades. With this
PR, we instead find the maximum version compatible with the specified range for
a dependency. This also means that modules with pinned version numbers will
stay on their given version.

In addition to these improvements, this PR fixes an old pet peeve of mine where
the command would also upgrade dependencies like `@sanity/block-tools`, which
might be used in import tooling or similar. It _should_ only upgrade studio
dependencies, but we don't have an explicit list of these modules, so we use
an exclude list instead for now.

Most important change of this PR is to respect the semver ranges specified:
If you're on `@sanity/base` @ `v2.23.0` and a new `v3.0.0` is released, we
can't blindly upgrade to it, as this might have backwards incompatible changes.
With these changes, we'll warn about semver-incompatible upgrades that are
available, with a special handling for `@sanity/base`, which will signify a
new major version of the studio.

### Notes for release

- `sanity upgrade` now behaves slightly differently:
  - It now respects the version ranges/pinned versions declared in `package.json`
  - Tries not to upgrade non-studio modules such as `@sanity/block-content-to-html`
  - Shows a more prominent message if there are major updates available, with a link to the changelog

### DIFF
--- a/packages/@sanity/cli/package.json
+++ b/packages/@sanity/cli/package.json
@@ -56,7 +56,7 @@
     "fs-extra": "^7.0.0",
     "gauge": "^2.7.4",
     "get-it": "^5.2.1",
-    "get-latest-version": "^1.0.0",
+    "get-latest-version": "^3.0.1",
     "git-user-info": "^1.0.1",
     "gitconfiglocal": "^2.0.1",
     "inquirer": "^6.0.0",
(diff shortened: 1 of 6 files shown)


bjoerge (approved):
Looks great, excellent work!

PR #3012, opened 2021-12-15
feat(form-builder): add workshop for FormBuilderInput

sjelfull:
### Description

This adds a workshop for `FormBuilderInput`, and allows us to test the form builder, schemas and fields sans some of the studio-specific logic that is in `SanityFormBuilder` and `FormView`.

image


packages/@sanity/form-builder/src/__workshop__/example.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const isUseMutator = useBoolean('Use Mutator', false, 'Props')
+  const [, setFocused] = useState(false)
+  const [focusPath, setFocusPath] = useState([])
+  const isReadOnly = useBoolean('Read-only', false, 'Props')
+  const isFilterFields = useBoolean('Filter Fields', false, 'Props')
+  const isHiddenGroup = useBoolean('Hidden Group', false, 'Props')
+  const includeUnknownField = useBoolean('Unknown Field in Value', false, 'Props')
+  const isDebug = useBoolean('Debug', false, 'Props')
+  const [documentValue, setDocumentValue] = useState<{[key: string]: any}>(getDummyDocument())
+  const [_fieldFilter, setFieldFilter] = useState<string>(``)
```
bjoerge, line 30: I think it would be more clear if this was named fieldFilterSource (or similar)

packages/@sanity/form-builder/src/__workshop__/example.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setFocusPath(path)
+    setFocused(true)
+  }, [])
+  const handleChangeFieldFilter = useCallback((value) => {
+    const handledValue = value && value.length > 0 ? value : ``
+
+    setFieldFilter(handledValue)
+    toast.push({status: 'success', title: `Updated field filter`})
+  }, [])
+  const fieldFilter = useCallback(
```
bjoerge, line 62: this will parse the function source every time we filter a field. Would it be a possible improvement/simplification to parse it in a useMemo that depends on the fieldFilter source string?

packages/@sanity/form-builder/src/FormBuilderContext.ts
```diff
@@ -49,14 +49,17 @@ function memoize(method) {
   }
 }
 
+export interface PatchChannelProps {
```
bjoerge, line 52: Nice one!

packages/@sanity/form-builder/src/utils/empty.ts
```diff
@@ -1,5 +1,4 @@
 export const EMPTY_OBJECT: any = Object.freeze({})
 export const EMPTY_ARRAY: any = Object.freeze([])
-export const emptyObject = <T>() => EMPTY_OBJECT as T
-
-export const emptyArray = <T>() => EMPTY_ARRAY as T
+export const emptyObject = <T extends unknown>() => EMPTY_OBJECT as T
+export const emptyArray = <T extends unknown>() => EMPTY_ARRAY as T
```
bjoerge, line 4: does it work to use T extends unknown[] here?

packages/@sanity/form-builder/src/__workshop__/_common/input.tsx
```diff
@@ -0,0 +1,153 @@
+import React, {
```
bjoerge, line 1: Not sure if the naming of this file (`input.tsx`) makes all that much sense to me. Could we try to come up with a better name? Alternatively would it make sense to split it into different component files?

bjoerge (changes requested):
Great work, @fred! I've left some comments and ideas for improvements that would be good to address before merge.

packages/@sanity/form-builder/src/__workshop__/_common/input.tsx
```diff
@@ -0,0 +1,153 @@
...
+          >
+            {props.children}
+          </FormBuilderContext>
+        </ToastProvider>
+      </LayerProvider>
+    </ThemeProvider>
+  )
+}
+
+export const FormDebugger = function FormDebugger(props: FormDebuggerProps) {
```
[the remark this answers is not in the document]
sjelfull: No reason, it was using `forwardRef` before

packages/@sanity/form-builder/src/FormBuilderContext.ts
sjelfull: I did rename it to `PatchChannelOptions` to match the others

packages/@sanity/form-builder/src/__workshop__/_common/input.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }))
+
+  return (
+    <DebugTextArea
+      padding={3}
+      radius={1}
+      readOnly
+      ref={rootRef}
+      rows={10}
+      value={JSON.stringify(props.value, null, 2)}
```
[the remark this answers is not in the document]
sjelfull: In theory this might be fine, since the input should only show up if you use an unknown type, so it is rare that you will see it.

packages/@sanity/form-builder/src/__workshop__/example.tsx
sjelfull: Memoized the filter 👍

packages/@sanity/form-builder/src/__workshop__/_common/input.tsx
sjelfull: I've split it up into separate component files and renamed input to `DebugInput`

bjoerge (approved):

PR #3014, opened 2021-12-15
Maintenance update for the Portable Text Editor

skogsmaskin:
### Description

This is a rather big PR with several updates and fixes for both  `@sanity/portable-text-editor` and `@sanity/form-builder`.

A summary of the changes in their respective packages:

#### `@sanity/portable-text-editor`
* `slate` and `slate-react` (Portable Text Editor dependencies) finally upgraded to latest versions.
Adds better Android and IME support, and generally better stability and performance.

With this chore in mind, I have taken this opportunity to make several new improvements to the portable-text-editor code base. 

* There are now tighter and cleaner types which makes the relation between the portable text editor and the slate react editor easier to understand and work with. A lot of determination code could also be replaced by convenience methods that check the said types better.

I found serious problems with how the selection was passed around and kept in separate states throughout the application. This is now boiled down and fixed so that we can `===` those selections and check if it is the exact same selection or not.

The selection object will now be the same object in:
* `PortableTextEditor.getSelection` method
* In the `usePortableTextEditorSelection` hook.
* Weak-mapped to the slate editor selection by said object.

Refactored handling of focus and blur to behave cleaner and to be easier to work with (less side effects).

Refactored and improved copy/paste. The editor now only deals with `application/x-portable-text` data besides HTML and plain text.

Made collaborative tests perform more reliably and faster, and added more of these tests.

#### `@sanity/form-builder`

The PortableText Input for the studio has also been updated with several improvements.

* Everything in the editor is now keyboard accessible.
* Opening and closing object editing modals and popovers have become much more responsive.
* Better tracking of focus, scrolling to the relevant stuff if needed (like when clicking validation messages scroll to the relevant editor content).

### What to review

I know this one is huge, so it's not easy to review. Would be great if you looked through the code though and see if there is something obvious that can be done better or differently.

### Notes for release
* Better Android and IME support in the Portable Text Editor.
* Better keyboard accessibility in the Portable Text Editor for editing objects.
* Maintenance update to the Portable Text Editor fixing several bugs and increasing performance and reliability.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
@@ -23,6 +23,7 @@ import {
   Scroller,
   ToolbarCard,
 } from './Editor.styles'
+
 interface EditorProps {
   initialSelection?: EditorSelection
   isFullscreen: boolean
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
@@ -95,18 +95,20 @@ export function Editor(props: EditorProps) {
     }
   }, [])
 
+  const renderPlaceholder = useCallback(() => <>Empty</>, [])
+
   const editable = useMemo(
     () => (
       <PortableTextEditable
         hotkeys={hotkeys}
         onCopy={onCopy}
         onPaste={onPaste}
-        placeholderText="Empty"
         ref={editableRef}
         renderAnnotation={renderAnnotation}
         renderBlock={renderBlock}
         renderChild={renderChild}
         renderDecorator={renderDecorator}
+        renderPlaceholder={renderPlaceholder}
         scrollSelectionIntoView={scrollSelectionIntoView}
         selection={initialSelection}
         spellCheck={false} // This is taken care of via renderBlock prop
@@ -120,6 +122,7 @@ export function Editor(props: EditorProps) {
       renderAnnotation,
       renderBlock,
       renderChild,
+      renderPlaceholder,
       scrollSelectionIntoView,
     ]
   )
--- a/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
@@ -88,7 +88,7 @@ export function Editor(props: EditorProps) {
   }, [isFullscreen, isTopLayer, onToggleFullscreen])
 
   // Keep the editor focused even though we are clicking on the background or the toolbar of the editor.
-  const keepFocused = useCallback((event: React.SyntheticEvent) => {
+  const handleMouseDown = useCallback((event: React.SyntheticEvent) => {
     if (event.target instanceof Node && !editableRef.current.contains(event.target)) {
       event.preventDefault()
       event.stopPropagation()
@@ -125,7 +125,7 @@ export function Editor(props: EditorProps) {
   )
 
   return (
-    <Root $fullscreen={isFullscreen} data-testid="pt-editor" onMouseDown={keepFocused}>
+    <Root $fullscreen={isFullscreen} data-testid="pt-editor" onMouseDown={handleMouseDown}>
       <ToolbarCard data-testid="pt-editor__toolbar-card" shadow={1}>
         <Toolbar
           isFullscreen={isFullscreen}
--- a/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
@@ -224,14 +224,16 @@ const PortableTextInputController = React.forwardRef(function PortableTextInputC
     return null
   }, [handleEditorChange, handleIgnoreValidation, invalidValue])
 
-  // Scroll to the field if we have focus and needed
+  // Scroll to *the field* (not the editor) into view if we have focus in the field.
+  // For internal editor scrolling see useScrollToFocusFromOutside and useScrollSelectionIntoView in
+  // the Compositor component.
   useEffect(() => {
     if (focusPath && focusPath.length > 0 && innerElementRef.current) {
       scrollIntoView(innerElementRef.current, {
         scrollMode: 'if-needed',
       })
     }
-  }, [focusPath, value])
+  }, [focusPath])
 
   return (
     <div ref={innerElementRef}>
--- a/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
@@ -11,6 +11,7 @@ import {
   PortableTextEditor,
   Type,
   HotkeyOptions,
+  EditorSelection,
 } from '@sanity/portable-text-editor'
 import {Subject} from 'rxjs'
 import {Box, Text, useToast} from '@sanity/ui'
@@ -165,11 +166,7 @@ const PortableTextInputController = React.forwardRef(function PortableTextInputC
           })
           break
         case 'selection':
-          if (
-            change.selection && // If we have something selected
-            PortableTextEditor.isObjectPath(ref.current, focusPath) === false && // Not if this is pointing to an embedded object
-            focusPath.slice(-1)[0] !== FOCUS_TERMINATOR // Not if in transition to open modal
-          ) {
+          if (shouldSetEditorFormBuilderFocus(ref.current, change.selection, focusPath)) {
             onFocus(change.selection.focus.path)
           }
           break
@@ -277,3 +274,15 @@ const PortableTextInputController = React.forwardRef(function PortableTextInputC
     </div>
   )
 })
+
+function shouldSetEditorFormBuilderFocus(
+  editor: PortableTextEditor,
+  selection: EditorSelection | undefined,
+  focusPath: Path
+) {
+  return (
+    selection && // If we have something selected
+    focusPath.slice(-1)[0] !== FOCUS_TERMINATOR && // Not if in transition to open modal
+    PortableTextEditor.isObjectPath(editor, focusPath) === false // Not if this is pointing to an embedded object
+  )
+}
--- a/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
@@ -124,12 +124,12 @@ const PortableTextInputController = React.forwardRef(function PortableTextInputC
   const patches$: Subject<EditorPatch> = useMemo(() => new Subject(), [])
   const patchObservable = useMemo(() => patches$.asObservable(), [patches$])
 
-  const innerElementRef = useRef(null)
+  const innerElementRef = useRef<HTMLDivElement | null>(null)
 
   const handleToggleFullscreen = useCallback(() => {
-    setIsFullscreen(!isFullscreen)
+    setIsFullscreen((v) => !v)
     PortableTextEditor.focus(ref.current)
-  }, [isFullscreen, ref])
+  }, [ref])
 
   // Reset invalidValue if new value is coming in from props
   useEffect(() => {
--- a/packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObjectPreview.tsx
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObjectPreview.tsx
@@ -1,12 +1,27 @@
 import {IntentLink} from '@sanity/base/router'
 import {EditIcon, LinkIcon, TrashIcon, EyeOpenIcon, EllipsisVerticalIcon} from '@sanity/icons'
-import {PortableTextBlock, Type} from '@sanity/portable-text-editor'
-import {Box, Button, Flex, Menu, MenuButton, MenuButtonProps, MenuItem} from '@sanity/ui'
-import React, {forwardRef, useMemo} from 'react'
+import {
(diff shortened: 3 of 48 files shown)


packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
```diff
@@ -170,13 +87,45 @@ export function Editor(props: EditorProps) {
...
+    [
+      hotkeys,
+      initialSelection,
+      onCopy,
+      onPaste,
+      renderAnnotation,
+      renderBlock,
+      renderChild,
+      scrollSelectionIntoView,
+    ]
```
mariuslundgard, line 124: I believe we don’t really need to wrap this in `useMemo`. It doesn’t add much perf improvement, but actually adds another operation instead. Think it can just be rendered in the `return` below.

PS. We should instead wrap the `PortableTextEditable` in a `memo` if we only want it to re-render on a prop change.

```tsx
const MemoPortableTextEditable = memo(PortableTextEditable)

// ...

<MemoPortableTextEditable ... />
```

packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
```diff
@@ -170,13 +87,45 @@ export function Editor(props: EditorProps) {
     }
   }, [isFullscreen, isTopLayer, onToggleFullscreen])
 
-  const handleScrollSelectionIntoView = useMemo(
-    () => createScrollSelectionIntoView(scrollElement),
-    [scrollElement]
+  // Keep the editor focused even though we are clicking on the background or the toolbar of the editor.
+  const keepFocused = useCallback((event: React.SyntheticEvent) => {
```
mariuslundgard, line 91: This should be named `handleMouseDown`

packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -84,12 +120,16 @@ const PortableTextInputWithRef = React.forwardRef(function PortableTextInput(
 
   const toast = useToast()
 
-  const editorId = useMemo(() => uniqueId('PortableTextInputRoot'), [])
   // Memoized patch stream
   const patches$: Subject<EditorPatch> = useMemo(() => new Subject(), [])
   const patchObservable = useMemo(() => patches$.asObservable(), [patches$])
 
-  const handleToggleFullscreen = useCallback(() => setIsFullscreen(!isFullscreen), [isFullscreen])
+  const innerElementRef = useRef(null)
```
mariuslundgard, line 127: Should be typed so it’s easier to migrate to strict TypeScript in the future:

```suggestion
  const innerElementRef = useRef<HTMLDivElement | null>(null)
```

packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -55,7 +91,7 @@ export type PortableTextInputProps = {
   value: PortableTextBlock[] | undefined
 }
 
-const PortableTextInputWithRef = React.forwardRef(function PortableTextInput(
+const PortableTextInputController = React.forwardRef(function PortableTextInputController(
   props: Omit<PortableTextInputProps, 'level'>,
   ref: React.RefObject<PortableTextEditor>
```
mariuslundgard, line 96: I believe the correct typing here is:
```suggestion
  ref: React.ForwardedRef<PortableTextEditor>
```

packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -84,12 +120,16 @@ const PortableTextInputWithRef = React.forwardRef(function PortableTextInput(
...
-  const editorId = useMemo(() => uniqueId('PortableTextInputRoot'), [])
   // Memoized patch stream
   const patches$: Subject<EditorPatch> = useMemo(() => new Subject(), [])
   const patchObservable = useMemo(() => patches$.asObservable(), [patches$])
 
-  const handleToggleFullscreen = useCallback(() => setIsFullscreen(!isFullscreen), [isFullscreen])
+  const innerElementRef = useRef(null)
+
+  const handleToggleFullscreen = useCallback(() => {
+    setIsFullscreen(!isFullscreen)
```
hermanwikner, line 130: Small thing - I prefer this approach for toggle boolean in a state: `setIsFullscreen((v) => !v)`

This might also be more performant, because `isFullscreen` can be removed from the `useCallback` dependency array, and the function will not be re-created when that value changes.

packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObjectPreview.tsx
```diff
@@ -22,8 +37,11 @@ const POPOVER_PROPS: MenuButtonProps['popover'] = {
...
 
 export function BlockObjectPreview(props: BlockObjectPreviewProps) {
-  const {value, type, readOnly, onClickingEdit, onClickingDelete} = props
+  const {focused, value, type, readOnly, onClickingEdit, onClickingDelete} = props
+  const editor = usePortableTextEditor()
   const menuButtonId = useId()
+  const menuButton = useRef<HTMLButtonElement>()
+  const isTabbing = useRef<boolean>(false)
   const isCustomPreviewComponent = Boolean(type.preview?.component)
   const layout = 'block'
```
hermanwikner, line 46: This constant could be moved to the top level of the component: `const LAYOUT = 'block'`

packages/@sanity/form-builder/src/inputs/PortableText/object/InlineObject.tsx
```diff
@@ -30,6 +38,11 @@ function rootStyle({theme}: {theme: Theme}) {
     box-shadow: inset 0 0 0 1px var(--card-border-color);
     height: calc(1em - 1px);
     margin-top: 0.0625em;
+    cursor: default;
+
+    & > div:after {
```
hermanwikner, line 43: What does this `:after` do here? :)

packages/@sanity/form-builder/src/inputs/PortableText/object/InlineObjectToolbarPopover.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onDelete: (event: React.MouseEvent<HTMLButtonElement>) => void
+  onEdit: (event: React.MouseEvent<HTMLButtonElement>) => void
+  open: boolean
+  referenceElement: HTMLElement
+  scrollElement: HTMLElement
+  setOpen: (open: boolean) => void
+  title: string
+}
+
+export function EditObjectToolTip(props: InlineObjectToolbarPopoverProps) {
```
hermanwikner, line 33: The component is exported as `EditObjectToolTip` but the file name is `InlineObjectToolbarPopoverProps` – should they be the same name?

packages/@sanity/portable-text-editor/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "rootDir": "./src",
     "outDir": "./dist/dts",
     "jsx": "react",
-    "types": ["jest", "puppeteer"]
+    "types": ["jest", "puppeteer", "node"]
```
mariuslundgard, line 9: Is the `types` array actually needed here? I believe TypeScript will pick up these automatically, if they are installed as `@types/*`.

skogsmaskin: They are for running the local dev server (only).

skogsmaskin: Haven't tried removing it all together though!

packages/@sanity/portable-text-editor/src/utils/values.ts
```diff
@@ -50,12 +57,20 @@ export function toSlateValue(
...
+      return keepObjectEquality(
+        {
+          _type,
+          _key,
+          children: voidChildren,
+          value: rest,
+        },
+        keyMap
+      )
+    }) as Descendant[]
```
mariuslundgard, line 73: It this type casting required? `as Descendant[]`

packages/@sanity/portable-text-editor/src/utils/values.ts
```diff
@@ -67,40 +82,47 @@ export function fromSlateValue(
...
-          ) as PortableTextBlock
+          throw new Error('Not a valid block type')
         }
-        throw new Error('Not a valid block type')
+        const {_key, _type} = block
+        const blockValue = 'value' in block && block.value
+        return keepObjectEquality(
+          {_key, _type, ...(typeof blockValue === 'object' ? blockValue : {})},
+          keyMap
+        ) as PortableTextBlock
```
mariuslundgard, line 123: Is the `as PortableTextBlock` type casting required?

packages/@sanity/portable-text-editor/src/utils/values.ts
```diff
@@ -67,40 +82,47 @@ export function fromSlateValue(
...
       }
-      const {_key, _type} = block as PortableTextBlock
-      const blockValue = block.value as PortableTextBlock | undefined
-      return keepObjectEquality(
-        {_key, _type, ...(typeof blockValue === 'object' ? blockValue : {})},
-        keyMap
-      ) as PortableTextBlock
+      return block as PortableTextBlock
```
mariuslundgard, line 125: Is the `as PortableTextBlock` type casting required?

packages/@sanity/portable-text-editor/src/types/slate.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+}
+
+declare module 'slate' {
+  interface CustomTypes {
+    Editor: BaseEditor & ReactEditor & PortableTextSlateEditor
+    Element: TextBlock | VoidElement
+    Text: TextSpan
+  }
+}
```
mariuslundgard, line 21: Is this the official way to extend Slate types? Can we add a link to the docs here?

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextSelections.ts
```diff
@@ -0,0 +1,40 @@
...
+      } else {
+        change$.next({type: 'selection', selection: null})
+      }
+    }
+
+    const {onChange} = editor
+    editor.onChange = () => {
+      onChange()
+      emitSelection()
+    }
```
mariuslundgard, line 37: Should this callback pass on arguments?
```suggestion
    editor.onChange = (...args) => {
      onChange(...args)
      emitSelection()
    }
```

packages/@sanity/form-builder/src/inputs/PortableText/text/textStyles.tsx
```diff
@@ -2,45 +2,50 @@ import {Heading, Text} from '@sanity/ui'
 import React from 'react'
 import styled from 'styled-components'
 
+// Without this container, editing with Android breaks due to how Text is styled via responsiveFont in @sanity/ui
+const TextContainer = styled('div')`
+  display: block;
+`
+
```
mariuslundgard, line 9: Seems like this can either be replaced by 
```tsx
const TextContainer = styled.div``
```

or simple a `<div>`, since `div` is `display: block` by default.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextSelections.ts
skogsmaskin: There aren't any args for this fn.

mariuslundgard (approved):
Seems solid – great work! 👏

packages/@sanity/form-builder/src/inputs/PortableText/Editor.tsx
skogsmaskin: I know we removed the memo in the last iteration, but then `value` and other stuff was part of the deps. The deps now are pretty much non-changing, and that's why I'm thinking a memo makes sense here now.

packages/@sanity/form-builder/src/inputs/PortableText/PortableTextInput.tsx
skogsmaskin: I'm getting error that it doesn't have `.current` when trying to use that one.

packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObjectPreview.tsx
skogsmaskin: Yeah, agree!

packages/@sanity/form-builder/src/inputs/PortableText/object/InlineObject.tsx
skogsmaskin: Nothing! I've removed it. Good catch.

packages/@sanity/form-builder/src/inputs/PortableText/object/InlineObjectToolbarPopover.tsx
skogsmaskin: Most def! Renamed it. Thanks.

packages/@sanity/portable-text-editor/tsconfig.json
skogsmaskin: Seems to be working fine when I remove the entry altogether.

packages/@sanity/portable-text-editor/src/utils/values.ts
skogsmaskin: Yes, here it actually is, because it's a util to convert values.

packages/@sanity/portable-text-editor/src/utils/values.ts
skogsmaskin: Yes, as it's a conversion util we have to manually assign the types here.

PR #3018, opened 2021-12-16
fix(form-builder): the PTE dialogs should not close when clicking outside

mariuslundgard:
### Description

- Disables click-outside-to-close for dialogs in the PTE.
- Wraps `editObjectNode` with the correct boundary element in the PTE.
- Passes a `boundaryElement` (the third argument) to `useClickOutside` to prevent popovers from closing when clicking outside of the boundary element (the current pane content).


skogsmaskin (approved):
Nice!

PR #3019, opened 2021-12-16
fix(base): uniq and chunk availability requests

bjoerge:
When requesting document availability we're collecting all requests within the same event loop tick and batches them into a single request to the endpoint that checks for availability reason. In cases where a document had a lot of references this could produce a request that was above the max request size. This commit splits the list of ids to check into chunks of 300 documents and fetches their availability in parallel, at most 10 requests at a time. These numbers are merely guesses about what would be reasonable limits.

Also includes a fixed typing issue with `debounceCollect` that didn't reflect real world behavior.

### Notes for release
- Fixes an issue that could cause a max request size exceeded error for documents with many references.

### DIFF
--- a/packages/@sanity/base/src/preview/availability.ts
+++ b/packages/@sanity/base/src/preview/availability.ts
@@ -1,8 +1,8 @@
 /* eslint-disable max-nested-callbacks */
-import {combineLatest, defer, Observable, of} from 'rxjs'
-import {distinctUntilChanged, map, switchMap} from 'rxjs/operators'
+import {combineLatest, defer, from, Observable, of} from 'rxjs'
+import {distinctUntilChanged, map, mergeMap, switchMap} from 'rxjs/operators'
 import shallowEquals from 'shallow-equals'
-import {keyBy} from 'lodash'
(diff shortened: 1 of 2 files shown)


packages/@sanity/base/src/preview/availability.ts
```diff
@@ -59,8 +59,17 @@ function observeDocumentAvailability(id: string): Observable<DocumentAvailabilit
 }
 
 const fetchDocumentReadability = debounceCollect(function fetchDocumentReadability(
-  ids: string[]
+  args: string[][]
 ): Observable<DocumentAvailability[]> {
+  const uniqueIds = [...new Set(flatten(args))]
+  return from(chunk(uniqueIds, 300)).pipe(
+    mergeMap(fetchDocumentReadabilityChunked, 10),
```
ricokahler, line 66: didn't know there was a second argument here. cool 😎

ricokahler (approved):
very sick

packages/@sanity/base/src/preview/availability.ts
bjoerge: It's the best second argument! <3

PR #3020, opened 2021-12-17
chore(form-builder): keep array item focus in sync with focusPath

bjoerge:
### Notes for release
- fixes a bug that caused focus loss on an array item after closing the edit dialog

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
+++ b/packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
@@ -15,9 +15,10 @@ import {Box} from '@sanity/ui'
 import PatchEvent from '../../../../PatchEvent'
 import {ArrayMember, ReferenceItemComponentType} from '../types'
 import {EMPTY_ARRAY} from '../../../../utils/empty'
-import {hasFocusWithinPath} from '../../../../utils/focusUtils'
+import {hasFocusAtPath, hasFocusWithinPath} from '../../../../utils/focusUtils'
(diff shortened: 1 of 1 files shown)


packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
```diff
@@ -66,6 +67,13 @@ export const ArrayItem = memo(function ArrayItem(props: ArrayInputListItemProps)
   const hasFocusWithin = hasFocusWithinPath(props.focusPath, props.value)
   useScrollIntoViewOnFocusWithin(innerElementRef, hasFocusWithin)
 
+  useDidUpdate(hasFocusAtPath(props.focusPath, props.value), (hadFocus, hasFocus) => {
```
ricokahler, line 70: i think it's better to store the result from hasFocusAtPath in a variable and then use a vanilla effect to call the focus

ricokahler (approved):
i threw an opinionated comment in there but hey it works 😄

packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/item/ArrayItem.tsx
bjoerge: Curious to hear why do you think it would be better :)

I personally feel this captures the intention pretty well. An `useEffect` here would have to keep track of the previous value since we're not merely responding to _change_ but the transition from `!hadFocus` to `hasFocus`.

Merging this for now, but I'm happy to improve this pattern and change it into something better later on.

PR #3021, opened 2021-12-17
fix(cypress): fix flaky e2e test for review changes

sjelfull:
### Description

This fixes the flaky e2e tests for Conditional Fieldsets. It some times takes a while for the Review Changes pane to appear, and this change makes the previously failing tests wait for the pane to appear (for max 10s) before continuing.

### What to review

That the tests now runs successfully again.

### Notes for release

N/a


cypress/integration/form-builder/conditionalFieldset.test.js
```diff
@@ -37,6 +37,12 @@ const doc = {
   singleReadOnlyCallbackTrue1: 'Lorem',
 }
 
+const waitForReviewChanges = () => {
```
vicbergquist, line 40: Instead of using this and calling it inside each test, you can probably use the `beforeEach` hook? Like this:

```
 beforeEach(() => {
    cy.contains('Changes', {timeout: 10000}).should('be.visible')
  })
```

vicbergquist (changes requested):

cypress/integration/form-builder/conditionalFieldset.test.js
sjelfull: That makes sense for this test, will update

RitaDias (approved):
Let's gooo 🎉

cypress/integration/form-builder/conditionalFieldset.test.js
```diff
@@ -37,6 +37,12 @@ const doc = {
...
+  // Sometimes it takes a while for Review Changes to appear when loading the page.
+  // This prevents the test from failing randomly
+  cy.contains('Changes', {timeout: 10000}).should('be.visible')
```
vicbergquist, line 43: What exactly does `Changes` represent? Can we not check if the element with the review changes test id is visible instead? `Changes` feels a little vague

sjelfull: Added testid to the review changes panel instead.

PR #3022, opened 2021-12-17
fix(base): prevent crash if type `*` is provided to structure

ricokahler:
### Description

This PR prevents a crash from the newer `unstable_useDocumentPermissions` API.

This comes from an internal bug report where the reporter was trying to create an `S.documentList()` with a `.filter('...')` and no `.schemaType('...')` in attempt to have the document list return more than one type.

Prior to this API, this was not supported and resulted in this: 

CleanShot 2021-12-16 at 16 02 30@2x

However, with the new API, you'll get a "the Desk Tool crashed" error which is much less clear.

### What to review

You can reproduce the error by creating a structure with `S.documentList().filter("_type in ['post', 'author']")`. When you try to load an item from that list in `next`, it will throw. With this patch, you should get an error similar to the screenshot above.

### Notes for release

- Fixes an issue that caused a crash instead of a warning screen when omitting `schemaType` from `S.documentList()`

### DIFF
--- a/packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
+++ b/packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
@@ -149,6 +149,15 @@ function getDocumentPairPermissions({
   type,
   permission,
 }: DocumentPermissionsOptions): Observable<PermissionCheckResult> {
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
LGTM!

PR #3024, opened 2021-12-20
fix(server): consider cwd in when resolving module paths

ricokahler:
### Description

I found that using npm workspaces doesn't always work with the studio if the package being resolved isn't in the root-level `node_modules`. All the details of this issue should be in the PR in code comments.

### What to review

Try running this in a monorepo setup. It should work :D

### Notes for release

Fixes an issue where modules would fail to resolve in certain monorepo/workspace setups.


ricokahler:
converted to draft for more testing actually

ricokahler:
nvm i think it's fine now 😓 . i turn on legacyPeerDeps and my issues went away

PR #3027, opened 2021-12-21
fix(cypress): improve e2e test speed

sjelfull:
### Description

This is a second pass of #3021 after it turned out tests were still flaky.

The most important improvement over #3021 is that I discovered two things:
- Most of the test time was spent waiting for the studio to reload before each test.
- All tests in a test group could be tested against the same version of the DOM, so there was no reason to reload the studio

The test speed went from a consistently 2+ minutes to running at around 16 seconds, or from 4 to 2 minutes for all tests.

### What to review

- That all tests runs as expected

### Notes for release

N/a


RitaDias (approved):
✅

PR #3029, opened 2021-12-22
fix(form-builder): add prop for passing the item component to array of objects input

bjoerge:
### Description

Adds a small  change to the array of objects input that allows for passing the array item component. Potentially something we could provide as part of an official API in the future, but I'd consider it internal for now.

### Notes for release

N/A (internal only)


rexxars (approved):
👍

PR #3030, opened 2021-12-23
fix(form-builder): fix ui issues in PTE toolbar

hermanwikner:
### Description

This PR fixes an issue where the toolbar sometimes gets double borders, and the expand button was incorrectly positioned when there were no action / insert items.

Screenshot 2021-12-23 at 13 00 57


skogsmaskin (approved):
Cool! Thanks for this @hermanwikner

Issue #3011, opened 2021-12-15
Singleton document type doesn't load


jfulse:
**Describe the bug**
On version 2.23.1 our current approach to displaying singleton documents in the studio no longer works; when clicking the menu item the page doesn't finish loading.

**To Reproduce**

1. The desk structure is something like:
```
const documentId = ...

export default S.list()
  .title("title")
  .items([
    S.documentTypeListItem("site")
      .title("Singleton document")
      .child(S.editor().id(documentId)),
  ]);
```
2. Click on the document in the left menu
3. The document doesn't load

**Expected behavior**

The document should load after a bit

**Screenshots**

![Skjermbilde 2021-12-15 kl  13 58 43](https://user-images.githubusercontent.com/13033324/146190847-283cf814-f21e-4676-8933-77dc487b2310.png)

**Which versions of Sanity are you using?**
@sanity/cli             2.19.1 (latest: 2.23.1)
@sanity/base            2.23.1 (up to date)
@sanity/cli             2.23.1 (up to date)
@sanity/core            2.23.1 (up to date)
@sanity/default-layout  2.23.1 (up to date)
@sanity/default-login   2.23.1 (up to date)
@sanity/desk-tool       2.23.1 (up to date)
@sanity/imagetool       2.22.3 (up to date)
@sanity/react-hooks     2.23.1 (up to date)
@sanity/vision          2.23.1 (up to date)

**What operating system are you using?**

macOS Big Sur

**Which versions of Node.js / npm are you running?**
8.1.3
v16.11.0

ricokahler:
Hi @jfulse, I'm investigating this now and it seems like you can work around this issue by passing the schemaType along with your ID.

```js
const documentId = ...

export default S.list()
  .title("title")
  .items([
    S.documentTypeListItem("site")
      .title("Singleton document")
      .child(
        S.editor()
          .id(documentId)
          .schemaType("site") // <-- add this
      ),
  ]);
```

We [deprecated](https://github.com/sanity-io/sanity/commit/e0d82f06fff8256d82938fcb3239ce37b118b90c#diff-5e010626f97d87b9f7f61cf0c1d5bf37a092de64b0ea7756a944ae36184eb104R151) support for omitting the schema type due to a similar issue and complications resolving the type. However, we did not intent to break support for this (since v2.23.0)

---

[...]

PR #3043, opened 2022-01-04
fix(desk-tool): fix issue with singletons with no type not loading

ricokahler:
### Description

Fixes the root cause of #3011.

This issue was occurring because the `isLoaded` boolean was never being set to true causing 

### What to review

Try the current broken singleton case in the test-studio with this branch. It should load now.

i.e. `/test/desk/custom;foo-bar`

### Notes for release

Fixes an issue that caused singleton documents to never load if no schemaType was provided.


bjoerge (approved):
👍

PR #3045, opened 2022-01-06
fix(form-builder): fix race condition during publish of referenced document that sometimes produced an invalid query

bjoerge:
### Description

This fixes a rather peculiar bug that caused the studio to occasionally produce an invalid query when fetching preview fields for a reference.

### What to review
- Create new document from a reference input
- Edit the new document and wait for a few seconds
- Publish the referenced document
- Verify that no error message appears

### Notes for release
- Fixes an issue that could occasionally cause a query error during publish of a new document created directly from the reference input


ricokahler (approved):
nice!

bjoerge:
Merged in f2224881aba3041f2ec61a84532a83fe288d3a7f and released in [v2.23.4](https://github.com/sanity-io/sanity/releases/tag/v2.23.4)

PR #3047, opened 2022-01-06
feat(util): add SANITY_STUDIO_PROJECT_NAME env variable

runeb:
### Description

This PR adds `SANITY_STUDIO_PROJECT_NAME` env variable for setting the configured Studio project name via environment variables. It also adds previously missing tests for env variable config overrides.

### What to review

Tests and the changed logic of util reduceConfig. I moved some env handling from the global scope of this file into the exported function. This change should affect all usages involving running, building or otherwise using the Studio config.

### Notes for release

Added `SANITY_STUDIO_PROJECT_NAME` environment config variable


rexxars (approved):
Thanks, this is great!

PR #3048, opened 2022-01-07
fix search limit issues

bjoerge:
### Description

This fixes a couple of oversights/unintentional changes from the references in place release:
1) the default limit on studio searches were reduced from 1000 to 100. Since the search results comes unranked from the API this could cause better matches to disappear from the result list in cases where you had a lot of documents.
2) when fetching for existence of draft/published counterparts from hits returned from the reference search we used a `debug` request tag. Also, upping the amount of displayed search results in the reference input from 50 to 100 (after ranking)

### What to review
That studio search works as expected, both global document search and reference search.

### Notes for release
- Fixed an issue that could in some cases cause matching documents to be excluded from the search results


ricokahler (approved):
nice 👍

PR #3049, opened 2022-01-07
fix(schema): improve default title resolution for references

### LINKED ISSUE #3046: Reference fields show the type name rather than the title
With the new Sanity design, using an array of references field like this will expose the internal type name to the editor, which is not user friendly.

```
type: 'array', of: [ {
    type: 'reference', to: [{
        type: 'salesforceAttributeValue'
    }]
}]
```

Sanity will show the internal type name (in this case "**salesforceattributevalue**") in the heading when clicking "Add item" to an array of references:

image

The expected behaviour would be that the type **title** would be shown instead, as this is the humanly readable display name.

This bug can be circumvented by adding a `title` attribute in the field definition, like this:

```
type: 'array', of: [ {
    type: 'reference', to: [{
        title: 'Salesforce attribute value',   //  <-- Display title duplicates title defined inside the type
        type: 'salesforceAttributeValue'
    }]
}]
```

This, however, should not be necessary, as the same name is already defined within the type itself.

Also, best practise, as shown in multiple examples in Sanity's own documentation, is to define references using **only** the `type` attribute.  We have 50+ places to update in our code if adding a title to all reference arrays  is now required.

bjoerge:
### Description
When no title is given for a referenced document type, we currently default to the type name as-is, which makes the UI less user friendly (as pointed out in #3046).

This PR fixes it so that we display the referenced type in this order:
- The title defined for the referenced type at the reference field/array item, if any
- The title specified for the referenced document type, if any
- A [startCase](https://lodash.com/docs/3.10.1#startCase)'d version of the referenced document's type name (note: this follows from how we already resolve the default title for object types)

### What to review
Verify that the UI displays titles of reference fields and array items in a user friendly way

### Notes for release

- Made the default title of referenced document types more user friendly

Fixes #3046

### DIFF
--- a/packages/@sanity/schema/src/legacy/types/reference.ts
+++ b/packages/@sanity/schema/src/legacy/types/reference.ts
@@ -36,18 +36,13 @@ function humanize(arr, conjunction) {
 }
 
 function buildTitle(type) {
-  if (type.title) {
-    return type.title
-  }
   if (!type.to || type.to.length === 0) {
     return 'Reference'
   }
   return `Reference to ${humanize(
-    arrify(type.to).map((toType) =>
-      (toType.title || toType.name || toType.type || '').toLowerCase()
-    ),
+    arrify(type.to).map((toType) => toType.title),
     'or'
-  )}`
+  ).toLowerCase()}`
 }
 
 export const ReferenceType = {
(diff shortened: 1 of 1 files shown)


rexxars (approved):
Needs a rebase, but looks good from what I can tell!

PR #3054, opened 2022-01-12
feat: eslint config

ricokahler:
### Description

The following PR:

1. creates a new shared eslint config package we can offer to studio users
2. creates an `.eslintrc` file that includes this config while bootstrapping new projects

Note: the scope of this is to _only_ add a working`.eslintrc` to new projects. A future improvement would be to add a CLI command that would run the linter and run the linter prior to builds but we'd probably also want to run a type checking for TypeScript files too.

### What to review

- Review the rules in the config. I tried to make it a good balance between unobtrusive and helpful. Try `npm pack`ing it and installing it in a fresh project to get a feel for it.
- Review the changes to the CLI. Try bootstrapping a new project.

### Notes for release

New package `@sanity/eslint-config-studio`!

We now offer a shared eslint config that you can use to get better editor feedback while writing code in the Studio.

- New projects bootstrapped with the CLI will now ship with a default `.eslintrc`


ricokahler:
Looking at these warnings, i'm thinking the eslint config might be better outside of the mono repo 🤔 
 
CleanShot 2022-01-11 at 22 04 20@2x

PR #3055, opened 2022-01-12
feat: add eslint to new studio projects

ricokahler:
### Description

Supersedes #3054 (I keep forgetting to add the sc back link lol)

The following PR:

~1. creates a new shared eslint config package we can offer to studio users~ This work has been moved [here](https://github.com/sanity-io/eslint-config-studio)
2. creates an `.eslintrc` file that includes this config while bootstrapping new projects

Note: the scope of this is to _only_ add a working`.eslintrc` to new projects. A future improvement would be to add a CLI command that would run the linter and run the linter prior to builds but we'd probably also want to run a type checking for TypeScript files too.

### What to review

- Review the rules in the config. I tried to make it a good balance between unobtrusive and helpful. Try `npm pack`ing it and installing it in a fresh project to get a feel for it. [See here](https://github.com/sanity-io/eslint-config-studio)
- Review the changes to the CLI. Try bootstrapping a new project.

### Notes for release

- New projects bootstrapped with the CLI will now contain a default `.eslintrc` that extends our new shared ESLint config [`@sanity/eslint-config-studio`](https://github.com/sanity-io/eslint-config-studio)


mariuslundgard:
- Did we consider using [`eslint-config-sanity`](https://github.com/sanity-io/eslint-config-sanity)? If it’s not a good match, can it be modified?
- In other projects (`sanity-io/sanity` and `sanity-io/design`) we use `@typescript-eslint/parser` instead of `@babel/eslint-parser`. Is that possible in this case as well?

ricokahler:
@mariuslundgard

i wanted to name this config more explicitly in case we have new variants we want to release separately. i was looking at next.js and they have more than one config available. their latest is next/core-web-vitals but the old one still remains

additionally, part of me is wondering if this config should live outside the monorepo so we can version independently. any thoughts there?

---

i think the typescript eslint parser requires typescript as a dependency in order to work correctly which we don't require with studio projects.

i mostly went with the babel parser separately because that's what next.js and create-react-app does too

mariuslundgard:
> i wanted to name this config more explicitly in case we have new variants we want to release separately. i was looking at next.js and they have more than one config available. their latest is next/core-web-vitals but the old one still remains

Sounds wise. While keeping the eslint config in `eslint-config-sanity` and naming the Studio preset `"sanity/studio"` we’d do exactly that. And we could add more presets e.g. `"sanity/plugin"` etc. later.

> i think the typescript eslint parser requires typescript as a dependency in order to work correctly which we don't require with studio projects.

Ah, I see! Yeah, don’t want to force people to install TypeScript if they don't use it 👍

ricokahler:
@mariuslundgard 

> Sounds wise. While keeping the eslint config in eslint-config-sanity and naming the Studio preset "sanity/studio" we’d do exactly that. And we could add more presets e.g. "sanity/plugin" etc. later.

ah i learned about preset yesterday and didn't think of using one. let's do that!

rexxars (approved):

PR #3036, opened 2021-12-30
feat: add languages to code input

rawkode:
### Description

Added some missing languages: Dockerfile, Elixir, Ini, Nix, and Rust.

Ensured that it was updated in config and editorSupport

### What to review

NA

### Notes for release

Added languages: Dockerfile, Elixir, Ini, Nix, and Rust.

Issue #2682, opened 2021-08-17
[code-input] Add language support for YAML and Ruby


llibdude:
**Describe the solution you'd like**
I would like support for YAML and Ruby for the @sanity/code-input plugin.

**Additional context**
I found this PR for adding YAML which was approved a while ago and then recently closed.
https://github.com/sanity-io/sanity/pull/1840

Both requested languages are supported by the brace library.
https://github.com/thlorenz/brace/tree/master/mode

Thanks for considering these languages.  I know adding languages is a WIP, so hopefully this issue bumps up these language priorities :)

judofyr:
At some point we should make it possible to customize the languages supported by the plugin. Doesn't _really_ make sense to hard code the supported languages.

I've opened a PR for supporting these languages here: https://github.com/sanity-io/sanity/pull/2686

bjoerge:
As of [v2.24.0](https://github.com/sanity-io/sanity/releases/tag/v2.24.0) you can now import the ace modes you need from your own Studio. See the [readme](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/code-input#add-support-for-more-languages) for an example.

Issue #2357, opened 2021-03-04
Add TSX support to @sanity/code-input


schlosser:
**Is your feature request related to a problem? Please describe.**
I want to use Typescript React code snippets in my blog posts but there are only `JSX` and `TS` options.

**Describe the solution you'd like**
Add `TSX` to `SUPPORTED_LANGUAGES`, using `brace/modes/tsx`.

**Describe alternatives you've considered**
Adding TSX to `languageAlternatives`  and having it unstyled in the CMS, forking `@sanity/code-input`.

**Additional context**
`brace/mode/tsx` is here: https://github.com/thlorenz/brace/blob/master/mode/tsx.js

bjoerge:
As of [v2.24.0](https://github.com/sanity-io/sanity/releases/tag/v2.24.0) you can now import the ace modes you need from your own Studio. See the [readme](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/code-input#add-support-for-more-languages) for an example.

Issue #2264, opened 2021-01-27
code-input: Add the option to send a language string


bdashore3:
**Is your feature request related to a problem? Please describe.**
I'm trying to use the code-input extension and I would like to use Rust (since I program in it a lot) as one of the input languages. However, I can't add a custom language string to select in studio, because if I manually add rust in the languages list, the output comes out to `unknown`. One solution that was outlined in the slack thread was to choose a language I'd never use and set it to use Rust in my website code.

**Describe the solution you'd like**
I'd like the option to add custom language strings within the code-input editor for languages that aren't supported by inline highlighting on the studio. If highlighting exists, it shows up when editing code. Language selection would be based on the syntax highlighter such as `react-syntax-highlighter`

**Describe alternatives you've considered**
Using a provided language I'd never use (such as SASS) and adding a if statement in my code to set `node.language = 'rust'`.

**Additional context**
Sanity slack thread: [https://sanity-io-land.slack.com/archives/C9Z7RC3V1/p1611765338105000](https://sanity-io-land.slack.com/archives/C9Z7RC3V1/p1611765338105000)

jaydanurwin:
+1 on this because I'd like to use Dart as a language for my blog posts about Flutter. Thank you for sharing the temporary workaround as well!

bjoerge:
As of [v2.24.0](https://github.com/sanity-io/sanity/releases/tag/v2.24.0) you can now import the ace modes you need from your own Studio. See the [readme](https://github.com/sanity-io/sanity/tree/next/packages/%40sanity/code-input#add-support-for-more-languages) for an example.

PR #3056, opened 2022-01-12
Configurable ace modes

bjoerge:
### Description

People (understandably) want @sanity/code-input to support more languages (see #3036, #2357, #2682, #2264). However, enabling a language by default increases the bundle size even if the language is not used.

This PR adds support for configuring additional language support from the host studio application (see readme for an example).

Edit: oh, and while I was at it I upgraded react-ace to latest and removed brace in favor of ace-builds (official builds).

### What to review
See that the change makes sense and think of any potential problems with this approach that I didn't anticipate.

### Notes for release

- Allow configuring support for additional languages in @sanity/code input


bjoerge:
> should we revisit the built in languages we bundle up? Not sure how much it adds to the size though.

Had the same thought, it would be a breaking change so didn't bother with it now, but we should probably consider deprecating all but maybe a very few defaults and plan to remove them in the long term.

These are the numbers for the language modes we currently include by default:
  Uncompressed: 1.24 MB
  Minified (uglify): 844 kB
  Minified and gzipped (level: default): 194 kB

rexxars (approved):
Nice!

rexxars:
I agree that we should cut down the amount of languages we support out of the box, but this _is_ a breaking change, so I'd hold off on that for now. Maybe we should move the code input out of the monorepo and version it independently? That would allow us to make this breaking change.

bjoerge:
> Maybe we should move the code input out of the monorepo and version it independently?

Yep, makes a lot of sense. Let's do it next time around? (nothing about this PR will complicate that as far as I can tell, so merged for now). Thanks for reviewing!

PR #3058, opened 2022-01-12
fix(base): fix permissions bugs by fallbacking back to published variant

ricokahler:
### Description

Fixes incorrect permissions in certain permissions cases + adds failing tests to cover this case in the future. Details in the comments of the code.

### What to review

Review the changes and the new tests.

### Notes for release

Fixes a bug that caused some false warnings with `duplicate` and `update` permissions.

### DIFF
--- a/packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
+++ b/packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
@@ -22,8 +22,8 @@ function getSchemaType(typeName: string): SchemaType {
 
 interface PairPermissionsOptions {
   permission: DocumentPermission
-  draft: SanityDocument
-  published: SanityDocument
+  draft: SanityDocument | null
+  published: SanityDocument | null
   liveEdit: boolean
 }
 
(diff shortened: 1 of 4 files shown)


packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
```diff
@@ -32,7 +32,17 @@ function getPairPermissions({
...
   liveEdit,
-}: PairPermissionsOptions): Array<[string, Observable<PermissionCheckResult>] | null> {
+}: PairPermissionsOptions): Array<[string, Observable<PermissionCheckResult>]> {
+  // this was introduced because we ran into a bug where a user with publish
+  // access was marked as not allowed to duplicate a document unless it had a
+  // draft variant. this would happen in non-live edit cases where the document
+  // pair only had a published variant with the draft variant being null.
+  //
+  // note: this should _not_ be used if the draft and published versions should
+  // be considered separately/explicitly in the permissions.
```
bjoerge, line 42: nit, but wondering if it would be better to move the variables here into the two respective blocks that uses them? It adds a bit of duplication, but think it's justified by increased cohesion and also keeping them out of scope for the other checks, and this comment can be removed.

bjoerge (approved):
Great work and kudos for tests! Just a minor nit, feel free to ignore and merge.

packages/@sanity/base/src/datastores/grants/documentPairPermissions.ts
ricokahler: i like it!

PR #3071, opened 2022-01-17
test(desk-tool): add tests for default document actions

vicbergquist:
### Description

Adding e2e tests for:
- Default document actions: publish, unpublish, duplicate, delete, discard changes

### What to review

If tests are as they should, are there ways to improve them?

### DIFF
--- a/cypress/helpers/constants.js
+++ b/cypress/helpers/constants.js
@@ -1,0 +1,7 @@
+export const DEFAULT_CONFIG = {
+  defaultCommandTimeout: 15000,
+  viewportWidth: 2000,
+  viewportHeight: 3500,
+}
+
+export const DATASET_NAME = 'test'
--- a/cypress/helpers/index.js
+++ b/cypress/helpers/index.js
@@ -1,0 +1,5 @@
+export * from './constants'
+export * from './createUniqueDocument'
+export {default as sanityClientConfig} from './sanityClientSetUp'
+
+export const getTestId = (name) => `[data-testid="${name}"]`
--- a/cypress/integration/desk-tool/documentActions.test.js
+++ b/cypress/integration/desk-tool/documentActions.test.js
@@ -1,0 +1,116 @@
+import {
+  createUniqueDocument,
+  DEFAULT_CONFIG,
+  getTestId,
+  DATASET_NAME,
+  sanityClientConfig,
+} from '../../helpers'
+
+const TITLE = 'Document actions [Cypress]'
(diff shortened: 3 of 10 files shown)


cypress/helpers/index.js
```diff
@@ -0,0 +1,5 @@
+export * from './constants'
+export * from './createUniqueDocument'
+export {default as sanityClientConfig} from './sanityClientSetUp'
+
+export const getTestId = (name) => `[data-testid="${name}"]`
```
ricokahler, line 5: i like this helper and googled if there was something similar built in and found [custom commands](https://docs.cypress.io/api/cypress-api/custom-commands) where they have this example:

```js
Cypress.Commands.add('getBySel', (selector, ...args) => {
  return cy.get(`[data-test=${selector}]`, ...args)
})

// usage looks like `cy.getBySel('myTestId')`
```

how would you feel about making this a custom command? it might not work with intellisense though

cypress/helpers/constants.js
```diff
@@ -0,0 +1,7 @@
+export const DEFAULT_CONFIG = {
+  defaultCommandTimeout: 15000,
+  viewportWidth: 2000,
+  viewportHeight: 3500,
+}
```
ricokahler, line 5: is there a way to configure this part globally?

ricokahler (commented):
I think it looks good but I'm not well-versed in cypress 😅. I added one comment in there regarding `deleteDocument` that might be worth a second look. nice work!

cypress/integration/desk-tool/documentActions.test.js
```diff
@@ -0,0 +1,116 @@
...
+  return sanityClientConfig.delete(payload)
+}
+
+describe('@sanity/desk-tool: base document actions', DEFAULT_CONFIG, () => {
+  let documentId
+  let documentPath
+
+  beforeEach(() => {
+    // Delete all the documents already created
+    cy.wrap(deleteDocument({query: `*[_type == "documentActionsCi"]`}))
```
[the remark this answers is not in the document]
vicbergquist: Yeah... Not sure either, will likely be solved if/when we move to playwright

cypress/helpers/index.js
vicbergquist: Great idea, yeah I was thinking about that! Would be better, seeing as we will use it across tests.

cypress/helpers/constants.js
vicbergquist: Yes I think so! will take a look

PR #3072, opened 2022-01-17
refactor(form-builder): use context for conditional read only

sjelfull:
### Description

This PR changes the `ConditionalReadOnlyField` field to use a context to pass the evaluated `readOnly` value instead of mapping over the children with `React.Children.map`. I wasn't able to prove that mapping over the children caused a performance regression, but it's still a good change to make.

### What to review

- That tests still run successfully

### Notes for release

n/a


vicbergquist (changes requested):
Great job! Looks good to me. This should also probably be added to the change resolver?

sjelfull:
> Great job! Looks good to me. This should also probably be added to the change resolver?

Ah yeah, missed that! Will sort out.

packages/@sanity/form-builder/src/inputs/common/conditionalReadOnly/ConditionalReadOnlyContextProvider.tsx
```diff
@@ -0,0 +1,26 @@
+import React, {useMemo} from 'react'
```
bjoerge, line 1: How does this relate to the `ConditionalReadOnlyContextProvider` in @sanity/base? Do we still need both?

packages/@sanity/form-builder/src/inputs/common/conditionalReadOnly/ConditionalReadOnlyContext.ts
```diff
@@ -0,0 +1,9 @@
+import {createContext} from 'react'
```
bjoerge, line 1: How does this relate to the ConditionalReadOnlyContext exported from @sanity/base? Do we still need both?

bjoerge (commented):
I'm wondering if we still need the ConditionalReadOnly context files in the form-builder package, but other than that LGTM!

packages/@sanity/form-builder/src/inputs/common/conditionalReadOnly/ConditionalReadOnlyContextProvider.tsx
sjelfull: I believe we don't need this any more, will remove them

RitaDias (approved):
🛳 do the thing

PR #3074, opened 2022-01-17
fix(core): provide ace globals when mocking browser env

rexxars:
### Description

It seems the code input is still breaking GraphQL deployments (at least I can reproduce this when adding the input to a studio and having a code field). It is likely that the way we are mocking a browser environment might lead ace to be somewhat confused, and attempting to use a global `ace` instance, which does not exist.

This PR attempts to find the `ace-builds` dependency as a studio dependency, and if it does, import it and set it to an `ace` global when mocking the environment. It's pretty hacky, but it's the best I could come up with short term.

### What to review

- Add `@sanity/code-input` to a studios `package.json` dependencies
- Add `@sanity/code-input` to `plugins` in `sanity.json`
- Add a field of type `code` to any document/object schema type
- Run `sanity graphql deploy`
- Ensure that it does not crash

### Notes for release

- Fixes a bug where using the `@sanity/code-input` might lead to `sanity graphql deploy` crashing due to an undefined `ace` variable

### DIFF
--- a/packages/@sanity/core/src/util/mockBrowserEnvironment.js
+++ b/packages/@sanity/core/src/util/mockBrowserEnvironment.js
@@ -1,5 +1,6 @@
 const pirates = require('pirates')
 const jsdomGlobal = require('jsdom-global')
+const resolveFrom = require('resolve-from')
 const pluginLoader = require('@sanity/plugin-loader')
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
LGTM!

PR #3077, opened 2022-01-18
fix(code-input): ace mode variable is interpreted wrongly

ISnackable:
### Description

Address the following issue of the changes from #3056.  Due to JavaScript [Operator precedence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#operator_precedence), the mode variable is interpreted wrongly, 

When the ace mode name is different from the language name, the following issue occurred. The mode is `c_cpp` while the language value is `cpp`.
image

Before:
image

After:
image

Note: This issue is not visible if the ace mode and language value is the same.

### What to review

- Check that the code input work with an alternative language that has a different ace mode name to the language value name.
- Check that the code input work with default supported languages
- See file `dev/test-studio/schema/plugins/code.js`

### Notes for release

- Fixed an issue causing the code-input plugin to wrongly assign the ace mode with the error message "Uncaught SyntaxError: Unexpected token '<'"

### DIFF
--- a/dev/test-studio/schema/plugins/code.js
+++ b/dev/test-studio/schema/plugins/code.js
@@ -2,6 +2,7 @@ import {CodeBlockIcon} from '@sanity/icons'
 
 // eslint-disable-next-line import/no-unassigned-import
 import 'ace-builds/src-noconflict/mode-rust'
+import 'ace-builds/src-noconflict/mode-c_cpp'
 
 export default {
   name: 'codeTest',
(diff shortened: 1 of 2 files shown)


bjoerge:
Great catch, thanks @ISnackable!

ISnackable:
Hi @bjoerge, I realised the same issue "Uncaught SyntaxError: Unexpected token '<'" occurs for `PreviewCode.tsx` when the code preview is rendered in the PortableText. However, in this case, it is because the code is not assigned to the `configured?.mode` as seen in the code snippet below.

```ts
const mode = value?.language || fixedLanguage || 'text'
```

I'm not sure on how to fix this as the `type` variable is not passed to the props. Could you kindly look into this? Thank you.

PR #3081, opened 2022-01-18
fix(desk-tool): prevent stale `actionHandlers`

### LINKED ISSUE #3025: UserComponentPane action handlers are broken
This is clear from the code at https://github.com/sanity-io/sanity/blob/d055458a8ff395fe24181400e650916481f78f80/packages/%40sanity/desk-tool/src/panes/userComponent/UserComponentPane.tsx#L34

```js
export function UserComponentPane(props: UserComponentPaneProps) {
  ...
  const userComponent = useRef<{
    actionHandlers?: Record<string, DeskToolPaneActionHandler>
  } | null>(null)
  ...
  return (
    <Pane ...>
      <UserComponentPaneHeader
        actionHandlers={userComponent.current?.actionHandlers}
        ...
      />
      <UserComponentPaneContent>
          ...
          createElement(component, {
            ...
            ref: userComponent,

```

The only way the `UserComponentPaneHeader` receives the correct `actionHandlers` is when this component is rerendered after the `ref` has received the `actionHandlers`.

Base rule: Never pass `ref.current` to another component, it will not rerender if changed.

How to solve? There are various ways to solve this problem. Without changing the signature of `UserComponentPaneHeader` it becomes a bit tricky.

One method might be to use a `Proxy` that will simply try to see if `userComponent.current.actionHandler[action]` can be called.

Workaround, unknown. I need to try to find a way to rerender the `UserComponentPane` from my component.

ricokahler:
### Description

Refactors `UserComponentPane` to `useState` instead of `useRef` so that `UserComponentPaneHeader` can re-render when the component does.

Closes #3025

### What to review

Try out a custom component pane and see if it still works. Then try changing the component instance with a different set of action handlers. It should re-render.

### Notes for release

Fixes a bug that caused some action handlers to never bind.

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/userComponent/UserComponentPane.tsx
+++ b/packages/@sanity/desk-tool/src/panes/userComponent/UserComponentPane.tsx
@@ -1,4 +1,4 @@
-import React, {createElement, isValidElement, useRef} from 'react'
+import React, {createElement, isValidElement, useState} from 'react'
 import {isValidElementType} from 'react-is'
 import {Pane} from '../../components/pane'
 import {usePaneRouter} from '../../contexts/paneRouter'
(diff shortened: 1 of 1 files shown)


rexxars (approved):
LGTM!

PR #3087, opened 2022-01-19
Fix permission issue panel

RitaDias:
### Description

Fix an [issue](https://sanity-io.slack.com/archives/CL088LK3K/p1642514167011200) where a user with a role and rule wasn't being allowed to edit a document when they should.

## Edit: Decided to drop the update to the `community-studio`
I've also added an updated version of the `community-studio` to the examples. My plan would have been to add this in a separate PR but it does make testing this fix much easier. So if we want (and likely should), I can drop the commit before pushing the fix so that it's just itself. Especially since there seems to be a lot of things failing because of it.

I don't know if there's a reason / situation where we needed the dummy in this situation. 🤔 

### What to review

1. install dependencies + run `npm run community-studio`
2. use `#_debug_roles=editor` in the url (this is to be able to go to a grant rule where you can edit your own profile)

So here is what needs to happen:
- Going to your user: you should be able to update
image
- Going to any other user: you should **not** be able to update
Screenshot 2022-01-19 at 18 23 05

### Notes for release
- Fix permission issue on update documents in the panel

### DIFF
--- a/packages/@sanity/desk-tool/src/panes/document/documentPanel/DocumentPanel.tsx
+++ b/packages/@sanity/desk-tool/src/panes/document/documentPanel/DocumentPanel.tsx
@@ -64,7 +64,7 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
   const requiredPermission = value._createdAt ? 'update' : 'create'
   const liveEdit = useMemo(() => Boolean(getSchemaType(documentType)?.liveEdit), [documentType])
   const docPermissionsInput = useMemo(() => {
-    return {...value, _id: liveEdit ? 'dummy-id' : 'drafts.dummy-id'}
+    return {...value, _id: liveEdit ? value._id : `drafts.${value._id}`}
   }, [liveEdit, value])
   const [permissions, isPermissionsLoading] = useDocumentValuePermissions({
     document: docPermissionsInput,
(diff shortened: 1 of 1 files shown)


packages/@sanity/desk-tool/src/panes/document/documentPanel/DocumentPanel.tsx
```diff
@@ -64,7 +64,7 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
   const requiredPermission = value._createdAt ? 'update' : 'create'
   const liveEdit = useMemo(() => Boolean(getSchemaType(documentType)?.liveEdit), [documentType])
   const docPermissionsInput = useMemo(() => {
-    return {...value, _id: liveEdit ? 'dummy-id' : 'drafts.dummy-id'}
+    return {...value, _id: liveEdit ? value._id : `drafts.${value._id}`}
```
ricokahler, line 67: we have some helpers to ensure IDs if you think it's appropriate to use them

```
import {getPublishedId, getDraftId} from 'part:@sanity/base/util/draft-utils'
```

ricokahler (approved):
Your bug fix commit looks good! Nice find!

I do think we should hold off on importing the community studio however. It's throw the CI off with deps and I'm not sure it's worth the fight to get it to pass

packages/@sanity/desk-tool/src/panes/document/documentPanel/DocumentPanel.tsx
ricokahler: ah yes, you can import it from `import {getPublishedId} from '@sanity/base/_internal'`

RitaDias: Makes sense, I've updated the code to reflect that 👍

ricokahler (approved):
ship it

PR #3090, opened 2022-01-20
feat(desk-tool): make inspect dialog pane specific

RitaDias:
### Description

Move the inspect dialog to be pane specific and allow for a user to use the actions in the footer
image

### What to review

- Open references in place and try to open the inspect dialog in both
- Change the tabs between the two
- Try to publish changes / revert changes / duplicate actions in the footer

### Notes for release

- Make inspect dialog pane specific and allow footer actions to be used when it is open


mariuslundgard (approved):
Nice, looks great ✨

PR #3095, opened 2022-01-24
Fix glitch in placeholder rendering

skogsmaskin:
Check that the leaf text is really empty before displaying the placeholder as the decoration array is lagging behind typing.


hermanwikner (approved):
Nice! 👍

PR #3098, opened 2022-01-26
fix(portable-text-editor): fix bug with insertNodePatch

skogsmaskin:
### Description

Fixes a bug introduced in d4705d406f52bea868ede20b3a48416f5de3770d where an error would be thrown when producing a insertNode patch.

Also added tests for this.


bjoerge (approved):

RitaDias (approved):
Let's go ✨

PR #3099, opened 2022-01-26
Add more tests and fix bug in insert blocks into empty editor

skogsmaskin:
### Description

This will add more tests for the `operationToPatches` functions.

Also, refactor two block determinators to be more spesific (test that it's actually text blocks we are dealing with).

### Notes for release

Doesn't need to be mentioned as this doesn't really change anything.


RitaDias (approved):
Linting is failing but apart from that 🟢

PR #3100, opened 2022-01-26
fix(form-builder): prevent race condition crashing the studio in arrays

sjelfull:
### Description

This PR fixes a race condition where required array input with primitives some times crashes while a document is being published/deleted.

Theory is that the code is assuming `value` is always an array, but in this particular case it is temporarily undefined while the prop is being updated.

### What to review

- That it makes sense to fallback to a empty array in this particular case

### Notes for release

- Fix a race condition where required arrays with primitive values some times crashes while a document is being published/deleted.


bjoerge (approved):
Great catch!

PR #3101, opened 2022-01-26
fix(form-builder): fix issue with editing inline object references in PTE

skogsmaskin:
There was an issue where if the input of a inline object in the Portable Text Editor set an empty focusPath,
it would close the object editing interface as the editor would interpret that as focus on the editor node itself.

In these cases, make sure we end the focusPath with the focus terminator symbol.

### What to review
That editing inline objects of type Reference is possible.

Create a inline object of type reference and set it to an actual reference. Close the editing interface and try to open it again. 

### Notes for release
Fixes a bug in the Portable Text Editor where opening inline objects of type reference would not open the editing interface if set already.


hermanwikner (approved):
Awesome, works great! 💯

PR #3103, opened 2022-01-27
chore(client): move client out of monorepo

rexxars:
### Description

This PR removes the `@sanity/client` package from the monorepo, as it now lives over here: https://github.com/sanity-io/client

In addition, it upgrades all our modules to use the new `^3.0.1` range (not currently tagged as `latest`, but will be once we give it some more testing within the studio). The reason for not pinning the version is because it now lives outside of the monorepo, and won't automatically update on changes. That means we either have to regularly keep it up to date with a pinned version, or simply rely on ourselves not making breaking changes.

I think we've done an excellent job at not breaking its API in the past, so I feel like a version range is the rational choice here - it is, after all, what our users will use.

In getting the new version of the client working, I discovered a typing bug in the new version, which incidentally uncovered that the document availability code depended on an `any`-typed response, so I took the liberty of improving the typing there while I was at it - apologies for the slightly unrelated commit.

### What to review

That nothing has "dangling" references to the client living within the monorepo, somehow, and that the studio still works as expected.

### Notes for release

- Upgraded the studio to use `@sanity/client` version 3.0.1


bjoerge (approved):
Great work! ✨

PR #3105, opened 2022-01-27
fix(form-builder): skip initial value resolution when inserting non-empty array items

bjoerge:
### Description
When duplicating an array item, an item's initialValues currently take precedence over the duplicated data. This wasn't  intentional and this patch fixes it.

### What to review
There's an example of an array of items with initial values here: https://test-studio-git-bug-sc-13780array-of-objects-initial-val-9183e9.sanity.build/test/desk/input-standard;arraysTest;8135c43b-9f28-46bb-8d87-3207901dced4%2Cpath%3DarrayWithInitialValue
 - Make sure initial value is *not* inserted when duplicating an item
 - Make sure initial value is still used when adding new items

### Notes for release

- Fixes a bug that caused "duplicate array item" to insert the initial value instead of a duplicate of the chosen item

### DIFF
--- a/dev/test-studio/schema/standard/arrays.js
+++ b/dev/test-studio/schema/standard/arrays.js
@@ -115,6 +115,24 @@ export default {
       ],
     },
     {
+      name: 'arrayWithInitialValue',
+      title: 'Array of types that has initial values defined',
+      type: 'array',
+      of: [
+        {
+          type: 'object',
+          title: 'Dog breed',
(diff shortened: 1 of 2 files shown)


SimeonGriggs:
LGTM!

SimeonGriggs (approved):

RitaDias (approved):
An absolute speed demon 😄

PR #3106, opened 2022-01-27
feat(default-layout): update login status design

jorngeorg:
### Description
This add a couple of things to the login status menu:
* details about the logged in user (name, email and login provider)
* a link to Manage for the project

Users can have multiple users and it can sometimes hard to discern what user you are logged in with. This helps with that.

Secondly, there currently is no obvious way to get to Manage from the studio. This adds a link to take you directly there. 

### What to review
* Try with different login provides. 
* Simulate a long name
* Is there a way to know the organization id upfront, now we need to go through a redirect in manage which adds extra friction


bjoerge (approved):
Nice work!

PR #3111, opened 2022-01-28
fix(portable-text-editor): disallow pointer events on placeholder

skogsmaskin:
The placeholder in the PTE should not take any pointer events. On Webkit this manifested as an issue setting focus in the editor  with some pointer events.

### Description

Even though it has `user-select: none;` it was still possible in Webkit to select this text for some weird reason, making it troublesome to set focus in the editor and start typing afterwards.

### What to review

That you always get focus and can type when using the mouse to set focus in an empty editor.

### Notes for release

Fixed an issue where the empty editor placeholder text sometimes would cause focus problems in Webkit.


hermanwikner (approved):
Nice, works good! 👍

PR #3113, opened 2022-01-28
fix(portable-text-editor): fix paste of listItems

geoffreyjball:
### Description

In certain environments, pasting an unordered list item followed by a paragraph into a portable text field would crash the desk.

The else block was throwing an error as `element.level` is read-only.

### What to review

Confirm that pasting into the portable text field works as expected.

### Notes for release

Fixes a regression when pasting from certain text editors into a portable text field.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Element.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Element.tsx
@@ -132,7 +132,7 @@ export const Element: FunctionComponent<ElementProps> = ({
       if (Number.isInteger(element.level)) {
         renderAttribs.level = element.level
       } else {
-        element.level = 1
+        renderAttribs.level = 1
       }
       className += ` pt-list-item pt-list-item-${element.listItem} pt-list-item-level-${element.level}`
     }


packages/@sanity/portable-text-editor/src/editor/Element.tsx
```diff
@@ -132,7 +132,7 @@ export const Element: FunctionComponent<ElementProps> = ({
       if (Number.isInteger(element.level)) {
         renderAttribs.level = element.level
       } else {
-        element.level = 1
+        renderAttribs.level = 1
       }
       className += ` pt-list-item pt-list-item-${element.listItem} pt-list-item-level-${element.level}`
```
skogsmaskin, line 137: We should be using `renderAttribs.level` on the className here and not `element.level` as it's still without a fallback.

skogsmaskin (changes requested):

skogsmaskin:
@geoffreyjball - I was unable to work with this branch as it's external, so I made a new PR here https://github.com/sanity-io/sanity/pull/3129

PR #3114, opened 2022-01-31
fix(form-builder): fix overflow when there are many types

RitaDias:
### Description

Added `constrainSize` to the MenuGroup popover to limit the overflow of type options that the array input can have

image

### What to review

- Add schema that adds a lot of different types
  - example in `array.js` schema:
 ```
{
      name: 'arrayOfALLTypesPopover',
      title: 'Array of ALL the types',
      type: 'array',
      of: range(0, 60).map((value) => {
        return {
          type: 'image',
          name: value,
        }
      }),
    },
```
- Go to any of the array tests 
- Make sure that regardless of how many types are available, that the menu is readable and scrollable

### Notes for release

Fix bug where when there were multiple types available in the array input "add before / add after" the options didn't scroll

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/InsertMenu.tsx
+++ b/packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/InsertMenu.tsx
@@ -9,7 +9,12 @@ interface Props {
   onInsert: (pos: 'before' | 'after', type: SchemaType) => void
 }
 
-const MENU_POPOVER_PROPS = {portal: true, tone: 'default', placement: 'left'} as const
+const MENU_POPOVER_PROPS = {
+  portal: true,
+  tone: 'default',
+  placement: 'left',
+  constrainSize: true,
+} as const
 
 export const InsertMenu = memo(function InsertMenu(props: Props) {
   const {types, onInsert} = props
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/InsertMenu.tsx [hunk 1]
```diff
@@ -9,7 +9,12 @@ interface Props {
   onInsert: (pos: 'before' | 'after', type: SchemaType) => void
 }
 
-const MENU_POPOVER_PROPS = {portal: true, tone: 'default', placement: 'left'} as const
+const MENU_POPOVER_PROPS = {
```
hermanwikner, line 12: Nice! It would be nice to get typings for this item:

```ts
const MENU_POPOVER_PROPS: MenuGroupProps['popover'] = {
  constrainSize: true,
  placement: 'left',
  portal: true,
  tone: 'default',
}
```

hermanwikner (approved):
Works great! 🙌  I just added a small comment :)

packages/@sanity/form-builder/src/inputs/arrays/ArrayOfObjectsInput/InsertMenu.tsx
RitaDias: Good point! I've added it :)

PR #3115, opened 2022-01-31
Guard that hotkeys for the PTE is a stable object

skogsmaskin:
There will be issues if hotkeys from props are defined inline and is always a new object (they are set up at every render).

Ensure that we only read this prop initially, and warn if the hotkeys are different across renderers.

### Notes for release

It doesn't need to be mentioned.


sonnymai (approved):
Nice one @skogsmaskin !

andrewspinks:
> Nice one @skogsmaskin !

👍

bjoerge (approved):
LGTM!

PR #3116, opened 2022-01-31
chore: externalize + upgrade generate-help-url

rexxars:
### Description

- Moves the generate-help-url module out of the monorepo (it now lives here: https://github.com/sanity-io/generate-help-url)
- Bumped version requirement to v3 and changed from default to named export

### Notes for release

None.


bjoerge (approved):
Nice!

PR #3119, opened 2022-02-02
fix(form-builder): unset image/file field when last subfield is cleared

rexxars:
### Description

When an image has a field on it (e.g., an alt text string field), a certain sequence of events causes the image not to be unset. This poses a problem when that field is required, as it will prevent the document from being published even when the user deletes the field's contents.

The fix is a bit verbose, but leans on the same logic that the asset removal logic uses - eg if an asset is removed and it is the last non-crop/hotspot field with a value, unset the entire object. In this case we check if the patch being emitted from the field is an unset patch (eg the `alt` field wants to unset itself) and if so, whether or not there are any other non-empty values left in the object.

Fixes [sc-12411]

### Steps to reproduce

Schema example:

```
export default {
  name: 'imageTest',
  title: 'Image Test',
  type: 'document',
  fields: [
    {
      name: 'someImage',
      title: 'Some Image',
      type: 'image',
      fields: [
        {
          name: 'alt',
          type: 'string',
          title: 'Alt',
          options: {
            isHighlighted: true,
          },
        }
      ]
    },
  ]
}
```

1. Upload or select an image.
2. Add alt text.
3. Remove the image.
4. Delete the alt text.

You should observe a validation error, and in the JSON, the image field will persist.

### What to review

- That the patch makes sense
- That there isn't a better way of doing this
- Whether or not there could be some abstraction/reusable way to write this (given the same logic goes for both files and images)

### Notes for release

- Fixes an issue where image or file fields with nested, required fields would trigger validation even when image/file field was missing an asset

### DIFF
--- a/dev/test-studio/schema/standard/images.js
+++ b/dev/test-studio/schema/standard/images.js
@@ -45,6 +45,7 @@ export default {
           options: {
             isHighlighted: true,
           },
+          validation: (Rule) => Rule.required(),
         },
         {
           name: 'detailedCaption',
--- a/packages/@sanity/form-builder/src/inputs/files/FileInput/FileInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/FileInput/FileInput.tsx
@@ -273,6 +273,24 @@ export default class FileInput extends React.PureComponent<Props, FileInputState
 
   handleFieldChange = (event: PatchEvent) => {
     const {onChange, type} = this.props
+
+    // When editing a metadata field for a file (eg `description`), and no asset
+    // is currently selected, we want to unset the entire file field if the
+    // field we are currently editing goes blank and gets unset.
+    //
+    // For instance:
+    // A file field with a `description` and a `title` subfield, where the file `asset`
+    // and the `title` field is empty, and we are erasing the `description` field.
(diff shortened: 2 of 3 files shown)


RitaDias (commented):
The code itself looks fine for what is currently the approach.

I've had this discussion with @bjoerge (how there are plenty of things in the File and Image Input that rely on the same - or very similar - things) however, after trying to pull at a thread it became clear that this is the sort of thing that can balloon and so it's best to be handled separately. I say that if we want we can move the logic to a separate file (like in the `files/common`) but I'd be ok with leaving it as you have it for now too.

Small question: If I delete the caption after having written a Detailed caption, the Detailed caption disappears. However, if I write once more in the Caption, the Detailed caption has the same value, is this expected?

bjoerge (commented):
I've got the same remarks as @RitaDias re:shared logic, but also think it's something we can do later. LGTM!

@RitaDias:
> Small question: If I delete the caption after having written a Detailed caption, the Detailed caption disappears. However, if I write once more in the Caption, the Detailed caption has the same value, is this expected?

I noticed the same thing, but realised it's because we're using conditional fields here. So confusing, but believe it's the correct behavior.

bjoerge (approved):

PR #3120, opened 2022-02-02
fix(cypress): use uuid instead of nanoid for document ids

rexxars:
### Description

nanoid can generate IDs that contain characters we dont allow in document IDs, such as a leading dash. This PR replaces it with `@sanity/uuid` for "safe" IDs.

### What to review

- That the code still runs as expected

### Notes for release

None.


hermanwikner (approved):
Nice, good to know! 👍

PR #3124, opened 2022-02-02
fix(portable-text-editor): fix normalization bug in adding marks + test

skogsmaskin:
### Description

Fixed a bug where adding marks to a portable text block child would  be normalized wrong, and loosing existing marks in some situations. 

Basically the same fix as this commit: a8918d5fed325a639ea4f26cd84486ebc6cb29ec only for `addMark` as well.

Added test for this.

### Notes for release

Fixed a bug where adding marks to a portable text block would cause loosing existing marks in some situations.


hermanwikner (approved):
👍

PR #3126, opened 2022-02-03
feat: improve `CollapseMenu` and add keyboard arrow navigation in PTE toolbar and Navbar tools menu

hermanwikner:
### Description

This pull request adds:
- Improvements to the `CollapseMenu` component.
  - General code improvement
  - Implement custom intersection observer logic and hence get rid of the `react-intersection-observer` dependency
  - Preserve focus on the current focused button when the layout changes from expanded to collapsed state
  - Slightly modified and improved API
- A hook (`useRovingFocus`) that enables adding arrow keyboard navigation. An example of when this hook is useful is in 
the PTE toolbar, where there are several elements combined together, which should have a unified arrow keyboard navigation logic. Usage example:

```ts
const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)

useRovingFocus({
  rootElement: rootElement,
  direction: 'horizontal'
})

<ToolbarRoot ref={setRootElement}>
  <StyleMenu />
  <ActionMenu />
  <InsertMenu />
  <ExpandButton />
</ToolbarRoot>

```

- Arrow keyboard navigation in PTE toolbar with the `useRovingFocus` hook
- Arrow keyboard navigation to the tools menu in the nav bar (`ToolCollapseMenu`) with the `useRovingFocus` hook

### What to review

**CollapseMenu**:
  - General code review
  - Focus should be preserved on the current focused button when the layout changes from expanded to collapsed state
    
 **useRovingFocus**:
  - General code review
  - Can there be any situations that the implementation does not take into account?
  - Are there any risks with using `MutationObserver`?
  - Does the hook handle focus as expected? Test in the workshop story, tools menu in the navbar or in the PTE toolbar
  
  **ToolCollapseMenu**:
  - Does it make sense to add arrow keyboard navigation in the tools menu?

### Notes for release
Add arrow keyboard navigation in PTE toolbar and tools menu

### DIFF
--- a/packages/@sanity/base/src/components/rovingFocus/__tests__/useRovingFocus.test.tsx
+++ b/packages/@sanity/base/src/components/rovingFocus/__tests__/useRovingFocus.test.tsx
@@ -1,0 +1,178 @@
+// eslint-disable-next-line import/no-unassigned-import
+import '@testing-library/jest-dom/extend-expect'
+import {render, fireEvent} from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import React, {useState} from 'react'
+
+import {Button, Card, studioTheme, ThemeProvider} from '@sanity/ui'
+import {RovingFocusProps, useRovingFocus} from '..'
+
+interface TestProps extends Pick<RovingFocusProps, 'direction' | 'initialFocus' | 'loop'> {
+  withDisabledButtons?: boolean
+}
+
+function RenderTestComponent(props: TestProps) {
+  const {direction, loop, withDisabledButtons, initialFocus} = props
+  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)
+
+  useRovingFocus({
+    direction: direction,
+    loop: loop,
+    rootElement: rootElement,
+    initialFocus: initialFocus,
+  })
+
+  return (
+    <ThemeProvider theme={studioTheme}>
+      <Card ref={setRootElement} id="rootElement">
+        <Button text="Test" disabled={withDisabledButtons} />
+        <Button text="Test" />
+        <Button text="Test" disabled={withDisabledButtons} />
+        <Button text="Test" />
+      </Card>
+    </ThemeProvider>
+  )
+}
+
+describe('base/useRovingFocus:', () => {
+  /**
+   * Horizontal direction
+   */
+  it('horizontal direction', () => {
+    const {container} = render(<RenderTestComponent />)
+    const rootElement = container.querySelector('#rootElement')
+    const buttons = rootElement.querySelectorAll('button')
+
+    // Focus button #0 on tab
+    userEvent.tab()
+    expect(buttons[0]).toBe(document.activeElement)
+
+    // Focus button #1 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[1]).toBe(document.activeElement)
+
+    // Focus button #2 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[2]).toBe(document.activeElement)
+
+    // Focus button #3 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[3]).toBe(document.activeElement)
+
+    // Focus button #0 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[0]).toBe(document.activeElement)
+
+    // Focus button #3 on arrow left
+    fireEvent.keyDown(rootElement, {key: 'ArrowLeft'})
+    expect(buttons[3]).toBe(document.activeElement)
+
+    // Focus button #2 on arrow left
+    fireEvent.keyDown(rootElement, {key: 'ArrowLeft'})
+    expect(buttons[2]).toBe(document.activeElement)
+  })
+
+  /**
+   * Vertical direction
+   */
+  it('vertical direction', () => {
+    const {container} = render(<RenderTestComponent direction="vertical" />)
+    const rootElement = container.querySelector('#rootElement')
+    const buttons = rootElement.querySelectorAll('button')
+
+    // Focus button #0 on tab
+    userEvent.tab()
+    expect(buttons[0]).toBe(document.activeElement)
+
+    // Focus button #1 on arrow down
+    fireEvent.keyDown(rootElement, {key: 'ArrowDown'})
+    expect(buttons[1]).toBe(document.activeElement)
+
+    // Focus button #2 on arrow down
+    fireEvent.keyDown(rootElement, {key: 'ArrowDown'})
+    expect(buttons[2]).toBe(document.activeElement)
+
+    // Focus button #3 on arrow down
+    fireEvent.keyDown(rootElement, {key: 'ArrowDown'})
+    expect(buttons[3]).toBe(document.activeElement)
+
+    // Focus button #0 on arrow down
+    fireEvent.keyDown(rootElement, {key: 'ArrowDown'})
+    expect(buttons[0]).toBe(document.activeElement)
+
+    // Focus button #3 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowUp'})
+    expect(buttons[3]).toBe(document.activeElement)
+
+    // Focus button #2 on arrow right
+    fireEvent.keyDown(rootElement, {key: 'ArrowUp'})
+    expect(buttons[2]).toBe(document.activeElement)
+  })
+
+  /**
+   * With disabled buttons
+   */
+  it('with disabled buttons', () => {
+    const {container} = render(<RenderTestComponent withDisabledButtons />)
+    const rootElement = container.querySelector('#rootElement')
+    const buttons = rootElement.querySelectorAll('button')
+
+    // Focus button #1 on tab
+    userEvent.tab()
+    expect(buttons[1]).toBe(document.activeElement)
+
+    // Focus button #3 on arrow right (skips #2 because it is disabled)
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[3]).toBe(document.activeElement)
+
+    // Focus button #1 on arrow right (skips #0 because it is disabled)
+    fireEvent.keyDown(rootElement, {key: 'ArrowRight'})
+    expect(buttons[1]).toBe(document.activeElement)
+  })
+
+  /**
+   * Without loop
+   */
+  it('without loop', () => {
+    const {container} = render(<RenderTestComponent loop={false} />)
+    const rootElement = container.querySelector('#rootElement')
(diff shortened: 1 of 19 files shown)


packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+const EMPTY_ARRAY = []
+
+const MUTATION_ATTRIBUTE_FILTER = ['aria-hidden', 'disabled', 'href']
+
+const FOCUSABLE =
+  'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'
+
+function getFocusableElements(element: HTMLElement) {
+  return [...(element.querySelectorAll(FOCUSABLE) as any)].filter(
+    (el) => !el.hasAttribute('disabled') && !el.getAttribute('aria-hidden="true"')
```
ricokahler, line 14: did you mean `el.getAttribute('aria-hidden') !== 'true'`? i'm not familiar with that usage

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+  ) as HTMLElement[]
+}
+
+export function useRovingFocus(props: RovingFocusProps, dependencies?: unknown[]): undefined {
+  const {direction = 'horizontal', initialFocus, loop = true, pause = false, rootElement} = props
+  const [focusedIndex, setFocusedIndex] = useState<number>(-1)
+  const [focusableElements, setFocusableElements] = useState<HTMLElement[]>([])
+
+  const deps = dependencies ? dependencies : EMPTY_ARRAY
```
ricokahler, line 23: 🍬 i think you could provide a default param instead of the ternary

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+        event.preventDefault()
+        setFocusedIndex((prevIndex) => {
+          const next = (prevIndex + lastFocusableIndex) % focusableLen
+
+          if (!loop && next === lastFocusableIndex) {
+            return prevIndex
+          }
+
+          return next
+        })
```
ricokahler, line 71: this is very tomato-tomato (so feel free to ignore) but you could use `Math.max` and `Math.min` to help handle the non-loop case e.g.

```js
if (!loop) return Math.max(0, prevIndex - 1)
```

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+    return () => {
+      mo.disconnect()
+    }
+  }, [focusableElements, handleSetElements, rootElement, ...deps])
+
+  /**
+   * Set focus on elements in focusableElements depending on focusedIndex
+   */
+  useEffect(() => {
+    ;(focusableElements as HTMLElement[]).forEach((el, index) => {
```
ricokahler, line 120: i don't think you need this cast
```suggestion
    focusableElements.forEach((el, index) => {
```

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+const FOCUSABLE =
+  'a[href], button, input, textarea, select, details, [tabindex]:not([tabindex="-1"])'
+
+function getFocusableElements(element: HTMLElement) {
+  return [...(element.querySelectorAll(FOCUSABLE) as any)].filter(
+    (el) => !el.hasAttribute('disabled') && !el.getAttribute('aria-hidden="true"')
+  ) as HTMLElement[]
+}
+
+export function useRovingFocus(props: RovingFocusProps, dependencies?: unknown[]): undefined {
```
ricokahler, line 18: can you add a JSDoc comment here explaining what this is with links back to the roving focus definition? i think it would help give future readers more context

packages/@sanity/base/src/components/rovingFocus/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface RovingFocusProps {
+  direction?: 'horizontal' | 'vertical'
+  initialFocus?: 'first' | 'last'
+  loop?: boolean
+  pause?: boolean
+  rootElement: HTMLElement | HTMLDivElement | null
```
ricokahler, line 6: for my own knowledge: I see some hook apis use refs and I see some other use elements like you have here? i've been wondering which kind of API is better under certain scenarios

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+   * Set focus on elements in focusableElements depending on focusedIndex
+   */
+  useEffect(() => {
+    ;(focusableElements as HTMLElement[]).forEach((el, index) => {
+      if (index === focusedIndex) {
+        el.setAttribute('tabIndex', '0')
+        el.setAttribute('aria-selected', 'true')
+        el.focus()
+        el.onfocus = () => handleFocus(index)
+        el.onblur = () => handleFocus(-1)
```
ricokahler, line 126: do you need to clean up these event listeners (e.g. with `el.addEventListener` and `el.removeEventListener`)?

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
...
+        el.setAttribute('aria-selected', 'false')
+        el.onfocus = () => handleFocus(index)
+      }
+    })
+
+    if (focusedIndex === -1 && focusableElements) {
+      const initialIndex = initialFocus === 'last' ? lastFocusableIndex : 0
+      focusableElements[initialIndex]?.setAttribute('tabIndex', '0')
+      focusableElements[initialIndex]?.setAttribute('aria-selected', 'true')
+    }
```
ricokahler, line 138: how would you feel about extracting this part into its own `useEffect` that sets `focusedIndex`? thinking something like this:

```js
useEffect(() => {
  if (focusedIndex === -1 || focusableElements) {
    setFocusedIndex(initialFocus === 'last' ? lastFocusableIndex : 0)
  }
}, [focusableElements, initialFocus, lastFocusableIndex])
```

the benefit here is that you get the re-use the logic from your other effect (though it requires an additional render). might be missing something though lmk!

packages/@sanity/form-builder/src/inputs/PortableText/toolbar/Toolbar.tsx
```diff
@@ -80,6 +81,13 @@ const InnerToolbar = memo(function InnerToolbar({
 
   const collapsed = rootElementRect ? rootElementRect?.width < 400 : false
 
+  useRovingFocus(
+    {
+      rootElement: rootElement,
+    },
+    [showActionMenu, showInsertMenu, collapsed]
```
ricokahler, line 88: what happens when you pass dependencies here? what does it run when they change?

ricokahler (commented):
this is looking really good! i think your approach with the mutation observer is definitely the way to go!

i think i might've caught something with event handler clean up and  `getAttribute` and i had a few other questions in there and suggestions for docs. feel free to DM if i'm confusing 😄

packages/@sanity/base/src/components/rovingFocus/types.ts
hermanwikner: Not sure actually. For me, it feels easiest to just pass the element directly. But I am not sure if it's the best approach. We use that approach for hooks in `@sanity/ui` such as `useClickOutside` (pass the element, not the ref). :)

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
hermanwikner: Good idea! However, we can't `setFocusedIndex` in the `useEffect` you're suggesting, because that will make the first/last button take the focus on mount. We only wan't to control the `tabIndex` to make it focusable when you tab to it.

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
hermanwikner: Hm, interesting, not sure actually! 👀

packages/@sanity/form-builder/src/inputs/PortableText/toolbar/Toolbar.tsx
hermanwikner: Good point. I thought it would be needed, but I think I was wrong. I can remove these dependencies!

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
hermanwikner: Yes, definitely, will do 👍

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
hermanwikner: Your suggestion is better – will update with `el.getAttribute('aria-hidden') !== 'true'` ! 👍

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+import {useCallback, useEffect, useState} from 'react'
+import {RovingFocusProps} from '.'
```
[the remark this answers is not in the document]
hermanwikner: Good point 🙂

packages/@sanity/base/src/components/rovingFocus/__tests__/useRovingFocus.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+        <Button text="Test" disabled={withDisabledButtons} />
+        <Button text="Test" />
+        <Button text="Test" disabled={withDisabledButtons} />
+        <Button text="Test" />
+      </Card>
+    </ThemeProvider>
+  )
+}
+
+describe('base/useRovingFocus:', () => {
```
ricokahler, line 37: these tests are great btw! i'm learning things reading them as well 😎

packages/@sanity/base/src/components/rovingFocus/useRovingFocus.ts
```diff
@@ -0,0 +1,171 @@
...
+        <div ref={setRootElement}>
+          <button>Button</button>
+          <button>Button</button>
+          <button>Button</button>
+        </div>
+      )
+  }
+  ```
+ */
+export function useRovingFocus(props: RovingFocusProps, dependencies: unknown[] = []): undefined {
```
ricokahler, line 39: do you think it's safe to remove the `dependencies` argument now?

ricokahler (approved):
looks great! I had one comment around the dependencies argument but that's about it. well done!

PR #3130, opened 2022-02-03
fix(form-builder): hide array actions if read-only

sjelfull:
### Description

This hides the array actions for row items if array is read-only.

Clipboard 2022-03-02 at 2 32 35 PM

### What to review

- That we are happy with UI

### Notes for release

- Fixed a bug where read-only arrays showed a empty actions menu


bjoerge (approved):
LGTM!

PR #3131, opened 2022-02-03
fix(form-builder): improve empty read-only array state

sjelfull:
### Description

This improves the state of read-only, empty arrays by showing a disabled Add button w/ an explanatory tooltip that tells the user that this field is read-only.

Before you would just see the title/description of a field with no array UI.

|Before|After|
|-|-|
||image|

### What to review

- UI

### Notes for release

- Improved empty read-only array state


bjoerge (approved):
Nice, LGTM!

PR #3134, opened 2022-02-04
Allow `null` Tools for conditional rendering

SimeonGriggs:
This ended up being _more_ than a one liner so I understand if it gets rejected / requires more discussion for a better approach!

### Description

Allows for Tools to be 
1. Conditionally rendered and 
2. Conditionally registered to the Router. 
3. 
4. This way a Tool may only appear in the Navbar and loaded if it satisfies a particular condition like current Dataset or a User's Role. 

For example:

```js
export default isTestDataset
  ? {
      title: 'Conditional Tool',
      name: 'test-conditional-tool',
      component: ConditionalTool,
    }
  : null
```

Changes have been made to no longer assume all registered Tools are objects and will filter out any _falsy_ Tools.

### What to review

In the test studio, `private` dataset:

* http://localhost:3333/playground/conditional-tool
* Should not render component
* "Conditional Tool" should not be in navbar

In the test studio, `test` dataset:

* http://localhost:3333/test/test-conditional-tool
* Should load component
* "Conditional Tool" should be in navbar

### Notes for release

Desk Tools can be conditionally rendered

### DIFF
--- a/dev/test-studio/parts/tools/test-conditional-tool.tsx
+++ b/dev/test-studio/parts/tools/test-conditional-tool.tsx
@@ -1,17 +1,24 @@
 import React from 'react'
-import {Container, Box, Text} from '@sanity/ui'
+import {Container, Stack, Text} from '@sanity/ui'
+import {CheckmarkIcon} from '@sanity/icons'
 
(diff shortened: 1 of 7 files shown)


rexxars (approved):
Did some minor tweaks, but this looks alright to me.

PR #3135, opened 2022-02-04
fix(form-builder): ensure asset sources is array before usage

rexxars:
### Description

The part names and export shapes are easy to get wrong, eg the distinction between `part:@sanity/form-builder/input/file/asset-source` and `part:@sanity/form-builder/input/file/asset-sources` is very subtle. If you happen to implement `part:@sanity/form-builder/input/file/asset-sources` and it exports a single asset source, the studio crashes in spectacular ways.

This PR checks that the part exports an array, and if not warns and uses the default asset sources instead. 

I also took the liberty of fixing a ref warning/error that was being thrown when opening the default asset source while I was at it, sorry for the slightly unrelated commit.

### Notes for release

- Added validation and warnings on incorrectly implemented asset sources part

### DIFF
--- a/dev/test-studio/parts/assetSources/noop.tsx
+++ b/dev/test-studio/parts/assetSources/noop.tsx
@@ -1,18 +1,20 @@
 import React from 'react'
 
+interface Props {
+  onClose: () => void
+}
+
 export default {
   name: 'noop-asset-source',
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Nice!

PR #3139, opened 2022-02-08
feat: changelog dialog

hermanwikner:
### Description

This PR replaces the previous dialog that listed outdated packages with a new dialog that instead displays the studio changelog. The changelog data being displayed in the dialog is authored in the internal admin-studio.

Screenshot 2022-02-23 at 15 09 44

**UI components**
The UI components are created in `@sanity/default-layout` under the `update` directory, and can be tested in isolation in the workshop environment.

**Changelog data**
The changelog data is being returned from the hook called `useModuleStatus`. This hook communicates with a proxy that handles logic to return:

- The `currentVersion` of the studio
- The `latestVersion` of the studio
- The `changelog` entries from `currentVersion` to (and including) `latestVersion`. Fetched from the admin-studio.

The `Navbar` has a new `ChangelogContainer` component that handles the data returned from the `useModuleStatus` hook – which it then passes on to UI components. The `ChangelogContainer` returns `null` if the request has error, is loading or returns no changelog data.

**How to test**
- Start the studio with `SANITY_INTERNAL_ENV=staging npm run dev` to run against staging api:s
- Navigate to `moduleStatus.ts` and set the `fakeOutdatedModule` constant to `true`. This will simulate that the studio is not up-to-date with the latest version.

### What to review

- Is there any copy that should be updated in the dialog to make it clearer what is actually displayed?
- Do we need to add pagination to the `ChangelogDialog` component? Alternatively render the items in a `VirtualList`? 
- Is there any test that could be written for this implementation? A test have been written for the logic in the proxy that determines current and latest versions.
- Does the `ChangelogDialog` display the expected changelog data?
- Does the filtering of changelog data in `ChangelogContainer` make sense?
- General code review

### Notes for release

Add changelog dialog


packages/@sanity/base/src/module-status/types.ts
```diff
@@ -1,5 +1,19 @@
 import type {SanityClient} from '@sanity/client'
 
+export interface ChangeItem {
+  changeType: 'bugfix' | 'feature'
+  description: any
```
rexxars, line 5: We should probably type this, as it is quite hard to reason about the shape without knowing what is in it?

Rough untested sketch based on what I can tell from the shape:
```ts
import type {SanityImageHotspot, SanityImageCrop, SanityImageDimensions} from '@sanity/asset-utils'
import type {PortableTextBlock} from '@portabletext/types' // Remember to add this one as a dependency!

interface CodeBlock {
  _type: 'code'
  code?: string
  language?: string
  filename?: string
  highlightedLines?: number[]
}

interface MaterializedImage {
  _type: 'image'
  alt?: string
  caption?: string
  hotspot?: SanityImageHotspot
  crop?: SanityImageCrop
  asset?: {
    _id: string
    path: string
    metadata: {
      dimensions: SanityImageDimensions
      lqip?: string
    }
  }
}

type StudioBlockContent = 
  | PortableTextBlock
  | CodeBlock
  | MaterializedImage
```

packages/@sanity/default-layout/src/update/PortableTextContent/PortableTextContent.tsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react'
+import {PortableText, PortableTextComponents} from '@portabletext/react'
+import {Box, Card, Code, Heading, Stack, Text} from '@sanity/ui'
+import urlBuilder from '@sanity/image-url'
+import {getProject} from '@sanity/asset-utils'
+import {ImageCard, ListBox, ListBoxOuter, RootBox, SpaceBox} from './PortableTextContent.styles'
+
+interface PortableTextContentProps {
+  value: any
```
rexxars, line 9: If we add typings for the block content shape we should make sure this one is also typed correctly

packages/@sanity/default-layout/src/update/PortableTextContent/PortableTextContent.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function Image({value}) {
```
rexxars, line 12: ...and if typed, we'd get the right type here - `MaterializedImage` (or whatever we want to call it)

rexxars (changes requested):
Great work! A few minor typing requests from my end, but otherwise good!

packages/@sanity/base/src/module-status/types.ts
```diff
@@ -1,5 +1,19 @@
 import type {SanityClient} from '@sanity/client'
 
+export interface ChangeItem {
```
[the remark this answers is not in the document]
hermanwikner: Yes, good idea to make it clearer what "change" this type is referring to 👍

rexxars (approved):
👍

mariuslundgard (approved):
Looks amazing – great work @hermanwikner.

One thing we should adjust before merging:
If the API returns no changelog items, but the Studio is outdated, we should still show the update status dialog:
image

PR #3144, opened 2022-02-09
build: add better release notes script

ricokahler:
### Description

I've attempted this before, but I think I got it this time. This PR adds a more automated release notes scripts.

How it works:

1. The PR template has much more clear instructions as well symbols it can use to parse info out of it
2. The release script goes through each commit and grabs the associated PR
3. It parses the bodies of all the PRs and returns a formatted release notes starter. This starter can then be edited by editorial like normal.

I also threw in a style guide for the release notes to make it more clear how to write these things (though it could use some improvement)

### What to review

- The script itself
- The updated PR template
- The new styled guide

### Notes for release

N/A


rexxars:
Will revisit this in the future - ideally we want to create a draft changelog entry in Sanity + have a webhook trigger creation/updates against the GitHub releases API based on that source data, in order to have a single source of truth. Leaving the branch since we'll reuse a lot of the parsing logic here!

PR #3145, opened 2022-02-10
chore: use playwright for e2e tests

vicmeow:


playwright/tests/desk-tool/documentActions.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+const DOC_TITLE = 'Document actions [Playwright]'
+const inputSelector = 'data-testid=input-title >> input'
+
+// The path of the network request we want to wait for
+const documentActionPath = (action) =>
+  `*/**/mutate/${DEFAULT_DATASET}?tag=sanity.studio.document.${action}*`
+
+// Helper to generate path to document created in a test
+const generateDocumentUri = (docId) =>
+  `/${DEFAULT_DATASET}/desk/input-ci;documentActionsCI;${docId}`
```
rexxars, line 18: The first part here isn't actually the dataset name, but rather the name of the space - maybe we can define a separate constant for it?

```suggestion
const generateDocumentUri = (docId) =>
  `/${STUDIO_SPACE_NAME}/desk/input-ci;documentActionsCI;${docId}`
```

playwright/helpers/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_DATASET = 'test'
```
rexxars, line 1: Maybe this would be an acceptable place to put a unique run ID as well?
And we will want to have the space name be defined in a single place, too.

```suggestion
import {uuid} from '@sanity/uuid'

export const TEST_RUN_ID = uuid()
export const STUDIO_SPACE_NAME = 'test'
export const DEFAULT_DATASET = 'test'
```

playwright/helpers/createUniqueDocument.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function createUniqueDocument({
+  _type,
+  _id,
+  ...restProps
+}: SanityDocumentStub): Promise<Partial<SanityDocument>> {
+  const doc = {
+    _type,
+    _id: _id || uuid(),
+    ...restProps,
```
rexxars, line 13: Could we add a unique ID for this test run here, for further isolation? We should probably add it to the CI schemas as a hidden string field as well, if we do.

```suggestion
    ...restProps,
    runId: TEST_RUN_ID,
```

playwright/helpers/sanityClient.ts
```diff
@@ -0,0 +1,13 @@
...
+export const testSanityClient = sanityClient({
+  projectId: 'ppsg7ml5',
+  dataset: 'test',
+  token: process.env.PLAYWRIGHT_SANITY_SESSION_TOKEN,
+  useCdn: false,
+})
+
+export async function deleteDocument(payload) {
+  return testSanityClient.delete(payload)
+}
```
rexxars, line 13: I feel like we can just as well use `testSanityClient.delete()` directly to get more control of the operation (second argument, for instance), or have this function do something more specific, eg `function deleteDocumentsForRun(typeName: string)`, which could generate a GROQ query with the correct type name and run ID. Thoughts?

playwright/tests/desk-tool/documentActions.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+async function getDocumentUri(_type: string, title?: string) {
+  const testDoc = await createUniqueDocument({_type, title})
+  return [generateDocumentUri(testDoc?._id), testDoc?._id]
+}
+
+test.describe('@sanity/desk-tool: document actions', () => {
+  // Before we run our tests, delete all previously created test documents to avoid clutter
+  test.beforeAll(async () => {
+    await deleteDocument({query: `*[_type == "documentActionsCI"]`})
+  })
```
rexxars, line 30: I feel like this is the wrong place to be doing a cleanup, and that we should rather clean up only this current test run when complete. What do you think? Could we change this to a `afterAll` with a more narrowly scoped query?

psuedo-code:

```suggestion
  // After the tests have been run, try to clean up the documents we created in this run, along with any documents of this type older than the defined threshold (to account for cancelled/failed tests)
  test.afterAll(async () => {
    const threshold = new Date(Date.now() - STALE_TEST_THRESHOLD_MS).toISOString()
    await deleteDocument({query: `*[_type == "documentActionsCI" && (runId == "${TEST_RUN_ID}" || _createdAt < "${threshold}")]`})
  })
```

playwright/tests/desk-tool/documentActions.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// Helper to generate path to document created in a test
+const generateDocumentUri = (docId) =>
+  `/${DEFAULT_DATASET}/desk/input-ci;documentActionsCI;${docId}`
+
+// Helper to get the path to the document the client creates for a test
+async function getDocumentUri(_type: string, title?: string) {
+  const testDoc = await createUniqueDocument({_type, title})
+  return [generateDocumentUri(testDoc?._id), testDoc?._id]
+}
```
rexxars, line 24: Maybe this function name could be more clear about the fact that it creates a document? And maybe it could return an object so it is easier to know what the properties returned represent?

```suggestion
async function createDocument(_type: string, title?: string): Promise<{studioPath: string; documentId: string}> {
  const testDoc = await createUniqueDocument({_type, title})
  return {studioPath: generateDocumentUri(testDoc?._id), documentId: testDoc?._id]}
}
```

playwright/tests/desk-tool/documentActions.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  test.beforeAll(async () => {
+    await deleteDocument({query: `*[_type == "documentActionsCI"]`})
+  })
+
+  test('edit and publish document', async ({page}) => {
+    const publishButton = page.locator('data-testid=action-Publish')
+
+    await getDocumentUri('documentActionsCI').then(([path]) => {
+      page.goto(path)
+    })
```
rexxars, line 37: With my above suggestion, this would be something like:

```suggestion
    const {studioPath} = await createDocument('documentActionsCI')
    await page.goto(path)
```

Currently it does not return the `goto`, which I assume is a promise

rexxars (changes requested):

playwright.config.ts
```diff
@@ -0,0 +1,102 @@
...
+    // {
+    //   name: 'Google Chrome',
+    //   use: {
+    //     channel: 'chrome',
+    //   },
+    // },
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  outputDir: 'test-results/',
```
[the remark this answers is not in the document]
vicbergquist: Yeah, that's a lot better 👍

playwright/tests/desk-tool/documentActions.spec.ts
vicbergquist: Yeah I thought about that, but isn't it nice to see the documents that were created after you've run your tests? In case you need to debug something 🤔

rexxars: Potentially, but you can't do that with the current code if someone (or some CI) runs the test suite before you have time to debug, as it will delete all documents of that type.

We need some sort of isolation. I guess the question is whether or not we should clean up documents at all, and if so, in what way. Maybe the [`testInfo.errors` property in `afterAll`](https://playwright.dev/docs/api/class-test#test-after-all) contains whether or not the suite failed? If so, maybe we could use that to keep failed test runs around, or something? 

If not we should probably just remove the cleanup for now.

playwright/tests/default-layout/navbar.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    const globalCreateButton = page.locator('data-testid=default-layout-global-create-button')
+    const createDialog = page.locator('data-testid=default-layout-global-create-dialog')
+    const paneHeader = page.locator('data-testid=pane-header')
+
+    await Promise.all([
+      globalCreateButton.click(),
+      expect(createDialog).toBeVisible(),
+      expect(createDialog).toHaveCSS('z-index', '500801'),
+      expect(paneHeader).toHaveCSS('z-index', '201'),
```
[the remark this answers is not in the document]
vicbergquist: Yeah, I can do that! This is just how it was written in cypress

playwright/tests/desk-tool/documentActions.spec.ts
vicbergquist: Yeah we can use `testInfo` to determine whether we should delete a document or not, will add! Great idea 👍

rexxars:
Closing this as #4143 has now landed, but will leave the branch around in case we might need it.

PR #3146, opened 2022-02-10
fix(portable-text-editor): fix spellcheck defaults issue

skogsmaskin:
In order to mitigate the Chrome v.96 spellcheck perf issue, spellcheck control was moved from the outer container of the PTE to just be active on the actual text blocks.

Since this is not longer a issue, go back to outer container spellcheck control.

This will also fix a bug that was introduced not respecting the browser default settings when spellcheck is undefined
on the schema type setting it to not enabled in that case.

### Notes for release
Fixed an issue where spellchecking in the Portable Text Editor would be disabled when not defined.


geoffreyjball (commented):
Looks good!

julesjazz:
That looks great, thank you!

geoffreyjball (approved):

PR #3147, opened 2022-02-10
Update createWithSchemaTypes.ts

jyk2000:
### Description
editor.isTextBlock() fails when blocks including ListItemBlock which does not require 'style'
But in this plugin, it's forcing to have style causing errors on multiple location 
ex) When deleting text in ListItemBlock, removeTextPatch throws error due to this bug. 

### What to review

Create document with portable text field. 
Add ListItemBlock with external links in it. 
And try to delete text with back space. Then you'll see error child not found w/o this patch. 

### Notes for release

"style" property is not required in TextBlock.


skogsmaskin:
Hi @jyk2000, first off thanks for contributing! :tada: 

`.style` is a required prop for TextBlocks (also list blocks) in the type checks, so this probably not the fix we are looking for to your problems.

Could you give me a clearer example of what / how it is failing, because it's not failing for me following your descriptions.

rexxars:
Closing until we have more context/information.

PR #3156, opened 2022-02-18
fix(base): fix collapse text option in `CollapseMenu`

hermanwikner:
### Description

The refactored `CollapseMenu` had a minor issue with the `collapseText` option. The `collapseText` option controls whether the text is to be hidden or displayed when the menu is "collapsed". In the PTE toolbar we do not want to collapse the text for the `InsertMenu` – which did not work after the refactor work. This PR adds a fix to this!

Screenshot 2022-02-18 at 16 18 18

### What to review

The `InsertMenu` in PTE toolbar should always show the text on buttons.

### Notes for release

Fix `collapseText` option in `CollapseMenu`


skogsmaskin (approved):
👍

PR #3162, opened 2022-02-22
fix(desk-tool): catch 404 and replace with empty result

ricokahler:
### Description

@bjoerge after some thinking, I think we actually want to handle the 404 instead of passing it the draft ID.

Since our drafts model never allows you to reference the prefixed `drafts.doc-id`, it wouldn't make sense to try to fetch references for it, even if the unprefixed `doc-id` doc hasn't been created yet.

### What to review

Is the logic right? Does it catch and re-throw appropriately?

### DIFF
--- a/packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts
+++ b/packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts
@@ -1,7 +1,8 @@
 import documentStore from 'part:@sanity/base/datastore/document'
 import client from 'part:@sanity/base/client'
+import {ClientError} from '@sanity/client'
 import {createHookFromObservableFactory, getPublishedId} from '@sanity/base/_internal'
-import {Observable, timer, fromEvent, EMPTY, from} from 'rxjs'
+import {Observable, timer, fromEvent, EMPTY, from, of} from 'rxjs'
 import {
   map,
   startWith,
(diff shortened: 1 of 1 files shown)


packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts
```diff
@@ -92,16 +101,27 @@ function fetchCrossDatasetReferences(
...
+                  .join(','),
+              }
+            : null),
+        })
+        .pipe(
+          catchError((e) => {
+            // it's possible that referencing document doesn't exist yet so the
+            // API will return a 404. In those cases, we want to catch and return
+            // a response with no references
+            if (isClientError(e)) return of({totalCount: 0, references: []})
```
bjoerge, line 121: Should probably check for `statusCode == 404` here as well?

packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts
```diff
@@ -92,16 +101,27 @@ function fetchCrossDatasetReferences(
...
+            throw e
```
bjoerge, line 122: This is the way to rethrow

Alternatively you could import `throwError` from `rxjs` and return `throwError(e)`, but it should be pretty much equivalent (the stack trace will be preserved in both cases afaik), so I'd prefer `throw`

packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts
ricokahler: ah this is good. nice catch!

bjoerge (approved):
Nice 🎉

PR #3166, opened 2022-02-25
feat(default-login): introduce dual auth mode

rexxars:
### Description

This builds upon #3155 but introduces a new "dual" authentication mode that attempts both a cookie and falls back to a SID in the case the cookie did not "stick". More details on internal Slack.

### Notes about changes

- I moved the auth and sid checks into `@sanity/base` (user store), as it was easier to provide a "fallback to token exchange" flow within that location, rather than outsourcing it to the default login package. Maybe a bit of concern mangling, but feels acceptable to me.
- The configuration changed from an array to a string (`dual`, `cookie`). I don't see any reason why anyone would want to disable cookies, and the flow would be a lot more complicated, so I felt this was an acceptable change.
- The cookie test utility is no longer needed, as I see it. Thus I removed it. 
- I rewrote the login wrapper in typescript using a functional component - since most of the logic is now gone and we have a current user hook it seemed like a good idea to me
- Removed the auth token exports from `@sanity/base` since all that logic is now contained within the module itself


skogsmaskin (changes requested):
Looks good! However, we should get the authentication state window sync in place again (which previously was enabled/disabled through that constant in the old JS version for LoginWrapper). As we have moved much of the logic  to the user store now it should probably be built into that. Also it will be easier now to have this sync work with both token and cookie logins.

rexxars:
> I don't think that the loginMethods are working as expected. 🤔 I tried setting it to cookie and I'm still being able to login in safari / firefox when it shouldn't.

Did you use `loginMethod` or `loginMethods`? I changed the name to reflect that it is not an array anymore but rather a setting (eg either `dual` or `cookie`)

rexxars:
> However, we should get the authentication state window sync in place again (which previously was enabled/disabled through that constant in the old JS version for LoginWrapper).

I have now reintroduced this. In order to cut down on the number of re-fetch operations and back-and-forth messages, I expanded the shape of the message to include a user ID - this way we can compare the local state against the remote, and only re-fetch if there is a difference. Theoretically I suppose we could serialize the entire user and send it, skipping the re-fetch, but I feel this is slightly safer for now.

This now works with both cookies and tokens - it is simply based on the current user state.

> I don't think that the loginMethods are working as expected. 🤔 I tried setting it to cookie and I'm still being able to login in safari / firefox when it shouldn't.

Did you use `loginMethod` or `loginMethods`? I changed the name to `loginMethod` in order to reflect that it is not an array anymore but rather a switch (eg either dual or cookie). Seems to work for me.

RitaDias:
> Did you use `loginMethod` or `loginMethods`? I changed the name to `loginMethod` in order to reflect that it is not an array anymore but rather a switch (eg either dual or cookie). Seems to work for me.

You're right 8) I didn't notice the missing s. It works 🙇

RitaDias:
However,  I don't seem to be able to get it working on older studios that don't have the `loginMethod` property. Before we had a fallback in case if it didn't have the `loginMethod` in the `default-login` it would behave as it would if it had `dual`. I think we are missing that here!

skogsmaskin (commented):
[WRONG RADIO BUTTON CLICKED]

skogsmaskin (approved):
Excellent!

Issue #3173, opened 2022-02-28
documentTypeList().apiVersion() has no effect in structure builder


barbogast:
Hi all,

I'm trying to set the `apiVersion` for a `filter` in the structure builder:
```ts
 S.documentTypeList('character')
  .id('characters')
  .filter(
    '_type == "character" && _id in *[_id == $bookId].characters[]._ref'
  )
  .params({bookId})
  .apiVersion('2022-02-09')
```

(Side note: The filter query seems to not work with API version "v1". Strange, because it doesn't look too special. Not the main issue, though.)

My problem is that the studio still sends the request with "v1", which doesn't return any result, so apparently using `apiVersion` doesn't have any effect.

I tried to dig through the code, and maybe the issue is that the `apiVersion` information arrives in [DocumentListPane](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/desk-tool/src/panes/documentList/DocumentListPane.tsx#L46) but doesn't get passed on into `useDocumentList()`?

-----
```
≻ sanity versions
@sanity/cli                   2.27.0 (up to date)
@sanity/base                  2.27.1 (up to date)
@sanity/core                  2.27.0 (up to date)
@sanity/default-layout        2.27.1 (up to date)
@sanity/default-login         2.27.0 (up to date)
@sanity/desk-tool             2.27.1 (up to date)
@sanity/eslint-config-studio   2.0.0 (up to date)
@sanity/vision                2.27.1 (up to date)
```

- MacOS 12.1
- node 16.13.2
- yarn 1.22.17

barbogast:
Yeah, passing `apiVersion` into `useDocumentList` and from there into `getQueryResults` fixes the issue.

PR #3174, opened 2022-02-28
Fix api version

barbogast:
This fixes `documentTypeList().apiVersion()` not having any effect on outgoing API requests.

Fix for #3173


barbogast:
Note that this PR is _not_ ready for merging, I just wanted to point out what I did locally in my `node_modules` folder.

I think Typescript will complain as `DocumentListPaneNode` has no `apiVersion` (the data is there, though).

rexxars (approved):
Thanks for your contribution! I've fixed the few typing issues that was preventing this from building.

PR #3176, opened 2022-03-03
fix(cli): improve `sanity upgrade` tag/range handling

rexxars:
### Description

This PR attempts to address a number of annoyances/bugs with the `sanity upgrade` command:

1. The command currently fails when using `--tag` altogether.
    - Fixed by upgrading `get-latest-version` and returning `undefined` instead of `'unknown'` when a package is not found with the given range. If we can't find it, we stick with the version as-is.
2. When using `--tag`, we incorrectly add a `^` prefix to the tag in `package.json`, unless `--save-exact` is specified.
    - Fixed by using the tag or range given as-is. When using `--save-exact`, we store the actual, resolved version numbers - otherwise we just store the tag/range.
3. The command currently fails if any of the declared packages are not installed when running `sanity upgrade`
    - Fixed by additional checks, try/catching and adding `--check-files` flag when running yarn
4. When upgrading to a tagged release, it would always see the versions as "new", because of a mismatch between the version on npm and the version in `package.json` locally (a `+hash` postfix locally, vs none remotely)
    - Fixed by stripping the hash when comparing
5. Currently you can specify specific modules to upgrade. This will in the vast majority of cases break things, as you will end up with conflicting/multiple versions of the same modules as dependencies.
    - "Fixed" by removing this option. I cannot picture anyone having used this successfully.

In addition to this, I took the liberty of improving the documentation for the command with examples.

Fixes [sc14620]

### How to review

You will probably want to go through the changes commit by commit - is is more digestable that way. 

To run the command, you will want to check out the branch and run `npm run build`, then from another studio run:
`~/path/to/sanity-checkout/packages/@sanity/cli/bin/sanity upgrade <whatever options you want to test>`

### Notes for release

Fixes a number of issues with the `sanity upgrade` command when using `--tag` or `--range`

### DIFF
--- a/packages/@sanity/cli/package.json
+++ b/packages/@sanity/cli/package.json
@@ -56,7 +56,7 @@
     "fs-extra": "^7.0.0",
     "gauge": "^2.7.4",
     "get-it": "^5.2.1",
-    "get-latest-version": "^3.0.1",
+    "get-latest-version": "^4.0.0",
     "git-user-info": "^1.0.1",
     "gitconfiglocal": "^2.0.1",
     "inquirer": "^6.0.0",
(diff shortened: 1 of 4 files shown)


skogsmaskin (approved):
Great improvement! I have read through the code and also tested it with upgrading a studio to a tag, which worked perfectly.

PR #3178, opened 2022-03-04
fix(base): ignore presence sessions with no user profile

rexxars:
### Description

When a user profile does not exist or cannot be fetched due to insufficient permissions, we should handle the case gracefully by removing the presence session altogether. Currently, the studio crashes.

### Review notes

- I removed a few unused rxjs imports while I was in these files. Gardening and all that.
- Included a non-existent user ID in the mock debug presence data to make it easier to test

### Notes for release

- Fixes a bug where the studio could crash if a user had insufficient permissions to fetch user profiles while collaborating with others


bjoerge (approved):
Looks good!

PR #3181, opened 2022-03-07
fix(desk-tool): correct document list ordering on first load

rexxars:
### Description

TL;DR: This fixes an issue where the initial load of the document list does not respect the ordering if more than 100 items are present in the list, as well as improving the ordering of datetime fields.

Long version:

We previously had a bit of an odd ordering implementation because of some internal inconsistencies and the lacking ability to order items in a case-insensitive manner. Since then, we've resolved these issues, which now means we can utilize the correct ordering.

To do so, we have to know whether or not a field that is being sorted is a string, since applying `lower()` to other types will yield `null` and break ordering of those. Since I already put this logic in place, a similar approach was implemented for datetime fields - we apply the `dateTime()` method for those (this isn't _usually_ needed as ISO-dates in UTC sort just fine, but in cases where people might have imported data or similar, timezones might come into play and break the ordering). 

The implementation has some caveats:
- We need to be able to extract the document type from the GROQ filter, if any, and can only apply the logic if we do.
    - If you have a filter without a type constraint, we cannot apply these fixes
    - If you have a multi-type filter, we are not able to do so either. This case is theoretically solvable if we implement a better GROQ filter parsing and can extract multi-type filters. We then would need to assure that the field being sorted on is of the same type in all the matched types
    - If you have a type filter written in some exotic/non-obvious way, we're not able to extract the type and thus won't apply the fix

### Notes for release

- Fixes a bug where document lists would not always show items in the correct order until scrolling down and back up again

### DIFF
--- a/packages/@sanity/desk-tool/package.json
+++ b/packages/@sanity/desk-tool/package.json
@@ -58,6 +58,7 @@
   },
   "devDependencies": {
     "@sanity/base": "2.27.2",
+    "@sanity/schema": "2.27.0",
     "@sanity/ui-workshop": "^0.3.12",
     "@testing-library/jest-dom": "^5.11.10",
     "@testing-library/react": "^11.2.5",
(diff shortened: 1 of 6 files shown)


bjoerge (approved):
This is great! ✨

PR #3182, opened 2022-03-07
feat(export): allow `outputPath` to be a writable stream

tsolbjor:
### Description

Add the possibility to reference a writable stream as 'outputPath'


rexxars (approved):
Thanks! The PR had some code issues -`isWritableStream` did not return it's value, for instance. Have fixed these now.

PR #3183, opened 2022-03-07
feat(schema): warn when providing superfluous image metadata props

rexxars:
### Description

The properties `isOpaque`, `hasAlpha` and `dimensions` are automatically extracted and populated for any image asset. Specifying these will make the backend reject the request, as they are not user configurable.

Because they are listed as part of the image metadata, it is likely that users _will_ provide these, and have the upload requests fail.

This PR warns when specifying these properties, and removes them from the array of metadata before returning the compiled schema type.

### What to review

That the warning makes sense from a user perspective, and that the code looks correct.

### Notes for release

- Added a check for superfluous image metadata properties being specified


geoffreyjball (approved):
Looks great!

PR #3187, opened 2022-03-14
Fix dragging issue in PTE with array input

RitaDias:
### Description

- I've added a test studio which allows you to directly test the use case (`dev/sanity-drag-handler-bug-14502`) which will be removed before merging into `next`

The issue was when moving items around in a PTE (adding an item to an array input), it was triggering the hovering of the array items beneath and preventing users from making any additional changes afterwards. Check video:

https://user-images.githubusercontent.com/6951139/158134688-f874f712-7974-498e-a6d5-f600793f4d84.mov

### What to review
Follow the same path as shown in the video and make sure that the behavior is no longer happening. 
Also try to drag an image on the image input (inside of the PTE) and make sure that it still doesn't trigger the same bug mentioned before.

### Notes for release

Fix drag issue with the array input with PTE items (not super sure about this note 🥲 )

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/common/fileTarget/fileTarget.tsx
+++ b/packages/@sanity/form-builder/src/inputs/common/fileTarget/fileTarget.tsx
@@ -121,13 +121,14 @@ export function fileTarget<ComponentProps>(Component: React.ComponentType<Compon
     const handleDragEnter = useCallback(
       (event: React.DragEvent) => {
         event.stopPropagation()
-        /* this is a (hackish) work around to have the drag and drop work when the file is hovered back and forth over it
+
+        if (onFilesOver) {
+          /* this is a (hackish) work around to have the drag and drop work when the file is hovered back and forth over it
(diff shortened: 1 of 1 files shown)


skogsmaskin (changes requested):
Do we really need an own studio under dev for this?

Other than that, I think this would be great to have in there!

RitaDias:
> Do we really need an own studio for this?
> 
> Other than that, I think this would be great to have in there!

Nope we don't need it :) The studio was just added to make it easier for you (who is reviewing) to test the use case. 
The studio was added in a separate commit so it can be more easily dropped before pushing (and it should!). 

This is part of an improvement of our dev processes with support that I've had, I will bring it up to the team on tomorrow's SU :)

skogsmaskin:
> > Do we really need an own studio for this?
> > Other than that, I think this would be great to have in there!
> 
> Nope we don't need it :) The studio was just added to make it easier for you (who is reviewing) to test the use case. The studio was added in a separate commit so it can be more easily dropped before pushing (and it should!).

If I had just read your description a bit better I would not have made this comment. My apologies :)

PR #3191, opened 2022-03-15
Change plugin callstack

skogsmaskin:
### Description

I noticed that the call order of the PTE plugin stack is in the reverse order of what I intended. The idea is to have the most core ones being called first and the selection emitter plugin last (as plugins may manipulate the selection).

Also updated a test with the correct prop for the placeholder (was refactored a while ago).

### Notes for release

Doesn't need to be mentioned.


rexxars (approved):
👍

PR #3192, opened 2022-03-15
fix(portable-text-editor): fix wrong type option sent to custom paste…

skogsmaskin:
… handler

The wrong type was sent in here, it should be the PT array type and not the block type to keep backward compatibility with the API.


bjoerge (approved):
Needs a rebase, but otherwise LGTM

PR #3195, opened 2022-03-16
fix(code-input): preview loading non-existant mode

ISnackable:
### Description

Address the following issue of the changes from #3056, a fix for the same issue in #3077 but not fixed in the code-input preview component. 

Similarly, when the ace mode name is different from the language name, the following issue occurred. The mode is c_cpp while the language value is cpp.
issue

This can be tested under Plugin inputs > Code test > _Some Document_ > Code in arrays in the Test studio.

Before         |  After
:-------------------------:|:-------------------------:
image  |  image

### What to review

- Check that the code input **preview** work with an alternative language that has a different ace mode name to the language value name.
- Check that the code input **preview** work with other default supported languages

### Notes for release

- Fixed an issue with the code-input plugin's preview assigning wrong ace mode for additional languages


ISnackable:
Hmmm.. eslint is failing on CodeInput.tsx due to react/no-unused-prop-types: https://github.com/sanity-io/sanity/runs/5565087937?check_suite_focus=true#step:6:45.

Should I remove this line? @bjoerge

ISnackable:
Screwed something up... Just going to open a new PR

PR #3196, opened 2022-03-16
fix(form-builder) fix image resizing on initial load

RitaDias:
### Description

When loading a document that has an image input and that image is a smaller one, then the input should be of the height of the image. 
If the image is large, then it should be set to a max height that we have set.

After some discussion with Design, it was also decided to remove the resizing of the input for keeping consistency with how we do other inputs

### What to review

These following states are meant to be like that when you first open a document (especially for smaller images)

| default | small image | big image (max) |
|---------|-------------|-----------|
|   image      |       image  |   image |

It should also work when you crop the image (so if it's a large image, but you've cropped it, then when you open the document, the image input will be the size of the cropped version)

### Notes for release

Update image input height based on image size

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
@@ -5,7 +5,7 @@ import {Box, Button, Card, Dialog, Menu, MenuButton, MenuItem, Stack, ToastParam
 import {get, groupBy, uniqueId} from 'lodash'
 import {Observable, Subscription} from 'rxjs'
 import {ChangeIndicatorForFieldPath} from '@sanity/base/change-indicators'
-import {ImageIcon, SearchIcon} from '@sanity/icons'
+import {ChevronDownIcon, ImageIcon, SearchIcon} from '@sanity/icons'
 import {
   ImageAsset,
   AssetFromSource,
@@ -523,7 +523,7 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
       assetSources && assetSources?.length === 0 ? null : (
         <MenuItem
           icon={SearchIcon}
-          text="Browse"
+          text="Select"
           onClick={() => {
             this.setState({isMenuOpen: false})
             this.handleSelectImageFromAssetSource(assetSources[0])
@@ -593,9 +593,10 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
           button={
             <Button
               mode="ghost"
-              text="Browse"
+              text="Select"
               data-testid="file-input-multi-browse-button"
               icon={SearchIcon}
+              iconRight={ChevronDownIcon}
             />
           }
           menu={
@@ -624,7 +625,7 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
     return (
       <Button
         fontSize={2}
-        text="Browse"
+        text="Select"
         icon={SearchIcon}
         mode="ghost"
         onClick={() => {
--- a/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
@@ -50,7 +50,6 @@ export type Props = {
   type: ImageSchemaType
   level: number
   onChange: (event: PatchEvent) => void
-  resolveUploader: UploaderResolver
   observeAsset: (documentId: string) => Observable<ImageAsset>
   onBlur: () => void
   onFocus: (path: Path) => void
--- a/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
@@ -29,8 +29,7 @@ export function ImagePreview(props: ComponentProps<typeof Card> & Props) {
 
   const onLoadChange = useCallback(({target: img}) => {
     const imgHeight = img.offsetWidth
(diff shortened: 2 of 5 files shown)


packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -523,7 +522,7 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
       assetSources && assetSources?.length === 0 ? null : (
         <MenuItem
           icon={SearchIcon}
-          text="Browse"
```
bjoerge, line 526: This change doesn't look relevant to the PR description. Was including it intentional.

packages/@sanity/form-builder/src/inputs/files/common/ActionsMenu.tsx
```diff
@@ -45,7 +45,7 @@ export function ActionsMenu(props: Props) {
       {browse}
 
       <MenuDivider />
-      <MenuItem as="a" icon={DownloadIcon} text="Download original file" href={downloadUrl} />
+      <MenuItem as="a" icon={DownloadIcon} text="Download" href={downloadUrl} />
```
bjoerge, line 48: Same as the above comment - this doesn't seem like it belongs to the PR, just checking if it was intentionally included here.

packages/@sanity/form-builder/src/sanity/inputs/SanityImageInput.tsx
```diff
@@ -38,7 +37,6 @@ export default React.forwardRef(function SanityImageInput(props: Props, forwarde
   return (
     <ImageInputWithValuePath
       {...props}
-      resolveUploader={resolveUploader}
```
bjoerge, line 41: why was this removed?

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
```diff
@@ -12,13 +18,45 @@ interface Props {
...
   const {drag, readOnly, isRejected, src, ...rest} = props
   const [isLoaded, setLoaded] = useState(false)
-
+  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)
+  const rootRect = useElementRect(rootElement)
+  const rootWidth = rootRect?.width || 0
   const acceptTone = isRejected || readOnly ? 'critical' : 'primary'
   const tone = drag ? acceptTone : 'default'
 
+  const maxHeightToPx = (MAX_DEFAULT_HEIGHT * document.documentElement.clientHeight) / 100 // convert from vh to px, max height of the input
```
bjoerge, line 45: While reading this code without the context it's a bit unclear to me why we're doing this. Could you add a short comment explaining what we want to achieve and why we're doing it this way?

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
```diff
@@ -30,7 +68,7 @@ export function ImagePreview(props: ComponentProps<typeof Card> & Props) {
   }, [])
 
   return (
-    <RatioBox {...rest} style={{height: '30vh'}} tone="transparent">
+    <RatioBox {...rest} ref={setRootElement} style={{height: rootHeight}} tone="transparent">
```
bjoerge, line 71: would be good to have the style here wrapped in a useMemo so it only changes whenever the rootHeight changes, esp. if it is fairly stable across renders.

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
bjoerge, line 45: Could we wrap `maxHeightToPx` in a useMemo here? It never change after initial render if I'm reading this correctly.

bjoerge (approved):
Nice work! Added a few comments/questions, no blockers for merge.

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
RitaDias: Sorry, I should have added a comment on it. Yes, it was intentional.
There was some feedback that the previous text was causing some confusion :)

packages/@sanity/form-builder/src/inputs/files/common/ActionsMenu.tsx
RitaDias: Same as before, I think it was because the text was considered too long? @jorngeorg

jorngeorg: I edited it because it's factually wrong. As long as we're not using `dl=raw` you're not downloading the original but a an altered copy of the image.

packages/@sanity/form-builder/src/sanity/inputs/SanityImageInput.tsx
RitaDias: I removed it because, at least from the imageInput file, the linting was saying that the prop itself was never being used. I don't recall specifically why, but we started using the resolverUploader directly from `sanity/uploads/resolveUploader` 🤔 I felt it didn't make sense to keep it since that was the case

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImagePreview.tsx
RitaDias: there is an inline comment (right after the 100). I will move it on top of the declaration of the variable. I will also add the following:
This is a formula to transform vh values (the max value we have set for the image input is 30vh) into pixels (which are the values that we receive the image size in.

bjoerge (approved):

PR #3204, opened 2022-03-18
feat: Implemented fallback component for language-filter

LiamMartens:
### Description
This change was already made in the i18n plugin so it can fallback to other language select implementations if i18n is not enabled for the selected schema type. This PR introduces the same behavior for the field level localization. 

### What to review
When the plugin `@sanity/language-filter` is below `@sanity/document-internationalization` both implementations should still work. The language filter should show for object level translation schemas and the language switch dropdown should show for document level translations.

### Notes for release
This change introduces support to fallback to different language select implementations when the current is not relevant

### DIFF
--- a/packages/@sanity/language-filter/src/SelectLanguageProvider.tsx
+++ b/packages/@sanity/language-filter/src/SelectLanguageProvider.tsx
@@ -1,10 +1,11 @@
 // @todo: remove the following line when part imports has been removed from this file
 ///<reference types="@sanity/types/parts" />
 
-import React, {useRef, useEffect, useState} from 'react'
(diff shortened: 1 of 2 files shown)


sjelfull (approved):
Looks good to me!

The type addition also looks good to me, i'm just not 100% sure about adding types to parts.
@bjoerge Does it look good to you?

bjoerge (approved):
LGTM - thank you @LiamMartens!

fabien:
This breaks Studio setups without explicit `documentTypes` config: https://github.com/sanity-io/sanity/issues/3406

PR #3208, opened 2022-03-22
fix(types): use type-only imports, drop react dependency

### LINKED ISSUE #3194: @Sanity/types react version 17.0.1 while everything else is 17.0.2
Types package has a fixed react version, where all other packages has a free minor version bump ^ symbol, so after installing node_modules we are left with two react versions in the dependency tree: 17.0.2 for all the packages and 17.0.1 for @Sanity/types.

rexxars:
### Description

The types module currently has `react` as a dependency, but it shouldn't have to, as we are only using the type definitions for it. I believe it might have been added because `depcheck` was complaining about it, but now that it allows using the `@types` module for type-only imports, we can work around this.

This PR changes from type imports to `import type` imports and drops the `react` dependency. 

Fixes #3194

### What to review

That things still build.


rexxars:
> Love it! Should we consider adding an eslint rule enforcing it too?

Probably, but I unfortunately don't have time to look into it right now.

PR #3209, opened 2022-03-22
Get started template

skogsmaskin:


skogsmaskin:
@robinpyon - I added you as a reviewer here too, because we are merging in the Shopify template. It's pretty much as in you left it in your branch, but with some eslint fixes.

skogsmaskin:
I have tested the shopify template, and it's working as it should.

sjelfull (approved):
Looks  good to me, with the caveat that I didn't review the templates.

PR #3211, opened 2022-03-23
fix(default-layout): prevent changelog crash by filtering out entries without data to display

hermanwikner:
### Description

- The changelog crashes when a changelog entry is passed to the `ChangelogList` without any `changeItems` data. This PR fixes this issue by filtering out all the changelog entries that does not contain any `changeItems` before the changelog is passed to the `ChangelogList`.
- Add missing `href` to the "See full changelog on Github" link

### What to review

n/a

### Notes for release

Prevent changelog crash by filtering out entries without data to display

### DIFF
--- a/packages/@sanity/default-layout/src/update/changelog/ChangelogDialog.tsx
+++ b/packages/@sanity/default-layout/src/update/changelog/ChangelogDialog.tsx
@@ -20,7 +20,8 @@ const StickyBox = styled(Box)`
 export function ChangelogDialog(props: ChangelogDialogProps) {
   const {changelog, currentVersion, dialogProps, latestVersion} = props
 
-  const hasChangelog = changelog.filter((c) => c?.changeItems).length > 0
+  const changelogWithChangeItems = (changelog || []).filter((c) => c?.changeItems?.length > 0)
+  const hasChangelog = changelogWithChangeItems.length > 0
 
   return (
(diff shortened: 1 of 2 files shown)


packages/@sanity/default-layout/src/update/changelog/ChangelogDialog.tsx
```diff
@@ -20,7 +20,8 @@ const StickyBox = styled(Box)`
 export function ChangelogDialog(props: ChangelogDialogProps) {
   const {changelog, currentVersion, dialogProps, latestVersion} = props
 
-  const hasChangelog = changelog.filter((c) => c?.changeItems).length > 0
+  const changelogWithChangeItems = (changelog || []).filter((c) => c?.changeItems)
```
rexxars, line 23: In theory, the changelog can have a `changeItems` property that _is_ an array, but is empty. I think it would make sense to check that it both exists and is not empty. Agreed?

```suggestion
  const changelogWithChangeItems = (changelog || []).filter((c) => c?.changeItems?.length > 0)
```

rexxars (changes requested):

packages/@sanity/default-layout/src/update/changelog/ChangelogDialog.tsx
hermanwikner: That's a good idea! Agree 👍

PR #2271, opened 2021-01-30
[@sanity/cli] add --organization flag and organization selection for new projects on sanity init

runeb:
**Type of change (check at least one)**

- [ ]  Bug fix (non-breaking change which fixes an issue)
- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  Documentation (fix or update to documentation)
- [ ]  Maintenance
- [ ]  Other, please describe:

**Does this change require a documentation update? (Check one)**

- [x]  Yes
- [ ]  No

**Current behavior**

Creating new projects with `sanity init` does not let users select which organization to place them into.

**Description**
This PR enhances the `sanity init` command with a `--organization` cli flag to specify organization id, alternatively displays a visual selection list of available organizations for new projects, if no flag was provided. It will not let you assign an existing project to an organization, instead referring you to https://manage.sanity.io. If the user has no organizations the selection does not appear.

**Note for release**

**Checklist** 

- [x]  I have read the [Contributing Guidelines](https://github.com/sanity-io/sanity/blob/next/CONTRIBUTING.md)
- [ ]  The PR title includes a link to the relevant issue
- [x]  The PR title is appropriately formatted: `[some-package] PR title (#123)`
- [ ]  The code is linted
- [ ]  The test suite is passing
- [ ]  Corresponding changes to the documentation have been made

PR #3212, opened 2022-03-23
feat(cli): allow selecting organization during project creation

rexxars:
### Description

When creating new projects using the `sanity init` command, this PR will allow you to attach it directly to an organization - either using the `--organization` flag or by prompting the user. It does require an organization to already exist and that the user has "attach project" grants to it. If no organizations are found with these permissions, it skips the prompt.

Closes [sc-13493]
Supersedes #2271

### What to review

That the behavior, messaging and code looks right

### Notes for release

- Allow selecting an organization to attach new projects to during `sanity init`


RitaDias (approved):
Fantastic ✨

PR #3213, opened 2022-03-24
PTE: fix issue with annotations when merging two blocks

skogsmaskin:
There was a case here where merging two blocks (A,B) where B had annotations, their `markDefs` would not be copied over to A if you would backspace from the start of block B or press delete from the end of block A.

We already have normalization for this if spans are merged, but in this case when the merged node was a block, this code would miss.

Also added a test for this.

Also fixed an issue where orphaned `markDefs` would exist in block A (containing annotations) if splitting that block at the start position. Added test for this as well.

As part of this work I have also refactored the `.delete` method have a default mode where it will delete the selected content and merge nodes across the selection conserving unselected content instead of deleting the whole related blocks or children fully. I renamed mode `block` to `blocks` to better align with the actual effect of the operation and introduced a new  mode `selected` (which is the said default). As this API is not yet released to the public or documented, I think it can be tolerated to introduce this change without furter due at this point.

### Notes for release
* Fixes an issue where merging two blocks (A,B) where B had annotations, their `markDefs` would not be copied over to A if you backspaced from the start of block B or press delete from the end of block A.
* Fixes an issue where orphaned `markDefs` would exist in block A if splitting that block at the very start of it.


bjoerge (approved):
LGTM!

PR #3219, opened 2022-03-28
fix(form-builder): fix the number input to accept decimal values when default

RitaDias:
### Description

Fix issue where the number input wasn't allowing for decimal numbers (when the input is active)

### What to review

Check that you can add any decimal number on a basic number schema

### Notes for release

Fix issue where the number input wasn't allowing for decimal numbers (when the input is active)


bjoerge (approved):
Nice!

PR #3220, opened 2022-03-28
chore(perf-studio): remove perf-studio workflow

bjoerge:
### Description

Due to the unpredictable nature of the VMs we're assigned by Github it turned out to be hard to get comparable numbers. This PR removes the perf-test workflow until we figure out a more reliable way for gathering perf metrics.

### Notes for release

n/a (internal)


rexxars:
Forgot this PR existed and was unmerged :/
The same changes have been done both in next and v3, so closing this.

PR #3224, opened 2022-03-30
Chore: Refresh README

runeb:
* Fixes broken Slack badge
* Header refresh
* Updates Studio screenshot
* Removes old graphics

# Before
Screenshot 2022-03-30 at 15 43 30

# After
Screenshot 2022-03-30 at 16 06 01


kmelve (approved):
Great stuff! We'll do another pass on this in not too long.

PR #3226, opened 2022-03-31
feat(core): add `dry-run` flag

RitaDias:
### Description

Add `dry-run` flag as an option when running `sanity graphql deploy` 

### What to review

To test:
- Running `sanity graphql deploy --dry-run` will not deploy anything
  - It won't ask for the playground and, if there are conflicts, will not confirm that you want to deploy
- Running `sanity graphql deploy` will work exactly as before
- Do breaking changes to the schema. When running with the dry-run flag you should have a list of the changes and you won't be prompted about deploying with breaking changes

### Notes for release

Add `dry-run` flag as an option when running `sanity graphql deploy`

### DIFF
--- a/packages/@sanity/core/src/actions/graphql/deployApiAction.js
+++ b/packages/@sanity/core/src/actions/graphql/deployApiAction.js
@@ -26,6 +26,7 @@ const isInteractive = process.stdout.isTTY && process.env.TERM !== 'dumb' && !('
 const ignoredWarnings = ['OPTIONAL_INPUT_FIELD_ADDED']
 const ignoredBreaking = []
 
+// eslint-disable-next-line complexity
 module.exports = async function deployApiActions(args, context) {
   // Reparsing CLI flags for better control of binary flags
   const flags = parseCliFlags(args)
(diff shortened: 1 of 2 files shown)


packages/@sanity/core/src/actions/graphql/deployApiAction.js
```diff
@@ -115,8 +120,33 @@ module.exports = async function deployApiActions(args, context) {
...
+      const shouldDeploy = await prompt.single({
+        type: 'confirm',
+        message: 'Do you want to deploy a new API despite the dangerous changes?',
+        default: false,
+      })
+
+      if (!shouldDeploy) {
+        return
+      }
+    }
```
rexxars, line 146: The `else` branch is not necessary, since process.exit() will immediately kill the process - kind of like an early return. Less indentation/branching makes it harder to read the code, in my opinion :)

```suggestion
    if (dryRun) {
      process.exit(1)
    }

    if (!isInteractive) {
      throw new Error(
        'Dangerous changes found - falling back. Re-run the command with the `--force` flag to force deployment.'
      )
    }

    const shouldDeploy = await prompt.single({
      type: 'confirm',
      message: 'Do you want to deploy a new API despite the dangerous changes?',
      default: false,
    })

    if (!shouldDeploy) {
      return
    }
```

rexxars (approved):
Looks good!

packages/@sanity/core/src/actions/graphql/deployApiAction.js
RitaDias: Fully agreed, committed the change :)

PR #3228, opened 2022-04-01
fix(portable-text-editor): fix pasted plain text normalization issue

skogsmaskin:
### Description

There was a rare bug when pasting plain texti into the PTE where it sometimes would error with "Children for text block with _key '659301731a2e' is empty".

The reason is we are normalizing the wrong value. We must normalize the portableText value here, as this is the one being validated and cause the error further down in the code.

### What to review

That plain text pasting works as expected. We haven't found a way of reproducing this bug though (which content actually triggers it), only got reports. But it's very obvious what is causing it. 

### Notes for release

Fix issue with pasting plain text into the Portable Text editor where it in some rare circumstances would error with "Children for text block is empty"


bjoerge (approved):
LGTM!

PR #3229, opened 2022-04-01
fix(portable-text-editor): fix order of outer plugin stack

skogsmaskin:
`withReact` must be the inner plugin here, as insertData and insertFragmentData are defined here and should be overwritten by the insertData plugin.

### Notes for release
* Fix regression from ` v2.29.3` where formatting was lost when pasting HTML into the Portable Text editor.


bjoerge (approved):
LGTM!

PR #3231, opened 2022-04-04
refactor(default-layout): add data attributes to elements in `Navbar`

hermanwikner:
### Description

Add data attributes to elements in `Navbar` to enable being targeted with CSS. 

### What to review

Does the attribute names make sense?

### Notes for release

Add data attributes to elements in Navbar


mariuslundgard (approved):
LGTM!

PR #3237, opened 2022-04-06
fix(base): use same-origin referrer policy

rexxars:
### Description

This PR adds the `same-origin` referrer policy on the studio level, while opting for `strict-origin-when-cross-origin` on requests to the image CDN (`cdn.sanity.io`).

Rationale:

- When sending requests for user avatars to third-parties such as Google (`googleusercontent.com`) and GitHub (`avatars.githubusercontent.com`), we don't want to include a referrer, as this is leaking information. It can apparently also cause Google to return 403s if they get enough requests from the same origin.
- My initial thought was to add a referrer policy for these requests explicitly, but the avatars are rendered using the SVG `<image>` tag, which for some reason does not support setting a referrer policy.
- For requests to our image CDN, we want to include the origin (but not the full path) in order to answer questions about bandwidth usage grouped by origin.

### What to review

- That things still render as expected

### Notes for release

- Requests to third-party domains from the studio will no longer send a `referrer` header by default

### DIFF
--- a/packages/@sanity/base/src/components/Document.tsx
+++ b/packages/@sanity/base/src/components/Document.tsx
@@ -78,6 +78,7 @@ export default function Document(props: DocumentProps) {
         <title>{title}</title>
         <meta name="viewport" content={viewport} />
         <meta name="robots" content="noindex" />
+        <meta name="referrer" content="same-origin" />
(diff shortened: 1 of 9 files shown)


bjoerge (approved):
Great addition. LGTM!

PR #3250, opened 2022-04-21
fix: do not send in token to apiClient specifically during unauthenticated requests

svirs:
### Description

The client should not send up a token when the request is specifically required to be unauthenticated.


RitaDias (approved):
Looks good ✅

sjelfull (approved):

PR #3251, opened 2022-04-21
Show the date and time when change is hovered

selbekk:
### Description

This pull request adds a hover effect that shows the date and time (or ISO timestamp) for a given change. 

This helps out users that are looking at changes made weeks or months ago, that need to know the precise moment a change occurred (or at least the actual date), instead of "2w", "1y" etc.

### What to review

Make sure the code compiles, and that you want to use the ISO format. This was the quickest way to provide as much detail as possible without cluttering up the code with date formatting.

### Notes for release

Add support for showing the exact date and time a timeline item happened, by hovering a given timeline item.


sjelfull:
Thanks for the PR @selbekk! - we decided to make this a tooltip in #3628 to align it with how we show timestamps elsewhere.

PR #3253, opened 2022-04-22
feat(base): add flag to ignore schema types in search

runeb:
### Description

This PR adds a boolean property `__experimental_search_ignore` to the ObjectSchemaType for indicating that a type should be ignored in Studio search.

This is added to give more control over studio search results. A common scenario would be system documents, or otherwise non-editor documents showing up in search and confusing editors. A solution for the use case without this PR requires the removal of the schema from the schema definitions, but this also excludes the document type(s) from Desk editing.

The change is introduced along the existing `__experimental_search` property as a stop-gap feature until the Studio has stable APIs for customising the search experience. The reason for a new property is that the existing `__experimental_search` is an array of paths and weights, and there was no natural way to add this new functionality to that API.

### What to review

The added unit test should illustrate the change well. Should however be tested in a running Studio.

One possible blocker for this PR is that I believe this code path is also used for searching for documents from reference fields, so if you wish to filter these documents only from global search and not from reference fields, more changes and updates to the calling sites will have to be done.

This PR, if merged, need not appear in the release notes, as it is an experimental feature only to be used in certain blocking cases.

### DIFF
--- a/packages/@sanity/base/src/search/common/utils.ts
+++ b/packages/@sanity/base/src/search/common/utils.ts
@@ -5,6 +5,8 @@ const isDocumentType = (type: SchemaType): type is ObjectSchemaType =>
 
 const isSanityType = (type: SchemaType): boolean => type.name.startsWith('sanity.')
 
+const isSearchable = (type: SchemaType): boolean => type.__experimental_search_ignore !== true
+
 export const getSearchableTypes = (
   schema: Schema
   // eslint-disable-next-line camelcase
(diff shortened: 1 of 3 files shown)


KittyGiraudel:
I just stumbled upon this pull-request looking for this very feature. I was about to suggest implementing it by using `__experimental_search: []` but as you mentioned in the description, this might cause some unrelated issues.

I went over the code and it looks good (from what I can say as someone external). Looking forward to having it merged. 😊

KittyGiraudel (approved):

fabien:
What about a callback function to provide a GROQ filter to apply (based on the type that is passed to it)?

rexxars:
Closing this as the findability and improved search project has partially addressed this and will continue to improve on the functionality going forward.

PR #3257, opened 2022-04-25
fix(cli): update condition to render tutorial component

hermanwikner:
### Description

Only render the tutorial component if the user has permission to the project (i.e. has any role in the project)

### What to review

n/a 

### Notes for release

n/a


RitaDias (approved):
Let's go! ✨

PR #3259, opened 2022-04-25
feat(default-layout): add empty schema copy in the create document button tooltip

hermanwikner:
### Description

This PR adds copy in the 'create new document' button tooltip when there's no schema defined. Currently, when there's no schema defined, the permission message is displayed in the tooltip.

Screenshot 2022-04-26 at 11 05 51

### What to review

- Start a studio without a defined schema and check that the correct tooltip copy is displayed. A good way to test is to start the `clean-studio` with `npm run clean-studio`.
- Review the code – does the use of `getNewDocumentOptions` make sense in this situation, or are there other more optimals ways to check if the schema is empty?

### Notes for release

Add empty schema copy in the create document button tooltip

### DIFF
--- a/packages/@sanity/default-layout/src/navbar/Navbar.tsx
+++ b/packages/@sanity/default-layout/src/navbar/Navbar.tsx
@@ -11,7 +11,7 @@ import config from 'config:sanity'
 import * as sidecar from 'part:@sanity/default-layout/sidecar?'
 import ToolMenu from 'part:@sanity/default-layout/tool-switcher'
 import styled from 'styled-components'
-import {TemplatePermissionsResult} from '@sanity/base/_internal'
+import {getNewDocumentOptions, TemplatePermissionsResult} from '@sanity/base/_internal'
 import {HAS_SPACES} from '../util/spaces'
 import {DatasetSelect} from '../datasetSelect'
 import {useDefaultLayoutRouter} from '../useDefaultLayoutRouter'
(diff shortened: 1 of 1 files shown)


packages/@sanity/default-layout/src/navbar/Navbar.tsx
```diff
@@ -236,22 +257,7 @@ export const Navbar = memo(function Navbar(props: NavbarProps) {
...
-                      operationLabel="create any document"
-                    />
-                  )}
-                </Box>
-              }
-            >
+            <Tooltip portal scheme="light" content={<Box padding={2}>{createTooltipContent}</Box>}>
               <SpacingBox marginRight={shouldRender.brandingCenter ? undefined : 2}>
                 <Button
                   aria-label="Create new document"
```
ninaandal, line 263: Should it be added here that it is disabled if the schema is empty? `disabled={!canCreateSome || noDocumentOptions}`

hermanwikner: Good idea, I thought about that too! But it seems like the `canCreateSome` condition is taking care of disabling the button for both permission and empty schema situations. So I don't think it is necessary to add that! 🙂

RitaDias (approved):
Looks good to me ✅

skogsmaskin (approved):
LGTM!

PR #3261, opened 2022-04-26
refactor: alias __experimental_search to searchConfig for CDRs

ricokahler:
### Description

The following PR aliases the `__experimental_search` schema property to `searchConfig` for **`crossDatasetReference`s only**. Cross-dataset references are special here because this configuration is required even though it shares the same API as `__experimental_search`

Note: we went with the key `searchConfig` instead of the simpler `search` so we could potentially use that key in the future for another API.

### What to review

- Does the alias work as expected? Are the example cross-dataset refs in the studio working correctly?

### Notes for release

N/A


rexxars (approved):
👍

PR #3263, opened 2022-04-27
fix(cli): correct checks for cli auth token

rexxars:
### Description

PR #3250 introduced some unintended behavior changes, as parts of the CLI depended on
being able to say `requireUser: false` when using the `clientWrapper` utility, and then
checking for a `token` in its config to do conditional logic.

With the token not being part of the config after #3250, this broke. This commit changes
the checks to use a `getCliToken()` utility instead, which does not depend on the client
configuration per se, but rather the stored CLI config, as well as any environment
variable.

### What to review

That the CLI still works, and that the change appears to be correct. In particular, the `sanity login`, `sanity logout`, `sanity debug` and `sanity init` commands should be tested.

### Notes for release

- Fixed an issue where `sanity debug` would give "unauthorized" errors when used
- Fixed an issue where `sanity login` would not clear existing session on a new login
- Fixed an issue where `sanity init` might crash when attempting to fetch current user name

### DIFF
--- a/packages/@sanity/cli/src/commands/debug/printDebugInfo.js
+++ b/packages/@sanity/cli/src/commands/debug/printDebugInfo.js
@@ -96,7 +96,7 @@ function printKeyValue(obj, context) {
 
 async function gatherInfo(context) {
   const baseInfo = await promiseProps({
-    globalConfig: gatherGlobalConfigInfo(context),
+    globalConfig: gatherGlobalConfigInfo(),
     projectConfig: gatherProjectConfigInfo(context),
   })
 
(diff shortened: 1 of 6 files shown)


packages/@sanity/cli/src/commands/debug/printDebugInfo.js
```diff
@@ -120,7 +121,7 @@ function getGlobalConfigLocation() {
   return path.join(configDir, 'sanity', 'config.json')
 }
 
-function gatherGlobalConfigInfo(context) {
+function gatherGlobalConfigInfo() {
```
svirs, line 124: there's an invocation of `gatherGlobalConfigInfo` on ln 98 which doesn't need to get passed the context anymore

svirs (approved):
lgtm! but I don't know the api well enough to know when a session is really needed or not for all calls that were being sent since we've been allowing token through to the endpoints until now. I did a search across the code for when we set `requireUser` to false and this seems like it covers all cases

RitaDias (approved):
Looks good! ✅

PR #3265, opened 2022-04-28
fix(cli): don't remove dash from output path

ninaandal:
### Description
If the project name is for instance brown-bee, the output path would exclude the dash.

### What to review
Test that running init with a project that includes a dash in the name also includes the dash in the output name. 

### Notes for release
Fixed a bug that would not include dash in the output path when initializing a new project.


RitaDias (approved):
✨

Issue #2665, opened 2021-08-10
Wrong 'sanity deploy' error


brandonleboeuf:
When I tried to run 'sanity deploy' with the role of "Editor", I would get the error ">> Hostname already taken" for any studio hostname I tried.

When my role was switched to "Admin", the same name that was previously giving an error, would work correctly.

I believe the error should have been a permissions error rather than a naming conflict error.

judofyr:
Looking at the code it seems that we treat all errors from the server as "Hostname already taken". This is definitely a bug.

https://github.com/sanity-io/sanity/blob/98ffed5545ad9c8ba646b38610b86a6567928236/packages/%40sanity/core/src/actions/deploy/deployAction.js#L165-L169.

rexxars:
Thanks for reporting! This was resolved in [v2.30.0](https://github.com/sanity-io/sanity/releases/v2.30.0)

PR #3269, opened 2022-05-02
fix(cli): show correct error msg for sanity deploy errors

ninaandal:
### Description

When a user with the role as "Editor" tried to run `sanity deploy`, an error message of "Hostname already taken" would be provided, even if the hostname was available. Link to issue: https://github.com/sanity-io/sanity/issues/2665. 

This commit improves the error messages provided to the user.
Instead of a hardcoded "Hostname already taken" for any `sanity deploy` error, a correct error message belonging to the error will be shown.

### What to review

Run `sanity deploy` with the role as "Editor, or when providing a Studio hostname that is already taken. Make sure that correct and appropriate error message is given. 

### Notes for release

Fixed an issue where `sanity deploy` would give "Hostname already taken" for all errors


rexxars (approved):
👍

PR #3271, opened 2022-05-03
fix: field presence tooltip issues

hermanwikner:
### Description

This PR fixes z-index issues in `PresenceTooltip`. In order to solve these issues without adding specific z-index values – the `PresenceTooltip` is now using the `portal` attribute. However,  the positioning of the `Tooltip` was not behaving correctly when using the default portal element provided by the `PortalProvider` in `DocumentPanel`. To solve this, I added another portal element with the `__unstable_elements` to be used by the `PresenceTooltip`.

Screenshot 2022-05-03 at 15 06 39

### What to review
- Does the use of `__unstable_elements` make sense, or could this be solved in another way?

### Notes for release
Fix field presence tooltip z-index issues


ricokahler (approved):
nice!

PR #3278, opened 2022-05-09
fix(validation): fix broken date validation message

hermanwikner:
### Description

This PR fixes an issue with the date formatting in `dateValidator`, which caused the validation message shown in the tooltip to be incorrect. The format function from `date-fns` got an invalid time value. 

**Before and after**

Screenshot 2022-05-09 at 14 02 59

**Schema**
```js
    {
      name: 'startDate',
      type: 'datetime',
      title: 'Start date',
      validation: (Rule) => Rule.required(),
    },
    {
      name: 'endTime',
      type: 'datetime',
      title: 'End date',
      validation: (Rule) => Rule.min(Rule.valueOfField('startDate')).error('My error msg'),
    }
```

### What to review

- Review the code
- Try out the schema above 

### Notes for release

Fix broken date validation message

### DIFF
--- a/dev/test-studio/schema/standard/datetime.js
+++ b/dev/test-studio/schema/standard/datetime.js
@@ -68,6 +68,17 @@ export default {
       placeholder: 'Enter a date here',
     },
     {
+      name: 'startDate',
+      type: 'datetime',
+      title: 'Start date',
+    },
+    {
+      name: 'endTime',
+      type: 'datetime',
+      title: 'End date',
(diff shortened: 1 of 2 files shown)


rexxars (approved):
I added a schema example to more easily test this in the future, and fixed the typings on `getFormattedDate` - I see that the validation package actually has an error where the date validators assumes the value passed to `min()` and `max()` is a number, which they're not in the case of dates (they are strings), but think that is beyond the scope of this PR to fix.

Anyway, LGTM!

Issue #2997, opened 2021-12-09
Improve "Create New" type selection


rshackleton:
**Is your feature request related to a problem? Please describe.**
The new "reference in place" feature is brilliant and streamlines the content editing process for related content. However, we're using Sanity to handle some marketing page content where the page is split into various sections which can be added to a list. Essentially a page builder. As the list of possible types is quite long there are some UI changes which would be beneficial.

**Describe the solution you'd like**
1. The new type selector becomes very tall with a lot of possible types, perhaps an option for this to be a dialog or something would be better.

![image](https://user-images.githubusercontent.com/13448395/145414290-0ec7fdf0-c27b-4a67-aac6-e9b73b307aad.png)

2. The list of possible types takes up a lot of space and isn't useful for the content editor (in this case). The option to remove/override this would be great too.

![image](https://user-images.githubusercontent.com/13448395/145414413-4832fb91-e107-47e3-8d15-6ea462cf3f90.png)

**Describe alternatives you've considered**

donnachadelong:
As an editor working in Sanity, this is a big issue. The list of widgets is too long to be viewed on 100% or even 90% browser window magnification. I need to zoom out to see the widgets at the bottom of the list. This makes editing very cumbersome and it looks ridiculous when doing training. At the very least, please add a scroll-bar to the list.

ninaandal:
Thank you for reporting! A fix has been released as `v2.30.0`.

rshackleton:
The new popout update is great, thanks for that change.

There was the second suggestion in my original issue regarding the list of types shown before - is it possible also add the option to hide that?

rshackleton:
Yes, this list of types. For a content editor it isn't a particularly useful message in all cases.

![image](https://user-images.githubusercontent.com/13448395/168394138-3fe4dbe3-f192-4b99-abe7-93a93460c6c2.png)

ninaandal:
This is default when you don't have a `title`. If you add a `title` to the reference type, then this list will not be shown. I hope this will fix your issue 😊

rshackleton:
![image](https://user-images.githubusercontent.com/13448395/168594014-e3f49a39-3c21-4afb-bb71-abcf4d37e606.png)

That is much better, thanks for letting me know about that. I'd clearly overlooked that in the docs.

PR #3281, opened 2022-05-10
fix(base): add scrolling to dropdown when adding new reference

ninaandal:
### Description
The type selector list when adding a new reference did not have a size constraint. A size constraint was therefore added to aid for scrolling.

Link to issue: https://github.com/sanity-io/sanity/issues/2997 

### What to review
Add a new reference and check that the list is scrollable when screensize is smaller than the list. 

### Notes for release
Add scrolling to the list of types when adding a new reference.


RitaDias (approved):
🚢🎉

PR #3287, opened 2022-05-13
feat(cli): allow specifying core path for development

dcilke:
### Description

Add a new environment variable for cli development `SANITY_CLI_CORE_PATH` to allow using the local core library for the CLI.

### What to review

`SANITY_CLI_CORE_PATH` is honored and log messages make sense

### Notes for release

None

### DIFF
--- a/packages/@sanity/cli/src/cli.js
+++ b/packages/@sanity/cli/src/cli.js
@@ -69,6 +69,22 @@ module.exports = async function runCli(cliRoot) {
 }
 
 function getCoreModulePath(workDir) {
+  /* eslint-disable no-process-env */
+  if (process.env.SANITY_CLI_CORE_PATH) {
+    const local = fse.existsSync(process.env.SANITY_CLI_CORE_PATH)
+    if (local) {
+      console.log(
+        chalk.yellow(`[WARN] Loading @sanity/core from: ${process.env.SANITY_CLI_CORE_PATH}`)
+      )
+      return process.env.SANITY_CLI_CORE_PATH
+    }
+    console.error(
+      chalk.red(`Could not resolve SANITY_CLI_CORE_PATH: ${process.env.SANITY_CLI_CORE_PATH}`)
+    )
+    process.exit(1)
+  }
+  /* eslint-enable no-process-env */
+
   const pkgPath = resolveFrom.silent(workDir, '@sanity/core')
   if (pkgPath) {
     return pkgPath


vquigley (approved):

packages/@sanity/cli/src/cli.js
```diff
@@ -69,6 +69,22 @@ module.exports = async function runCli(cliRoot) {
...
+  if (process.env.SANITY_CLI_CORE_PATH) {
+    const local = fse.existsSync(process.env.SANITY_CLI_CORE_PATH)
+    if (local) {
+      console.log(
+        chalk.yellow(`[WARN] Loading @sanity/core from: ${process.env.SANITY_CLI_CORE_PATH}`)
+      )
+      return process.env.SANITY_CLI_CORE_PATH
+    }
+    console.warn(
+      chalk.red(`[ERR] Could not resolve SANITY_CLI_CORE_PATH: ${process.env.SANITY_CLI_CORE_PATH}`)
```
RitaDias, line 82: Could we make the error message clearer / make it more "human" friendly and aligned with the rest of the messages in the cli (for example, removing the `[ERR]`)? 🙏

packages/@sanity/cli/src/cli.js
```diff
@@ -69,6 +69,22 @@ module.exports = async function runCli(cliRoot) {
...
+  /* eslint-disable no-process-env */
+  if (process.env.SANITY_CLI_CORE_PATH) {
+    const local = fse.existsSync(process.env.SANITY_CLI_CORE_PATH)
+    if (local) {
+      console.log(
+        chalk.yellow(`[WARN] Loading @sanity/core from: ${process.env.SANITY_CLI_CORE_PATH}`)
+      )
+      return process.env.SANITY_CLI_CORE_PATH
+    }
+    console.warn(
```
RitaDias, line 81: Also, if this is an error then perhaps it is best to use `output.error()` instead?

```suggestion
    output.error(```

RitaDias (changes requested):
I'm still going to go through and test it out but just thinking about messaging :)

packages/@sanity/cli/src/cli.js
dcilke: 👍  We have mixed uses of `[ERR]` in this file. Happy to go with whichever aligns with the rest of the CLI.

rexxars:
Would using the [symlinkDependencies script](https://github.com/sanity-io/sanity/blob/next/scripts/symlinkDependencies.js) serve the intended purpose? Just trying to cut down on the number of code paths and features to maintain

dcilke:
I'm not familiar with that script but it looks like it would solve the problem I have. I'm thinking of closing this in favor of documentation. Would `CONTRIBUTING.md` be the right place?

RitaDias:
> I'm not familiar with that script but it looks like it would solve the problem I have. I'm thinking of closing this in favor of documentation. Would `CONTRIBUTING.md` be the right place?

I think this could be the best place yes!
The way I've shared "how to use it" with people as been:

On your project:
`npm install

In the mono repo root
`./scripts/symlinkDependencies.js [path/to/project/studio] --all`
`npm run watch`

On your project:
`npm start / sanity start`

atombender:
Isn't this was `npm link` is supposed to solve?

RitaDias (approved):
This looks good! I did some tests and all seems to work as expected though I feel I could have done some more robust testing. Let me know if there is any path that I should have been more attentive to ✨

rexxars:
This is going to change with v3 (`@sanity/core` is no longer used), and the `symlinkDependencies` script should be enough to cover these cases, so I am going to close this one for now. If there is a case that is not covered by it, feel free to reopen.

PR #3289, opened 2022-05-13
feat(copy dataset): Allow API to determine if target dataset exists

vquigley:
### Description
Allow API to determine if the target dataset exists or if a copy job is currently in progress.
New functionality in API to determine the difference between existing dataset and running job.
Stops confusion when a copy job is in progress but the UI says "dataset name already exists".

### What to review

Simply removing a check against local storage.

### Notes for release
New functionality in API to determine the difference between existing dataset and running job.
CLI will now let the API determine the state of the target dataset.

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -172,7 +172,7 @@ export default {
       if (error.statusCode) {
         output.print(`${chalk.red(`Dataset copying failed: \n${error.response.body.message}`)}\n`)
       } else {
-        output.print(`${chalk.red(`Dataset copying failed: SECIBD >>>\n${error.message}`)}\n`)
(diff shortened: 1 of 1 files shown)


packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -142,9 +142,6 @@ export default {
 
     const targetDatasetName = await (targetDataset ||
       promptForDatasetName(prompt, {message: 'Target dataset name:'}))
-    if (existingDatasets.includes(targetDatasetName)) {
-      throw new Error(`Target dataset "${targetDatasetName}" already exists`)
-    }
```
dcilke, line 147: Do we need to make any changes to the error logging?

vquigley: No, it gets caught in the same set of statements we already use.

RitaDias (approved):
✅ This looks good!

PR #3292, opened 2022-05-16
feat(cli): list dataset copy jobs

tzhelyazkova:
### Description

Introduce a new cli command `sanity dataset jobs list` which prints out information about the dataset copy jobs related to the project.
This new command will help users find out the status of copy jobs without needing the job id.
[Story 18247](https://app.shortcut.com/sanity-io/story/18247/list-job-history-sanity-cli) 

### What to review

This PR introduces a new outputter function `printTable` and a new dataset action `listDatasetCopyJob`.
You can test it locally by running `sanity dataset jobs list` in `/dev/test-studio` for example. 

Screenshot 2022-05-18 at 16 50 17

### Notes for release

New CLI command `sanity dataset jobs list`.

### DIFF
--- a/packages/@sanity/cli/src/outputters/cliOutputter.js
+++ b/packages/@sanity/cli/src/outputters/cliOutputter.js
@@ -9,6 +9,10 @@ export default {
     console.log(...args)
   },
 
+  table(...args) {
+    console.table(...args)
+  },
+
   warn(...args) {
     console.warn(...args)
   },
--- a/packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
+++ b/packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
@@ -25,7 +25,7 @@ module.exports = async function listDatasetCopyJobs(args, context) {
     if (error.statusCode) {
       output.error(`${chalk.red(`Dataset copy list failed:\n${error.response.body.message}`)}\n`)
     } else {
-      output.print(`${chalk.red(`Dataset copy list failed:\n${error.message}`)}\n`)
+      output.error(`${chalk.red(`Dataset copy list failed:\n${error.message}`)}\n`)
     }
   }
 
--- a/packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
+++ b/packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
@@ -1,0 +1,72 @@
+import {parseISO, formatDistanceToNow, formatDistance} from 'date-fns'
+
+module.exports = async function listDatasetCopyJobs(args, context) {
+  const {apiClient, output, chalk} = context
+  const flags = args.extOptions
+  const client = apiClient()
+  const projectId = client.config().projectId
+  const query = {}
+  let response
+
+  if (flags.offset && flags.offset >= 0) {
+    query.offset = flags.offset
+  }
+  if (flags.limit && flags.limit > 0) {
+    query.limit = flags.limit
+  }
+
+  try {
+    response = await client.request({
+      method: 'GET',
+      uri: `/projects/${projectId}/datasets/copy`,
+      query,
+    })
+  } catch (error) {
+    if (error.statusCode) {
+      output.error(`${chalk.red(`Dataset copy list failed:\n${error.response.body.message}`)}\n`)
+    } else {
+      output.print(`${chalk.red(`Dataset copy list failed:\n${error.message}`)}\n`)
+    }
+  }
+
+  if (response && response.length > 0) {
+    output.print('Dataset copy jobs for this project:')
+    const print = []
+
+    response.forEach((job) => {
+      const {id, state, createdAt, updatedAt, sourceDataset, targetDataset, withHistory} = job
+
+      let timeStarted = ''
+      if (createdAt !== '') {
+        timeStarted = formatDistanceToNow(parseISO(createdAt))
+      }
+
+      let timeTaken = ''
+      if (updatedAt !== '') {
+        timeTaken = formatDistance(parseISO(updatedAt), parseISO(createdAt))
+      }
+
+      print.push({
+        'Job ID': id,
+        State: state,
+        History: withHistory,
+        'Time started': timeStarted === '' ? '' : `${timeStarted} ago`,
+        'Time taken': timeTaken,
+        'Source dataset': sourceDataset,
+        'Target dataset': targetDataset,
+      })
+    })
+
+    output.table(print, [
+      'Job ID',
+      'Source dataset',
+      'Target dataset',
+      'State',
+      'History',
+      'Time started',
+      'Time taken',
+    ])
+  } else {
+    output.print("This project doesn't have any dataset copy jobs")
+  }
+}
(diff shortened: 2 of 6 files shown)


packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
```diff
@@ -0,0 +1,82 @@
...
+        state: state,
+        history: withHistory,
+        'time started': `${timeStarted} ago`,
+        'time taken': timeTaken,
+        'source dataset': sourceDataset,
+        'target dataset': targetDataset,
+      })
+    })
+
+    output.printTable(print)
```
dcilke, line 57: We should provide an order to the rows in the table.
```suggestion
    output.printTable(print, ['jobId', 'source dataset', 'target dataset', 'state', 'history', 'time started', 'time taken'])
```

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
```diff
@@ -0,0 +1,82 @@
...
+  } else {
+    output.print("This project doesn't have any dataset copy jobs")
+  }
+}
+
+function convertTimestampToDaysHoursMinsSecs(timePassedInSeconds) {
```
dcilke, line 63: `sanity` already imports `Moment.js`, should we use that?

packages/@sanity/core/src/commands/dataset/listCopyDatasetJobs.js
```diff
@@ -0,0 +1,30 @@
+import lazyRequire from '@sanity/util/lib/lazyRequire'
+
+const helpText = `
+    Options
+        --offset Start position in the list of jobs
+        --limit Maximum number of jobs returned
```
dcilke, line 6: Do we have a default or maximum on limit?

packages/@sanity/cli/src/outputters/cliOutputter.js
```diff
@@ -9,6 +9,10 @@ export default {
     console.log(...args)
   },
 
+  printTable(...args) {
```
dcilke, line 12: nit: Just `table`

dcilke (changes requested):

packages/@sanity/core/src/commands/dataset/listCopyDatasetJobs.js
tzhelyazkova: added.

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
tzhelyazkova: It seems that Moment.js is considered a legacy project - https://www.npmjs.com/package/moment

I tried it out and it rounds up the time according to some internal rules. See https://momentjs.com/docs/#/displaying/fromnow/
Do we care if it says "10 hours ago" instead of "9 hours, 20 minutes, 44 seconds ago"? 🤔 

With `Moment.js`:
Screenshot 2022-05-23 at 17 40 34


With custom code:
Screenshot 2022-05-23 at 17 41 03

RitaDias: We already have some code available when it comes to reading time in `base` -> `useTimeAgo`. Though since it's a hook I don't think we could use it here anyway 🤔 

I don't know if hours are valuable here but my gut feeling tells me that we should use moment instead of adding more logic that I don't feel would add a lot? But I might be missing context and the hours make sense. Thoughts? @rexxars @skogsmaskin

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
```diff
@@ -0,0 +1,90 @@
...
+
+  try {
+    response = await client.request({
+      method: 'GET',
+      uri: `/projects/${projectId}/datasets/copy`,
+      query,
+    })
+  } catch (error) {
+    if (error.statusCode) {
+      output.print(`${chalk.red(`Dataset copy list failed:\n${error.response.body.message}`)}\n`)
```
RitaDias, line 24: I also just double checked with the team and it's preferable to use `output.error()` over the `output.print` :)

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
vquigley: I agree with Rita here. If we are already using Moment then it being legacy isn't as big an issue as if we were adding a new dependency. 

I would rather we display "10 hours ago" instead of "9 hours, 20 minutes, 44 seconds ago". In this context, the exact time is not relevant, i.e. I don't care about the seconds or minutes when the time was weeks ago. If they want exact timing then use the HTTP API.

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
```diff
@@ -0,0 +1,90 @@
+module.exports = async function listDatasetCopyJobs(args, context) {
```
rubioz, line 1: 🏳️ (don't read too much into this): 

The only comment I could leave here is that, perhaps we should test this a bit, as there are three different responsibilities:

1. Data fetching: are the parameters right? do communications succeed?
2. Data mapping: is the returned payload presentable? does it have the right information?
3. Data presenting: can we present the information?

Refactoring this into three different units will be overkill, but perhaps we can build some test around how it would behave when requests fail.

rubioz (approved):

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
tzhelyazkova: Thanks all! I removed the custom code and used `date-fns` as suggested. The result:
Screenshot 2022-06-02 at 12 02 19

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
tzhelyazkova: The only thing we can verify here with a unit test is presenting the information. I did try to write a test for that a while ago but I couldn't find a way to verify what is being printed to the console. 
The rest of the things you mentioned would be in an integration test I guess but that's out of scope for this PR. 🤔

vquigley (approved):

packages/@sanity/core/src/actions/dataset/listDatasetCopyJobs.js
```diff
@@ -0,0 +1,72 @@
...
+    response = await client.request({
+      method: 'GET',
+      uri: `/projects/${projectId}/datasets/copy`,
+      query,
+    })
+  } catch (error) {
+    if (error.statusCode) {
+      output.error(`${chalk.red(`Dataset copy list failed:\n${error.response.body.message}`)}\n`)
+    } else {
+      output.print(`${chalk.red(`Dataset copy list failed:\n${error.message}`)}\n`)
```
RitaDias, line 28: smallest nitpick, feels like this should also be output.error?

```suggestion
      output.error(`${chalk.red(`Dataset copy list failed:\n${error.message}`)}\n`)
```

RitaDias (approved):
Just a small nitpit but otherwise it's looking good :) 
edit: I've also rebased the branch, let's see if the errors go away

RitaDias:
@tzhelyazkova 👋 I was hopeful that rebasing it would make the tests pass but it doesn't seem the case. The tests would need to pass / stop screaming before merging :)

PR #3296, opened 2022-05-19
fix(desk-tool): prevent sortBy from overtaking the defaultOrdering

RitaDias:
### Description

Fix two bugs:
- Using `defaultOrdering` in the desk structure will now work.
- With this fix we revert a bug that was introduced in a past PR where the document list appeared to be loading forever when changing the sorting.

I will drop the last commit shown since it's only for pr testing purposes.

### What to review

1. Go to the test studio
2. At the bottom of the desk structure, you will find two authors: The first one has the `defaultOrdering` based on `role` while the second is a normal one based on name. *They should have the documents in different order*
3. Double check that the order of the documents is different
4. Double check that when changing sorting options that it will keep trying to load the documents forever

### Notes for release

- Using `defaultOrdering` in the desk structure will now work.
- Revert a bug that was introduced in a past PR where the document list appeared to be loading forever when changing the sorting.

### DIFF
--- a/dev/test-studio/parts/deskStructure.js
+++ b/dev/test-studio/parts/deskStructure.js
@@ -527,6 +527,10 @@ export default () =>
 
       S.divider(),
 
+      S.documentTypeListItem('author')
+        .id('test-sorting')
+        .child(S.documentTypeList('author').defaultOrdering([{field: 'role', direction: 'desc'}])),
+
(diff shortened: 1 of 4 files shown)


rexxars (approved):
👍

geball (approved):
Thanks Rita!

PR #3300, opened 2022-05-22
Ensure `margins` doesn't make its way to the DOM

KittyGiraudel:
### Description

The `WithIntersection` component spreads its props on the rendered `<div>` element which results in the `margins` prop making its way onto the DOM. This is not a valid prop and shouldn’t be rendered in the HTML.

### DIFF
--- a/packages/@sanity/base/src/presence/overlay/WithIntersection.tsx
+++ b/packages/@sanity/base/src/presence/overlay/WithIntersection.tsx
@@ -11,7 +11,7 @@ export interface WithIntersectionProps {
 export const WithIntersection = (
   props: WithIntersectionProps & React.HTMLProps<HTMLDivElement>
 ) => {
-  const {onIntersection, io, id, ...rest} = props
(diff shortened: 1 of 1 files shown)


KittyGiraudel:
Why was this closed?

rexxars:
Not sure why this got closed - might have been a result of deleting the fork or similar. Have re-opened and will review shortly.

KittyGiraudel:
GitHub used to maintain ghost forks to preserve open pull-requests — apparently this behavior has changed. I am sorry about this. Thank you for reopening though! ⭐

rexxars:
Thanks for raising this! After some investigation, it looks like `margins` were never declared as a prop for this component, which caused typescript to choke on this change. Since the prop was never meant to hit the DOM but was only used for styling purposes, I am thinking it is probably better to use the `$` prefix to indicate to `styled-components` that it should not reach the DOM.

3ff4182166be7352a5e70433dd39d05a64d32602 addresses this and will be included in the next release. Thanks again for raising!

PR #3314, opened 2022-06-02
fix(cli): throw a better error message in case of a failed dataset copy job

tzhelyazkova:
### Description

The error method receives an event object, which doesn't have an error property nor a message property.
The message is changed to print the event data in case of a failed job.

### DIFF
--- a/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
+++ b/packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
@@ -90,8 +90,9 @@ const followProgress = (jobId, client, output) => {
 
       spinner.text = `Copy in progress: ${currentProgress}%`
     },
-    error: (err) => {
-      spinner.fail(`There was an error copying the dataset: ${err.message}`)
+    error: (event) => {
+      spinner.fail()
+      throw new Error(`There was an error copying dataset: ${event.data}`)
     },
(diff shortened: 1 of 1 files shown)


j33ty (approved):

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -90,8 +90,9 @@ const followProgress = (jobId, client, output) => {
 
       spinner.text = `Copy in progress: ${currentProgress}%`
     },
-    error: (err) => {
-      spinner.fail(`There was an error copying the dataset: ${err.message}`)
+    error: (event) => {
+      spinner.fail()
+      throw new Error(`There was an error copying dataset with job id: ${event.data.job_id}`)
```
vquigley, line 95: This will return undefined if the Job Id does not exist. Can we just output event.data? That will output all the info.

image

image

vquigley (changes requested):

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
tzhelyazkova: huh I didn't know it's possible to not have a job id

vquigley (approved):

PR #3315, opened 2022-06-07
chore(cli): add output to inform that `--skip-history` flag exists

vquigley:
### Description

This has been added after investigations into long processing times. Copying with history is slow.

This information is an alternative to changing the default settings and breaking backwards compatibility.

### What to review

New output. Run copy process with and without flag to see output.

```sanity dataset copy production backup-1```
image

```sanity dataset copy production backup-2 --skip-history```
image

### Notes for release

N/A


RitaDias (approved):
✅ let's go

dcilke (approved):

PR #3318, opened 2022-06-09
fix(cli): use dataset copy command to list copy dataset jobs

tzhelyazkova:
Instead of having a separate command for listing dataset copy jobs, use the already existing `dataset copy` command by adding a new `--list` flag.


packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -3,13 +3,17 @@ import {Observable} from 'rxjs'
...
 import debug from '../../debug'
 
 const helpText = `
 Options
   --detach Start the copy without waiting for it to finish
   --attach <job-id> Attach to the running copy process to show progress
   --skip-history Don't preserve document history on copy
+  --list Lists all dataset copy jobs corresponding to a certain criteria.
+  --offset Start position in the list of jobs. Default 0.
+  --limit Maximum number of jobs returned. Default 10. Maximum 1000.
```
dcilke, line 16: 🤔 These flags are only relevant when `--list` is specified. Maybe something like:
```suggestion
  --offset Start position in the list of jobs. Default 0. With --list.
  --limit Maximum number of jobs returned. Default 10. Maximum 1000. With --list.
```
For clarity?

packages/@sanity/core/src/commands/dataset/copyDatasetCommand.js
```diff
@@ -105,12 +112,18 @@ export default {
   group: 'dataset',
   signature: '[SOURCE_DATASET] [TARGET_DATASET]',
   helpText,
-  description: 'Copies a dataset including its assets to a new dataset',
+  description:
+    'Manages dataset copy jobs, including starting a new copy job, listing copy jobs and following the progress of a running copy job',
```
dcilke, line 116:
```suggestion
    'Manages dataset copying, including starting a new copy job, listing copy jobs and following the progress of a running copy job',
```

dcilke (commented):
My two cents

PR #3347, opened 2022-06-16
refactor(sanity): make inline objects and annotations open in popover by default

mariuslundgard:
### Description

This PR makes it so the Portable Text input treats inline objects, annotations, and custom block objects differently:
- Inline objects open in a popover by default.
- Annotations open in a popover by default.
- Custom block objects open in a dialog by default,

### What to review

Make sure opening and closing of inline objects, annotations and custom block objects works as expected.

### DIFF
--- a/packages/sanity/src/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/form/inputs/PortableText/PortableTextInput.tsx
@@ -51,9 +51,8 @@ export type PTObjectMember = ArrayOfObjectsItemMember<
     ObjectSchemaType
   >
 >
-
 export interface PortableTextMemberItem {
-  kind: 'annotation' | 'object' | 'inlineObject'
+  kind: 'annotation' | 'textBlock' | 'objectBlock' | 'inlineObject'
   key: string
   member: PTObjectMember
   node: ObjectFormNode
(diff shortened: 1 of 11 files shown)


packages/sanity/src/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -158,56 +149,89 @@ export function PortableTextInput(props: PortableTextInputProps) {
...
+      // if (member.kind === 'error') {
+      // }
+
+      if (member.kind === 'item') {
+        if (!_isBlockType(member.item.schemaType)) {
+          result.push({kind: 'object', member, node: member.item})
+        }
+
+        if (member.item.validation.length > 0) {
+          result.push({kind: 'object', member, node: member.item})
```
skogsmaskin, line 169: This will potentially add members twice.

skogsmaskin (changes requested):
Had a comment about the portableTextMembers compilation.

Also, I'm not sure why we need the `.node` in those objects? For me this just complicates things.

Other than that, this looks good to me.

skogsmaskin:
Something also happened with the focus first field function in `ObjectEditModal`  - it no longer focuses first field on opening.

PR #3355, opened 2022-06-16
adding calendarTodayLabel

saiichihashimoto:
`calendarTodayLabel` appears inconsistently in the date and datetime documentation. Not sure it truly does anything #3354 but adding it to types regardless.

https://www.sanity.io/docs/date-type
https://www.sanity.io/docs/datetime-type


rexxars (approved):
Thanks, Saiichi!

PR #3362, opened 2022-06-19
feat(block-tools): use strict typings

rexxars:
### Description

The `@sanity/block-tools` package had a lot of implicit any types, and did not use strict null checks.
In the process of getting the full autogenerated documentation one step further, this needed to be addressed.

This PR attempts to provide explicit types wherever possible. As you can see, the test _code_ is basically left untouched, so if the test coverage was sufficient previously, this should be fairly safe.

### What to review

- That the changes generally makes sense
- Accepting the breaking changes this introduces:
  - The default export is removed. We now only use named exports. It seemed like we had a bit of a dual setup where we _both_ exported a default _and_ used named exports, which ESBuild didn't like.
  - The block normalization currently has what I feel [must be a bug](https://github.com/sanity-io/sanity/blob/v3/packages/%40sanity/block-tools/src/util/normalizeBlock.ts#L66-L70)? It looks like unused annotations are meant to be removed, but it only looks for _any_ mark definition and keeps it around if one exists.

### Notes for release

- Improved typings for the `@sanity/block-tools` module
- BREAKING: `@sanity/block-tools` now uses only named exports


skogsmaskin:
> * The block normalization currently has what I feel [must be a bug](https://github.com/sanity-io/sanity/blob/v3/packages/%40sanity/block-tools/src/util/normalizeBlock.ts#L66-L70)? It looks like unused annotations are meant to be removed, but it only looks for _any_ mark definition and keeps it around if one exists.

The idea here is to allow marks that are 1) Decorators allowed by the schema like `strong`. or 2) marks that have a matching key in the markDefs array on the root of the block.

skogsmaskin (approved):
Thanks for this! I think these changes should be allright.

The test coverage should be pretty good in this package, so if they pass I'm pretty confident.

PR #3363, opened 2022-06-20
Smaller PT-input improvements

skogsmaskin:
### Description

Various smaller fixes for the PT input. Most importantly [simplify scrollToFocus hook](https://github.com/sanity-io/sanity/pull/3363/commits/97b72acd4449e085b8680c0fd5546e5f412de62a), which will make this hook more reliable and simpler after we got the pt member kind property.


bjoerge (approved):
LGTM!

PR #3366, opened 2022-06-20
feat(cli): allow specifying --api to filter which APIs to deploy

rexxars:
### Description

When defining multiple GraphQL APIs, you might want to only deploy certain defined APIs instead of all of them (see https://github.com/sanity-io/sanity/discussions/3344)

This PR allows you to specify an `id` for each API, and then reference it using the `--api <api-id>` flag when running `sanity graphql deploy`.

It also updates the help text for this command, which didn't match the command in v3.

### What to review

- That the changes makes sense
- That it works. Example `sanity.cli.ts`:

```ts
import {createCliConfig} from 'sanity/cli'

export default createCliConfig({
  api: {
    projectId: 'ppsg7ml5',
    dataset: 'blog',
  },

  graphql: [
    {id: 'ancient', tag: 'ancient', generation: 'gen1'},
    {id: 'default', tag: 'default', generation: 'gen3'},
  ],
})
```

### Notes for release

- Allow specifying a specific GraphQL API to deploy using the new `--api` flag

### DIFF
--- a/packages/@sanity/cli/src/types.ts
+++ b/packages/@sanity/cli/src/types.ts
@@ -207,6 +207,12 @@ export interface SanityJson {
 
 export interface GraphQLAPIConfig {
   /**
+   * ID of GraphQL API. Only (currently) required when using the `--api` flag
+   * for `sanity graphql deploy`, in order to only deploy a specific API.
+   */
+  id?: string
+
+  /**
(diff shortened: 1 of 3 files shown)


skogsmaskin (approved):
LGTM!

PR #3380, opened 2022-06-28
feat(studio): add not authenticated screen

hermanwikner:
This PR adds a screen that is displayed when a user is signed in to a studio without any roles (i.e. is unauthorized). 

The previous `NotAuthenticatedScreen`, which handled authorization, has been renamed to `AuthenticateScreen` – and the new screen added in this PR is now called `NotAuthenticatedScreen` instead. 

Screenshot 2022-06-28 at 13 52 57

### DIFF
--- a/packages/sanity/src/studio/AuthBoundary.tsx
+++ b/packages/sanity/src/studio/AuthBoundary.tsx
@@ -28,7 +28,7 @@ export function AuthBoundary({
   }, [activeWorkspace.auth])
 
   useEffect(() => {
-    activeWorkspace.auth.state.subscribe({
+    const subscription = activeWorkspace.auth.state.subscribe({
       next: ({authenticated, currentUser}) => {
         if (currentUser?.roles?.length === 0) {
           setLoggedIn('unauthorized')
@@ -40,6 +40,10 @@ export function AuthBoundary({
       },
       error: handleError,
     })
+
+    return () => {
+      subscription.unsubscribe()
+    }
   }, [activeWorkspace])
 
   if (loggedIn === 'loading') return <LoadingComponent />
--- a/packages/sanity/src/studio/AuthBoundary.tsx
+++ b/packages/sanity/src/studio/AuthBoundary.tsx
@@ -1,22 +1,26 @@
 import React, {useEffect, useState} from 'react'
-import {LoadingScreen, NotAuthenticatedScreen} from './screens'
+import {LoadingScreen, AuthenticateScreen, NotAuthenticatedScreen} from './screens'
 import {useActiveWorkspace} from './activeWorkspaceMatcher'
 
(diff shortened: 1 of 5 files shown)


packages/sanity/src/studio/screens/NotAuthenticatedScreen.tsx
```diff
@@ -1,15 +1,61 @@
...
+  if (error) throw error
+
+  const {activeWorkspace} = useActiveWorkspace()
+
+  const handleLogout = useCallback(() => {
+    activeWorkspace.auth.logout?.()
+  }, [activeWorkspace])
+
+  useEffect(() => {
+    activeWorkspace.auth.handleCallbackUrl?.().catch(handleError)
```
bjoerge, line 19: It's a bit unclear to me why we call this in response to any change in workspace auth.
Is it possible to let the workspace auth (store?) take care of this? Or, if not, could we move it to the event handler instead?

packages/sanity/src/studio/AuthBoundary.tsx
```diff
@@ -25,16 +29,27 @@ export function AuthBoundary({
 
   useEffect(() => {
     activeWorkspace.auth.state.subscribe({
-      next: ({authenticated}) => setLoggedIn(authenticated ? 'logged-in' : 'logged-out'),
+      next: ({authenticated, currentUser}) => {
```
bjoerge, line 32: I realise this isn't introduced by this PR, but we need to unsubscribe in the effect cleanup:
```js
useEffect(() => {
  const subscription = activeWorkspace.auth.state.subscribe(/*…*/)
  return () => {
      subscription.unsubscribe()
    }
  }, [activeWorkspace])
```

packages/sanity/src/studio/screens/NotAuthenticatedScreen.tsx
```diff
@@ -1,15 +1,61 @@
...
+  }, [activeWorkspace.auth])
+
+  useEffect(() => {
```
bjoerge, line 22: Also  a missing unsubscribe in effect cleanup here.

bjoerge (changes requested):
Would be great if you have time to fix the couple of missing unsubscribes in useEffect cleanup :)

packages/sanity/src/studio/screens/NotAuthenticatedScreen.tsx
hermanwikner: Thx 👍

packages/sanity/src/studio/AuthBoundary.tsx
hermanwikner: Thx 👍

rexxars (approved):

PR #3387, opened 2022-07-01
refactor(form-builder): add condition for name in upload progress

RitaDias:
### Description

Make sure that it doesn't try to read the name from a file when it is not available.

### Notes for release

Fix potential error when trying to access a file that is not existent


bjoerge (approved):
LGTM!

PR #3394, opened 2022-07-05
Restore PTE spellchecking defaults for Chromium

skogsmaskin:
### Description

This will have Chromium based browsers default to spell checking when spellCheck prop is `undefined`.

For other browsers this works as expected (user settings when undefined).

### Notes for release

Fix issue with Chromium based browsers where spell check in the PortableText editor didn't follow user settings when unspecified in schema.


bjoerge (approved):
LGTM!

PR #3400, opened 2022-07-08
feat(base): support presence for cookieless auth users

rexxars:
### Description

The browser `WebSocket` API does not support specifying HTTP headers to the initial HTTP connection. This means we cannot send the `Authorization` header that we normally do for non-cookie authentication methods, meaning anything over the bifur socket (such as presence) fails to work.

We recently implemented support for sending authorization credentials through the first frame to the socket (only available on `v2022-06-30` and higher), and have released a new version of the bifur client to allow for retrieving the token through an observable.

This PR upgrades the bifur client, and provides the `authToken$` observable as an option, which ensures the client authorizes before sending any subscribe/requests to the server.

### What to review

- That the presence feature works as intended using both cookies (chrome) and cookieless (brave, safari)
- That there are no excessive connection attempts against the `/socket` endpoint

### Notes for release

- Added support for the presence feature when not using cookies for authentication

### DIFF
--- a/packages/@sanity/base/src/client/bifur.ts
+++ b/packages/@sanity/base/src/client/bifur.ts
@@ -1,8 +1,10 @@
-import type {ClientConfig, SanityClient} from '@sanity/client'
-import {fromSanityClient} from '@sanity/bifur-client'
-import {versionedClient} from './versionedClient'
+import {fromUrl} from '@sanity/bifur-client'
+import {authToken$} from '../datastores/authState'
+import {getVersionedClient} from './versionedClient'
 
-export const bifur = fromSanityClient(
-  // The global Sanity client is guaranteed to have a dataset, thus the type cast
-  versionedClient as SanityClient & {config(): ClientConfig & {dataset: string}}
-)
+const bifurVersionedClient = getVersionedClient('2022-06-30')
+const dataset = bifurVersionedClient.config().dataset
+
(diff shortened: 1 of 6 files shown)


dcilke:
Something seems off 
image

bjoerge (approved):
Looks great!

packages/@sanity/base/src/client/bifur.ts
```diff
@@ -1,8 +1,10 @@
...
+import {getVersionedClient} from './versionedClient'
 
-export const bifur = fromSanityClient(
-  // The global Sanity client is guaranteed to have a dataset, thus the type cast
-  versionedClient as SanityClient & {config(): ClientConfig & {dataset: string}}
-)
+const bifurVersionedClient = getVersionedClient('2022-06-30')
+const dataset = bifurVersionedClient.config().dataset
+
+const url = bifurVersionedClient.getUrl(`/socket/${dataset}`).replace(/^http/, 'ws')
```
dcilke, line 8: Nit: we have `fromSanityClient` in the `bifur-client`

dcilke (commented):
Where are we sending the authorization message?

packages/@sanity/base/src/client/bifur.ts
rexxars: We do, but it doesn't accept the token observable. Will  leave it like this for now.

PR #3402, opened 2022-07-08
refactor(mutator): use strict typescript settings

rexxars:
### Description

Apologies for the huge PR.

This is a first attempt at migrating the mutator to using strict typescript settings. Given it is such an intregal part of the Sanity studio, I feel this is one of those pieces that should most definitely be using the strictest settings possible.

For the most part this has been a game of discovery, figuring out how everything ties together and which functions accept what information. It is now in a state where all the old tests pass.

There are a number of smaller code flow changes - this is usually for type narrowing in typescript, but in some cases I did not understand the flow and needed to rewrite it to better understand it.

I also took the liberty of removing an outdated readme, removing exports that we did not use internally, and removed tap as a dependency (it was using a somewhat odd combination of jest and tap).

### What to review

- The naming of types are not perfect, but I feel it is better than nothing. At the very least we can now use codemods/refactorings in VS Code to iterate on it. All of this should be considered internals - so we can afford to be slightly less clear for now.
- Any modifications to the code flow needs to be read through properly to ensure I didn't make any silly mistakes
- A full multi-user browser test is probably not the worst idea in the world

### Notes for release

None. Internal change.

### DIFF
--- a/packages/@sanity/mutator/src/document/BufferedDocument.ts
+++ b/packages/@sanity/mutator/src/document/BufferedDocument.ts
@@ -7,6 +7,9 @@ import type {Doc, Mut} from './types'
 
 const ONE_MINUTE = 1000 * 60
 
+/**
+ * @internal
+ */
 export interface CommitHandlerMessage {
   mutation: Mutation
   success: () => void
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Fabulous work! I'm afraid I don't have time to do an in-depth review until after the holidays, so feel free to merge for now and I can have a closer look over the summer if needed.

PR #3446, opened 2022-07-27
feat(dashboard): add docs banner to sanity tutorials widget

vicbergquist:
### Description

This adds a banner at the top of the sanity tutorials videos with a CTA to our docs site.

### What to review

- Start the test studio and make sure everything works as it should in the dashboard with the banner/

### Notes for release

n/a

### DIFF
--- a/packages/@sanity/dashboard/src/widgets/sanityTutorials/SanityTutorials.js
+++ b/packages/@sanity/dashboard/src/widgets/sanityTutorials/SanityTutorials.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import {Flex, Grid, Stack, Heading, Container} from '@sanity/ui'
+import {Flex, Grid, Stack, Heading, Container, Card, Text, Box, Button} from '@sanity/ui'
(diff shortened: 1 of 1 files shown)


ninaandal (changes requested):
Maybe fix this when the screen gets small? 

Screenshot 2022-07-28 at 11 54 22

vicbergquist:
> Maybe fix this when the screen gets small?
> 
> Screenshot 2022-07-28 at 11 54 22

what size is this? I'm already adjusting the layout like this `<Flex direction={['column', 'column', 'row']}>` 🤔

ninaandal:
Below `950px` ish, but it is with the right side banner open
Screenshot 2022-07-28 at 14 14 19

vicbergquist:
> Below `950px` ish, but it is with the right side banner open
> Screenshot 2022-07-28 at 14 14 19

Hm does any important devices have that viewport width? If not, I'm not sure if it makes sense to fix it with container queries

ninaandal (approved):
Yeah, makes sense 👍 LGTM otherwise!

PR #3452, opened 2022-07-28
feat(base): clearly indicate referenced documents

ninaandal:
### Description
Indicators that a document is a referenced document is added. Any document that has a reference to it will have a blue header/footer, a reference icon next to the title and a dialog that with information when publishing changes to a referenced document. 

A popover will be shown once for every user for every browser. Once the user dismiss it, it will not be shown again. 

These changes were necessary to inform the user of the consequences of making changes to a referenced document. 
Link to task: https://app.shortcut.com/sanity-io/story/21975/indicate-a-referenced-document-in-document-pane-list-view 

Screenshot 2022-08-01 at 12 50 24

### What to review

Go any file that is referenced in the studio to see the new features to indicate that the document is a referenced document. 
Check that the one time popover works. Press the publish button to see the dialog with a warning when publishing changes to a referenced document. 

### Notes for release

feat(base): clearly indicate referenced documents


packages/@sanity/desk-tool/src/panes/document/useDocumentIsReferenced.tsx
```diff
@@ -0,0 +1,13 @@
+import {createHookFromObservableFactory, getPublishedId} from '@sanity/base/_internal'
+import documentStore from 'part:@sanity/base/datastore/document'
+import {Observable} from 'rxjs'
+
+export const useDocumentsIsReferenced = createHookFromObservableFactory((documentId: string) => {
+  const totalClause = 'count(*[references($documentId)]) > 0'
```
sjelfull, line 6: Will this work for referencing documents that you don't have permissions for?

ninaandal: I don't know.. @rexxars, do you know?

packages/@sanity/desk-tool/src/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -18,118 +18,125 @@ import {DocumentHeaderTitle} from './DocumentHeaderTitle'
...
-      views,
-    } = useDocumentPane()
-    const {revTime: rev} = historyController
-    const {features} = useDeskTool()
-    const {index, BackLink, hasGroupSiblings} = usePaneRouter()
-    const contextMenuItems = useMemo(() => menuItems.filter(isMenuButton), [menuItems])
-    const [isValidationOpen, setValidationOpen] = React.useState<boolean>(false)
-    const showTabs = views.length > 1
-    const showVersionMenu = features.reviewChanges
+  forwardRef(
```
vicbergquist, line 28: cant really see what changed below because the linting also changed. Is it possible to revert the changes not related to this PR?

packages/@sanity/desk-tool/src/panes/document/useDocumentIsReferenced.tsx
rexxars: Good point - it does not, unfortunately. I've escalated this internally, we'll discuss further there.

packages/@sanity/desk-tool/src/panes/document/useDocumentIsReferenced.tsx
```diff
@@ -0,0 +1,13 @@
+import {createHookFromObservableFactory, getPublishedId} from '@sanity/base/_internal'
+import documentStore from 'part:@sanity/base/datastore/document'
+import {Observable} from 'rxjs'
+
+export const useDocumentsIsReferenced = createHookFromObservableFactory((documentId: string) => {
```
rexxars, line 5: I was a bit hasty in suggesting we create a separate hook for this - it looks like we probably want to reuse the [useReferringDocuments](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/desk-tool/src/components/confirmDeleteDialog/useReferringDocuments.ts#L140) hook, since it also handles cross-project references.

Ideally we'd lift the hook out of `confirmDeleteDialog` since it would then be used in multiple locations.

packages/@sanity/desk-tool/src/panes/document/statusBar/dialogs/ConfirmDialog.tsx
```diff
@@ -19,15 +20,17 @@ export function ConfirmDialog(props: {
   const {dialog, referenceElement} = props
 
   return (
```
rexxars, line 22: Maybe add a commenting explaining the theme color provider usage?

```suggestion
  // Wrapping in ThemeColorProvider to avoid the ConfirmDialog inheriting tone of the footer
  return (
```

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
```diff
@@ -0,0 +1,74 @@
...
+    setHidePopover(true)
+  }
+
+  return (
+    <ThemeColorProvider tone="default">
+      <Box marginLeft={2}>
+        <Popover
+          content={<InnerPopover onClose={onClose} />}
+          placement="bottom-start"
+          referenceElement={document.getElementById('referenceTitle')}
```
rexxars, line 36: This should use a reference (`useRef`) - multiple document panes can be open at the same time, which will create multiple elements with the same ID

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
```diff
@@ -0,0 +1,74 @@
...
+          portal
+          open
+        >
+          <Box display="flex">
+            <TitleText id="referenceTitle" tabIndex={0} textOverflow="ellipsis" weight="semibold">
```
rexxars: See above comment

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Box>
+        </Popover>
+      </Box>
+    </ThemeColorProvider>
+  )
+}
+
+function InnerPopover({onClose}: {onClose?: () => void}) {
+  return (
+    <StyledBox as={'div'}>
```
rexxars, line 56: Stylistically, we prefer that static string values use the regular XML-like notation, even if it has no impact on the actual output.

```suggestion
    <StyledBox as="div">
```

packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
```diff
@@ -67,7 +68,8 @@ export const DocumentPane = memo(function DocumentPane(props: DocumentPaneProvid
   const [templatePermissions, isTemplatePermissionsLoading] = useUnstableTemplatePermissions(
     getNewDocumentOptions()
   )
-  const isLoaded = isDocumentLoaded && !isTemplatePermissionsLoading
+  const [isReferenced, isReferenceLoading] = useDocumentsIsReferenced(options.id)
```
rexxars, line 71: Seeing the amount of places this needs to be ["prop-drilled"](https://blog.logrocket.com/solving-prop-drilling-react-apps), I am wondering if perhaps it would make sense to lift this up to the [DocumentPaneProvider](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/desk-tool/src/panes/document/DocumentPaneProvider.tsx) and attach it to the [DocumentPaneContextValue](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/desk-tool/src/panes/document/DocumentPaneContext.ts#L13). That way, you could utilize the `useDocumentPane()` hook in order to get the information instead of relying on the prop being passed down everywhere you might need it.

packages/@sanity/desk-tool/src/panes/document/documentPanel/DocumentPanel.tsx
```diff
@@ -63,6 +64,7 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
   const headerRect = useElementRect(headerElement)
   const portalRef = useRef<HTMLDivElement | null>(null)
   const [documentScrollElement, setDocumentScrollElement] = useState<HTMLDivElement | null>(null)
+  const [isReferencedDocument] = useDocumentsIsReferenced(documentId)
```
rexxars, line 67: I believe this will end up creating two listeners for the same information - I would prefer that we either receive the `isReferencedDocument` through a prop, or as stated above, through the `DocumentPaneContext`

packages/@sanity/desk-tool/src/panes/document/documentPanel/header/DocumentPanelHeader.tsx
rexxars: Prettier is "at fault" here - not much to do about it, I think. One alternative would be to refactor the code a bit so prettier doesn't wrap it further, heh

packages/@sanity/desk-tool/src/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -18,118 +18,125 @@ import {DocumentHeaderTitle} from './DocumentHeaderTitle'
...
+    (
+      {rootElement, isReferencedDocument}: DocumentPanelHeaderProps,
+      ref: React.ForwardedRef<HTMLDivElement>
+    ) => {
```
rexxars, line 32: This is very nitpicky and only addressing Victorias note above - but theoretically I think you might force prettier to keep formatting if you do something like this:

```suggestion
  forwardRef((headerProps: DocumentPanelHeaderProps, ref: React.ForwardedRef<HTMLDivElement>) => {
  const {rootElement, isReferencedDocument} = headerProps
```

rexxars (changes requested):

packages/@sanity/desk-tool/src/panes/document/DocumentPane.tsx
ninaandal: This makes SO much sense - thanks for pointing this out 🙌

packages/@sanity/desk-tool/src/panes/document/useDocumentIsReferenced.tsx
ninaandal: Makes sense - I'll delete the other hook and use `useReferringDocuments` instead 🙌

packages/@sanity/desk-tool/src/actions/PublishAction.tsx
```diff
@@ -152,6 +189,15 @@ export const PublishAction: DocumentActionComponent = (props) => {
...
       : title,
     shortcut: disabled || publishScheduled ? null : 'Ctrl+Alt+P',
-    onHandle: handle,
+    onHandle: documentIsReferenced ? () => setShowConfirmPublishReferencedDocument(true) : handle,
+    dialog: showConfirmPublishReferencedDocument ? dialog : undefined,
   }
 }
+
+const LinkBox = styled.div`
+  border-color: #e8f1fe;
```
vicbergquist, line 198: I think this should use a color from the theme values instead. Same with the grey below for the border 👍

packages/@sanity/desk-tool/src/components/pane/PaneHeader.styles.tsx
```diff
@@ -79,3 +79,18 @@ export const TabsBox = styled(Box)(({theme}: {theme: Theme}) => {
     }
   `
 })
+
+export const LinkCircle = styled.div`
+  border: 1px solid grey;
```
vicbergquist, line 84: Same here, should use theme colors!

packages/@sanity/desk-tool/src/components/pane/PaneHeader.styles.tsx
```diff
@@ -79,3 +79,18 @@ export const TabsBox = styled(Box)(({theme}: {theme: Theme}) => {
...
+  border-radius: 50%;
+  width: 24px;
+  height: 24px;
+`
+
+export const HorizontalLine = styled.div`
+  border-top: 1px solid #ced2d9;
```
vicbergquist, line 91: should use theme color

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
```diff
@@ -0,0 +1,76 @@
...
+            </LinkCircle>
+          </Box>
+          <Heading size={1}>This is a referenced document</Heading>
+        </Flex>
+        <Text>Changes to this type of document may affect other documents that reference it</Text>
+      </Box>
+      <HorizontalLine />
+      <Box as="li" padding={3}>
+        <Stack space={2}>
+          <Button onClick={onClose} tone="primary" text="Got it" />
```
vicbergquist, line 71: It's currently not possible to close the popover with keyboard only, would be great if this button gets focus when it's rendered so that it becomes keyboard accessible

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <LinkCircle>
+              <LinkIcon fontSize={'24'} />
+            </LinkCircle>
+          </Box>
+          <Heading size={1}>This is a referenced document</Heading>
+        </Flex>
+        <Text>Changes to this type of document may affect other documents that reference it</Text>
+      </Box>
+      <HorizontalLine />
+      <Box as="li" padding={3}>
```
vicbergquist, line 69: Any reason this is an `li` element? If yes, where is the parent `ul` or `ol`? Should be added if it's on purpose 👍

vicbergquist (changes requested):
Great job! I have a couple of comments, mostly about using theme values for colors instead of hard-coded ones.

 Let me know if you have questions!

packages/@sanity/desk-tool/src/components/paneItem/ReferencedDocHeading.tsx
ninaandal: I'll add the `autoFocus` prop to it 😊

vicbergquist (approved):

rexxars (approved):
👍

PR #3453, opened 2022-07-29
feat(export): check sha1 hash first

dcilke:
### Description

Swap the logic for validating assets to prefer sha1 validation. This is to address an issue where the md5 of an asset is being misreported.

### What to review

Dataset export works


rexxars (approved):
👍

PR #3462, opened 2022-08-01
fix(export): allow either hash check

dcilke:
### Description

Allows export hash check to succeed if either sha1 or md5 are the same.

### What to review

Dataset export works


rexxars (approved):
👍

rexxars (commented):
Maybe leave a comment for future readers why we check both hashes?

PR #3465, opened 2022-08-02
V3 portable text editor

skogsmaskin:
### Description

Performance and stability refactor of the `portable-text-editor` package and the Sanity PT-input. Especially stability in a collaborative setting is improved.

* Refactor withPatches plugin to execute remote patches on debounced state document while typing. Convert remote patches to text operations. This way we don't need a separate selection adjustment, but text operations will make a natural selection handling through the operations. This also helps with performance as we don't need to re-render quite as much (we only need to check consistency when applying a new props value).

* Improve fullscreen and activation re-rendering performance.

* Various refactors and improvements.


packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
```diff
@@ -180,6 +185,7 @@ export default class CollaborationEnvironment extends NodeEnvironment {
               ])
             },
             pressKey: async (keyName: KeyInput, times?: number) => {
+              await delay(generateRandomInteger(0, 100))
```
bjoerge, line 188: Can this potentially give random test failures that are hard to debug? Would it be better to use timer mocks here?

bjoerge (approved):
Looks good to me! Kudos for the increased test coverage too!

packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
skogsmaskin: This is to make it a bit more unpredictable actually, to test that the collaborative tests still pass even if there are random delays in the tests users keypresses.

PR #3466, opened 2022-08-02
feat!(datastores): remove deprecated document APIs from v1

rexxars:
### Description

Removes the deprecated document document APIs from ancient times

### What to review

- That this is still okay to remove


bjoerge (approved):
Yes yes yes

PR #3469, opened 2022-08-03
fix(dashboard): use correct property for feed item keys

rexxars:
### Description

The tutorials dashboard widget prints two warnings currently - one caused by missing keys for feed items (it was using `.id` instead of `._id`, and one about a required proptype that was missing.

The widget seems to work just fine without the prop, and as we're moving away from the prop-types module I figured I might as well just remove it from use in this component.

### Notes for release

- Fixed a few warnings being emitted from the tutorials dashboard widget


ninaandal (approved):
lgtm 👍

PR #3470, opened 2022-08-03
Chunk document availability requests

skogsmaskin:
### Description

The URL for testing document availability builds up by appending document ids in a comma separated list in the document URI itself. This potentially builds up very large URLs resulting in 414 (URL too long) HTTP errors.

This will chunk those ids if they exceed the limit for a URL and do several requests from the needed chunks of ids.

### What to review

That it works like before and actually chunks when many ids are occurring in the list.

### Notes for release

Fixed a bug where asking for document availability for many document ids resulted in a 414 (URL too long) error.


rexxars:
Turns out there were already chunking in place, but the limit was set arbitrarily - 217c5586855fb643b02bab0be741a729f3aaf154 supersedes this PR by altering the existing chunking to use a strategy which accounts for the size of document IDs.

PR #3482, opened 2022-08-08
fix(desk): correct prop typings for pane view components

rexxars:
### Description

The `S.view.component()` builder has incorrectly typed component - it was `UserComponent`, which is for custom _panes_ - while what we actually want is a custom _document view_ component. This caused some confusion when wanting to create these components, but where the props passed to it was simply wrong. While fixing the typings, I also wanted to clean up the typings for `PaneView` into separate `FormView` and `ComponentView` types, which lead to some downstream changes - apologies for the slightly hard to read PR here.

### Notes for release

- Fixed incorrect types for the React component passed to `S.view.component()`


bjoerge (approved):
Nice! LGTM!

PR #3484, opened 2022-08-08
feat(desk): add reference count and learn more link to references indication

ninaandal:
### Description
Implemented references indicator design changes. 
Link to story: https://app.shortcut.com/sanity-io/story/22638/indicate-a-referenced-document-part-2 

The tooltip now include a reference count, stating how many documents reference the document that is viewed. The reference count is also shown in the one-time popover, stating how many documents will be affected if the document is edited.

The dialog for the PublishAction button that appeared when publishing a referenced document is also removed after user feedback. 

Screenshot 2022-08-09 at 09 51 41

Screenshot 2022-08-09 at 09 54 11

### What to review
A referenced document - the tooltip on link icon next to text and the popover text and "Learn more"-button.

### Notes for release
feat(base): Reference indicators now include a reference count for affected documents

### DIFF
--- a/packages/@sanity/desk-tool/src/actions/PublishAction.tsx
+++ b/packages/@sanity/desk-tool/src/actions/PublishAction.tsx
@@ -41,7 +41,7 @@ export const PublishAction: DocumentActionComponent = (props) => {
   const {publish}: any = useDocumentOperation(id, type)
   const validationStatus = useValidationStatus(id, type)
   const syncState = useSyncState(id, type)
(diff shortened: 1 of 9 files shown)


sjelfull (changes requested):
Looks good for the most part. Could we adjust the padding of the popover content slightly so that it matches the design?

image

If we:
1) Reduce padding so that content and buttons below is aligned
2) Make the gap between buttons the same as the padding around the buttons the same
It should be good to go

rexxars (approved):
👍

PR #3487, opened 2022-08-08
fix(form): broken array grid layout

hermanwikner:
### Description

This PR fixes so that the array items are rendered in a grid layout when the schema type has `options: { layout: 'grid'}`.

### Notes for release

fix(form): fix broken array grid layout


rexxars (commented):
👍 Thanks!

PR #3489, opened 2022-08-08
feat(studio): add cors origin error screen

hermanwikner:
### Description

This PR adds CORS origin error handling. A a dialog is displayed when the error occurs.

Screenshot 2022-08-09 at 11 06 27

### What to review

To test, run the studio in development mode on a port that is not configured in the projects CORS origin settings. Go to `sanity.cli.ts` and add e.g:

```js
  server: {
    port: 3008,
  },
```

### Notes for release
feat(studio): add cors origin error screen


packages/sanity/src/datastores/cors/checkCors.ts
```diff
@@ -0,0 +1,21 @@
...
+    .catch(() => false)
+
+  const promise = Promise.all([pingRequest, userRequest])
+    .then((res) => ({
+      isCorsError: res[0] && !res[1],
+      pingResponded: res[0],
+    }))
+    .catch((e) => {
+      throw e
+    })
```
bjoerge, line 18: This isn't necessary since the error will propagate on the promise anyway
```suggestion
```

packages/sanity/src/datastores/cors/checkCors.ts
```diff
@@ -0,0 +1,21 @@
...
+  const userRequest = client
+    .request({uri: '/users/me', withCredentials: false})
+    .then(() => true)
+    .catch(() => false)
+
+  const promise = Promise.all([pingRequest, userRequest])
+    .then((res) => ({
+      isCorsError: res[0] && !res[1],
+      pingResponded: res[0],
+    }))
```
bjoerge, line 15: I'd like to suggest using destructing here to improve readability :)
```suggestion
    .then(([pingResponded, userResponded]) => ({
      isCorsError: pingResponded && !userResponded,
      pingResponded,
    }))
```

packages/sanity/src/studio/screens/CorsOriginErrorScreen.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function CorsOriginErrorScreen(props: CorsOriginErrorScreenProps) {
+  const {projectId} = props
+
+  const corsUrl = `https://sanity.io/manage/project/${projectId}/settings/api`
+
+  const origin =
+    window.location.origin ||
+    window.location.href.replace(new RegExp(`${window.location.pathname}$`), '')
```
bjoerge, line 15: Under what circumstances (if any?) will `window.location.origin` not be defined?

bjoerge (approved):
Looks great! Left a couple of minor suggestions.

This has probably been considered - but a future improvement of this could be to do this check not only for the user request, but for all requests done by the Studio client. In principle the Studio could fetch data from multiple projects that may not have the current origin allowlisted. It may not even do any requests to the user endpoint at all. If that's the case it would be good to show the same helpful message to the user :)

But yeah, this is a great start and gives parity with v2!

packages/sanity/src/datastores/cors/checkCors.ts
hermanwikner: Agree, cleaner :)

packages/sanity/src/studio/screens/CorsOriginErrorScreen.tsx
hermanwikner: I think this is wrong by me – it should be `window.location.origin` only

PR #3490, opened 2022-08-08
fix(server): handle jsx in `.js` files

rexxars:
### Description

When using a custom `_document.js` file, the esbuild loader did not handle JSX. This was made worse by the importer silently swallowing all errors, instead of just the "not found" ones.

This PR tells esbuild to use the tsx loader even for .js files, and will throw an error on any error that is not "module not found" - eg syntax errors and such.

### What to review

- Custom `_document.js`/`_document.tsx` files work as expected

### Notes for release

- Fixed a bug where a custom `_document.js` with JSX would not work, and syntax errors would silently be swallowed


bjoerge (approved):
Nice!

PR #3492, opened 2022-08-09
fix: ignore blocks inside list items

snorrees:
### Description

This fixes a bug where p-tags (and other block elements) inside li elements where incorrectly hoisted out of the li element.

With this change blocks are ignored, and block children are added directly to the listItem instead.

### What to review

Testcase should cover the changes introduced; from the code comment it seems like this usecase was already accounted for, but was checking element tag name instead of parent tag name.

### Notes for release

[@sanity/block-tools] Blocks inside list items are now ignored. Child element will be added directly to the list item instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,10 @@ jspm_packages
 # macOS finder cache file
 .DS_Store
 
+# Intellij
+.idea
+*.iml
+
 # VS Code settings
 .vscode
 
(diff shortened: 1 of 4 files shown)


juice49 (approved):

skogsmaskin (approved):
Thank you for this.
LGTM!

PR #3493, opened 2022-08-09
fix(cli): work around edge-case when unable to resolve @sanity/core

rexxars:
### Description

In certain rare cases, like when initing a Sanity project inside of an existing yarn workspaces monorepo, the CLI fails to resolve the newly installed `@sanity/core` module during the init process. When this happens, it gives a less than helpful error message: 

> TypeError [ERR_INVALID_ARG_TYPE]: The "id" argument must be of type string. Received null

The only reason why we resolve `@sanity/core` is in order to use the `configcheck` and `dataset import` commands, which isn't technically speaking necessary for Sanity to work.

This PR therefore takes the lazy approach: If unable to resolve the module, it silently registers this happening and instead prompts the user to run `sanity dataset import` themselves after the init is complete. While not _ideal_, it is an edge case, and is fixed more properly in v3.

### What to review

That `sanity init` works as expected:
- Within and without yarn workspaces
- With and without having selected the import data choice
- Initing to the current directory, and initing to a different directory

### Notes for release

- Fixed a case where `sanity init` would fail if ran inside of a yarn workspaces/monorepo setup in certain cases


bjoerge (approved):
👍

sjelfull (approved):

PR #3496, opened 2022-08-10
feat(workspaces): improve validation, allow single workspace without `name`, `basePath`

rexxars:
### Description

I was looking in to an issue where insufficient validation lead to a rather confusing error message when one of many workspaces was missing a base path. While addressing, I realized there was a general lack of validation, so I took the liberty of adding validation for workspace names as well.

The types for workspaces was a bit strict - if you have a single workspace, we technically allowed you not to pass a name or basePath, but the types made it seem like they were required. This is now loosened - with a single workspace, you can skip the name. With the array form of defining workspaces, you will need a name and basePath.

I also moved the workspace validation to the `prepareConfig` step in order to catch errors earlier and in a central location, as well as tidied up the error screens for this error as well as "workspace not found" errors. They are not perfect, but at least look slightly more similar to the schema error screen.

### What to review

That workspaces still work as expected, and that the validation rules make sense (they do a bit of excessive looping, but it is likely that this will be run for less than 10 workspace, so I am not too worried about it).

### Notes for release

- Improved validation and types for workspaces


bjoerge (approved):
Looks great!

PR #3500, opened 2022-08-10
 feat(form): add correct field level UI

hermanwikner:
### Description
This PR adds logic to determine when a field should have the "level UI" (that is, padding to left + border). 

### What to review
Do the level UI appear on places where it is expected? (Tip: compare with v2)

### Notes for release
feat(form): add correct field level UI

### DIFF
--- a/packages/sanity/src/form/studio/inputResolver/inputResolver.tsx
+++ b/packages/sanity/src/form/studio/inputResolver/inputResolver.tsx
@@ -26,6 +26,8 @@ import {resolveArrayInput} from './resolveArrayInput'
 import {resolveStringInput} from './resolveStringInput'
 import {resolveNumberInput} from './resolveNumberInput'
 import {defaultInputs} from './defaultInputs'
+import {getArrayFieldLevel, getObjectFieldLevel} from './helpers'
+import {isObjectField} from '../../utils/asserters'
 
 function resolveComponentFromTypeVariants(
   type: SchemaType
@@ -106,10 +108,12 @@ function PrimitiveField(field: FieldProps) {
 }
 
 function ObjectOrArrayField(field: ObjectFieldProps | ArrayFieldProps) {
+  const level = isObjectField(field) ? getObjectFieldLevel(field) : getArrayFieldLevel(field)
+
   return (
     <FormFieldSet
       data-testid={`field-${field.inputId}`}
-      level={field.level}
+      level={level}
       title={field.title}
       description={field.description}
       collapsed={field.collapsed}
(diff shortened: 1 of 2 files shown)


packages/sanity/src/form/studio/inputResolver/inputResolver.tsx
```diff
@@ -106,10 +107,12 @@ function PrimitiveField(field: FieldProps) {
 }
 
 function ObjectOrArrayField(field: ObjectFieldProps | ArrayFieldProps) {
+  const level = getFieldLevel(field as ObjectFieldProps)
```
bjoerge, line 110: Would be a tad safer to do a type assertion here:
```suggestion
  const level = isObjectField(field) ? getFieldLevel(field) : field.level
```
(You'll need to add this import as well: `import {isObjectField} from '../../utils/asserters'`)

bjoerge (approved):
Looks good! Just a smol suggestion to use a type asserter instead of a static cast.

packages/sanity/src/form/studio/inputResolver/inputResolver.tsx
hermanwikner: Thanks!

However, that won't pass array fields to the `getFieldLevel` helper. I think it needs to be something like this:

`const level = isObjectField(field) || isArrayField(field) ? getFieldLevel(field) : field.level`

But that results TS errors. Any advice?

hermanwikner: Maybe better to split them up, like `getObjectFieldLevel` and `getArrayFieldLevel`. And do something like:

```js
  const level = useMemo(() => {
    if (isObjectField(field)) {
      return getObjectFieldLevel(field)
    }

    if (isArrayField(field)) {
      return getArrayFieldLevel(field)
    }

    return field.level
  }, [field])

```

bjoerge: >Maybe better to split them up, like getObjectFieldLevel and getArrayFieldLevel. And do something like:
> const level = isObjectField(field) ? getObjectFieldLevel(field) : getArrayFieldLevel(field)

Ah, yes, this would be the way to go - the TS errors you got was because array fields doesn't have a `fields` property, and that would actually become a runtime error with the earlier type cast.

hermanwikner: Updated! 👍

bjoerge (approved):
🙌🏼

PR #3502, opened 2022-08-10
fix(form): check for invalid image/file values, provide reset option

rexxars:
### Description

When an incorrect value is set for an image/file `asset` field, the inputs crash because it expects the shape to be correct. This PR upgrades the asset utils in order to use some looser type checks, and uses these to ensure the image asset is correct. If not, it will show a warning with a reset option, which unset the `asset` field. On images, it will also unset `crop` and `hotspot` since they are likely invalid for whatever image is selected next.

I also made the action menu for file/images more forgiving of a missing `copyUrl`/`downloadUrl` - automatically hiding these options if the values are undefined.

Screenshot:
image

### How to reproduce
- Change an image field to an object, and include an `asset` field which is a reference to something that is not an image
- Use the input to select a non-image reference
- Switch back to an image type, removing the `asset` field

### Notes for release

- Fixed an issue where incorrect image and file shapes (references to non-assets) would crash the studio

### DIFF
--- a/packages/sanity/src/form/inputs/files/common/ActionsMenu.tsx
+++ b/packages/sanity/src/form/inputs/files/common/ActionsMenu.tsx
@@ -11,8 +11,8 @@ interface Props {
   onReset: MouseEventHandler<HTMLDivElement>
   accept: string
   directUploads?: boolean
-  downloadUrl: string
-  copyUrl: string
+  downloadUrl?: string
+  copyUrl?: string
 }
 
 export function ActionsMenu(props: Props) {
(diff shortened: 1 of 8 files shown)


bjoerge (approved):
Very nice!

packages/sanity/src/form/inputs/files/common/ActionsMenu.tsx
```diff
@@ -11,8 +11,8 @@ interface Props {
   onReset: MouseEventHandler<HTMLDivElement>
   accept: string
   directUploads?: boolean
-  downloadUrl: string
-  copyUrl: string
+  downloadUrl?: string
+  copyUrl?: string
```
RitaDias, line 15: In what situation would we not want the downloadUrl / copyUrl be optional?

PR #3503, opened 2022-08-10
fix(form): use most specific title for schema type in modal header

rexxars:
### Description

When the schema type used for array members did not have a title, it displayed `Edit undefined` in the header. This PR makes it look up the most specific title in the schema type chain, falling back to the schema type name when unable to go any further. 

I also investigated a bit and it looks like `file` and `image` etc never had a title. Was this intentional? I've added titles to all the base schema types - maybe _that_ is the real fix here? Let me know if that is a bad idea, or if we should drop the title lookup helper and just stick with the titles in the schema definitions 👍  

Before:
image

After:
image

### Notes for release

- Fixed a bug where object modals would sometimes display "Edit undefined" in the header


bjoerge (approved):
🙌🏼

PR #3505, opened 2022-08-11
fix(form): focus issue on `ArrayItem`

hermanwikner:
### Description

This PR fixes an issue where the menu button in `ImageInput` lost focus when in an `ArrayItem`, and therefore could not be used. This can be considered a quick fix – and I think this is something we should try to improve at a later point.

The problem seemed to be that `ArrayItem` took focus, and the solution here is to only enable focus of `ArrayItem` when the dialog of that item is closed.

### What to review

1. Navigate to [this document in the test-studio](http://localhost:3333/test/content/input-standard;arraysTest;dc22742c-6b3f-4b68-a03c-d43c7c411e52) and scroll down to **Image array (with defaults)**.
2. Add and then open the item
3. The `Select` menu on the `ImageInput` should be usable.

### Notes for release

fix(form): focus issue on `ArrayItem`


bjoerge (commented):
This is a great pragmatic fix, although I now wonder why we render the dialog inside the item element in the first place. Curious whether moving the dialog to outside of the <ArrayItem would be an alternative (and possibly less hacky) way to fix the issue?

hermanwikner:
> This is a great pragmatic fix, although I now wonder why we render the dialog inside the item element in the first place. Curious whether moving the dialog to outside of the <ArrayItem would be an alternative (and possibly less hacky) way to fix the issue?

I agree, I don't really see why it is rendered inside `ArrayItem`. However, I tested to move the dialog outside of `ArrayItem` but the focus issue persisted. @bjoerge

bjoerge (approved):
Aight, thanks for checking :)

PR #3506, opened 2022-08-11
Make schema warnings available through navbar

bjoerge:
### Description
Currently we do not show any schema warnings in v3 studios *unless* there's also schema errors. This adds a button to easily view schema errors to the global navbar. This is only visible in dev mode (so you won't be able to see it in action on the preview deployment for this PR, but it it is enabled in this build: https://test-studio-itunv5byl.sanity.build/)
image

### Notes for release
- Schema warnings will now be accessible via the global nav bar in development mode


packages/sanity/src/studio/screens/schemaErrors/SchemaProblemGroups.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const ret = []
+
+    for (const problemGroup of problemGroups) {
+      for (const problem of problemGroup.problems) {
+        ret.push({group: problemGroup, problem})
+      }
+    }
+
+    return ret
+  }, [problemGroups])
```
rexxars, line 37: Might be easier to use `React.memo` on the component here, since it is the only prop received?

packages/sanity/src/studio/components/navbar/schemaWarnings/SchemaWarningsButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+    ) || []
+
+  const popoverProps = useMemo(
+    () => ({
+      constrainSize: true,
+      portal: true,
+      scheme: scheme,
+    }),
+    [scheme]
+  )
```
rexxars, line 39: I can't see this used anywhere?

packages/sanity/src/studio/components/navbar/schemaWarnings/SchemaWarningsButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <Inline padding={2} space={2}>
+          <InfoOutlineIcon />
+          <Text muted size={1}>
+            Note: this message is only displayed during development
+          </Text>
+        </Inline>
+      ),
+      onClickOutside: handleClose,
+      onClose: handleClose,
+      scheme: 'light',
```
rexxars, line 67: Always light? If intentional, can we put a comment here to explain why?

rexxars (changes requested):
This is a fantastic idea! Couple of thoughts:

1. This message is helpful, but hard to spot at the bottom:
_Note: this message is only displayed during development_
Perhaps we can move it to the top to be gain more visibility?

2. I would love for the warnings to be sorted somehow - I guess they might implicitly be sorted by schema type? If so, I suppose that's alright! If not, sorting by schema type or type of error would be amazing.
3. I could see us having _config warnings_, so it would be nice if the dialog reflected this instead of explicitly saying "schema errors".
4. The "schema warning" header on each warning is a bit repetitive - perhaps we could instead show a "Schema warnings" header above the list?

packages/sanity/src/studio/components/navbar/schemaWarnings/SchemaWarningsButton.tsx
bjoerge: Copy paste err, thanks for spotting!

packages/sanity/src/studio/screens/schemaErrors/SchemaProblemGroups.tsx
bjoerge: This was already there from earlier (this file has been slightly modified + renamed and git doesn't seem to have picked it up).

packages/sanity/src/studio/components/navbar/schemaWarnings/SchemaWarningsButton.tsx
bjoerge: Copy paste from the ChangelogDialog - I've asked @hermanwikner if he knows why we need it there - will remove & update with a comment.

bjoerge: Ah - I realised now why it's needed: the navbar is dark schemed by default, so the dialog will inherit dark mode from it. That's an interesting issue - I guess what we'd want to express here is more like `scheme: "opposite"` or `scheme: "reset"` or something… 🤔

packages/sanity/src/studio/components/navbar/schemaWarnings/SchemaWarningsButton.tsx
```diff
@@ -0,0 +1,101 @@
...
+        selected={isDialogOpen}
+      />
+      {isDialogOpen && (
+        <Dialog
+          header={`Found ${groupsWithWarnings.length} warnings with the current schema`}
+          width={2}
+          {...dialogProps}
+          id={dialogId}
+        >
+          <Stack space={4} padding={3}>
```
hermanwikner, line 94: This `Stack` is unnecessary since the `SchemaProblemsGroup` has its once `Stack`. 

I suggest changing it to a `Box` and bump to `padding` to `4`!

```js
<Box padding={4}>
  <SchemaProblemGroups problemGroups={groupsWithWarnings} />
</Box>
```

hermanwikner:
The logo is not centered in mobile mode. I suggest, as with the changelog, to not show these warnings in mobile mode. This could be fixed by updating the `shouldRender` condition to `mediaIndex > 1 && isDev` in `Navbar` :)

Screenshot 2022-08-12 at 10 41 21

bjoerge:
Great feedback, thanks!

> 1. Perhaps we can move it to the top to be gain more visibility?

Like so? 
image

> 2. I would love for the warnings to be sorted somehow

I think they are currently appearing in the same order they are added to the schema. Each schema type are grouped together. I can sort them alphabetically by schema type name if you think that's better. The underlying data structure we use here groups all issues from the same "path" together, but at some point a decision was made to list them out flat/ungrouped in this view. Let's improve on this in a future iteration?

> 3. would be nice if the dialog reflected this instead of explicitly saying "schema errors".

I've changed the name of this thing to be "Configuration issues" now. Makes a lot of sense - great suggestion!

> 4. The "schema warning" header on each warning is a bit repetitive - perhaps we could instead show a "Schema warnings" header above the list?

I've changed the title to display the name and type of the schema type with the issue, I think helps, although it still repeats for multiple warnings for the same schema type, but given that we decided at some point do show these ungrouped I'm hesitant to do too much work in this area.

image

bjoerge:
> The logo is not centered in mobile mode.

Great suggestion! Fixed.

bjoerge:
@rexxars, @hermanwikner Here's a new deployment with your feedback addressed: https://test-studio-no0kiw57n.sanity.build/

hermanwikner (approved):
💯

rexxars (approved):
👍 💯

PR #3507, opened 2022-08-12
feat(schema): Add validation to disallow array members with same name as built-in types

bjoerge:
### Description
Currently it's possible to define inline object types in arrays that shares the same name as a built-in type (e.g. reference, image) - this causes problems with input resolving, and its generally a bad idea to have, for example, an inline array member type named "reference".

image

This change introduces a hard studio error if we detect array members with the same name as a builtin type. It also adds a warning if we find an array member that has the same name as another root level schema type, since it might lead to confusion having two possibly incompatible data types that shares the same `_type` name, and therefore something we would advice against.

image

### What to review
Verify that defining array members with the same name as a built-in type fails the studio with an appropriate error.

For example, the following schema should fail:
```js
export default {
  name: 'author',
  type: 'document',
  title: 'Author',
  fields: [
    {
      type: 'array',
      name: 'contents',
      title: 'Contents',
      of: [
        {
          type: 'object',
          name: 'reference', // <-- 💥 
          title: 'Text',
          fields: [{name: 'title', type: 'string'}],
        },
      ],
    },
  ],
}
```

Help articles to review:
- https://www.sanity.io/help/schema-array-of-type-builtin-type-conflict
- https://www.sanity.io/help/schema-array-of-type-global-type-conflict

### Notes for release

- Add stricter handling of array members with same name as one of the built-in types


rexxars (approved):
Great addition, and really appreciate the thoughtful comments!

PR #3508, opened 2022-08-13
Fix v3 tests

rexxars:
This fixes a wide variety of currently failing tests because of refactors and such. Much of it is based on an expanded  TestProvider from the form tests in order to make it more reusable across test suites.

I also:
- Expanded on the `useDocumentType` tests with the unimplemented `.todo()` tests
- Silenced a few console.warn/console.error outputs from being printed during tests
- Brought in a few fixes from v2 to the `@sanity/export` package as they were causing failing tests
- Upgraded nock used in `@sanity/import` to silence a node deprecation message
- Refactored the CLI test because it needed to work with v2 and v3, and Vite has slightly different behavior
- Added some steps in GitHub actions workflow in order to download Chrome and allow the puppeteer tests to pass
- Temporarily allowed failures on the ETL job. It currently doesn't understand namespace members other than functions, and I am blocked on this until I can discuss which approach to take in extracting this in a way that doesn't require reimplementing every member definition afresh. Will revisit as early as possible.
- Refactored server code to not require a virtual module for the entry (less custom code is better code, and fixes a build issue I saw while working on the CLI tests)

The only two tests I have disabled or deleted are:
- [StudioProvider.test.tsx](https://github.com/sanity-io/sanity/commit/9a5419a59550fa6e22e76ade2069581df95826a0#diff-7c45c4f50cc9f1031e545b6ff4ad29d064ea1b91221b19cf105683dad3d03fe1) - it didn't seem to be testing anything in particular aside from not crashing. I would rather bring it back once we know what we want to test.
- A validation test for nested marks, which I couldn't quite figure out. This one we should investigate more and ensure that v3 validates this correctly, but I didn't feel it was worth blocking the rest of these fixes.


packages/sanity/src/conditional-property/useConditionalProperty.test.tsx
```diff
@@ -1,32 +1,10 @@
 import {renderHook} from '@testing-library/react-hooks'
```
bjoerge, line 1: we might be able to remove these tests - they should be obsolete after resolving conditional properties as part of preparing the form state now. Made a note to check if this is the case and can do that separately.

bjoerge (approved):
Really great work! Just had a couple comments while reading through - but good to merge from what I can tell.

packages/sanity/src/conditional-property/useConditionalProperty.test.tsx
rexxars: Awesome, thanks for checking!

scripts/testCli.js
```diff
@@ -68,23 +67,28 @@ if (!githubWorkspace && !skipDelete) {
   // For debugging clarity
   console.log('Base path is %s', basePath)
 
-  // Schedule cleanup tasks before exiting
-  process.on('exit', cleanup)
+  // Allow running the Sanity CLI tool without specifying absolute path every time
+  const binPath = path.join(basePath, 'packages', '@sanity', 'cli', 'bin', 'sanity')
 
   // Require a clean slate at startup
   cleanup()
```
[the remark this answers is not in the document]
rexxars: Good catch - yes it should!

PR #3511, opened 2022-08-15
fix(form): pass layout option from array input

bjoerge:
### Description

This change introduces an layout option to the array item component and passes it from the array schema type where it's configured

### What to review
Make sure arrays configured with grid layout still works, and that the regression introduced by the previous change is fixed

### Notes for release

N/A


hermanwikner (approved):
Looks good, thank you! 🙏

PR #3512, opened 2022-08-15
chore(pkg-utils): support tsconfig build target

mariuslundgard:
### Description

`pkg-utils bundle` will now respect the `target` defined in `tsconfig.json`. If not defined, it will default to `ESNext`.

### What to review

- Make sure that `@sanity/portable-text-editor` bundles files compatible with ES2019.
- Make sure that other packages build with the desired ES target.

### Notes for release

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
     "history-server": "^1.3.1",
     "husky": "^7.0.4",
     "jest": "^27.4.3",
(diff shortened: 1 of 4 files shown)


skogsmaskin (approved):
Works as expected! Tnx.

PR #3513, opened 2022-08-15
feat(sanity): add `document.unstable_languageFilter` to Config API

mariuslundgard:
### Description

This PR adds a new config property for configuring a custom language filter component in v3. This was possible in v2, using a custom part (which is now deprecated).

With this change, the language filter can be configured like this:

```tsx
import {SelectIcon, TranslateIcon} from '@sanity/icons'
import {Button} from '@sanity/ui'
import {createConfig} from 'sanity'

export default createConfig({
  document: {
    unstable_languageFilter: [CustomLanguageFilter]
  }
})

function CustomLanguageFilter() {
  return <Button icon={TranslateIcon} iconRight={SelectIcon} mode="ghost" text="en_US" />
}
```

This will render a language filter component in the document pane header:
image

### What to review

- Please review the implementation and the DX.


packages/sanity/src/config/configPropertyReducers.ts
```diff
@@ -161,3 +165,24 @@ export const imageAssetSourceResolver: ConfigPropertyReducer<AssetSource[], Conf
...
+ */
+export const _documentLanguageFilterReducer: ConfigPropertyReducer<
+  _DocumentLanguageFilterComponent[],
+  _DocumentLanguageFilterContext
+> = (prev, {document}, context) => {
+  const resolveDocumentLanguageFilter = document?.unstable_languageFilter
+  if (!resolveDocumentLanguageFilter) return prev
+
+  if (typeof resolveDocumentLanguageFilter === 'function')
+    return resolveDocumentLanguageFilter(prev, context)
```
snorrees, line 180: Nitpick: check that the function returned an array?

dev/test-studio/structure/constants.ts
```diff
@@ -1,4 +1,4 @@
-import {simpleArrayOfObjects} from '../schema/debug/simpleArrayOfObjects'
+// import {simpleArrayOfObjects} from '../schema/debug/simpleArrayOfObjects'
```
bjoerge, line 1: was commenting out this intentional?

packages/sanity/src/config/types.ts
```diff
@@ -157,9 +158,31 @@ export interface DocumentPluginOptions {
...
   actions?: DocumentActionComponent[] | DocumentActionsResolver
   productionUrl?: AsyncComposableOption<string | undefined, ResolveProductionUrlContext>
+  unstable_languageFilter?: _DocumentLanguageFilterResolver
   newDocumentOptions?: NewDocumentOptionsResolver
 }
 
+/**
+ * @internal
+ */
+export interface _DocumentLanguageFilterContext extends ConfigContext {
```
bjoerge, line 168: I take the `_`-prefix to mean internal or unstable here - given that we consider the language filter feature to be unstable - would it perhaps add clarity to use the _unstable_ prefix on the accompanying types as well?

bjoerge (approved):
Looks great! Just have a minor comment about using the `_` prefix for internal/unstable interfaces

packages/sanity/src/config/types.ts
mariuslundgard: It uses the `_` prefix since the member is `@internal`. This follows the convention / lint rule defined by API Extractor (Microsoft). Personally think we should follow this convention, but it’s not something we’ve discussed yet.

mariuslundgard: Reference: https://api-extractor.com/pages/messages/ae-internal-missing-underscore/

PR #3514, opened 2022-08-15
fix: string array issues

hermanwikner:
### Description

This PR fixes:
- Render tags input when an array has `options: {layout: 'tags'}`
- Update form field UI for tags array (no border or padding to the left of the field)
- Make it possible to have strings in addition to an object with `title` and `value` in `options.list`. E.g:
```
  options: {
    list: [{title: 'Title', value: 'value'}, 'Hello'],
  }
```
### What to review
Test that the above mentioned fixes works

### Notes for release

fix: string array issues

### DIFF
--- a/packages/sanity/src/form/studio/inputResolver/helpers.ts
+++ b/packages/sanity/src/form/studio/inputResolver/helpers.ts
@@ -8,6 +8,7 @@ export function getOption(type: SchemaType, optionName: string) {
 
 const PSEUDO_OBJECTS = ['array', 'file', 'image', 'reference']
 const HIDDEN_FIELDS = ['asset', 'crop', 'hotspot', '_ref', '_weak']
+const NO_LEVEL_LAYOUTS = ['tags']
 
 export function getObjectFieldLevel(field: ObjectFieldProps): number {
   const {type, fields, options} = field.schemaType
@@ -29,8 +30,9 @@ export function getArrayFieldLevel(field: ArrayFieldProps): number {
   const {options} = field.schemaType
 
   const hasListOptions = (options?.list || [])?.length > 0
+  const isNoLevelLayout = NO_LEVEL_LAYOUTS.includes(options?.layout || '')
 
-  if (hasListOptions) {
+  if (hasListOptions && !isNoLevelLayout) {
     return field.level
   }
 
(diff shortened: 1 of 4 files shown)


packages/sanity/src/form/studio/inputResolver/helpers.ts
```diff
@@ -29,8 +30,9 @@ export function getArrayFieldLevel(field: ArrayFieldProps): number {
   const {options} = field.schemaType
 
   const hasListOptions = (options?.list || [])?.length > 0
+  const isCustomArrayLayout = CUSTOM_ARRAY_LAYOUTS.includes(options?.layout || '')
```
bjoerge, line 33: I'm wondering if we could find a better name than "custom" here - usually "custom" means "user defined". Given that we only handle a single known layout here, maybe we could just write it explicitly?  

```suggestion
const isTagsArray = options?.layout === 'tags'
```

packages/sanity/src/form/studio/inputResolver/resolveArrayInput.ts
```diff
@@ -30,15 +30,15 @@ export function hasOptionsList(type: ArraySchemaType): boolean {
 }
 
 export function resolveArrayInput(type: ArraySchemaType): ComponentType<any> {
+  if (isStringArray(type) && isTagsArray(type)) {
```
bjoerge, line 33: just out of curiosity: was moving this to the top here required?

bjoerge (approved):
Nice work! Approved, but feel free to look at my comments :)

rexxars (approved):

packages/sanity/src/form/studio/inputResolver/helpers.ts
hermanwikner: Good point, I'll rename it 👍

packages/sanity/src/form/studio/inputResolver/resolveArrayInput.ts
hermanwikner: I moved it because the tag layout didn't render if you also had `list` in `options`. Not sure what is the right thing to do here. I'm going to move it back to how it was before for now 👍

PR #3522, opened 2022-08-16
fix(form-builder): make sure to forward ref for object input with tabs

sjelfull:
### Description

This change fixes the console warning _The input component for type "fieldGroups" has no associated ref element..._ for document types with field groups.

The bug is that that all inputs need to set the ref to an actual element, but in the case of field groups we didn't forward the ref anywhere. The ref is now set on the tab list.

### What to review

- Go to test studio and document type with field groups
- See that console warning _The input component for type "fieldGroups" has no associated ref element...._ is not triggered any more

### Notes for release

- Fixed a bug where we didn't forward the ref for object input with tabs, causing console warnings


ninaandal (approved):
👍

PR #3523, opened 2022-08-16
v3 PTE value handling improvements

skogsmaskin:
This will fix an issue with editing PT-members from the outside and have the current state immediately reflected in the editor rendering, and only render those blocks which actually changed.

These changes also improves performance and stability of the value handling (like removing a separate value copy in state, less consistency checking etc).


bjoerge (approved):
LGTM! Props for improved test coverage too 👏🏼

PR #3525, opened 2022-08-17
feat(default-layout): search improvements

robinpyon:
## Description

This PR adds some quality of life improvements to our current search experience.

### **Features**

- Basic filtering: filter search results by any number of document types. You can also use search as a basic browser to browse document types without entering a search query.
- Improved navigation: search is now toggleable with a hotkey (Command/Control-K) and fully navigable by keyboard. Inputs are treated as comboboxes which you can navigate with the up and down arrow keys.
- Improved performance: search results are now rendered in a virtual list and will no longer lock up the browser when dealing with large result sets. It *should* feel less janky when scrolling. This virtual list also has wraparound keyboard support.
- View recent searches: these are saved in local storage for future reference and can be individually or collectively deleted. Search terms are saved only when you interact with a search result (e.g. doing a search for “harry” and closing the panel will not count)
- Search will retain previous terms and the position in the virtual list too, which should make it more convenient to resume your search. Closing search will also return focus to the last focused element. Re-opening search with a list of existing results will snap you back to the last top-most visible index.
- Basic screen reader support: screen readers should correctly broadcast combo / listbox roles, even within virtual lists. Example video can be found [here](https://sanity-io.slack.com/archives/C03809QMR9S/p1660601027758749?thread_ts=1660600907.515999&cid=C03809QMR9S) (internal slack)

### **Breaking changes**

**Search items will no longer respect [custom preview components](https://www.sanity.io/docs/previews-list-views#d82fd035385f).** 

The main rationale for this change is that search results should strive to be consistent as possible across multiple document types, especially as we start to open up more filtering options in future. It also helps with virtual lists (as we can keep all item heights consistent).

### **Notable changes in this package**

1. Search results now use `SanityDefaultPreview` instead of `Preview` (from `part:@sanity/base/preview`)
    - This was done to render previews that don’t resolve custom input components (similar to desk pane items)
    - There is a bit of duplication of logic with pane items helpers ([search result item helpers](https://github.com/sanity-io/sanity/blob/findability/packages/%40sanity/default-layout/src/navbar/search/components/searchResultItem/helpers.tsx) vs [desk-tool pane item helpers](https://github.com/sanity-io/sanity/blob/findability/packages/%40sanity/default-layout/src/navbar/search/components/searchResultItem/helpers.tsx)
        - Q. Can / should these helpers be extracted into the `base` or `utils` package? (or similar)
2. Two key components are exported – `<SearchPopover>` (desktop / larger breakpoints) and `<SearchDialog>` (mobile / smaller breakpoints). 
3. A `<SearchProvider>` encapsulates the above two components and is used to retain search state once search is closed. It also uses a reducer to better corral various search-related actions and shared UI state.
4. A separate `<CommandListProvider>` has also been created to try and generalise the various command lists used in search (the main search input, as well the filters panel)
    - It’s a fairly meaty component which is responsible for a fair bit, including but not limited to: keyboard navigation, focus redirection, assigning aria attributes and pointer blocking (more on this below).
4. To ensure recent searches are correctly scoped to different workspaces in Local Storage, we use the following key convention of:  
`search-terms::recent__projectId:datasetName:currentUserId`
6. Navbar has now been [simplified a little](https://github.com/sanity-io/sanity/compare/next...findability#diff-f3837cbffa080c638b641b4814bc691f64f7273a54c52391b1dc9f3c8809ce4b) - some search-related logic and components have either been removed or are now encapsulated within the above components.

**New dependencies**

- `react-focus-lock` (in use in `form-builder`)
    - Used to lock focus when search is open, as well as returning focus when search is closed
- `react-virtual` (new)
    - Used to drive virtual lists in search results. Provides an API that allows you to navigate to either an index or offset in the virtual list, which we use to resume scroll position when search is re-opened and also provide wraparound navigation.
- `is-hotkey` (in use by multiple packages)
- `debug` (in use by multiple packages)

#### `@sanity/base`

[Changes to the search function in `createWeightedSearch`](https://github.com/sanity-io/sanity/blob/5c866e48261f481d7c295e170b103ce1f29f2313/packages/%40sanity/base/src/search/weighted/createWeightedSearch.ts#L18-L22). 

- It now supports both a single search string (previous behaviour) in addition to a `SearchTerms` object with support for specifying document types. [More context from Snorre](https://github.com/sanity-io/sanity/pull/3525#discussion_r947647180)
- It also contains an optional third argument – `searchComments` – which will accept an array string and prepend these as comments to the search query.
    - This was done to allow sending (multiple) custom comments for future measurement. An example of this is  `// findability-mvi:1\n// findability-selected-types:5\n*[_type in $__types \u0026\u0026...`
    - Possibly smelly, but seems like a low cost way to enable measurement. Open to alternatives here!

### Worth mentioning

1. [We hard-code the height of search result items](https://github.com/sanity-io/sanity/blob/findability/packages/%40sanity/default-layout/src/navbar/search/constants.ts#L20)
    - This helps drive a more performant virtual list. It’s possible to dynamically measure these, but it incurs a noticeable performance hit.
    - We use also this height to accurately determine (and store) the current virtual list scroll position when closing search.
2. We kill existing `:hover` states in command lists by conditionally displaying an overlay above the child container. 
    - Video of this in action below, with overlay areas denoted in red (note the position of the mouse cursor before arrow keys are pressed)
    - This is done to allow us to only have one selected item regardless of keyboard / mouse interaction. You can see similar on the front page of Google as well as other command palettes (e.g. Alfred or Dash on OSX). 
    - Admittedly this is pretty hacky, and arguably the proper way to do it would be to conditionally add/remove hover CSS classes. However I believe this may a little difficult to do with both our current `SanityDefaultPreview` and Sanity UI components as it’s not possible to declaratively remove hover states (afaik).  

https://user-images.githubusercontent.com/209129/185018974-850af601-ef73-40ef-86be-e871f8db9b9a.mp4

### Known issues

- ~~On mobile: the `isLoading` skeleton doesn’t seem to be showing (though each document type is rendering fine)~~
- ~~On mobile (iOS Safari): height issues when dealing with collapsing URL toolbar~~
- Invoking search via the shortcut (Command/Control-K) in a markdown field will invoke search as well as create a markdown link `[]()`
- The button to remove individual recent searches isn’t accessible via keyboard or on touch devices
    - Currently happy to live with this, as the button to remove all recent searches is accessible

### Things to test

- **Keyboard navigation:** toggling with CMD/Control+K, tabbing between groups, navigating command lists with arrow keys.
- **Recent searches:** removing individual and all items, try and break the recent search list with extra long queries and/or multiple document types. Try switch between workspaces – recent searches should be specific to the current user, dataset and project.
- **Mobile / smaller breakpoints:** general flow
- **General performance:** searching in the test studio for generic terms like “a”, scrolling through hundreds of items.
- **Resuming search:** conduct a search which yields a decent number of items, scroll down the list and then close / re-open. Does it behave as you’d expect?
- **Existing search functionality:** reference input fields (which also use `createdWeightedSearch`) should work as normal. 

### Updated search

findability-mvi-1

- Added support for filtering search results by any number of document types. You can also browse documents from any number of types without entering a search query.
- Search is now toggleable with a hotkey (Command/Control-K) and fully navigable by keyboard.
- Recent searches (including selected document types) are stored for later use
- Search terms and results are retained upon close: handy if you want to quickly revisit or amend your last search
- Improved performance when rendering and scrolling through large result sets
- BREAKING CHANGE: Documents listed in search results will no longer resolve [custom components for block content and array views](https://www.sanity.io/docs/previews-list-views#d82fd035385f)


packages/@sanity/base/src/search/weighted/createSearchQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export const DEFAULT_LIMIT = 1000
+
+const combinePaths = flow([flatten, union, compact])
+
+const pathWithMapper = ({mapWith, path}: SearchPath): string =>
+  mapWith ? `${mapWith}(${path})` : path
+
+export function createSearchQuery(
```
snorrees, line 35: FYI: Most of this code is hoisted as is from createWeightedSearch.ts to make it unit-testable, and for combining per-search-config with globally provided config. Unlike before, we are now always recreating searchSpec, unlike before when it was done once when creating the weighted search function. This is so we can change which types are included in the query.  

Another new bit is the offset param, which would allow for running paginated searches. It is unused atm.

packages/@sanity/base/src/search/weighted/types.ts
```diff
@@ -1,3 +1,15 @@
+import {ObjectSchemaType} from '@sanity/types'
+
+/**
+ * @internal
+ */
+export interface SearchTerms {
```
snorrees, line 6: Making this an object should allow us to extend with new terms in the future (specific fields, dates ect).

packages/@sanity/default-layout/src/navbar/Navbar.tsx
```diff
@@ -276,37 +230,16 @@ export const Navbar = memo(function Navbar(props: NavbarProps) {
...
-                      mode="bleed"
-                      onClick={handleToggleSearchOpen}
-                      ref={setSearchCloseButtonElement}
-                    />
-                  )}
-                </Flex>
-              </SearchCard>
-            )}
+            <SearchProvider>
+              {shouldRender.searchFullscreen && (
```
snorrees, line 234: I really like that these are two separate components now.

packages/@sanity/default-layout/src/navbar/search/SearchDialog.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  const dialogRef = useRef<HTMLDivElement>(null)
+
+  // Always hide filters on mount
+  useEffect(() => {
+    dispatch({type: 'FILTERS_HIDE'})
+  }, [dispatch])
+
+  // Force dialogs to be 100% height
+  // TODO: there has to be a better way to do this
```
snorrees, line 134: If we dont find a better way, do we need the todo?

snorrees: Also, the dep filtersVisible is not used. I guess its driving this effect to trigger, but then we perhaps only need to set height when filtersVisible is true?

packages/@sanity/default-layout/src/navbar/search/SearchField.tsx
```diff
@@ -1,196 +1,67 @@
...
+   */
+  const handleWindowResize = useCallback(() => {
+    setPopoverPosition(calcDialogPosition(placeholderElement))
+  }, [placeholderElement])
+
+  useEffect(() => {
+    setPopoverPosition(calcDialogPosition(placeholderElement))
+  }, [placeholderElement])
+
+  useEffect(() => {
```
snorrees, line 28: Suggestion: create a usePopoverPosition hook, move the sideefects and state into it and return popoverPosition; this will make the callsite here a bit more tidy?

packages/@sanity/default-layout/src/navbar/search/components/NoResults.tsx
```diff
@@ -0,0 +1,36 @@
...
+            <TypePills types={state.terms.types} />
+          </Inline>
+        </Flex>
+      )}
+    </Flex>
+  )
+}
+
+// TODO: use idiomatic sanity/ui styling - is there a better way to apply _inline_ text styles?
+const Semibold = styled.span(({theme}: {theme: Theme}) => {
```
snorrees, line 32: I _think_ could just work? 
```ts
const Semibold = styled.span`
  font-weight: ${(theme: Theme) => theme.sanity.fonts.text.weights.semibold};
`
```

packages/@sanity/default-layout/src/navbar/search/components/RecentSearches.tsx
```diff
@@ -0,0 +1,115 @@
...
+}: RecentSearchesProps) {
+  const {
+    dispatch,
+    state: {recentSearches},
+  } = useSearchState()
+
+  const mediaIndex = useMediaIndex()
+
+  const maxVisibleQueryChars = useMemo(() => {
+    return mediaIndex < 2 ? 20 : 40
```
snorrees, line 29: Use constants for 20 and 40 to increase readability and dry it a bit?

packages/@sanity/default-layout/src/navbar/search/components/SearchHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const SearchHeaderWrapper = styled(Box)`
+  border-bottom: 1px solid ${hues.gray[200].hex};
+  flex-shrink: 0;
+`
+
+const NotificationBadge = styled.div`
+  align-items: center;
+  background: ${hues.blue[700].hex};
+  color: ${white.hex};
```
snorrees, line 124: Are we sure we dont want to use colors from theme for this instead? Same goes for hues? I think using sanity color directly is discouraged. Will have to be rewritten for dark theme in v3 I think.

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -0,0 +1,157 @@
...
+   * Send react-virtual's `scrollToIndex` function to shared CommandList context
+   */
+  useEffect(() => {
+    setVirtualListScrollToIndex(scrollToIndex)
+    return () => setVirtualListScrollToIndex(null)
+  }, [setVirtualListScrollToIndex, scrollToIndex])
+
+  /*
+  // Load next page and focus previous sibling
+  const handleLoadMore = useCallback(
```
snorrees, line 59: Can probably delete this since we dont use it in this iteration?

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Add current search terms to recent searches, trigger child item click and close search
+   */
+  const handleResultClick = useCallback(() => {
+    dispatch({
+      terms,
+      type: 'RECENT_SEARCHES_ADD',
+    })
+    onChildClick?.()
```
snorrees, line 82: Conditional function invocation syntax? TIL!

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  text="More"
+                  title="Load more search results"
+                />
+              )*/}
+              </VirtualListChildren>
+            </VirtualList>
+          )}
+
+          {!result.hits.length && result.loaded && (
+            // (No results)
```
snorrees, line 128: Unnecessary comment imo (name of component is the same)

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -0,0 +1,157 @@
...
+                      onClick={handleResultClick}
+                      onMouseDown={onChildMouseDown}
+                      onMouseEnter={onChildMouseEnter(virtualRow.index)}
+                      key={hit.hit._id}
+                      virtualRow={virtualRow}
+                    />
+                  )
+                })}
+
+                {/*result.hasMore && (
```
snorrees, line 114: Delete imo

packages/@sanity/default-layout/src/navbar/search/components/TypePills.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {Box, Card, Flex, Text} from '@sanity/ui'
+import React, {useMemo} from 'react'
+import styled from 'styled-components'
+
+interface TypePillsProps {
+  availableCharacters?: number
+  types: SchemaType[]
+}
+
+export function TypePills({availableCharacters = 40, types}: TypePillsProps) {
```
snorrees, line 13: Make the default a const?

packages/@sanity/default-layout/src/navbar/search/components/TypePills.tsx
```diff
@@ -0,0 +1,73 @@
...
+    </Flex>
+  )
+}
+
+function typeTitle(schemaType: SchemaType) {
+  return schemaType.title ?? schemaType.name
+}
+
+const Pill = styled(Card)`
+  background: ${hues.blue[50].hex};
```
snorrees, line 70: Consider using a color from theme to make it portable to v3?

snorrees (commented):
Massive effort!

- **Keyboard navigation**: feels really good!
- **Recent searches**: checked on small and big breakpoints, all good.
- **Mobile / smaller breakpoints**: thunmbs up from me
- **General performance**: noticeable improvment on, especially when navigating with keyboard in the result list. Not laggy at all.
- **Resuming search**: it retains scroll position. This makes sense. For whatever reason I expected the result I interacted with to be on top or at least highlighted, but I'm not sure thats the right expectation. 

I think this is turning out to be a fantastic first iteration! I mostly have nitpicky comments that can be ignored as you see fit.

Do consider replacing use of sanity/color with a value from theme though, or get the theme key somewhere so we can easily find darkmode colors when we do the v3 port.

robinpyon:
> **Resuming search**: it retains scroll position. For whatever reason I expected the result I interacted with to be on top or at least highlighted, but I'm not sure thats the right expectation.

This is a reasonable expectation! It should both retain the last scroll position AND the last selected item, and if the last selected item isn't visible (because you made the viewport smaller) it should then select the top-most item by default.

packages/@sanity/default-layout/src/navbar/search/components/SearchHeader.tsx
robinpyon: You're definitely right here! Dropped the `@sanity/colors` dependency throughout and now pull from the theme.

packages/@sanity/default-layout/src/navbar/search/SearchDialog.tsx
robinpyon: Good point! Cleaned this to be slightly less offensive, also removed useEffect dependency as well as the redundant dispatch

packages/@sanity/default-layout/src/navbar/search/SearchField.tsx
robinpyon: Good shout, that's in!

packages/@sanity/default-layout/src/navbar/search/components/TypePills.tsx
robinpyon: That's in – the theme path I'm pulling from seems a little arbitrary though (` theme.sanity.color.selectable.primary.enabled.code.bg`). Seems like a smell, though I figure we'll give this another proper pass come V3

packages/@sanity/base/src/search/weighted/createWeightedSearch.ts
```diff
@@ -2,85 +2,60 @@
...
-  const params: Record<string, string> = {}
-  return terms.reduce((acc, term, i) => {
-    acc[`t${i}`] = `${term}*` // "t" is short for term
-    return acc
-  }, params)
+type ObjectSchema = {
+  name: string
+  // eslint-disable-next-line camelcase
+  __experimental_search: ObjectSchemaType['__experimental_search']
 }
```
bjoerge, line 16: This is duplicated both here and in `createWeightedSearch.ts` - and the naming is quite similar to `ObjectSchemaType` from `@sanity/types`. I'd suggest renaming it to `SearchableType` (or similar), and share it between the functions that accepts it).

packages/@sanity/base/src/search/weighted/createSearchQuery.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export interface SearchQuery {
+  query: string
+  params: SearchParams
+  options: Record<string, unknown>
+  searchSpec: SearchSpec[]
+  terms: string[]
+}
+
+type ObjectSchema = {
```
bjoerge, line 22: See earlier comment

bjoerge (approved):
Left a couple of comments/suggestions that you can consider, no blockers for merge. Really thorough work, well done! 👏

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -0,0 +1,130 @@
...
+    <SearchResultsWrapper aria-busy={result.loading} $loading={result.loading}>
+      <PointerOverlay ref={setPointerOverlayRef} />
+
+      {result.error ? (
+        <SearchError />
+      ) : (
+        <>
+          {!!result.hits.length && (
+            // (Has search results)
+            <VirtualList ref={childParentRef}>
```
hermanwikner, line 82: I think `VirtualList as="ul"` and `SearchResultItemWrapper as="li"` would be more semantic markup here :)

hermanwikner (approved):
Great work – looks and works really good! 💯 

I left a few comments, mostly nitpicking and some suggestions :) I wouldn't say that there are any blockers for merge, expect bumping the `radius` to `2` on the `PlaceholderSearchInput`!

packages/@sanity/default-layout/src/navbar/search/components/SearchHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+            smallClearButton
+            spellCheck={false}
+            value={terms.query}
+          />
+        </Box>
+
+        {/* Filter toggle */}
+        <Card borderLeft={!!onClose} padding={onClose ? 2 : 1}>
+          <FilterButton>
+            <Button
```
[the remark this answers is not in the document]
snorrees: I think this already has `aria-label="Filter"`. Renaming FilterButton to FilterBox to follow conventions though.

packages/@sanity/default-layout/src/navbar/search/components/SearchHeader.tsx
```diff
@@ -0,0 +1,136 @@
...
+    width: 20px;
+    vertical-align: bottom !important;
+  }
+`
+
+const FilterButton = styled(Box)`
+  position: relative;
+`
+
+const SearchHeaderWrapper = styled(Box)`
```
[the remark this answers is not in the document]
snorrees: Is this also the case for v3? We have to port this to v3 very soon, making it v3 compatible would be nice.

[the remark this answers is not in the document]
snorrees: I'll change this to Card with borderBottom, but in general I get the implication that we should prefer css-vars throughout for v2?

snorrees:
@bjoerge I pushed the fixes to search that I talk a bit with out the other day.
We are including it in this PR as it is very close to the code being changed. 

Could you do a double take on this commit: https://github.com/sanity-io/sanity/pull/3525/commits/7a2f05299cf0c8d6f0e8ac525d085d87cd07cdff 

Added a bit of context to the commit message.

rexxars (approved):
Looks good! Left a couple of small comments. Nothing here is a blocker.

packages/@sanity/base/src/search/weighted/createWeightedSearch.ts
```diff
@@ -1,86 +1,61 @@
...
-      )
-      .pipe(
-        options.unique ? map(removeDupes) : tap(),
-        map((hits: SearchHit[]) => applyWeights(searchSpec, hits, terms)),
-        map((hits) => sortBy(hits, (hit) => -hit.score))
-      )
+  // supports string as search param to be backwards compatible
+  return function search(searchParams, searchOpts = {}, searchComments = []) {
+    const searchTerms = getSearchTerms(searchParams, types)
+    const {query, params, options: o, searchSpec, terms} = createSearchQuery(searchTerms, {
```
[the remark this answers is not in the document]
snorrees: You are right, no need to alias here; I think it is a remnant after a refactor pass, fixing.

packages/@sanity/default-layout/src/navbar/search/SearchDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+    dispatch({type: 'FILTERS_HIDE'})
+  }, [dispatch])
+
+  return (
+    <FocusLock autoFocus={false}>
+      <StyledDialog
+        cardRadius={1}
+        header="Filter"
+        height="fill"
+        id="search-filter"
```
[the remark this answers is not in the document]
snorrees: sanity/ui dialog requires an id; I'll change it to use `import {useId} from '@reach/auto-id'` which is common in the studio code afaik.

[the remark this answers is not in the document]
snorrees: Lets keep the "id" around as data-testid.

bjoerge:
Reviewed 7a2f05299cf0c8d6f0e8ac525d085d87cd07cdff individually, @snorrees. Looks great! Thank you for the diligent explanatory comments and the added test coverage ✨

PR #3526, opened 2022-08-17
feat(form): introduce spreadable elementProps to reduce custom input boilerplate

bjoerge:
### Description

This passes a spreadable `elementProps` property to all input components. This currently includes things like `onFocus`/`onBlur` event handlers, `ref`, `value` (for primitive inputs) and an `onChange` that handles "native" events and emits the correct patches, and the idea is that you can safely pass it on to any "native" input element. This reduces the amount of boilerplate needed for custom inputs (especially those dealing with primitive values). For inputs dealing with complex values (objects and arrays), `value` is not included in `elementProps`, since it doesn't make sense to pass it to any native input.

This change also includes moving change indicator rendering from individual primitive input components to their corresponding *field* component instead. Another win for reduced boilerplate.

This change means that input components will receive two "values" - `props.value` will now be the of the "parsed" type defined in the schema (e.g. number), while `elementProps.value` will always be a string since its intended to being forwarded as an attribute to the native HTML input element.

I also snuck in a small refactoring of some of the member components improving the file structure.

### What to review
This is includes a lot of edits, but was a fairly quick refactor to do, so I'm fine with proceeding with not merging this, although I think it will be a significant improvement to the DX of writing custom inputs.

Apart from the regular code review I'd like some feedback on naming - ~whether `domProps` makes sense or if there's better alternatives (see also internal discussion).~. Update: settled on `elementProps`

### Notes for release

- Now passing a spreadable `elementProps` to input components that reduces required boilerplate needed for custom inputs


rexxars (approved):
This look amazing. I think `domProps` might not be the right naming, but I can't really find anything perfect here. `elementProps` maybe, or `inputProps` (but that is slightly overloaded now I guess?)

I also think we need to figure out a way to configure the change indicator handling (eg opt-out), but we can probably do that separately

bjoerge:
Merged in 2c769682f. Thanks for reviewing!

PR #3534, opened 2022-08-18
feat: add array schema validation when using tags layout with list

hermanwikner:
### Description
This PR adds:
- Schema validation when using `layout: 'tags'` together with a `list` with predefined values
- Renders the `TagsArrayInput` before `OptionsArray`. I think this is more expected behaviour.

### What to review
- Configure an array with tags layout + a list in `options` and the warning should appear in the list of configuration issues 
- Does the copy make sense?

### Notes for release
feat: add array schema validation when using tags layout with list


bjoerge (approved):
Super nice!

PR #3542, opened 2022-08-19
fix(server): work around npm peer dependency issue

rexxars:
### Description

Reported in Slack: https://sanity-io-land.slack.com/archives/C9Z7RC3V1/p1660809409071889

It seems that npm is having issues installing `@sanity/core` - it seems to be stuck in an infinite loop. After much digging, I believe this to be caused by a bug in npm, which I have [filed an upstream issue for](https://github.com/npm/cli/issues/5322).

However, as we cannot know if this will be fixed anytime soon, I figured we should try to work around it. It _appears_ as if the `@hot-loader/react-dom` peer dependency is what is causing issues. I have therefore republished it as `@sanity/hot-loader-react-dom`, removing its peer dependencies and rather implicitly depending on them. Since the parent (`@sanity/server`) already has that peer dependency defined, I think this should be fine.

### What to review

That the dev server starts as expected and accepts changes in the same way as it previously did. It is not a perfect hot-reload, but it never has been from what I can recall.

### Notes for release

- Fixed an issue potentially causing npm to go into an infinite loop when installing `@sanity/core`


bjoerge (approved):
Impressive detective work! 🕵🏼

skogsmaskin (approved):

PR #3545, opened 2022-08-20
refactor(datastores): optimize cors check, clear token on 401

rexxars:
### Description

Was observing some requests on studio load and came across a slight optimization opportunity; we're currently doing a CORS-check with both a ping request and a user request, but if the CORS-configuration is correct, we also re-fetch the user right after.

This PR only calls the `/ping` route in the event of a user fetch failure - in that case we already know the user request failed, so we don't need to re-run it. If the ping route succeeds, assume it is a CORS error and throw a more specific CORS error (as before), otherwise, re-throw the network error.

Didn't look like we were using the error properties aside from `projectId`, so I took the opportunity to remove them.

Oh, and I also saw the opportunity to address one of the the things we talked about lately: if the user request fails with a 401, clear the token from localStorage in order to not reuse an expired/invalid token further.

### What to review

- That the CORS-check still work as intended
- That clearing an invalid token works (`localStorage.__studio_auth_token_ppsg7ml5 = '{"token":"blatti","time":"2022-08-20T00:04:00.428Z"}'` in the devtools should be enough)

### Notes for release

- Fixed an issue where an invalid/expired session token would not be cleared


packages/sanity/src/datastores/authStore/createAuthStore.ts
```diff
@@ -84,12 +84,6 @@ const saveToken = ({token, projectId}: {token: string; projectId: string}): void
 }
 
 const getCurrentUser = async (client: SanityClient) => {
```
hermanwikner, line 86: Nice optimization! 👍 

But the CORS error screen is not displayed after these changes. I investigated it a bit and I think these changes will make it work. Let me know what you think!

```suggestion
const getCurrentUser = async (client: SanityClient) => {
  try {
    const user = await client
      .request({
        uri: '/users/me',
        withCredentials: true,
        tag: 'users.get-current',
      })
      .then((res) => res)
      .catch((err) => {
        // 401 means the user had some kind of credentials, but failed to authenticate,
        // we should clear any local token in this case and treat it as if the used was
        // logged out
        if (err.statusCode === 401) {
          clearToken(client.config().projectId || '')
        }
        return null
      })

    // Request failed for some other reason, see if this was a CORS-error by
    // checking the `/ping` endpoint, which allows all origins
    if (!user) {
      const validCorsConfig = await client
        .request({uri: '/ping', withCredentials: false, tag: 'cors-check'})
        .then(() => true)

      // Throw a specific error on CORS-errors, to allow us to show a customized dialog
      if (validCorsConfig) {
        throw new CorsOriginError({projectId: client.config()?.projectId})
      }
    }

    // If the user came back with an id, assume it's a full CurrentUser
    return typeof user?.id === 'string' ? user : null
  } catch (err) {
    // Some non-CORS error, re-throw
    throw err
  }
}
```

hermanwikner (changes requested):
Nice! But the CORS error screen is not displayed after these changes. I left a comment with a suggested change 👍

rexxars:
> Nice! But the CORS error screen is not displayed after these changes

Oops. Some inverted logic in there 😓  Thanks for catching.

Have updated with the correct logic, and also added special handling of network errors with no message, which seems to be the default.

hermanwikner:
The CORS-check still work as intended ✅ 👍 

When clearing an invalid token, I see the error screen in the image below. Do we want to show the error screen, or do we want to show the log in screen?

Screenshot 2022-09-08 at 13 59 48

rexxars:
> When clearing an invalid token, I see the error screen in the image below. Do we want to show the error screen, or do we want to show the log in screen?

Good catch! Have updated the PR to broadcast a change event when we clear the token.

hermanwikner (approved):
Nice! 👍

PR #3546, opened 2022-08-20
fix(config): include logo in resolved studio config

rexxars:
### Description

The `logo` config property was not getting passed on when resolving the configuration. This PR fixes this, and pre-renders (well... computes) it at the config level, so it is ever so slightly faster to render.

### What to review

That the `logo` property can be configured and is respected by the studio. The test-studio should now show "Test Studio™" instead of just "Test Studio", but in theory you should be able to put something else here. An SVG or an image etc.

### Notes for release

- Fixed an issue where a configured `logo` component for a workspace would not be respected


hermanwikner (approved):
Nice! 🙏

PR #3547, opened 2022-08-21
Typesafety experiments

snorrees:
PoC on how defineType could be changed to:

* test the API using ts-expect-error in "unit tests"
* ensures that the type passed to defineType is returned "correctly" and typenarrowed. Ie, type is the actual literal type and not string, for instance.
* ensure that type-safety is maintained for alias types
  *  needed to make a for-typescirpt-only concession for this, by allowing devs to pass an `alias` prop to narrow an alias-type to an intrinsic type. If this prop is omitted, safety falls back to "basic type"
* rewrote Intrinsic types from `type` to `interface` (directory.) This allows devs to extend it using declaration merging
   * added a testcase to illustrate how it can be done  

*Heads up*: this is not a full implementation, but hopefully could spar some discussion.

In the current API I dont think alias-types are correctly type narrowed, as `Exclude<string, Schema.Type>` does not work as expected in TypeScript:

```ts
//we expect this to not allow strings from Schema.Type
type TypeAlias = Exclude<string, Schema.Type>

// but this is still allowed
const typeAlias: TypeAlias = 'boolean'

// TypeAlias is regarded as string 
// even though we excluded 'boolean' via Schema.Type
```


snorrees:
I would be happy to make a full test-suit for the current implementation if that sounds interesting; then make some proposals that will make them all pass.

array.of and reference.to also need special consideration.

PR #3550, opened 2022-08-22
fix(form): update resolving of preview component

hermanwikner:
### Description

 This PR fixes so that `SanityPreview` is always returned from `defaultResolvePreviewComponent` in `inputResolver`. This change is made because `SanityPreview` [takes care of resolving what preview component to render](https://github.com/sanity-io/sanity/blob/v3/packages/sanity/src/preview/components/SanityPreview.tsx#L25), but also performs some logic that is important for a preview component, logic that was missed before when we conditionally returned `components.preview`.

### What to review
n/a

### Notes for release

fix(form): update resolving of preview component

### DIFF
--- a/packages/sanity/src/form/studio/defaults.ts
+++ b/packages/sanity/src/form/studio/defaults.ts
@@ -25,5 +25,5 @@ export const defaultRenderItem: RenderItemCallback = (props) => {
 }
 
 export const defaultRenderPreview: RenderPreviewCallback = (props) => {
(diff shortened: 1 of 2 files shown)


skogsmaskin:
I think this may have broken the rendering of custom preview's inside the PT-input:

Before (`v3` branch):
image

After:
image

bjoerge (approved):

skogsmaskin:
We found the above mentioned issue to be this: https://github.com/sanity-io/sanity/pull/3557

So go ahead and merge this 👍

PR #3554, opened 2022-08-23
feat: implement renderDiff

hermanwikner:
### Description

This PR adds

- form.renderDiff

```ts
const config = createConfig({
  ...,
  form: {
    renderDiff: (props, next) => {
      return next(props)
    }
  }
})
```

- Review changes test
  - With `components.diff`
  - With `form.renderDiff`
  - With default diff component
- Export of all diff components
- Revert changes tooltip UI fix

### What to review
- Review the new `form.renderDiff` and that it works as expected
- General code review

### Notes for release

feat: implement renderDiff


cypress/integration/field/reviewChanges.test.js
```diff
@@ -0,0 +1,128 @@
...
+  numberWithDefaultDiffComponent: undefined,
+  numberWithCustomDiffComponent: undefined,
+  numberWithRenderDiffComponent: undefined,
+  booleanWithDefaultDiffComponent: undefined,
+  booleanWithCustomDiffComponent: undefined,
+  booleanWithRenderDiffComponent: undefined,
+  slugWithDefaultDiffComponent: undefined,
+  slugWithCustomDiffComponent: undefined,
+  slugWithRenderDiffComponent: undefined,
+}
```
mariuslundgard, line 21: We could type this as `SanityDocument`

```suggestion
import {SanityDocument} from '@sanity/types'

const doc: SanityDocument = {
  _id: DOC_ID,
  _type: 'reviewChanges'
```

cypress/integration/field/reviewChanges.test.js
```diff
@@ -0,0 +1,128 @@
...
+    ).then(() => {
+      cy.get('[data-testid="string-field-diff"]').should('be.visible')
+      cy.get('[data-testid="boolean-field-diff"]').should('be.visible')
+      cy.get('[data-testid="slug-field-diff"]').should('be.visible')
+      cy.get('[data-testid="number-field-diff"]').should('be.visible')
+    })
+  })
+
+  // Configured in the reviewChanges schema
+  it('components.diff is visible', () => {
```
mariuslundgard, line 66: Hm, there is potential for race conditions between these tests. Since patching happens in multiple tests at the same time, there is potential for data to interfered between tests.

I think we should

hermanwikner: Are you sure? Because a new document is created for each test

package.json
```diff
@@ -88,6 +88,7 @@
     "@rollup/plugin-node-resolve": "^13.1.1",
     "@sanity/client": "^3.3.3",
     "@sanity/tsdoc-to-portable-text": "^0.4.1",
+    "@sanity/uuid": "^3.0.1",
```
mariuslundgard, line 91: Hm, I don't see `@sanity/uuid` being used in the added source code?

packages/sanity/src/field/types/datetime/diff/DatetimeFieldDiff.tsx
```diff
@@ -1,15 +1,27 @@
...
+  schemaType,
+}: DatetimeFieldDiffProps) => {
   return (
-    <DiffFromTo
-      align="center"
-      diff={diff}
-      schemaType={schemaType}
-      previewComponent={DatetimePreview}
-    />
+    <Box data-testid="datetime-field-diff">
```
mariuslundgard, line 18: Are we adding a `div` just for the sake of testing? If so, we should avoid that.

Could we do `<DiffFromTo data-testi="datetime-field-diff" ... />` instead?

packages/sanity/src/field/types/file/diff/FileFieldDiff.tsx
```diff
@@ -105,13 +114,13 @@ export const FileFieldDiff: DiffComponent<ObjectDiff<File>> = ({diff, schemaType
   )
 
   return (
-    <>
+    <Box data-testid="file-field-diff">
```
mariuslundgard, line 117: same as above

packages/sanity/src/field/types/image/diff/ImageFieldDiff.tsx
```diff
@@ -96,7 +104,7 @@ export const ImageFieldDiff: DiffComponent<ObjectDiff<Image>> = ({diff, schemaTy
   const imageDiff = <FromTo align="center" from={from} layout="grid" to={to} />
 
   return (
-    <>
+    <Box data-testid="image-field-diff">
```
mariuslundgard, line 107: same as above

packages/sanity/src/field/types/number/diff/NumberFieldDiff.tsx
```diff
@@ -1,15 +1,27 @@
...
+  schemaType,
+}: NumberFieldDiffProps) => {
   return (
-    <DiffFromTo
-      diff={diff}
-      schemaType={schemaType}
-      previewComponent={NumberPreview}
-      layout="inline"
-    />
+    <Box data-testid="number-field-diff">
```
mariuslundgard, line 18: same as above

packages/sanity/src/field/types/object/diff/ObjectFieldDiff.tsx
```diff
@@ -1,7 +1,21 @@
...
+  diff: ObjectDiff
+  schemaType: ObjectSchemaType
+}
+
+export const ObjectFieldDiff: DiffComponent<ObjectDiff> = ({
+  diff,
+  schemaType,
+}: ObjectFieldDiffProps) => {
+  return (
+    <Box data-testid="object-field-diff">
```
mariuslundgard, line 17: same as above

packages/sanity/src/field/types/reference/diff/ReferenceFieldDiff.tsx
```diff
@@ -1,17 +1,29 @@
...
   return (
-    <DiffFromTo
-      align="center"
-      diff={diff}
-      layout="grid"
-      path="_ref"
-      previewComponent={ReferencePreview}
-      schemaType={schemaType}
-    />
+    <Box data-testid="reference-field-diff">
```
mariuslundgard, line 18: same as above

packages/sanity/src/field/types/slug/diff/SlugFieldDiff.tsx
```diff
@@ -7,13 +9,23 @@ interface Slug {
...
+  schemaType,
+}: SlugFieldDiffProps) => {
   return (
-    <DiffFromTo
-      layout="inline"
-      diff={diff}
-      schemaType={schemaType}
-      previewComponent={SlugPreview}
-    />
+    <Box data-testid="slug-field-diff">
```
mariuslundgard, line 22: same as above

packages/sanity/src/field/types/url/diff/UrlFieldDiff.tsx
```diff
@@ -1,8 +1,19 @@
...
-export const UrlFieldDiff: DiffComponent<StringDiff> = ({diff, schemaType}) => {
-  return <DiffFromTo diff={diff} schemaType={schemaType} previewComponent={StringPreview} />
+export interface UrlFieldDiffProps {
+  diff: StringDiff
+  schemaType: StringSchemaType
+}
+
+export const UrlFieldDiff: DiffComponent<StringDiff> = ({diff, schemaType}: UrlFieldDiffProps) => {
+  return (
+    <Box data-testid="url-field-diff">
```
mariuslundgard, line 15: same as above

mariuslundgard (changes requested):
The potential for race conditions in the tests should be adressed before merging this (see comments in test files).

The feature seems to work really well, great work ✨

cypress/integration/field/reviewChanges.test.js
hermanwikner: I think my comment was posted before you updated your comment :) 

What I meant was: A unique document is already created in this test, and later deleted once it's completed. But, if I understand you correctly, you think we should create several documents for each case in this test?

rexxars:
Will defer reviewing this until we decide on the render callback vs component approach for studio and input components

cypress/integration/field/reviewChanges.test.js
mariuslundgard: Well, one document for each test should be sufficient 👍 

The reason is that jest runs tests in parallel – so we need to make sure the test code supports that.

PR #3557, opened 2022-08-24
String of options must return option value for the preview

skogsmaskin:
### Description

Consider the following type:

```
{
    type: 'string',
    name: 'tone',
    title: 'Tone',
    options: {
      list: [
        {value: 'default', title: 'Default'},
        {value: 'primary', title: 'Primary'},
        {value: 'positive', title: 'Positive'},
        {value: 'caution', title: 'Caution'},
        {value: 'critical', title: 'Critical'},
      ],
    },
  }
```
With the current code, the `options.list.title` is selected as value for the preview, which seems like a bug.

This will return the value instead, as it is in the actual data.


bjoerge (approved):
Yep, using title here seems like a bug.

PR #3560, opened 2022-08-24
fix(form-builder): check for invalid image/file values, provide reset option

sjelfull:
### Description

(Backport of v3 fix in [#3502](https://github.com/sanity-io/sanity/pull/3502))

> When an incorrect value is set for an image/file `asset` field, the inputs crash because it expects the shape to be correct. This PR upgrades the asset utils in order to use some looser type checks, and uses these to ensure the image asset is correct. If not, it will show a warning with a reset option, which unset the `asset` field. On images, it will also unset `crop` and `hotspot` since they are likely invalid for whatever image is selected next.

> I also made the action menu for file/images more forgiving of a missing `copyUrl`/`downloadUrl` - automatically hiding these options if the values are undefined.

In addition to the v3 fix I have updated the diff previews to handle invalid values.

Updated image input: 

image

Update image diff: 

image

Update file input: 

image

Updated file diff: 

image

### How to reproduce
- Change an image field to an object, and include an `asset` field which is a reference to something that is not an image
- Use the input to select a non-image reference
- Switch back to an image type, removing the `asset` field

### Notes for release

- Fixed an issue where incorrect image and file shapes (references to non-assets) would crash the studio and display invalid diffs

### DIFF
--- a/packages/@sanity/default-layout/package.json
+++ b/packages/@sanity/default-layout/package.json
@@ -32,7 +32,7 @@
   "dependencies": {
     "@portabletext/react": "^1.0.0",
     "@reach/auto-id": "^0.13.2",
-    "@sanity/asset-utils": "^1.2.0",
+    "@sanity/asset-utils": "^1.2.5",
     "@sanity/base": "2.31.0",
     "@sanity/client": "^3.3.3",
     "@sanity/generate-help-url": "^3.0.0",
--- a/packages/@sanity/form-builder/src/inputs/files/FileInput/InvalidFileWarning.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/FileInput/InvalidFileWarning.tsx
@@ -1,0 +1,36 @@
+import {ResetIcon, WarningOutlineIcon} from '@sanity/icons'
+import {Card, Flex, Box, Text, Stack, Button} from '@sanity/ui'
+import React from 'react'
+import styled from 'styled-components'
+
+type Props = {
+  onClearValue?: () => void
+}
+
+const ButtonWrapper = styled(Button)`
+  width: 100%;
+`
+
+export function InvalidFileWarning({onClearValue}: Props) {
+  return (
+    <Card tone="caution" padding={4} border radius={2}>
+      <Flex gap={4} marginBottom={4}>
+        <Box>
+          <Text size={1}>
+            <WarningOutlineIcon />
+          </Text>
+        </Box>
+        <Stack space={3}>
+          <Text size={1} weight="semibold">
+            Invalid file value
+          </Text>
+          <Text size={1}>
+            The value of this field is not a valid file. Resetting this field will let you choose a
+            new file.
+          </Text>
+        </Stack>
+      </Flex>
+      <ButtonWrapper icon={ResetIcon} text="Reset value" onClick={onClearValue} mode="ghost" />
+    </Card>
+  )
+}
--- a/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable import/no-unresolved */
-
 import {FormFieldSet, ImperativeToast} from '@sanity/base/components'
 import {Box, Button, Card, Dialog, Menu, MenuButton, MenuItem, Stack, ToastParams} from '@sanity/ui'
 import {get, groupBy, uniqueId} from 'lodash'
@@ -18,12 +16,8 @@ import {
 import React, {ReactNode} from 'react'
 import {FormFieldPresence, PresenceOverlay} from '@sanity/base/presence'
 import deepCompare from 'react-fast-compare'
-import {
-  ResolvedUploader,
-  Uploader,
-  UploaderResolver,
-  UploadOptions,
-} from '../../../sanity/uploads/types'
+import {isImageSource} from '@sanity/asset-utils'
+import {ResolvedUploader, Uploader, UploadOptions} from '../../../sanity/uploads/types'
 import {ImageToolInput} from '../ImageToolInput'
 import PatchEvent, {setIfMissing, unset} from '../../../PatchEvent'
 import UploadPlaceholder from '../common/UploadPlaceholder'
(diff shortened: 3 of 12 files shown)


packages/@sanity/default-layout/package.json
```diff
@@ -32,7 +32,7 @@
   "dependencies": {
     "@portabletext/react": "^1.0.0",
     "@reach/auto-id": "^0.13.2",
-    "@sanity/asset-utils": "^1.2.0",
+    "@sanity/asset-utils": "^1.2.5",
```
RitaDias, line 35: Do we need to add this dependency for the default-layout? 🤔

sjelfull: It's used in `PortableTextContent` already - this just changes the version

RitaDias: Of course, for some reason when I first saw it I didn't see the red on the previous version!

packages/@sanity/form-builder/src/inputs/files/common/ActionsMenu.tsx
```diff
@@ -11,8 +11,8 @@ interface Props {
   onReset: MouseEventHandler<HTMLDivElement>
   accept: string
   directUploads: boolean
-  downloadUrl: string
-  copyUrl: string
+  downloadUrl?: string
+  copyUrl?: string
```
RitaDias, line 15: I understand this will menu now is only being used in FileInputs (which is a different question that I brought up in the ImageInput). In what situations will we not have the downloadUrl / copyUrl for files?

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -552,30 +558,41 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
...
+                onUpload={this.handleSelectFiles}
+                browse={browseMenuItem}
+                onReset={this.handleRemoveButtonClick}
+                downloadUrl={imageUrlBuilder
+                  .image(value.asset)
+                  .forceDownload(
+                    assetDocument.originalFilename || `download.${assetDocument.extension}`
+                  )
+                  .url()}
+                copyUrl={imageUrlBuilder.image(value.asset).url()}
```
RitaDias, line 588: Since you've moved these out of the return (which is great!) I think we can just replace them with the set values on line 567 and 568 :)

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -552,30 +558,41 @@ export default class ImageInput extends React.PureComponent<Props, ImageInputSta
...
+          let downloadUrl: string | undefined
+
+          if (isImageSource(value)) {
+            const filename = assetDocument.originalFilename || `download.${assetDocument.extension}`
+            downloadUrl = imageUrlBuilder.image(assetDocument._id).forceDownload(filename).url()
+            copyUrl = imageUrlBuilder.image(assetDocument._id).url()
+          }
+
+          return (
+            <ImageActionsMenu
```
RitaDias, line 572: This might be more of a question for @rexxars since I see that this was also changed in the v3 version. But why did we split the menus specifically based on the image / file input? The menu (at least the one I recall) was with the same behavior?

RitaDias (changes requested):
Hi! i just have a couple of questions, great work 😄

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
sjelfull: Woops, forgot to move it

packages/@sanity/form-builder/src/inputs/files/common/ActionsMenu.tsx
sjelfull: It is still being used for both File and Image inputs. They can be undefined when called in `renderAssetMenu()` - although that doesn't seem to be called if this is an invalid image.

I mimicked the v3 change here - not sure why it is like that though.

packages/@sanity/form-builder/src/inputs/files/ImageInput/ImageInput.tsx
RitaDias: read @sjelfull earlier comment and I confused two different menus that are in the image input (confused `ImageActionsMenu` with `ActionsMenu` 🥲 ) nevermind this question ✨

packages/@sanity/form-builder/src/inputs/files/common/ActionsMenu.tsx
RitaDias: Ahh you're right, I just double checked and forgot that there are two "menu" type things in the image input. I resolve the other comment (unfortunate naming on my part for these things, whoops)

As for the optional props, I'd like to understand why (maybe if @bjoerge or @rexxars have the time) but it isn't a blocker for this :)

sjelfull: Ah, I believe the reason it is like this is that the v3 version has different setups for rendering preview vs field vs input, so this extra check makes sense. I'll change these to not be optional, since they are always set in v2.

RitaDias (approved):
Looks good to me! :sparkles:

packages/@sanity/form-builder/src/inputs/files/FileInput/InvalidFileWarning.tsx
```diff
@@ -0,0 +1,36 @@
...
+          <Text size={1}>
+            <WarningOutlineIcon />
+          </Text>
+        </Box>
+        <Stack space={3}>
+          <Text size={1} weight="semibold">
+            Invalid file value
+          </Text>
+          <Text size={1}>
+            The value of this field is not a valid file. Resetting this field will let you choose a
```
bjoerge, line 28: Are we able to provide more info about why it's not valid and what may have caused it? Wondering if it would be better to reword it into a bit more cautious terms to the editor aware of the impact and that it might require a bit of consideration (e.g. someone might be in the middle of a schema migration and blindly deleting it might not be the best cause of action).

bjoerge (approved):
Looks great! Just a small comment/question about the warning message

packages/@sanity/form-builder/src/inputs/files/FileInput/InvalidFileWarning.tsx
sjelfull: Worth considering for v3 / both v2 and v3, but I think it is kind of an edge case for now.

PR #3561, opened 2022-08-24
fix(form): read only fieldsets

hermanwikner:
### Description

This PR fixes an issue with read only in fieldsets. Looking at the schema example below – `myString` should be read only.

```ts
{
  name: 'myDocument',
  type: 'document',
  fieldsets: [
    {
      name: 'myFieldset',
      readOnly: true,
    }
  ],
  fields: [
    {
      name: 'myString',
      type: 'string',
      fieldset: 'myFieldset'
    }
  ]
}
```

### What to review

Make sure that fields inside a fieldset respect the fieldsets `readOnly` value.

### Notes for release

fix(form): read only fieldsets

### DIFF
--- a/packages/sanity/src/form/store/formState.ts
+++ b/packages/sanity/src/form/store/formState.ts
@@ -553,6 +553,13 @@ function prepareObjectInputState<T>(
       return []
     }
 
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
Good one, thanks!

PR #3562, opened 2022-08-25
Add alt text field for main image in blog schema

hidde:
### Description

The blog schema has a field for a main image, but no field to provide alternative text for that image. This change adds that.

This is from a suggestion by a community member and I feel it makes for a sensible default.

### What to review

Would this be a useful property and is it added in a satisfactory way?

### Notes for release

- The main image field in the blog schema now includes an alternative text field.


vicbergquist (changes requested):
lgtm

PR #3564, opened 2022-08-25
Lock document during publish

bjoerge:
### Description
This connects the "publish" document action with actual progress for publishing a document, and disables the editor if the current document is in the process of being published. This reduce the risk of getting into the situation where the client hasn't yet received the `createOrReplace` from the publish transaction and thus base new draft edits on an old and outdated local version of the published document.

Here's an illustration of how it will appear if there's a lot of latency (fake latency is effective in the deployment for this branch, to try live). The locking is normally a lot less disruptive, hardly noticeable when little latency.

NOTE: to make these changes production ready the following tasks needs to be done:

- Rewrite implementation of `isPublishMutation` and `allTransactionMessagesReceived` based on proper backend support for detecting
- Remove "fake latency"-step in `getPairListener.ts`
- Remove `publish-`-prefix from publish operation
- Update the publish operation test (currently disabled)

### What to review
Make sure everything related to publishing a document works as good as - or preferably *better* - than before.

### Notes for release
- Fixes a race condition that could happen during publish, causing the draft being overridden by an older version of the published document


bjoerge:
Superseded by https://github.com/sanity-io/sanity/pull/3588

PR #3565, opened 2022-08-25
fix(datastores): memoize edit state in document store instance

rexxars:
### Description

There is some global memoization going on in the document pair datastore utils. This isn't great:
1. Global state is a side effect. It can be hard to test these utils properly, and it could lead to memory leaks.
2. The keys used in the memoization is insufficient. For the edit state, it does not take into account changes in the schema - for instance `liveEdit`. It also does not take the source into account (projectId/dataset), which could lead to trouble in cases where document IDs and type names are shared (think production vs staging).

I want us to clean up all of these globally memoized things, but I figured I would open this initially so we could discuss/verify the approach.

The document store is already memoized with a key on `client, documentPreviewStore, historyStore, schema`, and I implemented a cache local to the document store for the edit state in particular. It looks like the listener gets unsubscribed as it should when constructing the document pair.

Thoughts? Is this the way to go, or should we address this differently?

### What to review

- Change in memoization
- To see the before/after:
  - `cd dev/starter-next-studio && yarn dev` - navigate to a document, then open `Studio.tsx` and change the schema type to be `liveEdit: true`. Watch how the footer does not reflect the change. Checkout this branch and try the same, and see it reflected.

### Notes for release

- Fixed certain schema properties not being reflected after a hot-module reload in Next.js/Webpack

### DIFF
--- a/packages/sanity/src/datastores/document/document-pair/editState.ts
+++ b/packages/sanity/src/datastores/document/document-pair/editState.ts
@@ -4,7 +4,6 @@ import {Observable} from 'rxjs'
 import {map, publishReplay, refCount, startWith} from 'rxjs/operators'
 import {HistoryStore} from '../../history'
 import {IdPair} from '../types'
-import {memoize} from '../utils/createMemoizer'
 import {operationArgs} from './operationArgs'
 import {isLiveEditEnabled} from './utils/isLiveEditEnabled'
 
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
This approach looks solid to me. I guess we could also consider to move memoization / caching out of the document store entirely at some point and instead cache it in the React/context layer. After all the reason we did this caching was to make sure we never instantiated more than one document store across the _runtime_, but that requirement has shifted to be about a _React tree_ now, I guess.

PR #3567, opened 2022-08-26
Connect the review changes button and its tooltip

KittyGiraudel:
At the moment, the “review changes” button which indicates the edit status of the open document, is not semantically connected to the tooltip it triggers. This pull-request intends to address this by creating a relationship between both nodes using `aria-describedby`. 

Screenshot 2022-08-26 at 13 21 41

**However**, this is not a perfect fix, because the tooltip is being added and removed from the DOM, but the `aria-describedby` attribute is not. This causes the toggle to reference an unknown ID with `aria-describedby` when the tooltip is not visible (since it’s the only time it’s actually in the DOM).

A simple fix for that would be for the toggle to be able to know if the tooltip is currently visible, but that doesn’t appear to be possible with the current tooltip component. I feel like the latter would benefit from a proper rewrite to use a hook or render props to convey its state to the toggle but that’s for another day.

After discussion with @hidde, we agreed that this would be marginally better than the current situation since some assistive technologies can pick up that newly formed relationship (but not all, due to the aforementioned issue).


vicbergquist (approved):
LGTM, thanks!

Issue #3495, opened 2022-08-10
Reference field is disabled initially


kmelve:
### Discussed in https://github.com/sanity-io/sanity/discussions/3391

<div type='discussions-op-text'>

<sup>Originally posted by **fabien** July  4, 2022</sup>
When no reference has been assigned yet, a field of type `reference` is disabled initially:

<img width="606" alt="image" src="https://user-images.githubusercontent.com/2749/177103335-c565f98f-41e5-497c-b4cd-393bad6fbb4f.png">

It's only after clicking 'Create new' and removing that draft again, that the search/autocomplete field will become enabled, in order to select a pre-existing item from the list.</div>

skogsmaskin:
This is now fixed.

PR #3568, opened 2022-08-26
Fix issue where reference input autocomplete was disabled

skogsmaskin:
Fixes https://github.com/sanity-io/sanity/issues/3495


bjoerge (approved):
Good catch! 🕵🏼

PR #3569, opened 2022-08-26
Add an accessible name to the presence menu button

KittyGiraudel:
The button that opens the presence menu (where one can see who else is interacting with the current document) does not have any accessible name. This is a problem for people using screen-readers. This pull-request addresses this issue by adding a customizable label.

Screenshot 2022-08-26 at 13 33 05

Related WCAG 2.1 SC: [4.1.2 Name, Role and Value](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.1#name-role-value)


RitaDias (approved):
Looks good! Thank you for your help ❤️

PR #3570, opened 2022-08-26
Remove unnecessary tabindex attribute

KittyGiraudel:
There is no reason for that title to be focusable. It’s not an interactive element and should not have a `tabindex` attribute.

If it needs to be focused programmatically (which it shouldn’t but we never know), `tabindex="-1"` should be used instead so it remains focusable via `.focus()`. Let me know if I should update the PR. :)


vicbergquist (approved):
LGTM, thank you!

PR #3577, opened 2022-08-29
feat(sanity): expose router

mariuslundgard:
### Description

- This change moves router-specific exports from `sanity/_unstable` to `sanity`.

```ts
// BEFORE THIS CHANGE
import {route, useRouter, useStateLink} from 'sanity/_unstable'

// WITH THIS CHANGE
import {route, useRouter, useStateLink} from 'sanity'
```

- This change revises the properties exposed by the various router hooks (rename `handleClick => onClick`).
- This change adds `@internal` to some exported members.

### What to review

- That the moved exported members are now importable using the `sanity` export instead of the `sanity/_unstable` export.

### Notes for release

n/a


rexxars (approved):
👍

PR #3580, opened 2022-08-29
fix(form): handle number and boolean changes in `elementProps`

rexxars:
### Description

The `onChange` handler of `elementProps` did not account for non-string types, emitting patches that used strings instead of number/boolean. This PR makes the change handler check the schema type for the item in question and emits patches of the corresponding type, using `valueAsNumber` for numbers and `checked` for booleans. Empty strings and `NaN` values are treated as "unset this field".

Also removes the custom `onChange` handler from the boolean input, which seems like a win?

### What to review

That all primitive inputs work as expected (numbers, strings, booleans)

### Notes for release

None, since the `elementProps` change has yet to be released.

### DIFF
--- a/packages/sanity/src/form/inputs/BooleanInput.tsx
+++ b/packages/sanity/src/form/inputs/BooleanInput.tsx
@@ -1,9 +1,6 @@
-/* eslint-disable import/no-unresolved */
-
-import React, {useCallback} from 'react'
+import React from 'react'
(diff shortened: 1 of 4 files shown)


hermanwikner (approved):
Thank you, works good! 👍

bjoerge (approved):
Yay!

PR #3590, opened 2022-09-02
fix(base): defer connection to presence client

RitaDias:
### Description

Hi all 👋 
Bjorge was kind enough to help me with a bug  regarding presence when people are in different datasets and go through the parts of the code that would be affected.

#### The bug
A project has two datasets: staging and production
User A is in a document in the production dataset and user B is in a document in the staging dataset.

User A goes to the presence menu, finds user B and attempts to travel to the document where they are only to be met with an error

image

#### The solution

To fix the actual navigation would be a large piece of work since the navigation/resolution of the document don't consider the dataset that the user is in. 
So, the solution for now is to hide users that are not in the same dataset and this way they won't appear on the list or be allowed to be navigated to if you are not on the same dataset as them. 

example:
public dataset on test project
image

private dataset on test project
image

### What to review

- Double check that you can still navigate to documents from the presence menu as usual
- Double check that if a user is in a different dataset that they won't appear on the list 

### Notes for release

Presence list will now only show users that are in the same dataset as the logged in user

### DIFF
--- a/packages/@sanity/base/src/client/bifur.ts
+++ b/packages/@sanity/base/src/client/bifur.ts
@@ -1,10 +1,13 @@
 import {fromUrl} from '@sanity/bifur-client'
+import {memoize} from 'lodash'
 import {authToken$} from '../datastores/authState'
 import {getVersionedClient} from './versionedClient'
 
-const bifurVersionedClient = getVersionedClient('2022-06-30')
-const dataset = bifurVersionedClient.config().dataset
+export const getBifur = memoize(() => {
+  const bifurVersionedClient = getVersionedClient('2022-06-30')
+  const dataset = bifurVersionedClient.config().dataset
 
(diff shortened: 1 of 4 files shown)


rexxars:
I'm missing _how_ this fixes the issue - is it because the bifur client is configured before the correct dataset is configured based on the URL routing?

rexxars (approved):
Approving based on the assumption that a late-configured client is the cause of this issue, and that deferring instantiation leads to the correct configuration being used.

RitaDias:
> I'm missing _how_ this fixes the issue - is it because the bifur client is configured before the correct dataset is configured based on the URL routing?

I should have made it clearer in the actual body, but yes your assumption on the approval message is right! Instantiating it later gives us enough time for the experimental spaces logic that sets the right dataset to run. Sorry for the confusion!

PR #3601, opened 2022-09-05
Schema config spike

snorrees:
### initialValue
`initialValue` functions are now supplied with a `ConfigContext` (contains `schema`, `client` and `currentUser`).

This is needed since createConfig -> schemaTypes config-api is resolved before client, user and schema is available.

We can now access context in initialValue functions like so:

```ts
  //async or not
  initialValue: async (params, context) => {
     // same params as fund in most createConfig callbacks
     const {client, schema, currentUser, projectId, dataset} = context
      
     // await client.fetch with data from currentUser for instance
  },
```

This change introduces breaking changes to the `resolveInitialValue` and `resolveInitialValueForType` functions, which as a bit of a cascading effect through the code. 

#### Considerations
We could consider making the context object optional with an empty object fallback, but that would make it a worse dev-experience when configuring initialValue as all context vals could be undefined. This would also make SanityArrayInput and the like a bit more risky to use. 

The implementation in this PR forces the use of `useResolveInitialValueForType` hook to capture the correct context, but if context is optional, `resolveInitialValueForType` could be passed directly, which would break things.

I feel breaking the API is worth it here for these functions, as they are kinda deeply buried and undocumented AFAIK.

**Note:** The typing of initialValue is not ideal via defineType. I have not changed the definition aside from adding the context, but making a note of it to see if we can improve it when taking a pass on defineType.

### Custom validation
`schema` was added to validation context. This was already close by and was easy to pass on.

Current user (and projectId/dataset) was not. The latter would break documentStore initialization function and make memoization of validation more difficult.

This is now possible:

```ts
   validation: (Rule) =>
        Rule.custom(
          async (value, context) =>
             const {client, schema} = context
             // schema.get(value) get another schema based on value, because why not
             // client was already part of context, but perhaps undocumented?
        ),
```

## What to review

* Are we ok with the breaking changes to `resolveInitialValue` and `resolveInitialValueForType` and cascading changes?
* Are we ok with _not_ adding currentUser to validation context?
* See explicit comments in the diff

### Notes for release

initialValue functions now has access to a context object with client, schema and currentUser:

```ts
  initialValue: async (params, context) => {
     const {client, schema, currentUser, projectId, dataset} = context
     // use client to determine initialValue, for instance
  }
```  

Custom validation function context now has access to schema alongside client:

```ts
   validation: Rule => Rule.custom(async (value, context) => {
      const {schema, client, ...restContext} = context
      // use schema and or client to validate
   }),
```


dev/test-studio/structure/groupByOption.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type StructureGroup = 'v3' // extend with union strings
+
+export interface StructureGroupOption {
+  structureGroup?: StructureGroup
+}
+
+type MaybeStructureOptions = StructureGroupOption | undefined
+
+export function typesInOptionGroup(
```
snorrees, line 12: Let me know if I should delete this. I just prefer defining structure from the schema 😅 ; this is a minimal approach with some typing to make find usage work.

package.json
```diff
@@ -119,7 +119,7 @@
     "dotenv-flow": "^3.2.0",
     "esbuild": "^0.14.43",
     "esbuild-register": "^3.3.2",
-    "eslint": "^8.23.0",
+    "eslint": "8.22.0",
```
snorrees, line 122: Note to self: revert this before merge; 8.23.0 does not work in IntelliJ atm.

packages/@sanity/types/src/schema/types.ts
```diff
@@ -123,7 +124,7 @@ export namespace Schema {
       preview?: React.ComponentType<any> // @todo: use `PreviewProps` here
     }
     // TODO
-    initialValue?: any
+    initialValue?: InitialValueProperty<any, any>
```
snorrees, line 127: Still a ways to go on this, but a slight improvement.

packages/sanity/src/desk/components/paneHeaderActions/InsufficientPermissionsMessageTooltip.tsx
```diff
@@ -22,7 +22,15 @@ export function InsufficientPermissionsMessageTooltip({
 
   return (
     <Tooltip
-      content={loading ? 'Loading…' : <InsufficientPermissionsMessage currentUser={currentUser} />}
+      content={
+        loading ? (
```
snorrees, line 26: Just noticed that tooltip for the create-document button in document list is unstyled when loading, this improves it.

packages/sanity/src/templates/resolve.ts
```diff
@@ -1,19 +1,19 @@
 import {
-  Schema,
   ArraySchemaType,
-  // InitialValueParams,
+  InitialValueProperty,
```
snorrees, line 3: This file is the most critical spot; many codepaths all of which needs context drilled.

dev/test-studio/structure/groupByOption.ts
```diff
@@ -0,0 +1,30 @@
+import {StructureBuilder} from 'sanity/lib/dts/src/_exports/desk'
```
mariuslundgard, line 1: this should probably be `sanity/desk` (this happens because of typescript links, i think)
```suggestion
import {StructureBuilder} from 'sanity/desk'
```

packages/@sanity/types/src/schema/types.ts
```diff
@@ -450,7 +451,18 @@ export interface ConditionalPropertyCallbackContext {
 export type ConditionalPropertyCallback = (context: ConditionalPropertyCallbackContext) => boolean
 export type ConditionalProperty = boolean | ConditionalPropertyCallback | undefined
 
-export type InitialValueResolver<Params, Value> = (params?: Params) => Promise<Value> | Value
+export interface SourceContext {
```
mariuslundgard, line 454: Hm, just wondering if `SourceContext` can be confused with `Source` (as returned by `useSource()`).

Could this be named `InitialValueResolverContext` for more clarity?

mariuslundgard, line 454: Hm, just wondering if `SourceContext` can be confused with `Source` (as returned by `useSource()`).

Could this be named `InitialValueResolverContext` for more clarity?

```suggestion
export interface InitialValueResolverContext {
```

packages/sanity/src/config/prepareConfig.ts
```diff
@@ -21,6 +22,7 @@ import {
   PreparedConfig,
   SingleWorkspace,
   WorkspaceOptions,
+  ConfigContext,
```
mariuslundgard, line 25: Hm – what is `ConfigContext` in relation to `SourceContext`?

packages/sanity/src/form/inputs/PortableText/toolbar/Toolbar.tsx
```diff
@@ -152,6 +152,7 @@ export function Toolbar(props: ToolbarProps) {
   const features = useFeatures()
   const editor = usePortableTextEditor()
   const selection = usePortableTextEditorSelection()
+  const resolveInitialValueForType = useResolveInitialValueForType()
```
mariuslundgard, line 155: Is this in use?

dev/test-studio/structure/groupByOption.ts
snorrees: Yeah, intellij autofix. I'll do a oass o er imports to ensure they are fixed.

packages/@sanity/types/src/schema/types.ts
snorrees: Sure. It is the same fields as ConfigContext but that type is in sanity package only afaict.

packages/sanity/src/config/prepareConfig.ts
snorrees: It is the same thing, but SourceContext is a redefinition due to package direction. Called it source since everything is pulled from current source, but see preview comment.

snorrees:
Renamed `SourceContext` to `InitialValueResolverContext` and used it everywhere (also where `ConfigContext` was previously used). Looking at the imports it seems more logically grouped now.

packages/@sanity/types/src/schema/types.ts
```diff
@@ -1,5 +1,6 @@
 // Note: INCOMPLETE, but it's a start
 import type {ComponentType} from 'react'
+import {SanityClient} from '@sanity/client'
```
mariuslundgard, line 3:
```suggestion
import type {SanityClient} from '@sanity/client'
```

packages/sanity/src/config/prepareConfig.ts
```diff
@@ -201,7 +201,8 @@ function getBifurClient(client: SanityClient, auth: AuthStore) {
 
   const url = bifurVersionedClient.getUrl(`/socket/${dataset}`).replace(/^http/, 'ws')
 
-  return fromUrl(url, auth.token ? {token$: auth.token} : {})
+  const options = auth.token ? {token$: auth.token} : {}
+  return fromUrl(url, options)
```
mariuslundgard, line 205: Is this change necessary?

packages/sanity/src/datastores/document/document-store.ts
```diff
@@ -30,7 +30,10 @@ function getIdPairFromPublished(publishedId: string): IdPair {
 
 export interface DocumentStore {
   checkoutPair: (idPair: IdPair) => Pair
-  initialValue: (opts: InitialValueOptions) => Observable<InitialValueMsg>
+  initialValue: (
+    opts: InitialValueOptions,
+    context: InitialValueResolverContext
```
mariuslundgard, line 35: Looking good!!! ✨

mariuslundgard (approved):
Looks great to me! See comments + one little nitpick with `import type {SanityClient}` 🧐

packages/sanity/src/config/prepareConfig.ts
snorrees: TypeScript was shouting that token$ is not allowed in literal type. Might just be IntelliJ shouting though, so I can revert if you want.

snorrees: Hmm I think maybe that was a fluke locally; reverting.

PR #3602, opened 2022-09-06
feat(cli): prefix commands if using npx

sjelfull:
### Description

This adds `npx` detection to the CLI, so all the help text will show the correct version of commands based on if you are using npx or not.

If running with npx it will look like this:

image

If running the `sanity` command as normal, it will look like this:

image

After doing research on the canonical way to detect `npx`, I ended up with a naive approach that checks if the folder that npx uses (`_npx`) is listed in `process.argv` or `process.env.npm_lifecycle_event` is `npx`. I have observed either when when testing.

### What to review

- Test that all the commands display either versions depending on if you use `npx` or not

### Notes for release

- CLI help text will now prefix commands with `npx @sanity/cli` instead of `sanity` if you are using `npx`


skogsmaskin (approved):
LGTM!

RitaDias (approved):
✨

PR #3608, opened 2022-09-07
perf(form-builder): skip passing parent for schema types with no conditional fields

bjoerge:
### Description
This optimizes field rendering by not passing the parent value to fields that has no conditional callbacks. In certain cases this improves rerendering speed with about 20%.

### What to review

- Verify that conditional fields works like before

### Notes for release

- Don't pass parent object to fields with no conditional callbacks


RitaDias (approved):
Looks good to me :) 

One thought that is more for the future: could we add tags to these prs (related to performance) so that we can have a look at them all in the future? 
I think it would be awesome if we could have a small brown bag about the changes & investigations we've done and the changes will be easier to find that way :) 

Great work!

PR #3610, opened 2022-09-07
fix(base): improve document operation typings

bjoerge:
### Description

This PR does not alter the runtime behavior and is merely a cleanup of the types used in the document operations that was originally done as part of the work in #3588, but submitted as a separate PR

### Notes for release

n/a internal only


rexxars (approved):
👍

PR #3611, opened 2022-09-07
refactor: optimize presence

bjoerge:
cc @rexxars 

### Description

Two PRs in one - they are closely related, so probably makes sense to review together. More details in each commit message.

### What to review

- Verify that presence works as before and does not trigger as many re-renders as we used to

### Notes for release

- Optimized Studio presence to trigger less re-renders


rexxars:
Note that the naming on these hooks and interfaces could probably have been improved, but we're going to take a closer look at this API in v3, so didn't want to spend too much effort on this right now.

PR #3612, opened 2022-09-07
feat: move to react `18`

stipsan:
### Description

Making the new baseline React `v18`.
This is a breaking change, and it requires a version of `@sanity/ui` that is also on react 18.
It sets the foundation for `v18`, but this PR tries its best to not introduce using new APIs like `startTransition`, `useTransition` or `useSyncExternalStore`. It's best to get this landed and adopt these new features gradually, and intentionally in cases where the impact and benefit can be measured.

### Other changes
- `renderStudio` now returns a function for unmounting instead of `void`. Useful for integrating with non-react libraries (vue, svelte, solid, angular, etc).
- React `StrictMode` is now enabled in development mode.


packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -158,6 +165,9 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
             </div>
             <Leaf
               {...lProps}
+              // @TODO: fix the typings on lProps.leaf and attributes
+              attributes={lProps.attributes as unknown as ElementAttributes}
+              leaf={lProps.leaf as unknown as SlateElement}
```
stipsan, line 170: Should this be in a follow-up shortcut?

packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -171,6 +181,9 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
       return (
         <Leaf
           {...lProps}
+          // @TODO: fix the typings on lProps.leaf and attributes
+          attributes={lProps.attributes as unknown as ElementAttributes}
+          leaf={lProps.leaf as unknown as SlateElement}
```
stipsan, line 186: Should this be in a follow-up shortcut?

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -701,6 +701,7 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
           </Grid>
         </Header>
         <SplitpaneContainer flex={1}>
+          {/* @ts-expect-error: https://github.com/tomkp/react-split-pane/pull/819 */}
```
stipsan, line 704: Should this be in a follow-up shortcut?

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -713,6 +714,7 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
                   - Disables resize if the container height is less then 500px
                   This should ensure that we mostly avoid a pane to take up all the room, and for the controls to not be eaten up by the pane
                 */}
+              {/* @ts-expect-error: https://github.com/tomkp/react-split-pane/pull/819 */}
```
stipsan, line 717: Should this be in a follow-up shortcut?

packages/sanity/src/components/collapseMenu/CollapseMenuButton.tsx
```diff
@@ -1,20 +1,18 @@
 import {Button, ButtonProps, TooltipProps} from '@sanity/ui'
 import React, {forwardRef} from 'react'
 
-export interface CommonProps extends Omit<ButtonProps, 'text' | 'icon' | 'iconRight'> {
+export type CommonProps = {
```
stipsan, line 4: This is rewritten to reuse `ButtonProps` parts instead of re-defining them. `@sanity/ui` have, for example, `React.ElementType | React.ReactNode` as the type for `icon`

packages/sanity/src/components/previews/helpers.tsx
```diff
@@ -33,5 +31,5 @@ export function renderPreviewNode<Layout = PreviewLayoutKey>(
     return createElement(value, {layout})
   }
 
-  return value || fallbackNode
+  return (value as React.ReactNode) || fallbackNode
```
stipsan, line 34: Not sure how to narrow these automatically

packages/sanity/src/components/previews/portableText/BlockImagePreview.tsx
```diff
@@ -43,25 +43,27 @@ export function BlockImagePreview(props: BlockImagePreviewProps) {
...
+            {(title || fallbackTitle) && (
+              <Text size={1} textOverflow="ellipsis" weight="semibold">
+                {title ? renderPreviewNode(title as any, 'block') : fallbackTitle}
+              </Text>
+            )}
 
             {subtitle && (
               <Text muted size={1} textOverflow="ellipsis">
-                {renderPreviewNode(subtitle, 'block')}
+                {renderPreviewNode(subtitle as any, 'block')}
```
stipsan, line 54: Didn't have time to figure out how to avoid using `as any` here.

packages/sanity/src/studio/renderStudio.tsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react'
-import {render} from 'react-dom'
+import React, {StrictMode} from 'react'
+import {createRoot} from 'react-dom/client'
```
stipsan, line 2: `createRoot` enables some of the new rendering features in React. But to use concurrent rendering we'll have to use new APIs like `startTransition`. The scope of this PR is to lay the groundwork and allow us to gradually adopt the new features in future releases.

packages/sanity/src/desk/actions/GetHookCollectionState.tsx
```diff
@@ -38,17 +34,17 @@ export function GetHookCollectionState<T, K>(props: GetHookCollectionStateProps<
     ricHandle.current = requestIdleCallback(() => {
       ricHandle.current = null
 
-      if (mountedRef.current) {
-        setTick((tick) => tick + 1)
-      }
```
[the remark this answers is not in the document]
snorrees: Oh how I have longed for this!

packages/sanity/src/desk/components/paneItem/PaneItem.tsx
```diff
@@ -91,7 +91,8 @@ export function PaneItem(props: PaneItemProps) {
   const LinkComponent = useMemo(
     () =>
       // eslint-disable-next-line @typescript-eslint/no-shadow
-      forwardRef(function LinkComponent(linkProps: any, ref: any) {
+      forwardRef(function LinkComponent(linkProps: ChildLinkProps, ref?: any) {
+        // @ts-expect-error - @todo fix `Expression produces a union type that is too complex to represent.ts(2590)`
```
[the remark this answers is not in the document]
snorrees: This stems from the `ComponentType` -> `ElementType` change in `ChildLink`.

Caused by the JSX property expansion in ElementType: it is simply massive and encapsulates the whole HTML spec 😨 
image

Doing a typeguard and fallback to React.createElement for the JSX bit seems to work though (only for TypeScript, hopfully runtime as well):
```ts
    forwardRef(function LinkComponent(linkProps: ChildLinkProps, ref: any) {
      if (typeof ChildLink === 'function') {
        return <ChildLink {...linkProps} childId={id} ref={ref} />
      }
      return React.createElement(ChildLink, {
        ...linkProps,
        childId: id,
        ref,
      })
    }),
```      

My question is this: Did we really have to change the router types from ComponentType?
Seems to me like these should be considered "internal" and that we should be able to ensure that its not passing JSX strings around?

packages/sanity/src/util/useThrottledCallback.ts
```diff
@@ -29,13 +29,5 @@ export function useThrottledCallback(
     [callback, options, wait]
   )
 
-  useEffect(
-    () => () => {
-      // cleanup: invoke the last value
-      throttledCallback.flush()
-    },
-    [throttledCallback]
-  )
```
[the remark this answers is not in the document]
stipsan: `useThrottledCallback` is only used by `GetHookCollectionState` which is used by:
- `RenderBadgeCollectionState`
- `RenderActionCollectionState`
In other words only for the document action buttons and menu, and list badges. I think it's ok to do it this way in this instance. As it just means that sometimes the callback can fire after unmounting (that's what the flush is preventing here) or it won't fire after unmount. Either case we're only talking about read state here and not write ops or input comps. 😌

packages/sanity/src/desk/components/paneItem/PaneItem.tsx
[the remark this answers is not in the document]
stipsan: I'm fine with attempting to reverting them, let's try

packages/sanity/src/config/types.ts
```diff
@@ -9,7 +9,7 @@ import type {
   SchemaTypeDefinition,
 } from '@sanity/types'
 import type React from 'react'
-import {ComponentType, ReactNode} from 'react'
+import {ElementType, ReactNode} from 'react'
```
rexxars, line 12:
```suggestion
import type {ElementType, ReactNode} from 'react'
```

packages/sanity/src/desk/components/paneItem/PaneItemPreview.tsx
```diff
@@ -15,74 +15,45 @@ import {getPreviewStateObservable, getValueWithFallback} from './helpers'
...
-        isPlaceholder={isLoading}
-        icon={icon}
-        layout={layout}
-        status={status}
-      />
-    )
-  }
+export function PaneItemPreview(props: PaneItemPreviewProps) {
+  const {icon, layout, presence, schemaType, value} = props
+  // const {title} = value
```
rexxars, line 27:
```suggestion
```

packages/sanity/src/studio/renderStudio.tsx
```diff
@@ -8,5 +8,11 @@ export function renderStudio(rootElement: HTMLElement | null, config: Config) {
     throw new Error('Missing root element to mount application into')
   }
 
-  render(<Studio config={config} />, rootElement)
+  const root = createRoot(rootElement)
+  root.render(
+    <StrictMode>
+      <Studio config={config} />
+    </StrictMode>
```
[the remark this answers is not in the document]
rexxars: Maybe we could add a third (optional) argument to `renderStudio()`?

```ts
export function renderStudio(
  rootElement: HTMLElement | null,
  config: Config,
  renderOptions?: {strict?: boolean}
) {
  // ...
}
```

I feel like it doesn't belong in the studio config itself

rexxars (approved):
Looks good - great work.

Let's make sure we have a proper release of `@sanity/ui`, `@sanity/icons`, `@sanity/color` and `@sanity/logos` we can use that is compatible and lets our jest tests pass before we merge this.

We should also prepare some release notes ahead of time explaining the rationale behind requiring React 18.

packages/@sanity/portable-text-editor/src/types/portableText.ts
```diff
@@ -45,12 +44,12 @@ export type PortableTextFeature = {
   value: string
   // Backward compatibility (blockEditor)
   blockEditor?: {
-    icon?: string | ComponentType<any>
-    render?: ComponentType<any>
+    icon?: string | React.ElementType
+    render?: React.ElementType
   }
```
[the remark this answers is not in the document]
skogsmaskin: Absolutely, I have already discussed this a bit with @snorrees.

We are now free to streamline this stuff with the rest, let's do that.

packages/@sanity/portable-text-editor/package.json
```diff
@@ -82,16 +83,16 @@
     "jest-environment-node": "^27.1.0",
     "node-ipc": "^9.2.0",
     "puppeteer": "^12.0.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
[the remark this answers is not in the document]
skogsmaskin: Wouldn't it be possible to be 17-18 in just that particular module?

packages/@sanity/vision/src/components/VisionGui.tsx
stipsan: Follow up task created https://app.shortcut.com/sanity-io/story/25140/replace-react-split-pane-with-allotment

mariuslundgard:
Just sending some 💘 to this PR. Great work here!

packages/sanity/src/desk/components/paneItem/PaneItem.tsx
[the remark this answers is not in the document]
stipsan: Reverting resolved the issue 👍

packages/@sanity/block-tools/src/types.ts
```diff
@@ -27,8 +26,8 @@ export interface BlockContentFeatures {
  * @beta
  */
 export interface BlockEditorSchemaProps {
-  icon?: string | ComponentType
-  render?: ComponentType
+  icon?: string | React.ElementType
```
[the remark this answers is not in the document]
stipsan: Reverted 👍

packages/@sanity/portable-text-editor/package.json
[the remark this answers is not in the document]
stipsan: Ok so it turns out we can't as the version of `@sanity/ui` that works with studios v2 are `^0.37` but react 18 will be published as `^0.39`. This dependency can obviously only use one or the other so the way I see it we can no longer use the same version of the editor in both studio major versions. 😅

packages/sanity/package.json
```diff
@@ -116,27 +116,27 @@
...
     "@sanity/types": "3.0.0-dev-preview.20",
-    "@sanity/ui": "^0.38.2",
+    "@sanity/ui": "0.38.3-next.18",
     "@sanity/util": "3.0.0-dev-preview.20",
     "@sanity/uuid": "^3.0.1",
     "@sanity/validation": "3.0.0-dev-preview.20",
     "@types/is-hotkey": "^0.1.7",
     "@types/react-copy-to-clipboard": "^5.0.2",
-    "@types/react-is": "^17.0.1",
+    "@types/react-is": "^17.0.3",
```
mariuslundgard, line 139: Is there no `^18` of `@types/react-is`?

stipsan: Not yet, no

packages/sanity/src/config/types.ts
```diff
@@ -9,7 +9,7 @@ import type {
   SchemaTypeDefinition,
 } from '@sanity/types'
 import type React from 'react'
-import {ComponentType, ReactNode} from 'react'
+import type {ComponentType, ReactNode} from 'react'
```
mariuslundgard, line 12: Just curious: is this change necessary? If so, there are a lot of places to update?

stipsan: @rexxars suggested this in his review:
https://github.com/sanity-io/sanity/pull/3612#discussion_r971501519
image

I don't have an opinion here and will follow the consensus 😄

mariuslundgard (approved):
Looks good to me. I think we should release `@sanity/ui@0.39` and update dependencies before releasing this in a new `dev-preview`.

Great work! 🙌

PR #3613, opened 2022-09-07
Define type 2.0

snorrees:
### Description
New implementation for `defineType`, `defineField` and `defineArrayMember`.

Adds a bunch of tests that uses `@ts-expect-error` to verify behavior of the new implementation

Changes some of the schema definition types, and makes the define functions strict-and-narrow.

### What to review

Skim over the implementation notes below. Does this seem ok?
Are you ok with the design concessions? This is all a battle between strictness and flexiblity and there is still a lot of room for improvement.

### Notes for release

`defineType`, `defineField` and `defineArrayMember` are now more narrowly typed.

# Implementation details

We make `defineType`, `defineField` and `defineArrayMember` available to increase type-safety and improve intellisense in IDEs . The latter two are optional, but helps with intellisense and type-narrowing.

## Design

Strict properties by default: Unknown properties should give compilation error.

### **Strict by default**

```tsx
defineType({
  type: 'string',
  name: 'custom-string',
  //@ts-expect-error preview does not exist in type StringDefinition
  preview: {},
  of: [],
})
```

**Opt-in to allow unknown props**

```tsx
defineType({
  type: 'string',
  name: 'custom-string',
  // strict: false, so now this is alloed
  preview: {},
  of: [],
}, {strict: false})
```

**Alias types use a best-effort, wide definition**

There is no way for us to know the underlying type of an alias type out-of-the-box. 
However, for most alias-types, a lot of properties cannot be overridden (`fields`, `of`, `to`).

This leaves us with a list of properties common across all types. We allow these in a broadened form, for instance any option is allowed.

```tsx
defineType({
	type: 'custom-string', 
	name: 'custom-string-redefined',
  // options can any anything
	options: {
	  // this is allowed, and makes sense for string
	  direction: 'horizontal',
	  // but so is this
	  slugify: () => null,
		andThis: true,
	},
  //@ts-expect-error thisIsStillNotAllowed is not in AliasTypeDefinition
  thisIsStillNotAllowed: true 
})
```

**Alias can be narrowed with a type-hint**
Second parameter in `defineType` can provide `alias` as a type hint to narrow the type of the definition.

```tsx
defineType({
	type: 'custom-string', 
	name: 'custom-string-narrowed-by-alias-parameter',
	options: {
	  direction: 'horizontal',
	  //@ts-expect-error slugify is not in StringDefinition
	  slugify: () => null
	},
}, {alias: 'string'})
```

## Narrow return type

We try to narrow the return type of `defineType` as much as possible. When defining a string schema, the returned type should be `StringDefinition` and not a union of all possible schema types.

```tsx
const stringDef: Schema.StringDefinition = defineType({
  type: 'string',
  name: 'custom-string',
})

//@ts-expect-error BooleanDefinition is not assignable to StringDefinition
const stringDef: Schema.BooleanDefinition = defineType({
  type: 'string',
  name: 'custom-string',
})
```

**Narrow initialValue and validation**

`initialValue` and `validation` are type per schema definition, so function values and chain-functions are narrowed to the correct type.

```tsx
defineType({
  type: 'string',
  name: 'custom-string',
  // initialValue will error if it does not resolve to string
  initialValue: () => Promise.resolve('string'),
  validation: (Rule) => [
    Rule.required()
      .min(1)
      .max(10)
      .length(10)
      .uppercase()
      .lowercase()
      .regex(/a+/)
      // value is string | undefined
      .custom((value) => (value?.toUpperCase() == 'SHOUT' ? 'Error' : true))
      .warning(),
    // @ts-expect-error greaterThan is not in StringRule
    Rule.greaterThan(5).error(),
  ],
})
```

**Type-safe select keys in preview.prepare**

`preview.prepare` value keys must come from the keys in `preview.select`.

```tsx
defineType({
  type: 'document',
  name: 'custom-document',
  preview: {
    select: {
      myTitle: 'myTitle',
    },
    // this will error as subtitle is not in keyof select ('myTitle')
    prepare: ({title}) => {
      return {title}
    },
  },
  fields: [{type: 'string', name: 'myTitle'}],
})
```

The `preview.prepare` values are typed as `any` and not as `unknown` . Here, convenience has ben chosen over strictness. An explicit type can still be provided for the values, if further type-safety is required.

```tsx
defineType({
  type: 'document',
  name: 'custom-document',
  preview: {
    select: {
      title: 'myTitle',
    },
    prepare: ({title}: {title?: string}) => {
      return {title}
    },
  },
  fields: [{type: 'string', name: 'myTitle'}],
})
```

## `interface` instead of `type` for intrinsics types

Previously, `IntrinsicTypes` was defined as a union type, It is now an interface (map) with type-name as key and type-definition as value. All schema types are derived from this map.

```tsx
//before
type IntrinsicTypeDefinition = ArrayDefinition | BlockDefinition | ...

//now
export interface IntrinsicTypeDefinition {
  array: ArrayDefinition
  block: BlockDefinition
  // ...
}
```

`type` is a closed definition in TypeScript; it cannot be extended. 

`interface` is an open definition and can be extended using [[Declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

This allows us to extend the `IntrinsicTypeDefinition` with user-types, which then again will give better type-safety and intellisense. 

For the code plugin it could look something like this (with a reduced feature set):

```tsx
export interface CodeOptions: {
   langauge: 'js' | 'ts
}

export type CodeDefinition = Omit<Schema.ObjectDefinition, 'type' | 'fields' | 'options'> &   {
  type: 'code'
  options: CodeOptions
}

declare module 'sanity' {
  export namespace Schema {
    export interface IntrinsicTypeDefinition {
      'code': CodeDefinition
    }
  }
}
```

 **All types have options** 

`options` has been added to all schema definitions, also for types that have no options. For these, an empty option type is defined. This is to support [[Declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) plugin options into any and all schema types.

```tsx

declare module 'sanity' {
  export namespace Schema {
    export interface DocumentOptions: {
		   myPluginAddedThis: boolean
		}
  }
}
```

### VS Code type autocomplete

IntelliJ will show a list of intrinsic types

Untitled (1)

But VS Code shows this:

Untitled (2)

Seems like IntelliJ uses the additional info in `Schema.Type | string` for intellisense, while VS Code just interprets it as a `string` (which is is, since Schema.Type is a subset of string).

### Exactly typed arrays (const array)

Ideally, we would return the *exact* type of `fields` as provided to `defineType` and not a union type, ie `[StringField<'fieldName1'>, AliasField<'fieldName2'>]` instead of `(StringField | AliasField)[]` . This would allow us to automatically generate content-lake types (the stored schema) based on the definition.

The current implementation does not do this unfortunately. `fields` (and `of` ) is typed using a broad union type `FieldDefinition` which only gives partial type-safety, and pretty bad intellisense. This is typically a problem because alias-types in this context collapse to `string` which impacts TypeScript inference in a bad way.

As a workaround the function `defineField` and `defineArrayMember` exists. If these are use, for everything, maximum safety and the best intellisense is achieved.

**Note on `defineField` for `image`**
Edit: isHighligheted is not a thing in v3, so this bit was removed.

### Partial generics

It would be nice if you could type hint using partial generics, for instance `defineType<'string'>({type: 'alias-type', /* smart type here */)` . Unfortunately this does not work as we want, as once you provide *any* generic argument explicitly, typescript will stop inferring the generic types and only use defaults for the once you omitted. In other words, it stops narrowing the type you pass to defineType automatically.

# TODO

We should get rid of Schema namespace and move each type definition into separate files.
Lets do that in a seperate PR.


rexxars:
I'm slightly unsure about the naming of the `defineArrayOf` - to me that sounds like you're either defining the array, or the `of` value, but you're actually defining an item _within_ the `of`. Maybe `defineArrayMember` or `defineArrayMemberType` or some such? Open to suggestions, but think it is slightly misleading as it stands.

snorrees:
> I'm slightly unsure about the naming of the `defineArrayOf` - to me that sounds like you're either defining the array, or the `of` value, but you're actually defining an item _within_ the `of`. Maybe `defineArrayMember` or `defineArrayMemberType` or some such? Open to suggestions, but think it is slightly misleading as it stands.

I tend to agree, what leaning towards `defineArrayMember` although also thinking about `defineArrayItem` as it is slightly shorter. MemberItem is established as input component for array items, so I guess I cant look there for the tiebreaker ;) I'm fine with either. `defineArrayMemberType` is a tad to long imo.

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * This function will narrow the schema type down to fields and options based on the provided
+ * type-string.
+ *
+ * Using `defineArrayOf` is optional, but should provide improved autocompletion in your IDE, when building your schema.
+ * Field-properties like `validation` and `initialValue`will also be more specific.
+ *
+ * See {@link defineType} for example usage.
+ *
+ * @param arrayOfSchema - should be a valid Sanity array.of type definition.
```
rexxars, line 174:
```suggestion
 * @param arrayOfSchema - should be a valid `array.of` member definition.
```

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,233 @@
...
+ * ## Note on type-safety in the current implementation
+ *
+ * Type-safety in arrays (`fields`, `of` ect) can only be guaranteed when {@link defineField} and {@link defineArrayOf}
+ * are used to wrap array-entries.
+ *
+ * For entries without a function-wrapper, TypeScript will resolve to a union type of all possible properties across
+ * all schema-types will be allowed.
+ *
+ * @param schemaDefinition - should be a valid Sanity schema type definition.
+ * @param defineOptions - optional param to provide typehints for the schemaDefinition.
```
rexxars, line 96:
```suggestion
 * @param defineOptions - optional param to provide type hints for `schemaDefinition`.
```

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,233 @@
...
+ * This function will narrow the schema type down to fields and options based on the provided
+ * type-string.
+ *
+ * Using `defineArrayOf` is optional, but should provide improved autocompletion in your IDE, when building your schema.
+ * Field-properties like `validation` and `initialValue`will also be more specific.
+ *
+ * See {@link defineType} for example usage.
+ *
+ * @param arrayOfSchema - should be a valid Sanity array.of type definition.
+ * @param defineOptions - optional param to provide typehints for the arrayOfSchema.
```
rexxars, line 175:
```suggestion
 * @param defineOptions - optional param to provide type hints for `arrayOfSchema`.
```

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,233 @@
...
+    MaybeAllowUnknownProps<TStrict>,
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  defineOptions?: DefineOptions<TStrict, TAlias>
+): typeof arrayOfSchema & WidenValidation & WidenInitialValue {
+  return arrayOfSchema
+}
+
+/**
+ * `typed` can be used to ensure that a object conforms to an exact interface.
```
rexxars, line 209:
```suggestion
 * `typed` can be used to ensure that an object conforms to an exact interface.
```

packages/@sanity/types/src/schema/defineTypes.ts
```diff
@@ -0,0 +1,83 @@
+import {PreviewConfig} from './preview'
+import {InitialValueProperty, Schema, SchemaValidationValue} from './types'
```
rexxars, line 2:
```suggestion
import type {PreviewConfig} from './preview'
import type {InitialValueProperty, Schema, SchemaValidationValue} from './types'
```

packages/@sanity/types/src/schema/defineTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export interface DefineOptions<
+  TStrict extends StrictDefinition,
+  TAlias extends Schema.Type | undefined
+> {
+  /**
+   * `strict: false` allows unknown properties in the schema.
+   * Use this when adding customizations to the schema that are not part of sanity core
```
rexxars, line 10: Perhaps add a note here suggesting extending the interface with options?

packages/@sanity/types/src/schema/defineTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  strict?: TStrict
+  /** Should be provided when type is a non-intrinsic type, ie type is a type alias*/
```
rexxars, line 13:
```suggestion
  /** Should be provided when type is a non-intrinsic type, ie type is a type alias */
```

packages/@sanity/types/src/schema/ruleBuilder.ts
```diff
@@ -0,0 +1,20 @@
+import {CustomValidator, FieldReference} from '../validation'
```
rexxars, line 1:
```suggestion
import type {CustomValidator, FieldReference} from '../validation'
```

rexxars (changes requested):
Some typos and nitpicks, but overall this is fantastic work!

A few notes what does not fit inline:

1. A field of type `text` should be allowed without specifying `options`, but is currently not I believe. Could you add a test and make `options` optional?
2. Looks like might previously have allowed `description` to be a `React.Element` (eg rendered JSX) - could you see if this is still the case, and if so update the typings?
3. There used to be a `blockEditor` option (now renamed to `portableText`, I believe) that allowed setting an icon and a render function on types. Is this still supported? If you are unsure, talk to PK and hear what the current status on this is? It might need to be added to the typings as well.

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * Define a Sanity type definition. This function will narrow the schema type down to fields and options based on the provided
+ *  type-string.
+ *
+ * Schemas defined using `defineType` should typically be added to the Studio config under `schema.types`.
+ * Defined types can be references by their `name`. This is referred to as a type-alias.
+ *
+ * When using type-aliases as `type`, `defineType` cannot know the base-type, so type-safety will be reduced.
+ * If you know the base type of the type-alias, provide `defineOptions.alias: <base type name>`.
```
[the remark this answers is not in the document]
snorrees: I like aliasFor

packages/@sanity/types/src/schema/defineTypes.ts
snorrees: Adding a chapter on declaration type merging with examples.

snorrees:
> 1. Could you add a test and make `options` optional?

Done!

> 2. Looks like might previously have allowed `description` to be a `React.Element` (eg rendered JSX) - could you see if this is still the case, and if so update the typings?

It indeed still works with JSX  stuff but not render functions, ie `string | ReactElement` as you say (but ReactNode).
Updated type and added a test.

> 3. There used to be a `blockEditor` option (now renamed to `portableText`, I believe) that allowed setting an icon and a render function on types. Is this still supported? If you are unsure, talk to PK and hear what the current status on this is? It might need to be added to the typings as well.

Changed typing from blockEditor -> portableText. Both are currently supported in v3. Made a task to revisit the API before GA.

packages/@sanity/types/src/schema/define.ts
```diff
@@ -0,0 +1,305 @@
...
+    Schema.FieldBase,
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  defineOptions?: DefineOptions<TStrict, TAlias>
+): typeof schemaField & WidenValidation & WidenInitialValue {
+  return schemaField
+}
+
+/**
+ * Define a an array item membeer type within an array definition `of`-array.
```
rexxars, line 236:
```suggestion
 * Define an array item member type within an array definition `of`-array.
```

rexxars (changes requested):

wimbarelds:
I realize this MR us closed, but I was reading this and thought I _might_ have a helpful insight.

Given that the intent seems to be to narrow the return type of `defineField` and similar functions; and the current returntype is `typeof schemaField & WidenValidation & WidenInitialValue`

I thought the following might be useful:
The return type of a function can be narrowed by typing your argument as a generic that extends your interface, rather than the interface itself. Ie:

```typescript
const myFunction = <T extends ArgumentInterface>(argument: T): T => argument;
```
Or as an extremely simplified example:
```
const str = <T extends string>(arg: T): T => arg;
const foo = str('bar');
```
The resulting type of `foo` here isn't `string`, it is "bar". [Link to Typescript Playerground example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBNBOMC8MA8AVGBTAHlLYAJhHFPAJZgDmAfABQCG8VAXDOgJRuZI0xNUA3AChQkWADMQIZKXh0A5ACMmCjoKA).

The same holds true for objects and the like. It lets you ensure that a value is in compliance with a type/interface, without "widening" its type to that interface, and in fact maximally narrowing it.

snorrees:
> I realize this MR us closed, but I was reading this and thought I _might_ have a helpful insight.
(...)

Hey, and thanks for the input :) Keep it coming! 

You are brushing on a couple of things that we have tried to address with the new helpers, and we _are_ actually using the functionality you are mentioning.

I'll try to explain the thought behind how it works right now, and why it is done that way 😅. There are some concessions done to improve autocomplete inference at the cost of  specificity in the return value. Hopefully this all makes sense, but feel free to disagree/chip in:


## Narrow / widen for  `defineField` 

`defineField` has to be compatible with ObjectDefinition `fields`(`FieldDefinition[]`). 

Unfortunately,`FieldDefinition` is a union-type that resolves to compromise type for validation and inititialValue, which makes intellisense less than stellar. You can see this if you try to provide fields to a document or object directly in `defineType`, without wrapping each in `defineField`. Without defineField, you are allowed to pass numbers as inititialValue to string, for instance. 

So, `defineField` exist to provides a narrower type for each field. These narrowed types are not strictly assignable to FieldDefinition (because of type-narrowed validation and initialValue). As a workaround, we do a cheeky widen in the return type to make it all line up. The `typeof schemaField` is there to reduce boilerplate; it captures the type of the input schema, then we tack on the wider compromise types that make it assignable to `FieldDefinition`.

In a perfect world, `defineType` around the outer definition would be sufficient, but we are not there yet. Happy to hear how to fix it :)

## On `T extends X`

It is correct that `T extends string` will narrow to an exact string literal. In fact, we are using this throughout the API to narrow `type` provided in `defineType` et al, so we can lookup the exact definition in `IntrinsicDefinitions` :)

However, when you use the same pattern for objects, things work a bit differently; you are allowed to provide unknown properties:

 ```
 interface MyObjectDef {
   someField: string
 }
 
 const objDef = <T extends MyObjectDef>(arg: T): T => arg;`
 const obj = objDef({
   someField: 'fieldvalue',
   andChuckThisOnForGoodMeasure: true // this will compile!
} )
 ```

Because we want the helpers to error when you have typos, we _dont_ want to allow unknown properties by default.

But what if you DO want to allow unknowns then? Well, we got you covered. `defineType` et al takes a second argument, which changes the behavior so that it DOES use the above T extends pattern:

```ts
defineType({
  type: 'string',
  name: 'a',
  error: true // compile error here
})
```

wimbarelds:
Didn't notice that there had been a response to this :(

Not wanting to accept unknown keys makes sense!
There are some options to work around that, like:

```typescript
interface ObjectDef {
  someField: 'a' | 'b'
}
type Only<Base, ArgType> = Base & {[k in Exclude<keyof ArgType, keyof Base>]: never};
type OnlyObjectDef<ArgType> = Only<ObjectDef, ArgType>;

const fn = <ArgType extends OnlyObjectDef<ArgType>>(arg: ArgType): ArgType => arg;

const a = fn({
  someField: 'a',
});

const b = fn({
  someField: 'a',
  someOtherField: 123, // Type 'number' is not assignable to type 'never'
});
```

But that only works 1 level deep and is of somewhat questionable elegance.

Thanks for the explanation!

PR #3616, opened 2022-09-08
feat(desk)!: drop `preserveInstance`, `urlParams` options on custom user panes

rexxars:
### Description

This PR introduces two breaking changes on custom user panes, since we've been wanting to get rid of these since v2. This has to be done before a GA v3 release so as not to cause another breaking change that could easily have been done in this release.

**BREAKING CHANGE**: we've been wanting to make `preserveInstance` the default behavior for custom panes for a while, but it has been a breaking change. Now that we have the ability to do so, lets reduce the complexity of things. User component panes should use `componentDidUpdate` or a `useEffect` with a dependency on the `itemId`/`childItemId` prop to react to updates.

**BREAKING CHANGE**: `urlParams` has been available through the `usePaneRouter()` hook as `params` since version 2.2.0. This commit drops this prop from being passed, since we prefer it to be retrieved through the hook instead.

### What to review

That user component panes still _work_, but do not remount on item ID changes anymore. This difference can be seen in the test studio - switch between "component pane 1" and "component pane 2". In the old version, the `Random ID` would change on every switch. In the new version, it stays the same.

Also note that the `urlParams` property is gone in the new version, but that the router still has access to it.

- Previous: https://test-studio-rfpvdjuqn.sanity.build/test/content/custom;component1
- This change: https://test-studio-eopmyw79f.sanity.build/test/content/custom;component1

### Notes for release

- BREAKING CHANGE: The `preserveInstance` option is now the default behavior for custom desk tool panes, meaning the components used needs to account for a change in `itemId`/`childItemId` and update accordingly
- BREAKING CHANGE: The `urlParams` property has been removed from custom desk tool panes - the same information has been available through the `usePaneRouter()` hook since v2.2.0 and is now the only way of retrieving the URL parameters for a pane

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -92,59 +92,55 @@ export const structure: StructureResolver = (S, {schema}) => {
                 .id('component1')
                 .title('Component pane (1)')
                 .child(
-                  Object.assign(
-                    S.component(DebugPane)
-                      .id('component1')
-                      .title('Component pane #1')
-                      .options({no: 1})
-                      .menuItems([
-                        S.menuItem()
-                          .title('From Menu Item Create Intent')
-                          .intent({
-                            type: 'create',
-                            params: {type: 'author', id: `special.${uuid()}`},
-                          }),
-                        S.menuItem()
-                          .title('Also Menu Item Create Intent')
-                          .intent({
-                            type: 'create',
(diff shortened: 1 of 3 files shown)


mariuslundgard (approved):
It works, looks good to me 👍

bjoerge (approved):
Don't have the bandwidth to review in depth but LGTM!

PR #3619, opened 2022-09-09
Improve PT-input readOnly mode

skogsmaskin:
### Description

This will fix a couple of thing when the PT-input is in readOnly mode.

* Improved colors based on theme marking the field as readOnly.
* Improved functionality when navigating embedded objects inside the PT when in readOnly.
* Upgraded the PTE to latest version.
image

image


packages/@sanity/form-builder/src/inputs/PortableText/Editor.styles.tsx
```diff
@@ -64,6 +67,7 @@ export const EditableWrapper = styled(Card)<{$isFullscreen: boolean; $readOnly:
   width: 100%;
   counter-reset: ${TEXT_LEVELS.map((l) => createListName(l)).join(' ')};
   overflow: hidden;
+  --card-bg-color: 'inherit';
```
hermanwikner, line 70: It is invalid to use quotes here, should be:

`--card-bg-color: inherit`

packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObject.styles.ts
```diff
@@ -125,12 +129,16 @@ export const BlockActionsInner = styled(Flex)`
...
-`
-export const BlockPreview = styled(Box)((props: {theme: Theme}) => {
-  const color = props.theme.sanity.color.input
+export const StyledButton = styled(Button)((props: {theme: Theme; readOnly: boolean}) => {
   return css`
-    background-color: ${color.default.enabled.bg};
+    --card-bg-color: ${props.readOnly
+      ? props.theme.sanity.color.input.default.disabled.bg
+      : 'inherit'};
+    --card-border-color: 'inherit';
```
hermanwikner, line 137: Should be without quotes. Not on the line above, because it's JS and not CSS!

hermanwikner: When we remove the quotes here, an issue is fixed with the focus outline not showing for this button when navigating with the keyboard! 👍

packages/@sanity/form-builder/src/inputs/PortableText/Editor.styles.tsx
hermanwikner, line 70: It is invalid to use quotes here, should be:

`--card-bg-color: inherit`

packages/@sanity/form-builder/src/inputs/PortableText/object/BlockObject.styles.ts
hermanwikner, line 137: Should be without quotes. Not on the line above, because it's JS and not CSS!

When we remove the quotes here, an issue is fixed with the focus outline not showing for this button when navigating with the keyboard! 👍

hermanwikner (commented):
Looks nice 👍 But I have some suggestions:

The `Card` primitive uses React context to pass down tone values to its children. When we modify the `Card` color values with CSS – we lose the inheritance of colors provided by the `Card`(s) – and e.g. `Button` doesn't look correct on hovering/focus/pressed states.

The `disabled` color that this PR uses is the same as the `Card`:s `transparent` tone. I suggest that we skip custom CSS on `Card`:s – and instead set the `tone` to "transparent" when we want the disabled styling.

Suggested changes:

**Editor.tsx**
```js
<EditableCard tone={readOnly ? 'transparent' : tone} />
```
And remove custom `--card-bg-color` in on `EditableCard` in `BlockObject.styles.ts`

**BlockObject.tsx**
```js
<Root tone={readOnly ? 'transparent' : tone} />
```

**BlockObjectPreview.tsx**

Replace `StyledButton` with a standard `Button`. The styling of the button will be handled "automatically" when using `tone` on the above mentioned `Card`:s

packages/@sanity/form-builder/src/inputs/PortableText/Editor.styles.tsx
skogsmaskin: Thank you!

skogsmaskin:
@hermanwikner - thank you for your suggestions. I've made a new commit with your simple solution.

hermanwikner (approved):
👍

PR #3620, opened 2022-09-09
feat(server): print error on process.env usage in development

rexxars:
### Description

Note: I am not 100% confident that this is a good idea, but I can't think of a better approach.

When migrating from Sanity v2 to v3, users might attempt to use the `process.env.SANITY_STUDIO_` approach for getting environment variable values. However, we are using Vite, and it instead uses `import.meta.env.SANITY_STUDIO_` for the same purpose.

We have a couple of options here:
- Leave it as-is, crashing when attempting to access `process`
- Declare all `SANITY_STUDIO_` prefixed env variables both in `import.meta.env` _and_ `process.env`
- Implement a proxy/object with getters at `process.env` that warns/errors on use

This PR uses the last approach, but as I wrote in the opening notes; I am not convinced it is a good idea. It leads to the side effect that `process` and `process.env` gets defined, which _might_ have unintended consequences.

Open for discussion.

### What to review

- Is this a good idea?
- Would it be better to just document the change?
- Would it be better to check this differently? (esbuild plugin?)

### DIFF
--- a/packages/@sanity/server/src/getViteConfig.ts
+++ b/packages/@sanity/server/src/getViteConfig.ts
@@ -96,6 +96,7 @@ export async function getViteConfig(options: ViteOptions): Promise<InlineConfig>
     resolve: {
       alias: getAliases({monorepo}),
     },
+    define: mode === 'development' ? {'process.env': getProcessEnvProxy()} : {},
   }
 
   if (mode === 'production') {
@@ -150,4 +151,14 @@ export function finalizeViteConfig(config: InlineConfig): InlineConfig {
       },
     },
   }
+}
+
+function getProcessEnvProxy() {
+  return `new Proxy({}, {
+    get: (target, key) => {
+      if (key.indexOf('SANITY_STUDIO_') === 0) {
+        console.warn('process.env is not available - use "import.meta.env.' + key + '" instead')
+      }
+    }
+  })`
 }


mariuslundgard (approved):
Nice touch 👍

bjoerge (approved):
I might be wrong, but I'm guessing that `if (process?.env) {...}` may not be a very common pattern and that most existing code paths out there would just blindly assume it's already there.

Would it make sense to warn for variables other than `SANITY_STUDIO_*`? E.g. if including a third party script that assumes the existence of `process.env`, we could catch it too and provide the same kind of useful information.

rexxars:
> Would it make sense to warn for variables other than `SANITY_STUDIO_*`? E.g. if including a third party script that assumes the existence of `process.env`, we could catch it too and provide the same kind of useful information.

I tried that approach, but there is actually quite a number of packages that check for the existence of `process && process.env` (in various ways) - `DEBUG`, `SC_ATTR` and such - I think there was easily 8 different environment variables that were attempted to be accessed. This is one of the reasons I am unsure whether or not this is a good idea altogether and that we should rather rely on an esbuild plugin or other approach.

skogsmaskin:
Re-opening this as it was closed when renaming branches yesterday.

rexxars:
Superseded by https://github.com/sanity-io/sanity/pull/4046

PR #3623, opened 2022-09-09
fix: move presence menu to leftmost position in navbar

rexxars:
### Description

This is done because the number of people present changes over time, causing occasional "jumping" when people enter/exit. With the changelog/warning button to the left of the presence menu, you can end up in situations where you click on the wrong thing by accident, since the layout shifts. It also causes less of a layout shift on update.

### What to review

That things still look alright, and that you agree that this makes sense.

### Notes for release

Probably not worth mentioning.


mariuslundgard (approved):

hermanwikner (approved):
Nice 👍

PR #3624, opened 2022-09-09
feat(form): future-proof patch types

rexxars:
### Description

We might want to provide alternative ways of specifying patches in the future. By adding a `patchType` symbol to the patches returned by our patch helpers, we can future-proof that we can differentiate between a "patch helper generated" patch and an arbitrary value.

Wasn't quite sure about the naming on this one - open to suggestions.

### Notes for release

Probably not worth mentioning.

### DIFF
--- a/packages/sanity/src/form/patch/types.ts
+++ b/packages/sanity/src/form/patch/types.ts
@@ -1,6 +1,19 @@
-/* eslint-disable camelcase */
+import type {Path} from '@sanity/types'
 
-import {Path} from '@sanity/types'
+/**
+ * @alpha
+ */
+export interface FormPatchBase {
+  /**
+   * A property used to identify this as a Sanity patch type, eg "set", "unset", "insert", etc.
+   * This allows us to potentially introduce new patch types in the future without breaking
+   * existing code. This is an internal property/implementation detail and should not be used by
+   * consumers.
+   *
+   * @internal
+   */
+  patchType: symbol
+}
 
 /**
  * @alpha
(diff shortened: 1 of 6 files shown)


packages/sanity/src/form/patch/types.ts
```diff
@@ -1,6 +1,11 @@
-/* eslint-disable camelcase */
+import type {Path} from '@sanity/types'
 
-import {Path} from '@sanity/types'
+/**
+ * @alpha
+ */
+export interface FormPatchBase {
+  patchType: symbol
+}
```
mariuslundgard, line 8: Should we instead mark the property as `@internal`?

React calls these `$$typeof` – maybe a convention we can use too?

```suggestion
/**
 * @public
 */
export interface FormPatchBase {
  /** @internal */
  $$typeof: symbol
}
```

mariuslundgard (approved):
Looks good to me. Nitpick: we can make the property `@internal` (see comment).

packages/sanity/src/form/patch/types.ts
rexxars: I'm alright with both of those changes - lets hear if @bjoerge agrees

bjoerge: +1 to making it internal. Not sure about adopting `$$typeof` - feels too much like relying on an internal React convention, and worried it might add confusion. I.e. during debugging, `$$typeof` provides a quick way to eyeball the react boundary (e.g. is this thing coming from React or our code?). I'd say keep it as `patchType`, but yeah, keep it `@internal`

bjoerge (approved):

PR #3625, opened 2022-09-09
feat(types): add missing `Geopoint` data type

rexxars:
### Description

Adds a typescript interface for the `Geopoint` data type that is represented by the [geopoint](https://www.sanity.io/docs/geopoint-type) schema type.

### What to review

- Comments and shape seems correct and does not contain typos

### Notes for release

- Add and export `Geopoint` typescript interface

### DIFF
--- a/packages/@sanity/types/src/schema/types.ts
+++ b/packages/@sanity/types/src/schema/types.ts
@@ -762,3 +762,30 @@ export interface ImageSchemaType extends Omit<ObjectSchemaType, 'options'> {
     sources?: AssetSource[]
   }
 }
+
+/**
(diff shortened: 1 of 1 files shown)


hermanwikner (approved):
Nice! 👍 

Shouldn't it be `GeoPointDefinition` and `GeoPoint` with a capitalized P?

rexxars:
> Shouldn't it be `GeoPointDefinition` and `GeoPoint` with a capitalized P?

This isn't the schema definition typings - those cannot actually be configured in any meaningful way (you just use `type: 'geopoint'`), and since the type name is `geopoint` and not `geoPoint`, I feel `Geopoint` is the correct casing.

PR #3632, opened 2022-09-13
feat(cli): use define helpers for template schema types

rexxars:
### Description

Makes the CLI templates use the new `defineType`, `defineField` and `defineArrayOf` (to be renamed) helpers. This was done as an exercise to see if all our own templates passed the new definitions, which they seem to do apart from the `text` type currently requiring `options`.

The shopify template shouldn't necessarily work as-is (it has yet to be ported to v3), but the others should work.

### What to review

That the usage is correct

### Notes for release

- New project templates now use the `defineType` and `defineField` helpers


snorrees (approved):
Very nice! 

Nice to have more examples of it working, although the real test is using them when writing schemas by hand. I have found some cases where IntelliJ refuse to autocomplete valid properties, but I guess that is up to the intellisense spirits to decide :shrug:

Hit the same eslint issue as CI is reporting , when updating defineArrayOf -> defineArrayMember. Pushed without githooks, but this seems to be a real error?

AFAICT "@sanity/eslint-config-studio" seems to be only intended for the templates themselves, but I am unsure if the `cli/templates/.eslintrc` file can be deleted (ESLint uses it when visiting the template directory), or if cli should add this as an devDependency.

Also unsure if a fix goes in this branch or later.

rexxars:
> Pushed without githooks, but this seems to be a real error?

Yes, but it's a minor annoyance when touching the templates - don't want to spend time on it now. Later commits not touching the templates won't fail because of it.

> I still maintain that keeping selection values explicit any is best DX, but making it unknown would have required adding types here. Are we ok with any as the default?

I am, yes.

PR #3635, opened 2022-09-13
feat(sanity): implement studio components API

hermanwikner:
### Description

This PR implements an API that makes it possible to wrap, modify or replace components in the studio. This PR enables this for `Layout`, `Logo`, `Navbar` and `ToolMenu`. Example usage:

```jsx
const config = createConfig({
  studio: {
    components: {
      Navbar: NavbarWithBanner,
      ToolMenu: ToolMenu
    }
  }
})

function NavbarWithBanner(props: NavbarProps) {
  const { roles } = useCurrentUser()
  const showBanner = roles.find((r) => r.name === 'editor')

  return (
    <Stack>
      {showBanner && <Banner />}
      {props.renderNavbar(props)}
    </Stack>
  )
}

function ToolMenu(props: ToolMenuProps) {
  const {renderToolMenu, tools} = props
  const nonDevTools = tools.filter((t) => t.name !== 'vision')

  const conditionalTools = isDev ? tools : nonDevTools

  return renderToolMenu({...props, tools: conditionalTools})
}
```

Note: this PR removes the `logo` property in config since it is replaced with `studio.components.Logo`.

### What to review
- Code review
- Test that the API works as expected (in `createConfig` and `createPlugin`, in combination)
- Should we add more components in this iteration, or remove any of the components in this PR?

### Notes for release
feat(sanity): implement studio components API


rexxars (approved):
This is great work - I love this approach:

1. It's just a React component in the simplest case
2. 4. You can use hooks right from the config, avoiding what could have been a common footgun
3. It's still composable (ergo defer to next in line)
4. Props are overridable

Ideally I'd like to see this same approach implemented for diff, field and input components as well, unless others have objections!

dev/test-studio/sanity.config.ts
```diff
@@ -79,19 +71,43 @@ export default createConfig([
...
     name: 'default',
     title: 'Test Studio',
-    logo: Branding,
     projectId: 'ppsg7ml5',
     dataset: 'test',
     plugins: [sharedSettings()],
     basePath: '/test',
+    studio: {
+      components: {
+        Logo: Branding,
```
mariuslundgard, line 80: I know this is just the `test-studio`, but would be good to rename `Branding` to `Logo` imo.

mariuslundgard (commented):
Looks great to me!

PR #3636, opened 2022-09-13
Bug/remove empty on close

snorrees:
### Description

Restores v2 behavior: "delete on close" for empty array items.

Simply checks if the current value only contains _key and _type and if so, invokes onRemove.

Note: not tested with renderItem as it seems to be unimplemented atm?

### What to review

All the changes ;)

### Notes for release

Empty array items are removed when the edit dialog is closed.

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -357,7 +357,8 @@ export const structure: StructureResolver = (S, {schema}) => {
           !STANDARD_PORTABLE_TEXT_INPUT_TYPES.includes(id) &&
           !PLUGIN_INPUT_TYPES.includes(id) &&
           !EXTERNAL_PLUGIN_INPUT_TYPES.includes(id) &&
-          !DEBUG_FIELD_GROUP_TYPES.includes(id)
+          !DEBUG_FIELD_GROUP_TYPES.includes(id) &&
+          !typesInOptionGroup(S, schema, 'v3').includes(id)
         )
       }),
     ])
--- a/packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
+++ b/packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
@@ -141,12 +141,13 @@ export function ArrayOfObjectsItem(props: MemberItemProps) {
     onPathOpen(member.item.path)
   }, [onPathOpen, member.item.path])
 
+  const isEmptyValue = !member.item.value || isEmpty(member.item.value)
   const handleClose = useCallback(() => {
-    if (!member.item.value || isEmpty(member.item.value)) {
+    if (isEmptyValue) {
       onRemove()
     }
     onPathOpen(member.item.path.slice(0, -1))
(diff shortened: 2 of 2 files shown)


dev/test-studio/structure/resolveStructure.ts
```diff
@@ -357,7 +357,8 @@ export const structure: StructureResolver = (S, {schema}) => {
           !STANDARD_PORTABLE_TEXT_INPUT_TYPES.includes(id) &&
           !PLUGIN_INPUT_TYPES.includes(id) &&
           !EXTERNAL_PLUGIN_INPUT_TYPES.includes(id) &&
-          !DEBUG_FIELD_GROUP_TYPES.includes(id)
+          !DEBUG_FIELD_GROUP_TYPES.includes(id) &&
+          !typesInOptionGroup(S, schema, 'v3').includes(id)
```
snorrees, line 361: Drive-by fix: v3 schemas was added twice in the test-studio structure.

packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
```diff
@@ -122,8 +142,11 @@ export function ArrayOfObjectsItem(props: MemberItemProps) {
   }, [onPathOpen, member.item.path])
 
   const handleClose = useCallback(() => {
+    if (!member.item.value || isEmpty(member.item.value)) {
+      onRemove()
+    }
     onPathOpen(member.item.path.slice(0, -1))
-  }, [onPathOpen, member.item.path])
+  }, [onPathOpen, member.item.path, member.item.value, onRemove])
```
snorrees, line 149: The callback will change every time value changes; I couldnt think of a way to avoid that.

packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
```diff
@@ -50,6 +51,25 @@ export function ArrayOfObjectsItem(props: MemberItemProps) {
     }
   })
 
+  const onRemove = useCallback(() => {
```
snorrees, line 54: Had to move onRemove up since it is now used by handleClose. 
Also moved onInsert since they belong together imo.

packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
mariuslundgard: Nitpick: It might not be worth it, but we could add a variable named `isEmptyValue` or something to avoid the callback change as often.

```ts
const isEmptyValue = !member.item.value || isEmpty(member.item.value)

const handleClose = useCallback(() => {
  if (isEmptyValue) {
    onRemove()
    onPathOpen(member.item.path.slice(0, -1))
  }
}, [isEmptyValue, member.item.path, onPathOpen])
```

mariuslundgard (approved):
Looks good to me!

packages/sanity/src/form/members/items/ArrayOfObjectsItem.tsx
snorrees: Love the idea!

PR #3638, opened 2022-09-13
fix: corrected object level indentation for slug and alias types

snorrees:
### Description

Slug had level indentation. This is now removed.
The code that handled level indentation for objects did not account for recursive alias types. This is now added.

Moved the resolution of levels into formState so that this behaves equally in all components.
Should also be a tiny pref boost, since we dont have to recalculate level every render.

Added a test-suit to try to account for most codepaths.

#### Before
image

image

#### After
image

image

### What to review

Evaluate the tests-cases.

image

Is the move from components to formState ok?

Can be tested in the studio: added alias for slug and reference. 

### Notes for release

Slug no longer has indentation next to the input.

### DIFF
--- a/packages/sanity/src/form/studio/inputResolver/helpers.test.ts
+++ b/packages/sanity/src/form/studio/inputResolver/helpers.test.ts
@@ -1,108 +1,181 @@
 /* eslint max-nested-callbacks: */
 
-import {getObjectFieldLevel, type PartialObjectFieldProps} from './helpers'
-
-type SchemaSnippet = PartialObjectFieldProps['schemaType']
-type SchemaTest = SchemaSnippet & {description: string}
-
-const noIndentation: SchemaTest[] = [
-  {description: 'any array', name: 'array'},
-  {description: 'file without additional fields', name: 'file', fields: [{name: 'asset'}]},
-  {
-    description: 'image without additional fields',
-    name: 'image',
-    fields: [{name: 'asset'}, {name: 'crop'}, {name: 'hotspot'}],
-  },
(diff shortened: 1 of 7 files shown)


packages/sanity/src/form/studio/inputResolver/helpers.test.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint max-nested-callbacks: */
+
+import {getObjectFieldLevel, type PartialObjectFieldProps} from './helpers'
```
snorrees, line 3: Rewriting to use createSchema, as it is not modelling the types correctly and the fix doesnt actually work 😆

packages/sanity/src/form/studio/inputResolver/helpers.ts
```diff
@@ -9,15 +9,30 @@ export function getOption(type: SchemaType, optionName: string) {
...
+  type?: PartialObjectSchema
+  fields?: {name: string}[]
+}
+
+export interface PartialObjectFieldProps {
+  schemaType: PartialObjectSchema
+  level: number
+}
+
+export function getObjectFieldLevel(field: PartialObjectFieldProps): number {
```
snorrees, line 24: I have this hunch that this information should be encapsulated in the field props/field definition time when you have more local context. This logic is very hard to understand imo. I can investigate if we can move this if you want, otherwise I'll just fix it (as written this is wrong since slug root type is 'object'.

hermanwikner (approved):
Looks and works good, great stuff! 💯

PR #3639, opened 2022-09-13
feat!: require api version when using client, freeze client config

rexxars:
### Description

**BREAKING CHANGE:** When using a Sanity client in custom code, the API version in use was unspecified and could theoretically change at any time if someone called the `config(newConfig)` method on the instance, or if the studio authors decided to use a more modern version.

Because we don't want peoples' code to unexpectedly break, we now require an API version to be specified when using `useClient()`, as well as in cases where a client would get passed through a context argument. In the case of contexts, there is now a `getClient()` property instead of a `client` property.

Note that these clients are reused based on the API version, so you are not creating new clients every time you ask for one. Because they are reused, we do not allow them to be reconfigured on the fly - calling `config(newConfig)` will now throw an error. If you need a new client with a different configuration, you can use the `withConfig(newConfig)` method, which clones the client it was called on and shallowly merges the new configuration. 

### What to review

- That `useClient({apiVersion: '2022-06-07'})` or equivalent works and returns the expected client
- That places previously using `client` in contexts (validation, config properties) now gets a `getClient()` instead

### Notes for release

- **BREAKING CHANGE**: When needing to use a Sanity client through validation and configuration contexts, you must now use `getClient({apiVersion: SOME_API_VERSION})` instead of accessing `context.client` - this is to ensure that we can easily upgrade the API version used throughout the studio without breaking existing code. Similarly, using the `useClient()` React hook in components now require you to specify an API version as well.


packages/@sanity/validation/src/validateDocument.ts
```diff
@@ -59,17 +59,44 @@ export default async function validateDocument(
...
+    value: doc,
+    path: [],
+    document: doc,
+    type: documentType,
+    getDocumentExists: context?.getDocumentExists,
+  }
+
+  // <TEMPORARY UGLY HACK TO PRINT DEPRECATION WARNINGS ON USE>
+  /* eslint-disable no-proto */
+  const wrappedClient = client as any
```
snorrees, line 77: Extract to a function to it easier to see the start and end of the hack?

packages/sanity/src/components/image/Image.tsx
```diff
@@ -42,7 +43,7 @@ export const Image = forwardRef(function Image(
   ref: React.ForwardedRef<HTMLImageElement>
 ) {
   const {dpr, fit, height, source, width, ...restProps} = props
-  const client = useClient()
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
```
snorrees, line 46: Did you consider `useDefaultStudioClient()`? Would reduce number of imports if nothing else.

packages/sanity/src/config/__tests__/resolveConfig.test.ts
```diff
@@ -144,15 +144,16 @@ describe('createWorkspaceFromConfig', () => {
     })
   })
 
-  it('allows overriding the `currentUser` and `client`', async () => {
+  it('allows overriding the `currentUser` and `getClient`', async () => {
```
snorrees, line 147: 👀 TIL, these are sneaky; not in the interface for config for good reason I presume 😈

packages/sanity/src/config/prepareConfig.ts
```diff
@@ -216,15 +217,51 @@ function resolveSource({
...
+    getClient,
     currentUser,
     dataset,
     projectId,
     schema,
   }
 
+  // <TEMPORARY UGLY HACK TO PRINT DEPRECATION WARNINGS ON USE>
+  /* eslint-disable no-proto */
+  const wrappedClient = client as any
```
snorrees, line 244: Possibly extract/reuse code. But its getting deleted soon so I guess it doesnt matter.

packages/sanity/src/config/useConfigContextFromSource.ts
```diff
@@ -9,10 +9,23 @@ import type {Source, ConfigContext} from './types'
...
-    return {projectId, dataset, schema, currentUser, client}
-  }, [projectId, dataset, schema, currentUser, client])
+    const client = getClient({apiVersion: '2021-06-07'})
+    const context = Object.defineProperty(
+      {projectId, dataset, schema, currentUser, getClient, client},
+      'client',
+      {
+        get() {
+          console.warn(
+            '`configContext.client` is deprecated and will be removed in the next version! Use `configContext.getClient({apiVersion: "2021-06-07"})` instead.'
```
snorrees, line 21: Possibly reuse error message from const. Again, doesnt matter.

snorrees (approved):
LGTM! 

Added some comments but they can be ignored 😅

packages/sanity/src/components/image/Image.tsx
rexxars: I did, but from a teaching perspective I find it clearer that we use the same hooks that other people would use. We can always refactor this later if it becomes annoying though.

packages/@sanity/validation/src/validateDocument.ts
rexxars: That would have been better. It is going away in the next release so will leave it as-is, but noted.

packages/sanity/src/config/__tests__/resolveConfig.test.ts
rexxars: `createWorkspaceFromConfig` should be considered internal - it is only for testing purposes that you'd want to do this :)

rexxars:
Merged and released in [v3.0.0-dev-preview.16](https://github.com/sanity-io/sanity/commit/7bbd903439fd70d71360f038cdd58c1f40d480f1)

PR #3641, opened 2022-09-14
Fix PT-input fullscreen issue

skogsmaskin:
### Description

This will fix an issue when closing modals from PT that contains a nested PT-editor when pressing `ESC`.

There was an attempt to always render the ExtendedLayer wrapped around the fullscreen version of the PT-input in order to optimize rendering. This seems not to actually give any particular performance wins though.

By only including the ExtendedLayer when in fullscreen fixes the issue described above as the `isTopLayer` get's the correct value now.

### What to review

That everything closes as expected using `ESC` on modals and fullscreen mode in the PT-input.

### Notes for release
* Fix an issue where Portable Text Input modals including a nested PT-input didn't close properly when pressing `ESC`.


hermanwikner (approved):
Nice! ✨

PR #3648, opened 2022-09-15
fix(base): tag bifur requests

sjelfull:
### Description

This tags Bifur requests with studio origin.

### What to review
- That the websocket is able to connect and handle messages as before

### Notes for release

n/a


sgulseth (approved):

rexxars (approved):
The code looks good, but this isn't working currently because the server doesn't accept the tag. Let's hold of on merging until it does.

PR #3649, opened 2022-09-15
Workaround weird Chrome bug related to flex layout

skogsmaskin:
### Description

There seem to be a bug in Chrome related to flex layout and spellchecking. By using a div with 'display: flex' here the spellcheck would use around 20% cpu idle. By using a Box with the same display value, the bug isn't introduced for some reason.

### What to review

That everything works as before and that the CPU idle use is gone.

### Notes for release
* Workaround Chrome bug related to spellcheck and flex layout.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
+++ b/packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
@@ -99,8 +99,12 @@ function textBlockStyle(props: TextBlockStyleProps & {theme: Theme}) {
 
 export const TextRoot = styled.div<TextBlockStyleProps>(textBlockStyle)
 
-export const TextBlockFlexWrapper = styled(Flex)`
+// Because of a weird bug in Google Chrome regarding the @sanity/ui Flex component and spellchecking,
+// this is set to be a Box with 'display: flex'. Using the Flex component here results in Chrome
+// using 20% CPU when idle when spellchecking is on for some reason.
+export const TextBlockFlexWrapper = styled(Box)`
   position: relative;
+  display: flex;
 `
 
 export const ListPrefixWrapper = styled.div`


packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
```diff
@@ -99,8 +99,9 @@ function textBlockStyle(props: TextBlockStyleProps & {theme: Theme}) {
 
 export const TextRoot = styled.div<TextBlockStyleProps>(textBlockStyle)
 
-export const TextBlockFlexWrapper = styled(Flex)`
+export const TextBlockFlexWrapper = styled(Box)`
```
bjoerge, line 102: This piece of code looks so much like an oversight to me now ("why is it a box when it clearly should have been a flex?"). Would be great to add a comment here so it doesn't accidentally gets changed back to Flex. Otherwise good!

bjoerge (changes requested):

packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
```diff
@@ -99,8 +99,9 @@ function textBlockStyle(props: TextBlockStyleProps & {theme: Theme}) {
...
   position: relative;
+  display: flex;
```
hermanwikner, line 104: This CSS will not be applied to the `Box` – so right now `TextBlockFlexWrapper` is a `Box` with `position: relative`. And maybe that's the fix? 🙂

But if you want `display: flex` on a `Box` you need to either:

- Use the `display` prop: `<Box display="flex"/>`
- Override display like this:

```
const FlexBox = styled(Box)`
  &:not([hidden]) {
    display: flex;
  }

```

packages/@sanity/form-builder/src/inputs/PortableText/text/TextBlock.styles.ts
skogsmaskin: Added comment about this now.

bjoerge (approved):

PR #3652, opened 2022-09-15
feat(default-layout): sortable search results, search phrases, experimental config

robinpyon:
image

# Description

### Features

- Order search results by relevance, created and last updated date.
- Search for documents containing an exact phrase by wrapping text inside double quotes.
- Add `__experimental_omnisearch_visibility: true` to your document schemas to hide them from omnisearch. Perfect for hiding workflow-related documents or configuration singletons that don’t need to be accessed frequently.
- View relevance scores for search results via a debug flag (`#_debug_search_score`)

### Bug fixes

- Fixes an issue where the scroll position of search results wasn’t being correctly retained in some instances.
- Fixes an issue where search would cause the studio to crash if Local Storage is unavailable.

# What to review

## Notable changes

1. Taking cues from the reference and cross dataset reference input components – the function to generate weighted search is now colocated within `default-layout` rather than being an export of `@sanity/base/search`
    - Similar to reference / cross reference input components, this now imports from `@sanity/base/_internal`
    - ~~Given that  `part:@sanity/base/search` and `part:@sanity/base/search/weighted` are not being used anywhere else in the studio, they’ve been removed~~
2. Logic for how we extract search queries and calculate scores has been expanded to accommodate searching for phrases in wrapped quotes:
    - Fundamentally, this is handled in a very similar way to how search worked prior. My understanding of search, and how phrases come into play can be found [here](https://www.notion.so/How-studio-search-roughly-works-a511a6a254884fa98aa41c00c5fe5f67) (internal doc)
    - The important thing to note is that how we handle non-quoted text searches hasn’t changed. **If you don’t engage with this feature then you won’t notice a difference here.**
3. A few more comments are now attached with each search query for measurement. Full example:
    
    ```jsx
    // findability-mvi:2
    // findability-recent-search:4
    // findability-selected-types:2
    // findability-sort:relevance
    ```
    
    In the above example, `findability-recent-search:4` indicates that this was the 5th item clicked in recent searches. All these comments will be removed once client-side instrumentation is in place.
    
4. In the test studio: a debug schema `experimentalOmnisearchVisibilityTest` has been added. These documents have `__experimental_omnisearch_visibility:false` and won’t appear in omnisearch.
5. Sorting by relevance will now sort by `_updatedAt desc` in GROQ before applying scoring on the client, leaning towards more recent documents in general. Previously, omnisearch was sorting on `_id asc`.
    - Search within reference + cross dataset reference fields are unchanged, and will continue to sort on `_id asc`.
6. Notable refactoring: 
    - Simplified `createWeightedSearch` signature slightly, and have moved more into the `SearchOptions` type
    - Some unused dev dependencies were removed from `default-layout` (`mocha`, `nyc`) and it now uses `react-testing-library` .
    - `default-layout` was also added to the root level `jest.config.js`
    - Search related tests in `@sanity/base` are now colocated and have been expanded on slightly.
    - Quite a bit of code was touched to ensure this is valid once we can enable `strictNullChecks: true`.

## Things to test

1. **Change the sorting of search results:**
    - Does this work and make sense to you? Bearing in mind scoring isn’t taken into consideration when not sorting on relevance.
2. **Search on phrases with quoted text:**
    - View relevance scoring in debug mode – do these scores seem reasonable to you?
    - Try search for strings with special characters – some good candidates to compare and contrast (test studio):
        1. `"winnie-the-pooh"` vs `winnie-the-pooh`
        2. `"go, team. go!"` vs `go, team. go!`

# Notes for release

### Search updates

- Order search results by relevance, created and last updated date.
- Search for documents containing an exact phrase by wrapping text inside double quotes.
- Hide documents from global search by specifying `__experimental_omnisearch_visibility: true`.

## Bugfixes

- Fixes an issue where the scroll position in global search wasn’t being correctly retained in some instances.
- Fixes an issue in global search that would cause the studio if Local Storage is unavailable.


packages/@sanity/default-layout/src/navbar/search/contexts/commandList/index.tsx
```diff
@@ -202,6 +202,34 @@ export function CommandListProvider({
     []
   )
 
+  const scrollToNextItem = useCallback(() => {
```
snorrees, line 205: Might be overkill/require passing to many variables, but is there a useScrollTo('next' | 'previous') abstraction hiding here that would eliminate all duplication execpt selecting +1 | 1 for moving the index?

packages/@sanity/default-layout/src/navbar/search/contexts/commandList/index.tsx
```diff
@@ -235,59 +263,62 @@ export function CommandListProvider({
...
-    virtualList,
-  ])
+  }, [childContainerElement, headerInputElement, scrollToNextItem, scrollToPreviousItem])
+
+  /**
+   * Listen to keyboard arrow events on the 'closest' parent [data-overflow] element to the child container.
+   * On arrow press: focus the header input element and then navigate accordingly.
+   *
+   * Done to account for when users focus a wrapping element with overflow (by dragging its scroll handle)
+   * and then try navigate with the keyboard.
```
snorrees, line 300: Oh the edge cases 🙈

packages/@sanity/default-layout/src/navbar/search/contexts/search/index.tsx
```diff
@@ -76,25 +83,40 @@ export function SearchProvider({children, currentUser}: SearchProviderProps) {
...
+        ordering?.customMeasurementLabel || `${ordering.sort.field} ${ordering.sort.direction}`
 
-    if (termsChanged || pageIndexChanged) {
       handleSearch({
-        ...terms,
-        limit: SEARCH_LIMIT,
-        offset: pageIndex * SEARCH_LIMIT,
+        options: {
+          // Comments prepended to each query for future measurement
+          comments: [
```
snorrees, line 98: Consider extracting the comments constructuin and adding some unit tests. Since its invisible to the user, we might not pick up if we break it (for instant if isRecentSearchTerms is changed without thinking about this usage).

packages/@sanity/default-layout/src/navbar/search/contexts/search/reducer.ts
```diff
@@ -218,12 +240,31 @@ export function searchReducer(state: SearchReducerState, action: SearchAction):
...
+ *
+ * This is done so we can better disambiguate between requests sent as a result of clicking a 'recent search'
+ * for purposes of measurement. __index is used to measure exactly which recent search item was clicked.
+ *
+ * @todo remove this (and associated tests) once client-side instrumentation is available
+ */
+function stripRecentIndexAndTimestamp(terms: RecentSearchTerms | SearchTerms) {
+  if (isRecentSearchTerms(terms)) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const {__index, __recentTimestamp, ...rest} = terms
```
snorrees, line 266: I think there is a case to be made for putting these in a wrapper object, so we only need to check for one property. If more of these get added in the future, it is likely that this callsite gets forgotten about.

packages/@sanity/default-layout/src/navbar/search/datastores/recentSearches.ts
```diff
@@ -1,14 +1,17 @@
...
 
 const SEARCH_TERMS_KEY = 'search-terms::recent'
 export const MAX_RECENT_SEARCHES = 5
 // might come in handy in the future
 const CURRENT_VERSION = 1
 
-export type RecentSearchTerms = SearchTerms & {__recentTimestamp: number}
+export type RecentSearchTerms = SearchTerms & {
+  __index: number
+  __recentTimestamp: number
```
snorrees, line 13: Consider wrapping in __recent or something (see previous comment)

packages/@sanity/default-layout/src/navbar/search/hooks/useSearch.ts
```diff
@@ -17,32 +18,46 @@ import {
...
-    query: terms.query.trim(),
+    ...request,
+    terms: {
+      ...request.terms,
+      query: request.terms.query.trim(),
+    },
   }
 }
 
+const searchWeighted = createWeightedSearch(getSearchableOmnisearchTypes(schema), searchClient, {
```
snorrees, line 56: This line wont port to v3 since its using the global schema. In v3 this must be done inside useSearch and the useSchema hook. Possibly think about how it can be memoized there now.

packages/@sanity/default-layout/src/navbar/search/utils/isRecentSearchTerms.ts
```diff
@@ -0,0 +1,12 @@
+import type {SearchTerms} from '@sanity/base'
+import type {RecentSearchTerms} from '../datastores/recentSearches'
+
+export function isRecentSearchTerms(
+  terms: SearchTerms | RecentSearchTerms
+): terms is RecentSearchTerms {
+  const recentSearchTerms = terms as RecentSearchTerms
+  return (
+    typeof recentSearchTerms.__recentTimestamp !== 'undefined' &&
```
snorrees, line 9: Another datapoint that wrapper object could be in order; this is another callsite that must change if RecentSearchTerms gets more __ props

snorrees (approved):
Looks good, thorough and well explained as always :D 

Just found some minor nitpick and open questions for myself.

Testing:
* Seems to work like I would expect
* As I mentioned elsewhere, the bug where preview does lags behind document updates confused my testing a bit. But this is not at all related to this code, and will be fixed when that issue is adressed. I also dont think preview not updating in search result is really a big problem as you typically wont be searching for already opened documents.
* Scores: Unless you know the underlying algorithm, these will probably feel a bit random to people. Title * 10, subtitle * 2, partial match / 2, ect ect. Sorting wise it feels good though!

packages/@sanity/default-layout/src/navbar/search/components/SearchHeader.tsx
```diff
@@ -28,8 +32,8 @@ const SearchHeaderCard = styled(Card)`
 
 const NotificationBadge = styled.div`
   align-items: center;
-  background: ${({theme}) => theme.sanity.color.selectable.primary.enabled.fg};
-  color: ${({theme}) => theme.sanity.color.selectable.primary.selected.fg};
+  background: ${({theme}) => theme?.sanity?.color?.selectable?.primary?.enabled?.fg};
```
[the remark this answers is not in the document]
robinpyon: This was me being lazy – these need to typed to use `@sanity/ui`'s `Theme` e.g.

```javascript
import type {Theme} from '@sanity/ui'
...
background: ${({theme}: {theme: Theme}) => theme.sanity.color.selectable?.primary.enabled.fg};
```

packages/@sanity/default-layout/src/navbar/search/components/SortMenu.tsx
```diff
@@ -0,0 +1,146 @@
...
+  Card,
+  Flex,
+  Inline,
+  Menu,
+  MenuButton,
+  MenuDivider,
+  MenuItem,
+  Text,
+} from '@sanity/ui'
+import {isEqual} from 'lodash'
```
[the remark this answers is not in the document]
robinpyon: I haven't applied this in [applyWeights](https://github.com/sanity-io/sanity/blob/6bdbafe627d0a2a38223be9be60abe27e5992ebd/packages/%40sanity/base/src/search/weighted/applyWeights.ts) (which uses _a lot_ of lodash), but agree with the sentiment and practice!

robinpyon:
Thanks @snorrees!

**Re: scores**: I agree! Especially with how search will use preview fields (which can be automatically inferred if you don't have a preview configuration) .... mix in the interplay between `__experimental_search` and it can all seem a bit much. It begs the question of whether this needs to be surfaced in documentation more.

packages/@sanity/default-layout/src/navbar/search/contexts/search/index.tsx
robinpyon: Would love a bit more direction here on specific tests you'd expect!

Right now, `comments` in search options has type safety, and there's [a test for createSearchQuery](https://github.com/sanity-io/sanity/blob/f698c924f9c26783e751bccc4f78e4d44171a273/packages/%40sanity/base/src/search/weighted/createSearchQuery.test.ts#L189-L203) that (loosely) covers prepended comments when creating queries

packages/@sanity/base/sanity.json
```diff
@@ -50,14 +50,6 @@
...
       "path": "assets/asset-url-builder"
     },
-    {
-      "name": "part:@sanity/base/search",
-      "description": "Parse the search query and fetch search results from the dataset"
-    },
-    {
-      "name": "part:@sanity/base/search/weighted",
-      "description": "Weighted search implementation"
-    },
```
rexxars, line 60: Technically speaking, this is a breaking change. I'm not sure if anyone is likely to have actually used these though - but there is a reference to them from this shared content demo, for instance: https://github.com/runeb/shared-content-demo/blob/main/schemas/components/ExternalReference.js

Would it be difficult to keep around a copy of the existing parts as-is? Given v3 is around the corner we'll be removing them then anyway, and it would be nice to to cause any breakage until then.

packages/@sanity/default-layout/src/navbar/search/components/SearchResults.tsx
```diff
@@ -74,57 +81,68 @@ export function SearchResults({
...
     onChildClick?.()
     onClose()
   }, [dispatch, onChildClick, onClose, recentSearchesStore, terms])
 
   return (
-    <SearchResultsDiv aria-busy={result.loading} $loading={result.loading}>
-      <PointerOverlay ref={setPointerOverlayRef} />
+    <SearchResultsFlex direction="column">
+      {/* Sort menu */}
+      {!!result.hits.length && <SortMenu small={small} />}
```
rexxars, line 105: Slightly subjective I suppose, but I prefer to assign these sort of checks to variables such as `hasSearchResults` to be more readable, or write them as `result.hints.length > 0`.

`!!0` works fine, but I find it takes slightly more brainpower to do the "integer to boolean, then negate it" conversion in my head 🙂

rexxars (approved):
This looks great. I do worry about potentially breaking someone's code by removing the search parts, however.

packages/@sanity/base/sanity.json
robinpyon: Great pick up – and definitely not difficult to keep! That part removal has since been reverted and that original exported weighted search function can still be used as-is.

PR #3656, opened 2022-09-16
feat(cli): add `--no-open` flag to login command

rexxars:
### Description

Adds a new `--no-open` flag (technically also `--open`, but that is the default behavior) to the `sanity login` command, which is useful in certain situations such as automation. Also documents the `--provider <id>` option, which is also useful for automation.

### What to review

- That `sanity login` still opens a browser by default
- That `sanity login --no-open` does not open browser, but prints login URL to screen

### Notes for release

- Added a new `--no-open` flag to the `sanity login` command in order to suppress the opening of browser automatically


RitaDias (approved):
🛳

vicbergquist (approved):
LGTM!

PR #3657, opened 2022-09-16
feat(cli): add `--force` flag to `sanity dataset delete` command

rexxars:
### Description

Adds a new `--force` flag to the dataset delete command, in order to aid automation.

The reparsing of the CLI args is done to work around an issue where `sanity dataset delete --force foo` would treat the `force` flag as a string flag instead of a boolean flag and fail.

### What to review

- That `sanity dataset delete my-dataset` still works and prompts for the name
- That `sanity dataset delete my-dataset --force` skips the prompt and deletes

### Notes for release

- Added a new `--force` flag to the `sanity dataset delete` command


RitaDias (commented):
Maybe I'm doing something wrong, but it looks like the `--force` flag is not working? 🤔 
image

RitaDias (approved):
🛳

PR #3659, opened 2022-09-16
feat(core): add `--force` flag to dataset alias delete command

rexxars:
### Description

This is getting sorta repetitive, ain't it? 😅 Apologies, should probably have bundled all of these together.

### What to review

- That the alias delete command works, with or without the flag

### Notes for release

- Added `--force` flag to `sanity dataset alias delete` command, in order to skip the confirmation prompt


RitaDias:
I see that in other prs you have helper text (https://github.com/sanity-io/sanity/pull/3657/files#diff-47801728e2ce66f1bdb09d666be0df0dbfc1d6470fce4c988499145591439305R7) to show how to use the flag (which I struggled with :') ) but this pr doesn't have it.

I had expected it to appear in the --help like we have for the unlink
image

RitaDias (approved):
Added a small comment, but otherwise 🚢

PR #3662, opened 2022-09-19
Throttle onChange in text inputs

skogsmaskin:
### Description

This will introduce a local editing state on String and Text inputs and then submits the produced patches throttled (1000 ms throttle time, non-leading).

This makes the performance and writing experience in these fields a lot better.

### Notes for release
* Better typing performance in String and Text inputs.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/StringInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/StringInput.tsx
@@ -1,8 +1,9 @@
-import React, {useMemo} from 'react'
+import React, {useCallback, useEffect, useMemo, useState} from 'react'
 import {useId} from '@reach/auto-id'
 import {StringSchemaType} from '@sanity/types'
-import {TextInput} from '@sanity/ui'
+import {TextInput, useForwardedRef} from '@sanity/ui'
 import {FormField} from '@sanity/base/components'
+import {throttle} from 'lodash'
 import PatchEvent, {set, unset} from '../PatchEvent'
 import {Props} from './types'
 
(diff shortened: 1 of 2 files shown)


packages/@sanity/form-builder/src/inputs/StringInput.tsx
```diff
@@ -19,29 +20,60 @@ const StringInput = React.forwardRef(function StringInput(
     [markers]
   )
 
-  const handleChange = React.useCallback(
+  const [localValue, setlocalValue] = useState(null)
```
bjoerge, line 23: nit: `setlocalValue` => `setLocalValue`

packages/@sanity/form-builder/src/inputs/StringInput.tsx
```diff
@@ -19,29 +20,60 @@ const StringInput = React.forwardRef(function StringInput(
...
       const inputValue = event.currentTarget.value
-      onChange(PatchEvent.from(inputValue ? set(inputValue) : unset()))
+      setlocalValue(inputValue)
+      throttledSubmit()
     },
-    [onChange]
+    [throttledSubmit]
   )
 
+  useEffect(() => () => throttledSubmit(), [throttledSubmit])
```
bjoerge, line 49: should be `throttledSubmit.flush()` here I think?

packages/@sanity/form-builder/src/inputs/TextInput.tsx
```diff
@@ -26,13 +27,34 @@ const TextInput = React.forwardRef(function TextInput(
     [markers]
   )
 
-  const handleChange = React.useCallback(
-    (event) => {
+  const [localValue, setlocalValue] = useState(null)
```
bjoerge, line 30: nit: `setlocalValue` => `setLocalValue`

packages/@sanity/form-builder/src/inputs/TextInput.tsx
```diff
@@ -26,13 +27,34 @@ const TextInput = React.forwardRef(function TextInput(
...
       const inputValue = event.currentTarget.value
-      onChange(PatchEvent.from(inputValue ? set(inputValue) : unset()))
+      setlocalValue(inputValue)
+      throttledSubmit()
     },
-    [onChange]
+    [throttledSubmit]
   )
+
+  useEffect(() => () => throttledSubmit(), [throttledSubmit])
```
bjoerge, line 56: Same as above, should be `throttledSubmit.flush()` here as well I believe

bjoerge (changes requested):

bjoerge:
Seems to work fine! Just a couple of nitpicks, and a comment about calling `.flush()` that would be good if you could look into :)

packages/@sanity/form-builder/src/inputs/StringInput.tsx
skogsmaskin: Yes! I was actually thinking about this in the weekend. My idea was just to call onChange there directly.

packages/@sanity/form-builder/src/inputs/TextInput.tsx
skogsmaskin: Yeah

PR #3668, opened 2022-09-20
Optimize validation flow/scheduling

bjoerge:
### Description
This PR optimizes how validation happens in the studio.

Currently, validation is being run as a debounced function of "document changes" (both local + remote). Once a certain time has passed since _last_ change we start a new validation pass. However, once a validation pass has started, there's no way to cancel it, and even though we ignore the result of any previous (and no longer relevant) validation passes, they still run to completion, taking up unnecessary resources. Validation is a pretty CPU heavy task since the whole document has to be traversed and validator functions has to be re-computed for every node. In particular on large documents this consumed a significant amount of resources, sometimes leading to UI freeze and bad editor experience.

This PR is twofold:

#### 1. Refactor coordination and scheduling of document validation

Even though our current debounced validation described above allows the editor to type continiusly for a while without validation getting in the way, it is still disruptive if the editor stops to think for a little while and then resumes typing when validation is ongoing. Another drawback of using debounce here is that you actually *have* to stop typing in order to get the new validation status. Now that validation happens in the background (see below), we can improve this situation further. This PR introduces a better flow using throttling instead of debouncing. This means we now can kick off a new validation pass almost immediately after first document change. For example, when you start writing in a required field, we'll now start a new validation pass almost immediately, and you'll see the field go from invalid to valid. As you keep typing, we will keep running validation in the background, but never more than a single validation pass at a time, and always a validation pass after the document is no longer being edited.

#### 2. Refactor `@sanity/validation` to use observables under the hood

Although not *strictly* neccesary given the way we still run validation passes to completion, this gives us the ability to cancelling ongoing validation passes, and gives us more flexibility for possible future optimizations. It also makes it easier to control things like how many items we validate in parallel. The `@sanity/validation` package now has an additional export named `validateDocumentObservable()`. The existing promise-based functions are now built on top of the observable counterparts, exports are still left intact, so no breaking changes.

In addition, validation is now considered a "background task". This means that validation of each item will now run in an idle callback, allowing the browser to prioritize UI updates and only run validation when there's available CPU.

### What to review

Make sure validation works as before (I've done quite a bit of testing, and it seems fine, fwiw).

#### Questions:
- Should we force through validation with a idle callback deadline if it takes too long? This will likely lead to UI freeze, and if we don't do this, editors on low-end devices may have to wait for a long time after typing before the validation errors appears.

### Notes for release

- Optimize execution and scheduling of document validation


skogsmaskin (approved):
Awesome work @bjoerge, thank you for this!

I have been testing this in Chrome and Safari (with the shim) and it seems to work very well 👍 

Codewise I have looked through, and as far as I can tell it looks great!

rexxars (approved):
Great work!

PR #3669, opened 2022-09-20
perf(desk-tool): optimize useTimeAgo

bjoerge:
### Description
Our `useTimeago` hook currently uses a combo of `useEffect` and `setState` to keep the value. This lead to a cases where we would re-render the hook twice (render => args change => useEffect => setState => render). Depending on the usage context, re-renders might be expensive.

I also changed from setInterval to a setTimeout continuation to avoid pile up (if the next render takes long, we end up waiting less than the actual desired delay, which, in principle (but probably only in extreme cases) could lead to a pile-up of invocations.

Also, to my surprise, passing `Infinity` as delay argument to `setInterval/setTimeout` is equivalent with a delay of `0`) and invokes it immediately, so I've changed the logic here slightly.

This PR refactors the `useTimeAgo` hook to address the above issues.

### What to review

Verify that the hook works like it used to :)

### Notes for release
N/A, not visible for end-users

### DIFF
--- a/packages/@sanity/base/src/time/useTimeAgo.ts
+++ b/packages/@sanity/base/src/time/useTimeAgo.ts
@@ -1,19 +1,19 @@
-import {useEffect, useState} from 'react'
+import {useEffect, useReducer} from 'react'
 import {
-  format,
-  differenceInSeconds,
-  differenceInMinutes,
-  differenceInHours,
   differenceInDays,
-  differenceInWeeks,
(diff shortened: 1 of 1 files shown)


bjoerge:
> We were checking Number.isFinite() so this shouldn't happen, but I guess this reads more clearly anyway 👍

Ah, I totally missed that, my bad! Thanks for reviewing!

PR #3670, opened 2022-09-20
refactor(form-builder): simplify cross dataset reference config

sjelfull:
### Description

This simplifies cross dataset references and makes it ready for GA, by:
- Removing support for cross project references and cross project tokens
- Derives the fields to search in from the preview config, instead of `__experimental_search`

Also touches [sc-23420]

### What to review

- That the cross dataset input search will search in the fields set in `preview` on each `type.to`
- That the logic that set the project ID from the client instead of from the schema makes sense
- That tests runs as expected


bjoerge (approved):
Great work! So good to get rid of the complexity here :) Would be good to remove the redundant combineLatest in useReferringDocuments berfore merge.

bjoerge (approved):
Fantastic work! Just found a few nitpicks, nothing that blocks merge.

rexxars (approved):

PR #3671, opened 2022-09-20
fix(vision): allow reconfiguration of vision client

rexxars:
### Description

Mistakenly forgot that the client passed to vision from the studio now has a frozen config, but we're attempting to mutate it when changing API versions. Since we're using `withConfig()` (which clones and applies a new config) on the passed client, it is okay to un-freeze the config of this _particular_ instance of the client.

### What to review

- That vision works inside the studio when switching between API versions

### Notes for release

- Fixes an issue where the vision plugin would fail to run queries


rexxars:
Fixes https://github.com/sanity-io/sanity/issues/3673

PR #3672, opened 2022-09-21
feat(form-builder): add withParent higher order component

bjoerge:
### Description

Looks like we started passing the `parent` value to all inputs in 2.23.0 (https://github.com/sanity-io/sanity/commit/945bb0fb768008f743ad09e1f5744c77de64bb7c). This seems like an unintentional change to me, and is also something that likely have a negative impact on perf.

AFAIK, the parent-prop has never been documented and we should try to replace it with something more performant. In order to unblock users that already got logic tied to it, I've made a `withParent`-higher order component that will make the parent value available to the custom input component.

Usage: 
```js
const MyCustomInput = withParent(
  forwardRef(function MyInputWithParent(props, ref) {
    
    // parent value is now available in props.parent
    console.log(props.parent)

    return (
      // …
    )
  })
)
```

### What to review
There's a debug/test input added that demonstrates the usage included in this PR. Can be tested [here](https://test-studio-git-bug-sc-24761parent-showing-as-undefined.sanity.build/test/desk/input-debug;withParentTest).

### Notes for release
- Replacing internal `parent` prop passed to input components with a `withParent` higher order component.


bjoerge:
(The lint errors isn't related to these changes, fwiw)

RitaDias (commented):
Should we be documenting this component somewhere? 🤔

bjoerge:
We will most likely replace it with something better in v3, and it's a stop-gap measure to provide a workaround to those who already have tied logic to the (undocumented) parent-property their input receives, so think it's actually ok to leave this one too as undocumented, at least for now.

sjelfull (approved):

PR #3687, opened 2022-09-27
Fix Spinner misalignments

skogsmaskin:
### Description

This will upgrade `@sanity/ui` to ` 0.37.22` and take care of some Spinner issues.

With the current performance optimization  done in `@sanity/ui` for Spinner animations, some spinners in the Studio code aligned wrong when animated. This was due to them being wrapped in a Text node.

I have updated all occurrences of this (I think) by implementing them according to the updated documentation at https://www.sanity.io/ui/docs/primitive/spinner#examples

For the ProjectUser dashboard widget I have used the Spinner Sanity-part to better align them with plugin code elsewhere.

For the Seach bar spinner I have used the spinner icon from `@sanity/icons` as the TextInput from `@sanity/ui` is expecting an icon and not a component. Then put on the same animation as `@sanity/ui` uses for its spinner.

### What to review

That all spinners inside the studio looks good.

### Notes for release

Doesn't need to be mentioned as this problem doesn't affect any production code and needed to be done in order to upgrade `@sanity/ui`.

### DIFF
--- a/dev/design-studio/package.json
+++ b/dev/design-studio/package.json
@@ -31,7 +31,7 @@
     "@sanity/logos": "^1.1.15",
     "@sanity/structure": "2.33.2",
     "@sanity/types": "2.33.2",
-    "@sanity/ui": "^0.37.21",
+    "@sanity/ui": "^0.37.22",
     "@sanity/vision": "2.33.3",
     "prop-types": "^15.6.0",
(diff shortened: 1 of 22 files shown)


mariuslundgard (approved):
Looks good to me 👍

PR #3692, opened 2022-09-28
fix(cli): use same process when executing scripts

rexxars:
### Description

The new bundling strategy broke `sanity exec` and only by accident works with GraphQL schema generation. This PR changes the exec script from spawning a new node process to instead doing everything in-process. It's not great, but I am not familiar enough with the new pkg-utils bundling strategy to come up with a good alternative at the moment.

### What to review

- `sanity exec` works as expected
- `sanity graphql deploy --dry-run` works as expected

### Notes for release

None. Transparent to the end user, as the bundling changes have yet to be released.


packages/sanity/src/cli/actions/exec/configClient.ts
```diff
@@ -1,5 +1,6 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
mariuslundgard, line 1: (disabling `import/no-extraneous-dependencies` will not be necessary soon)

packages/sanity/src/cli/actions/exec/configClient.ts
```diff
@@ -1,5 +1,6 @@
...
+import {getCliClient} from 'sanity/cli'
```
mariuslundgard, line 2: Can this be imported from `@sanity/cli` directly?

packages/sanity/src/cli/actions/exec/execScript.ts
```diff
@@ -36,23 +34,17 @@ const execScript: CliCommandAction<ExecFlags> = async function execScript(args,
...
+    await import('./registerBrowserEnv')
+  } else {
+    await import('./esbuild')
+  }
+
+  if (withUserToken) {
+    await import('./configClient')
+  }
+
+  await import(scriptPath)
```
mariuslundgard, line 47: I think these changes make a lot of sense – using dynamic imports to split up the runtime. But does this mean we are no longer running these processes in child threads?

rexxars: > How do you feel about not running these in child processes any more? We _could_ solve that by adding entry points for each of the child process scripts.

Don't feel great about it - so your suggested entry point approach looks great to me! I'll take a stab at it.

packages/sanity/src/cli/actions/exec/configClient.ts
rexxars: Yep, good catch.

PR #3698, opened 2022-09-30
test(cli): provide full cli tests for both v2 and v3

rexxars:
### Description

When we release v3, we need to ensure that the CLI tool still works with both v2 and v3 so that people can easily work on projects on both versions. This PR introduces CLI tests that tests all CLI commands on both a v2 and v3 studio.

There are some minor changes to the GraphQL commands in order to make it easier to test (and also provides a bit more functionality as a bonus).

### What to review

That the tests appear correct. They won't run locally unless a token is set up, but it is configured on GitHub actions.

### Notes for release

None.


mariuslundgard:
Took the liberty of rebasing this on `v3`.

mariuslundgard:
Hm, seems the GitHub action `test` just outputs `No SANITY_CI_CLI_AUTH_TOKEN set, skipping CLI tests`. Should this env variable exist?

mariuslundgard:
Also – seems none of the other tests are running (locally).

rexxars:
Running the tests in CI uncovered a few issues - I've squashed away the details, but should be good to re-review now.

It's not ideal that we need a token to run the tests - ideally we'd be able to at least run _some_ of the tests without a user or with our own CLI user, but many of the tests would require project membership. I suppose we could check whether or not the local CLI user has access to the project before running the tests, for instance. Will leave this for another iteration - at least now they'll run and uncover issues in CI.

packages/@sanity/cli/test/exportImport.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        'production',
+        testRunArgs.exportTarball,
+        '--overwrite',
+      ])
+      expect(result.stdout).toMatch(/export finished/i)
+      expect(result.code).toBe(0)
+
+      const stats = await stat(path.join(studiosPath, version, testRunArgs.exportTarball))
+      expect(stats.isFile()).toBe(true)
+      expect(stats.size).toBeGreaterThanOrEqual(5000)
```
mariuslundgard, line 28: What is `5000` here? Is this a threshold we can assume is accurate? Would it add clarity to give a it name, e.g. `const TARBALL_MIN_TYPICAL_SIZE = 5000` with comment?

packages/@sanity/cli/jest.config.cjs
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = createJestConfig({
+  displayName: require('./package.json').name,
+  testEnvironment: 'node',
+  rootDir: __dirname,
+  globalSetup: '<rootDir>/test/shared/globalSetup.ts',
+  globalTeardown: '<rootDir>/test/shared/globalTeardown.ts',
+  slowTestThreshold: 60000,
+  testTimeout: 30000,
+  setupFilesAfterEnv: ['<rootDir>/test/shared/setupAfterEnv.ts'],
```
mariuslundgard, line 13: Nitpick – personally I'm a fan of alphabetizing lists like this. It's probably just ocd 😅

packages/@sanity/cli/test/alias.test.ts
```diff
@@ -0,0 +1,96 @@
+import {describeCliTest, testConcurrent} from './shared/describe'
+import {getTestRunArgs, runSanityCmdCommand, studioVersions} from './shared/environment'
+
+describeCliTest('CLI: `sanity dataset alias`', () => {
+  describe.each(studioVersions)('%s', (version) => {
+    jest.setTimeout(30 * 1000)
```
mariuslundgard, line 6: Is this the way of defining timeout which is buggy in Jest?

packages/@sanity/cli/test/graphql.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    )
+    const client = testClient.withConfig({dataset: graphqlDataset})
+
+    testConcurrent('graphql deploy', async () => {
+      await client.createOrReplace(
+        {
+          _id: 'person_james-cameron',
+          _type: 'person',
+          name: 'James Cameron',
+        },
```
mariuslundgard, line 20: I enjoy seeing this kind of test data 😌

packages/@sanity/cli/test/alias.test.ts
rexxars: No, it's the `testTimeout` set in `jest.config.cjs` which isn't working when using jest projects - using `jest.setTimeout` still works, but is tedious when you have to define it for every single test. This one can be removed since it is the default, however - will fix.

packages/@sanity/cli/test/shared/devServer.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      timer = setTimeout(tryConnect, 1000)
+    }
+
+    async function tryConnect() {
+      let res: ResponseData
+      try {
+        res = await Promise.race([
+          request(`http://localhost:${port}/`),
+          new Promise<ResponseData>((_, rejectTimeout) => setTimeout(rejectTimeout, 500)),
```
mariuslundgard, line 65: Should `rejectTimeout` be called with an `Error` instance here?

packages/@sanity/cli/test/shared/globalSetup.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+async function prepareCliInstall(): Promise<void> {
+  // Prepare a folder for the CLI to be installed into
+  await mkdir(cliInstallPath, {recursive: true})
+  spawnSync(npmPath, ['init', '-y'], {cwd: cliInstallPath})
+}
+
+async function packCli(): Promise<string> {
+  // Run `npm pack` so we can create a fully isolated install, replicating what a user would get
```
mariuslundgard, line 157: I think this is a good idea – since this means we test what will actually be installed in `node_modules` out there.

packages/@sanity/cli/tsconfig.lib.json
```diff
@@ -6,7 +6,8 @@
     "./src/**/__mocks__",
     "./src/**/__workshop__",
     "./src/**/*.test.ts",
-    "./src/**/*.test.tsx"
+    "./src/**/*.test.tsx",
+    "./test/**/__fixtures__"
```
mariuslundgard, line 10: `"./test/**/__fixtures__"` is not needed here since `"./test"` is not included.

packages/@sanity/cli/test/alias.test.ts
mariuslundgard: OK, I see 🙏

mariuslundgard (approved):
Looks super solid to me. Just have minor comments that you can feel free to disregard. Great work!

packages/@sanity/cli/test/shared/devServer.ts
rexxars: I suppose. We don't use the error for anything - just schedule another connection attempt - but looks a little cleaner with an error I suppose.

packages/@sanity/cli/test/exportImport.test.ts
rexxars: Just me being lazy. I've replaced it with a proper tarball contents check (for an ndjson file + asset) - the functionality itself is provided by `@sanity/export`, so will leave the "proper" tests for that module.

packages/@sanity/cli/jest.config.cjs
rexxars: :) Fixed

PR #3700, opened 2022-09-30
Fix React 18 issue with PT-input

skogsmaskin:
Is seems like the PortableTextInput with React 18 is able to receive events from the PortableTextEditor even though it is unmounted and cleaned up at the time we receive the event inside PortableTextInput.

By checking the editorRef there before sending the patches, we can avoid that DocumentPane crashes at least when it receives the mutation.

I am however not sure if this is symtom of that something else is wrong, or it is just some of the things to expect with React 18.


stipsan (approved):
Maybe it scheduled it for unmounting but doesn't actually unmount right away. I'd have to run a profiling session on it to know for sure 🤔

Anyways, great work! This unblocks us from doing a new `dev-preview` release 😌

PR #3710, opened 2022-10-03
Feat/v3/slug context

snorrees:
### Description

Extends the context object passed to `options.source` and `options.slugify` with getClient, schema, currentUser (and projectId and dataset for consistency with initial value context).

Also fixes the context type for `options.isUnique` which was duplicated in the definition, and missing some properties (they where available at runtime already though, via validation context).

This is primarily to make it possible to use the client when using these options; a common thing in v2.

### What to review

All the code.

### Notes for release

Slug options `source` and `slugify` are now passed a `context` parameter as the last argument. It contains `getClient`, `schema` and `currentUser`.


snorrees:
Something wrong with this PR, recreating.

PR #3711, opened 2022-10-03
fix: added context with client and schema to slugify and source options

snorrees:
### Description

Extends the context object passed to `options.source` and `options.slugify` with getClient, schema, currentUser (and projectId and dataset for consistency with initial value context).

Also fixes the context type for `options.isUnique` which was duplicated in the definition, and missing some properties (they where available at runtime already though, via validation context).

This is primarily to make it possible to use the client when using these options; a common thing in v2.

### What to review

All the code.

### Notes for release

Slug options `source` and `slugify` are now passed a `context` parameter as the last argument. It contains `getClient`, `schema` and `currentUser`.

### DIFF
--- a/packages/sanity/src/form/inputs/Slug/SlugInput.tsx
+++ b/packages/sanity/src/form/inputs/Slug/SlugInput.tsx
@@ -1,5 +1,12 @@
 import React, {useCallback, useMemo} from 'react'
-import {Path, SanityDocument, SlugParent, SlugSchemaType, SlugSourceFn} from '@sanity/types'
+import {
+  Path,
+  SanityDocument,
+  SlugParent,
+  SlugSchemaType,
+  SlugSourceContext,
+  SlugSourceFn,
+} from '@sanity/types'
 import * as PathUtils from '@sanity/util/paths'
 import {Box, Button, Card, Flex, Stack, TextInput} from '@sanity/ui'
 import {PatchEvent, set, setIfMissing, unset} from '../../patch'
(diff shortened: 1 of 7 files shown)


packages/sanity/src/form/inputs/Slug/utils/slugify.ts
```diff
@@ -9,12 +9,16 @@ const defaultSlugify = (value: FIXME, type: SlugSchemaType): string => {
   return value ? speakingurl(value, slugifyOpts) : ''
 }
 
-export function slugify(sourceValue: FIXME, type: SlugSchemaType): Promise<string> {
+export function slugify(
```
bjoerge, line 12: suggestion: same as above re: converting async function. I think we want the error to propagate on the returned promise (I do realise that this isn't "new" behavior, so feel free to ignore)

packages/sanity/src/form/inputs/Slug/SlugInput.tsx
```diff
@@ -15,26 +23,36 @@ export type Slug = {
...
-) {
+  document: SanityDocument,
+  context: SlugContext
+): SlugSourceContext {
   const parentPath = valuePath.slice(0, -1)
   const parent = PathUtils.get(document, parentPath) as SlugParent
+  return {parentPath, parent, ...context}
+}
+
+function getNewFromSource(
```
bjoerge, line 36: suggestion: make this an async function? That way we make sure that any (current or future) sync errors triggered by either this function or the slugifier call are thrown on the returned promise. As a bonus, the Promise.resolve call in here can be removed too.

bjoerge (approved):
Nice! Noted a couple of non-required suggestions for improvement.

packages/sanity/src/form/inputs/Slug/SlugInput.tsx
snorrees: Will do. Funny seeing eslint fight itself:

image

image

🤷

bjoerge: Seems like we should turn off the `require-await` rule or replace it with `@typescript/require-await` which uses type information: https://typescript-eslint.io/rules/require-await/

PR #3716, opened 2022-10-04
chore(ci): reduce PR builds

stipsan:
### Description

Our PRs are a bit noisy atm as most CI tasks run double checks. With this it'll only run its CI things once per PR (without breaking fork PRs).


mariuslundgard (approved):
LGTM!

PR #3717, opened 2022-10-04
chore: fix double export

stipsan:
### Description

Seems like a merge conflict gone wrong, it fixes a related linter error 
image


mariuslundgard:
This will be fixed in API members PR as well

PR #3719, opened 2022-10-04
fix: object reconciler

snorrees:
### Description

Form state now correctly updates members when hidden state changes.

Underlying issue: immutableReconcile did not account for changes in number of keys in objects,
which resulted in it considering next object unchanged when a key had been removed.
This made the code retain the previous value incorrectly.

This had the effect of associating wrong schemaTypes with member fields, when hidden function
changed state (as one observed effect).

### What to review

Evaluate the code.

This was verified in the test-studio by reproducing a reported issue related to this (observed as wrong schema passed as props to renderInput) and confirming that it fixed.

### DIFF
--- a/packages/sanity/src/core/form/store/utils/__tests__/immutableReconcile.test.ts
+++ b/packages/sanity/src/core/form/store/utils/__tests__/immutableReconcile.test.ts
@@ -108,5 +108,6 @@ test('returns new array when previous and next has different length', () => {
   const lessItems = ['a']
 
   expect(immutableReconcile(moreItems, lessItems)).not.toBe(moreItems)
+
   expect(immutableReconcile(lessItems, moreItems)).not.toBe(lessItems)
 })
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/store/utils/__tests__/immutableReconcile.test.ts
```diff
@@ -87,3 +87,26 @@ test('does not mutate any of its input', () => {
...
+    key1: 'value1',
+    key2: 'value2',
+    key3: 'value3',
+  }
+  const lessKeys = {
+    key1: 'value1',
+    key2: 'value2',
+  }
+
+  expect(immutableReconcile(moreKeys, lessKeys)).not.toBe(moreKeys)
```
snorrees, line 102: This fails without the changes to immutableReconcile.ts

snorrees:
Dunno whats going on with the github checks here??

bjoerge (approved):
Thanks a bunch for digging into this. Great to have this sorted out 💯

rexxars (approved):
👏

PR #3721, opened 2022-10-05
feature(cli): add `reactStrictMode` option

stipsan:
## Description

Adds an option to `sanity.cli.js` that controls wether strict mode is enabled. It was previously always-on. Now it's opt-in. The test studio enables it by default.
Prior art: https://nextjs.org/docs/api-reference/next.config.js/react-strict-mode

## How to disable strict mode in the test studio without editing `renderStudio.tsx` or `dev/test-studio/sanity.cli.ts`
Either run the dev command with an env var override:
```bash
SANITY_STUDIO_REACT_STRICT_MODE=false yarn dev
```

Or run this once and carry on as usual:
```bash
cp dev/test-studio/.env.example dev/test-studio/.env.development
```

## Double console log calls

If you're seeing double console logs after enabling strict mode, check the `Hide logs during second render in Strict Mode` option in the React Devtools Components Settings modal:

Screenshot 2022-10-05 at 06 35 29

## What to review

In studios that have a `sanity.cli.ts` file, add `reactStrictMode: true` to enable it.
In studios without, it should be possible to opt-in by either `SANITY_STUDIO_REACT_STRICT_MODE=true sanity start` or by creating a `.env.development` file adjacent to `sanity.config.ts` with the same env var.
Other studios shouldn't see strict mode on.
Including custom studios that call `renderStudio` directly.

## Notes for release

You can opt-in to [React's Strict Mode](https://reactjs.org/docs/strict-mode.html) checks for your custom studio components in these ways:
- Add `reactStrictMode: true` to your `sanity.cli.ts` file.
- Create a `.env.development` file and add `SANITY_STUDIO_REACT_STRICT_MODE=true` to it.
- Set the env var while starting sanity: `SANITY_STUDIO_REACT_STRICT_MODE=true sanity start`

We strongly suggest you opt-in to better prepare your custom studio for the future of React. This only affects development mode, production mode or `sanity build` is unaffected by this setting.

### DIFF
--- a/dev/test-studio/.env.example
+++ b/dev/test-studio/.env.example
@@ -1,0 +1,4 @@
+# To disable strict mode on your local checkout:
+# `cp .env.example .env.development`
+# Then restart your `sanity start` or `yarn dev` command
+SANITY_STUDIO_REACT_STRICT_MODE=false
--- a/dev/test-studio/.gitignore
+++ b/dev/test-studio/.gitignore
@@ -2,3 +2,6 @@
 
 # Ignore generated source files
 /workshop/scopes.js
+
+.env.*
+!.env.example
--- a/dev/test-studio/sanity.cli.ts
+++ b/dev/test-studio/sanity.cli.ts
@@ -8,6 +8,10 @@ export default createCliConfig({
     dataset: 'test',
   },
 
+  // Can be overriden by:
+  // A) `SANITY_STUDIO_REACT_STRICT_MODE=false yarn dev`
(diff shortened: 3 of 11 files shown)


snorrees (commented):
Some nice info in the description here 👍 

Only have one suggestion: add a period in the last sentence of Notes for release. I read it as "only affects development mode, production mode  ... ie as a list of things it affects.

"This only affects development mode. Production mode or sanity build is unaffected by this setting."

snorrees (approved):

PR #3722, opened 2022-10-05
fix(sanity): fix controls for custom block components preview

sjelfull:
### Description

When a PTE object block has a custom `components.preview`, the dot-menu disappears. This fixes the check that adds the wrapper around the preview.

This is reproducible by adding a block to a PTE input with the following schema:
```ts
defineField({
  type: 'object',
  name: 'testPreview',
  fields: [
    {
      title: 'Title',
      name: 'title',
      type: 'string',
    },
  ],
  preview: {
    select: {
      title: 'title',
    },
  },
  components: {
    preview: () => <div>Yolo</div>,
  },
})
```

### What to review

- That the UI looks correct when you have a custom `components.preview` for a block preview

### Notes for release

- Fixed missing block action menu around custom block object previews


stipsan (approved):
LGTM 😄

PR #3723, opened 2022-10-05
Feat/v3/poc component props

snorrees:
PoC on how we can augment `@sanity/types` from `sanity` to avoid moving component props around.


skogsmaskin:
@snorrees - is this something we sohuld re-open?

PR #3724, opened 2022-10-05
feat(sanity): studio default global styles

mariuslundgard:
### Description

This change introduces an internal component named `<GlobalStyle />` that uses `createGlobalStyle` from `styled-components` to render global CSS that sets a background color to the `<html>` element and a `font-family` to the `#sanity` element.

This fixes the problem of incorrect font-family in the cases where you don't use `<Text>` from `@sanity/ui`.

NOTE: global styles will only be applied by default in a context where you use `sanity start` or `sanity build`. If you embed the `<Studio />` in e.g. Next.js, you'll have to enable global styles specifically:

```tsx
<Studio
  config={config}
  // tell the studio to render global styles
  unstable_globalStyles
/>
```

### What to review

- That this works in a Next.js app.
- That this works as intended in the default Studio runtime.

### Notes for release

- The Studio will now render a default font-family and background-color.

### DIFF
--- a/packages/sanity/src/core/studio/GlobalStyle.tsx
+++ b/packages/sanity/src/core/studio/GlobalStyle.tsx
@@ -1,0 +1,15 @@
+import {createGlobalStyle} from 'styled-components'
+
+export const GlobalStyle = createGlobalStyle(({theme}) => {
+  const {color, fonts} = theme.sanity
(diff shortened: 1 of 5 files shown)


rexxars (approved):
👍

PR #3725, opened 2022-10-05
fix(sanity): fix icon from schema in block previews

sjelfull:
### Description

This fixes rendering of icons from schema in `SanityPreview` by making sure the icon is pulled out from the schema type.

image

### What to review

- Does the change catch all schema types? Before we had [a special case for references](https://github.com/sanity-io/sanity/blob/817abbdf31e28180fea477070f6b96387cec751f/packages/@sanity/base/src/preview/components/RenderPreviewSnapshot.tsx#L31), do we still need that?

### Notes for release

- Fixed icons from schema in block previews


joaisa17:
This could previously be solved by defining `media` in preview, as the icon

hermanwikner (approved):
Looks good to me 👍 But would be good to have @skogsmaskin's eyes on this as well

skogsmaskin (approved):

PR #3730, opened 2022-10-06
fix: hook factory strict mode errors

stipsan:
### Description

Fixes issues in `createHookFromObservableFactory` that React Strict Mode 18 forced to the surface.
The fix is in two places, in the new `react-rx` version introduced in this PR. And the refactored version of `createHookFromObservableFactory` internals.
This PR also introduces a new command `yarn lint:exhaustive-deps` that lints with `--quiet` while setting the deps check rule to `error` so you can quickly identify potential problems.
It's really helpful when working in strict mode to uncover bugs.

Also also, `useMemoObservable`, `useObservableCallback` and `useAsync` now receives the same helpful linter warnings as we're used to in `useEffect`, `useMemo` etc.

### What to review

Search for anything using `createHookFromObservableFactory` and components that subscribe to the observable hooks produced by the factory. In Strict Mode you should no longer see any errors. And outside Strict Mode you should see a nice speed boost to their performance.

### Notes for release

Fixes bugs uncovered by running Strict Mode on our test studio.


bjoerge (approved):
Looks great!

packages/sanity/src/core/studio/components/navbar/workspace/WorkspaceAuth/WorkspaceAuth.tsx
```diff
@@ -17,10 +17,13 @@ export function WorkspaceAuth() {
...
-    workspaces.length === 1
-      ? workspaces[0]
-      : workspaces.find((workspace) => workspace.name === selectedWorkspaceName)
+  const selectedWorkspace = useMemo(
+    () =>
+      workspaces.length === 1
+        ? workspaces[0]
+        : workspaces.find((workspace) => workspace.name === selectedWorkspaceName),
+    [workspaces, selectedWorkspaceName]
+  )
```
rexxars, line 26: Workspaces will usually contain < 5 items - is looping over 5 items going to be slower than the overhead of a memo hook here?

packages/sanity/src/core/util/useUnique.ts
```diff
@@ -4,6 +4,7 @@ import {useRef} from 'react'
 /**
  * This React hook should be considered an escape hatch – to make sure that a value is the same
  * on every render. SHOULD NOT BE USED IN MOST CASES.
+ * @deprecated please use `useMemo` and `useCallback` strategies instead to make deps stable, this hook runs comparisons on every single render and whiel each comparison can be fast, it quickly adds up
```
rexxars, line 7:
```suggestion
 * @deprecated please use `useMemo` and `useCallback` strategies instead to make deps stable, this hook runs comparisons on every single render and while each comparison can be fast, it quickly adds up
```

rexxars (approved):

rexxars (changes requested):
Tests are failing. Thanks for adding the additional hook dependency check!

PR #3743, opened 2022-10-10
Fix handle change exception

skogsmaskin:
### Description

We had a bug where the FormView would throw errors when calling onChange from the PT-input while switching documents. This was quick fixed by this commit: https://github.com/sanity-io/sanity/pull/3700/commits/6f3f6ff1cfabc384f0e65e45d1eb9fd2add98bde

However, the real problem seems to be that handleChange is called for a a FormView in which the DocumentPane isn't ready yet, so checking for this inside DocumentPaneProvider seems to be the correct place to fix this with the current codebase.

I have also reverted the quick fix I did.


stipsan (approved):
LGTM 👍

PR #3746, opened 2022-10-11
feat: legacy theming

mariuslundgard:
### Description

Move logic from v2 for creating a Sanity UI theme object based on a record of CSS variables.

You can now use `buildLegacyTheme` to generate a Sanity UI theme:

```ts
import {buildLegacyTheme, createConfig} from 'sanity'

export default createConfig({
  // project configuration ...

  // Customize theming
  theme: buildLegacyTheme({
    '--black': '#000',
    '--gray': '#777',
    '--focus-color': '#00f',
  })
})
```

### What to review

- Open the [Test Studio](https://test-studio-git-feat-legacy-theme.sanity.build/test/content)
- Open the Google workspace
- Open the Vercel workspace
- Check that colors and contrasts look reasonable and good enough
- Check that light/dark mode switching is NOT supported for legacy themes

### Notes for release

- Use `buildLegacyTheme()` to migrate your theme from v2 => v3. NOTE: this will disable light/dark mode switching.

### DIFF
--- a/packages/sanity/src/core/studio/StudioThemeProvider.tsx
+++ b/packages/sanity/src/core/studio/StudioThemeProvider.tsx
@@ -1,7 +1,9 @@
+/* eslint-disable no-nested-ternary */
+
 import React from 'react'
 import {ThemeProvider, LayerProvider} from '@sanity/ui'
 import {useActiveWorkspace} from './activeWorkspaceMatcher'
-import {useColorScheme} from './colorScheme'
+import {ColorSchemeContext, useColorScheme} from './colorScheme'
 
 interface StudioThemeProviderProps {
   children: React.ReactChild
@@ -10,11 +12,14 @@ interface StudioThemeProviderProps {
 /** @internal */
 export function StudioThemeProvider({children}: StudioThemeProviderProps) {
   const theme = useActiveWorkspace().activeWorkspace.theme
-  const {scheme} = useColorScheme()
+  const colorScheme = useColorScheme()
+  const scheme = theme.__legacy ? (theme.__dark ? 'dark' : 'light') : colorScheme.scheme
 
   return (
-    <ThemeProvider scheme={scheme} theme={theme} tone="transparent">
-      <LayerProvider>{children}</LayerProvider>
-    </ThemeProvider>
+    <ColorSchemeContext.Provider value={{...colorScheme, scheme}}>
+      <ThemeProvider scheme={scheme} theme={theme} tone="transparent">
+        <LayerProvider>{children}</LayerProvider>
+      </ThemeProvider>
+    </ColorSchemeContext.Provider>
   )
 }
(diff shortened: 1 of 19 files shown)


packages/sanity/src/core/studio/StudioThemeProvider.tsx
```diff
@@ -10,11 +12,14 @@ interface StudioThemeProviderProps {
...
   const theme = useActiveWorkspace().activeWorkspace.theme
-  const {scheme} = useColorScheme()
+  const colorScheme = useColorScheme()
+  const scheme = theme.__legacy ? (theme.__dark ? 'dark' : 'light') : colorScheme.scheme
 
   return (
-    <ThemeProvider scheme={scheme} theme={theme} tone="transparent">
-      <LayerProvider>{children}</LayerProvider>
-    </ThemeProvider>
+    <ColorSchemeContext.Provider value={{...colorScheme, scheme}}>
```
rexxars, line 19: Should this be memoized? I guess there aren't too many things outside of the theme provider that causes rerenders, but when rendered inside of other apps it might be?

packages/sanity/src/core/theme/types.ts
```diff
@@ -1,4 +1,9 @@
 import {RootTheme} from '@sanity/ui'
 
-/** @beta */
-export type StudioTheme = RootTheme
+/** @public */
+export interface StudioTheme extends RootTheme {
+  /** @internal */
+  __dark?: boolean
+  /** @internal */
+  __legacy?: boolean
```
rexxars, line 8: Could there be a case for making this not about legacy theming but rather about only having one scheme defined? If so, perhaps this could be promoted to a public property.

packages/sanity/src/core/theme/_legacy/theme.ts
```diff
@@ -0,0 +1,134 @@
...
+  // Gray
+  '--gray': gray[500].hex,
+  '--gray-base': gray[500].hex,
+  // '--gray-darkest': gray[900].hex,
+  // '--gray-darker': gray[800].hex,
+  // '--gray-dark': gray[600].hex,
+  // '--gray': gray[500].hex,
+  // '--gray-light': gray[400].hex,
+  // '--gray-lighter': gray[200].hex,
+  // '--gray-lightest': gray[100].hex,
```
rexxars, line 84: What is the deal with these commented variables?

rexxars (approved):
Looks great! Couple of questions, but if we want to change anything we can do so later.

packages/sanity/src/core/theme/_legacy/theme.ts
mariuslundgard: They were configured in v2, but by adding TypeScript to this, it seems they're not being used. Will remove commonets.

packages/sanity/src/core/theme/types.ts
mariuslundgard: Yes, probably – but that's outside of scope here.

It's `@internal` so can be renamed later.

PR #3750, opened 2022-10-12
fix: persist split panes on reload

stipsan:
### Description

Fixes a long standing bug where if you reload the page when in split view the extra panes are closed

Before: https://test-studio-git-v3.sanity.build/test/content/author;grrm%7C%2Cview%3Djson
After: https://test-studio-git-bug-sc-25678split-pane-urls.sanity.build/test/content/author;grrm%7C%2Cview%3Djson

### Notes for release

- Split panes persists on page reloads. And can be linked to.


rexxars (approved):
👍

PR #3754, opened 2022-10-12
fix(router): reduce rerenders and prevent history.replace during render

stipsan:


skogsmaskin:
@stipsan - is this still on? It was closed when we renamed the branches yesterday.

stipsan:
Superseded by #4282

PR #3757, opened 2022-10-13
fix: only show "document published" toast once

stipsan:
### Description

Prevents toasts that have already been fired from being fired again.

### What to review

See if push toasts work as they should.

### Notes for release

Toast notifications no longer fire more than once for the same event.


mariuslundgard (approved):
LGTM

PR #3759, opened 2022-10-14
fix(desk-tool): fix issue with history pane not showing the earlier version

bjoerge:
### Description

Fixes a regression introduced in [v2.34.2](https://github.com/sanity-io/sanity/releases/tag/v2.34.2) that broke the revision history selector by making it always show the current version of the document.

### What to review

- Make sure navigating throuigh revision history will display the selected revision

### Notes for release

- Fixes a bug causing revision history selector to always display the current version instead of the selected revision


skogsmaskin (approved):
LGTM!

Issue #3761, opened 2022-10-15
Add "types" field to named exports on @sanity/types


james-camilleri:
**Describe the bug**
The latest versions of the "@sanity/types" package support named exports, but the named exports do not include a `"types"` field, which causes TypeScript to throw errors when using a project in `nodenext` module resolution mode.

**To Reproduce**
1. Create a new v3 Sanity studio project.
2. Add a new type definition with the `defineField` or `defineType` type helpers.
3. Update the `moduleResolution` compiler option to `nodenext`.
4. `defineField` and `defineType` input throw a type error: `Module '"sanity"' has no exported member 'defineType'.`

**Expected behavior**
Imports should work using the latest typescript resolution method, which is where the overall ecosystem is heading.

**Which versions of Sanity are you using?**
@sanity/types                 3.0.0-dev-preview.20 (latest: 2.34.0)

*Less important, but included anyway*
@sanity/cli                                 2.30.6 (latest: 2.34.2)
@sanity/eslint-config-studio                 2.0.0 (up to date)
@sanity/vision                3.0.0-dev-preview.20 (latest: 2.34.0)

**What operating system are you using?**
Windows 10.

**Which versions of Node.js / npm are you running?**
Node 16.17.1, npm 8.15.0

**Additional context**
I've attempted to update the `exports` field in the `@sanity/types` package.json, and it seems to have corrected the issue locally. I've opened a PR to this effect.

james-camilleri:
I've thrown together a quick commit that's seemed to solve the problem locally.:
https://github.com/sanity-io/sanity/pull/3762

mariuslundgard:
PS. See reason for closing in #3762.

PR #3762, opened 2022-10-15
Correct exports field for `nodenext` module resolution

james-camilleri:
### Description
Adds the existing type definition to the named `exports` field in the `package.json` file for `@sanity/types`, in order to solve #3761. This should stop type errors from being thrown when Typescript's `nodenext` module resolution is used.

### What to review
In order to reproduce the type error:
1. Create a new v3 Sanity studio project.
2. Add a new type definition with the `defineField` or `defineType` type helpers.
3. Update the moduleResolution compiler option to `nodenext`.
4. `defineField` and `defineType` input throw a type error: `Module '"sanity"' has no exported member 'defineType'.`

This PR should correct the above type errors.

### Notes for release
Add named export for typescript types.

### DIFF
--- a/packages/@sanity/types/package.json
+++ b/packages/@sanity/types/package.json
@@ -7,6 +7,7 @@
   "module": "./lib/index.js",
   "exports": {
     ".": {
+      "types": "./lib/dts/src/index.d.ts",
       "source": "./src/index.ts",
       "require": "./lib/index.cjs",
       "default": "./lib/index.js"


mariuslundgard:
Hello @james-camilleri – thanks for you PR 👍  Since we have removed the `"exports"` property from the `@sanity/types` package, this PR no longer applies unfortunately. So will close. Let us know there are still problems that are not addressed.

This also closes #3761.

PR #3764, opened 2022-10-16
feat(studio): sortable search results, search phrases, experimental config 

robinpyon:
# Description

This PR introduces the search improvements introduced in V2 [here](https://github.com/sanity-io/sanity/pull/3652) with a few notable differences:

- Dark mode support
- Search provider and context are now in separate files to workaround [HMR issues in vite](https://github.com/vitejs/vite/issues/3301)
- Added experimental search config types (`__experimental_search` and `__experimental_omnisearch_visibility`) to the document schema definition (so they should appear when using `defineType`)
- Added Workshop stories for `<SearchPopover>` and `<SearchDialog>`

## Notable refactoring

1. Functions to generate weighted search (also used by reference and cross dataset reference input components) have been moved into `src/core/search` (previously `src/core/studio/components/navbar/search/search`). This felt natural since generating weighted search is a cross-cutting concern and not necessarily tied to the navbar.
2. Some _internal_ shared preview functions used by both `<PaneItemPreview>` and `<SearchResultItemPreview>` have been moved into `core/preview/utils`. 

Perhaps in future there's an opportunity to export a general purpose document list preview component that also takes care of fetching / creating preview values – which could be useful for plugins such as `documents-pane` or scheduled publishing.

## Known issues

- ~~Text input fields in Omnisearch temporarily have borders. This is due to an issue where `border=false` on TextInput components aren't respected in the studio navbar. Currently investigating whether this is a bug or intentional.~~ (fixed in ` "@sanity/ui": "1.0.0-beta.30"`)
- Omnisearch works when changing between light / dark mode, but won't render the correct colours after changing the colour mode twice. 
- Virtualized lists don't appear to be as performant as in V2 (Try do a search for "a" in test studio and drag the scrollbar). There is notable 'pop-in' in V3 which wasn't present in V2, but it's still quite snappy and won't hang the main thread.


packages/sanity/src/core/studio/components/navbar/search/components/PointerOverlay.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+`
+
+/**
+ * Simple component which conditionally appears over command list items to cancel
+ * existing :hover states for all child elements.
+ *
+ * It should only appear if hover capabilities are available (not on touch devices)
+ */
+export const PointerOverlay = forwardRef<HTMLDivElement>(function PointerOverlay(_props, ref) {
```
snorrees, line 26: Think I missed this last time. Boy did you engineer a lot of stuff for this :D :mucle:

packages/sanity/src/core/studio/components/navbar/search/components/SearchHeader.tsx
```diff
@@ -0,0 +1,140 @@
...
+  position: relative;
+`
+
+const SearchHeaderCard = styled(Card)`
+  flex-shrink: 0;
+`
+
+const NotificationBadge = styled.div`
+  align-items: center;
+  background: ${({theme}: {theme: Theme}) => theme.sanity.color.selectable?.primary.enabled.fg};
```
snorrees, line 37: Optional; should there be an explicit fallback here?

packages/sanity/src/core/studio/components/navbar/search/components/searchResultItem/SearchResultItemPreview.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface SearchResultItemPreviewProps {
+  documentId: string
+  documentPreviewStore: DocumentPreviewStore
+  presence?: DocumentPresence[]
+  schemaType: SchemaType
+}
+
+/**
+ * Brute force all nested boxes on iOS to use `background-attachment: scroll`, as
```
snorrees, line 24: Does sanity/ui have a task for this?

snorrees (approved):
Feels a bit wierd to say "LGTM" on such a big PR, but I recognize a lot of the code and it is mostly moving files right?

Awesome job!

packages/sanity/src/core/studio/components/navbar/search/components/searchResultItem/SearchResultItemPreview.tsx
robinpyon: There's no task as this isn't so much an issue with `@sanity/ui`, but rather a temporary workaround to allow Skeleton components to render correctly within virtual lists, which is definitely an edge case. There's probably a better way to handle this.

This comment was also inaccurate re: `background-attachment: fixed` support on iOS and has been updated

PR #3773, opened 2022-10-19
fix(sanity): disable group tabs when review changes pane is open

hermanwikner:
### Description
The document group tabs should be disabled when the review changes pane is open.

### What to review
- Code review
- Should work for nested tabs

### Notes for release
fix(sanity): disable group tabs when review changes pane is open


bjoerge (approved):
Solid work!

PR #3778, opened 2022-10-20
fix(sanity): focus issues

hermanwikner:
### Description

This PR fixes some focus issues:

1. The `ImageActionsMenu` component used a global key-down event to close the menu and set focus to the button that opens the menu. This creates problems, because when you e.g. close the `PresenceMenu` with "Escape" – focus is set on the `ImageActionsMenu` component's menu button, and not the `PresenceMenu` menu button, because the global event is triggered. 
2. The `PresenceMenu` lost focus on the selected element when the list of users was updated. This PR introduces a fix to preserve focus on the selected element when the list is updated. This fix can be considered a temporary workaround, and we should probably update `sanity/ui` to handle this instead in the future.

### What to review
- Close the presence menu with "Escape" in a document with an image input. Focus should be set back on the presence menu button.
- Navigate in the presence menu – and focus should not be lost when the list of users is updated.
- Use the image input with the keyboard. Focus should be set back on the button(s) when you close the edit dialog(s) and menu button.

### Notes for release
fix(sanity): focus issues in presence menu and image input


bjoerge (approved):

bjoerge:
Great!

PR #3781, opened 2022-10-21
Add strict runtime validation of array choices

bjoerge:
### Description
This adds strict validation of list of predefined options provided for array types. This means you will now get a schema error if you provide list options that are not allowed by the array declaration.

This also _removes_ support for predefined choices on the shape of `{title: 'Some title', value: {_type: 'someObjectType', ...}}` since preview is better handled by delgating to the studio preview for the type. You can still provide display title for primitive values, as explained in the [help article](https://www.sanity.io/help/schema-predefined-choices-invalid) linked to the schema compile error.

Example:

image

### What to review
- Verify that the validation makes sense
- Verify that the errors are sensible and the information provided by the help url makes sense

### Notes for release

- Added strict validation of predefined list options for array types

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/types/array.ts
+++ b/packages/@sanity/schema/src/sanity/validation/types/array.ts
@@ -8,7 +8,27 @@ function isPrimitiveTypeName(typeName) {
   return typeName === 'string' || typeName === 'number' || typeName === 'boolean'
 }
 
-const pluralize = (arr: unknown[], suf = 's') => (arr.length === 1 ? '' : suf)
+function isAssignable(typeName, type) {
+  return (typeof type.name === 'string' ? type.name : type.type) === typeName
+}
+
+function quote(n) {
+  return `"${n}"`
+}
+
+function pluralize(arr: unknown[], suf = 's') {
+  return arr.length === 1 ? '' : suf
+}
+
+function format(value: unknown) {
+  if (Array.isArray(value)) {
+    return `array with ${value.length} entries`
+  }
+  if (typeof value === 'object' && value !== null) {
+    return `object with keys ${humanizeList(Object.keys(value).map(quote))}`
+  }
+  return quote(value)
+}
 
 export default (typeDef, visitorContext) => {
   // name should already have been marked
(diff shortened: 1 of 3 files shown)


packages/@sanity/schema/src/sanity/validation/types/array.ts
```diff
@@ -1,8 +1,35 @@
+import {flatten, partition} from 'lodash'
+import humanizeList from 'humanize-list'
 import {error, HELP_IDS, warning} from '../createValidationResult'
-import {flatten} from 'lodash'
 import {getDupes} from '../utils/getDupes'
 import {coreTypeNames} from '../../coreTypes'
 
+function isPrimitiveTypeName(typeName) {
```
snorrees, line 7: I thought v3 didn't allow implicit any?

Would love to see type info for all params, even if just inline definitions.

packages/@sanity/schema/src/sanity/validation/types/array.ts
```diff
@@ -106,6 +133,79 @@ export default (typeDef, visitorContext) => {
     )
   }
 
+  const [primitiveTypes, objectTypes] = partition(
```
snorrees, line 136: This code is kinda involved with a lot of nested ifs. Is it covered by unit-tests? Would make it a lot less scary to modify in the future.

packages/@sanity/schema/src/sanity/validation/types/array.ts
```diff
@@ -106,6 +133,79 @@ export default (typeDef, visitorContext) => {
...
+      isPrimitiveTypeName(visitorContext.getType(ofType.type)?.jsonType)
+  )
+
+  const isMixedArray = primitiveTypes.length > 0 && objectTypes.length > 0
+
+  if (isMixedArray) {
+    const primitiveTypeNames = primitiveTypes.map((t) => t.type)
+    const objectTypeNames = objectTypes.map((t) => t.type)
+    problems.push(
+      error(
```
snorrees, line 149: Would prefer of the error-message creation code (here and below) was extracted into functions; will make it easier to follow the flow/improve readability imo.

snorrees (approved):
Approved, if there exists tests, otherwise I would love to see a couple :D

packages/@sanity/schema/src/sanity/validation/types/array.ts
bjoerge: Good point - seems strict mode has been disabled for this package https://github.com/sanity-io/sanity/blob/2778e0ca004ddcf0e32fb8f4cf91b7323e640b1e/packages/@sanity/schema/tsconfig.json#L10

packages/@sanity/schema/src/sanity/validation/types/array.ts
bjoerge: I'm afraid that goes for almost everything in this package. I think it's better addressed by doing a separate quality  improvement iteration on this package (includes adding strict mode, upping test coverage etc).

PR #3791, opened 2022-10-24
chore(deps): update react-virtual

robinpyon:
### Description

Updates `react-virtual` to use `3.0.0-beta.18` (which lists React 18 as a peer dependency). This _should_ remove dependency installation warnings.

### What to review

We currently use `react-virtual` quite simply, and that's only to render search results in Omnisearch. 

Search results should scroll and be keyboard navigable as normal.

### Notes for release

N/A


rexxars (approved):
Grazie!

PR #3793, opened 2022-10-25
Improve PTE test stability

skogsmaskin:
After upgrading to React 18, and other dependencies (i.e. new packing system) the portable-text-editor's tests have been needing some improvements.

I have done the following to better support their async nature:

* Refactor some of the tests to use `waitFor` instead of `act` from react testing library. This get rid of those warnings we have been seeing which potentially also can be a source to instability in test results.
* Run the individual tests in sync. Though each collaborative editing test websocket payload is tagged with the instance of the test and it's value is mapped to that id on the server (in order to support running in parallell), running these in parallel still seems to somehow yield different results than running them in serial mode (`jest --runInBand`). This is proven by one test that will differ between the modes. The best option for now seems to be to run these so that only one test use the test web/socket-server at the time. We already use `--runInBand` on the CI server for the root test list.
* Upped timeout limits for DOM selectors in Puppeteer. Sometimes tests seems to struggle when little CPU time is available and the test would fail with a timeout error.

### DIFF
--- a/packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
+++ b/packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
@@ -18,9 +18,11 @@ const WEB_SERVER_ROOT_URL = 'http://localhost:3000'
 // eslint-disable-next-line no-process-env
 const DEBUG = process.env.DEBUG || false
 
-// Wait this long for selections and a new doc revision to appear on the clients
-const SELECTION_TIMEOUT_MS = 1500 // This will also be an indicator of the performance in the editor. Set it as low as possible without breaking the tests.
-const REVISION_TIMEOUT_MS = FLUSH_PATCHES_DEBOUNCE_MS + 1000 // 300 seems to be the limit for the doc patching to go full circle (increase this if tests starts to time out)
+// Wait this long for selections and a new doc revision to appear in the clients.
+const SELECTION_TIMEOUT_MS = 1500
+
+// How long to wait for a new revision to come back to the client(s) when patched through the server.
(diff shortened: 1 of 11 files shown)


packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
```diff
@@ -19,8 +19,8 @@ const WEB_SERVER_ROOT_URL = 'http://localhost:3000'
 const DEBUG = process.env.DEBUG || false
 
 // Wait this long for selections and a new doc revision to appear on the clients
-const SELECTION_TIMEOUT_MS = 1000 // This will also be an indicator of the performance in the editor. Set it as low as possible without breaking the tests.
-const REVISION_TIMEOUT_MS = FLUSH_PATCHES_DEBOUNCE_MS + 300 // 300 seems to be the limit for the doc patching to go full circle (increase this if tests starts to time out)
+const SELECTION_TIMEOUT_MS = 1500 // This will also be an indicator of the performance in the editor. Set it as low as possible without breaking the tests.
+const REVISION_TIMEOUT_MS = FLUSH_PATCHES_DEBOUNCE_MS + 1000 // 300 seems to be the limit for the doc patching to go full circle (increase this if tests starts to time out)
```
bjoerge, line 23: the comment says "300 seems to be the limit for the doc patching to go full circle (increase this if tests starts to time out)" - is _300_ still accurate?

bjoerge (approved):
Great work! Thank you for taking the time to look into these!

packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
skogsmaskin: I will edit these comments as they no longer are valid with the new values.

PR #3795, opened 2022-10-25
Feat/type improvments

snorrees:
## Description

Improvements to definition types:

### Explicit components props
Removed `components` from `@sanity/types` definitions.

Added explicit, per type components via  declaration merging in `sanity/form`.
This way we dont have to worry about cyclic dependencies.

It should no longer be possible to assign an input-component to a form-component now.

Added test-suite to check that all types support type-narrowed props.

### Simplified `fields` requirements
`fields` is now simply ` FieldDefinition[]` again.

@rexxars noted that using `fields` with filtered arrays and the like is painful, since` FieldDefinition[]` is not assignable to `[FieldDefinition, ...FieldDefinition[]]`. The type-saftey we get from the more strict  definition is not worth it.
There is already a runtime check that ensures that object and document require one field, so this is fine.

### Input placeholder text
`placeholder` (as in, HTML input placeholder text) added to the types that support it. This includes 
  * date
  * datetime
  * number
  * string
  * text
  * url

Slug does not support placeholder. Not adding support for it in this type-only PR even though it possibly should support it.
  
### Made options.filter optional on reference

Ensures that `filterParams` only can be used when `filter` is `string`, and allows just specifying `disableNew` as option and nothing else.

## What to review

Look at the code ;)

## Notes for release

Probably not needed.


packages/@sanity/types/src/reference/types.ts
```diff
@@ -35,7 +35,8 @@ export type ReferenceFilterResolver = (options: {
 
 /** @public */
 export interface ReferenceFilterResolverOptions {
-  filter: ReferenceFilterResolver
+  filter?: ReferenceFilterResolver
+  filterParams?: never
```
snorrees, line 39: Had to add this never bit to allow TS to differentiate `ReferenceFilterResolverOptions` and  `ReferenceFilterQueryOptions`

packages/sanity/src/core/form/types/definitionExtensions.ts
```diff
@@ -0,0 +1,254 @@
...
+  ArrayOfObjectsInputProps,
+  ArrayOfPrimitivesInputProps,
+  BooleanInputProps,
+  NumberInputProps,
+  ObjectInputProps,
+  StringInputProps,
+} from './inputProps'
+import {ObjectItem, ObjectItemProps, PrimitiveItemProps} from './itemProps'
+
+export interface ArrayOfObjectsComponents {
```
snorrees, line 29: A lot of these are identical, but having per-type explicit definitions feels more correct.

rexxars (approved):
👍

PR #3798, opened 2022-10-25
Feat/arrayitem chopup

snorrees:
### Description

PoC on how we can extract and export array-bits to make it easier to compose using existing functionality.

Making reusable `Components` is tricky due to minor styling variations, so I tried to extract behavior instead, in the form of hooks.

This is just a firstpass attempt but take a look at the `InlineItemComposedOfParts`

It gives us this
image

With very little copy paste imo. And focus/tab handling and drag-and-drop "just works" too!

Probably need a bit more work on naming things and add @beta/@alpha/@interanl tags, but I think this should go a long way. What do you think, worth pursuing?


bjoerge:
Fantastic effort, @snorrees!! I'll do my best to find time to give this a proper review, but might take a while before I get around, so I'm temporarily unassigning myself as reviewer to keep my notifications in place.

PR #3812, opened 2022-10-27
fix(react-hooks): issue with `useEditState` low priority

hermanwikner:
### Description
This PR fixes an issue that documents are not loaded when using the `useEditState` hook with "low priority" with `rxjs@7`. 

### What to review
- Test that documents load when using `rxjs@7`

### Notes for release
Fix issue with documents not loading when using `rxjs@7`


skogsmaskin (approved):
If it still works the way that it emits immediately first, then emits debounced by a second afterwards I am happy with this change.

As far as I can tell that's the way it works with the updated observable, but I haven tested it.

skogsmaskin:
I have confirmed it works the same way.

bjoerge (approved):

PR #3814, opened 2022-10-27
fix(server): configure `cacheDir` to avoid cache conflicts

mariuslundgard:
### Description

Set `viteConfig.cacheDir` to a location that is different from Vite's default location: `node_modules/.vite`. This makes sure that there will not be Vite cache conflicts if you happen to use Vite for something else than `sanity` in your project.

### What to review

- Remove any `node_modules/.vite` directories.
- Run `sanity dev` & `sanity build` and make sure the cache is stored in `node_modules/.sanity/vite`.

### Notes for release

- Fix an issue where Vite cache could be overwritten/corrupted.


rexxars (approved):
Seems to work nicely!

PR #3824, opened 2022-10-31
refactor: rename `createConfig` to `defineConfig`

rexxars:
### Description

To align better with other frameworks and our own methods (`defineType`, `defineField` et all), we want to change from `createConfig` to `defineConfig` (and `defineCliConfig` instead of `createCliConfig`). This PR makes this backwards compatible - we're only deprecating the old `create` methods. Prior to GA release, they will be removed.

### Notes for release

- `createConfig` renamed to `defineConfig`, and `createCliConfig` renamed to `defineCliConfig`


bjoerge (approved):
Great call, LGTM!

PR #3825, opened 2022-10-31
fix(schema): validate that component properties are react components

rexxars:
### Description

We currently have a bug where if `components` is defined but `input` is not, it will yield a validation warning about invalid input component being used. This PR fixes this, as well as adding validation for other components (field, item, preview for now).

Since it no longer deals with _input_ components specifically, I also renamed some methods/files to reflect this.

### What to review

- Schema validation works when not defining `input` in `components`.
- Schema validation gives warnings if items in `components` are not React components.

### Notes for release

- Fixed a bug where not specifying an `input` component in the `components` field would give a schema validation warning


bjoerge (approved):
Nice! LGTM

PR #3827, opened 2022-11-01
fix(desk-tool): fix: earlier revision not appearing in form

bjoerge:
This is a cherry pick of https://github.com/sanity-io/sanity/pull/3759 into `v3`, plus a fix for a v3-only issue with forms not becoming `readOnly` based on external factors like "viewing an older revision", or "current user has no access to edit".


rexxars (approved):
Good find! LGTM!

PR #3836, opened 2022-11-03
feat(cli): use typescript by default for new projects

rexxars:
### Description

After a lot of discussion, we've decided that we should be encouraging the use of typescript for new projects. This PR changes the default from javascript to typescript. People will still be prompted, but the default choice is to use typescript. You can use `--no-typescript` to explicitly opt out, should you want to (`--typescript` still works as before, skipping the prompt).

### What to review

- Initializing new projects works as before, but the default in the typescript prompt is "yes"
- `--no-typescript` works as expected

### Notes for release

- New projects now use TypeScript by default


bjoerge (approved):
👍🏼 Sensible

PR #3853, opened 2022-11-09
Type fixes

snorrees:
### Description

Adds missing email type.
Fixes some issues with typing to make for better DX overall.

### What to review

Are we ok with email just supporting `rule.required().custom()` typewise for validation? We _could_ add all the string methods, but I see email ships with built in `rule.email()` that cannot be disabled afaik.

### Notes for release

Probably not needed. These are bugfixes for types imo.


packages/@sanity/types/src/schema/definition/schemaDefinition.ts
```diff
@@ -110,6 +113,14 @@ export type TypeAliasDefinition<
   validation?: SchemaValidationValue
   initialValue?: InitialValueProperty<any, any>
   preview?: PreviewConfig
+
+  components?: {
```
snorrees, line 117: Reinstates components with any props for TypeAliasDefinition, ie, the fallback case when we cannot determine an exact type.

packages/@sanity/types/src/schema/definition/type/block.ts
```diff
@@ -18,10 +18,10 @@ export interface BlockRule extends RuleDef<BlockRule, any[]> {}
 export interface DecoratorDefinition {
   title: string
   value: string
-  icon?: ReactNode | ComponentType
-  portableText?: {
-    icon?: ReactNode | ComponentType
-    render?: ComponentType
+  icon?: ReactNode | ComponentType<any>
+  blockEditor?: {
```
rexxars, line 22: Is this correct, @skogsmaskin ? I guess these are about to change real soon?

rexxars (approved):
👍  from me

packages/@sanity/types/src/schema/definition/type/block.ts
skogsmaskin: They are! It's OK to merge it like this for now though.

snorrees: Yeah, changed these to what works for runtime in the current version. Using portableText crashes the studio with an unhelpful error message.

PR #3857, opened 2022-11-09
feat(form): add API affordance for transforming patches

bjoerge:
### Description
This adds support for intercepting patches from a subtree of input components and transform them. Usage example: https://github.com/sanity-io/sanity/blob/d7220c45250c9631fdb98b4d722260d070604de8/dev/test-studio/schema/standard/objects.tsx#L38-L51

### What to review
Any feedback on naming/ API ergonomics etc. are appreciated

### Notes for release
- adds API affordance for transforming patches from an input component subtree


snorrees (approved):
🔥

PR #3867, opened 2022-11-14
fix: input mode for number type

### LINKED ISSUE #3866: [V3] Field Type Number does not render the correct `inputmode`
**Describe the bug**

> Before I start, I want to point out that those changes impact me while using my CMS on mobile (iOS 16). On web it wouldn't matter due to the keyboard usage.

I have a bunch of schemas containing `number` fields.

In v2, a `number` field renders an input type `text` (which, let's be honest, isn't the best option semantic for an input):

**Latest Sanity v2**
Screenshot 2022-11-14 at 18 37 56

On mobile, when I focus on this field, it opens the keyboard with letters, numbers, and symbols. Far from optimal, but I still could find the decimal symbols and fill those numbers out.

When I clone my CMS using Sanity v3, I notice that the same field now renders in type `number` (great), however, it has a mode called `numeric`:

Screenshot 2022-11-14 at 18 37 32

And that's not right for a number field.

That's because, according to MDN:

```
"Numeric" input keyboard, but only requires the digits 0–9. Devices may or may not show a minus key.
```

I think you can guess that now using Sanity v3 when I open on mobile, it only opens the numeric keyboard, and I cannot type decimals:

That's really hard because every time I have to type something on mobile, I have to do it wrong, open it on my PC and fix those values.

**To Reproduce**

1. creates a sanity project using v3;
2. defines a schema with a field type number;
3. run the server;
4. open sanity desk on your mobile (I tried on iOS, but I believe it would also affect android. Emulator might work also);
5. See the numeric field

**Expected behavior**

I expect when I define a `numeric` field, it uses `inputmode="decimal"` so the right keyboard layout could be invoked.

**Screenshots**

On the problem description

On my v3 project:

```
@sanity/cli (global)  3.0.0-rc.2 (latest: 2.35.0)
sanity                3.0.0-rc.2 (latest: 2.33.2)
```

On my v2 project:

```
@sanity/cli             2.33.3 (latest: 2.35.0)
@sanity/base            2.33.3 (latest: 2.35.0)
@sanity/cli             2.33.3 (latest: 2.35.0)
@sanity/core            2.33.3 (latest: 2.35.0)
@sanity/data-aspects    2.33.2 (up to date)
@sanity/default-layout  2.33.3 (latest: 2.35.0)
@sanity/default-login   2.33.3 (latest: 2.35.0)
@sanity/desk-tool       2.33.3 (latest: 2.35.1)
@sanity/form-builder    2.33.3 (latest: 2.35.0)
@sanity/types           2.33.2 (latest: 2.35.0)
@sanity/vision          2.33.3 (latest: 2.35.0)
```

iOS 16

- pnpm: v7.14.0
- node: v16.17.0

**Additional context**

- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

raulfdm:
### Description

Closes #3866

This PR changes the `inputMode` property to be `decimal` instead `number`. See the implantation in the open Issue.

### What to review

Check the number input.

Obs.: I tried to add an unit test to ensure this is not broken but I couldn't get the `minRule` condition working. I don't have much context for solving this kind of problem.

### Notes for release

When `type` is `number`, uses HTML `inputmode` equals `decimal` instead number. This will show up on mobile the decimal keyboard.

### DIFF
--- a/packages/@sanity/form-builder/src/inputs/NumberInput.tsx
+++ b/packages/@sanity/form-builder/src/inputs/NumberInput.tsx
@@ -43,7 +43,7 @@ const NumberInput = React.forwardRef(function NumberInput(
       <TextInput
         type="number"
         step="any"
-        inputMode={onlyPositiveNumber ? 'numeric' : 'text'}
+        inputMode={onlyPositiveNumber ? 'decimal' : 'text'}
         id={id}
         customValidity={errors && errors.length > 0 ? errors[0].item.message : ''}
         value={value}
--- a/packages/@sanity/form-builder/src/inputs/__tests__/NumberInput.test.tsx
+++ b/packages/@sanity/form-builder/src/inputs/__tests__/NumberInput.test.tsx
@@ -20,7 +20,7 @@ const dummyDocument = {
   _rev: '5hb8s6-k75-ip4-4bq-5ztbf3fbx',
   _type: 'numberFieldTest',
   _updatedAt: '2021-11-05T12:34:29Z',
-  num: 0,
+  num: 1,
   title: 'Hello world',
 }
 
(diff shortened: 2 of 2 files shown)


bjoerge (approved):
Thanks a lot! I've fixed the tests and changed the logic a bit so it still uses `inputMode=numeric` for fields that have `.integer()` or `.precision(0)` validation rules

Merged to both v2 (13e14ab) and v3 (38df0fb) and will be included in upcoming releases.

PR #3870, opened 2022-11-16
refactor(core)!: rename document action `modal` prop to `dialog`

rexxars:
### Description

**BREAKING CHANGE**: Document actions has a `modal` prop which in v2 was named `dialog`. In this version, we are reverting this back to `dialog`, since it is a more correct term for the feature. To align with this change, we are also renaming types that are coupled to it.

For now, this is a backwards compatible change, but before we launch v3 into General Availability (GA), the support for `modal` will be removed.

### What to review

- That document actions still work as expected (default + custom)
- That the names of types and props make sense

### Notes for release

- **BREAKING CHANGE:** Renamed the `modal` property to `dialog` for document actions. This makes them backwards compatible with v2 document actions, and is a more correct term for the feature.


bjoerge (approved):
LGTM!

PR #3874, opened 2022-11-16
refactor(cli): always print dev vs preview warning

rexxars:
### Description

Previously, we only showed a warning about the `sanity start` vs `sanity dev` change when we could not find a static build.

This PR changes it so that we _always_ show a warning message, and in a little more stand-outish way:

image

There were some suggestions on making `sanity start` a noop, but I think we should rather optimize for aligning with other frameworks. This warning should help people spot the difference - we can probably remove it at some point in the future.

### What to review

- Message gets printed when running `npx sanity start` - regardless of whether or not the `dist` folder exists
- The message itself

### Notes for release

- `sanity start` now always prints a message indicating that it is serving a _static_ build. If you want to skip the warning, use `sanity preview` instead - which is an alias, but does not have the warning. This message is printed as a helpful guide for existing Sanity users - in the future this warning will go away, aligning with other frameworks.


RitaDias (approved):
Works like a charm ✨

PR #3882, opened 2022-11-18
refactor(inputs): refactor PT-input annotation popover

skogsmaskin:
When there is a Popover element opened in the editor, wait for the editor reference element to be set before we open it. Currently there is an issue with @sanity/ui and Floating.ui that will not open the popover properly if the reference element is given at a later time.

Also, improve `useScrollToOpenenedMember` to actually test if the form has focus, and not only the editor itself. The form value is what we want here as the editor in transition could have focus, but not form state focus (focusPath).

Also fixed an related issue where the toolbar popover for PT annotations would not properly show on newly created annotations.

Also moved some common editor blur logic to a more central place and renamed toolbar prop `onExpand` to `onOpenItem` to make it clearer what it does and easier to change the common behaviour.

### Notes for release
Fixed UI issues related to adding annotations in the Portable Text Editor.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
@@ -87,7 +87,7 @@ export function Compositor(props: InputProps) {
 
   // Scroll to the DOM element of the "opened" portable text member when relevant.
   useScrollToOpenedMember({
-    hasFocus,
+    hasFormFocus: focusPath.length > 0,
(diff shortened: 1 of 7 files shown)


rexxars (changes requested):
Our coding guidelines say we should use `on<Subject><Verb>`, so `onOpenMember` should actually be `onMemberOpen` - this ensures we group together related callbacks and stay consistent across the codebase.

rexxars (approved):
LGTM!

PR #3883, opened 2022-11-18
fix(core): unify API order

hermanwikner:
### Description

This PR unifies the order in which APIs are composed. More specifically, this PR changes so that the parent config is always called after the child config. This gives the parent config control over the output, and not the plugin.

- **child config** = a plugin config
- **parent config**  = the root config (in `defineConfig`) or the config of a plugin that has a plugin

The illustration below shows an example of a configuration with different plugins, and what the result would be. The result is an array representing the order in which these configs will be resolved. The resolving starts from index 0, and then loop forward, giving the root config the final say on what the output should be.

order-example

### Why is this update needed?
- This update is important because the behaviour is currently not consistent across all APIs. All APIs except the components API are resolved by starting in the root configuration and then looping the plugins array forward. But the components API starts in the last plugin in the plugin array, looping the array backwards and ending in the root config. This is inconsistent behaviour that can lead to unexpected results. All APIs should be resolved in the same order.
- It makes sense that the parent config controls the output, and not a plugin. We've already seen confusion around the document actions API, where the `deskTool` plugin configures document actions. But since the root config is not the last step, it is not possible to control the document actions from the root config, instead you need to create a plugin that is _after_ `deskTool` in the plugins array to control the actions.

### ⚠️ Heads up!
- This might cause breaking behaviour, since the root config used to be _first_ in the array, but now it is _last_ in the array (expect for the components API)
- In the components API, we have looped the plugins array backward, and then finished with the root config. Since this update means that we will be looping the plugins array forward for all APIs, this change might be especially breaking for the components API.

### What to review

Add APIs to the root config and plugins, with plugins, and make sure that you get the expected result

### Notes for release

feat(core): unify API order


mariuslundgard (approved):
LGTM. I would call this a "fix" and not a "breaking change". Since it doesn't change API surface, and previously this was as bug.

rexxars (approved):
👍

PR #3891, opened 2022-11-22
refactor: re-order methods in fieldpresence to reduce deadcode removal on next

RitaDias:
### Description

There is an issue with Vercel/Next where after they've made changes to their configuration keys (specifically `swcMinify` is now set to default to true) that causes issues with some parts of the code. 
This change came after a bug report came over the `FieldPresenceWithOverlay` was not found when running `start` but was fine when running the `dev` command). 

After speaking with @stipsan , he realised - from other bugs exactly like this one - that this is Next recognises that the method `FieldPresenceWithOverlay` is not being called anywhere so it can remove it completely. Changing the order of the methods should help reduce the amount of times that we run into this issue in the future.

Vercel/Next have also been made aware that this is happening.

### What to review

The code is exactly the same, I've only moved the code that is called on other methods over to the top (so if method `a` calls `b`, b is instantiated before `a` and not after)

### DIFF
--- a/packages/sanity/src/core/presence/FieldPresence.tsx
+++ b/packages/sanity/src/core/presence/FieldPresence.tsx
@@ -16,48 +16,6 @@ import {useReporter} from './overlay/tracker'
 import {FormNodePresence} from './types'
 
 /** @internal */
-export interface FieldPresenceProps {
-  presence: FormNodePresence[]
-  maxAvatars: number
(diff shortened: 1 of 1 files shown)


stipsan (approved):
Good fix, let's hope this is enough to hold us over until Next.js ships a fix 🙇

PR #3892, opened 2022-11-22
fix(errors): unify error boundaries, prevent duplicate toasts

rexxars:
### Description

We currently have two error boundaries - one on the studio level and one on the tool level. This _sorta_ makes sense, since by having a separate tool-level boundary, we can still move to different tools if one of them crashes (the global error boundary doesn't render any navigation bar). However, the two error boundaries look and behave differently, which isn't great.

This PR refactors things a bit:
- Moves the error logging bit (`console.error()` + pushing a toast) out of the error boundary and into its own little component (doesn't render anything, but attaches a listener using a `useEffect`). Could have been just a hook, but needs to be rendered below the toast provider in the tree, so might as well make it a component.
- Makes the studio error boundary slightly more reusable by allowing the configuration of a heading text (also refactors the conditionals to reduce the nesting of the actual rendered UI)
- Drops the custom error boundary on the tool level, instead reusing the `StudioErrorBoundary`
- Uses an `id` on the pushed toasts to prevent the same exact error message from being displayed more than once at a time (this happens in certain cases because of React re-rendering and such)

### What to review

- Still catches errors in tools and globally
- Switching tools works even if one tool crashes

### Notes for release

None. Basically transparent for the end user (slightly different look and feel, but generally the same).

### DIFF
--- a/packages/sanity/src/core/error/ErrorLogger.tsx
+++ b/packages/sanity/src/core/error/ErrorLogger.tsx
@@ -1,0 +1,67 @@
+import {useToast} from '@sanity/ui'
+import {useEffect} from 'react'
+import {ConfigResolutionError, SchemaError} from '../config'
+import {CorsOriginError} from '../store'
+import {globalScope} from '../util'
+
+const errorChannel = globalScope.__sanityErrorChannel
+
+/**
+ * Attaches a listener to the global error channel and displays a toast when a (unknown)
+ * error occurs. Compares the last displayed error message with the current
+ *
+ * @internal
+ */
+export function ErrorLogger(): null {
+  const {push: pushToast} = useToast()
+
+  useEffect(() => {
+    if (!errorChannel) return undefined
+
+    // errorChannel.subscribe() returns a unsubscriber function.
+    // By returning it from this `useEffect`, it'll unsubscribe on unmount.
+    return errorChannel.subscribe((msg) => {
+      // NOTE: Certain errors (such as the `ResizeObserver loop limit exceeded` error) is thrown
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/error/ErrorLogger.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+      console.error(msg.error)
+
+      pushToast({
+        // Use the error message as the ID in order to prevent duplicates from showing
+        // A bit of a hack, but serves
+        id: msg.error.message,
+        closable: true,
+        description: msg.error.message,
+        duration: 5000,
```
hermanwikner, line 43: Nitpick: 5000 is the default duration so this value is not really needed!

packages/sanity/src/core/error/ErrorLogger.tsx
```diff
@@ -0,0 +1,67 @@
...
+      if (!msg.error) {
+        return
+      }
+
+      // For errors that we "expect", eg have specific error screens for, do not push a toast
+      if (isKnownError(msg.error)) {
+        return
+      }
+
+      console.error(msg.error)
```
hermanwikner, line 35: Do we want to log the error even though it is a known error?

hermanwikner (approved):
Looks and works good! 👍 

I had some small comments

packages/sanity/src/core/error/ErrorLogger.tsx
rexxars: I don't think so, since the error screens should have a much clearer explanation of what the error is.

PR #3893, opened 2022-11-23
fix: sanity dev will now react to env var (sanity staging)

RitaDias:
### Description

`sanity dev` didn't respect SANITY_INTERNAL_ENV="staging". 

### What to review

- To test, past export `SANITY_INTERNAL_ENV="staging"` in the terminal
   - ❗ we're missing env file handling, hence why we need this for a test purpose and it will be solved outside of the scope of this task
- Set projectId to a project that exists in the staging environment of sanity (.work) in `sanity.config.ts` for the test studio 
   - The current test studio (projectId) doesn't exist in staging and so it will cause errors if you try to run the command
    - Run `yarn dev`

### Notes for release

This is for internal use and doesn't need to be highlighted in the changelog


packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
```diff
@@ -169,6 +169,12 @@ export function _createAuthStore({
         requestTagPrefix: 'sanity.studio',
         ignoreBrowserTokenWarning: true,
         allowReconfigure: false,
+        apiHost:
+          /* __SANITY_STAGING__ is a global variable set by the vite config */
+          // @ts-expect-error: __SANITY_STAGING__ is a global env variable set by the vite config
```
RitaDias, line 174: ts was giving an error because it couldn't find `__SANITY_STAGING__`

rexxars (approved):
👍

PR #3895, opened 2022-11-23
Various fixes related to number input when used inside arrays

### LINKED ISSUE #1694: Unable to delete predefined "0" when you have an array of numbers.
**Describe the bug**
When you have defined an array of numbers in the schema it is not possible to remove the predefiend 0, which makes the input experience pretty cumbersome.

**To Reproduce**

1. Go to any schema
2. Define an array of numbers
``` 
{
      title: 'Week split',
      name: 'weekSplit',
      type: 'array',
      of: [{ type: 'number' }]
},
```
3. Add an array item
4. Try removing the zero in the input field
5. See error

**Expected behavior**

When a user is trying to input numbers with relative swiftness. It is cumbersome to have to navigate behind the zero, input the number you actually want, then navigate to the end and remove the zero. The only workaround as I see it today is to either define a custom object (which I have tested works), but that unnecessarily obfuscates the schema structure. The second option is to define a custom input component, I haven't tested if this works yet.

**Screenshots**

```
@sanity/cli             0.147.3 (up to date)
@sanity/base            0.147.6 (up to date)
@sanity/components      0.147.6 (up to date)
@sanity/core            0.147.6 (up to date)
@sanity/default-layout  0.147.6 (up to date)
@sanity/default-login   0.147.0 (up to date)
@sanity/desk-tool       0.147.7 (up to date)
@sanity/storybook       0.147.0 (up to date)
@sanity/vision          0.147.6 (up to date)
```
macOS Catalina version 10.15.2

6.13.4
v12.9.1

bjoerge:
### Description

Sorry for bundling several things into one PR. The commits can be reviewed one by one.

1. Fixes a bug preventing array items with an invalid type from being deleted
2. Changes from triple-equals to Object.is equality check in @sanity/mutator when checking if a patch needs applying or not (needed in next step)
3. Using a signed zero (`-0`) trick to mark number input as "empty" and render the number input with an empty string instead of a zero you can't get rid of. If the value is `-0` we pass an empty string to the number input. This allows us to retain the type information even when the field is empty. Closes #1694
4. Set the number input range to `[Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER]`. Rationale:
If you store numbers larger than what can fit in 32 bits you are likely going to get some surprises and have unintended consequences if you try to do arithmetics, so we should make it hard to do this by accident and encourage people to store large numbers as strings instead.

### What to review

- Check that number input doesn't allow numbers outside of the `[-9007199254740991, 9007199254740991]` range
- Check that adding/editing/removing of numbers in arrays works as expected.

### Notes for release

- Fixed an issue causing numeric array items to disappear when clearing the input

### DIFF
--- a/packages/@sanity/mutator/src/patch/ImmutableAccessor.ts
+++ b/packages/@sanity/mutator/src/patch/ImmutableAccessor.ts
@@ -81,7 +81,7 @@ export class ImmutableAccessor implements Probe {
       throw new Error('setIndex only applies to arrays')
     }
 
-    if (value === this._value[i]) {
+    if (Object.is(value, this._value[i])) {
       return this
     }
 
@@ -131,7 +131,7 @@ export class ImmutableAccessor implements Probe {
       throw new Error('Unable to set attribute of non-object container')
     }
 
-    if (value === this._value[key]) {
+    if (Object.is(value, this._value[key])) {
       return this
     }
 
--- a/packages/sanity/src/core/form/inputs/NumberInput.tsx
+++ b/packages/sanity/src/core/form/inputs/NumberInput.tsx
@@ -8,7 +8,7 @@ import {NumberInputProps} from '../types'
  * @beta
  */
 export function NumberInput(props: NumberInputProps) {
-  const {schemaType, validationError, value, elementProps} = props
+  const {schemaType, validationError, elementProps} = props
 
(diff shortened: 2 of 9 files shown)


skogsmaskin (approved):
Nice! Thank you for this.

Issue #2947, opened 2021-11-22
.defaultOrdering() on documentList() does not seem to work 


tommhuth:
When using `documentList()`, the corresponding `defaultOrdering()` does not seem to have any effect. However, moving the ordering into `documentList` as a param does produce the expected sort order:

```js
S.documentList({
    options: {
        defaultOrdering: [{ field: "_updatedAt", direction: "desc" }],
        filter: `_type == "foo"`,
    },
    id: schemaType,
    title: "Foo",
})
```

 I see someone has already posted an [issue](https://github.com/sanity-io/sanity/issues/2534) regarding `defaultOrdering` on  `documentTypeListItem`, but I'm not quite sure if this is a duplicate of that,  as the method clearly does exist, it just doesn't seem to do anything.  

**Which versions of Sanity are you using?** 
@sanity/cli               2.6.0 (latest: 2.22.0)
@sanity/base            2.21.10 (latest: 2.22.0)
@sanity/cli             2.21.10 (latest: 2.22.0)
@sanity/components       2.14.0 (up to date)
@sanity/core            2.21.11 (latest: 2.22.0)
@sanity/default-layout  2.21.10 (latest: 2.22.0)
@sanity/default-login   2.21.10 (latest: 2.22.0)
@sanity/desk-tool       2.21.10 (latest: 2.22.2)
@sanity/vision          2.21.10 (latest: 2.22.0)

**What operating system are you using?**
Mac 11.6

**Which versions of Node.js / npm are you running?**
Node: v14.15.4
NPM: 6.14.10

guledmahdi:
This was previously working until we upgraded to the latest sanity build

straube:
I'm seeing the same behavior. Added `defaultOrdering( ... )` to a `documentTypeList()` and it's not working.

PR #3897, opened 2022-11-23
fix(desk): use custom sort order by default and only use defaultOrdering as fallback

vicmeow:
### Description

Custom sorts were broken and document list would not update with the correct sort order when changing sort order from the pane menu. 

### What to review

Make sure that ordering works as expected

### Notes for release

- Fixed an issue where custom sort orders were broken

### DIFF
--- a/packages/sanity/src/desk/panes/documentList/DocumentListPane.tsx
+++ b/packages/sanity/src/desk/panes/documentList/DocumentListPane.tsx
@@ -59,7 +59,7 @@ export const DocumentListPane = memo(function DocumentListPane(props: DocumentLi
   // Ensure that we use the defaultOrdering value from structure builder if any as the default
   const defaultSortOrder = useMemo(() => {
     return defaultOrdering?.length > 0 ? {by: defaultOrdering} : DEFAULT_ORDERING
-  }, [])
+  }, [defaultOrdering])
 
   const [sortOrderRaw, setSortOrder] = useDeskToolSetting<SortOrder>(
     typeName,
--- a/packages/sanity/src/desk/panes/documentList/DocumentListPane.tsx
+++ b/packages/sanity/src/desk/panes/documentList/DocumentListPane.tsx
@@ -76,7 +76,6 @@ export const DocumentListPane = memo(function DocumentListPane(props: DocumentLi
   const filterIsSimpleTypeContraint = isSimpleTypeFilter(filter)
 
   const {error, fullList, handleListChange, isLoading, items, onRetry} = useDocumentList({
-    defaultOrdering,
     filter,
     params,
     sortOrder,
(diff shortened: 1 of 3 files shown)


bjoerge:
Will this reincarnate #2947?

vicbergquist:
> Will this reincarnate #2947?

aah.. probably. Will look for a fix to make sure it doesn't happen

vicbergquist:
> > Will this reincarnate #2947?
> 
> aah.. probably. Will look for a fix to make sure it doesn't happen

I believe I found a fix. You can test it by adding `.defaultOrdering([{field: 'title', direction: 'asc'}])` on the document list for `Custom books list` in `resolveStructure.ts`, and then using the sort menu in the pane header.

skogsmaskin:
Rebased and set the correct branch for this.

skogsmaskin (approved):

rexxars:
@vicmeow I did a few minor changes, see the last three individual commits and tell me if they look good to you:

1. The default ordering option is no longer used in `useDocumentList`, so I removed the option to avoid confusion
2. The memoization did not account for an update in the desk structure `defaultOrdering` - it _shouldn't_ cause a lot of re-renderings to include it in the dependency array and should ensure we update the sort order on desk structure change
3. For some reason the default sort order code-wise was set to creation date, but as far as I can remember we've always defaulted to last _updated_ - agree?

rexxars (approved):

geball:
I tried testing this branch against test-studio and got some orderings I couldn't explain. For example, Author documents (`/test/content/author`) sorted by Name have a single document out of place at the top, and then appear to sort correctly.
Author documents sorted by 'asdddfxx !' then numbers, and finally by letters.

Book documents (`/test/content/book`) have a custom ordering in the schema, and Sort by Swedish title doesn't seem to be followed.

rexxars:
> For example, Author documents (`/test/content/author`) sorted by Name have a single document out of place at the top, and then appear to sort correctly. 

It is confusing - this is caused by the drafts vs published document: the published document starts with the characters `!!`, while the draft removes these. So it looks like it is wrong because what you're seeing the title of is the draft. This is of course less than ideal, but it is hard to work around since the backend returns all documents regardless of status, and we then have to collate them on the client side - leading to this situation. We could potentially client-side order them once they are collated, but this would lead to inconsistent behavior when scrolling (we're lazy-loading results).

> Book documents (`/test/content/book`) have a custom ordering in the schema, and Sort by Swedish title doesn't seem to be followed.

This has the same issue. There are other "user experience" ordering issues too; if a field starts with a space (or is just whitespace in general), it is sorted to the top of the list, since those characters sort prior to letters. This is usually not what you want (it can be hard to tell that they have leading whitespace).

I still feel like this PR improves on the _current_ situation, and then we can work on getting an improved ordering figured out, although it might require some creative queries or API changes.

geball:
I appreciate that explanation! Thanks @rexxars.

geball (approved):
Looks good, @vicmeow!

PR #3900, opened 2022-11-24
fix: pass `readOnly` to select element

mariuslundgard:
### Description

Pass on `readOnly` property to the `Select` element.

### What to review

Check that read only works as intended for string input with select layout.

### Notes for release

Bugfix: pass `readOnly` to select element


bjoerge (approved):
Good catch, thanks!

Issue #3217, opened 2022-03-25
Reference fields filters in Sanity Studio won't update unless the page is refreshed or the field is hidden and revealed again through conditional rendering. 


MichaelHelgesen:
**Describe the bug**

I have reference fields (**product series** and **product type**) that filters the results based on the input (or no input) in another reference field (**category**)

**Product series** and **product type** are only visible if **category** is set (conditionally rendering). 

When setting the **category** for the first time, the **product series** input is filtered. And when setting the **product series** field for the fist time, the **product type** is filtered. 

But if I change the **product series** input, or clear it, and clear the **product type field**, the **product type** field won't update its filtering. I can see in the JSON that the document is updated, but the Studio reference filtering is not reflecting the change.

If clear the **category** field, so that **product series** and **product type** are hidden, and set a new category, the filtering is correct again. It also gets fixed if I refresh the page.

**To Reproduce**

1. Create a reference field that filters the references based on the input in another reference field.
```
{
          name: 'productType',
          title: "Produkttype",
          type: 'reference',
          to: [{ type: 'productType' }],
          options: {filter: ({document}) => {
            return {
              filter: `${!document.productSeries ? "_type == $type && referenceToProductSeries[0]-> == null" : "_type == $type && references($ref)"}`,
              params: {
                type: "productType",
                ref: !document.productSeries ? "" : document.productSeries._ref
              }
            } 
          }},
          hidden: ({document}) => !document?.subCategory
        },
```
2. Clear the value in the reference field with the filter.
3. Change the value in the reference field the filter is based on.
4. Unless the page was refreshed or the reference field with filter is conditionally rendered, the filter will not register the change in the reference field the filter is based on.

**Expected behavior**

I expected that the ternery filtering in the reference field would update when the field on which the filtering is based on changes.

**Screenshots**

**Which versions of Sanity are you using?**
@sanity/cli              2.8.0 (latest: 2.28.0)
@sanity/base            2.29.2 (up to date)
@sanity/components      2.14.0 (up to date)
@sanity/core            2.28.1 (up to date)
@sanity/default-layout  2.29.2 (up to date)
@sanity/default-login   2.29.2 (up to date)
@sanity/desk-tool       2.29.2 (up to date)
@sanity/vision          2.29.2 (up to date)

**What operating system are you using?**
macOS Monterey

**Which versions of Node.js / npm are you running?**
8.5.3
v16.13.2

**Additional context**
[...]

PR #3901, opened 2022-11-24
fix(form-builder): remove distinctUntilChanged function from handleQueryChange in referenceInput

ninaandal:
### Description

Removed the `distinctUntilChanged` function call from the `handleQueryChange` in the `ReferenceInput` to fix the issue of filtering not being updated when the parent component is changed. 

Link to issue: https://github.com/sanity-io/sanity/issues/3217

### What to review

The book reference dropdown is updated when the decade is changed without the need of a hard refresh. 
ie try to go from 1980 to 10 (there are actual books for these decades to show) 

Screenshot 2022-11-24 at 17 25 50

### Notes for release
fix(form-build): Fix reference filtering updas


bjoerge (approved):
Good catch! Thanks for the thoughtful comment.

PR #3908, opened 2022-11-28
Fix/pte block preview

skogsmaskin:
### Description

There seems to be some legacy code laying around causing confusion and incorrect behaviour in the block previews for the Portable Text Editor.

This will streamline the preview code in there, using the same code for both default and custom previews.

### Note for release
- Fix issue with custom previews inside the Portable Text Editor.


snorrees (approved):
Lgtm!

PR #3916, opened 2022-11-29
feat(core): add warnings when using deprecated schema config

hermanwikner:
### Description

This PR adds warnings when using deprecated schema configs.

### What to review

- Are there any other warnings that should be added?

### Notes for release

Add warnings when using deprecated schema config

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/types/deprecated.ts
+++ b/packages/@sanity/schema/src/sanity/validation/types/deprecated.ts
@@ -1,0 +1,40 @@
+import {SchemaValidationResult} from '../../typedefs'
+import {warning} from '../createValidationResult'
+
+export function validateDeprecatedProperties(type): SchemaValidationResult[] {
+  const warnings = []
+
+  if (type?.inputComponent) {
+    warnings.push(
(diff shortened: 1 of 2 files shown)


packages/@sanity/schema/src/sanity/validation/types/deprecated.ts
```diff
@@ -0,0 +1,43 @@
+import {SchemaValidationResult} from '../../typedefs'
+import {warning} from '../createValidationResult'
+
+export function validateDeprecatedProperties(type): SchemaValidationResult[] {
+  const warnings = []
+
+  // Do not validate document types
+  if (type?.type === 'document') return []
```
snorrees, line 8: Would rather see a comment explaining _why_ we dont validate document types. Cause I was like, why not, here?

snorrees (approved):
These will be super useful when migrating I think!

packages/@sanity/schema/src/sanity/validation/types/deprecated.ts
hermanwikner: Good point

PR #3917, opened 2022-11-30
fix(core): hide style select when there is only one option

hermanwikner:
### Description

This PR fixes so that the block style select in PTE is only visible when there is more than one option.

### What to review

Make sure that the block style select is visible when there is more than one option and hidden when there is only one option.

### Notes for release

Only show the block style select in the portable text editor when there is more than one option


skogsmaskin (approved):
👍

PR #3924, opened 2022-12-01
chore: add studio v3 shopify template

thebiggianthead:
### Description

Update CLI template for Shopify to Studio v3 and re-enable the template in the CLI.

### What to review

Run the CLI template command for the Shopify template:

`sanity init --template shopify`

### Notes for release

Following release, the in-app Shopify instructions need to be updated, as in [this PR](https://github.com/sanity-io/shoebill/pull/40).


skogsmaskin (changes requested):
Seems to be a missing dep. for this template, `pluralize`.

I see you added it, but looks like it needs to be wrapped in `'`

thebiggianthead:
> Seems to be a missing dep. for this template, `pluralize`.
> 
> I see you added it, but looks like it needs to be wrapped in `'`

@skogsmaskin Looks like ESList is removing the `'` on this... I'd be apprehensive to change the ESLint config. Any thoughts or shall I just `eslint-disable-next-line` it?

skogsmaskin:
Oh wait, I might have been reading this wrong, I thought this was part of a template string!

Hmm..then why isn't it building 🤔

skogsmaskin:
Pushed a fix @thebiggianthead !

skogsmaskin:
It builds now, however I'm not able to select the Shopify template when running the CLI with `init`.

The curlpit seems to be here: https://github.com/sanity-io/sanity/blob/feat/v3-shopify-cli-template/packages/@sanity/cli/src/actions/init-project/initProject.ts#L568

After I added it there I'm able to select it.

When making a new Studio based on the template it works, however I have a warning in my studio:

image

skogsmaskin:
Reloading the page (CMD-R) when on a sub-section seems to error with this:

image

skogsmaskin:
> @skogsmaskin I believe the Shopify CLI template has been purposefully left out of the selection in the past and has to be specified with --template shopify when running the command. I'll double check whether we want to retain this behaviour or not.

Ah, yes, that's true, that's how we did it in v2. Nice to check up on that though, as the `ecommerce` template is currently no longer an option in v3.

thebiggianthead:
@skogsmaskin Added the fix for the crashing and updated options array to include Shopify template.

Color schema warning issue is pending: https://sanity-io.slack.com/archives/C03C8362FJP/p1669976070281609

packages/@sanity/cli/templates/shopify/components/hotspots/ProductTooltip.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface HotspotFields {
+  productWithVariant?: {
+    product: {
+      _ref: string
+    }
+  }
+}
+
+export default function ProductPreview({value, renderPreview}: HotspotTooltipProps<HotspotFields>) {
```
hermanwikner, line 13: Nitpick (feel free to ignore): I think this could be more readable like below. Also, we try to avoid using inline styling and prefer using a styled components outside the component scope

```tsx
import {useSchema} from 'sanity'
import {Box} from '@sanity/ui'
import {HotspotTooltipProps} from 'sanity-plugin-hotspot-array'

const StyledBox = styled(Box)`
  width: 200px;
`

export default function ProductPreview(props: HotspotTooltipProps<HotspotFields>) {
  const {value, renderPreview} = props
  const productSchemaType = useSchema().get('product')
  const hasProduct = value?.productWithVariant?.product?._ref && productSchemaType

  const previewProps = useMemo(
    () => ({
      value: value?.productWithVariant?.product,
      schemaType: productSchemaType,
      layout: 'default',
    }),
    [productSchemaType, value?.productWithVariant?.product]
  )

  return (
    <StyledBox padding={2}>
      {hasProduct ? renderPreview(previewProps) : `No product selected`}
    </StyledBox>
  )
}

```

packages/@sanity/cli/templates/shopify/components/inputs/CollectionHidden.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import {WarningOutlineIcon} from '@sanity/icons'
+import {StringFieldProps} from 'sanity'
+import {Box, Card, Flex, Stack, Text} from '@sanity/ui'
+
+export default function CollectionHiddenInput(props: StringFieldProps) {
+  return (
+    <Card padding={4} radius={2} shadow={1} tone="critical">
+      <Flex align="flex-start">
+        <Text size={2}>
```
hermanwikner, line 10: Nitckpick: `size={2}` is the default text size, so it is not really necessary to use here

packages/@sanity/cli/templates/shopify/components/hotspots/ProductTooltip.tsx
thebiggianthead: Leaving this for now as I have other pressing items to attend to and this is triggering some TS errors. May revisit.

packages/@sanity/cli/src/actions/init-project/templates/shopify.ts
```diff
@@ -1,51 +1,69 @@
...
 const shopifyTemplate: ProjectTemplate = {
   configTemplate,
-  importPrompt: 'Add a sampling of sci-fi movies to your dataset on the hosted backend?',
-  datasetUrl: 'https://public.sanity.io/moviesdb-2018-03-06.tar.gz',
   dependencies: {
-    '@sanity/dashboard': '^2.21.7',
+    '@sanity/asset-utils': '^1.3.0',
+    '@sanity/color-input': '^3.0.1',
     'lodash.get': '^4.4.2',
-    'pluralize-esm': '^9.0.2',
```
skogsmaskin, line 44: Is there a reason we are not using `pluralize-esm` anymore?

skogsmaskin: I think we should be using the`pluralize-esm` version here as it's the one we use elsewhere.

skogsmaskin (changes requested):

packages/@sanity/cli/templates/shopify/components/media/ColorTheme.tsx
```diff
@@ -0,0 +1,27 @@
...
+  background?: string
+  text?: string
+}
+
+const ColorThemePreview = (props: Props) => {
+  const {background, text} = props
+
+  return (
+    <div
+      style={{
```
hermanwikner, line 11: Prefer styled component instead of inline styling

```ts
import styled, {css} from 'styled-components'

const StyledSpan = styled.span(({$background}) => {
  const bg = $background || 'white'

  return css`
    align-items: center;
    background-color: ${bg};
    border-radius: inherit;
    display: flex;
    height: 100%;
    justify-content: center;
    overflow: hidden;
    width: 100%;
  `
})

```

packages/@sanity/cli/templates/shopify/components/media/ColorTheme.tsx
```diff
@@ -0,0 +1,27 @@
+type Props = {
```
hermanwikner, line 1: Nitpick: Prefer interfaces unless for a case that cannot be covered by them

skogsmaskin (approved):

packages/@sanity/cli/src/actions/init-project/templates/shopify.ts
```diff
@@ -1,51 +1,68 @@
...
-    slug: '^5.1.0',
-    'sanity-plugin-dashboard-widget-shopify': '^0.1.7',
-    'sanity-plugin-media': '^1.4.3',
+    'sanity-plugin-hotspot-array': '^1.0.0',
+    'sanity-plugin-media': '^2.0.2',
+    slug: '^8.2.2',
+    'styled-components': '^5.2.0',
+    '@types/lodash.get': '^4.4.7',
+    '@types/slug': '^5.0.3',
+    '@sanity/vision': '^3.0.0-rc.2',
```
hermanwikner, line 65: Is this the correct version? I think it should be `^3.0.0`

PR #3933, opened 2022-12-05
feat(cli): add vision as default plugin to templates

hermanwikner:
### Description

This PR adds `@sanity/vision` to all templates + removes the old e-commerce (replaced with shopify) and get-started templates.

### What to review

Make sure that all templates is bootstrapped correctly with the vision plugin.

### Notes for release

Add vision to templates


packages/@sanity/cli/src/studioDependencies.ts
```diff
@@ -4,6 +4,9 @@ export const studioDependencies = {
     // Official studio dependencies
     sanity: '^3.0.0',
 
+    // Official studio plugin dependencies
+    '@sanity/vision': '^3.0.0',
```
snorrees, line 8: This makes it so it ends up in all the templates?

snorrees (approved):

packages/@sanity/cli/src/studioDependencies.ts
hermanwikner: Yes, and that is what we want!

PR #3934, opened 2022-12-06
fix(core): use anchor tag and href on `SearchResultItem`

hermanwikner:


robinpyon:
Thanks @hermanwikner – can't believe I missed this quick win!

PR #3939, opened 2022-12-08
fix(config): respect `unstable_clientFactory`

mariuslundgard:
Brings back support for `unstable_clientFactory`:

```ts
import {createFakeClient} from './fake/sanityClient'

const sanityConfig = defineConfig({
  unstable_clientFactory: () => createFakeClient(),

  projectId: 'test',
  dataset: 'test',
})
```

This makes the Config API support mocking the client without using a test framework like `jest` – but instead through simple dependency injection. This means we can test rendering of Sanity components in UI Workshop (and similar) without having to rely on a connection to the Sanity backend.


rexxars (approved):
Thanks Marius, much appreciated!

PR #3953, opened 2022-12-10
build: add `bin` directory to distribution

mariuslundgard:
### Description

This change adds `bin` to the `files` array in `package.json`.


rexxars (approved):
😱

PR #3958, opened 2022-12-13
chore(server): upgrade to vite 4

rexxars:
### Description

[Vite 4](https://vitejs.dev/blog/announcing-vite4.html) has been released, and we want to make sure we stay on top of changes to our core dependencies. Nothing _too_ exciting in this release, but it does upgrade to Rollup 3, which the new version of pkg-utils is also based on. It thus makes sense for us to keep these in sync, and ensure that any customizations (like our build entry plugin) does not fall behind several releases.

### What to review

This one needs a bit of thorough testing and should probably not be merged until after the holidays. Might also warrant a canary release so we can test it in real-world studios a bit. From the top of my head, we need to make sure that:

- `sanity dev`, `sanity build` and `sanity preview` works as before, for (at least) the given scenarios:
  - hot reloads (in dev)
  - handles CSS module imports (`import styles from 'some.module.css'`)
  - handles regular CSS imports etc (`import 'some.global.css'`)
  - handles url imports (`import {theme} from 'https://themer.sanity.build/...'`)
  - error handling/boundaries work as expected (in dev)

### Notes for release

- Upgraded to Vite 4


mariuslundgard:
Thanks for pushing for this so soon 👏 

I'm getting errors when attempting to run the Test Studio development server.

image

This is caused by newer versions of `esbuild-register` not allowing `register()` to be called more than one. I have had this same issue in other projects (`@sanity/ui-workshop` and `@sanity/tsdoc`), so I quickly wipped up a possible solve here:
https://github.com/sanity-io/sanity/commit/899f54f21000ffa5d9ad02f63b9be75b36bd4f33

If want to include this fix, you can cherry pick these 2 commits:
```
03ef874f6a3e19ee0466c74e32173d243c7f3689
899f54f21000ffa5d9ad02f63b9be75b36bd4f33
```

rexxars:
> This is caused by [newer versions of `esbuild-register` not allowing `register()` to be called more than once per thread](https://github.com/egoist/esbuild-register/issues/82). I have had this same issue in other projects (`@sanity/ui-workshop` and `@sanity/tsdoc`), so I quickly wipped up a possible solve here: [899f54f](https://github.com/sanity-io/sanity/commit/899f54f21000ffa5d9ad02f63b9be75b36bd4f33)

Ah, I mistakenly seem to have forced a new lockfile which has brought in both newer typescript and other dependencies. It's probably wise to uncover these issues anyway though!

Thanks for the fixes, will cherry-pick!

mariuslundgard:
> Might also warrant a canary release so we can test it in real-world studios a bit.

Yep, this is a really good candidate for a canary release.

mariuslundgard (approved):
Great stuff, looks good to me. Once we have released a canary on npm and tested this in a few scenarios, I think it's good to go. From what I gather, the Vite and Rollup bump wasn't huge in terms breaking changes, but best to be on the safe side.

rexxars:
I decided to move the `@sanity/server` module into the `sanity` module, since it is the only place it was ever used. It prevents a (minor and functioning) circular dependency, and reduces maintenance surface since we never really wanted people to use the `@sanity/server` module in their own code.

bjoerge (approved):
Great work - happy to help out testing once the canary is out ✨ 

The etl CI task fails with `error [sanity] missing types path`, but I'm not sure if related to the changes here.

stipsan (approved):
LGTM, I'll help test the canary and report back 🤌

rexxars:
Have released a canary (that is, under the `canary` tag). Seems to be working fine in my preliminary tests.

mariuslundgard:
It indeed works well. Have tested it in a pretty basic Studio, but this works:
- Hot reloading
- CSS modules
- Global CSS
- Importing theme via URL import
- Error boundaries (afaict, only tested with invalid schemas)

Some praise: the Studio is hella snappy! ⚡

3200pro:
Thanks for getting into this so quickly. Is the canary for Vite 4 public?

mariuslundgard:
Hey @3200pro, yes you try it here:

```sh
npm install sanity@canary

# Should update depedency to "sanity": "3.0.7-canary.7"
```

3200pro:
Just a quick update. I tried the canary and I still get the following error:
Screen Shot 2022-12-15 at 3 28 33 PM

PR #3966, opened 2022-12-14
fix(cli): only require workspace name if more than 1 workspace

rexxars:
### Description

When you want to deploy a GraphQL API, we should only require you to define a workspace name if you have multiple defined. There was a check for this, but it mistakenly checks for more than zero instead of more than one. This PR fixes this check.

### What to review

- `sanity graphql deploy` with a single workspace does not throw an error
- `sanity graphql deploy` with multiple workspaces throws an error if no `workspace` name is given in GraphQL config

### Notes for release

- Fix issue where you had to provide a workspace name to deploy GraphQL APIs, even if only a single workspace was defined


skogsmaskin (approved):
👍

PR #3967, opened 2022-12-14
fix(config): throw error early if tools/template resolution fails

rexxars:
### Description

We were attempting to collect/batch as many errors as possible before showing them. However, when resolving tools and templates, we try to access the resolved `tools` property as if we could guarantee it was a valid array. This would fail if (for instance) the `tools` property was an object instead of an array.

This PR changes to throwing early - which I don't think should be a huge step back in any way, and will fix cases where users instead got an error with message `cannot read properties of undefined (reading 'some')` and similar.

### What to review

- Resolving configuration still works as expected
- Defining an object or non-array property for `tools` will give the intended error

### Notes for release

- Fixed error where providing an invalid `tools` configuration property would not provide a readable error

### DIFF
--- a/packages/sanity/src/core/config/prepareConfig.ts
+++ b/packages/sanity/src/core/config/prepareConfig.ts
@@ -284,7 +284,11 @@ function resolveSource({
     // TODO: validate templates
     // TODO: validate that each one has a unique template ID
   } catch (e) {
-    errors.push(e)
+    throw new ConfigResolutionError({
+      name: config.name,
+      type: 'source',
+      causes: [e],
+    })
   }
 
   let tools!: Source['tools']
@@ -297,7 +301,11 @@ function resolveSource({
       reducer: toolsReducer,
     })
   } catch (e) {
-    errors.push(e)
+    throw new ConfigResolutionError({
+      name: config.name,
+      type: 'source',
+      causes: [e],
+    })
   }
 
   // In this case we want to throw an error because it is not possible to have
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/config/prepareConfig.ts
```diff
@@ -284,7 +284,12 @@ function resolveSource({
     // TODO: validate templates
     // TODO: validate that each one has a unique template ID
   } catch (e) {
-    errors.push(e)
+    throw new ConfigResolutionError({
+      name: config.name,
+      // TODO: figure out this name
```
bjoerge, line 289: without any further context it's a bit unclear to me what the actionable thing is re this todo comment

packages/sanity/src/core/config/prepareConfig.ts
```diff
@@ -297,7 +302,12 @@ function resolveSource({
       reducer: toolsReducer,
     })
   } catch (e) {
-    errors.push(e)
+    throw new ConfigResolutionError({
+      name: config.name,
+      // TODO: figure out this name
```
bjoerge, line 307: same as above

bjoerge (approved):
Looks good - just a lil question about a todo comment I found a bit hard to understand

packages/sanity/src/core/config/prepareConfig.ts
rexxars: Copy+pasted from another location, I think we were figuring out the `space` vs `workspace` vs `source` thing. Will remove.

PR #3968, opened 2022-12-14
chore(vision): use react 18 compatible dependencies

rexxars:
### Description

Two `@sanity/vision` related changes:
- Currently has `@sanity/client` and `rxjs` listed as peer dependencies. We are importing types using `import type`, and otherwise not using these modules directly. It is therefore not correct to flag them as peer dependencies. Additionally, studios don't install these dependencies directly, so they will give off warnings about missing dependencies, which is very unfortunate.
- `react-split-pane` does not seem to be maintained anymore, and has react 16 as its supported peer dependency range. This leads to peer dependency warnings when installing the vision plugin. It works just fine with React 18 however, so I have forked the module and updated the peer dependency range to `^18`. Ideally we should replace the library with something different, but I don't have time to address this right now.

### What to review

- That `@sanity/vision` works as expected still (should be very minor changes - if it lets you install dependencies, builds and allows us to resize the different areas of vision, it works fine)

### Notes for release

- Fixed a few peer dependency warnings in `@sanity/vision`

### DIFF
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -27,7 +27,7 @@
     "@sanity/ui-workshop": "^1.0.0",
     "@sanity/util": "3.0.6",
     "@sanity/uuid": "^3.0.1",
-    "@sanity/vision": "3.0.1",
+    "@sanity/vision": "3.0.6",
     "@turf/helpers": "^6.0.1",
     "@turf/points-within-polygon": "^5.1.5",
     "@types/three": "^0.144.0",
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
LGTM! Could also consider replacing the `Subscription` import from rxjs with a locally defined interface with `{unsubscribe: () => void}` which seems to be the only interface we require it to implement

stipsan (approved):

rexxars:
> LGTM! Could also consider replacing the `Subscription` import from rxjs with a locally defined interface with `{unsubscribe: () => void}` which seems to be the only interface we require it to implement

Good point, that's even better.

PR #3971, opened 2022-12-15
Refactor block type render callbacks

skogsmaskin:
### Description

This will introduce a couple of changes to the block type render callbacks.

* Pure block type render callbacks (decorators, styles, lists) now only has a `.component` on it's definition, as opposed to `.compoments.item`. This is because there will be only one, and we better want to differentiate between these and general Studio type render callbacks (like for objects etc).

* The block types `.lists` definition is now validated by the schema validator.

### DIFF
--- a/packages/@sanity/schema/src/sanity/validation/types/block.ts
+++ b/packages/@sanity/schema/src/sanity/validation/types/block.ts
@@ -18,12 +18,14 @@ const allowedKeys = [
   'validation',
 ]
 const allowedMarkKeys = ['decorators', 'annotations']
-const allowedStyleKeys = ['title', 'value', 'components']
-const allowedDecoratorKeys = ['title', 'value', 'icon', 'components']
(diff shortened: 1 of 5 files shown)


packages/@sanity/schema/src/sanity/validation/types/block.ts
```diff
@@ -102,6 +108,48 @@ function validateMarks(marks, visitorContext, problems) {
   return {...marks, decorators, annotations}
 }
 
+function validateLists(lists, visitorContext, problems) {
```
snorrees, line 111: Does this have a unit test?

snorrees (approved):

packages/@sanity/schema/src/sanity/validation/types/block.ts
skogsmaskin: I can't find any existing tests for these validations. We should have them though. But I guess that's outside of the scope here right now.

PR #3973, opened 2022-12-15
fix(block-tools): ignore blocks inside list items (#3492)

juice49:
This fix has [previously been released in v2](https://github.com/sanity-io/sanity/pull/3492). There's some work to add it to v3 in the [chore/sc-21163/cherry-pick-changes-for-v3-on-next](https://github.com/sanity-io/sanity/compare/chore/sc-21163/cherry-pick-changes-for-v3-on-next) branch, but it has not yet been merged.

***

This fixes a bug where p-tags (and other block elements) inside li elements where incorrectly hoisted out of the li element.

With this change blocks are ignored, and block children are added directly to the listItem instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,10 @@ jspm_packages
 # macOS finder cache file
 .DS_Store
 
+# Intellij
+.idea
+*.iml
+
 # VS Code settings
 /.vscode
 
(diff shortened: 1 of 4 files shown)


skogsmaskin (approved):
Nice!

PR #3976, opened 2022-12-15
Refactor PT-input with updated schema types

skogsmaskin:
This is a update to the PT-input with the lastest schema changes from https://github.com/sanity-io/sanity/pull/3971

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
@@ -51,9 +51,8 @@ interface EditorProps {
 }
 
 const renderDecorator: RenderDecoratorFunction = (props) => {
-  const {value, renderDefault, type, focused, selected} = props
-  const CustomComponent = type.components?.item
-  const rendered = renderDefault(props)
+  const {children, value, type, focused, selected} = props
+  const CustomComponent = type.component
   if (CustomComponent) {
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
```diff
@@ -51,9 +51,13 @@ interface EditorProps {
 }
 
 const renderDecorator: RenderDecoratorFunction = (props) => {
-  const {value, renderDefault, type, focused, selected} = props
-  const CustomComponent = type.components?.item
-  const rendered = renderDefault(props)
+  const {children, value, type, focused, selected} = props
+  const CustomComponent = type.component
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const RenderDefault = useCallback(
```
bjoerge, line 57: why do we need to violate the rules of hooks here? Also unclear to me why  `renderDefault` is capitalized, is it intentional?

bjoerge (changes requested):
A couple of comments, and there's a bunch of inline errors that should be sorted out, otherwise LGTM

packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
skogsmaskin: No, this should not be like this 😅

PR #3979, opened 2022-12-15
fix(core): update logo 

hermanwikner:
### Description

This pull request moves the padding from the button that wraps the logo in the navbar to the default logo component. 

**Why?**
Today, if you configure a custom logo component, the navbar often becomes unnecessarily large because you have to render the logo inside a button with padding. By removing the padding from the button that wraps the logo, it is up to the implementation of the custom logo to decide how big the logo/button should be.

### What to review

n/a

### Notes for release

The default button padding surrounding the logo component is now gone. This means that the custom logo component now fully controls the size of the navbar button. Therefore, it may be necessary to update the custom logo component with padding. Below is an example of how to reproduce the same result as before:

```tsx
import {Box} from '@sanity/ui'

function MyLogo() {
  return (
    <Box padding={3}> // <-- Wrap your custom logo with a `Box` with `padding={3}`
      <CustomLogo />
    </Box>
  )
}

export default defineConfig({
  // ... rest of config

  studio: {
    components: {
      logo: MyLogo,
    },
  },
})

```

### DIFF
--- a/dev/test-studio/components/Branding.tsx
+++ b/dev/test-studio/components/Branding.tsx
@@ -1,6 +1,10 @@
-import {Text} from '@sanity/ui'
+import {Box, Text} from '@sanity/ui'
 import React from 'react'
 
 export function Branding() {
-  return <Text weight="bold">Test Studio&trade;</Text>
+  return (
+    <Box padding={3}>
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM!

PR #3983, opened 2022-12-15
feat(core): add new `documentIdEquals` utility method

rexxars:
### Description

Quite often I find myself needing to see if two documents/document IDs are the same, if you disregard their draft status. While you can use `getPublishedId()` manually and compare the two, it is _just_ a bit too repetetive for my taste. This PR adds a new `documentIdEquals()` method which does this comparison. I'm not sure on naming. Other alternatives could be:

- `documentIdsAreEqual()`
- `documentIdsEqual()`
- `documentIdsRepresentSameDocument()`
- `documentIdsAreEqualIfYouDisregardTheirPublishedStatus` (yes that was a joke)

Another alternative would be to just live with having to do the `getPublishedId()` calls manually, but in this case we should at least make the method public (it is currently set as internal).

### What to review

- Naming
- Implementation
- Tests

### Notes for release

- Added a new `documentIdEquals` utility method for checking equality of two document IDs, disregarding the draft/published status

### DIFF
--- a/packages/sanity/src/core/util/draftUtils.test.ts
+++ b/packages/sanity/src/core/util/draftUtils.test.ts
@@ -1,5 +1,5 @@
 import {SanityDocument} from '@sanity/types'
-import {collate, removeDupes} from './draftUtils'
+import {collate, documentIdEquals, removeDupes} from './draftUtils'
 
 test('collate()', () => {
   const foo = {_type: 'foo', _id: 'foo'}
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
I think the naming makes sense. We have used the term "document pair" to represent the `[draft, published]` tuple, but I think it might make sense to think about this tuple as _the_ document in many cases.

PR #3984, opened 2022-12-15
feat(core): add `userHasRole` utility method

rexxars:
### Description

This PR adds a new utility method, `userHasRole()`, which simplifies the task of checking if a user has a given role. I find myself writing code and examples which does the iteration over the `roles` array quite often and while it isn't challenging to write, it is a bit tedious and makes the code a bit harder to read.

### What to review

- Naming
- Implementation
- Tests

### Notes for release

- Added a new utility method, `userHasRole()`, for checking if the current user has a role with a given ID


packages/sanity/src/core/util/userHasRole.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * export function MyComponent() {
+ *   const user = useCurrentUser()
+ *   const hasAdminRole = userHasRole(user, 'administrator')
+ *   return <div>Is administrator: {hasAdminRole ? 'Yes' : 'No'}</div>
+ * }
+ * ```
+ * @public
+ */
+export function userHasRole(user: CurrentUser, roleId: string): boolean {
```
hermanwikner, line 22: Nice! 👍 

Maybe it would be nice to have support for an array of roles as well?

Something like:

```ts
export function userHasRole(user: CurrentUser, roleId: string | string[]): boolean {
  const {roles} = user

  if (Array.isArray(roleId)) {
    return roles.filter((r) => roleId.includes(r.name)).length === roleId.length
  }

  return roles.some((role) => role.name === roleId)
}
```

```ts
const hasRoles = userHasRole(user, ['developer', 'custom-role', 'another-role'])
```

hermanwikner (approved):

skogsmaskin (approved):
Nice! This will be handy. I support Herman's suggestion to be a ble to test for several roles at once.

bjoerge (approved):

packages/sanity/src/core/util/userHasRole.ts
rexxars: It is unclear to me in the array form (without reading the code) whether or not it is "the user needs all of these roles", or if it is "the user needs _one_ of these roles". Maybe it would be better to have separate methods for this use case, which could have more descriptive names?

rexxars: Will merge as-is for now and we can consider expanding the API since it is not a breaking change.

PR #3996, opened 2022-12-19
fix: allow global studio base path

### LINKED ISSUE #3972: basePath not applying during sanity build
**Describe the bug**

Sanity build command doesn’t apply the basePath.

**To Reproduce**

1. Setup a Next JS site with a studio folder in the file structure.
2. Add the /studio basePath to both the sanity.cli.js and sanity.config.js.
3. Run the sanity build command and specify and output path to your /public/studio Next folder.
4. View index.html in /public/studio to observe all reference urls having /static instead of /studio/static.

**Expected behavior**

Upon running ```sanity build ../public/studio -y``` I expect the index.html file to have urls that reference my basePath /studio/static.

**Screenshots**
n/a

@sanity/cli (global)  3.0.6 (up to date)
@sanity/vision        3.0.6 (up to date)
sanity                3.0.6 (up to date)

MacOS Monterey v12.0.

v16.10.0

**Additional context**

I just switched to Studio v3, with a Next site and everything else is working perfectly. The studio is isolated in a folder within the Next files. To build the site I apply the following command from my root package
``` 
"sanity:build": "cd studio && sanity build ../public/studio -y",
``` 
I have the basePath specified sanity.cli.js and sanity.config.js. Both of those files exist within the studio folder and their applications are as follows.

sanity.cli.js
```
  api: {
    projectId: "",
    dataset: "production",
  },
  project: {
    basePath: "/studio",
  },
});
```

sanity.config.js
```
export default defineConfig({
  name: "default",
  title: "",
  projectId: "",
  dataset: "production",
  basePath: "/studio",
  plugins: [deskTool({ structure }), media()],
  schema: {
    types: schemaTypes,
  },
});
```

rexxars:
### Description

There is a `basePath` option in the CLI configuration (`project.basePath`) which is also available through an environment variable (`SANITY_STUDIO_BASEPATH`). This wasn't properly threaded through to all locations, so the built javascript bundles (and preview server) did not respect it properly.

This PR addresses two separate points:

1. Makes sure the configured base path is threaded through to the correct server/build configuration
2. Prepends the configured studio basepath (if any) to the workspace base paths

I'm slightly unsure how I feel about the second point here. I do feel like you would expect the tooling to take it into account if configured, instead of having to manually configure all your workspaces to manually include it? But theoretically we could also leave it up to the user, as we do now. It is a bit of double-configuration, however, and unless you are using an environment variable to configure it, you need to define it more than once.

One could also discuss _where_ the application of the "root path" belongs - I've currently added it to `prepareConfig`, which applies it while normalizing base paths, but theoretically this could also be done earlier in the process - but then we'd need to account for single vs multiple workspaces in several locations, which I was hoping to avoid.

Would love opinions on this! 
 
Fixes #3972

### What to review

- Code
- `sanity build` + `sanity preview` with and without a configured base path (either through `sanity.cli.ts` (`project.basePath`) or through environment variable (`SANITY_STUDIO_BASEPATH`))
- `sanity dev` with and without configured base path

It's probably easiest to review this one commit at a time, since it deals with two slightly different things (build tooling vs automatically applying base path to workspaces)

### Notes for release

- Fixes an issue where a configured studio-wide `basePath` would not be respected in build output, and workspaces would have to manually include the prefix in each workspace base path

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
@@ -1,9 +1,9 @@
 import path from 'path'
 import {promisify} from 'util'
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore This may not yet be built.
 import chalk from 'chalk'
 import rimrafCallback from 'rimraf'
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore This may not yet be built.
 import type {CliCommandArguments, CliCommandContext} from '@sanity/cli'
 import {buildStaticFiles, ChunkModule, ChunkStats} from '../../server'
 import {checkStudioDependencyVersions} from '../../util/checkStudioDependencyVersions'
@@ -75,15 +75,15 @@ export default async function buildSanityStudio(
     spin.succeed()
   }
 
-  const basePath =
+  const specifiedBasePath =
     // Allow `sanity deploy` to override base path
-    overrides?.basePath ||
+    overrides?.basePath ??
     // Environment variables
-    process.env.SANITY_STUDIO_BASEPATH ||
+    process.env.SANITY_STUDIO_BASEPATH ??
     // `sanity.cli.ts`
-    cliConfig?.project?.basePath ||
-    // Fallback
-    '/'
+    cliConfig?.project?.basePath
+
+  const basePath = specifiedBasePath || '/'
 
   spin = output.spinner('Build Sanity Studio').start()
 
--- a/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
@@ -75,6 +75,16 @@ export default async function buildSanityStudio(
     spin.succeed()
   }
 
+  const basePath =
+    // Allow `sanity deploy` to override base path
+    overrides?.basePath ||
+    // Environment variables
+    process.env.SANITY_STUDIO_BASEPATH ||
+    // `sanity.cli.ts`
+    cliConfig?.project?.basePath ||
+    // Fallback
+    '/'
+
   spin = output.spinner('Build Sanity Studio').start()
 
   try {
@@ -82,7 +92,7 @@ export default async function buildSanityStudio(
     const bundle = await buildStaticFiles({
       cwd: workDir,
       outputDir,
-      basePath: overrides?.basePath || cliConfig?.project?.basePath || '/',
+      basePath,
       sourceMap: Boolean(flags['source-maps']),
       minify: Boolean(flags.minify),
       vite: cliConfig && 'vite' in cliConfig ? cliConfig.vite : undefined,
--- a/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
@@ -64,6 +64,28 @@ export default async function buildSanityStudio(
     })
   }
 
+  // Determine base path for built studio
+  let basePath = '/'
+  const envBasePath = process.env.SANITY_STUDIO_BASEPATH
+  const configBasePath = cliConfig?.project?.basePath
+
+  // Allow `sanity deploy` to override base path
+  if (overrides?.basePath) {
+    basePath = overrides.basePath
+  } else if (envBasePath) {
+    // Environment variable (SANITY_STUDIO_BASEPATH)
+    basePath = envBasePath
+  } else if (configBasePath) {
+    // `sanity.cli.ts`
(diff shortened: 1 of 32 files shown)


mariuslundgard:
The following does currently not work in this branch:

1. Configure `project.basePath = '/test/'` in `sanity.cli.ts`
2. Build the studio using `SANITY_STUDIO_BASEPATH="/foo/" sanity build`
3. Start the preview server `sanity start`

In this case the Studio does not render, since it's running on `/test/` but serving assets on `/foo/`.

I would expect defining a environment would override override other config.

--

PS. Would like us to have unit tests to make sure `basePath` and `SANITY_STUDIO_BASEPATH` works as intended, and doesn't unintentionally break in the future.

packages/sanity/src/_internal/cli/actions/build/buildAction.ts
```diff
@@ -75,14 +75,24 @@ export default async function buildSanityStudio(
...
+  const basePath =
+    // Allow `sanity deploy` to override base path
+    overrides?.basePath ||
+    // Environment variables
+    process.env.SANITY_STUDIO_BASEPATH ||
+    // `sanity.cli.ts`
+    cliConfig?.project?.basePath ||
+    // Fallback
+    '/'
+
```
mariuslundgard, line 87: Wondering if `??` would be a better operator than `||` here.

Let's say a user wants to reset the base path using env variable:
```sh
# This might not give the expected result, since it should be "/"
# Though we allow "" since we sanitize the value to make sure it ends with a slash
SANITY_STUDIO_BASEPATH="" sanity build
```

mariuslundgard:
We're also building the Studio in the `sanity deploy` action. We should probably also support `SANITY_STUDIO_BASEPATH="/root/" sanity deploy`?

packages/@sanity/server/src/buildStaticFiles.ts
```diff
@@ -160,3 +160,7 @@ async function writeWebManifest(basePath: string, destDir: string): Promise<void
     .writeFile(path.join(destDir, 'manifest.webmanifest'), content, 'utf8')
     .catch(skipIfExistsError)
 }
+
+function ensureTrailingSlash(urlPath: string) {
+  return urlPath.endsWith('/') ? urlPath : `${urlPath}/`
+}
```
mariuslundgard, line 166: Nitpick: move helper to separate module, .e.g `ensureTrailingSlash.ts`.

packages/@sanity/server/src/buildStaticFiles.ts
```diff
@@ -44,12 +44,12 @@ export async function buildStaticFiles(
...
 
   debug('Writing Sanity runtime files')
-  await writeSanityRuntime({cwd, reactStrictMode: false, watch: false})
+  await writeSanityRuntime({cwd, reactStrictMode: false, watch: false, basePath})
 
   debug('Resolving vite config')
   let viteConfig = await getViteConfig({
     cwd,
-    basePath,
+    basePath: ensureTrailingSlash(basePath || '/'),
```
mariuslundgard, line 52: We should now make `basePath` a require option for `getViteConfig` so we are not maintaing multiple default values.

packages/@sanity/server/src/renderDocument.ts
```diff
@@ -38,6 +38,7 @@ ${generateHelpUrl('custom-document-component')}
 `.trim()
 
 interface DocumentProps {
+  basePath?: string
```
mariuslundgard, line 41: Can this be a require prop now? `basePath: string`

packages/sanity/src/_internal/cli/util/servers.ts
```diff
@@ -77,3 +79,7 @@ function toInt(value: string | number | undefined, defaultValue: number): number
   const intVal = parseInt(`${value}`, 10)
   return Number.isFinite(intVal) ? intVal : defaultValue
 }
+
+function ensureTrailingSlash(path: string) {
```
mariuslundgard, line 83: Reuse same helper from `buildStaticFiles.ts`?

OlaAlsaker:
@rexxars do you have an estimate for when this PR will be merged in? We want to self-host our Sanity Studio, but in a sub-directory. The basePath-bug makes this impossible right now 😅

jeff-overhaul:
> @rexxars do you have an estimate for when this PR will be merged in? We want to self-host our Sanity Studio, but in a sub-directory. The basePath-bug makes this impossible right now 😅

Been anxiously checking this PR status since before Christmas as well

rexxars:
It is likely that it will land in a release next week, but we might push out a canary release you can try before the weekend. I'll keep you updated!

rexxars:
Feel free to try out the `next` tag of the `sanity` and `@sanity/cli` packages if you want to give this a go - it will most likely go out in a proper release early next week.

PR #3997, opened 2022-12-19
chore: deprecate v2 modules

rexxars:
### Description

We now have a range of npm modules that are no longer used as of Sanity Studio v3. While they still serve their original purpose for v2 studios, it is important for us to indicate that as of Sanity Studio v3, you no longer need for instance `@sanity/desk-tool`, since it is now part of the `sanity` module.

This PR adds a deprecation notice to the top of all deprecated `@sanity`-scoped modules that have been removed in the v3 release.

### What to review

- Copy
- Whether or not all removed modules have been covered
- Whether or not I have added the notice to any module that is _not_ deprecated in v3

### Notes for release

None.


packages/@sanity/server/README.md
```diff
@@ -0,0 +1,5 @@
+# ⚠️ THIS PACKAGE IS DEPRECATED
```
hermanwikner, line 1: We still have the server package in version 3 – is it correct to have a deprecation warning for this package?

bjoerge: cc @rexxars

bjoerge (approved):
LGTM! Would be good to have @hermanwikner's comment considered before merging.

packages/@sanity/server/README.md
rexxars: It is going to be removed in the vite 4 PR, so I think it's fine to leave this deprecation warning until then, especially since it is not meant to be used outside of our internals.

PR #4001, opened 2022-12-20
feature(schema): backward compatibility for breaking changes in block…

skogsmaskin:
### Description

This will keep backward compatibility for breaking changes in the block schema introduced in v3 since launch date in order not to break the contract for semantic releases.

It will allow using the old config keys for decorators, styles and annotations, but will produce a warning about them being deprecated.


bjoerge (approved):
Great! Thank you for doing this.

PR #4003, opened 2022-12-20
fix(core): add transition effect when sorting array of primitives

hermanwikner:
### Description

This pull request adds a transition effect when sorting an array of primitives. This fix should be considered a workaround, but right now there is no obvious way to implement this in a more stable way.

### What to review

Make sure that the array of primitives use transition and works as expected

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx
+++ b/packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx
@@ -14,10 +14,24 @@ import {nearestIndexOf} from './utils/nearestIndex'
 import {ItemRow} from './ItemRow'
 import {ArrayOfPrimitivesFunctions} from './ArrayOfPrimitivesFunctions'
 
+interface State {
+  disableTransition: boolean
+}
 // Note: this should be a class component until React provides support for a hook version of getSnapshotBeforeUpdate
 /** @beta */
-export class ArrayOfPrimitivesInput extends React.PureComponent<ArrayOfPrimitivesInputProps> {
+export class ArrayOfPrimitivesInput extends React.PureComponent<
+  ArrayOfPrimitivesInputProps,
+  State
+> {
   _element: HTMLElement | null = null
+
+  constructor(props: ArrayOfPrimitivesInputProps) {
+    super(props)
+
+    this.state = {
+      disableTransition: false,
+    }
+  }
 
   handleAppend = (itemValue: PrimitiveValue) => {
     const {value = [], onIndexFocus, onItemAppend} = this.props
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx
```diff
@@ -19,6 +19,14 @@ import {ArrayOfPrimitivesFunctions} from './ArrayOfPrimitivesFunctions'
 export class ArrayOfPrimitivesInput extends React.PureComponent<ArrayOfPrimitivesInputProps> {
```
bjoerge, line 19: Would be nice to add a type for the component state here:
```suggestion
export class ArrayOfPrimitivesInput extends React.PureComponent<
  ArrayOfPrimitivesInputProps,
  {disableTransition: boolean}
> {
```

packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx
```diff
@@ -154,13 +176,20 @@ export class ArrayOfPrimitivesInput extends React.PureComponent<ArrayOfPrimitive
...
                   gap={1}
                 >
                   {membersWithSortIds.map(({member, id}, index) => {
                     return (
-                      <Item key={member.key} id={id} sortable={isSortable} disableTransition>
+                      <Item
+                        key={member.key}
+                        id={id}
+                        sortable={isSortable}
+                        disableTransition={(this.state as any)?.disableTransition}
```
bjoerge, line 191: with the typing suggestion above you can simplify this (state is never nullish)
```suggestion
                        disableTransition={this.state.disableTransition}
```

packages/sanity/src/core/form/inputs/arrays/common/list.tsx
```diff
@@ -61,22 +61,24 @@ function sortingStrategy(axis: Axis) {
...
-  const {items, axis, onItemMove, children, ...rest} = props
+  const {items, axis, onItemMove, onItemMoveStart, onItemMoveEnd, children, ...rest} = props
 
   const sensors = useSensors(useSensor(PointerSensor), useSensor(KeyboardSensor, SENSOR_OPTIONS))
 
   const handleDragEnd = useCallback(
     (event: DragEndEvent) => {
       const {active, over} = event
 
+      onItemMoveEnd?.()
```
bjoerge, line 72: Feels like`onItemMoveEnd` should be invoked *after* `onItemMove`, so that the order becomes move-start, move, move-end?

bjoerge (changes requested):
Seems to work well! It's a clever workaround and I think we can live with the extra complexity it adds, esp. since it only touches internal code.

Left some comments & suggestions as usual, PTAL :)

packages/sanity/src/core/form/inputs/arrays/common/list.tsx
hermanwikner: Good point 👍

bjoerge (approved):

PR #4006, opened 2022-12-21
fix: apply `config.form.image.directUploads`

stipsan:
### Description

The official [Custom Asset Sources guide for turning off file uploads and only use external sources doesn't work](https://www.sanity.io/docs/custom-asset-sources#8a7448076430:~:text=directUploads%3A%20false%2C). It's because internally we have a typo causing `config.form.file.directUploads` to apply for `image` while it should only apply to `file`.

### What to review

A config like:
```js
import {deskTool} from 'sanity/desk'
import {unsplashAssetSource} from 'sanity-plugin-asset-source-unsplash'
import {schemaTypes} from './schemas'

export default defineConfig({
  name: 'default',
  projectId: '<projectId>',
  dataset: 'production',
  plugins: [deskTool()],
  form: {
    image: {
      assetSources: () => [unsplashAssetSource],
      directUploads: false,
    },
  },
  schema: {
    types: schemaTypes,
  },
})
```
should now disable the Upload button:
image

### Notes for release

- fix: apply `config.form.image.directUploads` correctly

### DIFF
--- a/packages/sanity/src/core/config/prepareConfig.ts
+++ b/packages/sanity/src/core/config/prepareConfig.ts
@@ -491,7 +491,7 @@ function resolveSource({
         directUploads:
           // TODO: consider refactoring this to `noDirectUploads` or similar
           // default value for this is `true`
-          config.form?.file?.directUploads === undefined ? true : config.form.file.directUploads,
+          config.form?.image?.directUploads === undefined ? true : config.form.image.directUploads,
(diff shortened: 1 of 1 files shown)


rexxars (approved):
Oops! Thanks for uncovering and fixing this!

mariuslundgard (approved):

PR #4010, opened 2022-12-21
Chore/upgrade rxjs

bjoerge:
### Description

Opening a draft PR to get preview deployment and to do some testing. Let's pick it up in the new year.


bjoerge:
Continued in #4047

PR #4012, opened 2022-12-22
fix(deps): upgrade to `@sanity/pkg-utils@2`

mariuslundgard:
### Description

- Upgrades `@sanity/pkg-utils` from v1 &rarr; v2
- Sorts `package.json` files using `prettier-plugin-packagejson`
- Adds `exports` maps to all packages
- Adds auto-generated Node ESM wrapper to all exports

### Why?

`@sanity/pkg-utils@2` is ready for Node ESM support.

### DIFF
--- a/packages/groq/package.json
+++ b/packages/groq/package.json
@@ -1,6 +1,5 @@
 {
   "name": "groq",
-  "private": "true",
   "version": "3.1.4",
   "description": "Tagged template literal for Sanity.io GROQ-queries",
   "types": "./lib/dts/src/groq.d.ts",
--- a/packages/groq/package.json
+++ b/packages/groq/package.json
@@ -2,31 +2,6 @@
   "name": "groq",
   "version": "3.1.4",
   "description": "Tagged template literal for Sanity.io GROQ-queries",
-  "types": "./lib/dts/src/groq.d.ts",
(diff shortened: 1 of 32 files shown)


packages/groq/package.json
```diff
@@ -1,6 +1,7 @@
 {
+  "private": true,
+  "type": "commonjs",
   "name": "groq",
-  "private": "true",
```
mariuslundgard, line 3: I changed this from `"private": "true"` to `"private": true`. But seems like this should be removed instead?

bjoerge: not sure what the intention was behind marking it as private, but looks like it could have been a temporary measure (cc @rexxars)

bjoerge (approved):

packages/groq/package.json
mariuslundgard: I removed `"private": true` – which is hardly what we want :-)

mariuslundgard:
@bjoerge – I re-requested your review since I changed this quite a bit. Good news is that this should give us [Node ESM support](https://github.com/sanity-io/sanity/issues/4013) (in theory).

stipsan (approved):
Great work! ✨✨

PR #4015, opened 2022-12-22
Fix fallback render issue with custom PT-styles

skogsmaskin:
### Description

We recently introduced a regression where a custom style in the PT-input  would error if no render component was defined for that custom style. 

Before (v3.0.6) we rendered the Normal style in that case, and we should continue to do that in order to no break this behaviour unnecessary.

Also there was a goof which removed the optionality of the `.component?` schema prop for that style (also relevant for decorators and list items).

Regression commits:
https://github.com/sanity-io/sanity/commit/0902fa59ab0b2648ad1e80a01ba86cfc7e000820
https://github.com/sanity-io/sanity/commit/f1152f740a3427cb4eaeb731402d031dd0377437

### Notes for release

Fixed a regression where a custom style in the portable text input would not fallback to be rendered as the normal style without defining a rendering-component for that style.


bjoerge (approved):
LGTM!

PR #4017, opened 2022-12-22
fix(base): wait for all chunked requests to arrive before continuing

bjoerge:
### Description

Our previous approach of chunking document availability requests did not wait for all the chunks to finish before returning the results. The `debounceCollect` utility expects an observable that returns all results in one go, but because we returned a stream that would emit the chunks one-by-one as they arrived, we would only handle the first emission, which would not include the whole result set, breaking assumptions further down the line.

This patch fixes the issue by reducing over each emission, successively add them to an array. (the `reduce` operator will wait for the stream to finish before emitting a single reduced value).

### What to review
- Check that code looks ok
- Make sure existence validation works as before
- Make sure chunked existence requests are handled properly (this is rather tricky to verify without setting up lots of test data, but I've verified this for the repro case I was looking at, fwiw)

### Notes for release

- Fixes an issue that could give false validation errors for complex documents


rexxars (approved):
Nice find, lgtm

PR #4020, opened 2022-12-23
fix: allow use of `Rule.valueOfField` inside fields using `defineField`

fritz-c:
### Description

Currently, the following schema definition is allowed by TypeScript:
```ts
const product = defineType({
  name: "product",
  type: "document",
  fields: [
    {
      name: "compareAtPrice",
      type: "number",
      validation: (Rule) => Rule.greaterThan(Rule.valueOfField("price")),
    },
    // ...
  ],
})
```

But the following, with only the addition of the `defineField` call, is not allowed*:
```ts
const product = defineType({
  name: "product",
  type: "document",
  fields: [
    defineField({
      name: "compareAtPrice",
      type: "number",
      validation: (Rule) => Rule.greaterThan(Rule.valueOfField("price")),
    }),
    // ...
  ],
})
```
<details><summary><i>* There are cases where that same call is allowed, such as under a `defineConfig` or `definePlugin` call</i></summary>

  ```ts
  const config = defineConfig({
    // ...
    schema: {
      types: [
        {
          type: "document",
          name: "mydoc",
          fields: [
            defineField({
              type: "string",
              name: "mystring",
              validation: (Rule) =>
                Rule.max(Rule.valueOfField("some-other-field")),
            }),
          ],
        },
      ],
    },
  });
  ```

</details>

This PR updates the types used in validation rules such that TypeScript no longer considers use of `Rule.valueOfField()` results as invalid arguments to functions that accept field references. I also updated the signature for `Rule.length` and `Rule.precision` calls to reflect the fact that those can accept field references as well.

### What to review

I added TypeScript test cases to go with each change. You can undo a couple of the type changes locally and see that a type error occurs in the corresponding test case.

### Notes for release

- Fixes type issue where calling `Rule.valueOfField` inside validation defined in `defineField` resulted in type errors.
- The typing for validation rules `Rule.length` and `Rule.precision` now allows referencing values from other fields with `Rule.valueOfField`

### DIFF
--- a/packages/@sanity/types/src/schema/definition/type/array.ts
+++ b/packages/@sanity/types/src/schema/definition/type/array.ts
@@ -1,3 +1,4 @@
+import {FieldReference} from '../../../validation'
 import {RuleDef, ValidationBuilder} from '../../ruleBuilder'
 import {InitialValueProperty, SchemaValidationValue} from '../../types'
 import {IntrinsicDefinitions, TypeAliasDefinition, IntrinsicTypeName} from '../schemaDefinition'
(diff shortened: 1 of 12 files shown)


snorrees (approved):
Tests and all :D Looks good!, thanks!

PR #4038, opened 2023-01-04
fix: field groups between docs with different field names

legowhales:
### Description

Changing the way the active group name is retrieved. As the issue came from this line:

```javascript
const isSelected = group.name === (props.fieldGroupState?.value || defaultGroupName)
```
In some situation, props.fieldGroupState?.value is not undefined but should be (as is it not included in schemaTypeGroupConfig).

See https://github.com/sanity-io/sanity/discussions/3794

### What to review

Field groups are still working as expected.

### Notes for release

Fixing field groups selection between docs with different field names.


bjoerge:
Looking at the code change here it looks like it fixes the symptom rather than the cause. We just merged #4050, which should address the root cause of this. It will be included in the next release. Thank you for your detailed description of the issue and the effort to fix it!

PR #4039, opened 2023-01-04
fix(desk): set the form to read-only when reconnecting

hermanwikner:
### Description

This PR makes the form read-only when reconnecting to avoid unexpected behavior. Since the Studio is not fully functional when offline yet, this ensures that the connection is ready before allowing changes to be made.

This PR also adds missing dependencies in the `useFormState` and `useConnectionState` hooks.

### What to review

- Review the code
- Make sure that it works as expected

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
@@ -317,14 +317,28 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   const isNonExistent = !value?._id
 
   const readOnly = useMemo(() => {
+    const hasNoPermission = !isPermissionsLoading && !permissions?.granted
+    const updateActionDisabled = !isActionEnabled(schemaType!, 'update')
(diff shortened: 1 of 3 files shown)


packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -317,14 +317,28 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   const isNonExistent = !value?._id
 
   const readOnly = useMemo(() => {
+    const hasNoPermission = !isPermissionsLoading && !permissions?.granted
+    const updateActionDisabled = !isActionEnabled(schemaType!, 'update')
+    const createActionDisabled = isNonExistent && !isActionEnabled(schemaType!, 'create')
+    const reconnecting = connectionState === 'reconnecting'
+
```
RitaDias, line 324: very good clearing up of the code to make it more readable 👍

bjoerge: Second this ^

bjoerge (approved):
Great work!

PR #4040, opened 2023-01-04
fix(cli): reintroduce deprecated graphql deployment flags

rexxars:
### Description

This PR reintroduces the `--playground`, `--generation` and `--non-null-document-fields` flags for the `sanity graphql deploy` command. While they ideally should still be configured in `sanity.cli.ts`, it seemed somewhat arbitrary which flags we allowed and did not allow.

With a hopefully clear message during deployment, we will allow using these flags, but indicate that they apply to _all_ defined GraphQL APIs.

While researching this, I also discovered that we were requiring a `projectId` and `dataset` to be defined in the CLI configuration, which is unnecessary as we are extracting this from the studio source configuration anyway. I have thus removed this requirement, which also means you can run `sanity graphql deploy` without a CLI configuration file present - same as in v2.

### What to review

- `sanity graphql deploy` still works (`npx sanity graphql deploy` in `examples/blog-studio` for instance)
- Flags can be used
- Code makes sense

### Notes for release

- Reintroduced the `--playground`, `--generation` and `--non-null-document-fields` flags for `sanity graphql deploy`
- Fixed an issue where a CLI configuration file was _required_ in order to deploy a GraphQL API

### DIFF
--- a/packages/@sanity/cli/test/graphql.test.ts
+++ b/packages/@sanity/cli/test/graphql.test.ts
@@ -25,7 +25,7 @@ describeCliTest('CLI: `sanity graphql`', () => {
       expect(result.stdout).toContain('https://')
       expect(result.code).toBe(0)
 
-      const [graphqlUrl] = result.stdout.match(/(https:\/\/.*(\s|$))/) || []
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
LGTM!

PR #4042, opened 2023-01-05
fix(core): add default media to new document dialog previews

hermanwikner:
### Description

This PR adds a default icon to the items in the new document dialog. This is a regression in v3.

### What to review

n/a

### Notes for release

n/a


bjoerge (approved):
👍🏼

PR #4046, opened 2023-01-05
feat(server): support studio env vars through `process.env`

rexxars:
### Description

Supersedes #3620 - the previous approach warned when using `process.env`, this one actually defines the environment variables as `process.env.SANITY_STUDIO_X`.

Rationale:
- There is no standard for environment variables across environments.
  - Node.js uses `process.env`.
  - Vite uses `import.meta.env`.
  - Deno uses `Deno.env`.
  - Cloudflare workers uses global variables.
- Sanitys tooling runs on Node.js. It is difficult/annoying to write things like configuration files, where they need to work on both the server/cli and on the client side.
- By defining these as `process.env.SANITY_STUDIO_X`, they are both accessible as `import.meta.env.SANITY_STUDIO_X` _and_ under `process.env`. This means that you can use the `process.env` approach in configuration files and (if you want/need to) use `import.meta.env` in code.

We should still encourage people to use environment variables only on the very "edge" of an application, eg in configuration, instead of deeply inside project code. This makes it easier to switch between environments where `process.env` might not be available.

[sc-19342]

### What to review

- See that a defined environment variable is accessible through `process.env`
  - Define a studio env var (`export SANITY_STUDIO_SOMETHING=yep`)
  - Use the env var in studio code - eg in a studio config or similar (`process.env.SANITY_STUDIO_SOMETHING`)
  - Use the env var in studio code from `import.meta.env`
- Ensure only `SANITY_STUDIO_` prefixed variables are accessible

### Notes for release

- Allow accessing studio-prefixed environment variables through `process.env`

### DIFF
--- a/packages/sanity/src/_internal/cli/server/getStudioEnvVars.ts
+++ b/packages/sanity/src/_internal/cli/server/getStudioEnvVars.ts
@@ -1,0 +1,20 @@
+/* eslint-disable no-process-env */
+
+/**
+ * Get environment variables prefixed with SANITY_STUDIO_, as an object.
+ * Specify a `prefix` to add a prefix to the environment variable keys,
+ * eg: `getStudioEnvVars('process.env.')`
+ *
+ * @param prefix - Prefix to add to the environment variable keys
+ * @returns Object of studio environment variables
+ * @internal
+ */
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
A very pragmatic and sensible solution. Great work!

PR #4049, opened 2023-01-06
Fix duplication of initial block in empty PTE

skogsmaskin:
### Description

There is an issue in the PTE where if you have editor A and editor B, and A deletes all the content, a raise condition may occur:

If client B receives the deletion, and afterwards client A writes in new content - this initial block is duplicated in client B's editor and we risk a duplicate keys error in all clients as the duplicated blocks are already propagated to the document store through patching being done by client B.

The reason this happens is because we had an invalid test for equality between the server value and the editor state in the case where the editor value is longer (array) than the server value. To add to the problem there are also side effects by creating the placeholder block (a non existing block for the user to type into in an empty editor) by text operations which will produce insert patches.

The solution is to fix the equality test defect,  and to create the placeholder block without using operations  (but rather just put it straight to the editor's children) - then patch from anything there  (as opposed to creating that placeholder block by actual transformations / patches).

Related to this I also found an issue with the selection for a local client is not being reset properly when the document is emptied remotely by another client.

I also moved the stored selection in PortableTextEditor  to a ref instead, as we potentially want to use it even in a unmounted component state (we hade some occasional React dev warnings about this debugging the above issue).

### What to review

That the duplication error is no longer reproducible.

### Notes for release
* Fixes a potential raise condition with duplicate keys when an editor deletes all the content of a portable text editor, while other clients are within that document, and then starts to type new content.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/PortableTextEditor.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/PortableTextEditor.tsx
@@ -32,7 +32,7 @@ import {
 } from '../types/editor'
 import {validateValue} from '../utils/validateValue'
 import {debugWithName} from '../utils/debug'
-import {getValueOrInitialValue, isEqualToEmptyEditor, toSlateValue} from '../utils/values'
+import {getValueOrInitialValue, toSlateValue} from '../utils/values'
 import {KEY_TO_SLATE_ELEMENT, KEY_TO_VALUE_ELEMENT} from '../utils/weakMaps'
 import {FLUSH_PATCHES_DEBOUNCE_MS} from '../constants'
 import {PortableTextEditorContext} from './hooks/usePortableTextEditor'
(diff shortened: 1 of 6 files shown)


packages/@sanity/portable-text-editor/src/editor/PortableTextEditor.tsx
```diff
@@ -255,9 +256,11 @@ export class PortableTextEditor extends React.Component<
     }
     // Sync value from props, but not when we are responding to incoming patches
     // (if this is the case, we sync the value after the incoming patches has been processed - see createWithPatches plugin)
+    const isPristineEditor =
+      !prevProps.value && this.slateInstance.children === this.state.initialSlateEditorValue
     if (
       this.props.value !== prevProps.value &&
-      (!prevProps.value || this.readOnly || !this.props.incomingPatches$)
```
skogsmaskin, line 260: This logic was wrong, because prevProps.value could be `undefined` when someone empties the editor.

We want this to only be true on the initial load when in a collaborative setting.

skogsmaskin (commented):
Added some comments.

ninaandal (approved):
🥳

PR #4052, opened 2023-01-06
feat(cli): merge user vite config object

nkgentile:
### Description

- (Fix?) Respect user configuration when running `sanity preview`
- Allow user to pass configuration object that will be merged into default Vite configuration
- Align user configuration function with Vite's API by [passing context as second argument](https://vitejs.dev/config/#conditional-config) and accept a Promise:
```ts
import {defineCliConfig} from 'sanity/cli'
import {mergeConfig} from 'vite'

  // ...rest of config

  async vite(config, {command, mode}){
    const target = await getProxyTarget()

    return mergeConfig(config, {
      server: {
        proxy: {
          "/api": {
            target,
            changeOrigin: true,
            secure: false,
            rewrite: (path) => path.replace(/^\/api/, ""),
          },
        },
      },
    })
  }
})
```

### What to review

1. Bootstrap a new Sanity Studio v3 project, with TypeScript
2. In `sanity.config.ts` add a `vite` property to check Intellisense and add:
```ts
import {defineCliConfig} from 'sanity/cli'

  // ...rest of config

  vite: {
    server: {
      open: true
    }
  }
})
```

### DIFF
--- a/packages/@sanity/cli/src/types.ts
+++ b/packages/@sanity/cli/src/types.ts
@@ -3,6 +3,7 @@ import type chalk from 'chalk'
 import type {Ora} from 'ora'
 import type {SanityClient} from '@sanity/client'
 import type {Separator, DistinctQuestion, Answers, ChoiceCollection} from 'inquirer'
+import type {InlineConfig} from 'vite'
 import type {ClientRequirements} from './util/clientWrapper'
 import type {CliConfigResult} from './util/getCliConfig'
 import type {CliPackageManager} from './packageManager'
@@ -284,6 +285,5 @@ export interface CliConfig {
 
   graphql?: GraphQLAPIConfig[]
 
-  // @todo
-  vite?: (config: any) => any
+  vite?: InlineConfig | ((config: InlineConfig) => InlineConfig)
 }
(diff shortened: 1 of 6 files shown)


nkgentile:
In the example above:
```ts
import {defineCliConfig} from 'sanity/cli'
import {mergeConfig} from 'vite'

  // ...rest of config

  async vite(config, {command, mode}){
    const target = await getProxyTarget()

    return mergeConfig(config, {
      server: {
        proxy: {
          "/api": {
            target,
            changeOrigin: true,
            secure: false,
            rewrite: (path) => path.replace(/^\/api/, ""),
          },
        },
      },
    })
  }
})
```
I had to add `vite` as a `devDependency` to use `mergeConfig`, but I wonder if the function signature could provide it in the arguments somewhere? 🤔

rexxars:
I really want to land this, but it would mean having to ship vite as a non-dev dependency in `@sanity/cli` in order for the types to ship. Would love to find a solution to this, but don't currently have the bandwidth. Any thoughts?

nkgentile:
Is it a no-no to add the missing exports in `vite.d.ts` and then rely on declaration merging? Worst case, a user only has `@sanity/cli` installed and they don't get autocomplete in `defineCliConfig`?

rexxars:
> Is it a no-no to add the missing exports in `vite.d.ts` and then rely on declaration merging? Worst case, a user only has `@sanity/cli` installed and they don't get autocomplete in `defineCliConfig`?

Hmm, no, that seems like a decent enough option.

rexxars (approved):

PR #4064, opened 2023-01-10
fix(desk): improve saving indicator

hermanwikner:
### Description

This PR fixes so that the "saved" status in the document footer can only be set when the document is not syncing. This is done by using the boolean returned from the `useSyncState` hook when deciding whether to allow the state to be set to "saved". There have been reports of the "saved" status being displayed even though the document was not actually saved due to poor network connectivity.

### What to review

- Make sure that the document status indicator works as before

### Notes for release

n/a


bjoerge (approved):
LGTM!

PR #4066, opened 2023-01-10
Fix focus navigation issue inside of block in PTE

RitaDias:
### Description

This fixes the tab navigation focus inside of a block within PTE

### What to review

First test:
1. In the test studio go to "Block Test"
2. Pick a document and scroll down to `With geopoints` portable text editor
3. Add geographical point -> let the modal open
4. Use tab to navigate the inputs and make sure that you can and it doesn't escape to the PTE beneath

https://user-images.githubusercontent.com/6951139/211536689-bf96256d-5046-4598-9e3b-39c8605ddbd2.mov

Second test:
1. 1. In the test studio go to "Block Test"
2. Pick a document and scroll down to `With geopoints` portable text editor
3. Add geographical point -> close the modal
4. Single click on the geopoint component (select it, but don't open it)
5. Use tab to navigate through the menu items

https://user-images.githubusercontent.com/6951139/211536988-b69a1ff5-a701-42c4-b9ef-ea6d2cb4d2fc.mov

### Notes for release

Fix focus navigation when in a component within Portable Text Editor

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
@@ -115,6 +115,7 @@ export function BlockObject(props: BlockObjectProps) {
         onClickingEdit={openItem}
         readOnly={readOnly}
         renderPreview={renderPreview}
+        isOpen={memberItem?.member.open}
         type={type}
         value={block}
       />
--- a/packages/sanity/src/core/form/inputs/PortableText/object/BlockObjectPreview.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/BlockObjectPreview.tsx
@@ -32,6 +32,7 @@ interface BlockObjectPreviewProps {
   renderPreview: RenderPreviewCallback
   type: ObjectSchemaType
   value: PortableTextBlock
+  isOpen?: boolean
 }
 
 const POPOVER_PROPS: MenuButtonProps['popover'] = {
@@ -51,6 +52,7 @@ export function BlockObjectPreview(props: BlockObjectPreviewProps): ReactElement
     renderPreview,
     type,
     value,
+    isOpen,
   } = props
   const {isTopLayer} = useLayer()
   const editor = usePortableTextEditor()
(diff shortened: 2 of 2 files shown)


ninaandal (approved):
lgtm ✨

packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
```diff
@@ -115,6 +115,7 @@ export function BlockObject(props: BlockObjectProps) {
         onClickingEdit={openItem}
         readOnly={readOnly}
         renderPreview={renderPreview}
+        isOpen={memberItem?.member.open}
```
RitaDias, line 118: Even when the default is "undefined" it will still translate as false which is fine because this component will update once the member is finally opened / closed.

skogsmaskin: You could do a `!!memberItem?.member.open` to keep it all `Boolean`.

skogsmaskin, line 118: `memberItem?.member.open` must be part of the hook dependency below.

I suggest you do a:

`const isOpen = !!memberItem?.member.open` above this, and then use `isOpen` as hook dependency and variable inside this hook.

packages/sanity/src/core/form/inputs/PortableText/object/BlockObjectPreview.tsx
```diff
@@ -85,7 +87,7 @@ export function BlockObjectPreview(props: BlockObjectPreviewProps): ReactElement
           PortableTextEditor.focus(editor)
         }
         if (event.key === 'Tab') {
-          if (menuButton.current && !isTabbing.current) {
+          if (menuButton.current && !isTabbing.current && !isOpen) {
```
skogsmaskin, line 90: `isOpen` must be part of the hook dependency below.

skogsmaskin (changes requested):

skogsmaskin:
I think we also need to reset the `isTabbing` ref. when `isOpen` changes to false again. Or it will only work once.

Inside BlockObjectPreview:

```
  useEffect(() => {
    if (isOpen) {
      isTabbing.current = false
    }
  }, [isOpen])
```

packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
RitaDias: You're right, weird that my linting didn't complain about the missing dependency 🤔 
I'll update these 👍

skogsmaskin: Yeah, maybe you have to restart the TS server inside VSCode.

skogsmaskin (approved):
Seems to work great now!

PR #4068, opened 2023-01-10
fix: rename various npm scripts publish => release

bjoerge:
### Description
Currently, running e.g. `yarn publish-current-branch` will trigger the `publish` script after packages has been successfully published (as a lifecycle script), and cause the whole release process to fail.

Not 100% sure why this happens and why it suddenly has become a problem. (I suspect it might always have been the behavior, but effectively been a noop before and now causing errors e.g. after some external tooling updates).

This PR addresses the issue by renaming the various `publish`-scripts to `release` (which I guess is sort of more accurate as well since it does more than a mere publish).

### What to review
- Apart from our internal release guides – do we know about any other places (e.g. code sites) that needs to be updated accordingly?

### Notes for release

n/a - internal only


mariuslundgard (approved):
Looks good to me!

PR #4070, opened 2023-01-10
refactor(core): deprecate WithReferringDocuments HOC in favor of a new hook

bjoerge:
### Description
This deprecates the WithReferringDocuments HOC and replaces it with a `useReferringDocuments()` hook. I've kept the old HOC around since removing it would strictly be a breaking change (although it's marked as internal and not visible to the type system).

With this change we no longer use the `react-props-stream` package, so I've removed that as a dependency too.

I've marked the new `useReferringDocuments()` hook as `@internal` for now as we should probably have a quick discussion about what it's public API should be.

### What to review
- Make sure all the places we display the list of referring documents still works as before (delete document confirmation dialog and the asset usage dialog)

### Notes for release
n/a


RitaDias (approved):
Everything looks as expected and seems to work exactly as we want it 👍

PR #4087, opened 2023-01-12
fix(core): improve read only visibility of tags array

hermanwikner:
### Description

This PR improves the read only visibility of the tags input array.
tags

### What to review

- Does the UI look good?

### Notes for release

Improve read only visibility of tags array input


mikolajdobrucki (approved):
the changes are good to go, however:

the colour contrast of the text elements in tags in read only state does not meet AA standards by WCAG. I would we create another story for that and treat it as a separate issue.

PR #4090, opened 2023-01-12
fix(form/inputs): fix issues with custom style component rendering 

skogsmaskin:
### Description

The background for this PR is that we needed to wrap all text content inside the Portable Text Editor as @sanity/ui text for them to render properly in the editor. When a Sanity/UI Text element was missing, the text would be impossible to put the cursor in by the mouse, and the text were hidden behind the error/warning overlay when those were displayed.

We can not require that the implementor of a custom style component knows about, or should need to care about using Sanity/UI texts in their custom components. Currently we have worked our way around t his by wrapping the custom component in a Text. But this causes issues the moment you would like to use like a Sanity/UI `Heading` in your custom component (`Text` takes presedence).

The problem turned out to be an issue with the z-index on the text block error/warning overlays (thanks @hermanwikner).
By removing the Text workaround and specifying a z-index on the overlay, everything is now working as it should. 

Notes for release:
* Fixes a styling issue with custom style components for text blocks in the Portable Text Input.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/text/textStyles.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/text/textStyles.tsx
@@ -10,6 +10,12 @@ export const TextContainer = styled.div`
   display: block;
 `
 
+/**
+ * Component for wrapping children inside a user's custom style component.
+ * This ensures that we have a TextContainer there (see comment above).
+ */
+export const Custom = ({children}: TextStyleProps) => <TextContainer>{children}</TextContainer>
+
 export const Normal = ({children, ...rest}: TextStyleProps) => (
   <Text data-testid="text-style--normal" {...rest}>
     <TextContainer>{children}</TextContainer>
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/PortableText/text/textStyles.tsx
```diff
@@ -10,6 +10,12 @@ export const TextContainer = styled.div`
   display: block;
 `
 
+/**
+ * Component for wrapping children inside a user's custom style component.
+ * This ensures that we have a TextContainer there (see comment above).
+ */
+export const Custom = ({children}: TextStyleProps) => <TextContainer>{children}</TextContainer>
```
hermanwikner, line 17: Since `TextContainer` is already exported – could we remove this one and instead use `TextContainer` in `Style.tsx`?

hermanwikner (approved):
Looks good to me! 👍

packages/sanity/src/core/form/inputs/PortableText/text/textStyles.tsx
skogsmaskin: I was giving this some thoughts yesterday and I wouldnt t leak that logic with the android issue out other places. Conceptually it is easier to deal with various styles. Internally we use Custom when user style.

PR #4099, opened 2023-01-17
fix(cli): exit if user does not confirm v3 init continuation

rexxars:
### Description

When running `sanity init` with the v3 CLI, you will be prompted with a warning indicating you're following v2 installation instructions, and a confirmation prompt will ask the user whether or not to continue initing with a v3 studio. When answering "no" on this dialog, the CLI still continues the init process as if answering yes, however.

This PR fixes this issue, and also adjusts the border look of the message/warning dialog to match branding guideline. Also reduces the amount of nested code - apologies for hard to read diff, but code shouldn't be too hard to follow.

### What to review

- `sanity init` will show prompt
  - answering no will exit
  - answering yes will continue

### Notes for release

- Fixed an issue where answering "no" when being asked whether or not to create a new v3 project would not actually stop the init process


skogsmaskin (approved):
Great to have this looked over. This was something we had to rush in there right before the release,  so execution might not have been the best :)

bjoerge (approved):

PR #3628, opened 2022-09-12
fix(desk-tool): add full timestamp tooltip to timeline items

sjelfull:
### Description

This adds the full change timestamp to a tooltip when you hover over the relative time stamp in the version timeline.

https://user-images.githubusercontent.com/10508/190409236-557b0d86-1115-448e-b9ed-32e36a080bad.mp4

Closes #3251

### What to review

- Is this the best UI solution?

### Notes for release

- Added full timestamp to version timeline

PR #4100, opened 2023-01-17
fix(desk): add full timestamp tooltip to timeline items

robinpyon:
### Description

V3 port of #3628 with a few minor differences:
- Tooltips are now correctly nested inside (and not around) `MenuItem` components
  - This was previously causing an issue where tooltips were appearing immediately when the Menu was opened, and other odd side-effects (tooltip not appearing on hover)
- History timeline menu can now be dismissed with ESC
- Added workshop story

### Notes for release

- The document history timeline now displays full dates in tooltips on hover

### DIFF
--- a/packages/sanity/src/desk/panes/document/timeline/__workshop__/DefaultStory.tsx
+++ b/packages/sanity/src/desk/panes/document/timeline/__workshop__/DefaultStory.tsx
@@ -1,0 +1,54 @@
+import React, {useMemo} from 'react'
+import {Box, Card, Inline, Stack, Text} from '@sanity/ui'
+import {DeskToolProvider} from '../../../../DeskToolProvider'
+import type {DocumentPaneNode} from '../../../../types'
+import {DocumentPaneProvider} from '../../DocumentPaneProvider'
+import {TimelineMenu} from '../timelineMenu'
+
+const DOCUMENT_ID = 'test'
+const DOCUMENT_TYPE = 'author'
+
+export default function DefaultStory() {
+  const pane: DocumentPaneNode = useMemo(
+    () => ({
+      id: DOCUMENT_ID,
+      options: {
+        id: DOCUMENT_ID,
+        type: DOCUMENT_TYPE,
+      },
+      type: 'document',
+      title: 'Workshop',
+    }),
+    []
+  )
+
+  return (
+    <DeskToolProvider>
+      <DocumentPaneProvider index={0} itemId={DOCUMENT_ID} pane={pane} paneKey={DOCUMENT_ID}>
+        <Box padding={2}>
+          <Stack space={2}>
+            <Card padding={3} shadow={1} tone="primary">
+              <Stack space={3}>
+                <Inline space={1}>
+                  <Text size={1} weight="medium">
+                    Document ID:
+                  </Text>
+                  <Text size={1}>{DOCUMENT_ID}</Text>
+                </Inline>
+                <Inline space={1}>
+                  <Text size={1} weight="medium">
+                    Document Type:
+                  </Text>
+                  <Text size={1}>{DOCUMENT_TYPE}</Text>
+                </Inline>
+              </Stack>
+            </Card>
+            <Card padding={2}>
+              <TimelineMenu chunk={null} mode="rev" />
+            </Card>
+          </Stack>
+        </Box>
+      </DocumentPaneProvider>
+    </DeskToolProvider>
+  )
+}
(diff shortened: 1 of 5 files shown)


RitaDias (approved):
This looks good to me 🎉

packages/sanity/src/desk/panes/document/timeline/__workshop__/DefaultStory.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {DeskToolProvider} from '../../../../DeskToolProvider'
+import type {DocumentPaneNode} from '../../../../types'
+import {DocumentPaneProvider} from '../../DocumentPaneProvider'
+import {TimelineMenu} from '../timelineMenu'
+
+const DOCUMENT_ID = 'test'
+const DOCUMENT_TYPE = 'author'
+
+export default function DefaultStory() {
+  const pane: DocumentPaneNode = useMemo(
```
hermanwikner, line 12: I think `useMemo` is a bit unnecessary here since this is a static object. Could be moved to a `const PANE` outside the component scope instead.

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
```diff
@@ -54,7 +57,17 @@ export function TimelineMenu({chunk, mode, placement}: TimelineMenuProps) {
     handleClose()
   }, [handleClose])
 
+  const handleGlobalKeyDown = useCallback(
```
hermanwikner, line 60: In addition to `Escape` – the menu should be closed also when pressing `Tab`.  Also, we should set the focus back on the button when closing the menu.

I suggest making this update to the `handleGlobalKeyDown` function:

```ts
  const handleGlobalKeyDown = useCallback(
    (event: KeyboardEvent) => {
      if (open && (isEscape(event) || event.key === 'Tab')) {
        handleClose()
        button?.focus()
      }
    },
    [handleClose, button, open]
  )

```

packages/sanity/src/desk/panes/document/timeline/timelineItem.tsx
```diff
@@ -66,34 +72,48 @@ export function TimelineItem(props: {
       data-selection-top={isSelectionTop}
       onClick={handleClick}
     >
-      <div
-        // eslint-disable-next-line react/jsx-no-bind
-        onMouseEnter={() => setHovered(true)}
-        // eslint-disable-next-line react/jsx-no-bind
-        onMouseLeave={() => setHovered(false)}
+      <Tooltip
```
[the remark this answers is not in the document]
robinpyon: This is what the original PR did ([example](https://test-studio-git-bug-sc-24090document-version-history-tim-053d79.sanity.build/test/desk/untitledRepro;grrm)), though it does cause a few unintended side-effects – e.g. upon clicking the button, the selected tooltip will appear immediately, and then hovering over that item will cause it to disappear). 

Fair point on the tooltip not appearing when focusing via the keyboard – do you know of any working examples of Menu/MenuItems that utilise Tooltips correctly?

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
robinpyon: Great pickup re: re-focusing!

Re: dismissing on tab, it's worth mentioning that `MenuButton` doesn't handle this behaviour either. Will add, but we should consider making that the default behaviour of `MenuButton` as well

hermanwikner: `MenuButton` does actually close the menu when pressing tab, see [this Arcade](https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJytkDEOwyAMRXdOYXEBlo7A0KFqhx4irRxCRUjVOlIilLvXECTUrUO3%2Fx%2F%2FWzb6FHARAK57mnTYWA3o3UBG9j4Eyb4L3kUj7xgJXxk85jf5fm3IMtRXjPNxJpoiO4BbkSbpnQHhwjNbSIKyW0mOzEwqso6x1VR7IRxr%2F4whTLn6Q%2BI%2Fkfqu2l556737fVu7S2iVf9WKD1kcYv4%3D)!

robinpyon: It doesn't for me 😓, neither in the studio either – let's address separately!

PR #4101, opened 2023-01-18
refactor(core): remove commented code and deal with "todo" config error screen

RitaDias:
### Description

Cleaning up unused code that was related to a error config message.
This no longer happens as the studio lets you know what has gone wrong. Since we don't use it and the code is commented, I think it's best that we remove it!

### What to review

Double check that the removal of the code is clean and won't affect other parts of the studio

### DIFF
--- a/packages/sanity/src/core/studio/screens/ConfigErrorsScreen.tsx
+++ b/packages/sanity/src/core/studio/screens/ConfigErrorsScreen.tsx
@@ -8,6 +8,7 @@ import React from 'react'
 // `
 
 export function ConfigErrorsScreen() {
+  /* This screen is not currently being used anywhere. We're keeping it as a basis for future work */
   return <>TODO: implement config errors screen</>
   // return (
   //   <Root forwardedAs={Flex}>
--- a/packages/sanity/src/core/studio/workspaceLoader/WorkspaceLoader.tsx
+++ b/packages/sanity/src/core/studio/workspaceLoader/WorkspaceLoader.tsx
@@ -158,7 +158,7 @@ function WorkspaceLoaderBoundary({ConfigErrorsComponent, ...props}: WorkspaceLoa
   // TODO: implement this
   // const errors = useMemo(() => flattenErrors(error, []), [error])
 
-  //TODO: implement config error screen
+  //TODO: implement config error screen - a story has been created for this
   // if (error instanceof ConfigResolutionError) return <ConfigErrorsComponent />
 
   // otherwise hand off to other boundaries


packages/sanity/src/core/studio/screens/ConfigErrorsScreen.tsx
```diff
@@ -1,44 +0,0 @@
...
-  //       <Button
-  //         tone="primary"
-  //         // eslint-disable-next-line react/jsx-no-bind
-  //         onClick={() => window.location.reload()}
-  //         type="button"
-  //         text="Retry"
-  //       />
-  //     </Content>
-  //   </Root>
-  // )
```
RitaDias, line 43: I just noticed that this is actually used on StudioProvider. While I wasn't able to trigger this message to appear in the way that the issue described it, I feel that then the best approach here would be to put the interface that is commented out instead? 

Because the TODO just looks bad :/ Unless we have a different approach that was left along the way? @hermanwikner

hermanwikner: This screen can't be displayed at this time because it's commented out here, and if there's a configuration error we currently fall back to the default error screen: https://github.com/sanity-io/sanity/blob/5abfeccd0e7d861c4a52eb7e6103d9b3cd710b11/packages/sanity/src/core/studio/workspaceLoader/WorkspaceLoader.tsx#L162

The commented JSX might work but I think it needs a bit more work and design to match other error screens we have! I like that we clean up unused code but we should implement this screen at some point, so I think it might be nice to still keep this code until we start that work :)

RitaDias: I'm wondering if it can't though :/ 
Because when I looked up [ConfigErrorsScreen.tsx](https://github.com/sanity-io/sanity/pull/4101/files/41d435e79070b926f9516227ad9648efda497905#diff-c27b006e5e8002d9e4bc0214c11e2ff00166b6ece3607618d7b23504d78d86da) I found it being used in the `StudioProvider`.

https://github.com/sanity-io/sanity/blob/053e47b66d118cb7952b92e500d7e051d96a3105/packages/sanity/src/core/studio/StudioProvider.tsx#L75

I haven't been able to replicate this appearing naturally while using the product but this screen appearing right now in any instance would look really bad. Do you know if it's a mistake that this screen is being used and something else should be used instead?

hermanwikner: The `ConfigErrorsScreen` component is passed to the `WorkspaceLoader`'s prop called `ConfigErrorsComponent`. But `WorkspaceLoader` do not render that component (since it is commented out) – so it can't be rendered currently!

RitaDias: Sigh of course 🤦 
I don't even have an excuse for missing this 😂 

Great, then I think I'll add a comment on [ConfigErrorsScreen.tsx](https://github.com/sanity-io/sanity/pull/4101/files/41d435e79070b926f9516227ad9648efda497905#diff-c27b006e5e8002d9e4bc0214c11e2ff00166b6ece3607618d7b23504d78d86da) just about the plan to eventually clean it up and create a shortcut task :)

RitaDias: @hermanwikner created a story on shortcut and added a comment for more visibility. if you're good with it we can go with a ✅

packages/sanity/src/core/studio/workspaceLoader/WorkspaceLoader.tsx
```diff
@@ -158,9 +158,6 @@ function WorkspaceLoaderBoundary({ConfigErrorsComponent, ...props}: WorkspaceLoa
   // TODO: implement this
   // const errors = useMemo(() => flattenErrors(error, []), [error])
 
-  //TODO: implement config error screen
```
hermanwikner, line 161: I think we should keep this comment as it is a way to prepare for the actual implementation. This is the right place for the component to be rendered, so the comment will be helpful once we implement this screen for real in the future!

RitaDias: fair enough, I'll add the same comment to this as we have on the configerrorscreen 👍

hermanwikner (approved):

PR #4102, opened 2023-01-18
refactor(.github): add lock workflow for inactive issues

RitaDias:
### Description

Add automated method of locking issues if they are older than 60 days with an automated message.

### What to review

Review that the yml file is set up properly and that nothing was missed that we might want to add in the steps

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -1,0 +1,27 @@
+name: Lock Threads
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
+
+concurrency:
+  group: lock
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v4.0.0
+        with:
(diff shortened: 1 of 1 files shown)


stipsan (approved):
It looks correct 👍 
But I'm not sure if 60 days is enough time. These bots works well if they help clear up issues that are rotting because OP have moved on or figured it out on their own and no longer need help. I've seen it cause a ton of frustration in OSS projects where it's the OP who ends up waiting for the maintainers to respond, and the bot is seen as too aggressive/unhelpful by the community 😅

RitaDias:
> It looks correct 👍 But I'm not sure if 60 days is enough time. These bots works well if they help clear up issues that are rotting because OP have moved on or figured it out on their own and no longer need help. I've seen it cause a ton of frustration in OSS projects where it's the OP who ends up waiting for the maintainers to respond, and the bot is seen as too aggressive/unhelpful by the community 😅

Yeah I'm also not sure if 60 is enough. Tbh we're basing this according to a new process we're trying to get off the ground to address github issues but I agree 60 might not be enough. What do you think of 90?

stipsan:
90 sounds reasonable

bjoerge (approved):

PR #4116, opened 2023-01-19
chore: upgrade @sanity/client, get-it to latest versions

rexxars:
### Description

This PR upgrades the `get-it` and `@sanity/client` dependencies to their latest versions (8.x and 5.x respectively). A few changes was made to account for removed methods, as well as the new named export (`createClient`).

Note: Once we've released the new major version of `@sanity/client` I'll update the version specifiers to use it.

### What to review

- Code changes looks right
- Studio still works as expected (a lot of surface here, but document lists, mutations and history view are probably the most important to check)
- Sanity CLI works
- `sanity dataset import` works
- `sanity dataset export` works

### Notes for release

- Upgraded studio client to use `@sanity/client@5`


stipsan (approved):
LGTM ✨

packages/@sanity/types/src/validation/types.ts
```diff
@@ -232,10 +232,6 @@ export type RuleSpecConstraint<T extends RuleSpec['flag']> = ConditionalIndexAcc
  * @public
  */
 export type ValidationContext = {
-  /**
-   * @deprecated Will be removed in the next version! Use `getClient({apiVersion: '2021-06-07'})` instead
-   */
-  client: SanityClient
```
bjoerge, line 238: Assuming _next version_ means `v3` here? Otherwise this would technically be a breaking change, right?

bjoerge: We got some downstream TS compile errors in the validation tests after this was removed. Pushed a fix for those in 4e9badf084

bjoerge (commented):
I did a canary release of this (tag: `client-v5`) and got this error:
<details>
<summary>
TypeError: getIt is not a function
    at Object.<anonymous> (/Users/bjoerge/.npm/_npx/cff7ba8e10088552/node_modules/@sanity/cli/lib/_chunks/cli-6d251d45.js:418:88)
</summary>
<pre>
npx sanity@client-v5 -v
Need to install the following packages:
  sanity@3.2.5-client-v5.11
Ok to proceed? (y) y
/Users/bjoerge/.npm/_npx/cff7ba8e10088552/node_modules/@sanity/cli/lib/_chunks/cli-6d251d45.js:418
response.statusCode<300&&!isJson(response.headers['content-type']));}const httpRequest=getIt([jsonResponse({force:true}),httpErrors(),debug$1({namespace:'get-latest-version'}),promise$1(),retry$2({shouldRetry})]);async function getLatestVersion(pkgName,opts){const scope=pkgName.split('/')[0];const regUrl=registryUrl(scope);const pkgUrl=url.resolve(regUrl,encodeURIComponent(pkgName).replace(/^%40/,'@'));const options=typeof opts==='string'?{range:opts,auth:true}:Object.assign({range:'latest',auth:true},opts);const authInfo=options.auth&&registryAuthToken(regUrl,{recursive:true});const request=options.request||httpRequest;const headers={accept:'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*'};if(authInfo){headers.authorization="".concat(authInfo.type," ").concat(authInfo.token);}let res;try{res=await request({url:pkgUrl,headers});}catch(err){if(err.response&&err.response.statusCode===404){throw new Error("Package `".concat(pkgName,"` doesn't exist"));}throw err;}const data=res.body;const range=options.range;if(data['dist-tags'][range]){return options.includeLatest?{latest:data['dist-tags'].latest,inRange:data['dist-tags'][range]}:data['dist-tags'][range];}if(data.versions[range]){return options.includeLatest?{latest:data['dist-tags'].latest,inRange:range}:range;}const versions=Object.keys(data.versions);const version=semver$2.maxSatisfying(versions,range);if(!version){throw new Error("No version exists that satisfies \"".concat(range,"\""));}return options.includeLatest?{latest:data['dist-tags'].latest,inRange:version}:version;}getLatestVersion.request=httpRequest;var src=getLatestVersion;var semverCompare=function cmp(a,b){var pa=a.split('.');var pb=b.split('.');for(var i=0;i<3;i++){var na=Number(pa[i]);var nb=Number(pb[i]);if(na>nb)return 1;if(nb>na)return-1;if(!isNaN(na)&&isNaN(nb))return 1;if(isNaN(na)&&!isNaN(nb))return-1;}return 0;};var pTimeoutExports={};var pTimeout$1={get exports(){return pTimeoutExports;},set exports(v){pTimeoutExports=v;}};let TimeoutError$1=class TimeoutError extends Error{constructor(message){super(message);this.name='TimeoutError';}};const pTimeout=(promise,milliseconds,fallback,options)=>{let timer;const cancelablePromise=new Promise((resolve,reject)=>{if(typeof milliseconds!=='number'||milliseconds<0){throw new TypeError('Expected `milliseconds` to be a positive number');}if(milliseconds===Infinity){resolve(promise);return;}options={customTimers:{setTimeout,clearTimeout},...options};timer=options.customTimers.setTimeout.call(undefined,()=>{if(typeof fallback==='function'){try{resolve(fallback());}catch(error){reject(error);}return;}const message=typeof fallback==='string'?fallback:"Promise timed out after ".concat(milliseconds," milliseconds");const timeoutError=fallback instanceof Error?fallback:new TimeoutError$1(message);if(typeof promise.cancel==='function'){promise.cancel();}reject(timeoutError);},milliseconds);(async()=>{try{resolve(await promise);}catch(error){reject(error);}finally{options.customTimers.clearTimeout.call(undefined,timer);}})();});cancelablePromise.clear=()=>{clearTimeout(timer);timer=undefined;};return cancelablePromise;};pTimeout$1.exports=pTimeout;// TODO: Remove this for the next major release
                                                                                       ^

TypeError: getIt is not a function
    at Object.<anonymous> (/Users/bjoerge/.npm/_npx/cff7ba8e10088552/node_modules/@sanity/cli/lib/_chunks/cli-6d251d45.js:418:88)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/bjoerge/.npm/_npx/cff7ba8e10088552/node_modules/@sanity/cli/lib/run.js:4:11)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)

Node.js v18.12.1
</pre>
</details>

Edit: looks like CI fails for the same reason

rexxars:
Interesting turn of events:
1. Upgraded `get-it` used in `get-latest-version` to 8.x, which dedupes and ensures there is only a single version of get-it used in the monorepo
2. This revealed this error in the tests - turns out jest 27 doesn't understand conditional package exports:
     > Cannot find module 'get-it/middleware' from '../../node_modules/@sanity/client/dist/index.cjs'
3. Upgrading jest makes `import` statements in node_modules break

@stipsan do you have any knowledge of the right way of solving this?

stipsan:
@rexxars pushed an attempt at fixing it. Basically we have to setup Jest to transform any package in `node_modules` that ships ESM syntax so that `babel-jest` changes it back to CJS.
Or we could turn `legacyExports: true` in `get-it` and publish a new version so we can stay on jest v27 for now and have `get-it/middleware` working?

rexxars:
Alright - new attempt in place. Fixed a few cases where the `@sanity/cli` API client would break when used in Sanity v2 projects. Please re-review!

bjoerge (approved):
I did a new tagged release and tested both by setting up a new studio and installing in a personal studio, and didn't find any issues. Amazing work! :sparkles:

Issue #4063, opened 2023-01-10
Unpublish dialog asks you to confirm deletion


runesandnes:
**Describe the bug**

The unpublish dialog says "Are you sure you want to delete Foo"?, which is a bit too dramatic.

**Screenshots**
![sanity-unpublish](https://user-images.githubusercontent.com/103252104/211506183-e23b0496-f0aa-4e50-98ae-16f2c3cb3a7f.png)

**Which versions of Sanity are you using?**
@sanity/cli (global)          3.1.4 (up to date)
@sanity/dashboard             3.1.1 (latest: 3.1.2)
@sanity/eslint-config-studio  2.0.1 (up to date)
@sanity/table                 1.0.1 (up to date)
@sanity/vision                3.0.6 (latest: 3.1.4)
sanity                        3.1.4 (up to date)

ninaandal:
Thank you for reporting @runesandnes 🙌 A fix has been included in [v3.2.5](https://github.com/sanity-io/sanity/releases/tag/v3.2.5)

PR #4118, opened 2023-01-20
fix(desk): Change hardcoded delete word in action dialog to the correct action

ninaandal:
### Description
The action dialog should say "unpublish" and not "delete" when you unpublish a document. 
Changed to use `action` instead of hardcoded `delete`.

https://github.com/sanity-io/sanity/issues/4063

### What to review
Try to unpublish/delete document and see that the text is correct.


hermanwikner (approved):
Nice! 👍

Issue #3947, opened 2022-12-09
[v3][Bug] Read-only reference field isn't.


Snugug:
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

Creating a read-only reference field will remove the ability to create a new field, but does not prevent the existing field, if empty, from being set.

**To Reproduce**

1. Create a read-only reference field in a document:

```js
    defineField({
      name: 'reference',
      title: 'Reference',
      type: 'reference',
      to: [{ type: 'post' }],
      readOnly: () => true,
    }),
```

2. Notice that both the search-ahead and the drop-down for selecting referenced content still works. Select a piece of content
3. Choose a piece of content. Notice how the field and that content is now set to read-only and cannot be changed

**Expected behavior**

When a reference field is set to read-only, its value shouldn't be able to be set if it's empty

**Screenshots**
![Screen Shot 2022-12-09 at 9 24 30 AM](https://user-images.githubusercontent.com/377188/206723595-d35edab8-f625-45c7-9287-68503c07f5c7.png)

![image](https://user-images.githubusercontent.com/377188/206723679-b0760ca7-a71d-41ac-a6f6-6d95f4a6523d.png)

![image](https://user-images.githubusercontent.com/377188/206723748-8a43467e-babf-49ce-9e67-6d28d386f9d4.png)

![image](https://user-images.githubusercontent.com/377188/206723838-4904e50b-2a32-4994-980e-316f6731af56.png)

![image](https://user-images.githubusercontent.com/377188/206723923-89151b14-bb56-4674-b069-816df85607ad.png)

**Which versions of Sanity are you using?**

```
@sanity/cli (global)          3.0.6 (up to date)
@sanity/eslint-config-studio  2.0.1 (up to date)
sanity                        3.0.6 (up to date)
```

**What operating system are you using?**

Chrome 108.0.5359.94 (browser problem not CLI problem)

**Which versions of Node.js / npm are you running?**

```
pnpm -v && node -v
7.16.1
v14.21.1
```

ninaandal:
Thank you for reporting @Snugug 🙌 A fix has been included in [v3.2.5](https://github.com/sanity-io/sanity/releases/tag/v3.2.5)

PR #4123, opened 2023-01-23
fix(core): add readonly prop to RefenceInput

ninaandal:
### Description
An empty read-only reference field should not show the dropdown and be able to set a reference. It should be read-only for an empty field as well.

https://user-images.githubusercontent.com/44635000/214015457-2dda9a5c-74e2-479c-83db-7f590629afb1.mov

https://github.com/sanity-io/sanity/issues/3947

Added `readOnly` field for the `ReferenceAutocomplete`.

### What to review
- Make sure that it truly is read-only for the field - both when empty and when there is a reference

### Notes for release
Fixes bug where the read-only reference field was not read-only for empty field

### DIFF
--- a/dev/test-studio/schema/standard/references.ts
+++ b/dev/test-studio/schema/standard/references.ts
@@ -23,6 +23,7 @@ export default defineType({
       type: 'reference',
       description: 'Some description',
       to: {type: 'referenceTest'},
+      readOnly: () => true,
     },
     {
       name: 'selfOrEmpty',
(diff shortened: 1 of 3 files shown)


ninaandal:
> I think the "Create new" button should still be visible in read only mode. 🤔 It doesn't really make much sense in read only mode since it can't be used – but I think it becomes clearer that it is a reference input that is read only. Currently, I think it looks like a string input field that is read only.
> 
> Screenshot 2023-01-23 at 13 05 11

Yes, I agree 😄 I have fixed that

hermanwikner (approved):
LGTM! 👍

PR #4127, opened 2023-01-24
fix(core): non valid file types for image and file inputs will be greyed out

RitaDias:
### Description

In the past when using the default asset source, you could pick any of the files / images there regardless of whether the input had `accept` limitations (type limitations)

You can still see their usage and delete them, just not upload them to the input.

| Image (png) | file (pdfs) |
|-------|-------|
|  image | image  |

Fix issue where the file type was "overflowing" into the next table cell in the file input
Pasted Graphic

#### Image - Without `accept`
1. Go to test-studio
2. Standard inputs -> image -> Main Image field
4. Press select -> uploaded images
5. You should be able to select and upload any item

#### Image - With `accept`
1. Go to test-studio
2. Standard inputs -> image -> PNG image field
3. Press select -> uploaded images
4. Check that all assets are specifically pngs (you can check by trying to delete them and seeing the extension, for example)

#### File - Without `accept`
1. Go to test-studio
2. Standard inputs -> file -> Simple file (field)
4. Press browse
5. You should be able to select and upload any item

#### File - With `accept`
1. Go to test-studio
2. Standard inputs -> file -> File only accepts pdf field
3. Press browse
4. Check that all assets are pdfs

#### File - With multiple `accept` (png, jpg)
1. Go to test-studio
2. Standard inputs -> image -> JPEG & PNG image
4. Press browse
5. Check that all assets are either png or jpg (you can check by trying to delete them and seeing the extension, for example)
 
Check that it looks good on both *light and dark mode*

### Notes for release

While browsing, the File and Image input will only allow you to select items that are of a valid type


packages/sanity/src/core/form/studio/assetSource/AssetRow.tsx
```diff
@@ -42,12 +42,18 @@ const CustomCard = styled(Card)<RowProps>`
       --card-muted-fg-color: var(--card-bg-color);
       --card-fg-color: var(--card-bg-color);
     `}
+
+  ${({disabled}) =>
+    disabled &&
+    css`
+      --card-muted-fg-color: var(--card-border-color);
```
RitaDias, line 49: make the text (beyond the title) greyed out

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -24,8 +29,9 @@ const buildQuery = (start = 0, end = PER_PAGE, assetType = ASSET_TYPE_IMAGE) =>
     mimeType,
     extension,
     size,
-    metadata {dimensions}
-  }
+    metadata {dimensions},
+    "acceptedType": ${acceptParam.length <= 0 ? 'true' : `mimeType == "${acceptParam}"`}
```
RitaDias, line 33: when the type is "any" then the file and image input send in '' strings which means that the type will be accepted

rexxars (changes requested):
Fantastic that we're addressing this!

A few notes on the implementation:

1. The HTML `accept` attribute is quite loose. It can accept values such as `image/png, image/jpeg` (to allow multiple types), `image/*` (wildcards) or single values (`image/png`). The current implementation does not take this into account, and requires a complete match.  
   While we theoretically _could_ make GROQ match these (with enough work), I don't think it is the right solution. I would rather the assets are returned as-is, and that we do the matching of `mimeType` against the `accept` attribute on the client side.
    - Roughly speaking: lowercase the `accept` attribute, split on `,`, trim whitespace, then match either the full mime type or the start of the mime type (if a wildcard is present).
2. If no `accept` attribute is present, it never loads any images/files
3. I feel like we could improve the disabled state of the not-accepted images even more, but I am not sure how to do so design-wise, so will leave that for @mikolajdobrucki to decide

RitaDias:
> Fantastic that we're addressing this!
> 
> A few notes on the implementation:
> 
> 1. The HTML `accept` attribute is quite loose. It can accept values such as `image/png, image/jpeg` (to allow multiple types), `image/*` (wildcards) or single values (`image/png`). The current implementation does not take this into account, and requires a complete match.
>    While we theoretically _could_ make GROQ match these (with enough work), I don't think it is the right solution. I would rather the assets are returned as-is, and that we do the matching of `mimeType` against the `accept` attribute on the client side.
>    
>    * Roughly speaking: lowercase the `accept` attribute, split on `,`, trim whitespace, then match either the full mime type or the start of the mime type (if a wildcard is present).
> 2. If no `accept` attribute is present, it never loads any images/files
> 3. I feel like we could improve the disabled state of the not-accepted images even more, but I am not sure how to do so design-wise, so will leave that for @mikolajdobrucki to decide

1. Ahh this makes sense. My downfall here was going to check if the accept field on our image/field schema was string (or an array of strings) without considering the native accept fields. 
   1.1. This makes sense, question re: using groq vs client side: I also considered this and what made me go with the groq approach was the thought of making it load faster vs having to rely on the client having to filter things out. This might be a non-concern though, would love to hear your thoughts on this
2. For some reason when I did these tests locally before creating the PR I could swear they worked but now trying them again they don't 🙃 that's less than stellar I'll fix that 👍

bjoerge:
We use the [attr-accept](https://www.npmjs.com/package/attr-accept) package to help us verify the mime type during drag+drop, so might serve us well to use it here too?

RitaDias:
> We use the [attr-accept](https://www.npmjs.com/package/attr-accept) package to help us verify the mime type during drag+drop, so might serve us well to use it here too?

I see that we are using a `accepts` in [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/studio/uploads/accepts.ts) but it's not from that package but seems to have the same behavior and being used in the way you mentioned. 🙂  I'll explore a bit

RitaDias:
> We use the [attr-accept](https://www.npmjs.com/package/attr-accept) package to help us verify the mime type during drag+drop, so might serve us well to use it here too?

Thank you for sharing this! I had a go at it! This works really well for all instances (no accept, one accept, multiple accepts) and it would be client side. 
I guess my question now is: do we want to continue relying on client side rendering versus using the groq filters. Usually I would prefer to go with an approach that pushes these sort of concerns to the back end but I feel I might be missing some context 👁️

packages/sanity/src/core/form/studio/assetSource/AssetRow.tsx
bjoerge: > > We use the [attr-accept](https://www.npmjs.com/package/attr-accept) package to help us verify the mime type during drag+drop, so might serve us well to use it here too?
> 
> I see that we are using a `accepts` in [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/studio/uploads/accepts.ts) but it's not from that package but seems to have the same behavior and being used in the way you mentioned. 🙂 I'll explore a bit

Ah, yeah, that's the one! Looks like we have inlined it and added types :)

robinpyon:
One argument for matching these in GROQ is that we could fetch (and render) _only_ valid assets when opening the default asset source. 

The current implementation means it'd be possible for users to try select an asset for a field (with custom `accept`) and then be required to manually scroll / load multiple pages to try and find qualifying assets. In the worst case, they may page through dozens of pages only to realise their dataset doesn't have anything valid.

rexxars:
I would also argue that the user experience of only showing allowed file types is better. If we went down that route, we'd want to push this to a GROQ-filter, obviously.

However, going this route means we'll have to display some sort of message indicating that the displayed files/images are a subset of the uploaded files/images, eg something like "Showing only allowed file types".

_Ideally_ you'd want to display _what_ is allowed, eg "Showing only SVG images", but it can be hard to translate something like `image/png,image/svg` or `application/vnd.openxmlformats-officedocument.wordprocessingml.document` to a printable message.

RitaDias:
Thank you all for your thoughts!
@mikolajdobrucki, what are your thoughts on the ux of switching this solution to hide the types that aren't part of the accepted parameter? :)

RitaDias:
@rexxars @bjoerge @robinpyon I spoke with Mikolaj on Friday last week, I was hopeful he would write his thoughts here but I'll do it (and he can add if I've missed something)
- He doesn't like removing them fully from the list because it is not super obvious and it might introduce confusion and people being unsure if it is the system that is not showing the images vs an error that they've done
- The messaging and being able to parse in a neat way is also part of his concern (I brought this up given your comment Espen).

With this in mind (keeping the images / files as disabled but still showing) are we ok with filtering it out with GROQ (the current solution on this PR) and if not what is the reasoning behind it?

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -207,6 +233,18 @@ const DefaultAssetSource = function DefaultAssetSource(
...
       __unstable_autoFocus={hasResetAutoFocus}
     >
+      {showAcceptMessage && (
+        <Card tone="primary" marginTop={4} marginX={4} padding={[3, 3, 4]} border radius={2}>
+          <Flex gap={[3, 4]} align="center">
+            <Text>
+              <InfoOutlineIcon />
+            </Text>
+            <Text size={1}>
+              Only showing assets of accepted types: <b>{acceptTypes}</b>
```
RitaDias, line 243: After bringing up in the Standup, it was decided to go with the printing of the types accepted as a first go.

robinpyon:
Thanks @RitaDias, this is getting there! What you've proposed will make it much easier for users to find relevant assets. The callout displaying the current accept option is also great and easy to understand.

However, the current approach makes the assumption that only MIME types will be supplied where the `accept` option can take a range of [unique file type specifiers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers). This means that currently, accept values like `.png` or `audio/*` will fail.

Based on the above link, the `accept` option can be:
> 1. A valid case-insensitive filename extension, starting with a period (".") character. For example: .jpg, .pdf, or .doc.
> 2. A valid MIME type string, with no extensions.
> 3. The string audio/* meaning "any audio file".
> 4. The string video/* meaning "any video file".
> 5. The string image/* meaning "any image file".

With this in mind, one approach you could consider is:
- Splitting the `accept` string and then process accordingly: lowercase, trim, dedupe and omit empty values
- Partition these results into three buckets, ignoring everything else:
	1. Values with the _specific_ wildcard values (e.g. `audio/*, video/*, image/*`). 
	2. MIME types (e.g. in the format `type/subtype`)
	3. File extensions (e.g. `.png`)
- Iterate through all MIME types and remove any values that have already been covered by a 'wildcard' selector (e.g. remove `audio/aac` if `audio/*` is present)
- Construct a GROQ query that can accommodate items across all three buckets, e.g.
	1. Add `mimeType` filters with the `match` operator for wildcard values (`mimeType match "image/*"`)
	2. Add `mimeType` filters with equality operators for exact values (`mimeType == 'video/mp4'`)
	3. Add `extension` filters with equality operators for exact values (`extension == 'png'`) 

So putting this together, the following `accept` string on a file input:
```
.doc,.docx,.xml,application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,,,     .jpeg,image/jpg,audio/*,  cat/*,,,not-valid,42,audio/aac
```

... could build a GROQ string similar to:
```
*[
  _type == 'sanity.fileAsset' &&
  (
    extension in [
      'doc',
      'docx',
      'jpeg',
      'xml'
    ] ||
    mimeType in [
      "application/msword",
      "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
    ] ||
    mimeType match "audio/*"
  )
] {
  ... 
}
```

Finally, it's just a question of what this callout says. Using the above example, it could potentially read something like:

```
Only showing assets of type: application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document,audio/*,.doc,.docx,image/jpg,.jpeg,.xml
```

This could be an alphabetised list of _valid_ `accept` values. We'd also have the option of surfacing whether invalid values have been removed, but this may not be necessary if we're already showing the full valid list.

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -44,7 +88,7 @@ const DefaultAssetSource = function DefaultAssetSource(
   ref: React.ForwardedRef<HTMLDivElement>
 ) {
   const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
-  const versionedClient = useMemo(() => client.withConfig({apiVersion: '1'}), [client])
+  const versionedClient = useMemo(() => client.withConfig({apiVersion: '2023-02-14'}), [client])
```
RitaDias, line 91: Needed to be updated since version 1 of the api was giving me the wrong data

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -58,7 +102,16 @@ const DefaultAssetSource = function DefaultAssetSource(
...
+    accept,
   } = props
+  const acceptTypes = accept
+    ? accept
+        .split(',')
+        .map((a) => a.trim())
+        .join(', ')
+    : ''
+  const showAcceptMessage = !isLoading && accept && accept.length > 0
+  const isImageOnlyWildCard = accept && accept === 'image/*' && assetType === 'image'
```
RitaDias, line 114: Needed otherwise, the default of the "image input" always returning "image/*" would trigger the accept message to always appear. And in that use case it is not necessary.

RitaDias:
@rexxars @robinpyon @bjoerge hi all 👋 it's been 84 years but I'm finally back with the updated changes after Rob's last comment :)

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
+
+      // all that remains is then the mime types, so we build that part
+      const mime = `'${acceptR}', ${acceptTypes.mimes}`
+      return {...acceptTypes, ...{mimes: mime}}
+    },
+    {mimes: '', extensions: '', wildcards: ''}
+  )
+
+  // when no accept filter is set, we don't need to add the filter condition
+  return `&&
```
bjoerge, line 43: wondering if it makes sense to leave this `&&` out here and move it to the query instead - would make the function more future proof and reusable I believe

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -70,16 +123,18 @@ const DefaultAssetSource = function DefaultAssetSource(
...
 
-      fetch$.current = versionedClient.observable
-        .fetch(buildQuery(start, end, assetTypeParam), {}, {tag})
-        .subscribe((result) => {
-          setIsLastPage(result.length < PER_PAGE)
-          // eslint-disable-next-line max-nested-callbacks
-          setAssets((prevState) => prevState.concat(result))
-          setIsLoading(false)
-        })
+      if (typeof accept !== 'undefined') {
```
bjoerge, line 126: I'm struggling to understand how this fits in. Is it not valid to pass `accept: undefined`? If `accept` _is_ `undefined`, should we not still fetch assets? If it (for some reason) goes from `!undefined` to `undefined`, will it not then keep `isLoading` forever? I might be missing something obvious here.

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
+const buildFilterQuery = (acceptParam: string) => {
+  const WILDCARD_ACCEPT = ['image/*', 'audio/*', 'video/*']
+  const acceptItems = acceptParam.split(',').map((accept) => accept.trim())
+
+  const typesForFilter: {[key: string]: string} = acceptItems.reduce(
+    (acceptTypes: {[key: string]: string}, acceptR: string) => {
+      // builds the wildcard part of the groq query
+      if (WILDCARD_ACCEPT.includes(acceptR)) {
+        const wild = `mimeType match '${acceptR}' || ${acceptTypes.wildcards}`
+        return {...acceptTypes, ...{wildcards: wild}}
```
bjoerge, line 26: The last spread here isn't necessary, and I think I would prefer the variable to be inlined for better readability:

```suggestion
{
  ...acceptTypes,
  wildcards: `mimeType match '${acceptR}' || ${acceptTypes.wildcards}`
}
```

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
+      }
+
+      // builds the extension part of the groq query (and removes the .)
+      if (acceptR.indexOf('.') === 0) {
+        const exte = `'${acceptR.replace('.', '')}', ${acceptTypes.mimes}`
+        return {...acceptTypes, ...{extensions: exte}}
```
bjoerge, line 32: Would be good if this could be simplified the same way as mentioned above

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
+      }
+
+      // all that remains is then the mime types, so we build that part
+      const mime = `'${acceptR}', ${acceptTypes.mimes}`
+      return {...acceptTypes, ...{mimes: mime}}
```
bjoerge, line 37: And here, same as above :)

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
+    },
+    {mimes: '', extensions: '', wildcards: ''}
+  )
+
+  // when no accept filter is set, we don't need to add the filter condition
+  return `&&
+  (
+    ${typesForFilter.wildcards} 
+    extension in [${typesForFilter.extensions}] ||
```
bjoerge, line 46: if there's no extension this will result in `extension in ['']` - will that work as expected?

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
```diff
@@ -14,8 +14,51 @@ const PER_PAGE = 200
...
 const ASSET_TYPE_FILE = 'sanity.fileAsset'
 
-const buildQuery = (start = 0, end = PER_PAGE, assetType = ASSET_TYPE_IMAGE) => `
-  *[_type == "${assetType}"] | order(_updatedAt desc) [${start}...${end}] {
+const buildFilterQuery = (acceptParam: string) => {
+  const WILDCARD_ACCEPT = ['image/*', 'audio/*', 'video/*']
+  const acceptItems = acceptParam.split(',').map((accept) => accept.trim())
+
+  const typesForFilter: {[key: string]: string} = acceptItems.reduce(
+    (acceptTypes: {[key: string]: string}, acceptR: string) => {
```
bjoerge, line 22: Could `acceptR` be written in full? I don't understand what the trailing `R` stands for :sweat-smile:

dev/test-studio/schema/standard/files.ts
```diff
@@ -74,5 +74,33 @@ export default defineType({
...
+      },
+    },
+    {
+      name: 'fileWithMultipleAccept',
+      title: 'File for audio/*, video/*, .zip, image/jpeg, image/png',
+      type: 'file',
+      description:
+        'Should not accept other files types than audio/*, video/*, .zip, image/jpeg, image/png',
+      options: {
+        accept: 'audio/*, video/*, .zip, image/jpeg, image/png',
```
bjoerge, line 93: Thank you for adding these! :heart:

mikolajdobrucki (approved):
lgtm!

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
RitaDias: The reason why I kept it here was because in my line of thinking if there is not accept params then it would not have a "loose" && in the query itself. Unless I'm misunderstanding your suggestion 😄

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
RitaDias: I think it still needs the spreader because it needs to merge "two objects" together. without the spreader the editor is telling me that "property assignement is missing". Unless I missed something? 🤔

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
RitaDias: You're absolutely right, leftovers from my first pass on the code 🙃

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
RitaDias: Part of an old solution that needed to be removed :) 
Sorry about that, removed ✅

packages/sanity/src/core/form/studio/assetSource/DefaultSource.tsx
RitaDias: Yes, it will!
(so in the use case `audio/*, video/*, image/jpeg, image/png`, it'll show all audios, videos, jpegs and pngs).
I understand it makes the query a bit noisy, I considered adding a specific condition to tackle these extensions but thought it might just add noise considering it works as expected 🤔 What do you think?

bjoerge (approved):
Great work! 👏🏼

Issue #4124, opened 2023-01-23
`readonly` property not working inside fieldsets


KittyGiraudel:
**Describe the bug**

The `readOnly` property does not work if the field lives inside a fieldset (regardless whether collapsed or not) in version 3. It used to work fine in version 2.

**To Reproduce**

1. Define a field on a document.
2. Make this field part of a fieldset.
3. Mark this field as readonly.

```js
{
  type: 'document',
  name: 'docName',
  fieldsets: [
    { name: 'fieldsetName', options: { collapsible: true, collapsed: true }, }
  ],
  fields: [
    { name: 'fieldName', type: 'string', fieldset: 'fieldsetName', readOnly: true }
  ]
}
```

**Expected behavior**

I expect the `readOnly` property to be properly honored. 

**Which versions of Sanity are you using?**

v3.2.4

**What operating system are you using?**

macOS, although it shouldn’t matter.

**Which versions of Node.js / npm are you running?**

v16.15.0

PR #4129, opened 2023-01-25
fix(form): make sure fields inside fieldsets inherits the object's readOnly state

bjoerge:
### Description

The bug described in #4124 happened because we ignored use the `readOnly`-state for fields inside fieldsets. This PR makes sure we properly forward it as the field's formState.

### What to review
- Make sure readOnly fields inside fieldsets actually becomes readOnly

### Notes for release
- Fixes an issue that ignored the `readOnly` state for fields inside fieldsets

### DIFF
--- a/packages/sanity/src/core/form/store/formState.ts
+++ b/packages/sanity/src/core/form/store/formState.ts
@@ -492,7 +492,7 @@ function prepareObjectInputState<T>(
     return null
   }
 
-  const readOnly =
+  const objectIsReadOnly =
     props.readOnly ||
     resolveConditionalProperty(props.schemaType.readOnly, conditionalPropertyContext)
 
@@ -524,7 +524,7 @@ function prepareObjectInputState<T>(
   const parentProps: RawState<ObjectSchemaType, unknown> = {
     ...props,
     hidden,
-    readOnly,
+    readOnly: objectIsReadOnly,
   }
 
   // note: this is needed because not all object types gets a ´fieldsets´ property during schema parsing.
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/store/formState.ts
```diff
@@ -492,7 +492,7 @@ function prepareObjectInputState<T>(
     return null
   }
 
-  const readOnly =
+  const objectIsReadOnly =
```
bjoerge, line 495: Renamed this variable to make it clear that it's the parent object's readOnly state

packages/sanity/src/core/form/store/formState.ts
```diff
@@ -576,18 +576,21 @@ function prepareObjectInputState<T>(
         field,
         parent: parentProps,
         index,
-
         // the explicit type cast here is ok - we know that a fieldset can not have fieldsets
       }) as FieldMember | FieldError | null
+
+      if (fieldState?.kind === 'error') {
```
bjoerge, line 582: This is also not strictly related to the fix, but the early return replaces the `as FieldMember` below

packages/sanity/src/core/form/store/formState.ts
```diff
@@ -576,18 +576,21 @@ function prepareObjectInputState<T>(
...
         return []
       }
 
       const fieldStateWithReadOnly = {
         ...fieldState,
         field: {
-          ...(fieldState as FieldMember)?.field,
-          readOnly: readOnly || fieldsetReadOnly,
+          ...fieldState.field,
+          readOnly: objectIsReadOnly || fieldState.field.readOnly || fieldsetReadOnly,
```
bjoerge, line 593: Including `fieldState.field.readOnly` here is what fixes the issue :)

PR #4130, opened 2023-01-25
fix(core): only notify subscribers when `currentUser` value changes in `createAuthStore`

hermanwikner:
### Description

This PR addresses an issue discovered when two browser windows are open in the same studio at the same time, where reloading one causes the other to re-render. The solution implemented in this PR is to only notify subscribers of the authentication stream when there is a change in the `currentUser` object by using the `distinctUntilChanged` operator to compare previous and next states.

### What to review

1. Open the same studio in two different browser windows
2. Reload one of the browser windows
3. The studio in the other browser window should not re-render.

### Notes for release

n/a


bjoerge (approved):
LGTM! Not too familiar with this code, so would be great to have @rexxars's eyes on it too

PR #4135, opened 2023-01-27
fix(core): array popover editing improvements

hermanwikner:
### Description
This pull request addresses several problems with the popover editing feature for array items.

**Context**
It is possible to configure an array field to use a popover instead of a dialog for editing items, as shown in the following code example:

```js
{
  type: 'array',
  name: 'myArray',
  options: {
    modal: {type: 'popover', width: 2}, // <-- 👈 edit inside a popover instead of a dialog
  },
  of: [
    // fields...
  ]
}
```

**List of fixes:**
- Ensures that the width property functions as intended (currently the popover does not respect `width`)
- Restores focus on the array item upon closing the popover
- Traps focus within the popover content
- Includes a sticky header in the popover + header UI fixes
- Constrains the popover's size to the boundary element (i.e the form view)
- Makes the popover scrollable
- General user interface improvements

**Note:** There is currently an issue where the popover flashes and is difficult to use when an image input is rendered inside the popover and the width is set to `width: 'auto'`. It is likely that this is an issue that should be addressed in the Sanity UI library rather than in the Studio. However, this will be treated as a separate issue.

### What to review

Please review the list of fixes above. While there are many areas that could be improved when it comes to popover editing in arrays, this pull request focuses on addressing the most critical issues.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/components/popoverDialog/PopoverContainer.tsx
+++ b/packages/sanity/src/core/components/popoverDialog/PopoverContainer.tsx
@@ -1,0 +1,40 @@
+import {
+  Container,
+  ContainerProps,
+  rem,
+  ResponsiveWidthStyleProps,
+  Theme,
+  useArrayProp,
+  _responsive,
+} from '@sanity/ui'
+import React, {Ref} from 'react'
+import styled from 'styled-components'
+
+// This is a workaround to make sure that the Container gets the correct width when used inside a popover.
+// The default Container uses `maxWidth` which doesn't work well with popovers because the popover
+// calculates its width based on the content width.
+const StyledContainer = styled(Container)((props: ResponsiveWidthStyleProps & {theme: Theme}) => {
+  const {theme} = props
+  const {container, media} = theme.sanity
+
+  return _responsive(media, props.$width, (val) => ({
+    // Make sure that the Container gets the correct width when used inside a popover.
+    width: val === 'auto' ? 'none' : rem(container[val]),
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
Great work!

We are already using [react-focus-lock](https://www.npmjs.com/package/react-focus-lock) to trap focus inside datepicker popovers, did you consider using it here as well? https://github.com/sanity-io/sanity/blob/19d0e70c01c33ba8c30a43eb1cc3751c10d29287/packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx#L2

hermanwikner:
> Great work!
> 
> We are already using [react-focus-lock](https://www.npmjs.com/package/react-focus-lock) to trap focus inside datepicker popovers, did you consider using it here as well?
> 
> https://github.com/sanity-io/sanity/blob/19d0e70c01c33ba8c30a43eb1cc3751c10d29287/packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx#L2

Good point @bjoerge  👍  I wasn't aware that we had `react-focus-lock`. Updated PR to use it instead!

robinpyon:
Really nice @hermanwikner! Works fantastically

Possibly out of scope for this issue, but perhaps there's something in conditionally using regular dialogs on smaller / mobile breakpoints to prevent the following? (it's still completely functional, you just have very limited real estate)

image

robinpyon (approved):

PR #4138, opened 2023-01-30
Refactor PortableTextEditor: simplify and isolate concerns

skogsmaskin:
### Description

This is a maintenance refactoring of the PortableTextEditor in order to better isolate different concerns and make  contributions and development easier in the future. Hopefully it will also lead to better stability and fewer bugs as it's easier to reason about the code and the different effects. Writing the logic into functional components also let us use hooks and new React features.

* Broke apart `PortableTextEditor` main component into several new components:
  - `Validator` - responsible for validating new values
  - `SlateContainer` - responsible for setting up (and compartmentalize / hide) the Slate instance.
  - `Synchronizer` - responsible for synchronizing the value and selection state. Also provides the contexts we need.
  - `useSyncValue` - hook used for comparing a value against the current editor state and update it if needed.

`PortableTextEditor` is now pretty much only the static API methods and not all of the above into one huge (React class) component. 

I have refactored how the editor plugins can subscribe to stuff. Previously they subscribed automatically, however it now has it's own subscribe/unsubscribe function that can be called by React lifecycle hooks.

Fixed a bug where the value sometimes would not be properly unset and we got duplicated blocks going from an empty editor to a non-empty one.

Finally fixed an issue with our collaborative test environment which have been resulting in flaky tests. This was because it put out the new value before the patches over the websocket connection.

I have also replaced `puppeteer` with `playwright` as we are going to start using it anytime soon.

### DIFF
--- a/packages/@sanity/cli/test/build.test.ts
+++ b/packages/@sanity/cli/test/build.test.ts
@@ -8,7 +8,7 @@ describeCliTest('CLI: `sanity build` / `sanity deploy`', () => {
     const testConcurrentV3 = version === 'v3' ? testConcurrent : test.skip
 
     // Builds can take a bit of time with lots of concurrent tasks and slow CIs
-    jest.setTimeout(120 * 1000)
+    jest.setTimeout(240 * 1000)
 
     const studioPath = path.join(studiosPath, version)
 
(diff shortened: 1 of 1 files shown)


skogsmaskin:
Succeeded by https://github.com/sanity-io/sanity/pull/4150

PR #4139, opened 2023-01-30
fix(core): focus handling and dialog behavior

hermanwikner:
### Description
This pull request implements:
- Consistent dialog behavior. In order to be able to do this, `@sanity/ui` has been updated to `1.2.0` where improvements to the `Dialog` component were introduced. 
- Some focus issues in the image and file input.

### What to review
- When several dialogs are open at the same time with split panes, you should only be able to close one of the dialogs (with "Escape" or click outside)
- Image and file input is usable with keyboard

### Notes for release
n/a

### DIFF
--- a/dev/design-studio/package.json
+++ b/dev/design-studio/package.json
@@ -33,7 +33,7 @@
     "@sanity/color": "^2.1.20",
     "@sanity/icons": "^2.1.0",
     "@sanity/logos": "^2.0.2",
-    "@sanity/ui": "^1.0.0",
+    "@sanity/ui": "^1.2.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "sanity": "3.5.0",
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -24,7 +24,7 @@
     "@sanity/image-url": "^1.0.2",
     "@sanity/portable-text-editor": "3.5.0",
     "@sanity/types": "3.5.0",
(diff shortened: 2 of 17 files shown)


mikolajdobrucki (approved):
LGTM! 🚀

skogsmaskin:
I'm testing the following, but I'm not sure if it's part of what we are trying to solve here, but:

1) I open a image crop tool modal
2) I open up "Inspect document" modal.
3) Both modals are shown
4) I press ESC
5) Both modals are closed.

I would at least argue that ESC should not close both of them?

Should the image modal close when I open the "inspect" modal?

mikolajdobrucki (approved):

mikolajdobrucki:
> I'm testing the following, but I'm not sure if it's part of what we are trying to solve here, but:
> 
> 1. I open a image crop tool modal
> 2. I open up "Inspect document" modal.
> 3. Both modals are shown
> 4. I press ESC
> 5. Both modals are closed.
> 
> I would at least argue that ESC should not close both of them?
> 
> Should the image modal close when I open the "inspect" modal?

I would agree that  ESC should not close both dialogs at the same time. 
It is not a regression though (afaik it was always like this) so I would treat that as a separate issue not blocking for release
therefore accepted the PR 👍

bjoerge:
Took the liberty of rebasing against next

bjoerge (approved):
Did a round of testing and works great from what I can tell! ✨

PR #4141, opened 2023-01-30
chore: add script for setting npm `latest` tag on v2 modules

rexxars:
### Description

When we run a v2 release, we now set the npm dist tag to `v2`. This is fine for the modules that are present in both v2 and v3, such as `@sanity/cli`, but it is unfortunate that new versions of 2.x modules are not set as `latest`.

This PR introduces a script which loops through all the modules in the `@sanity` packages folder, runs `npm view <pkgName> version` on them to see what the `latest` tag is set to on npm, and if that version is in the 2.x range, _and_ a _newer_ version is available locally (and is on the 2.x range) - it will set the `latest` tag on the npm module to the local version.

Kind of hacky, but think it should do the trick.

Also set this up as a post-publish task. 

### Notes for release

Nothing. Transparent for the end user.


bjoerge (approved):
This is great, thank you!

PR #4142, opened 2023-01-31
feat(config): allow specifying `apiHost` for source

rexxars:
### Description

This PR allows users to configure a custom API host for each source (usually for the entire studio, but it makes sense to do it on a source level, which workspaces inherit from). Enables the use of custom CNAMEs, which some of our customers have, as well as for testing purposes.

### What to review

- Specifying API host makes studio use the configured host. Prefixes with project ID.
- Code makes sense

### Notes for release

- A custom API hostname can now be configured for the studio through the `apiHost` configuration parameter. Note that the  configured API host must behave the same as Sanity's API, thus this is mostly useful for CNAMEs.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -110,6 +110,16 @@ export default defineConfig([
     basePath: '/playground',
   },
   {
+    name: 'staging',
+    title: 'Staging',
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Nice, thanks for adding such helpful comments too!

PR #4144, opened 2023-01-31
refactor: use named export for schema package

rexxars:
### Description

We want to move away from default exports, as they are hard to work with in a dual CommonJS/ESM environment. The schema package currently only exposes the `Schema` class on the default export. This PR moves the default Schema class export to a _named_ export (`Schema`), _and_ introduces a "deprecated" class for the default export, which only exists to warn about the deprecated default export.

I decided to warn with a stack trace to make it easier to find where the warning is appearing from.

### Notes for release

- Use named export for the `@sanity/schema` module - the default export is now deprecated and will give a warning when used

### DIFF
--- a/packages/@sanity/block-tools/README.md
+++ b/packages/@sanity/block-tools/README.md
@@ -13,7 +13,7 @@ sanity exec path/to/script.js
 Let's start with a complete example:
 
 ```js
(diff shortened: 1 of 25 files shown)


bjoerge (approved):
💯 🙇🏼

PR #4150, opened 2023-02-01
Simplify and isolate concerns in the PTE

skogsmaskin:
### Description

The `PortableTextEditor` main component was very big. It contained not only all the static API methods for the PTE, but also a lot of logic and functions related to context and how we sync the value in and out of the editor.

This has made it hard to understand, maintain and further develop this piece of crucial code of the editor. It also hindered us from using modern React features like concurrency and the power of hooks.

`PortableTextEditor` is now pretty much left with only the static API class methods while everything else is moved into new functional React components with different isolated responsibilities:

  - `Validator` - responsible for validating new values and offer resolution to errors.
  - `SlateContainer` - sets up (and encapsulates) the Slate instance.
  - `Synchronizer` - synchronizes the external value and keeps the selection state. Provides contexts for various variables we need like like selection, value, readonly etc.
  - `useSyncValue` - hook that returns a function to compare a value against the current editor state and gently update it through operational transforms if need be. Used by `Synchronizer`.
 
* Refactored how the editor plugins can subscribe and unsubscribe to stuff (like patches). Previously they subscribed automatically (as part of the constructor for PortableTextEditor  setting up the plugin chain - which became another problem in React 18 btw). Now they all can be subscribed/unsubscribed to in component hooks.

* Moved various internal components from the root to an own `./components` folder to keep things tidy.

* Fixed an issue in the realtime collaborative tests websocket server that caused flakyness (value was emitted before patches).

* Added a Portable Text Editor API method: `insertBreak` to programatically split a block at the current selection (like pressing the enter key does). It's related because we relied on this functionality in a test, where we called this method in Slate directly through the instance variable in PortableTexteEditor. This is luckily no longer possible to do! So I added it to our own API as it's a nice high level function to have anyway.

Sorry for the size of this PR, but the refactoring touched upon a lot of different aspects and parts of the code which was a bit hard to isolate.

This PR will go into this larger one: https://github.com/sanity-io/sanity/pull/4147

### What to review

* Are the names of the new components descriptive and conventional enough?
* Is the performance similar or maybe even better than before?

### Notes for release

Most of these changes are internal code quality improvements, however there is one user facing bugfix:

* Fixed a bug that could lead to duplicate inserts of blocks in the Portable Text Editor.


packages/@sanity/portable-text-editor/src/editor/plugins/createWithPatches.ts
```diff
@@ -78,192 +78,195 @@ export interface PatchFunctions {
...
-          }
-        })
-      }
-      const {apply} = editor
-      editor.apply = (operation: Operation): void | Editor => {
-        if (editor.readOnly) {
-          apply(operation)
-          return editor
+            }
+            // if (patches.length > 0) {
```
bjoerge, line 147: Should these be removed? If not, add a comment about why it's useful to keep?

packages/@sanity/portable-text-editor/src/utils/patchToOperations.ts
```diff
@@ -45,11 +45,11 @@ export function createPatchToOperations(
             ? (parsed.start1 || 0) + parsed.diffs[0][1].length
             : (parsed.start2 || 0) + parsed.length2 - distance,
       }
-      debug(
-        `DiffMatchPatch (${distance < 0 ? 'remove' : 'insert'}) at ${JSON.stringify(slatePath)}}: `,
-        JSON.stringify(point, null, 2),
-        JSON.stringify(parsed, null, 2)
-      )
+      // debug(
```
bjoerge, line 48: seems useful to have this enabled? (if put inside a function the function will only run if debug is enabled)

packages/@sanity/portable-text-editor/src/utils/patchToOperations.ts
```diff
@@ -345,8 +352,8 @@ export function createPatchToOperations(
 
   return function (editor: Editor, patch: Patch): boolean {
     let changed = false
-    debug('\n\nNEW PATCH =============================================================')
-    debug(JSON.stringify(patch, null, 2))
+    // debug('\n\nNEW PATCH =============================================================')
```
bjoerge, line 355: comment in or remove?

packages/@sanity/portable-text-editor/src/utils/patchToOperations.ts
```diff
@@ -393,6 +400,6 @@ function findLastKey(path: Path): string | null {
 }
 
 function debugState(editor: Editor, stateName: string) {
-  debug(`Children ${stateName}:`, JSON.stringify(editor.children, null, 2))
-  debug(`Selection ${stateName}: `, JSON.stringify(editor.selection, null, 2))
+  // debug(`Children ${stateName}:`, JSON.stringify(editor.children, null, 2))
```
bjoerge, line 403: comment in or remove?

bjoerge (approved):
Massive! Just had a few minor comments when reading through. Feel free to ignore. Great work!

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPatches.ts
skogsmaskin: Should be removed. Tnx!

PR #4154, opened 2023-02-03
fix(core): change design for clickable present users in dropdown

ninaandal:
### Description
The LinkIcon for clickable users in the presence menu has been replaced with a new design. The users that are not clickable are now greyed out and has text under. Users that can be clicked on, are not. 

Screenshot 2023-02-06 at 11 24 05

### What to review
Click on the presence menu and make sure users that have a presence can be clicked on. 

### Notes for release
Design change of the presence menu where non-clickable users are greyed out

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
@@ -98,16 +98,23 @@ export function PresenceMenu(props: PresenceMenuProps) {
       menu={
         <StyledMenu padding={1}>
           {hasPresence && (
-            <Stack space={2}>
-              {presence.map((item) => (
-                <PresenceMenuItem
-                  focused={focusedId === item.user.id}
-                  key={item.user.id}
-                  onFocus={handleItemFocus}
-                  presence={item}
-                />
-              ))}
-            </Stack>
+            <>
+              <Box margin={2}>
+                <Text muted size={1} weight="semibold">
+                  SEE WHERE THEY ARE
+                </Text>
+              </Box>
+              <Stack space={2}>
+                {presence.map((item) => (
+                  <PresenceMenuItem
+                    focused={focusedId === item.user.id}
+                    key={item.user.id}
+                    onFocus={handleItemFocus}
+                    presence={item}
+                  />
+                ))}
+              </Stack>
+            </>
           )}
 
           {!hasPresence && (
--- a/packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
@@ -99,23 +99,16 @@ export function PresenceMenu(props: PresenceMenuProps) {
       menu={
         <StyledMenu padding={1}>
           {hasPresence && (
-            <>
-              <Box paddingX={2} paddingY={3}>
-                <Label muted size={1}>
-                  SEE WHERE THEY ARE
-                </Label>
-              </Box>
-              <Stack space={2}>
-                {presence.map((item) => (
-                  <PresenceMenuItem
-                    focused={focusedId === item.user.id}
-                    key={item.user.id}
-                    onFocus={handleItemFocus}
-                    presence={item}
-                  />
-                ))}
-              </Stack>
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/studio/components/navbar/presence/PresenceMenuItem.tsx
```diff
@@ -79,20 +78,12 @@ export const PresenceMenuItem = memo(function PresenceMenuItem(props: PresenceLi
     >
       <Flex align="center">
         <AvatarCard>
-          <UserAvatar size={1} key={presence.user.id} user={presence.user} />
+          <UserAvatar size={1} key={presence.user.id} user={presence.user} blur={!hasLink} />
```
hermanwikner, line 81: Instead of adding a new `blur` prop to `UserAvatar` – I suggest using the already existing `status` property.

```ts
<UserAvatar status={hasLink ? 'editing' : 'inactive'} />
```

ninaandal: Uuh thank you!!

packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
```diff
@@ -98,16 +98,23 @@ export function PresenceMenu(props: PresenceMenuProps) {
...
-                <PresenceMenuItem
-                  focused={focusedId === item.user.id}
-                  key={item.user.id}
-                  onFocus={handleItemFocus}
-                  presence={item}
-                />
-              ))}
-            </Stack>
+            <>
+              <Box margin={2}>
```
hermanwikner, line 102: I suggest updating the JSX of this header slightly to use this:

```ts
<Box paddingX={3} paddingY={3}>
  <Label size={1} muted>
    See where they are
  </Label>
</Box>
```

This would also better match the design we have in the workspaces menu! (cc: @mikolajdobrucki)

Screenshot 2023-02-03 at 14 03 13

mikolajdobrucki: great catch @hermanwikner, the right version looks much better

ninaandal: Yes, that makes sense, I didn't think about `Label`, but that seems to be more in line with the Figma as well

packages/sanity/src/core/studio/components/navbar/presence/PresenceMenuItem.tsx
```diff
@@ -73,26 +72,31 @@ export const PresenceMenuItem = memo(function PresenceMenuItem(props: PresenceLi
...
 
         <Box flex={1} marginLeft={4}>
           <Text textOverflow="ellipsis">{presence.user.displayName}</Text>
+          {!hasLink && (
+            <Stack paddingTop={2}>
+              <Text size={0} muted textOverflow="ellipsis">
+                Not in a document
+              </Text>
+            </Stack>
+          )}
```
hermanwikner, line 98: Nitpick (feel free to ignore): In this situation I would wrap the display name and subtitle with `Stack` to create the vertical space. It is arguably a more correct use of `Stack` in this situation. :) 

```tsx
<Stack space={2}>
  <Text textOverflow="ellipsis">{presence.user.displayName}</Text>

  {!hasLink && (
    <Text size={0} muted textOverflow="ellipsis">
      Not in a document
    </Text>
  )}
</Stack>
```

packages/sanity/src/core/studio/components/navbar/presence/PresenceMenu.tsx
```diff
@@ -4,6 +4,7 @@ import {
   Box,
   Button,
   Card,
+  Label,
```
hermanwikner, line 7: Unused import of `Label`!

packages/sanity/src/core/studio/components/navbar/presence/PresenceMenuItem.tsx
ninaandal: Good catch! Fixed this now :)

mikolajdobrucki (approved):
lgtm! 🚀

PR #4156, opened 2023-02-06
fix(desk): focusFirst in document and dialog behavior

ninaandal:
### Description
Adds focus to the first field when opening a document. This will be the case even if the field is `readOnly`. 
This also fixes the issue of a dialog in an array being open even after switching documents. It will also make the presence update when a user navigates to a document. 

### What to review
- Open a document and see that the first field is focused. 
Also: 
- Open an array item popover in an array with `modal.type popover`
- Open another document from the document list
- Reopen the original document from the document list
- The popover is should not still be open

### Notes for release
Adds focus to the first field in a document, close dialog when navigating documents

### DIFF
--- a/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
@@ -442,7 +442,8 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   useEffect(() => {
     // Reset focus path
     setFocusPath(params.path ? pathFromString(params.path) : [])
-  }, [documentId, params.path])
+    onSetOpenPath([])
+  }, [params.path, documentId])
 
   return (
     <DocumentPaneContext.Provider value={documentPane}>{children}</DocumentPaneContext.Provider>
--- a/packages/sanity/src/desk/panes/document/documentPanel/documentViews/FormView.tsx
+++ b/packages/sanity/src/desk/panes/document/documentPanel/documentViews/FormView.tsx
@@ -1,8 +1,7 @@
 /* eslint-disable no-nested-ternary */
 
-import {isActionEnabled} from '@sanity/schema/_internal'
-import {Box, Container, Flex, Spinner, Text} from '@sanity/ui'
-import React, {useCallback, useEffect, useMemo} from 'react'
+import {Box, Container, Flex, Spinner, Text, focusFirstDescendant} from '@sanity/ui'
+import React, {useEffect, useMemo, useRef} from 'react'
 import {tap} from 'rxjs/operators'
 import {useDocumentPane} from '../../useDocumentPane'
 import {Delay} from '../../../../components/Delay'
@@ -11,7 +10,6 @@ import {
   DocumentMutationEvent,
   DocumentRebaseEvent,
   FormBuilder,
-  PatchEvent,
   PatchMsg,
   PresenceOverlay,
   createPatchChannel,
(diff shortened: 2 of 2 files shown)


packages/sanity/src/desk/panes/document/documentPanel/documentViews/FormView.tsx
```diff
@@ -105,6 +103,12 @@ export function FormView(props: FormViewProps) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [hasRev])
 
+  const formRef = useRef<null | HTMLDivElement>(null)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
hermanwikner: Yes, it would also work to just have `onSetOpenPath([])` there instead!

However, I think it would be nice to have the focus management logic in one place instead of two. Right now focus management is done both in `FormView` (with `focusFirstDescendant`) and in `DocumentPaneProvider` (with `setFocusPath`). Having this logic in one place makes it easier to maintain going forward, and minimizes the risk of the focus logic in each component interfering with each other.

Using `focusFirstDescendant` will in most cases result in an input being focused, which will result in the presence location being updated. However, if the first focusable element in the form is not an input that is part of the schema, but perhaps some custom element at the top of the form, the presence location will not be updated when the document is opened. In order for the presence location to be updated, an input must be focused in the form that is part of the schema. Therefore, it may be better to use `focusPath` instead of `focusFirstDescendant` to set the initial focus to ensure presence is updated correctly.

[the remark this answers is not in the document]
ninaandal: Okay, I'm convinced 😊 I'll commit the code that you suggested instead then 👍

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -438,11 +438,21 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
   useEffect(() => {
-    // Reset focus path
-    setFocusPath(params.path ? pathFromString(params.path) : [])
-  }, [documentId, params.path])
+    // The path that is specified in the URL
+    const paramsPath = params.path ? pathFromString(params.path) : []
+
+    // The first field in the schema
+    // TODO: consider adding logic to check if the first field is enabled
+    const firstField = schemaType?.fields?.[0].name
```
hermanwikner, line 447: I think we could do something like this to make sure that the field is not hidden. The `members` array do not contain any fields that are hidden.

```ts
const firstMemberFieldName = (formStateRef.current?.members?.[0] as FieldMember)?.name
```

ninaandal: I'm still not convinced that we should set the focus element in `DocumentPaneProvider.tsx` instead of in the `FormView.tsx`. 
In your suggestion here: if the first member does not have a `name` field, i.e. if it instead has objects, it will be `undefined`. I think it will be cleaner to use `focusFirstDescendent` to avoid covering edge cases. Even if this will affect the presence location

ninaandal: `focusFirstDescendant` is currently used in `Dialog` as well, so using in the `FormView` is more consistent

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -442,7 +442,8 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   useEffect(() => {
     // Reset focus path
     setFocusPath(params.path ? pathFromString(params.path) : [])
-  }, [documentId, params.path])
+    onSetOpenPath([])
+  }, [params.path, schemaType?.fields, documentId])
```
bjoerge, line 446: Unless I'm missing something, `schemaType?.fields` is not needed as a dependency here?
```suggestion
  }, [params.path, documentId])
```

bjoerge (approved):
Just a minor comment, otherwise LGTM!

PR #4161, opened 2023-02-07
fix(core): add readOnly prop to sortable array items

ninaandal:
### Description
It was possible to sort arrays and publish even when `readOnly` was set. Fixed by adding the prop `readOnly` to `sortable`. 

### What to review
Add `readOnly` prop to an array and make sure it is not sortable. 

### Notes for release
Disables sorting of arrays when `readOnly` is set.


bjoerge (approved):
Great!

PR #4174, opened 2023-02-13
fix: revert client `getUrl` and `getDataUrl` usage

robinpyon:
# Description

This PR fixes a regression introduced in 4b4dc5c which would prevent users from being able to delete documents.

[This specific line](https://github.com/sanity-io/sanity/commit/4b4dc5c12609959d4998c949d16c6d4bcd0a6cf8#diff-929baa0fbd6c67766add67f4af183d565c28bf30659b9ecbc26b21cca4201fe0R98) is the cause of the issue, which expects the project dataset in its path.

However, since `getUrl` and `getDataUrl` were [re-introduced in @sanity/client 5.2.0](https://github.com/sanity-io/client/releases/tag/v5.2.0), all this PR does is revert earlier changes which previously removed calls to `client.getUrl` and `client.getDataUrl`.

# What to review

Touched areas should work as expected:
- You should be able to delete documents
- Vision fetching / listening
- Cross dataset reference inputs

# Notes for release

(None)


rexxars (approved):
Grazie!

PR #4177, opened 2023-02-15
fix(portable-text-editor): lacking .inlineBlock PT-input customization

skogsmaskin:
### Description

This will fix an issue with lacking support for the PT-input's  custom `.inlineBlock`-rendering via the `.components` schema extensions for object like types.

While doing this work, I've also used to opportunity to streamline and enhance these PT-input related rendering components.

Previously `renderDefault` just returned a static default rendering in these components, and the props it received were very limited. The props are now more complete and aligned with how the array customization support works.

These components will now also receive a `children` props which is the editing interface for the attached object similar to the array customization support.

Here is an example:

```ts
    inlineBlock: (props: BlockProps) => {
      return (
        <Box>
          {props.renderPreview(...)}
          {
            <Popover
              boundaryElement={props.__unstable_boundaryElement}
              constrainSize
              open={props.focused}
              placement="top"
              portal="editor"
              referenceElement={props.__unstable_referenceElement}
              scheme="light"
              content={
                <Box padding={1}>
                  <Inline space={1}>
                    <Button mode="bleed" padding={2} onClick={props.onOpen}>
                      Open
                    </Button>
                    <Button mode="bleed" padding={2} onClick={props.onRemove}>
                      Remove
                    </Button>
                    <Button
                      mode="bleed"
                      padding={2}
                      onClick={() => alert(JSON.stringify(props.value))}
                    >
                      Display value
                    </Button>
                  </Inline>
                </Box>
              }
            />
          }
          {props.open && (<Dialog>{props.children)}</Dialog>)}
        </Box>
      )
    },
```

Also in this PR is a fix to `portable-text-editor` where the `focused` and `selected` props were not correctly given to the `renderAnnotation` callback and thus also became wrong in the customization support as we forward these. We need to track the state of this on our own as `useSelected` in Slate only is for Element nodes and not Leaf nodes.

### Note
https://github.com/sanity-io/sanity/pull/4179 is also included in this PR, as the work overlapped on the files being worked on at the moment.

### What to review

That the code is following good practices and if there are improvements to be made.

That the default editing interfaces are working correctly for blocks, inline-blocks and annotations in the PT-input.

That the customized editing interfaces are working correctly for blocks, inline-blocks and annotations in the PT-input in various compositions.

### Notes for release
* Improve render customization support for Portable Text blocks, inline blocks and text annotations.
* Fix an issue where dragging inline objects inside the Portable Text Input would cause these elements to be copied and not moved where you dropped them.


packages/@sanity/portable-text-editor/src/types/editor.ts
```diff
@@ -383,11 +383,9 @@ export interface BlockListItemRenderProps {
   editorElementRef: React.RefObject<HTMLElement>
   focused: boolean
   level: number
-  nextItem?: string
   path: Path
-  previousItem?: string
```
skogsmaskin, line 388: The `nextItem` and `previousItem` was never meant to be there. Experimental types that accidentially got checked in previously.

packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -227,7 +227,7 @@ export function TextBlock(props: TextBlockProps) {
     setTimeout(() => PortableTextEditor.focus(editor))
   }, [editor, path])
 
-  const CustomComponent = type.components?.block as ComponentType<BlockProps>
+  const CustomComponent = schemaType.components?.block as ComponentType<BlockProps>
```
bjoerge, line 230:
```suggestion
  const CustomComponent = schemaType.components?.block as (ComponentType<BlockProps> | undefined)
```

bjoerge (approved):
Left a few suggestions/comments, otherwise lgtm!

bjoerge (commented):
Really great work! I love that we're now reaching feature parity and API alignment with the array input customization surface. Left a few comments inline, please take a look - the only thing I'd say is a blocker is whether the `type` => `schemaType` rename in various places is breaking and needs addressing.

packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
```diff
@@ -105,15 +105,70 @@ export function useSyncValue(
...
-                  debug('Replacing changed block', oldBlock)
-                  const currentSel = slateEditor.selection
-                  Transforms.deselect(slateEditor)
-                  Transforms.removeNodes(slateEditor, {at: [i]})
-                  withPreserveKeys(slateEditor, () => {
-                    Transforms.insertNodes(slateEditor, c, {at: [i]})
-                  })
-                  if (currentSel) {
-                    Transforms.select(slateEditor, currentSel)
+                  // If the blocks have the same _key, update it.
```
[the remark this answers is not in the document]
skogsmaskin: Yeah, this became rather hard to read. I have now made an effort to make it a little more readable by extracting to  functions, more meaningful variable names and some comments.

PR #4179, opened 2023-02-16
fix(portable-text-editor): dragging inline objects

skogsmaskin:
### Description

This code is left over from a time when Slate-react handled drag and drop differently.
It seems to be causing inline objects to be duplicated, and that is not the behaviour we want.

When dragging and dropping inline objects in the PT-input the edit popover was open in situations it should not. Also fix this.

### What to review

That inline-blocks can be dragged and dropped as expected.

### Notes for release

* Fix an issue where dragging inline objects inside the Portable Text Input would cause these elements to be copied and not moved where you dropped them.


hermanwikner (approved):
Nice, thank you! It works as expected 👍

PR #4183, opened 2023-02-17
fix(core): move theme logic to component and use in both navbars

ninaandal:
### Description
The "Switch to light/dark mode" and "Manage project" options disappeared when resizing the window. The logic for the theme has been moved in separate component to be reused in both navbars, both for full screen and smaller screens. 

Clipboard 2023-16-02 at 3 48 48 PM

### What to review
Make sure the options appear in both small screens and big screens and that the buttons work as expected. 

### Notes for release
Adds theme options and manage project option to navbar on small screens


hermanwikner:
The design is a bit broken when it comes to spacing (see before and after picture)

Screenshot 2023-02-20 at 13 39 40

packages/sanity/src/core/studio/components/navbar/appearance/appearance.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <Stack as="ul" space={[1, 2]}>
+        <Box padding={2}>
+          <Label size={1} muted>
+            Appearance
+          </Label>
+        </Box>
+
+        <Stack as="li" key="System">
+          <StyledButton
```
hermanwikner, line 26: This can't be a `Button` since it breaks the keyboard navigation in `UserMenu`.

packages/sanity/src/core/studio/components/navbar/tools/ToolVerticalMenu.tsx
```diff
@@ -16,32 +16,34 @@ export function ToolVerticalMenu(props: ToolVerticalMenuProps) {
 
   return useMemo(
     () => (
-      <Stack as="ul" space={[1, 2]}>
-        {tools.map((tool) => {
-          const title = tool?.title || startCase(tool.name) || undefined
+      <>
```
hermanwikner, line 19: Why do we need to wrap this in a fragment?

packages/sanity/src/core/studio/components/navbar/appearance/appearance.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (!theme?.__legacy)
+    return (
+      <Stack as="ul" space={[1, 2]}>
+        <Box padding={2}>
+          <Label size={1} muted>
+            Appearance
+          </Label>
+        </Box>
+
+        <Stack as="li" key="System">
```
robinpyon, line 25: Is `key` necessary here? 

minor: I also wonder if using `<Grid>` over `<Stack>` is more appropriate here as it may be a better signifier of intent. Seeing a `<Stack>` component makes me think that this _could_ contain multiple buttons within, which would never be the case.

hermanwikner:
Since we want to use `MenuItem` in `UserMenu` and `Button` in `NavDrawer` – I think it's best to implement them differently in these two places since they should behave different. I suggest we update `useColorScheme` with the following `colorSchemeOptions` (not sure about the name):

```tsx
interface ColorSchemeOption {
  icon: React.ComponentType
  name: 'light' | 'dark' | 'system'
  onSelect: () => void
  selected: boolean
  title: string
}

export function ColorSchemeProvider() {
  const colorSchemeOptions = useMemo(() => {
    const options: ColorSchemeOption[] = [
      {
        title: 'System',
        name: 'system',
        selected: usingSystemScheme,
        onSelect: clearStoredScheme,
        icon: DesktopIcon,
      },
      {
        title: 'Dark',
        name: 'dark',
        selected: scheme === 'dark' && !usingSystemScheme,
        onSelect: () => handleSetScheme('dark'),
        icon: MoonIcon,
      },
      {
        title: 'Light',
        name: 'light',
        selected: scheme === 'light' && !usingSystemScheme,
        onSelect: () => handleSetScheme('light'),
        icon: SunIcon,
      },
    ]

    return options
  }, [clearStoredScheme, handleSetScheme, scheme, usingSystemScheme])

  return (
    <ColorSchemeContext.Provider
      value={{
        colorSchemeOptions,
        // ...rest
      }}
    >
      {/* ... */}
    </ColorSchemeContext.Provider>
  )
}
```

Then it will be easy to loop out these options with `MenuItem` or `Button`:

```tsx
function UserMenu() {
  const {colorSchemeOptions} = useColorScheme()

  return (
    <Stack>
      {/* // stuff... */}

      {colorSchemeOptions.map((option) => (
        <MenuItem
          aria-label={`Use ${option} appearance`}
          icon={option.icon}
          iconRight={option.selected && <CheckmarkIcon />}
          key={option.name}
          onClick={() => option.onSelect()}
          pressed={option.selected}
          text={option.title}
        />
      ))}

      {/* // stuff... */}
    </Stack>
  )
}
```

```tsx
function NavDrawer() {
  const {colorSchemeOptions} = useColorScheme()

  return (
    <Stack>
      {/* // stuff... */}

      {colorSchemeOptions.map((option) => (
        <Button
          aria-label={`Use ${option} appearance`}
          icon={option.icon}
          iconRight={option.selected && <CheckmarkIcon />}
          key={option.name}
          onClick={() => option.onSelect()}
          selected={option.selected}
          text={option.title}
        />
      ))}

      {/* // stuff... */}
    </Stack>
  )
}
```

packages/sanity/src/core/studio/components/navbar/appearance/appearance.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import {Box, Button, Label, Stack, useRootTheme} from '@sanity/ui'
+import {CheckmarkIcon, DesktopIcon, MoonIcon, SunIcon} from '@sanity/icons'
+import styled from 'styled-components'
+import {useColorScheme} from '../../../colorScheme'
+import {StudioTheme} from '../../../../theme'
+
+const StyledButton = styled(Button)`
+  cursor: pointer;
```
robinpyon, line 9: The current convention in the studio is that only buttons with URLs should have a hand cursor. Given that these are commands, we can skip creating a custom style and remove this altogether.

packages/sanity/src/core/studio/components/navbar/userMenu/UserMenu.tsx
```diff
@@ -114,44 +104,7 @@ export function UserMenu() {
...
-                onClick={clearStoredScheme}
-                pressed={usingSystemScheme}
-                text="System"
-                iconRight={usingSystemScheme && <CheckmarkIcon />}
-              />
-
-              <MenuItem
-                aria-label="Use dark appearance"
-                icon={MoonIcon}
-                onClick={() => setScheme('dark')}
```
robinpyon, line 139: minor: we should try and avoid using inline functions (eslint should display a warning here)

robinpyon:
@mikolajdobrucki Do you think there's potential for us to remove the more spaced out layout in the drawer (as demonstrated in the below video) and simply use the small padding value across all breakpoints?

Rationale being that we don't have button arrangements in the studio with that much padding (to my knowledge at least!)

https://user-images.githubusercontent.com/209129/220135458-2c40353c-85a9-4b26-82c5-91a83e262031.mp4

robinpyon:
We'll also need to consider overflow on constrained heights! 

Ideally, the header (with name and workspace select) as well as the footer (sign out button) are always visible, and any overflow content in-between being scrollable.

image

mikolajdobrucki (approved):
Looks great, good to go from my end! 🚀

robinpyon (approved):
Thank you @ninaandal, this looks good!

A few comments explaining some of my recent commits:
- The `StyledCard` isn't necessary as you can set `flex="none"` on Sanity UI `Card` elements. (Sorry – I gave you bad information here, though in fairness the docs may need some updating!)
- Conditionally added `tabIndex` on appearance buttons in `NavDrawer`, which are based off the drawer's `open` state. (These were previously set to `0` which made them untabbable)
- Also removed the `as="a"` from appearance `<Button>` elements – these aren't necessary as we don't need to treat them as `<a>` elements (they're not external links but rather application _commands_). A by-product of using `as="a"` without a corresponding `href` also meant that you couldn't use your keyboard ENTER key to trigger these.

PR #4186, opened 2023-02-18
fix(cli): provide consistent dotenv loading behavior

rexxars:
### Description

Our dotenv (`.env`) handling currently only loads the `.env.[mode]` file, not `.env` nor `.env.local` and similar. We want it to be consistent with how Vite loads them, since it is a reasonable implementation:

```
.env                # loaded in all cases
.env.local          # loaded in all cases, ignored by git
.env.[mode]         # only loaded in specified mode
.env.[mode].local   # only loaded in specified mode, ignored by git
```

This PR ensures:

1. New projects gets an ignore pattern for `.local` files, mirroring vite projects
2. Loads environment variables from dotenv files on CLI startup
    - In v3 projects and above, it will use the "Vite" way, loading and merging all found dotenv files
    - In v2 projects, it will use the current behavior
3. Environment/"mode" is determined in the following way:
    1. If `SANITY_ACTIVE_ENV` is set, use it
    2. If the command being run is `build` or `deploy`, use `production`
    3. If `NODE_ENV` is set to `production`, use `production`
    4. Otherwise, use `development`
4. The environment variables can be used in scripts using `sanity exec`
5. Users can get the studio-scoped variables (and dotenv handling) in their own apps by using the new `getStudioEnvironmentVariables()` method
6. Always defines `process.env.MODE`, mirroring Vite's `import.meta.env.MODE`, and ensures that the `process.env` object is defined in development even if no `SANITY_STUDIO_` prefixed variables are set, which calls for less surprises.
7. The behavior is tested in both `sanity dev`, `sanity build` and `sanity exec`. Also ensures the old behavior is kept in v2.

### Review notes

1. The updated dotenv and dotenv-expand dependencies are needed to mirror Vite's behavior
2. There is some weirdness when Jest tries to use worker threads, so we're checking for it and not forking. This means that there's a slightly different code path being run when loading the document, which means we do not have the loaders available to do full TS/JSX transformations and such. Thus, there is a `createElement` based `EnvDocument.tsx` in the test fixtures which will look very odd. I want to address this at some point, but cannot afford to spend any more time on it right now.

### Notes for release

- Makes the behavior of loading `.env` files consistent with Vite, and ensures environment variables set through these files can be used in the CLI and studio configuration, as well as in scripts executed through `sanity exec`.
- Introduces a new `process.env.MODE` variable to the Vite build, mirroring `import.meta.env.MODE`. This has the side effect of fixing a difference in behavior when there are/aren't any defined `SANITY_STUDIO_` prefixed environment variables, and attempting to read one.
- Added a new `getStudioEnvironmentVariables()` method to the `sanity` module. It can optionally be told to load the `.env` files. This can come in handy if you want to mirror how Sanity loads environment variables outside of Sanity/Vite, for instance if using a different bundler.


bjoerge (approved):
Looks super solid! Thank you for taking the time to write tests as well <3

packages/sanity/src/_internal/cli/server/getStudioEnvironmentVariables.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @public
+ */
+export function getStudioEnvironmentVariables({
+  prefix = '',
+  envFile = false,
+  jsonEncode = false,
+}: {
+  prefix?: string
+  envFile?: {mode: string; envDir?: string} | false
+  jsonEncode?: boolean
```
robinpyon, line 31: minor: Should it be possible to call `getStudioEnvironmentVariables` without any params?

```suggestion
  prefix,
  envFile,
  jsonEncode,
}: {
  prefix?: string
  envFile?: {mode: string; envDir?: string} | false
  jsonEncode?: boolean
} = {
  prefix: '',
  envFile: false,
  jsonEncode: false,
}
```

robinpyon (approved):
This looks great! I'm not super familiar with the CLI package, but came across no issues when testing locally

Issue #4074, opened 2023-01-11
[v3] Safari doesn't render custom JSX document icons


genox:
I use custom JSX icons extensively. Safari doesn't show them:

<img width="713" alt="Screenshot 2023-01-11 at 11 13 33" src="https://user-images.githubusercontent.com/5611407/211778949-e46e671e-3bb4-4ac3-b77e-c19fe76ccab7.png">

Here's an example icon component:

```
export const IconPerson = () => (
  <svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" strokeWidth={1}>
    <path
      d="M6.750 6.000 A5.250 5.250 0 1 0 17.250 6.000 A5.250 5.250 0 1 0 6.750 6.000 Z"
      fill="none"
      stroke="currentColor"
      strokeLinecap="round"
      strokeLinejoin="round"
      vectorEffect="non-scaling-stroke"
    />
    <path
      d="M2.25,23.25a9.75,9.75,0,0,1,19.5,0"
      fill="none"
      stroke="currentColor"
      strokeLinecap="round"
      strokeLinejoin="round"
      vectorEffect="non-scaling-stroke"
    />
  </svg>
);
```

The icons look fine in Chrome and Firefox on Mac, the same icons also render perfectly fine in NextJS on Safari.

I use the embedded version of Sanity within NextJS.

sanity version: 3.1.4

PR #4197, opened 2023-02-22
fix(core): explicitly set SVG size when used as preview media

pauloborges:
### Description

This PR makes small CSS adjustments to the `MediaWrapper` component used in list previews.

The first commit addresses the issue #4074, in which SVGs need to have their sizing "explicitly" defined in Safari.

The second commit takes the opportunity to do a small cleanup on the same `MediaWrapper` component, this time for images, removing "unused" CSS rules.

### What to review

This PR should not change any existing behavior, apart from making SVGs visible on list previews when using Safari.

I tested it:

* With SVGs of different view box sizes, including non-squared ones.
* With different layouts on the list component (compact and detailed view).
* On the different "lists" across Sanity Studio (e.g. array fields, references).

### Notes for release

Something along the lines of:

> Fixes usage of SVG elements inside list previews.

### DIFF
--- a/packages/sanity/src/core/components/previews/_common/Media.styled.ts
+++ b/packages/sanity/src/core/components/previews/_common/Media.styled.ts
@@ -37,7 +37,7 @@ export const MediaWrapper = styled.span<{
     }
 
     & svg {
-      display: block;
+      flex: 1;
       font-size: calc(21 / 16 * 1em);
     }
 
--- a/packages/sanity/src/core/components/previews/_common/Media.styled.ts
+++ b/packages/sanity/src/core/components/previews/_common/Media.styled.ts
@@ -15,7 +15,6 @@ export const MediaWrapper = styled.span<{
   const iconSize = PREVIEW_ICON_SIZE[$layout]
 
   return css`
-    position: relative;
     width: ${$responsive ? '100%' : rem(width)};
     height: ${$responsive ? '100%' : rem(height)};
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/components/previews/_common/Media.styled.ts
```diff
@@ -26,17 +25,14 @@ export const MediaWrapper = styled.span<{
...
-      top: 0;
       width: 100%;
       height: 100%;
       object-fit: contain;
       border-radius: inherit;
     }
 
     & svg {
-      display: block;
+      flex: 1;
```
hermanwikner, line 35: Thank you! @pauloborges

I think changing from `display: block` to `flex: 1` should be enough to fix the issue. Removing `position: relative` and the CSS from `& img` introduces some issues with the absolute positioned `span` that renders the outline around the preview media when it's an `img` (see image):

Screenshot 2023-06-29 at 10 27 43

hermanwikner (changes requested):

packages/sanity/src/core/components/previews/_common/Media.styled.ts
pauloborges: Hello! Thanks for reviewing it. Do you mind taking this over? You may only need to drop the last commit.

hermanwikner:
Closing this as the issue is addressed in #4838.

PR #4201, opened 2023-02-23
fix(core): invalid reference link

hermanwikner:
### Description

This PR fixes an issue where an error is thrown if a reference link is clicked while it is loading.

### What to review

Click a reference link when it is loading. An error should not be thrown.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
@@ -25,10 +25,10 @@ import {useScrollIntoViewOnFocusWithin} from '../../hooks/useScrollIntoViewOnFoc
 import {useDidUpdate} from '../../hooks/useDidUpdate'
 import {set, unset} from '../../patch'
 import {AlertStrip} from '../../components/AlertStrip'
-import {ReferencePreviewCard} from './ReferenceItem'
 import {useReferenceInput} from './useReferenceInput'
 import {useReferenceInfo} from './useReferenceInfo'
 import {PreviewReferenceValue} from './PreviewReferenceValue'
+import {ReferenceLinkCard} from './ReferenceLinkCard'
 import {IntentLink} from 'sanity/router'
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
```diff
@@ -273,7 +281,7 @@ export function ReferenceField(props: ReferenceFieldProps) {
             <Flex gap={1} align="center">
               <ReferencePreviewCard
                 flex={1}
-                forwardedAs={EditReferenceLink as any}
+                forwardedAs={forwardedAs}
                 tone="inherit"
                 radius={2}
                 data-as="a"
```
robinpyon, line 287:
```suggestion
                data-as={refType?.name ? 'a' : undefined}
```

What do you think about the above? Conditionally setting `data-as` will mean that the preview card won't show a hand cursor prior to the reference type being known

Example video below (red = no ref type, green = ref type loaded)

https://user-images.githubusercontent.com/209129/221065866-24659d16-2ba1-4ccc-b2df-e19e2fa24733.mp4

_Ideally_ the link wouldn't be clickable until the full preview is loaded in its entirety, but that's probably a story for another time!

robinpyon:
Thanks @hermanwikner! 

Just tested and didn't have issues, though the issue persists outside of reference fields (e.g. references in arrays). As such, it may be worth applying this change in the [ReferenceItem](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceItem.tsx) component?

packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
hermanwikner: Good point, will add that 👍

robinpyon (approved):
Looks great @hermanwikner, thank you!

Issue #4192, opened 2023-02-21
Not possible to type 0 after decimal in number input


rexxars:
**Describe the bug**

When attempting to type a number that has a zero after the decimal in the default number input, the zero is removed.

**To Reproduce**

1. Enter `1.` in a number input
2. Enter a zero in the input
3. See the field revert back to `1.`

**Expected behavior**

Can enter a zero after the decimal (at least until I am able to type the next digit)

**Which versions of Sanity are you using?**

v3.5.0

**What operating system are you using?**

MacOS

**Which versions of Node.js / npm are you running?**

Not relevant, but Node 18, npm 8

PR #4206, opened 2023-02-24
fix(form): Allow fractional part of a number to have leading zeros

pauloborges:
### Description

This PR addresses the issue #4192. In which is impossible to type numbers that have leading zeros in their fractional part, e.g. `1.0001`.

This happens because every input change re-renders the form using a "normalized" value. For the case of a number, that is `HTMLInputElement. valueAsNumber`, which will happily parse a string of value `"1.000"` into the number `1`.

The first commit adds a few unit tests around the `PrimitiveField` component, to establish some base ground for refactoring. It also includes a test that captures this behaviour, and which fails without the proposed fix.

The second commit implements the fix, by introducing a piece of local state to hold the input value, as typed. On re-renders, this value is only updated if the underlying number differs from the component incoming prop.

(I haven't added tests for the `ArrayOfPrimitivesItem` component. If this proposed solution is accepted, and if you think it's valuable, I can definitely add them!)

Here's a video demonstrating the result of the proposed fix:

https://user-images.githubusercontent.com/424015/221186323-4b0584a8-f42b-420b-a606-aab851fbc9c1.mov

### What to review

It's important to verify that `PrimitiveField` and `ArrayOfPrimitivesItem` still work as expected for `boolean` and `string` values.

### Notes for release

Something along the lines of:

> Allow fractional part of a number to have leading zeros.

### DIFF
--- a/packages/sanity/src/core/form/inputs/NumberInput.tsx
+++ b/packages/sanity/src/core/form/inputs/NumberInput.tsx
@@ -31,6 +31,7 @@ export function NumberInput(props: NumberInputProps) {
       pattern={onlyPositiveNumber ? '[d]*' : undefined}
       max={Number.MAX_SAFE_INTEGER}
       min={Number.MIN_SAFE_INTEGER}
+      data-testid="number-input"
     />
   )
 }
--- a/packages/sanity/src/core/form/members/array/items/ArrayOfPrimitivesItem.tsx
+++ b/packages/sanity/src/core/form/members/array/items/ArrayOfPrimitivesItem.tsx
@@ -1,5 +1,5 @@
-import {isBooleanSchemaType, isNumberSchemaType} from '@sanity/types'
-import React, {useCallback, useMemo, useRef} from 'react'
+import {isBooleanSchemaType, isNumberSchemaType, isStringSchemaType} from '@sanity/types'
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import {FIXME} from '../../../../FIXME'
 import {ArrayOfPrimitivesItemMember} from '../../../store'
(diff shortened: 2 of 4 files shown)


bjoerge (commented):
Hi @pauloborges - really appreciate this! Thank you for taking the time to add tests as well <3

I've reviewed and have a few comments:
- I understand that there's an inconsistency in how `value` is passed to both boolean inputs and string/number inputs, but  I'm not sure if we have any practical issue with it and - unless I'm missing something obvious - it's not strictly related to the issue described in #4192, so I'd prefer to keep it out of this PR.
- I noticed that if I type 0.0000, the number input renders as empty if I reload the page (or watch it from another tab). It should render `0`.
- The string representation of the numeric input value should only be forwarded as `elementProps.value` only. `inputProps.value` should always be the normalized numeric value, otherwise this would be a breaking change in that it changes the type from number to string.
- I also wonder if we can get rid of the useEffect that syncs the native value and instead use the native value whenever we see that it's different but normalizes to the same number as the one received through props.

Hope that makes sense! I'd be happy to do these changes myself unless you have time and want to give it a go.

pauloborges:
Hi @bjoerge, thanks a lot for the detailed review!

I've now pushed a couple of `FIXUP` commits to address your comments. I kept them as separate commits to help clarify what I changed, but I intend to squash them to the original commits if they get approved.

As for the suggestion to remove the synchronization effect: excellent idea! One drawback is that the code, at least for me, needs a bit extra time to parse/understand. I implemented it in a different commit. Please have a look!

Happy to discuss it more!

bjoerge:
Great! I've done some minor tweaks, squashed your commits and merged to main. This fix will be included in next week's release. Thanks again for taking the time, really appreciate it 🙇🏼

PR #4207, opened 2023-02-24
fix(desk): document list menu items selected state

hermanwikner:
### Description

This PR adds a selected state on the document list menu items.

Screenshot 2023-02-24 at 14 58 22

### What to review

n/a

### Notes for release

n/a


robinpyon:
Thanks @hermanwikner, this is sorely needed!

I did a pass on the test studio and have noticed a few visual issues –

### Both layout options are marked as selected when no Local Storage key for this pane exists

image

You should be able to re-create this by clearing all `desk-tool:` prefixed keys in LS.

Ideally, we should be able to infer the default pane layout if no LS key exists (which is currently `"default"`)

This could be potentially done by extracting the `defaultLayout` pane prop [here](https://github.com/sanity-io/sanity/blob/d5166e52d1669170036adb98b8f88804c7e2c337/packages/sanity/src/desk/panes/documentList/DocumentListPane.tsx#L40) and export them from document lists' [constants](https://github.com/sanity-io/sanity/blob/d5166e52d1669170036adb98b8f88804c7e2c337/packages/sanity/src/desk/panes/documentList/constants.ts).

### Pane sort order doesn't appear to be respected until the `::sortOrder` key is set

Example video below. 

https://user-images.githubusercontent.com/209129/221527803-a2b3efd3-3324-45fd-b966-f90201fc310c.mp4

### Both last edited  and created sort options are both marked as selected when either is enabled

image

Similar to the first point, we should be able to infer the default sort order (exported in [constants](https://github.com/sanity-io/sanity/blob/d5166e52d1669170036adb98b8f88804c7e2c337/packages/sanity/src/desk/panes/documentList/constants.ts)) and mark that as selected if no LS key exists for this document list pane just yet.

hermanwikner:
Good catch, thanks @robinpyon! 👍 I've updated the PR

robinpyon (approved):
Looks great @hermanwikner, thank you!

PR #4216, opened 2023-02-27
fix(core): search performance improvements

robinpyon:
### 1. Prevent search field definitions from being re-created on every render

An oversight with `useMemo` dependencies in `<SearchProvider>` meant that global search was regenerating search field definitions _on every render_.

This wasn't noticeable in most average sized studios (<1K fields), but was causing larger studios (15K+ fields) to slow to a crawl on every re-render of the component containing `<SearchProvider>`. (In this case, `<StudioNavbar>`, which currently re-renders on router state / media query changes).

This has been fixed and should now correctly only generate those search field definitions once. 

### 2. Avoid unnecessary `Array.find` lookups when working with search definitions (fields, filters, operators)

Definitions are now stored as dictionaries and subsequent lookup is now much faster.

Random sampling of `filterGroup()` execution time (test studio, ~600 fields):

```
Before:

2.7999999970197678 ms
2.5
1.4000000059604645 ms
4.5 ms
1.5 ms
1.4000000059604645 ms
1.3999999910593033 ms
1.7999999970197678 ms
1.5 ms
1.4000000059604645 ms

---

After:

0.09999999403953552 ms
0.09999999403953552 ms
0.10000000894069672 ms
0.09999999403953552 ms
0.10000000894069672 ms
0.09999999403953552 ms
0.09999999403953552 ms
0.10000000894069672 ms
0.19999998807907104 ms
0.10000000894069672 ms
```

### What to review

On extremely large studios (thousands of fields+)
- There shouldn't be any noticeable delay when `<StudioNavbar>` is re-rendered
- Interacting with global search UI (document types, add filters) should feel _much_ snappier

### Notes for release

- General performance improvements on larger studios (containing thousands of fields)


packages/sanity/src/core/studio/components/navbar/search/definitions/fields.ts
```diff
@@ -51,15 +72,30 @@ export function createFieldDefinitions(
...
+  fieldDefinitions: SearchFieldDefinition[]
+): SearchFieldDefinitionDictionary {
+  return fieldDefinitions.reduce<SearchFieldDefinitionDictionary>((acc, val) => {
+    acc[val.id] = val
+    return acc
+  }, {})
+}
+
+export function generateFieldId(field: SearchFieldDefinition): string {
+  return Md5.hashStr(
```
bjoerge, line 88: Don't have the full context here, but could this be simplified to `[field.documentTypes.join(','), field.fieldPath, field.filterName, field.type].join('-')`?

bjoerge (approved):
Fantastic improvements ⚡️⚡️⚡️

Just a tiny comment/suggestion, otherwise LGTM!

packages/sanity/src/core/studio/components/navbar/search/definitions/fields.ts
robinpyon: Great callout – this hashing is completely unnecessary, it's not used for any comparison and the joined string is already unique by design. That's been changed!

PR #4220, opened 2023-02-28
Fix preserve key issue

skogsmaskin:
### Description

There was a peculiar issue in the Portable Text Editor which would cause wrong keys being set on split nodes when typing fast and at the same time toggling decorators (like strong, italic) as you typed. This bug was introduced in v.3.4.0. as the changes there introduced usage of `withPreserveKeys` in syncing and resolving the server value to the editor state.

The way the helper method was, it used the prev state as toggle for some reason (it's been like this for a long time), but with potentially overlapping calls to this helper, it could resolve the wrong state.

The fix is to hardcode the value either true or false, and not rely on the prev state.

I also moved the `withPreserveKeys` scope in the  syncValue function to apply as narrow as possible so we don't use this scope unnecessary.

### What to review

In the current version you can trigger this bug by writing fast and at the same time toggling between bold and regular with the keyboard shortcut Cmd+B.

With this fix the bug should no longer possible to reproduce.

### Notes for release
* Fixed a bug in the Portable Text Editor introduced in v. 3.0.4 that could occur when typing fast and toggling text decorators as you typed (like bold).


robinpyon:
Thanks @skogsmaskin!

I'm currently lacking the PTE knowledge to give meaningful feedback on what you've written – however, can confirm that the main issue is no longer reproducible.

bjoerge (approved):
LGTM!

PR #4222, opened 2023-02-28
feat(cli): use local callback server for auth instead of listener

rexxars:
### Description

Replaces the current login flow, which is based on an EventSource listener, with a more standard approach:

1. User initiates login procedure with `sanity login`
2. User is prompted for login provider
3. An HTTP server is started in the background on the users computer
4. User is sent to login provider URL, with an `origin` (redirect uri) set to `http://localhost:port/callback`
5. Once authenticated, user is redirected to callback URL, with a temporary session ID in URL params
6. Temporary session ID is exchanged for a token from API
7. Local HTTP server is closed
8. User's browser is redirected to `https://www.sanity.io/login/success`

[sc-29620]

### What to review

- Login flow works (`sanity login`)
- SSO login flow works (`sanity login --sso :organization-slug`)

To run, you need to build the CLI and run the sanity CLI from the right location:
```
cd ~/path/to/sanity-checkout
yarn build
./packages/@sanity/cli/bin/sanity login
```

### Notes for release

Probably not worth mentioning, as to the end user the behavior is the same as it was previously.


packages/@sanity/cli/src/actions/login/login.ts
```diff
@@ -106,77 +130,91 @@ export async function login(
...
+      const tokenPath = `${tokenPathName}${tokenUrl.search}`
+
+      let token: TokenDetails
+      try {
+        token = await apiClient({requireUser: false, requireProject: false})
+          .clone()
+          .request({uri: tokenPath})
+      } catch (err) {
+        failLoginRequest()
+        rejectToken(err)
```
dcilke, line 184: This feels a bit heavy handed given we have a fully qualified url for fetching the token. If there are additional benefits to using the `apiClient` instead of fetch, we can craft the `/auth/fetch` uri manually as one of the url parameters is the `sid`. So

```js
const sessionId = url.searchParams.get('sid')
if (sessionId) {
	failLoginRequest()
  return
}

let token: TokenDetails
try {
	token = await apiClient({requireUser: false, requireProject: false})
  	.clone()
    .request({uri: `/auth/fetch?sid=${sid}`})
} catch (err) {
	failLoginRequest()
	rejectToken(err)
}
```

packages/@sanity/cli/src/actions/login/login.ts
```diff
@@ -106,77 +130,91 @@ export async function login(
...
-  const values = {secret: '', url: ''}
   return new Promise((resolve, reject) => {
-    es.addEventListener('error', (err: MessageEvent | EventWithMessage) => {
-      es.close()
-      const message = 'message' in err && typeof err.message === 'string' ? err.message : ''
-      reject(new Error(`Unable to get authorization info: ${message}`))
-    })
+    const server = http.createServer(async function onCallbackServerRequest(req, res) {
+      function failLoginRequest() {
+        res.writeHead(303, 'See Other', {Location: `https://${domain}/login/error`})
```
dcilke, line 153: If there are any error messages we would want to display to the user, the error page accepts various error codes. (i.e `/login/error?error=NO_SESSION_ID`)

dcilke (approved):
Looks great! I've confirmed it works on both the default providers and SSO providers. 🎉

packages/@sanity/cli/src/actions/login/login.ts
dcilke: 💡 good to know!

PR #4223, opened 2023-02-28
fix(core): changing workspace with same document id causes error

### LINKED ISSUE #3992: Bug when switching workspaces in Studio
Hi, 

I'm having quite a big issue and can't understand why. 
I have a studio with 2 datasets : staging & production. I created a workspaces for each one of them with the following config : 

```
export default defineConfig([
	{
		name: "production",
		title: "Coucool - Public",
		icon: () => <RocketIcon onResize={undefined} onResizeCapture={undefined} />,
		projectId: "PROJECT_ID",
		dataset: "production",
		basePath: "/production",
		subtitle: "production",
		plugins: [sharedConfig()],
	},
	{
		name: "staging",
		title: "Coucool - Test",
		icon: () => <UserIcon onResize={undefined} onResizeCapture={undefined} />,
		projectId: "PROJECT_ID",
		dataset: "staging",
		basePath: "/staging",
		subtitle: "staging",
		plugins: [sharedConfig()],
	},
]);
```

<details>
  <summary>Click here to see the sharedConfig code </summary>

```
const sharedConfig = definePlugin({
	name: "sharedConfig",
	document: {
		actions: (prev, context) => {
			return context.schemaType === "siteSettings"
				? [
						...prev.filter(
							(originalAction) =>
								originalAction.action !== "delete" && originalAction.action !== "duplicate"
						),
				  ]
				: prev;
		},
	},
	plugins: [
		colorInput(),
		deskTool({
			structure: (S, context) =>
				S.list()
					.title("Contenu")
					.items([
						S.listItem()
							.title("Paramètres événement")
							.child(S.document().schemaType("siteSettings").documentId("siteSettings"))
							.icon(CogIcon),
						orderableDocumentListDeskItem({
							type: "section",
							S,
							context,
							title: "Sections",
							icon: () => <MasterDetailIcon onResize={undefined} onResizeCapture={undefined} />,
						}),
					]),
		}),
		visionTool(),
	],
	schema: {
		types: schemaTypes,
	},
});
```
</details>

The issue is that everytime I'm switching from one workspaces to another, it's like the content on the studio isn't updating (the URL is correctly updating btw) which leads to big errors. I'm also having this error snackbar everytime : 
Capture d’écran 2022-12-19 à 00 51 47

1. Create a new project
2. Add the same config that I sent you above
3. Try to switch workspaces

**Expected behavior**

When I switch workspaces, I should see the refreshed content and no error snackbar.

```
@sanity/cli (global)             3.0.6 (up to date)
@sanity/color-input              3.0.1 (up to date)
@sanity/eslint-config-studio     2.0.1 (up to date)
@sanity/icons                    2.1.0 (up to date)
@sanity/orderable-document-list  1.0.2 (up to date)
@sanity/ui                       1.0.1 (latest: 1.0.2)
@sanity/vision                   3.0.6 (up to date)
sanity                           3.0.6 (up to date)
```

macOS

`nom -v: 8.19.2 / node -v: v18.12.1`

Thanks a lot

binoy14:
### Description

Fixes issue where switching workspaces and navigating to document that shares the same id causes an error and didn't show document history

Fixes #3992

### What to review

1. Fix issue where the memoizedPair did not account for sanity client config changing
2. Fix multiple levels of memoization for the `remoteSnapshots`

### Notes for release

- Fix issue when changing workspaces on documents that have same id didn't show document history

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
@@ -8,18 +8,16 @@ import {checkoutPair, type Pair} from './checkoutPair'
 export const memoizedPair: (
   client: SanityClient,
   idPair: IdPair,
-  _typeName?: string
+  typeName: string
 ) => Observable<Pair> = memoize(
-  (client: SanityClient, idPair: IdPair, _typeName?: string): Observable<Pair> => {
+  (client: SanityClient, idPair: IdPair, _typeName: string): Observable<Pair> => {
     return new Observable<Pair>((subscriber) => {
       subscriber.next(checkoutPair(client, idPair))
     }).pipe(publishReplay(1), refCount())
   },
-  (client: SanityClient, idPair: IdPair, typeName?: string) => {
+  (client, idPair, typeName) => {
     const config = client.config()
 
-    return `${config.dataset ?? ''}-${config.projectId ?? ''}-${idPair.publishedId}-${
-      typeName ?? ''
-    }`
+    return `${config.dataset ?? ''}-${config.projectId ?? ''}-${idPair.publishedId}-${typeName}`
   }
 )
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
@@ -4,6 +4,7 @@ import {publishReplay, refCount} from 'rxjs/operators'
 import type {IdPair} from '../types'
 import {memoize} from '../utils/createMemoizer'
 import {checkoutPair, type Pair} from './checkoutPair'
+import {memoizeKeyGen} from './memoizeKeyGen'
 
 export const memoizedPair: (
   client: SanityClient,
@@ -15,9 +16,5 @@ export const memoizedPair: (
       subscriber.next(checkoutPair(client, idPair))
     }).pipe(publishReplay(1), refCount())
   },
-  (client, idPair, typeName) => {
-    const config = client.config()
-
-    return `${config.dataset ?? ''}-${config.projectId ?? ''}-${idPair.publishedId}-${typeName}`
-  }
+  memoizeKeyGen
 )
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
@@ -1,9 +1,9 @@
-import {SanityClient} from '@sanity/client'
+import type {SanityClient} from '@sanity/client'
 import {Observable} from 'rxjs'
 import {publishReplay, refCount} from 'rxjs/operators'
-import {IdPair} from '../types'
+import type {IdPair} from '../types'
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
```diff
@@ -15,5 +15,11 @@ export const memoizedPair: (
       subscriber.next(checkoutPair(client, idPair))
     }).pipe(publishReplay(1), refCount())
   },
-  (_client, idPair) => idPair.publishedId
+  (client: SanityClient, idPair: IdPair, typeName?: string) => {
+    const config = client.config()
+
+    return `${config.dataset ?? ''}-${config.projectId ?? ''}-${idPair.publishedId}-${
```
bjoerge, line 21: Passing a client without dataset/projectId here will probably cause problems down the line, so I'm wondering if it makes sense to throw an error here instead if that's the case. WDYT?

packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
```diff
@@ -15,5 +15,11 @@ export const memoizedPair: (
...
+      typeName ?? ''
```
bjoerge, line 22: I might be missing something, but since IDs are always unique and two documents of different types can't share the same ID, I'm not sure if we need to have `typeName` as part of the key here?

packages/sanity/src/core/store/_legacy/document/document-pair/remoteSnapshots.ts
```diff
@@ -1,16 +1,17 @@
...
 /** @internal */
-export const remoteSnapshots = memoize(
-  (client: SanityClient, idPair: IdPair, typeName: string) => {
-    return memoizedPair(client, idPair, typeName).pipe(
-      switchMap(({published, draft}) => merge(published.remoteSnapshot$, draft.remoteSnapshot$))
-    )
-  },
-  (_client, idPair, typeName) => idPair.publishedId + typeName
-)
+export const remoteSnapshots = (
```
bjoerge, line 9: We still need the memoize wrapper here, and it should use the same key generator as memoizedPair. 

Although removing memoization here will still benefit from using the memoized observable returned from `memoizedPair`, it will now make `remoteSnapshot` return a new observable instance every time it's called, which might break some expectations down the line.

bjoerge (changes requested):
Great find and proposed fix!

I think there's a couple of minor issues that needs to be sorted before merge, see my comments.

packages/sanity/src/core/store/_legacy/document/document-pair/memoizedPair.ts
rexxars: A document can be deleted and recreated with a different type. Would this create problems? I feel like its safer to include the typename.

bjoerge: Ah, good point, and that's something that happens when people run migration scripts renaming types, so not that rare either. Talked with @binoy14 yesterday and it seems like it's straightforward to make `typeName` required, so we will use it as part of the cache key from now on. Thanks!

PR #4224, opened 2023-02-28
fix(desk): cyrillic characters in deskStructure crashes studio

### LINKED ISSUE #4185: Sanity crashes if title contains cyrillic characters in deskStructure
Please add cyrillic support

```javascript
import {FiSettings} from 'react-icons/fi'

export const deskStructure = (S) =>
  S.list()
    .title('Content')
    .items([
      S.listItem()
        .title('Настройки')
        .icon(FiSettings)
        .child(S.document().schemaType('siteSettings').documentId('siteSettings')),
      ...S.documentTypeListItems().filter(
        (listItem) => !['siteSettings'].includes(listItem.getId())
      ),
    ])
```

> Error: Structure node id cannot contain character

image

binoy14:
### Description

Fix issue where adding Cyrillic characters in title for deskStructure crashes Sanity

Fixes #4185 

### What to review

1. Created a new file to generate id from either title or provided id since it was being used in multiple place
   - _Note: Not sure if the name of the file makes sense, I would love some suggestions_
2. Does the location of test make sense?
 

### Notes for release

- Allow Cyrillic characters in Desk Structure titles

### DIFF
--- a/packages/sanity/src/desk/structureBuilder/Component.ts
+++ b/packages/sanity/src/desk/structureBuilder/Component.ts
@@ -1,10 +1,10 @@
-import {camelCase} from 'lodash'
 import {SerializeOptions, StructureNode, Serializable, Child} from './StructureNodes'
 import {SerializeError, HELP_URL} from './SerializeError'
 import {MenuItem, MenuItemBuilder, maybeSerializeMenuItem} from './MenuItem'
 import {MenuItemGroup, MenuItemGroupBuilder, maybeSerializeMenuItemGroup} from './MenuItemGroup'
 import {validateId} from './util/validateId'
 import {UserComponent} from './types'
+import {getStructureNodeId} from './util/getStructureNodeId'
 
 /** @beta */
 // TODO: rename to `StructureComponent` since it clashes with React?
@@ -51,7 +51,7 @@ export class ComponentBuilder implements Serializable<Component> {
   }
 
   title(title: string): ComponentBuilder {
-    return this.clone({title, id: this.spec.id || camelCase(title)})
+    return this.clone({title, id: getStructureNodeId(title, this.spec.id)})
   }
 
   getTitle(): BuildableComponent['title'] {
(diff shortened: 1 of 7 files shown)


packages/sanity/src/desk/structureBuilder/util/getStructureBuilderId.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (id) {
+    return id
+  }
+
+  const [disallowedChar] = title.match(disallowedPattern) || []
+  if (disallowedChar) {
+    return camelCase(getSlug(title))
+  }
+
+  return camelCase(title)
```
rexxars, line 15: Two things here:

1. The matching must be tested against the camelcased version, since the title is likely to include spaces and similar. Otherwise I think this would match too often. Not the end of the world, but it would make some backwards incompatible changes to existing IDs that may be bookmarked already.
2. I think it is probably easier to use `regex.test()` here, since we're not using the matched character

```suggestion
  const camelCased = camelCase(title)
  return disallowedPattern.test(camelCased) ? camelCase(getSlug(title)) : camelCased
```

packages/sanity/src/desk/structureBuilder/util/getStructureBuilderId.ts
```diff
@@ -0,0 +1,16 @@
+import {camelCase} from 'lodash'
+import getSlug from 'speakingurl'
+import {disallowedPattern} from './validateId'
+
+export function getStructureBuilderId(title: string, id?: string): string {
```
rexxars, line 5: I suggest `getStructureNodeId()` - the "builder" part is kind of an implementation detail (would rename the file also to match, once the method is renamed).

```suggestion
export function getStructureNodeId(title: string, id?: string): string {
```

rexxars (changes requested):
This is great! Few minor suggestions on naming and order of operations

rexxars (approved):

PR #4227, opened 2023-03-01
fix(portable-text-editor): fix broken up front validation 

skogsmaskin:
### Description

During the last refactor of the main PortableTextEditor component released in the recent 3.4.0., something happened to the up front validation of the value before passing it on to the actual editor. In the refactor we moved validation into it's own component.

Part of the problem is how the internal change$ observable subscription  in SyncWrapper now becomes unsubscribed if the value doesn't validate anymore (as Validator can't safely return it's children at this point).

We therefore must pass this to `props.onChange` ourselves in the validator effect in Validator, for it to reach the outside consumer.

Because of this, displaying the validation error resolver tool in the PT-input never got shown.

To make things worse, the initialValue was not properly validated either, so it got propagated into the editor causing it to crash.

This PR will change:
* Introduce a validation context to be used inside the PTE.
* Make sure that initial value is properly validated.
* Emit the `invalidValue` change to the outside whenever the value becomes invalid.

### What to review
* That the user can recover from an invalid value given to the PT input

### Notes for release
* Fix a regression from v.3.4.0 where invalid content for the PT-input would cause the studio to crash.


skogsmaskin:
Closing this in favour to a new one that will include some more things.

PR #4229, opened 2023-03-01
feat(cli): use strict mode schemas for shopify online template

rexxars:
### Description

This addresses slightly more than the title indicates:

- Formats template code with prettier (editing files after bootstrapping template will change the formatting, which seems odd)
- Uses strict mode schemas - accordionGroup, inventory and priceRange types are moved to separate, top-level object types
- Imports React in TSX files to maximize cross-environment support for JSX
- Replaces the custom `blocksToText` utility with `toPlainText` from `@portabletext/toolkit`
- Adds titles to some fields that were missing it, for less schema warnings

Needs to be paired with https://github.com/sanity-io/sanity-plugin-shopify-assets/pull/2 in order to actually achieve full strict mode schemas. 

### What to review

That template works as expected. Can be tested by building and running the CLI:

```
cd ~/your/branch/checkout
yarn && yarn build
./packages/@sanity/cli/bin/sanity init --template shopify-online-storefront
```

### Notes for release

- Made schemas in Shopify online storefront template compatible with GraphQL deployments


thebiggianthead (approved):

rexxars:
> Looks good - just added a `name` to the array of product options to ensure this matches the name provided by the data Sanity Connect provides.

Hmm, but that actually makes it a new inline type, so it is no longer strict mode compatible.

_Ideally_, this would have always been named `productOption`, but given that we can't easily change Sanity Connect to use a new type name (since it would break existing studios), I think the only thing we can do here is to rename the `productOption` type to simply `option`. That way things are still `_type: 'option'` in the data.

I don't love the fact that it has such a generic name in the global namespace, but I can't think of any other backwards compatible approach here.

PR #4232, opened 2023-03-02
fix(core): focus issue in nav drawer

hermanwikner:
### Description

This pull request fixes an issue where elements in the `NavDrawer` were focusable even though it was closed. The problem was caused by the `useRovingFocus` hook taking control of `tabIndex` which led to our own logic to set `tabIndex` to `-1` on elements being overridden by that hook.

The problem is fixed by using `AnimatePresence` so that the `NavDrawer` is only in the DOM when it is open, and replacing `useRovingFocus` with `FocusLock` (which is an already existing dependency).

### What to review

- Make sure that the elements in `NavDrawer` is not focusable when it is closed
- Make sure that the `NavDrawer` works as before

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/NavDrawer.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/NavDrawer.tsx
@@ -1,15 +1,4 @@
-import {
-  Layer,
-  Card,
-  Flex,
-  Text,
-  Box,
-  Button,
-  Stack,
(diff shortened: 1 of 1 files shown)


RitaDias (approved):
This looks good to me! 🙂 
Thank you! 🙏

robinpyon (approved):
Great pickup, thank you @hermanwikner!

PR #4235, opened 2023-03-03
chore(build): remove cypress

bjoerge:
### Description
We are no longer running the cypress tests (after the release of v3), and they seem to be largely outdated and the setup is just adding noise without providing any value, so I suggest we remove it for now. The right thing to do long term is probably to migrate the still relevant ones to playwright?

### Notes for release
Not relevant - internal tooling


vicmeow (approved):
We might be able to use some of the helper functions in our playwright tests, but it'll be no problem finding back to this or creating new ones from scratch when we need to. They'll probably have to be changes anyway. But lgtm, thanks!

PR #4241, opened 2023-03-06
fix(core): ensure filter function on reference field is always fired when menu is opened

binoy14:
### Description

Previously, when a reference field with a filter function was opened, the function would only be triggered on the first occasion. However, with this fix, the filter function will be executed every time the select menu is opened.

### Notes for release

Ensures that the filter function of a reference field is always executed upon opening of the select menu.

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
@@ -149,7 +149,6 @@ export function ReferenceInput(props: ReferenceInputProps) {
     (inputValue$: Observable<string | null>) => {
       return inputValue$.pipe(
         filter(nonNullable),
-        distinctUntilChanged(),
         switchMap((searchString) =>
           concat(
             of({isLoading: true}),


packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
```diff
@@ -149,7 +149,8 @@ export function ReferenceInput(props: ReferenceInputProps) {
     (inputValue$: Observable<string | null>) => {
       return inputValue$.pipe(
         filter(nonNullable),
-        distinctUntilChanged(),
+        // ensures that the filter function is always fired when the button is clicked
+        distinctUntilChanged((prev, cur) => prev !== '' && cur !== ''),
```
bjoerge, line 153: Would it be just as good to remove the whole `distinctUntilChanged` here? Since it's a stream of values emitted from the text input I don't think we will ever get two following identical values (the input element will only emit onChange when it's `value` actually changes, right?), or am I missing something?
```suggestion
        // ensures that the filter function is always fired when the button is clicked
        distinctUntilChanged((prev, cur) => prev !== '' && cur !== ''),
```

binoy14: That was my first thought, but I wasn't sure if it's there for a reason, I am happy to remove it if it makes more sense

bjoerge: let's remove it - it seems to work exactly the same without it ✂️

bjoerge (approved):

PR #4244, opened 2023-03-07
fix(desk): attempt to resolve missing document type automatically

rexxars:
### Description

In some cases, a link will be generated that does not include a document type (because only the document ID is loaded until the document type has been resolved, for instance). When following the link, if the default document child resolver was used, it would throw an error. This PR improves on the situation by attempting to resolve the document type. If we're unable to find the document, we still throw - but otherwise we can recover from this state.

[sc-30793]

### What to review

- Example URL that fails in `next`: https://test-studio.sanity.build/test/content/author;grrm;fed42f93-66c5-4a47-a4dd-2a03c965a021%2CparentRefPath%3DfavoriteBooks%5B_key%3D%3D%2297724180be7c%22%5D
- Should now resolve in preview build: https://test-studio-git-fix-structure-auto-resolve-type.sanity.build/test/content/author;grrm;fed42f93-66c5-4a47-a4dd-2a03c965a021%2CparentRefPath%3DfavoriteBooks%5B_key%3D%3D%2297724180be7c%22%5D

### Notes for release

- Fixes an issue where the desk tool might crash if expanding a referenced document before its type was resolved

### DIFF
--- a/packages/sanity/src/desk/structureBuilder/Document.ts
+++ b/packages/sanity/src/desk/structureBuilder/Document.ts
@@ -1,5 +1,6 @@
 import {uniq} from 'lodash'
 import {SchemaType} from '@sanity/types'
+import {resolveTypeForDocument} from './util/resolveTypeForDocument'
 import {ChildResolver} from './ChildResolver'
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM!

PR #4250, opened 2023-03-07
fix(core): fix crash when history does not include valid user information

binoy14:
### Description

When a dataset is copied without history it stores userId as <system> which is not a valid userId. This change fixes studio to show a dummy user information when invalid ids are present.

### What to review

Making change to v2 version

### Notes for release

Fix crash in studio when viewing history for dataset that is copied without history


rexxars (approved):
LGTM! Disregard the failing e2e test, it seems to be consistent in the v2 branch, so is unrelated to this PR.

PR #4251, opened 2023-03-08
refactor(core): initial generalised command list

robinpyon:
# Description

This PR is a first pass at separating the virtualized [combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) used in global search (colloquially referred to as a _Command List_) to a 'core' component.

This initial work is in service of eventually enabling this functionality in other areas of the studio that deal with large lists requiring both virtualization and non-focusing keyboard navigation  – such as document history and upcoming changes to the create new popover. Maybe even filterable desk panes in future.

Some minor tidying has also been done to some existing studio workshop stories – they're now organised in a manner that more closely resembles the `sanity` package file structure. Core components of any reasonable complexity (or with accompanying workshop stories) now also get their own folder.

Finally, the manual bump to Sanity UI [1.3.0](https://github.com/sanity-io/ui/releases/tag/v1.3.0) is there to fix a regression brought about by a dependency update, and allow these command lists to programatically set hover styling on some Sanity UI components.

# What to review

- [Command List workshop stories](https://test-studio-git-refactor-command-list.sanity.build/test/workshop/core;components;command-list;kitchen-sink?scheme=dark) (`core/components/CommandList`) which outline basic scenarios (filtering, selecting, disabled items, infinite scrolling)
  - How can these components be made more ergonomic?
  - Some things to try when using these command lists: a combination of fast scrolling, keyboard navigation with up/down arrow keys and dragging scroll handles.
- Global search – quite a bit has been reshuffled but functionally it remains the same!
- Workshop story structure – does this make sense?

# Notes for release

(N/A)


packages/sanity/src/core/studio/components/navbar/search/SearchDialog.tsx
```diff
@@ -42,148 +39,40 @@ const SearchDialogBox = styled(Box)`
...
-        )}
-
-        <CommandListContainer>
-          {hasValidTerms ? <SearchResults onClose={handleClose} /> : <RecentSearches />}
-        </CommandListContainer>
-      </InnerCard>
-    </SearchDialogBox>
+    <SearchWrapper hasValidTerms={hasValidTerms} onClose={onClose} onOpen={onOpen} open={open}>
+      <Portal>
+        <FocusLock autoFocus={false} returnFocus>
```
hermanwikner, line 54: There is some focus issue with the search text input. When I start typing, the focus is lost when the search results popover appear and the search text input expands. Using `autoFocus` seems to solve the problem, but that might introduce other issues?

packages/sanity/src/core/studio/components/navbar/search/components/SearchPopover.tsx
```diff
@@ -71,179 +64,63 @@ export function SearchPopover({
...
-        {hasValidTerms ? (
-          <SearchResults onClose={handleClose} />
-        ) : (
-          <RecentSearches showFiltersOnClick />
-        )}
-      </CommandListContainer>
-    </SearchPopoverCard>
+    <SearchWrapper hasValidTerms={hasValidTerms} onClose={onClose} onOpen={onOpen} open={open}>
+      <Portal>
+        <FocusLock autoFocus={false} disabled={disableFocusLock} returnFocus>
```
hermanwikner, line 94: There is some focus issue with the search text input. When I start typing, the focus is lost when the search results popover appear and the search text input expands. Using `autoFocus` seems to solve the problem, but that might introduce other issues?

hermanwikner: This focus issue seems to be related to "Recent Searches". When I have recent searches, focus works as expected.

robinpyon: Great spot, in global search we have one shared input used across two command lists.

The issue here was that `autoFocus` was being set on command lists – however, it's possible to have NO visible command lists, e.g. if you have no recent search results OR search query

Changes:
- In global search: `autoFocus` is now set on the search header itself and not on the individual command lists (we check for touch support so we don't auto focus on mobile to avoid the software keyboard popping up un-asked)
- Command lists will no longer check for touch support when calling `autoFocus`. They shouldn't have been opinionated in this way to begin with

hermanwikner:
I discovered a small focus issue: when you close the search popover, the focus does not return to the search field

robinpyon:
> I discovered a small focus issue: when you close the search popover, the focus does not return to the search field

@hermanwikner This is intentional! Closing search will return focus to the previously selected element, which _may_ be the search field but in most cases will probably be a field in the document pane.

hermanwikner:
> > I discovered a small focus issue: when you close the search popover, the focus does not return to the search field
> 
> @hermanwikner This is intentional! Closing search will return focus to the previously selected element, which _may_ be the search field but in most cases will probably be a field in the document pane.

Alright, thanks! @robinpyon 

I'm not entirely sure I understand, or if I'm doing something wrong. But if I move the focus to the search field and press "Enter" to open the search, and then press "Escape" to close the search, I lose focus orientation, and when I press "Tab" the navbar logo gets focused. I would have expected that if I open search via the search field that focus is restored to the search field when search is closed.

robinpyon:
@hermanwikner You understand correctly and were right the first time, this was indeed a regression! 

`<FocusLock>` now handles auto focus for us (rather than setting that on the nested `<TextInput>`) – this should be fixed.

hermanwikner (approved):
This is great, solid work! And I love the new workshop structure 💯

packages/sanity/src/core/field/__workshop__/index.ts
```diff
@@ -2,8 +2,8 @@ import {defineScope} from '@sanity/ui-workshop'
 import {lazy} from 'react'
 
 export default defineScope({
-  name: 'sanity/field',
```
bjoerge, line 5: great call changing from `sanity` => `core` here

bjoerge (changes requested):
Reviewed code only and it looks great, solid work! The commit  1d8072245ce282a3bd90f3a77e4f6b7d74e31d84 was removed from `next` and should be removed from this branch as well before merging. Otherwise good to merge from what I can tell.

packages/sanity/src/core/field/__workshop__/index.ts
robinpyon: Thanks @bjoerge – that's done!

bjoerge (approved):

PR #4252, opened 2023-03-08
feat(desk): document inspectors (`@beta`)

mariuslundgard:
**tl;dr**
- New Document inspector API
- Review changes => document inspector
- Validation => document inspector

---

#### NEW: Document inspectors `@beta` API

This PR introduces the `@beta` ability to configure custom document inspectors in the Studio.

```ts
// sanity.config.ts

import {RocketIcon} from '@sanity/icons'
import {defineConfig, defineDocumentInspector} from 'sanity'

const customInspector = defineDocumentInspector({
  name: 'custom',
  menuItem: {
    icon: RocketIcon,
    title: 'Custom',
  },
  component: lazy(() => import('./inspectors/custom')),
  showAsAction: true,
})

export default defineConfig({
  // ...

  document: {
    inspectors: (prev) => [customInspector, ...prev],
  },
})
```

> NOTE: It may be preferable to let document inspector components be lazy, so they don't bloat the Studio bundle, but load dependencies as the component is rendered instead.

Then, to create the component for your custom document inspector:

```tsx
// inspectors/custom.tsx

import {CloseIcon} from '@sanity/icons'
import {Box, Button, Card, Flex, Text} from '@sanity/ui'
import {DocumentInspectorProps} from 'sanity'
import {DocumentInspectorHeaderCard} from 'sanity/desk'

export default function CustomInspector(props: DocumentInspectorProps) {
  const {onClose} = props

  return (
    <Flex direction="column" height="fill" overflow="hidden">
      <DocumentInspectorHeaderCard as="header" flex="none">
        <Flex as="header">
          <Box flex={1} padding={4}>
            <Text as="h1" size={1} weight="semibold">
              Custom
            </Text>
          </Box>
          <Box padding={3}>
            <Button
              aria-label="Close Custom inspector"
              fontSize={1}
              icon={CloseIcon}
              mode="bleed"
              onClick={onClose}
              padding={2}
            />
          </Box>
        </Flex>
      </DocumentInspectorHeaderCard>

      <Card flex={1} overflow="auto">
        <Box padding={4}>
          <Text size={1}>(Inspector contents)</Text>
        </Box>
      </Card>
    </Flex>
  )
}
```

In the document pane within the Desk tool, the custom document inspector will be rendered as a menu button in the header. When clicked, the inspector component will be rendered in the side panel to the right of the document view.

Document inspector demo

> NOTE: It's possible to render the document inspector inside the header menu (the vertical ellipsis) by omitting `showAsAction` or using `showAsAction: false`.

#### Review changes => document inspector

With this PR, the _Review changes_ panel has been converted to a document inspector:

Review changes inspector

#### Validation => document inspector

With this PR, the _Validation_ dropdown list has been converted to a document inspector:

Validation inspector

### What to review

- The source code and the API of the new document inspector feature.
- QA: functionality of Review changes inspector.
- QA: functionality of Validation inspector.

### Notes for release

- NEW: Document inspectors `@beta` API


dev/test-studio/inspectors/ai/SparklesIcon.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
```
mariuslundgard, line 1: This should be moved to `@santity/icons`

dev/test-studio/inspectors/ai/inspector.tsx
```diff
@@ -0,0 +1,31 @@
+import {CloseIcon} from '@sanity/icons'
+import {Box, Button, Flex, Text} from '@sanity/ui'
+import React, {useCallback} from 'react'
+import {useDocumentPane} from 'sanity/desk'
+
+export default function AIInspector() {
+  const {closeInspector} = useDocumentPane()
+  const handleClose = useCallback(() => closeInspector(), [closeInspector])
```
mariuslundgard, line 8: Document inspectors should receive `props.onClose` for simpler use.

packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/FieldGroupTabs.tsx
```diff
@@ -37,7 +37,7 @@ const GroupTabs = ({
   shouldAutoFocus = true,
   disabled,
 }: FieldGroupTabsProps) => (
-  <TabList space={2} data-testid="field-group-tabs">
+  <TabList space={1} data-testid="field-group-tabs">
```
mariuslundgard, line 40: This change should not be part of this PR.

packages/sanity/src/desk/panes/document/inspectors/validation/inspector.tsx
```diff
@@ -0,0 +1,172 @@
...
+  WarningOutlineIcon,
+} from '@sanity/icons'
+import {Box, Button, Card, CardTone, ErrorBoundary, Flex, Stack, Text} from '@sanity/ui'
+import {
+  ObjectSchemaType,
+  Path,
+  SanityDocument,
+  SchemaType,
+  ValidationMarker,
+  // isValidationErrorMarker,
```
mariuslundgard, line 15: Remove these

packages/sanity/src/core/config/document/inspector.ts
```diff
@@ -0,0 +1,88 @@
...
+/** @beta */
+export interface DocumentInspectorCallbackContext {
+  validation: ValidationMarker[]
+}
+
+/** @beta */
+export interface DocumentInspectorMenuItem {
+  hidden?: boolean | ((props: DocumentInspectorCallbackContext) => boolean)
+  hotkeys?: string[]
+  icon?: ComponentType
```
hermanwikner, line 23: I think it would make sense if `icon` was not optional. Without an icon, the inspector button won't be visible, and without a TS error, it can be difficult to figure out why the button isn't visible. Alternatively, we could have a fallback icon?

Screenshot 2023-05-26 at 14 12 50

packages/sanity/src/desk/panes/document/inspectors/changes/index.ts
```diff
@@ -0,0 +1,18 @@
+import {RestoreIcon} from '@sanity/icons'
+import {lazy} from 'react'
+import {DocumentInspector} from 'sanity'
+
+export const changesInspector: DocumentInspector = {
+  name: 'changes',
```
hermanwikner, line 6: Could it be good to have a more unique name for this inspector, so that we minimize the risk of someone creating an inspector with the name that interferes with this inspector?  Perhaps a name like `desk/changes` would be a good name, so you understand that it is an inspector that `sanity/desk` configures in e.g. the `document.inspectors` resolver?

packages/sanity/src/desk/panes/document/inspectors/validation/index.ts
```diff
@@ -0,0 +1,28 @@
+import {CheckmarkCircleIcon, ErrorOutlineIcon, WarningOutlineIcon} from '@sanity/icons'
+import {lazy} from 'react'
+import {DocumentInspector, FormNodeValidation, isValidationError, isValidationWarning} from 'sanity'
+
+export const validationInspector: DocumentInspector = {
+  name: 'validation',
```
hermanwikner, line 6: Same as previous comment – rename to e.g `desk/validation`?

packages/sanity/src/desk/panes/document/documentInspector/DocumentInspectorHeaderCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+    left: 0,
+    bottom: -1,
+    right: 0,
+    borderBottom: '1px solid var(--card-border-color)',
+    opacity: 0.5,
+  },
+})
+
+/** @internal */
+export function DocumentInspectorHeaderCard(
```
hermanwikner, line 23: I think it would be nice if the close button was part of the `DocumentInspectorHeaderCard` as it would help maintain consistent design across the inspector headers. I suggest an update where`onClose` and`title` are added as props, and that `children` is an optional prop used in situations such as for e.g. the review changes inspector.

Example:

```ts
 <DocumentInspectorHeaderCard onClose={onClose} title="My inspector" />
```

```ts
<DocumentInspectorHeaderCard onClose={onClose} title="My inspector">
  <Content />
</DocumentInspectorHeaderCard>
```

packages/sanity/src/desk/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -98,14 +101,9 @@ export const DocumentPanelHeader = memo(
...
 
-            {validation.length > 0 && (
-              <ValidationMenu
-                boundaryElement={rootElement}
-                isOpen={isValidationOpen}
-                key="validation-menu"
-                setOpen={setValidationOpen}
-              />
-            )}
+            {actionItems.map((item, itemIndex) => (
```
hermanwikner, line 104: Possibly an edge case: but I configured 100 inspectors, and the UI can't handle that many inspectors. I wonder if we could solve it by rendering the inspector buttons in a menu when they exceed a certain number (not the validation inspector). Something similar to the image below (not sure which icon might be appropriate, I used the `RobotIcon` as a placeholder icon).

Screenshot 2023-05-26 at 17 08 28

hermanwikner: I realized that this can be handled by setting `showAsAction` to `false`!

test/e2e/tests/desk/inspectors.spec.ts
```diff
@@ -0,0 +1,60 @@
+import {test, expect} from '@playwright/test'
+
+test.describe('sanity/desk: document inspectors', () => {
+  test('open and close custom inspector', async ({page}) => {
+    await page.goto('/test/content/input-debug;inspectorsTest;inspectors-test')
```
hermanwikner, line 5: I suggest we create a unique document when running these tests, instead of relying on already existing documents. We test review changes (the document must be a draft) and validation (the document needs validation errors) - and if someone edits or publishes the documents currently used, then a test might fail. 

```ts
import {uuid} from '@sanity/uuid'
import {testSanityClient} from '../../helpers'

test('open "Review changes" inspector', async ({page}) => {
  const id = uuid()
  await testSanityClient.create({
    _type: 'stringsTest',
    _id: id,
  })

  await page.goto(`/test/content/input-standard;stringsTest;${id}`)

  // ...

  await testSanityClient.delete(id)
})

```



We have a dedicated folder (`/schema/ci`) in test-studio, I think it would be good to move the tests to that folder!

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -152,10 +160,29 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     }
   }, [documentId, presenceStore])
 
+  const inspectors: DocumentInspector[] = useMemo(
+    () => inspectorsResolver({documentId, documentType}, [validationInspector, changesInspector]),
```
hermanwikner, line 164: Is there any reason or benefit to configuring these as default inspectors here instead of adding them to `document.inspectors` in the `deskTool` plugin definition?


```ts
import {changesInspector, validationInspector} from '.path/to/inspectors'

export const deskTool = definePlugin(() => ({
  name: '@sanity/desk-tool',
  document: {
    inspectors: [changesInspector, validationInspector],
  },
  // ...
}))

```

packages/sanity/src/desk/panes/document/documentInspector/DocumentInspectorPanel.tsx
```diff
@@ -0,0 +1,40 @@
...
+        minWidth={DOCUMENT_INSPECTOR_MIN_WIDTH}
+      >
+        {createElement(inspector.component, {onClose: handleClose})}
+      </Resizable>
+    )
+  }
+
+  return (
+    <Box as="aside" data-ui="DocumentInspectorPanel" flex={flex}>
+      {createElement(inspector.component, {onClose: handleClose})}
```
hermanwikner, line 37: Should we consider in addition to `onClose` also including `documentType` and `documentId` in the props?

packages/sanity/src/desk/panes/document/documentInspector/Resizer.tsx
```diff
@@ -0,0 +1,89 @@
...
+import styled from 'styled-components'
+
+const Root = styled.div`
+  position: absolute;
+  top: 0;
+  left: -4px;
+  bottom: 0;
+  width: 9px;
+  z-index: 201;
+  cursor: col-resize;
```
hermanwikner, line 11: The resize UI looks different between panes and inspectors and we use different cursors (`col-resize` here and `ew-resize` in panes) – should we unify this?

Screenshot 2023-05-29 at 14 28 16

rexxars (changes requested):
Initial functionality-only review:

- Clicking the change indicator, pencil icon in footer, or validation button on mobile devices crashes the studio with a maximum update depth exceeded error.
- Split pane behavior is somewhat broken with the inspectors:
  - When opening the validation on the left side, the right side also opens the validation panel.
  - When opening the review changes panel on the left side, the right side also opens the panel, but the "since" parameter is not set, and so yields no changes. 
- Validation breadcrumbs differ from the review changes breadcrumbs:
  - Validation: `Array of slugs / Slug Embed / Title`
  - Review changes: `Array of slugs / #2 / Title`
- Thought: With validation panel mimicking/looking like the review changes panel, I for some reason expected connectors to appear. Maybe this is something to consider for the future?

packages/sanity/src/desk/panes/document/inspectors/changes/index.ts
mariuslundgard: Yes, true. Changing to `sanity/desk/history`

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
mariuslundgard: Ohh, good catch!

mariuslundgard: It works 👍

mariuslundgard:
@rexxars 
> Thought: With validation panel mimicking/looking like the review changes panel, I for some reason expected connectors to appear. Maybe this is something to consider for the future?

Yes, would love that! Think we should make the Connectors API usable in other contexts, as a general UI pattern for connecting 2 things within a document pane.

Will address your other comments now.

mariuslundgard:
@rexxars 

> Clicking the change indicator, pencil icon in footer, or validation button on mobile devices crashes the studio with a maximum update depth exceeded error.

Indeed – this is strange. It seems to be resolved when removing the `lazy()` wrapper around the inspector components. Do you have any idea what could cause that? 🤔 It was working earlier.

Anyways, I’ll remove the `lazy()` wrappers as they're problematic in that they cause the Desk tool to reload and make users loose context.

mariuslundgard:
@rexxars

> 

This is caused by the `exclusiveParams` in `packages/sanity/src/desk/router.ts` not containing the `inspect` parameters:
```
const exclusiveParams = ['view', 'since', 'rev', 'inspect']
```

If wonder if we should flip this logic so params are exclusive by default, and instead do:
```
const sharedParams = ['template']
```

The other option is to make it possible for plugin authors to declare how their params should behave 🤔 

I will however add `'inspect'` to `exclusiveParams` for now.

packages/sanity/src/desk/panes/document/documentInspector/DocumentInspectorHeaderCard.tsx
mariuslundgard: Addressed! And renamed the component to `DocumentInspectorHeader`.

packages/sanity/src/desk/panes/document/documentInspector/Resizer.tsx
mariuslundgard: You're totally right – will align on with the pane resizer 👍

rexxars (changes requested):
Great work - wonderful addition!

First off: Appreciate that this is made `@beta`/`@internal` for now until we have tested the API more and decided it's the way we want it going forward.

The review changes panel used to be disabled on small screen sizes, but it is now enabled. Is this a conscious decision? I'm wondering whether or not we've tested the usability of it - for instance if users have enough context to really undo changes with only the diff showing. 

If we do decide to go ahead with it, there are a couple of things I'd want to consider fixing first:

The confirm dialog appears outside the screen:
image

The "revert changes" buttons are somewhat small for touching, and one has to know what they mean since there is no hover state to help you, as on desktop:
image

Apart from that, this looks and works great!

mariuslundgard:
@rexxars 

> The review changes panel used to be disabled on small screen sizes, but it is now enabled. Is this a conscious decision? I'm wondering whether or not we've tested the usability of it - for instance if users have enough context to really undo changes with only the diff showing.

Very good point. I think we should continue to disable Review changes on narrow devices until this is more considered. So I added a commit for that just now.

mariuslundgard:
@rexxars @hermanwikner 

> The E2E tests can probably be skipped for now - they are timing out due to slow CI - we need to look into "priming" the tests before running to prevent them from timing out.

I will go ahead and skip the e2e tests.

hermanwikner (approved):
Looks good, great work @mariuslundgard! 👏

PR #4256, opened 2023-03-08
fix(core): fix deleting invalid reference items in array

binoy14:
### Description
Fixes invalid reference type in array fields

### What to review

Not sure if this is the correct way to fix it but seems like it since 
https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/members/object/fields/ArrayOfObjectsField.tsx#L93 

the `handleChange` function is adding the name prefix so it should avoid double prefixing

### Notes for release

Fixes deleting invalid reference items in array fields


bjoerge (approved):
Great catch, thank you!

PR #4258, opened 2023-03-09
feat(cli): use local callback server for auth instead of listener

binoy14:
### Description

Backports auth changes changes from #4222 to v2

### What to review

Making sure everything is backported and I didn't miss anything

### DIFF
--- a/packages/@sanity/cli/src/actions/login/login.js
+++ b/packages/@sanity/cli/src/actions/login/login.js
@@ -1,13 +1,28 @@
 import os from 'os'
-import urlParser from 'url'
-import crypto from 'crypto'
+import http from 'http'
 import open from 'opn'
 import chalk from 'chalk'
-import EventSource from 'eventsource'
-import {parseJson} from '@sanity/util/lib/safeJson'
 import getUserConfig from '../../util/getUserConfig'
 import canLaunchBrowser from '../../util/canLaunchBrowser'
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/login/login.js
```diff
@@ -23,47 +38,57 @@ export default async function login(args, context) {
...
 
-  output.print(`\n${actionText} ${loginUrl}\n`)
+  output.print(`\n${actionText} ${loginUrl.href}\n`)
   const spin = output
     .spinner('Waiting for browser login to complete... Press Ctrl + C to cancel')
     .start()
 
   if (shouldLaunchBrowser) {
-    open(loginUrl, {wait: false})
+    open(loginUrl.href)
```
rexxars, line 75: This still needs the `wait: false` flag, since it is using an old version of the `opn` dependency, which waits by default. Currently the login command hangs after logging in, because it is waiting for the browser to close.

rexxars (changes requested):
One issue (commented inline), apart from that this looks good

packages/@sanity/cli/src/actions/login/login.js
binoy14: ah that makes sense, thanks for catching that!

rexxars (approved):

PR #4259, opened 2023-03-09
docs: add docblock for useClient hook

rexxars:
### Description

I figured since we already talked about this and I had an example ready, I might as well _actually_ add the docblock ;) 

### Notes for release

- Marked the `useClient()` hook as public (stable)

### DIFF
--- a/packages/sanity/src/core/hooks/useClient.ts
+++ b/packages/sanity/src/core/hooks/useClient.ts
@@ -2,7 +2,24 @@ import type {SanityClient} from '@sanity/client'
 import type {SourceClientOptions} from '../config'
 import {useSource} from '../studio'
 
-/** @beta */
+/**
+ * React hook that returns a configured Sanity client instance based on the given configuration.
+ * Automatically uses the correct project and dataset based on the current active workspace.
+ *
+ * @public
+ * @param clientOptions - Options for the client. Specifying
+ *   {@link https://www.sanity.io/docs/api-versioning | apiVersion} is required in order to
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/hooks/useClient.ts
```diff
@@ -2,7 +2,24 @@ import type {SanityClient} from '@sanity/client'
...
+ * React hook that returns a configured Sanity client instance based on the given configuration.
+ * Automatically uses the correct project and dataset based on the current active workspace.
+ *
+ * @public
+ * @param clientOptions - Options for the client. Specifying
+ *   {@link https://www.sanity.io/docs/api-versioning | apiVersion} is required in order to
+ *   prevent breaking changes if studio changes the API version used in other places.
+ * @returns A configured Sanity client instance
+ * @remarks The client instance is automatically memoized based on API version
+ * @example Instantiating a client
```
vicmeow, line 15: Do you mean "initiating" here?

vicmeow (changes requested):
LGTM! Just a comment on the use "instantiating"

packages/sanity/src/core/hooks/useClient.ts
rexxars: Instantiating as in "Creating an instance of". Maybe a bit technical, but "initiating" sounds wrong too. How about "Creating a client" or "Getting a client"?

vicmeow: Aha. Maybe "setting up" or "actualizing"? If not we could just stick with what you wrote first!

PR #4265, opened 2023-03-10
feat: add 'workspace' to conditional readOnly + hidden context

SimeonGriggs:
### Description

Adds the current `workspace` to the context parameter in `readOnly` and `hidden` functions.

### What to review

That there aren’t missing places for this sort of conditional logic. This probably shouldn't be merged until it's in other places that share the same context object like desk structure, rule.custom validations, etc.

Also the current workspace object is so big (it includes schema, theme, and more) that it might not be a good idea to include all of it – or some trimmed-down version.

### Notes for release

Adds the current `workspace` to the context parameter in `readOnly` and `hidden` functions.

### DIFF
--- a/packages/@sanity/types/src/schema/types.ts
+++ b/packages/@sanity/types/src/schema/types.ts
@@ -106,6 +106,7 @@ export interface ConditionalPropertyCallbackContext {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   value: any
   currentUser: Omit<CurrentUser, 'role'> | null
+  workspace: any
 }
 
 /** @public */
--- a/dev/test-studio/schema/author.ts
+++ b/dev/test-studio/schema/author.ts
@@ -1,5 +1,5 @@
 import {UserIcon as icon} from '@sanity/icons'
-import {Rule} from 'sanity'
+import {defineField, Rule} from 'sanity'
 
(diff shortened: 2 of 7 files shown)


packages/@sanity/types/src/schema/types.ts
```diff
@@ -106,6 +106,7 @@ export interface ConditionalPropertyCallbackContext {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   value: any
   currentUser: Omit<CurrentUser, 'role'> | null
+  workspace: any
```
bjoerge, line 109: > Also the current workspace object is so big (it includes schema, theme, and more) that it might not be a good idea to include all of it – or some trimmed-down version.

Agree with this. I think this trimmed-down version should be defined here instead of `any` (since this is the user-facing API).

What information about the workspace would be the most useful to include here?

bjoerge (changes requested):
Thanks @SimeonGriggs! A few quick comments:

1. Would you mind doing a rebase against next?
2. It would be great to have the interface of the "trimmed down" version clearly defined (see my comment below)
3. There's quite a few TypeScript errors that would be great to have sorted out

jordanl17:
This is being tracked in Studio team's backlog and is on the cards on day. Closing this PR and referencing in the appropriate todo

PR #4266, opened 2023-03-10
fix(core): clear unused references in an array

binoy14:
### Description

Clears unused references in an array on blur and pressing escape key, this works in v2 so copied the same functionality

### What to review

Copies the functionality from v2 not sure if this is the way to go about fixing the issue https://github.com/sanity-io/sanity/blob/f0eff1356962f376f1dd059e268d7d2b0aff7157/packages/%40sanity/form-builder/src/inputs/ReferenceInput/ArrayItemReferenceInput.tsx#L403-L414 
### Notes for release

Fixes an issue where unused references get cleared on clicking outside the array input or pressing escape key


bjoerge (approved):
Great work! LGTM

PR #4267, opened 2023-03-10
feat(core): provide `getClient()` method to reference filter context

rexxars:
### Description

This PR adds the `getClient()` method to the context passed to reference filter resolvers.
Brings back functionality that was possible to do in v2, and does so in a way that is compatible with other v3 contexts.

[sc-25303]
[sc-25082]

### What to review

- Changes makes sense
- Client can be retrieved (try asyncFilter example on reference test type - it might not return any results, but should not crash)

### Notes for release

- Added `getClient()` method to reference filter context


bjoerge (approved):
Nice!

PR #4268, opened 2023-03-13
fix(core): remove return null when hidden field

ninaandal:


robinpyon:
Thanks @ninaandal and @bjoerge!

I did have a play and was able to find two issues:  

**1. If you've selected any group (that isn't "All fields") and open the changes panel:**
  - Both "All fields" and the previously selected group are shown as selected
  - Non "all fields" groups on the root level aren't disabled

**Screenshot:**
image

I was able to work around this locally with the following, though not yet sure if there are other unintended side effects:

```ts
  const groups = [ALL_FIELDS_GROUP, ...schemaTypeGroupConfig].flatMap((group): FormFieldGroup[] => {
    const groupHidden = resolveConditionalProperty(group.hidden, conditionalPropertyContext)
    const isSelected = group.name === (props.fieldGroupState?.value || defaultGroupName)

    // Set the "all-fields" group as selected when review changes is open to enable review of all
    // fields and changes together. When review changes is closed - switch back to the selected tab.
    const selected = props.changesOpen ? group.name === ALL_FIELDS_GROUP.name : isSelected
    // Also disable non-selected groups when review changes is open
    const disabled = props.changesOpen ? !selected : false

    return groupHidden
      ? []
      : [
          {
            disabled,
            icon: group?.icon,
            name: group.name,
            selected,
            title: group.title,
          },
        ]
  })
```

**2. Groups that consist of only hidden fields are now marked as disabled** 
In these instances, I believe users should still be able to click on group tabs – even if no nested fields are revealed. It may not be completely clear to users that group tabs are disabled due to all their children being hidden. WDYT?

bjoerge:
@ninaandal @robinpyon sorry this got left hanging!

@robinpyon I've incorporated your suggestion and it works smoothly! Thanks for catching! From what I can tell this should be good to merge now?

bjoerge:
Our test suite caught an error here - just pushed a fix.

@ninaandal, @robinpyon could you confirm that the original issue is now fixed?

ninaandal:
> Our test suite caught an error here - just pushed a fix.
> 
> @ninaandal, @robinpyon could you confirm that the original issue is now fixed?

This lgtm 🎉 Thank you for fixing this

robinpyon (approved):
Thank you both!

PR #4270, opened 2023-03-13
Revert "fix(core): clear unused references in an array (#4266)"

binoy14:
### Description

This reverts commit 3b0e4cf3f8070e2787089d33d38372d5d54e6846.

Using the fix below actually breaks the menu item buttons since they are in a portal which the component does not have access to. Proper fix to be added in future PR


bjoerge (approved):
Thanks you for testing and catching this! Given that 3b0e4cf3f8070e2787089d33d38372d5d54e6846 did not get released yet, I've removed it from `next`

PR #4274, opened 2023-03-14
fix(core): fix typings to reflect that the second argument to patch.execute is optional

bjoerge:
### Description
As per the types it currently looks like providing a second argument to `patch.execute` is required when it isn't (passing `undefined` works because `{...undefined}` is valid)

This PR fixes modifies the typings to reflect this.

### Notes for release

- Changed typings of `patch.execute()` to reflect that the second argument is optional


vicmeow (approved):
LGTM!

PR #4275, opened 2023-03-14
fix(core): remove button to convert to reference for arrays

binoy14:
### Description

When converting from array of references to a singular reference convert to reference would show which would convert it incorrectly making the document invalid

### What to review

~I still kept `typeof value === 'object'` check to keep typescript happy~

### Notes for release

Remove convert to reference button when converting array of references to singular reference

### DIFF
--- a/packages/sanity/src/core/form/inputs/InvalidValueInput/InvalidValueInput.tsx
+++ b/packages/sanity/src/core/form/inputs/InvalidValueInput/InvalidValueInput.tsx
@@ -6,6 +6,7 @@ import {Details} from '../../components/Details'
 import {isDev} from '../../../environment'
 import {converters as CONVERTERS, ValueConverter} from './converters'
 import {UntypedValueInput} from './UntypedValueInput'
+import {isPlainObject} from 'lodash'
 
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/InvalidValueInput/InvalidValueInput.tsx
```diff
@@ -59,7 +60,12 @@ export const InvalidValueInput = forwardRef(
       [value, actualType, validTypes]
     )
 
-    if (typeof value === 'object' && value !== null && !('_type' in value)) {
+    if (
+      isPlainObject(value) &&
+      typeof value === 'object' &&
+      value !== null &&
+      !('_type' in value)
```
bjoerge, line 67: Not that it probably matters all that much _in practice_, but we're kinda doing the same job twice here to make TypeScript happy. I think it's perfectly fine to use `!('_type' in (value as object))` here

bjoerge (changes requested):

bjoerge (approved):

PR #4279, opened 2023-03-14
chore: drop unused mock files

rexxars:
### Description

Just doing some cleanup - seems to me like there were a bunch of mock files that probably dates back to v2, that no longer seem to be in use.

### Notes for release

Nothing. Transparent to the end user.


bjoerge (approved):
Nice, thank you! ✨

PR #4281, opened 2023-03-15
fix(core): handle 403 responses when calling `userStore.getUser()`

robinpyon:
## Description

This PR ensures that `userStore.getUser()` catches 403 responses - needed when trying to fetch users when your current role doesn't have access to _project members_. Previously, this was crashing the studio when opening document history changes.

This only really affects `<UserAvatar>`, and the reason why presence doesn't crash in these instances is that it uses `DataLoader.loadMany()` under the hood, which handles responses containing errors a little differently ([they always resolve](https://github.com/graphql/dataloader#loadmanykeys)).

This is a quick fix, and in future we could potentially look to obviate these requests entirely if your role doesn't have sufficient access.

## What to review

Opening document history shouldn't crash the studio when accessing the studio with a role that does not have read access to either:
- Project members
- Project details

## Notes for release

Fixes an issue where the studio may crash when accessing history changes with reduced permissions.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/user/__tests__/userStore.test.ts
+++ b/packages/sanity/src/core/store/_legacy/user/__tests__/userStore.test.ts
@@ -1,0 +1,31 @@
+import {SanityClient} from '@sanity/client'
+import {createUserStore} from '../userStore'
+
+export class HttpError extends Error {
+  statusCode?: number
+}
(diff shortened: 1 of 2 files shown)


rexxars (approved):
Thanks! Appreciate the tests 🙏

PR #4283, opened 2023-03-15
fix(SSR): handle `scheme` & `onSchemeChange` props edge cases

stipsan:
### Description

Fixes SSR hydration mismatch bugs caused by `ColorSchemeProvider` reading from localStorage (`const storedScheme = getItem?.(LOCAL_STORAGE_KEY) as ThemeColorSchemeKey | undefined`) during render. Rewritten to use `useSyncExternalStore` and other patterns we use for solving this in `@sanity/ui` with hooks like `usePrefersDark` and `useMediaIndex`.
Also rewrites `colorScheme.tsx` internals to break `ColorSchemeContext` into smaller, separate contexts to speed up re-renders. In [Themer](https://themer.sanity.build/) the slow path is sometimes extremely noticeable, but now it's far less of a problem.

And finally, the `<StudioProvider scheme={"light" | "dark"}>` works as expected. And support for `<StudioProvider scheme="system">` is added. Like before, setting the `scheme` prop acts as an initial value.

But if you add `onSchemeChange` it now acts as "controlled" and the appearance menu is hidden, in the same way it's hidden if you use a legacy color theme.

### What to review

Check the new `dev/test-next-studio` for tests that demonstrate the hydration errors are fixed (there's no console warnings or errors 🎉 ).

### Notes for release

fix(SSR): hydration mismatch bugs related to the `scheme` and `onSchemeChange` props on `<Studio />` are fixed!

### DIFF
--- a/dev/test-next-studio/.depcheckignore.json
+++ b/dev/test-next-studio/.depcheckignore.json
@@ -1,0 +1,3 @@
+{
+  "ignore": ["styled-components", "react", "react-dom", "sanity"]
+}
--- a/dev/test-next-studio/.gitignore
+++ b/dev/test-next-studio/.gitignore
@@ -1,0 +1,2 @@
+.next
+.vscode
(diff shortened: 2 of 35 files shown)


bjoerge (approved):
LGTM, awesome work, thank you!

PR #4285, opened 2023-03-15
fix(core): remove array from document when it's only item is unset

binoy14:
### Description

Removes array from document when it's only item is unset

### Notes for release

Removes array from document when it's only item is unset


packages/sanity/src/core/form/members/object/fields/ArrayOfObjectsField.tsx
```diff
@@ -90,9 +91,25 @@ export function ArrayOfObjectsField(props: {
 
   const handleChange = useCallback(
     (event: PatchEvent | PatchArg) => {
+      const patches = PatchEvent.from(event).patches
+      // if the patch is an unset patch that targets an item in the array (as opposed to unsetting a field somewhere deeper)
+      const hasUnset = patches.some((patch) => patch.type === 'unset' && patch.path.length === 1)
```
bjoerge, line 96: Re-reading this I realize that it would be more accurate to name this `isRemovingItem` or similar, given that it reflects whether an item is removed from the array, not whether there is an unset patch in the array.

packages/sanity/src/core/form/members/object/fields/ArrayOfPrimitivesField.tsx
```diff
@@ -179,9 +180,25 @@ export function ArrayOfPrimitivesField(props: {
 
   const handleChange = useCallback(
     (event: PatchEvent | PatchArg) => {
+      const patches = PatchEvent.from(event).patches
+      // if the patch is an unset patch that targets an item in the array (as opposed to unsetting a field somewhere deeper)
+      const hasUnset = patches.some((patch) => patch.type === 'unset' && patch.path.length === 1)
```
bjoerge, line 185: same as previous comment :)

bjoerge (approved):
LGTM! (you can consider renaming the `hasUnset` variable if you think it makes sense)

packages/sanity/src/core/form/members/object/fields/ArrayOfObjectsField.tsx
binoy14: named it `isRemovingLastItem` let me know if that makes sense

PR #4287, opened 2023-03-15
fix(vision): fix results getting cut off in vision

binoy14:
### Description

- Fix result cut off in vision result
- Fix scroll bar being hidden in vision result

[sc-26725]
[sc-26731]

### What to review

Before:
Screenshot 2023-03-15 at 3 03 59 PM

After:
Screenshot 2023-03-15 at 3 05 42 PM

### Notes for release

- Fix result cut off in vision result
- Fix scroll bar being hidden in vision result

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.tsx
+++ b/packages/@sanity/vision/src/components/VisionGui.tsx
@@ -828,26 +828,24 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
                   tone={error ? 'critical' : 'default'}
                   $isInvalid={Boolean(error)}
                 >
-                  <Result padding={3} paddingTop={5} overflow="auto">
+                  <Result overflow="auto">
                     <InputBackgroundContainer>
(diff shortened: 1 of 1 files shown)


rexxars:
Any chance you are able to fix (eg remove) the side scroll too?

rexxars (approved):
Would be awesome if we were able to remove the side-scroll too (when there is no content that forces a scroll), but unless it is a quick fix: LGTM

binoy14:
Seems like the library itself is adding some scroll bars but I can't figure out how or where it's happening https://lapple.github.io/react-json-inspector/

I can change [this](https://github.com/sanity-io/sanity/pull/4287/files#diff-c8afd2229894d93767b11a43554cfaaaaf701574510b2582ddb8a8016960008bR831) to only overflow on Y and hidden on X which fixes it. I tested by adding a deeply nested data till it reached the width as well as adding a really large single line text and it just clipped to new line but not sure if there any other concerns on changing it.

PR #4288, opened 2023-03-16
feat(core): improve global create menu

hermanwikner:
### Description

This pull request introduces improvements to the experience of creating a new document in studio with the 'create new' option in the nav bar.

### What to review
-
### Notes for release

-


packages/sanity/src/core/studio/components/navbar/new-document/NewDocumentButton.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const handleOpenButtonKeyDown = useCallback(
+    (e: React.KeyboardEvent<HTMLButtonElement>) => {
+      if (e.key === 'ArrowDown' && !open) {
+        setOpen(true)
+      }
+    },
+    [open]
+  )
+
+  // Close popover on escape or tab
```
robinpyon, line 89: Closing on tab is a nice touch we should throughout the studio!

packages/sanity/src/core/studio/components/navbar/new-document/filter.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  // Split the query into words (separated by spaces, dashes or underscores)
+  const words = lowercaseQuery.split(/[\s\-_]+/)
+
+  return words.every((word) => lowercaseTitle.includes(word))
+}
+
+// Sort the results based on relevance. The result should be sorted based on the following criteria:
+// 1. If the title starts with the query, it should be first
+// 2. If the title includes the query, it should be second
```
robinpyon, line 15: Also great attention to detail, global search should do this

robinpyon:
This is great work @hermanwikner! Thank you for including tests.

> Should we hide the filtering when there aren't that many document types?

Personally I don't think so. I think there's value in signalling that capability to users (even if they're not in a position to take full advantage of it).

A few issues with the `<Tooltip>` component on disabled options worth calling out – 1. adjacent items don't always trigger and 2. they don't dismiss when only scrolling. I don't think these are blocking though, and can probably be addressed in separately in Sanity UI.

Finally – I don't think the filter should persist after the popover / dialog is closed. WDYT?

robinpyon (approved):

bjoerge (commented):
Code looks great, really solid work!

Got a couple of comments/questions:
- In the old menu/dialog we displayed the icon next to the document type/template. Was it a deliberate choice to remove it?
- What are a typical _high_ number of document types? Tens of thousands of document types doesn't sound very realistic and I'm wondering if virtualization is worth the extra complexity in this case. E.g. how many document types can you have before you get a noticeable slowdown?

bjoerge (approved):

packages/sanity/src/core/studio/components/navbar/new-document/NewDocumentButton.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const [popoverElement, setPopoverElement] = useState<HTMLDivElement | null>(null)
+  const [buttonElement, setButtonElement] = useState<HTMLButtonElement | null>(null)
+  const [searchInputElement, setSearchInputElement] = useState<HTMLInputElement | null>(null)
+
+  const {scheme} = useColorScheme()
+  const currentUser = useCurrentUser()
+
+  const hasNewDocumentOptions = options.length > 0
+  const disabled = !canCreateDocument || !hasNewDocumentOptions
+  const placeholder = `Search ${options.length} document types`
```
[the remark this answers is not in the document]
hermanwikner: That is a very good point. I'll make sure we have a more accurate copy here!

mikolajdobrucki (changes requested):
Two minor comments regarding the UI:

1. The the search input is focused automatically when you open the popover. As the focus ring creates enough separation, we probably don't need a border between the label "create new document" and the input field. We can also decrease the spacing between the label and the input to make the header smaller.

2. The blue focus state for the list items doesn't work well as it puts too much emphasis on the list item the moment the popover opens.

Currently, many different parts of the Studio follow different patterns when it comes to focus, hover and selected states. I hope we can take a more holistic view at that soon and find a more long-term solution.

In the meantime, we can use the same pattern as the global search by using light grey as a hover and selected color. This will make the component less noisy and more consistent with a similar pattern used in the closest proximity (search).

Mockup presenting the suggested changes 👇 

image

mikolajdobrucki (approved):
Looks great, people are going to love it!

PR #4290, opened 2023-03-16
feat(cli): add `--bare` and `--env` flags to `sanity init`

benja:
### Description

It's currently not possible to create a Sanity project without generating boilerplate code. So far, the _"hack"_ to quickly create a project has been to init a new project, go through the flow, and then abort the process when it asks for an output path.

To make this flow more convenient for developers setting up Sanity with existing projects (e.g. starter templates), I've added two new flags to the `init` command. 

These flags are:

- `--bare`: This flag only outputs the project ID and dataset to stdout, without any template code.
- `--env <path?>`: This flag detects the framework and creates a `.env` file with the correct prefix for environment variables. Path is optional; the default value is `.env`, but you can change it to anything beginning with `.env`, for example `.env.local.example`.

Screenshot 2023-03-16 at 15 50 41

### What to review
Please feel free to comment on the code and provide suggestions. I've tried to recognize code patterns, but let me know if I've deviated too much. Also, feel free to improve any messages printed to the console!

I didn't find any existing tests for the `sanity init` command (other than a smoke test), so I didn't write any tests for the new flags. However, I did run existing tests to ensure everything is still working as intended. Let me know where I should possibly write tests.

### Notes for release

Added `--bare` flag to output project details to stdout.
Added `--env <path?>` flag to create or append project environment variables to `.env` (or another path, e.g. `.env.local`).


runeb:
> more convenient for developers setting up Sanity with existing projects

I assume you mean existing Studio code, but need to create a Sanity project? On V2 you used to be able to run sanity init in an existing Studio folder and it would re-configure the Studio with a new Sanity project

Just my 2 cents

benja:
> > more convenient for developers setting up Sanity with existing projects
> 
> I assume you mean existing Studio code, but need to create a Sanity project? On V2 you used to be able to run sanity init in an existing Studio folder and it would re-configure the Studio with a new Sanity project
> 
> Just my 2 cents

This is going to allow us to create v3 templates where the Studio code and the app (any framework) is set up for you, but you don't yet have a Sanity project to go with it.

It would also be more convenient for those who have existing projects where they quickly want to set up a new Sanity project (or use an existing one) and switch out the `projectId` and `dataset` in their code.

packages/@sanity/cli/package.json
```diff
@@ -62,6 +62,8 @@
   },
   "dependencies": {
     "@babel/traverse": "^7.19.0",
+    "@vercel/frameworks": "^1.3.2",
```
stipsan, line 65: As `@vercel/fs-dectors` already depend on `@vercel/frameworks` you should be able to drop it :smil'
```suggestion
    "@vercel/frameworks": "^1.3.2",
```

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+      const framework: Framework | null = await detectFrameworkRecord({
+        fs: new LocalFileSystemDetector(workDir),
+        frameworkList: frameworks as readonly Framework[],
+      })
+
+      if (framework && framework.slug !== 'sanity') {
+        print(
+          `\nDetected framework ${chalk.blue(framework?.name)}, using prefix '${
+            framework.envPrefix
+          }'`
```
stipsan, line 799: What happens if it detects a framework that doesn't have a prefix, like `remix`?
Does it print:
```
Detected framework Remix, using prefix ''
```
?

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+        )
+      }
+
+      await writeToEnv(filename, variables, {
+        framework,
+      })
+    } catch (err) {
+      print(err)
+      throw new Error('An error occurred while creating .env')
```
stipsan, line 808: It's nice to use [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) so it's a little easier to traverse the stack trace in your debugger:
```suggestion
      throw new Error('An error occurred while creating .env', {cause: err})
```

stipsan (approved):
I literally can't wait for this to launch ✨ Stellar work diving straight into the deep end @benja, respect 🙇‍♂️

stipsan:
> I assume you mean existing Studio code, but need to create a Sanity project? On V2 you used to be able to run sanity init in an existing Studio folder and it would re-configure the Studio with a new Sanity project
> 
> Just my 2 cents

With these commands you still get to decide wether you want to create a new Sanity project, or use an existing one. A new dataset, or one you already have.
And instead of writing to a `sanity.json`, it writes to `.env`. I agree it would be nice to bring back the convenient reconfigure feature from v2. But this is a great first step, it makes sense to keep iterating on this and use more of the `dotenv` toolchain as a stretch goal, reconfiguring it if you may, as future improvements :relaxed:

runeb:
> With these commands you still get to decide wether you want to create a new Sanity project, or use an existing one. A new dataset, or one you already have.
> And instead of writing to a `sanity.json`, it writes to `.env`. I agree it would be nice to bring back the convenient reconfigure feature from v2. But this is a great first step, it makes sense to keep iterating on this and use more of the `dotenv` toolchain as a stretch goal, reconfiguring it if you may, as future improvements ☺️

I was just a little confused because I thought the cli already had the capability of reconfiguring existing code with new projectId and dataset. Looks like a good addition if not, and especially the new .env stuff!

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: Good suggestion! Wasn't aware of that.

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -792,7 +792,7 @@ export default async function initSanity(
         frameworkList: frameworks as readonly Framework[],
       })
 
-      if (framework && framework.slug !== 'sanity') {
+      if (framework && framework.slug !== 'sanity' && framework.envPrefix) {
```
benja, line 795: We could have conditionally rendered the text `, using prefix...` and showed `Detected framework` regardless, but I think there's little value to showing a detected framework if nothing _really_ happens. I think we should show it once we add new increments that depend on the framework of choice.

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: I forgot to check the truthiness of prefix. Good catch!

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: | Remix app | Vue (using Vite) |
|--------|--------|
| image | image

benja:
> I was just a little confused because I thought the cli already had the capability of reconfiguring existing code with new projectId and dataset. Looks like a good addition if not, and especially the new .env stuff!

I agree we should bring back reconfiguring of existing project, perhaps even dabble further into outputting project files you're missing (depending on the framework you're in). I think this is a good first step in the right direction! :)

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+      framework?: Framework | null
+    }
+  ) {
+    const prefix = options.framework?.slug === 'sanity' ? '' : options.framework?.envPrefix ?? ''
+    const envVariables = Object.keys(object)
+      .map((key) => `${prefix}${key}="${object[key]}"`)
+      .join('\n')
+
+    try {
+      // TO-DO: instead of appending if file exists, let's check whether variables are already set so we can just update them
```
bjoerge, line 825: I worry that the "just append it" approach will lead to confusion and potential undesired surprises (like writing to the wrong dataset). The config the studio will end up using will now be determined by how the env vars are parsed (does it pick the first or last in case of duplicates?). Ideally, we should ask the user if they want to replace any existing variables, or, at the very least show a clear warning after appending in case of duplicates.

bjoerge (commented):
Love the feature! Left a couple of comments inline. I also think it would be great to have @rexxars' eye on this given that he's more familiar with it.

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -167,6 +175,24 @@ export default async function initSanity(
...
   debug(`Dataset with name ${datasetName} selected`)
 
+  // If user doesn't want to output any template code
+  const bareOutput = args.argv.find((el) => el === '--bare')
+  if (bareOutput) {
+    print(`\n${chalk.green('Success!')} Below are your project details:\n`)
+    print(`Project ID: ${chalk.cyan(projectId)}`)
+    print(`Dataset: ${chalk.cyan(datasetName)}`)
+    print(
+      `\nYou can find your project on Sanity Manage — https://www.sanity.io/manage/personal/project/${projectId}\n`
```
rexxars, line 185: Projects can be optionally be attached to an organization, so `personal` is not always correct here. Use the redirect route?
```suggestion
      `\nYou can find your project on Sanity Manage — https://www.sanity.io/manage/project/${projectId}\n`
```

packages/@sanity/cli/package.json
```diff
@@ -62,6 +62,8 @@
...
+    "@vercel/fs-detectors": "^3.8.4",
```
rexxars, line 66: We need to clarify the license and stability of these packages with Vercel before we can merge this. They are shipped with Apache 2 license files, but are marked as `UNLICENSED` and `MIT` in the `package.json` files. I've opened a PR to address the UNLICENSED case, but not sure what the correct value is for the other: https://github.com/vercel/vercel/pull/9696

As for the stability: we should either ensure that these are semver stable, or pin them.

packages/@sanity/cli/src/actions/init-project/initProject.ts
rexxars: +1, think this should make it into the current iteration

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+    const prefix = options.framework?.slug === 'sanity' ? '' : options.framework?.envPrefix ?? ''
+    const envVariables = Object.keys(object)
+      .map((key) => `${prefix}${key}="${object[key]}"`)
+      .join('\n')
+
+    try {
+      // TO-DO: instead of appending if file exists, let's check whether variables are already set so we can just update them
+
+      // flag 'a' appends if exists
+      await fs.writeFile(filename, envVariables.concat('\n'), {flag: 'a', encoding: 'utf-8'})
```
rexxars, line 828: Using the `a` flag here may lead to invalid env files if the already existing env file does not end with a newline character. Maybe we should write with a leading `\n`?

rexxars (changes requested):
The text we print on init has this phrasing:
> [...] Then we'll install an open-source JS content editor that connects to [...]

This isn't true when using the `--bare` or `--env` options. We should check if either of these flags are used and print some alternative text.

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+      throw new Error('An error occurred while creating .env', {cause: err})
+    }
+  }
+
+  async function writeToEnv(
+    filename: string,
+    object: Record<string, string>,
+    options = {} as {
+      framework?: Framework | null
+    }
```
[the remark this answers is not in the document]
benja: Ah! At first I only passed down the prefix (if it existed) and forgot to update the type. Thanks!

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: Replacements were supposed to be done at a later stage, but it turns out they are more significant than first thought. I'll get to it now.

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: Thanks! Wasn't aware 🙇🏽‍♂️

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +778,59 @@ export default async function initSanity(
...
+      })
+    } catch (err) {
+      print(err)
+      throw new Error('An error occurred while creating .env', {cause: err})
+    }
+  }
+
+  async function writeToEnv(
+    filename: string,
+    object: Record<string, string>,
```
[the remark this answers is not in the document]
benja: Thanks! I agree that `envVars` (or the like) is much clearer to the reader.

packages/@sanity/cli/src/actions/init-project/initProject.ts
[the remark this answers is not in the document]
benja: Good suggestion. It does add another manual step (prompt) to configure the output path, but despite that, I can still see it being useful.

For our new starter examples that are decoupled into `app` and `studio`, having support to provide an output path could allow one command to look like so:

`npm create sanity@latest --env --output-path app` &&
`npm create sanity@latest --env --output-path studio` to create env vars for `studio/.env`, `app/.env`.

Saying that, they would still have to go through the same flow and select the same project twice.

What about adding support for multiple dirs, such as `npm create sanity@latest --env --output-path app studio` to automatically write to both `app/` and `studio/` in the same command? @rexxars

benja: Good catch!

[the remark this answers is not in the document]
rexxars: > It does add another manual step (prompt) to configure the output path, but despite that, I can still see it being useful.

I'm not sure if we need to prompt for an output dir, was mostly wondering if we should ignore the flag if specified, or respect it (eg prepend the env var file path with it, if present).

Not sure it's worth the extra complexity - if we did support multiple output paths, it would probably be by specifying the flag multiple times. Theoretically paths can contain commas, so seems a bit error prone to split on it. It would also have to error when outputting a studio and not an env file, since it doesn't make sense to output it to multiple places. My two cents.

bjoerge (changes requested):
Some comments inline and some general suggestions:

- The logic that looks for existing env vars and replaces them seems a bit too brittle. Could we use the `dotenv` package for parsing existing dotenv file and look for exact match instead?

Replacing the env vars needs to support the various possible / common ways to define env vars (quoted, unquoted, with comments, etc.). Part of me thinks that the safes bet would be to parse with dotenv, replace or add variables to the parsed object and write it back to the file. Although this would be lossy (e.g. comments, whitespace will get lost, it feels like the added safety might weigh up for it.

Given the edge-casey nature of this, It would also be great to have unit tests for the variable substitution logic so that we make sure we cover the known ways to define env files.

benja:
The parsing and replacement logic is now enhanced, @bjoerge.

To demonstrate, running the `--env` flag on an `.env` like this will replace the following only:
```diff
SOME_KEY="ABCABC"
NEXT_PUBLIC_SANITY_PROJECT_ID_TWO="ABCABC"
NEXT_PUBLIC_SANITY_PROJECT_ID_THREE= 'ABCABC'
+NEXT_PUBLIC_SANITY_DATASET: production
-NEXT_PUBLIC_SANITY_DATASET: development
TEST_KEY="ABCABC"
NEXT_PUBLIC_SANITY_DATASET_TWO: ABCABC
+NEXT_PUBLIC_SANITY_PROJECT_ID=7l3e6r0c
-NEXT_PUBLIC_SANITY_PROJECT_ID=aj4e20da
NEXT_PUBLIC_SANITY_DATASET_THREE=ABCABC
```

I believe it's now ready for release.

bjoerge (changes requested):
Super clever approach to parse with dotenv and replace with regex, love it! 🙌🏼

This is very close to shippable now as far as I can tell, but I'd like to reduce the chance of accidentally overwrite user's existing env files re: my comment below.

packages/@sanity/cli/package.json
```diff
@@ -62,6 +62,8 @@
   },
   "dependencies": {
     "@babel/traverse": "^7.19.0",
+    "@vercel/frameworks": "1.3.2",
+    "@vercel/fs-detectors": "3.8.4",
```
rexxars, line 66: These packages now have the correct license fields, and I also landed a PR to update a dependency that was causing snyk to report a ReDoS vuln, so we should update:

```suggestion
    "@vercel/frameworks": "1.3.4",
    "@vercel/fs-detectors": "3.8.9",
```

rexxars (changes requested):
Might be worth breaking the env files into a separate file, so it can more easily be tested in isolation?

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +782,125 @@ export default async function initSanity(
...
+        print(`Appended ${key}="${envVars[key]}"`)
+        continue
+      }
+
+      print(`Found existing ${key}, replacing value.`)
+      updatedKeys[key] = value
+    }
+
+    // clone fileContents and replace existing keys by string matching
+    let updatedEnv = fileContents.repeat(1)
```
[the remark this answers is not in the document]
benja: thanks!

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +782,125 @@ export default async function initSanity(
...
+
+    // time to update or create the file
+    try {
+      const existingEnv = await fs.readFile(fileOutputPath, {encoding: 'utf8'})
+      const updatedEnv = parseAndUpdateEnvVars(existingEnv, envVars)
+      await fs.writeFile(fileOutputPath, updatedEnv, {
+        encoding: 'utf8',
+      })
+    } catch (err) {
+      await fs.writeFile(
```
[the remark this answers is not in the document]
benja: If an error occurs while parsing the file, wouldn't that break out of the try-catch and throw an error, and never actually call the `writeFile`?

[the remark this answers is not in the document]
benja: Good point. What if we catch the `ENOENT` code like so:

```ts
try {
      const existingEnv = await fs.readFile(fileOutputPath, {encoding: 'utf8'})
      const updatedEnv = parseAndUpdateEnvVars(existingEnv, envVars)
      await fs.writeFile(fileOutputPath, updatedEnv, {
        encoding: 'utf8',
      })

      print(`\n${chalk.green('Success!')} Environment variables written to ${fileOutputPath}`)
    } catch (err) {
      if (err.code === 'ENOENT') {
        await fs.writeFile(
          fileOutputPath,
          Object.keys(envVars)
            .map((key) => `${key}="${envVars[key]}"`)
            .join('\n')
            .concat('\n'),
          {
            encoding: 'utf8',
          }
        )
        print(`\n${chalk.green('Success!')} Environment variables written to ${fileOutputPath}`)
      }

      throw new Error(
        `\n${chalk.red('Error!')} Something went wrong while writing environment variables`,
        {
          cause: err,
        }
      )
    }
  }
```

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -752,6 +782,125 @@ export default async function initSanity(
...
+    for (const [key, value] of Object.entries(updatedKeys)) {
+      if (existingKeys[key]) {
+        const existingValue = existingKeys[key]
+
+        updatedEnv = updatedEnv
+          .split('\n')
+          .map((line) => {
+            if (
+              !line.trim().startsWith('#') && // ignore comments
+              new RegExp(`(^\s?${key})[= :]`).test(line) // match key
```
rexxars, line 890:
```suggestion
              new RegExp(`(^\\s*${key})((: )|( *=))`).test(line) // match key
```

I assume this was meant to be "one whitespace character, or none, at the start of the line, followed by the key, followed by a colon/equal sign. However, the current regex says:

One or none `s` character at the start of the line, followed by the key, followed by either an equal sign, a space or a colon.

My suggestion (if I got it right) says "any number of whitespace characters at the start of the line, followed by the key, followed by either a colon and a space, OR any number of spaces followed by an equals sign". The colon/equal rules and spaces I pulled together from a brief test of the dotenv parser: the colon syntax does not support spaces after the key, whereas the equal syntax does.

Whether or not you accept my suggestion, the `s` character seems like a bug - I'm sure it was just one too few escapes.

rexxars (changes requested):

packages/@sanity/cli/src/actions/init-project/initProject.ts
benja: good catch! :)) i don't speak regex too well, but ran this through some manual testing and it seems to work as good as before 👍🏼

packages/@sanity/cli/src/actions/init-project/initProject.ts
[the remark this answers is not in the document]
bjoerge: Yes works perfectly - thank you! 🙏🏼

packages/@sanity/cli/src/actions/init-project/initProject.ts
bjoerge: Oh, that `\s?` can probably be blamed on me from an earlier suggestion 🙈

bjoerge (approved):

PR #4294, opened 2023-03-20
chore(CI): add prettier-if-needed action

stipsan:
### Description

Changes `yarn check:format`, that runs on PRs, to only check prettier on changed files.
And to catch cases where something like a `prettier` update, a config change or whatever, leads to some files needing to run prettier then a bot will make a PR that ran prettier.

The bot uses a new setup we've developed that allows GitHub actions to create PRs that still trigger `push` and `pull_request` events that are declared in `.github/workflows/*.yml` by using https://github.com/apps/ecospark.
Here's what the PR looks like in action: https://github.com/sanity-io/renovate-presets/pull/75

The `tibdex/github-app-token` + `ecospark` + `peter-evans/create-pull-request` can be used to automate any chore that needs to go through the PR + actions workflow.

### What to review

Check if the PR title, commit message, PR label etc in the workflow makes sense

### Notes for release

No notes


bjoerge:
Thank you! ✨

PR #4301, opened 2023-03-21
chore: refactor e2e test setup

rexxars:
This PR tweaks the end-to-end test setup a bit to prepare for future changes:

- Moves/renames the playwright folder to `test/e2e`. Playwright is the tech used, but doesn't describe the role of the folder. End-to-end tests is commonly abbreviated to e2e, and should be somewhat more descriptive. Placing it within the test folder also makes this more obvious.
- Prefer that people use `.env.local` instead of a `.env` file. The `.env` file _should_ be able to live in git, while a `.env.local` file should not. This mirrors the behavior of vite, and thus the behavior of Sanity Studio v3.
- Actually add the example env file referenced in the e2e readme.
- Rename the environment variable used for the test. Again, playwright is the tech, e2e tests is the purpose of the token.
- Allow setting an env var to override headless setting. By default, we should use headless mode, but the user should be able to override this if they want to see the browsers doing the work.
- Move the generated HTML report, artifacts and state into the `test/e2e` folder (in subdirs) to prevent cluttering the root of the monorepo.
- Changed the default viewport to be something slightly more normal (desktop size).
- Add a (high) action timeout.
- Fixed client instantiation in helpers, and reused defined constants.

It also sets up a CI task for running the end to end tests. Currently they are _extremely_ limited, but might as well start somewhere.

Future PRs will improve on/create a framework for test isolation (to be able to do mutations and similar in many concurrent builds) and define some helpers to prevent boilerplate from leaking into individual tests.


vicmeow (approved):
lgtm

bjoerge (approved):
LGTM!

PR #4310, opened 2023-03-24
(form/inputs): PT: fix various path and selection issues

skogsmaskin:
### Description

This PR will fix some regressions in the PT-input from the recently merged [customization branch](https://github.com/sanity-io/sanity/pull/4177). Symptoms include problems with selecting and deleting objects inside the PT-input. 

Whenever there was focus on an inline block or annotation inside a text block, `useScrollToOpenedMember` hook was incorrectly setting a selection into the start of it, when it shouldn't causing trouble when deleting inline objects.

In addition PortableTextInput was incorrectly using the wrong focus path (relative instead of full) in some places further causing problems with this.

There was also an issue where an inline object would not be rendered into in the editor before a form member is emitted from formState. This made poor response when dragging, pasting inline objects into the editor, so allow the inline object to be rendered without a member too.

I have also fixed an issue related to changing block style and how focus is set back to the editor again.

### What to review

* That you can delete inline and block objects without loosing your selection.
* That copying and dragging inline objects works as expected and feels snappy.
* That you can toggle fullscreen (with the mouse) without loosing your editor selection.
* That you can change a block style without loosing your selection.

### Notes for release

Doesn't need to be mentioned that much as it's fixing regressions in code that's not released.
However there is one thing:
* Fix issue where focus is not set back to the editor when changing the style of a Portable Text block.


packages/sanity/src/core/form/inputs/PortableText/object/BlockObjectActionsMenu.tsx
```diff
@@ -24,7 +24,6 @@ import {IntentLink} from 'sanity/router'
 
 interface BlockObjectActionsMenuProps extends PropsWithChildren {
   focused: boolean
-  isActive?: boolean
```
skogsmaskin, line 27: This was incorrectly sent as a always true variable (line 320 above).

Remove it and use prop `focused` instead.

PR #4314, opened 2023-03-27
Fix issue with pasting into empty PT-input

skogsmaskin:
### Description

This will fix an regression in the PT-input when pasting content into an empty editor.

This happened because a `set` patch was sent targeting a block before a `setIfMissing` and `insert` patch was made when inserting the fragment into the content. I have rewritten the patch creation logic a bit to be both safer (remove use of `unset`) and made sure that we avoid the premature patch problem.

I have also finally found a way to test native pasting into the editors, so I have added some basic tests for pasting also.
This will test that the user can successfully paste into the editor and the content is successfully updated elsewhere.

### What to review

* That you can paste into an empty editor.
* That you can write into an empty editor.
* That yo can select all and paste over the existing content.
* That the editor field is reset (entirely) in the document when deleting content (select all + delete, or backspace delete).

### Notes for release
* Fix a recent regression in the PT-input when pasting content into an empty editor would error.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
+++ b/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
@@ -77,11 +77,13 @@ export function useSyncValue(
       // If empty value, remove everything in the editor and insert a placeholder block
       if (!value || value.length === 0) {
         debug('Value is empty')
-        const hadSelection = !!slateEditor.selection
         withoutSaving(slateEditor, () => {
           withoutPatching(slateEditor, () => {
             Editor.withoutNormalizing(slateEditor, () => {
-              Transforms.deselect(slateEditor)
+              const hadSelection = !!slateEditor.selection
+              if (hadSelection) {
+                Transforms.deselect(slateEditor)
+              }
               const childrenLength = slateEditor.children.length
               slateEditor.children.forEach((_, index) => {
                 Transforms.removeNodes(slateEditor, {
@@ -89,6 +91,7 @@ export function useSyncValue(
                 })
               })
               Transforms.insertNodes(slateEditor, slateEditor.createPlaceholderBlock(), {at: [0]})
+              // Add a new selection in the top of the document
               if (hadSelection) {
                 Transforms.select(slateEditor, [0, 0])
               }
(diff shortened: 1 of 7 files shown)


packages/@sanity/portable-text-editor/src/editor/plugins/createWithPlaceholderBlock.ts
```diff
@@ -33,25 +34,6 @@ export function createWithPlaceholderBlock({
         ],
       }
     }
-    const {onChange} = editor
```
skogsmaskin, line 36: Removing this code, as we have a similar code in the `useSyncValue` hook now and the code is more transparent in there.

packages/@sanity/portable-text-editor/test/__tests__/writingTogether.collaborative.test.ts
```diff
@@ -117,22 +117,49 @@ describe('collaborate editing', () => {
       ]
     `)
     const selA = await editorA.getSelection()
-    expect(selA).toMatchInlineSnapshot(`null`)
```
skogsmaskin, line 120: This is now changed because we set the selection again when a editor becomes empty and the user had a selection from before.

Also, different keys here as we have skipped one key generation call by removing the insert placeholder code from  `editor.onChange`.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPatches.ts
```diff
@@ -237,8 +230,12 @@ export function createWithPatches({
         // Do nothing
       }
 
-      // Unset the value if the operation made the editor empty
-      if (editorIsEmpty && ['remove_text', 'remove_node'].includes(operation.type)) {
+      // Unset the value if a operation made the editor empty
+      if (
+        !editorWasEmpty &&
+        editorIsEmpty &&
+        ['set_node', 'remove_text', 'remove_node'].includes(operation.type)
```
skogsmaskin, line 237: The unset that previously was needed on line 176 is now taken care of here instead, which is much safer.

packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
```diff
@@ -77,18 +77,27 @@ export function useSyncValue(
...
                 Transforms.removeNodes(slateEditor, {
                   at: [childrenLength - 1 - index],
                 })
               })
               Transforms.insertNodes(slateEditor, slateEditor.createPlaceholderBlock(), {at: [0]})
+              if (selectionFocusOnBlock) {
+                Transforms.select(slateEditor, [0, 0])
+              }
             })
-            Editor.normalize(slateEditor)
```
skogsmaskin, line 91: Will be normalized in the end anyway.

packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
```diff
@@ -80,15 +80,22 @@ export function useSyncValue(
         withoutSaving(slateEditor, () => {
           withoutPatching(slateEditor, () => {
             Editor.withoutNormalizing(slateEditor, () => {
+              const hadSelection = !!slateEditor.selection
```
skogsmaskin, line 83: If the value has become empty, and the user already had a selection on the previous value we don't want to loose the cursor.

bjoerge (approved):
LGTM!

PR #4319, opened 2023-03-28
chore(typings): remove `get-it` overrides

stipsan:
### Description

The `get-it` package is now shipping with typings, so it shouldn't be necessary to provide them here.

### What to review

If TypeScript builds then we're in the clear.

### Notes for release

N/A - internals only


RitaDias (approved):
Someone else might want to have a look at it, but everything looks clear :)

PR #4320, opened 2023-03-29
Fix issue with pasting empty blocks into empty editor

skogsmaskin:
### Description

This will fix an issue where if you paste content starting with multiple empty lines into an empty editor, the placeholder block will be split by `editor.insertFragment` and the latter block will have the same  _key of the first resulting in a duplicate key error message. If we in the case of an empty editor instead first split the block ourselves, this problem doesn't happen.

I have written a test for this. I have also improved test support for pasting html (besides plain text) through this work.

### What to review

* Copy two empty lines from a editor.
* Fully empty the editor
* Paste the two empty lines into the editor
* Confirm that there is no error message about duplicate keys.

### Notes for release
* Fix an issue with duplication of block _key when pasting empty lines into an empty Portable Text input.


packages/@sanity/portable-text-editor/src/utils/values.ts
```diff
@@ -123,7 +123,7 @@ export function fromSlateValue(
 
 export function isEqualToEmptyEditor(
   children: Descendant[] | PortableTextBlock[],
-  types: PortableTextMemberSchemaTypes
```
skogsmaskin, line 126: Renamed this param just to align with the rest of the codebase.

PR #4321, opened 2023-03-29
feature(core): show to-blockstyle in pt-diff

skogsmaskin:
When block styles are changed, show both from and to style in the history diff:

image

They are not 1:1 represented in the diff preview in visual size, so it's hard to know what it changed to.

### Notes for release
* Show name of block style a block changed into in review changes panel.

### DIFF
--- a/packages/sanity/src/core/field/types/portableText/diff/components/Block.tsx
+++ b/packages/sanity/src/core/field/types/portableText/diff/components/Block.tsx
@@ -68,7 +68,9 @@ export default function Block(props: {
             annotations={[diff.origin.fields.style?.annotation]}
             diff={diff.origin.fields.style}
           >
-            <Text size={0}>Changed block style from '{fromStyle}'</Text>
+            <Text size={0}>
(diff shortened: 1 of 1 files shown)


binoy14:
❓How can I test this? Not sure how to get to the view in your screenshot

skogsmaskin:
Sorry it wasn't clear. Publish something with PT. Change a sentence to 'h1' style. View history.

binoy14 (approved):
Thanks for providing the test step, was able to test it and looks good to me.

robinpyon (approved):
Thanks @skogsmaskin!

In future, I wonder if it'd be worth rendering block styles 'plainly' in the changes panel, since it's possible for users to customize these styles, e.g.

image

skogsmaskin:
> Thanks @skogsmaskin!
> 
> In future, I wonder if it'd be worth rendering block styles 'plainly' in the changes panel, since it's possible for users to customize these styles, e.g.
> 

@robinpyon - yeah, I think it does. Very much so for plain text changes at least.

PR #4322, opened 2023-03-29
fix(core): add fieldset within field to expandOperation

ninaandal:
### Description
Validation error would not focus fields in collapsed fieldsets that existed within a field. 
This was due to the fieldset not being found and expanded within the `getExpandOperations.ts`. 
By adding a fieldset within a field to the expand operation, the field can get focused when clicking the validation error. 

### What to review
Make sure that all validation error links works as expected, especially within a field and a fieldset within a field. 

### Notes for release

Fixes bug where validation error would not focus on collapsed fieldset

### DIFF
--- a/dev/test-studio/schema/debug/fieldGroupsWithFieldsetsAndValidation.js
+++ b/dev/test-studio/schema/debug/fieldGroupsWithFieldsetsAndValidation.js
@@ -1,0 +1,76 @@
+export default {
+  name: 'fieldGroupsWithFieldsetsAndValidation',
+  title: 'With fieldsets and validation',
+  type: 'document',
+  groups: [
+    {name: 'group1', title: 'Group 1'},
+    {name: 'group2', title: 'Group 2'},
+  ],
+  fieldsets: [{name: 'fieldset1', title: 'Fieldset 1', options: {collapsed: true}}],
+  fields: [
+    {
+      name: 'title',
+      type: 'string',
+      group: 'group1',
+    },
+    {
+      name: 'field1',
+      type: 'string',
+      group: 'group2',
+      description: 'Required',
+      fieldset: 'fieldset1',
+      validation: (Rule) => Rule.required(),
+    },
+    {name: 'field2', type: 'string', group: 'group1'},
+    {name: 'field3', type: 'string', group: ['group1', 'group2']},
+    {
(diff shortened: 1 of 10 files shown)


bjoerge (approved):
Great detective work 🕵🏼  While reviewing this one I realized that we wasn't traversing into nested fields, so attempting to expand anything nested more than 1 level deep did not work, so I took a stab at fixing that in a separate commit. Also realized that it's a bit hard to read this code, so did an attempt at adding some clarifying comments as well

robinpyon:
Thanks @bjoerge and @Ninaandal!

I did manage to find a minor edge case – try visit [this document](https://test-studio-git-bug-sc-13539validation-error-won-t-focus-76bc20.sanity.build/test/content/input-debug;field-groups;fieldGroupsWithFieldsetsAndValidation;b64c63a9-85de-4acb-aea3-2a73afa591d8), select **Group 2** and then click the last validation error. It will correctly switch to **All fields** but won't focus or scroll to the field in question.

bjoerge:
>I did manage to find a minor edge case – try visit [this document](https://test-studio-git-bug-sc-13539validation-error-won-t-focus-76bc20.sanity.build/test/content/input-debug;field-groups;fieldGroupsWithFieldsetsAndValidation;b64c63a9-85de-4acb-aea3-2a73afa591d8), select Group 2 and then click the last validation error. It will correctly switch to All fields but won't focus or scroll to the field in question.

That was a great catch @robinpyon! Turns out this happens because when we determine what fieldsets and groups needs expanding we traverse the form state which has already had hidden members removed. The removed hidden members also includeded members that are filtered out because they are not in the currently selected group. As a result, when expanding a path, we won't find fields in other groups than the one currently selected. The workaround for now is to add an internal `_allMembers`-property to the form state for object nodes that also includes hidden members. This allows us to traverse all members, even if they are currently not in the selected group.

We should probably consider not removing hidden members, but instead pass them on to the view layer and let the view layer filter them out, but that would be a breaking change, so added this workaround for now. I've tested quite extensively and think this should be good to merge now. Thanks for reviewing and testing 🙇🏼

bjoerge (approved):

robinpyon (approved):
Thank you @bjoerge! Just tested and all LGTM

PR #4327, opened 2023-03-29
Add support for block form components in config and plugins

skogsmaskin:
### Description

This work includes `block`, `inlineBlock` and `annotation` form component support.

This means it's now possible to write plugins or workspace configuration that change how these objects are rendered in the studio.

Brings parity with the support we have for array items on this.

This work also contains some performance optimizations with the block rendering. I have discovered that some props from formState are changing  when they should not. See own comments on these. This is crucial because these custom components could contain slow code causing editor performance to degrade, so we want the aboslute minimum of renders.

💡 We might want to create some sort of testing of re-renders here, and warn if custom code (or our own) is causing it to re-render too much or too slow.

### What to review

* Does this introduce any breaking changes?
* Does`blockProps` and `BlockAnnotationProps` contain what you would expect from such a component? I have included all the render methods we have from `ArrayOfObjectItem`, other than that they are mostly specific to the block type. We can always add later I guess, but are we happy with what we currently have?

### Notes for release
* Add support for `block`, `inlineBlock` and `annotation` form components. You can now write plugins or create workspace configuration that change how these types are rendered in the studio.


packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -98,7 +96,12 @@ export function PortableTextInput(props: PortableTextInputProps) {
     },
   }))
 
-  const {subscribe} = usePatches({path})
+  // TODO: why are these not stable???
+  const _path = useMemo(() => props.path, [])
+  const _onChange = useMemo(() => props.onChange, [])
+  const _onItemRemove = useMemo(() => props.onItemRemove, [])
+
```
skogsmaskin, line 103: Rumors says these are currently being worked on having stable (props from formState).

packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -147,9 +168,11 @@ export function TextBlock(props: TextBlockProps) {
 
   const isOpen = Boolean(memberItem?.member.open)
   const parentSchemaType = editor.schemaTypes.portableText
-  const presence = memberItem?.node.presence || EMPTY_ARRAY
+  const presence =
+    (memberItem?.node.presence.length || -1) > 0
+      ? memberItem?.node.presence || EMPTY_ARRAY
+      : EMPTY_ARRAY
```
skogsmaskin, line 175: This optimization could done in a more central place. Preferably in formState.

packages/sanity/src/core/form/types/inputProps.ts
```diff
@@ -77,12 +79,21 @@ export interface ObjectInputProps<
   /** @beta */
   onFieldClose: (fieldName: string) => void
 
+  /** @beta */
+  renderAnnotation: RenderAnnotationCallback
+
+  /** @beta */
+  renderBlock: RenderBlockCallback
```
bjoerge, line 86: The introduction of these callbacks is a breaking change, so we need to find a way to make them optional

packages/sanity/src/core/form/types/inputProps.ts
```diff
@@ -77,12 +79,21 @@ export interface ObjectInputProps<
...
+
   /** @beta */
   renderInput: RenderInputCallback
 
   /** @beta */
   renderField: RenderFieldCallback
 
+  /** @beta */
+  renderInlineBlock: RenderBlockCallback
```
bjoerge, line 95: breaking change

bjoerge (changes requested):
Great work! The introduction of the new (required) render callbacks will be a breaking change, so we should look into how we can make them optional. Arguably, the forwarding of all the different render callbacks has become a bit unwieldy, so would be good if we can find a way to make them optional at every level.

skogsmaskin:
> For the array api, props.children is the actual form editor for the item itself. In the blocks api, props.children is empty. In arrays, this allows us to render the item editor inline, in a custom component. Any thoughts on doing the same here? (probably a problem because of content editable?)

It should do this?

It was part of the work done here: https://github.com/sanity-io/sanity/pull/4177

skogsmaskin:
Added a patch to always include the form for the member items. Previously it was set to only include the form for opened items, which is making it hard to provide a inline editing interface for block objects for example.

skogsmaskin:
[Added a commit yesterday](https://github.com/sanity-io/sanity/pull/4327/commits/459f165b8e6e6d3c54a09cd121bfce36c9e0b05c) to workaround that the `onChange` prop given to the inputs are unstable and can't be used as part of the rendering callback dependencies.

To work around this I used the the `useFormCallbacks` hooks instead so we don't have to pass this prop around internally.

skogsmaskin:
Rebased on `next` and changed from `useMemo` to `useState` for the remaining unstable formState prop (`onItemRemove`).

RitaDias:
👋 
Having a look at the code now but meanwhile I had a question: Are there any specific use cases in the studio that I should be on the lookout for and try out (like for example how this could impact / improve the `config`)? :)

packages/sanity/src/core/form/types/inputProps.ts
RitaDias: I had a small question which might have been answered in separate conversations before: Do we have a reason why these new callbacks need to be mandatory?

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -24,21 +24,22 @@ import {Subject} from 'rxjs'
...
 import {FormInput} from '../../components'
 import {FIXME} from '../../../FIXME'
+import {useFormCallbacks} from '../../studio'
 import {Compositor, PortableTextEditorElement} from './Compositor'
 import {InvalidValue as RespondToInvalidContent} from './InvalidValue'
 import {usePatches} from './usePatches'
 import {RenderBlockActionsCallback} from './types'
 import {PortableTextMarkersProvider} from './contexts/PortableTextMarkers'
 import {PortableTextMemberItemsProvider} from './contexts/PortableTextMembers'
-import {_isArrayOfObjectsFieldMember, _isBlockType} from './_helpers'
```
RitaDias, line 41: potential nitpick but a thought: I understand that we are now making them explicitly `internal` on its definition, but shouldn't we keep the naming convention to more easily identify it as internal still?

skogsmaskin: > potential nitpick but a thought: I understand that we are now making them explicitly `internal` on its definition, but shouldn't we keep the naming convention to more easily identify it as internal still?

Good question. Do we even use this naming convention still? I think it looks very ugly. And we got TS to take care of it like you say.

skogsmaskin:
> 👋 Having a look at the code now but meanwhile I had a question: Are there any specific use cases in the studio that I should be on the lookout for and try out (like for example how this could impact / improve the `config`)? :)
> 
> edit: I just noticed the formComponents and the changes to config, I'm assuming that those are where the "current changes" are most visible 😄 In those use cases (in the playground usecase, should I be trying out and looking out for specific things?)

I think most of all we need to figure out about the `inputProps` and how to make all of these render callbacks optional there, like discussed above.

@snorrees and team have been testing the functionality itself pretty extensively on a real world project, so I'm pretty confident in that department.

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
RitaDias: ~~I also don't think it looks pretty don't get me wrong 😬~~ 
I'm only asking because I myself had this question somewhat recently but ended up not needing to do it.
@sanity-io/studio-dx we might move this discussion out of github but quick thoughts on this (if it's defined as `@internal` in tsdocs then perhaps we don't need to keep the naming convention for internal things)?

RitaDias: 👋 @skogsmaskin just had a chat with the team and the consensus is as long as it’s tagged with `@internal` then the naming convention is dropped. 

marking this as resolved ✨

skogsmaskin:
@bjoerge I have made all the new props optional, can you please have another look?

packages/sanity/src/core/form/types/inputProps.ts
skogsmaskin: I made them mandatory to follow the same patterns as with `renderItem` which is mandatory. All of these should probably be optional at some point, but we can change that later.

packages/sanity/src/core/form/types/inputProps.ts
skogsmaskin: These are now optional.

packages/sanity/src/core/form/types/inputProps.ts
```diff
@@ -209,6 +229,15 @@ export interface ArrayOfPrimitivesInputProps<
...
   onIndexFocus: (index: number) => void
 
+  /** @beta */
+  renderAnnotation: RenderAnnotationCallback
+
+  /** @beta */
+  renderBlock: RenderBlockCallback
+
+  /** @beta */
+  renderInlineBlock: RenderBlockCallback
```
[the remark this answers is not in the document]
skogsmaskin: These are now optional.

RitaDias (approved):
:shipit: let's gooo

PR #4334, opened 2023-03-30
refactor(portable-text-editor): ignore void edits

skogsmaskin:
### Description
Values inside void objects (block objects or inline blocks) are not the editor's responsibility to deal with, and the data doesn't interfere with the editor's selection state.

Just ignore these kind of patches. They are not edited by the editor.

`useSyncValue` will make sure to update this content when it's changed anyway.

Started out by making a failing test for this which is now passing.

There are no issues with this with normal usage of the editor, however if you send external patches to the document, the editor might crash because how it acts these patches.

### What to review

* That all regular text editing operations works as they should (should be covered by tests pretty good).

### Notes for release

Doesn't need to be mentioned I think. Doesn't affect normal/expected usage.


bjoerge (approved):
LGTM!

PR #4335, opened 2023-03-30
fix(perf): various perf improvements

bjoerge:
### Description
This adds a handful of perf opmitizations that's especially noticeable on large arrays.

### Before
image

### After
image

### What to review
Editing in the studio should work exactly as before, just feel smoother and faster

### Notes for release

- Optimized typing performance of array items. This should make editing feel a lot faster, especially for large arrays.

### DIFF
--- a/packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/List/ListArrayInput.tsx
+++ b/packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/List/ListArrayInput.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable react/jsx-handler-names */
 import {Card, Stack, Text} from '@sanity/ui'
 import React, {useCallback, useMemo} from 'react'
+import shallowEquals from 'shallow-equals'
 import {Item, List} from '../../common/list'
 import {ArrayOfObjectsInputProps, ObjectItem} from '../../../../types'
 import {ArrayOfObjectsItem} from '../../../../members'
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/List/ListArrayInput.tsx
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable react/jsx-handler-names */
 import {Card, Stack, Text} from '@sanity/ui'
 import React, {useCallback, useMemo} from 'react'
+import {isEqual} from 'lodash'
```
binoy14, line 4: 💭 should we use `react-fast-compare` instead since it's supposedly faster?

packages/sanity/src/core/form/studio/contexts/FormCallbacks.tsx
```diff
@@ -24,8 +24,66 @@ export const FormCallbacksContext = React.createContext<FormCallbacksValue | nul
...
+    ref.current.onSetPathCollapsed(path, collapsed)
+  }, [])
+  const onSetFieldSetCollapsed = useCallback((path: Path, collapsed: boolean) => {
+    ref.current.onSetFieldSetCollapsed(path, collapsed)
+  }, [])
+  const onFieldGroupSelect = useCallback((path: Path, fieldGroupName: string) => {
+    ref.current.onFieldGroupSelect(path, fieldGroupName)
+  }, [])
+
+  const contextValue: FormCallbacksValue = useMemo(
```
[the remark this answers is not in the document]
bjoerge: It's needed unless the context provider performs a shallow equal on the passed value (which I don't think it is?). Without the useMemo it will return a new object upon every render (even if the values stays the same)

packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/List/ListArrayInput.tsx
bjoerge: Actually I think `shallow-equals` makes more sense here and is a little bit clearer. I doubt there will be any noticeable differences between react-fast-compare and shallow-equals here since they both essentially do the same thing when comparing arrays of strings

PR #4341, opened 2023-03-31
fix(form/inputs): fix issues with opening annotations in the PT-input

skogsmaskin:
### Description

This fixes a recent regression where annotations within nested content (inside array items or simiar) was unable to be edited (it would not open the popover toolbar).

This happened because the path created for the `block.markDefs` object was sliced incorrectly, and it the member item with the element reference for the popovers could not be resolved, neither could the item be opened correctly by using said path.

### What to review
* That objects can be properly edited within a PT-input that is nested inside other content (like array items).

### Notes for release
* Fix issue where Portable Text annotations within nested content could not be properly edited.


ninaandal (approved):
Nice! Tested in safari, firefox, chrome and seems to work well now! Well done 💪

PR #4282, opened 2023-03-15
fix: upgrade `history` to v5

stipsan:
### Description

Bumps `history` to `v5.3.0`. There are a lot of changes since `v4.6.3`, for detailed information see the release notes:
- [v5.3](https://github.com/remix-run/history/releases/tag/v5.3.0)
- [v5.2](https://github.com/remix-run/history/releases/tag/v5.2.0)
- [v5.0](https://github.com/remix-run/history/releases/tag/v5.0.0)

Changes that are relevant for us:
- Removed unfixable warnings about pushing the same path in hash history
- Fixed some long-standing encoding issues with `location.pathname`
- Both browser and hash history have support for iframes (custom `window` objects)
- About 50% smaller than v4 (and no dependencies)
- Full TypeScript declarations and IntelliSense docs
- Support for native ESM consumption of all exports

While bumping `history` I've refactored its call sites to fix long standing issues we've had with `history.push` and `history.replace`, as well as issues when using the `unstable_history` API. For public and stable APIs there are no changes.
While those using the `unstable_history` prop might be affected by changes to the `history` API, if they are the improved TypeScript definitions should help them upgrade.

Since the rest are implementation details I'll annotate the diff view directly were relevant, to keep the PR description brief.

### What to review

In general all things related to the Router should work exactly the same as before, except more resilient in React race condition scenarios. With the biggest potential impact being on resolving intent links and default redirects. In practice this means, in the context of our Test Studio:

- [/](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/) redirects to the first workspace [/test](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/test)
- [/playground](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/playground) doesn't redirect to [/test](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/test)
- [/404](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/404) shows "Workspace not found" and clicking "Go to default workspace" takes you to [/test](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/test)
- [/test](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/test) redirects to the first tool [/test/content](https://test-studio-git-bug-sc-31148panes-lost-on-resize.sanity.build/test/content)
[...]

PR #4344, opened 2023-04-03
fix(router): resolve intent resolver regression since `v3.7.1`

stipsan:
### Description

The router refactor in #4282, released with [`v3.7.1`](https://github.com/sanity-io/sanity/releases/tag/v3.7.1) introduced a regression bug in intent link resolving.
The regression manifests itself when trying to create a new document, or navigating to a search result from the navbar, as demonstrated in this video:

https://user-images.githubusercontent.com/81981/229625967-efa1a7dd-d8d5-43a0-9197-47e926e34296.mp4

The reason why this happens is because the old router used to eagerly resolve intent links during render. The new router changed this flow to resolving them inside a `useEffect` flow so that React renders no longer have side-effects during render.
This means that when links are clicked there's an intermediary loading state that causes everything to unmount and remount, as demonstrated in the video above.
The fix solves this by adding intent resolving to the `onNavigate` handler in addition to the existing `useEffect`. Since `onNavigate` is called in response to user events such as clicking on an search result link, or a "create new..." button, it's safe to resolve intents here.
And by resolving them early we avoid the intermediary render state and unmounts only happen if the intent link resolves to a different tool or another pane tree that don't have a shared parent.

### What to review

The video below demonstrates how the intent links no longer causes dramatic unmounts, context is preserved and the UI no longer changes more than it absolutely needs to in order to resolve the intent.
There are some commented out `console.debug` calls, please ignore them as I intend to make a follow up PR after this one that resolves other router bugs I found. Priority 1 is to get a release that fixes the regression introduced in `v3.7.1` while the follow-up PR might take longer to create as it'll deal with long standing bugs, not recent regressions.

### DIFF
--- a/packages/sanity/src/router/RouterProvider.tsx
+++ b/packages/sanity/src/router/RouterProvider.tsx
@@ -52,15 +52,7 @@ export interface RouterProviderProps {
  * @public
  */
 export function RouterProvider(props: RouterProviderProps): React.ReactElement {
-  // TODO: can we do nested routes?
   const {onNavigate, router: routerProp, state} = props
-
-  const navigateUrl = useCallback(
-    (opts: {path: string; replace?: boolean}) => {
-      onNavigate(opts)
-    },
-    [onNavigate]
-  )
 
   const resolveIntentLink = useCallback(
     (intentName: string, parameters?: IntentParameters): string => {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/router/RouterProvider.tsx
```diff
@@ -52,16 +52,8 @@ export interface RouterProviderProps {
  * @public
  */
 export function RouterProvider(props: RouterProviderProps): React.ReactElement {
-  // TODO: can we do nested routes?
```
stipsan, line 55: Not yet, it requires a pretty significant refactor that require a lot of parallel work in the Structure Builder internals 😭

rexxars (approved):
Looks great, thanks!

PR #4345, opened 2023-04-04
fix(desk): compare selected on sortOrderRaw instead of sortOrder

ninaandal:
### Description
When adding selected state based on `extendedProjection`, it could cause several sort orderings to be selected, as these are not unique. Two sort orderings can have the same `extendedProjection` if they are sorting by the same field. 

This PR removes selected state based on `extendedProjection` and is only based on `sortOrderRaw` instead of `sortOrder`. 
This is necessary as a `sortOrder` based on type `string` field includes a `mapWith` prop, while the `item.params` does not add the `mapWith` prop, hence the `isEqual` check will fail and a sort option by a `string` will never be selected. 

### What to review
Make sure the correct sorting option is highlighted in the dropdown menu for sorting document lists. 

### Notes for release
Fixes issue where several sorting options could be selected concurrently in dropdown.

### DIFF
--- a/dev/test-studio/schema/standard/numbers.ts
+++ b/dev/test-studio/schema/standard/numbers.ts
@@ -66,4 +66,16 @@ export default defineType({
       readOnly: true,
     },
   ],
+  orderings: [
+    {name: 'asc', title: 'Asc', by: [{field: 'myNumberField', direction: 'asc'}]},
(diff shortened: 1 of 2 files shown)


hermanwikner (approved):
Works great! 👍

PR #4347, opened 2023-04-04
fix(core,desk): document timeline performance improvements

robinpyon:
## Description

This PR brings a number of minor fixes, UX updates and performance improvements to the document timeline and changes panel.

**Document timelines:**
- These are now virtualized and are much more performant on initial open as well as when scrolling through transactions.
- Tooltips have been removed and we now only display absolute dates, making it easier to visually parse.
- Transactions are now lazy loaded when you scroll near the bottom of the list.
  - Previously, full document transaction history would be loaded _in full_ upon opening the timeline menu. For documents with extended history, this could result in dozens of potentially unnecessary requests. [Example](https://user-images.githubusercontent.com/209129/229933026-2a435a59-a445-44ce-aa07-f95e231e793f.mp4)
- The display and positioning of the spinner has been tweaked to be less obtrusive.
- The 'since' timeline (changes panel) has been simplified – we no longer display chunks that appear _before_ the current document's revision, nor highlight the 'range' between both points.
- Other minor visual tweaks – "Current version" now reads "Latest version", we no longer use time ago-style dates and menu button labels are much more predictable.
- Selecting a timeline entry will no longer close the popover. This is counter to how popover menus typically work in the studio, but this makes it much more ergonomic to browse document revisions without having to constantly re-open the timeline.

**Changes panel:**
- We now always display the changes panel if a `since` URL param is present, regardless of how much transaction data has loaded.
  - In the event you have the changes open without sufficient transaction data (e.g. directly visit a document URL with a `since` param), the changes panel will still appear immediately and display a loading spinner. [Example](https://user-images.githubusercontent.com/209129/229933627-a2c22a73-5246-4d56-b7a8-3a08985fb4ca.mp4)
  - Previous behaviour here was to display the current live document whilst document transaction history loaded in the background. Only once enough transaction history was displayed would the changes panel then be displayed. 
    - However, this would lead to scenarios where you'd initially see an editable document, which would snap to a read-only state (potentially mid-edit!) and the changes panel would then open once enough transaction history had loaded. [Example](https://user-images.githubusercontent.com/209129/229934410-409b2928-b220-426b-a567-8343b0516e20.mp4)

**Document pane:**
- **As a result of the above, we now wait for the document timeline controller to be a 'ready' state before displaying the editable document**
  - This means that documents now wait for 1. a 'connected' connectionState, 2. a 'ready' edit state and now 3. a valid range selection timeline controller state
- We no longer re-render the entire document pane on _all_ updates to `TimelineController`. 
  - Previously, `<DocumentPaneProvider>` would re-render on every controller change, which would be especially expensive when dealing with documents with lots of history. These re-renders would also continue even after the timeline menu had been closed. [Example before](https://user-images.githubusercontent.com/209129/229938421-51a81d76-bc5e-4cdc-ab91-852f280ca83d.mp4), [example after](https://user-images.githubusercontent.com/209129/229939117-9663c76c-3c4a-436e-9951-80183c14eba0.mp4)
- Fixed issues with double fetching in strict mode, which could sporadically lead to broken timelines in development
- Transaction fetching is suspended when the document pane is closed. 
  - Previously, these would carry on even after the document pane had unmounted. [Example](https://user-images.githubusercontent.com/209129/229947231-008e6850-3b16-407b-8848-6bf70d2c1d98.mp4) (with request delay for emphasis)

## What to review

- Timeline menus: these should feel snappy, render quickly and respond well when navigating with up / down arrow keys. Transactions should lazy load as expected.
- Directly visiting a document with `since` and `rev` params should show the changes panel immediately
- The studio should be resilient when quickly navigating between URLs with document history params (e.g. when using the browser back and forward buttons)
- Loading document history shouldn't crash the studio in development
- The document pane / form should be more resilient to TimelineController errors (e.g. if you block all requests to the transactions endpoint, the document should still be editable – even if you can't select document revisions)

## Notes for release

- Performance improvements to document history


packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -119,22 +114,38 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
-    () => historyStore.getTimeline({publishedId: documentId, enableTrace: isDev}),
-    [documentId, historyStore]
-  )
   const [timelineMode, setTimelineMode] = useState<'since' | 'rev' | 'closed'>('closed')
-  // NOTE: this emits sync so can never be null
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const {historyController} = useMemoObservable(
-    () => historyStore.getTimelineController({client, documentId, documentType, timeline}),
-    [client, documentId, documentType, timeline]
-  )!
```
robinpyon, line 132: The primary culprit of unnecessary document re-renders, this was previously triggering re-renders on _every_ `handler` callback within `TimelineController`

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -119,22 +114,38 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     params.path ? pathFromString(params.path) : []
   )
   const activeViewId = params.view || (views[0] && views[0].id) || null
-  const timeline = useMemo(
-    () => historyStore.getTimeline({publishedId: documentId, enableTrace: isDev}),
-    [documentId, historyStore]
-  )
```
robinpyon, line 125: `Timeline` objects are only consumed by `TimelineController`s and aren't used elsewhere throughout the studio, so this has been abstracted out

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -119,22 +114,38 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+      next: (controller) => {
+        setOnOlderRevision(controller.onOlderRevision())
+        setRevTime(controller.revTime)
+        setSinceAttributes(controller.sinceAttributes())
+        setTimelineDisplayed(controller.displayed())
+        setTimelineReady(!['invalid', 'loading'].includes(controller.selectionState))
+      },
+    })
+    return () => subscription.unsubscribe()
+  }, [timelineController$])
```
robinpyon, line 148: Plenty of room for improvement here, it's quite verbose but explicit. I tried implementing selectors here with `useSyncExternalStore`, though came up short.

packages/sanity/src/desk/panes/document/timeline/timelineItem.styled.tsx
```diff
@@ -35,71 +28,28 @@ export const IconWrapper = styled(Flex)(({theme}: IconWrapperProps) => {
...
-        }
-      }
-    `}
-
-      ${state === 'disabled' &&
-    css`
-      [data-ui='Avatar'] {
-        opacity: 0.2;
-      }
-    `}
```
robinpyon, line 79: We no longer display the 'withinSelection' range, which simplifies this massively – this also means we don't need to render transactions in the changes panel that users cannot interact with.

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
robinpyon: This was also doubling up fetches in strict mode (as `getTimelineController` was also responsible for fetching document pairs, in addition to creating an observable controller)

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
bjoerge: Agreed - but no rush as this can always be improved on later. LMK if you want to team up to look into how to improve it.

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -336,7 +354,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     isPermissionsLoading,
     permissions?.granted,
     ready,
-    rev,
+    revTime,
```
bjoerge, line 357: good call for clarity, thank you!

packages/sanity/src/desk/panes/document/DocumentPaneContext.ts
```diff
@@ -66,7 +64,7 @@ export interface DocumentPaneContextValue {
   setTimelineMode: (mode: TimelineMode) => void
   setTimelineRange(since: string | null, rev: string | null): void
   source?: string
-  timeline: Timeline
+  timelineController$: Subject<TimelineController>
```
bjoerge, line 67: Does the outside world need this to be a Subject? (E.g. is there anything externally that calls `.next()` on it, or could it be an `Observable<TimelineController>` instead?

bjoerge (approved):
I have reviewed code only and it looks great to me - really solid work! As always, I've left some comments and suggestions but didn't spot anything that should block merge :)

packages/sanity/src/core/store/_legacy/document/document-pair/remoteSnapshots.ts
```diff
@@ -17,9 +18,5 @@ export const remoteSnapshots = memoize(
       switchMap(({published, draft}) => merge(published.remoteSnapshot$, draft.remoteSnapshot$))
     )
   },
-  (client, idPair, typeName) => {
```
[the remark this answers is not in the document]
robinpyon: Not removed! Just replaced with `memoizeKeyGen` which is also being used in a few places throughout the studio. I thought it'd make sense to reference that here too, though can easily revert if you think this would need to be keyed differently in future.

packages/sanity/src/desk/panes/document/timeline/timeline.tsx
```diff
@@ -1,61 +1,85 @@
...
+      return true
+    })
+  }, [chunks, disabledBeforeFirstChunk, firstChunk])
 
-    if (loadingElement && rootEl) {
-      const {offsetHeight, scrollTop} = rootEl
-      const bottomPosition = offsetHeight + scrollTop + LOAD_MORE_OFFSET
-      const isVisible = loadingElement.offsetTop < bottomPosition
+  const selectedIndex = useMemo(
+    () => filteredChunks.findIndex((c) => c.id === lastChunk?.id),
```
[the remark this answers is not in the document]
robinpyon: Great pick up, thanks!

packages/sanity/src/desk/panes/document/DocumentPaneContext.ts
robinpyon: It doesn't – that's been updated!

packages/sanity/src/core/store/_legacy/__workshop__/HistoryTimelineStory.tsx
```diff
@@ -42,42 +39,49 @@ export default function HistoryTimelineStory() {
...
-    historyController.setRange(params.since || null, params.rev || null)
-    forceUpdate()
-  }, [forceUpdate, historyController, params.rev, params.since])
-
-  const changesOpen = historyController.changesPanelActive()
-
-  const compareValue: Partial<SanityDocument> | null = changesOpen
-    ? historyController.sinceAttributes()
-    : null
+    const subscription = timelineController$.subscribe((controller) => {
```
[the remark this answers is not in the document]
robinpyon: I considered this! My understanding is that `useState` updates with object literals will always trigger re-renders (unlike `this.setState` which merges), so such an approach may require a custom hook or a separate reducer.

Also related to [this comment](https://github.com/sanity-io/sanity/pull/4347#discussion_r1162923574) – ideally we could use a custom hook that uses `useSyncExternalStore` under the hood or something similar which enables redux-like selectors. Would love to try and land this.

mikolajdobrucki (changes requested):
The performance of the new version is amazing 😍  Sooo much better than before!

Design-wise there is one thing I would suggest we change:

It is hard to talk about standards in Sanity Studio when it comes to dropdown lists as it seems like almost each of them is styled a bit differently.

However one thing that seems to be a consistent pattern is a padding around the list as a whole. I would suggest we use it also in this case:

**What I suggest**
image

**What it looks like now**
image

robinpyon:
Thanks @mikolajdobrucki – that's a fair suggestion, and it's been added!

packages/sanity/src/core/store/_legacy/history/useTimelineController.ts
```diff
@@ -0,0 +1,225 @@
...
+    return () => {
+      if (snapshotsSubscriptionRef.current) {
+        snapshotsSubscriptionRef.current.unsubscribe()
+        snapshotsSubscriptionRef.current = null
+      }
+    }
+  }, [client, controller, documentId, documentType])
+
+  /**
+   * Custom hook which wraps around `useSyncExternalStore`. Accepts a selector function which can be used
```
stipsan, line 152: I have a bad feeling about hooks that create hooks that are then called. It's risky magic and can easily break with React concurrent features.

I'd encourage you to consider a 2 hooks pattern where you instead pass the store:

```tsx
import {useSyncExternalStoreWithSelector} from 'use-sync-external-store/with-selector'

function Component() {}
  const {timelineStore} = useDocumentPane()
  const onOlderRevision = useTimelineSelector(timelineStore, (state) => state.onOlderRevision)
}

function useTimelineSelector(timelineStore, selector, isEqual?) {
  return useSyncExternalStoreWithSelector(timelineStore.subscribe, timelineStore.getSnapshot, timelineStore.getServerSnapshot, selector, isEqual)
}
```

We also ought to stop putting more stuff on the already giant `DocumentPaneContext` and start splitting it into multiple, separate, context providers if we are to properly solve todays performance bottlenecks in the Studio.
The pattern above can be ugpraded to the multi-context pattern:
```tsx
import {useSyncExternalStoreWithSelector} from 'use-sync-external-store/with-selector'

function Component() {}
  const onOlderRevision = useTimelineSelector((state) => state.onOlderRevision)
}

function useTimelineSelector(selector, isEqual?) {
  const timelineStore = useTimelineStore()
  return useSyncExternalStoreWithSelector(timelineStore.subscribe, timelineStore.getSnapshot, timelineStore.getServerSnapshot, selector, isEqual)
}

function useTimelineStore() {
  return useContext(TimelineStoreContext)
}
```

bjoerge: Agree with cody about the pattern of hooks creating hooks.

We are well aware that we need to set aside time to split up the DocumentPaneProvider/context, but don't necessarily think it should get in the way of merging this PR unless there's evidence it has any additional negative impact on perf?

robinpyon: @stipsan Really appreciate the feedback here

> I'd encourage you to consider a 2 hooks pattern where you instead pass the store:

Much better – that's been updated. We no longer have a hook returning a hook, nor do we store one in DocumentPaneProvider.

> We also ought to stop putting more stuff on the already giant DocumentPaneContext and start splitting it into multiple, separate, context providers

Completely agree and totally on board with this. I've opted to keep it as-is _for now_, but I think there's definitely an argument for breaking this up and we could look to do this holistically, provided that DX give the green light.

Some other notes:
- We create a `timelineStore` once in `DocumentPaneProvider` and store that in context. Components down the chain can retrieve that store to either subscribe to changes via `useTimelineSelector`, or alternatively call select methods which are called on the underlying `TimelineController` (mainly, obtaining ranges and loading more transactions)
- I think this is still an improvement over what was happening prior (storing the entire `TimelineController` class instance in context). But as you've suggested, a dedicated Timeline provider / context could go a long way in cleaning this up.
- Do we need `use-sync-external-store` if we're only supporting React 18?

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
```diff
@@ -0,0 +1,209 @@
...
+         * Until then, this workaround waits until the next call stack before calling `setRange`.
+         */
+        setTimeout(() => {
+          innerController.setRange(since || null, rev || null)
+          timelineController$.next(innerController)
+        }, 0)
+      }
+    }
+    controller.resume()
+    return () => controller.suspend()
```
bjoerge, line 132: Will this potentially re-use controllers when navigating between workspaces, or will it be suspended forever and a new one created if you come back to a previous workspace?

packages/sanity/src/core/store/_legacy/history/history/TimelineController.ts
```diff
@@ -126,11 +128,6 @@ export class TimelineController {
     return Boolean(this._rev) && (this.selectionState === 'range' || this.selectionState === 'rev')
   }
 
-  /** Returns true when the changes panel should be active. */
-  changesPanelActive(): boolean {
-    return Boolean(this._since) && this.selectionState === 'range'
-  }
-
```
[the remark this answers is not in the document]
bjoerge: :100:

packages/sanity/src/core/store/_legacy/document/document-pair/remoteSnapshots.ts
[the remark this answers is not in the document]
bjoerge: Ah, missed that - thanks for clearing that up!

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
```diff
@@ -0,0 +1,209 @@
...
+              selectionState: innerController.selectionState,
+              sinceAttributes: innerController.sinceAttributes(),
+              sinceTime: innerController.sinceTime,
+              timelineDisplayed: innerController.displayed(),
+              timelineReady: !['invalid', 'loading'].includes(innerController.selectionState),
+            })),
+            // Only emit (and in turn, re-render) when values have changed
+            distinctUntilChanged(deepEquals),
+            tap((timelineState) => {
+              timelineStateRef.current = timelineState
```
bjoerge, line 189: Since this happens before `catchError`, the current state won't be set to `INITIAL_TIMELINE_STATE` in case of an error - is this the behavior we want?

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
```diff
@@ -0,0 +1,209 @@
...
+        }, 0)
+      }
+    }
+    controller.resume()
+    return () => controller.suspend()
+  }, [rev, since, controller, timelineController$])
+
+  /**
+   * Fetch document snapshots and update the mutable controller.
+   * Unsubscribes on clean up, preventing double fetches in strict mode.
```
bjoerge, line 137: Would be interesting to hear @stipsan's view on adding workarounds for this type of strict mode behavior. I feel double fetches in strict mode is not something we should actively fight against because that's how strict mode works by design and for a reason, but I don't feel confident enough about this to have strong opinions.

bjoerge (approved):
I left a couple of minor comments/questions, but other than that this is good to go from my side. Really solid work, @robinpyon!

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
robinpyon: Cripes, it certainly isn't – that's been updated!

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
robinpyon: It won't – controllers are created on a per-document basis and are never reused. Switching workspaces will unmount the document pane, suspend any future transaction fetching and create a new controller instance.

This was prior behaviour too, though can certainly see a future where this instance is re-used and shared throughout.

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
robinpyon: I agree, this adds a lot of noise – and this is more of a quick workaround to prevent `TimelineController` from sporadically crashing due to multiple snapshot fetches. I should have marked it up as such.

At the very least – if we do choose to handle this behaviour – I imagine it's something we'd want to do within fetch helpers / custom hooks and not within consuming components (like the above).

stipsan: I didn't see this, we should never ever fight strict mode 😅 We should refactor this ASAP, and make sure we're not doing it elsewhere. If it's a problem we should instead disable Strict Mode, or only enable strict mode on parts of the render tree that falls outside the crash you're mentioning.

THere's no way to detect if React is doing a double useEffect setup + teardown + setup + teardown, or if there if it's in a background render, transition or a valid race condition 😰

stipsan: For example, this is a more idiomatic way to solve it, that also reduces strict mode noise:
```ts
 useEffect(() => {
    const subscription = remoteSnapshots(
      client,
      {draftId: `drafts.${documentId}`, publishedId: documentId},
      documentType
    ).subscribe((ev: RemoteSnapshotVersionEvent) => {
      controller.handleRemoteMutation(ev)
    })
    return () => {
      subscription.unsubscribe()
    }
  }, [client, controller, documentId, documentType])
```

Where instead of using the "need to fix this thing in strict mode as it's noisy", causing the API to use a `useRef` and more complicated book keeping. The angle is "React might cancel this side effect and trigger clean up at any time and we should anticipate it".

robinpyon: Thanks @stipsan. As above, I should have been more explicit in commenting this as a temporary fix. 

The workaround you've suggested will show very subtle side effects in strict mode (you'll see doubled-up entries in the timeline), but I believe this is more an issue with the underlying `TimelineController` and `handleRemoteMutation` not being resilient to handling multiple fetches more than anything.

When time allows – we'll aim to remove that bookkeeping and not fight strict mode as suggested!

PR #4353, opened 2023-04-07
test(cli): resolve user email for configured test token

rexxars:
### Description

Current, the tests rely on running them with a specific test user token. With this change, the tests can be run as any admin user for the project. At some point we may want to make this even more configurable so external users can utilize their own projects here too, but will leave that for a separate PR.

### Notes for release

None - internal change.


rexxars:
> Thanks! code makes sense but looks like the tests are failing 🤔

Yeah, looks like an internal server error when creating datasets. Appears to be a passing issue - will re-run the tests.

PR #4354, opened 2023-04-10
fix(cli): fix basePath not assigned to css files

### LINKED ISSUE #4254: basePath not respected for bundled css during sanity build
**Describe the bug**

Using sanity build (3.2.6) in my team, the basePath is respected everywhere (as far as we can see) except a bundled css file in index.html:

```html
<link rel="icon" href="/studio/static/favicon.ico" sizes="any"/>
...
<link rel="stylesheet" href="/static/app-a8eba63f.css"/>
```

This is likely an oversight of PR https://github.com/sanity-io/sanity/pull/3996 (issue https://github.com/sanity-io/sanity/issues/3972), of which other users report the same issue with version 3.3.1 also.

**To Reproduce**

 - Setup a Next JS site with a studio folder in the file structure.
 - Add the `/studio` basePath to both the sanity.cli.js and sanity.config.js.
 - Run the sanity build command and specify and output path to your `/public/studio` Next folder.
 - View index.html in `/public/studio` to observe the reference URL for the `app-*.css` file having `/static` instead of `/studio/static`.

**Expected behaviour**

The app css file is linked correctly as with other assets.

```
@sanity/cli (global)   3.2.6 (latest: 3.6.0)
@sanity/components    2.14.0 (up to date)
@sanity/ui            1.0.14 (latest: 1.3.0)
@sanity/vision         3.0.0 (latest: 3.6.0)
sanity                 3.2.6 (latest: 3.6.0)
```

NPM 8.19.3
NodeJS v18.13.0
Though, I was likely using v16 at the time I first saw this issue.

binoy14:
### Description

Fixes issue where css files did not get a basePath when the option was passed

fixes #4254 

### What to review

Does the change cover all use cases?

### Notes for release

- Fixes issue where basePath option is not respected for css files


packages/sanity/src/_internal/cli/server/renderDocument.ts
```diff
@@ -185,7 +185,7 @@ function getDocumentHtml(studioRootPath: string, props?: DocumentProps): string
       // If the URL is absolute, we don't need to prefix it
       return new URL(url).toString()
```
binoy14, line 186: ❓ does this also need to account for basepath? I am thinking absolute url is `https://some-site.com/style.css`

rexxars: I don't think it needs to account for it - URLs like `/foo/bar` and `foo/bar` will both throw, so this is primarily for full HTTP(S) imports

rexxars (changes requested):

packages/sanity/src/_internal/cli/server/renderDocument.ts
```diff
@@ -185,7 +185,7 @@ function getDocumentHtml(studioRootPath: string, props?: DocumentProps): string
...
     } catch {
-      return url.startsWith('/') ? url : `/${url}`
+      return `${props.basePath}${url.startsWith('/') ? '/' : ''}${url}`
```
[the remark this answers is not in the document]
binoy14: ah yeah I'll fix that

binoy14:
I added tests and created a new internal function for handling different scenarios. Let me know if the location and naming convention of things make sense

rexxars (approved):
👍

PR #4356, opened 2023-04-10
chore(perf): add perf test for large document

binoy14:
### Description

Adds perf test for a large document

- Fix deleting drafts on tests

### What to review

- Does the change in delete logic make sense?

### Notes for release

N/A (internal tooling)

### DIFF
--- a/perf/tests/arrayOf1kItems.test.ts
+++ b/perf/tests/arrayOf1kItems.test.ts
@@ -33,7 +33,10 @@ export default {
     const doc = await client.create(arrayOf1kItems)
     return {
       data: {documentId: doc._id},
-      teardown: () => client.delete(doc._id),
+      teardown: async () => {
+        await client.delete(doc._id)
+        return client.delete(`drafts.${doc._id}`)
+      },
     }
   },
   version: 1,
--- a/perf/runner/utils/testIds.ts
+++ b/perf/runner/utils/testIds.ts
@@ -23,6 +23,7 @@ type KnownTestId =
   | 'large-array-editing-another-field'
   | 'array-of-1k-string-items'
   | 'array-of-1k-reference-items'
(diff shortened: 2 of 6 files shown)


perf/tests/arrayOf1kItems.test.ts
```diff
@@ -33,7 +33,10 @@ export default {
     const doc = await client.create(arrayOf1kItems)
     return {
       data: {documentId: doc._id},
-      teardown: () => client.delete(doc._id),
+      teardown: async () => {
+        await client.delete(doc._id)
+        return client.delete(`drafts.${doc._id}`)
```
bjoerge, line 38: A slight improvement here could be to do these in a transaction e.g.:
```
return client.transaction().delete(doc._id).delete(`drafts.${doc._id}`).commit()
```

perf/tests/arrayOf1kStringItems.test.ts
```diff
@@ -29,7 +29,10 @@ export default {
     const doc = await client.create(arrayOf1kStringItems)
     return {
       data: {documentId: doc._id},
-      teardown: () => client.delete(doc._id),
+      teardown: async () => {
```
bjoerge, line 32: Same as above re: transaction.

bjoerge (approved):
Thank you for this! LGTM!

rexxars (approved):
Unless we are depending on the result of the transaction being queryable immediately after the mutation succeeds, you can speed up the `commit()` calls by [making the visibility async](https://github.com/sanity-io/client#mutation-options):

```diff
-someTransaction.commit()
+someTransaction.commit({visibility: 'async'})
```

The studio generally only needs to wait for them to be queryable in document lists, whereas navigating straight to a document ID will use a different endpoint that can immediately "see" the documents.

PR #4360, opened 2023-04-11
fix(core): fix permissions when initialValue is empty object

binoy14:
### Description

Adds `_type` when the initialValue is empty object when statically or in a function

### What to review

Migrates https://github.com/sanity-io/sanity/pull/4346 to v3 as well as adds tests for it. Let me know if the test structure makes sense

### Notes for release

Fixes permission issue when initialValue is set to empty object


rexxars (approved):
This is great!

PR #4361, opened 2023-04-11
fix(core): clear unused references in an array when clicked outside

binoy14:
### Description

This PR fixes issue where unused references don't clear when clicked outside which works in v2. This second attempt makes sure that the options menu works as intended and does not trigger cancel when clicking on actions like duplicate, etc.

### What to review

https://user-images.githubusercontent.com/6476108/231271452-62208359-e99e-40c4-abd5-624e919b5cfe.mov

The solution to fix this isn't ideal IMO but it is the only way I can think this could work. Since the ReferenceInput and ReferenceItem don't necessarily have a strict parent child relation and the menu popover is located in the ReferenceItem the only way for ReferenceInput to know if that is being clicked is when a ref is shared using context.

### Notes for release

Fixes issue where unused references in array were not cleared when clicking outside

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
@@ -3,8 +3,7 @@
 
 import React, {KeyboardEvent, FocusEvent, useCallback, useRef, useState, useMemo} from 'react'
 import {concat, Observable, of} from 'rxjs'
-
-import {catchError, distinctUntilChanged, filter, map, scan, switchMap, tap} from 'rxjs/operators'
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceInput.tsx
```diff
@@ -252,8 +260,31 @@ export function ReferenceInput(props: ReferenceInputProps) {
       })),
     [searchState.hits]
   )
+
+  const isEditing = focusPath.length === 1 && (focusPath[0] === '_ref' || focusPath[0] === '$')
```
bjoerge, line 264: I don't think we need to check for `$` anymore(?) since we got rid of that idiosyncrasy in v3

PR #4363, opened 2023-04-12
fix: refactor desk tool intent resolver to reduce remounts

stipsan:
### Description

Primarily refactors intent resolving on the DeskTool level to no longer unmount the desk structure as it resolves an intent. It now does it in parallel, to preserve as much of the current UI as possible. In a future PR I want to explore fully persisting the current UI before transitioning to the resolved intent and completely eliminate the intermediary states (demonstrated in the videos in "What to review")

Also fixes up a lot of minor issues related to how components handle `useRouter` and `useRouterState` hooks and derivatives.

### What to review

Attaching before/after videos below that demonstrate the most important changes to review.

##### Before

https://user-images.githubusercontent.com/81981/231619124-f7213109-b4c0-4a27-ac95-4bcba1c1669a.mov

##### After

https://user-images.githubusercontent.com/81981/231745112-86dda8be-535d-4190-be74-77ce80d93269.mov

##### Before

https://user-images.githubusercontent.com/81981/231618734-1235c836-8119-4abc-812c-625390f59817.mov

##### After

https://user-images.githubusercontent.com/81981/231745414-6d299ff2-86ad-45d6-99d7-010fb42d3690.mov

#### Loading a resolved edit intent on Slow 3G

Demonstrates how "popping" is reduced, with the changes to `LoadingPanes` in this PR as well as the impact of no longer doing full unmounts in the DeskTool when resolving an intent.

##### Before
https://user-images.githubusercontent.com/81981/231618039-dbbde3da-4380-40ca-97d2-5f79b3f8dac6.mov

##### After

https://user-images.githubusercontent.com/81981/231618525-b928fc05-263c-4e21-8454-18a613e797f3.mov

Other than these we should go over and test the minor changes done to hooks and components that were changed in this PR, making use of `useMemo`, and `useRouterState(selector)`.

### Notes for release

- fix: refactor desk tool intent resolver to reduce remounts


RitaDias:
👋 
Still reviewing the code, but from reviewing the videos I do have a concern specific to "Slow resolve (shows a toast that indicates progress)". 

It looks to me that we're introducing a new way of "loading" content (toast) that we don't have anywhere else in the studio and could introduce some confusion. As far as I can tell we use the spinner pretty consistently across the studio with the intent of "loading".

I will leave it to @mikolajdobrucki but it felt like something that should be brought up.

packages/sanity/src/core/studio/StudioLayout.tsx
```diff
@@ -110,9 +117,11 @@ export function StudioLayout() {
   )
 }
 
+// @TODO re-use `LoadingComponent`, which is `LoadingScreen` by default, to reduce "popping" during initial load
```
RitaDias, line 120: Why in the future and not now? It would require more refactoring and rewriting?

packages/sanity/src/desk/components/deskTool/intentResolver/Redirect.tsx
RitaDias, line 1: Why are we deleting this file?

packages/sanity/src/core/studio/StudioLayout.tsx
stipsan: Out of scope, don't have the time as our deadline is rapidly approaching 💀

packages/sanity/src/desk/components/deskTool/intentResolver/Redirect.tsx
stipsan: No longer used anywhere ☺️

stipsan:
@RitaDias it's not meant to be a new way to load content. Our routing currently don't have the ability to know ahead of time what'll happen as an intent link is resolved. 
As mentioned in the PR, the next PR will address this and the fallback toast will be removed and we'll be back on spinners 😅

packages/sanity/src/desk/components/deskTool/DeskToolBoundary.tsx
```diff
@@ -20,15 +17,6 @@ export function DeskToolBoundary({tool: {options}}: DeskToolBoundaryProps) {
...
 
-  const {state: routerState} = useRouter()
-  const intent = useMemo(() => {
-    const intentName = typeof routerState.intent === 'string' ? routerState.intent : undefined
-    const params = isRecord(routerState.params) ? routerState.params : EMPTY_RECORD
-    const payload = routerState.payload
-
-    return intentName ? {intent: intentName, params, payload} : undefined
-  }, [routerState])
-
```
RitaDias, line 31: We don't need this anymore? Wondering about it since looking at the comment on line 46 🤔 
My assumption is that we don't need it because we are resolving this at a different time*?

stipsan: Yes, this data was only used by `<IntentResolver>` so I moved it into that component. This reduces the re-render burden for `DeskToolBoundary` since it no longer has to re-render on every single URL update 😌

RitaDias: Really cool 🙇

stipsan:
@RitaDias pushed a commit that removes the toast and uses a standard `<LoadingPane>` instead. Updated the "After" footage to reflect what it looks like now 😄

RitaDias (approved):
This looks good to me, I think that I just want @mikolajdobrucki to have a look at the videos and description before merging :)

mikolajdobrucki (approved):
Looks great! 🚀 

One small comment that is not a blocker for release but would be great to fix in the next round:

Let's try to make the spinners more consistent, following the rules:

1. Spinner at the top, label below
2. Label in `Sentance case` ending with `…`

+ also check if size of the spinner, font style of the label and spacing between them is the same in all cases


image

stipsan:
Thanks @mikolajdobrucki, I'll definitely come back and work on those after we got the other priorities out of the way 🙏

PR #4364, opened 2023-04-12
Fix issue with invalid Portable Text value causing crash

skogsmaskin:
### Description

This will fix an issue where the syncing of a invalid value would continue to be iterated on, resulting in insert operations
with invalid paths that would cause the editor to crash.

When an invalid block is discovered, don't continue to iterate on the remaining blocks of the value, and don't output a new value to the change observable.

Added a test for this.

### What to review

That invalid blocks inside the editor would produce a nice error and resolve message, and not cause the editor to crash.

### Notes for release
* Fixed an issue where a invalid Portable Text value would cause the editor to crash.


robinpyon (approved):
Thanks @skogsmaskin  – LGTM!

PR #4366, opened 2023-04-12
chore(cli): fix basePath test flakiness

binoy14:
### Description

I was reading the css file from wrong directory so based on if v2 test run first or not it would fail. Now that it is reading the files from correct directory it should not fail flakily 

### Notes for release

N/A


rexxars (approved):
🙏

PR #4368, opened 2023-04-12
fix: exhaustive dependency arrays

rexxars:
### Description

PR to fix exhaustive dependency arrays for hooks. Feel free to push fixes to this branch.

### What to review

- Affected areas still work as expected

### Notes for release

None.


packages/sanity/src/core/form/inputs/CrossDatasetReferenceInput/useReferenceInfo.ts
```diff
@@ -42,27 +42,35 @@ export function useReferenceInfo(
...
+      docInfo._id
+        ? getReferenceInfo(docInfo).pipe(
             map(
               (result) =>
                 ({
                   isLoading: false,
                   result,
                   error: undefined,
                   retry,
+                  retryAttempt,
```
stipsan, line 57: Adding `retryAttempt` here makes it natural for the linter to treat it as part of the deps array.

packages/@sanity/vision/src/components/ParamsEditor.tsx
```diff
@@ -25,9 +25,15 @@ export function ParamsEditor(props: ParamsEditorProps) {
   const {onChange} = props
   const {raw: value, error, parsed, valid} = eventFromValue(props.value)
   const [isValid, setValid] = useState(valid)
+  const [init, setInit] = useState(false)
 
   // Emit onChange on very first render
-  useEffect(() => onChange({parsed, raw: value, valid: isValid, error}), [])
+  useEffect(() => {
+    if (!init) {
```
stipsan, line 32: Using `setState` is the canonical way in React today for "run exactly once, on mount" use cases.

packages/sanity/src/core/form/inputs/PortableText/hooks/useHotKeys.tsx
```diff
@@ -7,7 +7,7 @@ export function useHotkeys(hotkeys: HotkeyOptions): HotkeyOptions {
 
   // Guard that hotkeys from props will be a stable object.
   // If this props is defined inline and is always a new object, there will be issues with key handling and cursor!
-  const initialHotkeys = useMemo(() => hotkeys, [])
+  const [initialHotkeys] = useState(() => hotkeys)
```
stipsan, line 10: To save an initial snapshot of a prop it's best to use `useState`, as it's guaranteed to persist it for the lifecycle of the component.
While `useMemo` isn't guaranteed to behave this way, even though it's given an empty deps array. React reserves the right to release it to free up memory at any point in time.

skogsmaskin: Yeah, it's important that this is only the initial prop or the warning will not work.

bjoerge (approved):
LGTM, thanks for doing this!

PR #4370, opened 2023-04-13
chore(CI): create yarn lockfile maintenance action

stipsan:
### Description

On this `prettier-if-needed.yml` PR: https://github.com/sanity-io/sanity/pull/4369
I noticed it committed a change to `yarn.lock`, which is odd. It shouldn't. 
I ran `yarn` myself on `git checkout origin next` and can confirm the lockfile is currently out of sync. 
This can happen due to PR queues and how merge conflicts or ordering happens.
In any case this PR introduces an action that will create a PR that fixes `yarn.lock` whenever needed.
The benefit of this is to avoid multiple PRs now changing `yarn.lock` as authors run `yarn` locally and the lockfile is currently stale in `next`.

### What to review

I didn't commit a fixed `yarn.lock` on purpose so that we can test if the bot works and is able to fix the currently stale `yarn.lock` on `next`.
Merging this PR should also trigger the prettier bot to auto-close: https://github.com/sanity-io/sanity/pull/4369

### Notes for release

- N/A

### DIFF
--- a/.github/workflows/prettier-if-needed.yml
+++ b/.github/workflows/prettier-if-needed.yml
@@ -46,7 +46,7 @@ jobs:
           key: prettier-${{ hashFiles('yarn.lock') }}
       - run: yarn prettier --write
       - name: GitHub blocks PRs from automation that alter workflows in any way
-        run: git restore .github/workflows
+        run: git restore .github/workflows yarn.lock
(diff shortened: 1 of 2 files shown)


RitaDias (approved):
📑 🔍

PR #4385, opened 2023-04-14
fix(core): capitalize type name if item type lacks title in pte

ninaandal:
### Description
In schema, when a field only has a `name` the `title` is automatically generated and should be uppercase in the PTE. 
Before: 
image

### What to review
Make sure that if a field in PTE does not have a `title` field, the automatically generated title is uppercase. 

### Notes for release
Fixes automatically generated title in PTE to be uppercase.


skogsmaskin (commented):
You could be using `upperFirst` from 'lodash' here also.

I think it could be as simple as:

`const title = item.type.title || upperFirst(item.type.name)`

Not sure why it was using the `item.type?.type` here. With this, it will be `reference` as a result here for references, and not the actual type name. Maybe it's something that made sense in v2 and not for v3.

skogsmaskin (approved):
I believe this is the way! 👍

PR #4386, opened 2023-04-14
fix(perf-runner): add support for running tests against recent tags only

bjoerge:
### Description
Small change that adds support for running perf tests against --count recent tags (typically releases).

For example, to run the perf tests against the 5 most recent tags: `yarn perf:test --tagsOnly --count=5`

Note: The copy edits are mainly to make the output more accurate (it's no longer just "branch" or "branches))

### What to review
Check that the code makes change

### Notes for release

n/a: internal tooling


binoy14 (approved):
Thanks for doing this!

PR #4391, opened 2023-04-18
feat(core): internal form field actions

mariuslundgard:

### DIFF
--- a/dev/test-studio/form/FormField.tsx
+++ b/dev/test-studio/form/FormField.tsx
@@ -1,0 +1,20 @@
+import {EllipsisVerticalIcon} from '@sanity/icons'
+import {Button} from '@sanity/ui'
+import React from 'react'
+import {FieldProps} from 'sanity'
+import {useDocumentPane} from 'sanity/desk'
+
+export function FormField(props: FieldProps) {
+  const {documentType} = useDocumentPane()
+
+  if (documentType === 'fieldActionsTest') {
+    return props.renderDefault({
+      ...props,
+      actions: (
+        <Button fontSize={1} icon={EllipsisVerticalIcon} mode="bleed" padding={2} tabIndex={-1} />
+      ),
+    })
+  }
+
+  return props.renderDefault(props)
+}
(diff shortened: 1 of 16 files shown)


packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
```diff
@@ -98,16 +138,37 @@ export function defaultResolveFieldComponent(
...
     return ImageOrFileField as React.ComponentType<Omit<FieldProps, 'renderDefault'>>
   }
+
   if (typeChain.some((t) => isReferenceSchemaType(t))) {
     return ReferenceField as React.ComponentType<Omit<FieldProps, 'renderDefault'>>
   }
 
+  if (isDescendentOfType(schemaType, 'slug')) {
+    return PrimitiveField as React.ComponentType<Omit<FieldProps, 'renderDefault'>>
+  }
```
rexxars, line 155: How does this change relate to this PR? If it is still required, could we use the same approach as above (eg using `typeChain`)?

dev/test-studio/form/FormField.tsx
```diff
@@ -0,0 +1,20 @@
...
+    return props.renderDefault({
+      ...props,
+      actions: (
+        <Button fontSize={1} icon={EllipsisVerticalIcon} mode="bleed" padding={2} tabIndex={-1} />
+      ),
+    })
+  }
+
+  return props.renderDefault(props)
+}
```
rexxars, line 20: To make it clear that this is only demoing the field actions capability, could we use a different icon and add a description? Currently you have to read the code to understand that an ellipsis button with no action is actually the _intended_ behavior here (I know this is test-studio only, but still provides clarity).

```suggestion
import {ApiIcon} from '@sanity/icons'
import {Box, Button, Text, Tooltip} from '@sanity/ui'
import React from 'react'
import {FieldProps} from 'sanity'
import {useDocumentPane} from 'sanity/desk'

export function FormField(props: FieldProps) {
  const {documentType} = useDocumentPane()

  if (documentType === 'fieldActionsTest') {
    return props.renderDefault({
      ...props,
      actions: (
        <Tooltip
          content={
            <Box padding={2}>
              <Text>Demo: Field actions extension point</Text>
            </Box>
          }
        >
          <Button fontSize={1} icon={ApiIcon} mode="bleed" padding={2} tabIndex={-1} />
        </Tooltip>
      ),
    })
  }

  return props.renderDefault(props)
}
```

rexxars (approved):
Overall this looks good, left a few inline comments.

Could you briefly describe the field groups change?

rexxars (changes requested):
Oh, tests needs a new snapshot to account for UI changes it seems.

PR #4393, opened 2023-04-18
fix(cli): use more recent version ranges for new projects

rexxars:
### Description

_Note: This is not a complete fix, but will _mitigate_ problems until we can properly fix it._

pnpm 8 has switched resolution to use the _lowest_ matching version when installing dependencies. This is apparently done for performance reasons, but creates problems for us since it means pnpm users will get Sanity 3.0.0 instead of the latest. The same goes for other root-level dependencies.

There is no flag to set resolution mode, and installing a pnpmrc seems a little invasive, so I think the real solution is to resolve the latest (compatible) version of dependencies upon install and write that to the project `package.json`. Until we can get around to that, this will at least bring versions up to a much more recent version of Sanity, and avoids a number of react hooks warning on the dev server startup when using pnpm.

### DIFF
--- a/packages/@sanity/cli/src/studioDependencies.ts
+++ b/packages/@sanity/cli/src/studioDependencies.ts
@@ -2,16 +2,16 @@ export const studioDependencies = {
   // Dependencies for a default Sanity installation
   dependencies: {
     // Official studio dependencies
-    sanity: '^3.0.0',
+    sanity: '^3.9.0',
 
     // Official studio plugin dependencies
-    '@sanity/vision': '^3.0.0',
+    '@sanity/vision': '^3.9.0',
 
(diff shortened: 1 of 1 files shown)


binoy14 (approved):
Looks good, just one question. Does this mean we would need to bump this version every release or this is just temporary?

rexxars:
> Looks good, just one question. Does this mean we would need to bump this version every release or this is just temporary?

This is temporary - we should make sure that during `sanity init`/`npm create sanity` we resolve the latest versions and write those to the `package.json` file, which will "fix" the issue. I personally think this change in behavior is kind of weird by pnpm and should be left to lockfiles, but 🤷‍♂️ - not much we can do about it.

PR #4397, opened 2023-04-19
fix(validation): Remove recursive validation check for annotations in PTE

ninaandal:
### Description
During the v3 work, the type for annotations were changed to be arrays. 
The check where validation is added to arrays will therefore include annotations as well now, and a separate check for annotations are not necessary. By removing this check for annotations, the validation messages are now only shown once. 

### What to review
That validation for annotations work as expected and are now duplicated anymore. 

### Notes for release
Fixes bug where validation messages on annotations were shown twice.

### DIFF
--- a/packages/@sanity/validation/src/validateDocument.ts
+++ b/packages/@sanity/validation/src/validateDocument.ts
@@ -210,36 +210,6 @@ function validateItemObservable({
     )
   }
 
-  // markDefs also do no run nested validation if the parent object is undefined
-  // for a similar reason to arrays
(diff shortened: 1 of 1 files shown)


skogsmaskin (approved):
Yes, I think we are getting these double now because the block schema was more properly defined during the v3 work and there are two different codepaths doing the validation currently.

I think general validation code is able to figure out that markDefs have to be validated now as a array of objects. And we don't need this custom code, which is a leftover from v2 (probably).

It works the way it's supposed to by removing this code, also recursively and nested.

PR #4398, opened 2023-04-19
fix(desk): Change message shown after successfully deleting document

ninaandal:
### Description
Simplifies message shown after successfully deleting document. 
Changing the copy of the toast to a longer, more descriptive explanation would provide more context but most likely would make the experience more confusing. How to retrieve a deleted document should be covered in docs instead. 

### Notes for release
Changes message shown after successfully deleting a document.


mikolajdobrucki (approved):
LGTM 👍

PR #4400, opened 2023-04-19
fix(core): fix focus issues when a new reference field is added

binoy14:
### Description

The issues are related to when an application is not running in strict mode. This is important because in [StrictMode](https://react.dev/reference/react/StrictMode#strictmode) every effect gets runs twice but when not using strict mode, i.e in production the hook only gets called once. 

Issue 1: When a new reference item is created the focus is not set into the AutoComplete
Issue 2: In a virtualized list if there is a validation error clicking on the error on top of the document panel does not scroll to the validation error

### What to review

The way to review this PR is looking at the tests first and if the assertions makes sense.
I have added tests for `usePrevious`, `useOnClickOutside` and `useScrollIntoViewOnFocusWithin` hooks while I was in the folder without any change to the existing hooks. 

#### Whats Changed

`useUpdate` Hook: Currently the hook checks for an initial mount and does not run the compare function when it's an initial mount. The hook also utilized `usePrevious` hook to store the previous value. This is an issue because when array of reference field is created it is by default focused, i.e (focused = true) and that does not change until a user interaction occurs. See code before

https://github.com/sanity-io/sanity/blob/c6464400f2d971982668c1e66d53fd05944e6bb5/packages/sanity/src/core/form/members/object/fields/ArrayOfObjectsField.tsx#L60-L64

Since the value it is listening for does not change and it ignores the first render you run into situation where previous and the current value becomes the same and it never fires the compare function in the file above hence never focusing on the input field.

❓Does this change affect anything that I overlooked? From my testing this does not seem to be an issue and is helping with scroll issues where it was previously not working

### Notes for release

- Fixes issue where unused references are cleared on clicking outside
- Fixes issue where the validation error is not focused when clicking on error in the panel

### DIFF
--- a/packages/sanity/src/core/form/hooks/__tests__/useDidUpdate.test.ts
+++ b/packages/sanity/src/core/form/hooks/__tests__/useDidUpdate.test.ts
@@ -1,0 +1,125 @@
+/* eslint-disable max-nested-callbacks */
+import {renderHook} from '@testing-library/react'
+import {StrictMode} from 'react'
+import {useDidUpdate} from '../useDidUpdate'
+
+describe('useDidUpdate', () => {
+  describe('Normal Mode', () => {
+    it('calls the didUpdate callback when the value changes', () => {
+      const previousValue = {foo: 'bar'}
+      const currentValue = {foo: 'baz'}
+      const didUpdate = jest.fn()
+
+      const {rerender} = renderHook(({value}) => useDidUpdate(value, didUpdate), {
+        initialProps: {value: previousValue},
+      })
+
+      // The first time, didUpdate should be called
+      expect(didUpdate).toHaveBeenCalledTimes(1)
+
+      rerender({value: currentValue})
+
+      expect(didUpdate).toHaveBeenCalledWith(previousValue, currentValue)
+    })
+
+    it('calls the compare function when it is passed', () => {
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/form/hooks/__tests__/useDidUpdate.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      const currentValue = {foo: 'baz'}
+      const didUpdate = jest.fn()
+
+      const {rerender} = renderHook(({value}) => useDidUpdate(value, didUpdate), {
+        initialProps: {value: previousValue},
+        wrapper: StrictMode,
+      })
+
+      // The first time, didUpdate should be called. StrictMode runs hooks twice https://react.dev/reference/react/StrictMode#strictmode
+      expect(didUpdate).toHaveBeenCalledTimes(2)
```
RitaDias, line 77: Very good comment to make sure this doesn't get lost and get another person wondering why 😙 👌

RitaDias (approved):
Looked through the tests and did some manual look around as well. Everything looks good as far as I can tell!
Nothing comes to mind in terms of changes that you might have overlooked but I'll keep an eye if Bjørge or Espen think of something 👁️

PR #4403, opened 2023-04-20
chore(.github): add code owners

RitaDias:
### Description

Add code owners to specific parts of the monorepo in the `next` branch, this is just a beginning stage and more owners will be added as we go along.

### What to review

the owners

### Notes for release

n/a (internal only)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,31 @@
+# These owners can be changed and added to in the future,
+
+# for now these are the main people that we are required to have eyes on PRs
+
+# that changed items in the given spaces
+
+# ------------- Packages
+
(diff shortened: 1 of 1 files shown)


binoy14:
❓ Does it make sense to use `* studio-dx` to default add group or is that too much noise?

RitaDias:
> ❓ Does it make sense to use `* studio-dx` to default add group or is that too much noise?

_Potentially_, but I also think that it would make sense to do it 😂 
I guess this is also in part of something that we would like to run by Rob when he comes back because EX might like to iron things out between themselves before having to do DX. I feel that can be added once we have that chat?

binoy14 (approved):
Thanks for doing this!

PR #4405, opened 2023-04-20
Update ReferenceChangedBanner.tsx

evenwestvang:


packages/sanity/src/desk/panes/document/documentPanel/ReferenceChangedBanner.tsx
```diff
@@ -166,8 +169,11 @@ export const ReferenceChangedBanner = memo(() => {
           ) : (
             <>
               <Box flex={1} marginLeft={3}>
-                <TextOneLine title="This reference has been removed since you opened it." size={1}>
-                  This reference has been removed since you opened it.
+                <TextOneLine
+                  title="The reference you followed to this document was removed."
+                  size={1}
+                >
+                  The reference you followed to this document was removed.
```
mikolajdobrucki, line 176: How about making it consistent with the update above?

> The reference to this document was removed.

I would consider is as potentially more readable.

evenwestvang: There can be many references to a document. It's specifically the one you followed that changed. The suggestion is also more vague as it's not clear what "it" refers to: document or reference. You followed a reference to the document. It has now been changed.

evenwestvang: If we think that is too vague we should perhaps do away with the warning?

rexxars: Did we make a decision here? Would be nice to move this one along in some direction.

rexxars:
Closing as this is out of date. If we still want to make changes to this text, we should update the i18n resources now.

PR #4408, opened 2023-04-21
Allow for skipping visibility in preview element (fix flickering issue)

RitaDias:
### Description

A bug was reported where someone using the `<preview>`  with the default layout was getting a flickering when there was a re-render. The data was the same, but the interface flickered. The original use case was in a drag and drop situation where the re-render made sense to exist. The example below is a simple way of reproducing it.

The error:

https://user-images.githubusercontent.com/6951139/233350609-7c684353-88c8-4b6e-a21e-84042749b584.mov

*The solution* that I arrived to was after trying different things. The main issue is that because a re-render is happening, the ref of the element in the preview is changed so the visibilty needs to be re-rendered multiple times. After looking at the way that the fetching of the data (when necessary) and the visibility were set up there wasn't any indication that they were slow (they were all miliseconds fast). 

The best solution for this particular issue, after discussing with Bjørge and trying different solution was to allow for an internal prop that can skip the visibility. We already have an exception when it comes to some `layouts` (`isPTE` renamed to `shouldSkipVisbility`), this just adds an external way to allow users to use it.

The error and fix:

https://user-images.githubusercontent.com/6951139/233351365-8c8f2745-380e-4ddf-b166-ea172765d274.mov

### What to review

1. In the test-studio, go to the Components API playground workspace (`custom-components/content/book`)
2. Go into any of the books and press the re-render to check behavior in different layouts

### Notes for release

n/a internal updates tooling

### DIFF
--- a/dev/test-studio/components/formComponents.tsx
+++ b/dev/test-studio/components/formComponents.tsx
@@ -1,11 +1,64 @@
-import React from 'react'
-import {Card} from '@sanity/ui'
-import {definePlugin, InputProps, FieldProps, ItemProps, PreviewProps} from 'sanity'
+import React, {useState} from 'react'
+import {Card, Stack, Button, Text} from '@sanity/ui'
+import {definePlugin, InputProps, FieldProps, ItemProps, Preview} from 'sanity'
 
 export function Input(props: InputProps) {
+  const [count, setCount] = useState(0)
+
   return (
     <Card data-testid="input-config-component" padding={2} border tone="primary">
-      {props.renderDefault(props)}
+      <Stack space={2}>
+        <Button onClick={() => setCount((c) => c + 1)} text={`Re-render: ${count}`} />{' '}
+        <Text>Default - no skip</Text>
+        <Preview
+          key={`default-no-${count.toString()}`}
+          layout="default"
+          value={props.value}
+          schemaType={props.schemaType}
+        />
+        <Text>Default - with skip</Text>
+        <Preview
+          key={`default-${count.toString()}`}
+          layout="default"
+          value={props.value}
+          schemaType={props.schemaType}
+          __internal_skip_visibility_check
+        />
+        <Text>No Layout - no skip</Text>
+        <Preview
+          key={`no-layout-no-${count.toString()}`}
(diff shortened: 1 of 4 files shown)


dev/test-studio/components/formComponents.tsx
RitaDias, line 1: Will be reverted once ✅

packages/sanity/src/core/preview/components/PreviewLoader.tsx
```diff
@@ -27,21 +27,31 @@ export function PreviewLoader(
...
+    component,
+    style: styleProp,
+    schemaType,
+    __internal_skip_visibility_check,
+    ...restProps
+  } = props
 
   const [element, setElement] = useState<HTMLDivElement | null>(null)
-  const isPTE = layout && ['inline', 'block', 'blockImage'].includes(layout)
+  const shouldSkipVisbility =
```
RitaDias, line 41: While discussing with Bjørge we spoke about removing this from the component and replacing this in whichever parts made sense in PTE (given the name of the const)

However, after spending some more time on it, I realised that it's not only PTE situations where this would be an issue. The naming itself i misleading because while it probably has most import in PTE it also means that, if we were to remove this, if I was to create a preview element with these layouts they would start flickering as well which I don't think is what we want to do.

Given this I renamed the const

RitaDias: After discussing this a bit with @bjoerge we made the decision that the fact that the layout was impacting the "visibility" is incidental and it shouldn't impact it in the way that it does. After some investigation we found the places to make sure that PTE would work as before despite the changes to the PreviewLoader (cc @skogsmaskin )

860a6e39fd80375cdacf934039597b4d77618e5c

It is important that if people are creating and using the <preview> in the way that was given as an example that they might need to refactor the component or simply use the new provided prop. The layouts impacting the way they did shouldn't be taken into account.

skogsmaskin (approved):
I have evaluated and approved the changes in regard to the PT-input 👍

bjoerge (approved):
LGTM, great work! 👏🏼

PR #4410, opened 2023-04-21
fix(core): list virtualization is broken inside modals

### LINKED ISSUE #4396: Virtualized arrays of objects not rendering all items in 3.9.0
**Describe the bug**

Virtualized arrays of objects not rendering all items. The array in the screenshot has 4 elements. I can confirm all 4 being available in the studio by inspecting the document.
This works again when downgrading to 3.8.3

**To Reproduce**

In an array of objects that contain arrays of objects just open the popover to edit one element and look at the list of items.
For me this list is outside the viewport of the popover initially and I have to scroll down. This might affect the sizing logic perhaps.

**Expected behavior**

Should work like 3.8.3 with non-virtualized lists.

**Screenshots**

Screenshot 2023-04-19 at 13 59 23

sanity 3.9.0

macOs ventura.
Arc browser, hence Chrome really.

pnpm 7.26.3
node v18.13.0

binoy14:
### Description

fixes #4396 

When virtualized list is inside the modal it does not render all the content

Before
Screenshot 2023-04-21 at 10 17 42 AM

After
Screenshot 2023-04-21 at 10 17 56 AM

### What to review

Since the virtualized list always looks for the scroll element that is on the DocumentPanel, when it is inside it is still using the DocumentPanel to get the offset. The change that is made is to wrap the EditPortal in BoundaryElement, since that would be the closet parent it would use the scroll reference from it to set the correct top offset.

❓Are there other instances that you can think of that would have similar issues where arrays are not rendered inside DocumentPanel and have their own scroll positioning?

### Notes for release

Fixes virtualized list not rendering properly when inside a modal

### DIFF
--- a/packages/sanity/src/core/form/components/EditPortal.tsx
+++ b/packages/sanity/src/core/form/components/EditPortal.tsx
@@ -1,4 +1,4 @@
-import React, {type ReactNode, useRef} from 'react'
+import React, {type ReactNode, useState} from 'react'
 import {BoundaryElementProvider, Box, Dialog, ResponsiveWidthProps} from '@sanity/ui'
 import {PresenceOverlay} from '../../presence'
 import {PopoverDialog} from '../../components'
@@ -26,7 +26,7 @@ export function EditPortal(props: Props): React.ReactElement {
     type,
     width,
   } = props
-  const documentScrollElement = useRef<HTMLDivElement>(null)
+  const [documentScrollElement, setDocumentScrollElement] = useState<HTMLDivElement | null>(null)
 
   const contents = (
     <PresenceOverlay margins={PRESENCE_MARGINS}>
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/components/EditPortal.tsx
```diff
@@ -35,26 +36,32 @@ export function EditPortal(props: Props): React.ReactElement {
 
   if (type === 'dialog') {
     return (
-      <Dialog
+      <BoundaryElementProvider element={documentScrollElement.current}>
```
bjoerge, line 39: should this be in state rather than a mutable ref? As it is now we won't trigger a new re-render if it changes, right?

bjoerge (approved):
Great find, thank you!

packages/sanity/src/core/form/components/EditPortal.tsx
binoy14: 🤔 hmm not sure to be honest it seems to be working but I'll still change to keep it consistent with DocumentPanel implementation and possibly what you mentioned

rexxars (approved):

PR #4411, opened 2023-04-21
test(e2e): simplify auth setup, fix env var handling

rexxars:
### Description

The global setup that was responsible for setting auth cookie was a bit cumbersome and slow, as it fired up a browser before the tests ran. Additionally, the storage state was shared across all runs. The new setup (in this PR) simply injects the storage state into the config, which is much faster. I also switched to using localStorage, since the cookie approach was failing to work on Safari locally.

The only downside (as I see it), is that it now hardcodes the project ID, whereas previously it extracted it from the auth request, but I think it is an acceptable solution _for now_.

Also set reduced motion preference to minimize animations, which may or may not speed up tests at some point.

### What to review  

That the changes makes sense :)

### Notes for release

None.


bjoerge (approved):
LGTM!

binoy14 (approved):
🔥

PR #4412, opened 2023-04-21
fix: use utf8 indices for diff-match-patches

rexxars:
### Description

Note: builds on #4411 

There is a lot of context here, but the TL;DR is:

- The `diff-match-patch` JS library uses UCS-2 indices when generating patches
- This is usually not a big deal since the algorithm is lenient and allows for some differences in position
- For texts with a large amount of multibyte unicode, however, this can end up skewing the position far from where it technically is, on a byte level
- When this happens, Content Lake (which uses a different language/library for diff-match-patch) cannot find a match for the given patch, leading to an inconsistent state between the studio and the backend
- This PR changes to use the new `@sanity/diff-match-patch` library, which includes a fix for this (it now always uses UTF-8 indices) as well as some surrogate pair fixes we have identified previously.
- The `sanity-diff-patch` library has also been updated to use the new library, preventing us from having two different versions of the dependency.

### What to review

- Do lots of edits in text fields and portable text fields - preferably in multiple browsers in parallel. Ensure that things stay in sync. Use both regular text and unicode multi-bytes such as emojis, kanji and similar. Every once in a while, reload the page and check that changes does not disappear.
- The tests are a bit hacky: I've identified several things that we should try to solve/create helper methods for, but for now I'll keep it hacky until we can find the time to refactor. Initially I thought the bug had to do with partial unicode characters, and so the tests take a very heavy handed approach in terms of typing "slowly", navigating the caret in the textarea, inputting more text etc - this could probably just be replaced by `.fill()` calls for this _particular_ test, since the bug had to do with indices and not partial edits. I can do this refactor later, however.

### Notes for release

- Fixes an issue where multibyte unicode characters could lead to patches not being applied


dev/test-studio/schema/standard/texts.ts
```diff
@@ -10,6 +10,12 @@ export default defineType({
   fields: [
     {
       name: 'title',
+      type: 'string',
+      title: 'Title',
+      description: 'NOT a string field',
```
RitaDias, line 15: https://media.tenor.com/CowGNQSUsOYAAAAC/confused-math.gif

RitaDias:
👋 
While testing PTE with throttling Network to slow 3G I noticed that there's some strange behaviour with the patches. I saw this happening with writing Kanji, two chromes open and copy pasting in the same PTE

It didn't delete the content but it "reverted" it to the state before my deleting and pasting.

https://user-images.githubusercontent.com/6951139/234226852-77a8f569-576d-46cd-a947-fc9ab5d58638.mov

When I tried to delete the content one more time the content tool crashed. Refreshing the page from this will leave the PTE with the previous state before the deleting of the content
image

__

I tried replicating this on an older deployment of next and it's not happening there with the same network conditions. I had this branch open in a separate browser while testing the older deployment and it crashed there (though I can't reload the document, this error always appears unless I remove the kanji characters from PTE)
image


As for Bjørge's error ☝️ you might have fixed it already but I wasn't able to reproduce it in text fields or in PTE :)

rexxars:
> Not sure at what level this happens (have not investigated), but seeing ,true appearing in text fields when typing in the same field in two different browser windows.

Oops. Had some very basic issues that the unit tests caught - was in a bit of a rush when I pushed this. Fixed.

> It didn't delete the content but it "reverted" it to the state before my deleting and pasting.

The "Invalid patch string" has been fixed, but I'm still seeing errors in PTE. Following up outside of GitHub with @skogsmaskin for further investigation.

> I had this branch open in a separate browser while testing the older deployment and it crashed there

Do you remember/know which document ID this is?

skogsmaskin:
Got this one today while testing in the PTE in this (freshly pulled) branch:

* Mark some (normal EN-us) text
* Cut it (CMD-X)

image

rexxars:
Since I last commented, I have updated this PR through force-pushes:
- Use a new version of the diff-match-patch library which now uses UTF-8 byte lengths for stringified patches. This means that patches should be more compatible with other environments (eg content lake)
- Fixed a few incorrect usages which triggered the "Invalid array operation" and similar

There are a few places in the portable text editor that is parsing and using the patch contents to do operations such as maintaining cursor positions and similar. These need to be adjusted a bit to account for new byte lengths/offsets. I'll have a look early next week.

rexxars:
Since my last comment, I've done some work on the PTE to make it work with the new patches. I also discovered a few other issues with the way it applied patches. Some rationale for the commits, as they might not be obvious

- 89e896e - hoisting these functions makes it easier to optimize them, and prevents an additional unnecessary scope 
- 7365955 - there were two unused arguments that was passed around - simplifying the API seemed like a good idea
- 7b5fbd9 - rearranges the DMP patch application logic to return as early as possible, to avoid unnecessary nesting and extra work being performed
- 603d3c0 - slight performance gain by skipping `JSON.stringify` when debugging is not enabled. Could also use `%o` formatters or similar instead of manually stringifying.
- 39941a3 - the brunt of the change: instead of applying the operations based on the string indices, it applies the patch to the local value, then creates a new diff based on that local value - which now holds the full context - including the individual segments of unchanged/added/removed diffs. It loops through them and creates editor operations from them.

There is still the [createWithUndoRedo](https://github.com/sanity-io/sanity/blob/fix/sc-32401/kanji-dmp-fix/packages/%40sanity/portable-text-editor/src/editor/plugins/createWithUndoRedo.ts#L239-L245)

skogsmaskin (approved):
Great work! I have tested it today, and it seems to be working well as far as I can tell.

rexxars:
Last batch of tests were promising. I'm marking this as ready to review again.

bjoerge:
Unassigning myself as reviewer for now - let me know when you want me to have a look at it again.

PR #4414, opened 2023-04-24
fix(core): add max height to popover

ninaandal:
### Description
The popover for annotation links were not scrollable due to the height of it being the entire document pane height. I could not find an optimal fix for this issue and all input is highly appreciated, but setting a `max-height` to the popover will at least make it more user friendly and scrollable. 

### What to review
That the annotations popover is scrollable so that all references are accessible. 

### Notes for release
Fixes issue where annotation popovers were not scrollable in PTE.

### DIFF
--- a/dev/test-studio/schema/standard/portableText/simpleBlock.tsx
+++ b/dev/test-studio/schema/standard/portableText/simpleBlock.tsx
@@ -45,7 +45,53 @@ export default defineType({
         defineArrayMember({
           type: 'block',
           marks: {
-            annotations: [linkType, myStringType],
+            annotations: [
+              {
+                name: 'link',
+                type: 'object',
+                title: 'link',
+                fields: [
+                  {
+                    name: 'url',
+                    type: 'url',
+                    validation: (Rule) =>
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
Change LGTM 

Small Note, probably not related to this but the tooltips inside PTE are hidden behind PTE vs outside where it's user readable. See screenshot below

Screenshot 2023-04-24 at 12 13 15 PM

RitaDias:
> Visit Preview

👋 I did some testing as well, unless I'm not trying the exact approach I've not run into this 
image

In any case this feels like something outside of the scope of this PR, if you can reliably reproduce that @binoy14 do you mind creating a story for it please? :)

RitaDias (approved):
Looks good to me ✅

skogsmaskin (approved):
After looking into this with Nina today, I think this is probably the best we can do right now. The complication is that we don't know if the popover will be displaying upwards or downwards (or at least hard to make it available), making it hard to calculate the potential dynamic max-width. Nina tried this approach.

When looking at the reasons you would want a popover-editing interface in the first place, it seems uncommon to want to edit a huge document inside a popover type interface. With that in mind, I think this is an acceptable solution for this exact problem for now.

Thank you @ninaandal!

Issue #4415, opened 2023-04-24
Clearing and updating a block content field results in an "Non-unique keys" error


studiodrimmel:
**Description**
I created a block type field, while only allowing users to enter a bold text. When a user tries to clear the field with an existing, and partially bold, text and enters a new text the studio returns a "Non-unique keys" error.

**To Reproduce**
1. Create a new block content field, and only allow strong decorators:
```js
defineField({
  title,
  name,
  type: 'array',
  of: [
    {
      type: 'block',
      marks: {
        decorators: [{title: 'Strong', value: 'strong'}],
        annotations: [],
      },
      lists: [],
      styles: [],
    },
  ],
  validation: (rule) => rule.required(),
})
```

2. Add a text to the field in the studio
<img width="320" alt="Screenshot 2023-04-24 at 15 09 15 1" src="https://user-images.githubusercontent.com/101345863/234015803-84d5cde2-2653-4399-bd86-e8ca21438ba6.png">

3. Publis your change, everything should work fine. Now do select the entire text (Mac: command + A, windows: Control + A) and click the backspace button. Type in another text and the warning/error pops up. 

**Expected behavior**
Should not show the error/warning, a user should be able to enter a new text. It works fine whenever I do not have a partially bold text.

**Screenshots**
[Video of the issue](https://user-images.githubusercontent.com/101345863/234012904-5ec7c414-f4c2-4d15-ad7c-f9296f9a9105.mov)
<img width="631" alt="Screenshot 2023-04-24 at 14 54 34" src="https://user-images.githubusercontent.com/101345863/234015110-ac21c92b-a88a-45a3-8784-90d1f3805c57.png">

**Which versions of Sanity are you using?**

@sanity/cli (global)          3.9.0 (up to date)
@sanity/code-input            4.1.0 (up to date)
@sanity/eslint-config-studio  2.0.1 (up to date)
@sanity/vision                3.9.0 (up to date)
sanity                        3.9.0 (up to date)

**What operating system are you using?**

MacOS Ventura 13.0.1

**Which versions of Node.js / npm are you running?**
8.11.0
v16.15.1

**Additional context**
-

skogsmaskin:
@studiodrimmel - we released a fix for this in [v3.9.1](https://github.com/sanity-io/sanity/releases/tag/v3.9.1). Closing this.

PR #4416, opened 2023-04-24
Fix issue with PTE merge operation not producing unset patch

skogsmaskin:
### Description

There is a bug in the Portable Text Editor, where if you write only one sentence in the editor, mark half of it bold, then select all and (over)write something, it will produce two duplicate blocks as a result giving a "duplicate key error".

The reason is that `merge_node` isn't included in the list of operations that potentially can make the editor empty and should lead to a unset patch being produced.

In detail:

The follow operations are done:

1. The text is set to empty in both of the spans.
2. The mark is removed from the second span.
3. The now empty identical spans are merged into one single empty span, that equals to an empty editor.
4. The merge_node operation made the editor "empty", and a unset patch should be produced.

Since the unset patch is never produced, the insert patch is produces correctly afterwards will then duplicate the block.

Also added a test for this.

### What to review

That https://github.com/sanity-io/sanity/issues/4415 is no longer happening.

### Notes for release
* Fixed a bug in the Portable Text Input which could duplicate single sentence content with mixed marks in the editor when overwritten by another text.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/plugins/createWithPatches.ts
+++ b/packages/@sanity/portable-text-editor/src/editor/plugins/createWithPatches.ts
@@ -234,7 +234,7 @@ export function createWithPatches({
       if (
         !editorWasEmpty &&
         editorIsEmpty &&
-        ['set_node', 'remove_text', 'remove_node'].includes(operation.type)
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM, kudos for adding tests!

PR #4419, opened 2023-04-25
fix(portable-text-editor): ignore void edits inside markDefs

skogsmaskin:
### Description

There was a problem where creating and inserting a value into a text annotation object would error on a listening editor.

Values inside `markDefs` (void) objects of regular text blocks are not the editor's responsibility to deal with here. So if a nested path inside a text block is anything else than `children` in the second segment, just ignore it.

`useSyncValue` will make sure to update this content when it's changed.

### What to review

* That adding a text annotation with a value (link with href val.) will not crash an editor that have the same document and editor opened.

### Notes for release

* Fix an issue where creating a text annotation and inserting a value inside it, would make a listening text editor crash.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/utils/__tests__/patchToOperations.test.ts
+++ b/packages/@sanity/portable-text-editor/src/utils/__tests__/patchToOperations.test.ts
@@ -92,7 +92,7 @@ describe('operationToPatches', () => {
       ]
     `)
   })
-  it('will not create patches for insertion inside object blocks', () => {
+  it('will not create operations for insertion inside object blocks', () => {
     editor.children = [
       {
         _type: 'someType',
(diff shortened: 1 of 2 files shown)


RitaDias (approved):
Just to double check: I tried creating annotations with URLs, copy and pasting those texts (always with the same doc opened in a different browser window) and everything was fine. Was there any specific use case I should be looking out for?

Apart from the clarification, everything looks good :)

skogsmaskin:
> Just to double check: I tried creating annotations with URLs, copy and pasting those texts (always with the same doc opened in a different browser window) and everything was fine. Was there any specific use case I should be looking out for?
> 

No I think that should cover it @RitaDias. Thank you.

PR #4420, opened 2023-04-25
fix(core): fix scrolling issue with documents

binoy14:
### Description

Fixes an issue where the document would automatically scroll when a user is not scrolling. From investigation, the reason that happens is when the element first measured it gets stored in a cache but if the element's height changes then there is a code in react-virtual that is to fix the delta by doing a scroll.

https://github.com/TanStack/virtual/blob/623ac63988f16e6ea6755d8b2e190c123134501c/packages/virtual-core/src/index.ts#L653-L672

### What to review

To fix the issue we are making sure that the preview items in the virtualized list are not hidden when outside of the scroll window so it doesn't keep changing heights based on scrolling. There is still an issue if you have a new reference autocomplete and you get to a specific height it would do a scroll, we will probably fix that in future PR and probably investigate the library logic for a more future proofing fix.

### Notes for release

Fixes an issue with documents would automatically scroll

### DIFF
--- a/dev/test-studio/schema/debug/scrollBug.tsx
+++ b/dev/test-studio/schema/debug/scrollBug.tsx
@@ -1,0 +1,461 @@
+import {defineArrayMember, defineField, defineType, PreviewProps} from 'sanity'
+import {
+  ActivityIcon,
+  ClipboardIcon,
+  ComponentIcon,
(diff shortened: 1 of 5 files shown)


bjoerge (approved):
Great find!

PR #4425, opened 2023-04-26
fix(cli): skip v3 init prompt in unattended mode

rexxars:
### Description

When running the CLI `init` command in unattended mode (`sanity init --yes ...`), we don't want any prompt to appear, instead defaulting to whatever the default is. The prompt introduced after the v3 launch which suggests running `npm create sanity` did not respect this mode, and stopped at the prompt.

This PR makes it bypass the prompt (but show the warning) if passing `--yes`.

### What to review

`~/path/to/sanity/packages/@sanity/cli/bin/sanity init --yes --dataset=test --project=ppsg7ml5 --output-path=~/some-path` does not show prompt (requires building first (`yarn build`))

### Notes for release

- Resolves issue where unattended mode (`--yes` flag) for `sanity init` would still show a confirmation prompt about running a v2-like CLI command

### DIFF
--- a/packages/@sanity/cli/src/commands/init/initCommand.ts
+++ b/packages/@sanity/cli/src/commands/init/initCommand.ts
@@ -75,6 +75,7 @@ export const initCommand: CliCommandDefinition<InitFlags> = {
   action: async (args, context) => {
     const {output, chalk, prompt} = context
     const [type] = args.argsWithoutOptions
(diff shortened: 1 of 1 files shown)


binoy14 (approved):
LGTM!

RitaDias (approved):
🚢

PR #4426, opened 2023-04-26
fix(cli): prefer package manager that is currently executing

rexxars:
### Description

When running the CLI `init` command (and similar) through a package manager (eg `npm create sanity`, `yarn create sanity`, `pnpm create sanity`, `npx create-sanity`, `npx @sanity/cli init` etc), it seems logical to me that the user would want to install dependencies for the project using the same package manager. In other words, when running `yarn create sanity`, automatically suggest using yarn to install dependencies, instead of defaulting to npm.

I decided not to be _too_ presumptuous - we don't _skip_ the prompt, we just default to the same one that is being run. The exception to this rule is when running in unattended mode, in which case we _do_ presume that this is the right choice.

### What to review

Slightly hard to test this, _but_:
1. Create an empty `package.json` somewhere
2. Add a script that runs the CLI command, eg:
```
{
  "scripts": {
    "init-test": "~/path/to/sanity/packages/@sanity/cli/bin/sanity init"
  }
}
```
3. Run `yarn init-test`/`pnpm init-test`/`npm run init-test`
4. Once you get to the "choose package manager" step, observe that the package manager defaults to the one you ran the init-test command with

### Notes for release

- When initializing new projects, determine default package manager choice based on currently running package manager

### DIFF
--- a/packages/@sanity/cli/src/packageManager/packageManagerChoice.ts
+++ b/packages/@sanity/cli/src/packageManager/packageManagerChoice.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-process-env */
 import path from 'path'
 import which from 'which'
 import preferredPM from 'preferred-pm'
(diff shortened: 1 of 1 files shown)


RitaDias:
Hmm, looks like at least with yarn this isn't working 🤔 
Was I meant to anything between creating the `package.json` and running the command? 🤔 
image

bjoerge:
@RitaDias you might need to run `yarn build:cli` first - It worked for me after doing that.

packages/@sanity/cli/src/packageManager/packageManagerChoice.ts
```diff
@@ -143,3 +147,33 @@ function getCommandPath(cmd: string, cwd?: string): Promise<string | null> {
...
+    return 'yarn'
+  }
+
+  if (agent.includes('pnpm')) {
+    return 'pnpm'
+  }
+
+  // Both yarn and pnpm does a `npm/?` thing, thus the slightly different match here
+  // Theoretically not needed since we check for yarn/pnpm above, but in case other
+  // package managers do the same thing, we'll (hopefully) catch them here.
```
bjoerge, line 173: thank you for the thoughtful explanation

bjoerge (approved):
Smooth quality of life improvement. Thank you!

RitaDias (approved):
✅ Bjørge's suggestion worked!

PR #4431, opened 2023-04-27
test(cli): reduce dangling entity cleanup threhold to 16 hours

rexxars:
### Description

There are a few dangling entities left by the tests, mainly dataset copy jobs. To increase the likelyhood of these getting cleaned up, this PR makes sure that the dangling cleanup is _always_ run, instead of 70% of the time. In addition, the threshold for deleting old entities is lowered from 48h to 16h. To have the required precision for this, I needed to change the format from using dates (`20220131` to rounded unix timestamps (`168262061`). 

Note: This may interrupt ongoing CI jobs - but feel like a workaround for that case was not worth it.

### What to review

Cleanup logic looks sane.

### Notes for release

None.

### DIFF
--- a/packages/@sanity/cli/test/shared/globalSetup.ts
+++ b/packages/@sanity/cli/test/shared/globalSetup.ts
@@ -140,12 +140,14 @@ async function prepareDatasets() {
     const datasets = [args.documentsDataset, args.graphqlDataset, args.aclDataset]
 
     await Promise.all(
-      datasets.map((ds) =>
-        client.datasets.create(ds, {aclMode: 'public'}).catch((err) => {
+      datasets.map((ds) => {
+        // eslint-disable-next-line no-console
+        console.log(`Creating dataset ${ds}...`)
+        return client.datasets.create(ds, {aclMode: 'public'}).catch((err) => {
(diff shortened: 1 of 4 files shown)


binoy14 (approved):
Makes sense to me

rexxars:
Seems the setup is run in a different process, so the test ID doesn't remain consistent since it gets a new timestamp. Will have to rethink this 🤔

rexxars:
Ready to re-review - by setting an environment variable, we're able to share the timestamp across workers/threads

packages/@sanity/cli/test/shared/globalSetup.ts
```diff
@@ -140,12 +140,14 @@ async function prepareDatasets() {
     const datasets = [args.documentsDataset, args.graphqlDataset, args.aclDataset]
 
     await Promise.all(
-      datasets.map((ds) =>
-        client.datasets.create(ds, {aclMode: 'public'}).catch((err) => {
+      datasets.map((ds) => {
+        // eslint-disable-next-line no-console
+        console.log(`Creating dataset ${ds}...`)
```
RitaDias, line 145: Do we want to leave the console log in this case?

PR #4436, opened 2023-04-28
feat: document list search

hermanwikner:
## Description

This PR introduces search to all desk document lists along with a number of QoL improvements.

- Search bars have been added to all document lists
- Improved desk list accessibility: it's now possible to tab between panes and use the up / down arrow keys to navigate list items. It's also possible to tab to pane headers and expand / collapse them with the keyboard alone.
- Improved the perceived performance of desk pane lists: we no longer show a separate loading spinner and opt for loading skeletons

image

## What to review

- Searching within document lists should work as expected. Text queries within custom document lists should have their searches constrained to lists' custom GROQ filters.
- You should be able to tab between document lists and headers. Note that by design, you cannot collapse the last pane header in the desk list, as such it will not be focusable.
- `focus-visible` outlines within list panes should only appear on keyboard interaction
- To generate a search query, we need to know the document types present in a given list. In certain cases, we can identify these types by examining the `filter` and `params`. For example, if a filter reads `_type == "author" ` (a "simple filter"), we know that the search query should look for author documents. However, a filter can be highly customized and take a form like `_id in path("drafts.**")`, which lists all draft documents. In such cases, we can't determine the relevant document types to search for. **In these cases we make an additional request** with the filters and any parameters to get the relevant document types, after which we can create a search query.
- The document list query is too large to be used as a listening query, so this implementation sets up a listener that is notified when mutations occur that matches the filter and any parameters. When a mutation is received, the document list query is refetched.
- Search in document lists currently respects paths defined by `__experimental_search`, but NOT individual path weighting. We don't apply weighted search in document lists as it's technically considered its own sort order ('Relevance').
  - There’s a possible workaround here in future, where we use weighted search and only apply ‘Relevance’ sorting when viewing search results, but otherwise default to the document lists ‘native’ order when the search query is empty.

### Improvements to desk document lists
It's now possible to search within document lists. We've also improved keyboard accessibility, making it much easier to navigate between lists and the document form editor itself.


bjoerge (changes requested):
Great work!

I pushed a suggestion on how to improve the document list query here: https://github.com/sanity-io/sanity/commit/c3983111d423ed2c29837b7cd2461953fe3e34c0

During testing I also found that the error handling/retry logic isn't working (try e.g. blocking the listener request in devtools, and you'll see that the document list loads forever)

packages/sanity/src/desk/panes/documentList/useDocumentList.ts
```diff
@@ -1,111 +1,211 @@
...
+
+  // Get the type name from the filter, if it is a simple type filter.
+  const typeNameFromFilter = useMemo(
+    () => getTypeNameFromSingleTypeFilter(filter, paramsProp),
+    [filter, paramsProp]
+  )
 
-    return `*[${filter}]|order(${order})[0...${limit}]{${finalProjection}}`
-  }, [filter, fullList, sortOrder])
+  const isLoading =
```
robinpyon, line 78: Really minor – a clarifying comment here would be great

i.e. why is `isLoading = true` when `Boolean(loading && result === null && !error)`? Logically I'd assume that `isLoading` would be true even when `loading` is true, even with a `null` result. Perhaps I'm getting thrown off by the variable name itself.

Additionally – is `isInitialLoading` still being used?

robinpyon (approved):
Really great work @hermanwikner! Will defer to @bjoerge on `listenSearchQuery` (though we also spoke about this offline).

mikolajdobrucki (approved):
Looks amazing! So much work that went into. I'm so excited we're finally making it! 🔥🔥🔥

PR #4440, opened 2023-04-28
fix(form): prevent autofocus on array dialogs where focus is on a deeper value

bjoerge:
### Description

This will skip autofocus on first interactive descendant when opening dialogs to edit array objects unless focus is on  the object itself. Prevents an issue setting focus on the close button even if focus path mandated focus for a node deeper down in the hierarchy.

### What to review
Open %2Cpath%3DsomeArray%5B_key%3D%3D"01d8a490d0d1"%5D.focusTest.someObject.first and observe that the right field inside the dialog gets focused
Here's `next` for comparison: 

### Notes for release

- Fixes an issue causing focus to be set on wrong element when  deep linking to a field inside an array


bjoerge:
Superseded by #4443

PR #4441, opened 2023-04-28
`findability-4` focus handling updates

robinpyon:
## Description

This PR introduces the following changes to `findability-4`

### CommandList

- `canReceiveFocus?: boolean` – whether the virtual list container can receive focus
- `onlyShowSelectionWhenActive?: boolean` - only show child selected states when the command list either has focus or is hovered
- `focusRingOffset?: number` - Pixel offset of the virtual list focus ring. Negative values will cause the focus ring to appear inset (default is `0`)

### Other notable changes
- After collectively deciding that the additional logic and potential for unnecessary re-renders wasn't worth it, we no longer wrap desk-tool panes with `<InputTypeProvider>` – this has been removed
- As a result of the above, desk panes will no longer autofocus. We now aim for a much simpler, less opinionated approach that doesn't cause new list panes to steal focus. 
- Regular pane lists will always default to index `0` when no selected items are found, keeping in-line with document pane list behaviour (previously these would initialize with an initialIndex of `-1`)
- In command lists, we no longer store the previous `onEndReached` value as this would lead to edge cases where you wouldn't be able to load the next page if you were within the threshold _after_ the next page had loaded.
- Some minor refactoring: regular list panes now break out into separate header and content components (to match how document list panes are structured)
- Search queries in document lists will now append findability-specific comments
- A minor fix to ensure header titles in _both_ document and regular lists now conditionally accept focus, based on whether they're the last visible pane and collapsed state

## What to review
- Desk pane focus ring should render as expected (using back / forward browser buttons shouldn't trigger it)
- `focus-visible` events on desk panes should appear when expected (via keyboard interaction only)
- List pane headers should conditionally receive focus based on their collapsed and `isLast` state


packages/sanity/src/desk/panes/list/ListPaneContent.tsx
```diff
@@ -0,0 +1,130 @@
...
+        />
+      )
+    },
+    [childItemId, isActive, layout, shouldShowIconForItem]
+  )
+
+  return (
+    <PaneContent overflow={layoutCollapsed ? undefined : 'auto'}>
+      {items && items.length > 0 && (
+        <CommandList
```
robinpyon, line 112: This should contain `wrapAround=false`

packages/sanity/src/core/components/commandList/CommandList.tsx
```diff
@@ -539,25 +573,24 @@ export const CommandList = forwardRef<CommandListHandle, CommandListProps>(funct
...
-  // If the input element is present, we want to ensure that the virtual list element is not focusable.
-  // This is to prevent the virtual list from being focused when the input element is focused.
-  const rootTabIndex = inputElement ? -1 : tabIndex
+  const rootTabIndex = canReceiveFocus ? 0 : -1
 
   return (
     <VirtualListBox
-      data-focus-visible={focusVisible}
       id={getCommandListChildrenId()}
+      onMouseEnter={handleVirtualListMouseEnter}
```
hermanwikner, line 585: I wonder if these mouse events should be added to the `VirtualListChildBox` instead to avoid this from happening? 

https://user-images.githubusercontent.com/15094168/235634281-c1cd87ee-88a5-423d-8807-ecaad415a373.mov

hermanwikner (approved):

PR #4451, opened 2023-05-03
fix(form): improve focus handling for image and file inputs

bjoerge:
### Description
This includes a couple of improvements for image and file inputs with regards to focus handling:

  - Makes sure image and file fields support deep linking + scroll to through focus path (fixes deep linking + validation)
  - Fixes a bug breaking the ability to paste a file into a file field that has a value (we currently had two focusable elements in a file field that had a value (the preview wrapper + the FileTarget)

### What to review
Make sure deep linking (either via intent link w/path, from presence menu or validation menu) works
- Deep linking to file field from intent link: https://test-studio-git-fix-improve-asset-inputs-focus.sanity.build/test/intent/edit/id=66890ca9-0354-4a16-9ec0-e25ff5679295;path=FileWithTwoAccept/
- Deep linking to image field from intent link: https://test-studio-git-fix-improve-asset-inputs-focus.sanity.build/test/intent/edit/id=6a2f0a14-0b52-4022-9194-647eb3fc1ee1;path=noAssetSource/

### Notes for release
- Makes sure image and file fields support deep linking
- Fixes a bug that broke the ability to paste a file into a file field that already has got a value

### DIFF
--- a/packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx
+++ b/packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx
@@ -39,7 +39,6 @@ import {PatchEvent, setIfMissing, unset} from '../../../patch'
 import {MemberField, MemberFieldError, MemberFieldSet} from '../../../members'
 import {ImperativeToast} from '../../../../components'
 import {ChangeIndicator} from '../../../../changeIndicators'
-import {FIXME} from '../../../../FIXME'
 import {CardOverlay, FlexContainer} from './styles'
 import {FileActionsMenu} from './FileActionsMenu'
 import {FileSkeleton} from './FileSkeleton'
@@ -72,15 +71,10 @@ export interface BaseFileInputState {
   isMenuOpen: boolean
 }
 
-/** @internal */
-export type Focusable = {
-  focus: () => void
-}
 const ASSET_FIELD_PATH = ['asset']
 
 /** @internal */
 export class BaseFileInput extends React.PureComponent<BaseFileInputProps, BaseFileInputState> {
-  _focusRef: Focusable | null = null
   _assetFieldPath: Path
   uploadSubscription: Subscription | null = null
   browseButtonElement: HTMLButtonElement | null = null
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx
```diff
@@ -303,17 +297,16 @@ export class BaseFileInput extends React.PureComponent<BaseFileInputProps, BaseF
...
-  hasFileTargetFocus() {
-    return this.props.focusPath?.[0] === 'asset'
-  }
-
-  handleFileTargetFocus = (event: FIXME) => {
+  handleFileTargetFocus = (event: React.FocusEvent) => {
     // We want to handle focus when the file target element *itself* receives
     // focus, not when an interactive child element receives focus. Since React has decided
     // to let focus bubble, so this workaround is needed
     // Background: https://github.com/facebook/react/issues/6410#issuecomment-671915381
```
RitaDias, line 304: 👋 does this comment still hold given the changes?

packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx
```diff
@@ -352,12 +345,14 @@ export class BaseFileInput extends React.PureComponent<BaseFileInputProps, BaseF
           isChanged={changed}
         >
           {/* not uploading */}
-          {!value?._upload && (
+          {value?._upload ? (
+            this.renderUploadState(value._upload)
+          ) : (
             <FileTarget
               {...elementProps}
+              onFocus={this.handleFileTargetFocus}
```
bjoerge, line 353: We need this special handling in order to prevent interactive elements _inside_ the file target from triggering focus on the file input itself (see implementation of `this.handleFileTargetFocus`)

packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx
bjoerge: It does! We still need this check to catch focus events triggered by interactive child elements

bjoerge:
>Can I have some partial links to know that I am testing the deep linking properly

@RitaDias Updated OP

RitaDias (commented):
Great work! 
I think the only thing I noticed which can be odd is, especially on the second link (and opening on a new tab) which doesn't seem to happen on the first one. My first assumption is that it's probably related to the size of the input before and the fact that it's the last one?

https://user-images.githubusercontent.com/6951139/235877352-46f2935a-b6f5-4c10-b588-5c9d2b780a36.mov

This being said when you refresh the link it goes fully down (thought not necessarily when you copy and paste the link to a new window where the same error is visible)

bjoerge:
Yeah, that one is weird. I suspect there's something interfering with scrolling at some level, but don't think it's related to these changes, so we can look into that separately. The important thing is that the right element gets focus.

binoy14 (approved):

PR #4456, opened 2023-05-04
fix(core): prevent number primitive fields from triggering change callbacks when numbers are out of range

robinpyon:
## Description

This PR updates `PrimitiveField` and prevents `onChange` callbacks when number values exceeds safe ranges in JS (`Number.MAX_SAFE_INTEGER` and `Number.MIN_SAFE_INTEGER` respectively)

Currently, entering an integer outside the ranges in a `number` field will crash the studio.

## What to review

- Entering large or small values outside this range in `number` fields will no longer cause its `onChange` callbacks to trigger – meaning the input won't update with the out of range value. In these cases, we don't store an intermediate value or show any sort of validation error.
- There's existing precedence for this in the studio – [primitives within arrays do this same check](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/members/array/items/ArrayOfPrimitivesItem.tsx#L68-L70), so this is just replicating that behaviour for number fields.

## Notes for release

Fixes an issue where out of range integers in number fields could crash the studio

### DIFF
--- a/packages/sanity/src/core/form/members/object/fields/PrimitiveField.test.tsx
+++ b/packages/sanity/src/core/form/members/object/fields/PrimitiveField.test.tsx
@@ -173,6 +173,28 @@ describe('PrimitiveField', () => {
       expect(input).toBeInstanceOf(HTMLInputElement)
       expect(input.value).toEqual('1.00')
     })
+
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
Great catch, thank you!

ninaandal (approved):

PR #4457, opened 2023-05-04
fix(form): set focus according to focus path on inputs inside block objects

bjoerge:
### Description

Currently, if you follow a deep link to a block object field (opened in a dialog) it doesn't put focus on it, but rather sets focus on the first field inside the dialog ([repro](https://test-studio.sanity.build/test/intent/edit/id=97a43753-a8ab-46eb-81e1-33067bedf2bb;path=body%5B_key%3D%3D%22b016cc30e71e%22%5D.otherstring/) - as you can see from the url: the path points at `otherstring`)

This PR removes the logic that placed focus on first field with the following logic:

If the block itself has focus (i.e. not a child of the block), we use the default behavior of Sanity UI Dialogs (which is to place focus at the first interactive element inside the dialog, e.g. the close button if the dialog has one). If a child of the block has focus (according to the current focus path), Sanity UI dialog autofocus behavior is disabled in favor of the standard focus handling of inputs/field wrappers.

### What to review
- Make sure deep linking to fields and inputs inside blocks in portable text editor works as expected
- Here's the above mentioned url with this PR applied: https://test-studio-git-bug-sc-32925fix-dialog-autofocus.sanity.build/test/intent/edit/id=97a43753-a8ab-46eb-81e1-33067bedf2bb;path=body%5B_key%3D%3D%22b016cc30e71e%22%5D.otherstring/

### Notes for release
- Fixes an issue with focus placement when deep linking to fields contained in Portable Text blocks

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
@@ -161,6 +161,7 @@ export function BlockObject(props: BlockObjectProps) {
       children: inputRef.current,
       focused,
       markers,
+      nodeFocused: memberItem?.node.focused,
       onClose,
       onOpen,
       onPathFocus,
@@ -183,6 +184,7 @@ export function BlockObject(props: BlockObjectProps) {
       isOpen,
       markers,
       memberItem?.elementRef,
+      memberItem?.node.focused,
       memberItem?.node.path,
       onClose,
       onOpen,
@@ -283,8 +285,8 @@ export const DefaultBlockObjectComponent = (props: BlockProps) => {
     onClose,
     onOpen,
     onRemove,
+    nodeFocused,
     open,
-    path,
     readOnly,
     renderPreview,
     selected,
@@ -292,6 +294,7 @@ export const DefaultBlockObjectComponent = (props: BlockProps) => {
     value,
     validation,
   } = props
+
   const isImagePreview = schemaType.name === 'image'
   const hasError = validation.filter((v) => v.level === 'error').length > 0
   const hasWarning = validation.filter((v) => v.level === 'warning').length > 0
@@ -326,7 +329,7 @@ export const DefaultBlockObjectComponent = (props: BlockProps) => {
         {renderPreview({
           actions: (
             <BlockObjectActionsMenu
-              focused={focused}
+              blockFocused={focused}
               isOpen={open}
               onOpen={onOpen}
               onRemove={onRemove}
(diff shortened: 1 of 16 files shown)


dev/test-studio/schema/standard/portableText/simpleBlock.tsx
```diff
@@ -24,7 +24,10 @@ const linkType = defineArrayMember({
 const myStringType = defineArrayMember({
   type: 'object',
   name: 'test',
-  fields: [{type: 'string', name: 'mystring', validation: (Rule) => Rule.required()}],
+  fields: [
+    {type: 'string', name: 'mystring', validation: (Rule) => Rule.required()},
+    {type: 'string', name: 'otherstring', validation: (Rule) => Rule.required()},
```
bjoerge, line 29: adding a second field here provides a repro for the issue

packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
```diff
@@ -151,7 +151,7 @@ export function BlockObject(props: BlockObjectProps) {
       __unstable_boundaryElement: boundaryElement || undefined,
       __unstable_referenceElement: memberItem?.elementRef?.current || undefined,
       children: input,
-      focused,
+      focused: memberItem?.node.focused,
```
bjoerge, line 154: I'm not 100% sure about this one, the previous `focused` prop seems to come from the Portable Text editor and is based on selection. Would love to have your eyes on this @skogsmaskin.

packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
```diff
@@ -319,7 +318,7 @@ export const DefaultBlockObjectComponent = (props: BlockProps) => {
         {renderPreview({
           actions: (
             <BlockObjectActionsMenu
-              focused={focused}
+              blockFocused={focused}
```
bjoerge, line 321: Renamed this prop to make it clearer that it refers to whether the surrounding _block_ is having focus (can revert if it has unintended consequences)

stipsan (approved):
Code looks good 👍 I can help test this with Visual Editing if you hand me a `canary` @bjoerge 💖

bjoerge:
@stipsan released a canary @`3.10.1-canary.1+a4b0c68abe`

skogsmaskin:
Should probably merge in this first? https://github.com/sanity-io/sanity/pull/4466

skogsmaskin (changes requested):
Will be great to have this in!

However I think we should leave the "focused" prop here as is because it's something different than the object (formstate) focus. All of these object components take `focused` prop meaning "this node is currently the focused point in the editor selection".

skogsmaskin:
Closing this one in favour to https://github.com/sanity-io/sanity/pull/4524 which includes these fixes.

PR #4458, opened 2023-05-04
fix: avoid scroll on `overflow: hidden`

stipsan:
### Description

When `overflow: hidden` browsers turn off scrollbars and the user can no longer scroll. But it can still be scrolled if the APIs `.focus()` and `.scrollIntoView`, as well as with other native browser behaviour as browsers still consider them scrolling boxes.

This video demonstrates how `overflow: hidden` is being scrolled into breaking the Studio layout when opening an edit intent link on iOS Safari that ends up in a field inside a dialog. 

https://user-images.githubusercontent.com/81981/236252831-515d3752-6957-465d-9e25-84b7544ace06.mp4

What makes matters worse is that since `overflow: hidden` disables user scroll the user can't recover from this broken state in a lot of cases. And there are many other ways it can break the layout and UI.

The new `overflow: clip` fixes this and turns of all scrolling. By adding it after `overflow: hidden` it's applied in browsers that support it, while those who don't continue to use `overflow: hidden`.

~~As [`@sanity/ui`](https://github.com/sanity-io/ui/pull/1098) also need a fix this PR is in draft status. Once `@sanity/ui` is published with the fix I'll update this PR and move it out of draft.~~

It's out! PR ready ☺️ 

### What to review

That what happens in the video no longer happens. No matter what triggers the soft keyboard on iOS and Android in dialogs or other initially hidden, or below-the-fold elements, shouldn't be able to break the layout.

### Notes for release

fix: edit intent links that result in iOS softkeyboards to open while opening dialogs no longer breaks the layout.

### DIFF
--- a/dev/design-studio/package.json
+++ b/dev/design-studio/package.json
@@ -33,7 +33,7 @@
     "@sanity/color": "^2.1.20",
     "@sanity/icons": "^2.3.0",
     "@sanity/logos": "^2.0.2",
-    "@sanity/ui": "^1.3.0",
+    "@sanity/ui": "^1.3.3",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "sanity": "3.10.1",
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -24,7 +24,7 @@
     "@sanity/image-url": "^1.0.2",
     "@sanity/portable-text-editor": "3.10.1",
     "@sanity/types": "3.10.1",
-    "@sanity/ui": "^1.3.0",
+    "@sanity/ui": "^1.3.3",
     "@sanity/ui-workshop": "^1.0.0",
     "@sanity/util": "3.10.1",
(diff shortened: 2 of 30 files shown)


RitaDias:
👋 
Thanks for setting this up Cody 🙏 

I was testing it on my android (Samsung) and (I might be testing this wrong) I think I still can see it happening there?

https://user-images.githubusercontent.com/6951139/236410602-db380884-2174-4088-8ac7-4f691b488ffe.mp4

bjoerge (approved):
Thanks a million @stipsan! 💖

RitaDias (approved):
Just saw the update on slack ✅

mariuslundgard (approved):
I haven't tested all edge cases, but this should be totally good in theory, so 💚 from me 😊

skogsmaskin (approved):
Tested it on my Android phone with Chrome and it seems to be working well.

PR #4461, opened 2023-05-06
Export `extract` from `@sanity/mutator`

nkgentile:
- feat(mutator): export `extract`
- docs(mutator): adding inline docs for `extract`


nkgentile:
In userland, I'm having to do, since I don't need the paths:
```ts
const extract = (...args: Parameters<typeof extractWithPath>) =>
  extractWithPath(...args).map(({value}) => value);

const productGids = extract(`..[_type == "productWithVariant"].gid`, page);
```

packages/@sanity/mutator/src/jsonpath/extract.ts
```diff
@@ -1,5 +1,13 @@
 import {extractAccessors} from './extractAccessors'
 
+/**
+ * Extracts a value for the given JsonPath
+ *
+ * @param path - Path to extract
+ * @param value - Value to extract from
+ * @returns An array of objects from the given path
+ * @internal
```
rexxars, line 9: Since the method is exported, it needs to be marked as either `public` or `beta` - this one has remained unchanged for a long time, so I think we're good with marking this as public.
```suggestion
 * @public
```

packages/@sanity/mutator/src/jsonpath/extract.ts
```diff
@@ -1,5 +1,13 @@
 import {extractAccessors} from './extractAccessors'
 
+/**
+ * Extracts a value for the given JsonPath
```
rexxars, line 4:
```suggestion
 * Extracts values matching the given JsonPath
```

packages/@sanity/mutator/src/jsonpath/extract.ts
```diff
@@ -1,5 +1,13 @@
...
+ *
+ * @param path - Path to extract
+ * @param value - Value to extract from
+ * @returns An array of objects from the given path
```
rexxars, line 8:
```suggestion
 * @returns An array of values matching the given path
```

rexxars (changes requested):

RitaDias (approved):
I think there might be something wrong with the tests config wise but overall this looks good to me

rexxars (approved):

PR #4466, opened 2023-05-09
wip

skogsmaskin:


RitaDias:
I might be testing this wrong but clicking the inputs is not updating the links (and so not linking me to the right place?) Or are we meant to be testing this with an already built link that will focus the path?

https://github.com/sanity-io/sanity/assets/6951139/fef117f2-f741-4d89-8779-f3aa6755c197

skogsmaskin:
> I might be testing this wrong but clicking the inputs is not updating the links (and so not linking me to the right place?) Or are we meant to be testing this with an already built link that will focus the path?
> 

Sorry for unclear instructions Rita. The location bar href will not be updating on it self no.

If you run two studios (preferably two different users, or sessions) you can with one user stand on the input and then with the other click user click on the avatar in the presence top menu. You should then be able to go to the single span the other user is currently placing their cursor.

RitaDias:
👋 @skogsmaskin thank you for clarifying!!

While testing I came across two odd cases that I expected to work in this case:

1. Writing content from user 1, user 2 using the presence and landing on the same PTE and span where user 1 selected the text

https://github.com/sanity-io/sanity/assets/6951139/902a5ea3-9fd6-4020-81f5-3fe5905e683c

2. When user 1 is pressing (**not** in the modal) a block in a PTE, I expected user 2 when using the presence to travel to the doc to land on the PTE not **inside** the modal

https://github.com/sanity-io/sanity/assets/6951139/ec6ebe6f-a106-41da-b90e-5a6debd6c975

skogsmaskin:
@RitaDias - yeah, I see that too. I would also say it's kind of unexpected.

skogsmaskin:
@RitaDias - yeah, I see that too. I would also say it's unexpected. Will look into it.

RitaDias:
> @RitaDias - can you test again please? I have fixed the issue where it doesn't focus back to the object when you close t he modal.

@skogsmaskin 👋 
That one is fixed but the other two use cases are still happening
- user 2 hovering over plain text on PTE and user 1 not being scrolled into the right field

https://github.com/sanity-io/sanity/assets/6951139/54c5e0f0-0230-4b06-9c30-78fea9a33aa8


- user 2 ending up on the open the modal when user 1 didn't open the modal (just had the block pressed, visible on video 2)

RitaDias (approved):
Nice work! Let's go! :shipit:

skogsmaskin:
Just cleaned up the commits a bit before merge.

PR #4467, opened 2023-05-09
fix(portable-text-editor): fix sync issue with text spans

skogsmaskin:
### Description

When a span text is changed, it can't be updated with setNodes only. Use insertText to update the actual text. This is a result of how the Slate transformations are working. Do this by removing and inserting text into the node when it is a text span.

Added a test for this as well.

### What to review

That sending patches to a document that are replacing span texts are being properly updated, or by simulating the same through a new  props.value where a single span text is changed.

### Notes for release
* Fixed an issue with the Portable Text Editor where single span texts could become out of sync with the server value.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.test.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.test.tsx
@@ -1,0 +1,130 @@
+/**
+ * \@jest-environment ./test/setup/jsdom.jest.env.ts
+ */
+/* eslint-disable no-irregular-whitespace */
+// eslint-disable-next-line import/no-unassigned-import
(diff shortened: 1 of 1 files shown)


rexxars:
Seems to have broken the "will let B stay on same line when A inserts a new line above" test, but does seem to fix the bug 👍

skogsmaskin:
> Seems to have broken the "will let B stay on same line when A inserts a new line above" test, but does seem to fix the bug 👍

Yeah, looks like we must delete the text first. Should be fixed now.

skogsmaskin:
Force pushed some changes tonight.

skogsmaskin:
Added a explicit test for this as well since we didn't seem to pick up on it.

bjoerge (approved):
LGTM!

PR #4469, opened 2023-05-09
fix(desk): remove path param from url after initially read

bjoerge:
### Description

When following a intent link or a link to a field/input at a specific path in the Studio (e.g from presence menu or through Visual Editing), the path remains in the url. This means that if you reload the page, you are taken (likely surprisingly) back to the path you arrived at initially. This PR fixes that by clearing the path parameter from the URL after it's been set as the initial focus path.

#### Old behavior:
- Go to https://test-studio.sanity.build/test/content/input-standard;objectsTest;cc5c90bd-1184-44ae-9a94-d3395ac7ebec%2Cpath%3Drecursive.recursive.events%255B_key%253D%253D%25220123f5fdca78%2522%255D.where
- Move around in the document, make some edits
- Reload page
- You're back to the field you arrived at earlier

#### New behavior:
- Go to https://test-studio-git-fix-remove-initial-path-param.sanity.build/test/content/input-standard;objectsTest;cc5c90bd-1184-44ae-9a94-d3395ac7ebec%2Cpath%3Drecursive.recursive.events%255B_key%253D%253D%25220123f5fdca78%2522%255D.where
- Move around in the document, make some edits
- Reload page
- It behaves as if you opened the document as usual

### Notes for release

- When following a link to a document field, the field path is now cleared from the url after initial load

### DIFF
--- a/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
@@ -451,12 +451,15 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   // Reset `focusPath` when `documentId` or `params.path` changes
   useEffect(() => {
     if (ready && params.path) {
-      const pathFromUrl = resolveKeyedPath(formStateRef.current?.value, pathFromString(params.path))
+      const {path, ...restParams} = params
+      const pathFromUrl = resolveKeyedPath(formStateRef.current?.value, pathFromString(path))
       // Reset focus path when url params path changes
       setFocusPath(pathFromUrl)
       setOpenPath(pathFromUrl)
+      // remove the `path`-param from url after we have consumed it as the initial focus path
+      paneRouter.setParams(restParams)
     }
-  }, [params.path, documentId, setOpenPath, ready])
+  }, [params, documentId, setOpenPath, ready, paneRouter])
 
   return (
     <DocumentPaneContext.Provider value={documentPane}>{children}</DocumentPaneContext.Provider>


RitaDias (commented):
While testing out I did come across a use case that would be my concern about this change:
1. Open the new link
2. Stay on the modal but **don't** anything
3. Refresh
4. The document opens, but not on the modal which where I might have expected to be since I didn't close it

I think this might also cause the same level of confusion, couldn't we just keep whatever the old "path" was? (and in that case it would preserve the "outside" the modal or inside the modal depending on where the user went) 🤔  

https://github.com/sanity-io/sanity/assets/6951139/0a0b4d91-ffda-4ef0-b948-e6afe6352f1e

bjoerge:
Thanks @RitaDias - I think it's a fair point. I'm not sure what users would normally expect here. I think you can also make the argument that the current behavior is inconsistent in the way that dialog state is not normally something we preserve in the URL, so if you follow a normal click sequence to open the dialog, then a reload will not keep you in the dialog.

Would love to hear @mikolajdobrucki's take here :)

mikolajdobrucki (approved):
I would agree with Rita in a way that ideally, after Refresh you should stay in the same place on the page as you were – no matter if you performed any action on the page since the initial load or not. So if you didn't do anything, after a Reload you should still be where you initially landed. If you closed a modal, scrolled somewhere else, after a Reload you also should stay there, in the new place you reached. 

Unfortunately, neither is true In the Studio. Reload makes the Studio forget where you were and always move you to the top of the page (always apart from the case this PR is addressing ofc).

So in fact, neither of the behaviours is optimal. But if I had to choose between these two I would say that this PR is a welcomed fix as it keeps things more as they were (not great, but at least consistent).

RitaDias (approved):
✅

RitaDias (approved):
✅

PR #4473, opened 2023-05-10
Fix issue with PTE-undo/redo shared history

skogsmaskin:
### Description

There is an issue with the undo/redo functions for the Portable Text Editor that leads to a invalid undo/redo history for users that are within the same document at the same time.

We should not be considering the `isPatching` state here in the undo/redo plugin, as the code that produces the unwanted operations are properly  scoped with  the `withoutSaving` scope elsewhere in the code. This would exclude them from the history, as the intention of this removed return originally was - and is the way it should be.

By returning this state there, the `isSaving` test could get stuck in a state where we are returning the previous value from `isSaving`. This is what happened when we had actual remote patching going on causing local operations to be missed.

### What to review

That if you are two users writing on the same document, your undo/redo history will work acceptably.

### Notes for release
* Fixed issue with shared undo/redo history for the Portable Text Input


skogsmaskin:
This will probably be part of a new incoming PR from the `fix/sc-32401/kanji-dmp-fix` branch. But leaving it as draft for now.

Issue #4432, opened 2023-04-27
Field group is hidden if all fields are inside fieldsets


rylanharper:
**Describe the bug**

I upgrade to the latest version of Sanity (`3.9.1`) and notice that the first [Field Group](https://www.sanity.io/docs/field-groups) object no longer displays the corresponding group data or tab within the Sanity Studio.

**To Reproduce**

I have a simple a set of field groups:
```
// Current field groups
groups: [
  {
    title: 'Content',
    name: 'content',
    default: true
  },
  {
    title: 'Upsells',
    name: 'upsells'
  }
],

// Schema fields
fields: [
    {
      title: 'Block Title',
      name: 'blockTitle',
      type: 'string',
      group: 'content'
    },
    {
      title: 'Upsell Image',
      name: 'upsellImage',
      type: 'image',
      description: 'An upsell image showcasing a product or collection',
      group: 'upsells'
    }
  ],
```

The first group `content` no longer displays its tab and data within the array. You can also get rid of the second group set (`upsells`) and it will still not work.

**Expected behavior**

I should see the following field groups at the top of the document: `All fields`, `Content`, `Upsells`. Instead I only see `All fields`, `Upsells` and the first group data no longer displays. You can see this visually in the screenshot below ⬇️

**Screenshots**

<img width="375" alt="Screenshot 2023-04-27 at 12 10 35 PM" src="https://user-images.githubusercontent.com/16216160/234971186-a76b002e-ddc8-42e0-9c48-6229c921cc2a.png">

**Which versions of Sanity are you using?**

Sanity `3.9.1`

thierryteyssier:
Indeed I have the same issue.
And if I have fields without an assigned group, these fields are visible on all groups and not only in "All fields".

rylanharper:
Yes, this is still present in the latest `3.10.0` as well.

thobas-dnvgl:
Ping @bjoerge because you've updated the groups feature in 3.9.1.

sebmor:
+1 Fields without groups are now visible for all groups.

thobas-dnvgl:
On 3.10.1 I still have the issue (first group declared missing from the UI). Let me try to have a reduced test case.

bjoerge:
Thank you @thobas-dnvgl! - I'll take a look. Let's track https://github.com/sanity-io/sanity/issues/4471 separately.

rylanharper:
[...]

Issue #4474, opened 2023-05-10
Fields assigned both a group and fieldset do not display correctly after version 3.9.0


LinusStern:
I want to report a bug with the sanity interface present in all versions after v3.9.0.

When a field is given both a group and a fieldset in v3.9.1, the group this field is part of is no longer rendered at the top, and in v3.10.0+ the group is both not rendered and the field is shown even when another group is selected.

For context: In the images below I've got a style badge group which contains an icon and text field. These two fields are also in a 2-column fieldset.

![v3-9-0](https://github.com/sanity-io/sanity/assets/35244606/5f8f3c63-43e3-4726-9398-181c4cba6d06)
![v3-9-1](https://github.com/sanity-io/sanity/assets/35244606/45585c66-8d7e-47d3-9c4b-da73d604c53f)
![v3-10-0](https://github.com/sanity-io/sanity/assets/35244606/8eb019c9-8c97-4426-95c5-a3557bb1f29d)

bjoerge:
Fixed in v3.11.0

PR #4477, opened 2023-05-11
refactor(form): fix issue with hiding a field group that only have fieldsets

bjoerge:
### Description

As described in #4432, when a field group only has fieldsets, the group becomes hidden.
 
#### Before
When a field group has fieldsets only, it disappears. Reproducible here: https://test-studio-cbwgkvfea.sanity.build/test/content/input-debug;field-groups;fieldGroupsWithFieldsets;96fcdb05-ed51-41b7-a189-94eae745eae2%2Ctemplate%3DfieldGroupsWithFieldsets

#### After
The same case as above with the fix from this PR applied: https://test-studio-git-fix-fieldset-field-group-bug.sanity.build/test/content/input-debug;field-groups;fieldGroupsWithFieldsets;96fcdb05-ed51-41b7-a189-94eae745eae2

In particular, the following cases should hold true after these changes:
- Fields inside fieldsets should be filtered based on the currently active field group
- A field group should be selectable even if it consists entirely of fieldsets
- If a fieldset has no fields in the active group, it should be hidden
- If all fields (both in fields and fieldsets) in the active group are hidden, the field group should be hidden
- Validation and conditional fields should work as usual

### Notes for release

- Fixes an issue causing field groups with only fieldsets to be hidden

### DIFF
--- a/dev/test-studio/schema/debug/fieldGroupsWithFieldsets.js
+++ b/dev/test-studio/schema/debug/fieldGroupsWithFieldsets.js
@@ -1,0 +1,54 @@
+import {CogIcon} from '@sanity/icons'
+
+export default {
+  name: 'fieldGroupsWithFieldsets',
+  title: 'Field groups with all fields inside fieldsets',
+  type: 'document',
+  groups: [
+    {
+      name: 'group1',
+      title: 'Group 1',
+      icon: CogIcon,
+    },
+    {
+      name: 'group2',
+      title: 'Group 2',
+    },
+  ],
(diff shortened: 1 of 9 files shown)


ninaandal (approved):
Good job with all the fieldset/fieldgroup work 😎 
From what I can tell, this PR also fixes [this issue](https://github.com/sanity-io/sanity/issues/4474) 🙌

robinpyon (approved):
Thanks @bjoerge!

PR #4486, opened 2023-05-12
Add boundary provider for nested content inside PT-input dialogs

skogsmaskin:
### Description

Wrap PT-input dialog content in a own boundary element provider, so that content inside can use the dialog content container as the boundary element (as opposed to the editor boundary itself).

This is needed to correctly calculate the height of it for virtual array lists which will use the closest boundary for this.

### What to review

That long arrays will correctly show when loaded inside a PT-input modal (both popover and regular dialogs).

### Notes for release
* Fix issue where arrays inputs inside dialogs inside the Portable Text input would not properly render.


robinpyon (approved):
LGTM, thank you both!

skogsmaskin:
@bjoerge I have updated the code to use state here instead as you proposed based on the experience with the array item dialogs.

bjoerge (approved):

bjoerge (approved):

PR #4487, opened 2023-05-12
fix(form/inputs): only scroll and select focusPath when needed

skogsmaskin:
### Description

The PT-input is doing focusPath tracking in some situations it should not, which can cause issues when editing embedded objects inside through the dialogs.

### What to review

* That we don't make a new selection (or scrollIntoView) in the editor when an object is being edited or multiple times.
* That studio links pointing to content inside a PT-input is properly selected, focused and scrolled to
* That the selection is correct when toggling fullscreen in the PT-input
* That clicking on top navigation validation items for the PT-input is properly selected, focused and scrolled to

### Notes for release

Doesn't need to be mentioned, it's a unreleased regression from https://github.com/sanity-io/sanity/pull/4466

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/hooks/useTrackFocusPath.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/hooks/useTrackFocusPath.tsx
@@ -1,7 +1,9 @@
 import {PortableTextEditor, usePortableTextEditor} from '@sanity/portable-text-editor'
 import {Path} from '@sanity/types'
-import {useEffect} from 'react'
+import {useEffect, useRef} from 'react'
 import scrollIntoView from 'scroll-into-view-if-needed'
+import {isEqual} from 'lodash'
+import {PortableTextMemberItem} from '../PortableTextInput'
 import {usePortableTextMemberItems} from './usePortableTextMembers'
 
 interface Props {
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/PortableText/hooks/useTrackFocusPath.tsx
```diff
@@ -27,6 +27,13 @@ export function useTrackFocusPath(props: Props): void {
       .filter((item) => item.member.open)
       .sort((a, b) => b.member.item.path.length - a.member.item.path.length)[0]
     if (memberItem && memberItem.elementRef?.current) {
+      // Don't do anything if we have already "closed" a textBlock (see below) or have a object open for editing.
+      const isTextBlockClosed = memberItem.kind === 'textBlock' && !memberItem.member.open
+      const isObjectOpen = memberItem.kind !== 'textBlock' && memberItem.member.open
```
skogsmaskin, line 32: Will this be true with the case where you only have an object "selected" in the editor, but not having focus inside?

Ref. what we talked about offline @bjoerge

skogsmaskin (commented):
I'm a bit unsure how this will act  if we decide to change the formState `open` property for the member.

skogsmaskin:
Force pushed a new solution that probably works better.
Just do it once per opened `memberItem`.

skogsmaskin:
I have identified another regression here too, when creating new annotations and inline objects, the selection screwed up from the focus tracking.

skogsmaskin:
Force pushed a new approach that doesn't interfere with creating links and inline objects.

.gitignore
```diff
@@ -71,3 +71,7 @@ etc
 /test/e2e/report
 /test/e2e/results
 /test/e2e/state
+
+# Temporary files made by yarn when running canary releases
+.yarn
+.yarnrc
```
RitaDias, line 77: Do these also ignore the .cjs files that are appearing bellow on the files changes? 👇

RitaDias (approved):
After testing it around, it looks good :)

bjoerge (approved):
LGTM!

skogsmaskin:
These changes will be part of another PR that @bjoerge will merge in later on.

Closing this for now.

skogsmaskin:
This will be superseded by Superseded by https://github.com/sanity-io/sanity/pull/4524

PR #4499, opened 2023-05-15
fix(core): editing reference input in PTE does not show value

binoy14:
### Description

This reverts the change in #4105 where it fixes an issue where clicking Replace in reference inputs would still retain the search query. This is because the change broke reference inputs in PTE. 

### What to review

Following scenario work
1. Editing reference field in PTE shows correct value
2. ~In an array list clicking on replace removes the current value from the input~ We will tackle this in future iteration as fixing the above issue is more critical

Does the code change make sense?

### Notes for release

- Fixes an issue where editing reference inputs in Portable Text Editor does not show current value

### DIFF
--- a/dev/test-studio/schema/debug/ptReference.ts
+++ b/dev/test-studio/schema/debug/ptReference.ts
@@ -1,0 +1,24 @@
+import {defineType} from 'sanity'
+
+export default defineType({
+  name: 'ptReference',
+  title: 'Portable Text Reference',
+  type: 'document',
+  fields: [
+    {
+      name: 'body',
+      title: 'Body',
+      type: 'array',
+      of: [
+        {
+          type: 'block',
+          marks: {
+            annotations: [{type: 'reference', to: {type: 'author'}}],
+          },
+          of: [{type: 'reference', to: {type: 'author'}}],
+        },
+        {type: 'reference', to: {type: 'author'}},
+      ],
+    },
+  ],
+})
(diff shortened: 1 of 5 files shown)


RitaDias:
Do we have an example of the first scenario mentioned on the test studio? 👀 
___

Also, while testing the second scenario: I am concerned a specific use case when it comes to user expectation and data permanence (I'd love to hear @mikolajdobrucki 's thoughts).

In a scenario where:
1. I click "replace" on an reference field
2. But after looking at my options I actually end up not replacing it (either I didn't find what I wanted or whatever the reason might be)
3. I end up with the empty field (an in the case if it's in an array fully removing it from the list)

You can still recover it **if** the data you are trying to recover has been **published** 

https://github.com/sanity-io/sanity/assets/6951139/d10c7e45-6d94-41ea-a69b-b5cdaa985bf0

But if the data is **draft** (either by you or someone else) then the drafted item in the array will be lost
Case where I created a draft item on a user 2 and then tried to replace it with user 1. 
I think this specific situation will bring up a lot of confusion (plus potential draft data loss which might be minor but still a thought to keep in mind)

https://github.com/sanity-io/sanity/assets/6951139/6e359cdd-8133-4db6-b78f-344fe8c963ac

binoy14:
> Do we have an example of the first scenario mentioned on the test studio? 👀

I will add debug inputs, I don't think it currently exists

re: second use case, that makes sense I will take a look at it again to achieve the expected behavior without compromising on data loss

RitaDias (approved):
Looks good! 👍

RitaDias (approved):
Looks good! 👍

robinpyon (approved):
Thanks @binoy14, LGTM!

PR #4506, opened 2023-05-16
Fix re-rendering issues

skogsmaskin:
### Description

Both `onChange` and `onItemRemove` sent in to an `arrayOfObjects` input will be a new object every time an array item changes because it's callback is dependent on the field value.

Use a `ref` for this dependency instead. This also fixes the instability of `onItemRemove` which again is dependent on the `onChange` callback.

I have also fixed a snag with presence causing re-rendering for the PT-input as they are new empty arrays each time presence is returned back. Use the same empty array object in this case.

### What to review

* [Set this constant](https://github.com/sanity-io/sanity/blob/fix/re-rendering-issue/packages/sanity/src/core/form/inputs/PortableText/debugRender.ts#L1) to true then write some Portable Text. The only colour that should change is the current block you're typing in, or where a new presence is made.

### Notes for release
* Improved performance for arrays and portable text inputs.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
@@ -1,7 +1,7 @@
 import {PortableTextEditor, usePortableTextEditor} from '@sanity/portable-text-editor'
 import {ObjectSchemaType, Path, PortableTextObject} from '@sanity/types'
 import {Tooltip} from '@sanity/ui'
-import React, {ComponentType, useCallback, useMemo} from 'react'
+import React, {ComponentType, useCallback, useMemo, useRef} from 'react'
 import {pathToString} from '../../../../field'
 import {BlockAnnotationProps, RenderCustomMarkers} from '../../../types'
 import {DefaultMarkers} from '../_legacyDefaultParts/Markers'
@@ -99,13 +99,14 @@ export function Annotation(props: AnnotationProps) {
   const presence = memberItem?.node.presence?.length ? memberItem.node.presence : EMPTY_ARRAY
 
   const isOpen = Boolean(memberItem?.member.open)
-  const input = memberItem?.input
+  const inputRef = useRef(memberItem?.input)
+  inputRef.current = memberItem?.input // Update at every render as other Annotation props updates
 
   const componentProps = useMemo(
     (): BlockAnnotationProps => ({
       __unstable_boundaryElement: boundaryElement || undefined,
       __unstable_referenceElement: memberItem?.elementRef?.current || undefined,
-      children: input,
+      children: inputRef.current,
       focused,
       markers,
       onClose,
@@ -128,7 +129,7 @@ export function Annotation(props: AnnotationProps) {
       boundaryElement,
       editor.schemaTypes.block,
       focused,
-      input,
+      inputRef,
       isOpen,
       markers,
       markersToolTip,
--- a/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
@@ -96,7 +96,7 @@ export function Annotation(props: AnnotationProps) {
     [Markers, markers, renderCustomMarkers, text, validation]
   )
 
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
```diff
@@ -96,7 +96,11 @@ export function Annotation(props: AnnotationProps) {
     [Markers, markers, renderCustomMarkers, text, validation]
   )
 
-  const presence = memberItem?.node.presence || EMPTY_ARRAY
+  const presence =
+    memberItem?.node.presence && memberItem?.node.presence.length > 0
```
binoy14, line 100: 🤔 Could this work instead?

```suggestion
    memberItem?.node.presence?.length
```

packages/sanity/src/core/form/inputs/PortableText/object/BlockObject.tsx
```diff
@@ -123,7 +123,10 @@ export function BlockObject(props: BlockObjectProps) {
   const parentSchemaType = editor.schemaTypes.portableText
   const hasMarkers = Boolean(markers.length > 0)
 
-  const presence = memberItem?.node.presence || EMPTY_ARRAY
+  const presence =
+    memberItem?.node.presence && memberItem?.node.presence.length > 0
```
binoy14, line 127: same as above

packages/sanity/src/core/form/inputs/PortableText/object/InlineObject.tsx
```diff
@@ -76,7 +76,10 @@ export const InlineObject = (props: InlineObjectProps) => {
 
   const isOpen = Boolean(memberItem?.member.open)
   const input = memberItem?.input
-  const presence = memberItem?.node.presence || EMPTY_ARRAY
+  const presence =
+    memberItem?.node.presence && memberItem?.node.presence.length > 0
```
binoy14, line 80: same as top comment

packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -147,7 +147,10 @@ export function TextBlock(props: TextBlockProps) {
 
   const isOpen = Boolean(memberItem?.member.open)
   const parentSchemaType = editor.schemaTypes.portableText
-  const presence = memberItem?.node.presence || EMPTY_ARRAY
+  const presence =
+    memberItem?.node.presence && memberItem?.node.presence.length > 0
```
binoy14, line 151: same as top comment

binoy14 (approved):
Works great from my testing, code looks good just left a small comment

packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
skogsmaskin: Indeed much cleaner! Thanks, committing a fix.

packages/sanity/src/core/form/members/object/fields/ArrayOfObjectsField.tsx
```diff
@@ -99,7 +102,7 @@ export function ArrayOfObjectsField(props: {
 
       if (isRemovingLastItem) {
         // apply the patch to the current value
-        const result = applyAll(member.field.value || [], patches)
+        const result = applyAll(valueRef.current || [], patches)
```
skogsmaskin, line 105: I'm gonna update the value ref here also as according to what was done by @RitaDias here: https://github.com/sanity-io/sanity/pull/4518/files#diff-bb9bedb1a67a3a5e89f968e104a98a4df4770f736cb430c59d53a93af8170242R46-R49

skogsmaskin:
Pushed some additional code that will align with what https://github.com/sanity-io/sanity/pull/4518 did, where we update the ref from the patched result as well.

skogsmaskin (commented):
I'm thinking now that the empty array stuff for presence should be dealt with in formState? @binoy14

RitaDias (approved):
I see there are still some questions being asked but I tested out use cases (the ones brought up in the description and the older ones related to emptying of arrays) and reviewed the code and it looks good so I'll leave a ✅ :)

PR #4520, opened 2023-05-22
chore(perf): support passing a comma-separated list of test-ids when running perf suite

bjoerge:
### Description

Adds support for specifying a comma separated list of test-ids to run when running the performance tests

Example:
```
yarn perf:test --tagsOnly --count=10 --testIds=deeply-nested-objects-test,large-document-editing
```

### What to review
Does it make sense?

### Notes for release

n/a internal tooling

### DIFF
--- a/perf/cli.ts
+++ b/perf/cli.ts
@@ -18,6 +18,7 @@ async function main(args: {
   tagsOnly?: boolean
   pattern?: string
   local?: boolean
+  testIds?: string
   count?: string
   label?: string
 }) {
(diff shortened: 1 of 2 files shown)


binoy14 (approved):

RitaDias (approved):
👋 The code seems to make sense to me!

I only have one question which is: when trying to run this locally I got an error `Missing required environment variable "PERF_TEST_METRICS_TOKEN"` and I'm assuming we probably want to write down somewhere that in order to run this you need to set up a env variable?

bjoerge:
> I'm assuming we probably want to write down somewhere that in order to run this you need to set up a env variable?

It's in the readme here: https://github.com/sanity-io/sanity/tree/next/perf#prerequisites, but that _might_ not be the natural place to look for it. Open for suggestions.

Thanks for reviewing!

PR #4522, opened 2023-05-23
fix(core): add `PortableTextInputProps` to `InputProps` union

hermanwikner:
### Description

This pull request adds `PortableTextInputProps` to the `InputProps` union. This is done because `renderDefault` expects `InputProps`, so when doing things like in the example below, you will get TS errors.

```ts
import {PortableTextInputProps, defineType} from 'sanity'

function MyEditor(props: PortableTextInputProps) {
}

export const myDocument = defineType({
  // ...
  type: 'document',
  name: 'myDocument',

  fields: [
    {
      name: 'body',
      type: 'array',
      components: {
        input: MyEditor,
      },
      of: [
        // ...
      ],
    },
  ],
})

```

### What to review

Make sure that you don't get TS errors

### Notes for release

n/a


skogsmaskin (approved):
Thanks @hermanwikner

This LGTM

packages/sanity/src/core/form/inputs/PortableText/_legacyDefaultParts/__tests__/BlockExtras.test.tsx
```diff
@@ -2,8 +2,8 @@ import {fireEvent, waitFor} from '@testing-library/react'
 import React from 'react'
 import {PortableTextMarker} from '../../../../types'
 // import {renderInput} from '../../../../test/renderInput'
-import {PortableTextInput, PortableTextInputProps} from '../../PortableTextInput'
-import {portableTextType} from './schema'
+// import {PortableTextInput, PortableTextInputProps} from '../../PortableTextInput'
+// import {portableTextType} from './schema'
```
bjoerge, line 6: can these be removed?

bjoerge (approved):
Thanks @hermanwikner!

packages/sanity/src/core/form/inputs/PortableText/_legacyDefaultParts/__tests__/BlockExtras.test.tsx
hermanwikner: Yes, probably! But there were a few things commented out in that file that used these types, so I thought they might as well be left commented out

bjoerge: Ah, I see, makes sense to keep them, then. Thnx for clarifying!

PR #4526, opened 2023-05-24
fix: whitelist all standard text decorators in block-tools

sjelfull:
### Description

This updates the whitelist of block-tools to allow highly requested decorators like `sup/sub`, as well as any other similar HTML elements that will "just work": `mark, ins, small` (we already had `del` in there).

It also updates the README to be more specific that we only will convert a subset of HTML elements, and adds a link to the internal whitelist.

### What to review

- Review that the docs update makes sense to you and makes it clearer what tags we whitelist
- Review the list of whitelisted decorators: Any objects to the additions?

### Notes for release

- Added support for `<sup>`, `<sub>`, `<ins>` and `<mark>` in `@sanity/block-tools`

### DIFF
--- a/packages/@sanity/block-tools/README.md
+++ b/packages/@sanity/block-tools/README.md
@@ -68,13 +68,34 @@ This will deserialize the input html (string) into blocks.
 
 #### Params
 
+##### `html`
+
+The stringified version of the HTML you are importing
+
 ##### `blockContentType`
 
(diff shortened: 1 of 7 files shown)


skogsmaskin (approved):
LGTM! Thank you for this @sjelfull!

robinpyon (approved):
Thank you @sjelfull!

PR #4529, opened 2023-05-25
feat(core): add document action custom dialog component

hermanwikner:
### Description
This pull request introduces a new `"custom"` option to `dialog.type` along with a `dialog.component` which makes it possible to use a custom dialog component in a document action.

There are situations when it can be useful to configure a custom dialog component in a document action, where an example of this is in the delete document action in `sanity/desk`. This document action currently has a workaround to achieve this, where a completely custom dialog is passed to `dialog.content`, which results in two dialogs being rendered on top of each other, first the default document action dialog and then the custom dialog. This results in there being two backdrops whose opacity is multiplied, which results in a backdrop that is too dark for the delete dialog.

Example of usage:
```ts
import React, {useCallback, useState} from 'react'
import {DocumentActionComponent} from 'sanity'
import {Dialog, Text} from '@sanity/ui'

export const MyDocumentAction: DocumentActionComponent = () => {
  const [open, setOpen] = useState<boolean>(false)
  const toggleOpen = useCallback(() => setOpen((v) => !v), [])

  return {
    label: 'My Document Action',
    onHandle: toggleOpen,
    dialog: {
      type: 'custom',
      component: open && (
        <Dialog onClose={toggleOpen}>
          <Text>My custom dialog</Text>
        </Dialog>
      ),
    },
  }
}
```

### What to review

- Make sure that all the document actions in `sanity/desk` works as expected
- Does this change make sense?

### Notes for release

Add document action custom dialog component

### DIFF
--- a/dev/test-studio/documentActions/actions/TestCustomComponentAction.tsx
+++ b/dev/test-studio/documentActions/actions/TestCustomComponentAction.tsx
@@ -1,0 +1,35 @@
+import {Button, Card, Dialog, Stack, Text} from '@sanity/ui'
+import {DocumentActionComponent} from 'sanity'
+import React, {useCallback, useState} from 'react'
+
+export const TestCustomComponentAction: DocumentActionComponent = () => {
+  const [open, setOpen] = useState<boolean>(false)
+  const toggleOpen = useCallback(() => setOpen((v) => !v), [])
+
+  return {
+    label: 'Custom modal',
+    tone: 'primary',
+    onHandle: toggleOpen,
+    dialog: {
+      type: 'custom',
+      component: open && (
+        <Dialog
+          header="Custom action component"
+          id="custom-modal"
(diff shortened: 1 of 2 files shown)


packages/sanity/src/desk/constants.ts
```diff
@@ -5,3 +5,5 @@ export const _DEBUG = false
 
 export const EMPTY_PARAMS = {}
 export const LOADING_PANE = Symbol('LOADING_PANE')
+
+export const DOCUMENT_PANEL_PORTAL_ELEMENT = 'documentPanelPortalElement'
```
bjoerge, line 9: Great call doing it this way

bjoerge (approved):
Thank you @hermanwikner - this is very nice!

robinpyon (approved):
Thanks @hermanwikner!

PR #4530, opened 2023-05-25
fix(core/form/inputs): fix rendering regression for PT-input

skogsmaskin:
### Description

There was a rendering regression from v3.11.2 in 75dfb9ca1bbb6bc21cad9c8139e4a9d87f9522ad where nested PT-inputs (PT in PT) would not properly update their form inputs. It was probably a bit naive to use refs like that. 😔 

This will revert the use of refs for the input in the PT node components, and let the PortableTextInput figure out if we need to re-render their inputs instead.

I have added a condition that will only re-render the input if the member is open _or_ it's value has changed.

This way we keep the intended performance optimisations by not render all the object inputs with every cycle,
but making sure we do so initially, when the object is open, or it's value has changed.

### What to review
* [Set this constant](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/PortableText/debugRender.ts#L1) to true then write some Portable Text. The only colour that should change is the current block you're typing in, or where a new presence or value is made.
* That custom block rendering is able to render the inputs at any time even when displayed directly in the block rendering of the PT-input.
```
        {
          type: 'image',
          name: 'image',
          components: {
            block: (props: BlockProps) => <div contentEditable={false}>{props.children}</div>,
            // props.children here is the input
          },
        },
```
* That one can edit a PT annotation in a nested Portable Text Input (PT block object containing a new PT-input on one of it's properties).

### Notes for release
* Fix a recent regression from v3.11.2 with the PT-input where text annotations inside nested PT-inputs could not properly be edited.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -199,9 +199,13 @@ export function PortableTextInput(props: PortableTextInputProps) {
       }
 
       if (existingItem) {
+        // Only update the input if the node is open or the value has changed
+        // This is a performance optimization.
+        if (item.member.open || existingItem.node.value !== item.node.value) {
+          existingItem.input = input
+        }
         existingItem.member = item.member
         existingItem.node = item.node
-        existingItem.input = input
         return existingItem
       }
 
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
LGTM!

RitaDias (approved):
Looks good to me :)

PR #4531, opened 2023-05-25
Keep stable child presence

skogsmaskin:
### Description

When using `useChildPresence` it will return a new object even though the presence data is equal.

Keep stable child presence by using immutableReconcile on them.

I have also updated the PT-input nodes to get their presence from this hook, rather than getting them from the array member item. This is also how the Array Input does it, and we get the added performance there as well.

### What to review
* That there is a new presence list only when something actually changed in the list.

### Notes for release
* Improved presence rendering performance

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
@@ -6,6 +6,7 @@ import {
   PortableTextEditor,
   usePortableTextEditor,
 } from '@sanity/portable-text-editor'
+import {isEqual} from '@sanity/util/paths'
 import {BlockProps, RenderCustomMarkers, RenderPreviewCallback} from '../../../types'
 import {PatchArg} from '../../../patch'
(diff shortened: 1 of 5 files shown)


bjoerge (approved):

packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -79,6 +80,12 @@ export function TextBlock(props: TextBlockProps) {
   const memberItem = usePortableTextMemberItem(pathToString(path))
   const editor = usePortableTextEditor()
 
+  const presence = useChildPresence(path, true)
+  const rootPresence = useMemo(
```
binoy14, line 84: ❓`rootPresence` is an unused, should it be used?

skogsmaskin: Thank you for spotting this @binoy14 

I have updated the code slightly (amened last commit), please refresh.

PR #4534, opened 2023-05-26
fix: update popover position when panes change

sjelfull:
### Description

The Popover component from `@sanity/ui` and `floating-ui` does a good job of updating the position when things like its reference element or boundary changes. This is handled by the built-in [`autoUpdate`](https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/autoUpdate.ts) middleware, setting up a `ResizeObserver` + listening for scroll events.

However we have at least one instance in the studio where this isn't enough. If you have an open popover inside of the PTE, the boundary element will be the PTE itself. If any panes is collapsed/expanded, the position of the popover should be updated as well, but the boundary element doesn't change in any way that can be detected automatically.

We don't have a way to access the root boundary element either, since the closest boundary will be the PTE, instead of the document pane.

`floating-ui` has a `update` method for cases like this where you need to manually update the position. Here we listen to the `panes` value, and trigges an update event.

This will only affect any open popovers, and shouldn't have any negative performance impact.

This PR relies on sanity-io/ui#1099 being released, that includes the `usePopover` hook.

### What to review
- Is there any downside to listening to the `panes` value and triggering an update?
- Open the test studio -> Open a document inside `Standard Inputs -> Portable Text -> Blocks test`, create a link, open the edit popover, and then try to toggle some panes. See if the popover keeps being anchored to the text its linked to.

### Notes for release

- Fixes an issue with misaligned popovers when panes get expanded or collapsed

### DIFF
--- a/dev/design-studio/package.json
+++ b/dev/design-studio/package.json
@@ -33,7 +33,7 @@
     "@sanity/color": "^2.1.20",
     "@sanity/icons": "^2.3.0",
     "@sanity/logos": "^2.0.2",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "sanity": "3.11.3",
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -24,7 +24,7 @@
     "@sanity/image-url": "^1.0.2",
     "@sanity/portable-text-editor": "3.11.3",
     "@sanity/types": "3.11.3",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "@sanity/ui-workshop": "^1.0.0",
     "@sanity/util": "3.11.3",
     "@sanity/uuid": "^3.0.1",
--- a/examples/ecommerce-studio/package.json
+++ b/examples/ecommerce-studio/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@sanity/cli": "3.11.3",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "react": "^18.2.0",
     "react-barcode": "^1.4.1",
     "react-dom": "^18.2.0",
--- a/packages/@sanity/portable-text-editor/package.json
+++ b/packages/@sanity/portable-text-editor/package.json
@@ -75,7 +75,7 @@
     "slate": "0.81.1"
   },
   "devDependencies": {
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@types/debug": "^4.1.5",
--- a/packages/@sanity/vision/package.json
+++ b/packages/@sanity/vision/package.json
@@ -67,7 +67,7 @@
     "@rexxars/react-split-pane": "^0.1.93",
     "@sanity/color": "^2.1.20",
     "@sanity/icons": "^2.3.0",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "@uiw/react-codemirror": "^4.11.4",
     "hashlru": "^2.3.0",
     "is-hotkey": "^0.1.6",
--- a/packages/sanity/package.json
+++ b/packages/sanity/package.json
@@ -151,7 +151,7 @@
     "@sanity/portable-text-editor": "3.11.3",
     "@sanity/schema": "3.11.3",
     "@sanity/types": "3.11.3",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
     "@sanity/util": "3.11.3",
     "@sanity/uuid": "^3.0.1",
     "@sanity/validation": "3.11.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3982,10 +3982,22 @@
     segmented-property "^3.0.3"
     vite "^4.1.1"
 
-"@sanity/ui@^1", "@sanity/ui@^1.0.0", "@sanity/ui@^1.3.3":
+"@sanity/ui@^1", "@sanity/ui@^1.0.0":
   version "1.3.3"
   resolved "https://registry.yarnpkg.com/@sanity/ui/-/ui-1.3.3.tgz#ec3b9aeb8be79f53d94406bbf80c806af1a47e7f"
   integrity sha512-fp5eyc6U760Mvg6wmiv0wOnX/uiPpFoPf4Ne9cJX6X80mk2/u2pOkXCuMuhfXmty3KP7uPZxkapD8atd7BaWIw==
+  dependencies:
+    "@floating-ui/react-dom" "1.1.1"
+    "@sanity/color" "^2.2.3"
+    "@sanity/icons" "^2.2.2"
+    csstype "^3.1.1"
+    framer-motion "^10.0.0"
+    react-refractor "^2.1.7"
+
+"@sanity/ui@popover-fixes-canary":
+  version "1.4.0"
+  resolved "https://registry.yarnpkg.com/@sanity/ui/-/ui-1.4.0.tgz#63c5cf642c5a1d089479ad3332da61d0006bb77d"
+  integrity sha512-9pcBJ3kUFCvBVYsbKam6mKgWT72HcZ0yRSxdZPsjJD1Nwo4jKWSAaBfqKZ7W2zWkV2TBAfoSF85Hw5foGAho1Q==
   dependencies:
     "@floating-ui/react-dom" "1.1.1"
     "@sanity/color" "^2.2.3"


RitaDias:
Hi @sjelfull 👋 thanks for setting this up!

Very quickly: I noticed that a tooltip (not the url itself but the tooltip with the edit button and trash button) is misaligned. Is this something that should be in the scope of this PR or something separate? 

You'll be able to find what I mean in the last few seconds of the video:

https://github.com/sanity-io/sanity/assets/6951139/73d11d4e-9ad1-411a-9e49-117feedc81ec

dev/design-studio/package.json
```diff
@@ -33,7 +33,7 @@
     "@sanity/color": "^2.1.20",
     "@sanity/icons": "^2.3.0",
     "@sanity/logos": "^2.0.2",
-    "@sanity/ui": "^1.3.3",
+    "@sanity/ui": "popover-fixes-canary",
```
RitaDias, line 36: Leaving a small comment here to not forget to revert this version before merging 😄 
(I've totally never done it myself, nope)

packages/sanity/src/desk/components/pane/PaneLayout.tsx
```diff
@@ -50,6 +51,16 @@ export function PaneLayout(
...
   }, [collapsed, onCollapse, onExpand])
 
+  // Force popover position update when panes change
+  // This is needed to ensure that any open popovers is positioned correctly if a pane is toggled
+  // between collapsed and expanded. We normally use the boundary element, but that won't work
+  // inside the PTE
+  useEffect(() => {
+    if (state.panes.length) {
+      forcePopoverUpdate?.()
+    }
```
RitaDias, line 61: Would it make sense / is there a way of checking if there are even popups open? I can see that this could be called multiple times without necessarily needing it if no PTE ends up having any popovers open.

Or perhaps the `forcePopoverUpdate` (which might be better served with a more descriptive name) already checks for this case?

sjelfull:
> Hi @sjelfull 👋 thanks for setting this up!
> 
> Very quickly: I noticed that a tooltip (not the url itself but the tooltip with the edit button and trash button) is misaligned. Is this something that should be in the scope of this PR or something separate? 
> 
> You'll be able to find what I mean in the last few seconds of the video:
> 
> https://github.com/sanity-io/sanity/assets/6951139/73d11d4e-9ad1-411a-9e49-117feedc81ec
> 
> 

Good catch! I have only looked at popovers specifically, but tooltips are using the same library and have the same weakness. I'll look into it and see how we want to tackle it.

packages/sanity/src/desk/components/pane/PaneLayout.tsx
sjelfull: There isn't a way currently, but there might be a way we can constrain this to only run if the document panel is resized.

For what it is worth, the only thing the method does is update the state in the context, and won't do anything unless any popovers is listening to the update.

hermanwikner: We could do something like this to prevent the update function from running when there are no popovers in the DOM:

```ts
function hasPopovers() {
  const attr = '[data-ui="Popover"]'
  const popovers = document.querySelectorAll(attr)
  return popovers.length > 0
}
```

```ts
  useEffect(() => {
    if (state.panes.length && hasPopovers()) {
      forcePopoverUpdate?.()
    }
  }, [state.panes, forcePopoverUpdate])

```

packages/sanity/src/desk/components/pane/PaneLayout.tsx
```diff
@@ -50,6 +51,16 @@ export function PaneLayout(
     if (!collapsed && onExpand) onExpand()
   }, [collapsed, onCollapse, onExpand])
 
+  // Force popover position update when panes change
+  // This is needed to ensure that any open popovers is positioned correctly if a pane is toggled
+  // between collapsed and expanded. We normally use the boundary element, but that won't work
+  // inside the PTE
+  useEffect(() => {
```
hermanwikner, line 58: We could do something like this to prevent the update function from running when there are no popovers in the DOM:

```ts
function hasPopovers() {
  const attr = '[data-ui="Popover"]'
  const popovers = document.querySelectorAll(attr)
  return popovers.length > 0
}
```

```ts
  useEffect(() => {
    if (state.panes.length && hasPopovers()) {
      forcePopoverUpdate?.()
    }
  }, [state.panes, forcePopoverUpdate])

```

hermanwikner:
The current `useEffect` in `PaneLayout` seems to work great! 👍 

I'm just thinking that there might be other situations (maybe not now, but in the future, or in custom implementations) that might affect the size of a pane and when we want to run the update function. Therefore, I _think_ it would be more "safe" to add a resize observer to the `Pane` component – and trigger the update function from there (only if there are any popovers in the pane).

Something like this:

```ts
function Pane() {
  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)
  const {forceUpdate} = usePopover()

  // ...

  const {
    collapsed,
    // ...
  } = usePaneLayout()

  // ...

  const popoverForceUpdate = useCallback(() => {
    // If the pane is collapsed, we don't need to force update
    if (collapsed || !rootElement) return

    // If there are no popovers in the pane, we don't need to force update
    const popovers = rootElement.querySelectorAll('[data-ui="Popover"]')
    if (popovers.length === 0) return

    forceUpdate()
  }, [collapsed, forceUpdate, rootElement])

  useEffect(() => {
    if (!rootElement) return undefined

    const ro = new ResizeObserver(popoverForceUpdate)

    ro.observe(rootElement)

    return () => {
      ro.disconnect()
      ro.unobserve(rootElement)
    }
  }, [popoverForceUpdate, rootElement])

  // ...

  return <Root ref={setRootElement}>{/* ... */}</Root>
}
```

packages/sanity/src/desk/components/pane/PaneLayout.tsx
sjelfull: I agree that this might be a good idea, but is there precedence in the studio of querying for HTML attributes like this?

hermanwikner:
The `useEffect` in `PaneLayout` works well for repositioning popovers when pane state is updated. However, you may still encounter situations when the popover is not positioned correctly. See video with an example using a custom component. This particular situation can be solved by adding a resize observer to `FormView` and use the update function as a callback.

```ts
  const {forceUpdate} = usePopover()

  useEffect(() => {
    const el = formRef.current

    if (!el) return undefined

    const ro = new ResizeObserver(forceUpdate)

    ro.observe(el)

    return () => {
      ro.disconnect()
      ro.unobserve(el)
    }
  }, [forceUpdate, formRef])
```

https://github.com/sanity-io/sanity/assets/15094168/bf2454b6-8cb2-4f64-b464-cf2d8cebd1ea

sjelfull:
> 
> Very quickly: I noticed that a tooltip (not the url itself but the tooltip with the edit button and trash button) is misaligned. Is this something that should be in the scope of this PR or something separate?
> 

@RitaDias Turns out that the tooltip is handled by a Popover. The complication is that the position is actually controlled by [a virtual element](https://floating-ui.com/docs/virtual-elements), calculated from the [rect of the range](https://github.com/sanity-io/sanity/blob/39081c27b68628ebbc3397b823a9fd476e92c348/packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx#L101-L110) selected in the PTE.

This is a separate issue, and won't be affected by the proposed fix in `@sanity/ui`. I'll discuss with the team how we can tackle this.

RitaDias (approved):
Apart from reverting the version that we have for sanity ui I'd be ok with merging this unless we're still waiting for something 👍

packages/sanity/src/desk/components/pane/PaneLayout.tsx
```diff
@@ -50,6 +51,16 @@ export function PaneLayout(
     if (!collapsed && onExpand) onExpand()
   }, [collapsed, onCollapse, onExpand])
 
+  // Force popover position update when panes change
+  // This is needed to ensure that any open popovers is positioned correctly if a pane is toggled
+  // between collapsed and expanded. We normally use the boundary element, but that won't work
+  // inside the PTE
```
skogsmaskin, line 57: This comment ending could read: `for popovers where the position is calculated from elements inside of the input itself` to be more clear.

packages/sanity/src/desk/components/pane/PaneLayout.tsx
```diff
@@ -22,6 +22,7 @@ export function PaneLayout(
     CardProps &
     Omit<React.HTMLProps<HTMLDivElement>, 'as' | 'height' | 'ref' | 'wrap'>
 ) {
+  const {forceUpdate: forcePopoverUpdate} = usePopover()
```
skogsmaskin, line 25: Are we sure this is not causing any perf. issues and it only re-renders when strictly necessary?

skogsmaskin (commented):
I had some comments. Most importantly the performance question.

packages/sanity/src/desk/components/pane/PaneLayout.tsx
sjelfull: As far as I can tell, it wouldn't. There is two parts to this: the context value, and the hook that listens to updates.
Whenever you call `forceUpdate`, it will set the context value with a new random string. The [`useOnForcedUpdate `](https://github.com/sanity-io/ui/pull/1099/files#diff-c0c8b59e352a97f55536df16b5ef9f4d5f333c9171e1eea39ab2dc26ae4c5a72) hook will listen to that change, and make sure to use the `update()` method to recalculate the position. It will trigger a re-render of the popover component itself, and there might be "unnecessary" re-renders if the panes state change a lot, but this can't be avoided. It shouldn't trigger any re-renders if any popovers isn't open.

I have a branch locally where I have a slightly debounced version that runs every 15ms, that also fixes the other issue with the annotation tooltip. I'll push this, and we can test and see if we are happy.

RitaDias:
@sjelfull this one is quite old so I think it's safe to close? If you want you can reopen in but since there's so many conflicts it's just best to start from a new thing of main if we still feel it's a problem :)

jordanl17:
> @sjelfull this one is quite old so I think it's safe to close? If you want you can reopen in but since there's so many conflicts it's just best to start from a new thing of main if we still feel it's a problem :)

Think this is sensible - there also seems to be a dependency on a sanity/ui update that never materialised, so could take more lift to get this out

sjelfull:
@jordanl17 Yes thanks, have no idea if this is still an issue :)

PR #4538, opened 2023-05-30
fix(core): add change indicator to boolean fields

vicmeow:
### Description

Change indicators were no longer showing up for boolean fields.

### What to review

I see that booleans are resolved with a pass through component, but it was missing the change indicator wrapper, so I added it. Seeing as last time I worked in the Studio was v2, so not sure if this is the right way to go about this fix. Please let me know if I need to change anything.

### Notes for release

Fixed an issue where change indicators were no longer showing up on boolean fields


packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
```diff
@@ -8,8 +8,16 @@ import {FormField, FormFieldSet} from '../../components'
 import {ChangeIndicator} from '../../../changeIndicators'
 import {getTypeChain} from './helpers'
 
-function PassThrough({children}: {children: React.ReactNode}) {
-  return <>{children}</>
+function PassThrough(field: FieldProps) {
```
bjoerge, line 11: Maybe rename this to `BooleanField`? I think the idea was at some point that could be useful for other type of fields too, but it seems to be only used by boolean fields.

bjoerge (approved):
Nice, thank you! ✨

packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
vicmeow: Ok!

PR #4540, opened 2023-05-30
fix(block-tools): preserve whitespace in preprocess for certain html tags

sjelfull:
### Description

This fixes an issue in `@sanity/block-tools` where whitespace was removed from content within certain HTML tags like `<pre>`, where whitespace should be preserved.

Before this change, whitespace in the whole document was normalized before passing the html to be processed.

In this PR we replace that approach with another preprocessor that loops over the DOM tree and process each text node as long as its not inside a HTML tag that should preserve whitespace.

While this fix was focused on `<pre>` tags, I think it makes sense to  include `<textarea>` and `<code>` tags in the whitelist, even though we don't process these in any special way currently.

### What to review
- That whitespace in `<pre>` tags is now preserved (there is tests for this now)

### Notes for release

- Fixed an issue where `@sanity/block-tools` would remove whitespace from `<pre>` tags


skogsmaskin (changes requested):
Seems to have broken some tests?

PR #4541, opened 2023-05-30
chore: add tsdoc to sanity package

RitaDias:
### Description

WIP

### What to review

- [x] Have tsdoc installed
- [x] Have `@hidden` tag
- ↪️  Automate way to have permissions to run publish for tsdoc #4584 (non blocker)
- ↪️  Write `@public` docs (out of scope for this PR)


packages/sanity/src/core/config/document/actions.ts
```diff
@@ -49,47 +57,55 @@ export interface DocumentActionConfirmDialogProps {
...
+/**
+ * @hidden
+ * @beta */
 export interface DocumentActionPopoverDialogProps {
   type: 'popover'
   content: React.ReactNode
   onClose: () => void
 }
 
-/** @beta */
```
binoy14, line 79: seems like this got wiped during a rebase I am guessing causing build failure

packages/sanity/tsdoc.json
```diff
@@ -0,0 +1,23 @@
...
+        "@hidden": true,
+        "@beta": true,
+        "@internal": true,
+        "@public": true,
+        "@experimental": true,
+        "@see": true,
+        "@link": true,
+        "@example": true,
+        "@deprecated": true,
+        "@alpha": true
```
binoy14, line 21: I believe this is missing `@param` and `@returns` tag causing eslint error

```suggestion
        "@alpha": true,
        "@param": true,
        "@return": true
```

packages/sanity/src/core/config/document/actions.ts
RitaDias: Eagle eyes, thank you 🙏

RitaDias: Should be re-added now, once more thanks 🙏

packages/sanity/tsdoc.json
binoy14: made a typo in code suggestion 🤦 pushed a fix

binoy14:
This should be good to review and eventually merge, thoughts @RitaDias ?

If this is merged writing more docs would be easier to test directly against `next` and can be incremental instead of making this PR very large to review both by the team and by Martin

RitaDias:
> This should be good to review and eventually merge, thoughts @RitaDias ?
> 
> If this is merged writing more docs would be easier to test directly against `next` and can be incremental instead of making this PR very large to review both by the team and by Martin

I'm all for it, I think it has little risk. there will be some things that we probably want to change in the test-studio but again it's the test studio so I'm not concerned.

@bjoerge would you be ok with that?

dev/test-studio/package.json
```diff
@@ -24,6 +24,7 @@
     "@sanity/image-url": "^1.0.2",
     "@sanity/portable-text-editor": "3.12.1",
     "@sanity/types": "3.12.1",
+    "@sanity/tsdoc": "1.0.0-alpha.22",
```
binoy14, line 27: Probably should upgrade to latest alpha everywhere before merging

binoy14 (approved):
✅

bjoerge (approved):
Great work! 👏🏼

PR #4544, opened 2023-05-30
refactor(portable-text-editor): improve onPaste typings

skogsmaskin:
### Description

This will improve the DX a bit on the `PortableTextEditor` `onPaste` prop which is reused by the `PortableTextInput`

First I noticed this function is being called with the wrong path value.  This is supposed to be a keyed PortableText path (and not a Slate path)

Then there was a too strictly typed property on the callback function, it was typed to `PortableTextBlock` which will require a `_key` which isn't needed here. Use `TypedObject` which is the correct one.

### What to review

* That the path is called with the correct relative keyed path where the focus is at.
* That one are able to write this without TS-error:
```ts
const onPaste: PortableTextInputProps['onPaste'] = ({event, schemaTypes}) => {
  return {
    // Only a type is required here
    insert: [
      {
        _type: 'myType',
        foo: 'bar',
      },
    ],
  }
}

```

### Notes for release
* Improve typings on PortableTextInputProps and fix incorrect `path` value for `onPaste` callback function

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -8,6 +8,7 @@ import {
   useSlate,
 } from '@sanity/slate-react'
 import {noop} from 'lodash'
+import {PortableTextBlock} from '@sanity/types'
 import {
   EditorChange,
   EditorSelection,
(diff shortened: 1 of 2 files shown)


RitaDias (approved):
:shipit:

bjoerge (approved):
LGTM ✨

PR #4545, opened 2023-05-31
fix(core): add Help&Resources button to navbar

ninaandal:
### Description
The Learnability project introduces a new `Help & Resources` button in the navbar that exposes a dropdown with various resources. Most of the resources are external links, but they can also trigger internal actions within the studio. 

The resources are entered in to the [admin studio](https://admin.sanity.io/desk/studio). 

Castaway has also been extended with an additional endpoint, so everything with the following path` api.sanity/v1/help` is directed to castaway.

The `Changelog` has been removed from the navbar and is replaced by an external link to this [changelog](https://www.sanity.io/changelog) accessible in the new `Help & Resources` button. 

### What to review

The new `?` button in the navbar. 

Screenshot 2023-06-13 at 12 26 46

### Notes for release

Adds a Help and resources button to studio navbar and removes Changelog button


skogsmaskin (changes requested):
Great work @ninaandal!

However, a matter of visual design, the button looks a bit cramped up on my screen:

image

Since there is only one category, I though the category header here was actually a link. It probably helps when there are more categories, but maybe some way to have this a bit more clear (difference between category title and actual links):

image

Also, would it make sense to have this go into a working branch for the new initiative, and not `next` right now?

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
```diff
@@ -0,0 +1,123 @@
...
+              if (!subSection) return null
+              return <SubSection key={subSection._key} subSection={subSection} />
+            })}
+
+          {/* Studio version information */}
+          <Box padding={3}>
+            <Text size={1} muted weight="medium" textOverflow="ellipsis">
+              Sanity Studio version {SANITY_VERSION}
+            </Text>
+            {!error && !isLoading && latestStudioVersion && (
```
rexxars, line 38: `isLoading` can never be true here, can it? Given the opening ternary?

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
```diff
@@ -0,0 +1,123 @@
...
+interface ResourcesMenuItemProps {
+  error: Error | null
+  isLoading: boolean
+  value: ResourcesResponse | undefined
+}
+
+export function ResourcesMenuItems({error, isLoading, value}: ResourcesMenuItemProps) {
+  const sections = value?.resources?.sectionArray
+  const latestStudioVersion = value?.latestVersion
+
```
rexxars, line 15: I would consider doing early returns instead of inline conditionals - it makes it easier to read the code, since you already know the different cases are accounted for. Not a blocker, but for future reference (psuedo-code):

```
if (isLoading) {
  return <LoadingUI />
}

return <NonLoadingUiStuffHere />
```

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
```diff
@@ -0,0 +1,123 @@
...
+              </Box>
+            )}
+          </Box>
+        </>
+      )}
+    </>
+  )
+}
+
+const FallbackLinks = () => (
```
rexxars, line 52: This can/should be made into a constant instead of a function - saves React from creating and diffing it on each render.

Psuedo-code:
```diff
-const FallbackLinks = () => <>...menuItems...</>
+const fallbackLinks = <>...menuItems...</>

- <div><FallbackLinks /></div>
+ <div>{fallbackLinks}</div>
```

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/helpResources.ts
```diff
@@ -0,0 +1,27 @@
+import type {Observable} from 'rxjs'
+import {map} from 'rxjs/operators'
+import {SanityClient} from '@sanity/client'
+import {SANITY_VERSION} from '../../../../../version'
+import {ResourcesResponse} from './types'
+
+/**
+ * @internal Not a stable API yet
+ */
+
```
rexxars, line 10: Avoid newlines between TSDoc and member it is documenting
```suggestion
```

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/helpResources.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getHelpResources(client: SanityClient): Observable<ResourcesResponse> {
+  const request$ = client.observable
+    .request<ResourcesResponse>({
```
rexxars, line 13: Not critical since it is currently internal, but it would be safer to enforce an API version here rather than hope the client passed is already using the version we want. Otherwise the `ResourcesResponse` type could theoretically be incorrect.

```suggestion
  const request$ = client
    .withConfig({apiVersion: '1'})
    .observable.request<ResourcesResponse>({
```

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/helpResources.ts
```diff
@@ -0,0 +1,27 @@
...
+      url: '/help',
+      /*
+      query and tag is used by analytics for tracking.
+      Builds to: `{m: ['sanity@3.0.2']}' and serializes to: `?m=sanity@3.0.2`.
+      Final format will be e.g. https://api.sanity.io/v1/help?tag=sanity.studio.module.version-check&m=sanity%400.0.0-development
```
rexxars, line 18:
```suggestion
      Final format will be e.g. https://api.sanity.io/v1/help?tag=sanity.studio.module.version-check&m=sanity%403.0.2
```

packages/sanity/src/core/studio/components/navbar/resources/ResourcesButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const {scheme} = useColorScheme()
+
+  const {value, error, isLoading} = useGetHelpResources()
+
+  return (
+    <Flex>
+      <Tooltip
+        content={
+          <Box padding={2}>
+            <Text size={1}>Help and Resources </Text>
```
rexxars, line 24: Is the Title Case intentional here? It sounds kinda like a product, but... it isn't? :)
Not sure if this is a discussed and decided point - feel free to ignore if so.

```suggestion
            <Text size={1}>Help and resources</Text>
```

packages/sanity/src/core/studio/components/navbar/resources/ResourcesButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+`
+
+export function ResourcesButton() {
+  const {scheme} = useColorScheme()
+
+  const {value, error, isLoading} = useGetHelpResources()
+
+  return (
+    <Flex>
+      <Tooltip
```
rexxars, line 21: Tiny annoyance - this may not be easy to fix, so could potentially be skipped if so:
image

A pixel or two of this tooltip ("Help and Resources" title) shows below the menu when open. _Ideally_ the tooltip would be hidden once the menu is open.

rexxars (approved):
Left a few comments inline - none of these are blockers, so thus approving this PR, but would like to see some of them addressed in either this or a follow-up PR.

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
```diff
@@ -0,0 +1,123 @@
+import {Box, Card, Flex, Label, MenuDivider, MenuItem, Spinner, Text} from '@sanity/ui'
+import React from 'react'
+import {SANITY_VERSION} from '../../../../version'
+import {ResourcesResponse, Section} from './helper-functions/types'
+
+interface ResourcesMenuItemProps {
+  error: Error | null
+  isLoading: boolean
+  value: ResourcesResponse | undefined
```
binoy14, line 9: small nit pick, can be also written as 
```suggestion
  value?: ResourcesResponse
```

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const FallbackLinks = () => (
+  <>
+    <MenuItem
+      as="a"
+      text="Join our community"
+      size={0}
+      href="https://www.sanity.io/exchange/community"
+      target="_blank"
```
binoy14, line 59: ❓should the external links also have `rel=”noreferrer noopener”` tags as well. Since they all point to our site I am not 100% sure if it's needed

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/helpResources.ts
```diff
@@ -0,0 +1,27 @@
+import type {Observable} from 'rxjs'
+import {map} from 'rxjs/operators'
+import {SanityClient} from '@sanity/client'
+import {SANITY_VERSION} from '../../../../../version'
+import {ResourcesResponse} from './types'
+
+/**
+ * @internal Not a stable API yet
```
binoy14, line 8: `@hidden` tag is also needed to make sure it gets hidden from the docs

```suggestion
 * @internal Not a stable API yet
 * @hidden
```

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/hooks.ts
```diff
@@ -0,0 +1,18 @@
+import {useMemo} from 'react'
+import {LoadableState, useLoadable} from '../../../../../util'
+import {useClient} from '../../../../../hooks'
+import {ResourcesResponse} from './types'
+import {getHelpResources} from './helpResources'
+
+/**
+ * Fetch help resources (content for the navbar help menu) from `/help`.
+ *
+ * @internal
```
binoy14, line 10: `@hidden` tag is also need to make sure it gets hidden from the docs

```suggestion
 * @internal
 * @hidden
```

packages/sanity/src/core/studio/components/navbar/resources/helper-functions/types.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+interface InternalAction extends Item {
+  _type: 'internalAction'
+  type?: InternalActionType
+}
+
+type InternalActionType = 'show-welcome-modal'
+
+export interface ResourcesResponse {
```
binoy14, line 54: 💭 Should this have tsdoc comment, if it's internal then I would add the suggestion

```suggestion
/**
 * @hidden
 * @beta */
export interface ResourcesResponse {
```

binoy14 (approved):
Great work! ✅ added small suggestions

packages/sanity/src/core/studio/components/navbar/resources/ResourcesMenuItems.tsx
robinpyon: I don't believe it's necessary, as these are trusted links and referrer information may be useful. `rel="noopener"` is also [added by default in modern browsers](https://caniuse.com/mdn-html_elements_a_implicit_noopener) when `target="_blank"` is specified.

packages/sanity/src/core/studio/components/navbar/resources/ResourcesButton.tsx
robinpyon: Great pickup – this UX bug is also present on the existing workspaces button.

Let's potentially address this as a fast follow in Sanity UI? I don't believe tooltips should persist on click

binoy14:
Is this supposed to have release notes?

skogsmaskin (approved):
Looks much better now in regard to my previous comment! ✨

PR #4550, opened 2023-06-01
chore: fix installation command in release note template

kmelve:
Correct the installation command and bring some more helpful context to the upgrade command

### Description

Corrected the installation command in the release note template.

### What to review

Does the MD render as it should?

### Notes for release

No notes are needed for this.


bjoerge (approved):
🙇🏼

PR #4552, opened 2023-06-02
Fix issue with removing array item immediately after it's been added

### LINKED ISSUE #4539: Array of objects does not open "Add item" modal for new documents in version 3.11.3
**Describe the bug**

A schema containing an array of objects with references does not open the add item modal if the array is empty. It works adding new objects to an existing array, but if the document does not have the array yet, it does not work to add objects to it. This is a bug introduced in version 3.11.3, downgrading to 3.11.1 resolves this issue.

**To Reproduce**

1. Create a schema with an array of objects. The objects should have a reference field, but I'm not sure if that is required to recreate the bug.
2. Click on "add item" on the array in Studio. It sends a bunch of requests, saves the doc, but does not open add item modal.

**Example schema:**
```javascript
import {defineType, defineField} from 'sanity'

export default defineType({
  name: 'teamLineup',
  type: 'object',
  title: 'Lagoppstilling',
  fields: [
    defineField({
      name: 'players',
      type: 'array',
      title: 'Spillere',
      of: [{type: 'lineupPlayer'}],
    }),
  ],
})

const lineupPlayer = defineType({
  name: 'lineupPlayer',
  type: 'object',
  title: 'Spiller',
  fields: [
    defineField({
      name: 'playerRef',
      type: 'reference',
      title: 'Spiller',
      to: [{type: 'player', title: 'Spiller'}],
    }),
    defineField({
      name: 'captain',
      type: 'boolean',
      title: 'Kaptein',
    }),
    defineField({
      name: 'position',
      type: 'string',
      title: 'Posisjon',
    }),
  ],
})
```

**Expected behavior**

An empty array can be updated with new objects containing references.

3.11.3

MacOs

bjoerge:
### Description
As reported in #4539, when adding new items to an array, the dialog doesn't open after (note: happens only for arrays at level >= 1).

This happened because the function we use to check whether the array value is empty looked recursively at the value, so when adding a new item, this check would identify the newly added item as empty. I've done a little refactoring to make things a bit clearer:
- `isEmpty` renamed to `isDeepEmpty` to make it clear that it looks recursively at the value.
- Introduced a much simpler `isShallowEmptyObject` which is what we want to use when checking whether to unset empty values.

Fixes #4539

### What to review
- Make sure there's no breaking changes
- Make sure the "unset empty values" logic still works for array items and object fields
- The fix for #4539 is in f845fcb350fe90931bdacece7e0640aed1d8aef8

### Notes for release

- Fixes an issue that could sometimes cause newly inserted array items to be immediately removed


packages/@sanity/util/src/content/isDeepEmpty.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @deprecated Use `isDeepEmpty` instead
+ * todo: remove in v4
+ */
+export const isEmpty = isDeepEmpty
+
+/**
+ * @deprecated Use `isDeepEmpty` instead
+ * todo: remove in v4
+ */
+export const isEmptyObject = isDeepEmptyObject
```
bjoerge, line 63: leaving these here to avoid breaking change

packages/sanity/src/core/form/members/object/fields/ObjectField.tsx
```diff
@@ -85,7 +85,7 @@ export const ObjectField = function ObjectField(props: {
         pendingValue.current = applyAll(pendingValue.current || {}, patches)
 
         // if the result after applying the patches is empty, then we should unset the field
-        if (isEmpty(pendingValue.current)) {
+        if (pendingValue.current && isShallowEmptyObject(pendingValue.current)) {
```
bjoerge, line 88: This fixes #4539

packages/sanity/src/core/templates/resolve.ts
```diff
@@ -189,7 +189,7 @@ export async function resolveInitialObjectValue<Params extends Record<string, un
 
   const merged = deepAssign(fieldValues, initialObject)
 
-  if (isEmpty(merged)) {
+  if (isDeepEmpty(merged)) {
```
bjoerge, line 192: I'm not 100% sure we need to do a deep emptyness check here, but leaving it as-is just to be sure

packages/@sanity/util/src/content/isDeepEmpty.ts
RitaDias: Thank you for adding the comments and tags 🙏

RitaDias (approved):
Did a round of the object, image, file and array tests. Everything seems to work as intended.
Code looks good, good catch! 🎉

packages/@sanity/util/src/content/isDeepEmpty.ts
binoy14, line 1: Does it make sense to add unit tests for this? Below are some AI generated tests that might just work or might need some tweaking 

```tsx
describe('isDeepEmpty', () => {
  it('returns true for undefined', () => {
    expect(isDeepEmpty(undefined)).toBe(true);
  });

  it('returns true for null', () => {
    expect(isDeepEmpty(null)).toBe(true);
  });

  it('returns true for an empty object', () => {
    expect(isDeepEmpty({})).toBe(true);
  });

  it('returns true for an object with only empty values', () => {
    expect(isDeepEmpty({ a: {}, b: { c: {} } })).toBe(true);
  });

  it('returns true for an empty array', () => {
    expect(isDeepEmpty([])).toBe(true);
  });

  it('returns true for an array with only empty values', () => {
    expect(isDeepEmpty([[], [[]]])).toBe(true);
  });

  it('returns false for a non-empty object', () => {
    expect(isDeepEmpty({ a: 1 })).toBe(false);
  });

  it('returns false for a non-empty array', () => {
    expect(isDeepEmpty([1])).toBe(false);
  });

  it('ignores _type and _key properties', () => {
    expect(isDeepEmpty({ _type: 'foo', _key: 'bar', a: {} })).toBe(true);
  });
});

describe('isDeepEmptyObject', () => {
  it('returns true for an empty object', () => {
    expect(isDeepEmptyObject({})).toBe(true);
  });

  it('returns true for an object with only empty values', () => {
    expect(isDeepEmptyObject({ a: {}, b: { c: {} } })).toBe(true);
  });

  it('returns false for an object with non-empty values', () => {
    expect(isDeepEmptyObject({ a: {}, b: { c: {} }, d: 1 })).toBe(false);
  });

  it('ignores _type and _key properties', () => {
    expect(isDeepEmptyObject({ _type: 'foo', _key: 'bar', a: {} })).toBe(true);
  });
});

describe('isDeepEmptyArray', () => {
  it('returns true for an empty array', () => {
    expect(isDeepEmptyArray([])).toBe(true);
  });

  it('returns true for an array with only empty values', () => {
    expect(isDeepEmptyArray([[], [[]]])).toBe(true);
  });

  it('returns false for an array with non-empty values', () => {
    expect(isDeepEmptyArray([[], [[]], 1])).toBe(false);
  });
});
```

binoy14 (approved):
Code looks good thanks for doing this! 

Added a comment with some unit tests if you think it's valuable to add

PR #4553, opened 2023-06-02
fix(perf): add flag to list known test ids

bjoerge:
### Description
Lil' change to add a perf cli flag for listing current testids

### Notes for release

n/a - internal tooling


binoy14:
Do the commits look correct? Seems like it's adding more things then just the testids?

PR #4556, opened 2023-06-05
fix(form): pass the correct data to conditional fields contained inside fieldsets

bjoerge:
### Description

Currently, conditional fields on fields contained inside fieldsets doesn't get passed the correct arguments. Instead of getting the current field value, it receives the parent's field values.

### What to review

- Make sure fields + field groups and conditional fields works as they should

### Notes for release

- Fixes a bug that made conditional field callbacks inside fieldsets to get passed a wrong parent value

### DIFF
--- a/packages/sanity/src/core/form/store/formState.ts
+++ b/packages/sanity/src/core/form/store/formState.ts
@@ -603,9 +603,7 @@ function prepareObjectInputState<T>(
 
       const fieldsetMembers = fieldSet.fields.flatMap(
         (field): (FieldMember | FieldError | HiddenField)[] => {
-          const hidden = resolveConditionalProperty(field.type.hidden, conditionalPropertyContext)
-
-          if (fieldsetHidden || hidden) {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/store/formState.ts
```diff
@@ -603,26 +603,13 @@ function prepareObjectInputState<T>(
...
-          if (fieldsetHidden || hidden) {
-            return [
-              {
-                kind: 'hidden',
-                key: `field-${field.name}`,
-                name: field.name,
-                index: index,
-              },
-            ]
-          }
```
bjoerge, line 617: Removing this fixes the issue, since prepareFieldMember does the conditional hidden check (passing the correct context data) and returns a `HiddenField` if the conditional hidden callback returns true.

RitaDias (approved):
It works :shipit:

skogsmaskin (approved):
LGTM! Nice catch.

PR #4557, opened 2023-06-06
refactor(core/form/inputs): PT-input track irregular span node focusPath

skogsmaskin:
The focusPath coming from the outside can end in a `.text`  for PT-span nodes.

Usually span node paths are referred to by a path ending by their `_key`.

This will make sure we can also track focus on those paths.

### What to review

* Have one user have focus inside a PT-text node.
* Have another user copying that presence link (don't click it)
* Add `.text` to the end of the url (but remove the ending `/`, making it end like: `...22%5D.text`
* Go to that URL, you should be focused on that text span node.


bjoerge (approved):

RitaDias (approved):
Works like a charm :)

PR #4558, opened 2023-06-06
fix(form): fix bug making fields not inherit document level readOnly state

bjoerge:
### Description

This fixes what looks to have been a regression in v3.11.1 that broke document-level readOnly

### What to review
Make sure the editor makes documents of the following type readOnly:
```js
{
  name: 'test',
  type: 'document',
  fields: [
    {
      name: 'title',
      type: 'string',
    }
  ],
  readOnly: true
}
```

### Notes for release
- Fixes a regression from v3.11.1 that broke document-level readOnly


RitaDias (approved):
Great catch!

PR #4559, opened 2023-06-06
fix(core, desk): add accessible labels to presence indicators, publish status, and review changes

vicmeow:
### Description

Currently when a screen reader is using Sanity, no labels are read out about presence indicators, or understandable document status timestamps. This fixes that to some extent.

### What to review

Test the with a screen reader where possible.
1. Navigate to a document
2. Use the document sparkline or pane header to get information about the document

### Notes for release

Fixed various accessibility label issues in the desk tool

### DIFF
--- a/packages/sanity/src/core/presence/DocumentPreviewPresence.tsx
+++ b/packages/sanity/src/core/presence/DocumentPreviewPresence.tsx
@@ -70,7 +70,7 @@ export function DocumentPreviewPresence(props: DocumentPreviewPresenceProps) {
   return (
     <Tooltip content={tooltipContent} {...PRESENCE_MENU_POPOVER_PROPS}>
       <AvatarStackCard scheme={selected ? invertedScheme : undefined} $selected={selected}>
-        <AvatarStack maxLength={2}>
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/presence/DocumentPreviewPresence.tsx
```diff
@@ -70,7 +70,7 @@ export function DocumentPreviewPresence(props: DocumentPreviewPresenceProps) {
   return (
     <Tooltip content={tooltipContent} {...PRESENCE_MENU_POPOVER_PROPS}>
       <AvatarStackCard scheme={selected ? invertedScheme : undefined} $selected={selected}>
-        <AvatarStack maxLength={2}>
+        <AvatarStack maxLength={2} aria-label="Who is here">
```
hermanwikner, line 73: Could we use the string returned from `getTooltipText` as aria label?

vicmeow: Yes, will update it! Thanks 😊

bjoerge (approved):
Thanks for doing this!

rexxars (approved):
Nice - I've rebased and resolved conflicts due to new changes panel

PR #4560, opened 2023-06-06
fix(portable-text-editor): show editable when invalid value

skogsmaskin:
Always show the editable also when there is an invalid value.

This is making a better experience when choosing to ignore the validation error for a PT text block.

The invalid value is now removed from the greater value when displayed in the editor, making it possible to still edit everything else if you choose to. This was not previously possible, but became possible with the introduction of the useSyncValue hook. Enabling this behaviour now.

Note that you have to explicitly ignore a warning in order to continue editing on the remaining value.


rexxars:
@skogsmaskin @christianhg Another one for you to discuss :)

skogsmaskin:
Closing this, as it should be solved by allowing to render errors in the specific blocks containing the error. Thus it can still render the editor and you can edit anything else except those blocks containing error.

PR #4562, opened 2023-06-07
feat(desk): add `useMenuItem` to document inspector API

hermanwikner:
### Description

This pull request updates the document inspector API so that the menu item is configured with a React hook (`useMenuItem`). The pull request also moves the `showAsAction` property to be part of the menu item value. This update opens up possibilities to perform custom logic in `useMenuItem` and return a dynamic menu item object (see the validation inspector for example).

Before:
```ts
const inspector = defineDocumentInspector({
  // ...
  
  menuItem: {
    title: 'Inspector',
  },
  showAsAction: true,
})
```

After:
```ts
const inspector = defineDocumentInspector({
  // ...
  
  useMenuItem: () => {
  // ...
  
    return {
      showAsAction: true,
      title: 'Inspector',
    }
  },
})
```

### DIFF
--- a/packages/sanity/src/core/config/document/inspector.ts
+++ b/packages/sanity/src/core/config/document/inspector.ts
@@ -1,10 +1,11 @@
-import {ValidationMarker} from '@sanity/types'
 import {ButtonTone} from '@sanity/ui'
 import {ComponentType} from 'react'
 
 /** @beta */
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface DocumentInspectorProps {
+  documentId: string
+  documentType: string
   onClose: () => void
 }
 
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/config/document/inspector.ts
```diff
@@ -1,26 +1,30 @@
...
+  documentType: string
 }
 
 /** @beta */
 export interface DocumentInspectorMenuItem {
-  hidden?: boolean | ((props: DocumentInspectorCallbackContext) => boolean)
+  hidden?: boolean
   hotkeys?: string[]
   icon?: ComponentType
+  showAsAction?: boolean
```
hermanwikner, line 26: Consider making this a string instead of a boolean to make it more configurable in the future?

`location: 'default' | 'menu' | ...`

mariuslundgard (approved):
Looks great – thanks for doing this! ✨

PR #4563, opened 2023-06-07
fix(desk): close document panes when document is deleted

sjelfull:
### Description

This fixes an issue where the document pane of the deleted document was left open after the document was deleted, including any open panes to the right.

After this change, all open panes to the right including the deleted one will close.

We will also show a banner similar to the permissions banner that includes a restore button if another user has the document open.

Example video:

https://github.com/sanity-io/sanity/assets/209129/1276151f-dad4-48db-87c9-bcbc6fc52504

### What to review

- That all panes to the right including the deleted document pane closes when you delete the document
- That the delete confirmation toast shows
- Another user has the deleted document open: That the Deleted document banner shows for the user
- Another user has the deleted document open: That the user is able to restore the document from the Restore button in the banner
- Another user has the deleted document open: That it is possible to open the timeline menu without seeing an error when the timeline menu is open

### Questions
- Is checking if the history timeline has entries in combination with no document pair a solid way to detect if a document was deleted while the user was viewing it?
- Is it fine to close the pane in `DocumentOperationResults.tsx`? I tried using the `useDocumentOperationEvent` hook in the document pane to handle closing there, but there was race conditions where the pane would close before the toast  was triggered.

### Notes for release

- Improved UX when deleting a document

### DIFF
--- a/packages/sanity/src/desk/components/paneRouter/PaneRouterContext.tsx
+++ b/packages/sanity/src/desk/components/paneRouter/PaneRouterContext.tsx
@@ -25,6 +25,7 @@ export const PaneRouterContext = createContext<PaneRouterContextValue>({
   handleEditReference: () => missingContext(),
   ParameterizedLink: () => missingContext(),
   replaceCurrent: () => missingContext(),
+  closeCurrentAndAfter: () => missingContext(),
   closeCurrent: () => missingContext(),
   duplicateCurrent: () => missingContext(),
   setView: () => missingContext(),
--- a/packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
@@ -2,6 +2,7 @@ import {useToast} from '@sanity/ui'
 import React, {memo, useEffect, useRef} from 'react'
 import {useDocumentPane} from './useDocumentPane'
 import {useDocumentOperationEvent} from 'sanity'
+import {usePaneRouter} from '../../components'
 
 function getOpErrorTitle(op: string): string {
   if (op === 'delete') {
@@ -38,6 +39,7 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
   const {documentId, documentType} = useDocumentPane()
   const event: any = useDocumentOperationEvent(documentId, documentType)
   const prevEvent = useRef(event)
+  const paneRouter = usePaneRouter()
 
   useEffect(() => {
     if (!event || event === prevEvent.current) return
@@ -65,8 +67,16 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
       })
     }
 
+    /**
+     * If the document was deleted successfully, close the pane.
+     * This lives here instead of in the DocumentPaneProvider to avoid race conditions.
+     */
+    if (event.type === 'success' && event.op === 'delete') {
+      paneRouter.closeCurrentAndAfter()
+    }
+
     prevEvent.current = event
-  }, [event, pushToast])
+  }, [event, paneRouter, pushToast])
 
   return null
 })
--- a/packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
@@ -44,6 +44,8 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
   useEffect(() => {
     if (!event || event === prevEvent.current) return
 
+    let cleanupId: number | undefined
+
     if (event.type === 'error') {
       pushToast({
         closable: true,
@@ -69,13 +71,16 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
 
     /**
      * If the document was deleted successfully, close the pane.
-     * This lives here instead of in the DocumentPaneProvider to avoid race conditions.
      */
     if (event.type === 'success' && event.op === 'delete') {
-      paneRouter.closeCurrentAndAfter()
+      // Wait until next tick to allow deletion toasts to appear first
+      cleanupId = setTimeout(() => paneRouter.closeCurrentAndAfter(), 0) as any as number
     }
 
     prevEvent.current = event
+
+    // eslint-disable-next-line consistent-return
+    return () => clearTimeout(cleanupId)
   }, [event, paneRouter, pushToast])
 
   return null
--- a/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
@@ -44,7 +44,6 @@ import {
   useDocumentValuePermissions,
   useTimelineStore,
   useTimelineSelector,
-  useDocumentOperationEvent,
   DocumentFieldAction,
   DocumentInspectorMenuItem,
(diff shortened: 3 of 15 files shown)


packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
```diff
@@ -58,15 +60,18 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
     }
 
     if (event.type === 'success' && !IGNORE_OPS.includes(event.op)) {
-      pushToast({
+      // If this is a delete operation, we want to push the toast to the root, since the current pane
+      // will be unmounted, and the toast will be lost
+      const toastFunc = event.type === 'delete' ? rootPushToast : pushToast
```
hermanwikner, line 65: When testing this, I do not get any toast message when deleting a document. It looks like `event` is always `undefined`. Since `DocumentOperationResults` will be unmounted when a document is deleted – will this effect ever run? Do we need to run this is a clean up function in the effect?

sjelfull: I'm pretty sure it was working when I was testing, but it could be a race condition of some sort. I'll look into it.

packages/sanity/src/desk/components/paneRouter/PaneRouterContext.tsx
```diff
@@ -23,6 +23,7 @@ export const PaneRouterContext = createContext<PaneRouterContextValue>({
   handleEditReference: () => missingContext(),
   ParameterizedLink: () => missingContext(),
   replaceCurrent: () => missingContext(),
+  closeToTheRight: () => missingContext(),
```
robinpyon, line 26: **suggestion(non-blocking):**`closeToTheRight` could be renamed `closeCurrentAndAfter` or similar? (layout agnostic and a little clearer that it also closes the _current_ pane)

Alternatively, an argument could be added to `closeCurrent` to facilitate this so we don't need another method

robinpyon (commented):
Thanks @sjelfull! Left a few some comments – mostly around the addition of another provider, and prematurely closing the pane, also happy to pair on this

packages/sanity/src/desk/components/paneRouter/PaneRouterContext.tsx
sjelfull: I agree with renaming it to `closeCurrentAndAfter`. I'd prefer a new method over adding an argument, since the former is more explicit and clear.

bjoerge (commented):
Not sure if this is ready for review given the wip commit, but some general/early feedback:

- I'm struggling to see how the operation lock logic is relevant for closing panes - should it be a separate PR?
- Not sure about the naming of operation hooks. To me it implies that I can register a hook that will be executed before or after a certain operation is happening in order to do some last minute/after the fact operations, but this doesn't seem to be the case. I get that I can _react_ to etiher before or after, but I'm not sure how/or even if it's advisable to do any additional operations that touches the document. Feels a bit scary to e.g. mutate the document in a "before" publish event. Ideally I think we should offer a rather constrained/declarative/non-sideeffectsy API for pre/post operation hooks. I'm curious to learn more about the problem it currently solves and more than happy to help seek out a more speciailized solution.

rexxars:
Converted this to a draft PR until the feedback above is addressed

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -44,13 +44,15 @@ import {
...
   useTimelineSelector,
+  useDocumentOperationEvent,
   DocumentFieldAction,
   DocumentInspectorMenuItem,
   FieldActionsResolver,
   EMPTY_ARRAY,
   DocumentFieldActionNode,
   FieldActionsProvider,
 } from 'sanity'
+import {useRouter} from 'sanity/router'
```
robinpyon, line 55:
```suggestion
```

**nitpick**: remove if it's not being used

robinpyon (commented):
Thanks @sjelfull!

Works beautifully – the only thing I came across is a false positive when creating new documents in the _exceptionally_ rare case when the timeline isn't able to be loaded for whatever reason (e.g. blocked requests). I have a suggestion for this in a [separate PR](https://github.com/sanity-io/sanity/pull/4818).

Will defer to DX on your other questions (though I do think that the logic for closing panes shouldn't be contained with `DocumentOperationResults.tsx`)

sjelfull:
@bjoerge Thanks for the feedback. Went with a simpler approach by setting a value in the DocumentPaneProvider.

robinpyon (approved):

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -616,6 +646,17 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
   }, [connectionState, pushToast])
 
+  /**
+   * If the document was deleted successfully, close the pane.
+   */
+  useEffect(() => {
+    if (!documentOperationEvent) return
+    if (documentOperationEvent.type === 'success' && documentOperationEvent.op === 'delete') {
+      // Wait until next tick to allow deletion toasts to appear first
+      setTimeout(() => paneRouter.closeCurrentAndAfter(), 0)
```
bjoerge, line 656: even though it's unlikely to be used it would be good if we could cancel the timeout in effect cleanup (I'm wary of subtle react strict mode issues)

packages/sanity/src/desk/types.ts
```diff
@@ -251,7 +251,7 @@ export interface PaneMenuItem extends DeskToolMenuItem {
   isDisabled?: boolean
   shortcut?: string
   selected?: boolean
-  tone?: 'primary' | 'positive' | 'caution' | 'critical'
+  tone?: ButtonTone
```
bjoerge, line 254: Is there any benefit to this change? Feels a bit coincidental that it's the same set of tones as the Sanity UI button supports.

sjelfull: You are right. There was a leftover `ButtonTone` imported there (see [`next`](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/desk/types.ts#L3C17-L3C17)), so I thought a merge conflict had lost the `tone` change. Looks like its intentional to not use it there, even though its passed directly to Sanity UI components. I'll just remove this from the PR.

packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -616,6 +646,17 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     }
   }, [connectionState, pushToast])
 
+  /**
+   * If the document was deleted successfully, close the pane.
+   */
+  useEffect(() => {
+    if (!documentOperationEvent) return
+    if (documentOperationEvent.type === 'success' && documentOperationEvent.op === 'delete') {
+      // Wait until next tick to allow deletion toasts to appear first
```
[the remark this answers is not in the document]
sjelfull: Moved it back to the provider because it felt like `DocumentOperationResults` was only concerned with toasts. I believe using a `setTimeout()` does avoid the race condition, so we can do either.

bjoerge (approved):
Great work!

PR #4565, opened 2023-06-07
chore: prototype debug tool for CSM

stipsan:
This is something I threw together just to make it easier to debug CSM in the Test Studio. 
It shouldn't be merged in its current state. But hopefully it gives enough context and examples for @sanity-io/studio-dx to dive into edit intent link e2e testing for visual editing.

When `resultSourceMap` exists in the query response the Query Result render changes to:

image

Sample query:
```
{
  "ptetest": {
    "one": *[_type == "pt_allTheBellsAndWhistles"][0],
    "two": *[_type == "pt_allTheBellsAndWhistles"][1]
  },
  "author": *[_type == "author"][0],
  "book": *[_type == "book"][0],
}
```

Query result is moved to the `result` key like it is transmitted by Content Lake.
The Content Source Map is available on the `resultSourceMap`, to quickly see what information is available to `@sanity/preview-kit/client` when it comes to what kind of affordances we have when trying to differentiate arrays and PTE. 
And finally, `resultEditLinks` uses the `@sanity/preview-kit/client` internals to generate intent links, but instead of using `@vercel/stega` to embed them in strings, it clones `result` and replaces strings that can be source mapped to an edit intent link and replaces it.


stipsan:
@SocketSecurity ignore-all 

These are our own packages

bjoerge:
Don't have any bandwith to look into this, so unassigning myself as reviewer for now - will pick it up over the summer!

PR #4567, opened 2023-06-07
chore(deps): unblock update to @sanity/pkg-utils to v2.3.1

stipsan:
### Description

Updating `@sanity/pkg-utils`, either explicitly as in #4546, or as a side-effect of refreshing `yarn.lock` as seen in #4549, result in this error being thrown:
```bash
yarn depcheck
yarn run v1.22.19
$ lerna exec \$LERNA_ROOT_PATH/scripts/depcheck.js --no-bail --parallel
/home/runner/work/sanity/sanity/node_modules/wide-align/align.js:2
var stringWidth = require('string-width')
                  ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/sanity/sanity/node_modules/string-width/index.js from /home/runner/work/sanity/sanity/node_modules/wide-align/align.js not supported.
Instead change the require of index.js in /home/runner/work/sanity/sanity/node_modules/wide-align/align.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/wide-align/align.js:2:19)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/gauge/render-template.js:2:13)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/gauge/plumbing.js:3:22)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/gauge/index.js:2:16)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/npmlog/log.js:3:13)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/@lerna/cli/index.js:4:13)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/lerna/index.js:3:13)
    at Object.<anonymous> (/home/runner/work/sanity/sanity/node_modules/lerna/cli.js:11:3) {
  code: 'ERR_REQUIRE_ESM'
}
```

Looking at `string-width` it made the breaking change from CJS to Pure ESM in v5. But `wide-align` have a semver range that should be pulling in `v4`:
https://github.com/iarna/wide-align/blob/c1bf09df8a2c549d68a7a0e65315db89d0eff457/package.json#L23-L26

How is it possible that by installing a newer `@sanity/pkg-utils` it somehow installs the wrong version of `string-width`???

There's only two clues to this mystery:
1. The CI throws a strange warning related to `string-width` during `yarn install`: warning `Pattern ["string-width@^1.0.2 || 2 || 3 || 4"] is trying to unpack in the same destination "/home/runner/.cache/yarn/v6/npm-string-width-cjs-4.2.3-269c7117d27b05ad2e536830a8ec895ef9c6d010-integrity/node_modules/string-width-cjs" as pattern ["string-width-cjs@npm:string-width@^4.2.0"].
2. It only happens once `@sanity/pkg-utils` is updated, which pulls in a new version of `rimraf`, which is using a new version of `glob` which uses `@isaacs/cliui` and it deals with the same Pure ESM issue of v5 but in an entirely different way: 
  ``` "dependencies": {
       "string-width": "^5.1.2",
       "string-width-cjs": "npm:string-width@^4.2.0",
     },
   ```

It's my theory that the usage of both `npm:string-width@^4.2.0` and `^1.0.2 || 2 || 3 || 4` is somehow breaking good old yarn v1.
Given that Yarn v1 no longer receives bugfixes I worry about this rabbit hole being the beginning of a trend. Basically any dependency could start adopting the `npm:` prefixed, `-cjs` suffixed pattern and you don't need a whole lot of dependencies doing this before it'll be less time spent migrating away from yarn v1 and over to pnpm or npm workspaces.
For now the `overrides` fix in this PR buys us some time, but it's not a silver bullet and for certain types of dependencies it can cause other problems :sweat:

### What to review

Build output shouldn't have breaking changes, and build performance should be the same as before.

### Notes for release

N/A

Closes #4549 and fixes #4546

cc @mariuslundgard

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
   },
   "resolutions": {
     "**/@types/react": "^18.0.25",
-    "@types/react": "^18.0.25"
+    "@types/react": "^18.0.25",
+    "wide-align/string-width": "^4.2.0"
   },
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.18.6",
@@ -84,7 +85,7 @@
     "@optimize-lodash/rollup-plugin": "^4.0.1",
     "@playwright/test": "^1.31.2",
     "@sanity/client": "^6.1.1",
-    "@sanity/pkg-utils": "^2.1.0",
+    "@sanity/pkg-utils": "^2.3.1",
     "@sanity/tsdoc-to-portable-text": "^0.4.3",
     "@sanity/uuid": "^3.0.1",
     "@types/jest": "^27.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -41,7 +41,12 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.21.7.tgz#61caffb60776e49a57ba61a88f02bedd8714f6bc"
   integrity sha512-KYMqFYTaenzMK4yUtf4EW9wc4N9ef80FsbMtkwool5zpwl4YrT1SdWYSTRcT94KO4hannogdS+LxY7L+arP3gA==
 
-"@babel/core@^7.1.0", "@babel/core@^7.11.1", "@babel/core@^7.12.3", "@babel/core@^7.16.0", "@babel/core@^7.19.6", "@babel/core@^7.20.12", "@babel/core@^7.21.3", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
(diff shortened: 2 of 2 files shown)


rexxars (approved):

binoy14 (approved):
Very detailed description, ran into same issue in another repo now I know the culprit. Thanks!

PR #4568, opened 2023-06-07
chore(ci): replace yarn v2 install flags with v1

stipsan:
### Description

Turns out yarn v1 will let you use whatever flag you want `yarn install --foo-bar`, including ones that are for v2 https://yarnpkg.com/cli/install#options
I migrated them back to v1, `--frozen-lockfile` is roughly the same thing as `--immutable`. 

With that fixed and the CI guarding with `--frozen-lockfile` there's no longer a need for `yarn-if-needed.yml` to check if it should sync the lockfile as PRs that forget to do that will fail the build. Now it runs the same deduping heuristics as recommended by Renovate: https://github.com/scinos/yarn-deduplicate#duplicated-packages
It fixes the Yarn v1 problem with duplicate dependencies piling up over time, without the downsides of often having to recreate the entire lockfile with all that churn.

And there is no equivalent to `--immutable-cache`. While `--check-cache` can be recreated using `--update-checksums --force` but it's very slow, and so I'm only doing it in the `yarn-if-needed.yml` action, which is skipped unless a commit pushed to `main` edits `yarn.lock`. 

### What to review

Everything builds like before, but if you try to sneak in a dependency change in a `package.json` the CI should no longer allow you to merge that PR until `yarn install` sync the lockfile.

### Notes for release

N/A


rexxars (approved):

bjoerge (approved):
Thanks!

PR #4571, opened 2023-06-09
chore: correct toast text for copied URL

SimeonGriggs:
### Description

Fixes casing for "URL" on toast popup when copying an asset's URL to the clipboard

### What to review

That you agree URL should be uppercase?

### Notes for release

Fixes casing for "URL" on toast popup when copying an asset's URL to the clipboard


bjoerge (approved):
ty!

PR #4572, opened 2023-06-09
feat(portable-text-editor): add value snapshot to mutation change

skogsmaskin:
### Description

This will add a snapshot of the current editor value with the mutation change that is emitted when the user edits something.

The PortableTextEditor only emits a value change when the props.value is changed, so this snapshot can be used to set the outside value without needing to perform any patching.

### What to review

* That the snapshot is provided in the mutation event is enabling what is needed to simply set the value on the outside and sending it in through `props.value` again.

### Notes for release
* Added  `snapshot` of the editor value to the mutation changes the Portable Text Editor is emitting.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/components/Synchronizer.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/components/Synchronizer.tsx
@@ -61,9 +61,10 @@ export function Synchronizer(props: SynchronizerProps) {
         finalPatches.length,
         pendingPatches.current.length
       )
-      onChange({type: 'mutation', patches: finalPatches})
+      const editorValue = PortableTextEditor.getValue(portableTextEditor)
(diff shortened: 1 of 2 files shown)


hermanwikner (approved):
Nice, thanks @skogsmaskin! It seems to work very well. I tested it by doing this:

```tsx
import {Schema} from '@sanity/schema'
import {PortableTextEditor, PortableTextEditable, EditorChange} from '@sanity/portable-text-editor'

export function Editor() {
  const [value, setValue] = useState<PortableTextBlock[]>()

  const handleChange = useCallback((change: EditorChange) => {
    if (change.type === 'mutation' && change.snapshot) {
      setValue(change.snapshot)
    }
  }, [])

  return (
    <PortableTextEditor onChange={handleChange} schemaType={schema.get('body')} value={value}>
      <PortableTextEditable />
    </PortableTextEditor>
  )
}

```

bjoerge (approved):
LGTM

PR #4584, opened 2023-06-14
feat: add etl layer

RitaDias:
### Description

this is based on a branch that @mariuslundgard created ([marius/tsdoc-etl](https://github.com/sanity-io/sanity/blob/marius/tsdoc-etl/scripts/etl.ts))

What does this allow for?
- Changes to the setting of sanity docs (such as changing the slugs if we want to) to be done in a layer that is abstracted from the tsdoc package. Meaning faster and easier changes done.
- Allows for running `yarn etl sanity` to create and upload the documents straight from the root (as long as the .env.local) are set up as intended
- Updated the version for tsdoc

### What to review

- Make sure the code looks good
- When running locally (with the .env.local set up the way that we have in the notion but for the package/sanity) that you can make the changes as I described them ☝️ 

### Notes for release

No notes


.github/workflows/etl.yml
```diff
@@ -40,4 +40,4 @@ jobs:
       - name: Extract API docs
         env:
           EXTRACT_SANITY_API_TOKEN: ${{ secrets.EXTRACT_SANITY_API_TOKEN }}
-        run: yarn etl --quiet
+        run: yarn etl sanity
```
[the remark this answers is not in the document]
mariuslundgard: The env variables are not set. I suggest adding a `.env` in the root of the monorepo with:

```sh
# Used by `yarn etl`
EXTRACT_SANITY_PROJECT_ID="<project-id-for-tsdoc>"
EXTRACT_SANITY_DATASET="<dataset-for-tsdoc>"
```

And then change the env variable names in `packages/sanity/tsdoc.config.ts` to:

```ts
import {defineConfig} from '@sanity/tsdoc'
import pkgConfig from './package.config'

export default defineConfig({
  extract: pkgConfig.extract,
  output: {
    sanity: {
      projectId: process.env.EXTRACT_SANITY_PROJECT_ID,
      dataset: process.env.EXTRACT_SANITY_DATASET,
      token: process.env.EXTRACT_SANITY_WRITE_TOKEN,
    },
  },
})
```

.github/workflows/etl.yml
binoy14, line 1: ❓ Should this only run when a merge happens on `current` We probably don't want things getting merged on `next` to appear on the production marketing site.

Follow up: It would be nice to see things in a staging site on the marketing site for things merged into next, would help with writing docs and reviewing as well. Not sure if the marketing site has that ability

scripts/etl.ts
binoy14, line 1: I am curious what is the difference in this file and the once exposed by tsdoc [here](https://github.com/sanity-io/tsdoc/blob/alpha/src/cli/etl/etlCommand.ts) Shouldn't we use a unified approach ?

package.json
```diff
@@ -86,7 +86,7 @@
     "@playwright/test": "^1.31.2",
     "@sanity/client": "^6.1.1",
     "@sanity/pkg-utils": "^2.1.0",
-    "@sanity/tsdoc": "1.0.0-alpha.22",
+    "@sanity/tsdoc": "1.0.0-alpha.25",
     "@sanity/tsdoc-to-portable-text": "^0.4.3",
```
binoy14, line 90: ❓Looks like `@sanity/tsdoc-to-portable-text` is unused, is that needed?

.github/workflows/etl.yml
RitaDias: Yep this should be only in "current" I think 

As for the follow up: I'm not sure either. Maybe as a first step let's write a story (for the next mvi) and bring it up in the channel with Jon? 🤔

scripts/etl.ts
RitaDias: When I asked Marius he mentioned that it would take the code from the TSDoc into account as well (though I could have misunderstood). @mariuslundgard can you just confirm/clarify in case if I'm wrong? 🙂

package.json
RitaDias: I would be for removing it but I'm also unsure if it's being used somewhere hidden. @mariuslundgard any reason to keep it?

mariuslundgard: Should be safe to remove – it's an older incarnation of `@sanity/tsdoc`

scripts/etl.ts
mariuslundgard: The reason for rolling a separate `etl` script in the monorepo, is to do customizations that we don't want to bundle in the general `@sanity/tsdoc` library. E.g. how to deal with custom tags, remove certain members from the APIs, etc.

.github/workflows/etl.yml
RitaDias: Made the change (to make it only run on current)

.github/workflows/etl.yml
```diff
@@ -39,5 +39,7 @@ jobs:
 
       - name: Extract API docs
         env:
+          EXTRACT_SANITY_PROJECT_ID: ${{ secrets.EXTRACT_SANITY_PROJECT_ID }}
+          EXTRACT_SANITY_DATASET: ${{ secrets.EXTRACT_SANITY_DATASET }}
           EXTRACT_SANITY_API_TOKEN: ${{ secrets.EXTRACT_SANITY_API_TOKEN }}
```
RitaDias, line 44: Using current dev secrets

.github/workflows/etl.yml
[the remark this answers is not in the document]
RitaDias: @mariuslundgard thanks for this, it was very helpful!  
I managed to get it working, however, can you just double check that with the changes I've made (especially here https://github.com/sanity-io/sanity/blob/c34f20c3e149c891c5a1067992643f859c9be7bf/.github/workflows/etl.yml#L6) will make sure that it only runs when it's merged into `current`?

Because in our case we don't create a PR when we're releasing, it's a rebase I'm unsure if this would still trigger then? 🤔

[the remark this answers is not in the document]
mariuslundgard: Yes, and you may also use an `if` condition to make sure it will not run if a PR were to be created mistakenly.

However – I’m wondering if what we want is to let it run on every PR, except for the _writing to Sanity_ part. That would be achieved by not including the token env variable. In that case, it will do the extraction and transformation, but not load it to Sanity. This could be a good way of testing that the API is extractable before releasing.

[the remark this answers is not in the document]
RitaDias: > However – I’m wondering if what we want is to let it run on every PR, except for the writing to Sanity part. That would be achieved by not including the token env variable. In that case, it will do the extraction and transformation, but not load it to Sanity. This could be a good way of testing that the API is extractable before releasing.

I would say yes if we currently had a way of easily previewing the data. Because otherwise I'm not sure if there was a clear way to get feedback that it's not been "extracted properly"? I suppose the CI could just complain "I wasn't able to extract `X` api" and we wouldn't need to wait until the release for something to potentially go wrong... 🤔 

That could indeed be cool, let me see what I can do 🤔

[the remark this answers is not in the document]
RitaDias: @mariuslundgard 
👋 I figured out something which works: https://github.com/sanity-io/sanity/blob/411195ad2922a73ed7ec4b495e6258daf2f66c9b/.github/workflows/etl.yml#L42-L44

There's basically a condition that checks if: you're running this on Push AND if the branch. If it's not in current (just doing the usual pushing to a PR) then it will use the `dev_`. otherwise it will use the other secret.
I haven't tested it on the "current" branch, but it does work with the dev information

**Why this over preventing it from uploading all together?** One of the topics we have been talking about is to have a connection with the website to more easily (visually) see the changes. What this means is that we will need to have a set up for a dev env at some point anyway. This feels better 👍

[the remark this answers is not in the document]
bjoerge: Sweet! Next step is to figure out how we can get a render of whats in the dev dataset somewhere for us to have a preview (strictly speaking it doesn't even have to be the website as long as it can show us that it can render nicely, if that makes it easier somehow). But yeah, we can tackle that separately.

Great work @RitaDias!!! ✨

mariuslundgard (approved):
LGTM!

binoy14 (approved):
✅ 🚢

PR #4597, opened 2023-06-16
docs(cli): update docs for getStudioEnvironmentVariables

RitaDias:
### Description

Brush up the tsdoc comments for `getStudioEnvironmentVariables`

The reason why the params obj was moved out into its own interface and exported is due to how some limitations with TSdoc. As it was before there was a huge bulk of text on the `getStudioEnvironmentVariables` which made it hard to quickly parse through.

Now we link to an exported interface (set as public since `getStudioEnvironmentVariables` is also public).

### What to review

Make sure that the information makes sense. The text itself wasn't changed much from the previous iteration, just an example added and some words shuffled

### Notes for release
 
N/A (Not sure if we want to include this in the release)


packages/sanity/exports/cli.ts
```diff
@@ -2,4 +2,5 @@ export {createCliConfig, defineCliConfig, getCliClient} from '@sanity/cli'
 
 export type {CliClientOptions, CliConfig} from '@sanity/cli'
 
+export type {studioEnvVariablesOptions} from '../src/_internal/cli/server/getStudioEnvironmentVariables'
 export {getStudioEnvironmentVariables} from '../src/_internal/cli/server/getStudioEnvironmentVariables'
```
rexxars, line 6: Prefer a single re-export per file:
```suggestion
export {getStudioEnvironmentVariables, type studioEnvVariablesOptions} from '../src/_internal/cli/server/getStudioEnvironmentVariables'
```

packages/sanity/src/_internal/cli/server/getStudioEnvironmentVariables.ts
```diff
@@ -4,29 +4,46 @@ import {loadEnv} from '@sanity/cli'
...
- * in the same way the studio does. A `mode` must be specified, usually `development`
- * or `production`, which will load the corresponding `.env.development` or `.env.production`.
- * To specify where to look for the dotenv files, specify `options.envFile.envDir`.
+ * The params for the `getStudioEnvironmentVariables` function that gets Studio focused environment variables.
  *
- * Specify `options.jsonEncode` to JSON-encode the values, which is handy if you want to pass
- * this to a bundlers hardcoded defines, such as Vite's `define` or Webpack's `DefinePlugin`.
+ * @public
+ */
+export interface studioEnvVariablesOptions {
```
rexxars, line 11: Interfaces should be PascalCase
```suggestion
export interface StudioEnvVariablesOptions {
```

packages/sanity/src/_internal/cli/server/getStudioEnvironmentVariables.ts
```diff
@@ -4,29 +4,46 @@ import {loadEnv} from '@sanity/cli'
...
+   * this to a bundlers hardcoded defines, such as Vite's `define` or Webpack's `DefinePlugin`.
+   */
+  jsonEncode?: boolean
+}
+
+/**
+ * Get environment variables prefixed with SANITY_STUDIO_, as an object.
  *
  * @param options - Options for the environment variable loading
+ *  {@link studioEnvVariablesOptions}
```
rexxars, line 35: Do we need this? I thought TSDoc was smart enough to know which type it was, and thus automatically link it without a specific mention of the type?

rexxars (changes requested):

packages/sanity/src/_internal/cli/server/getStudioEnvironmentVariables.ts
binoy14: It does not automatically link to it. The `@link` tag would be needed for anything that we want to link it to.

binoy14: Also updating the link when the interface is changed 
```suggestion
 *  {@link StudioEnvVariablesOptions}
```

packages/sanity/exports/cli.ts
```diff
@@ -2,4 +2,4 @@ export {createCliConfig, defineCliConfig, getCliClient} from '@sanity/cli'
 
 export type {CliClientOptions, CliConfig} from '@sanity/cli'
 
-export {getStudioEnvironmentVariables} from '../src/_internal/cli/server/getStudioEnvironmentVariables'
+export {getStudioEnvironmentVariables, type studioEnvVariablesOptions} from '../src/_internal/cli/server/getStudioEnvironmentVariables'
```
binoy14, line 5: I think linter is complaining about spacing here

bjoerge (approved):
LGTM!

binoy14 (approved):
✅

PR #4600, opened 2023-06-16
fix(core): a11y issue in command list

hermanwikner:


packages/sanity/src/desk/panes/documentList/DocumentListPaneContent.tsx
```diff
@@ -224,7 +226,7 @@ export function DocumentListPaneContent(props: DocumentListPaneContentProps) {
         <CommandListBox>
           <CommandList
             activeItemDataAttr="data-hovered"
-            ariaLabel="Document list"
+            ariaLabel={`List of ${paneTitle} documents`}
```
bjoerge, line 229: Is `paneTitle` the right name here?

I'm reading this as a way that I can specify the full pane title, when in reality I can only specify a portion of it. So if I pass, let's say _All published documents_, it will read _List of All published documents documents_. I suggest either renaming the property or making the default `List of ${documentType} documents`

bjoerge (changes requested):

packages/sanity/src/core/components/commandList/CommandList.tsx
```diff
@@ -271,10 +271,12 @@ export const CommandList = forwardRef<CommandListHandle, CommandListProps>(funct
     const activeIndex = activeIndexRef?.current
     if (items.length > 0) {
       inputElement?.setAttribute('aria-activedescendant', getChildDescendantId(activeIndex))
+      virtualListElement?.setAttribute('aria-activedescendant', getChildDescendantId(activeIndex))
     } else {
       inputElement?.removeAttribute('aria-activedescendant')
+      virtualListElement?.setAttribute('aria-activedescendant', getChildDescendantId(activeIndex))
```
robinpyon, line 277:
```suggestion
      virtualListElement?.removeAttribute('aria-activedescendant')
```

**suggestion**: clear the attribute if no items are present

bjoerge (approved):
Thank you @hermanwikner!

packages/sanity/src/core/components/commandList/CommandList.tsx
```diff
@@ -567,7 +569,12 @@ export const CommandList = forwardRef<CommandListHandle, CommandListProps>(funct
     inputElement?.setAttribute('aria-expanded', 'true')
     inputElement?.setAttribute('aria-controls', getCommandListChildrenId())
     inputElement?.setAttribute('role', 'combobox')
-  }, [getCommandListChildrenId, inputElement])
+
+    virtualListElement?.setAttribute('aria-autocomplete', 'list')
+    virtualListElement?.setAttribute('aria-expanded', 'true')
+    virtualListElement?.setAttribute('aria-controls', getCommandListChildrenId())
+    virtualListElement?.setAttribute('role', 'combobox')
+  }, [getCommandListChildrenId, inputElement, virtualListElement])
```
robinpyon, line 577:
```suggestion
  }, [getCommandListChildrenId, inputElement])
```

**suggestion**: remove the addition of aria attrs to `virtualListElement`

I don't believe these additions are necessary, as these aria attributes are specific to the input / combobox itself. From what I can tell it should work as expected with these removed – I've noticed that with these enabled, OS X VoiceOver will erroneously report on each item twice when navigating, e.g.
- "Item, Selected, 12/17"
- "Item, Clickable"

When it should just return once, i.e.
- "Item, Selected, 12/17"

robinpyon (changes requested):
Thanks for the quick turnaround @hermanwikner! Added a few minor comments, nothing significant

robinpyon (approved):

PR #4606, opened 2023-06-19
fix(core): improve PTE popover behavior

mariuslundgard:
### Description

This PR updates how popovers are rendered and behave, making use of the new props in`@sanity/ui`'s `Popover`: `floatingBoundary` and `referenceBoundary`.

- `referenceBoundary` is the HTML element that is used as boundary for the element which the popover points to. The annotation or inline object in the case of PTE. Specifying this boundary will make the popover disappear when an annotation or inline object is scrolled out of view (i.e. out the scrollable editable area).
- `floatingBoundary` is the HTML element that is used as boundary for the popover itself. When this is set to the document pane's scrollable area, the popover will grow outside of the PTE when needed (which is good!).

Before:
image

After:
image

### What to review

- That PTE popovers are accessible.
- That PTE popovers respect the `options.width` property from the schema definition.

### Notes for release

- Improves behavior of popover in the Portable Text editor.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
@@ -9,7 +9,7 @@ import {
   BlockAnnotationRenderProps,
 } from '@sanity/portable-text-editor'
 import {Path, PortableTextBlock, PortableTextTextBlock} from '@sanity/types'
-import {Box, Portal, PortalProvider, usePortal} from '@sanity/ui'
+import {Box, Portal, PortalProvider, useBoundaryElement, usePortal} from '@sanity/ui'
 import {ArrayOfObjectsInputProps, RenderCustomMarkers} from '../../types'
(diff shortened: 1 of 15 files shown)


snorrees (approved):
Various tests in the test studio 

Validation ✅ :
image
Inline edit popover ✅ 
image
Inline editor ✅ 
image
Block menu  ✅ 
image

Are these in a/the latest canary btw? I think some might have had issues in the inspector PTE in assist (validation comes to mind), but need to double check.

I know I tested modal width from schema config in assist ok before.

mariuslundgard:
@snorrees 
> Are these in a/the latest canary btw? I think some might have had issues in the inspector PTE in assist (validation comes to mind), but need to double check.

I think the validation tooltips have been fixed on the `next` in meantime.

skogsmaskin (approved):
Nice to have the `floatingBoundary` as well here. LGTM!

PR #4607, opened 2023-06-20
fix: add missing tags

RitaDias:
### Description

Fix issue with the ESLinting (on next)
https://github.com/sanity-io/sanity/actions/runs/5320741764/jobs/9634918275#logs

@rexxars 
Main question I have about the workflow (Should the linter fix? 🤔) is if there is a reason why we shouldn't have this running on each Push? When merging the tsdoc changes into next we got an error from the linting because of missing tags, but these would have be useful to have as we push. Thoughts? :thinking_face:

### What to review

Make sure the question asked ☝️ is answered and that we're ok with it

### Notes for release

n/a

### DIFF
--- a/packages/sanity/tsdoc.json
+++ b/packages/sanity/tsdoc.json
@@ -20,6 +20,8 @@
     "@alpha": true,
     "@param": true,
     "@returns": true,
-    "@remarks": true
+    "@remarks": true,
+    "@throws": true,
+    "@defaultValue": true
   }
 }


bjoerge:
But yeah, I'd prefer to have it check PRs as well so it can fix eslint errors _before_ they land in next, so +1 for this.

bjoerge (approved):

RitaDias:
> I'd prefer https://github.com/sanity-io/sanity/blob/next/.github/workflows/lintChanged.yml to be fixed. `lint-fix-if-needed.yml` is designed to catch cases that existing linter CI tooling won't, such as:
> 
> * a PR that updates an ESLint plugin or preset that includes a new rule that is enabled by default and supports `--fix`.
> * a PR that changes the ESLint configuration to enable a new rule that has `--fix`.
> 
> I believe we have another workflow that catches everything `lintChanged.yml` doesn't, but don't remember its name.
> 
> So to reiterate:
> 
> 1. on PRs we have a fast pass that runs `lintChanged.yml` only on changed files that takes much shorter time to run.
> 2. On commits to `next` we have an action (don't remember its `.yml` name) that lints everything, it catches everything that `lintChanged.yml` can't (like config changes and dependency updates).
> 3. Should any error or warning catched in step 2. be possible for a bot to do for you and create a PR to ease your workflow, then `lint-fix-if-needed.yml` will do that.

I'm pretty new on this workflow things but my main question (for these sort of instances) apart from fixing the lintChanged.yml is that, from my perception, this is not something that the eslint would be able to fix automatically right?

stipsan:
>I'm pretty new on this workflow things but my main question (for these sort of instances) apart from fixing the lintChanged.yml is that, from my perception, this is not something that the eslint would be able to fix automatically right?

Correct, it would not. The `lint-fix-if-needed.yml` action should only run on pushes to `next`, and it failing should continue to be treated as an unexpected event.
We already run `lintChanged.yml` on PRs, which is `ESLint but fast`. In this PR, by changing `lint-fix-if-needed.yml` to also run on PRs, we now have to wait for `ESLint the slowest possible way` before merging and `lintChanged.yml` no longer have a purpose.

A failing build of `lint-fix-if-needed.yml` is a symptom of `lintChanged.yml` allowing PRs to merge that it shouldn't.
I don't have bandwidth to suggest a full fix for `lintChanged.yml`, but high-level this is how it should flow:
- it's safe to only diff changed files when:
	- there are no changes to `yarn.lock` (which might cause `ESLint` plugins or presets to have changed)
	- no changes to the `.eslintignore` file
	- no changes to a `.eslintrc` file
	- no changes to a `.eslintrc.cjs` file
- if unsafe, lint everything on the PR, even though it's slow, to guard linter errors from sneaking into `main`.

Until `lintChanged.yml` is fixed I'd recommend anyone touching ESLint configurations and dependencies have to manually make sure there are no linter errors before merging.

binoy14 (approved):

RitaDias:
> > I'm pretty new on this workflow things but my main question (for these sort of instances) apart from fixing the lintChanged.yml is that, from my perception, this is not something that the eslint would be able to fix automatically right?
> 
> Correct, it would not. The `lint-fix-if-needed.yml` action should only run on pushes to `next`, and it failing should continue to be treated as an unexpected event. We already run `lintChanged.yml` on PRs, which is `ESLint but fast`. In this PR, by changing `lint-fix-if-needed.yml` to also run on PRs, we now have to wait for `ESLint the slowest possible way` before merging and `lintChanged.yml` no longer have a purpose.
> 
> A failing build of `lint-fix-if-needed.yml` is a symptom of `lintChanged.yml` allowing PRs to merge that it shouldn't. I don't have bandwidth to suggest a full fix for `lintChanged.yml`, but high-level this is how it should flow:
> 
> * it's safe to only diff changed files when:
>   
>   * there are no changes to `yarn.lock` (which might cause `ESLint` plugins or presets to have changed)
>   * no changes to the `.eslintignore` file
>   * no changes to a `.eslintrc` file
>   * no changes to a `.eslintrc.cjs` file
> * if unsafe, lint everything on the PR, even though it's slow, to guard linter errors from sneaking into `main`.
> 
> Until `lintChanged.yml` is fixed I'd recommend anyone touching ESLint configurations and dependencies have to manually make sure there are no linter errors before merging.

Thank you so much for the suggestion 🙏 
I don't have time for handling it right now, but I am adding this to a shortcut story so that we can tackle it afterwards :) 🙏

PR #4609, opened 2023-06-20
feat(cli): print all current user roles in debug output

rexxars:
### Description

Currently, the `sanity debug` command prints the value of `user.role`, which can only ever contain the built-in roles we used to support, or the string "custom". This is not great for debugging, since a user can now have a number of different roles.

This PR changes to use a newer API version for the project info request, which includes the full `roles` array. Additionally, it moves the `User role` from the project info to the user info, since this makes more sense to me.

Note that we only print the name (id) of the roles, not the descriptions. I think this usually good enough to debug with.

### What to review

- `sanity debug` in global context works (skips outputting the `Roles` key)
- `sanity debug` in project context works (outputs the array of user roles)

### Notes for release

- `sanity debug` now outputs all the roles of the currently authenticated user


binoy14 (approved):

RitaDias (approved):
🚢

PR #4612, opened 2023-06-21
docs(sanity): structure/desktool items docs

RitaDias:
### Description
 
it ain't much but it's honest work

- [x] Write docs for desktool.
All apis moved from `@beta` to `@public` were done so because otherwise the five items that we had initially set as public wouldn't have made sense to exist and be exposed. 

### What to review

- These were meant to be 5 items which turned into 80ish when moving the needed stuff from `@beta` to `@public` :eye: 
- What does this mean? It means that a lot of items were moved because of tsdoc requesting the tags to match. Please review because some items might not make sense to be exposed in the docs website itself (so, while remaining `@public` should be `@hidden`)

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/templates/types.ts
+++ b/packages/sanity/src/core/templates/types.ts
@@ -182,6 +182,7 @@ export interface TemplateItem {
    *
    * @experimental
    * @beta
+   * @hidden
    */
   initialDocumentId?: string
 
--- a/packages/sanity/src/core/templates/types.ts
+++ b/packages/sanity/src/core/templates/types.ts
@@ -57,11 +57,11 @@ export interface InitialValueTemplateItem extends TemplateResponse {
 }
 
 /**
- * Type for template response
+ * Interface for template response
  *
  * @public
  */
-export type TemplateResponse = {
+export interface TemplateResponse {
   /** Template id */
   templateId: string
   /** Template title */
--- a/packages/sanity/src/core/templates/types.ts
+++ b/packages/sanity/src/core/templates/types.ts
@@ -48,9 +48,7 @@ export type TemplateArrayFieldDefinition = TemplateFieldDefinition & {
  *
  * @public
  */
-export interface InitialValueTemplateItem extends TemplateResponse {
-  /** Initial value id */
-  id: string
+export interface InitialValueTemplateItem extends TemplateItem {
   type: 'initialValueTemplateItem'
   /** Initial value schema type */
   schemaType: string
@@ -61,8 +59,10 @@ export interface InitialValueTemplateItem extends TemplateResponse {
  *
  * @public
  */
-export interface TemplateResponse {
-  /** Template id */
+export interface TemplateItem {
+  /**
+   * ID for the template the item references
+   */
   templateId: string
   /** Template title */
   title?: string
--- a/packages/sanity/src/core/templates/types.ts
+++ b/packages/sanity/src/core/templates/types.ts
@@ -1,20 +1,78 @@
-import {InitialValueProperty, SchemaType} from '@sanity/types'
+import type {InitialValueProperty, SchemaType} from '@sanity/types'
+import type {ElementType, ReactElement} from 'react'
 
-/** @public */
+/**
+ * An initial value template is a template that can be used to create a new documents.
+ *
+ * This allows a document type to have multiple different starting values while having the same
+ * shared schema definition. Using parameters allows for dynamic template values.
+ *
+ * As the name implies, these are _initial_ values, not _default_ values. The distinction is that
+ * the initial value is only set when the document is created - it is not "merged" into existing
+ * documents that may lack values for fields.
+ *
+ * All document types will by default (automatically, behind the scenes) have an initial value
+ * template generated for them, which will have the same ID as the schema type name. The value of
+ * this template will be the value of the `initialValue` property on the schema type definition,
+ * or an empty object if none is set.
+ *
+ * @public
+ */
 export interface Template<Params = any, Value = any> {
+  /**
+   * Template ID. Automatically generated templates will have the same ID as the schema type name.
+   */
   id: string
+
+  /**
+   * Template title.
+   */
   title: string
-  description?: string
+
+  /**
+   * Schema type name the template belongs to. For the automatically generated templates,
+   * this will be equal to the `id` property.
+   */
   schemaType: string
+
+  /**
+   * Template icon. Rendered in places such as the "new document" dialog. Optional.
+   * Inferred from the schema type icon if not set.
+   */
   icon?: SchemaType['icon']
+
+  /**
+   * Value to use as initial value. Can either be a static object value, or a function that
+   * resolves _to_ an object value. If using a function, it can be given a set of parameters,
+   * which can then determine the value that is returned.
+   */
   value: InitialValueProperty<Params, Value>
+
+  /**
+   * Array of parameters the template accepts. Currently not used (any parameters are accepted),
+   * but by defining parameters, the templates that require parameters can be identified and
+   * excluded from UIs that do not provide them.
+   */
   parameters?: TemplateParameter[]
+
+  /**
+   * Template description. Rendered in places such as the "new document" dialog. Optional.
+   *
+   * @deprecated No longer used
+   */
+  description?: string
 }
 
-/** @public */
+/**
+ * Parameter for a template. Closely resembles API used to define fields for object schema types.
+ *
+ * @public
+ */
 export type TemplateParameter = TemplateFieldDefinition | TemplateArrayFieldDefinition
 
-/** @public */
+/**
+ * @public
+ */
 export interface TypeTarget {
   type: string
 }
@@ -25,17 +83,40 @@ export interface TemplateReferenceTarget {
   to: TypeTarget | TypeTarget[]
 }
 
-/** @public */
+/**
+ * Field definition for a template parameter.
+ * Closely resembles API used to define fields for object schema types.
+ *
+ * @public
+ */
 export interface TemplateFieldDefinition {
+  /**
+   * Parameter name. Must be unique within the template.
+   */
   name: string
+
+  /**
+   * Parameter type, eg `string`, `number`, `boolean` etc.
+   */
   type: string
+
+  /**
+   * Parameter type. Will be attempted to be automatically set if not given,
+   * by title-casing the `name` property.
+   */
   title?: string
+
+  /**
+   * Description for the parameter. Optional.
+   * May be used in the future to explain the parameter in UIs.
+   */
   description?: string
+
+  /**
+   * Optional bag of options for the parameter. Currently unused.
+   */
   options?: {[key: string]: any}
 }
-
-/** @internal */
-export type ReferenceFieldDefinition = TemplateFieldDefinition & TemplateReferenceTarget
 
 /** @public */
 export type TemplateArrayFieldDefinition = TemplateFieldDefinition & {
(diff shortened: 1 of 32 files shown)


packages/sanity/src/desk/structureBuilder/Document.ts
```diff
@@ -34,60 +34,110 @@ const createDocumentChildResolver =
...
 }
 
 /**
- * @hidden
- * @beta */
+ * A `DocumentBuilder` is used to build a document node.
+ *
+ * @public */
 export class DocumentBuilder implements Serializable<DocumentNode> {
+  /** component builder option object */
```
binoy14, line 79:
```suggestion
  /** component builder option object See {@link PartialDocumentNode} */
```

packages/sanity/src/desk/structureBuilder/DocumentList.ts
```diff
@@ -48,90 +48,149 @@ const createDocumentChildResolverForItem =
...
 
 /**
- * @hidden
- * @beta */
+ * Type for partial document list
+ *
+ * @public
+ */
 export interface PartialDocumentList extends BuildableGenericList {
+  /** Document list options */
```
binoy14, line 56:
```suggestion
  /** Document list options. See {@link DocumentListOptions} */
```

packages/sanity/src/desk/structureBuilder/DocumentList.ts
```diff
@@ -48,90 +48,149 @@ const createDocumentChildResolverForItem =
...
 
 /**
- * @hidden
- * @beta */
+ * Interface for document list input
+ *
+ * @public
+ */
 export interface DocumentListInput extends GenericListInput {
+  /** Document list options */
```
binoy14, line 68:
```suggestion
  /** Document list options. See {@link DocumentListOptions} */
```

packages/sanity/src/desk/structureBuilder/DocumentList.ts
```diff
@@ -48,90 +48,149 @@ const createDocumentChildResolverForItem =
...
 /**
- * @hidden
- * @beta */
+ * Interface for document list
+ *
+ * @public
+ */
 export interface DocumentList extends GenericList {
   type: 'documentList'
+  /** Document list options */
```
binoy14, line 79:
```suggestion
  /** Document list options. See {@link DocumentListOptions} */
```

packages/sanity/src/desk/structureBuilder/GenericList.ts
```diff
@@ -22,132 +22,224 @@ export const shallowIntentChecker: IntentChecker = (intentName, params, {pane, i
...
+ * Interface for base generic list
+ *
+ * @public
+ */
 export interface BaseGenericList extends StructureNode {
+  /** List layout key */
   defaultLayout?: PreviewLayoutKey
+  /** Can handle intent */
   canHandleIntent?: IntentChecker
+  /** List display options */
```
binoy14, line 43:
```suggestion
  /** List display options. See {@link ListDisplayOptions} */
```

packages/sanity/src/desk/structureBuilder/GenericList.ts
```diff
@@ -22,132 +22,224 @@ export const shallowIntentChecker: IntentChecker = (intentName, params, {pane, i
...
- * @hidden
- * @beta */
+ * Interface for base generic list
+ *
+ * @public
+ */
 export interface BaseGenericList extends StructureNode {
+  /** List layout key */
   defaultLayout?: PreviewLayoutKey
+  /** Can handle intent */
```
binoy14, line 41:
```suggestion
  /** Can handle intent. See {@link IntentChecker} */
```

packages/sanity/src/desk/structureBuilder/GenericList.ts
```diff
@@ -22,132 +22,224 @@ export const shallowIntentChecker: IntentChecker = (intentName, params, {pane, i
...
   title: string
+  /** Input menu items */
   menuItems?: (MenuItem | MenuItemBuilder)[]
+  /** Input menu items groups */
   menuItemGroups?: (MenuItemGroup | MenuItemGroupBuilder)[]
+  /** Input initial value */
   initialValueTemplates?: (InitialValueTemplateItem | InitialValueTemplateItemBuilder)[]
+  /** Input default layout */
   defaultLayout?: PreviewLayoutKey
+  /** If input can handle intent */
```
binoy14, line 98:
```suggestion
  /** If input can handle intent. See {@link IntentChecker} */
```

packages/sanity/src/desk/structureBuilder/Intent.ts
```diff
@@ -2,45 +2,61 @@ import {PartialDocumentList, getTypeNamesFromFilter} from './DocumentList'
...
   template?: string
 }
 
 /** @internal */
 export const DEFAULT_INTENT_HANDLER = Symbol('Document type list canHandleIntent')
 
 /**
- * @hidden
- * @beta */
+ * Intent parameters
```
binoy14, line 28:
```suggestion
 * Intent parameters
 * See {@liink BaseIntentParams} and {@link IntentJsonParams}
```

packages/sanity/src/desk/structureBuilder/Intent.ts
```diff
@@ -2,45 +2,61 @@ import {PartialDocumentList, getTypeNamesFromFilter} from './DocumentList'
...
+ */
 export type IntentJsonParams = {[key: string]: any}
 
 /**
- * @hidden
- * @beta */
+ * Type for base intent parameters
+ *
+ * @public */
 export type BaseIntentParams = {
```
binoy14, line 15: could be converted to interface

```suggestion
export interface BaseIntentParams {
```

packages/sanity/src/desk/structureBuilder/Intent.ts
```diff
@@ -2,45 +2,61 @@ import {PartialDocumentList, getTypeNamesFromFilter} from './DocumentList'
...
 /**
- * @hidden
- * @beta */
+ * Interface for intents
+ * @public */
 // TODO: intents should be unified somewhere
 export interface Intent {
+  /** Intent type */
   type: string
+  /** Intent parameters */
```
binoy14, line 41:
```suggestion
  /** Intent parameters. See {@link IntentParams} */
```

binoy14 (commented):
Thanks for making this massive change! I made some comments on some links and converting `types` to `interfaces`. I possibly might have also missed some things that could be useful to have links on. 

I am gonna leave the actual content review to someone who actually knows this more

packages/sanity/src/desk/structureBuilder/Component.ts
```diff
@@ -19,93 +21,164 @@ export interface Component extends StructureNode {
...
+  /** Component */
   component?: UserComponent
+  /** Component child */
   child?: Child
+  /** Component options */
   options?: {[key: string]: unknown}
+  /** Component menu items */
   menuItems?: (MenuItem | MenuItemBuilder)[]
+  /** Component menu item groups */
   menuItemGroups?: (MenuItemGroup | MenuItemGroupBuilder)[]
```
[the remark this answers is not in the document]
RitaDias: It's all good, thanks for bringing these things up. I wanted to update these yesterday but body said no 📉 
This also helped me see that I missed other places. thank you 🙏

packages/sanity/src/core/templates/types.ts
```diff
@@ -44,25 +44,38 @@ export type TemplateArrayFieldDefinition = TemplateFieldDefinition & {
...
   subtitle?: string
+  /** template description */
   description?: string
+  /** template parameters */
   parameters?: {[key: string]: any}
+  /** template icon */
   icon?: React.ElementType | React.ReactElement
   /**
+   * The id of the document to use as initial value
    * @experimental
```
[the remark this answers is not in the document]
RitaDias: Good point, I see that it's rendering it normally, I'm going to add a banner there the same way that we have @beta 🤔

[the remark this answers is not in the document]
RitaDias: Added it on this commit (added link to my async checkpoint)

packages/sanity/src/desk/structureBuilder/GenericList.ts
```diff
@@ -22,132 +22,224 @@ export const shallowIntentChecker: IntentChecker = (intentName, params, {pane, i
...
 }
 
 /**
- * @hidden
- * @beta */
+ * Interface for buildable generic list
+ *
+ * @public
+ */
 // Used internally in builder classes to make everything optional
```
RitaDias, line 71: Given this comment not sure if it should be changed to @internal

packages/sanity/src/desk/structureBuilder/Intent.ts
```diff
@@ -9,10 +9,10 @@ import {StructureNode} from './StructureNodes'
 export type IntentJsonParams = {[key: string]: any}
 
 /**
- * Type for base intent parameters
+ * Interface for base intent parameters
  *
  * @public */
-export type BaseIntentParams = {
```
binoy14, line 12: Looked at the reason why the typecheck is failing and it seems to be cause of this change from `type` to `interface` My guess is because 

`export type IntentParams = BaseIntentParams | [BaseIntentParams, IntentJsonParams]` 

and `export type IntentParameters =
  | Record<string, unknown>
  | [Record<string, unknown>, Record<string, unknown>]` are same when the sub type is a type but if it's an interface it's not equal. I don't know why it does it but someone with more TS experience might know

packages/sanity/src/core/templates/types.ts
```diff
@@ -25,58 +83,117 @@ export interface TemplateReferenceTarget {
...
+  icon?: ElementType | ReactElement
+
   /**
-   * The id of the document to use as initial value
+   * Experimental: not fully supported yet
+   * Hints at what the document ID for the new document should be.
+   * Leave undefined to let the system decide.
+   *
    * @experimental
+   * @beta
```
RitaDias, line 184: @rexxars should we add @hidden on this as well? Or are we ok with keeping it beta & experimental?

packages/sanity/tsdoc.json
```diff
@@ -5,6 +5,11 @@
       "tagName": "@hidden",
       "syntaxKind": "block",
       "allowMultiple": true
+    },
+    {
+      "tagName": "@todo",
+      "syntaxKind": "block",
+      "allowMultiple": true
```
RitaDias, line 12: Not sure if we need this particular block since, as far as I understand it, we only need to add this to tsdoc (have it collected) if we intend on rendering it in the front end. I think that adding it in the "supportForTags" is probably enough

packages/sanity/src/desk/structureBuilder/Intent.ts
```diff
@@ -19,6 +19,12 @@ export interface BaseIntentParams {
   id?: string
   /* Intent template */
   template?: string
+  /**
+   * Experimental field path
+   * @beta
+   * @experimental
```
RitaDias, line 25: Same comment as I made in a separate place, given that it's @beta & @experimental do we want it showing right now or should we add @hidden ?

packages/sanity/tsdoc.json
RitaDias: While trying something else I realised that this is overall breaking as it is (still doesn't recognise the todo tag 🤔 )

rexxars: > I think that adding it in the "supportForTags" is probably enough

Good catch! Updated

RitaDias:
👋 
Thanks for reviewing this, I made some of the changes but others I didn't (unless Espen or Binoy feel particularly strongly about them as well, in which case I can change them). My line of thinking with the comments as they are now (and speaking with the team) is: the comments will be worked on and don't need to be perfect on the first go.

> ### `TemplateResponse`
> Similar to types, I don't think interface comments need to start with "Interface for ...". In this case could be written for less duplication as:
> 
> > The base interface for initial value templates.
> 

Didn't think the change would be truly make such a huge different that would require this change.

> ```ts
>   /** Template id */
>   templateId: string
>   /** Template title */
>   title?: string
>   /** template subtitle */
>   subtitle?: string
>   /** template description */
>   description?: string
>   /** template parameters */
>   parameters?: {[key: string]: any}
>   /** template icon */
>   icon?: React.ElementType | React.ReactElement
> ```
> 
> The comments on the properties don't add much in addition to the property names. Would skip the comments here, or rewrite so they expand on the property names.
> 

True, they don't, however from a rendering in the front end perspective I think that having some of the comments (even if they are /right now/ not super useful is still better than having nothing (which might bring up the question of: did they even look at this section or did they forget about it?)

> ### `ChildResolverOptions`
> Should we mention that this belongs to the Structure API?
> 
> Same goes for:
> 
> * `ChildResolverOptions`
> * `ItemChild`
> * `ChildObservable`
> * `Component`
> * `ComponentInput`
> * `BuildableComponent`
> * `ComponentBuilder`
> * `DocumentOptions`
> * `PartialDocumentNode`
> * `DocumentBuilder`
> * `PartialDocumentList`
> * `DocumentListInput`
> * etc
> 
> Ideally, all these members would be prefixed with `Structure*` imho.
> 

I would strongly suggest against renaming anything (breaking changes). As for mentioning I don't think we necessarily need to do now, but if we do then I would need to know the items that would be considered part of the Structure API (unless it's just "all of them"?)

> ### `StructureResolverContext`
> > Type for a desk tool context. Extends from ConfigContext.
> 
> could be:
> 
> > Context passed as the second argument to the `structure` resolver. Extends {@link ConfigContext}.

This item is hidden currently, but I've made this change anyway :)

binoy14 (approved):
✅ 🙌

rexxars (approved):
We can iterate on the types after merging - this is a fantastic start

PR #4613, opened 2023-06-21
Add blockEditor definition to BlockDecoratorDefinition

ONGChain:
### Description
Adds a new interface definition `BlockEditorDefinition` to `packages/@sanity/types/src/schema/definition/type/block.ts` to define the blockEditor property. This will allow TypeScript users to simply use the blockEditor property without having to have `as BlockDecoratorDefinition` appended to the declaration.

### What to review
Defining a block field in TypeScript with the blockEditor property. Overall, this is no more than a hotfix with really nothing being affected, as the only changes are to the block type interfaces and the blockEditor property is added as an optional property on the `BlockDecoratorDefinition`.

### Notes for release
Added the optional blockEditor property to the BlockDecoratorDefinition interface.

### DIFF
--- a/packages/@sanity/types/src/schema/definition/type/block.ts
+++ b/packages/@sanity/types/src/schema/definition/type/block.ts
@@ -15,10 +15,17 @@ export interface BlockOptions {
 export interface BlockRule extends RuleDef<BlockRule, any[]> {}
 
 /** @public */
+export interface BlockEditorDefinition {
+  icon: ReactNode | ComponentType
+  render: ReactNode | ComponentType
+}
+
+/** @public */
 export interface BlockDecoratorDefinition {
   title: string
   value: string
   icon?: ReactNode | ComponentType
+  blockEditor?: BlockEditorDefinition
 }
 
 /** @public */


snorrees (approved):
LGTM, but would be nice if you could add extend the existing test for this type.

skogsmaskin:
@ONGChain, thank you for this, and I understand why you want to add this (currently this props is talked about in our documentation).

However, this is now a legacy prop in `v3`, but we still support it so it during the transition period of v3, by taking the new block stuff out of @beta, marking them as @public and updating the documentation. This is something we are currently working on, and I'm sorry this is a bit confusing right now!

The new `v3` way of doing this, would be to set `icon` directly on the decorator type for icons, and use `component` for rendering the actual decorator. This also can be rendering `props.renderDefault(props)` like other render callbacks in v3 does.

```ts
defineArrayMember({
  type: 'block',
  marks: {
    decorators: [
      {
        title: 'Bold',
        value: 'bold',
        icon: BoldIcon,
        component: (props) => (
          // Children is here the text. You could also use `props.renderDefault(props)`, for the default rendering
          <strong style={{fontWeight: 700}}>{props.children}</strong>
        ),
      },
    ],
    ...rest
})

```

We aim to get up to date on the documentation for this very soon as it's actively being worked on right now. Thank you for your patience and your efforts with this PR.

I have to close it though, as we don't want people to use this property in the future, when we move the new stuff out of @beta.

PR #4617, opened 2023-06-21
chore(router): add docs for sanity/router package

binoy14:
### Description

Adds documentation for `sanity/router` package

### What to review

- Accuracy for the documentation
- If more details are needed in anything
- If examples need more details
- Also converts some `type` to `interface` 

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/router/route.ts
+++ b/packages/sanity/src/router/route.ts
@@ -38,6 +38,7 @@ export interface RouteObject {
   /**
    * Creates a new router.
    * Returns {@link Router}
+   * See {@link RouteNodeOptions} and {@link RouteChildren}
    */
   create: (
     routeOrOpts: RouteNodeOptions | string,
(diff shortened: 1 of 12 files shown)


packages/sanity/src/router/route.ts
```diff
@@ -8,27 +8,87 @@ import {decodeParams, encodeParams} from './utils/paramsEncoding'
...
   scope?: string
 }
 
 /**
+ * Interface for the {@link route} object.
+ *
  * @public
  */
-export const route: {
+export interface RouteObject {
```
binoy14, line 37: This probably needs a better name. There is already an interface with name `Route`

packages/sanity/src/router/types.ts
```diff
@@ -36,24 +61,67 @@ export type RouteChildren =
...
 
 /**
  * @public
  */
 export interface Router extends RouterNode {
+  /**
+   * Indicates whether this router is a route.
+   * @internal
+   */
   _isRoute: boolean
```
binoy14, line 93: Not sure if this is actually `@internal` but I marked it assuming the use of `_isRoute` on tsdoc it gets ignored right now. Let me know if I need to change this

mariuslundgard:
## [WIP] Content comments

### Comments on React component parameters

I feel like some of these comments are superfluous. It's the `props` to give to the component, but that's a React thing.

Also – it would be more accurate as "Props to pass to the `Link` component."

image

rexxars (approved):

RitaDias (approved):
Let's goooo :shipit:

PR #4618, opened 2023-06-21
test(e2e): retry once on failures, record video on retry

rexxars:
### Description

When the end to end tests fail, it can be hard to debug what is actually causing it to fail - especially when it fails only on CI but not locally. This PR adds retrying on failures, and when that happens - also enables tracing and video recording.

### What to review

- Config looks correct

### Notes for release

None

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -37,6 +37,8 @@ export default defineConfig({
   /* Maximum time one test can run for. */
   timeout: 30 * 1000,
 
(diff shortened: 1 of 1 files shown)


binoy14 (approved):
Curious on where do the actual videos get stored.

rexxars:
> Curious on where do the actual videos get stored.

They get stored in the test output directory. On CI they should be attached to the run as artifacts.

PR #4619, opened 2023-06-22
chore(perf): add support for --excludeTestIds and fail hard on unknown test-id

bjoerge:
### Description

Adds support for excluding one or more tests, and fails the test runner if an unknown test-id is given for either `--testIds` or 
`--excludeTestIds`

### Notes for release

N/A internal tooling


RitaDias (approved):
🚀

PR #4637, opened 2023-06-25
Add the ability for VScode to auto complete the IntrinsicTypeName

ickynavigator:
### Description

This PR introduces the ability to use vscode auto complete in the defineField function (specifically the type option passed in).

### What to review

Trying the autocomplete function on vscode as well other IDEs.
```ts
type IntrinsicTypeName = "string" | "number" | "boolean" | "object" | "image" | "text" | "array" | "block" | "date" | "datetime" | "document" | "file" | "geopoint" | "reference" | "crossDatasetReference" | "slug" | "url" | "email"

type TType = IntrinsicTypeName | (string & {})

const foo = (a: TType) => {}
function bar(a: TType) {}
function foobar<T extends TType>(a: T) {}
function foobarfoo<T extends IntrinsicTypeName | (string & {})>(a: T) {}

const testString: TType = ""
foo('string');
bar('email');
foobar('')
foobarfoo('')
```


packages/@sanity/types/src/schema/define.ts
```diff
@@ -170,7 +170,7 @@ import {FieldDefinitionBase, IntrinsicTypeName} from './definition'
  * @beta
  */
 export function defineType<
-  TType extends string | IntrinsicTypeName, // IntrinsicTypeName here improves autocompletion in _some_ IDEs (not VS Code atm)
+  TType extends IntrinsicTypeName | (string & {}),
```
snorrees, line 173: Awesome that this works, but this seems like is a real head-scratcher. Could you add a comment indicating why it is there/why it is necessary for future maintainers? (without it it could possibly be deleted under the guise of "this is redundant / what does string & {} even mean)

snorrees (changes requested):

packages/@sanity/types/src/schema/define.ts
ickynavigator: I'll add a comment but if I'm being completely honest with you I also don't know why it works 😅

snorrees (approved):

ickynavigator:
> Good stuff, and thanks!

Happy to help!!

rexxars (approved):

rexxars:
I prematurely merged this - apologies. It seems the change is causing some issues. I've removed the commit from `next` and pushed a branch here: https://github.com/sanity-io/sanity/tree/feat/intrinsic-type-name-improved-type

See failing type check run here: https://github.com/sanity-io/sanity/actions/runs/5392798133/jobs/9791642898

ickynavigator:
I'll check it out and see if I can get them to work

snorrees:
I'm sorry too, didnt think to check the tests. (oh how easy it is to forget stuff to the veil of time)

A couple of thoughts after doing a quick looksee:
1) If we can get this to work, this change should be applied to defineField and definedArrayMember
2) We should be able to explain what we really are doing when defining `string & {}`. My gut feeling is that "this can never happen".
3) This change might have to be propagated through ALL the type-narrowing helper types, which, to be honest, I dont think is worth it.
4) When we decide to make TypeScript 5 be minimum version for sanity, we can rewrite the helpers to use const params, which should simplify typings alot.

As is, I'm not sure if this change is safe, unless we have a firm explanation for 2) imo.

snorrees:
> I'll check it out and see if I can get them to work

Seems like this might get fixed in typescript 5.2 🤞

ickynavigator:
> > I'll check it out and see if I can get them to work
> 
> Seems like this might get fixed in typescript 5.2 🤞

You're right. This is the typescript [issue](https://github.com/microsoft/TypeScript/issues/29729)

PR #4638, opened 2023-06-26
chore: dedupe `@sanity/icons` and update snapshot

stipsan:
Follows up on @binoy14 [from](https://github.com/sanity-io/sanity/pull/4599#issuecomment-1607597138):
> the tests are failing. From looking at it, it seems like it's due to @sanity/icons version change. stroke-linejoin="round" is added to the svgs

This PR fixes that. Looks like it's a result from when @mariuslundgard updated `@sanity/icons` but it appears it the lockfile introduced a duplicate of `@sanity/icons`, which is why it didn't break the snapshot test at the time https://github.com/sanity-io/sanity/commit/0b2c40e9697f2712d541ab4f05c99e25b4094940


binoy14 (approved):
Thanks for fixing it  ✅

rexxars (approved):

PR #4646, opened 2023-06-27
fix(core): increase button hit area of collapsible fieldset

ninaandal:
### Description

The hit area for the collapsable field appears tiny with the little arrow (even though you can press the text as well). This PR increases this area. 

### What to review
Screenshot 2023-06-27 at 14 54 22

Expanding the fieldset has an increased hit area without affecting the design. 

### Notes for release
Fixes issue with small hit area for collapsible fieldsets.

### DIFF
--- a/packages/sanity/src/core/form/components/formField/FormFieldSetLegend.tsx
+++ b/packages/sanity/src/core/form/components/formField/FormFieldSetLegend.tsx
@@ -31,6 +31,7 @@ const ToggleButton = styled(Flex).attrs({forwardedAs: 'button'})((props: {theme:
     font: inherit;
     outline: none;
     border-radius: ${rem(radius[1])};
+    position: relative;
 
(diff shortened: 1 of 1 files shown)


hermanwikner (approved):
Nice! 👍

binoy14 (approved):
Curious on the thoughts of making the button width `100%` so clicking anywhere in that line toggles the collapse. Similar to how [Material UI](https://mui.com/material-ui/react-accordion/#basic-accordion) does it?

ninaandal:
> Curious on the thoughts of making the button width `100%` so clicking anywhere in that line toggles the collapse. Similar to how [Material UI](https://mui.com/material-ui/react-accordion/#basic-accordion) does it?

Considering we do not have a cursor and there isn't any indication that it's clickable on the entire width, I won't do this in this PR and leave it to design if we want to update it 😃

PR #4656, opened 2023-06-29
build(etl): improve output of etl command

rexxars:
### Description

When an ETL token was present, we did not output data to the `etc` folder, but the output _said it did_. This was very confusing to me. 

Given that outputting to filesystem takes ~15ms on my computer, I didn't think it was worth doing the either/or approach, and instead we just do both. Open to changing this if there's a reason though.

Additionally, I added some progress indicators with timings.

### What to review

- ETL still runs
- If token is present, outputs to filesystem _and_ remote
- If token is _not_ present, gives a warning to let developer know it is an option

### Notes for release

None.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
     "mkdirp": "^1.0.4",
     "nanoid": "^3.1.30",
     "npm-run-all": "^4.1.5",
+    "ora": "^6.3.1",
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
✅

PR #4668, opened 2023-07-05
fix(core): restore not working on nested panes

binoy14:
### Description

When an item is restored in a nested pane or in a filtered view it does not actually restore after the confirmation and reverts back to the original state. 

### What to review

I am not sure if this is the correct fix honestly but from my testing this seems to fix the issue where we run the complete actions first before navigating to an intent. Please let me know if this makes sense or if there is some better way to solve this

### Notes for release

- Fixes an issue where restore document does not work in nested pane view


RitaDias (commented):
I tested this the same way that the video you had describing the issue was! Looks good to me, but I'm going to wait until Espen ✅ since I also don't have a lot of context in this area :)

RitaDias (approved):
Everything works with the setTimeout in place as well. Re-added Espen for the final ✅

PR #4672, opened 2023-07-06
Feat/range decorations

skogsmaskin:
### Description

This will add support for wrapping a selection inside the PT-input with a custom wrapper component, which comes in handy for search highlighting, presence, validation and other things that are not part of the actual editable content.

image

This PR also brings support to put custom meta data on the validation items, which can be used in combination with this feature, and for any other input that want to reference custom things like selections or regex-match data or whatever.

I have also made the `span` type a core type, and made it extendable with it's own validation etc. Not needed here, so I'll probably pull that out into a separate PR when this is out of draft.

Can be tested in the test studio by navigating to here:
image


rexxars:
@skogsmaskin @christianhg Could you figure out a way forward for this? I'm okay to leave branches if they're helpful/have a future, but it would be nice to close open PRs that are stale

skogsmaskin:
This was released as part of [this PR](https://github.com/sanity-io/sanity/pull/5871)

PR #4685, opened 2023-07-11
fix(core): fix issue with change indicator and path lines

RitaDias:
### Description

As described in a reported bug: String Arrays and Tag arrays weren't showing the change indicator nor the path line in the changes pane. 
This PR adds those as well as fixes other issues with missing path lines in other array inputs.

**Other issues related to the path lines were identified and filed in a separate story and have been left outside of the scope for this PR.**

### What to review

- [x] Array of strings (such as `Array of strings` field in an "Array Test" document) will show the change indicator and path line
- [x] Array of tags (such as `Tags` field in an "Array Test" document) will show the change indicator and path line

These changes also impact the following array (and fix some issues that were identified) such as:

- [x] Array of emails (such as Array of email addresses in an "Array Test" document)
- [x] Array of type string with options list (such as Array of strings with legacy format on lists in an "Array Test" document)
- [x] Array of type strings (with other types) (such as Array with primitive types - string in an "Array Test" document)

### Notes for release

- Array of strings (array of type string) will show the change indicator and path line to the correct changes in the change's pane
- Array of tags (array of type string with layout: 'tags') will show the change indicator and path line to the correct changes in the change's pane

- Array of emails (array of type email) will now show the path line to the correct changes in the change's pane
- Array of strings (array of type string with options.list) will now show the path line to the correct changes in the change's pane


packages/sanity/src/core/form/store/formState.ts
```diff
@@ -791,7 +791,7 @@ function prepareArrayOfPrimitivesInputState<T extends (boolean | string | number
     schemaType: props.schemaType,
     focused: isEqual(props.path, props.focusPath),
     focusPath: trimChildPath(props.path, props.focusPath),
-    path: props.path,
+    path: props.path.concat([0]),
```
RitaDias, line 794: This is is what fixes the broken path lines on the array items 
Screenshot 2023-07-13 at 02 45 46

I was trying to find a neater way (like we have for the review changes in the `BuildChangeList` which actually builds the right paths) of setting this up since I don't know if there is the possibility for complications for "prepareArrayOfPrimitivesInputState" specifically (I don't think that there will be since it doesn't go any "deeper" like into a popover or something) but any pointers would be swell 🥲

binoy14:
Not seeing path lines for strings

Screenshot 2023-07-17 at 8 57 04 AM

as well as for tags

Screenshot 2023-07-17 at 8 57 37 AM

It seems like strings is working on `next` but not in the preview

RitaDias:
> * Array of strings

I'm not sure what's going on but I've just opened the preview and it looks like it works? The pathline was nonexistent when hovering the change indicator from the docs? This video is from the preview 👇 

https://github.com/sanity-io/sanity/assets/6951139/0a1765b5-1335-4773-a2ca-423c6bf46ec8

This is from the next branch 👇 
https://github.com/sanity-io/sanity/assets/6951139/7bbd93c4-4c1e-47f1-8f6a-b20ea77c37aa

binoy14:
Steps to reproduce the issue I am seeing

1. Create a new array test doc. Checkmark cats, notice pathmark works
2. Check multiple items, pathmark works
3. Unselect the other two and only select cats and publish the doc
4. After publish checkmark horses, notice pathmark doesn't work
5. checkmark dogs, pathmark works

RitaDias:
> Steps to reproduce the issue I am seeing
> 
> 1. Create a new array test doc. Checkmark cats, notice pathmark works
> 2. Check multiple items, pathmark works
> 3. Unselect the other two and only select cats and publish the doc
> 4. After publish checkmark horses, notice pathmark doesn't work
> 5. checkmark dogs, pathmark works

This should be fixed with the newest commits :) thanks for bringing it up 🙏

packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx
```diff
@@ -192,27 +196,33 @@ export class ArrayOfPrimitivesInput extends React.PureComponent<
                 >
                   {membersWithSortIds.map(({member, id}, index) => {
                     return (
-                      <Item
+                      <ChangeIndicator
                         key={member.key}
-                        id={id}
-                        sortable={isSortable}
-                        disableTransition={this.state.disableTransition}
+                        path={member.kind === 'item' ? member.item.path : path}
```
RitaDias, line 201: Needed to avoid typing issues. Decided to use the same logic as what we use in line 210

bjoerge: I don't think this will render correctly for error items - think it would be better if you could move the <Item .. element to inside the {member.kind === 'item'} check below and wrap it there instead?

bjoerge (changes requested):

RitaDias:
One thing I'm trying to fix before this is ready.

With the addition of https://github.com/sanity-io/sanity/blob/67e1caed932e9ac18ad69f593811ed169827d9bb/packages/sanity/src/core/form/store/formState.ts#L846

https://github.com/sanity-io/sanity/blob/67e1caed932e9ac18ad69f593811ed169827d9bb/packages/sanity/src/core/form/store/formState.ts#L789

I noticed in instances where an unpublished document adds one array item, then reverts the change, the change indicator doesn't change. (see video)


https://github.com/sanity-io/sanity/assets/6951139/dc2d1709-24e4-42ca-919b-7cd6750f53cf

This is because the "props.changed" which relies on `isChangedValue` in https://github.com/sanity-io/sanity/blob/67e1caed932e9ac18ad69f593811ed169827d9bb/packages/sanity/src/core/form/store/formState.ts#L399 which is returning the wrong values when an array goes from `undefined` to an empty array (array with 0 length).

I'm leaving this here just in case if any from the DX side stumbles upon it while I'm out and is wondering if it's safe. It needs a bit more love just to iron out some edges ✨

RitaDias:
👋 Ready to be reviewed once more :)

binoy14 (approved):
Looks good from my testing now, great work, thanks!

PR #4689, opened 2023-07-12
feat(core): add breadcrumbs to document header

binoy14:
### Description

This is the first pass at implementing the breadcrumbs on the title document header. This is not the complete solution but getting the pieces to iterate on improving this in future tickets. 

#### Things included in this PR
- Show tooltip if text is overflowing
- Clicking on the breadcrumb item navigates to the correct pane

#### Things not included in this PR (Future PRs)
- Dynamic adjustment of breadcrumb size
- Mobile styles
- Opening Reference item in sidebar should show it's structure

Note: The test is skipped and I will come back to it when everything is finalized

### What to review

- Check the breadcrumb in the [workshop](https://test-studio-git-feature-sc-35253.sanity.build/test/workshop/desk;components;Breadcrumbs;normal?scheme=dark&value=eJyrrgUAAXUA%2BQ%3D%3D) 
- The breadcrumb component is somewhat copied from [sanity/ui](https://github.com/sanity-io/ui/blob/main/src/components/breadcrumbs/breadcrumbs.tsx) with tweaking to customize it for the needs of studio.

### Notes for release

N/A (Will be part of larger release announcement)

### DIFF
--- a/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
+++ b/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
@@ -30,7 +30,7 @@ export const BreadcrumbItem = forwardRef(function BreadcrumbItem(
       isTitle={isTitle}
       ref={ref}
       mode="bleed"
-      padding={2}
+      padding={1}
       textAlign="left"
       justify="flex-start"
       {...restProps}
--- a/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
+++ b/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
@@ -1,0 +1,56 @@
+import {Text, Tooltip, Box, Button, type ButtonProps} from '@sanity/ui'
+import React, {forwardRef, type HTMLProps, type ReactNode, type ForwardedRef} from 'react'
+import styled from 'styled-components'
+
+type ParentTextProps = ButtonProps & Omit<HTMLProps<HTMLButtonElement>, 'ref' | 'as'>
+
+/**
+ * @internal
+ * @hidden
+ */
+export interface BreadcrumbItemProps extends ParentTextProps {
+  children: ReactNode
+}
+
+const StyledButton = styled(Button)`
+  width: 100%;
+`
+
+/**
+ * @internal
+ * @hidden
+ */
+export const BreadcrumbItem = forwardRef(function BreadcrumbItem(
+  props: BreadcrumbItemProps,
+  ref: ForwardedRef<HTMLDivElement>
+) {
+  const {children, ...restProps} = props
+
+  return (
+    <StyledButton
+      ref={ref}
+      mode="bleed"
+      padding={1}
+      textAlign="left"
+      justify="flex-start"
+      {...restProps}
+      text={
+        <Tooltip
+          content={
+            <Box padding={2} style={{maxWidth: '500px'}}>
+              <Text muted size={1}>
+                {children}
+              </Text>
+            </Box>
+          }
+          padding={1}
+          placement="bottom"
+          fallbackPlacements={['top', 'left', 'right']}
+          portal
+        >
+          <span>{children}</span>
+        </Tooltip>
+      }
+    />
+  )
+})
--- a/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
+++ b/packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
@@ -1,6 +1,5 @@
-import {Text, Tooltip, Box, type ButtonProps} from '@sanity/ui'
+import {Text, Tooltip, Box, Button, type ButtonProps} from '@sanity/ui'
 import React, {forwardRef, type HTMLProps, type ReactNode, type ForwardedRef} from 'react'
-import styled, {css} from 'styled-components'
 
 type ParentTextProps = ButtonProps & Omit<HTMLProps<HTMLButtonElement>, 'ref' | 'as'>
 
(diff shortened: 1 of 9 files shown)


packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
RitaDias, line 1: Small question, do we want to add a todo directly here about (eventually) moving this to sanity/ui (as I'm assuming that is the end goal of this comment - even if in the distant future? Or perhaps it's not worth it? cc as well @mikolajdobrucki )

packages/sanity/src/desk/components/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,101 @@
...
+            mode="bleed"
+            onClick={open ? collapse : expand}
+            padding={2}
+            ref={setExpandElement}
+            selected={open}
+            text="…"
+          />
+        </Popover>,
+        ...rawItems.slice(len - maxLength, len - afterLength),
+        ...rawItems.slice(len - afterLength),
```
RitaDias, line 77: I have to admit I'm having a hard time following this section, is it to solely keep this at a specific number of breadcrumb items?

RitaDias (commented):
Great job! :) left a few question that popped up as I was reading through and testing things out :)

packages/sanity/src/desk/components/breadcrumbs/Breadcrumbs.tsx
binoy14: Yes, this to support "maxItems" so that items move to a popover once it's over the max items. The logic is mostly taken from [sanity/ui](https://github.com/sanity-io/ui/blob/main/src/components/breadcrumbs/breadcrumbs.tsx#L41-L78) and modified a little bit to support the design case. I will add some comments to this piece to make more sense of it. Thanks for flagging it!

packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <Box padding={2} style={{maxWidth: '500px'}}>
+            <Text size={1}>{children}</Text>
+          </Box>
+        }
+        padding={1}
+        placement="bottom"
+        fallbackPlacements={['top', 'left', 'right']}
+        portal
+        // We only want to show tooltip if the text is overflowing
+        disabled={clientWidth >= scrollWidth}
```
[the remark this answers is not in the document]
binoy14: Two things here that I am assuming is causing this issue. 

1. The breadcrumb is supposed to dynamically resize based on space. The title has min width of 220px and if it's getting smaller than that then the rest of the items are supposed to go inside a popover which is not yet implemented yet which is causing some weird styling issues
2. The tooltip is only supposed to show when the items have ellipsis truncating happening. This logic might be a little brittle and need some more testing after the first bullet is implemented since the truncating happens with CSS and we are trying to figure out if it's truncating using JS which may or may not have performance issues.

Thanks for pointing this out, I have noted these requirements in the ticket for it

packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
binoy14: From my understanding this is not supposed to go into sanity UI but would love to hear Mikolaj's thoughts on this

packages/sanity/src/desk/components/breadcrumbs/Breadcrumbs.tsx
binoy14: I simplified it, hopefully it makes more sense now

packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
mikolajdobrucki: Afaik we plan to add it to Sanity UI in the foreseeable future

mikolajdobrucki (changes requested):
The UI of the breadcrumbs needs some tweaking. Please review it against the Figma and adjust:
* font weight of breadcrumb items
* padding of breadcrumb items
* left padding of the whole breadcrumbs

Is 👇 
image

Should be 👇 
image


Notice that the font weight should be adjusted not only in breadcrumbs but also in all other panes so the weight of the title in e.g. list panes and document list panes is the same as weight of the text in the breadcrumb:
image

More tweaks needed:

image

* paddings of the popover should be smaller
* the focus ring is cut off

breadcrumbs are supposed to be hidden on mobile:
image
Link to figma spec: https://www.figma.com/file/zJxe14PiS9uadieQKTJzRM/Design-Updates-%E2%80%93-Specs?type=design&node-id=264-35458&mode=design&t=Zz1GcKZqMjHEfb8J-4

packages/sanity/src/desk/components/breadcrumbs/BreadcrumbItem.tsx
RitaDias: Fantastic, if this is the case should we add a story for it (probably in a MVI connected to this as to keep it connected to something)? 🤔

packages/sanity/src/desk/components/breadcrumbs/Breadcrumbs.tsx
RitaDias: Think it's better! Thank you! :)

RitaDias (approved):
Think this looks good! But I just want to have someone else from DX also have a look :)

binoy14:
@mikolajdobrucki fixed most of the things, here are things that are not fixed and will be in future PRs

> the focus ring is cut off

I think this is something worth fixing when the dynamic breadcrumb is implemented

> breadcrumbs are supposed to be hidden on mobile:

Mobile styles are not done, I will make a ticket for it

let me know if you see anything else

binoy14:
Going to merge this, if there is any follow up feedback, it can be addressed in future PRs

PR #4694, opened 2023-07-14
chore: drop unused root-level dependencies

rexxars:
### Description

While investigating why we had the `@vercel/node` dependency, I saw a bunch of other deps that I don't _think_ we are using on the _root level_ of the monorepo. This is for things like npm scripts and similar.

I could be wrong, of course, but I'll let CI run its course, and I'd like to hear back from y'all if you know that any of these are being used. Picking individuals here since I want your eyes on it :)

### What to review

- Do all the expected scripts and stuff still work?

### Notes for release

None.


binoy14 (approved):
LGTM

RitaDias (approved):
🚢

bjoerge (approved):
Thanks for doing this, lgtm!

PR #4700, opened 2023-07-14
chore: prettier v3

rexxars:
### Description

Upgrades to  prettier v3, and formats according to new rules.
The trailing commas takes some getting used to, but... I suppose they're okay?
We always have the option of sticking with the old setting (`es5`), which leads to a smaller number of changes.

No rush reviewing this since the prettification is just running a script, but figured I'd throw this up here since we generally want to stay up to date with dependencies.

### What to review

- Do we want to keep the `es5` setting for trailing commas, or go with the new default (`all`)
- Is this the right time to do this? I suspect we may want to wait until a number of larger PTE + DMP PRs land, probably in a few weeks time.

### Notes for release

None.


bjoerge:
Thanks for doing this!

I'm personally in favor `trailingCommas: "all"` and been using it for personal projects for a long time.

Applying formatting in a branch can be done quite painless during rebase by using the following command:
```
rebase --strategy-option=theirs --exec 'npx prettier --write `git show --name-only --pretty= HEAD | xargs ls -d 2>/dev/null` && git add `git show --name-only --pretty= HEAD | xargs ls -d 2>/dev/null` && git commit --amend --no-edit'
```

I recommend aliasing this to `pretty-rebase` like so:
```
 git config --global alias.pretty-rebase "rebase --strategy-option=theirs --exec 'npx prettier --write \`git show --name-only --pretty= HEAD | xargs ls -d 2>/dev/null\` && git add \`git show --name-only --pretty= HEAD | xargs ls -d 2>/dev/null\` && git commit --amend --no-edit'"
```

bjoerge (approved):

RitaDias (approved):
Giving the green but I agree with this:

> Is this the right time to do this? I suspect we may want to wait until a number of larger PTE + DMP PRs land, probably in a few weeks time.

The only thing is to keep in mind if PTE / DMP slides in the release window just so we don't have such a huge PR open for longer than it needs :)

Thanks for doing this!

rexxars:
With the diff-match-patch/PTE changes having landed, I am merging this. See @bjoerge's comment on how to more easily rebase your branches: https://github.com/sanity-io/sanity/pull/4700#issuecomment-1637797149

PR #4703, opened 2023-07-17
fix(core/desk): remove muted prop in tooltip content

ninaandal:
### Description
Some tooltip content had a `muted` prop, leading to potential accessibility issues. It should be darker. 

### What to review
That all the tooltips in the studio does not have a muted prop. 

Note: there are a lot of muted props in the tooltips in the search component. I have not changed this component, as this seems to be intentional. 

### Notes for release
Fixes accessibility issues in tooltip text color contrast

### DIFF
--- a/packages/sanity/src/desk/panes/document/statusBar/sparkline/PublishStatus.tsx
+++ b/packages/sanity/src/desk/panes/document/statusBar/sparkline/PublishStatus.tsx
@@ -32,7 +32,7 @@ export function PublishStatus(props: PublishStatusProps) {
         portal
         content={
           <Stack padding={3} space={3}>
-            <Text size={1} muted>
+            <Text size={1}>
               {liveEdit ? (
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Thanks @ninaandal!

packages/sanity/src/desk/panes/document/statusBar/sparkline/PublishStatus.tsx
```diff
@@ -32,7 +32,7 @@ export function PublishStatus(props: PublishStatusProps) {
         portal
         content={
           <Stack padding={3} space={3}>
```
robinpyon, line 34:
```suggestion
          <Stack padding={2} space={2}>
```

**suggestion:** Really minor, but the padding for these tooltips should be consistent with others. This could arguably be a `<Box>` component too – since we never render more than one child here.

We should definitely consider creating a studio-specific tooltip (with set overrides) if we continue to see these sorts of inconsistencies!

robinpyon (approved):
Thanks @ninaandal – added one minor comment but otherwise looks good

PR #4706, opened 2023-07-17
fix(schema): remove projectId and __experimental_search from CDR type definitions

bjoerge:
### Description

#3670 removed the ability to specify `projectId` and `__experimental_search` for cross dataset references (CDRs), but looks like we forgot to remove these from `defineField` schema typings as well. This PR simply removes both as valid type options.

### What to review
- Is it worth keeping them around as deprecated? They have no effect, so I lean towards making the type checker complain if they are provided.

### Notes for release
- Removes `projectId` and `__experimental_search` options for cross dataset reference types


rexxars:
I seem to have unintentionally superseded this one with #4713 - sorry about that.

bjoerge:
No worries :)

PR #4713, opened 2023-07-19
fix(import): rewrite cross-dataset references, add option to skip

rexxars:
### Description

Cross-dataset references no longer allow referencing other projects, which means when we are importing a previous export to a different project, the import will fail with a mutation error.

There are two possible fixes here:
1. Assume the new project has datasets with the same names. This is the new default behavior - any project IDs found during import and rewritten to the new. We then check if all the referenced datasets actually exist in the project, and throw an error if they do not.
2. Skip the cross-dataset references altogether. New options have been added:
    - `@sanity/import`: `skipCrossDatasetReferences: true`
    - `@sanity/import-cli`: `--skip-cross-dataset-references`
    - `@sanity/cli`: `--skip-cross-dataset-references` (`sanity dataset import …`)

Along with this change, 05aeed4 also adds `@deprecated` to the `projectId` property of the CDR definitions, as well as removing this property from our fixtures and tests.

Note that content lake still (currently) requires both a `_projectId` and a `_dataset` to be specified - otherwise I would remove the property from being stored altogether.

### What to review

- Cross-dataset references still work as expected in studio
- Import commands look and work correctly

### Notes for release

- Fixed an issue where exported datasets with cross-dataset references could not be imported to different projects. Note that referenced datasets must exist within same project - otherwise use `--skip-cross-dataset-references`.

### DIFF
--- a/dev/test-studio/schema/debug/allNativeInputComponents.ts
+++ b/dev/test-studio/schema/debug/allNativeInputComponents.ts
@@ -129,7 +129,6 @@ export const allNativeInputComponents = defineType({
     defineField({
       type: 'crossDatasetReference',
       name: 'crossDatasetReference',
-      projectId: 'ppsg7ml5',
       dataset: 'test',
       to: [
         {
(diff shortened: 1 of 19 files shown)


rexxars:
Oops, just realized that #4706 covers part of this. However, I feel like it's better that we deprecate the property for now, since removing it completely will cause breaking builds.

rexxars:
Updated with a commit to cover the experimental search as well, to fully supersede #4706.

bjoerge (approved):
Excellent work!

My only question was whether we could skip writing (the now redundant) `_projectId` entirely (or even consider removing it if it's already there), but see you have already asked about it internally.

binoy14 (approved):
Looks good from what I can gather

PR #4716, opened 2023-07-19
refactor(desk): drop unused validation menu component

rexxars:
### Description

Could not find a single place this was used - I assume it's a leftover from migrating from the dropdown to the inspector panel.

### What to review

- Component is not used anywhere, validation still works

### Notes for release

None


packages/sanity/src/desk/panes/document/documentPanel/header/ValidationMenu.tsx
```diff
@@ -1,93 +0,0 @@
...
-import {
-  isValidationInfoMarker,
-  isValidationWarningMarker,
-  isValidationErrorMarker,
-  Path,
-} from '@sanity/types'
-import {Button, ButtonProps, Menu, MenuButton} from '@sanity/ui'
-import React, {useCallback, useMemo, useId} from 'react'
-import {useDocumentPane} from '../../useDocumentPane'
-import {ValidationList} from 'sanity'
```
binoy14, line 11: Correct me if I am wrong but this was the only component using `ValidationList` https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/components/validation/validationList.tsx 

and seems like `ValidationList` is the only component using `ListItem` https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/components/validation/listItem.tsx#L24

binoy14 (approved):
✅ left a comment of other components that might be unused if we want to nuke them as well

packages/sanity/src/desk/panes/document/documentPanel/header/ValidationMenu.tsx
rexxars: Thanks for being thorough! Removed those components too!

packages/sanity/src/core/components/index.ts
```diff
@@ -21,6 +21,5 @@ export * from './textWithTone'
 export * from './TooltipOfDisabled'
 export * from './transitional'
 export * from './userAvatar'
-export * from './validation'
```
RitaDias, line 24: Perhaps silly question, but since we are exporting it is there a risk that this would then be a breaking change? 🤔

binoy14 (approved):
🗑️

packages/sanity/src/core/components/index.ts
jonjohansen: Seems like the scheduled publishing plugin was using this: 

https://github.com/sanity-io/sanity-plugin-scheduled-publishing/issues/80 😬

RitaDias: I guess this is as good time as any to understand what is our process / stance when internally tagged tools are used and then we do changes / remove them 😬

PR #4719, opened 2023-07-20
fix(core): add missing validation at input level validation (boolean)

RitaDias:
### Description

Add input validation for boolean inputs.
Before it wasn't showing in the document that the boolean had missing / breaking validations.

### What to review

**Boolean**
- Go to V3 APIs -> V3 validations
- Pick any document or create a new one
- Check that the validation works for booleans once those are set.

### Notes for release
 
- Boolean inputs will now show validation errors at the input level (in document)

### DIFF
--- a/packages/sanity/src/core/form/inputs/BooleanInput.tsx
+++ b/packages/sanity/src/core/form/inputs/BooleanInput.tsx
@@ -19,7 +19,7 @@ const ZeroLineHeightBox = styled(Box)`
  * @beta
  */
 export function BooleanInput(props: BooleanInputProps) {
-  const {id, value, schemaType, readOnly, elementProps} = props
+  const {id, value, schemaType, readOnly, elementProps, validation} = props
   const layout = schemaType.options?.layout || 'switch'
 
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/BooleanInput.tsx
```diff
@@ -46,7 +46,7 @@ export function BooleanInput(props: BooleanInputProps) {
           <FormFieldHeaderText
             description={schemaType.description}
             inputId={id}
-            // validation={validation}
```
bjoerge, line 49: _Extremely_ curious as to why this was commented out in the first place, but afraid we'll never know

bjoerge (approved):
Seems to work fine! Thanks @ritadias!

packages/sanity/src/core/form/inputs/BooleanInput.tsx
rexxars: 😅

rexxars (approved):

PR #4721, opened 2023-07-20
refactor(validation): move `@sanity/validation` into monopackage

rexxars:
### Description

**[Low priority]**

Rationale: the exported API has never really been built for external consumption. It requires a number of internal things that is very cumbersome to pull together outside of a Sanity application - schema, client getters, document existence checks and whatnot.

By moving it to the monopackage, we can more easily treat it as what it is; an internal API. With time, we can instead work on exposing a simplified API meant for external consumption in node scripts and similar.

Aside from moving it to the monopackage, a few other alterations/tidy-ups have been done:

- Uses named exports
- Uses `type` keyword for imports where applicable
- Moves exported functions to the top of the file where possible
- Uses the `export ... from ...` syntax for the index instead of `import mod from ...`  + `export {mod}`

**Note:** At some point we need to publish a version of `@sanity/validation` that indicates it is deprecated/not used internally anymore - but the current version should continue to work just fine for a while longer.

### What to review

- Validation still works!
- Code looks good

### Notes for release

None.


binoy14 (approved):
Looks good to me!

packages/sanity/src/core/validation/inferFromSchema.ts
```diff
@@ -15,5 +15,3 @@ function inferFromSchema(schema: Schema): Schema {
 
   return schema
 }
-
-export default inferFromSchema
```
RitaDias, line 19: Might be a non issue but: Could these sort of changes (removing `defaults` for this particular instance for example) be considered a breaking change?

packages/sanity/src/core/validation/validators/booleanValidator.ts
RitaDias, line 1: Interesting that we seemingly didn't have a validator for boolean 🤔 
I would have expected to just be a typical move from the /validators/booleanValidator to this. While the code looks similar I think it's a github thing and not a "our code" thing 🤔

bjoerge (approved):
LGTM, great work!

rexxars:
> I'm ok with approving this, I guess my only question really is related to [#4721 (comment)](https://github.com/sanity-io/sanity/pull/4721#discussion_r1280402631). Mainly asking not from our code base but from someone / some plugin whenever we push this new version out. Makes me think the imports will probably break, easy fix from their side but annoying?

With this change, we're not actually even exposing (exporting) the validation logic anymore - but people are free to continue using the old `@sanity/validation` package until further notice. At some point we'll expose better APIs that are meant for public consumption - currently it's a bit too tied up in internals: configs, schemas, i18n etc etc.

packages/sanity/src/core/validation/validators/booleanValidator.ts
rexxars: I found a few things in there that I tweaked briefly, so I think git treated it as a remove+add instead of move. I should probably have done it in two different commits, but.. oh well.

packages/sanity/src/core/validation/validateDocument.ts
```diff
@@ -50,7 +47,17 @@ export function resolveTypeForArrayItem(
...
-export default async function validateDocument(
+/**
+ * Validates a document against the given schema
+ *
+ * @param getClient - Function that returns a Sanity client, based on a passed configuration
+ * @param doc - Document to validate
+ * @param schema - Schema to validate against
+ * @param context - Context to use for validation
+ * @returns A promise that resolves to an array of validation markers
+ * @beta
```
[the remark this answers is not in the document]
rexxars: Will add after merge, but this is only for internal consumption - won't be exported :)

packages/sanity/src/core/validation/validateDocument.ts
```diff
@@ -59,6 +66,16 @@ export default async function validateDocument(
...
 
+/**
+ * Validates a document against the given schema, returning an Observable
+ *
+ * @param getClient - Function that returns a Sanity client, based on a passed configuration
+ * @param doc - Document to validate
+ * @param schema - Schema to validate against
+ * @param context - Context to use for validation
+ * @returns An Observable that produces an array of validation markers
+ * @beta
```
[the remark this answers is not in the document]
rexxars: Will add after merge, but this is only for internal consumption - won't be exported :)

PR #4722, opened 2023-07-20
fix(core): optimize wheel event on CommandList virtual element

rexxars:
### Description

Small change: sets the mouse wheel listener on the virtual list element to be passive ([background](https://chromestatus.com/feature/5745543795965952)). This potentially speeds up the scrolling, and should be safe since we never call `preventDefault()` in the attached listener.

### What to review

- Lists work as before

### Notes for release

None.


robinpyon (approved):
Good catch, thanks @rexxars!

PR #4740, opened 2023-07-24
fix(core): allow `null` and `role`-less user in `userHasRole` fn

rexxars:
### Description

This allows it to be used directly from a conditional property callback, where `null` is valid since a studio may not require authenticated users. In addition, the conditional property callback context have dropped the legacy `user` property, which meant the signature did not match.

### Notes for release

- Made the `userHasRole` function more relaxed on user value, allowing easier use in places such as conditional callbacks


binoy14 (approved):

bjoerge (approved):
LGTM!

PR #4743, opened 2023-07-26
fix(desk): multiple desk support

andparsons:
### Description

With the work done in #4742, two constants (HISTORY_INSPECTOR_NAME & VALIDATION_INSPECTOR_NAME) were introduced that cause a conflict when trying to use more than one desk in a workspace.

This change introduces replacement functions that use the name defined in the config of the desk tool which is used as the url segment, ensuring uniqueness.

### What to review

This has a minimal surface area to test, and intoduces no new functionality.
The intent is if no desk name is defined in the config, there is no change from the original default.

### Notes for release

To support multiple desks, tool name now used in validator names


rexxars:
Closing in favor of #4748

PR #4746, opened 2023-07-27
feat(types): add `isSlug` asserter method

rexxars:
### Description

Will be used in the upcoming validation changes, for instance. Figured it was worth a separate PR for easier reviewing.

### What to review

- Assertion looks okay (I don't think we need to go deeper - this matches the overall shape we need) 

### Notes for release

- Added `isSlug()` method to `@sanity/types` for type-safe asserting of a slug object


binoy14 (approved):

bjoerge (approved):
Thanks, lgtm!

PR #4718, opened 2023-07-20
feat(i18n): provide internationalization capabilities

rexxars:
This provides most of the "framework", as a first draft for review. Many of the ideas here are based on #4502, but it has been reduced in scope. Please see internal slack for a high-level API overview - it's probably easier to start there, and then look at this for implementation details.

There are some "leftovers" from the PoC implementation that I know needs to be rethought/cleaned up - especially the time `timeAgo` is a particularly gnarly one, given [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) and how that probably is a good use for it.

For this PR, I want you to focus on the implementation details of the _framework_. The idea is to land this in a feature branch (`feat/i18n`) and keep working on internationalizing the UI, but provide smaller PRs to review as we move. I realize this _is_ a larger PR, but with a PoC already in place, I felt it would make sense to start from there and adjust, rather than build brick by brick.

I suggest you review this as a whole rather than per-commit, since there's a bit of exploration going on.

We use [i18next](https://www.i18next.com/) under the hood, which is... okay. There are a lot of implementation details that feels a bit off to me (numerous ways of doing the same thing, undocumented minor alterations to code flow), and the documentation leaves a lot to be desired. I've made an attempt of using the good parts, but leaving it open for us to switch to a different framework/write a simplified version ourselves, but there is still some work to be done in this area - in particular, the `useTranslation` hook and the `t` function is still fully typed from `i18next`, which means we have to keep support their API. Not ideal, but with some help we might be able to limit this before release.

PR #4747, opened 2023-07-27
feat(i18n): localized validation messages

rexxars:
### Description

This provides localized validation messages, using the i18n framework from #4818.

Roughly speaking, we pass the `LocaleSource` down to all validators, which allows them to use the `i18n.t()` method to translate messages. This is done internally for all the built-in validators.

Users can provide custom messages in three ways:
1. Override existing validation messages by implementing a resource bundle (see the [Norwegian example](https://github.com/sanity-io/sanity/commit/4c099598be85185c3d62cc01839ea2a331699658) in the test studio.
2. Return a localized string of their own choosing from a `Rule.custom()` method, using the `i18n.t()` function from context
3. Provide an object of localization strings to each `.error()`/`.warn()`:
    ```ts
    {
      // ...
      validation: Rule => Rule.required().error({
        'no-NB': 'Denne må du fylle ut',
        'en-US': 'You need to fill out this one'
      })
    }
    ```

**Note:** ~this builds on #4718 and includes #4721 and #4746 - it's probably best to review those first, and then come back to this. Once stuff lands in `next`, I'll rebase branches for easier reviewing.~

**Update**: The above PRs have been merged, this should be ready for reviewing now

### What to review 

- Implementation looks correct
- Can switch between locales and have validation reflected in language of choice

### Notes for release

None for now.


packages/sanity/test/validation/__snapshots__/numbers.test.ts.snap
```diff
@@ -194,7 +194,7 @@ Array [
   Object {
     "item": ValidationError {
       "children": undefined,
-      "message": "Value is required",
+      "message": "Required",
```
binoy14, line 197: ❓ should this still be `Value is required` vs just `Required` ?

binoy14 (approved):

binoy14:
Not sure if it's related to this but changing the locales does not show the validation messages anymore

https://github.com/sanity-io/sanity/assets/6476108/a861a46e-3c0d-4ae6-adad-ccc99aec4f6f

packages/sanity/test/validation/__snapshots__/numbers.test.ts.snap
rexxars: I took the liberty of changing it - all the other validators were using simply "Required", so I felt it was too inconsistent to leave this one with a different message.

rexxars:
> Not sure if it's related to this but changing the locales does not show the validation messages anymore

Thanks for reporting! While it does show itself through this PR, it is not _caused by it_ - it does reveal that we have a problem where validation is not recomputed/retained when the React tree re-renders pretty far up, which needs to be fixed. I'll create a separate story to address this.

dev/test-studio/plugins/locale-no-nb/bundles/studio.ts
```diff
@@ -59,7 +59,7 @@ const studioResources: Record<StudioLocaleResourceKeys, string> = {
   'timeAgo.seconds.ago_one': 'ett sekund siden',
   'timeAgo.seconds.ago_other': '{{count}} second ago',
   /* Relative time, granularity: seconds, using a minimal format*/
```
RitaDias, line 61: Super small nitpick: when reading the comment about this (granularity: seconds) and then seeing hte `m` on the seconds minimal, I got a bit confused. What is the "m" meant to stand for here? "miliseconds" / "minutes" / "seconds" ? 😬

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -60,7 +65,7 @@ const studioLocaleStrings = {
   'timeAgo.seconds.ago_one': '{{count}} minute ago',
   'timeAgo.seconds.ago_other': '{{count}} second ago',
   /* Relative time, granularity: seconds, using a minimal format*/
-  'timeAgo.seconds.minimal': '{{count}m',
+  'timeAgo.seconds.minimal': '{{count}}m',
```
RitaDias, line 68: Same question as [here](https://github.com/sanity-io/sanity/pull/4747/files#r1285529916)

packages/sanity/src/core/validation/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './types'
```
RitaDias, line 1: Are we sure we want to export literally everything? Just double checking since we've been pretty reduced with the number of things we export beneath. 🤔

packages/sanity/src/core/validation/types.ts
```diff
@@ -0,0 +1,23 @@
...
+   * module in the first place, but is now unfortunately part of the public API and thus cannot
+   * be changed easily.
+   *
+   * This is a temporary solution until we can remove the `ValidationContext` from the types module,
+   * which is likely to happen at the next major version.
+   *
+   * @public
+   */
+  interface ValidationContext {
+    i18n: LocaleSource
```
RitaDias, line 19: If I'm not mistaken, this will probably appear as a property in the rendering of the docs ui, we might want to just add a small comment / linking it back to the LocaleSource type

packages/sanity/src/core/i18n/fallback.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Returns a fallback source for internationalization that can be used in cases where you need
+ * access to the "framework" of translation, but do not have access to an actual source.
+ *
+ * Memoized - only initializes the i18n instance once, and maintains the identity of the source.
+ *
+ * ⚠️ NOTE: This will only have the base studio strings defined - no custom bundles or plugins.
+ *
+ * @returns The fallback source
```
RitaDias, line 21: Do we want to link it to the type `LocaleSource`? 🤔

packages/sanity/src/core/i18n/fallback.ts
```diff
@@ -0,0 +1,64 @@
...
+    fallbackLng: defaultLocale.id,
+    lng: defaultLocale.id,
+    supportedLngs: [defaultLocale.id],
+    debug: false,
+    load: 'currentOnly',
+    resources: staticResources,
+    interpolation: {
+      // If we're in a browser, assume this is running inside of the studio, eg a React app,
+      // and that values returned will be escaped by the framework (eg React) automatically.
+      escapeValue: shouldEscape,
```
RitaDias, line 61: This is for my own curiosity: why do we need this? 👀

packages/sanity/src/core/i18n/localeNamespaces.ts
```diff
@@ -5,3 +5,10 @@
  * @hidden
  */
 export const studioLocaleNamespace = 'studio' as const
+
+/**
+ * The namespace for validation messages.
+ *
+ * @beta
```
RitaDias, line 12: Add a `@hidden`?

packages/sanity/src/core/validation/validators/dateValidator.ts
```diff
@@ -12,6 +12,7 @@ const isoDate =
 // eslint-disable-next-line no-warning-comments
 // TODO (eventually): move these to schema type package
 interface DateTimeOptions {
+  // @todo accept object representing `Intl.DateTimeFormatOptions`? how do we localize?
```
RitaDias, line 15: Do we have a shortcut story for this? 👁️

RitaDias (commented):
I left some comments (half of them are some questions I had myself as a looked at the code and half are pure curiosity), I haven't approved it because I'd like @bjoerge to have a final look at this and I'll keep an eye :eye:

dev/test-studio/plugins/locale-no-nb/bundles/studio.ts
rexxars: Eh, I admit this one was rushed - I ported this directly from the earlier PoC, only doing changes to _syntax errors_.  This one should be `s`. It is likely to be replaced by a full "relative time" implementation from the `Intl` API in an upcoming PR

RitaDias: Makes sense, I guess one main question I have upon a second look is if we need the {{count}} / placeholder types in the translations for the studio. This might be splitting hairs but a thought that I came to (for these specific situations where time is involved) is that it might easier to simply add (in the implementation):

const timeLapsed = (...)

return `{{timeLapsed}} {{context.i18n.t('timeAgo.seconds.ago_other')}}`

At the same time I can totally see that this implementation is more flexible for the overall scenarios (like where the "placeholder" is in the middle of the sentence) which does make me think it's just the overall better approach so there aren't any potential confusion depending on use cases 🤔 .

rexxars: Eh, these are ready to be replaced at some point near in time - don't mind them - they are directly ported from the PoC. I only touched a syntax error when I came across it. This one should have been `s`, not `m`, fwiw.

rexxars: I don't want to dive too deeply into this time example, but for other places, the best practice is to not do that type of interpolation. i18next has [_some_ documentation](https://www.i18next.com/principles/best-practices#impact-of-interpolation-on-localization) on it.

packages/sanity/src/core/validation/index.ts
rexxars: It's only three members, all of them public. But yes, this might be a good time to start limiting the exposed types :)
I'll switch to explicit exports after merging!

packages/sanity/src/core/validation/types.ts
rexxars: Will add after merging, thanks for pointing that out

packages/sanity/src/core/i18n/fallback.ts
rexxars: Given this is internal, should be enough to follow the return type

packages/sanity/src/core/i18n/fallback.ts
rexxars: Honestly, we probably don't, and I'm not sure we actually _want_ escaping. I can't see us passing HTML down in these strings, and even if we did, I'm not sure what the fallback would be used for - certainly wouldn't expect it to be used for rendering HTML directly. I don't feel confident enough changing it right now, since escaping should be safer - but I'm open to changing it in the future if we encounter issues with it.

packages/sanity/src/core/i18n/localeNamespaces.ts
rexxars: Yep will do.

packages/sanity/src/core/validation/validators/dateValidator.ts
rexxars: Not yet, will create one :) 
It's not really a must-have, but I thought it might be interesting to explore it.

PR #4748, opened 2023-07-27
fix(desk): prevent duplicate validation/change inspectors

rexxars:
### Description

When multiple desks are configured for a studio, you ended up with duplicate change and validation inspectors.
This PR addresses this with the same approach as the badges/actions: it only adds those plugin elements if not already present, as part of the plugin configuration.

Also refactored the duplication checks to use a Set instead, as it is a little less verbose.

Supersedes #4743 if approved (will give credit to author as part of release notes, even though we didn't specifically use their code)

### What to review

- No duplicate document inspectors, badges, or actions exists

### Notes for release

- Fixed an issue where using multiple desk tools in the same studio would cause duplicate "review changes" actions and validation lists to appear. Thanks @andparsons!

### DIFF
--- a/packages/sanity/src/desk/deskTool.ts
+++ b/packages/sanity/src/desk/deskTool.ts
@@ -27,6 +27,8 @@ const documentActions = [
 
 const documentBadges = [LiveEditBadge]
 
+const inspectors = [validationInspector, changesInspector]
+
 /**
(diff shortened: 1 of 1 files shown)


binoy14 (approved):
✅ very nice!

bjoerge (approved):
LGTM - thanks for adding thoughtful comments <3

PR #4749, opened 2023-07-28
fix(desk): use configured api version when fetching document lists

bjoerge:
### Description
Currently, specifying `apiVersion` on either `documentList()` or `documentTypeList()` has no effect. This PR fixes it by forwarding the option to the `useClient()`-call in `useDocumentList()`

### What to review
Make sure specifying `apiVersion(<version>)` on documentList and documentTypeList results in `<version>` actually being used when fetching documents

### Notes for release
- Fixed a bug that made specifying `apiVersion` on `documentList()` and `documentTypeList()` not work

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -196,7 +196,7 @@ export const structure: StructureResolver = (S, {schema, documentStore}) => {
                     options: {
                       filter: '_id in path("drafts.**")',
                     },
-                  }),
+                  }).apiVersion('2023-07-28'),
                 ),
 
               S.listItem()
@@ -317,6 +317,7 @@ export const structure: StructureResolver = (S, {schema, documentStore}) => {
                 child: () =>
                   S.documentTypeList('author')
                     .title('Developers')
+                    .apiVersion('2023-07-27')
                     .filter('_type == $type && role == $role')
                     .params({type: 'author', role: 'developer'})
                     .initialValueTemplates(S.initialValueTemplateItem('author-developer')),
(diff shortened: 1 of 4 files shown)


robinpyon (commented):
Thanks @bjoerge!

I had a look and it appears this change causes other document lists to load with an older version of the API (`2021-06-07`) by default, as opposed to the version defined by `DEFAULT_STUDIO_CLIENT_OPTIONS.apiVersion`. 

Custom document lists (such as [this one](https://test-studio-git-bug-sc-35794.sanity.build/test/content/custom;anythingWithATitle)) will also fail to load.

I believe the reason for this is that Structure Builder is assigning an `apiVersion` further up the chain [based off the filter type provided](https://github.com/sanity-io/sanity/blob/3dffb111b81143d2c499eb0e5942bd3abc655753/packages/sanity/src/desk/structureBuilder/DocumentList.ts#L244). In the above case, since the filter is deemed non-simple, it reverts to `v1` which presumably doesn't support the `defined` function

- In this instance, should we change `2021-06-07` to point to `DEFAULT_STUDIO_CLIENT_OPTIONS.apiVersion` instead?
- When it comes to handling non-simple type filters, do we still want to default to v1?
  - If so, we should make a note to improve this error message for people who supply their own custom document list filters

bjoerge:
@rexxars I've addressed your feedback in [a86d0cb](https://github.com/sanity-io/sanity/pull/4749/commits/a86d0cb965b64c67f53b472cd794f6eb0562d7f1), please take a look.

- I've switched to using default studio options for simple filters
- Kept the fallback to `v1` for custom filters unless an API version is set
- Added a console warning if no `apiVersion` is explicitly set, pointing at [this help article](https://www.sanity.io/help/structure-api-version-required-for-custom-filter)

rexxars (approved):
Great!

PR #4750, opened 2023-07-28
fix(core): change cdr tooltip text

ninaandal:
### Description
The cross dataset reference tooltip text was changed to avoid any confusion and to be consistent with the icon and text that is used in other places, for example in arrays (see screenshot)
Screenshot 2023-07-28 at 10 45 14

### What to review
The tooltip text for CDR 

### Notes for release
Tooltip text for CDR was changed


bjoerge (approved):
Thank you!

PR #4756, opened 2023-08-01
Various improvements to the PTE and PT-input

skogsmaskin:
### Description

First of all, sorry there's a bit of changes of various characters in this PR, but as they are built upon each other over a longer period, it wasn't too easy to extract them into smaller PRs.

This PR will change the following:

* Use the new diffMatchPatch library in the `@sanity/portable-text-editor` according to the work being done in [this PR ](https://github.com/sanity-io/sanity/pull/4412)
* Change from sending patches debounced to throttled, making sure that changes are synchronized more often with the server
* Upgrade slate and slate-react to the latest versions, getting rid of our own fork of slate-react.
* Setting the Portable Text editor in readOnly mode when the internet connection is down.
* Improved undo/redo functionality and more tests for this.
* Minor selection fixes to the PT-input


packages/@sanity/portable-text-editor/test/web-server/components/Editor.tsx
```diff
@@ -71,6 +71,7 @@ export const Editor = ({
   const selectionString = useMemo(() => JSON.stringify(selectionValue), [selectionValue])
   const editor = useRef<PortableTextEditor>(null)
   const keyGenFn = useMemo(() => createKeyGenerator(editorId.substring(0, 1)), [editorId])
+  const [readOnly, setReadOnly] = useState(!window.navigator.onLine)
```
bjoerge, line 74: I'm a fan of using a local state override instead of syncing props with state, e.g. instead of updating based on a new props value and triggering a re-render, I'd suggest changing this state to `isOffline` and use that later on with `readOnly={isOffline || readOnlyFromProps}`, which I think also reads better.

packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
```diff
@@ -31,6 +31,8 @@ import {useScrollSelectionIntoView} from './hooks/useScrollSelectionIntoView'
 import {Style} from './text/Style'
 import {ListItem} from './text/ListItem'
 
+const noOutlineStyle: React.CSSProperties = {outline: 'none'}
```
bjoerge, line 34: Consider removing the dependency to React.CSSProperties by changing to

```suggestion
const noOutlineStyle = {outline: 'none'} as const
```

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -89,6 +89,7 @@ export function PortableTextInput(props: PortableTextInputProps) {
   const [invalidValue, setInvalidValue] = useState<InvalidValue | null>(null)
   const [isFullscreen, setIsFullscreen] = useState(false)
   const [isActive, setIsActive] = useState(false)
+  const [readOnly, setReadOnly] = useState(readOnlyFromProps)
```
bjoerge, line 92: Same as earlier re:syncing props with state / local override

bjoerge (approved):

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
skogsmaskin: Good point, I've refactored these to get rid of the `readOnly` state var (use directly from props) and introduced an own state var for the network status to keep these concepts untangled.

bjoerge (approved):

hermanwikner (approved):

PR #4768, opened 2023-08-01
fix(form): improve reference error states

bjoerge:
### Description

If you have a reference to a nonexistent document, we currently display a confusing/unhelpful message. This PR refactors some of the logic we already had for handling the various cases that may occur with references to make them show up as originally intended.

**Before**
image

**After**
image

(Note: the reference strength mismatch is irrelevant in this case, since it's been deleted while still being referenced it can only be a weak ref, and can't be made into a strong reference)

Also snuck in a fix for an issue causing references of invalid types to never load:
**Before**
image
(forever)

**After**
image
(note: "empty" is the actual type name here)

### What to review

Make sure error states for reference inputs renders correctly:
- A (weak) reference that can't be found
- Referenced document is inaccessible by current role (this one might be hard to test, so feel free to skip)
- Weak/strong mismatch between schema and value
- Referenced document is of invalid type

### Notes for release

- Fixed an issue where references to undeclared document types would never load
- Improved error UIs for reference values

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
@@ -171,13 +171,16 @@ export function ReferenceField(props: ReferenceFieldProps) {
             <Text as="p" muted size={1}>
               {schemaType.weak ? (
                 <>
-                  It will not be possible to delete the "{preview?.title}"-document without first
-                  removing this reference.
+                  It will not be possible to delete the{' '}
+                  {preview?.title ? <>"{preview?.title}"-document</> : <>referenced document</>}{' '}
+                  without first removing this reference.
                 </>
               ) : (
                 <>
-                  This makes it possible to delete the "{preview?.title}"-document without first
-                  deleting this reference, leaving this field referencing a nonexisting document.
+                  This makes it possible to delete the{' '}
+                  {preview?.title ? <>"{preview?.title}"-document</> : <>referenced document</>}{' '}
+                  without first deleting this reference, leaving this field referencing a
+                  nonexisting document.
                 </>
               )}
             </Text>
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceItem.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceItem.tsx
@@ -307,13 +307,16 @@ export function ReferenceItem<Item extends ReferenceItemValue = ReferenceItemVal
             <Text as="p" muted size={1}>
               {schemaType.weak ? (
                 <>
-                  It will not be possible to delete the "{preview?.title}"-document without first
-                  removing this reference.
+                  It will not be possible to delete the{' '}
+                  {preview?.title ? <>"{preview?.title}"-document</> : <>referenced document</>}{' '}
+                  without first removing this reference.
                 </>
               ) : (
                 <>
-                  This makes it possible to delete the "{preview?.title}"-document without first
-                  deleting this reference, leaving this field referencing a nonexisting document.
+                  This makes it possible to delete the{' '}
+                  {preview?.title ? <>"{preview?.title}"-document</> : <>referenced document</>}{' '}
+                  without first deleting this reference, leaving this field referencing a
+                  nonexisting document.
                 </>
               )}
             </Text>


packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
```diff
@@ -81,3 +153,45 @@ export function PreviewReferenceValue(props: {
...
+
+  if (list.length === 2) {
+    return [list[0], <Fragment key="comma"> {conjunction} </Fragment>, list[1]]
+  }
+
+  const subList = list.slice(0, -1)
+  return [
+    ...subList.map((item, i) => <Fragment key={i}>{item}, </Fragment>),
+    <Fragment key="last">
+      {conjunction} {list.at(-1)}
```
rexxars, line 194: We still use ES2017 as our build target, but `Array.prototype.at` is not included until ES2022.
```suggestion
      {conjunction} {list[list.length - 1]}
```

rexxars (changes requested):

rexxars (approved):

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
```diff
@@ -57,21 +59,91 @@ export function PreviewReferenceValue(props: {
...
+                </Text>
+              </Stack>
+            }
+          >
+            <TextWithTone tone="default">
+              <HelpCircleIcon />
+            </TextWithTone>
+          </Tooltip>
+        </Box>
+      </Inline>
```
robinpyon, line 141:
```suggestion
      <Flex align="center" justify="flex-start">
        <Box padding={1}>
          <Flex align="center">
            <Box flex={1} paddingY={2}>
              <Text muted>Type of referenced document not declared in schema</Text>
            </Box>
          </Flex>
        </Box>
        <Box>
          <Tooltip
            portal
            content={
              <Stack space={3} padding={3}>
                <Text size={1}>
                  Referenced document (<code>{value._ref}</code>) is of type{' '}
                  <code>{referenceInfo.result.type}</code>.
                </Text>
                <Text size={1}>
                  According to the schema, referenced documents can only be of type{' '}
                  {humanizeList(
                    type.to.map((toType) => <code key={toType.name}>{toType.name}</code>),
                    'or'
                  )}
                  .
                </Text>
              </Stack>
            }
          >
            <Box padding={2}>
              <TextWithTone tone="default">
                <HelpCircleIcon />
              </TextWithTone>
            </Box>
          </Tooltip>
        </Box>
      </Flex>
```

**suggestion**: Using a wrapping `<Flex>` rather than `<Inline>` to ensure that the icon / context menu wraps gracefully at smaller breakpoints, e.g.

image

vs.

image

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
```diff
@@ -57,21 +59,91 @@ export function PreviewReferenceValue(props: {
...
+              portal
+              content={
+                notFound ? (
+                  <UnavailableMessage title="Not found" icon={HelpCircleIcon}>
+                    The referenced document does not exist
+                    <br />
+                    (id: <code>{value._ref}</code>)
+                  </UnavailableMessage>
+                ) : (
+                  <UnavailableMessage title="Insufficcient permissions" icon={AccessDeniedIcon}>
```
robinpyon, line 88:
```suggestion
                  <UnavailableMessage title="Insufficient permissions" icon={AccessDeniedIcon}>
```

robinpyon (approved):
Thanks @bjoerge! Left a few comments, nothing major, all cosmetic

I did also notice that references that both have a reference strength AND schema type mismatch will not render a title, e.g. (note `""-document` )

image

Ideally this would fall back to a generic title (e.g. "this document"), but this is super edge case stuff

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
```diff
@@ -81,3 +153,45 @@ export function PreviewReferenceValue(props: {
...
+          <Text as="p" muted size={1}>
+            {props.children}
+          </Text>
+        </Box>
+      </Box>
+    </Flex>
+  )
+}
+
+const humanizeList = (list: React.ReactNode[], conjunction: string) => {
```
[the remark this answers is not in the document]
rexxars: > **question:** what qualifies for inclusion in `@sanity/utils`?

Pretty sure we want to move away from `@sanity/utils` altogether. Utilities should either be internal to the `sanity` module, or exported publicly from the `sanity` module if they're reusable enough that it makes sense for our users to use them.

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
mikolajdobrucki: @robinpyon which one is 'before' and which one is 'after' your change?
I guess that the icon should *not* wrap under the text?

robinpyon: That's correct – first image is 'after', this could have been clearer!

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
```diff
@@ -57,21 +59,94 @@ export function PreviewReferenceValue(props: {
...
     return (
-      <Stack space={2} padding={2}>
-        The referenced document is of invalid type: ({refTypeName || 'unknown'})
-        <pre>{JSON.stringify(value, null, 2)}</pre>
-      </Stack>
+      <Flex align="center" justify="flex-start">
+        <Box padding={1}>
+          <Flex align="center">
+            <Box flex={1} paddingY={2}>
+              <Text muted>Type of referenced document not declared in schema</Text>
```
mikolajdobrucki: I would suggest to change this copy to "Document of invalid type"

bjoerge: Agreed - fixed by 544c368153

mikolajdobrucki (approved):
Amazing work! Looks good to me! 🚀

packages/sanity/src/core/form/inputs/ReferenceInput/PreviewReferenceValue.tsx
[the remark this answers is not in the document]
bjoerge: It just didn't feel worth it to include an external dependency for just this single case. Should be easy to replace with something from npm or turn it into it's own package at a later point if we need it elsewhere too.

bjoerge:
Thanks for review everyone :)


@robinpyon - and thanks for catching the ""-document issue - I've fixed it so it display the document title if available, or "the referenced document" if not:
image

image

While doing a last round of testing I found another visual issue (also very edge casey, but still):
If you create a new reference in place of a certain document type and later remove that type from the list of valid reference types in the schema, it would show up like this:
image

I've fixed this in 544897bbff so it looks and behaves the same as the other type mismatch message:
image

PR #4771, opened 2023-08-01
chore(deps): remove unused typedoc dependency

binoy14:
### Description

Removes unused dependency of `typedoc` 

### What to review

Making sure the dependency is not used anywhere

### Notes for release

N/A


rexxars:
It is currently in use by the portable text editor for generating docs, but I am not sure it is actually _needed_ since those docs (as far as I am aware) are not actually presented anywhere. @skogsmaskin ?

skogsmaskin:
That's right. It can be removed. Will be replaced by the new tsdoc packages we use elsewhere at some point.

rexxars (approved):

PR #4772, opened 2023-08-01
feat(auth): allow using explicit `token` mode for authentication

rexxars:
### Description

In certain rare cases, such as browsers inconsistently serving cookies, it may be helpful to explicitly turn off cookie-based authentication. This PR adds a new `token` option to the `loginMethod` property, which enables this.

I had to rewrite the flow in the auth store a bit to allow reuse between the `dual` and `token` options, but I think it reads slightly better now anyway. 

### What to review

- Authentication works as before in Chrome/Firefox/Safari
- Setting the `loginMethod` to `token` and `cookie` works (`cookie` _may_ not work in some browsers, but you should be seeing the same behavior as _before this change_)

  You can do so by using `createAuthStore()` - `sanity.config.ts`:
  
  ```ts
  import {defineConfig, createAuthStore} from 'sanity'

  export default defineConfig({
    // ...
    auth: createAuthStore({
      projectId: '...',
      dataset: '...',
      loginMethod: 'token'
    })
  })
  ```

### Notes for release

- Added new `token` login method for rare cases when the more secure cookie approach is not viable/wanted

### DIFF
--- a/packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
+++ b/packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
@@ -34,7 +34,7 @@ export interface AuthStoreOptions {
    * - `cookie` - explicitly disable `localStorage` method, relying only on
    *   cookies
    */
-  loginMethod?: 'dual' | 'cookie'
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
Looks solid - great work!

PR #4773, opened 2023-08-01
feat(auth): allow specifying options without createAuthStore()

rexxars:
### Description

~**Note**: builds on #4772 - please only review [this commit](https://github.com/sanity-io/sanity/commit/d0fab5cf4ebf4659f7c594eaaf00d52d91c9c448).~

When configuring the authentication aspects, you currently have to give it an `AuthStore`. This is not very ergonomic/developer friendly, and made worse by the fact that it requires you to duplicate project ID and dataset, as well as `createAuthStore` actually being marked as `@internal`.

With this change, I propose that we instead take a partial object of configuration parameters, such as `loginMethod`, `providers` etc, and merge these into the default options and internally call `createAuthStore()`. I left the `AuthStore` as an option because it can be useful for mocking and backwards compatibility, but we should discourage people from using it generally.

### What to review

- Code looks okay
- Regular authentication (without specifying a config) works as expected
- Authentication using an object works (staging workspace in test studio configured to use it, check for `withSid` parameter in login URL to see that it works)
- Providing an auth store still works (tests should take care of this, as they use it for mocking)

### Notes for release

- Allow specifying authentication options using an object instead of `createAuthStore()`

### DIFF
--- a/packages/sanity/src/core/config/auth/types.ts
+++ b/packages/sanity/src/core/config/auth/types.ts
@@ -1,0 +1,74 @@
+/**
+ * Authentication options
+ *
+ * @public
+ */
+export interface AuthConfig {
+  /**
+   * Login method to use for the studio the studio. Can be one of:
+   * - `dual` (default) - attempt to use cookies where possible, falling back to
+   *   storing authentication token in `localStorage` otherwise
+   * - `cookie` - explicitly disable `localStorage` method, relying only on cookies. May fail due
+   *   to cookies being treated as third-party cookies in some browsers, thus the default is `dual`.
+   * - `token` - explicitly disable cookies, relying only on `localStorage` method
+   */
+  loginMethod?: 'dual' | 'cookie' | 'token'
+
+  /**
+   * Whether to append the providers specified in `providers` with the default providers from the
+   * API, or replace the default providers with the ones specified.
+   */
+  mode?: 'append' | 'replace'
+
+  /**
+   * If true, the "Choose login provider" (eg "Google, "GitHub", "E-mail/password") screen
+   * will be skipped if only a single provider is configured in the `providers` array -
+   * instead it will redirect unauthenticated users straight to the authenticatino URL.
+   */
+  redirectOnSingle?: boolean
+
+  /**
+   * Array of authentication providers to allow.
+   * If not set, the default providers will be used.
+   */
+  providers?: AuthProvider[]
+
+  /**
+   * The API hostname for requests. Should usually be left undefined,
+   * but can be set if using custom cname for API domain.
+   */
+  apiHost?: string
+}
+
+/**
+ * A provider of authentication.
+ *
+ * By default, a list of providers for a project will be fetched from the
+ * {@link https://api.sanity.io/v1/auth/providers | Sanity API}, but you may choose to limit this
+ * list by explicitly defining the providers you want to allow, or add additional custom providers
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/config/types.ts
```diff
@@ -408,32 +413,36 @@ export interface SourceOptions extends PluginOptions {
   apiHost?: string
 
   /** @internal */
-  auth?: AuthStore
+  auth?: AuthConfig | AuthStore
```
bjoerge, line 416: Since we're mentioning this in release notes - should this now be marked as `@beta`?

bjoerge: I guess ideally we'd only want to mark `AuthStore` as internal

packages/sanity/src/core/config/auth/types.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Login method to use for the studio the studio. Can be one of:
+   * - `dual` (default) - attempt to use cookies where possible, falling back to
+   *   storing authentication token in `localStorage` otherwise
+   * - `cookie` - explicitly disable `localStorage` method, relying only on cookies. May fail due
+   *   to cookies being treated as third-party cookies in some browsers, thus the default is `dual`.
+   * - `token` - explicitly disable cookies, relying only on `localStorage` method
+   */
+  loginMethod?: 'dual' | 'cookie' | 'token'
+
+  /**
```
bjoerge, line 17: Just a thought, but would it be slightly better if we typed `providers` as `Provider[] | {mode: 'append' | 'replace'; providers: Provider[]}`?

Reasoning: `mode` is a pretty broad term, and having it on the authconfig makes it seem like it's about "auth mode" (whatever that would be), so maybe more intuitive to make it part of the providers config key

bjoerge (approved):
Looks great - tested the example and it works well.

Added a few questions/suggestions, otherwise LGTM!

packages/sanity/src/core/config/auth/types.ts
```diff
@@ -0,0 +1,74 @@
...
+   * Whether to append the providers specified in `providers` with the default providers from the
+   * API, or replace the default providers with the ones specified.
+   */
+  mode?: 'append' | 'replace'
```
bjoerge, line 21: Just a thought, but would it be slightly better if we typed `providers` as `Provider[] | {mode: 'append' | 'replace'; providers: Provider[]}`?

Reasoning: `mode` is a pretty broad term, and having it on the authconfig itself makes it seem like it's about "auth mode" (whatever that would be), so maybe more intuitive to make it part of the providers config key

packages/sanity/src/core/config/types.ts
rexxars: True, thanks. Will fix.

packages/sanity/src/core/config/auth/types.ts
rexxars: I agree - I think this is directly ported from v2, so the old names stick around. I'll see if I can deprecate the old way of doing things.

rexxars: After thinking about this some more, I actually liked the idea of making this a _reducer_ better, to align with other APIs. I've deprecated the `mode` property, and `providers` is now _either_ a static array (functioning as before, in combination with `mode` for backwards compatibility) or a function that returns an array/promise that resolves to an array.

Let me know if you agree with this approach!

PR #4782, opened 2023-08-03
feat(core): adds title to document panel

binoy14:
### Description

Adds a document title to the document panel. Note this does not include the title treatment when editing just shows the title. 

#### Long Title
Screenshot 2023-08-03 at 11 24 59 AM

#### Short Title
Screenshot 2023-08-03 at 11 25 09 AM

### What to review

1. Does the hook makes sense
2. Does the location of the hook make sense

### Notes for release

N/A (Part of bigger release announcement)

### DIFF
--- a/packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
+++ b/packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
@@ -1,32 +1,12 @@
 import React, {ReactElement} from 'react'
-import {useDocumentPane} from '../../useDocumentPane'
-import {unstable_useValuePreview as useValuePreview} from 'sanity'
+import {useDocumentTitle} from '../../useDocumentTitle'
 
 export function DocumentHeaderTitle(): ReactElement {
-  const {connectionState, schemaType, title, value: documentValue} = useDocumentPane()
-  const subscribed = Boolean(documentValue) && connectionState === 'connected'
-
-  const {error, value} = useValuePreview({
-    enabled: subscribed,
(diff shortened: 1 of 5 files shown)


packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
```diff
@@ -1,32 +1,18 @@
...
-    return <>New {schemaType?.title || schemaType?.name}</>
-  }
-
   if (error) {
-    return <>Error: {error.message}</>
+    return <>{error}</>
   }
 
-  return <>{value?.title || <span style={{color: 'var(--card-muted-fg-color)'}}>Untitled</span>}</>
+  return <>{title || <span style={{color: 'var(--card-muted-fg-color)'}}>Untitled</span>}</>
```
RitaDias, line 17: Small nitpick: Can we move the style out of the inline document?

RitaDias (approved):
Nice!

I guess for my side because I am not used to seeing the title duplicated (both at right the top and on the document itself) it comes off as strange but that's just my two half cents 😂 

Everything looks good to me :)

image

packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
binoy14: actually, after looking at this, I realized that I will be refactoring this in https://github.com/sanity-io/sanity/pull/4689 so it might not be worth changing it in this PR

bjoerge (approved):
LGTM

PR #4795, opened 2023-08-04
fix: delay refetching for listener events arriving with visibility != query

bjoerge:
### Description
The listener requests we use for document lists are set up with visibility=query, however, this does not guarantee that the listener events arrive after their mutations have been made visible to queries.

From our docs about [listener visibility](https://www.sanity.io/docs/listening#visibility-c4786e55c3ff):
> (...) listeners with `query` may in certain cases (notably with deferred transactions) receive events that are not yet visible to queries. The visibility event field will indicate the actual visibility.

This will often enough lead to document lists in the Studio not reflecting the actual reality of the data store. One fairly consistent way of reproducing the issue is to create a new document, make a single change (e.g. type a single character in a text field) and watch the newly created document *not* appearing in the document list before another change is made.

### What to review
- Make sure the document lists updates more reliably when documents are added/removed

### Notes for release

- Fixed an issue that would sometimes cause document lists to not refresh properly after adding or removing a document

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/listenQuery.ts
+++ b/packages/sanity/src/core/store/_legacy/document/listenQuery.ts
@@ -1,5 +1,5 @@
 import {SanityClient} from '@sanity/client'
-import {asyncScheduler, defer, merge, Observable, of, partition, throwError} from 'rxjs'
+import {asyncScheduler, defer, merge, Observable, of, partition, throwError, timer} from 'rxjs'
 import {filter, mergeMap, share, take, throttleTime} from 'rxjs/operators'
 import {exhaustMapWithTrailing} from 'rxjs-exhaustmap-with-trailing'
 import {MutationEvent, ReconnectEvent, WelcomeEvent} from './types'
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
Tested the change and working as intended in preview, code makes sense but will definitely defer to Espen for final word. Thanks!

packages/sanity/src/core/store/_legacy/document/listenQuery.ts
```diff
@@ -92,11 +92,23 @@ export function listenQuery(
...
       throttleTime(options.throttleTime || 1000, asyncScheduler, {leading: true, trailing: true})
     )
-  ).pipe(exhaustMapWithTrailing(() => fetch(client, fetchQuery, params, options)))
+  ).pipe(
+    exhaustMapWithTrailing((event) => {
+      if (event.type === 'mutation' && event.visibility !== 'query') {
+        // Even though the listener request specifies visibility=query, the events are not guaranteed to be delivered with visibility=query
+        // If the event we are responding to arrives with visibility != query, we add a little delay to allow for the updated document to be available for queries
+        // See https://www.sanity.io/docs/listening#visibility-c4786e55c3ff
+        return timer(1200).pipe(mergeMap(doFetch))
```
rexxars, line 109: Can you briefly explain why this using a timer here works?
> After given duration, emit numbers in sequence every specified duration
What is making this only do a single refetch instead of one every 1200ms?

bjoerge: `timer(duration)` (with a single argument) waits for the `duration`ms, emits `0` and then completes. Sounds like the docs you refer to here is about the timer(duration, interval) overload, which first waits for `duration`, then emits at every `interval`?

rexxars: Ah, yes. Thanks!

rexxars (approved):

PR #4798, opened 2023-08-07
fix: add `node.module` export condition

stipsan:
### Description

The ESM saga never ends 💃
This time there's trouble in build tools that understand the `node` and `import` conditions but don't support the CJS re-export pattern we use for node to protect against the "Dual Package Hazard" among other things.
How do we solve it?
By adding a condition that bundlers understand, but node ignores: `node.module`.

### What to review

If it builds it works, if anything the bundlesize on a lot of systems will be smaller if they previously used `node.import` as the ESM to CJS detour adds some fluff. Now they'll cut the fluff and be able to stay in ESM land.

### Notes for release

- fix(esm): add `node.module` conditions, improves compatibility with Webpack, Astro, Svelte and Vue


rexxars (approved):

rexxars:
Converted to draft until the next issues are resolved.

rexxars:
> A fix has landed in a recent `canary`: [vercel/next.js#53831](https://github.com/vercel/next.js/pull/53831)

Does this mean this PR is ready for review now, or that it is no longer necessary?

stipsan:
> > A fix has landed in a recent `canary`: [vercel/next.js#53831](https://github.com/vercel/next.js/pull/53831)
> 
> Does this mean this PR is ready for review now, or that it is no longer necessary?

It means we're ready for review :) I'll rebase and fix the conflicts 🙌

bjoerge:
@stipsan is this still relelvant? Anything else needs doing here?

stipsan:
> @stipsan is this still relevant? Anything else needs doing here?

We're seeing signal that it is increasingly relevant in embedded studios for plugins. [We recently needed to add it to `@sanity/assist`](https://github.com/sanity-io/assist/commit/7d987671a51dd550278d245e904900c9fb419088) for example to fix a build crash.
We haven't seen similar signal for `sanity` itself at this time, and so it's just a minor enhancement.
At the same time we've seen this pattern cause problems in some build setups, [such as the maintainers of `nuxt` in `@sanity/client](https://github.com/sanity-io/client/pull/318).`

Because of the benefit being so small, and there being a risk that it might throw a wrench in someone's build, I'm not sure we should merge this 🤔

bjoerge:
Thanks for clarifying. Let's leave it open for now and revisit later. I've unassigned the dx team as reviwewers, feel free to re-assign if/when you think we should pick it up again.

skogsmaskin (approved):
LGTM as code owner (not really relevant here)

PR #4806, opened 2023-08-08
feat(core, desk): convert review changes panel strings to use i18n primitives

RitaDias:
### Description

Convert review changes panel strings to use i18n primitives + define some basic guidelines (link in async check point)

### What to review

That all items/strings within the changelog / review changes are not broken and work as expected:
- Fix broken issue with curly brackets that was making the time look broken "`({count}m`" https://github.com/sanity-io/sanity/pull/4806/commits/8f0b61fc1fddc500498db84baa61421c4d50c8a4
- Add i18n primitives to items in review panel (and timeline menus by extension)

Items to keep in mind: 
- Possible duplication of keys when it's not needed (and how we want to have it across different packages - `loading` / `cancel`) and implications that might have
- Actions / naming conventions

### Notes for release

N/A


packages/sanity/src/core/field/diff/components/ChangeList.tsx
```diff
@@ -133,7 +143,7 @@ export function ChangeList({diff, fields, schemaType}: ChangeListProps): React.R
               <Button
                 tone="critical"
                 mode="ghost"
-                text="Revert all changes"
+                text={t('core.review-changes.revert-button-revert-all')}
```
RitaDias, line 146: This button and the button on line 130 didn't used to have the same text. However, as I was doing this, it feels like in the spirit of
1. keeping the number of translated keys down (given how minimal the change is) 
2. the spirit of the buttons are basically the same

I made the button have the same text, however, this has visual implications which are not great (check the table at the bottom). I'm tempted to simply revert the change and make them two separate keys as they were before, however, I'm sure this question will come up somewhat frequently as I go forward 🤔  Thoughts? (cc @rexxars @mikolajdobrucki )

| Before | After |
|--------|-------|
|  image |     image |

RitaDias: I was also considering on changing the text on the button within the popper to "Revert changes" since that is more widely used in other places too.

RitaDias: I was also thinking of changing the text on the button within the prompt popper to "revert changes" since it's more widely used in other places as well 🤔

RitaDias: this question also sort of applies to situations like while this message (revert all changes) shows the count (2 changes like the example ☝️ ) there are situations (like in arrays when you want to revert the whole array - image) where we don't show the count but the message is pretty similar. 

image

DX wise I can see the argument from both sides where it's annoying having to create so many translations for such similar prompts (and expensive depending on how many languages companies pay translation agencies for) but also it might make sense to keep them as separate things?

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -60,9 +60,38 @@ const studioLocaleStrings = {
   'timeAgo.seconds.ago_one': '{{count}} minute ago',
   'timeAgo.seconds.ago_other': '{{count}} second ago',
   /* Relative time, granularity: seconds, using a minimal format*/
-  'timeAgo.seconds.minimal': '{{count}m',
+  'timeAgo.seconds.minimal': '{{count}}m',
```
RitaDias, line 63: @rexxars I made this change just for the sake of fixing the interface.
However, this whole section (how we're doing the time) is something that I strongly suggest we change / minimise when it comes to the number of keys. 
From a DX perspective (the number of things I as a dev would need to write) and from a Business perspective (how expensive I see this becoming) I feel is what makes sense.

I will probably keep it outside of the scope of this PR and move it to its own story but a thought to keep in mind.

packages/sanity/src/core/field/diff/components/ChangeList.tsx
RitaDias: @mikolajdobrucki just to double check: the path we're choosing is then to change to "revert all" such as the image, yes?

image

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -42,6 +42,39 @@ const deskLocaleStrings = {
...
+  'desk.review-changes.title': 'Review changes',
+
+  /** Label for the close button label in Review Changes pane */
+  'desk.review-changes.close-label': 'Close review changes',
+
+  /** Label and text for differences tooltip that indicates the authors of the changes */
+  'desk.review-changes.changes-by-author': 'Changes by',
+
+  /** Loading changes in Review Changes Pane */
+  'desk.review-changes.loading-changes': 'Loading changes',
```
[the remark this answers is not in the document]
RitaDias: As well as 'desk.review-changes.loading-author' (https://github.com/sanity-io/sanity/pull/4806/files#diff-5c6b73dd5c1114ae88bf772a5e5f8cb88d3b696adfed0df4bc2e450c6ada078fR86) except this is in a different package all together

packages/sanity/src/core/i18n/bundles/studio.ts
rexxars: Yeah, I also think that this is _potentially_ the wrong approach for relative time anyway - there is a `Intl.RelativeTime` API that I think is probably a better candidate. I want to see if we can use that instead of manually defining these strings - so we should create a story for "Investigate best approach for relative time"

RitaDias: Created a story for this 👍

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
```diff
@@ -136,10 +138,12 @@ export function TimelineMenu({chunk, mode, placement}: TimelineMenuProps) {
...
   const revLabel = chunk
-    ? `${upperFirst(formatTimelineEventLabel(chunk.type))}: ${timeLabel}`
-    : 'Latest version'
+    ? `${upperFirst(t(`desk.timeline.${chunk.type}`))}: ${timeLabel}`
+    : t('desk.timeline.latest-version')
 
-  const sinceLabel = chunk ? `Since: ${timeLabel}` : 'Since: unknown version'
+  const sinceLabel = chunk
+    ? t('desk.timeline.since', {timeLabel: timeLabel})
+    : t('desk.timeline.since-version-missing')
```
rexxars, line 146: These ones are somewhat complicated. Ideally we'll want to remove the  `useFormattedTimestamp`  here and use the [`datetime` formatter](https://www.i18next.com/translation-function/formatting#datetime) (based on `Intl.DateTimeFormat`).

Something like:
```ts
const sinceLabel = chunk
  ? t('timeline.since', {
      timestamp,
      formatParams: {timestamp: {dateStyle: 'medium', timeStyle: 'short'}},
    })
  : t('timeline.since-version-missing')
```
And for the locale resource:
```ts
const resources = {
  // ...
  'timeline.since': 'Since: {{timestamp, datetime}}'
}
```

This will format the date according to the current locale and calendar, eg in US English it would give `Aug 22, 2021, 1:15 AM` while in Norwegian it would give `22. aug. 2021, 01:15`, and in Thai `22 ส.ค. 2564 01:15`.

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
```diff
@@ -136,10 +138,12 @@ export function TimelineMenu({chunk, mode, placement}: TimelineMenuProps) {
   const timeLabel = useFormattedTimestamp(chunk?.endTimestamp || '')
 
   const revLabel = chunk
-    ? `${upperFirst(formatTimelineEventLabel(chunk.type))}: ${timeLabel}`
-    : 'Latest version'
+    ? `${upperFirst(t(`desk.timeline.${chunk.type}`))}: ${timeLabel}`
```
rexxars, line 141: This doesn't need/shouldn't have the `upperFirst` call anymore, since we can now localize it specifically for the exact location it is used. It is also unclear whether or not all languages have the same casing rules.

rexxars (changes requested):
I know the previous localized strings were not really following any established pattern - treat them as "developer tests" rather than "well thought out naming patterns" 😅 With that in mind:

- Few thoughts/notes on naming:
  1. Do we need/want the `core` prefix? I like the `review-changes` prefix, but I don't quite know what is "core" and not. My hunch is to remove it.
  2. I'm wondering if we should have some sort "action" subgroup and naming convention. I can imagine we'll end up with a number of these "message, confirm, cancel" groups for prompts etc. 
  3. There is a bit of a mixture of the pattern on some of these, eg `review-changes.revert-button-cancel` and `review-changes.close-label`. I prefer not to "button" in the name - I'd perhaps go for "action" instead?
  4. I honestly find it hard to figure out quite what the right "syntax"/shape of these keys are - it would be great if you could try to think of some guidelines/rules for how we shape these throughout the studio so we can be consistent. I'd take a brief look through some of the other areas of the studio that needs localization to see if you can find some repeated patterns in terms of what you are translating, and then build a pattern from it. I can try to help if you want me to, but would be great if you took an initial stab at it!
- The "namespace" for each "resource bundle" is already part of the the strings, implicitly.
  In it's most bare form, all the strings within the "desk" namespace looks like: `desk:action.publish.waiting`, so adding the `desk.` prefix in front of each string ends up like this: `desk:desk.review-changes.title`. 
- All the locale resources for a given feature should be in the same namespace. It is unfortunate that code-wise we have parts of the feature in desk and part of it in core. I've raised this internally for discussion - perhaps we could split it into "diff" and "review changes".
- The comments above the individual resource keys are meant to tell both a human and (ideally) an AI what it needs to translate it. We shouldn't over-prioritize this point, but keep it in mind when writing these labels (eg an AI doesn't (necessarily) have a clue what Sanity or a "Review Changes pane" is, but "Text displayed when there are no changes to review" might be.

RitaDias:
👋 Going to go through the points so it's easier to identify what I'm talking about!

>  
>   1. Do we need/want the `core` prefix? I like the `review-changes` prefix, but I don't quite know what is "core" and not. My hunch is to remove it.

My line of thinking was to make it easier to identify in which packages they exist in. However, I was not aware of the following point you mentioned further down on the list: 
>In it's most bare form, all the strings within the "desk" namespace looks like: `desk:action.publish.waiting`, so adding the `desk.` prefix in front of each string ends up like this: `desk:desk.review-changes.title`.

Given that, I went ahead and removed the prefixes (both the `core` and `desk`) that I had  :) c5cc1f99755580122e4d391f16063e3ae95fe07f 4942bb1c52bd8ab709260761e235aa00232ec6bd

>   2. I'm wondering if we should have some sort "action" subgroup and naming convention. I can imagine we'll end up with a number of these "message, confirm, cancel" groups for prompts etc.
>   3. There is a bit of a mixture of the pattern on some of these, eg `review-changes.revert-button-cancel` and `review-changes.close-label`. I prefer not to "button" in the name - I'd perhaps go for "action" instead?
>   4. I honestly find it hard to figure out quite what the right "syntax"/shape of these keys are - it would be great if you could try to think of some guidelines/rules for how we shape these throughout the studio so we can be consistent. I'd take a brief look through some of the other areas of the studio that needs localization to see if you can find some repeated patterns in terms of what you are translating, and then build a pattern from it. I can try to help if you want me to, but would be great if you took an initial stab at it!

My problem with "action" is that it's pretty vague (in definition when looking at the key without knowledge of what it might mean internally for those that set the system up). Because my first instinct is that it's solely based on (buttons, prompts) but I'm unsure if my assumption is correct (or if it's the same assumption that everyone that comes across it will have). So I guess my question is: as part of the developing guidelines, do we want that to be the definition for "action" subgroup?

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -42,6 +42,67 @@ const deskLocaleStrings = {
...
+  /** Label for determining since which version the changes for timeline menu dropdown are showing.
+   * Receives the time label as a parameter.
+   */
+  'desk.timeline.since': 'Since: {{timeLabel}}',
+
+  /** Label for missing change version for timeline menu dropdown are showing */
+  'desk.timeline.since-version-missing': 'Since: unknown version',
+
+  /** Title for error when the timeline for the given document can't be loaded */
+  'desk.timeline.error-title': 'An error occurred whilst retrieving document changes.',
```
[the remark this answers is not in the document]
RitaDias: Replied to this in part in the following comment https://github.com/sanity-io/sanity/pull/4806#issuecomment-1676834014 :)

bjoerge (approved):
Looks great!

binoy14 (approved):
very nice!

rexxars (approved):

PR #4809, opened 2023-08-09
chore: upgrade typescript to 5.x

bjoerge:
### Description
This upgrades TypeScript to latest 5.x. Very few changes needed to be done to the code base except from adjusting and adding few expect-error directives (kept in a separate commit 909b93fc26367e3cd8671cb8638f789c99f47f58).

### What to review
- Can this have unintended consequences? From what I can tell by reading the [release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#breaking-changes-and-deprecations) there doesn't seem to be any breaking changes with the declaration output, so types generated by 5.x should still work in projects using 4.x

### Notes for release

N/A - transparent to end users/devs

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     "pretty-quick": "^3.1.3",
     "rimraf": "^3.0.2",
     "semver": "^7.3.5",
-    "typescript": "^4.9.4",
+    "typescript": "^5.1.6",
     "yargs": "^17.3.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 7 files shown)


skogsmaskin (approved):

rexxars (approved):

bjoerge:
> why upgrade now

Mostly a matter of keeping our stuff updated, but also that TS5 has seen quite significant speed and memory optimizations, but I also did this as part of debugging a bundler issue, so figured I could just take it into main :)

PR #4812, opened 2023-08-09
chore(ci): add turborepo to speed up the `yarn build:cli` task significantly

stipsan:
### Description

A ton of time is spent, locally, on the GitHub Actions CI, as well as Vercel test-studio builds, on the `yarn build:cli` task. On a fast M1 Pro Mac it usually takes at least 2 minutes to build every single time.

Turborepo have a lot of features, but it's speed, especially combined with Remote Caching, is such an improvement to the build process that it makes sense to start off with just that slow build task and later slowly expand as we find new bottlenecks.

### What to review

Builds should be exceptionally faster. Especially after you enable [Remote caching](https://turbo.build/repo/docs/core-concepts/remote-caching) on your local machine:
```bash
yarn
yarn turbo login
yarn turbo link # use the Sanity Vercel org
yarn build # Reports if remote cache is enabled
```

If you're setup correctly then you should be able to run these steps:
```bash
yarn build
rm -rf ./node_modules/.cache/turbo
yarn clean
yarn build
```
And still see `FULL TURBO` at the end.

On Vercel builds a new "Run Summary" button appears in the "Building" accordion:
image

Clicking it shows detailed information on the build
image

And on GitHub Actions each workspace in a Turborepo task is nicely grouped and collapsed by default:
image

Expanding them lets you see the complete build output of the job. It doesn't matter if the cache is hit or miss you always have access to the build log:
image

### Notes for release

N/A


package.json
```diff
@@ -12,9 +12,10 @@
     "packages/groq",
     "packages/sanity"
   ],
+  "packageManager": "yarn@1.22.17",
```
stipsan, line 15: Settings this property makes `turbo` faster, and prevents someone from accidentally doing `npm install` on modern node.

bjoerge (approved):
This is fantastic @stipsan <3

bjoerge:
@rexxars, @skogsmaskin - thoughts?

skogsmaskin:
> @rexxars, @skogsmaskin - thoughts?

I would like having faster builds that's for sure, and the speedup is pretty damn great by using this :) 

I'm positive, especially if just for the CLI build task for now. Then reevaluate after some time used.

Any known downsides by using this?

skogsmaskin (approved):

stipsan:
> > @rexxars, @skogsmaskin - thoughts?
> 
> 
> Any known downsides by using Turborepo (or similar)?
> 
> 

Only that if the build output in a package changes, say, it starts using a `/dist` folder instead of `/lib`, then you'll also have to update the `turbo.json` with that change so it's in the cache.
It's not a big deal though because if it's not in the cache you'll get errors on the CI.
Same is true if the build starts using new process.env variables that should invalidate a build. Basically anything that matters to the build and that turborepo can't "see" by just looking at the files under git alone, has to be declared and kept in sync as the monorepo evolves and changes how packages are bundled and published over time. 

I feel that's an ok trade off when you look at all the time you save by speeding up these builds 🤯

rexxars (approved):

PR #4813, opened 2023-08-10
fix(desk): handle edge cases when documents are deleted on a specific revision

robinpyon:
### Description

This PR handles primarily addresses two issues:

1. An edge case where the timeline menu wouldn't be accessible:
    - User 1 selects an older document revision
    - User 2 deletes the document
    - User 1 is able to restore the document (to the most recent version), but is then unable to access the timeline menu
2. Review changes' dropdowns not working (due to the timeline menu being disabled) in deleted documents 

Other changes have been outlined as comments!

Finally, a video demonstrating some of the paths that were tested:

https://github.com/sanity-io/sanity/assets/209129/f028ead1-3203-46bf-b6a6-63c5fc1682a2

### DIFF
--- a/packages/sanity/src/desk/panes/document/documentPanel/DeletedDocumentBanner.tsx
+++ b/packages/sanity/src/desk/panes/document/documentPanel/DeletedDocumentBanner.tsx
@@ -1,46 +1,49 @@
-import React, {useCallback, useMemo} from 'react'
+import React, {useCallback} from 'react'
 import {Button, Card, Container, Flex, Text} from '@sanity/ui'
 import {ReadOnlyIcon} from '@sanity/icons'
 import styled from 'styled-components'
 import {useDocumentPane} from '../useDocumentPane'
-import {useDocumentOperation, useTimelineSelector} from 'sanity'
+import {useDocumentOperation} from 'sanity'
+import {useRouter} from 'sanity/router'
 
 const Root = styled(Card)`
   position: relative;
   z-index: 50;
 `
-export function DeletedDocumentBanner() {
-  const {documentId, documentType, timelineStore} = useDocumentPane()
-  const chunks = useTimelineSelector(timelineStore, (state) => state.chunks)
-  const lastRevisionId = useMemo(() => {
-    const lastRevision = chunks.filter((chunk) => !['delete', 'initial'].includes(chunk.type))
 
-    return lastRevision?.[0]?.id
-  }, [chunks])
+interface DeletedDocumentBannerProps {
+  revisionId?: string | null
+}
+
+export function DeletedDocumentBanner({revisionId}: DeletedDocumentBannerProps) {
+  const {documentId, documentType} = useDocumentPane()
   const {restore} = useDocumentOperation(documentId, documentType)
+  const {navigateIntent} = useRouter()
   const handleRestore = useCallback(() => {
-    restore.execute(lastRevisionId)
-  }, [restore, lastRevisionId])
+    if (revisionId) {
+      restore.execute(revisionId)
+      navigateIntent('edit', {id: documentId, type: documentType})
+    }
+  }, [documentId, documentType, navigateIntent, restore, revisionId])
 
   return (
     <Root data-testid="deleted-document-banner" shadow={1} tone="transparent">
-      <Container paddingX={4} paddingY={lastRevisionId ? 2 : 3} sizing="border" width={1}>
+      <Container paddingX={4} paddingY={revisionId ? 2 : 3} sizing="border" width={1}>
         <Flex align="center">
           <Text size={1}>
             <ReadOnlyIcon />
           </Text>
 
           <Flex align="center" gap={2} flex={1} marginLeft={3}>
-            <Text size={1}>This document is deleted and can’t be edited.</Text>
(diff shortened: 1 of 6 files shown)


packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
```diff
@@ -161,7 +155,7 @@ export function TimelineMenu({chunk, mode, placement}: TimelineMenuProps) {
       ref={setPopover}
     >
       <Button
-        disabled={!ready || isDeleted}
+        disabled={!ready}
```
robinpyon, line 158: **suggestion**: we don't disable timeline menu items _for now_, as doing so means you're unable to navigate review changes on deleted documents

there's probably a greater question over whether review changes should even be visible in deleted documents to begin with – this whole area is probably due for a bit of a visual overhaul

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
```diff
@@ -45,12 +45,6 @@ export function TimelineMenu({chunk, mode, placement}: TimelineMenuProps) {
     setOpen(false)
   }, [setTimelineMode])
 
-  useEffect(() => {
-    if (open && isDeleted) {
-      handleClose()
-    }
-  }, [open, isDeleted, handleClose])
```
robinpyon, line 52: **suggestion**: consider removing, as this can play havoc when using review changes on deleted documents (see below).

packages/sanity/src/desk/panes/document/timeline/timelineItem.tsx
```diff
@@ -43,9 +41,7 @@ export function TimelineItem({
 }: TimelineItemProps) {
   const iconComponent = getTimelineEventIconComponent(type)
   const authorUserIds = Array.from(chunk.authors)
-  const isSelectable = useMemo(() => {
-    return !NOT_SELECTABLE_TYPES.includes(type)
-  }, [type])
```
robinpyon, line 48: **question:** what's the rationale behind making 'discarded draft' revs non selectable?

**suggestion:** we allow discarded drafts to be selectable as it was prior – there are cases where users may have a discarded draft as their most recent version, and will need to select it to return to the most up to date / recent version

Making deleted revisions non-selectable is a great shout though!

packages/sanity/src/desk/panes/document/documentPanel/DeletedDocumentBanner.tsx
```diff
@@ -1,46 +1,49 @@
...
+      <Container paddingX={4} paddingY={revisionId ? 2 : 3} sizing="border" width={1}>
         <Flex align="center">
           <Text size={1}>
             <ReadOnlyIcon />
           </Text>
 
           <Flex align="center" gap={2} flex={1} marginLeft={3}>
-            <Text size={1}>This document is deleted and can’t be edited.</Text>
-            {lastRevisionId && (
+            <Text size={1}>This document has been deleted.</Text>
```
robinpyon, line 38: **suggestion**: use terser language, it should (hopefully) be self evident that the document can't be edited!

packages/sanity/src/desk/panes/document/documentPanel/DeletedDocumentBanner.tsx
```diff
@@ -1,46 +1,49 @@
...
+            {revisionId && (
               <Button
-                radius={0}
```
robinpyon, line 37: **nitpick**: AFAIK we don't have a precedent for using buttons with no radii

packages/sanity/src/desk/panes/document/documentPanel/DeletedDocumentBanner.tsx
```diff
@@ -1,46 +1,49 @@
...
   z-index: 50;
 `
-export function DeletedDocumentBanner() {
-  const {documentId, documentType, timelineStore} = useDocumentPane()
-  const chunks = useTimelineSelector(timelineStore, (state) => state.chunks)
-  const lastRevisionId = useMemo(() => {
-    const lastRevision = chunks.filter((chunk) => !['delete', 'initial'].includes(chunk.type))
 
-    return lastRevision?.[0]?.id
-  }, [chunks])
```
robinpyon, line 19: **suggestion:** this was removed in favour of creating a targeted selector that only returns the last non-deleted revision ID. 

Whilst not super applicable to this banner (as it's only visible when the document is deleted), selecting `chunks` can cause a bunch of unnecessary renders as it updates very frequently

packages/sanity/src/desk/panes/document/timeline/timelineItem.tsx
sjelfull: Great point, I believed it was similar to the deleted event, and thought it didn't make sense to select it.

sjelfull (approved):

packages/sanity/src/desk/panes/document/timeline/timelineMenu.tsx
sjelfull: Agreed, we can tackle this later. Seems to work well regardless.

PR #4815, opened 2023-08-10
fix(core): timeouts randomly being cancelled

### LINKED ISSUE #4814: Sanity code randomly cancelling my `setTimeout` timeouts
**Describe the bug**

Sanity Studio code is calling `clearTimeout` with values that are returned by `requestIdleCallback`. `clearTimeout` should only be called with IDs returned by `setTimeout. This is resulting in the timeouts that I set in my code being randomly cancelled and never running.

**To Reproduce**

I have have custom code like this in my Sanity Studio codebase:

```javascript
const timeoutId = setTimeout(() => {
  console.log("timeout!!!")
}, 10000)

console.log("timeout set", timeoutId)
```

I see "timeout set 2" in the console but my timeout function never runs. After some debugging I discovered that [this line](https://github.com/sanity-io/sanity/blob/v3.15.0/packages/sanity/src/core/components/hookCollection/requestIdleCallback.ts#L15) is calling `clearTimeout` with my ID of `2`. Sanity's code got the number `2` from a call to `requestIdleCallback`, not `setTimeout`.

**Expected behavior**

I would expect Sanity's code to not randomly cancel my timeouts.

**Screenshots**

I monkey patched `window.clearTimeout` for debugging purposes. I made it throw an exception if `clearTimeout` was called my timeout ID.

image

3.15.0

Latest MacOS

balexand:
Fixes #4814

### DIFF
--- a/packages/sanity/src/core/components/hookCollection/requestIdleCallback.ts
+++ b/packages/sanity/src/core/components/hookCollection/requestIdleCallback.ts
@@ -18,7 +18,7 @@ const cancelIdleCallbackShim: Window['cancelIdleCallback'] = (handle: unknown) =
 const _requestIdleCallback =
   typeof requestIdleCallback === 'undefined' ? requestIdleCallbackShim : requestIdleCallback
 const _cancelIdleCallback =
-  typeof cancelIdleCallback === 'undefined' ? cancelIdleCallbackShim : cancelIdleCallbackShim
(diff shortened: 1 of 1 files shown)


rexxars:
Wild! Amazing work debugging to find this issue ❤️
Much appreciated!

rexxars (approved):

binoy14 (approved):
Great catch, thank you!

PR #4818, opened 2023-08-10
fix(desk): prevent isDeleted false positives when timeline isn't available, colocate pane deletion logic in DocumentPaneProvider

robinpyon:


packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -640,6 +646,17 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
   }, [connectionState, pushToast])
 
+  /**
+   * If the document was deleted successfully, close the pane.
+   */
+  useEffect(() => {
+    if (!documentOperationEvent) return
+    if (documentOperationEvent.type === 'success' && documentOperationEvent.op === 'delete') {
+      // Wait until next tick to allow deletion toasts to appear first
+      setTimeout(() => paneRouter.closeCurrentAndAfter(), 0)
```
sjelfull, line 656: Unsure if this is totally safe from race conditions? Will do some testing.

sjelfull: Seems to work well!

sjelfull (approved):

PR #4824, opened 2023-08-14
feat(core): field actions improvements

hermanwikner:
### Description

This pull request introduces updates to the Field Actions API and UI improvements to field actions and presence in form fields.

**Field Actions API update**
Currently, the field actions node is resolved at a higher level and then passed on to the field component as a prop. This means that you can use `renderDefault` for a form field component and modify the action node to be whatever you want. For example:

```tsx
import {definePlugin} from 'sanity'

const myField = definePlugin({
  name: 'my-field',
  form: {
    components: {
      field: (props) => {
        const actions = (
          <div>
            {props.actions}
            <button onClick={() => alert('Hello!')}>My action</button>
          </div>
        )

        return props.renderDefault({
          ...props,
          actions,
        })
      },
    },
  },
})
```

Although the current approach has some advantages, it introduces challenges in terms of UI maintenance for field actions. To have more control over the UI, this pull request updates so that field actions are always configured as a field action definition, and not as a React node. This means that the example above would instead work like this:

```tsx
import {definePlugin, defineDocumentFieldAction} from 'sanity'

const myAction = defineDocumentFieldAction({
  name: 'my-action',
  useAction() {
    return {
      title: 'My action',
      onAction: () => alert('Hello!'),
      type: 'action',
    }
  },
})

const myField = definePlugin({
  name: 'my-field',
  form: {
    components: {
      field: (props) => {
        const nextActions =  [...props.actions, myAction]

        return props.renderDefault({
          ...props,
          actions: nextActions,
        })
      },
    },
  },
})

```

This update not only improves the handling and maintenance of the UI, but also provides more control over which field actions are passed on to a form field component. This is because you can now filter the field actions array based on the eg the `name` field – which is not possible today. 

```tsx
import {definePlugin} from 'sanity'

const myField = definePlugin({
  name: 'my-field',
  form: {
    components: {
      field: (props) => {
        const nextActions = props?.actions?.filter((a) => a.name !== 'some-action')

        return props.renderDefault({
          ...props,
          actions: nextActions,
        })
      },
    },
  },
})
```

**Form Field UI update**
Currently, there are some problems with how presence and field actions are rendered together in a form field. This pull request improves this so that presence and field actions never overlap, and instead render next to each other when both exist in a field.

This pull request also updates so that presence and field actions are absolutely positioned, which means that they cannot cause any layout changes when they become visible, which can happen today if, for example, a field has a long description or title.

Untitled

### Notes
- There are some other improvement areas for field actions, in particular accessibility, which the focus has not been to improve in this pull request
- The `assist` plugin uses the old Field Actions API which makes the Studio crash. Therefore, that plugin is temporarily disabled.

### What to review
- Field actions should work as before
- Could the changes made to presence (e.g `FieldPresence`) introduce any problems elsewhere in Studio?

### Notes for release

n/a


bjoerge (approved):

packages/sanity/src/core/form/components/formField/FormFieldHeader.tsx
```diff
@@ -1,75 +1,69 @@
...
-  // due to their default vertical alignment being baseline.
-  lineHeight: 1,
-})
+  /** @hidden */
+  fieldHovered: boolean
+  /** @hidden */
+  fieldFocused: boolean
+}
 
 export const FormFieldHeader = memo(function FormFieldHeader(props: FormFieldHeaderProps) {
```
robinpyon, line 41: **question**: is `FormFieldHeader` being imported anywhere? From what I can gather it isn't – if so, we should consider removing it

packages/sanity/src/core/presence/FieldPresence.tsx
```diff
@@ -60,35 +56,30 @@ export const FieldPresenceInner = memo(function FieldPresenceInner({
...
-                    zIndex: 100 - i,
-                  }}
-                >
-                  {av.element}
-                </div>
-              )
-          )}
-        </InnerBox>
-      </PresenceTooltip>
+      <InnerBox direction="row-reverse">
```
robinpyon, line 63: **question:** really minor, but was there a reason for disabling presence tooltips when `DISABLE_OVERLAY = true`?

robinpyon (approved):
Thanks @hermanwikner! 

Looks good to me! And good call with cleaning up `<FieldActionsResolver>` usage. Left some questions but nothing blocking

Upon testing I did notice one minor virtual artfiact (video below – focus on the top-right most presence avatar). It only seems to happen on fields with actions, but it's _really minor_ – and something we should be able to address if/when we simplify presence throughout.

https://github.com/sanity-io/sanity/assets/209129/57773af7-4ac8-4519-9392-12122421f750

packages/sanity/src/core/form/components/formField/FormFieldHeader.tsx
hermanwikner: I'll remove it 👍

packages/sanity/src/core/presence/FieldPresence.tsx
hermanwikner: I had to move the tooltip to `FieldPresenceWithOverlay` to make it work together with `onMouseEnter` and `onMouseLeave` interactions to display field actions.

bjoerge (approved):

PR #4826, opened 2023-08-14
fix(core): close navbar tooltips when menubutton is open

ninaandal:
### Description
Tooltips remained on screen after opening a `MenuButton`  (most notably on both the Workspaces + Help navbar buttons) in the navbar. 
To prevent the tooltip from showing underneath the `MenuButton` when it is open, the `tooltip` is `disabled` when the `MenuButton` is open. 

### What to review
The `Workspaces` and the `Help and Resources` dropdowns - make sure that they work as expected. 

### Notes for release
Fixes annoyance where tooltip was showing under menubutton.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
@@ -137,6 +137,8 @@ export function StudioNavbar() {
     setDrawerOpen(true)
   }, [])
 
+  const [isWorkspaceMenuOpen, setIsWorkspaceMenuOpen] = useState(false)
+
   return (
     <RootLayer zOffset={100} data-search-open={searchFullscreenOpen}>
       <RootCard
@@ -179,9 +181,10 @@ export function StudioNavbar() {
                   placement="bottom"
                   portal
                   scheme={scheme}
+                  disabled={isWorkspaceMenuOpen}
                 >
                   <Box>
-                    <WorkspaceMenuButton />
+                    <WorkspaceMenuButton setIsMenuOpen={setIsWorkspaceMenuOpen} />
                   </Box>
                 </Tooltip>
               </Box>
--- a/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
@@ -137,8 +137,6 @@ export function StudioNavbar() {
     setDrawerOpen(true)
   }, [])
 
-  const [isWorkspaceMenuOpen, setIsWorkspaceMenuOpen] = useState(false)
-
   return (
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/studio/components/navbar/workspace/WorkspaceMenuButton.tsx
```diff
@@ -89,6 +93,8 @@ export function WorkspaceMenuButton(props: ButtonProps) {
             })}
         </StyledMenu>
       }
+      onClose={() => props.setIsMenuOpen && props.setIsMenuOpen(false)}
```
hermanwikner, line 96: I recommend a slight rewrite of this code to enhance readability and to align with our ESLint rule that discourages the use of arrow functions in JSX props (`react/jsx-no-bind`).  

I suggest this update:  

```tsx
export function WorkspaceMenuButton(props: WorkspaceMenuButtonProps) {
  const {onMenuOpenChange} = props

  // ... rest of code

  const handleOnOpen = useCallback(() => {
    onMenuOpenChange?.(true)
  }, [onMenuOpenChange])

  const handleOnClose = useCallback(() => {
    onMenuOpenChange?.(false)
  }, [onMenuOpenChange])

  return (
    <MenuButton
      // ...
      onClose={handleOnClose}
      onOpen={handleOnOpen}
    />
  )
}
```

ninaandal: Thanks for the input! Fixed this 😄

packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
```diff
@@ -179,9 +181,10 @@ export function StudioNavbar() {
                   placement="bottom"
                   portal
                   scheme={scheme}
+                  disabled={isWorkspaceMenuOpen}
```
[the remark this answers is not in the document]
ninaandal: Considering that the WorkspaceMenuButton is also used in the NavDrawer and this doesn't have a tooltip, I think the current solution is an okay solve

RitaDias (approved):
:shipit:

packages/sanity/src/core/studio/components/navbar/resources/ResourcesButton.tsx
```diff
@@ -1,6 +1,7 @@
+/* eslint-disable react/jsx-no-bind */
```
hermanwikner, line 1: Nitpick: would be nice to get rid of this comment

```ts
  const handleOnOpen = useCallback(() => setMenuOpen(true), [])
  const handleOnClose = useCallback(() => setMenuOpen(false), [])
```

hermanwikner (approved):
LGTM! 👍

PR #4828, opened 2023-08-16
feat(core): avoid layout shift when editing title

binoy14:
### Description

This PR enables so that when you are editing the form the title does not do a layout shift. It only changes it's height when any of the form fields are not focused.

FIXES SDX-672

### What to review

~This was a little complicated to implement, the way this works if there are two elements created. One is the actual display and the other one is a `visibility:hidden` element, the reason to keep visibility is so it still occupies the height so we can calculate it. When you focus on the form it changes hidden title's style to text overflow and then uses it's `scrollWidth` and `clientWidth` to calculate the height. This height is then set to the title that is displayed and it has an overflow of hidden so it does not cause a layout shift when the content is greater then the height.~

~I would love to know if there is a different and/or better way to implement this.~

The above has changed to use `line-clamp` instead. When you land on the document, the height of the title is calculated and it does not change unless you visit another document or do a reload, this would avoid a layout shift if the title is using a field that is at the bottom of the page.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/desk/panes/document/documentPanel/documentViews/styles.ts
+++ b/packages/sanity/src/desk/panes/document/documentPanel/documentViews/styles.ts
@@ -1,6 +1,65 @@
-import {Heading} from '@sanity/ui'
-import styled from 'styled-components'
+import {Box, Theme} from '@sanity/ui'
+import styled, {css} from 'styled-components'
 
-export const Title = styled(Heading)`
-  word-break: break-word;
+export const Root = styled(Box)`
+  position: relative;
 `
+
+export const Title = styled.h6(({theme}: {theme: Theme}) => {
+  const {fontSize, lineHeight} = theme.sanity.fonts.heading.sizes[5]
+
+  return css`
+    font-size: ${fontSize}px;
+    margin: 0;
+    line-height: ${lineHeight}px;
+    ${({
+      $titleHeight,
+      $isEditingForm,
+      $muted,
+    }: {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/desk/panes/document/documentPanel/documentViews/styles.ts
```diff
@@ -1,6 +1,65 @@
...
+      $titleHeight?: number
+      $isEditingForm?: boolean
+      $muted: boolean
+    }) => {
+      return css`
+        ${$titleHeight ? `height: ${$titleHeight}px` : ''};
+        ${$isEditingForm
+          ? css`
+              overflow: hidden;
+              text-overflow: ellipsis;
```
mikolajdobrucki, line 29: `text-overflow: ellipsis` won't work in this case as it only applies to single-line text elements.
Also, it won't take effect if you allow the text to wrap, over when applying `overflow:hidden`.

I would suggest to try an implementation using CSS's 'line-clamp' or just Javascript.

mikolajdobrucki (changes requested):
Agree we don't want any "layout shit" ;) 

The current CSS implementation doesn't seem to work. Please find a suggestion in my comments below.

bjoerge (approved):
This turned out to be a very elegant solution! Great work!

mikolajdobrucki:
Works great! Fantastic work @binoy14! 🔥

PR #4832, opened 2023-08-17
chore: convert various scripts to typescript

bjoerge:
### Description

This converts various maintenance JS scripts to TS. Also removes some old cruft from the design-studio that was no longer in use.

### What to review
Make sure scripts works as before:
- `yarn check:versions`
- `yarn depcheck`
- `yarn normalize-pkgfields`
- `yarn release-notes`

### Notes for release

N/A Internal only


rexxars (approved):
Did a few changes due to the type-check failing, looks good to me.
Some of these scripts don't work _exceptionally_ well right now, but... better to have them in TS than in JS at least :)

bjoerge:
> Did a few changes due to the type-check failing, looks good to me.
Oh, sorry about that. And thanks!

> Some of these scripts don't work exceptionally well right now, but... better to have them in TS than in JS at least :)

Yeah I noticed, but figured I'd stay away from that yak for now :)

robinpyon (approved):
No issues here, thanks @bjoerge!

(Would be great to do an audit on some of these other dev studios in the not too distant future)

PR #4834, opened 2023-08-17
feat(core): convert navbar strings to use i18n primitives

RitaDias:
### Description

Convert parts of the navbar strings to use i18n primitives

Out of Scope:
- Date filtering parts of the search
- Debug overlays / components
- NavBar -> Search -> Definitions -> Operators

### What to review

That all items/strings within the vision are not broken and work as expected.

### Notes for release

Description
Convert vision tool strings to use i18n primitives

What to review
That all items/strings within the vision are not broken and work as expected.

Notes for release

N/A


packages/sanity/src/core/studio/components/navbar/search/components/Instructions.tsx
```diff
@@ -1,16 +1,18 @@
...
+import {useTranslation} from '../../../../../i18n'
 
 export function Instructions() {
+  const {t} = useTranslation()
+
   return (
     <Flex align="center" direction="column" gap={4} paddingX={4} paddingY={5}>
       <Inline space={3}>
-        <Text muted>Use</Text>
+        <Text muted>{t('navbar.search.instructions')}</Text>
```
RitaDias, line 12: 👋 @rexxars I had to change this slightly from what it was before (with rendering an icon as a react component), I tried looking around and I don't see a very neat way of doing this apart from altering the text slightly. However, I might be wrong and there's a neater way to keep the text as it was originally, do you see any way?

RitaDias: Discovered you can do something like this: https://stackoverflow.com/questions/61201001/using-i18next-to-replace-a-variable-with-a-reactnode-element

https://react.i18next.com/latest/trans-component

I'll replace it here with the right approach

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+
+  /** Description for error when no valid asset source is found, describes that only the default asset is supported */
+  'navbar.search.error.no-valid-asset-source-only-default-description':
+    'Currently, only the default asset source is supported.',
+
+  /** Description for error when no valid asset source is found, describes that you should cehck the the current studio config */
+  'navbar.search.error.no-valid-asset-source-check-config-description': `Please ensure it's enabled in your studio configuration file.`,
+
+  /** Title for error when a filter cannot be displayed */
+  'navbar.search.error.display-filter-title': 'An error occurred whilst displaying this filter.',
```
ninaandal, line 239: Not an english expert, but since we are using `while` in this `'navbar.search.error.something-went-wrong-title': 'Something went wrong while searching',` - maybe use it here as well?

packages/sanity/src/core/studio/components/navbar/search/components/Instructions.tsx
RitaDias: b0c1c40ae80646873637d08920388d12f1978b66

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+  'navbar.helpResources.latest-sanity-version': `Latest version is {{latestVersion}}`,
+
+  /** Label for "join our community" call to action */
+  'navbar.helpResources.action.join-our-community': `Join our community`,
+
+  /** Label for "help and support" call to action */
+  'navbar.helpResources.action.help-and-support': `Help and support`,
+
+  /** Label for "contact sales" call to action */
+  'navbar.helpResources.action.contact-sales': `Contact sales`,
```
robinpyon, line 321: **note:** these are titles for _fallback_ links in the event the help + resources endpoint isn't able to be fetched. Help and resources links and copy are _not_ localized in the admin studio side, so we'll probably want to address this at some point!

**suggestion:** considering updating comment and/or the names to reflect the fallback nature of these

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+  /** Label for action to choose a different workspace */
+  'navbar.workspace-menu.action.choose-another-workspace': 'Choose another workspace',
+
+  /** Label for action to add a workspace */
+  'navbar.workspace-menu.action.add-workspace': 'Add workspace',
+
+  /** --- New Document --- */
+
+  /** Placeholder for the "search" input within the new document menu */
+  'navbar.new-document.search': 'Search',
```
robinpyon, line 109: **question:** should we consider creating a common path / namespace for single words? Something like `Search` is bound to be used across various points in the studio. This applies to other words too, such as `Loading`, `Clear`,  `True`, `False`, etc

packages/sanity/src/core/studio/screens/schemaErrors/SchemaProblemGroups.tsx
```diff
@@ -57,7 +59,8 @@ export function SchemaProblemGroups(props: {problemGroups: SchemaValidationProbl
                 <Text size={1} weight="semibold">
                   {schemaType ? (
                     <>
-                      {capitalize(schemaType.type)} type "{schemaType.name}"
+                      {capitalize(schemaType.type)} {t('navbar.configuration.type-label')} "
```
robinpyon, line 62: **suggestion:** this could be a good candidate for the `<Trans>` component? Currently this renders `Document` `type` `"book"` which may need to be re-ordered in specific locales?

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/reference/ReferenceAutocomplete.tsx
```diff
@@ -152,7 +154,8 @@ export const ReferenceAutocomplete = forwardRef(function ReferenceAutocomplete(
                     <Box padding={4}>
                       <Flex align="center" height="fill" justify="center">
                         <StyledText align="center" muted>
-                          No results for <strong>“{searchState.terms.query}”</strong>
+                          {t('navbar.new-document.no-results')}{' '}
```
robinpyon, line 157: **suggestion:** consider wrapping this in a `<Trans>` component

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/AssetSourceError.tsx
```diff
@@ -12,13 +15,13 @@ export function AssetSourceError(props: ResponsivePaddingProps) {
...
+            {t('navbar.search.error.no-valid-asset-source-title')}
           </TextWithTone>
           <TextWithTone size={1} tone="caution">
-            Currently, only the default asset source is supported.
+            {t('navbar.search.error.no-valid-asset-source-only-default-description')}
           </TextWithTone>
           <TextWithTone size={1} tone="caution">
-            Please ensure it's enabled in your studio configuration file.
+            {t('navbar.search.error.no-valid-asset-source-check-config-description')}
           </TextWithTone>
```
robinpyon, line 25: **question:** could this also be better served by `<Trans>`? 

Which could potentially allow breaking this up into two tokens instead of three: `navbar.search.error.no-valid-asset-source-title` and `navbar.search.error.no-valid-asset-source-description`

robinpyon (commented):
Thanks @RitaDias – monster effort!

I'm not too familiar with `i18next` best practices, so please take suggestions on `<Trans>` usage with a grain of salt.

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+  'navbar.workspace-menu.community-title': 'Community',
+
+  /** Title for "docs" link within the workspace menu */
+  'navbar.workspace-menu.docs-title': 'Docs',
+
+  /** Title for "privacy" link within the workspace menu */
+  'navbar.workspace-menu.privacy-title': 'Privacy',
+
+  /** Title for "sanity io website" link within the workspace menu */
+  'navbar.workspace-menu.sanity-io-title': 'sanity.io',
```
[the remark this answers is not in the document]
RitaDias: Good catch, this was before I realised that they were being used in a const situation and I went with a different approach. Removing them 👍

packages/sanity/src/core/i18n/bundles/studio.ts
RitaDias: This is something I have brought up with @rexxars in the past but we weren't able to get to a specific conclusion before the hackaton and leave.

Because these are used cross packages (and that means across different languages and potentially different use cases) it's a hard one.

Here's why: In portuguese there might be some cases where the word that is most commonly translated to "Search" is used only in some cases but not in others where _technically_ in english it would all end up being the same word. There are some other languages which I'm sure have even more details of these kinds which I'm concerned we're being blind to. At the same time, adding more keys can add more noise (and more business expense to those that are looking to translate the studio).

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+  'navbar.search.document-types-label': 'Document types',
+
+  /** Label for when no documents are found for a filter are found */
+  'navbar.search.no-matches-found': `No matches for {{filter}}`,
+
+  /** Accessibility action label for clearing document type filters */
+  'navbar.search.action.clear-type-filters': 'Clear checked filters',
+
+  /** Clear action label for clearing document type filters */
+  'navbar.search.action.clear-type-label': 'Clear',
```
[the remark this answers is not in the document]
RitaDias: I thought about this for a while and I considered it, the reason why I didn't go with it was: There are places where the labels are used both in accessbility instances AND in the interface itself. Which means, to me, that it's probably preferrable that it's in the comments to those keys explicitly said when it's also used the an accessibility setting vs have a separate accessibility key for it.

I don't have a strong opinion on this, however, I was trying to avoid a situation of possible confusion because I've come across at least a handful of cases where this has happened (the situation I pointed above)

cc @rexxars

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+
+  /** Accessibility action label for deleting a search filter */
+  'navbar.search.action.delete-filter': 'Delete filter',
+
+  /** Action label for adding a search filter */
+  'navbar.search.action.add-filter': 'Add filter',
+
+  /** Accessibility Label for filter search input in the filtering section of search */
+  'navbar.search.filter-by-title-label': 'Filter by title',
```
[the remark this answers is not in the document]
RitaDias: Replied in the [comment above](https://github.com/sanity-io/sanity/pull/4834/files#r1307263197)

packages/sanity/src/core/i18n/bundles/studio.ts
RitaDias: I copied the text directly to what the messaging was there previously. I don't have a strong opinion on this so I'm all for changing one way or another :)

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -68,6 +68,310 @@ export const studioLocaleStrings = {
...
+  /** Title for error when a filter cannot be displayed */
+  'navbar.search.error.display-filter-title': 'An error occurred whilst displaying this filter.',
+
+  /** Description for error when a filter cannot be displayed, describes that you should check the schema */
+  'navbar.search.error.display-filter-description':
+    'This may indicate invalid options defined in your schema.',
+
+  /** Label for error when the useSearchState is not used within the SearchProvider */
+  'navbar.search.error.use-search-state-not-used-within-provider':
+    'useSearchState must be used within an SearchProvider',
```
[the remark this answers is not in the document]
RitaDias: Ah this is a good point that I din't consider 🤔 
I think we probably need to but probably not at the client level (because you wouldn't want typos on specific hooks for example). But @rexxars what are your thoughts on this?

packages/sanity/src/core/i18n/bundles/studio.ts
RitaDias: Updating the comment to reflect this 👍

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/reference/ReferenceAutocomplete.tsx
RitaDias: This was before I figured out how to use the `<Trans>` component, so yes! definitely 😄

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/AssetSourceError.tsx
RitaDias: Usually the Trans component is to be used when you have variables that require different react components to be used within. 
I am not sure what would be the benefit in this case 🤔

packages/sanity/src/core/studio/screens/schemaErrors/SchemaProblemGroups.tsx
RitaDias: Good point! Replacing it!

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/AssetSourceError.tsx
robinpyon: (To summarise what we spoke about separately)

This is more an issue rooted in the original source markup, which has led to having two keys which are essentially 'description-line-1' and 'description-line-2'. 

We agreed that we should try and avoid similar markup in future, ideally representing this as the folllowing (and allow text to line break naturally):
```
<TextWithTone size={1} tone="caution">
  {t('navbar.search.error.no-valid-asset-source-only-description')}
</TextWithTone>
```

robinpyon (approved):

PR #4838, opened 2023-08-18
fix(core): svg preview issue in safari

hermanwikner:
### Description

This pull request addresses an issue where media for list previews wouldn't always be visible in Safari when a custom SVG element is configured (as reported in [#4074](https://github.com/sanity-io/sanity/issues/4074)). Currently, the media is visible only if the SVG element has both width and height attributes. This pull request ensures that the SVG is visible even when these attributes are absent.

```ts
const myDocument = defineType({
  name: 'myDocument',
  type: 'document',
  title: 'My document',
  icon: () => <svg>...</svg>,
})
```

Untitled

### What to review

- Make sure that media is rendered as expected

### Notes for release

Fixes media visibility issue in Safari for list previews with custom SVG element

### DIFF
--- a/packages/sanity/src/core/components/previews/_common/Media.styled.ts
+++ b/packages/sanity/src/core/components/previews/_common/Media.styled.ts
@@ -38,6 +38,7 @@ export const MediaWrapper = styled.span<{
 
     & svg {
       display: block;
+      flex: 1;
       font-size: calc(21 / 16 * 1em);
     }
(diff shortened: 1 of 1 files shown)


robinpyon (approved):
Thanks @hermanwikner!

rexxars (approved):

binoy14 (approved):

PR #4856, opened 2023-08-22
test(playwright): fix hotkey activation + helper method

skogsmaskin:
### Description

This will fix the Playwright form tests to work on Linux (and CI). There were two issues:

* Platform name was returned incorrectly sometimes (confirmed on local Linux)
* To trigger hotkeys (at least on the PT-input), you have to do that on Linux with a combination of metaDown + hotkeyPress + metaUp (`playwright.page.press('Meta+b')` will not work on Linux). I suspect it's either the way the hotkey event listeners are implemented in the PTE, or it could be a Linux-specific behavior in how keyboard events are handled in the browser.

I have moved the 'focusPTE' method into two separate functions that can target either the PortableTextInput element as a whole (`getFocusedPortableTextInput`), and another that targets the Portable Text Editor within the input (`getFocusedPortableTextEditor`).

With the latter, we can create simpler tests code as most tests before had some lines in each test to do this in order to get to the actual text input.

I have also renamed the `typeInPTEWithDelay` function to `writePortableText` and removed the delay function. The delay can be written more in particular for tests that need to simulate human typing or similar, but it should not be needed in general (I know this helper function has changed a bit through time). Webkit tests are passing now as well, so I have enabled them.

I have disabled the Webkit tests only when platform is `linux` at this time as we don't fully support it.
They are passing on Apple Webkit, so I have enabled them there.

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
@@ -82,13 +82,13 @@ test.describe('Annotations', () => {
       testInfo,
     })
     await mount(<PortableTextInputStory />)
-    const $pteTextbox = await getFocusedPortableTextEditor('field-body')
+    const $pte = await getFocusedPortableTextEditor('field-body')
 
-    await insertPortableText('Now we should insert a link.', $pteTextbox)
+    await insertPortableText('Now we should insert a link.', $pte)
 
     // Assertion: Wait for the text to be rendered
     await expect(
-      $pteTextbox.locator('[data-slate-string="true"]', {
+      $pte.locator('[data-slate-string="true"]', {
         hasText: 'Now we should insert a link.',
       }),
     ).toBeVisible()
@@ -102,7 +102,7 @@ test.describe('Annotations', () => {
       .click()
 
     // Assertion: Wait for link to be re-rendered / PTE internal state to be done
-    await expect($pteTextbox.locator('[data-slate-node="text"] span[data-link]')).toBeVisible()
+    await expect($pte.locator('span[data-link]')).toBeVisible()
 
     // Now we check if the edit popover shows automatically
     await expect(page.getByLabel('Link').first()).toBeAttached({timeout: 10000})
@@ -114,10 +114,13 @@ test.describe('Annotations', () => {
     await expect(page.getByLabel('Link').first()).toBeFocused()
 
     // Type in the URL
-    await typeWithDelay('https://www.sanity.io')
+    await page.keyboard.type('https://www.sanity.io')
 
     // Close the popover
     await page.keyboard.press('Escape')
+
+    // Expect the editor to have focus after closing the popover
+    await expect($pte).toBeFocused()
   })
 })
 
@@ -129,7 +132,7 @@ test.describe('Blocks', () => {
     const {getFocusedPortableTextInput} = testHelpers({page, testInfo})
     await mount(<PortableTextInputStory />)
 
-    const $portableTextField = await getFocusedPortableTextInput('field-body')
+    const $portableTextInput = await getFocusedPortableTextInput('field-body')
 
     await page
       .getByRole('button')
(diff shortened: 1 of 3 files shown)


packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
```diff
@@ -91,5 +90,14 @@ export function testHelpers({
 
       await new Promise((resolve) => setTimeout(resolve, delay || DEFAULT_TYPE_DELAY))
     },
+    /**
+     * Will create a keyboard event of a given meta/hotkey combination
+     * @param key
+     */
+    useHotKey: async (hotkey: string, metaKey: string | undefined = undefined) => {
```
sjelfull, line 97: Nice catch! Can we name this something more descriptive to match the others? The `useX` looks like a React hook.

Like `pressHotkey()`? `toggleHotKey()`?

skogsmaskin: Agreed, I have changed it to `toggleHotKey`

sjelfull (approved):

packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
```diff
@@ -35,29 +35,47 @@ export function testHelpers({
   page: PlaywrightTestArgs['page']
   testInfo: TestInfo
 }) {
-  const platformName = getPlatformName(testInfo.project.name) || process.platform
```
robinpyon, line 38: **suggestion:** given that `getPlatformName` is no longer being used, we can remove it entirely from `testHelpers.tsx`

packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
```diff
@@ -36,62 +36,55 @@ test.describe('Activation', () => {
...
 test.describe('Decorators', () => {
   test('Render default styles with keyboard shortcuts', async ({mount, page}, testInfo) => {
-    const {getModifierKey, focusPTE, typeInPTEWithDelay} = testHelpers({
-      page,
-      testInfo,
-    })
+    const {getModifierKey, getFocusedPortableTextEditor, insertPortableText, toggleHotkey} =
+      testHelpers({
+        page,
+        testInfo,
```
robinpyon, line 42: **suggestion:** given that `testInfo` is no longer being used in `testHelpers`, consider removing it across the board (which should make these specs a tiny bit slimmer!)

robinpyon (approved):
Thanks @skogsmaskin! Agree with your take on delayed keypresses

All looks good to me! Only minor point is that we can probably remove the inclusion of `testInfo` across all tests (since that's no longer used in `testHelpers`)

packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
```diff
@@ -300,40 +290,44 @@ test.describe('Blocks', () => {
...
     for (const icon of ICONS) {
       await expect(
-        $pteField.getByRole('button').locator(`[data-sanity-icon="${icon}"]`),
+        $portableTextInput.getByRole('button').locator(`[data-sanity-icon="${icon}"]`),
       ).toBeVisible()
     }
 
     // Assertion: ???
-    await expect($pteField.locator('button#block-style-select')).not.toBeVisible()
+    await expect($portableTextInput.locator('button#block-style-select')).not.toBeVisible()
```
robinpyon, line 306: **suggestion:** consider breaking this out into a separate test? E.g.

```
  test('Should not display block style button when no block styles are present', async ({
    mount,
    page,
  }, testInfo) => {
    const {getFocusedPortableTextInput} = testHelpers({page, testInfo})
    await mount(<PortableTextInputStory />)
    const $portableTextInput = await getFocusedPortableTextInput('field-bodyStyles')

    // Assertion: Block style button should be hidden
    await expect($portableTextInput.locator('button#block-style-select')).not.toBeVisible()
  })
```

packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
skogsmaskin: Yeah, I was thinking about that, but didn't want to change too much :)

I think the simpler we can keep this the better.

packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
skogsmaskin: I have added the test @robinpyon 👍

packages/sanity/playwright-ct/tests/formBuilder/PortableTextInput.spec.tsx
skogsmaskin: Removed that dependency now.

packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
skogsmaskin: Removed it

PR #4858, opened 2023-08-22
fix(deps): support `styled-components` v6

stipsan:
### Description

We've tested `styled-components` v6 and added support for it in `@sanity/ui` and `@sanity/ui-workshop`. It seems like we're able to support v6 without losing support for v5. The breaking change we expect to [possibly affect userland](https://github.com/styled-components/styled-components/releases/tag/v6.0.0) is "dropped automatic prop filtering; use transient props ($ prefix) for stuff you don't want to be passed to child component / HTML" but this isn't cause for concern or a blocker, as adding `$` prefixes are backwards compatible with v5 and has been best practice in our own codebase for quite some time.

### What to review

Custom UI written with `styled-components` directly, or code that overrides/extends `@sanity/ui` with techniques such as `styled()` and the `as` prop should be checked to see it [doesn't pass down props to the underlying HTML](https://github.com/styled-components/styled-components/releases/tag/v6.0.0#:~:text=dropped%20automatic%20prop%20filtering%3B%20use%20transient%20props%20(%24%20prefix)%20for%20stuff%20you%20don%27t%20want%20to%20be%20passed%20to%20child%20component%20/%20HTML).
Components that simply use primitives from `@sanity/ui`, like `Card`, doesn't need to be thoroughly reviewed as we just did that.

### Notes for release

fix(deps): added support for `styled-components` v6. Upgrading to it is optional, we'll continue to support v5 for the lifecycle of Studio v3.


stipsan:
Converted back to draft, as it looks like there's some TypeScript work to be done.

bjoerge:
Superseded by #4917

PR #4859, opened 2023-08-22
fix(core): add check for cdr schematype and use PrimitiveField

ninaandal:
### Description
Cross dataset references had the wrong level/depth. There was no check for cdr schematype in the `fieldResolver.tsx`, which resulted in the field type for cdr to be `ObjectOrArrayField`. This PR adds a check for `isCrossDatasetReferenceSchemaType` and rendering a `PrimitiveField` in that case, removing the indent for a cdr.  

Screenshot 2023-08-22 at 15 13 37

### What to review
Make sure that the `PrimitiveField` works as intended for a cross dataset reference. 

### Notes for release
Fixes indent bug for cross dataset references


bjoerge (approved):
Great find! Technically a cross dataset reference is an object, since it only contains a single input it makes sense to use the primitive field component for it. Thanks @ninaandal!

hermanwikner (approved):

PR #4861, opened 2023-08-23
fix(core): use primitiveField for slug schema type

ninaandal:
### Description
There was no check for the `slug` schematype, resulting in it rendering an `ObjectOrArrayField`. Hence it was lacking a `<label>` for the `title`. This PR adds a check for `slug` schematype and renders a `PrimitiveField` instead. 

### What to review
The `slug` schematype - that it works as expected and that the title is wrapped in a `label`

### Notes for release
Change field type of slug from `ObjectOrArrayField` to `PrimitiveField`

### DIFF
--- a/packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
+++ b/packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
@@ -179,6 +179,10 @@ export function defaultResolveFieldComponent(
     return PrimitiveField as React.ComponentType<Omit<FieldProps, 'renderDefault'>>
   }
 
+  if (typeChain.some((t) => t.name === 'slug')) {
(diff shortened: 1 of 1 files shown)


RitaDias (commented):
👋 
The code looks fine! I would just want to try to understand what is the use case for this so I can test it myself? 👀

ninaandal:
> 👋 The code looks fine! I would just want to try to understand what is the use case for this so I can test it myself? 👀

Basically just testing that the `slug` fields work as expected. 
I'm assuming this issue was reported due to accessibility reasons, that the `title` of a `slug` field should be wrapped in a `label`. And for object/array field types, this was not the case. But for the `PrimitiveField`, which is used for e.g. a `string` field, the `title` is wrapped in a `label`. 

Screenshot 2023-08-24 at 11 02 22

RitaDias (approved):
💪

bjoerge (approved):
LGTM - thanks @ninaandal!

PR #4866, opened 2023-08-24
feat(core): convert array input actions strings to use i18n primitives 

RitaDias:
### Description

Convert array input strings to use i18n primitives

~~TO DO still in PR:~~

~~ArrayOfPrimitivesInput~~
- ~~ArrayOfPrimitivesInput (class, can’t use the hook, not sure where to send in the prop from)~~
- ~~getEmptyValue (tried sending the t in through params, but used in ArrayOfPrimitives which is a class)~~


packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/Grid/IncompatibleItemType.tsx
```diff
@@ -3,6 +3,8 @@ import React, {useCallback} from 'react'
 import {BulbOutlineIcon, UnknownIcon} from '@sanity/icons'
 import {resolveTypeName} from '@sanity/util/content'
 import styled from 'styled-components'
+import {Trans} from 'react-i18next'
```
bjoerge, line 6: Unused import
```suggestion
```

bjoerge (changes requested):
Looks good! Found an unused import that would be good to remove before merge.

PR #4870, opened 2023-08-25
feature(schema): improve validation of block members

skogsmaskin:
### Description

Block members can not be given a type that is not an object-like type, but nothing is stopping you today and can lead to strange errors.

For instance, a type like `{type: 'string'}` doesn't make sense inside block children, as they have to be wrapped in an object. We do however have core types that are object-like, like image, file, and reference that should be
supported.

We also have the core type 'span', which I have included in the core types in this PR (I think it was an oversight). This should however not be supported here (as it's injected automatically into the schema). Thus we also need
an allow list for those core object types that actually are supported.

I have identified the following supported core types for the block schema type `of`:
```
['file', 'image', 'object', 'reference']
```

Also, previously defined object types can be used here with a shorthand, like:  `{type: 'author'}` like before.

Everything else will lead to a validation error like this:

image

I have also added a rule that will check that the name of an object definition here is not the same as any previously defined types (like we do with arrays). This is only a warning though.

### What to review

Is the error message ok? Do we need a special help article?

Check that we have allowed all the supported object-like types and report problems on the unsupported ones.

I have written some tests for this in the `@sanity/schema` package, but it should be tested e2e also because
the code path is quite different.

We can write e2e tests for this as well with playwright-ct for the PortableText Input.

You can test it out in the studio by making schemas like this:

```ts
defineField({
  title: 'Blocks',
  name: 'blocks',
  type: 'array',
  of: [
    {
      type: 'block',
      of: [
        // Should produce error
        {type: 'string', name: 'foo'},
        // Should produce warning
        {type: 'object', name: 'validObject', fields: [{type: 'string', name: 'foo'}]},
        // Should be allowed
        {type: 'image', name: 'image'},
        // Should be allowed
        {type: 'reference', name: 'reference', to: {type: 'author'}},
        // Should produce warning
        {type: 'object', name: 'reference', fields: [{type: 'string', name: 'foo'}]},
        // Should produce warning
        {type: 'object', name: 'image', fields: [{type: 'string', name: 'foo'}]},
        // Should produce warning
        {type: 'object', name: 'file', fields: [{type: 'string', name: 'foo'}]},
        // Should produce warning
        {type: 'object', name: 'span', fields: [{type: 'string', name: 'foo'}]},
        // Should not be allowed
        {type: 'span', name: 'something', fields: [{type: 'string', name: 'foo'}]},
        // Should be allowed
        {type: 'reference', name: 'reference', to: {type: 'author'}},
      ],
    },
  ],
}),
```
This should now fail because the type `string` is a core type that is not object-like.

Even though the `span` type is a core object-like type, it is not supported because it
isn't in the allow-list of core object types that are supported.

### Notes for release
* Improve validation of block schema members

### DIFF
--- a/packages/@sanity/schema/test/validation/validation.test.ts
+++ b/packages/@sanity/schema/test/validation/validation.test.ts
@@ -1,12 +1,5 @@
 import {flatten} from 'lodash'
-import {traverseSchema} from '../../src/core/traverseSchema'
 import {validateSchema} from '../../src/sanity/validateSchema'
-import array from '../../src/sanity/validation/types/array'
-import block from '../../src/sanity/validation/types/block'
-import file from '../../src/sanity/validation/types/file'
-import image from '../../src/sanity/validation/types/image'
-import object from '../../src/sanity/validation/types/object'
-import reference from '../../src/sanity/validation/types/reference'
 
 describe('Validation test', () => {
   test('assigns/populates `_problems` property', () => {
@@ -27,21 +20,7 @@ describe('Validation test', () => {
       },
     ]
 
-    const coreTypes = [
-      {name: 'array', type: 'type'},
-      {name: 'string', type: 'type'},
-      {name: 'object', type: 'type'},
-    ]
-
-    const visitors = {
-      array: {visit: array},
-      object: {visit: object},
-    }
-
-    const validation = traverseSchema(schemaDef, coreTypes, (typeDef, visitorContext) => {
-      const visitor = visitors[typeDef.type]
-      return visitor ? visitor.visit(typeDef, visitorContext) : typeDef
-    })
+    const validation = validateSchema(schemaDef)
 
     const myArray = validation.get('myArray')
     expect(myArray._problems.length).toBeGreaterThan(0)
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
This is great ❤️

packages/@sanity/schema/test/validation/validation.test.ts
```diff
@@ -89,4 +94,116 @@ describe('Validation test', () => {
...
+                  // Should be allowed
+                  {type: 'reference', name: 'reference', to: {type: 'author'}},
+                ],
+              },
+            ],
+          },
+        ],
+      },
+    ]
+    const coreTypes = [
```
pedrobonamin, line 153: Could this be imported from [/src/sanity/coreTypes.ts](https://github.com/sanity-io/sanity/pull/4870/files#diff-985c6b5eb4dfacfe03546f5ebcd798bebb2bd29a562995de498c9a691b76b7e0) ?

skogsmaskin: Yeah, this file was ready for a refactor that's for sure. Thanks for noticing :green_heart: 

I have added a suggestion in a new [commit](https://github.com/sanity-io/sanity/pull/4870/commits/0151177f3376af5549726f5e57c92d5cead42d59) that for sure removes a lot of this cruft, and just use the public `validateSchema` function instead.

I can't see any downsides (but there are upsides) by doing this instead?

@bjoerge - can you have another look?

pedrobonamin: Thanks for this update and refactor. I can't think of any downside on using the public function, it's better to do it that way IMHO, if you add any breaking change to that public function the tests will catch it.

pedrobonamin (approved):

bjoerge (approved):

PR #4871, opened 2023-08-25
fix(form/inputs): Track focusPath for a span's .text prop + harden test condition

skogsmaskin:
### Description

* Track editor focus for a focusPath that points to a span's text property.
* Harden test for span text by testing that the _type property is actually a 'span' too.
  I found and issue where inline blocks having a .text property also triggered editor focus here, which they shouldn't.
* Only focus the editor if focusPath is actually pointing to a editor focusable node.
* Improve test for `hasFocus` variable used in the PT-input (also test that the opened member is a editor node).
* Added component test for this

### What to review
* That the Portable Text Input will correctly track focus paths that terminate with a span's `text` property.
* That the Portable Text Input will not take focus if you have an inline object containing a root `.text` property.

### Notes for release
* Fixed a bug where the Portable Text Input would capture focus trying to edit an inline object's field named `.text`.

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
@@ -58,51 +58,61 @@ test.describe('Portable Text Input', () => {
   })
   test.describe('Should track focusPath', () => {
     test(`for span .text`, async ({mount, page}) => {
-      const initialPath = ['body', {_key: 'c'}, 'children', {_key: 'd'}, 'text']
       const component = await mount(
-        <FocusTrackingStory document={document} focusPath={initialPath} />,
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'c'}, 'children', {_key: 'd'}, 'text']}
+        />,
       )
-      await waitForFocusPath(page, initialPath)
-      const $portableTextInput = component.getByTestId('field-body')
-      const $pteTextbox = $portableTextInput.getByRole('textbox')
-      await expect($pteTextbox).toBeFocused()
-      expect(await getFocusedNodeText(page)).toEqual('Bar')
-      await setFocusPathFromOutside(page, ['body', {_key: 'e'}, 'children', {_key: 'f'}, 'text'])
-      expect(await getFocusedNodeText(page)).toEqual('Baz')
+      await waitForFocusedNodeText(page, 'Bar')
+      await component.update(
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'e'}, 'children', {_key: 'f'}, 'text']}
+        />,
+      )
+      await waitForFocusedNodeText(page, 'Baz')
     })
     test(`for span child root`, async ({mount, page}) => {
-      const initialPath = ['body', {_key: 'c'}, 'children', {_key: 'd'}]
       const component = await mount(
-        <FocusTrackingStory document={document} focusPath={initialPath} />,
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'c'}, 'children', {_key: 'd'}]}
+        />,
       )
-      await waitForFocusPath(page, initialPath)
-      const $portableTextInput = component.getByTestId('field-body')
-      const $pteTextbox = $portableTextInput.getByRole('textbox')
-      await expect($pteTextbox).toBeFocused()
-      expect(await getFocusedNodeText(page)).toEqual('Bar')
-      await setFocusPathFromOutside(page, ['body', {_key: 'e'}, 'children', {_key: 'f'}])
-      expect(await getFocusedNodeText(page)).toEqual('Baz')
+      await waitForFocusedNodeText(page, 'Bar')
+      await component.update(
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'e'}, 'children', {_key: 'f'}]}
+        />,
+      )
+      await waitForFocusedNodeText(page, 'Baz')
     })
     test(`for inline objects with .text prop`, async ({mount, page}) => {
-      const initialPath = ['body', {_key: 'g'}, 'children', {_key: 'i'}, 'text']
       const component = await mount(
-        <FocusTrackingStory document={document} focusPath={initialPath} />,
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'g'}, 'children', {_key: 'i'}, 'text']}
+        />,
       )
-      await waitForFocusPath(page, initialPath)
       const $portableTextInput = component.getByTestId('field-body')
       const $pteTextbox = $portableTextInput.getByRole('textbox')
       await expect($pteTextbox).not.toBeFocused()
       const inlineObjectTextInput = page.getByTestId('inlineTextInputField').getByRole('textbox')
       await expect(inlineObjectTextInput).toBeFocused()
-      await setFocusPathFromOutside(page, ['body', {_key: 'e'}, 'children', {_key: 'f'}])
+      await component.update(
+        <FocusTrackingStory
+          document={document}
+          focusPath={['body', {_key: 'e'}, 'children', {_key: 'f'}]}
+        />,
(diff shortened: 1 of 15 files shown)


skogsmaskin:
@cody, are you able to test the visual editing stuff with this?

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -218,7 +218,10 @@ export function PortableTextInput(props: PortableTextInputProps) {
     return items
   }, [members, props])
 
-  const hasFocus = focused || isEditorFocusablePath(focusPath)
+  const hasFocus =
```
bjoerge, line 221: does this actually mean "has focus within"? If so, would it be better if the variable was named to reflect that?

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -342,15 +345,11 @@ export function PortableTextInput(props: PortableTextInputProps) {
           </PortableTextMemberItemsProvider>
         </PortableTextMarkersProvider>
       )}
+      <pre>{JSON.stringify(focusPath)}</pre>
```
bjoerge: this should probably be removed before merge :)

bjoerge (changes requested):

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
skogsmaskin: Hehe, yup I saw it and rebased it away right after I opened this PR up 😅

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
skogsmaskin: Yeah, that's a clearer name for what it is. 👍 I've pushed a new commit with this change.

bjoerge (approved):

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTrackingStory.tsx
```diff
@@ -0,0 +1,83 @@
...
+}) {
+  const [focusPath, setFocusPath] = useState<Path>(focusPathFromProps || [])
+  useEffect(() => {
+    ;(window as any).__setFocusPath = (path: Path) => {
+      setFocusPath(path)
+    }
+    return () => {
+      ;(window as any).__setFocusPath = undefined
+    }
+  }, [])
```
robinpyon, line 73: **question:** is this adding complexity where we may not need it? Here we're adding an intermediate state layer, and the associated specs need to be aware of  `window.__setFocusPath` too.

Since the associated specs are checking the value of `data-focus-path` on the _story_ itself, it feels like we're leaning towards _testing the test_ rather than the underlying implementation.

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+      window.localStorage.debug = 'sanity-pte:*'
+    })
+  })
+  test.describe('Should track focusPath', () => {
+    test(`for span .text`, async ({mount, page}) => {
+      const initialPath = ['body', {_key: 'c'}, 'children', {_key: 'd'}, 'text']
+      const component = await mount(
+        <FocusTrackingStory document={document} focusPath={initialPath} />,
+      )
+      await waitForFocusPath(page, initialPath)
```
robinpyon, line 65: **question:** is it a smell that `waitForFocusPath` checks for a data attribute on the corresponding _story wrapper_ rather than the underlying PTE implementation?

More specifically, shouldn't `window.getSelection()` return the correct value immediately after mounting if `focusPath` is provided?

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
```diff
@@ -0,0 +1,147 @@
...
+      const $portableTextInput = component.getByTestId('field-body')
+      const $pteTextbox = $portableTextInput.getByRole('textbox')
+      await expect($pteTextbox).toBeFocused()
+      expect(await getFocusedNodeText(page)).toEqual('Bar')
+      await setFocusPathFromOutside(page, ['body', {_key: 'e'}, 'children', {_key: 'f'}, 'text'])
```
robinpyon, line 70: **question:** instead of using `setFocusPathFromOutside` – would it instead be preferable to re-render the component with new props? e.g.

```
const newPath = ['body', {_key: 'e'}, 'children', {_key: 'f'}, 'text']
await component.update(<FocusTrackingStory focusPath={newPath} />)
expect(await getFocusedNodeText(page)).toEqual('Baz')
```

robinpyon (commented):
Thanks @skogsmaskin! Left some comments surrounding potentially simplifying these tests, would love to hear your thoughts

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTrackingStory.tsx
skogsmaskin: @rob, it's not actually testing the value from there.  Testing the actual focusPath value is out of the scope of these tests, which is rather testing that it's getting the precise cursor location as expected.

The code added here is for the tests to know that the `focusPath` has been propagated to form, in order not to test this condition "too early".

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
skogsmaskin: The editor will sync the value and eventually put the focusPath in effect in the editor through hooks in an asynchronous way. That's what makes this a bit tricky to test, and why I wrote the "helper" code to confirm that the update to the focusPath has been propagated to the browser before we continue the test.

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
skogsmaskin: Yes, this is the problem, because they are not synchronous. They are dependent on the editor showing the correct value, before it applies the focusPath (or we will not be able to create the correct window.selection).

skogsmaskin:
> Thanks @skogsmaskin! Left some comments surrounding potentially simplifying these tests, would love to hear your thoughts

I appreciate your feedback here @robinpyon, and I understand where you're coming from. But it's harder to test this stuff than one should think because of the asynchronous nature of the editor, and that the value must be fully rendered in the editor at the time we transform the focusPath into an editor selection.

This is why the Editable component for the PTE has a  prop "selection" that can be used to set an initial selection for when the editor value is propagated and the editor is ready. At mount time we set the PT-input focusPath to this, and it will turn into a cursor selection when the value is fully propagated.  Then the PT-input has a hook that will propagate changes to the PT-input's focusPath when set from something else than the editor itself (the PT-input updates the focusPath based on the editor selection and is most of the time the actual source of the current focusPath).

I do agree with you that the test helper stuff for the focusPath is confusing, and it would be great if we could find a better way to test this. I didn't want to spend too much time on this either, but I think it's worth having another look.

skogsmaskin:
@rob, I had another look at this today and I was able to simplify the tests using `component.update` as you suggested!

It turned out there was actually a bug in our code with Webkit in regard to the synchronicity of hooks.
We have to use `useLayoutEffect` for the `useTrackFocusPath` hook as it seems to be colliding with other things happening in hooks inside slate's Editable element.  I discovered that in Webkit it sometimes would not set the correct selection, but default to the top document selection if you tried to call the same focusPath (but different objects) fast after each other (like clicking on a validation message several times in a row). This was the root of the flaky webkit tests as well, which I tried to mitigate with the test code discussed above.

Thank you for looking so thoroughly into the PR and your suggestions.

PR #4873, opened 2023-08-27
fix(portable-text-editor): fix issue with sequencing mark toggles

skogsmaskin:
### Description

Found a bug where if you toggle different decorators in sequence while
typing along, it can sometimes reset your applied decorator when reconciling with
the server value.

This was happening because we didn't specify the `_type` when splitting text
nodes with a new mark, and the new node would miss the span type.

The normalizer function would then assign the 'span' type to it however,
but it is not setting the marks (it's not the responsibility of that normalization).

This will make sure we don't need to normalize in this situation in the first place,
and that the type is correctly set in the split operation when toggling the mark.

### What to review
That you can perform the following sequence and keep the marks intact:

* Empty editor
* Press hotkey `cmd+b`
* Write a word
* Press hotkey `cmd+b`
* Press hotkey `cmd+i`
* Write a word
* Confirm that it is still italic when the new server value is returned

### Notes for release
* Fix a problem with toggling text decorators in succession while typing in the Portable Text Input.


sjelfull (approved):

bjoerge (approved):
LGTM!

PR #4874, opened 2023-08-27
test(portable-text-editor): improve test env. and upgrade packages

skogsmaskin:
### Description

This will improve the test framework for the Portable Text Editor.

Most importantly, I have identified and fixed a couple of bugs and problems in the real-time collaborative testing environment, which can definitely explain the flakiness we have seen in these tests.

These problems were mostly about how the mutations, values, and revisions were handled by the test web server and propagated to the clients. I have gone through the whole chain of events and made sure they are correctly performed.
Some of the tests also had missing `await` statements, so that also contributed to flakiness.

The collaborative tests now run both stable (and faster) 🎉  

Other improvements:

* Decreased PTE throttle MS for test environment for faster tests.
* Updated the `jest-dev-server` package and improved startup scripts.
* Alias `node-ipc` to use a safe fork of the compromised package, instead of the older original version which can be tempting to try to upgrade.  See more: https://github.com/node-ipc/node-ipc
* Got rid of `js-dom` dependency which no longer is needed.
* Upgraded to latest `@portabletext/toolkit` used in the tests.
* Aligned versions of `react-dom` and `jest` with the same versions as elsewhere in the mono repo.

### What to review
That the PTE tests runs stable every time in every environment.

### Notes for release
N/A


packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
```diff
@@ -24,29 +18,71 @@ const WEB_SERVER_ROOT_URL = 'http://localhost:3000'
...
 
 export default class CollaborationEnvironment extends NodeEnvironment {
   private _browserA?: Browser
   private _browserB?: Browser
   private _pageA?: Page
   private _pageB?: Page
+  private _contextA?: BrowserContext
+  private _contextB?: BrowserContext
+
+  // public async setup(): Promise<void> {
```
bjoerge, line 37: can these comments be removed? or add a comment about why they need to stay?

bjoerge (approved):
> What to review
> That the PTE tests runs stable every time in every environment.

But that will take for*ever* to verify!

Code changes looks good - solid work!

packages/@sanity/portable-text-editor/test/setup/collaborative.jest.env.ts
skogsmaskin: I want to keep them there for future reference. Added a comment.

PR #4876, opened 2023-08-29
test(playwright-ct): add to tests

skogsmaskin:
### Description

Refactor PT-input overlay activation to reuse code, and get around an issue on Webkit/Linux where tests would randomly
fail.

* Test decorator default icons and customizing decorators with your own icon.
* Enable previously skipped decorator tests for Webkit on Linux (issues resolved).
* Simplify an assert statement
* Simplify some locator queries
* Await text when `insertPortableText`
* Test toggling of decorator buttons (should be rendered active)

### DIFF
--- a/packages/sanity/playwright-ct.config.ts
+++ b/packages/sanity/playwright-ct.config.ts
@@ -35,7 +35,7 @@ export default defineConfig({
   ],
 
   /* Maximum time one test can run for. */
-  timeout: 120 * 1000,
+  timeout: 10 * 1000,
   expect: {
     // Maximum time expect() should wait for the condition to be met.
     timeout: 5 * 1000,
--- a/packages/sanity/playwright-ct.config.ts
+++ b/packages/sanity/playwright-ct.config.ts
@@ -38,7 +38,7 @@ export default defineConfig({
   timeout: 120 * 1000,
   expect: {
     // Maximum time expect() should wait for the condition to be met.
-    timeout: 20 * 1000,
+    timeout: 5 * 1000,
   },
 
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
@@ -37,7 +37,7 @@ test.describe('Portable Text Input', () => {
       // Write some text
       await insertPortableText('Hello there', $pte)
       // Assertion: placeholder was removed
-      expect(await $placeholder.count()).toEqual(0)
+      expect($placeholder).toHaveCount(0)
     })
   })
 })
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
@@ -33,11 +33,12 @@ test.describe('Portable Text Input', () => {
       const $pte = await getFocusedPortableTextEditor('field-body')
       const $placeholder = $pte.getByTestId('pt-input-placeholder')
       // Assertion: placeholder is there
+      await expect($placeholder).toBeVisible()
       await expect($placeholder).toHaveText('Empty')
       // Write some text
       await insertPortableText('Hello there', $pte)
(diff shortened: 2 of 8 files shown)


packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
```diff
@@ -37,7 +37,7 @@ test.describe('Portable Text Input', () => {
       // Write some text
       await insertPortableText('Hello there', $pte)
       // Assertion: placeholder was removed
-      expect(await $placeholder.count()).toEqual(0)
+      expect($placeholder).toHaveCount(0)
```
sjelfull, line 40: Looks good, just have to wait for this async assertion

```suggestion
      await expect($placeholder).toHaveCount(0)
```

sjelfull (changes requested):
Mostly looks good. Added a small request that can cause flakyness.

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Input.spec.tsx
skogsmaskin: Nice catch!

packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
```diff
@@ -20,10 +28,11 @@ export function testHelpers({page}: {page: PlaywrightTestArgs['page']}) {
      * @returns The Portable Text Input element
      */
     getFocusedPortableTextInput: async (testId: string) => {
+      // Wait for field to get ready (without this tests fails randomly on Webkit)
+      await page.locator(`[data-testid='${testId}']`).waitFor()
```
sjelfull, line 32: Same as above comment for `.waitFor()`.

packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
```diff
@@ -38,10 +47,11 @@ export function testHelpers({page}: {page: PlaywrightTestArgs['page']}) {
      * @returns The PT-editor's contentEditable element
      */
     getFocusedPortableTextEditor: async (testId: string) => {
+      // Wait for field to get ready (without this tests fails randomly on Webkit)
+      await page.locator(`[data-testid='${testId}']`).waitFor()
```
sjelfull, line 51: Same as above comment for `.waitFor()`.

sjelfull (changes requested):
Left some questions, mainly around what locator methods is being used.

packages/sanity/playwright-ct.config.ts
```diff
@@ -35,10 +35,10 @@ export default defineConfig({
   ],
 
   /* Maximum time one test can run for. */
-  timeout: 120 * 1000,
+  timeout: 10 * 1000,
   expect: {
     // Maximum time expect() should wait for the condition to be met.
-    timeout: 20 * 1000,
+    timeout: 5 * 1000,
   },
```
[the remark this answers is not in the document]
skogsmaskin: I think they are set too high and don't play well with TDD. You have to wait 20 seconds to get the test result back which is super annoying.

packages/sanity/playwright-ct/tests/utils/testHelpers.tsx
```diff
@@ -12,6 +12,14 @@ export const TYPE_DELAY_HIGH = 150
 export type MountResult = Awaited<ReturnType<ComponentFixtures['mount']>>
 
 export function testHelpers({page}: {page: PlaywrightTestArgs['page']}) {
+  const activatePTInputOverlay = async ($pteField: Locator) => {
+    const $overlay = $pteField.getByTestId('activate-overlay')
+    if ((await $overlay.count()) > 0) {
+      await $overlay.focus()
+      await page.keyboard.press('Space')
+    }
+    await $pteField.locator(`[data-testid='pt-editor__toolbar-card']`).waitFor()
```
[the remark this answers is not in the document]
skogsmaskin: Mind that this is not in a test but in a helper function. This wouldn't know about `expect`.

packages/sanity/playwright-ct.config.ts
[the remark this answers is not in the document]
sjelfull: Alright, that is a good point. Let's keep an eye on the flake though.

sjelfull (approved):

PR #4880, opened 2023-08-30
fix(core): add maxLength to breadcrumb and change symbol

ninaandal:
### Description
The breadcrumps for the review changes could get very long if there were several objects. Added a `maxLength` and changed the symbol from `/` to `>` as a separator to make it consistent with the new breadcrumb styling. 

Before: 
Screenshot 2023-08-22 at 13 59 48

After: 

Screenshot 2023-08-30 at 11 41 55

### What to review
Review changes -> the headings of which document has been editedrr

### Notes for release
Fixed a visibility issue with breadcrumbing which document was edited in Review Changes


kaylasanity (approved):

packages/sanity/src/core/field/diff/components/ChangeBreadcrumb.tsx
```diff
@@ -1,17 +1,24 @@
...
+  const maxLength = useMemo(() => {
+    if (change && change?.titlePath.length > 3) {
+      return 4
+    }
+    return undefined
+  }, [change])
 
   return (
     <Breadcrumbs
+      maxLength={maxLength}
```
binoy14, line 18: 💭 instead of a function here, can't we always define a `maxLength=4` ? Since it will only trigger that if the items are greater than the maxLength. 

Eg: https://www.sanity.io/ui/arcade?mode=jsx&jsx=eJyzccqvUMhNLErPzLOtNqm141KAAhunotTElOSi0tykYrigAlBphU9qXnpJBkg1knhxakFiUWJJfpFtNZIo0JiQ1IoShdzSktQUheLMqlTbakMkSyCgWslOqRZVlz5IG7I6hDyyqE1KZpmdR2pOTr6CoY0%2BiINd0gifpDE%2BSRN8kqb4JM3wSZqjSdroIwU3TBgomF9hxwXjjciYwhMwAHr1qQ0%3D

packages/sanity/src/core/field/diff/components/ChangeBreadcrumb.tsx
```diff
@@ -1,17 +1,24 @@
...
       separator={
         <Text muted size={1}>
-          /
+          {'>'}
```
robinpyon, line 21: **suggestion:** Use `<ChevronRightIcon />` (from `@sanity/icons`) instead of a greater than symbol

robinpyon (commented):
Thanks @ninaandal!

Minor point, but we should make sure to use correct icons rather than the text equivalent.

Whilst not a regression introduced by this PR, breadcrumbs in this space can be unnecessarily truncated. An example can be found [here](https://test-studio.sanity.build/test/content/input-debug;recursiveObjectTest;f27078e9-aec1-4cee-a654-05535c4b20dc%2Csince%3D%2540lastPublished%2Cinspect%3Dsanity%252Fdesk%252Fhistory).

image

I believe this is happening because `<ChangeTitleSegment>` will always set `maxWidth = 100` [when its supplied string is over 30 chars](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/field/diff/components/ChangeTitleSegment.tsx#L26-L34). 

We should consider only truncating this text when we know we're dealing with a breadcrumb containing _multiple_ segments.

packages/sanity/src/core/field/diff/components/ChangeBreadcrumb.tsx
ninaandal: Why do we even have that `maxWidth` set in any condition? Looks better without it? 

Screenshot 2023-09-04 at 12 45 40

Screenshot 2023-09-04 at 12 45 33

robinpyon: `maxWidth` probably made more sense before this `<Breadcrumbs>` component used `maxLength`. I agree in that it's probably no longer necessary.

robinpyon (approved):
Thanks @ninaandal!

binoy14 (approved):
Great job! thank you!

PR #4881, opened 2023-08-30
fix(portable-text-editor): emit new selection on mark toggle

skogsmaskin:
### Description

There is a regression in the latest release where the PT-input's toolbar button for a decorator in two cases
doesn't reflect the state of the editor. This is when you are toggling a decorator on a block containing only one span node, or if you have the cursor at a position without any text selected, and then press the decorator hotkeys.

This happens because the editor doesn't emit a new selection object in these situations, because it's the same selection.

In the Input the toolbar buttons are memo'ed on this selection object (we try to reuse the object identity of selections as much as possible for performance reasons).

However, in this case, we should force a new selection as it's very convenient for implementations to have the selection as a dependency when determining when to update state for toolbar buttons and similar.

With these changes, we will always emit a new selection object when toggling marks, even though the selection might still end up being the same value.

I have added a test for this in the Portable Text editor package.

I have also added a Playwright component test for the Portable Text Input to test this, which will
be followed up in a separate PR.

### What to review

That the above regression is fixed. With the PT-input:
* Have a empty PT-input. Focus it, but don't write anything. Toggle a decorator (for example ctrl+b). This should be reflected in the toolbar button state.
* Write a couple of words in an empty PT-input (important that it is one single span in the value). Select all the text. Toggle marks (ctrl+b). The toolbar button for that mark should reflect the state.

### Notes for release
* Fixed a regression in the Portable Text Input toolbar where it didn't reflect the active decorators state in some situations.


bjoerge (approved):
LGTM

PR #4886, opened 2023-08-31
feat(desk): implement comments (beta)

hermanwikner:
### Description

This pull request introduces a new comments feature, providing the ability to add comments to specific fields within documents. The comments specific logic is implemented in `desk`, and some updates have been made to `core` to enable comments to be configured from the `desk` plugin. Please note that this pull request aims to introduce comments as a beta feature and there is thus room for improvement in the implementation that will be adressed before GA.

The changes made to `core` and `desk` respectively are divided into two sections below.

### What to review: core (`DX` + `EX`)
- **Field Comments:**
  - Introduces a new `__internal_comments` prop for fields. This prop allows the configuration of fields with a comment action that is rendered along with Field Actions. The `desk` plugin uses this prop, as demonstrated in the following example:

    ```ts
    import {FieldProps} from 'sanity'
    
    export function CommentField(props: FieldProps) {
      return props.renderDefault({
        ...props,
        __internal_comments: {
          button: <CommentButton />, // The button rendered along with field actions
          hasComments: false, // Needed to know if we should display the comment button when the field has comments
          isAddingComment: false, // Needed to keep the field actions visible when authoring a comment
        },
      })
    }
    ```

- **Document Comments API:**
  - Adds a new (unstable) API to enable/disable the comments feature for documents. In the initial beta release, the comments feature is opt-in, requiring explicit configuration. The API provides both a simple boolean option to enable/disable comments for all documents and a more granular configuration by using the `documentType` and/or `documentId` values provided in the context parameter.

    **Boolean example – Enable comments for all documents:**
    ```ts
    
    export default defineConfig({
      // ... 
    
      document: {
        unstable_comments: {
          enabled: true,
        },
      },
    })
    ```

    **Context example – Enable comments only for 'article' documents:**
    ```ts

    export default defineConfig({
      // ...
    
      document: {
        unstable_comments: {
          enabled: (ctx) => {
            return ctx.documentType === 'article'
          },
        },
      },
    })
    ```

- **New exports from the sanity package:**
  - The `useDidUpdate` hook.
  - The `resolveConditionalProperty` function: We use this function to determine if a comment should be visible or not, ensuring that only comments related to visible fields are displayed.
  - The `useFeatureEnabled` hook: This hook is essential to check if the comments feature is enabled for the current plan.

##

#### Summary of critical elements in the implementation:

- **`CommentsProvider`**: This provider is used in `DocumentPaneProvider` and is responsible for handling all CRUD operations.

- **`CommentsSetupProvider`**: This provider wraps the entire studio through `studio.components.layout`. It is responsible for the setup of comments (using the setup endpoint) and configuring a client for the addon dataset. This client is accessed with `useCommentsSetup` in `CommentsProvider` to enable CRUD operations on the addon dataset.

- **`CommentsOnboardingProvider`**: This provider is relatively straightforward and is responsible for storing a value in local storage if the user dismisses the comments onboarding. It is implemented as a provider because, in the future, there may be multiple instances where onboarding needs to be displayed.

- **`useCommentOperations`**: This hook manages the logic for creating, editing, updating, or deleting comments.

- **`useCommentsStore`**: This hook initiates the initial fetch of all comments, subscribes to a real-time listener, and manages the state using a reducer. The `dispatch` function in this hook is exported to facilitate immediate and optimistic state updates with callbacks provided by the `useCommentOperations`. That is, whenever e.g. a `create` operation is executed, we optimistically update the state in the reducer to make the experience feel snappy and fast.

- **`useMentionOptions`**: This hook returns a list of users who can be mentioned in the current document. The hook performs the following tasks:
    - Retrieves all grant filters using `'*[_type == "system.group”]'`.
    - Evaluates each user's grant filter against the document's value to determine if the user has `read` access. If this is the case, it becomes possible to mention the user.

- **`useCommentsEnabled`**: This hook checks whether the comments feature flag (`studioComments`) are enabled for the project and resolves `document.unstable_comments.enabled` to determine if the comment feature should be enabled. The default value in `document.unstable_comments.enabled` should be `false`, and the user needs to explicitly enable comments in the config to try it out. This hook is used in three different locations:

    1. In `CommentField` to decide whether to render a field with a comment button or defer to the default field rendering.

    2. In the `useMenuItem` hook within the document inspector definition to determine whether the inspector button should be set as `hidden: true`.

    3. In `CommentsProvider` to determine whether to set up listeners, and so forth, for comments. If not, the provider is still rendered with nullified values.
    
- **`buildCommentBreadcrumbs`**: This function builds the breadcrumbs for a comment and evaluates the comment against the document value to determine whether it should be displayed. This evaluation is performed because we only want to display comments that have a corresponding field. If there is no associated field (e.g., if the schema definition has changed or an array item has been removed), the comment should not be displayed. This also applies if a conditional `hidden` callback returns `false`.

- **`buildCommentThreadItems`**: This function maps over all the comments an groups them in a format that is easier to work with in the UI. 

### Notes for release

n/a


pedrobonamin (commented):
Great job! This is going to be a fantastic feature. 
Left some questions and a suggestion, happy to discuss.

Looking forward to this !

packages/sanity/src/desk/comments/src/context/comments/CommentsProvider.tsx
```diff
@@ -0,0 +1,374 @@
...
+  edit: noopOperation,
+  getComment: () => undefined,
+  getCommentPath: () => null,
+  mentionOptions: EMPTY_MENTION_OPTIONS,
+  remove: noopOperation,
+  selectedPath: null,
+  setSelectedPath: noop,
+  setStatus: noop,
+  status: 'open',
+  update: noopOperation,
```
stipsan, line 78: I would love it if these were all separate context providers, so we can avoid building up the same kind of perf problems we've seen in the `DocumentPaneContext`.
It's always cheaper in React to have many many many context providers in a tree, where you only pay for the providers that are only subscribed, and the `useContext` only cause re-renders in components that need them to re-render.
Compared to super contexts where any change to any of the properties inside the `CommentsContext.Provider` `value` prop will cause expensive re-renders 😅

hermanwikner: Thanks for the feedback! @stipsan 

Can you clarify what you mean by "_these **were all** separate context providers_"? We could definitely move the `selectedPath` state to its own provider (+ the `mentionOptions`) – but I'm not sure we can move much else to their own providers since they need access to the comments data.

stipsan: Sure they can :)
First of all, a lot of times a state setter doesn't need to know what the current state is. For example `useState` has a state setter that returns the previous value in its updater.
So if this was the case before:
```tsx
import React from 'react'

const StateContext = React.createContext(false)
const SetStateContext = React.createContext(() => {})

export default Provider({children}) {
  const [value, _setValue] = React.useState(false)
  const setValue = React.useCallback(() => {setValue(!value)}, [value])

  return <StateContext.Provider value={value}><SetStateContext.Provider value={setValue}>{children}</SetStateContext.Provider></StateContext.Provider>
}
```
```tsx
import React from 'react'

const StateContext = React.createContext(false)
const SetStateContext = React.createContext(() => {})

export default Provider({children}) {
  const [value, _setValue] = React.useState(false)
  const setValue = React.useCallback(() => {setValue(value => !value)}, [])

  return <StateContext.Provider value={value}><SetStateContext.Provider value={setValue}>{children}</SetStateContext.Provider></StateContext.Provider>
}
```
The Studio codebase also has an example of how to do this with something that uses `useSyncExternalStore` under the hood (`ColorSchemeSetValueContext`, `ColorSchemeValueContext`, and `ColorThemeProvider`) and shows an optimized case.

Secondly, by splitting them up you make hard dependencies explicit and a lot easier to track and understand the flow, and how changes propagate.
And the way React works is that it is able to break up work into units of component renders. It can't partially render a component, it can only partially render a tree. By splitting things up you're allowing React to much better optimize and memoize parts of the tree and reap the full benefits of v18 features. By having super context providers and wrapping them in a `useMemo` is basically the best way to make React run in almost the slowest possible way, and its a lot harder to debug since React devtools only highlight changed props and hooks during profiling, it doesn't highlight which slice of a changed context value prop changed.

This is largely why the React docs when introducing `createContext` showcase this pattern by default: https://react.dev/learn/scaling-up-with-reducer-and-context#moving-all-wiring-into-a-single-file

packages/sanity/src/desk/comments/src/components/pte/config.ts
```diff
@@ -0,0 +1,34 @@
...
+  type: 'object',
+  fields: [
+    {
+      name: 'userId',
+      type: 'string',
+    },
+  ],
+})
+
+const blockType = defineField({
```
skogsmaskin, line 15: Consider adding this, denying decorators and annotations from pasted content.

```
  marks: {
    annotations: [],
    decorators: [],
  },
```

What about links? Should they be allowed to be pasted?

skogsmaskin (commented):
We should have another discussion around what the comment schema contains for this first iteration as links and decorators (strong, bold) is allowed now (but we don't have a way of editing them).

packages/sanity/src/desk/comments/src/components/pte/config.ts
robinpyon: Nope, since we don't render them! I think this closes [EDX-554](https://linear.app/sanity/issue/EDX-554/pasted-links-will-display-default-annotations-at-author-time)?

packages/sanity/src/desk/comments/src/context/onboarding/CommentsOnboardingProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+const VERSION = 1
+const LOCAL_STORAGE_KEY = `sanityStudio:comments:inspector:onboarding:dismissed:v${VERSION}`
+
+const setLocalStorage = (value: boolean) => {
+  window.localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(value))
+}
+
+const getLocalStorage = (): boolean => {
+  const value = window.localStorage.getItem(LOCAL_STORAGE_KEY)
+  return value ? JSON.parse(value) : false
```
bjoerge, line 14: I would recommend wrapping this in a try catch and silently ignore any parse error. If, for some reason the local storage value isn't valid stringified json this will throw a disruptive error. The last thing we want is to ask people to clear their localstorage. Localstorage can be finicky, and sensitive to various security policies in incognito/private mode, so I would also consider adding error handling around every localstorage call making sure it basically cannot throw an error.

packages/sanity/src/desk/comments/src/context/setup/CommentsSetupProvider.tsx
```diff
@@ -0,0 +1,128 @@
...
+export function CommentsSetupProvider(props: CommentsClientProviderProps) {
+  const {children} = props
+  const {dataset, projectId} = useWorkspace()
+  const originalClient = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
+  const [addonDatasetClient, setAddonDatasetClient] = useState<SanityClient | null>(null)
+  const [isRunningSetup, setIsRunningSetup] = useState<boolean>(false)
+
+  const getAddonDatasetName = useCallback(async (): Promise<string | undefined> => {
+    const res = await originalClient.withConfig({apiVersion: 'vX'}).request({
+      uri: `/projects/${projectId}/datasets?datasetProfile=comments&addonFor=${dataset}`,
```
bjoerge, line 25: Should there be a  tag added to this request too? e.g. `sanity.studio.(...)` (you're in a better position to judge what the particular tag should be)

packages/sanity/src/desk/comments/src/hooks/use-mention-options/helpers.ts
```diff
@@ -0,0 +1,72 @@
+import {SanityDocument} from '@sanity/client'
+import {evaluate, parse} from 'groq-js'
+import {MentionOptionUser} from '../../types'
+import {EvaluationParams, Grant, DocumentValuePermission, PermissionCheckResult} from 'sanity'
+
+// TODO:
+// This code is copied from the `grantsStore.ts` and slightly modified to work with the `MentionOptionUser` type.
+// TODO: Refactor this to be a shared function between the `grantsStore.ts` and this file.
```
bjoerge, line 8: +1 do we/should we have a matching ticket for this?

packages/sanity/src/desk/comments/src/hooks/use-mention-options/useMentionOptions.ts
```diff
@@ -0,0 +1,118 @@
...
+        })),
+      ),
+    )
+
+    // 3. Get all the system groups. Use the cached response if it exists to avoid unnecessary requests.
+    const cached = cachedSystemGroups
+    const systemGroup$ = cached ? of(cached) : client.observable.fetch('*[_type == "system.group"]')
+
+    // 4. Check if the user has read permission on the document and set the `canBeMentioned` property
+    const grants$: Observable<MentionOptionUser[]> = forkJoin([users$, systemGroup$]).pipe(
```
bjoerge, line 64: As `forkJoin` will wait for both streams to complete,  this stream will only emit a single (tuple) value and then complete. Just making sure this is the desired behavior, as it means a user will have to reload the studio to be able to mention a newly added user (e.g. a user that was added to the project after this ran initially) – in the future we may want to ma ke it reactive?

packages/sanity/src/desk/comments/src/hooks/useCommentsEnabled.ts
```diff
@@ -0,0 +1,68 @@
...
+  // 2. Config check
+  const {enabled} = useSource().document.unstable_comments
+  const enabledFromConfig = enabled({documentType, documentId: getPublishedId(documentId)})
+
+  const commentsEnabled = useMemo(() => {
+    if (isLoading) {
+      return {
+        isEnabled: false,
+        reason: 'loading',
+      } satisfies CommentsEnabled
```
bjoerge, line 44: is there any benefit from using satisfies vs typing the return value of this function as `CommentsEnabled`? (e.g. changing from `useMemo(() => {…` to `useMemo((): CommentsEnabled => {`  above)

packages/sanity/src/desk/comments/src/types.ts
```diff
@@ -0,0 +1,187 @@
+import {PortableTextBlock, User} from '@sanity/types'
+
+/**
+ * @beta
+ * @hidden
```
bjoerge, line 5: would be good if we could add a todo here that calls out documenting this as a remaining task for this feature

bjoerge (changes requested):
This looks fantastic! I've mostly focused on reviewing the core changes, but had a look at the desk changes as well and added some (hopefully helpful) comments!

packages/sanity/src/core/form/components/formField/FormFieldBaseHeader.tsx
```diff
@@ -43,72 +38,91 @@ const SlotBox = styled(Box)<{
...
+  // Determine if floating card with actions should be shown
+  const shouldShowFloatingCard = showFieldActions || hasComments
 
-  const presenceMaxWidth = calcAvatarStackWidth(MAX_AVATARS)
+  // Determine the shadow level for the card
+  const shadow = (showFieldActions && hasActions) || !hasComments ? 3 : undefined
 
-  // Use the width of the floating card to offset the presence box
+  // Determine if there are no comments or actions
+  const noCommentsOrActions = !comments?.button && !hasActions
```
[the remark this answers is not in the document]
hermanwikner: Good point, I have fixed this now and added a comment that clarifies why we check `comment.button`! 👍

packages/sanity/src/desk/comments/plugin/field/CommentFieldButton.tsx
```diff
@@ -0,0 +1,222 @@
...
+  width: 320px;
+`
+
+const StyledPopover = styled(Popover)(({theme}) => {
+  const {space} = theme.sanity
+  const offsetX = space[1] * 2
+  const offsetY = -3
+
+  return css`
+    /* &[data-placement='right-start'] {
```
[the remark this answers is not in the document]
hermanwikner: This CSS should not be used – I'll remove it 👍

packages/sanity/src/desk/comments/src/context/onboarding/CommentsOnboardingProvider.tsx
hermanwikner: Thanks, good point – I have now wrapped these in try/catch 👍

packages/sanity/src/desk/comments/src/hooks/use-mention-options/helpers.ts
hermanwikner: Yes, I'll create a ticket for this!

packages/sanity/src/desk/comments/src/hooks/use-mention-options/useMentionOptions.ts
hermanwikner: This is the intended behavior and I think it's fine – what do you think? It would be nice to make it reactive in the future – but I don't think we need to worry about it right now!

packages/sanity/src/desk/comments/src/hooks/useCommentsEnabled.ts
hermanwikner: No, there's no benefit! I'll change to `(): CommentsEnabled` instead of `satisfies CommentsEnabled` since that's a much cleaner in this situation 👍

packages/sanity/src/desk/comments/src/context/setup/CommentsSetupProvider.tsx
hermanwikner: Good question 🤔 I am not sure – will check with the team!

packages/sanity/src/desk/comments/src/types.ts
hermanwikner: I am not sure I understand, can you elaborate? :)

packages/sanity/src/core/form/hooks/useDidUpdate.ts
```diff
@@ -14,6 +14,11 @@ import {usePrevious} from './usePrevious'
  * })
  * ```
  */
+
+/**
```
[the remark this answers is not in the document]
hermanwikner: Good catch – will update! 👍

packages/sanity/src/desk/comments/src/context/setup/CommentsSetupProvider.tsx
hermanwikner: Yes, this should be the case. Fixed! Great catch 👍

robinpyon (commented):
Thanks @hermanwikner, this is an enormous effort! Really appreciate the breakdown you've provided and the inclusion of workshop stories / tests too.

I've just left some initial comments that touch on core (nothing significant) and a minor bug in desk. Will continue to review desk components shortly, but hopefully this is helpful for now.

packages/sanity/src/desk/comments/src/context/comments/CommentsProvider.tsx
bjoerge: FWIW, we have started splitting up this provider in #4938, but that work has been put on hold due to other priorities.

packages/sanity/src/desk/comments/src/types.ts
bjoerge: Even if it's considered internal/hidden for now, I think having some rudimentary api docs would be great and also helpful for us internally :)

robinpyon (approved):
Great work @hermanwikner! As mentioned, another pass on desk components will be forthcoming, but all the core changes LGTM

packages/sanity/src/core/form/types/fieldProps.ts
```diff
@@ -26,12 +26,21 @@ import {
   StringInputProps,
 } from './inputProps'
 
+/** @internal @deprecated DO NOT USE */
+export interface FieldCommentsProps {
+  hasComments: boolean
+  button: React.ReactNode
+  isAddingComment: boolean
```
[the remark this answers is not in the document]
bjoerge: Agree, had the same thought ^

bjoerge (approved):
🚢 🚀 ✨ 💬

PR #4887, opened 2023-08-31
feat(titles): Update titles inside desk tool

pedrobonamin:
When the user is moving through documents and panes, update the title property to be  more specific to the current route

### Description

Update the `<title>` element to be more specific to the current route every time a user follows a link in the studio.
Changes titles to use pipes as divider, in the title structure, mention first the most unique thing.

- Previously: `Test Studio - Content`
- Updated:  `Content | Test Studio`

Uses the following pattern:
- Desk root: `{{tool}} | {{desktoolTitle}}`
- Document list: `{{paneTitle}} | {{desktoolTitle}}`
- Document: `{{documentTitle}} | {{desktoolTitle}}`
- Document (new): `New {{schemaTitle}} | {{desktoolTitle}}`

### What to review

Titles inside the studio, navigating through documents in the desk and also navigating between tools.

### Notes for release

Update the <title> element to be more specific to the current route every time a user follows a link in the studio.

https://github.com/sanity-io/sanity/assets/46196328/5fd43734-5f2c-4533-ac36-c083c780701a

### DIFF
--- a/packages/sanity/src/core/config/types.ts
+++ b/packages/sanity/src/core/config/types.ts
@@ -142,6 +142,11 @@ export interface Tool<Options = any> {
   title: string
 
   /**
+   * Determines whether the tool will control the `document.title`.
+   */
+  controlsDocumentTitle?: boolean
+
+  /**
    * Gets the state for the given intent.
    *
    * @param intent - The intent to get the state for.
--- a/packages/sanity/src/core/studio/StudioLayout.tsx
+++ b/packages/sanity/src/core/studio/StudioLayout.tsx
@@ -89,14 +89,14 @@ export function StudioLayout() {
 
     return mainTitle
   }, [activeToolName, name, title])
+  const toolControlsDocumentTitle = !!activeTool?.controlsDocumentTitle
 
   useEffect(() => {
-    if (activeToolName === 'content') {
-      // Will be handled by sanity/src/desk/components/deskTool/DeskTitle.tsx
+    if (toolControlsDocumentTitle) {
       return
     }
     document.title = documentTitle
-  }, [documentTitle, activeToolName])
+  }, [documentTitle, toolControlsDocumentTitle])
 
   const handleSearchFullscreenOpenChange = useCallback((open: boolean) => {
     setSearchFullscreenOpen(open)
--- a/packages/sanity/src/core/studio/StudioLayout.tsx
+++ b/packages/sanity/src/core/studio/StudioLayout.tsx
@@ -84,15 +84,19 @@ export function StudioLayout() {
     const mainTitle = title || startCase(name)
 
     if (activeToolName) {
-      return `${mainTitle} – ${startCase(activeToolName)}`
+      return `${startCase(activeToolName)} | ${mainTitle}`
     }
 
     return mainTitle
   }, [activeToolName, name, title])
 
   useEffect(() => {
+    if (activeToolName === 'content') {
+      // Will be handled by sanity/src/desk/components/deskTool/DeskTitle.tsx
+      return
+    }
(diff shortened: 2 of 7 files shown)


packages/sanity/src/core/studio/StudioLayout.tsx
```diff
@@ -84,15 +84,19 @@ export function StudioLayout() {
...
     if (activeToolName) {
-      return `${mainTitle} – ${startCase(activeToolName)}`
+      return `${startCase(activeToolName)} | ${mainTitle}`
     }
 
     return mainTitle
   }, [activeToolName, name, title])
 
   useEffect(() => {
+    if (activeToolName === 'content') {
```
bjoerge, line 94: A potential issue is that tool names can be customized per studio instance, so it's not guaranteed that the desk tool is named `content`. If it's customized, we'll then end up with two code paths fighting to set the document title. Wondering if it would make sense to let tools themselves declare that they controls `document.title`? This will also enable other tools to declare that they take control over updating `document.title` without risking getting into a fight.

I.e. introduce an optional boolean property to the [Tool](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/config/types.ts#L113) interface (e.g. `controlsDocumentTitle` or something), and then replace this check against `tool.name` with a check against this property instead?

(_Ideally_ perhaps it should be a function that we could call from here to get the title, but since it requires so much state I think exposing a flag is better, at least for now)

pedrobonamin: Thanks @bjoerge that makes sense. 
Added `controlsDocumentTitle` property to the tool and set it as true in the Desk tool. 
I agree with you in the ideal behavior but in this case, given it needs to access the panes and resolve them, and also resolve the document if needed, I couldn't make it fit into a function like that. 

commit: https://github.com/sanity-io/sanity/pull/4887/commits/aefe83b63166b6405ee4cd7c3d5cfe0f15d0a9f9

packages/sanity/src/core/config/types.ts
```diff
@@ -141,6 +141,11 @@ export interface Tool<Options = any> {
    */
   title: string
 
+  /**
+   * Determines whether the tool will control the document title.
```
bjoerge, line 145: very tiiiiny nit: but for clarity maybe write `document.title`? "Document" is such an overloaded word in sanity 😄 

```suggestion
   * Determines whether the tool will control `document.title`.
```

bjoerge (approved):
Just a tiny tiny suggestion, otherwise looks great! Thanks @pedrobonamin!

packages/sanity/src/core/config/types.ts
pedrobonamin: sure!! Thanks!
Tiny things can have a big impact later

bjoerge: Very true :)

robinpyon (commented):
Thanks @pedrobonamin – great work!

One minor issue is that the last document title isn't always displayed in the instance where multiple document panes are open. Here's an [example link](https://test-studio-git-feature-edx-92-all-pages-have-the-same-t-ff0eed.sanity.build/test/content/untitledRepro;grrm;03d08532-319f-44b6-a655-15f2e7bea8f6%2Ctype%3Dauthor%2CparentRefPath%3DbestFriend) which should include "Zinken Hopsf!" in `document.title`

Once that's in I think this is good to go! I also submitted a [PR](https://github.com/sanity-io/sanity/pull/4898) against this which adds the above and refactors `DeskTitle` slightly – would love your thoughts, but also feel free to disregard

robinpyon (approved):

PR #4893, opened 2023-09-02
feat: make splitpanes responsive for minimal breakpoint

kmelve:
### Description

This PR makes the Vision plugin somewhat more responsive for narrow(er) screens by:
- Tuning the responsive grid column values so the 3-4 option boxes collapse more neatly
- Adjust the vertical/horizontal props and sizing to the split panes so they change direction beneath the narrowest breakpoint (600px).

This is useful when you want to test a query on mobile or with a narrow studio in a browser‘s split pane. The balance between the query/param/results panes could probably be tuned a tiny bit more, but it's at least functional.

localhost_3333_test_vision(iPhone 12 Pro)

### What to review

I did try to see if there was a way to reuse the Sanity UI logic to get the breakpoints but ended up doing a simple `window.innerWidth < 600` wrapped in the `narrowBreakpoint()` function. 

### Notes for release

Responsive improvements to the Vision plugin to accommodate narrower screens.

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.tsx
+++ b/packages/@sanity/vision/src/components/VisionGui.tsx
@@ -77,9 +77,13 @@ interface PaneSizeOptions {
   maxSize: number
 }
 
+function narrowBreakpoint(): boolean {
+  return typeof window !== 'undefined' && window.innerWidth > 600
+}
+
 function calculatePaneSizeOptions(rootHeight: number): PaneSizeOptions {
   return {
-    defaultSize: rootHeight / 2,
+    defaultSize: rootHeight / (narrowBreakpoint() ? 2 : 1),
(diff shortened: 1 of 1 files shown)


packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -773,9 +777,14 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
             )}
           </Grid>
         </Header>
-        <SplitpaneContainer flex={1}>
+        <SplitpaneContainer flex={'auto'}>
```
bjoerge, line 780: nit:
```suggestion
        <SplitpaneContainer flex="auto">
```

bjoerge (approved):
Nice - thanks @kmelve!

sjelfull (approved):
Thanks, LGTM!

PR #4894, opened 2023-09-05
fix(desk): use correct dialog positioning at smaller breakpoints

robinpyon:
### Description

This PR ensures that object dialogs are correctly positioned at smaller breakpoints (when the desk collapses into a single pane layout).

There was previously a mismatch between theme breakpoints (used by `<DialogProvider>`) and the `minWidth` value in `<StyledPaneLayout>`. Desk would collapse into a single pane at `512px` whilst dialogs would only use `fixed` positioning at `< 360px`. 

This PR ensures the desk tool will collapse at a theme media breakpoint instead of a hardcoded value, so it should be in lock-step with `<DialogProvider>` moving forward.

### What to review

Collapse your viewport enough so that the studio desk collapses into a single column layout.

In the following examples, scroll to the bottom of the page and click either the 'Sydney' or 'Melbourne' events

- [Before](https://test-studio.sanity.build/test/content/input-standard;objectsTest;0759e8b3-04a8-4e40-872f-e447856e8bda)
- [After](https://test-studio-git-feature-edx-197-studio-dialogs-are-incor-5862fd.sanity.build/test/content/input-standard;objectsTest;0759e8b3-04a8-4e40-872f-e447856e8bda)

In the after case, object dialogs should appear correctly when scrolled down the page.

One byproduct of this change is that the studio collapses to a single column breakpoint a little earlier than previously (600px vs 512px). I believe this is an acceptable trade-off, but please chime if you think otherwise @kaylasanity 

### Notes for release

Fixes an issue where dialogs wouldn't correctly appear at smaller breakpoints

### DIFF
--- a/packages/sanity/src/desk/components/deskTool/DeskTool.tsx
+++ b/packages/sanity/src/desk/components/deskTool/DeskTool.tsx
@@ -1,4 +1,4 @@
-import {PortalProvider, useToast} from '@sanity/ui'
+import {PortalProvider, useTheme, useToast} from '@sanity/ui'
 import React, {memo, Fragment, useState, useEffect, useCallback} from 'react'
 import styled from 'styled-components'
 import isHotkey from 'is-hotkey'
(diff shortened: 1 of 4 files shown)


ninaandal (approved):
This is great, thank you Rob 😃

kaylasanity (approved):

bjoerge (approved):

PR #4896, opened 2023-09-05
feat(WCAG): Associate field descriptions to inputs

pedrobonamin:
### Description

In the editor, some fields have a description in addition to just a label. This description is not associated with the field in code. This makes it harder to discover for some users, including screenreader users.

With this changes, when fields have a description, this description will be associated to the input by using `aria-describedby`. This is discoverable by screen readers.

**Note**: Couldn't make it work in portable text editor correctly, as the focused element in PTE is not the same one that receives the aria-describedby value when passing the prop , a follow up ticket will be created for this.

### What to review

Is there any missing field type to consider?
To trigger the descriptions using voice over functionality you have to:
1) Focus on the field, voiceover will let you know it has more information available.
2) Click `cmd+shift+option+/` to show the information. 

### Notes for release

WCAG : When fields have a description, associate them to the input by using `aria-describedby`.

https://github.com/sanity-io/sanity/assets/46196328/d072b1f9-1761-49c0-b84b-686c59e13193

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -67,6 +67,7 @@ export type PortableTextEditableProps = {
   scrollSelectionIntoView?: ScrollSelectionIntoViewFunction
   selection?: EditorSelection
   spellCheck?: boolean
+  describedBy?: string
 }
 
 /**
@@ -91,6 +92,7 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
     selection: propsSelection,
     scrollSelectionIntoView,
     spellCheck,
+    describedBy,
     ...restProps
   } = props
 
@@ -405,6 +407,7 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
         renderLeaf={renderLeaf}
         style={props.style}
         scrollSelectionIntoView={scrollSelectionIntoViewToSlate}
+        aria-describedby={describedBy}
       />
     ),
     [
@@ -420,6 +423,7 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
       renderElement,
       renderLeaf,
       scrollSelectionIntoViewToSlate,
+      describedBy,
     ],
   )
 
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -52,7 +52,10 @@ const EMPTY_DECORATORS: BaseRange[] = []
 /**
  * @public
  */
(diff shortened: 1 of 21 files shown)


packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -67,6 +67,7 @@ export type PortableTextEditableProps = {
   scrollSelectionIntoView?: ScrollSelectionIntoViewFunction
   selection?: EditorSelection
   spellCheck?: boolean
+  describedBy?: string
```
skogsmaskin, line 70: Could we rather support arbitrary props forwarding here, as there are already quite a few props to this component?

That way we could send in our own HTML props whatever they might be, and don't have to be overly specific about props we don't really care about in this context.

So something like this:

```
      <SlateEditable
        {...props}
        autoFocus={false}
        className={props.className || 'pt-editable'}
        decorate={decorate}
        onBlur={handleOnBlur}
        onCopy={handleCopy}
        onDOMBeforeInput={handleOnBeforeInput}
        onFocus={handleOnFocus}
        onKeyDown={handleKeyDown}
        onPaste={handlePaste}
        readOnly={readOnly}
        renderElement={renderElement}
        renderLeaf={renderLeaf}
        scrollSelectionIntoView={scrollSelectionIntoViewToSlate}
      />
```

skogsmaskin (changes requested):

packages/@sanity/portable-text-editor/src/editor/Editable.tsx
bjoerge: > Could we rather support arbitrary props forwarding here

For that to work we'll also have to make PortableTextEditableProps extend `ComponentProps<typeof SlateEditable>`, right? I don't have the full picture, but seems nice to keep the props interface minimal here? Guess it's also something we can always consider doing later too.

On a related note: Any reason to not name it `aria-describedby` here? That would help us avoid ambiguity if we decided to support arbitrary props forwarding some time in the future.

bjoerge (approved):
Looks great! Thank you @pedrobonamin 🙌🏼

skogsmaskin:
@bjoerge, @pedrobonamin - I have made a suggestion on how we can forward those props in a PR targeting this branch:

https://github.com/sanity-io/sanity/pull/4923

skogsmaskin (changes requested):
Please consider my [suggestion](https://github.com/sanity-io/sanity/pull/4896#issuecomment-1717472277) on how we can support whatever html props.

bjoerge:
> Please consider my https://github.com/sanity-io/sanity/pull/4896#issuecomment-1717472277 on how we can support whatever html props.

I'm OK with this approach 👍🏼

skogsmaskin (approved):

PR #4897, opened 2023-09-05
chore(ci): fix remote caching on gh actions

stipsan:
### Description

Remote caching in turborepo is working everywhere but GH Actions, turns out it's because of an expired token.

### What to review

GH Actions should now report "Cache hit" on `yarn build` a lot of the time, and if re-running the action and its previous build didn't error then it should have "Cache hit" 100% of the time.

### Notes for release

N/A


binoy14 (approved):
Thanks for the fix!

PR #4900, opened 2023-09-06
fix(import): increase concurrency of patching, document creation

rexxars:
### Description

We had some concurrency issues when strengthening references in parallel, but this hasn't been the case for years now. This improves the performance of the import by doing more strengthening in parallel.

### What to review

- Importing still works as expected

### Notes for release

- Improved performance of dataset imports using the `sanity dataset import` command/`@sanity/import` module


bjoerge (approved):

binoy14 (approved):

bjoerge:
@rexxars is this good to merge?

rexxars:
> @rexxars is this good to merge?

Just wanted to do a few tests to ensure we don't hit the rate limiter

PR #4902, opened 2023-09-07
refactor: remove unused `useConditionalReadOnly` hook

rexxars:
### Description

Found this hook while refactoring, but did not find any place we were rendering the context provider - so assuming it is unused.

### What to review

- Conditional read only fields still work as expected

### Notes for release

None, internal change.


sjelfull (approved):
I believe this is unused, yes. This was replaced by `resolveConditionalProperty` in v3.

bjoerge (approved):
Good catch, thanks!

PR #4903, opened 2023-09-07
fix(portable-text-editor): normalize user insert data

skogsmaskin:
### Description

This will fix a couple of problems that I discovered in the Portable Text Editor code.

* Improved normalization of nodes to always assign _key to nodes that are missing keys. I discovered that in some cases children of text blocks (inline blocks) would not get a key assigned like they were supposed to. Also span nodes was assigned keys in a separate patch after they were created. This ensures they get the key assigned at creation time.
* Added value validation rule to test that a text block's children are an actual array.
* Allow a text block's `markDefs` to be undefined. According to the PT-spec and TS-type, this is optional, and we should allow them to be missing from an input value. Added tests for this.
* Don't prefix void child `_key` with block key as it's not necessary and just complicates things. These nodes doesn't exist in the real data, only in the temporary editor state.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
+++ b/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
@@ -320,7 +320,7 @@ function _updateBlock(
             debug('Updating changed inline object child', currentBlockChild)
             Transforms.setNodes(
               slateEditor,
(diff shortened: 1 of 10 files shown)


rexxars:
@skogsmaskin @christianhg Could you find out if this PR is still relevant?

skogsmaskin:
Most of this has been handled through other PRs. Closing.

PR #4904, opened 2023-09-07
chore(docs): compare PRs against next branch instead of current

binoy14:
### Description

- fixes token mismatch when deleting dataset
- better message when no changes to documentation coverage

### What to review

The actions conditionals are valid

### Notes for release

N/A

### DIFF
--- a/.github/workflows/docReport.yml
+++ b/.github/workflows/docReport.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
   # Build on commits pushed to branches without a PR if it's in the allowlist
   push:
-    branches: [current]
+    branches: [current, next]
 
 jobs:
   report:
--- a/.github/workflows/docReport.yml
+++ b/.github/workflows/docReport.yml
@@ -40,11 +40,11 @@ jobs:
         env:
           NODE_OPTIONS: --max_old_space_size=8192
 
-      - name: Create Docs Report on current
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/current' }}
+      - name: Create Docs Report on current or next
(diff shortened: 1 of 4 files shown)


.github/workflows/docReport.yml
```diff
@@ -40,11 +40,11 @@ jobs:
         env:
           NODE_OPTIONS: --max_old_space_size=8192
 
-      - name: Create Docs Report on current
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/current' }}
+      - name: Create Docs Report on current or next
+        if: ${{ github.event_name == 'push' && (github.ref == 'refs/heads/current' || github.ref == 'refs/heads/next') }}
```
binoy14, line 44: Not sure if the `event_name` is still push when a PR is merged. Does anyone know?

.github/workflows/docReport.yml
```diff
@@ -40,11 +40,11 @@ jobs:
...
         env:
           DOCS_REPORT_TOKEN: ${{ secrets.DOCS_REPORT_DATASET_TOKEN }}
-          DOCS_REPORT_DATASET: production
+          DOCS_REPORT_DATASET: ${{ github.ref == 'refs/heads/current' && production || next}}
```
binoy14, line 47: Also is this syntax valid?

.github/workflows/docReport.yml
bjoerge: Seems fair to assume so, esp. since there doesn't seem to be a `merge` event.

bjoerge (approved):
Nice!

binoy14:
oh just realized the `next` branch is not in the allowed list for the workflow to run. Will add it

.github/workflows/docReport.yml
binoy14: It might be trial and error. I will merge and check

PR #4906, opened 2023-09-07
chore(docs): fix dataset delete script

binoy14:
### Description

- add more logging in delete script
- use PR numbers as dataset names instead of github ref
- fixes turbo token in docs actions 

### What to review

Seems like when the PR is closed the ref actually points to the next branch so it ends up deleting the next dataset. It seems like the `github.event.number` should be more consistent. [Reference](https://shipit.dev/posts/trigger-github-actions-on-pr-close.html) 

### Notes for release

N/A

### DIFF
--- a/.github/workflows/docReport.yml
+++ b/.github/workflows/docReport.yml
@@ -51,14 +51,14 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         env:
           DOCS_REPORT_TOKEN: ${{ secrets.DOCS_REPORT_DATASET_TOKEN }}
-          DOCS_REPORT_DATASET: ${{ github.event.number }}
+          DOCS_REPORT_DATASET: pr-${{ github.event.number }}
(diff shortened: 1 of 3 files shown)


.github/workflows/docReport.yml
```diff
@@ -51,14 +51,14 @@ jobs:
         if: ${{ github.event_name == 'pull_request' }}
         env:
           DOCS_REPORT_TOKEN: ${{ secrets.DOCS_REPORT_DATASET_TOKEN }}
-          DOCS_REPORT_DATASET: ${{ github.ref_name }}
+          DOCS_REPORT_DATASET: ${{ github.event.number }}
```
bjoerge, line 54: should we prefx it so it's easy to identify as a temporary dataset?

binoy14: sure I can take a look at how to do that

bjoerge (approved):

PR #4915, opened 2023-09-11
fix(core): fixes issue with restoring document in nested panes

binoy14:
### Description

This PR fixes the hacky workaround in https://github.com/sanity-io/sanity/pull/4668 in a more predictable way where it only navigates to the intent when the operation is successfully completed.

### What to review

The solution makes sense and there would be no side effects of this change. 

### Notes for release

- Fixes issue with restoring document history in nested panes

### DIFF
--- a/packages/sanity/src/desk/documentActions/HistoryRestoreAction.tsx
+++ b/packages/sanity/src/desk/documentActions/HistoryRestoreAction.tsx
@@ -1,20 +1,36 @@
 import {RestoreIcon} from '@sanity/icons'
-import React, {useCallback, useMemo, useState} from 'react'
-import {DocumentActionComponent, DocumentActionDialogProps, useDocumentOperation} from 'sanity'
+import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react'
+import {
+  DocumentActionComponent,
+  DocumentActionDialogProps,
+  useDocumentOperation,
+  useDocumentOperationEvent,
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
Great fix, @binoy14!

packages/sanity/src/desk/documentActions/HistoryRestoreAction.tsx
```diff
@@ -1,20 +1,36 @@
...
-    // wrapping in setTimeout gives the onComplete time to finish before navigating
-    setTimeout(() => navigateIntent('edit', {id, type}), 0)
-  }, [restore, revision, navigateIntent, id, type, onComplete])
+  }, [restore, revision, onComplete])
+
+  /**
+   * If the restore operation is successful, navigate to the document edit view
+   */
+  useEffect(() => {
+    if (!event || event === prevEvent.current) return
```
ricokahler, line 28: I don't see you updating `prevEvent.current`. Is that intended?

ricokahler: ah are you just using the ref so it doesn't update the effect?

binoy14: ah missed it, good catch! will update it

ricokahler (approved):

PR #4919, opened 2023-09-12
chore(docs): adds docs for useCurrentUser hook

binoy14:
### Description

Adds docs for useCurrentUser hook

### What to review

This hook changes from `internal` to `public` are there any concerns about that? 

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/store/user/hooks.ts
+++ b/packages/sanity/src/core/store/user/hooks.ts
@@ -22,7 +22,22 @@ export function useUser(userId: string): LoadingTuple<User | null | undefined> {
   return useUserViaUserStore(useMemo(() => [userStore, userId], [userId, userStore]))
 }
 
-/** @internal */
+/**
+ * Retrieves the current user from the source.
+ *
+ * @returns The current user or null if not available.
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
Looks good, thanks! Would be nice if the docs report changes could be merged separately.

packages/sanity/src/core/store/user/hooks.ts
```diff
@@ -22,7 +22,22 @@ export function useUser(userId: string): LoadingTuple<User | null | undefined> {
   return useUserViaUserStore(useMemo(() => [userStore, userId], [userId, userStore]))
 }
 
-/** @internal */
+/**
+ * Retrieves the current user from the source.
```
rexxars, line 26: To third party developers, I don't think it is clear what a "source" is, and as we don't really utilize the concept for anything at the moment, I think we should steer away from using it.

```suggestion
 * Retrieves information about the currently authenticated user.
```

rexxars (changes requested):

rexxars (approved):

PR #4921, opened 2023-09-12
feat(i18n): implement a very simple tags/components parser to use in a Translate component

bjoerge:
Implements a simple tags/components (not html!) parser for use with i18n, plus a <Translate component that takes a translation key and a component map, looks up the a translation string associated with the key, and replaces any `<Component>` parts of the translation string with the matching component provided through the `componentMap`.

- See unit tests for the [parser](https://github.com/sanity-io/sanity/blob/13dcc260cf483f194e5b53c22ae2c08d5c07540e/packages/sanity/src/core/i18n/__tests__/simpleParser.test.ts) or the [Translate](https://github.com/sanity-io/sanity/blob/13dcc260cf483f194e5b53c22ae2c08d5c07540e/packages/sanity/src/core/i18n/__tests__/Translate.test.tsx) component for more details
- Example usage [here](https://github.com/sanity-io/sanity/blob/de9a406d58a7314b11c640011c7cda93206ccd17/dev/test-studio/components/TranslateExample.tsx) with live [example here](https://test-studio-git-feature-sdx-598simple-translate-component.sanity.build/test/content/translate)

### DIFF
--- a/dev/test-studio/components/TranslateExample.tsx
+++ b/dev/test-studio/components/TranslateExample.tsx
@@ -1,0 +1,27 @@
+import {Translate, useTranslation} from 'sanity'
+import React from 'react'
+import {Card, Text} from '@sanity/ui'
+import {InfoFilledIcon} from '@sanity/icons'
+
+export function TranslateExample() {
+  const {t} = useTranslation('testStudio')
+  return (
+    <Card padding={4}>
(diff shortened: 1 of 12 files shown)


rexxars (approved):
This is great - thank you so much

PR #4929, opened 2023-09-14
feat(comments): minor UI updates

robinpyon:
### Description

- Tooltip delays (500 in, 0 out), `<TooltipDelayGroupProvider>` wraps comment actions
- Optically aligned icons (still temporary, changes forthcoming in `@sanity/icons`)
- Display total comment (not thread) count per field
- Minor changes to open/resolved selects, comment context menu alignment + field headers
- Primary tone on new comment container, also broken out into a separate component
- Other super minor tweaks, slightly more condensed comments, PTE container
- Bump `@sanity/ui` to `1.8.2`
- Uses current dataset name when querying for metacontent comments


packages/sanity/src/core/comments/components/list/CommentsListItem.tsx
```diff
@@ -186,8 +186,7 @@ export function CommentsListItem(props: CommentsListItemProps) {
             onChange={setValue}
             onEditDiscard={cancelEdit}
             onSubmit={handleSubmit}
-            placeholder="Reply…"
-            ref={replyInputRef}
```
hermanwikner, line 190: We should keep this ref so that `replyInputRef.current?.focus()` still works in `handleSubmit`.

hermanwikner (approved):
Thanks Rob, looks great 👍

PR #4930, opened 2023-09-14
fix(core): fix array items not showing in field groups

binoy14:
### Description

This PR adds a debug schema, the issue is when you have an array of items in a field group which is smaller then the height of the window which does not provide a scroll bar the virtualizer does not trigger a re-render causing items to not be recalculated and missing items in the array. 

### What to review

The fix adds a key to the scroll element which will cause it to change when the field group changes. This change in the scrollbar should trigger a recalculation in the virtualization library according to this lines of [code](https://github.com/TanStack/virtual/blob/beta/packages/virtual-core/src/index.ts#L376)   

### Notes for release

Fixes an issues with array items not showing all items inside a field group under specific conditions

### DIFF
--- a/dev/test-studio/schema/debug/virtualizationDebug.tsx
+++ b/dev/test-studio/schema/debug/virtualizationDebug.tsx
@@ -19,8 +19,8 @@ export function MyField(props: FieldProps) {
   )
 }
 
-const getAuthors = (client: SanityClient) => {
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
This will probably just fix it for root level groups(?). We might want to look into a fix that also works for groups inside groups, but this lgtm as a first step.

PR #4935, opened 2023-09-15
fix(comments): only calculate current caret element on demand

robinpyon:
### Description

A temporary fix which only requests the element at the current _caret_ on demand rather than listening to all `selectionchange` events. 

This exposes `activeCaretElement` on `CommentInputProvider`, and will be replaced with an alternate implementation in the near future (which selectively listens only when the mentions menu is open).


hermanwikner (approved):
Thank you! 🙏

PR #4936, opened 2023-09-15
Feat/comments mvi 1 fix mentions menu

skogsmaskin:


packages/sanity/src/core/comments/components/pte/comment-input/CommentInputProvider.tsx
```diff
@@ -60,124 +57,154 @@ export function CommentInputProvider(props: CommentInputProviderProps) {
...
+        isPortableTextTextBlock(block) &&
+        (block?.children || [])?.some((c) => (isPortableTextSpan(c) ? c.text : c.userId)),
     )
   }, [value])
 
   const hasChanges = useCommentHasChanged(value)
 
-  const focusLastBlock = useCallback(() => {
-    const block = PortableTextEditor.focusBlock(editor)
+  const focusEditor = useCallback(() => PortableTextEditor.focus(editor), [editor])
```
hermanwikner, line 76: Would it be possible to update this function to set the cursor at the end if there is already a value on focus? Currently, when you click "Edit Comment", the cursor is placed at the beginning of the message. But it would be nice if it was placed at the end!

packages/sanity/src/core/comments/components/mentions/MentionsMenu.tsx
```diff
@@ -25,23 +20,39 @@ const FlexWrap = styled(Flex)({
...
   loading: boolean
+  inputElement?: HTMLDivElement | null
   onSelect: (userId: string) => void
   options: MentionOptionUser[] | null
 }
 
 export const MentionsMenu = React.forwardRef(function MentionsMenu(
   props: MentionsMenuProps,
-  ref: React.Ref<HTMLDivElement>,
+  ref: React.Ref<FIXME>,
```
hermanwikner, line 36: I think this can be changed to: `React.Ref<MentionsMenuHandle>` and the `ref={ref}` added to the `Flex` which will get an TS error can be removed (it is not being used for anything)

hermanwikner (approved):
Thank you, this is great 💯 I left a few minor comments, but otherwise I think this is good to go!

packages/sanity/src/core/comments/components/pte/comment-input/Editable.tsx
```diff
@@ -22,7 +24,7 @@ const EditableWrapStack = styled(Stack)(() => {
 export const StyledPopover = styled(Popover)(() => {
   return css`
     // Position the Popover relative to the @
-    transform: translateX(6px); // todo: improve
+    transform: translate(6px, 6px); // todo: improve
```
sjelfull, line 27: Ideally we would configure this through a prop on the `Popover` component ([`offset`](https://floating-ui.com/docs/offset)), but `@sanity/ui` doesn't support this yet.

skogsmaskin:
I forced pushed a changed that gets rid of the `selectCurrent` imperative handle for the MentionsMenu. Works out of the box.

packages/sanity/src/core/comments/components/mentions/MentionsMenu.tsx
skogsmaskin: Yes, well spotted! I've force-pushed a change now that does exactly this.

packages/sanity/src/core/comments/components/pte/comment-input/CommentInputProvider.tsx
skogsmaskin: Yes, I'll look into it. I think it preferably should be an explicit function that gets called on edit - and not a general focus function here.

skogsmaskin: @hermanwikner - see my two lastest commits (new)

packages/sanity/src/core/comments/components/pte/comment-input/Editable.tsx
hermanwikner: I'll look into this as a separate thing once this is merged

PR #4954, opened 2023-09-27
fix(core): make sure to re-emit current edit state when re-subscribing to document validation

bjoerge:
### Description

This fixes an issue that currently happens when editing a document that has a reference to an unpublished document, switching to another document and then switching back. Sometimes this could lead to the validation error for unpublished document reference being ignored.

Steps to reproduce:
1. Go to https://test-studio.sanity.build/test/content
2. Navigate to `Standard Inputs`  > `Reference Test` > `REPRO SDX-736 #1`" and observe that the reference field has the correct validation error
3. Switch to `REPRO SDX-736 #2`
4. Switch back to `REPRO SDX-736 #1`" and notice that the validation error has disappeared.

### What to review
Go to the branch deployment (https://test-studio-git-sdx-736.sanity.build/test/content) and make sure the above steps does not cause the validation error to disappear

I also took the opportunity to fixe a couple of deprecation warnings while at it.

### Notes for release
- Fixes an issue that could sometimes make reference validation errors disappear when switching between documents

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/validation.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/validation.ts
@@ -17,10 +17,7 @@ import {
   groupBy,
   map,
   mergeMap,
-  publishReplay,
-  refCount,
   scan,
-  share,
   shareReplay,
   skip,
   throttleTime,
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/validation.ts
```diff
@@ -105,7 +106,7 @@ export const validation = memoize(
         // so only pass on documents if _other_ attributes changes
         return shallowEquals(omit(prev, '_rev', '_updatedAt'), omit(next, '_rev', '_updatedAt'))
       }),
-      share(),
+      shareLatestWithRefCount(),
```
bjoerge, line 109: This is the line that fixes the issue. Since we were using share, re-subscribing to this observable didn't publish the current edit state of the document, leaving the validation in a pending state

binoy14 (approved):
The test looks good to me but it would be good to have someone else's eyes on it as well since I am not very familiar with this

ricokahler (approved):

PR #4955, opened 2023-09-27
fix(block-tools): fix Google Docs preprocessing issue on Windows

skogsmaskin:
### Description

This will fix an issue on Windows when serializing a Google Docs HTML fragment into blocks.

The clipboard HTML was slightly different on Windows and our code didn't take that into account.

I have made some xpath matchers more specific and resilient now so they work both places. 

I have also added tests for this.

### What to review

That pasting from Google Docs into the PTE on Windows and MacOS will work correctly, without adding everything as bold.

### Notes for release
* Fix an issue where pasting from Google Docs into the Portable Text editor would make everything 'bold' on Windows computers.


rexxars (approved):
👍

PR #4957, opened 2023-09-28
fix(core): unset changes in reverse order in patch operation

ninaandal:
### Description
There was an issue with reverting changes in arrays of primitives. The list would not be cleared when trying to revert all changes. 
This had to do with shifting indices of the items, but not on the patch operations, so that triggering this operation 
```
{
  unset: ['arrayOfStrings[0]', 'arrayOfStrings[1]', 'arrayOfStrings[2]']
}
```
will remove the first and last item, but the `arrayOfStrings[1]` will be left out of the patch operation. 

By reversing the order of the items in the `unset` operations, the issue with shifted indices will not cause problems. 

`{ unset: ['arrayOfStrings[2]', 'arrayOfStrings[1]', 'arrayOfStrings[0]'] }`

### What to review
Reverting changes, especially for arrays, are working as expected. 

### Notes for release
Fixed issue where reverting all changes did not clear the list

### DIFF
--- a/packages/sanity/src/core/field/diff/changes/undoChange.ts
+++ b/packages/sanity/src/core/field/diff/changes/undoChange.ts
@@ -1,3 +1,4 @@
+import {toArray} from 'rxjs/operators'
 import {
   isIndexSegment,
   isKeyedObject,
@@ -56,10 +57,7 @@ export function undoChange(
       .forEach((child) => undoChange(child, rootDiff, documentOperations))
 
     patches.push(
-      ...buildUnsetPatches(
-        rootDiff,
-        unsetChanges.reverse().map((unsetChange) => unsetChange.path),
-      ),
+      ...buildUnsetPatches(rootDiff, unsetChanges.map((unsetChange) => unsetChange.path).reverse()),
     )
   } else if (change.diff.action === 'added') {
     // The reverse of an add operation is an unset -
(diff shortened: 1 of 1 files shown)


bjoerge (commented):
Great catch! To reduce the potential for future bugs, maybe either use the non-mutating [`.toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) instead of `.reverse()`?

ninaandal:
> Great catch! To reduce the potential for future bugs, maybe use the non-mutating [`.toReversed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) instead of `.reverse()`?

Thanks @bjoerge! 

Using `toReversed()` works when I run it, but I get an error in VSCode when trying to use it... It keeps suggesting `reverse()` instead. 
Error code: `Property 'toReversed' does not exist on type 'FieldChangeNode[]'. Did you mean 'reverse'?` 

Any idea why it's giving an error? Is it pretty new array method, could that be the issue?

bjoerge:
Ah, right, it's included in ES2023, which I don't think we have enabled in tsconfig yet. Moving .reverse() to after map() would go a long way for now, since .map() already creates a new array.

bjoerge (approved):

PR #4958, opened 2023-09-29
fix(core): accessibility improvements on document lists

ninaandal:
### Description
Items in document lists were marked up as `div` elements, but semantically, they are lists. The list items are now `li`, while the wrapper is a `ul`. 

### What to review
Make sure that document lists work as intended and that semantically, they are correct. 

### Notes for release
Improves accessibility for document lists


bjoerge (approved):
Thanks @ninaandal!

PR #4961, opened 2023-09-29
test(e2e): prep work for making e2e more isolated

jtpetty:
### Description

Updating our e2e test bench to not explicitly use the `testSanityClient` but rather the client to use be passed into the test. This will set us up for better test isolation going forward.

### What to review

The overall approach. The next step would be to do something similar with creating documents.

### Notes for release

This is a test bench refactor and is not relevant in the release notes


bjoerge (approved):
LGTM!

binoy14 (approved):

PR #4963, opened 2023-10-02
fix(core): set default placements for certain studio popovers

bjoerge:
### Description

The new default fallback  placements for popovers released in Sanity UI v1.8.2 are good for most use cases, but in the Studio it has been causing issues for popovers used by the reference input, cross dataset reference input and the document type selector used by the array input.

This PR fixes these issues by explicitly choosing better fallback placements.

## Before
image

### What to review

Make popovers used by the following components places themselves where they should and renders nicely:
- reference input,
- cross dataset reference input
- the document type selector used by the array input.

### Notes for release

- Fixed rendering issues with certain popovers being caused by wrong placement


ninaandal (approved):
lgmt 🚀

PR #4968, opened 2023-10-03
feat(i18n): convert document actions strings to use i18n primitives

bjoerge:
### Description
This converts various document action strings into 18n-keys

Also includes some minor cleanup. Commits can be reviewed one-by-one.

### DIFF
--- a/packages/sanity/src/desk/i18n/resources.ts
+++ b/packages/sanity/src/desk/i18n/resources.ts
@@ -77,10 +77,10 @@ const deskLocaleStrings = {
 
   /** --- DUPLICATE ACTION --- */
   /** Tooltip when action is disabled because the operation is not ready   */
-  'action.duplicate.disabled.notReady': 'Operation not ready',
+  'action.duplicate.disabled.not-ready': 'Operation not ready',
 
   /** Tooltip when action is disabled because the document doesn't exist */
-  'action.duplicate.disabled.nothingToDuplicate':
+  'action.duplicate.disabled.nothing-to-duplicate':
     "This document doesn't yet exist so there‘s nothing to duplicate",
 
   /** Label for the "Duplicate" document action */
--- a/packages/sanity/src/desk/i18n/resources.ts
+++ b/packages/sanity/src/desk/i18n/resources.ts
@@ -4,7 +4,11 @@
  * @internal
  */
 const deskLocaleStrings = {
-  /** Label for the "Publish" document action when there are pending changes.*/
+  /** --- PUBLISH ACTION --- */
+  /** Tooltip when action is disabled because the studio is not ready.*/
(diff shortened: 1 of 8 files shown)


binoy14 (approved):

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -37,11 +41,80 @@ const deskLocaleStrings = {
...
+  /** Fallback tooltip for the Unpublish document action when publish is invoked for a document with live edit enabled.*/
+  'action.unpublish.liveEdit.disabled':
+    'This document has live edit enabled and cannot be unpublished',
+
+  /** --- RESTORE ACTION --- */
+  /** Label for the "Restore" document action */
+  'action.restore.label': 'Restore',
+
+  /** Fallback tooltip for when user is looking at the initial version */
+  'action.restore.disabled.cantRestoreInitial': "You can't restore to the initial version",
```
rexxars, line 110: I suggest we do not use contractions in identifiers, so I would write this as `action.restore.disabled.cannot-restore-initial`

rexxars (changes requested):
We should align on camelCase vs kebab-case for the i18n resources, eg `action.delete.disabled.notReady` vs `action.delete.disabled.not-ready`.

I am personally leaning towards kebab-case, but open to input.

packages/sanity/src/desk/i18n/resources.ts
bjoerge: oh, good point - guess we should stick to snake-case as well. Fixed in 8dde36dee0e58072869656ba530a8f8c62286004

bjoerge:
> I am personally leaning towards kebab-case, but open to input.

Agree. Unless we think these will appear as object keys I don't see any point of using camel case

rexxars (changes requested):
Since we agree, it would be great if you changed the keys here to be kebabcase, so `action.restore.confirmDialog.confirmDiscardChanges` => `action.restore.confirmDialog.confirm-discard-changes` etc.

bjoerge:
> Since we agree, it would be great if you changed the keys here to be kebabcase, so action.restore.confirmDialog.confirmDiscardChanges => action.restore.confirmDialog.confirm-discard-changes etc. 

Fixed in c07fd19 - thanks for pointing out!

rexxars (changes requested):
Still seems to be using camelcase? Examples:

- discardChanges => discard-changes
- confirmDialog => confirm-dialog
- notPublished => not-published
- notReady => not-ready
- nothingToDelete => nothing-to-delete

There's more, but you get the idea

bjoerge:
🤦🏼

aight, new attempt. Also kebab-cased some existing keys for the publish action.

rexxars (approved):

PR #4975, opened 2023-10-04
feat(cli): add CDR metadata to GraphQL json schema

ricokahler:
### Description

The following PR adds CDR metadata to the GraphQL json schema to enable CDR support in the `gatsby-source-sanity` plugin.

There's a new key under the `ConvertedType` interface labeled `crossDatasetReferenceMetadata` that includes the dataset name and the type names.

Important note: Our graphql service _must_ be updated before this gets merged else all GraphQL deployments with crossDatasetReferences will fail because of the extra key.

Fixes SDX-768 
Fixes SDX-765

### What to review

See the updated types and tests to see how the GraphQL JSON schema will change. 

### Notes for release

N/A (more of an internal add for the time being)

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/graphql/types.ts
+++ b/packages/sanity/src/_internal/cli/actions/graphql/types.ts
@@ -45,6 +45,10 @@ export interface ConvertedType {
   interfaces?: string[]
   originalName?: string
   isReference?: boolean
+  crossDatasetReferenceMetadata?: {
+    dataset: string
+    typeNames: string[]
+  }
 }
 
 export interface ConvertedDocumentType extends ConvertedType {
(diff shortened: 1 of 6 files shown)


packages/sanity/src/_internal/cli/actions/graphql/types.ts
```diff
@@ -45,6 +45,10 @@ export interface ConvertedType {
   interfaces?: string[]
   originalName?: string
   isReference?: boolean
+  crossDatasetReferenceMetadata?: {
```
binoy14, line 48: ❓should this also include `projectId: string` ?

binoy14 (approved):
LGTM! Tested it and everything works great!

rexxars (approved):

PR #4976, opened 2023-10-05
refactor(i18n): couple of tweaks

rexxars:
### Description

Three minor changes:

- Drop some unused imports from `useLocale` hook
- Make `components` prop required for `<Translate />` component in order to discourage using it when not needed (can always bypass with `components={{}}` if necessary)
- Allow string arrays as values to `<Translate />`. This allows us to use the [list formatting function](https://www.i18next.com/translation-function/formatting#list).


bjoerge (approved):
lgtm, thanks!

PR #4983, opened 2023-10-09
fix(core/comments): ensure authentication config on comment client

skogsmaskin:
### Description

Re-use the authentication configured for the default client. Authentication method vary if the browser supports cookies or not.


robinpyon (approved):
Thanks @skogsmaskin

PR #4985, opened 2023-10-10
fix(block-tools): enforce list support in schema

skogsmaskin:
### Description

This will make sure that HTML is not deserialized to list blocks where there is no such support in the given schema.

### What to review
That pasting HTML lists into Portable Text will not produce any list blocks with the current schema:
```
{
  type: 'block',
  lists: [],
}
```

### Notes for release
* Fixed a bug where pasting HTML lists into the Portable Text Input would disregard schema configuration for list blocks.

### DIFF
--- a/packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
+++ b/packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
@@ -14,19 +14,17 @@ import {
 } from '../../constants'
 import {BlockEnabledFeatures, DeserializerRule} from '../../types'
 
-export function resolveListItem(listNodeTagName: string): string {
-  let listStyle
-  switch (listNodeTagName) {
-    case 'ul':
-      listStyle = 'bullet'
-      break
-    case 'ol':
-      listStyle = 'number'
-      break
-    default:
-      listStyle = 'bullet'
+export function resolveListItem(
+  listNodeTagName: string,
+  enabledListTypes: string[],
+): string | undefined {
+  if (listNodeTagName === 'ul' && enabledListTypes.includes('bullet')) {
+    return 'bullet'
   }
-  return listStyle
+  if (listNodeTagName === 'ol' && enabledListTypes.includes('number')) {
+    return 'number'
+  }
+  return undefined
 }
 
 export default function createHTMLRules(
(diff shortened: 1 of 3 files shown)


robinpyon (approved):
Thanks @skogsmaskin – LGTM!

packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
```diff
@@ -14,17 +14,24 @@ import {
...
-      listStyle = 'number'
+      if (enabledListTypes.includes('number')) {
+        listStyle = 'number'
+      }
       break
     default:
-      listStyle = 'bullet'
+      listStyle = undefined
   }
   return listStyle
```
rexxars, line 36: This function seems a bit cumbersome to me after the conditionals were added, maybe we could simplify it? 
```ts
  if (listNodeTagName === 'ul' && enabledListTypes.includes('bullet')) {
    return 'bullet'
  }
  if (listNodeTagName === 'ol' && enabledListTypes.includes('number')) {
    return 'number'
  }
  return undefined
```

packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
```diff
@@ -188,15 +195,19 @@ export default function createHTMLRules(
...
     {
-      deserialize(el, next) {
+      deserialize(el, next, block) {
         const tag = tagName(el)
         const listItem = tag ? HTML_LIST_ITEM_TAGS[tag] : undefined
         const parentTag = tagName(el.parentNode) || ''
+        const enabledListItem = resolveListItem(parentTag, options.enabledListTypes)
         if (!listItem || !el.parentNode || !HTML_LIST_CONTAINER_TAGS[parentTag]) {
           return undefined
         }
```
rexxars, line 205: This can be moved down below the conditional, in order to not do unnecessary work:

```suggestion
        if (!listItem || !el.parentNode || !HTML_LIST_CONTAINER_TAGS[parentTag]) {
          return undefined
        }
        
        const enabledListItem = resolveListItem(parentTag, options.enabledListTypes)
```

rexxars (approved):
Nice. Left some small suggestions on code style.

packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
skogsmaskin: Better! I've changed it, tnx!

packages/@sanity/block-tools/src/HtmlDeserializer/rules/html.ts
skogsmaskin: Yes, changed this as well!

PR #4990, opened 2023-10-13
feat(i18n): use Intl APIs for `TimeAgo` component

rexxars:
### Description

Note: review this commit by commit to more easily understand things.

1. Introduces an internal `intlCache`. We will need to instantiate several different instances of the Intl APIs with differing options, and this is a semi-costly operation. The cache ensures we reuse instances where possible. 
2. Made the `useIntlListFormat` hook utilize said cache to prove it works (also, why not)
3. Alters the `useTimeAgo` hook to allow for both past and future dates (_X days ago_, _in Y hours_)
4. Makes the `useTimeAgo` hook use i18n primitives
  - Initially (in 3cce07c), I made it up to the locales to define all the different variations of strings. This was mainly done in order to match the previous output from this hook.
  - After some discussion with @mikolajdobrucki we landed on the fact that longer strings actually makes more sense - especially when English is not the only target. 
  - In e75db31 I reimplemented the hook using only the native Intl APIs, apart from a single "just now" string. I left the old commit around in case we decide to revert to that, if the native Intl approach is not flexible enough.
5. Renames `useTimeAgo` and the `TimeAgo` component to `useRelativeTime` and `RelativeTime`, respectively. This is done because it more closely aligns with the `Intl.RelativeTimeFormat` API, as well as indicating that it can now show dates in the _future_ as well.

Closes SDX-677


packages/sanity/src/core/hooks/__tests__/useTimeAgo.test.tsx
```diff
@@ -51,22 +51,22 @@ describe('useTimeAgo', () => {
       expect(result.current).toBe('Jun 3, 2022')
     })
 
-    it('outputs eg "3w" if less than a month apart', () => {
+    it('outputs eg "3 wks" if less than a month apart', () => {
```
bjoerge: Does this replace the "3w" short format with the slightly longer "3 wks" all over? Since these are used in confined areas of the UI, will this risk running into issues with text no longer fitting the space it's given?

bjoerge (approved):
This is solid, lgtm!

binoy14 (approved):

packages/sanity/src/core/hooks/__tests__/useTimeAgo.test.tsx
rexxars: It does! @mikolajdobrucki approved this from the design side, but we should still have a look around and see if anything breaks - but we can easily change this after merging if problematic.

PR #4993, opened 2023-10-13
feat(i18n): convert review changes button to use i18n primitives

jtpetty:
### Description

This PR adds i18n support for the ReviewChangesButton

### DIFF
--- a/packages/sanity/src/desk/i18n/resources.ts
+++ b/packages/sanity/src/desk/i18n/resources.ts
@@ -26,7 +26,7 @@ const deskLocaleStrings = {
 
   /** Fallback tooltip for the "Publish" document action when publish is invoked for a document with live edit enabled.*/
   'action.publish.live-edit.publish-disabled':
-    'Cannot publish since liveEdit is enabled for this document type',
+    'Cannot publish since Live Edit is enabled for this document type',
 
   /** Tooltip when the "Publish" document action is disabled due to validation issues */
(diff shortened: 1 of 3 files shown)


packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -116,6 +116,22 @@ const deskLocaleStrings = {
...
+  'status-bar.review-changes-button.status.syncing.text': 'Saving...',
+
+  /** Label for button when status is saved */
+  'status-bar.review-changes-button.status.saved.text': 'Saved!',
+
+  /** Primary text for tooltip for the button */
+  'status-bar.review-changes-button.tooltip.text': 'Review changes',
+
+  /** Text for the secondary text for tooltip for the button */
+  'status-bar.review-changes-button.tooltip.changes-saved': 'Changes saved',
```
rexxars, line 131: For consistency I guess the `.tooltip` should be at the end?

rexxars (approved):
As we talked about over PM, maybe the `.tooltip` should be at the end? No strong preferences, but should be consistent at least.

packages/sanity/src/desk/i18n/resources.ts
jtpetty: maybe? it's the same tool tip, but this tool tip has a primary and secondary text.

bjoerge (approved):
LGTM

PR #4994, opened 2023-10-13
fix(i18n): camelCase to kebab-case for resource ids

jtpetty:
### Description

Fixes some resource ids that were changed from camelCase to kebab-case but usages were not changed accordingly.

### DIFF
--- a/packages/sanity/src/desk/documentActions/DiscardChangesAction.tsx
+++ b/packages/sanity/src/desk/documentActions/DiscardChangesAction.tsx
@@ -14,9 +14,9 @@ import {
 } from 'sanity'
 
 const DISABLED_REASON_KEY = {
-  NO_CHANGES: 'action.discardChanges.disabled.noChanges',
-  NOT_PUBLISHED: 'action.discardChanges.disabled.notPublished',
-  NOT_READY: 'action.discardChanges.disabled.notReady',
+  NO_CHANGES: 'action.discard-changes.disabled.noChanges',
+  NOT_PUBLISHED: 'action.discard-changes.disabled.not-published',
+  NOT_READY: 'action.discard-changes.disabled.not-ready',
 } as const
 
 /** @internal */
@@ -54,7 +54,7 @@ export const DiscardChangesAction: DocumentActionComponent = ({
         tone: 'critical',
         onCancel: onComplete,
         onConfirm: handleConfirm,
-        message: t('action.discardChanges.confirmDialog.confirm-discard-changes'),
+        message: t('action.discard-changes.confirmDialog.confirm-discard-changes'),
       },
     [handleConfirm, isConfirmDialogOpen, onComplete, t],
   )
@@ -68,7 +68,7 @@ export const DiscardChangesAction: DocumentActionComponent = ({
       tone: 'critical',
       icon: ResetIcon,
       disabled: true,
-      label: t('action.discardChanges.label'),
+      label: t('action.discard-changes.label'),
       title: (
         <InsufficientPermissionsMessage
           operationLabel="discard changes in this document"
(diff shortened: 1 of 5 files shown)


packages/sanity/src/desk/documentActions/DiscardChangesAction.tsx
```diff
@@ -14,9 +14,9 @@ import {
 } from 'sanity'
 
 const DISABLED_REASON_KEY = {
-  NO_CHANGES: 'action.discardChanges.disabled.noChanges',
-  NOT_PUBLISHED: 'action.discardChanges.disabled.notPublished',
-  NOT_READY: 'action.discardChanges.disabled.notReady',
+  NO_CHANGES: 'action.discard-changes.disabled.noChanges',
```
rexxars, line 17:
```suggestion
  NO_CHANGES: 'action.discard-changes.disabled.no-changes',
```

packages/sanity/src/desk/documentActions/DiscardChangesAction.tsx
```diff
@@ -54,7 +54,7 @@ export const DiscardChangesAction: DocumentActionComponent = ({
         tone: 'critical',
         onCancel: onComplete,
         onConfirm: handleConfirm,
-        message: t('action.discardChanges.confirmDialog.confirm-discard-changes'),
+        message: t('action.discard-changes.confirmDialog.confirm-discard-changes'),
```
rexxars, line 57:
```suggestion
        message: t('action.discard-changes.confirm-dialog.confirm-discard-changes'),
```

packages/sanity/src/desk/documentActions/UnpublishAction.tsx
```diff
@@ -13,8 +13,8 @@ import {
 } from 'sanity'
 
 const DISABLED_REASON_KEY = {
-  NOT_PUBLISHED: 'action.unpublish.disabled.notPublished',
-  NOT_READY: 'action.unpublish.disabled.notReady',
+  NOT_PUBLISHED: 'action.unpublish.disabled.not-published',
+  NOT_READY: 'action.unpublish.disabled.not-ready',
   LIVE_EDIT_ENABLED: 'action.unpublish.disabled.liveEditEnabled',
```
rexxars, line 18:
```suggestion
  LIVE_EDIT_ENABLED: 'action.unpublish.disabled.live-edit-enabled',
```

rexxars (changes requested):
I spy a few more camelcased resource IDs (the locale bundle needs to be updated to match, and ideally also take a look at `dev/test-studio/plugins/locale-no-nb/bundles/desk.ts` and see if you can align those as well)

jtpetty:
@rexxars - Thanks for finding the ones I missed. I think I got those now. I look at all the remaining calls to `t()` and I only see `timeAgo` as camel case. Are you fixing that in your PR?

bjoerge:
Sorry for not catching those. I was under the impression that these would trigger a compile error if they didn't match valid resource ids, but see now that that only works if you define them as a record of resource keys at usage site. Would be nice if we could find an easy way of enforcing that. Not trying to dodge blame, but this almost feels like a too easy mistake to do 😅 (and something that should be possible to catch statically?)

jtpetty:
> Sorry for not catching those. I was under the impression that these would trigger a compile error if they didn't match valid resource ids, but see now that that only works if you define them as a record of resource keys at usage site. Would be nice if we could find an easy way of enforcing that. Not trying to dodge blame, but this almost feels like a too easy mistake to do 😅 (and something that should be possible to catch statically?)

I completely agree with you. There should be some way for this to trigger an error somewhere.

PR #5002, opened 2023-10-17
fix(login): fix issue where logo is not shown on custom auth providers

binoy14:
### Description

Fixes an issue where custom logo would not show up on custom auth providers unless the name of the provider was `custom` and even in that case it would throw an error since the provider prop was not passed to the component.

Screenshot 2023-10-17 at 11 35 56 AM

### What to review

Change makes sense

### Notes for release

- Fixes an issue where logo is not shown on custom auth providers


bjoerge (approved):
Nice one!

skogsmaskin (approved):
Nice!

PR #5004, opened 2023-10-18
fix(comments): handle text overflow in editable surfaces, optically align field titles

robinpyon:
### Description

This PR adds a max-height (20% of current viewport height) to comment inputs to better handle long text / overflow scenarios. 

It also optically aligns thread titles with the document inspector title.

### What to review

Try create comments with massive amounts of text – comment footer actions should still be accessible in these scenarios and/or when on smaller viewports (at both the field and inspector level).


hermanwikner (approved):
Nice, thank you @robinpyon! 🙏

PR #5006, opened 2023-10-18
chore(e2e): Run e2e tests in shards

binoy14:
### Description

This PR enables e2e tests to run in shards as well as takes the report from each shard and merges into one final html report. See it [here](https://github.com/sanity-io/sanity/actions/runs/6563629567?pr=5006)

### Shard based on browsers so it will run

- chromium (1, 2)
- chromium (2,2)
- firefox (1, 2)
- firefox(2, 2)

This should reduce the CPU constraint per shard 

Fixes SDX-787

### Notes for release

- N/A

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,10 +15,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: [chromium, webkit, firefox]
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM - solid work!

rexxars (approved):
Nice - great work!

PR #5009, opened 2023-10-18
test(cli): reject returned promise from exec() upon non-zero exit code

bjoerge:
### Description

I had a hard time debugging an issue with failing CLI tests today that turned out to be caused by the `exec()` helper not taking the error route when executed process failed. This again caused two separate issues that made it hard to figure out why the certain CLI tests were failing:

1) exec of symlinkDependencies during cli tests failed silently after the file was renamed from `.js` to `.ts`
2) errors during `npm install` in the "fixture"-studios failed silently and was never surfaced in test output/logs. Instead, multiple cli tests were failing at later steps due to dependencies not having been installed, which disguised the root cause.

This PR makes sure that if the `exec()`'ed process fails with a non-zero exit code, the promise gets rejected and awaiting it will trigger an unhandled exception. 

### What to review
- This includes the changes from #5008, which don't need to be reviewed here.
- Make sure the changes make sense.
- Also included a small [modification](https://github.com/sanity-io/sanity/compare/fix/cli-tests-throw-on-exec-errors?expand=1#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R82) to the unit test workflow that will silence console.log statements - there's just too much noise currently (see [here](https://github.com/sanity-io/sanity/actions/runs/6560556679/job/17818440186#step:10:1676) for an example).
- Also includes a workflow change (0d45c81b6fe6aea1ca927543bf01a665d782c4be) that triggers unit tests for all branches - will remove before merging.

### Notes for release

N/A – Internal


rexxars (approved):
Nice! Thank you so much <3

bjoerge:
Merged via [a0f2c27](https://github.com/sanity-io/sanity/commit/a0f2c2713ef719a73ec4a4bd428513a588f146bc)

PR #5010, opened 2023-10-18
feat(i18n): add environment variable flag for debugging i18n

rexxars:
### Description

Adds a flag based on the `SANITY_STUDIO_DEBUG_I18N` environment variable, which enables the debug mode in `i18next`, as well as adding "markers" (◤ this is translated ◢) around anything translated by the `t` function. I think this is what you meant, @jtpetty ?

We can (probably) remove this before this goes live, but helpful for spotting parts that are not yet localized.

Example:

image

The user menu options are translated, but the locale names are not (they shouldn't be, I guess).
You can also see the tooltip on the configuration issues button is not localized (this one we said we would wait with, but proves the concept of spotting unlocalized areas)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "deploy:movies": "yarn build && cd examples/movies-studio && sanity deploy",
     "deploy:test": "yarn build && cd dev/test-studio && sanity deploy",
     "dev": "yarn start",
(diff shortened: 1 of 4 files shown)


jtpetty (approved):
LGTM - thanks for doing this. I think this is a valuable tool for devs doing plugins and other customization.

PR #5011, opened 2023-10-18
fix(i18n): add missing calendar labels

rexxars:
### Description

Found a few inconsistencies and missing translations for the calendar:

- "Go to next month" was not localized (_previous_ was)
- "Select minute" was not localized (_hour_ was)
- As these labels are used for aria-labels, I think they should read "Go to next month" instead of just "Next month", but I may be wrong. Please advice (or commit a change directly to this branch). Similarily, I am wondering if these perhaps should be suffixed with a `-aria-label` for consistency - but as they are under an `action` prefix I wasn't sure. Opinions welcome.

### DIFF
--- a/dev/test-studio/plugins/locale-no-nb/bundles/studio.ts
+++ b/dev/test-studio/plugins/locale-no-nb/bundles/studio.ts
@@ -114,16 +114,20 @@ const studioResources: Record<StudioLocaleResourceKeys, string> = {
 
   /** --- DateTime (and Date) Input --- */
 
+  /** Action message for navigating to next month */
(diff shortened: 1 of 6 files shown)


bjoerge (commented):
Thanks for catching these!

I agree that they should say "*Go to* next mont", etc..

Don't really have any well formed opinions about whether the keys should be  suffixed with `-aria-label` or not, but I would I assume it can give useful context to translators, so +1 from me I guess. We should strive for consistency, so we might need to go over and change some already existing keys if we agree we want to establish that as a pattern.

rexxars:
My hunch is to go with `-aria-label` to make it clearer. Will change to that in a separate PR.

PR #5012, opened 2023-10-18
fix: ensure created comments create URLs with inspect and comment params

robinpyon:
### Description

This PR ensures created comments store URLs based on current `window.location.origin` plus the current path based off current pane params.

This is similar to how the current 'copy link to comment' works - the only difference is that we force `inspect` and `comment` params when comments are created directly on fields.

There is a minor point to note which is when links are created on _active panes that are not the most-RHS desk pane_. It's outlined in comments but worth mentioning. E.g. if you have 5 panes open and the second one is active and you create a comment on it, it'll create a link containing the _full_ pane layout. 

We can probably improve on this in future and look to filter out non-active panes (beyond the current active one). The same logic could potentially be applied to 'copy link to comment' too – but the current implementation may be good enough for now.

### What to review

Create a comment (mention / reply) and wait for the notification email. Clicking the "View in Studio" link should take you directly to the comment with the inspector open.

### DIFF
--- a/packages/sanity/src/desk/comments/src/hooks/useCommentOperations.ts
+++ b/packages/sanity/src/desk/comments/src/hooks/useCommentOperations.ts
@@ -9,6 +9,8 @@ import {
   CommentPostPayload,
 } from '../types'
 import {useNotificationTarget} from './useNotificationTarget'
+import {useWorkspace} from 'sanity'
+import {useRouterState} from 'sanity/router'
 
 export interface CommentOperationsHookValue {
   operation: CommentOperations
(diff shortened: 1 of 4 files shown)


packages/sanity/src/desk/comments/src/hooks/useCommentOperations.ts
```diff
@@ -52,18 +54,28 @@ export function useCommentOperations(
...
+    useCallback(
+      (routerState) => (typeof routerState.tool === 'string' ? routerState.tool : undefined),
+      [],
+    ),
+  )
+  const {tools} = useWorkspace()
+  const activeTool = useMemo(
+    () => tools.find((tool) => tool.name === activeToolName),
+    [activeToolName, tools],
+  )
```
robinpyon, line 67: This was moved out of `useNotificationTarget` since the tool name isn't part of the notification object

packages/sanity/src/desk/comments/src/types.ts
```diff
@@ -69,7 +69,11 @@ export interface CommentPath {
   field: string
 }
 
-interface CommentContext {
+/**
+ * @beta
+ * @hidden
+ */
+export interface CommentContext {
```
robinpyon, line 76: Exporting this since I wanted to avoid something like `NonNullable<CommentDocument['context']>['notification']` as a return type for `getNotificationValue` – but maybe that's OK

packages/sanity/src/desk/comments/src/hooks/useNotificationTarget.ts
```diff
@@ -45,15 +43,25 @@ export function useNotificationTarget(
...
-  const deskToolSegment = currentUrl.pathname.split('/').slice(2, 3).join('')
-  currentUrl.pathname = `${basePath}/${deskToolSegment}/__edit__${documentId},type=${documentType}`
-  const notificationUrl = currentUrl.toString()
+  const handleGetNotificationValue = useCallback(
+    ({commentId}: {commentId: string}) => {
+      // Generate a path based on the current pane params.
+      // We force a value for `inspect` to ensure that this is included in URLs when comments
+      // are created outside of the inspector context (i.e. directly on the field)
+      // @todo: consider filtering pane router params and culling all non-active RHS panes prior to generating this link
+      const path = createPathWithParams({
```
hermanwikner, line 52: Nice that this function was useful here as well! 👍 

We don't do the "force thing" when copying the comment URL, we should maybe add it there as well.

hermanwikner (approved):
Looks good, thanks @robinpyon!

PR #5014, opened 2023-10-19
feat: provide formatted duration hook (`useFormattedDuration`)

rexxars:
### Description

Provides a new hook, `useFormattedDuration`, which uses the localization primitives we have in place and formats a duration (eg number of milliseconds) into something that can be shown to the user. It also gives an ISO8601 duration string, which can be used in a `<time>` tag or similar, for better machine readability.


bjoerge (approved):
Solid and thorough, great work!

PR #5016, opened 2023-10-19
feat(facelift): create StudioUI tooltip with delay and replace

ninaandal:
### Description

Starting the work of Studio UI - this PR adds the StudioUI tooltip with an globally added `delay`, but passes on the other `props`. 
Since the StudioUI tooltuip now have a default `padding`, this PR also removes the paddings added to the tooltips for consistency. 

### What to review
Since this PR changes all tooltips in the studio, all of them should be properly tested to make sure they are looking as they should

Adding some videos to show where many of the tooltips can be found: 

https://github.com/sanity-io/sanity/assets/44635000/183c760e-b5b4-4ffc-9287-8c56ea92e445

https://github.com/sanity-io/sanity/assets/44635000/91a7d783-baec-468d-8ca6-c9b29361abec

https://github.com/sanity-io/sanity/assets/44635000/e4e0e01b-78d0-4e18-b1c8-6c025671d829

Screenshot 2023-10-24 at 20 52 57

https://github.com/sanity-io/sanity/assets/44635000/a303c534-d181-4a54-a0a9-c06b6477901b

Screenshot 2023-10-24 at 21 40 03


packages/sanity/src/ui/tooltip/Tooltip.tsx
```diff
@@ -21,15 +12,6 @@ export interface TooltipProps
...
-    <UITooltip
-      content={
-        <Text size={1} weight="medium">
-          {text}
-        </Text>
-      }
-      {...rest}
-    />
-  )
+export const Tooltip = ({...rest}: UITooltipProps) => {
```
pedrobonamin, line 15: Great job @ninaandal thanks!
I think you will need to update here the props to `ComponentProps<typeof UITooltip>` so you can support the cases that are failing due to ts checks. 

What that does is expose all the props the original component receives, so you can add things like refs

packages/sanity/src/ui/tooltip/Tooltip.tsx
```diff
@@ -21,15 +14,22 @@ export interface TooltipProps
...
+export const Tooltip = forwardRef(function Tooltip(
+  props: TooltipProps,
+  ref: React.ForwardedRef<HTMLDivElement>,
+) {
+  const {content, ...rest} = props
+
+  if (typeof content === 'string') {
+    return (
+      <UITooltip
+        content={<Text size={1}>{content}</Text>}
```
robinpyon, line 26:
```suggestion
        content={<Text size={1} weight="medium">{content}</Text>}
```

**suggestion:** minor one, these need to be medium!

robinpyon (commented):
Thanks @ninaandal – interesting to see exactly _how many_ instances there are (and how much variance there is layout wise).

Upon using it, maybe we'll want to tweak that delay slightly (but this change makes it much easier to do so globally)

Left a minor comment on tooltip styling but otherwise LGTM!

robinpyon (approved):

PR #5019, opened 2023-10-19
feat(i18n): add localization for UnknownPaneType

jtpetty:
### Description

Adds localized strings for UnknownPaneType


packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -165,6 +165,16 @@ const deskLocaleStrings = {
 
   /** --- "PRODUCTION PREVIEW", eg link to content --- */
   'production-preview.menu-item.title': 'Open preview',
+
+  /** -- UNKNOWN PANE TYPE */
+
+  /** The text to display when type is missing */
+  'panes.unknown-pane-type.missing-type.text':
+    'Structure item is missing required <code>type</code> property.',
```
rexxars, line 173: HTML is escaped when using `t`, so this will be rendered as plain text. You'll need to use the `Translate` component like this: https://github.com/sanity-io/sanity/blob/feat/i18n/packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/Grid/IncompatibleItemType.tsx#L53-L58

@bjoerge I am wondering if we should perhaps allow simple HTML tags through `<Translate>` without having to configure `components` explicitly - either automatically, or by passing an `allowedTags` array or something. Thoughts?

packages/sanity/src/desk/panes/document/timeline/__workshop__/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+    "rules": {
+      "i18next/no-literal-string": "off"
+    }
+}
```
rexxars, line 5: Nitpick, but preferably add newlines at the end of files (the [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) VS Code extension automatically does this for you based on our config, fwiw).

rexxars (changes requested):

packages/sanity/src/desk/i18n/resources.ts
bjoerge: > I am wondering if we should perhaps allow simple HTML tags through <Translate> without having to configure components explicitly - either automatically, or by passing an allowedTags array or something. Thoughts?

Makes a ton of sense, +1 from me.

bjoerge:
(rebased this branch)

packages/sanity/src/desk/i18n/resources.ts
jtpetty: I agree! I'll hold off making any changes until that is settled. I may cherry pick the `.eslintrc` changes in a separate PR

jtpetty:
@rexxars - this should be ready for review again

ricokahler (approved):

rexxars (approved):

PR #5020, opened 2023-10-19
test(cli): symlinkDependencies before running `npm install`

bjoerge:
### Description

#4917 adds styled-components@6.1 as default to newly created Sanity Studios. The CLI tests are currently installing dependencies first, before symlinking in the local dependencies, but this now fails because the current `@sanity/cli` from npm doesn't have `styled-components@6` in it's accepted peer-dependency range. This addresses the issue (and future similar issues) by symlinking the local dependencies before running npm install.

Note: We might want to consider looking into replacing this with installing from local tarballs of the local packages instead (potentially via something like [yalc](https://github.com/wclr/yalc)), to make sure dependencies are installed in a way that is more like a "real-life" npm install, but this seems to work fine for now.

### Notes for release
N/A – internal

### DIFF
--- a/packages/@sanity/cli/test/shared/globalSetup.ts
+++ b/packages/@sanity/cli/test/shared/globalSetup.ts
@@ -79,6 +79,7 @@ function prepareStudios() {
       const destinationPath = path.join(studiosPath, version)
       const customDocStudioPath = path.join(studiosPath, `${version}-custom-document`)
 
+      await mkdir(destinationPath, {recursive: true})
       await copy(`${sourceStudioPath}/**/{*,.*}`, destinationPath, {dereference: true})
 
       if (version === 'v2') {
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/test/shared/globalSetup.ts
```diff
@@ -79,16 +79,18 @@ function prepareStudios() {
       const destinationPath = path.join(studiosPath, version)
       const customDocStudioPath = path.join(studiosPath, `${version}-custom-document`)
 
-      await mkdir(destinationPath, {recursive: true})
```
rexxars, line 82: I assume this is because the `copy()` function below creates it?

rexxars (approved):

packages/@sanity/cli/test/shared/globalSetup.ts
bjoerge: Good question - looks like cpx [creates](https://github.com/mysticatea/cpx/blob/master/lib/utils/copy-file.js#L104C10-L104C10) it, but I reverted this line to make sure we don't depend on undocumented/unspecified behavior.

PR #5043, opened 2023-10-23
feat(facelift): implement Dialog into StudioUI

pedrobonamin:
Code changes ready, awaiting confirmation from design team for not updated dialogs.

### Description

- Implements `<Dialog />` element in studio UI.
- Replaces uses of `@sanity/ui` Dialog for studio UI Dialog.
- Updates delete document dialog styles to match facelift.

### What to review

Usage of Dialog inside the studio. 
Cases found are detailed in the related ticket.


packages/sanity/src/ui/dialog/dialog.tsx
```diff
@@ -0,0 +1,53 @@
+import {Button, Flex, Dialog as UIDialog, DialogProps as UIDialogProps, Box} from '@sanity/ui'
+import React, {ComponentProps, forwardRef} from 'react'
+
+/** @internal */
+export interface DialogProps extends Omit<UIDialogProps, 'footer'> {
```
robinpyon, line 5: To enforce visual consistency, props for Studio UI components should be as restrictive as possible. We should consider selectively _picking_  props rather than omitting, as currently it's possible to make per-instance overrides with custom `cardRadius`, `shadowRadius` etc.

packages/sanity/src/ui/dialog/dialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+              />
+            )}
+            {footer.confirmButton && (
+              <Button mode="default" text="Confirm" tone="critical" {...footer.confirmButton} />
+            )}
+          </Flex>
+        )
+      }
+    >
+      <Box paddingX={3}>{children}</Box>
```
robinpyon, line 50: Similarly, the default padding for children should accommodate all use cases within the studio. Ideally, `children` of the Studio UI dialog shouldn't need to specify any custom padding at all. This custom padding value should ensure that child text are horizontally aligned with header text.

robinpyon: Upon further reflection, there are cases where it makes sense for dialog `children` to be completely flush within the body of the dialog (e.g. document inspect, default asset source – cases where there's sticky elements and/or a need for x-axis overflow).

Perhaps Studio UI tooltip `children` could always default to specific padding to cover the vast majority of cases (confirming / warning dialogs, editing document forms) whilst we provide an escape hatch for users to bypass this? 

Ideally, studio developers shouldn't even need to think about the proper padding required for dialog `children`, barring unique cases

pedrobonamin: Hey @robinpyon thanks for this review.
The `paddingX={3}` in the body is coming from the design file.
If you check that, all the dialogs displayed have 2 different paddings in the body.
The first one encapsulates the body in a padding 3, to make it match with the header and footer padding. 

Screenshot 2023-10-30 at 16 14 31

Then, depending on the content it has more or less padding, but always keeping the initially defined padddingX=3
E.g 
- The ConfirmDeleteDialog has a `paddingX=3` here.
Screenshot 2023-10-30 at 16 14 45

-  Form dialogs use a `paddingX=5` 
Screenshot 2023-10-30 at 16 21 26

packages/sanity/src/ui/dialog/dialog.tsx
pedrobonamin: Updated! Thanks!

packages/sanity/src/desk/components/confirmDeleteDialog/ConfirmDeleteDialog.tsx
```diff
@@ -73,27 +74,21 @@ export function ConfirmDeleteDialog({
 
   return (
     <Dialog
-      width={1}
+      width={0}
```
ninaandal, line 77: Is there a reason you are setting this to be `0`? I think it looks better with `width={1}`
Screenshot 2023-11-01 at 08 14 00
Screenshot 2023-11-01 at 08 13 52

robinpyon (approved):
Thanks @pedrobonamin!

PR #5044, opened 2023-10-23
chore(e2e): create a test studio for e2e tests

binoy14:
### Description

This PR creates a new studio instance that is inheriting the schema from the test studio but works in isolation with the idea being this studio is solely used for e2e testing. This studio also takes the projectId and dataset as env variables with the idea being that we would create a new dataset per PR which is removed after the PR is merged (future PR for this) this makes sure that the tests are run in true isolation and not affected by any side effects during development or testing. 

Other changes in this PR is that now the e2e tests in CI run on the production build of the studio which removes the CI flakyness out of the equation trying to serve a page. Since it would be the production build it should also match very closely to what is deployed.

### What to review

- Imports in studio-e2e-testing makes sense. Specifically [here](https://github.com/sanity-io/sanity/pull/5044/files#diff-e87db3bd660336e6760805705bd877f08b1765527795e45f7d1dcfa85234ccb6R9-R17)

### Notes for release

- N/A


bjoerge (approved):
Solid! LGTM

PR #5048, opened 2023-10-23
fix(mutator): match deep set/setIfMissing behavior

rexxars:
### Description

There seems to be a few mismatches between `@sanity/mutator` and Content Lake:
 
- In Content Lake, a `set`/`setIfMissing`-patch targeting a deep key with missing parents will recursively create an object structure for it
- In `@sanity/mutator`, it will stop once it encounters a key that is not present in the document.

This PR attempts to align the behavior with what Content Lake does. Please review carefully - my knowledge of the mutator is limited and there may be (edge) cases I have not considered.

Fixes SDX-772.

Edit: I see that arrays also gets overwritten in a similar vein as primitives, but I am not quite sure yet how to resolve it.

### What to review

- Mutations work as expected

### Notes for release

- Fixed an issue where certain remote patches would not cleanly apply in the studio, requiring a reload of the document in order to see the change

### DIFF
--- a/packages/@sanity/mutator/src/patch/Patcher.ts
+++ b/packages/@sanity/mutator/src/patch/Patcher.ts
@@ -3,6 +3,8 @@ import {Matcher} from '../jsonpath'
 import {parsePatch} from './parse'
 import {ImmutableAccessor} from './ImmutableAccessor'
 import {PatchTypes, SingleDocumentPatch} from './types'
+import {SetPatch} from './SetPatch'
+import {SetIfMissingPatch} from './SetIfMissingPatch'
 
 export interface Patch {
   id: string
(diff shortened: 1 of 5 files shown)


bjoerge (approved):
Wasn't aware of this behavior with nested set patches. Feels like it _should_ trigger an error, but for now I agree it's more important to align with content lake behavior. Thanks!

rexxars:
> Wasn't aware of this behavior with nested set patches. Feels like it _should_ trigger an error, but for now I agree it's more important to align with content lake behavior. Thanks!

I agree, but I'd say it is unlikely to change in Content Lake given this behavior has always been this way, and maintaining backwards compatible patches is the priority for now.

PR #5057, opened 2023-10-24
fix(core/form/inputs): fix issue with tabbing to popover toolbar buttons in PT-input

skogsmaskin:
### Description

There is currently an issue with reaching the annotation and inline object edit toolbar buttons with the keyboard leaving annotation editing impossible with the keyboard.

This will make sure that when these toolbars shows, it will be possible to press `Tab` in order to get to the buttons with the keyboard.

https://github.com/user-attachments/assets/6a928a83-980b-47ba-83b3-ad1e49cb7d74

Also added component tests for this.

### What to review

That you are able to tab into the buttons when the toolbar is showing.

### Notes for release
* Fixed an issue with keyboard navigation to edit buttons for annotation and inline objects in the Portable Text Input.

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Annotations.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Annotations.spec.tsx
@@ -48,8 +48,25 @@ test.describe('Portable Text Input', () => {
       // Expect the editor to have focus after closing the popover
       await expect($pte).toBeFocused()
 
+      const $toolbarPopover = page.getByTestId('annotation-toolbar-popover')
+
       // Assertion: the annotation toolbar popover should be visible
       await expect(page.getByTestId('annotation-toolbar-popover')).toBeVisible()
+      await expect($toolbarPopover).toBeVisible()
+
+      // Assertion: tab works to get to the toolbar popover buttons
+      await page.keyboard.press('Tab')
+      await expect(page.getByTestId('edit-annotation-button')).toBeFocused()
+      await page.keyboard.press('Tab')
+      await expect(page.getByTestId('remove-annotation-button')).toBeFocused()
+      await page.keyboard.press('Escape')
+      await expect($pte).toBeFocused()
+      await expect($toolbarPopover).toBeVisible()
+      await page.waitForTimeout(100)
+      await page.keyboard.press('Escape')
+      await page.waitForTimeout(100)
+      // Assertion: escape closes the toolbar popover
+      await expect($toolbarPopover).not.toBeVisible()
     })
 
     test('Can create, and then open the existing annotation again for editing', async ({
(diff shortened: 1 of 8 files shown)


robinpyon (commented):
Thanks @skogsmaskin, this is a huge improvement.

Just had a look and noticed a few things:

1. When pressing TAB for the first time over an annotation, it will try focus the next element briefly before then focusing buttons within the popover.

If you look at this video and pay attention to the 'No items' box, you'll see what I mean:

https://github.com/sanity-io/sanity/assets/209129/276ab455-7f18-4fac-9317-c3fcfdec4bc4

I'm very certain this is the case due to `buttonRef.current.focus()` being wrapped within a `setTimeout`. This does seem like a bit of a smell!

One possible alternative is that rather than focusing the button directly, we create an invisible anchor element (with `tabIndex=-1`), and then on TAB:
- initially focus the anchor element. Since it has `tabIndex=-1`, it will be ignored in sequential keyboard navigation
- the browser will then 'naturally' tab to the next button in the list

Here's an example video with that approach:

https://github.com/sanity-io/sanity/assets/209129/3a55d8d7-992d-4228-b4de-44b94619e2cd

2. This popover doesn't trap focus – making it very easy to accidentally tab or shift-tab away completely out of the form. We should probably trap focus here.

3. Focus isn't restored when closing the popover – ideally, closing the popover with ESC returns you exactly where you were.

We could probably punt on 2 + 3, but should capture these if we do.

RitaDias:
@skogsmaskin I'm assuming that this PR can be closed? :eyes:

skogsmaskin:
This one is revived from the dead! 🧟 

I have addressed all the feedback from @robinpyon, and there are also some other keyboard improvements like focus locking inside the popover modal from https://github.com/sanity-io/sanity/pull/8016 which this branch is now branched out from.

Here's a video showing the possible keyboard navigation for editing links and inline objects:


https://github.com/user-attachments/assets/f5aa4aa5-3cfa-4db0-8bba-65d4c0fc18ff

juice49:
It looks like the issue Robin raised has been resolved now, but I noticed this change causes focus to get trapped in the Portable Text Editor.

In the following video, I tab into a PTE field and keep pressing tab. Currently, the user would be able to exit the PTE field by tabbing, but in this branch focus is trapped. Is that intentional?

https://github.com/user-attachments/assets/81f23575-8c4d-499e-ba74-1a974b6fad92

skogsmaskin:
> It looks like the issue Robin raised has been resolved now, but I noticed this change causes focus to get trapped in the Portable Text Editor.
> 
> In the following video, I tab into a PTE field and keep pressing tab. Currently, the user would be able to exit the PTE field by tabbing, but in this branch focus is trapped. Is that intentional?
> 
>  Screen.Recording.2024-12-16.at.09.46.08.mov

It is intentional from my side yes, so that you get back to editor again (which is probably what you want in 99% of cases). Closing the popover (ESC) will break out from the focus trap and let you move tab like normal.

juice49:
Okay, thanks for explaining! I just wanted to make sure it was on purposes. In that case, this looks good to me!

juice49 (approved):

PR #5058, opened 2023-10-24
fix(portable-text-editor): allow user event handlers on Editable component

skogsmaskin:
### Description

This will allow the consumer to specify their own `onKeyDown`, `onFocus`, and `onBlur` event handlers on the `Editable` component of the PTE in addition to the internal ones.

Also fixed a prop typing (`onBeforeInput`) that was problematic to use.

### What to review

That the code looks alright.

### Notes for release
N/A


bjoerge (approved):
LGTM

PR #5059, opened 2023-10-24
doc(form): add doc and example for components extention of block type

skogsmaskin:
### Description

This will add some TS-doc with examples to the block component extension.

### What to review

That it's correctly written and makes sense.

### Notes for release
N/A


bjoerge (approved):
Great!

PR #5060, opened 2023-10-24
fix(portable-text-editor): fix issues for .delete fn and add tests

skogsmaskin:
### Description

This will fix bugs related to deleting content in the PTE through the `PortableTextEditor.delete` function.

* Test that the range produced from the `PortableTextEditor.delete` method is valid before deleting anything. There was an issue where if you gave it an invalid path, the resulting path for the produced range would be `[]` which would then delete content not targeted for deletion.
* If you deleted block by block with this method until there were no blocks left, it would not properly produce a new placeholder block causing issues with inserting new content.
* While writing tests for this I discovered some additional bugs with the `block` mode when providing a selection containing multiple blocks.

### What to review
Code looks good

### Notes for release
* N/A


bjoerge (approved):
LGTM

PR #5061, opened 2023-10-24
fix(form/inputs): fix invalid path used with onRemove for TextBlock

skogsmaskin:
### Description

This will fix an issue with the wrong path being sent to the PTE inside the block component's `onRemove` function. It should be the relative path, not the form builder path.

Also,  `.focus` can now be called without a `setTimeout` here.

### What to review
I have already tested this with a custom component, but if you want to test it yourself, you can use this schema:

```
    {
      name: 'foo',
      title: 'Foo',
      type: 'array',
      of: [
        defineArrayMember({
          type: 'block',
          components: {
            block: (blockProps) => {
              return (
                <Flex>
                  <Box flex={1}>{blockProps.renderDefault(blockProps)}</Box>
                  <Box contentEditable={false} style={{userSelect: 'none'}}>
                    <Button
                      icon={TrashIcon}
                      onClick={(event) => {
                        event.preventDefault()
                        blockProps.onRemove()
                      }}
                    />
                  </Box>
                </Flex>
              )
            },
          },
        }),
      ],
    },
```

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
@@ -130,11 +130,14 @@ export function TextBlock(props: TextBlockProps) {
   }, [onItemOpen, memberItem])
 
   const onRemove = useCallback(() => {
-    const sel: EditorSelection = {focus: {path, offset: 0}, anchor: {path, offset: 0}}
+    const point = {path: path.slice(-1), offset: 0}
+    const sel: EditorSelection = {
+      focus: point,
+      anchor: point,
+    }
     PortableTextEditor.delete(editor, sel, {mode: 'blocks'})
-    // Focus will not stick unless this is done through a timeout when deleted through clicking the menu button.
-    setTimeout(() => PortableTextEditor.focus(editor))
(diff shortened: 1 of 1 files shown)


binoy14 (commented):
Not sure if it is related to this but I was able to create a weird case where PTE would break. 

### Steps to reproduce

1. Use the custom component schema in PR descript
2. Add couple lines and delete them
3. Once the last item is deleted click on the PTE input and start typing

### Expected
Works normally without error

### Actual
Throws bunch of errors

https://github.com/sanity-io/sanity/assets/6476108/bf0cba39-77ec-4d78-a153-f2c1dbe324b2

skogsmaskin:
> Not sure if it is related to this but I was able to create a weird case where PTE would break.

Yeah,[ my other (previous PR)](https://github.com/sanity-io/sanity/pull/5060) should have been merged first.

skogsmaskin:
@binoy14 - I've merged the other PR now, and rebased this branch on `next`. Could you try again?

binoy14 (approved):
The test is working for me now but I am not familiar with this part of the code maybe @bjoerge could have a peek as well?

PR #5062, opened 2023-10-24
fix(mutator): apply `setIfMissing` prior to `set`

rexxars:
### Description

Fixes an issue where the order of operations of set/setIfMissing is reversed from what Content Lake does. By applying `setIfMissing` first, it ensures we can "prepare" structures that a deep `set` operation might use, for instance.

### What to review

- Mutations still work and apply as before

### Notes for release

- Fixes an issue where certain patches would apply differently in the studio and in Content Lake, leading to the studio UI not reflecting new values for certain fields


bjoerge (approved):
This is great, thank you <3

PR #5064, opened 2023-10-24
chore(i18n): add no-literal-string lint exceptions for __workshop__ dirs

jtpetty:
### Description

Since we are not translating files in `__workshop__` directories, this PR disables the `no-literal-string` rule for these directories.


ricokahler (approved):
Not a blocking comment but there is an [`overrides`](https://eslint.org/docs/latest/use/configure/configuration-files#how-do-overrides-work) configuration setting for eslint that we could use to ignore that rule for certain files when ran at a top-level. Would that be better?

PR #5065, opened 2023-10-24
feat(core): use platform-aware hotkeys

rexxars:
### Description

Introduces a new `<Hotkeys />` component in the `sanity` module which automatically changes `alt` to `option` and vice versa, but otherwise just reuses what's in `@sanity/ui`. For more context, see https://github.com/sanity-io/ui/pull/1168

### What to review

- Shows correct hotkey for publishing and similar on apple devices

### Notes for release

- Hotkeys are now device aware, showing `Option` instead of `Alt` on Apple devices


packages/sanity/src/core/components/Hotkeys.tsx
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @public
+ */
+export type HotkeysProps = UIHotkeysProps & {
+  /**
+   * Whether to make the keys platform-aware (eg `alt` to `option` on Apple devices).
+   *
+   * @defaultValue true
+   */
+  makePlatformAware?: boolean
```
bjoerge, line 15: out of curiosity and lack of imagination: in what cases would you want to make it not platform aware?

bjoerge (commented):
Very nice, thank you! <3

bjoerge (approved):

binoy14 (approved):
LGTM

packages/sanity/src/core/components/Hotkeys.tsx
rexxars: Potentially if showing both alternatives, eg something like:
```
Windows: Ctrl+Alt+I
Mac: Ctrl+Opt+I
```

PR #5067, opened 2023-10-25
fix(desk): edit logic for published- and edit status message in document list 

ninaandal:
### Description
The logic for published and edits message in the document list was wrong and would be incorrect in certain instances. 
This was also the case for the `PublishAction` tooltip. 

Before: 
Screenshot 2023-10-25 at 11 14 59
Screenshot 2023-10-25 at 11 14 54
Screenshot 2023-10-27 at 13 38 59

Instead of using `<TimeAgo />`, we can use `useTimeAgo()` and show label with abbreviations and suffix. 

### What to review
The document list published and edits status. 

### Notes for release
Fixed published and edits message in document list


robinpyon:
Great find @ninaandal 

Since this issue lies with the `<TimeAgo>` component – perhaps it makes sense to find other places that's used (I think there are just 1-2), apply this same fix and then remove that component altogether?

bjoerge (approved):
Nice, good catch & thanks for fixing!

ninaandal:
> Great find @ninaandal
> 
> Since this issue lies with the `<TimeAgo>` component – perhaps it makes sense to find other places that's used (I think there are just 1-2), apply this same fix and then remove that component altogether?

There are several `<TimeAgo />` components and I think it makes sense to keep as is for the preview components (reference and CDR reference) as this shows how long ago it was (like `almost two years ago` instead of the date it was published). But I'll remove the one that is used for the `PublishAction` as this can be replaced with `useTimeAgo()`

robinpyon (approved):
Thanks @ninaandal!

> I think it makes sense to keep as is for the preview components (reference and CDR reference) as this shows how long ago it was (like almost two years ago instead of the date it was published).

(nitpicking and not blocking) I would opt for consistency here – as rolling over these publish icons in different contexts currently yields different values (document list previews say "published yesterday", whilst reference inputs say "published 1 day ago"). Whilst we may know they're different components, it may not be apparent to users. 

https://github.com/sanity-io/sanity/assets/209129/180dde79-2106-4bbe-a526-fb967f4595a6

ninaandal:
> Thanks @ninaandal!
> 
> > I think it makes sense to keep as is for the preview components (reference and CDR reference) as this shows how long ago it was (like almost two years ago instead of the date it was published).
> 
> (nitpicking and not blocking) I would opt for consistency here – as rolling over these publish icons in different contexts currently yields different values (document list previews say "published yesterday", whilst reference inputs say "published 1 day ago"). Whilst we may know they're different components, it may not be apparent to users.
> 
>  Screenflick.Movie.15.mp4

We should make a decision of which format we would like then, as the reference preview one uses `date-fns/formatDistanceToNow` to format, while the document list items does not.
Do we want it to have the format: `Published about 1 years ago` for example, or just show the date when it was published (see the two screenshots)? 
Maybe @kaylasanity wants to chime in here? 

Screenshot 2023-10-30 at 09 25 32
Screenshot 2023-10-30 at 09 25 24

ninaandal:
Had a chat with @kaylasanity and decided to go for the timestamp format ✨

packages/sanity/src/core/form/inputs/ReferenceInput/ReferencePreview.tsx
```diff
@@ -104,13 +110,11 @@ export function ReferencePreview(props: {
...
-                          Edited <TimeAgo time={preview.draft._updatedAt} />
-                        </>
-                      ) : (
-                        <>No unpublished edits</>
-                      )}
+                      <Text size={1}>
+                        {preview.draft?._updatedAt
+                          ? `Edited ${timeSinceEdited}`
+                          : 'No unpublished edits'}
+                      </Text>
```
robinpyon, line 117:
```suggestion
  {preview.draft?._updatedAt
    ? `Edited ${timeSinceEdited}`
    : 'No unpublished edits'}
```

No need to double wrap `<Text>`, as doing so messages with padding here:

image

PR #5068, opened 2023-10-25
refactor(comments): refactor how input and focus is handled

skogsmaskin:
### Description

* Enter will submit the comment
* Input field is reset after submission or discard
* Setting focus is handled through the same function everywhere.
* Input key events are propagated to the parent components

Also made a performance optimization making sure the thread isn't re-rendered whenever there is a new value in the `CommenListItem` input.

Also added a test for comment submission.


hermanwikner (approved):
Great stuff @skogsmaskin 🙏

PR #5076, opened 2023-10-26
chore(dev): remove the CRA starter dev studio

bjoerge:
### Description
I don't think this provides us with any practical value, and it's causing dependency issues with jest since it depends on react-scripts which doesn't seem actively maintained anymore (it's last publish was April last year).

### What to review
- Any problems with doing this?

### Notes for release

n/a - chore


binoy14 (approved):
👋

PR #5079, opened 2023-10-26
fix(schema/types): fix wrong key being used for block schema def.

skogsmaskin:
### Description

This will fix some issues with the block type schema, where it is missing definition for level, and the list definition is using the wrong key as illustrated by editing a PT-field with the regular array input:

image

The correct key here is `listItem`. 

After:

image

### What to review
This change is trivial to do (renaming the schema field). I think it is, as all tests are passing and I can't find any other references to this field other than those I have already changed.

### Notes for release
N/A - not a user-facing problem.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/utils/getPortableTextMemberSchemaTypes.ts
+++ b/packages/@sanity/portable-text-editor/src/utils/getPortableTextMemberSchemaTypes.ts
@@ -73,9 +73,9 @@ function resolveEnabledDecorators(spanType: ObjectSchemaType) {
 }
 
 function resolveEnabledListItems(blockType: ObjectSchemaType) {
-  const listField = blockType.fields?.find((btField) => btField.name === 'list')
+  const listField = blockType.fields?.find((btField) => btField.name === 'listItem')
   if (!listField) {
-    throw new Error("A field with name 'list' is not defined in the block type (required).")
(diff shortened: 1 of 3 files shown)


packages/@sanity/portable-text-editor/src/utils/getPortableTextMemberSchemaTypes.ts
```diff
@@ -73,9 +73,9 @@ function resolveEnabledDecorators(spanType: ObjectSchemaType) {
 }
 
 function resolveEnabledListItems(blockType: ObjectSchemaType) {
-  const listField = blockType.fields?.find((btField) => btField.name === 'list')
+  const listField = blockType.fields?.find((btField) => btField.name === 'listItem')
```
bjoerge, line 76: this is strictly speaking a breaking change, right? How likely is it that someone has `list` but not `listItem` and will get this error after upgrading?

skogsmaskin: No, as this isn't really affecting the user schemas. They stay exactly the same. This is more the internal structure of the schema.

bjoerge (approved):

skogsmaskin:
I had to revert the merge to `next` as the tests failed there. I have followed up these problems in https://github.com/sanity-io/sanity/pull/5152

PR #5080, opened 2023-10-26
ci: move PTE e2e tests out in own workflow

skogsmaskin:
### Description

Moved these tests into their own workflow to have them run more in isolation and speed up our unit tests (playwright dep. no longer needed).

### Notes for release
N/A


.github/workflows/e2e-pte.yml
```diff
@@ -0,0 +1,100 @@
...
+          PWTEST_BLOB_REPORT_NAME: ${{ matrix.project }}
+        run: cd packages/@sanity/portable-text-editor && yarn test:e2e --silent --shard=${{ matrix.shardIndex}}/${{ matrix.shardTotal }}
+
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: playwright-report
+          path: blob-report
+          retention-days: 30
+  merge-reports:
```
skogsmaskin, line 75: Is this really needed here?

I copied what was in `test.yml` regarding this.

.github/workflows/test.yml
```diff
@@ -63,27 +63,6 @@ jobs:
...
-        id: cache-playwright-browsers
-        uses: actions/cache@v3
-        with:
-          path: ~/.cache/ms-playwright
-          key: playwright-browsers-${{ env.PLAYWRIGHT_VERSION }}
-
-      - name: Install Playwright Browsers
-        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
-        run: npx playwright install --with-deps
-
```
skogsmaskin, line 86: This can now be removed from here, as none of the unit tests are now using Playwright.

bjoerge (commented):
Sweet! Thanks a lot <3

.github/workflows/e2e-pte.yml
binoy14: Does it generate a playwright report? That is only needed if a blob playwright report is generated and it will merge into one html report. Looking at the run I don't think it's needed and it can be removed

.github/workflows/e2e-pte.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Install Playwright Browsers
+        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
+        run: npx playwright install --with-deps
+
+      - name: Run end-to-end tests
+        env:
+          # Missing in docs but in use
+          # here https://github.com/microsoft/playwright/blob/main/packages/playwright/src/reporters/blob.ts#L108
+          PWTEST_BLOB_REPORT_NAME: ${{ matrix.project }}
```
binoy14, line 66: This env var can also be removed

binoy14 (commented):
Looks good besides a couple comments. Thanks for doing this!

PR #5085, opened 2023-10-27
feat(i18n): adds translation for desk panes

jtpetty:
### Description

Adds translation for files in `packages/sanity/src/desk/panes/`


packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -165,6 +165,136 @@ const deskLocaleStrings = {
...
+
+  /** The loading message for the document not found pane */
+  'panes.document-pane.document-not-found.loading': 'Loading document…',
+
+  /** The title of the document not found pane if the schema is known */
+  'panes.document-pane.document-not-found.title': 'The document was not found',
+
+  /** The text of the document not found pane if the schema is known */
+  'panes.document-pane.document-not-found.text':
+    'The document type is not defined, and a document with the <Code>{{id}}</Code> identifier could not be found.',
```
bjoerge, line 225: did we merge the change that allowed for basic tags? if so can these now be changed to `<code>{{id}}</code>`?

packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
```diff
@@ -51,10 +51,10 @@ export const DocumentOperationResults = memo(function DocumentOperationResults()
         closable: true,
         duration: 30000, // 30s
         status: 'error',
-        title: getOpErrorTitle(event.op),
+        title: t(getOpErrorTitle(event.op), {op: event.op}),
```
bjoerge, line 54: is this a good candidate for [context](https://www.i18next.com/translation-function/context)? that way we could write all these as `panes.document-operation-results.operation.error_delete`, `panes.document-operation-results.operation.error_publish`, etc.

bjoerge (approved):
Nice - just a couple of questions/suggestions, no blockers

packages/sanity/src/desk/i18n/resources.ts
jtpetty: we have not merged that yet. When that goes in, I'll loop back and update these

packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
jtpetty: Oh! Neat! Have we used that anywhere else?

ricokahler (approved):

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -165,6 +165,136 @@ const deskLocaleStrings = {
 
   /** --- "PRODUCTION PREVIEW", eg link to content --- */
   'production-preview.menu-item.title': 'Open preview',
+
+  /** -- DESK PANES -- */
+  /** The tool tipe for the split pane button on the document panel header */
+  'document-panel-header.split-pane-button.tooltip': 'Split pane right',
```
rexxars, line 171: We want to avoid "locations" in keys such as "header" and "footer", as these elements may move around.

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -165,6 +165,136 @@ const deskLocaleStrings = {
...
+
+  /** The title of the sync lock toast on the form view */
+  'document-view.form-view.sync-lock-toast.title': 'Syncing document…',
+
+  /** The description of the sync lock toast on the form view */
+  'document-view.form-view.sync-lock-toast.description':
+    'Please hold tight while the document is synced. This usually happens right after the document has been published, and it should not take more than a few seconds',
+
+  /** The title of the reconnecting toast */
+  'panes.document-pane-provider.reconnecting-toast.title': 'Connection lost. Reconnecting…',
```
rexxars, line 215: Similar here - ideally we would avoid "toast" as the UI element for displaying this message may change.

packages/sanity/src/desk/i18n/resources.ts
```diff
@@ -165,6 +165,136 @@ const deskLocaleStrings = {
...
+    'Successfully performed {{op}} on document',
+
+  /** The text used in the document header title if creating a new item */
+  'header.document-header-title.new.text': 'New {{item}}',
+
+  /** The text used in the document header title if there is an error */
+  'header.document-header-title.error.text': 'Error: {{error}}',
+
+  /** The text used in the document header title if no other title can be determined */
+  'header.document-header-title.untitled.text': 'Untitled',
```
rexxars, line 297: Same here. We'll have to think of a naming regime that works a little better.

packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
```diff
@@ -1,35 +1,34 @@
 import {useToast} from '@sanity/ui'
 import React, {memo, useEffect, useRef} from 'react'
 import {useDocumentPane} from './useDocumentPane'
-import {useDocumentOperationEvent} from 'sanity'
+import {useDocumentOperationEvent, useTranslation} from 'sanity'
 import {usePaneRouter} from '../../components'
+import {deskLocaleNamespace} from '../../i18n'
 
 function getOpErrorTitle(op: string): string {
```
rexxars, line 8: Would be nice to rename this to indicate that it returns a key and not an actual title. Additionally, we could import and use the `DeskLocaleResourceKeys` type as the return value for better type safety

packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
rexxars: Agree with @bjoerge about using context here - see [some preliminary docs here](https://www.notion.so/sanityio/WIP-i18n-documentation-565cab7c1e4a4569a1ae0e3ee19e26c4#7c28e580ed1b4b35a8680d8c63b7a043) - you could also search for `{context: `, eg [this usage](https://github.com/sanity-io/sanity/blob/feat/i18n/packages/sanity/src/core/form/inputs/files/common/PlaceholderText.tsx#L50)

Unfortunate that `event` is not typed, but digging through the types it appears that these are the defined operations:

`'commit' | 'delete' | 'del' | 'publish' | 'patch' | 'discardChanges' | 'unpublish' | 'duplicate' | 'restore'`

packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
```diff
@@ -21,12 +23,24 @@ export function DocumentHeaderTitle(): ReactElement {
   }
 
   if (!documentValue) {
-    return <>New {schemaType?.title || schemaType?.name}</>
+    return (
+      <>
+        {t('header.document-header-title.new.text', {item: schemaType?.title || schemaType?.name})}
```
rexxars, line 28: Would prefer if this had a more specific name
```suggestion
        {t('header.document-header-title.new.text', {schemaType: schemaType?.title || schemaType?.name})}
```

packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
```diff
@@ -21,12 +23,24 @@ export function DocumentHeaderTitle(): ReactElement {
...
+      </>
+    )
   }
 
   if (error) {
-    return <>Error: {error.message}</>
+    return <>{t('header.document-header-title.error.text', {error: error.message})}</>
```
rexxars, line 34: Wondering how to reason about these errors that will most likely be in english. Fine for now, but should we flag these somehow for later review?

rexxars (approved):
Great work!

We need to rework some keys, but would be open to doing it as part of a separate PR so marking this as approved - feel free to merge or address in this PR based on your preference

packages/sanity/src/desk/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
jtpetty: Maybe we can add a `todo` here. Maybe something like `// @todo: i18n review second pass`?

packages/sanity/src/desk/panes/document/DocumentOperationResults.tsx
jtpetty: I switched to using contexts and got rid of these methods altogether

rexxars (approved):

PR #5086, opened 2023-10-27
chore: add eslint-plugin-no-attribute-string-literals

ricokahler:
### Description

This PR overhauls the ESLint config to:
1. Utilize the typescript config by default. Most of our codebase is in TypeScript now and it doesn't seem necessary to have two separate configs when typescript-eslint works for both javascript and typescript files. This simplifies the overrides section.
2. The `overrides` section now includes new patterns for matching and disables localization string rules in those areas.
3. `eslint-plugin-no-attribute-string-literals` was added to target string literals inside of JSX attributes with an extensive config to prevent false positives. We can also configure this plugin to target certain elements instead of how it works now. 

### What to review

Try out the config. Ensure things work correctly.

### Notes for release

N/A


jtpetty:
LGTM. It seems the lint issue is due to a literal string in `sanity/src/core/studio/components/navbar/NavDrawer.tsx`. I think the other failure is a flake.

rexxars (approved):
Would have preferred to see the unification of TS/JS configs as a separate PR since this makes it harder to test, but approving given we can address any issues that may occur after the merge.

PR #5090, opened 2023-10-30
feat(facelift): create menu item StudioUI

pedrobonamin:
### Description

Introduces `StudioUI` menu item, adding defaults to create the menu items.
Updates all the uses of menu items to use the new menu item. 
Removes the need to define complex children when using menu items within the Studio.
tries to discourage the use of `children` prop for the menu item, making developers specifically opt out. 

### What to review

- Menu item uses.
- Publish actions menu items, using the `renderMenuItem` prop to support custom wrappers (e.g. tooltip)
- `Workspace` menu item continues using the old design, confirm if it should be updated or not. 

### Notes for release
Introduces `StudioUI` menu item, adding defaults to create the menu items.


ninaandal (approved):
lgtm 🎉

packages/sanity/src/ui/menuItem/MenuItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+   * Badge text is only supported in `size="large"` menu items.
+   */
+  badgeText?: undefined
+  /**
+   * Avatar is only supported in `size="large"` menu items.
+   */
+  avatar?: undefined
+}
+
+const fontSize = 1
```
robinpyon, line 46:
```suggestion
const FONT_SIZE = 1
```

**suggestion(non-blocking):** consider screaming snake case, which better indicates intent and is a pattern used throughout the studio

packages/sanity/src/ui/menuItem/MenuItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+  avatar?: React.ReactNode
+  /**
+   * Hotkeys are only supported in `size="small"` menu items.
+   */
+  hotkeys?: undefined
+  /**
+   * Icon is only supported in `size="small"` menu items.
+   */
+  icon?: undefined
+}
```
robinpyon, line 27: **question:** after reviewing current usage of `size="large"` in the studio, there's only two cases I can see – 1. string lists in global search and 2. the presence menu within the nav bar. `badgeText` is not utilised at all.

Is this enough signal / usage to justify us varying on size? 

There could be an opportunity for us to simplify here – given that we already provide an escape hatch (with `children`) that we discourage – perhaps we could simply remove the small / large type variation and default all menu items to use the previous 'small' layout? 

It is a bit of a smell that varying on `size` changes which props are valid. If you're not familiar with the design system, it's not immediately clear why I wouldn't be able to render a 'large' menu item with hotkeys or an icon.

Genuinely curious to know what you think here!

robinpyon (commented):
Thanks @pedrobonamin – left some comments, would love your thoughts here

packages/sanity/src/ui/menuItem/MenuItem.tsx
pedrobonamin: I think it's a discussion worth having, I see pros and cons in both scenarios, but given that at this time we are aiming to simplify the consumer uses, my take will be to keep both possibilities here, as it simplifies the consumer side having the ability to set the size to large and pass some more props to accomplish what you need. 
Probably the workshop menu item will be updated to use one of this, (added a comment in linear asking for this) and I think this item is the one that will use the badge to show the status.

The only menu item I see in a future that will continue using children is the [BlockStyleSelect](packages/sanity/src/core/form/inputs/PortableText/toolbar/BlockStyleSelect.tsx)

pedrobonamin: As discussed today and taking as base the other PR you created, finally we are going for the approach of allowing all props in the same menu item, including stories for not recommended uses. 
It will make the menu items more flexible and easier to use by other developers. 

Children won't be supported, to use children you will need to import from `@sanity/ui`. 

Resolving this conversation.

robinpyon (approved):
Thanks @pedrobonamin, this looks great!

PR #5091, opened 2023-10-30
feat(tab): add StudioUI Tab component

ninaandal:
### Description

Introduces `StudioUI Tab`, adding defaults to create the tabs. Introduces a `small` and `default` size. 
Removes the possibility of adding `icons`, custom `padding` and `fontSize`. 
Updates all the uses of tabs to use the ones.

### What to review
Tab uses (can be tested in field groups)

### Notes for release
Introduces `StudioUI Tab` components, adding defaults to create them.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -228,6 +228,7 @@ module.exports = {
                   'DialogProps',
                   'MenuItem',
                   'MenuItemProps',
+                  'Tab',
                   'Tooltip',
                   'TooltipProps',
                 ],
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -229,6 +229,7 @@ module.exports = {
                   'MenuItem',
                   'MenuItemProps',
                   'Tab',
+                  'TabProps',
                   'Tooltip',
                   'TooltipProps',
                 ],
--- a/packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/GroupTab.tsx
+++ b/packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/GroupTab.tsx
@@ -5,6 +5,7 @@ interface GroupType {
   'aria-controls': string
   autoFocus?: boolean
   disabled?: boolean
+  icon?: React.ComponentType
   name: string
   onClick?: (value: string) => void
   selected: boolean
@@ -29,6 +30,7 @@ export const GroupTab = forwardRef(function GroupTab(
       id={`${props.name}-tab`}
       label={props.title}
       ref={ref}
+      size="small"
       {...props}
       onClick={handleClick}
     />
--- a/packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/GroupTab.tsx
+++ b/packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/GroupTab.tsx
@@ -1,11 +1,11 @@
 import React, {forwardRef} from 'react'
-import {Tab} from '@sanity/ui'
+import {Tab} from '../../../../../ui'
+//import {Tab} from '@sanity/ui'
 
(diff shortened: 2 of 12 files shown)


.eslintrc.cjs
```diff
@@ -221,7 +221,7 @@ module.exports = {
             paths: [
               {
                 name: '@sanity/ui',
-                importNames: ['Tooltip', 'TooltipProps'],
+                importNames: ['Tooltip', 'TooltipProps', 'Tab'],
```
pedrobonamin, line 224: We should also include "TabProps" here, so the imports of TabProps are now from `StudioUI`

packages/sanity/src/core/form/inputs/ObjectInput/fieldGroups/GroupTab.tsx
```diff
@@ -1,11 +1,11 @@
 import React, {forwardRef} from 'react'
-import {Tab} from '@sanity/ui'
+import {Tab} from '../../../../../ui'
+//import {Tab} from '@sanity/ui'
```
pedrobonamin, line 3: 👀

packages/sanity/src/ui/tab/Tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+  UITabProps,
+  'id' | 'label' | 'aria-controls' | 'tone' | 'selected' | 'focused'
+> & {
+  size?: 'default' | 'small'
+}
+
+const smallTabProps = {
+  padding: 2,
+  muted: true,
+  tone: 'positive' as ButtonTone,
```
pedrobonamin, line 21: Not sure about setting the `tone` here, it won't allow you set a tone different to positive as it will be always overridden.
Maybe we just need to `tone = "positive"` in the default props

packages/sanity/src/ui/tab/Tab.tsx
```diff
@@ -0,0 +1,50 @@
...
+export type TabProps = Pick<
+  UITabProps,
+  'aria-controls' | 'focused' | 'icon' | 'id' | 'label' | 'selected' | 'tone'
+> & {
+  size?: 'default' | 'small'
+}
+
+const SMALL_TAB_PROPS = {
+  padding: 2,
+}
```
robinpyon, line 19: Changed screaming snake case for constants outside the component definition, which is a pattern we use throughout (and helps communicate intent). Also removed `muted: true` since this was shared across both.

packages/sanity/src/ui/tab/Tab.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Studio UI components are opinionated `@sanity/ui` components meant for internal use only.
+ * Props and options are intentionally limited to ensure consistency and ease of use.
+ *
+ * @internal
+ */
+export const Tab = forwardRef(function Tab(
+  {
+    size = 'default',
+    tone = 'default',
```
robinpyon, line 36: Set to `default` – we can revisit once new color theming lands

packages/sanity/src/ui/tab/Tab.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React, {forwardRef} from 'react'
+
+/**
+ * @internal
+ *
+ * Padding and font sizes are fixed in Studio UI <Tab> components.
+ */
+export type TabProps = Pick<
+  UITabProps,
+  'aria-controls' | 'focused' | 'icon' | 'id' | 'label' | 'selected' | 'tone'
```
robinpyon, line 12: Re-added `icon` support for now, since this would be considered a breaking change (as users can configure custom icons for groups and document views)

robinpyon (approved):
Thanks @ninaandal – made some minor changes and left comments below!

PR #5093, opened 2023-10-30
refactor: use transient props in various styled components

robinpyon:
### Description

This PR migrates a number of styled components to now use [transient props](https://styled-components.com/docs/api#transient-props), and remove a number of warnings that were shown in localhost as a result of upgrading to styled-components to `6.x`.

A fully exhaustive audit wasn't conducted, but this should address most of the more common / egregious examples.

### What to review

Touched components should be unaffected in function or style. We should no longer see a glut of transient prop related warnings in localhost.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/changeIndicators/ElementWithChangeBar.styled.tsx
+++ b/packages/sanity/src/core/changeIndicators/ElementWithChangeBar.styled.tsx
@@ -6,54 +6,54 @@ interface ThemeContext {
 }
 
 interface RootProps {
-  focus: boolean
-  hover: boolean
-  changed?: boolean
-  isReviewChangeOpen: boolean
-  disabled?: boolean
+  $changed?: boolean
+  $disabled?: boolean
+  $isReviewChangeOpen: boolean
   $withHoverEffect?: boolean
 }
 
 const animationSpeed = 250
 
-export const ChangeBarWrapper = styled.div<RootProps>(({changed, disabled, isReviewChangeOpen}) => {
-  if (disabled)
+export const ChangeBarWrapper = styled.div<RootProps>(
+  ({$changed, $disabled, $isReviewChangeOpen}) => {
+    if ($disabled)
+      return css`
+        ${ChangeBar} {
+          display: none;
+        }
+      `
+
     return css`
-      ${ChangeBar} {
-        display: none;
+      --change-bar-offset: 2px;
+
+      display: flex;
+      position: relative;
+
+      @media (hover: hover) {
+        &:hover {
+          z-index: 10;
+        }
       }
+
+      /* hide when field is not changed */
+      ${!$changed &&
+      css`
+        ${ChangeBar} {
+          opacity: 0;
+          pointer-events: none;
+        }
+      `}
+
+      /* hide hover effect when review changes is open */
+      ${$isReviewChangeOpen &&
+      css`
+        ${ChangeBarButton} {
+          opacity: 0;
+        }
+      `}
     `
-
-  return css`
-    --change-bar-offset: 2px;
-
-    display: flex;
-    position: relative;
-
-    @media (hover: hover) {
-      &:hover {
-        z-index: 10;
-      }
-    }
-
-    /* hide when field is not changed */
-    ${!changed &&
-    css`
-      ${ChangeBar} {
-        opacity: 0;
-        pointer-events: none;
-      }
-    `}
-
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/changeIndicators/ElementWithChangeBar.styled.tsx
```diff
@@ -6,54 +6,54 @@ interface ThemeContext {
 }
 
 interface RootProps {
-  focus: boolean
-  hover: boolean
```
robinpyon, line 10: These were removed as the styled component wasn't doing anything with these values

packages/sanity/src/core/changeIndicators/ElementWithChangeBar.tsx
```diff
@@ -35,34 +30,21 @@ export function ElementWithChangeBar(props: {
             aria-label="Review changes"
             data-testid="change-bar__button"
             onClick={isReviewChangesOpen ? undefined : onOpenReviewChanges}
-            onMouseEnter={handleMouseEnter}
-            onMouseLeave={handleMouseLeave}
```
robinpyon, line 39: Similarly, this internal state has been removed as `ChangeBarWrapper` isn't varying on the value of `hover`

packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
```diff
@@ -29,7 +29,14 @@ export const ReferenceLinkCard = forwardRef(function ReferenceLinkCard(
   props: ReferenceLinkCardProps & React.HTMLProps<HTMLElement>,
   ref: React.ForwardedRef<HTMLElement>,
 ) {
-  const {documentType, as: asProp, ...restProps} = props
+  const {
+    as: asProp,
+    // We exclude `documentId` from spread props to avoid passing it to the Card component
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    documentId,
```
robinpyon, line 36: Would love to know if there's a more elegant way to do this!

packages/sanity/src/core/field/diff/components/DiffTooltip.tsx
```diff
@@ -56,13 +56,7 @@ function DiffTooltipWithAnnotation(props: DiffTooltipWithAnnotationsProps) {
 
   return (
     <LegacyLayerProvider zOffset="paneFooter">
-      <Tooltip
-        content={content}
-        // data-placement={restProps.placement}
-        portal
-        allowedAutoPlacements={['top', 'bottom']}
```
robinpyon, line 63: Removed as it's now deprecated in Sanity UI, and including it would yield similar transient prop warnings

packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
bjoerge: I think you should be able to assingn it to `_` and have eslint ignore it, e.g. `documentId: _`

bjoerge (approved):
Great - thanks a lot @robinpyon! 🙇🏼

packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
robinpyon: I considered this! Though I believe we need to amend our eslint config slightly to accommodate, e.g. 

```
'@typescript-eslint/no-unused-vars': [
  'warn',
  {
    varsIgnorePattern: '^_',
  },
],
```

Happy to add if you feel appropriate! Otherwise we can just leave as-is

pedrobonamin (approved):
looks good to me. Thanks!

packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
bjoerge: Since `const {foo: _, bar: _, ...rest} = …` will yell at you for redeclaring consts, will that rule allow omitting more than one property? Maybe we could configure it to be `_ignore_${propertyName}` (or maybe that's too much?).

There's also [no-unused-vars#ignorerestsiblings](https://eslint.org/docs/latest/rules/no-unused-vars#ignorerestsiblings), which is more specific to this particular case.

PR #5094, opened 2023-10-31
feat(router): add support for search params

bjoerge:
### Description
This PR adds support for a special key on the url state: `_searchParams` , which is an array of `[key: string, value: string]` tuples (similar to what you'd get from `URLSearchParams.entries()`).

It uses the scope (aka namespace) feature of the router, so when you're within a scope you don't need to worry about how your own search params fits in with the global environment. In the url, they will be namespaced with `/some/route?scope[param]=value`. Params in nested scopes will be `scopeA[scopeB][param]=value` (see added unit tests for details).

### Usage example

Given the route `/path/to/:someParam`

#### Read search params

Given the url: `/path/to/something?foo=bar&bar=baz`

```ts
const router = useRouter()

console.log(router.state)
```
Will log
```ts
{
  someParam: 'something',
  _searchParams: [['foo', 'bar'], ['bar', 'baz']]
//^ see note about prefix below 
}
```

#### Navigate w/search params

```ts
const router = useRouter()

router.navigate({
  someParam: 'something',
  _searchParams: [['foo', 'bar'], ['bar', 'baz']]
})
```
Will navigate to:
```
/path/to/something?foo=bar&bar=baz
```
### What to review
- Does the API make sense?
- I chose the `_`-prefix, which isn't *strictly* speaking safe as someone could potentially (although quite unlikely) have an existing route with `_searchParams` as a param (e.g. `/path/to/:_searchParams`), which will then be in conflict. If we want to play it extra safe, we can consider changing to `$searchParams`. We currently don't allow using `$` in url params (i.e. defining a route as `/path/to/:$param` throws an error).

Note: this PR also recovers the test suite and readme that somehow didn't make it when we moved this from `@sanity/state-router` over to the `sanity`-package

### Notes for release
- Adds support for search params in `sanity/router`

### DIFF
--- a/packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
+++ b/packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
@@ -57,7 +57,7 @@ function useRouterFromWorkspaceHistory(
       routerBasePath === '/' ? true : routerBasePathRegex.test(pathname)
     return {
       subscribe: (onStoreChange: () => void) => history.listen(onStoreChange),
-      getSnapshot: () => history.location.pathname,
+      getSnapshot: () => history.location.pathname + history.location.search ?? '',
       // Always return null for the server snapshot, as we can't know how to resolve intents until after authentication is done, which is browser-only
       getServerSnapshot: () => null,
       selector: (pathname: string | null) =>
--- a/packages/sanity/src/router/README.md
+++ b/packages/sanity/src/router/README.md
@@ -1,0 +1,331 @@
+# `sanity/router`
+
+## Features
+
+Based on a routing schema:
+
+- A state object can be derived from the current pathname
+- A state object can be used to generate a path name
+
+## API Usage
+
+Define the routes for your application and how they should map to application state
+
+```js
+import {route} from '@sanity/state-router'
+
+const router = route.create('/', [route.create('/products/:productId'), route.create('/users/:userId'), route('/:page')])
+
+router.encode({})
+// => '/'
+router.decode('/')
+// => {}
+
+router.encode({productId: 54})
+// => '/products/54'
+
+router.decode('/products/54')
+// => {productId: 54}
+
+router.encode({userId: 22})
+// => '/users/22'
+
+router.decode('/users/54')
+// => {userId: 54}
+
+router.encode({page: 'about'})
+// => '/about'
+
+router.decode('/about')
+// => {page: about}
+```
+
+## React usage
+
+### Setup routes and provider
+
+```jsx
+import {route} from '@sanity/state-router'
+import {RouterProvider, withRouter} from '@sanity/state-router/components'
+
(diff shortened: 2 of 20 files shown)


rexxars (approved):

mariuslundgard (commented):
I'm worrying that the API is unconventional. Did you consider this approach?

> Given the url: /path/to/something?foo=bar&bar=baz
> 
> ```ts
> const router = useRouter()
> 
> console.log(router.state, router.searchParams)
> ```
>
> Will log
> 
> ```ts
> {someParam: 'something'}, {foo: 'bar', bar: 'baz'}
> ```

bjoerge:
Can you elaborate on what you mean by unconventional, and what problems you think it may cause?

I did consider it! But the url params _is_ part of the state, so it felt natural to keep it there. It will make it easier to pass it around as a unified object representing the full picture of the current extracted url state.

This, in addition to the fact that it would be harder to implement in a backwards compatible way (router.encode/decode is part of our public API), made me go for this solution, but open to reconsider if you can help me understand what real problems it can cause.

If we didn't have to care about backwards compat, I think a better option would be if the state object itself had a distinction between url params and search params, e.g. using the example above:

Given the url: `/path/to/something?foo=bar&bar=baz`

```ts
const router = useRouter()

console.log(router.state)
```
Will log (something like)
```ts
{
  pathParams: {
    someParam: 'something'
  },
  searchParams: [['foo', 'bar'], ['bar', 'baz']]
}
```

Noting this as a potential future improvement

mariuslundgard:
@bjoerge This part looks especially unconvential/foreign:
```ts
_searchParams: [['foo', 'bar'], ['bar', 'baz']]
```

Would be more readable like this:
```ts
_searchParams: {foo: 'bar', bar: 'baz'}
```

And ideally also just be outside of the `state` object so it doesn't need the `_` or `$` prefix at all.

bjoerge:
> ```_searchParams: {foo: 'bar', bar: 'baz'}```

It's completely valid to have multiple search params with the same name (which is why `URLSearchParams` isn't a "map-like", structure, but has, for example, a `getAll('<param>') => string[]` method.

If we made it a record/map-like it would have to be:
```_searchParams: {foo: ['bar'], bar: ['baz']}```

But as noted in the PR description, this format is aligned with what you get from `URLSearchParams().entries()`, (and also other `.entries()` apis, etc.), so I it's a pretty common way of representing key/value pairs. You can also pass it as-is to the `URLSearchParams` [constructor](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#options.), e.g. `new URLSearchParams([['foo', 'bar'], ['bar', 'baz']])`, which you can't with a record.

Given that this PR is approved, I'll merge it for now. It should be trivial to convert the entries array into something more convenient, e.g. by using `Object.fromEntries(searchParams)` (if you known you're not dealing with multiple params) or 
`new URLSearchParams(searchParams)` if you do.

PR #5107, opened 2023-11-02
refactor: split Studio UI MenuItem into separate components

robinpyon:
### Description

- Rather than varying on `size`, this PR splits `<MenuItem>` into two separate components – `<MenuItem>` and `<PreviewMenuItem>`
- Workshop stories have been created for the above, indicating optionality
- Workspace preview component has been updated to match updated layout
- Support for `children` has been removed from Studio UI `<MenuItem>` variants – in these (rare) instances, users can just use regular Sanity UI components. Known cases:
   - [WorkspaceMenuButton](https://github.com/sanity-io/sanity/blob/EDX-635-menu-items-split-components/packages/sanity/src/core/studio/components/navbar/workspace/WorkspaceMenuButton.tsx) – this uses WorkspacePreview (which is reappropriated in other non-menu contexts, such as the workspace auth screen)
   - [BlockStyleSelect](https://github.com/sanity-io/sanity/blob/EDX-635-menu-items-split-components/packages/sanity/src/core/form/inputs/PortableText/toolbar/BlockStyleSelect.tsx) - this needs to render custom PT styles
- Renamed some props – opting against hungarian notation (e.g. `subText => subtitle`) and making some other values more agnostic (`avatar => preview`)

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -221,7 +221,7 @@ module.exports = {
             paths: [
               {
                 name: '@sanity/ui',
-                importNames: ['Tooltip', 'TooltipProps', 'MenuItem', 'MenuItemProps'],
+                importNames: ['MenuItem', 'MenuItemProps', 'Tooltip', 'TooltipProps'],
                 message:
                   'Please use the (more opinionated) exported components in sanity/src/ui instead.',
               },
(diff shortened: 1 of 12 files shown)


pedrobonamin (commented):
Hey @robinpyon thanks for this changes, appreciate the detailed review.

I think I wouldn't go for the change in the StudioUI `<MenuItem />` component.
With previous implementation, developers trying to build a menu had only one place to import the menu item from, with these changes they will have 3 different places.

They will now have to know that if they want to use something like preview, they need to import the PreviewComponent, instead of MenuItem or they would not understand why children isn't working and how to bypass that. 

Also, this will allow users to create a MenuItem with text, subtitle and icon, which by design it shouldn't be allowed.

For the rest of the changes, I would go for it. Having the stories is a great addition, but I would remove the story that includes the icon and subtitle. 

Added some changes to the original branch to make the `WorkspacePreview` use the menu item with badge. https://github.com/sanity-io/sanity/pull/5090/commits/a119ef0c097d75cafadd20177e6ea6e351e7816e

robinpyon:
Thanks @pedrobonamin! As we discussed offline, closing this in favour of an alternate approach, will open a new PR here shortly

PR #5109, opened 2023-11-02
fix(desk): skip waiting for welcome event before fetching list items

rexxars:
### Description

_Note; this was a quick stab at something that's been at the back of my mind for a bit, no rush - please review thoroughly and think through any consequences this might have._

I _believe_ we originally introduced the pattern of "set up a listener, wait for a welcome event, then fetch the results" in order to know that the list is _guaranteed_ to include all documents. To clarify; the case we are trying to prevent with this approach is where:

- We perform the query and set up a listener _at the same time_
- The query completes/executes _before_ the listener has been "established", and we have received a `welcome` event
- In between the two events, a mutation happens that creates, updates or deletes a document that should make the document appear/disappear from the list

In this case, we obviously end up with a list that isn't _quite_ correct. But the _downside_ of this approach is having to wait for the `welcome` event every single time you view/navigate to a list. The time this takes depends on many factors, like shard load, network conditions, location etc - but on bad days I may wait between 200 and 600 milliseconds for this to happen.

This PR does a trade-off: it prioritizes speed over consistency. In this particular case I think it is an acceptable one, given we are not talking about _editing_ consistency or such, but I would be open to discussing.

_Ideally_, this would be addressed by having listeners be able to get an initial "snapshot" upon establishing the listener - basically a "query and listen" all in one. But until we can build an API or that in Content Lake, maybe this can improve the user experience a bit.

### What to review

- Are there any unforeseen consequences with this approach?
- Do document lists still function as well as before, in terms of throttling, loading etc?
- Do you agree with the trade-off described?
- Are there alternative solutions to this that we could implement that I have not considered?

### Notes for release

- Improved speed when doing initial load of document lists

### DIFF
--- a/packages/sanity/src/desk/panes/documentList/listenSearchQuery.ts
+++ b/packages/sanity/src/desk/panes/documentList/listenSearchQuery.ts
@@ -1,27 +1,26 @@
-import {SanityClient, SanityDocument} from '@sanity/client'
+import type {SanityClient, SanityDocument} from '@sanity/client'
 import {
   asyncScheduler,
   defer,
+  filter as filterEvents,
   merge,
   mergeMap,
   Observable,
   of,
-  partition,
   share,
-  take,
   throttleTime,
   throwError,
   timer,
 } from 'rxjs'
 import {exhaustMapWithTrailing} from 'rxjs-exhaustmap-with-trailing'
-import {SortOrder} from './types'
+import type {SortOrder} from './types'
 import {
   createSearchQuery,
-  Schema,
-  SearchableType,
-  SearchOptions,
-  SearchTerms,
-  WeightedSearchOptions,
+  type Schema,
+  type SearchableType,
+  type SearchOptions,
+  type SearchTerms,
+  type WeightedSearchOptions,
 } from 'sanity'
 
 interface ListenQueryOptions {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/desk/panes/documentList/listenSearchQuery.ts
```diff
@@ -69,12 +68,11 @@ export function listenSearchQuery(options: ListenQueryOptions): Observable<Sanit
...
 
-  const [welcome$, mutationAndReconnect$] = partition(events$, (ev) => ev.type === 'welcome')
-
   return merge(
-    welcome$.pipe(take(1)),
-    mutationAndReconnect$.pipe(throttleTime(1000, asyncScheduler, {leading: true, trailing: true})),
+    of({type: 'welcome', visibility: null}),
+    events$.pipe(filterEvents((ev) => ev.type !== 'welcome')),
   ).pipe(
+    throttleTime(1000, asyncScheduler, {leading: true, trailing: true}),
```
bjoerge, line 75: I think the current logic addresses an edge case that rarely happens, so I'm not against merging this in it's current shape.

> Are there any unforeseen consequences with this approach?

I find it hard to say with confidence - it will _rarely_ be an issue for sure, but what downstream effects there are _when_ it happens is hard to say. I'm not sure how much third parties rely on this function and the current guarantees it provides.

> Do document lists still function as well as before, in terms of throttling, loading etc?

Yes, the logic seems fine and it worked as intended when I tested locally

> Do you agree with the trade-off described?

Given that it's such an edge case, I think it's fair, but I'd expect it to occationally display a slightly outdated list for a tiny percentage of studio users, which is probably ok.

> Are there alternative solutions to this that we could implement that I have not considered?

Not sure if you considered, but if we instead make the trade-off be _an extra initial request_, we can keep the current guarantees by doing this:
```suggestion
  const [welcome$, mutationAndReconnect$] = partition(events$, (ev) => ev.type === 'welcome')

  return merge(
    of({type: 'init', visibility: null}),
    welcome$.pipe(take(1)),
    mutationAndReconnect$.pipe(throttleTime(1000, asyncScheduler, {leading: true, trailing: true})),
  ).pipe(
```

PR #5112, opened 2023-11-02
chore: fix lint errors

bjoerge:
Fixes various lint errors that we got after [#5600](https://github.com/sanity-io/sanity/pull/5100)


binoy14:
Should we maybe look into merging https://github.com/sanity-io/sanity/pull/5033 instead? That should fix the errors

rexxars (approved):

bjoerge:
> Should we maybe look into merging https://github.com/sanity-io/sanity/pull/5033 instead? That should fix the errors

Good call – I've merged #5033 instead

PR #5113, opened 2023-11-02
chore(lint): use turbo for linting all the packages

binoy14:
### Description

This enables using turbo for linting packages which would run linting in all packages in parallel so on every change all the files are correctly checked and it gets cached so it should be super fast in subsequent runs

- removes the root level `.eslintignore` since each package needs it's own `.eslintignore` as turbo runs everything in the working directory
- updates the lintIfChanged action to run linting across all packages in the monorepo on every PR and merge on next

### What to review

Configuration changes make sense

### Notes for release

- N/A


package.json
```diff
@@ -49,8 +49,7 @@
     "etl": "node -r dotenv-flow/config -r esbuild-register scripts/etl",
     "init": "lerna clean --yes && run-s bootstrap build",
     "lerna:clean": "lerna clean",
-    "lint": "eslint --cache .",
-    "lint-next": "eslint `git diff --name-only next --diff-filter ACMRTUXB`",
+    "lint": "turbo run lint --continue -- --quiet",
```
binoy14, line 52: TIL: `--continue` stops turbo from shortcuting onthe first error. This might be useful to see all the linting error happening in your PR. 

https://turbo.build/repo/docs/reference/command-line-reference/run#--continue

binoy14:
It is expected to fail since the linting action as there are currently issues on `next`. This makes sure that PRs are not introducing bad linting or some linter errors in `next`

rexxars (approved):
I'm a little confused by the need to move the ignore files to individual directories, given the eslint configs have not changed and seem to still inherit from the parent?

Would it work if we moved from an `.eslintignore` to `ignorePatterns` [defined in the root config](https://eslint.org/docs/latest/use/configure/ignore#ignorepatterns-in-config-files)? 

Would be nice not to duplicate the ignore file throughout, but if it can't be helped, fair enough.

binoy14:
> I'm a little confused by the need to move the ignore files to individual directories, given the eslint configs have not changed and seem to still inherit from the parent?

If I read [this](https://eslint.org/docs/latest/use/configure/ignore#the-eslintignore-file) correctly. It only uses a single ignore file.

> Only one .eslintignore file can be used at a time, so .eslintignore files other than the one in the current working directory are not used

Since turbo is not running lint in each working directory of the package. The root one is ignored, Sadly it does not work like the `.eslintrc.js` where it traverser to the root


> Would it work if we moved from an .eslintignore to ignorePatterns [defined in the root config](https://eslint.org/docs/latest/use/configure/ignore#ignorepatterns-in-config-files)?

that's a good point. I think that would work for common files and each package can define specific ignores if needed. I will give that a shot

binoy14:
Was able to remove duplicate `.eslintignore` in packages. Only packages that need override have those files

bjoerge (approved):
This will be a huge time saver, thank you!

PR #5114, opened 2023-11-02
feat(i18n): allow simple HTML tags when using Translate component

rexxars:
### Description

It's a bit tedious to pass in components like `<Code>` that just maps to `<code>`. This PR allows using this predefined list of tags without specifying it directly: `<abbr>`, `<address>`, `<cite>`, `<code>`, `<del>`, `<em>`, `<ins>`, `<kbd>`, `<q>`, `<samp>`, `<strong>`, `<sub>`, `<sup>`.

The reason for these specifically is that they are inline elements that do not require any attributes for use, and is not deprecated by the HTML5 spec.

I'm not sure if my solution in the parser was the best, but I didn't want to spend more time finding a more elegant way to approach it - open to suggestions.

### What to review

- Using simple tags with the `<Translate/>` component works as it should

### Notes for release

None.

### DIFF
--- a/packages/sanity/src/core/i18n/__tests__/simpleParser.test.ts
+++ b/packages/sanity/src/core/i18n/__tests__/simpleParser.test.ts
@@ -90,14 +90,14 @@ describe('simpleParser - errors', () => {
       'Expected closing tag for <Red>, but found closing tag </Blue> instead. Make sure each opening tag has a matching closing tag.',
     )
   })
-  test('tags must be title cased', () => {
-    expect(() => simpleParser('foo <red> bar</Blue>')).toThrow(
-      'Invalid tag "<red>". Tag names must start with an uppercase letter and can only include letters and numbers.',
+  test('does not allow camelCased tag names', () => {
+    expect(() => simpleParser('foo <camelCased>bar</camelCased>')).toThrow(
+      'Invalid tag "<camelCased>". Tag names must be lowercase HTML tags or start with an uppercase letter and can only include letters and numbers.',
     )
   })
   test('tags cant contain whitespace or special characters', () => {
     expect(() => simpleParser('foo <Em@ail> bar</Em@ail>')).toThrow(
-      'Invalid tag "<Em@ail>". Tag names must start with an uppercase letter and can only include letters and numbers.',
+      'Invalid tag "<Em@ail>". Tag names must be lowercase HTML tags or start with an uppercase letter and can only include letters and numbers.',
     )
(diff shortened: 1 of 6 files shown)


ricokahler (approved):
nice

packages/sanity/src/core/i18n/__tests__/simpleParser.test.ts
```diff
@@ -90,14 +90,14 @@ describe('simpleParser - errors', () => {
...
-      'Invalid tag "<red>". Tag names must start with an uppercase letter and can only include letters and numbers.',
+  test('does not allow camelCased tag names', () => {
+    expect(() => simpleParser('foo <camelCased>bar</camelCased>')).toThrow(
+      'Invalid tag "<camelCased>". Tag names must be lowercase HTML tags or start with an uppercase letter and can only include letters and numbers.',
     )
   })
   test('tags cant contain whitespace or special characters', () => {
     expect(() => simpleParser('foo <Em@ail> bar</Em@ail>')).toThrow(
-      'Invalid tag "<Em@ail>". Tag names must start with an uppercase letter and can only include letters and numbers.',
+      'Invalid tag "<Em@ail>". Tag names must be lowercase HTML tags or start with an uppercase letter and can only include letters and numbers.',
```
bjoerge, line 100: should we mention the subset of tags we allow here?

packages/sanity/src/core/i18n/simpleParser.ts
```diff
@@ -28,6 +29,24 @@ function matchCloseTag(input: string) {
...
+  }
+
+  const isValidHtmlTagName = VALID_HTML_TAG_NAME_RE.test(tagName)
+  if (isValidHtmlTagName) {
+    return
+  }
+
+  throw new Error(
+    tagName.trim() === tagName
+      ? `Invalid tag "<${tagName}>". Tag names must be lowercase HTML tags or start with an uppercase letter and can only include letters and numbers.`
```
bjoerge, line 45: re:previous comment: maybe this would be a better place to mention what subset of tags we allow?

bjoerge (approved):
just a question/comment, otherwise great!

packages/sanity/src/core/i18n/simpleParser.ts
rexxars: Maybe. It does print an error in the Translate component, so I think that is fine for now. I tried refactoring, but ran into some parser shenanigan holes that I don't want to reason about at the moment. We can get back to it later if it becomes a problem.

PR #5116, opened 2023-11-02
fix(core): reference links not opening correct pane

### LINKED ISSUE #5105: References Forced to Reload When Opened Under Parent
**Describe the bug**

Child references are forced to reload when opening them under a parent. This bug only occurs on version 3.19.0, with the expected behavior occurring on earlier versions.
"This reference has changed since you opened it." is what I get every time I open a reference.

image

**Expected behavior**

References should load without having to reload the reference.

v3.19.0
All latest (command not working for me for some reason)

Windows 11

npm v10.2.2
node v21.1.0

rexxars:
### Description

Fixes #5105

#5093 introduced a regression where references can no longer be opened in-place.
This is caused by removing the `documentId` from [being passed to the state link](https://github.com/sanity-io/sanity/pull/5093/files#diff-f08509f8474b2f3f1c0301d9ca9693b77fa019a86d6ece6c45013ba55f5f7a77).

This revert _this_ change, but reintroduces warnings about passing it to DOM elements, so it is not a perfect solution. I struggled to figure out what is actually being passed to what here - a lot of `as`/`data-as`/Card/StyledCard magic and layers of abstraction, so I would appreciate if either of you had time to fix this properly and push to this branch.

This warrants a hotfix release once addressed.

### What to review

- That it fixes the problem
- What the proper solution is

### Notes for release

- Fixed an issue where references would not open in-place, instead yielding an error about being changed after opening

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceLinkCard.tsx
@@ -29,14 +29,7 @@ export const ReferenceLinkCard = forwardRef(function ReferenceLinkCard(
   props: ReferenceLinkCardProps & React.HTMLProps<HTMLElement>,
   ref: React.ForwardedRef<HTMLElement>,
 ) {
-  const {
-    as: asProp,
-    // We exclude `documentId` from spread props to avoid passing it to the Card component
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    documentId,
-    documentType,
-    ...cardProps
-  } = props
+  const {as: asProp, documentId, documentType, ...cardProps} = props
   const dataAs = documentType ? 'a' : undefined
 
   // If the child link is clicked without a document type, an error will be thrown.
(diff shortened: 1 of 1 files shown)


bjoerge (approved):

robinpyon:
Apologies – I should have been much more stringent here, especially given the presence of the polymorphic `as` prop. 

I did have a quick look into this and I believe the following workaround removes the error, though would appreciate your eyes @bjoerge 

```
// ReferenceLinkCard.tsx

// Before
const as = documentType ? asProp : 'div'

// After
const as = documentId || documentType ? asProp : 'div'

return (
  <StyledCard
    ...
    forwardedAs={as}
    ...
  />
)
```

**What's happening:**
- `ReferenceLinkCard` renders multiple times and receives both `documentId` and `documentType`, **but not both to start**. It renders multiple times, and receives `documentId` first (`documentType` will be `undefined`)
- During this render pass where only `documentId` is valid, `as` will still be set to `div`, yielding the errors about `documentId` not being a valid DOM element
- At a later render where `documentType` has a valid, `as` will then be set to `asProp`, which is `EditReferenceLink` (which ends up being [ReferenceChildLink](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/desk/components/paneRouter/ReferenceChildLink.tsx))
- At this point, `ReferenceChildLink` renders, using `documentId` and `documentType` props (and doesn't pass these to the underlying DOM element)
- I believe this explains why we don't receive errors for `documentType`

PR #5120, opened 2023-11-03
refactor(desk): comment highlight

hermanwikner:
### Description

This pull request moves the logic to set the "selected path" in comments to its own provider + makes a few update. Issues fixed in this PR: `EDX-714`, `EDX-708` and `EDX-719`.

### What to review

- Clicking on the "new thread" card/input should highlight the field and the card/input
- Comment highlight should disappear on click outside
- Comment highlight should be persistent on the form field when clicking on a comment item in the inspector
- When re-opening a comment (ie change status) – we should scroll to the comment in the open view

### Notes for release

n/a


packages/sanity/src/desk/comments/src/context/selected-path/CommentsSelectedPathProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface CommentsSelectedPathProviderProps {
+  children: React.ReactNode
+}
+
+export const CommentsSelectedPathProvider = React.memo(function CommentsSelectedPathProvider(
+  props: CommentsSelectedPathProviderProps,
+) {
+  const {children} = props
+  const [selectedPath, setSelectedPath] = useState<CommentsSelectedPath>(null)
```
robinpyon, line 13: (To preface: this new behaviour feels so much more intuitive, well done!)

As `setSelectedPath` receives an object literal, any `setSelectedPath` call will always trigger a _full_ re-render of the document pane, regardless of whether any path values (`fieldPath`, `threadId` etc) have changed or not.

This is especially noticeable when clicking between different threads within a field path. On local, this previously used to feel _instant_ but now there's a definite hitch, which is likely due to the entire document form re-rendering. 

This issue is also exacerbated by the fact that `setSelectedPath` is being called whenever you click _anywhere_ in the comment. Making a text selection in a comment will trigger two renders, which we probably don't want!

Rather than passing the `useState` set function to context – perhaps we can pass a custom callback which does a shallow comparison before setting internal state?

robinpyon, line 13: As `setSelectedPath` receives an object literal, any call will always trigger a full re-render of the document pane, regardless of whether any path values (`fieldPath`, `threadId` etc) have changed or not.

This is especially noticeable when clicking between different threads within a field path. On local, this previously used to feel instant but now there's a definite hitch, which is likely due to the entire document form re-rendering.

This issue is also exacerbated by the fact that `setSelectedPath` is being called whenever you click anywhere in the comment. Making a text selection in a comment will trigger two renders, which we probably don't want!

Rather than passing the `useState` set function to the context's value – perhaps we can pass a custom callback which does a shallow comparison before setting internal state?

robinpyon (commented):
Thanks @hermanwikner – this new behaviour is so much more intuitive, well done

There's one perf issue that I wanted to call out, otherwise everything generally looks good!

packages/sanity/src/desk/comments/src/context/selected-path/CommentsSelectedPathProvider.tsx
hermanwikner: Thanks, I will look into this 👍 

However, in the previous approach, the selected comment was also in the shape of an object, so nothing has really changed in that regard!

> This issue is also exacerbated by the fact that setSelectedPath is being called whenever you click anywhere in the comment. Making a text selection in a comment will trigger two renders, which we probably don't want!

I just pushed a fix to prevent this from happening!

hermanwikner: The issue here appears to be related to the "on click outside" behavior, which often triggers two state updates when clicking on an item. That is, if one item is already selected and you click on another item, you first update the state due to the click outside the initially selected item, and then a second update occurs because you've clicked on another element.

packages/sanity/src/desk/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -236,23 +237,32 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
...
 
-  const avatar = <CommentsAvatar user={user} />
+  useClickOutside(
+    useCallback(() => {
+      if (!hasChanges) {
+        cancelEdit()
+      }
+    }, [cancelEdit, hasChanges]),
+    [rootElement],
+  )
```
robinpyon, line 265: **question:** WDYT about pulling out these 'inline' `useCallback` cases and giving them named functions (`handleXXX`)? 

I think it can better communicate intent and makes the file easier to parse – it's nice being able to get an overview of all event callbacks associated with this component in one place, and makes it easier to identify opportunities to make this DRYer, too.

**For example:**

```
  useDidUpdate(
    isEditing,
    useCallback(() => {
      setMenuOpen(false)
    }, []),
  )
```

could be replaced with

```
  useDidUpdate(isEditing, handleMenuClose)
```

hermanwikner: Totally agree – I can tix this!

robinpyon (approved):
Awesome work @hermanwikner!

PR #5121, opened 2023-11-03
fix(desk): comment mention list issue when using the '@' button

hermanwikner:
### Description

This pull request fixes an issue when opening the mentions menu when inserting an '@' character in the comments input with the button. 

### What to review

 Insert a '@' character with the mention button and make sure that you can navigate the menu.


robinpyon (approved):
Thanks @hermanwikner - LGTM! 

Added a basic playwright-ct test which captures this behaviour too

PR #5123, opened 2023-11-03
fix(portable-text-editor): fix various

skogsmaskin:
### Description

This will improve and fix a couple of things with the PTE:

* Simplify the key name used for the "virtual" text node of inline objects and object blocks in the state for easier tests and value comparisons. This key doesn't need to be unique and can be put in a constant. This will fix an issue where the external value would need to be synced after inserting inline objects when it really isn't needed.

* When normalizing a Slate node that haven't been assigned the span type, generate a key and spread the existing node data (possibly .marks) in the normalization step. This will save us other normalization steps (`._key` normalization, and `.marks` normalization) down the line.
* Fix an issue with the `PortableTextEditor.insertChild`  function, where if you inserted a new span child when having focus on a inline object, it would try to insert it into the inline-object's "virtual" text node instead as a child of the block itself.

* Added and updated tests for these things.

### What to review
That the changes look good. The tests should cover most of the changes I think.
 

### Notes for release
N/A - mostly internal and non-user-facing improvements.


bjoerge (approved):
Looks great, LGTM!

PR #5125, opened 2023-11-03
feat(core): add `BetaBadge` component

hermanwikner:
### Description

This pull request: 
- Adds a `BetaBadge` component from the `sanity` package to be used to highlight beta features
- Refactors the comments implementation to use the new beta badge component

This makes it possible for other plugins/features to `import {BetaBadge} from 'sanity'`.

Screenshot 2023-11-03 at 14 54 58

### What to review

- Make sure that everything looks good

### Notes for release

n/a


packages/sanity/src/core/components/BetaBadge.tsx
```diff
@@ -0,0 +1,34 @@
...
+    color: ${fg};
+  `
+})
+
+/** @internal */
+export type BetaBadgeProps = Omit<BadgeProps, 'mode' | 'tone'>
+
+/** @internal */
+export function BetaBadge(props: PropsWithChildren<BetaBadgeProps>) {
+  const {fontSize: fontSizeProp, children: childrenProp} = props
```
bjoerge, line 24: Tiny nit, but you could apply defaults this way instead:
```suggestion
  const {fontSize = 1, children = 'Beta', ...rest} = props
```

packages/sanity/src/core/components/BetaBadge.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  const fontSize = fontSizeProp || 1
+  const children = childrenProp || 'Beta'
+
+  return (
+    <StyledBadge {...props} fontSize={fontSize}>
```
bjoerge, line 30: Slightly better to avoid repeating props, and reducing risk of accidentally overwriting passed vs default props if we move things around.

```suggestion
    <StyledBadge {...rest} fontSize={fontSize}>
```

bjoerge (approved):
Couple of suggestions, otherwise lgtm

PR #5126, opened 2023-11-03
fix(desk/comments): insert space after mention insert

skogsmaskin:
### Description

This will conveniently insert a space after a mention is inserted into the comment input.


hermanwikner (approved):
Works great, thanks 👍

PR #5129, opened 2023-11-03
feat(i18n): translate/localize diff components

rexxars:
### Description

Localizes all built-in diff components and related diff utilities.
Snuck in some changes to how units are displayed in the file diff, since I had to use the unit formatter anyway. It now chooses a more appropriate unit than megabytes if it makes sense.

### What to review

The usual

### Notes for release

None


bjoerge (approved):
Excellent work!

bjoerge:
(type check is failing though)

PR #5130, opened 2023-11-06
fix(desk): various comments UX papercuts

robinpyon:
### Description

(These changes only affect comments within desk)

- Forces gray on all comment dates (edx-701)
- Fixes mention tooltip layout (edx-709)
- Updates thread delete copy (edx-711)
- Updates coachmark copy (edx-716)

### What to review

That all of the above looks as it should!

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/desk/comments/src/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/desk/comments/src/components/list/CommentsListItemLayout.tsx
@@ -1,3 +1,4 @@
+import {hues} from '@sanity/color'
 import {
   TextSkeleton,
   Flex,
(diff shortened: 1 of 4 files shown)


packages/sanity/src/desk/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -65,6 +65,12 @@ const RetryCardButton = styled(Card)`
 
 const StyledCommentsListItemContextMenu = styled(CommentsListItemContextMenu)``
 
+// Container which displays comment dates in gray.
+// Transparent tone cards still contain a background, which we unset to ensure parent colors take precedence.
+const TransparentDateCard = styled(Card).attrs({tone: 'transparent'})`
+  background: unset;
```
hermanwikner, line 71: Is this really using 800 as tint value (as requested from design)? Should we access the color from `@sanity/color` instead – to really make sure that the color is always gray and has the correct tint?


```ts
import {hues} from '@sanity/color'

const TimeText = styled(Text)(({theme}) => {
  const isDark = theme.sanity.color.dark
  const fg = hues.gray[isDark ? 50 : 800].hex

  return css`
    min-width: max-content;
    --card-fg-color: ${fg};
  `
})

```

```ts
<TimeText size={0} />
```

hermanwikner (approved):
Thanks @robinpyon, looks good! 👍 

I had a question about the color change of the text. If we really want to force the text to always be gray with specific tints - I think we should access the color from `@sanity/color` instead since the colors provided by `Card` is themable.

robinpyon:
Thanks @hermanwikner  – ended up going with your suggestion!

We briefly touched on this the other day, but in future we should probably look to discourage cherry picking colors / accessing `@sanity/color` directly in the studio since it can easily lead to inconsistency in our UI. 

If we know `<Text>` within the studio is only going to ever take a few tints, we would ideally abstract that out into a component which simplifies this (which I believe `TextWithTone` sort of does this, with both `muted` and `dimmed` props). Something for us to consider in facelift!

PR #5132, opened 2023-11-06
feat(desk): improve comments mention list filtering

hermanwikner:
### Description

This pull request improves the filtering logic in the comments mention list by sorting the results with what the user has searched for first

### What to review

- Make sure that you get most relevant results first in the list when you search for a user

### DIFF
--- a/packages/sanity/src/desk/comments/src/components/mentions/MentionsMenu.tsx
+++ b/packages/sanity/src/desk/comments/src/components/mentions/MentionsMenu.tsx
@@ -66,9 +66,20 @@ export const MentionsMenu = React.forwardRef(function MentionsMenu(
   const filteredOptions = useMemo(() => {
     if (!searchTerm) return options || EMPTY_ARRAY
 
-    const filtered = options?.filter((option) => {
-      return option?.displayName?.toLowerCase().includes(searchTerm.toLowerCase())
-    })
+    const filtered = options
+      ?.filter((option) => {
+        return option?.displayName?.toLowerCase().includes(searchTerm.toLowerCase())
+      })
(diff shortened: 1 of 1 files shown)


packages/sanity/src/desk/comments/src/components/mentions/MentionsMenu.tsx
```diff
@@ -66,9 +66,20 @@ export const MentionsMenu = React.forwardRef(function MentionsMenu(
   const filteredOptions = useMemo(() => {
     if (!searchTerm) return options || EMPTY_ARRAY
 
-    const filtered = options?.filter((option) => {
-      return option?.displayName?.toLowerCase().includes(searchTerm.toLowerCase())
-    })
+    const filtered = options
+      ?.filter((option) => {
```
hermanwikner, line 70: Maybe this is a better approach:

```ts
?.filter((option) => {
  // Split the option.displayName into words and check if any of them start with the search term
  const words = option?.displayName?.split(' ') || []
  return words.some((word) => word.toLowerCase().startsWith(searchTerm.toLowerCase()))
})
```

robinpyon (approved):
Thanks @hermanwikner – this is a good improvement and works!

Currently, this doesn't handle diacritics, meaning a search for a search for `pal` won't pick up `pål`, etc – but it's not a big issue for now!

In the short term, we could potentially [deburr](https://lodash.com/docs/4.17.15#deburr) all strings during comparison (in addition to lowercasing), though ideally we would lean more on existing APIs like `String.prototype.localeCompare` and/or `Intl.Collator`. This could manifest in potential `localeContains` or `localeStartsWith` helper functions which take raw values and handles everything else for us.

PR #5136, opened 2023-11-06
fix(portable-text-editor): selection validation and perf. improvement

skogsmaskin:
### Description

This will improve the selection management of the Portable Text Editor's Editable component when used in unknown contexts where the underlying context might uncontrollably cause re-renders to be made.

For instance when having a PortableText Input inside an object that is inside another Portable Text Input.

In this case, we would have two nested form contexts that each will update themselves as soon as the value from the nested editor is propagated to the root PT-input, and the `component.input`'s for both of them will then be re-rendered.

Though none of the actual editor nodes are re-rendered, this may cause the window selection to change and the editor will try to make an editor selection out of it.

If the user then is in the middle of making changes (like splitting a block into a new one) this may cause the selection to become out of sync and act weird.

To solve this problem I have improved the DOMMutation observer function we already have in the `Editable` component to:

* Actually _work_ in React non-strict mode (the way the observer was set up with a ref dependency, never engaged the MutationObserver in production mode as it was only set once (to null). Use a state var to control this rather.
* Test that the editor is selected and active before actually doing something (perf. optimization)
* Test the current DOM Range against the DOMRange produced from the editor selection. If they differ, set the actual DOM Range to be the one based on the editor selection.

BONUS: while debugging this, I was also able to significantly optimize performance on rendering the form members' inputs which is pretty notable in the editing experience. 🎉 

I have also made some other minor improvements. See the commits.

* Improved accuracy of some tests
* Refactor away a redundant function in the Synchronizer (it's still doing the exact same thing, just less code).
* Call `editor.onChange` on hotkey actions that should produce a change instantly (like splitting a block with enter).

### What to review
That the cursor is stable and correct when editing Portable Text in nested Portable Text Inputs.

You can test with [the recursive inputs here](https://test-studio-git-edx-727.sanity.build/test/content/input-standard;portable-text;blocksTest;).

### Notes for release
* Fixed an issue with cursor instability when using the Portable Text Input inside another Portable Text Input's embedded objects.
* Improved performance in the Portable Text Input

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -26,7 +26,7 @@ import {
 import {HotkeyOptions} from '../types/options'
 import {fromSlateValue, isEqualToEmptyEditor, toSlateValue} from '../utils/values'
 import {normalizeSelection} from '../utils/selection'
-import {removeAllDocumentSelectionRanges, toPortableTextRange, toSlateRange} from '../utils/ranges'
+import {toPortableTextRange, toSlateRange} from '../utils/ranges'
 import {debugWithName} from '../utils/debug'
 import {usePortableTextEditorReadOnlyStatus} from './hooks/usePortableTextReadOnly'
 import {usePortableTextEditorKeyGenerator} from './hooks/usePortableTextEditorKeyGenerator'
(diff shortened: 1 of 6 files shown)


robinpyon (approved):
Thanks @skogsmaskin – LGTM!

PR #5137, opened 2023-11-06
feat(i18n): translate form field components

rexxars:
### Description

Translates the form field components

### DIFF
--- a/packages/sanity/src/core/form/components/formField/FormFieldValidationStatus.tsx
+++ b/packages/sanity/src/core/form/components/formField/FormFieldValidationStatus.tsx
@@ -1,11 +1,10 @@
-/* eslint-disable camelcase */
-
+import React from 'react'
 import {hues} from '@sanity/color'
 import {ErrorOutlineIcon, InfoOutlineIcon, WarningOutlineIcon} from '@sanity/icons'
-import {FormNodeValidation} from '@sanity/types'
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/form/components/formField/FormFieldValidationStatus.tsx
```diff
@@ -156,25 +130,68 @@ function FormFieldValidationStatusItem(props: {validation: FormNodeValidation})
...
+  const {t} = useTranslation()
   return (
-    <Text size={1}>
-      {errorText || ''}
-      {hasBoth && <> and </>}
-      {warningText || ''}
-    </Text>
+    <ErrorOutlineIcon
+      data-testid="input-validation-icon-error"
+      aria-label={t('form.validation.has-error-aria-label')}
```
ricokahler, line 168: this is a comment on the old code but i'm unsure if an aria-label is necessary if there is `aria-hidden` and `role="presentation"` on there

ricokahler (approved):
nice clean up too!

bjoerge (approved):

packages/sanity/src/core/form/components/formField/FormFieldValidationStatus.tsx
jtpetty: agreed, but probably something we should look at later when we are focusing on accessibility

PR #5139, opened 2023-11-07
feat(i18n): fix confirm delete dialog

ricokahler:
### Description

A follow up from #5114 so that these components use the `context` prop instead of passing it through `values`


jtpetty (approved):
LGTM!

PR #5140, opened 2023-11-07
feat(i18n) localize insufficient permissions message

ricokahler:
~~EDIT: Alright second try here. I opted to keep the interpolation value in the roles string but simplify the english phrase so that there is a clearer and stronger "action" clause. Additionally. I removed the roles from the phrase and move it into another that always displays it as a unit list. I think the result should translate better and has some nice consistencies. See new screenshot:~~

Edit edit: talked to @jtpetty who clarified and went with another approach that uses known context keys instead of the interpolation.

CleanShot 2023-11-14 at 16 13 34@2x

### DIFF
--- a/packages/sanity/src/core/components/InsufficientPermissionsMessage.tsx
+++ b/packages/sanity/src/core/components/InsufficientPermissionsMessage.tsx
@@ -2,24 +2,26 @@ import {AccessDeniedIcon} from '@sanity/icons'
 import {CurrentUser} from '@sanity/types'
 import {Text, Inline, Box} from '@sanity/ui'
 import React from 'react'
+// note: these are both available from the `../i18n` export but importing through
+// that export fails the build. may be due to a circular reference.
+import {useTranslation} from '../i18n/hooks/useTranslation'
+import {Translate} from '../i18n/Translate'
 
 /** @internal */
 export interface InsufficientPermissionsMessageProps {
-  title?: string
   operationLabel?: string
   currentUser?: CurrentUser | null
 }
 
 /** @internal */
 export function InsufficientPermissionsMessage(props: InsufficientPermissionsMessageProps) {
+  const {t} = useTranslation()
   const {
     currentUser,
-    title = 'Insufficient permissions',
-    operationLabel = 'access this feature',
+    operationLabel = t('insufficient-permissions-message.default-operation-label'),
   } = props
 
   const roles = currentUser?.roles || []
-  const plural = roles.length !== 1
 
   return (
     <Box padding={2}>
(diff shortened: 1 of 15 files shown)


bjoerge (approved):
LGTM

packages/sanity/src/core/components/InsufficientPermissionsMessage.tsx
```diff
@@ -2,47 +2,49 @@ import {AccessDeniedIcon} from '@sanity/icons'
...
-              Your role{plural && 's'}{' '}
-              {join(
-                roles.map((r) => <code key={r.name}>{r.title}</code>),
-                <>, </>,
-              )}{' '}
-              do{plural || 'es'} not have permissions to {operationLabel}
-            </>
-          )}
+          <Translate
+            i18nKey="insufficient-permissions-message.roles"
```
rexxars, line 37: To allow more control of the sentences, I would actually prefer if we broke this into separate translation keys instead of injecting `operationLabel` as a interpolation value. Yes, it will be a bit repetitive, but I think it is better to give translators full freedom of the sentence building.

rexxars (changes requested):

packages/sanity/src/core/components/InsufficientPermissionsMessage.tsx
ricokahler: sounds good. i was about to go that route but the change was bigger but i'll give it another go!

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1587,6 +1583,29 @@ export const studioLocaleStrings = {
...
+  /** --- Insufficient permissions message --- */
+
+  /** The title for the insufficient permissions message component */
+  'insufficient-permissions-message.title': 'Insufficient permissions',
+
+  /** Informs the users that their current roles do not have the required permissions to do a particular action */
+  'insufficient-permissions-message.not-authorized': 'Not authorized to {{action}}.',
+
+  /** Appears after the not-authorized message. Lists the current roles. */
+  'insufficient-permissions-message.roles': 'Your roles: <Roles/>',
```
jtpetty, line 1596: what if the user only has one Role?

packages/sanity/src/core/studio/components/navbar/new-document/NewDocumentButton.tsx
```diff
@@ -170,7 +170,7 @@ export function NewDocumentButton(props: NewDocumentButtonProps) {
     return (
       <InsufficientPermissionsMessage
         currentUser={currentUser}
-        operationLabel="create any document"
+        action={t('insufficient-permissions-message.action.create-any-document')}
```
jtpetty, line 173: I think a better approach here is to use the `operationLabel` to pass a context to the translation key

packages/sanity/src/core/i18n/bundles/studio.ts
ricokahler: i thought it would be okay to leave it plural. my thought process: our roles system is not mutually exclusive so it signifies to the user that they can have more than one role and may need to ask for an additional one

jtpetty (approved):
👍 LGTM

rexxars (approved):

PR #5149, opened 2023-11-07
fix(router): memoize `navigate` callback

mariuslundgard:
### Description

This change memoizes the `navigate` callback from the state router when using scoped routes.

Without this fix, depending on `navigate` in hooks could lead to infinite loops (e.g. in Presentation):

```ts
const {navigate} = useRouter()

const [viewParam, setViewParam] = useState(null)

useEffect(() => {
  // navigate because some param was changed
  navigate({
    // ...
  })
}, [navigate, viewParam])
```

### What to review

The changes to the code.

### Notes for release

- fix(router): memoize `navigate` callback


stipsan (approved):
All good 👍

PR #5150, opened 2023-11-07
feat(desk): add `onFocusPath` callback property

mariuslundgard:
### Description

This introduces a `onFocusPath` callback property on `DocumentPane`. This is needed by the new Presentation tool in order to subscribe to currently focused path within a document pane.

It also introduces an optimization to prevent unecessary `blur` events from triggering when the `path` is changed from the outside of `DocumentPane`.

### What to review

- Read the code changes.
- Consider what sorts of unit tests we might add for this.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
@@ -58,7 +58,7 @@ import {
  */
 // eslint-disable-next-line complexity, max-statements
 export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
-  const {children, index, pane, paneKey} = props
+  const {children, index, pane, paneKey, onFocusPath} = props
   const schema = useSchema()
   const templates = useTemplates()
   const {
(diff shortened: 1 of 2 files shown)


packages/sanity/src/desk/panes/document/DocumentPaneProvider.tsx
```diff
@@ -645,18 +662,33 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+        focusPathRef.current = pathFromUrl
+        onFocusPath?.(pathFromUrl)
+      }
+
+      setTimeout(() => {
+        disableBlurRef.current = false
+      }, 0)
+
       // remove the `path`-param from url after we have consumed it as the initial focus path
       paneRouter.setParams(restParams)
```
stipsan, line 689: To avoid race conditions in Strict Mode as well as Concurrent mode we should make sure we cancel the timer on unmount:

```suggestion
      const timeout = setTimeout(() => {
        disableBlurRef.current = false
      }, 0)

      // remove the `path`-param from url after we have consumed it as the initial focus path
      paneRouter.setParams(restParams)
      
      return () => clearTimeout(timeout)
```

stipsan (changes requested):
Looks solid, once the race condition is fixed I'll approve 👍

stipsan (approved):

PR #5152, opened 2023-11-08
fix(schema/types): fix block schema internal structure

skogsmaskin:
### Description

This follows up https://github.com/sanity-io/sanity/pull/5079. 

This will update some test snapshots that I don't quite understand why didn't fail in the original PR but failed when rebasing on current `next`.

This also fixes an issue with an invalid list of fields array for the block type. Not sure what happened to this during merge either, as tests were passing in the original PR.

I have also strengthened the code in `block-tools` to not rely on array order to find schema fields for the block type but rather to find them exclusively.

I have removed https://github.com/sanity-io/sanity/pull/5079 from `next` and included the merge commit in this new PR.

### What to review
* Are the changes to the Graphql schema OK? It is now correct, but maybe there's some logic attached to this inconsistency in the current Graphql code?

### DIFF
--- a/packages/@sanity/block-tools/test/tests/util/__snapshots__/blockContentTypeFeatures.test.ts.snap
+++ b/packages/@sanity/block-tools/test/tests/util/__snapshots__/blockContentTypeFeatures.test.ts.snap
@@ -391,7 +391,7 @@ Object {
               },
             },
             Object {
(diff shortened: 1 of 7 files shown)


rexxars (approved):
Looks good to me

PR #5153, opened 2023-11-08
fix(comments): truncate long field names in placeholder

sjelfull:
### Description

Fixes instances where long field names will cause the comment input placeholder to be wrapped over multiple lines, causing a overflow.

CleanShot 2023-11-08 at 14 51 24@2x

### What to review

- That long names is truncated and looks good in different scenarios
- That placeholders in normal PTE inputs outside of comments is properly position as before

### Notes for release

- Fixes instances where long field names in placeholders would cause overflow in comments input


hermanwikner (approved):
Thank you @sjelfull! 👍

robinpyon (approved):

PR #5157, opened 2023-11-08
feat(core): support weighted `template`, `mode` intent parameters

rexxars:
### Description

Attempts to solve the problem where you have multiple tools that can handle the same intent, and you want to weight them differently based on a parameter.

#### Requirements:
- One should be able to link from inside and outside of a studio to an edit intent, and give it a `mode` that is _better_ handled by one tool than another, but still be able to fall back if that "better" tool is not installed
- The currently active tool should be given priority if multiple tools are _equally_ good at handling the intent
- If the current active tool _can_ handle the intent, but a _better_ one is found, we should navigate to it

#### Example:
You have a "visual editing" tool and a "structured content" tool. They can both modify (edit) documents, but one is better at the schematic modification ("structured") and the other is better at the visual modifications ("visual").
- The structured content tool should be able to generate an intent link to a "visual edit" tool, and wise versa.
- If the structured content tool is open and you use the global search to find a document and open a result, the structured tool should open it, since it is the active tool. The same would be true for the visual editing tool.
  - If the global search included a preference for a "visual" editing mode, the visual editing tool should answer the intent, even if the structured content tool is open

#### Additional thoughts

I intentionally scoped down the parameters supported to the `template` param for the `create` intent and the `mode` param for the `edit` intent, since these are the ones I could see the most use for in the current landscape. 

In the future, it would be nice if the same (or a similar) approach could be used to score the intent based on other parameters. For instance:

- Two desk tools could argue about which one is best to handle an intent based on the structure definition (eg which one has the deepest/most specific point to edit it in)
- Two tools ("Desk" and "DealCreator") could argue about which one is the best to handle a create intent for a "deal" schema type.

I left this out of scope since it needs some more thinking on how these rankings would work - arbitrary score numbers? Still returned by `canHandleIntent` (seems oddly named if so)?
 
### What to review

- Does the approach solve the explained scenarios?
- Is this a forward-compatible approach?

### Notes for release

(Not sure if we need to feature this, as it is quite deep in tool/intent territory, but in case we do)

- Added the ability for tools to indicate that they can handle the `mode` and `template` parameters for intents, to give higher/lower priority to tools in certain scenarios

### DIFF
--- a/packages/sanity/src/core/config/types.ts
+++ b/packages/sanity/src/core/config/types.ts
@@ -165,12 +165,22 @@ export interface Tool<Options = any> {
   /**
    * Determines whether the tool can handle the given intent.
    *
+   * Can either return a boolean, or an object where the keys represent the parameters that
+   * can/can not be handled. This will be used to determine whether or not a tool is the best
+   * suited to handle an intent. Note that an object of only `false` values (or an empty object)
+   * is treated as `true`, so you want to explicitly return `false` if you know the intent cannot
+   * fulfill the intent request.
+   *
    * @param intent - The intent to check.
    * @param params - The parameters for the intent.
    * @param payload - The payload for the intent.
-   * @returns `true` if the tool can handle the intent, `false` otherwise.
+   * @returns Boolean: whether it can handle the intent. Object: Values representing what specific parameters can be handled.
    */
-  canHandleIntent?: (intent: string, params: Record<string, unknown>, payload: unknown) => boolean
+  canHandleIntent?: (
+    intent: string,
+    params: Record<string, unknown>,
+    payload: unknown,
+  ) => boolean | {[key: string]: boolean}
 }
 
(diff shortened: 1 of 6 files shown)


jtpetty:
Do we need to change https://github.com/sanity-io/sanity/blob/65eb8c2bfbca78bc1ed5443e55dd629b1f233c00/packages/sanity/src/desk/structureResolvers/resolveIntent.ts#L117 as well?

rexxars:
> Do we need to change
> 
> https://github.com/sanity-io/sanity/blob/65eb8c2bfbca78bc1ed5443e55dd629b1f233c00/packages/sanity/src/desk/structureResolvers/resolveIntent.ts#L117
> 
> as well?

No, the pane routing shouldn't care about these parameters.

mariuslundgard (approved):
LGTM!

PR #5160, opened 2023-11-09
fix(form): open hotspot editor using `focusPath`

mariuslundgard:
### Description

Prior to this change, the hotspot editor has been opened using `onFieldOpen`. This caused an issue with the hotspot editor not being able to maintain its open state when used in the Presentation tool.

This change involves using `onPathFocus` instead, which seems to be more idiomatic. This also means the hotspot editor is deep-linkable.

### What to review

- The code changes.

### Notes for release

- n/a


stipsan (approved):
Excellent :D

stipsan:
> > This caused an issue with the hotspot editor not being able to maintain its open state when used in the Presentation tool.
> 
> @mariuslundgard Can you elaborate on this? What does "not being able to maintain its open state" mean specifically? Just want to make sure we fix the right thing.

Sent you a repro link over DM that demonstrates the issue

PR #5172, opened 2023-11-10
feat(router): add support for tools to bypass search param scoping

bjoerge:
### Description
This task enables a tool to declare that it fully "owns" and controls the search params, as a way to conditionally opt-out of the search parameter scoping, resulting in cleaner urls (see #5148 for backround info about search params support).

Declaring this on the tool level makes it possible to detect if there are multiple tools that declare they own the search params, and enables us to add safeguards/dev warnings if we want to do this later.

### What to review
Whether the change makes sense. I'm not too satisfied with having to add special handling in so many different places, so ideally we should look into refactoring this behavior into core router utilities.

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/studio/router/helpers.ts
+++ b/packages/sanity/src/core/studio/router/helpers.ts
@@ -74,21 +74,29 @@ export function resolveIntentState(
   )
 
   if (matchingTool?.getIntentState) {
-    const toolState = matchingTool.getIntentState(
+    const _toolState = matchingTool.getIntentState(
       intent,
       params as any,
       prevState && (prevState[matchingTool.name] as any),
       payload,
-    )
+    ) as Record<string, unknown>
 
+    const {_searchParams, ...toolState} = _toolState
+
+    const nextUrlState: Record<string, unknown> = {
+      ...prevState,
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/studio/router/helpers.ts
```diff
@@ -74,21 +74,29 @@ export function resolveIntentState(
...
+
+    const nextUrlState: Record<string, unknown> = {
+      ...prevState,
+      tool: matchingTool.name,
+      [matchingTool.name]: toolState,
+    }
+    if (matchingTool.router?.__unsafe_disableScopedSearchParams) {
+      nextUrlState._searchParams = _searchParams
+    } else {
+      toolState._searchParams = _searchParams
```
skogsmaskin, line 94: Would this assignment be effective here?

bjoerge: It will! Because it's being assigned to `nextUrlState[matchingTool.name]` on line 89 above

skogsmaskin (approved):
LGTM!

mariuslundgard (approved):
👍

PR #5173, opened 2023-11-10
fix(router): improve search params encoding

bjoerge:
### Description
This adds support for skipping certain characters when encoding urls using sanity/router. This results in more readable urls. For now, only brackets and forward slashes are preserved, so instead of encoding the url like this
```
/section/foobar?tool%5Bpage%5D=%2Fproducts
```
it will be encoded like this
```
/section/foobar?tool[page]=/products
```

### What to review
Does it make sense? Any unintended consequences of this that I'm forgetting?

### Notes for release

N/A


skogsmaskin (approved):
LGTM!

mariuslundgard (approved):
LGTM!

PR #5179, opened 2023-11-13
fix(comments): improve scroll behavior

hermanwikner:
### Description

This pull request aims to stabilize and improve the scroll behavior in comments. The scroll functionality is intended for the following scenarios:

- When clicking on the comment button with existing comments, the corresponding comment thread item should scroll into view in the inspector and be highlighted.
- When clicking on a comment thread item in the inspector, the related comment item should be highlighted, and the corresponding field in the form should scroll into view and be highlighted.
- When utilizing the "Copy link to comment" feature, the comment should be scrolled to in the inspector when navigating to the URL

### What to review

- Ensure that the scroll functionality works correctly in various scenarios.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/desk/comments/plugin/field/CommentField.tsx
+++ b/packages/sanity/src/desk/comments/plugin/field/CommentField.tsx
@@ -48,7 +48,6 @@ export function CommentField(props: FieldProps) {
 const SCROLL_INTO_VIEW_OPTIONS: ScrollIntoViewOptions = {
   behavior: 'smooth',
   block: 'start',
-  inline: 'nearest',
 }
 
 const HighlightDiv = styled(motion.div)(({theme}) => {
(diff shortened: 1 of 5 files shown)


skogsmaskin (approved):
Tested it. Works well! 👍

robinpyon (approved):
Thanks @hermanwikner!

FWIW the listed scenarios were already working for me on `next` – is there perhaps a specific failing test case I can try repro? Otherwise works and looks great!

hermanwikner:
> FWIW the listed scenarios were already working for me on `next` – is there perhaps a specific failing test case I can try repro? Otherwise works and looks great!

Thanks, @robinpyon!

This pull request fixes an issue where scrolling to a comment from the URL in the inspector occasionally got interrupted. This occurred when trying to scroll to both a comment and a form field at the same time. With this PR, the scroll now prioritizes the comment item when navigating to a URL with a comment parameter (i.e., origin: 'url'), rather than focusing on the form field.

Additionally, I've refined the overall scrolling behavior in this PR. While scrolling to comments generally behaved as expected, there were instances where it didn't work consistently as intended (e.g, did not always scroll all the way down to a comment item, etcetera).

PR #5181, opened 2023-11-13
feat: rename desk => structure in public facing UI

bjoerge:
### Description

This renames the public facing part of the "Desk"-tool to "Structure". For now, only labels and default tool name is changed. This PR also adds a notice about the rename that will appear in the nav bar unless desk tool has been given a custom name.

**Onboarding tooltip horizontal nav bar**
image

**Onboarding tooltip vertical nav bar (narrow screens)**
image

### What to review
Unless desk-tool has been given a custom name:
- Make sure navigating to `/desk` redirects to `/structure`
- Make sure the onboarding message appears on desk tool menu button
- Make sure the tooltip links to the blog post announcement (blog post TBD)
- Make sure the tooltip can be dismissed and don't appear again after reloading

Note: to make the dialog re-appear, remove the localstorage key `sanityStudio:desk:renameDismissed` and reload page

```js
// paste into console to remove the key
localStorage.removeItem('sanityStudio:desk:renameDismissed')
```

### Notes for release

TBD


packages/sanity/src/core/studio/StudioLayout.tsx
```diff
@@ -113,13 +115,35 @@ export function StudioLayout() {
 
   const Navbar = useNavbarComponent()
 
+  /**
+   * Handle legacy URL redirects from `/desk` to `/structure`
+   */
+  const isLegacyDeskRedirect =
```
jtpetty, line 121: I think this makes sense. I think the only corner case I can think of is if a customer had a custom tool named 'structure' and so they explicitly set the deskTool name to 'desk', this will cause an unwanted redirect, but maybe that is not likely.

jtpetty (approved):
LGTM

packages/sanity/src/core/studio/StudioLayout.tsx
bjoerge: Good point! I just tested, but it seems to work as expected 🤔

runeb: Does the redirect kick in if someone purposefully names their structure tool `desk` because they want `/desk` to be the path?

bjoerge: It doesn't! You can explicity set `name: 'desk', title: 'Desk'`to get the old behavior back.

rexxars (approved):

PR #5184, opened 2023-11-14
feat(facelift): create CollapseTabList for redesigned navbar

pedrobonamin:
### Description

Adds `<CollapseTabList />` component, which works similar to ColapseMenu but instead of removing the text it hides the elements that don't fit and renders them into a `<Menu />`

https://test-studio-git-edx-742.sanity.build/test/workshop/core;components;collapseTabList;collapse-tab-list?scheme=light

Screenshot 2023-11-14 at 08 54 39
Screenshot 2023-11-14 at 08 53 59

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
@@ -160,7 +160,7 @@ export function StudioNavbar() {
       <RootCard borderBottom data-testid="navbar" data-ui="Navbar" padding={2} sizing="border">
         {/** Left flex */}
         <Flex align="center" justify="space-between">
-          <Flex align="center" gap={2}>
+          <Flex align="center" gap={2} style={{flexShrink: 0}}>
             <Flex align="center" gap={1}>
               {/* Menu button */}
               {!shouldRender.tools && (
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
```diff
@@ -160,7 +160,7 @@ export function StudioNavbar() {
       <RootCard borderBottom data-testid="navbar" data-ui="Navbar" padding={2} sizing="border">
         {/** Left flex */}
         <Flex align="center" justify="space-between">
-          <Flex align="center" gap={2}>
+          <Flex align="center" gap={2} style={{flexShrink: 0}}>
```
ninaandal, line 163: Should we move this to a styled component for consistency?

robinpyon: Yes! In general we try and avoid inline styles 

Though we'll likely want to move to a CSS grid approach for the navbar as we want the tool menu to be horizontally centered in the viewport regardless of the width of either the LHS or RHS sections. We can address this separately!

robinpyon (approved):
Thanks @pedrobonamin – the intention behind naming this `<CollapseTabList>` was that it would be a `<TabList>` component that supports nested `<Tab>` components.

We may need to revisit this once we get confirmation on final navbar layout. If these are in fact to be rendered as tabs, we'll want to make the changes above – otherwise, we'll may want to consider renaming this.

PR #5185, opened 2023-11-14
refactor(core): improve navbar collapse menu

bjoerge:
### Description
This changes the approach we use to check whether there's enough space to show menu options as expanded or collapsed or whether we need to move menu options to the menu:
1. We keep two "shadow" (hidden) navbars: One is rendering all items as expanded and the other is rendering all items as collapsed.
2. For each of these "shadow" navbars, we keep track of intersections for each rendered item in a dictionary on the format `Record<element.key, {intersecting: boolean, element: ReactElement}>`
3. This enables us to easily check whether we have received the intersections for all the current menu options, which again enables us to defer the first render until we have got all the intersection data we need.

This should also pave the way for future work that let's us track the individual widths (bounding boxes) of each menu element, and improve the navbar by gradually collapsing items from the right.

### What to review
Does the changes look ok?
I've done quite extensive testing in all major browsers (Chrome, Safari, Firefox) and it seems to work great, but there might be edge cases I don't know about here, so please test the cases you know.

### Notes for release

- Improved initial rendering of the Studio navigation toolbar

### DIFF
--- a/packages/sanity/src/core/components/collapseMenu/CollapseMenu.tsx
+++ b/packages/sanity/src/core/components/collapseMenu/CollapseMenu.tsx
@@ -1,10 +1,14 @@
 import {EllipsisVerticalIcon} from '@sanity/icons'
 import {Box, Button, Flex, MenuButtonProps, Text, Tooltip} from '@sanity/ui'
 import React, {
+  Children,
   cloneElement,
+  ForwardedRef,
   forwardRef,
+  Fragment,
   memo,
   ReactElement,
+  ReactNode,
   useCallback,
   useMemo,
   useState,
@@ -17,14 +21,14 @@ import {CollapseMenuDivider} from './CollapseMenuDivider'
 
 /** @internal */
 export interface CollapseMenuProps {
-  children: React.ReactNode
+  children: ReactNode | ReactNode[]
   collapsed?: boolean
   collapseText?: boolean
   disableRestoreFocusOnClose?: boolean
   gap?: number | number[]
   menuButtonProps?: Omit<MenuButtonProps, 'id' | 'menu' | 'button'> & {
     id?: string
-    button?: React.ReactElement
+    button?: ReactElement
   }
   onMenuClose?: () => void
 }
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/components/collapseMenu/CollapseMenu.tsx
```diff
@@ -209,45 +297,74 @@ export const CollapseMenu = forwardRef(function CollapseMenu(
...
                         'aria-hidden': hidden,
                       })}
                     </Flex>
                   </Tooltip>
+                </React.Fragment>
+              )
+            })}
+        </RowFlex>
+        {/* Rendered hidden in order to calculate intersections for expanded menu options */}
+        <RowFlex data-hidden aria-hidden="true" gap={gap} overflow="hidden">
```
hermanwikner, line 311: I suggest unifying the rendering of the hidden rows in a function called e.g `renderHiddenRow` that receives an array of items (`React.ReactElement[]`) and an `onIntersectionChange` callback. 

```ts
  const renderHiddenRow = useCallback(
    (
      items: ReactElement[],
      onIntersectionChange: (e: IntersectionObserverEntry, child: React.ReactElement) => void,
    ) => {
      return (
        <RowFlex data-hidden aria-hidden="true" gap={gap} overflow="hidden">
          {items.map((child, index) => {
            const {dividerBefore} = child.props
            return (
              <React.Fragment key={child.key}>
                {dividerBefore && index !== 0 && <CollapseMenuDivider hidden />}

                <OptionObserveElement
                  options={intersectionOptions}
                  // eslint-disable-next-line react/jsx-no-bind
                  onIntersectionChange={(e) => onIntersectionChange(e[0], child)}
                >
                  <Flex>
                    {cloneElement(child, {
                      disabled: true,
                      'aria-hidden': true,
                    })}
                  </Flex>
                </OptionObserveElement>
              </React.Fragment>
            )
          })}
        </RowFlex>
      )
    },
    [gap, intersectionOptions],
  )

```

And then do this in the JSX:

```ts
function AutoCollapseMenu() {
  // ... rest of code

  const renderHiddenRow = useCallback(() => {
    // Return the hidden row
  }, [])

  return (
    <OuterFlex>
      <RootFlex>
        <RowFlex>
          {/* The visible row here */}

          {/* Rendered hidden in order to calculate intersections for expanded menu options */}
          {renderHiddenRow(menuOptions, handleExpandedIntersection)}
          
          {/* Rendered hidden in order to calculate intersections for collapsed menu options */}
          {renderHiddenRow(collapsedElements, handleCollapsedIntersection)}
        </RowFlex>
      </RootFlex>
    </OuterFlex>
  )
}
```

hermanwikner (approved):
Thanks @bjoerge! 🙏 

I have a suggestion regarding the rendering of hidden rows. However, this is not a blocker for merging!

packages/sanity/src/core/components/collapseMenu/CollapseMenu.tsx
bjoerge: Thanks @hermanwikner. That's is a great suggestion! I implemented a slight modification of it (using a component instead of inline function). See ca6f97f67d21ea756b5f426b3e1de4dfaab387ac

PR #5186, opened 2023-11-14
fix(facelift): add wrapper to publish action button

pedrobonamin:
### Description
Adds wrapper to Publish Action button, it's needed to render the Tooltip correctly.
The button was the direct child of the tooltip, so the modified onMouseEnter action was passed to that button, given the button is disabled, the onMouseEnter action won't be triggered.


robinpyon (approved):
Thanks @pedrobonamin!

PR #5199, opened 2023-11-15
chore(ci): turn off "require dashboard approval" renovatebot req

stipsan:
### Description

Enables Renovatebot to open PRs without requiring approving the PR on https://github.com/sanity-io/sanity/issues/4030


jtpetty:
@stipsan - I'm not sure I fully understand what this PR does.

binoy14:
Have a few concerns regarding this

1. This would create a lot of PRs hogging vercel builds org wide. Can we limit the number of PRs it creates 
2. Does it make sense to change the rebase strategy to only do it when conflicted instead of behind base so avoid the above issue
3. We need first fix the dataset creation issue with cli tests before introducing this.

PR #5201, opened 2023-11-15
feat(facelift): force icon-only buttons to use a tooltip

pedrobonamin:
### Description

- Updates SanityUI Button component to force icon-only buttons to define a tooltip, forcing consistency.
- Refactors the components that are using the button to include a tooltip.
- Adds `TooltipDelayGroupProvider` to buttons that are part of the same menu, allowing to jump between tooltips fast.

https://github.com/sanity-io/sanity/assets/46196328/a531b460-1e30-434a-bba1-d2e4e372c187

- Pending to refactor:  https://github.com/sanity-io/sanity/compare/feat/facelift-mvi-2...EDX-758/button-with-tooltip?expand=1#diff-80454173037198dc078017b9fb13971b2f8b43f4d95c93007924964233c1367a

### What to review
Uses of buttons in general inside the studio.


packages/sanity/src/core/changeIndicators/ElementWithChangeBar.tsx
```diff
@@ -25,15 +26,16 @@ export function ElementWithChangeBar(props: {
...
-
-          <ChangeBarButton
-            aria-label="Review changes"
-            data-testid="change-bar__button"
-            onClick={isReviewChangesOpen ? undefined : onOpenReviewChanges}
-            tabIndex={-1}
-            type="button"
-            $withHoverEffect={withHoverEffect}
-          />
+          <Tooltip content="Review changes" portal>
```
pedrobonamin, line 29: This is an special button, not using the StudioUI Button, added tooltip for consistency.

packages/sanity/src/core/studio/components/navbar/tools/ToolCollapseMenu.tsx
```diff
@@ -41,7 +41,7 @@ export function ToolCollapseMenu(props: ToolCollapseMenuProps) {
   const children = useMemo(
     () =>
       tools.map((tool, index) => {
-        const title = tool?.title || startCase(tool.name) || undefined
+        const title = tool?.title || startCase(tool.name)
```
pedrobonamin, line 44: This should not return an undefined.

packages/sanity/src/desk/comments/plugin/field/CommentFieldButton.tsx
```diff
@@ -139,37 +147,42 @@ export function CommentFieldButton(props: CommentFieldButtonProps) {
...
   return (
-    <Tooltip
-      portal
-      placement="top"
-      content={`View comment${count > 1 ? 's' : ''}`}
-      fallbackPlacements={['bottom']}
-    >
-      <Button aria-label="Open comments" mode="bleed" onClick={onClick} size="small">
+    <Tooltip portal placement="top" content={`View comment${count > 1 ? 's' : ''}`}>
+      <SanityUIButton
```
pedrobonamin, line 172: This button has a "special" way of handling children, couldn't reuse the Studio UI Button component

packages/sanity/src/desk/comments/src/components/list/CommentsListItem.tsx
```diff
@@ -61,7 +61,7 @@ const ExpandButton = styled(Button)(({theme}) => {
   `
 })
 
-const GhostButton = styled(Button)`
+const GhostButton = styled.button`
```
pedrobonamin, line 64: This is just a hidden button, no need to use the StudioUI button

packages/sanity/src/desk/components/paneHeaderActions/PaneHeaderCreateButton.tsx
```diff
@@ -87,6 +87,8 @@ export function PaneHeaderCreateButton({templateItems}: PaneHeaderCreateButtonPr
           mode="bleed"
           disabled
           data-testid="action-intent-button"
+          // This button handles the tooltip in a special way, won't reuse the forced tooltip.
```
pedrobonamin, line 90: This buttons couldn't be refactored to the new tooltip, as they have a special behaviour for the tooltip content in https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/components/InsufficientPermissionsMessage.tsx

packages/sanity/src/core/form/inputs/arrays/common/DragHandle.tsx
```diff
@@ -10,21 +10,25 @@ const DragHandleButton = styled(Button)<{grid?: boolean}>`
...
+export const DragHandle = function DragHandle(props: DragHandleProps) {
   const id = useContext(SortableItemIdContext)!
   const {listeners, attributes} = useSortable({id})
+  const {mode = 'bleed', ...rest} = props
 
   return (
     <DragHandleButton
       icon={DragHandleIcon}
-      mode="bleed"
+      tooltipProps={{content: 'Drag to reorder'}}
```
robinpyon, line 27: Despite this technically being an icon-only button, adding a tooltip here can be especially distracting when you're trying to use it – suggest that we remove the tooltip here.

packages/sanity/src/core/form/inputs/files/FileInput/FileActionsMenu.tsx
```diff
@@ -132,6 +132,7 @@ export function FileActionsMenu(props: Props) {
               mode="bleed"
               onClick={handleClick}
               ref={setOptionsButtonRef}
+              tooltipProps={{content: 'Open file options menu'}}
```
robinpyon, line 135:
```suggestion
              tooltipProps={{content: 'Show more'}}
```

If only to be consistent with other context menu buttons

packages/sanity/src/core/form/inputs/files/ImageInput/ImageActionsMenu.tsx
```diff
@@ -82,34 +89,38 @@ export function ImageActionsMenu(props: ImageActionsMenuProps) {
...
-          aria-label="Open image edit dialog"
-          data-testid="options-menu-edit-details"
-          icon={CropIcon}
-          mode="ghost"
-          onClick={onEdit}
-          ref={setHotspotButtonElement}
-        />
-      )}
-      {/* Using a customized Popover instead of MenuButton because a MenuButton will close on click
+      <TooltipDelayGroupProvider delay={{open: 400}}>
```
robinpyon, line 92: The delay used for these delay providers should be consistent across the studio – suggest that we avoid a magic number here and define a shared constant

packages/sanity/src/core/form/inputs/files/ImageInput/ImageActionsMenu.tsx
robinpyon, line 1: Please check the padding of image action buttons, as they should not be butted up against each other

packages/sanity/src/core/form/inputs/files/common/FileInputButton/FileInputButton.tsx
```diff
@@ -2,7 +2,7 @@ import React, {useId} from 'react'
 import {ButtonProps} from '../../../../../../ui'
 import {FileButton} from './styles'
 
-export interface FileInputButtonProps extends ButtonProps {
+export type FileInputButtonProps = ButtonProps & {
```
robinpyon, line 5: Possibly an unnecessary change – please [prefer interfaces over types](https://www.notion.so/sanityio/Studio-Development-Guidelines-f48f9a9fe62843018ccd16c043c238cf?pvs=4#0240ed3590d8483ab8fabc1081ac5e97)!

packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
```diff
@@ -225,43 +227,45 @@ export function StudioNavbar() {
...
-                  />
-                </BoundaryElementProvider>
-              </SearchProvider>
-            </LayerProvider>
-
-            {/* Search button (desktop) */}
-            {!shouldRender.searchFullscreen && (
-              <SearchButton onClick={handleOpenSearch} ref={setSearchOpenButtonEl} />
-            )}
+          <TooltipDelayGroupProvider delay={{open: 400}}>
```
robinpyon, line 230: See above re: magic numbers and shared delay values

packages/sanity/src/core/studio/components/navbar/search/components/SearchHeader.tsx
```diff
@@ -89,7 +89,13 @@ export const SearchHeader = forwardRef<HTMLInputElement, SearchHeaderProps>(func
         {/* (Fullscreen) Close button */}
         {fullscreen && (
           <Card>
-            <Button aria-label="Close search" icon={ArrowLeftIcon} mode="bleed" onClick={onClose} />
+            <Button
+              aria-label="Close search"
+              icon={ArrowLeftIcon}
+              mode="bleed"
+              onClick={onClose}
+              tooltipProps={{content: 'Close search'}}
```
robinpyon, line 97: This tooltip currently doesn't render correctly, but may be addressed once recent tooltip changes are merged into Sanity UI `alpha` – something to keep an eye on!

packages/sanity/src/desk/comments/src/components/list/CommentsListItemContextMenu.tsx
robinpyon, line 1: I believe there's been a possible visual regression here? Note inconsistent padding around these buttons:

image

packages/sanity/src/desk/panes/document/statusBar/sparkline/PublishStatus.tsx
```diff
@@ -69,13 +69,8 @@ export function PublishStatus(props: PublishStatusProps) {
...
           icon={liveEdit ? PlayIcon : PublishIcon}
-          text={
-            collapsed ? null : (
-              <abbr aria-label={a11yLabel}>
-                {liveEdit && lastUpdated ? lastUpdatedTime : lastPublishedTime}
-              </abbr>
-            )
-          }
+          // eslint-disable-next-line no-nested-ternary
+          text={collapsed ? '' : liveEdit && lastUpdated ? lastUpdatedTime : lastPublishedTime}
```
robinpyon, line 73: Can this be rewritten in a way to avoid a nested ternary?

packages/sanity/src/desk/panes/documentList/DocumentListPaneHeader.tsx
```diff
@@ -57,7 +57,15 @@ export const DocumentListPaneHeader = memo(
...
         backButton={
           features.backButton &&
-          index > 0 && <Button as={BackLink} data-as="a" icon={ArrowLeftIcon} mode="bleed" />
+          index > 0 && (
+            <Button
+              as={BackLink}
+              data-as="a"
+              icon={ArrowLeftIcon}
+              mode="bleed"
+              tooltipProps={{content: 'Back'}}
```
robinpyon, line 66: This currently doesn't render correctly, please verify once the `alpha` branch of Sanity UI is up to date

packages/sanity/src/desk/panes/list/ListPaneHeader.tsx
```diff
@@ -23,7 +23,15 @@ export const ListPaneHeader = ({index, menuItems, menuItemGroups, title}: ListPa
...
       backButton={
         features.backButton &&
-        index > 0 && <Button as={BackLink} data-as="a" icon={ArrowLeftIcon} mode="bleed" />
+        index > 0 && (
+          <Button
+            as={BackLink}
+            data-as="a"
+            icon={ArrowLeftIcon}
+            mode="bleed"
+            tooltipProps={{content: 'Back'}}
```
robinpyon, line 32: As above!

packages/sanity/src/ui/button/Button.tsx
```diff
@@ -41,10 +58,35 @@ export const Button = forwardRef(function Button(
...
+  }: ButtonProps & Omit<React.HTMLProps<HTMLButtonElement>, 'as' | 'size' | 'title'>,
   ref: React.ForwardedRef<HTMLButtonElement>,
 ) {
+  if (tooltipProps) {
+    return (
+      <Tooltip
+        // Force the tooltip to render in portal
+        portal
+        {...tooltipProps}
+      >
```
robinpyon, line 72: Once those tooltip changes are in Sanity UI `alpha`, let's verify if we can force all tooltips to render in portals without boundary elements and consider simplifying exposed `tooltipProps`

robinpyon (commented):
Thanks @pedrobonamin - massive effort! 

Left a bunch of comments below – let's aim to re-validate once recent tooltip updates are brought into the `alpha` branch of Sanity UI

packages/sanity/src/ui/button/Button.tsx
pedrobonamin: Yes, I think it makes sense!

packages/sanity/src/core/form/components/tagInput/tagInput.tsx
```diff
@@ -304,7 +304,13 @@ function Tag(props: {
...
         {enabled && (
           <Box marginLeft={1}>
-            <Button icon={CloseIcon} mode="bleed" onClick={handleRemoveClick} size="small" />
+            <Button
+              icon={CloseIcon}
+              mode="bleed"
+              onClick={handleRemoveClick}
+              size="small"
+              tooltipProps={{content: 'Remove'}}
+            />
```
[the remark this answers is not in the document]
pedrobonamin: Thinking on a proposal, updated a bit the component structure to make it work with the tooltip.
It changes the space the button takes but keeps the tag in the same size, allowing to reuse the same button. Seems redundant but consistent. 

Screenshot 2023-11-16 at 08 56 19
Screenshot 2023-11-16 at 08 56 10

Let me know what you think, haven't pushed this.

packages/sanity/src/core/form/field/actions/FieldActionMenu.tsx
```diff
@@ -13,8 +13,9 @@ export interface FieldActionMenuProps {
   onMenuOpenChange: (open: boolean) => void
 }
 
-const STATUS_BUTTON_TOOLTIP_PROPS: StatusButtonProps['tooltip'] = {
+const STATUS_BUTTON_TOOLTIP_PROPS: StatusButtonProps['tooltipProps'] = {
   placement: 'top',
+  portal: true,
 }
```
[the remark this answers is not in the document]
pedrobonamin: I've added `portal:true` as default, will remove it from here.
Now I am thinking if the default for the studio tooltips should be `bottom` or `top`, in SanityUI the default is `bottom`. Let's validate it

packages/sanity/src/core/form/inputs/DateInputs/base/calendar/Calendar.tsx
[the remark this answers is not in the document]
pedrobonamin: I'm updating them to be larger and have the same size as **pre facelift**, given the more appealing look is to have the same height as the select elements.

packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx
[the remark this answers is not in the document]
pedrobonamin: Seems the `<Inline>` element is making tooltips position in a incorrect place. 
Changing it to `<Flex>` and positioning the tooltip in the top makes it look a better, still not 100% convinced on them. Let me know what you think. 
Screenshot 2023-11-16 at 09 36 09

packages/sanity/src/core/form/inputs/arrays/common/DragHandle.tsx
pedrobonamin: What do you think about modifying the IconOnly button ts definition to accept:
`  tooltipProps: TooltipProps | 'ICON_BUTTON_NO_TOOLTIP'`
So,  in the icon only buttons that we don't want a tooltip we can use that to disable the tooltip.

```
    <DragHandleButton
      tooltipProps="ICON_BUTTON_NO_TOOLTIP" // Avoid using tooltip in this button.
      icon={DragHandleIcon}
      mode={mode}
      data-ui="DragHandleButton"
      {...attributes}
      {...rest}
      {...listeners}
    />
```

packages/sanity/src/core/form/inputs/files/common/FileInputButton/FileInputButton.tsx
pedrobonamin: We are going to need it in this case, with the change of the button props to be an intersection it can be extended, because it won't know which of the two intersections to extend.
```
An interface can only extend an object type or intersection of object types with statically known members.ts(2312)
```

packages/sanity/src/desk/comments/src/components/list/CommentsListItemContextMenu.tsx
pedrobonamin: Good catch. yes it's generated by the span wrapping the elements when a tooltip is added to the buttons. 
Updated in

packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx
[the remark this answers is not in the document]
robinpyon: Positioning is much better! Though I share your opinion – not super convinced here! I think the tooltip being as large as the popover is the problem here. 

For now let's disable the tooltip and capture this with design (there's also the issue of this using a dark scheme, which it probably shouldn't)

packages/sanity/src/desk/panes/document/statusBar/sparkline/PublishStatus.tsx
pedrobonamin: `text={collapsed ? '' : `${liveEdit && lastUpdated ? lastUpdatedTime : lastPublishedTime}`}`
Just joking :smile

packages/sanity/src/core/form/inputs/arrays/common/DragHandle.tsx
robinpyon: I'm not too fond of the above approach, as i wouldn't intuitively expect `tooltipProps` to receive a string to disable it

There is value to having a tooltip for these drag handles (and in future, it's very possible that clicking these will invoke a contextual menu too). 

What might be a simpler solve is if we override `delay` in this instance and set it to something high `delay: { open: 1000 }`. WDYT?

packages/sanity/src/core/form/components/tagInput/tagInput.tsx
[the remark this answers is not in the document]
robinpyon: 👍 if it looks visually consistent with what we have currently, and is correctly aligned when users start typing

See my comment below on drag handles – re: tooltip omission, i may have spoken too hastily here! We could potentially just increase the delay in this instance to reduce noise here. WDYT?

packages/sanity/src/core/form/inputs/arrays/common/DragHandle.tsx
pedrobonamin: We could add that delay but I think we should still provide a safe opt-out from the tooltip, but making it explicit.
Maybe ` tooltipProps: TooltipProps | null ` ?

pedrobonamin: Given that the developer should have a good reason to opt out from the tooltip in the case of icon-only buttons

robinpyon: That's fair – I think `tooltipProps: TooltipProps | null` works!

packages/sanity/src/desk/panes/documentList/DocumentListPaneHeader.tsx
pedrobonamin: Looks correct (positioned at the bottom) after last fallback placements pr update

robinpyon (approved):
Thanks @pedrobonamin!

PR #5203, opened 2023-11-15
fix(comments): context menu layer issue

hermanwikner:
### Description

This pull request fixes a layer issue, as can be seen in the screenshot. The solution here is to render the popover in the `MenuButton` in a portal, and we rely on the `isTopLayer` boolean from `useLayer` to determine whether we should deselect the selected comment in `CommentsInspector`.

This pull request also makes sure that the menu of the root comment is always visible when hovering over the thread item card.

Screenshot 2023-11-15 at 17 05 12

### What to review

- Make sure that the hover styles are correctly applied to the thread item card
- Make sure that the menu of root comment is always visible when hovering over the thread item card

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/desk/comments/plugin/inspector/CommentsInspector.tsx
+++ b/packages/sanity/src/desk/comments/plugin/inspector/CommentsInspector.tsx
@@ -1,6 +1,7 @@
-import {Flex, useClickOutside, useToast} from '@sanity/ui'
+import {Flex, Layer, useClickOutside, useLayer, useToast} from '@sanity/ui'
 import React, {Fragment, useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import * as PathUtils from '@sanity/util/paths'
+import styled from 'styled-components'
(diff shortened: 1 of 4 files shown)


robinpyon (dismissed):

skogsmaskin:
I was testing this, and I could not reproduce the issue as described in the PR description.

However, I found something else (not very critical though). If you click open the comment context menu and then click the menu button again to close it, it becomes "stuck" in the visible state even after hovering/de-hovering it.

image

hermanwikner:
> I was testing this, and I could not reproduce the issue as described in the PR description.
> 
> However, I found something else (not very critical though). If you click open the comment context menu and then click the menu button again to close it, it becomes "stuck" in the visible state even after hovering/de-hovering it.
> 
> image

Thanks, @skogsmaskin – good catch! The button is visible when it has focus for keyboard navigation, and clicking it keeps it focused. I will look into it! 👍

robinpyon (approved):

PR #5205, opened 2023-11-15
fix(form/inputs): optimize rendering of member inputs

skogsmaskin:
### Description

There were some issues here with member form inputs not being re-rendered properly when rendered inline in the editor
This makes sure we re-render those every time we need to (members change).

This can now be achieved by object equality checks after https://github.com/sanity-io/sanity/commit/a0f93cba7430a4584d01256a55e078d52a05c459

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -13,7 +13,6 @@ import React, {
   useCallback,
   useRef,
   useImperativeHandle,
-  createRef,
   ReactNode,
   startTransition,
 } from 'react'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -194,16 +203,28 @@ export function PortableTextInput(props: PortableTextInputProps) {
       const isObject = item.kind !== 'textBlock'
       let input: ReactNode
 
-      // Only render the input if the item is open or new
-      if (isObject && (item.member.open || !existingItem)) {
-        const inputProps = {...(props as FIXME), absolutePath: item.node.path}
-        input = <FormInput {...inputProps} />
+      if (isObject && (!existingItem || !isEqual(item.member, existingItem?.member))) {
```
bjoerge, line 206: `isEqual(item.member, existingItem?.member)` will do a deep equality check - is that intentional? Given that these has (or at least _should_) already have gone through a reconciliation step that preserves object identity, a shallow equal should suffice, no?

skogsmaskin: Doesn't seem like it. I don't get the preferred result with a shallow comparison of the members.
It works with shallow if I only compare the value. Probably validation, presence, or similar objects perhaps?

skogsmaskin: Works great now, I can compare object equality!

skogsmaskin:
Moved that horrendous function into a hook instead of having it in the `PortableTextInput` component itself.

bjoerge (approved):

bjoerge:
LGTM!

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
bjoerge: Feels great to get to the bottom of that one 😮‍💨

PR #5208, opened 2023-11-15
fix: use event.pull_request.number instead

ricokahler:
### Description

It seems like we've been creating too many datasets due to usage of `github.event.number`. It seems like the intent was to use `github.event.pull_request.number` instead.

### What to review

Take a look at the logs when the dataset is created. You should see the same PR dataset used on every run. Additionally when this PR is closed, the dataset should be deleted as well.

### Notes for release

N/A (internal only)


ricokahler:
i'm not sure this is actually doing anything

PR #5209, opened 2023-11-15
feat(desk): use localized date time format for timeline

rexxars:
### Description

The items in the review changes timeline had a timestamp that was always displayed in US format.
This PR changes the items to instead use the format native to the chosen locale. When using `en-US`, the format is just a tiny bit different in that it does not pad hours with leading zeros. Apart from that, it remains the same.

Examples of formatting:
- `en-US`: `Oct 16, 2023, 2:17 PM`
- `no-NB`: `16. okt. 2023, 14:17`

In addition to the formatting, I also changed from using a `span` element to using `time` with a `datetime` attribute, as it is slightly more semantically correct and also provides screen readers with options.


jtpetty (approved):
LGTM

PR #5210, opened 2023-11-15
test(e2e): disable dataset creation per PR

ricokahler:
### Description

This is a temporary change to prevent explosive dataset growth in our CI tests.

### What to review

Do the CI the tests still pass? We wanted each run to have nearly its own dataset but our current suite should work fine with just one.

### Notes for release

N/A


jtpetty (approved):
👍

PR #5212, opened 2023-11-16
fix(core): field actions hidden issue

hermanwikner:
### Description

This pull request addresses an issue where hidden field actions caused the horizontal scroll bar to be visible within dialogs.

### What to review

- Confirm that the horizontal scroll bar in dialogs is no longer visible.
- Verify that field actions still work as expected.

### Notes for release

N/A


skogsmaskin (approved):
👍

PR #5213, opened 2023-11-16
feat(desk): add conditional borders to PaneHeader

ninaandal:
### Description
Adds conditional borders to the pane header

https://github.com/sanity-io/sanity/assets/44635000/c088a0e7-bf30-4d45-9462-cf6fd1f137f5

### What to review
The border for lists/documents lists/the document panel

FYI: I haven't added border logic to the components where the content is just one sentence and the scroll wasn't able to trigger (`ErrorPane.tsx `and `UnknownPaneType.tsx`)

### DIFF
--- a/packages/sanity/src/desk/components/pane/PaneHeader.styles.tsx
+++ b/packages/sanity/src/desk/components/pane/PaneHeader.styles.tsx
@@ -1,21 +1,32 @@
-import {Box, Flex, Layer, rgba, TextSkeleton, Text, Theme, Card} from '@sanity/ui'
+import {Box, Flex, Layer, rgba, TextSkeleton, Text, Theme, Card, LayerProps} from '@sanity/ui'
+import {HTMLProps, RefAttributes} from 'react'
 import styled, {css} from 'styled-components'
 
-export const Root = styled(Layer)`
-  line-height: 0;
-  position: sticky;
-  top: 0;
+interface CustomLayerProps extends LayerProps {
+  borderBottom?: boolean
+}
 
-  &:not([data-collapsed]):after {
-    content: '';
-    display: block;
-    position: absolute;
-    left: 0;
-    right: 0;
-    bottom: -1px;
-    border-bottom: 1px solid var(--card-shadow-outline-color);
-  }
-`
+export const Root = styled(Layer)((
+  props: CustomLayerProps & Omit<HTMLProps<HTMLDivElement>, 'as'> & RefAttributes<HTMLDivElement>,
+) => {
+  const {borderBottom = true} = props
+
+  return css`
+    line-height: 0;
+    position: sticky;
+    top: 0;
+
+    &:not([data-collapsed]):after {
+      content: '';
+      display: block;
+      position: absolute;
+      left: 0;
+      right: 0;
+      bottom: -1px;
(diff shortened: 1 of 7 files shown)


packages/sanity/src/desk/components/pane/PaneHeader.styles.tsx
```diff
@@ -1,21 +1,42 @@
...
+    &:not([data-collapsed]):after {
+      content: '';
+      display: block;
+      position: absolute;
+      left: ${BORDER_OFFSET_X}px;
+      right: ${BORDER_OFFSET_X}px;
+      bottom: -1px;
+      border-bottom: ${$borderBottom ? '1px solid var(--card-shadow-outline-color)' : 'none'};
+      opacity: ${() => ($isContentScrollable && $hasScrolledFromTop ? 1 : 0)};
+      transition: opacity 200ms ease-in;
```
robinpyon, line 36: This is close! But doesn't provide the transition that we're after –

It is applying a transition on `opacity` values, which we're varying by based on the value of `$isContentScrollable && $hasScrolledFromTop` – but we're also setting the `border-bottom` to `none`. As such, the border is 'snapping' in and out.

This could be alternatively re-written as:

```
      border-bottom: 1px solid ${$borderBottom ? 'var(--card-shadow-outline-color)' : 'transparent'};
      opacity: 1;
      transition: border-color 200ms ease-in;
```

packages/sanity/src/desk/components/pane/PaneHeader.styles.tsx
```diff
@@ -1,21 +1,42 @@
...
+ * Left/right border offsets for header components (in px).
+ */
+const BORDER_OFFSET_X = 12
+
+interface RootProps {
+  $borderBottom?: boolean
+  'data-collapsed'?: string
+  'data-testid'?: string
+  $isContentScrollable?: boolean
+  $hasScrolledFromTop?: boolean
```
robinpyon, line 14: These can safely be omitted, as `$borderBottom` is already derived from `isScrollable` and `hasScrolledFromTop` values, so we're probably sending more than we need.

Non JS identifiers (like `data-*` attributes) don't need to be manually typed, so these can also be omitted.

packages/sanity/src/desk/components/pane/PaneHeader.tsx
```diff
@@ -50,9 +52,31 @@ export const PaneHeader = forwardRef(function PaneHeader(
...
+
   return (
     <LayerProvider zOffset={100}>
-      <Root data-collapsed={collapsed ? '' : undefined} data-testid="pane-header" ref={ref}>
+      <Root
+        $borderBottom={isScrollable && hasScrolledFromTop}
+        data-collapsed={collapsed ? '' : undefined}
+        data-testid="pane-header"
+        $hasScrolledFromTop={hasScrolledFromTop}
+        $isContentScrollable={isScrollable}
```
robinpyon, line 77: See above, no need to pass 3 props to our styled component when the single `$borderBottom` is sufficient!

robinpyon (commented):
Thanks @ninaandal! Left a few comments below, let me know if you have any questions here

robinpyon (approved):
Thanks @ninaandal – looking great!

PR #5214, opened 2023-11-16
fix(desk): add capitalizedAction as button text in delete dialog

ninaandal:
### Description
Remove hardcoded Delete text and use capitalized action instead for the delete/unpublish dialog

### What to review
The dialog when deleting/unpublishing a document


pedrobonamin (approved):
Good catch!

PR #5218, opened 2023-11-16
fix(desk): add border always on document header, remove padding

ninaandal:
### Description
Adds a permanent border to the document panes and conditional borders to the document lists


robinpyon (approved):
Thanks @ninaandal!

PR #5221, opened 2023-11-16
fix(router): use base64url for encoding parameter payloads

rexxars:
### Description

Fixes an issue where using characters outside of the Latin1 range in intent or structure tool pane payloads would crash the studio.

### What to review

- Works with non-Latin1 characters such as `€`
- Is backwards compatible with old base64 encoded URLs

### Notes for release

- Fixed an issue where using characters outside of the Latin1 range in intents or structure tool payload URLs would crash the studio


packages/sanity/src/router/utils/base64url.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/**
+ * Decodes a base64url-encoded string
+ *
+ * @param str - String to decode
+ * @returns Decoded string
+ * @internal
+ */
+export function decodeBase64Url(str: string): string {
```
jtpetty, line 28: the encode removes `=` chars, but the decode does not add them back. I'm assuming that is intentional? Should that be explicitly documented?

jtpetty (approved):
👍

packages/sanity/src/router/utils/base64url.ts
rexxars: Shouldn't need to? [RFC 4648 § 5 ](https://datatracker.ietf.org/doc/html/rfc4648#section-5) states:
> The pad character "=" is typically percent-encoded when used in an URI [[9](https://datatracker.ietf.org/doc/html/rfc4648#ref-9)], but if the data length is known implicitly, this can be avoided by skipping the padding; see [section 3.2](https://datatracker.ietf.org/doc/html/rfc4648#section-3.2).

As long as we always attempt to decode using base64url _prior_ to attempting base64, this should work fine

ricokahler (approved):
super nice

PR #5225, opened 2023-11-17
feat(i18n): translate portable text editor

rexxars:
- Closes SDX-605
- Closes SDX-606
- Closes SDX-607

### DIFF
--- a/packages/@sanity/schema/src/legacy/types/blocks/defaults.ts
+++ b/packages/@sanity/schema/src/legacy/types/blocks/defaults.ts
@@ -2,7 +2,7 @@ export const DEFAULT_LINK_ANNOTATION = {
   type: 'object',
   name: 'link',
   title: 'Link',
-  i18nTitle: 'inputs.portable-text.annotation.link',
+  i18nTitleKey: 'inputs.portable-text.annotation.link',
   options: {
     modal: {type: 'popover'},
   },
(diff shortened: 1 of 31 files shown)


skogsmaskin (commented):
I'm wondering if we should write i18n into the `portable-text-editor` package at all? It doesn't really have any presentation concerns. Seems like it's most of all the validation error messages in the `portable-text-editor`  we are interested in translating here. Could these rather be wrapped or translated by the consumer and presenter (the PT-Input in this case)?

Just a thought, it's also nice to support translation of these directly. I guess there's no added deps or anything involved here? It's a point to be able to use the module without too many deps to other things (for external usage elsewhere). If this is the case then I guess it's all good.

Seems like a lot of tests break currently? What is the plan for having those pass?

packages/@sanity/schema/src/legacy/types/blocks/defaults.ts
```diff
@@ -1,6 +1,8 @@
 export const DEFAULT_LINK_ANNOTATION = {
   type: 'object',
   name: 'link',
+  title: 'Link',
+  i18nTitle: 'inputs.portable-text.annotation.link',
```
jtpetty, line 5: nit: definitely not worth blocking over, but should `i18nTitle` be `i18nTitleKey`? It's not actually the title, but the key to look up the title.

jtpetty (approved):
Wow! I thought we were going to hold back on all of PTE but looks like you found an easy solution to bring this home!

rexxars:
> I'm wondering if we should write i18n into the `portable-text-editor` package at all? It doesn't really have any presentation concerns. Seems like it's most of all the validation error messages in the `portable-text-editor` we are interested in translating here. Could these rather be wrapped or translated by the consumer and presenter (the PT-Input in this case)?

I intentionally didn't add any i18n to the `@sanity/portable-text-editor` module for that reason, aside from a optional added keys. I feel like that's not going to cost any maintenance or issues. How would you suggest we wrap it?

A different approach might be to give an arbitrary key to the validation messages and then map those to i18n keys in the `sanity` package, but we'd still need to pass along the context (eg key, index, expected type names and whatnot) - so not sure we are saving ourselves a lot of work.

Given that this is all internals, I think we can refactor later should we want to.

> Just a thought, it's also nice to support translation of these directly. I guess there's no added deps or anything involved here? It's a point to be able to use the module without too many deps to other things (for external usage elsewhere). If this is the case then I guess it's all good.

No new dependencies (in the `@sanity/portable-text-editor` package anyway) :)

> Seems like a lot of tests break currently? What is the plan for having those pass?

I will fix those before merging 🫡

packages/@sanity/schema/src/legacy/types/blocks/defaults.ts
rexxars: Good point. Will rename.

PR #5227, opened 2023-11-17
fix(facelift): update pane header create button tooltip

pedrobonamin:
### Description
Update the PaneHeaderCreateButton to include the tooltip in cases where it has only 1 template item type.
Incorrectly assumed in latest changes that the `<InsufficientPermissionsMessageTooltip >` was rendering in both cases. 

Now, if it doesn't have permissions it will render that template, in case it has permissions it will show the `Create new document` tooltip
Screenshot 2023-11-17 at 09 47 01


robinpyon (approved):
Thanks @pedrobonamin!

PR #5228, opened 2023-11-17
fix(core): fix padding issue with dateinput

ninaandal:
### Description
Fixes `padding` being off for the date input. 

image

Wrapping it in a `Box` for this for some reason was very off with the padding, even when setting the styling manually and not using the `padding` values for the component. I'm not sure if this is the desired approach for this, but setting a manual style for `Flex` seems to be an easy solution?


pedrobonamin (approved):
Thanks @ninaandal 
The fontSize of the icon changed from , and that is making the necessity to change the padding of this element to match the text input height. It was 27x27 px and now is 25x25, and we don't have a padding available with 5px, but we need that. 

In this case I think this is the right solution, thanks for fixing it!

PR #5232, opened 2023-11-17
fix(comments): reply input focus issue

hermanwikner:
### Description

This pull request fixes an issue where the reply comment input is getting focused when clicking outside of it. The reason for this behavior is not immediately apparent, but replacing `styled(Stack)` with a plain div resolves the problem. In addition, this pull request refactors some logic in the editable component in the comment input to make it more readable.

https://github.com/sanity-io/sanity/assets/15094168/903d5e45-bd85-4141-8b8f-78d755c2be88

### What to review

- Make sure that the comment input don't get focused when clicking outside of it

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/desk/comments/src/components/pte/comment-input/Editable.tsx
+++ b/packages/sanity/src/desk/comments/src/components/pte/comment-input/Editable.tsx
@@ -1,5 +1,5 @@
 import React, {KeyboardEvent, useCallback, useEffect, useMemo, useRef, useState} from 'react'
-import {Popover, PortalProvider, Stack, useClickOutside, useGlobalKeyDown} from '@sanity/ui'
+import {Popover, PopoverProps, useClickOutside} from '@sanity/ui'
 import {
   EditorSelection,
   PortableTextEditable,
@@ -13,17 +13,9 @@ import {renderBlock, renderChild} from '../render'
 import {useCommentInput} from './useCommentInput'
 import {useCursorElement} from './useCursorElement'
 
+const POPOVER_FALLBACK_PLACEMENTS: PopoverProps['fallbackPlacements'] = ['bottom', 'top']
 const INLINE_STYLE: React.CSSProperties = {outline: 'none'}
-
-const EditableWrapStack = styled(Stack)(() => {
-  return css`
-    & > div:first-child {
-      [data-slate-node='element']:not(:last-child) {
-        margin-bottom: 1em; // todo: improve
-      }
-    }
-  `
-})
+const EMPTY_ARRAY: [] = []
 
 const PlaceholderWrapper = styled.span(() => {
   return css`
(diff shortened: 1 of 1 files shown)


packages/sanity/src/desk/comments/src/components/pte/comment-input/Editable.tsx
```diff
@@ -199,49 +173,57 @@ export function Editable(props: EditableProps) {
...
-          renderChild={renderChild}
-          renderPlaceholder={renderPlaceholder}
-          selection={initialSelectionAtEndOfContent}
-          style={INLINE_STYLE}
-          tabIndex={focusLock ? 0 : undefined}
-        />
-      </EditableWrapStack>
-    </>
+    <div ref={rootElementRef}>
+      <StyledPopover
```
hermanwikner, line 200: We previously used a portal to render the `Popover` with the`rootElementRef` element as the portal element. However, this is unnecessary, and we can just place the `Popover` directly inside the `rootElementRef` without using a portal (the same result).

packages/sanity/src/desk/comments/src/components/pte/comment-input/Editable.tsx
```diff
@@ -97,43 +90,23 @@ export function Editable(props: EditableProps) {
...
-
-  // Close mentions if the user selects text
-  useEffect(() => {
-    if (mentionsMenuOpen && selection && !isEqual(selection.anchor, selection.focus)) {
-      closeMentions()
-    }
-  }, [mentionsMenuOpen, closeMentions, selection])
-
-  // Close mentions if the menu itself has focus and user press Escape
-  useGlobalKeyDown((event) => {
```
hermanwikner, line 127: This `useGlobalKeyDown` is not necessary since we will close the mentions menu in `handleKeyDown` (case 'Escape').

packages/sanity/src/desk/comments/src/components/pte/comment-input/Editable.tsx
```diff
@@ -13,17 +13,9 @@ import {renderBlock, renderChild} from '../render'
 import {useCommentInput} from './useCommentInput'
 import {useCursorElement} from './useCursorElement'
 
+const POPOVER_FALLBACK_PLACEMENTS: PopoverProps['fallbackPlacements'] = ['bottom', 'top']
 const INLINE_STYLE: React.CSSProperties = {outline: 'none'}
-
-const EditableWrapStack = styled(Stack)(() => {
-  return css`
-    & > div:first-child {
```
hermanwikner, line 20: This CSS is not used anymore since a comment can only contain one block. This CSS was previously used to add spacing between individual blocks in the editor.

robinpyon (approved):
Thanks @hermanwikner – great spot!

PR #5244, opened 2023-11-20
fix: globally replace compose with add icon

robinpyon:
### Description

This PR changes the icon for create new document buttons (from the 'compose' to the 'add' icon). This button is found in the studio navbar and document lists.

This was a proposed change in upcoming facelift work, but has been brought forward to reduce ambiguity with the just released _Presentation_ tool, as that also uses the compose icon.

### What to review

That the correct 'add' icon is being rendered in the above instances.

### Notes for release

N/A


mariuslundgard (approved):
Thank you 🙇‍♂️

bjoerge (approved):

PR #5247, opened 2023-11-21
fix(tooltips): wrap button collections within a TooltipDelayGroupProvider

pedrobonamin:
### Description

Wraps button collections within a `<TooltipDelayGroupProvider>`
Wrapped tooltips will only have a open delay for the first tooltip that opens from the group, for an improved experience, related buttons are wrapped together, so once the first tooltip shows, hovering on the next ones will show the tooltip without delays. 

Groups the following:

- Navbar – LHS (workspace, create new)
- Navbar – RHS (search, help, presence)
- Document pane header actions
- Document list pane header actions
- Document footer - publish button / review changes buttons
- Document actions - menu items
- PortableText toolbar.
- Comments input actions. (mention user and send message)
- Reference preview
- Pane preview item

https://github.com/sanity-io/sanity/assets/46196328/5f858887-3176-44b6-a4da-e4d85e7a2470

### What to review
Is there a missing group that should also be wrapped?

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Editor.tsx
@@ -12,8 +12,15 @@ import {
   RenderListItemFunction,
 } from '@sanity/portable-text-editor'
 import {Path} from '@sanity/types'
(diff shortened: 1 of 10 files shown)


robinpyon (approved):
Thanks @pedrobonamin! Can't think of any other groups for now!

PR #5249, opened 2023-11-21
fix(form): only allow one PTE toolbar menu to be open

hermanwikner:
### Description

This pull request addresses an issue where two menus in the PTE toolbar could be open simultaneously.

### What to review

1. Open one menu in the PTE toolbar.
2. Open the second menu.
3. Confirm that the first menu closes when the second one is opened.


sjelfull:
LGTM, but do we know why we were stopping propagation in that instance to begin with?

sjelfull (approved):

skogsmaskin:
> LGTM, but do we know why we were stopping propagation in that instance to begin with?

This is because the editor shouldn't lose the cursor focus.

skogsmaskin (approved):

PR #5251, opened 2023-11-21
fix(core): close PTE fullscreen when changing tab

ninaandal:
### Description
Fixes issue where when you entered fullscreen in pte and switched tabs (for example from editor to json tab), the fullscreen would still be active in the other pane, by unmounting the form inputs when the tab is not active.

**To be considered**
Would it be better to unmount the full component? There are some effects and subscriptions in `packages/sanity/src/desk/panes/document/documentPanel/documentViews/FormView.tsx` which we are not sure if it will be better to keep or not, that's why the change is unmount the `<FormBuilder>` instead of form view. 

### What to review

- Have two views / panes (editor, preview)
- Open PTE fully
- Try to open the second pane (ex: preview)

### Notes for release
Closes fullscreen PTE when changing tabs


jtpetty (approved):
LGTM

PR #5253, opened 2023-11-21
fix(portable-text-editor): handle undefined block children

sjelfull:
### Description

This fixes an issue where the PTE would crash if you inserted a block with undefined `children`.

We will now show the correct "Invalid Portable Text Editor value" message.

CleanShot 2023-11-21 at 15 38 51

How to reproduce:

```typescript
// someScript.ts
import {getCliClient} from 'sanity/cli'

getCliClient().create({
  _type: 'blocksTest',
  first: [{_key: 'a', _type: 'block'}],
})
```

```bash
// In the test studio folder
sanity exec someScript.ts --with-user-token
```

### What to review

- Create a document with the repro above
- Navigate to the document in Standard inputs -> Portable Text -> Blocks and find your document
- See the "Invalid Portable Text Editor value" message

### Notes for release

- Fixed a issue where the PTE would crash  if you inserted a block with undefined `children`

### DIFF
--- a/packages/@sanity/portable-text-editor/src/utils/__tests__/values.test.ts
+++ b/packages/@sanity/portable-text-editor/src/utils/__tests__/values.test.ts
@@ -58,22 +58,23 @@ describe('toSlateValue', () => {
       {schemaTypes},
     )
     expect(result).toMatchInlineSnapshot(`
-      Array [
-        Object {
-          "_key": "123",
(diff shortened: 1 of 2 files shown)


skogsmaskin (approved):
hehe, approving my own shit :)

pedrobonamin (approved):

PR #5254, opened 2023-11-21
fix(pte): clear unused markdefs when removing annotations

pedrobonamin:
### Description

When a mark is extended to more than one word, removing the annotation wasn't clearing the markDefs array if that annotation is no longer in use.

https://github.com/sanity-io/sanity/assets/46196328/2c702699-d68d-40e7-8e5a-516115fe2665

Remove unused markDefs when removing an annotation.


dev/test-studio/schema/standard/portableText/allTheBellsAndWhistles.ts
```diff
@@ -33,7 +32,7 @@ export const ptAllTheBellsAndWhistlesType = defineType({
                 type: 'object',
                 name: 'link',
                 title: 'Link',
-                components: {input: LinkAnnotationInput},
+                // components: {input: LinkAnnotationInput},
```
pedrobonamin, line 35: This is making the preview crash

skogsmaskin: There is something wrong with that schema in the test studio, should be removed or fixed.

pedrobonamin: I think it's a problem in the custom component, the schema seems correct. The component has lots of "TODO" in it, that's why I decided to remove it from here.
Maybe we can remove the component entirely from the codebase, not sure the intention of that component, is 2 years old and has been in TODO since then. 

Adding this change
`  const {schemaType: type, value, onChange, onBlur, onFocus, validation} = props` 
in https://github.com/sanity-io/sanity/blob/next/dev/test-studio/schema/standard/portableText/LinkAnnotationInput.tsx#L30 will fix the issue, but it will render just all TODO boxes

pedrobonamin: Screenshot 2023-11-22 at 16 21 30

skogsmaskin: At some point, this was a test studio type for  `v2`, and it hasn't been properly ported over to v3 I think. Feel free to change or even remove this type.

skogsmaskin (commented):
I'm a bit surprised that the [normalization in `withPortableTextMarksModel`](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextMarkModel.ts#L396) didn't catch these.

dev/test-studio/schema/standard/portableText/allTheBellsAndWhistles.ts
pedrobonamin: Removed, thanks!

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextMarkModel.ts
```diff
@@ -389,37 +389,40 @@ export function createWithPortableTextMarkModel(
...
+ */
+export function normalizeMarkDefs(
+  editor: PortableTextSlateEditor,
+  types: PortableTextMemberSchemaTypes,
+): void {
+  const {selection} = editor
+  if (selection) {
+    const blocks = Editor.nodes(editor, {
+      at: selection,
+      match: (n) => (n as unknown as Descendant)._type === types.block.name,
```
skogsmaskin, line 405: Bonus: The casting here (`(n as unknown as Descendant)`) is no longer needed, would be great to have it removed.

skogsmaskin (approved):
Great! And thanks again for the tests!

I've added a comment about a bonus change that could be added to the code while we are at it.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextMarkModel.ts
pedrobonamin: removed! Thanks

PR #5257, opened 2023-11-22
fix(pte): disable multiple blocks annotations

pedrobonamin:
### Description
When selecting multiple blocks in the PTE editor and then adding an annotation to it, the PTE editor breaks, because annotations are not meant to be multi-block.

Annotations require a `markDef` and that `markDef` needs to exist in the block in which the annotation is referred. 
To avoid issues, disable the `annotations` when selecting multiple blocks

- If block is empty, annotations will be disabled (this is already happening) and will show a tooltip message : Cannot apply {{annotationTitle}} to empty block
- If is selecting multiple blocks, annotation will be disabled and will show tooltip message Cannot apply {{annotationTitle}} to multiple blocks
- Selecting 1 block with content, annotation enabled and message: {{annotationTitle}}

Screenshot 2023-11-22 at 13 02 56
Screenshot 2023-11-22 at 13 02 49
Screenshot 2023-11-22 at 13 02 40

### What to review

Annotations behaviour

### Notes for release

Disabled multiple blocks annotations, adds disabled annotations information on tooltip.


skogsmaskin (approved):
Great! Thanks.

PR #5258, opened 2023-11-22
feat(desk): make sure comments don't affect performance when disabled

hermanwikner:


packages/sanity/src/desk/comments/src/context/comments/CommentsProvider.tsx
```diff
@@ -43,66 +42,11 @@ export interface CommentsProviderProps {
...
  */
 export const CommentsProvider = memo(function CommentsProvider(props: CommentsProviderProps) {
   const {children, documentId, documentType} = props
-
-  const {isEnabled} = useCommentsEnabled({
-    documentId,
-    documentType,
-  })
-
-  if (!isEnabled) {
```
hermanwikner, line 91: This can be removed now since we make sure to only render the `CommentField` component and the `CommentsDocumentInspector` (which uses `useComments`) if comments is enabled by looking at the enabled value from `CommentsEnabledContext.

packages/sanity/src/desk/panes/document/DocumentPane.tsx
```diff
@@ -144,24 +145,26 @@ function DocumentPaneInner(props: DocumentPaneProviderProps) {
...
-    >
-      {/* NOTE: this is a temporary location for this provider until we */}
-      {/* stabilize the reference input options formally in the form builder */}
-      {/* eslint-disable-next-line react/jsx-pascal-case */}
-      <ReferenceInputOptionsProvider
-        EditReferenceLinkComponent={ReferenceChildLink}
-        onEditReference={handleEditReference}
-        initialValueTemplateItems={templatePermissions}
-        activePath={activePath}
+    <CommentsEnabledProvider documentId={options.id} documentType={options.type}>
```
hermanwikner, line 148: Added here so that we can access the context value in `DocumentPaneProvider` and conditionally add the comment providers if comments is enabled

bjoerge (approved):
Works great! Thank you @hermanwikner 🙏🏼

bjoerge (changes requested):
I see now that the tests are failing - would be great if you could look into making them green 🙏🏼

bjoerge (approved):

PR #5264, opened 2023-11-23
fix(core): ensure that `StudioLayout` utilizes the components API

hermanwikner:
### Description

This pull request ensures that `StudioLayout` utilizes the Components API (`studio.components.layout`) when used in isolation. Recent features, such as Comments and AI assist, utilize this property in the Components API to wrap the layout in context providers. However, in an embeddable context, one might import the `StudioLayout` directly, like this:

```ts
import {StudioProvider, StudioLayout, defineConfig} from 'sanity'

const config = defineConfig({
  // ...
})

export function MyStudio() {
  return (
    <StudioProvider config={config}>
      <StudioLayout />
    </StudioProvider>
  )
}
```

With this approach, you import the _default_ `StudioLayout` component, which does not include any modifications added to `studio.components.layout`. This leads to errors when components in `StudioLayout` try to access values from missing context providers.

### What to review

- Run the new embedded dev app with `yarn dev:embedded-studio` and make sure that the comments feature works.
- Do you see any issues with these changes?

### Notes for release

Ensures that `StudioLayout` utilizes the components API when used in isolation

### DIFF
--- a/dev/embedded-studio/.depcheckignore.json
+++ b/dev/embedded-studio/.depcheckignore.json
@@ -1,0 +1,3 @@
+{
+  "ignore": ["styled-components", "react", "react-dom", "sanity"]
+}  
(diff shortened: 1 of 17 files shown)


stipsan (approved):
Perfect 😌

PR #5267, opened 2023-11-23
fix(pte): removing annotations cleans incorrect path

pedrobonamin:
### Description
Fixes issue in which splitting the paths when removing annotations and creating new spans was not always selecting the correct span to remove the annotations from.
See video:

https://github.com/sanity-io/sanity/assets/46196328/ee99bd69-be52-4a29-99e0-30d70ac3d996

### What to review

Annotations, adding and removing.

### Notes for release
Fixes issue in which splitting the paths when removing annotations and creating new spans was not always selecting the correct span to remove the annotations from


skogsmaskin (approved):
Nice! Thanks for adding a test as well! 👍

PR #5268, opened 2023-11-24
feat(perf): add cli flag for excluding patch releases

bjoerge:
### Description
Adds a flag to perf test runner that lets you exclude patch releases when running perf tests against earlier versions (tags)

### What to review
- Does it look sensible?

### Notes for release

n/a - internal tooling


skogsmaskin (approved):
LGTM!

PR #5270, opened 2023-11-24
fix(form): various typing speed improvements

bjoerge:
### Description
This PR includes a couple of commits that individually address performance issues. The main theme here is splitting up context providers and avoiding subscriptions to contexts provided by components that updates frequently (e.g. at every keystroke).

Note: while looking into this I also discovered that recent comments work introduced some of the same issues that is fixed in this PR, @hermanwikner has already done a great job addressing these separately in #5258, which is now merged into this branch)

Charts comparing (from left to right) v3.17, v3.18, v3.19, v3.20 and the branch of this PR (including other recent optimizations already merged to next)
image
(lower is better)

### What to review
- Does the changes make sense?
- Any potential for breaking changes? From what I can tell, this only touches internal APIs

### Notes for release
- Fixes some performance issues related to typing in text inputs. Typing in Studio inputs should now be noticably faster, in particular for schema types with a lot of deeply nested fields.


packages/sanity/src/core/form/FormBuilderProvider.tsx
```diff
@@ -118,10 +117,7 @@ export function FormBuilderProvider(props: FormBuilderProviderProps) {
   } = props
 
   const documentValueRef = useRef(documentValue)
-
-  useEffect(() => {
-    documentValueRef.current = documentValue
-  }, [documentValue])
+  documentValueRef.current = documentValue
```
hermanwikner, line 120: Why do we not want to set the value of this ref in a `useEffect` anymore?

packages/sanity/src/core/form/contexts/DocumentIdProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+  return ctx.id
+}
+
+export function useDocumentIds() {
+  return {draft: useDraftId(), published: usePublishedId()}
+}
+
+/**
+ * @deprecated Use `usePublishedDocumentId` instead
```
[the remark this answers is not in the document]
bjoerge: yes! thanks for catching

packages/sanity/src/core/form/FormBuilderProvider.tsx
bjoerge: Great question! Counter question: Why do we want to set it in a useEffect in the first place?

The intention behind doing it this way wasn't very clear from reading the code and it very much stands out as an error to me. If I'm not mistaken, idea was that we're keeping it in a ref so that we can look it up at the time the `getDocument()` function is called, to prevent it from becoming part of the render cycle. Since it should always point at the current document, I can't see any reason to wait for it to update it in an effect. But maybe more important: the way it was, would caused child components calling getDocument() in their render to lag one iteration behind.

That said, it doesn't look like we're using `useFormBuilder().__internal.getDocument()` at all anymore (the only prior usage I could find was by the slug input), so I've removed it completely in ada38de750f1e874d2f04f4fbb2e7cd6d5c7906c.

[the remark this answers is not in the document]
bjoerge: See answer above

bjoerge: Realized my code search skills failed me, and that useFormBuilder().__internal.getDocument() was still in use by the slug input, which made ada38de break. As an alternative I introduced a separate context that provides a `getDocument(path: Path)` function which is stable throughout the life cycle, and can be used to read the current document value in a callback function (e.g. `getNew[Slug]FromSource()`).

This made it possible to remove value entirely from FormBuilderContext, which is a nice little cleanup too. Take a look at c4964ed9992031cfcedc50e7cf1a2115cc72b538

packages/sanity/src/core/form/contexts/DocumentIdProvider.tsx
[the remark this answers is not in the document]
bjoerge: Fixed

packages/sanity/src/core/form/studio/FormBuilder.tsx
```diff
@@ -175,6 +136,50 @@ function RootInput() {
     [onSetFieldSetCollapsed],
   )
 
+  const {file, image} = useSource().form
```
hermanwikner, line 139: This can be removed since it is not used anywhere!

packages/sanity/src/core/form/contexts/GetFormValue.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {getValueAtPath} from '../../field'
+import type {FormDocumentValue} from '../types'
+
+/**
+ * @internal
+ * @hidden
+ */
+export type GetFormValueContextValue = (path: Path) => unknown
+
+const GetFormValueContext = createContext<GetFormValueContextValue | null>(null)
```
hermanwikner, line 12: No blocker for merge, but I think it is nice to split the provider, context, and hook(s) into different files, mainly because that makes it a bit easier and faster to find things in the codebase by utilizing the file search in the editor (at least in VSCode)!

bjoerge: Great point! Should be better now after db64732b74ba3b28f3984f9d8c8c99cb40921fbb

packages/sanity/src/core/form/studio/FormBuilder.tsx
bjoerge: Nice catch! Removed

packages/sanity/src/core/form/contexts/GetFormValue.tsx
bjoerge: Actually, looks like db64732b74ba3b28f3984f9d8c8c99cb40921fbb needs some more work, so I'm reverting that. Doesn't feel critical, can address it later :)

PR #5271, opened 2023-11-26
fix(pte): removes empty text block when inserting a new block in that position

pedrobonamin:
### Description

When inserting a new block the PTE editor was creating empty text blocks.
This PR adds a check to remove the text block in the user selection if it's empty before inserting a new block.
So, if you have a empty editor and try to insert any new block, no empty text blocks will be created.
If the editor has values, and you are in an empty text block, no new blocks will be added.

https://github.com/sanity-io/sanity/assets/46196328/03afca8c-5f02-439b-87ef-58fd07e8a568

### What to review

Adding text blocks should continue working as expected.

### Notes for release

Fixes issue in which new empty text blocks where added when adding a block in the portable text editor.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withEditableAPIInsert.test.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withEditableAPIInsert.test.tsx
@@ -20,10 +20,30 @@ const initialValue = [
     style: 'normal',
   },
 ]
-
 const initialSelection = {
   focus: {path: [{_key: 'a'}, 'children', {_key: 'a1'}], offset: 7},
   anchor: {path: [{_key: 'a'}, 'children', {_key: 'a1'}], offset: 7},
+}
+
+const emptyTextBlock = [
+  {
+    _key: 'emptyBlock',
+    _type: 'myTestBlockType',
+    children: [
+      {
+        _key: 'emptySpan',
+        _type: 'span',
+        marks: [],
+        text: '',
+      },
+    ],
+    markDefs: [],
+    style: 'normal',
+  },
+]
+const emptyBlockSelection = {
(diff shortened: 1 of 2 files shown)


packages/@sanity/portable-text-editor/src/editor/plugins/createWithEditableAPI.ts
```diff
@@ -180,6 +180,24 @@ export function createWithEditableAPI(
...
           portableTextEditor,
         )[0] as unknown as Node
+        const [focusBlock] = Array.from(
+          Editor.nodes(editor, {
+            at: editor.selection.focus.path.slice(0, 1),
+            match: (n) => n._type === types.block.name,
+          }),
+        )[0] || [undefined]
+
+        const isEmptyTextBlock =
```
skogsmaskin, line 190: You could use the `isEqualToEmptyEditor` helper function here, as it will also check if you did any styling or similar to the block. This is what we use throughout the code for this test.

skogsmaskin (changes requested):
Thanks for this @pedrobonamin!

Just had a suggestion for testing for "empty editor".

packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withEditableAPIInsert.test.tsx
```diff
@@ -137,3 +157,188 @@ describe('plugin:withEditableAPI: .insertChild()', () => {
     })
   })
 })
+
```
skogsmaskin, line 160: Great with more tests! ❤️

packages/@sanity/portable-text-editor/src/editor/plugins/createWithEditableAPI.ts
pedrobonamin: Thanks @skogsmaskin updated

skogsmaskin (approved):
Thanks for this @pedrobonamin

PR #5275, opened 2023-11-27
fix(comments): performance issue in `CommentField`

hermanwikner:
### Description

This pull request removes the use of `useDocumentPane` in the `CommentField` component to improve performance. The hook was used to open the comments inspector, which is now replaced with props passed to the `CommentsProvider`.


bjoerge (approved):
Fantastic!
Just did a comparison locally using the perf studio with comments enabled and the difference is huge!

For clarity: This is comparing v3.19.0 and v3.20.0 with comments _disabled_ vs this branch with comments _enabled_!

image

bjoerge:
Rebased against next. Good to merge now!

PR #5280, opened 2023-11-28
fix(core): prevent annotation popover from jumping when opening

ninaandal:
### Description
The annotation toolbar popover would wrongly appear at the top left corner of the parent container when you selected text over the annotation. 

https://github.com/sanity-io/sanity/assets/44635000/f5454db0-3216-4419-bb82-68992c5f7280

### What to review
Annotation popover and dialog logic. 

### Notes for release
Fixes issue where annotation popover open when selecting annotation text.


bjoerge (approved):
LGTM!

PR #5281, opened 2023-11-28
feat: added string/text to block converters

RostiMelk:
### Description

Added converter for when a schema field changes  type from string/text to portable text

### What to review

- Create a field of type string/test
- Add content to the field
- Publish document
- Change field type to array with of  type 'block'
- You should now see a "Convert to portableText" button
- Click it
- The data should be transformed into portable text spans

For text that is multiline, there should be multiple portable text spans.

### Notes for release

tbd

### DIFF
--- a/packages/sanity/src/core/form/inputs/InvalidValueInput/converters.ts
+++ b/packages/sanity/src/core/form/inputs/InvalidValueInput/converters.ts
@@ -1,4 +1,5 @@
 import isValidDate from 'date-fns/isValid'
+import {randomKey} from '@sanity/block-tools'
 
 const TRUTHY_STRINGS = ['yes', 'true', '1']
 const FALSEY_STRINGS = ['false', 'no', 'false', '0', 'null']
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/InvalidValueInput/converters.ts
```diff
@@ -56,6 +57,32 @@ export const converters: {[fromType: string]: {[toType: string]: ValueConverter}
         }
       },
     },
+    portableText: {
+      test: TRUE,
+      convert: (value: string) => [
+        {
+          _type: 'block',
```
rexxars, line 64: Users can alias a block type to a different name, so `block` is not safe to use - we need to find and pass through the expected type and pick the `name` property from the schema type in question.

packages/sanity/src/core/form/inputs/InvalidValueInput/converters.ts
```diff
@@ -56,6 +57,32 @@ export const converters: {[fromType: string]: {[toType: string]: ValueConverter}
...
+        },
+      ],
+    },
+  },
+  text: {
+    portableText: {
+      test: TRUE,
+      convert: (value: string) => [
+        {
+          _type: 'block',
```
rexxars, line 77: Same as above - `block` is not guaranteed to be the block type name

packages/sanity/src/core/form/inputs/InvalidValueInput/converters.ts
```diff
@@ -56,6 +57,32 @@ export const converters: {[fromType: string]: {[toType: string]: ValueConverter}
...
+          _key: randomKey(12),
+          children: value
+            .split('\n')
+            .map((line) => ({_type: 'span', text: line, _key: randomKey(12)})),
```
rexxars, line 81: This isn't quite right, it will convert:

```
This
is

a test
```

to:
```
thisisatest
```

A span is a "chunk of text", in a line. We should _either_ leave the newlines as-is, or try to map it to multiple blocks.

Here's a `convert` function that converts to individual blocks instead (note that it only matches _double_ newlines - single newlines are treated as "soft newlines". We could change to single newlines if preferable.

```ts
convert: (value: string) => value.split('\n\n').map(line => ({
  _type: 'block', // or `blockSchemaType.name`, actually
  _key: randomKey(6),
  children: [{ _type: 'span', _key: randomKey(6), text: line }],
  markDefs: []
}))
```

rexxars (changes requested):
Love the feature, but left some comments on things that needs adjusting.
Do we have any tests for these converters? Might be worth adding if not.

jordanl17:
This PR seems stale, so going to close. Please flag if it's still relevant and you'd like to progress with it

PR #5282, opened 2023-11-28
fix(dialog): drag events in editPortal are applied to underlying elements

pedrobonamin:
### Description

Contain onDragEnter event to the dialog only, as it's propagating to the elements behind it causing unexpected behaviours in the elements.

## With error

https://github.com/sanity-io/ui/assets/46196328/c2661532-44de-4fd5-bef9-a693e34eb224

## Fixed:

https://github.com/sanity-io/ui/assets/46196328/c6b317b1-cd5e-42a8-997b-93263ede1c06

### What to review

EditPortal modal behaviour. 

### Notes for release

Fix error in which working in a PTE which is opened in a modal over an array, drag'n'dropping blocks will affect the underlying array (error:"Can't upload this file here")

### DIFF
--- a/packages/sanity/src/core/form/components/EditPortal.tsx
+++ b/packages/sanity/src/core/form/components/EditPortal.tsx
@@ -18,6 +18,10 @@ interface Props {
   autofocus?: boolean
 }
 
+function onDragEnter(event: React.DragEvent<HTMLDivElement>) {
+  return event.stopPropagation()
+}
+
 export function EditPortal(props: Props): React.ReactElement {
   const {
     children,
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/components/EditPortal.tsx
```diff
@@ -42,22 +46,24 @@ export function EditPortal(props: Props): React.ReactElement {
...
-      >
-        <Dialog
-          header={header}
-          id={id || ''}
-          onClickOutside={onClose}
-          onClose={onClose}
-          width={width}
-          contentRef={setDocumentScrollElement}
-          __unstable_autoFocus={autofocus}
+      <div onDragEnter={onDragEnter}>
```
hermanwikner, line 49: Do we need to wrap this in a div? When I test this, the issue seems to be fixed by adding the `onDragEnter` function to the `Dialog` component. However, there might be some situations that I am not aware of where the div is needed!

pedrobonamin: No, you are absolutely right, I didn't realised the dialog was spreading the props. thanks!

hermanwikner (approved):
Nice, thanks! @pedrobonamin 🙏

PR #5285, opened 2023-11-28
fix(deps): upgrade `vite` to v5

stipsan:
### Description

# [Vite v5 is full of goodies, perf and fixes.](https://vitejs.dev/blog/announcing-vite5#vite-5-0-is-out)

### What to review

`sanity dev` should hot reload faster, `sanity build` should be significantly faster. The existing public API shouldn't break and v3 plugins should continue to bundle.
It can be tested using `pnpm i sanity@3.65.2-upgrade-vite-v5.31`

### Notes for release

fix(deps): upgrade to `vite` v5 for speedier `sanity dev` and `sanity build` DX ✨


rexxars:
Getting errors of `The CJS build of Vite's Node API is deprecated`. Not sure _quite_ how to resolve this - esbuild-register kind of messes up the stack trace, and the way we bundle everything into a single file kind of makes things even more complicated.

I think ideally we'd move towards an ESM-first CLI that does not bundle build the CLI to a single file (previously this was done because npm2 used an absurd amount of time installing dependencies - that story is now better, and we have less dependencies than back then too).

rexxars:
Having said that, if we can make it still make it work with today's solution, I'd be more than happy to look at the altered approach.

skogsmaskin:
Is this still on?

stipsan:
@skogsmaskin, yes, I'll update the PR again when I have time between larger tasks, or waiting for a review on a larger PR 😄

stipsan:
@skogsmaskin @rexxars ready for review now 👍

bjoerge:
@stipsan is this still relevant now that vite 6 is out? or can it be closed?

rexxars:
> @stipsan is this still relevant now that vite 6 is out? or can it be closed?

Still relevant. I'm working on it.

PR #5286, opened 2023-11-28
fix: remove circular dependency between `useGlobalPresence` and `usePresenceStore`

stipsan:
### Description

When running the build command on the test studio I noticed this warning:
image
It appears to be caused by the circular dependency fixed in this PR.

### What to review

Warning should be gone, and nothing else should break/change as a result. The generated bundle should stay the same as none of the publicly exported paths changed.

### Notes for release

None


rexxars (approved):
Thanks for finding this!

PR #5291, opened 2023-11-28
fix(pte): focus error in readonly editor

pedrobonamin:
### Description

When the PTE editor is in `readOnly` mode the onBlur event is not triggered when opening a internal dialog, for example, when we want to edit an inner object.
This is causing that when clicking on other object, the focus or selection of the PTE editor changes and creates a strange behaviour, focusing on multiple elements at the same time and showing an error in the screen.

With this changes, I'm disabling the selection when the editor is in readOnly mode.

### What to review

Is there anything to consider in why we would want to keep the selection in the readOnly mode?
Is there a way to trigger the onBlur event even though the editor is in readOnly mode?

### Notes for release

Fixes issue in which the PTE editor throws an error when opening dialogs if the user is viewing old documents.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -145,6 +145,9 @@ export function PortableTextInput(props: PortableTextInputProps) {
           }
           break
         case 'selection':
+          // In readOnly mode the selection is creating incorrect selections because `blur`is not triggered when opening
+          // a inner dialog.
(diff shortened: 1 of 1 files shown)


skogsmaskin (changes requested):
The problem seems to be that the object edit modals are requesting focus on the editor when they are clicked? With these terms in mind, I don't think these changes will fix the actual problem

I think we need to have a closer look at this.

pedrobonamin:
It's not possible to reproduce this issue anymore. Closing pull request.

PR #5292, opened 2023-11-28
test(ci): move cli test to separate workflow without sharding

binoy14:
### Description

This PR moves the CLI tests to it's own workflow without jest sharding so that it does not create a lot of datasets. The total number of datasets it should create per PR should be  (datasets(5) + copy) * studio version (2) * node versions (2) = **24** which is not ideal and hence long term solution would be moving some of the API calls but this should make the tests functional

### What to review

If the changes makes sense

### Notes for release

N/A


jtpetty (approved):
👍

PR #5297, opened 2023-11-29
fix(import): pass down all cli options to the import script

rneatherway:
### Description

The `--skip-cross-dataset-references` option is currently ignored by the CLI. As an example:

```
❯ sanity dataset import admin.ndjson production --skip-cross-dataset-references --allow-assets-in-different-dataset

... <snip>

Unhandled rejection: Error: The data to be imported contains one or more cross-dataset references, which refers to a dataset that do not exist in the target project.
Missing dataset: "v2-docs"
Either create this dataset in the given project, or use the `skipCrossDatasetReferences` option to skip these references.
    at validateCdrDatasets (/<snip>/node_modules/@sanity/import/src/validateCdrDatasets.js:25:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```


packages/sanity/src/_internal/cli/commands/dataset/importDatasetCommand.ts
```diff
@@ -105,8 +105,15 @@ const importDatasetCommand: CliCommandDefinition = {
...
-    const {allowAssetsInDifferentDataset, allowFailingAssets, assetConcurrency, replaceAssets} =
-      flags
+    const {
+      allowAssetsInDifferentDataset,
+      allowFailingAssets,
+      assetConcurrency,
+      skipCrossDatasetReferences,
+      replaceAssets,
+      replace,
+      missing,
```
rexxars, line 115: Same as below, these flags aren't actually used. Not sure why we're parsing them here. Maybe the `getMutationOperation` logic should be moved in here instead?

packages/sanity/src/_internal/cli/commands/dataset/importDatasetCommand.ts
```diff
@@ -242,8 +249,11 @@ const importDatasetCommand: CliCommandDefinition = {
         onProgress,
         allowFailingAssets,
         allowAssetsInDifferentDataset,
+        skipCrossDatasetReferences,
         assetConcurrency,
         replaceAssets,
+        replace,
+        missing,
```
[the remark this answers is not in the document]
rneatherway: Oh yes, I see that now, thanks!

packages/sanity/src/_internal/cli/commands/dataset/importDatasetCommand.ts
rneatherway: I see yes. I'm happy to just remove them from here, I don't see any big benefit to inlining `getMutationOperation`

rexxars (approved):

PR #5299, opened 2023-11-29
refactor(core): grants permission logic

hermanwikner:
### Description

This pull request refactors the grants permission logic in `core` to make it reusable in the comments implementation. Currently, the grants permission check logic expects to receive the `CurrentUser`. This causes type issues in the comments implementation because permissions need to be checked for each user in a list, and these users are not in the shape of `CurrentUser`.

In the current grants permission logic in `core`, only the `id` value of the user is used, and there's no need to pass the entire user object. Therefore, simplifying this by passing only the user's id makes it possible to use this logic in the comments implementation. The shape of each user in the comments implementation is different from `CurrentUser` (i.e., `MentionOptionUser`), and this change allows for the removal of duplicated logic in the comments implementation.

### What to review

- Make sure that everything works as usual

### Notes for release

n/a


packages/sanity/src/core/store/_legacy/grants/grantsStore.ts
```diff
@@ -65,11 +61,11 @@ async function matchesFilter(
 /** @internal */
 export interface GrantsStoreOptions {
   client: SanityClient
-  currentUser: CurrentUser | null
+  userId: string | null
 }
 
 /** @internal */
-export function createGrantsStore({client, currentUser}: GrantsStoreOptions): GrantsStore {
+export function createGrantsStore({client, userId}: GrantsStoreOptions): GrantsStore {
```
jtpetty, line 68: I know these are marked `@internal` but that doesn't mean a customer has not taken a dependency on them. I like this approach as it is cleaner, but we may need to be safer. A couple of alternatives could be:
* accept either userId or user.id if passed in
* use reflection to see if `user` is a string and use it as an id, otherwise use it as a User object
* have two separate methods, one that takes an id and one that takes a User object

Am I overthinking this @rexxars ?

jtpetty: Also, is there an interface that can be leveraged for the different classes of User?

bjoerge:
@hermanwikner what's the status on this one?

hermanwikner:
> @hermanwikner what's the status on this one? Is it still relevant?

Yes, this PR is still relevant for the comments feature

packages/sanity/src/core/store/_legacy/grants/grantsStore.ts
jtpetty: That feels safer. Can I ask, do we need two separate interfaces?

pedrobonamin (approved):
Looks good to me!

rexxars (approved):

ricokahler (approved):
Nice! I love these kinds of refactors.

PR #5300, opened 2023-11-30
fix(i18n): sort keys and adjust inconsistently named properties

rexxars:
### Description

To ease automation and findability, it makes sense for the keys to be alphabetically sorted. Once I made this change, I discovered a few inconsistently named items, which I have fixed.

There were also a few camelCased and dynamic strings being used, which does not align with the rest of the identifiers - I took the liberty of changing these and putting them under a subgroup (`timeline.operation`)

There may be some inconsistencies in the Norwegian translations, but these will be cleared up as they are moved out of the monorepo soon.
 
### Notes for release

None

### DIFF
--- a/dev/test-studio/plugins/locale-no-nb/bundles/structure.ts
+++ b/dev/test-studio/plugins/locale-no-nb/bundles/structure.ts
@@ -319,7 +319,7 @@ const structureLocaleStrings: Record<StructureLocaleResourceKeys, string> = {
   'buttons.action-menu-button.aria-label': 'Åpne dokumenthandlinger',
 
   /** the placeholder text for the search input on the inspect dialog */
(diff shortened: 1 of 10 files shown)


jtpetty:
> Nice! Would it make sense to add a lint rule to keep the sort order enforced?

I agree. This could be done in a follow up though. I found this lint rule: https://eslint.org/docs/latest/rules/sort-keys#:~:text=This%20rule%20checks%20all%20property,all%20variables%20are%20sorted%20alphabetically.

rexxars:
> Nice! Would it make sense to add a lint rule to keep the sort order enforced?

There is! `/* eslint sort-keys: "error" */` is at the top of all these files :)

PR #5302, opened 2023-11-30
fix(core): not open popover when clicking link icon

ninaandal:
### Description

Creating an annotation will briefly show a non-editable popover, before then showing the 'full' popover. 

Steps to recreate:
- Select any text in a PTE field and click the 'link' button
- Observe how you'll see an initial popover, then it'll disappear and another one will appear in its place

The second popover (with editable fields) should appear immediately. When the editable popover is visible, clicking anywhere outside (even the selected text) should dismiss it.

### What to review
That the annotation popover and the `Edit link` dialog show when expected.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
@@ -264,6 +264,7 @@ export const DefaultAnnotationComponent = (props: BlockAnnotationProps) => {
   const hasError = validation.some((v) => v.level === 'error')
   const hasWarning = validation.some((v) => v.level === 'warning')
   const hasMarkers = markers.length > 0
+  const [showPopover, setShowPopover] = useState(false)
 
   const {t} = useTranslation()
   const toneKey = useMemo(() => {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
```diff
@@ -281,28 +282,50 @@ export const DefaultAnnotationComponent = (props: BlockAnnotationProps) => {
...
+      }
+    }
+
+    document.addEventListener('keydown', handleKeyDown)
+
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown)
+    }
+  }, [])
+
```
skogsmaskin, line 302: This should be dropped. It's taken care [of here](https://github.com/sanity-io/sanity/blob/edx-813/packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx#L80)

skogsmaskin (commented):
See my comment about the `useEffect`, other than that it LGTM.

Next steps here should be to try to speed up the loading of the Popover on new annotation objects. It's so slow...

packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
skogsmaskin: I can't see any different behaviour when removing it, but maybe I'm missing something?

skogsmaskin:
Would love for us to have some component tests on this.

packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx
ninaandal: No, you're right 🤔

PR #5305, opened 2023-11-30
feat(i18n): localize PTE activation message

binoy14:
### Description

Localizes activation message for PTE on hover or on tap hover.

There seems to be uncommitted `yarn.lock` from next that got included as well

### What to review

If keys makes sense and logic makes sense

### Notes for release

(probably part of larger i18n release notes)

### DIFF
--- a/packages/sanity/src/core/form/components/ActivateOnFocus/ActivateOnFocus.tsx
+++ b/packages/sanity/src/core/form/components/ActivateOnFocus/ActivateOnFocus.tsx
@@ -61,16 +61,20 @@ export function ActivateOnFocus(props: ActivateOnFocusProps) {
 
   const msg = useMemo(() => {
     const isTouch = isTouchDevice()
-    let activateVerb = isTouch
-      ? t('inputs.portable-text.activate-on-focus.tap-verb')
-      : t('inputs.portable-text.activate-on-focus.click-verb')
-    if (focused && !isTouch) {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/components/ActivateOnFocus/ActivateOnFocus.tsx
```diff
@@ -59,13 +61,19 @@ export function ActivateOnFocus(props: ActivateOnFocusProps) {
...
   const msg = useMemo(() => {
     const isTouch = isTouchDevice()
-    let activateVerb = isTouch ? 'Tap' : 'Click'
+    let activateVerb = isTouch
+      ? t('inputs.portable-text.activate-on-focus.tap-verb')
+      : t('inputs.portable-text.activate-on-focus.click-verb')
     if (focused && !isTouch) {
-      activateVerb += ' or press space'
+      activateVerb += t('inputs.portable-text.activate-on-focus.focused-verb')
     }
```
jtpetty, line 69: This should probably be two separate strings with a context passed in as you can't just concatenate words like this in every language. Similar to [this](https://github.com/sanity-io/sanity/pull/5140/files) PR

jtpetty (dismissed):
for this, we should use context instead of concatenating strings

rexxars (approved):
Looks good - thanks!

PR #5308, opened 2023-11-30
feat(i18n): localize menu items

ricokahler:
### Description

This PR enables the translation of menu items, menu item groups, schema orderings, templates, initial template values, lists, list items, field actions, and user defined components.

### What to review

I introduced a lot of `i18n` keys in the structure builder. Does it all propagate correctly? Did I miss anything?

### Notes for release

N/A


packages/sanity/src/core/i18n/i18nConfig.ts
```diff
@@ -71,8 +71,10 @@ function createI18nApi({
       },
       loadNamespaces(namespaces: string[]): Promise<void> {
         const missing = namespaces.filter((ns) => !i18nInstance.hasLoadedNamespace(ns))
-        return missing.length === 0 ? Promise.resolve() : i18nInstance.loadNamespaces(namespaces)
+        return missing.length === 0 ? Promise.resolve() : i18nInstance.loadNamespaces(missing)
```
rexxars, line 74: Oops, thanks for catching!

rexxars (changes requested):
1. One alternative approach to resolving the title in the actual structure builder setup is to defer that to render time, at which point the namespace will be loaded, or if it isn't, it will trigger a suspense boundary to be shown.
<br/>This is going to be more work to implement, as it will require anything that renders a structure nodes' title to be aware of the i18n keys etc - but that _may_ be an acceptable trade-off? Alternatively, could we resolve the keys as part of a pre-rendering step? Let's discuss.

3. Wonder if we should have an importable component that loads namespaces instead doing the `i18nNamespacesToLoad` approach.
<br/>I could see us wanting to implement the same thing on different surfaces, and having to come up with a way to pipe these down to the locale provider seems a bit tedious.
<br/>A component that takes an array of namespaces to load and checks whether or not they are, then calls `loadNamespaces` and throws a promise to trigger suspense would be my suggestion. The structure (desk) tool could then utilize that.
3. If the user specifies an i18n key for custom sort items, they have to use the namespace prefix (`myNamespace:`) for the keys, which is fair enough - but unless they "break into" the `structure` namespace, that custom namespace will not be loaded. The solution outlined in my first point would solve this, I believe.

rexxars (approved):

PR #5314, opened 2023-12-01
fix(form): recalculate virtualized descendants when selected group changes

### LINKED ISSUE #4687: Items dissapear from Array fields, only when viewed via a tab, since upgrading to latest Sanity studio
A very weird bug here. I have upgraded to the latest sanity: + sanity@3.14.1
and now, my array fields lose some of the items, temporarily when viewing via a tab. When I open a modal, or create a new array item, the missing items appear behind again, but then re-dissapear when the modal is closed. When i view, via the 'all it
I have recorded a video to explain this better as its very strange, but really annoying.
I can confirm there is no custom publish actions or anything like that on this project. 

https://github.com/sanity-io/sanity/assets/36191462/2e355836-a06d-4bfc-acdb-9628dbb03f78

I have looked in the console, and can confirm they are missing from the dom, not just from view somehow. as you can see, there are only two nodes here, not four.
Screenshot 2023-07-11 at 15 46 35

juice49:
### Description

This branch uses the same approach Binoy employed to fix list virtualisation at the document level, but applies it to all object inputs. It wraps object inputs in a React fragment that uses the selected group name as its key. This causes a unique element to be rendered for each selected group, and therefore all virtualised descendants are recalculated when the selected tab changes.

Adding the key at this level solves the problem both at the document level, and inside modals.

This branch also removes Binoy's fix, which is now redundant.

Fixes #4687.

**Top-level groups before fix**

https://github.com/sanity-io/sanity/assets/1454914/6529d228-63c7-4770-9a09-2e3e0c997a5d

**Top-level groups after fix**

https://github.com/sanity-io/sanity/assets/1454914/573e66ca-17e1-486f-b816-f334058e2772

**Modal groups before fix**

https://github.com/sanity-io/sanity/assets/1454914/508ff955-778f-43bb-b099-4dc66dc51a8b

**Modal groups after fix**

https://github.com/sanity-io/sanity/assets/1454914/84b2f08d-9d49-401a-b389-f05b2da3f1b8

### What to review

Open [the "Virtualisation Debug" section](http://localhost:3333/test/structure/input-debug;virtualizationDebug;01de0cf0-1686-49f5-8212-a33b94edbfbf) in the Test Studio, and navigate between groups inside the example documents. Regardless of whether viewing a top-level document or a modal, there should never be any erroneously hidden array items. This is demonstrated in the videos above.

### Notes for release

Fixed virtualization issue which caused array items to erroneously be hidden when navigating between groups inside a modal.

### DIFF
--- a/packages/sanity/src/core/form/inputs/ObjectInput/ObjectInput.tsx
+++ b/packages/sanity/src/core/form/inputs/ObjectInput/ObjectInput.tsx
@@ -1,4 +1,4 @@
-import React, {memo, useCallback, useMemo} from 'react'
+import React, {Fragment, memo, useCallback, useMemo} from 'react'
 import {Grid, Stack} from '@sanity/ui'
 import {ObjectInputProps} from '../../types'
 import {ObjectInputMembers} from '../../members'
@@ -47,6 +47,8 @@ export const ObjectInput = memo(function ObjectInput(props: ObjectInputProps) {
     return <UnknownFields fieldNames={unknownFields} value={value} onChange={onChange} />
   }, [onChange, schemaType.fields, value])
 
+  const selectedGroup = useMemo(() => groups.find(({selected}) => selected), [groups])
+
   const renderObjectMembers = useCallback(
     () => (
       <ObjectInputMembers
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
Nice one! Thank you @juice49 🙏🏼

PR #5323, opened 2023-12-04
fix(i18n): key being used instead of translated value

rexxars:
### Description

An oversight on my part - when selecting a revision, we were not actually passing the i18n key to the `t` function, but rather showing it as-is. This fixes that.

### What to review

Review changes shows the correct label when selecting an operation.

### Notes for release

None


binoy14 (approved):

robinpyon (commented):
Thanks @rexxars – ideally, this button text should display the selected date _in addition_ to the action. Without it, users have to either manually re-open the dropdown to determine the date of the current revision (or look in the URL)

We already do this with revision changes right now, would it be possible for this to follow suit too? (To match how it was previously in `3.20.2`)

`next`

image

`3.20.2`

image

rexxars:
> We already do this with revision changes right now, would it be possible for this to follow suit too? (To match how it was previously in `3.20.2`)

Certainly, I missed that. Will fix.

PR #5326, opened 2023-12-05
fix(i18n): correct permission check banner interpolation

rexxars:
### Description

Fixes the "permission check" banner at the top of a document when you do not have the sufficient permissions to create or edit a document. Previously it showed something like this:

image

This is because interpolating JSX elements won't work.

This PR:

1. Uses the list formatter to split into parts and makes it available as `<Roles/>`
2. Uses context to define the different permissions available (`create`, `update`) so they can be translated to other languages
3. Uses the `count` parameter to differentiate between one and multiple roles, to align with other pluralization keys

### What to review

Permission checks now render correctly

### Notes for release

None


jtpetty (approved):
LGTM

PR #5327, opened 2023-12-05
chore(deps): upgrade jsdom dependency

### LINKED ISSUE #5316: [sanity] Update dependency jsdom to remove deprecated packages.
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

The normal Sanity npm package (`sanity@3.20.2`) uses `jsdom@20.0.3`, which depends on two deprecated packages: `abab` and `domexception`, which both have native variants now. `jsdom@23.0.0` removed both of these dependencies, and as such should be preferred. `jsdom-global` should work just fine with the new version.

**To Reproduce**

Simply installing the package will produce deprecation errors, like the following:
image

**Expected behavior**

No warnings when installing `sanity`.

3.20.2

Windows 10

Node: `v20.3.1`
npm: `9.6.7`
pnpm: `8.6.6`

rexxars:
### Description

Fixes #5316 (deprecated packages warning)

### What to review

Block tools and CLI exec command and family still works

### Notes for release

None

### DIFF
--- a/packages/@sanity/block-tools/package.json
+++ b/packages/@sanity/block-tools/package.json
@@ -63,7 +63,7 @@
     "@sanity/types": "3.20.2",
     "@types/jsdom": "^20.0.0",
     "@types/react": "^18.2.37",
-    "jsdom": "^20.0.0"
+    "jsdom": "^23.0.1"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
Thanks!

PR #5328, opened 2023-12-05
fix(i18n): use correct keys for file asset source headings and accepted types

robinpyon:
### Description

This PR ensures correct i18n keys are used within the default file asset source. It also fixes an issue with accepted types not displaying correctly.

### What to review

- When browsing assets in the default file asset source (on mobile breakpoints) – file headers and the show usage button should render correct localized text
- Accepted types should also render correctly

**Before**

image

**After**

image

### Notes for release

N/A


rexxars (approved):
Much appreciated!

PR #5338, opened 2023-12-07
fix(i18n): prefer narrow relative time formatting in `useRelativeTime`

robinpyon:
### Description

This PR updates `useRelativeTime` to prefer `narrow` styling for both Intl.RelativeTimeFormat's styling and Intl.NumberFormat's unitDisplay when `minimal=true`.

In other words, it'll display `3 wk. ago` as `3w ago`, which was how relative dates were displayed prior to recent i18n work.

### What to review

That relative dates in the studio display their more condensed versions. 

Whilst these minimal dates aren't used much in the current studio (primarily in the document status bar / sparkline), they're a little more prevalent in upcoming facelift work.

### Notes for release

- Fixed an issue where relative dates were not displaying fully condensed versions


juice49 (approved):
Looks great!

PR #5343, opened 2023-12-10
feat: expose all locale info from `useCurrentLocale` hook

rexxars:
### Description

The `useCurrentLocale` hook currently only exposes the locales' ID. As we mature the API, this won't be all the information we want to get from it, and thus we should change this now so we can expand in the future.

Introduces a new `Locale` type that is basically `LocaleDefinition` but without `bundles`. Again, as we mature we may want to split what the user passes as input for the locale from what the framework exposes, such as computed values (calendar info, date formatting rules, pluralization info etc)

### Notes for release

None (not yet public)


jtpetty (approved):
LGTM

ricokahler (approved):

PR #5348, opened 2023-12-11
fix(core): fields actions position shift

pedrobonamin:
### Description

Update how the`<FieldPresence>` element is positioned and remove the` width:0 `from the `<FieldActionFloatingCard>` as it is causing layout to shift in an unexpected way.

Now the Root will add the position to the field presence element, this was moved to the Root because we need to handle the case of focus-within. This is the nearest parent we can choose. 

https://github.com/sanity-io/sanity/assets/46196328/cff41096-dc96-4ca6-92c8-57af45ce98ae

https://github.com/sanity-io/sanity/assets/46196328/e6fce327-1f37-4283-ade3-35b04751bd78

### DIFF
--- a/packages/sanity/src/core/form/components/formField/FormFieldBaseHeader.tsx
+++ b/packages/sanity/src/core/form/components/formField/FormFieldBaseHeader.tsx
@@ -15,19 +15,29 @@ import {FieldCommentsProps} from '../../types'
 
 const TOOLTIP_GROUP_DELAY: TooltipDelayGroupProviderProps['delay'] = {open: 500}
 
-const Root = styled(Flex)`
-  /* Prevent buttons from taking up extra vertical space */
-  line-height: 1;
-  /* For floating actions menu */
-  position: relative;
-`
-
-const PresenceBox = styled(Box)<{$right: number}>(({theme, $right}) => {
+const Root = styled(Flex)<{
+  $floatingCardWidth: number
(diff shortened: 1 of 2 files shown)


hermanwikner:
It works great on hover, but I discovered two issues with presence positioning on **touch devices** and when using **keyboard navigation** to focus on field actions.

**Touch devices:**
In touch devices, the field actions should always be visible, but presences is not correctly positioned in relation to the field actions.

actions


**Focusing on field actions with the keyboard:**
When using the keyboard to focus on field actions, the position of presence is not correctly updated in relation to the field actions.


https://github.com/sanity-io/sanity/assets/15094168/26b4e93b-f8f0-44c2-9b68-ccaf59273e61


https://github.com/sanity-io/sanity/assets/15094168/73d6f2ea-5a44-45a6-941d-b65011ca2dd2

juice49:
I grabbed the latest code, and I seem to be experiencing an issue in Chrome 120.0.6099.71.

https://github.com/sanity-io/sanity/assets/1454914/a4a5f43f-7770-4a49-ad4f-51a952c59b03

pedrobonamin:
@hermanwikner I'm merging this as it's going to facelift and the solution is what we discussed in the meeting. 
Any extra feedback thing we should consider can be checked in the facelift branch before deploying that one.

PR #5351, opened 2023-12-11
feat(core): make calendars aware of locale week start day

rexxars:
### Description

*Note*: builds on #5343

Currently, the date/datetime/date filter calendars always start the shown weeks on Sundays. This is fine for some locales, but  for others it is more natural to show the week starting on Mondays.

This PR introduces a `weekInfo` object to locale definitions. Theoretically it should be inferable using `new Intl.Locale('en-US').weekInfo` / `new Intl.Locale('en-US').getWeekInfo()`, but it seems browser support is less than ideal, to say the least, so better to define this up front.

### What to review

Calendars still start on Sunday (test studio currently does not have additional locales installed, so hard to test).

### Notes for release

None

### DIFF
--- a/packages/sanity/src/core/i18n/types.ts
+++ b/packages/sanity/src/core/i18n/types.ts
@@ -149,6 +149,34 @@ export interface Locale {
    * The title of locale, eg `English (US)`, `Norsk (bokmål)`, `ไทย`…
    */
   title: string
+
+  /**
+   * Week information for this locale. Based on the `Intl.Locale['weekInfo']` type.
+   */
+  weekInfo: LocaleWeekInfo
+}
+
+/**
+ * An object representing week information associated with the Locale data specified in
+ * {@link https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Patterns_Week_Elements | UTS 35's Week Elements }
+ *
+ * @public
+ */
+export interface LocaleWeekInfo {
+  /**
+   * An integer indicating the first day of the week for the locale. Can be either 1 (Monday) or 7 (Sunday).
+   */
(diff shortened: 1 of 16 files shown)


packages/sanity/src/core/i18n/types.ts
```diff
@@ -137,6 +136,42 @@ export interface LocaleDefinition {
...
+ * An object representing week information associated with the Locale data specified in
+ * {@link https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Patterns_Week_Elements | UTS 35's Week Elements }
+ *
+ * @public
+ */
+export interface LocaleWeekInfo {
+  /**
+   * An integer indicating the first day of the week for the locale. Can be either 1 (Monday) or 7 (Sunday).
+   */
+  firstDay: 1 | 7
```
ricokahler, line 155: what happens if another number besides 1 or 7 is provided? would it be better to have these as strings e.g. `firstDay: 'monday' | 'sunday'`?

ricokahler (approved):
very interesting. i like that you included the link to [this](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Patterns_Week_Elements) in the comments too!

packages/sanity/src/core/i18n/types.ts
rexxars: > what happens if another number besides 1 or 7 is provided? would it be better to have these as strings e.g. `firstDay: 'monday' | 'sunday'`?

Could maybe validate it in the `defineLocale` call? This is just following the `Intl` types. While I agree that spelling it out seems easier to read, I want to automate this information, so I'll stick with what that API gives us.

PR #5352, opened 2023-12-12
fix(core): add timestamps to all preview components

ninaandal:
### Description
Unless explicitly added, the `updatedAt` prop was not included in the previews, hence not showing a timestamp for when it was last edited/published.

### What to review
That the list previews (and all previews really) show a correct timestamp of last edited/published

### Notes for release
Fixes issue where timestamp was not included in last edited/published in list previews

### DIFF
--- a/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
+++ b/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
@@ -100,11 +100,7 @@ export function getReferenceInfo(
             } as const)
           }
 
-          const previewPaths = [
-            ...(getPreviewPaths(refSchemaType?.preview) || []),
-            ['_updatedAt'],
-            ['_createdAt'],
-          ]
+          const previewPaths = [...(getPreviewPaths(refSchemaType?.preview) || [])]
 
           const draftPreview$ = documentPreviewStore.observePaths(draftRef, previewPaths).pipe(
             map((result) =>
--- a/packages/sanity/src/core/form/studio/inputs/crossDatasetReference/datastores/getReferenceInfo.ts
+++ b/packages/sanity/src/core/form/studio/inputs/crossDatasetReference/datastores/getReferenceInfo.ts
@@ -74,11 +74,7 @@ export function createGetReferenceInfo(context: {
           (candidate) => candidate.type === resolvedDoc._type,
         )
 
-        const previewPaths = [
-          ...(getPreviewPaths(refSchemaType?.preview) || []),
-          ['_updatedAt'],
-          ['_createdAt'],
-        ]
+        const previewPaths = [...(getPreviewPaths(refSchemaType?.preview) || [])]
 
         const publishedPreview$ = documentPreviewStore
           .observePaths(doc as Previewable, previewPaths, apiConfig)
--- a/packages/sanity/src/core/preview/utils/getPreviewPaths.ts
+++ b/packages/sanity/src/core/preview/utils/getPreviewPaths.ts
@@ -1,15 +1,18 @@
 import {SchemaType} from '@sanity/types'
 import {PreviewPath} from '../types'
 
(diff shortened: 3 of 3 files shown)


pedrobonamin (approved):

packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
```diff
@@ -100,11 +100,7 @@ export function getReferenceInfo(
             } as const)
           }
 
-          const previewPaths = [
-            ...(getPreviewPaths(refSchemaType?.preview) || []),
-            ['_updatedAt'],
-            ['_createdAt'],
-          ]
+          const previewPaths = [...(getPreviewPaths(refSchemaType?.preview) || [])]
```
bjoerge, line 103: Can be simplified to:
```suggestion
          const previewPaths = getPreviewPaths(refSchemaType?.preview) || []
```

packages/sanity/src/core/form/studio/inputs/crossDatasetReference/datastores/getReferenceInfo.ts
```diff
@@ -74,11 +74,7 @@ export function createGetReferenceInfo(context: {
           (candidate) => candidate.type === resolvedDoc._type,
         )
 
-        const previewPaths = [
-          ...(getPreviewPaths(refSchemaType?.preview) || []),
-          ['_updatedAt'],
-          ['_createdAt'],
-        ]
+        const previewPaths = [...(getPreviewPaths(refSchemaType?.preview) || [])]
```
bjoerge, line 77: Can be simplified to:
```suggestion
        const previewPaths = getPreviewPaths(refSchemaType?.preview) || []
```

binoy14 (commented):
I am not really sure what changed could you maybe share a screenshot if possible just for my understanding

packages/sanity/src/core/preview/utils/getPreviewPaths.ts
```diff
@@ -1,11 +1,18 @@
...
 
   if (!selection) return undefined
 
-  return Object.values(selection).map((value) => String(value).split('.'))
+  // Transform the selection dot-notation paths into array paths.
+  // Example: ['object.title', 'name'] => [['object', 'title'], ['name']]
+  const paths = Object.keys(selection).map((value) => String(value).split('.')) || []
+
+  // Return the paths with the default preview paths appended.
+  return paths.concat(DEFAULT_PREVIEW_PATHS)
```
bjoerge, line 17: This will add `_createdAt` and `_updatedAt` to objects types as well as documents. It's not a huge deal (I think), but would be preferable if we could add them only for document types.

bjoerge (commented):
Nice catch. I left a couple of suggestions. Especially the one about selecting `_createdAt` and `_updatedAt` for documents only would be great if you could look into. Not a blocker for merge if it turns out to be tricky.

ninaandal:
> I am not really sure what changed could you maybe share a screenshot if possible just for my understanding

Hi @binoy14  - sorry for being vague here! The problem was that when hovering over the document list, it didn't add a timestamp to the popover for when it was last changed unless the preview explicitly included the `updatedAt` prop. I'll add screenshots to the PR description as well

Screenshot 2023-12-13 at 12 01 10

packages/sanity/src/core/preview/utils/getPreviewPaths.ts
ninaandal: I see and I've been looking into this - but I'm wondering if it's actually necessary, as previews will only be created for documents and not for objects, no?

bjoerge: Ah, you're absolutely right. This happens right before fetching (I initially thought this happened at the "generate default preview config" phase), and we never have to fetch anything when previewing objects. Thanks for clearing that up!

bjoerge (approved):

PR #5354, opened 2023-12-12
feat(cli): add `schemaId` flag for initialising projects from schema builder

RostiMelk:
### Description

This PR adds a hidden `schemaId` flag to the CLI, specifically for the `init` command. This flag allows a user to initialize a project from the schema builder, overring the default behavior of prompting the user for a template.

This was initially supposed to be a hidden flag, but after discussion I have added a new help text to the `init` command
```shell
--schemaId <schemaId>: Init studio from a Schema Builder schema. Start your schema design at: https://schema.club. (alpha)
```

### What to review

Run `sanity init --schemaId ZaEh6uHn` and verify that the project is initialized from the schema builder.

### Notes for release

~No notes for release. This is a hidden flag, so it should not be documented in the CLI help.~

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/bootstrapTemplate.ts
+++ b/packages/@sanity/cli/src/actions/init-project/bootstrapTemplate.ts
@@ -11,10 +11,16 @@ import {createCliConfig} from './createCliConfig'
 import {createStudioConfig, GenerateConfigOptions} from './createStudioConfig'
 import type {ProjectTemplate} from './initProject'
 import templates from './templates'
+import {
+  assembeIndexContent,
+  builderSchemaToFileContents,
+  fetchBuilderSchema,
+} from '../../util/builderSchema'
 
 export interface BootstrapOptions {
   packageName: string
   templateName: string
+  builderId?: string
   outputPath: string
   useTypeScript: boolean
   variables: GenerateConfigOptions['variables']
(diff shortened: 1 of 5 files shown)


kmelve:
Would `--schemaId` be more intuitive?

jtpetty:
> Would `--schemaId` be more intuitive? It's not the “builder” that you're IDing. And folk's mental model here is that they have a "schema" they want to add to the studio?

I agree with that. Also, this is an open source repo so nothing is truly "hidden". Is there a way to explicitly add a comment that this flag is experimental and should not be used externally? Not documenting something is ambiguous. Is it not documented to "hide" it or is it not documented because someone forgot to document it? Being explicit is preferred in this situation I think.

RostiMelk:
> Would `--schemaId` be more intuitive? It's not the “builder” that you're IDing. And folk's mental model here is that they have a "schema" they want to add to the studio?

Renamed now.

RostiMelk:
> > Would `--schemaId` be more intuitive? It's not the “builder” that you're IDing. And folk's mental model here is that they have a "schema" they want to add to the studio?
> 
> I agree with that. Also, this is an open source repo so nothing is truly "hidden". Is there a way to explicitly add a comment that this flag is experimental and should not be used externally? Not documenting something is ambiguous. Is it not documented to "hide" it or is it not documented because someone forgot to document it? Being explicit is preferred in this situation I think.

Added helper

packages/@sanity/cli/src/actions/init-project/bootstrapTemplate.ts
```diff
@@ -49,6 +55,23 @@ export async function bootstrapTemplate(
     await fs.copyFile(path.join(sharedDir, 'tsconfig.json'), path.join(outputPath, 'tsconfig.json'))
   }
 
+  // If we have a builder ID, the template is assembled from the builder schema
```
jtpetty, line 58: nit: this comment should probably be updated

jtpetty:
> > > Would `--schemaId` be more intuitive? It's not the “builder” that you're IDing. And folk's mental model here is that they have a "schema" they want to add to the studio?
> > 
> > 
> > I agree with that. Also, this is an open source repo so nothing is truly "hidden". Is there a way to explicitly add a comment that this flag is experimental and should not be used externally? Not documenting something is ambiguous. Is it not documented to "hide" it or is it not documented because someone forgot to document it? Being explicit is preferred in this situation I think.
> 
> Added helper

My apologies, my intention wasn't to add it to the help text, but to document it in the code that we are explicitly not officially supporting this yet. This is an experiment and maybe down the road we officially support it once schema club is out of alpha.

RostiMelk:
> > > > Would `--schemaId` be more intuitive? It's not the “builder” that you're IDing. And folk's mental model here is that they have a "schema" they want to add to the studio?
> 
> > > 
> 
> > > 
> 
> > > I agree with that. Also, this is an open source repo so nothing is truly "hidden". Is there a way to explicitly add a comment that this flag is experimental and should not be used externally? Not documenting something is ambiguous. Is it not documented to "hide" it or is it not documented because someone forgot to document it? Being explicit is preferred in this situation I think.
> 
> > 
> 
> > Added helper
> 
> 
> 
> My apologies, my intention wasn't to add it to the help text, but to document it in the code that we are explicitly not officially supporting this yet. This is an experiment and maybe down the road we officially support it once schema club is out of alpha.

Ah okay, this makes a lot more sense. I'll do that, thanks 👍

packages/@sanity/cli/src/actions/init-project/bootstrapTemplate.ts
RostiMelk: Updated.

PR #5356, opened 2023-12-12
fix(cli): add (experimental) support for bun as package manager

bjoerge:
### Description
Adds bun as (experimental) package manager to use for sanity projects

image

### What to review
Is marking it as experimental necessary? My reasoning was that we don't expclitly test with `bun` and (afaik) haven't aquired much data/confidence about adoption/error rates.

### Notes for release

- Add bun as possible package manager to use for new sanity projects

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -503,6 +503,7 @@ export default async function initSanity(
     yarn: 'yarn dev',
     npm: 'npm run dev',
     pnpm: 'pnpm dev',
+    bun: 'bun dev # experimental',
     manual: 'npm run dev',
   }
   const devCommand = devCommandMap[pkgManager.chosen]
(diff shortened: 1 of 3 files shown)


ricokahler (approved):
nice

bjoerge:
> Did you run this?

😅 

I definitely should, and have now! I pushed a couple of commits based on your suggestions and published a tagged release,  ran it e2e and it works now!

Can be tested using `bun create sanity@canary`
I've verified that:
- It detects that bun is installed on the system
- It picks up bun as the currently running package manager and pre-selects it
- Dependencies are installed correctly in the target project
- Running `bun dev` in the studio folder runs the studio locally

Note: when running the cli, it currently says at the end:

> First: cd <project> - to enter project’s directory
> Then: bun dev **# experimental** - to run Sanity Studio

I've removed the _#experimental_ part here since it's already mentioned during project creation.

rexxars (approved):

PR #5361, opened 2023-12-13
fix(portable-text-editor): fix safari focus issue on empty editor

sjelfull:
### Description

This fixes a issue in Safari where focus in a activated, empty PTE input a second time won't allow you to type unless you click the "Empty" placeholder precisely.

### What to review

- That focus behaviour behaves the same was as before in all the normal browsers
- That focus activates properly on repeated tabbing into the same empty input

### Notes for release

- Fixes an issue where focus on a empty Portable Text Input won't work consistently in Safari


skogsmaskin (changes requested):
Seems to break stuff on Chrome, see my screenshot.

rexxars:
@sjelfull Is this still relevant? /cc @christianhg

PR #5369, opened 2023-12-14
Edx 802

hermanwikner:
### Description

This pull request updates the implementation that displays the popover informing about the renaming of `desk` to `structure`. Currently, this is handled by providing a custom icon component, which is wrapped with a popover. However, the icon will soon be removed from the tool menu buttons, and we will, therefore, need to update how this popover is rendered.

The solution that this pull request implements is to utilize the tool options property in `deskTool` to configure the popover component. Then, inside the tool menu component(s), we check if this popover is configured and render it.

Note that this solution is not very elegant and should be considered a temporary fix that will soon be removed. However, from exploring how to approach this, this seems to be the most reliable and stable solution for now.

### What to review

- The popover should be rendered as before, however, the positioning might be slightly different, as the reference element is no longer the icon but the children of the button. 
- Could this be solved differently?

### Notes for release

N/A


pedrobonamin (approved):
Thanks for this @hermanwikner !

packages/sanity/src/desk/deskTool.ts
```diff
@@ -119,7 +119,11 @@ export const deskTool = definePlugin<DeskToolOptions | void>((options) => {
         getIntentState,
         // Controlled by sanity/src/desk/components/deskTool/DeskTitle.tsx
         controlsDocumentTitle: true,
-        options,
+        options: {
+          ...options,
```
pedrobonamin, line 123: My only concern here is that options takes any as prop.
And this spread will return you something you won't be expecting.
Is it possible to update the options type to something like:

```
interface Options = {
 [key:string]: any;
 /**
 * @internal
 * /
 __internal_toolMenuButtonWrapper?: React.ReactNode
}
```

binoy14 (approved):
LGTM

ninaandal (approved):
Thank you Herman 🙌

packages/sanity/src/desk/deskTool.ts
hermanwikner: [discussed this topic in a meeting]

PR #5374, opened 2023-12-15
fix(facelift): self serve Inter font

pedrobonamin:


robinpyon:
We should use the web optimised WOFF 2.0 versions of this typeface, as it provides much better compression and is [well supported today](https://caniuse.com/woff2).

PR #5375, opened 2023-12-15
fix(desk): add tooltip on menuitem when disabled only

ninaandal:
### Description
A tooltip should only be visible if the menu item is disabled


pedrobonamin (approved):
Thanks @ninaandal 🙌🏻

PR #5381, opened 2023-12-16
feat(i18n): use last registered locale as default

rexxars:
### Description

Until we land schema localization, choosing a different language has limited value. For now, select the last registered locale and use that, disabling the choice. Will fall back to English (US) on missing strings.

### What to review

- Enabling a locale plugin will use that as the default (last one defined wins)

### Notes for release

None.


jtpetty (approved):
👍

PR #5382, opened 2023-12-16
feat(i18n): resolve preferred locale from browser languages

rexxars:
### Description

Note: builds on #5381 

Uses browsers' `navigator.languages` to resolve the preferred locale to use. Explicit user preference still has preference, and we still fall back to the default locale (English (US)) if none of the preferred locales can be found.  

It also adds the ability for users to explicitly set an array of preferred languages (in order of preference), or use the reducer pattern we've established in configuration to "dynamically" resolve it.

Made somewhat more complicated than I hoped because browsers use macro languages such as `no` if not explicitly setting "Norwegian (Bokmål)" for instance, so had to introduced a `macroLanguage` prop to locales that will need to be set for `nb-NO` and `nn-NO` packs (and possibly others).

Note: This also reverts back to showing all available locales, and is likely to linger here until we land schema localization, thus marked as draft.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -95,7 +95,7 @@ const sharedSettings = definePlugin({
     },
   },
   plugins: [
-    // ...localePlugins,
+    ...localePlugins,
     deskTool({
       icon: BookIcon,
       structure,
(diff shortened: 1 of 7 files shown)


jordanl17:
Closing until schema localization work is undertaken

PR #5384, opened 2023-12-18
feat: conditionally set maximum search depth in studio search

sjelfull:
### Description

This adds support for partial indexing, and limits the max field depth we use when searching in studio omnisearch. By default we search up to 5 levels deep.

If enabled, the studio will pull the max field depth from the the index settings API on startup, and use that instead. This makes sure we won't search fields that isn't indexed. Search performance will be impacted negatively if we try to search fields that isn't indexed.

To enable partial indexing support in studio:
```js
// ... rest of workspace/plugin config
search: {
  unstable_partialIndexing: {
    enabled: true,
  },
},
```

The document list, reference and cross dataset reference input search has been updated to respect the field depth limit.

Note: For CDR you have to enable partial indexing in the workspace you are in for it to check if the target dataset has partial indexing turned on.

### What to review

- That all fields up to 5 levels deep is searched within in the normal test studio workspace.
- That fields up to 3 levels deep is searched within the new Partial Indexing workspace.
- That the document list in the desk tool respects the same settings
- That the reference input respects the same settings
- That the CDR input respects the settings in the target dataset

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/theme/_legacy/theme.ts
+++ b/packages/sanity/src/core/theme/_legacy/theme.ts
@@ -1,5 +1,7 @@
 import {black, blue, gray, green, red, white, yellow} from '@sanity/color'
 import {studioTheme as defaults} from '@sanity/ui'
+// eslint-disable-next-line camelcase
+import {v0_v2} from '@sanity/ui/theme'
 import {StudioTheme} from '../types'
 import {buildColor} from './color'
 import {buildFonts} from './fonts'
@@ -38,11 +40,11 @@ export function buildLegacyTheme(partialLegacyTheme: Partial<LegacyThemeProps>):
 
   const color = buildColor(legacyPalette, legacyTones)
   const fonts = buildFonts(legacyTheme)
-
-  return {
+  const theme: StudioTheme = {
     __dark: _isDark(color.light.default.base.bg, color.light.default.base.fg),
     __legacy: true,
     ...defaults,
+    v2: undefined,
     color,
     focusRing: {
       offset: -1,
@@ -56,6 +58,9 @@ export function buildLegacyTheme(partialLegacyTheme: Partial<LegacyThemeProps>):
       parseInt(legacyTheme['--screen-xlarge-break'], 10) || 1600,
     ],
   }
+  // Override v2 default theme with the adapter from v0 theme to v2 theme
+  theme.v2 = v0_v2(theme)
+  return theme
 }
 
 const defaultCustomProperties: LegacyThemeProps = {


packages/sanity/src/core/studio/components/navbar/search/hooks/useSearchMaxFieldDepth.ts
```diff
@@ -0,0 +1,103 @@
...
+  const maxFieldDepth = (indexSettings?.settings?.partialIndexSettings?.maxFieldDepth ?? 1) - 1
+
+  /**
+   * If the maxFieldDepth is not a number or is -1 or 5, return the default value.
+   */
+  if (!isNumber(maxFieldDepth) || maxFieldDepth < 0) {
+    return DEFAULT_MAX_FIELD_DEPTH
+  }
+
+  return Math.min(maxFieldDepth, DEFAULT_MAX_FIELD_DEPTH)
```
sgulseth, line 102: Could this resolver be 1-1 with the api, where it returns a value between 0-5? And then we can apply the subtraction(-1) where we are using the field depth. The concern seems clearer that way

```suggestion

  const maxFieldDepth = indexSettings?.settings?.partialIndexSettings?.maxFieldDepth
  
  if (!isNumber(maxFieldDepth)) {
    return LOWER_MAX_FIELD_DEPTH
  }

  // Ensure we return a value between LOWER_MAX_FIELD_DEPTH and UPPER_MAX_FIELD_DEPTH
  return Math.max(Math.min(maxFieldDepth, UPPER_MAX_FIELD_DEPTH), LOWER_MAX_FIELD_DEPTH)
```

packages/@sanity/schema/src/legacy/searchConfig/resolve.ts
```diff
@@ -76,14 +89,16 @@ export function deriveFromPreview(type: {
       weight: PREVIEW_FIELD_WEIGHT_MAP[fieldName],
       path: select[fieldName].split('.'),
     }))
+    .filter((spec) => spec.path.length <= maxDepth)
```
sgulseth, line 92: Don't we need to subtract 1 since maxDepth here referes to the depth where 0 is the first element. Also check for `maxDepth !== -1`? 🤔 
```suggestion
    .filter((spec) => maxDepth === -1 || spec.path.length - 1 <= maxDepth)
```

sgulseth, line 92: nit: We also end up with three iteration on each preview select, we could optimize this by just using a for-loop:

```suggestion
const fields: {weight: number; path: (string | number)[]}[] = []
  for (const fieldName of Object.keys(select)) {
    if (!(fieldName in PREVIEW_FIELD_WEIGHT_MAP)) {
      continue
    }

    const path = select[fieldName].split('.')
    if (maxDepth > -1 && path.length - 1 > maxDepth) {
      continue
    }

    fields.push({
      weight: PREVIEW_FIELD_WEIGHT_MAP[fieldName],
      path,
    })
  }

  return fields
```

sgulseth (changes requested):

packages/@sanity/schema/src/legacy/searchConfig/resolve.ts
sjelfull: Good catch, thanks! Added tests to catch these.

sjelfull:
Have updated the PR with work-in-progress support for CDRs, will wrap that up over the holidays. The rest can be reviewed as is.

bjoerge (approved):

sgulseth (approved):

PR #5387, opened 2023-12-18
refactor(free-trial): fault tolerance, code conventions

rexxars:
### Description

Few refactorings;

- We should always check for the presence and ability to use localStorage before calling it. Strict browsers will throw an error when you try to access it with localStorage disabled.
- The request for the trial endpoint should be cancelled on unmount
- If the trial endpoint returns Portable Text blocks/marks/annotations that are not known, the default behavior is to warn. This warning is not something we want to expose to third-party users of the studio however, so lets disable it.
- We generally try to keep the context and providers as separate files, and have their exports match the filename where applicable. Also, the preferred naming convention is `SomethingContext` (actual React context returned from `createContext`) and `SomethingContextValue` (the interface/type that defines the shape of the context). 
- If an error occurs on the backend request, we don't want to emit a warning to the user - instead, let us suppress the error

### What to review

- Code changes make sense and is understood
- Code still does what was originally intended

### Notes for release

None


pedrobonamin (approved):

pedrobonamin:
Thanks for this @rexxars will take in consideration for next PRs.

PR #5394, opened 2023-12-19
Feat: draggable text blocks

skogsmaskin:
### Description

This change will allow any object type to be draggable in the PTE by returning `true` from `isDraggableType(schemaType)` which can be passed to the `Editable` props.

This will allow us to implement custom drag handles on regular text blocks, but not change the default.

It could be implemented for text blocks like this with the `renderBlock` prop for `Editable`:

```
renderBlock: (blockProps) => {
    return (
      <BlockDragWrapper draggable onDragStart={handleBlockDragStart}>
        <span data-custom-block-drag-handle="true" />
        {blockProps.children}
      </BlockDragWrapper>
    )
},
```

And the handleBlockDragStart

```
  // Only allow our own block drag handle to initiate block dragging.
  const handleBlockDragStart = useCallback((event: React.DragEvent<HTMLDivElement>) => {
    if (
      event.target === event.currentTarget &&
      !event.currentTarget.getAttribute('data-custom-block-drag-handle')
    ) {
      event.preventDefault()
    }
  }, [])
```

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -8,7 +8,7 @@ import {
   useSlate,
 } from 'slate-react'
 import {noop} from 'lodash'
-import {PortableTextBlock} from '@sanity/types'
+import {PortableTextBlock, SchemaType} from '@sanity/types'
 import {
   EditorChange,
   EditorSelection,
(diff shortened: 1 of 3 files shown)


rexxars:
@skogsmaskin @christianhg Could you figure out what the road forward for this PR is?

skogsmaskin:
This could still be relevant I think.

PR #5397, opened 2023-12-19
fix(cli): include command context in trace.completed/trace.failed

bjoerge:
### Description
Includes contextual info with all telemetry trace events

### Notes for release

n/a

### DIFF
--- a/packages/@sanity/cli/src/__telemetry__/cli.telemetry.ts
+++ b/packages/@sanity/cli/src/__telemetry__/cli.telemetry.ts
@@ -16,9 +16,9 @@ interface CLITraceData {
    */
   extraArguments: string[]
   coreOptions: {
-    help: boolean
-    debug: boolean
-    version: boolean
+    help?: boolean
+    debug?: boolean
+    version?: boolean
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/@sanity/cli/src/__telemetry__/cli.telemetry.ts
```diff
@@ -16,9 +16,9 @@ interface CLITraceData {
    */
   extraArguments: string[]
   coreOptions: {
```
bjoerge, line 18: Also excludes these ones if they are false/not given

packages/@sanity/cli/src/cli.ts
```diff
@@ -127,17 +127,17 @@ export async function runCli(cliRoot: string, {cliVersion}: {cliVersion: string}
   }
 
   const cliRunner = getCliRunner(commands)
-  const cliCommandTrace = telemetry.trace(CliCommand)
+  const cliCommandTrace = telemetry.trace(CliCommand, {groupOrCommand: args.groupOrCommand})
```
svirs, line 130: Could we throw in 
```
    extraArguments: args.extraArguments,
    commandArguments: args.argsWithoutOptions,
```
too?

bjoerge: Done

svirs (approved):

PR #5400, opened 2023-12-19
fix(cli): uniquify temp directory used during dataset import

jtpetty:
### Description

This change addresses an issue where a failed import could leave a temp directory around that would be used by subsequent invocations of the import command.

### What to review

I used the same method for uniquification that the export command uses. Should these be the same?


ricokahler (approved):
nice! I think this will fix the issue and shouldn't cause much more disk usage on re-runs because [it should delete the temp folder when it completes](https://github.com/sanity-io/sanity/blob/f136850d7f0cc07fe4299763210106806e31fccf/packages/%40sanity/import/src/importFromStream.js#L72)

PR #5409, opened 2023-12-20
feat(comments): implement reactions

hermanwikner:
### Description

This pull request introduces reactions to the comments feature. In the first iteration, there will be a fixed set of reactions to select from, however, in future iterations, we may introduce the complete palette of reaction options. The reactions are stored in the comment document in the 'reactions' array. Each individual reaction is an object with the following properties:

- `_key` - A unique key for the reaction. The key is constructed by combining the user ID and the short name of the reaction.
- `shortName` - The short name of the reaction. The short names follow [the naming convention listed here](https://projects.iamcal.com/emoji-data/table.htm).
- `userId` - The ID of the user who added the reaction.
- `addedAt` - The date when the reaction was added.

reactions

**Why is the `_optimisticState` needed?**

This flag is required to ensure that reactions removed optimistically but not yet processed on the server are not displayed. This scenario arises when a user removes **Reaction A** and, while the API request for that removal is still processing, removes **Reaction B**. There is a chance that **Reaction B** might temporarily reappear in the UI when the updated list of reactions is received. This is because the updated list, received after the removal of **Reaction A**, may not yet reflect the removal of **Reaction B**. Consequently, **Reaction B** might temporarily reappear in the UI until the updated list, including the removal of **Reaction B**, is received.

To address this issue, we utilize the `_optimisticState` flag to track the removal or addition of reactions locally. When the server responds with an updated list of reactions, we filter out reactions marked with the `_optimisticState` set to "removed." This prevents reactions that were removed optimistically but are still being processed from reappearing in the UI.

### What to review

- Make sure that the feature works as expected.
- Make sure that the data model is solid and future-proof.
- Make sure that the implementation is accessible.
- Test the resilience by intentionally removing and adding reactions frequently.

### Notes for release

Make it possible to react to comments


packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
```diff
@@ -129,7 +131,7 @@ export function CommentThreadLayout(props: CommentThreadLayoutProps) {
       </HeaderFlex>
 
       {canCreateNewThread && (
-        <ThreadCard onClick={handleNewThreadClick} tone={isSelected ? 'primary' : undefined}>
+        <ThreadCard onClick={handleNewThreadClick} tone={isSelected ? 'caution' : undefined}>
```
pedrobonamin, line 134: Just to confirm, is this change intentional?

packages/sanity/src/structure/comments/src/components/list/CommentsListItemContextMenu.tsx
```diff
@@ -9,9 +9,22 @@ import {
...
+import {ReactionIcon} from '../icons'
 import {ContextMenuButton} from 'sanity'
 
+const renderMenuButton = ({open}: {open: boolean}) => (
+  <Button
+    aria-label="Add reaction"
+    icon={ReactionIcon}
+    mode="bleed"
+    selected={open}
+    tooltipProps={{content: 'Add reaction'}}
```
pedrobonamin, line 24: I think this should be added as a translation

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -10,33 +10,25 @@ import {
...
-  }, [])
+const stopPropagation = (e: React.MouseEvent<HTMLDivElement>) => e.stopPropagation()
 
-  return (
-    <ContextMenuBox data-root-menu={isParent ? 'true' : 'false'} onClick={handleClick}>
-      {children}
-    </ContextMenuBox>
-  )
-}
+const ContextMenuBox = styled(Box)``
```
pedrobonamin, line 26: Do we need to continue defining this this way? Maybe just use box?

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -156,6 +150,20 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
   const hasChanges = useCommentHasChanged(value)
   const hasValue = useMemo(() => hasCommentMessageValue(value), [value])
 
+  // Filter out reactions that's been optimistically removed from the comment.
+  const reactions = useMemo(
+    () =>
+      (comment?.reactions?.filter((r) => r?._optimisticState !== 'removed') || EMPTY_ARRAY).filter(
+        (r) => {
+          // Filter out reactions that might have to incorrect format
+          return 'userId' in r && 'shortName' in r
```
pedrobonamin, line 159: Curious in how we could get to this situation and if we should be adding a safeguard like this.

packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
hermanwikner: Yes, it is intentional!

packages/sanity/src/structure/comments/src/components/list/CommentsListItemContextMenu.tsx
hermanwikner: Nothing is translated in the comments feature – we will treat that as a separate task and translate all text!

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
hermanwikner: It is unlikely that we that comments will be in the wrong format – but since the comment document can be mutated via e.g. the client, one might add a reaction item that is in the wrong shape.

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
hermanwikner: Yes, it is actually needed, because we target the `ContextMenuBox` element in CSS like this:

```js
${ContextMenuBox} {
  &:focus-within {
    opacity: 1;
  }
}
```

hermanwikner: We could refactor this to use an data attribute instead, like:

```js

[data-ui="ContextMenuBox"] {
  &:focus-within {
    opacity: 1;
  }
}

```

packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsUsersTooltip.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </Fragment>
+    ))
+
+    return (
+      <>
+        {others} <InlineText>and</InlineText>{' '}
+        <UserDisplayName currentUserId={currentUser.id} userId={last} />
+      </>
+    )
+  }, [currentUser, userIds])
```
pedrobonamin, line 112: I found it hard to read, what do you think about doing something like this?
```suggestion
  const content = useMemo(() => {
    const len = userIds.length

    if (len === 0 || !currentUser) return null

    return userIds.map((id, index) => (
      <Fragment key={id}>
        {/* If is last user, and there are more than one user, add 'and ' before the name. */}
        {index === len - 1 && len > 1 && (
          <>
            <InlineText>and </InlineText>{' '}
          </>
        )}
        <UserDisplayName
          currentUserId={currentUser.id}
          isFirst={index === 0}
          separator={index < userIds.length - 1}
          userId={id}
        />
      </Fragment>
    ))
  }, [currentUser, userIds])
```

Requires also to update the <UserDisplayName> to
```
function UserDisplayName(props: UserDisplayNameProps) {
  const {currentUserId, isFirst, userId, separator} = props
  const [user] = useUser(userId)

  const isCurrentUser = currentUserId === userId
  const you = isFirst ? 'You' : 'you'
  const content = isCurrentUser ? you : user?.displayName ?? 'Unknown user'
  const text = separator ? `${content}, ` : content

  return (
    <>
      <InlineText weight="medium">{text}</InlineText>
      {separator ? <> </> : null}
    </>
  )
}
`

File: packages/sanity/src/structure/comments/src/constants.ts [hunk 6]
```diff
@@ -0,0 +1,38 @@
...
+    shortName: ':heavy_plus_sign:',
+    title: 'Heavy plus sign',
+  },
+  {
+    shortName: ':-1:',
+    title: 'Thumbs down',
+  },
+]
+
+export const COMMENT_REACTION_EMOJIS: Record<CommentReactionShortNames, string> = {
```
pedrobonamin, line 31: any reason why this is not part of the `COMMENT_REACTION_OPTIONS` array? 
Have we considered exposing this into the plugin as configurable? Maybe it will be easier if it's all in one array.

packages/sanity/src/structure/comments/src/hooks/use-comment-operations/removeOperation.ts
```diff
@@ -0,0 +1,17 @@
...
+  id: string
+  onRemove?: (id: string) => void
+}
+
+export async function removeOperation(props: RemoveOperationProps): Promise<void> {
+  const {client, id, onRemove} = props
+  onRemove?.(id)
+
+  await Promise.all([
+    client.delete({query: `*[_type == "comment" && parentCommentId == "${id}"]`}),
```
pedrobonamin, line 14: ohh nice! I didn't knew this was a possiblity

pedrobonamin (approved):
This is awesome! Thanks @hermanwikner 
Left a suggestion and some questions, but none are blockers.

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
pedrobonamin: Missed that part! It's ok! Don't worry about updating it. Having the relation through the `ContentMenuBox` is less error prone, as removing it will show an error, but using the `data-ui` approach won't.
Thanks!

packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsUsersTooltip.tsx
pedrobonamin: 💯 thanks!

robinpyon (approved):
Fantastic work @hermanwikner!

PR #5411, opened 2023-12-20
feat(block-tools): add notion handling for single block decorators

sjelfull:
### Description

This creates a new ruleset for Notion, and handles block decorators for bold/emphasis/italic.

Normally Notion will export semantic HTML, but this behaviour changes when you have only copied a single block.

Then the exported HTML will contain spans with inline styles.

Fixes EDX-818.

### What to review

- That copy pasting a single block from Notion containing italic, emphasis or bold will be correctly formatted 

### Notes for release

- Fixed bold/emphasis/italic formatting in PTE when copying a single block from Notion


jtpetty (approved):
some of this seems pretty fragile, but I'm guessing there isn't really a better way to map styles of one platform to our format. 👍

PR #5412, opened 2023-12-20
fix(form): fix missing subtitles in array previews

sjelfull:
### Description

Fixes missing subtitle in object and reference previews inside arrays.

References:
CleanShot 2023-12-21 at 11 13 26

Objects:
CleanShot 2023-12-21 at 11 11 36

### Notes for release

- Fix missing subtitle in object and reference previews inside arrays


bjoerge (approved):
LGTM. The PR is still a draft though – is there more left to do here?

sjelfull:
> LGTM. The PR is still a draft though – is there more left to do here?

It should be ready, just wanted to verify that it fixes inline object previews - which it does.

PR #5413, opened 2023-12-21
fix(core): add missing deprecated tag to workspace logo

ninaandal:
### Description
The workspace logo was missing deprecated tag, as this is no longer supported


bjoerge (approved):
Thank you @ninaandal!

PR #5419, opened 2023-12-21
Optimize graphql deployment

sgulseth:
### Description

We're seeing an exponential slowdown when we are generating graphql schemas for sanity schemas that has many types referering to each other. This PR adds an, opt-in, cache that speeds up generating these schemas. Don't be alarmed by the diff, it's 95% snapshots 😅 

In the test schema case there's a 95% performance gain when generating with the cache:
```
withUnionCache: true:  88 ms
withUnionCache: false: 1732 ms
```

---

There were also some cases of looping over the same objects multiple times, I refactored the code into single loops.

### What to review

This change should, even with the flag on, generate the same schema. Is there a case where it might not? Any other test cases that might be beneficial?

### Notes for release

GraphQL: Add mode to optimize GraphQL deployment when schemas has many self referencing documents

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/graphql/deployGraphQLAPICommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/graphql/deployGraphQLAPICommand.ts
@@ -16,6 +16,7 @@ configuration file. Tread with caution!
   --non-null-document-fields Use non-null document fields (_id, _type etc)
   --playground Enable GraphQL playground for easier debugging
   --no-playground Disable GraphQL playground
+  --with-union-cache *Experimental:* Enable union cache that optimizes schema generation for schemas with many self referencing types
 
 Examples
   # Deploy all defined GraphQL APIs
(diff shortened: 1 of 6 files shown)


packages/sanity/src/_internal/cli/commands/graphql/deployGraphQLAPICommand.ts
```diff
@@ -16,6 +16,7 @@ configuration file. Tread with caution!
   --non-null-document-fields Use non-null document fields (_id, _type etc)
   --playground Enable GraphQL playground for easier debugging
   --no-playground Disable GraphQL playground
+  --with-union-cache *Experimental:* Enable union cache that optimnizes schema generation for schemas with many self referencing types
```
rexxars, line 19:
```suggestion
  --with-union-cache *Experimental:* Enable union cache that optimizes schema generation for schemas with many self referencing types
```

rexxars (changes requested):

rexxars (approved):

PR #5421, opened 2023-12-21
chore(e2e): update docs and make missing env more clear

binoy14:
### Description

Makes it more clear when env vars are missing to run e2e tests on what to do. Also updates docs, let me know if more clarity is needed. (Adding both EX and DX for awareness) 

### What to review

If I am missing anything

### Notes for release

N/A


bjoerge (approved):
Thank you!

PR #5422, opened 2023-12-22
fix(core): left align button text in PTE

ninaandal:
### Description
The `Normal` button in the PTE was center aligned instead of left aligned. 

**Before**: 
Screenshot 2023-12-22 at 11 04 55

**After**:
Screenshot 2023-12-22 at 11 04 24

### What to review
The buttons in the PTE toolbar


packages/sanity/src/core/form/inputs/PortableText/toolbar/BlockStyleSelect.tsx
```diff
@@ -136,6 +136,7 @@ export const BlockStyleSelect = memo(function BlockStyleSelect(
         mode="bleed"
         onClick={preventDefault}
         text={menuButtonText}
+        justify="flex-start"
```
sjelfull, line 139: I think we should make sure the caret is always right-aligned, so that it doesn't change position when the text on the left changes.

```suggestion
        justify="space-between"
```

sjelfull (changes requested):

packages/sanity/src/core/form/inputs/PortableText/toolbar/BlockStyleSelect.tsx
ninaandal: Thank you @sjelfull , this makes sense 🙇

sjelfull (approved):

PR #5424, opened 2023-12-22
fix(core): wait until the ref is available before showing trial popover

bjoerge:
### Description
This defers trial popover visibility until the anchor element is present in the DOM instead of using a 300ms delay, which turns out to not always be enough.

### What to review
- A bit tricky to test, but I've verified that it works locally

### Notes for release

- Fixes occasional wrong placement of free trial popover


jtpetty (approved):
👍

hermanwikner (approved):
Thank you!

Issue #5432, opened 2024-01-02
Vision tool crashes when selecting other api version in translated Studio. 


jorngeorg:
This is due to the change handler relying on the value of api version field being English – "other".  While in a translated Studio it can be any value. 

```
if (apiVersion?.toLowerCase() === 'other') {
...
}
```
https://github.com/sanity-io/sanity/blob/a63a20a2430345af74cb5cc060d6d20a70ffa91a/packages/%40sanity/vision/src/components/VisionGui.tsx#L394C39-L394C39

To reproduce:
1.  Open a studio and select any other language that English
2.  Open the Vision plugin
3.  Try selecting "other" api version
4.  The plugin crashes with the error `Error: Invalid API version string, expected 1 or date in format YYYY-MM-DD`
5.  The plugin will crash until you delete the localStorage value

PR #5433, opened 2024-01-02
fix(vision): use untranslated value for other api version

jorngeorg:
Fixes https://github.com/sanity-io/sanity/issues/5432

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.tsx
+++ b/packages/@sanity/vision/src/components/VisionGui.tsx
@@ -694,7 +694,7 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
                   {API_VERSIONS.map((version) => (
                     <option key={version}>{version}</option>
                   ))}
-                  <option key="other" value={t('settings.other-api-version-label')}>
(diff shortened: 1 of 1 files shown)


packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -694,7 +694,7 @@ export class VisionGui extends React.PureComponent<VisionGuiProps, VisionGuiStat
                   {API_VERSIONS.map((version) => (
                     <option key={version}>{version}</option>
                   ))}
-                  <option key="other" value={t('settings.other-api-version-label')}>
+                  <option key="other" value="other">
```
juice49, line 697: Thanks Jørn, this is great! Please can you also change the `value` prop of the `Select` component to use the constant value `'other'` on line `690`?

```tsx
<Select
  value={customApiVersion === false ? apiVersion : 'other'}
  onChange={this.handleChangeApiVersion}
>
```

Without this change, the API Version select shows the incorrect value when "Other" is chosen.

rexxars (approved):
Grazie!

PR #5438, opened 2024-01-02
feat(structure): deprecate `desk` in favor of `structure`

rexxars:
### Description

In [v3.20.0](https://github.com/sanity-io/sanity/releases/tag/v3.20.0), we "soft" renamed the "desk tool" to "structure tool" to make it clearer what its purpose what. We did not, however, do any changes to the _code_, which leaves a bit of a discrepancy and a potentially confusing surface (new users may be wondering why you are importing a `deskTool`, but ending up with a `Structure`).

This PR:
- Removes the "desk is now structure" popover
- Introduces a new `sanity/structure` export, which has a new `structureTool` function, along with all the other types and symbols previously exported under `sanity/desk`
  - Renamed: `DeskToolOptions` → `StructureToolOptions`
  - Renamed: `deskTool` → `structureTool`
  - Renamed: `useDeskTool` → `useStructureTool`
- Deprecates all exports from `sanity/desk` - they should now be imported from `sanity/structure` instead (will continue to work until the next major version, but suggesting that people migrate)
- Introduces a new codemod, `sanity codemod deskRename`, which will attempt to automatically rewrite any `sanity/desk` import to `sanity/structure` and change renamed symbols

Internal changes:
- _(Internal): Moves the file structure from `desk` directory to `structure`_
- _(Internal): Renames all(?) "Desk" references in symbols to "Structure"_
- _(Internal) Renamed: `DeskToolContextValue` → `StructureToolContextValue`_
- _(Internal) Renamed: `DeskToolFeatures` → `StructureToolFeatures`_
- _(Internal) Renamed: `DeskToolMenuItem` → `StructureToolMenuItem`_
- _(Internal) Renamed: `DeskToolPaneActionHandler` → `StructureToolPaneActionHandler`_
- _(Internal) Renamed: `DeskToolProvider` → `StructureToolProvider`_
- _(Internal) Renamed: `DeskToolProviderProps` → `StructureToolProviderProps`_

Breaking changes:
- None

### What to review

- Renames make sense
- Deprecation strategy works
- Codemod makes sense/works

### Notes for release

The `deskTool` export, along with any other API members from `sanity/desk` is now deprecated in favor of the `structureTool` from the `sanity/structure` path.

A codemod is available to migrate any code that may be using it - `npx @sanity/cli codemod deskRename`. Be sure to check in any local changes to version control _before_ running the codemod in case it should fail.

Note: If you have automated tests targeting the `desk-tool-list-pane` testid attribute, it needs to be changed to `structure-tool-list-pane`.


packages/sanity/src/structure/useStructureToolSetting.ts
```diff
@@ -4,23 +4,26 @@ import {useSettingsStore} from 'sanity'
...
   namespace: string | null,
   key: string,
   defaultValue?: ValueType,
 ): [ValueType | undefined, (_value: ValueType) => void] {
   const settingsStore = useSettingsStore()
   const [value, setValue] = useState<ValueType | undefined>(defaultValue)
 
-  const deskToolSettings = useMemo(() => settingsStore.forNamespace('desk-tool'), [settingsStore])
+  const structureToolSettings = useMemo(
+    () => settingsStore.forNamespace('structure-tool'),
```
juice49, line 16: I think you likely considered this (and it's probably an acceptable trade-off), but any desk tool settings currently stored in localStorage (such as sort orders) will be disregarded after this change.

packages/sanity/src/structure/panes/list/ListPane.tsx
```diff
@@ -22,7 +22,7 @@ export function ListPane(props: ListPaneProps) {
   return (
     <Pane
       currentMaxWidth={350}
-      data-testid="desk-tool-list-pane"
+      data-testid="structure-tool-list-pane"
```
juice49, line 25: It may be useful to explicitly mention this change in the release notes, in case any folks are relying on this test id for their own tests.

packages/sanity/src/structure/useStructureToolSetting.ts
rexxars: Yeah. It is a slightly "breaking" change in that settings for panes (ordering and display mode) will be lost, but I think it is acceptable given all you have to do is to set the ordering once more.

packages/sanity/src/structure/panes/list/ListPane.tsx
rexxars: Good point, will add it to the release notes

PR #5442, opened 2024-01-03
fix(theme): update studio theming, support only color and fonts customisation.

pedrobonamin:
### Description

The theme provided by `Themer` and `buildLegacyTheme` is too low level and allows for multiple customisation like margins, spaces, focus outlines etc. 
With these changes, those customisations possibilities are removed from the Studio and users will only be allowed to customise **fonts** and **colors**.

This is required because the **facelift** update introduced changes into areas like sizing, margins, spaces, etc and old themes created using themer or manually by users may not be updated with this, producing discrepancies in how the studio looks in sizes, fonts, spaces etcs.

Requires: https://github.com/sanity-io/themer/pull/311/files (merged ✅ )

### Example of broken navbar layout, theme created by themer.

Screenshot 2024-01-03 at 16 58 25
(Notice how the navbar is bigger)

### Expected navbar layout

Screenshot 2024-01-03 at 16 58 43

### After fixes
Screenshot 2024-01-03 at 17 00 30

### What to review

- Deprecations updates
- Legacy theme updates


packages/sanity/src/core/theme/index.ts
```diff
@@ -1,10 +1,7 @@
-import {buildTheme} from '@sanity/ui/theme'
-import {StudioTheme} from './types'
+import {RootTheme, buildTheme} from '@sanity/ui/theme'
 
 export * from './_legacy'
 export * from './types'
 
 /** @internal */
-export const defaultTheme: StudioTheme = buildTheme({
-  container: [320, 640, 960, 1280, 1600, 1920],
```
pedrobonamin, line 9: removed this "customisation" as it won't be possible to change it through the studioThemeProvider, anyways this is the same value as the default Sanity UI value.

packages/sanity/src/core/theme/themer/themerFonts.ts
```diff
@@ -0,0 +1,231 @@
...
+        fontSize: 22,
+        iconSize: 33,
+        lineHeight: 29,
+        letterSpacing: 0,
+      },
+    ],
+  },
+}
+
+export function usesLegacyThemerFonts(fonts?: StudioTheme['fonts']): boolean {
```
pedrobonamin, line 228: This will be updated to check for a `_themer: true` flag in the theme.

robinpyon (approved):

PR #5452, opened 2024-01-04
feat(desk): add __experimental_formPreviewTitle to documents

pedrobonamin:
### Description

With facelift a new title was introduced to all the documents, this title takes some space and doesn't work well with some custom implementations users have done to the document pane.

With this new flag `__experimental_form_preview_title ` users will be able to opt out from the new title.
__experimental_form_preview_title

### Notes for release

Allows users to hide the the large preview title is displayed in the document pane form

### DIFF
--- a/packages/@sanity/types/src/schema/definition/type/document.ts
+++ b/packages/@sanity/types/src/schema/definition/type/document.ts
@@ -29,8 +29,8 @@ export interface DocumentDefinition extends Omit<ObjectDefinition, 'type'> {
   /** @alpha */
   __experimental_omnisearch_visibility?: boolean
   /**
-   * Hides the document title introduced at the document pane
+   * Determines whether the large preview title is displayed in the document pane form
    * @alpha
    * */
-  __experimental_hide_form_title?: boolean
+  __experimental_form_preview_title?: boolean
 }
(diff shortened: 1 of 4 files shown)


packages/@sanity/types/src/schema/definition/type/document.ts
```diff
@@ -28,4 +28,9 @@ export interface DocumentDefinition extends Omit<ObjectDefinition, 'type'> {
   __experimental_search?: {path: string; weight: number; mapWith?: string}[]
   /** @alpha */
   __experimental_omnisearch_visibility?: boolean
+  /**
+   * Hides the document title introduced at the document pane
+   * @alpha
+   * */
+  __experimental_hide_form_title?: boolean
```
robinpyon, line 35:
```suggestion
  /**
   * Determines whether the large preview title is displayed in the document pane form
   * @alpha
   */
  __experimental_form_preview_title?: boolean
```

I would suggest renaming this to `__experimental_form_preview_title` to be a little more explicit – underscoring that this is a _preview_ title, which is shared across other previews too.

It also require users to set this to `false` to hide it, which I think reads a little better. WDYT?

pedrobonamin: Sounds good! Updated

robinpyon (approved):
Thanks @pedrobonamin!

binoy14 (approved):

PR #5458, opened 2024-01-05
fix(performance): add back `useMemo` on `DocumentPaneContext` value

stipsan:
### Description

Seems like `documentPane` [used to have a `useMemo`](https://github.com/sanity-io/sanity/commit/49a07c89ea2e1e56437c47be7cffae75ced5683e#diff-71d84315d36afd91007989c6e025ab7aded8fe1e2b336cd6baa2be74ace1f27a), but it was removed in #2939:
image

Looking at #2939 I can see a lot of rebase force pushes, and I'm making the assumption that the removal of `useMemo` wasn't intentional, it was likely an unfortunate side-effect of resolving many, very large, merge conflicts. @ricokahler can you confirm, or was the removal of `useMemo` intentional?

In any case, without the `useMemo` the `documentPane` variable, that `DocumentPaneProvider` passes to the `value` prop for `DocumentPaneContext.Provider`, gets a new identity every single time `DocumentPaneProvider` renders. It renders a lot as it listens to basically everything under the sun that is related to rendering a document pane.
For each new render, all components that have a `useDocumentPane` hook will also re-render, and even if they have their own memoization and techniques to detect if it should re-render or not, this work still has a cost. And these components are hot paths such as `FormView`, `DocumentPanel`, `DocumentLayout`, `CommentsInspector`, `DocumentInspectorPanel`, `ValidationInspector`, `TimelineMenu`, `ChangesInspector` and many many more.

### What to review

Now that `useMemo` is used we'll no longer have a `useDocumentPane` cascade of re-renders **every single time** that `DocumentPaneProvider` renders, so performance should be improved. However, since `DocumentPaneProvider` is packing so many, frequently changing, properties on `documentPane`, this bad cascade will still happen a lot, expect the improvement to be minor.
We need to verify that we don't have components that are relying on this broken behavior that can now break since the agressive rendering is reduced. Incorrect usage of `useRef`, and other similar flaws, can be "masked", or "fixed", by the old behaviour of `useDocumentPane`.
Affected components will feel "slugish" with the changes of this PR, that said I couldn't find any instances of that in my testing.
If such components are found we should fix said components by correcting their React hooks usage in this PR before merging. I doubt there are any, as theoretically affected components are already quite optimised.

### Notes for release

fix: improve memoization of the render context for document panes in the structure tool


packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
```diff
@@ -67,7 +67,6 @@ export interface DocumentPaneContextValue {
   setTimelineMode: (mode: TimelineMode) => void
   setTimelineRange(since: string | null, rev: string | null): void
   setIsDeleting: (state: boolean) => void
-  source?: string
```
stipsan, line 70: Removing it as it's never defined in the context value given to `DocumentPaneContext` and seems to have been this way for at least the last 2 years

ricokahler (approved):
Hey @stipsan! Apologies for taking a minute to review. Been going through my slack saved for later but finally had some time to run it through our perf studio. Here are the charts. The group on the right is this branch!

CleanShot 2024-01-19 at 16 00 07@2x

CleanShot 2024-01-19 at 16 00 35@2x

PR #5461, opened 2024-01-05
fix(cli): update esbuild options for better compat

ricokahler:
### Description

I ran into issues with the `mockBrowserEnvironment` function when re-using in a different branch.

Setting the format to `'cjs'` enforces that esbuild outputs code compatible with the commonjs module loader used to mock the browser environment on the fly and updating the extensions ensures that those files get ran through esbuild and converted into cjs compatible code.

### What to review

- Does the graphql command still work? Does this increase compatibility?

### Notes for release

Fixes an issue where certain files would not get loaded correctly in the mock browser environment used for GraphQL deploys.


binoy14 (approved):
don't have enough context on this but graphql command tests still work so I guess it's okay? I'd defer to @rexxars for final ✅

rexxars (approved):

PR #5468, opened 2024-01-08
fix(core): form fieldset border

hermanwikner:
### Description

This pull request addresses an issue with the left border of fieldsets.

Screenshot 2024-01-08 at 14 18 33

### What to review
- Is there any situation when we want border radius on the field set?

### Notes for release

N/A


sjelfull (approved):
LGTM. As far as I know there is no reason there should be a border-radius set there.

PR #5469, opened 2024-01-08
fix: mark builds as completed/errored and indicate empty cli runs

svirs:
### Description
Fix an error where builds weren't marked as completed or errored. Also explicitly marks a simple `sanity` cli command that is not run without any arguments/commands as an empty command

### What to review
CLI commands run without failure and downstream telemetry outputs are correct

### Notes for release


jtpetty (approved):
👍 - code looks fine, but would love to actually have a description in the PR

PR #5471, opened 2024-01-09
fix(cli): telemetry: only log first 10 elements in args array

bjoerge:
### Description
Caps the args list we log at max 10 entries

### What to review

I'm wondering if we could also remove the (duplicated) log statement, see my comment below.

### Notes for release
n/a

### DIFF
--- a/packages/@sanity/cli/src/cli.ts
+++ b/packages/@sanity/cli/src/cli.ts
@@ -130,7 +130,7 @@ export async function runCli(cliRoot: string, {cliVersion}: {cliVersion: string}
   const cliCommandTrace = telemetry.trace(CliCommand, {
     groupOrCommand: args.groupOrCommand,
     extraArguments: args.extraArguments,
-    commandArguments: args.argsWithoutOptions,
+    commandArguments: args.argsWithoutOptions.slice(0, 10),
     ...(!args.groupOrCommand && {emptyCommand: true}), // user did not entry a command
   })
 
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/cli.ts
```diff
@@ -130,14 +130,14 @@ export async function runCli(cliRoot: string, {cliVersion}: {cliVersion: string}
   const cliCommandTrace = telemetry.trace(CliCommand, {
     groupOrCommand: args.groupOrCommand,
     extraArguments: args.extraArguments,
-    commandArguments: args.argsWithoutOptions,
+    commandArguments: args.argsWithoutOptions.slice(0, 10),
   })
 
   cliCommandTrace.start()
   cliCommandTrace.log({
```
bjoerge, line 137: @svirs This is basically redundant now that we pass (most of) this as trace context. Would it be ok for us to remove explicitly logging this? The only thing we don't currently include in the context is `coreOptions`, but we can move it over so we have it there too.

svirs: That makes sense as long as we capture that in the context

binoy14 (approved):

packages/@sanity/cli/src/cli.ts
bjoerge: Cool - I've removed the redundant logging in 0825cac3b2063d30d0e94b0c253fdac99c608ef9

PR #5476, opened 2024-01-09
test(cli): fix flaky dev server test

juice49:
### Description

We've experienced some flakiness in the CLI dev server tests when they run in CI. This appears to be caused by port collisions that can occur when running the tests using multiple versions of node simultaneously.

The dev server now uses a different port depending on whether it's running in node LTS or node current.

### What to review

That the CLI tests pass in CI.


bjoerge (approved):
LGTM - thanks for digging into this!

PR #5478, opened 2024-01-10
fix: update reverse trial button styles, add tooltips and localize nav button

robinpyon:
### Description

This PR makes a few minor changes to the free trial buttons:
- These are more inline with facelift styles (using the same sized buttons / hit areas and tooltips)
- Free trial buttons now use tooltips (like all other text-less Studio buttons). These are also now localized.
- Fixes a small issue with buttons that weren't correctly spaced in the mobile nav drawer

image

### What to review

Free trial buttons should look like the above, and all tooltips should be localized as expected.

### Notes for release

N/A


jtpetty (approved):
👍

pedrobonamin (commented):
Good catch! 
Looks good to me!

PR #5479, opened 2024-01-10
fix: update search button navbar position

robinpyon:
### Description

This PR moves the navbar search button to the LHS (next to the create new button) in order to make it a little more discoverable. Mobile lockup is unchanged. That's it!

### What to review

That the search button appears on the LHS in the navbar.

### Notes for release

Updated search button navbar position to aid discoverability

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/StudioNavbar.tsx
@@ -181,6 +181,10 @@ export function StudioNavbar() {
                   {...newDocumentOptions}
                   modal={shouldRender.newDocumentFullscreen ? 'dialog' : 'popover'}
                 />
+                {/* Search button (desktop) */}
+                {!shouldRender.searchFullscreen && (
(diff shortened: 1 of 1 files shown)


pedrobonamin (approved):
Thanks @robinpyon I think this will make a difference for users.
Also, what do you think about doing a darker overlay? I noticed that in light mode we went too light with the overlay and it's difficult to get in a glance where you should be focusing on, it's happening in all modals in light mode.

Non blocking btw, this update should be done to the theme in sanity ui

## Pre facelift
Screenshot 2024-01-10 at 08 39 53

## Current facelift
Screenshot 2024-01-10 at 08 39 40

## Darker overlay in facelift 
Screenshot 2024-01-10 at 08 39 47

robinpyon:
@pedrobonamin 

> Also, what do you think about doing a darker overlay? I noticed that in light mode we went too light with the overlay and it's difficult to get in a glance where you should be focusing on, it's happening in all modals in light mode.

I think this is a great idea and we've already received signal here – the differences in light mode are _especially_ slight (though it's slightly more palatable in dark mode). As you've mentioned, it's a change we'll need to do in UI (updating `base.backdrop`).

PR #5489, opened 2024-01-11
feat(core): schema and field deprecation

binoy14:
### Description

This PR adds ability to deprecate documents and fields. 

#### Example to deprecate a document 
```tsx
defineType({
  name: 'deprecatedDocument',
  title: 'Deprecated Document',
  type: 'document',
  deprecated: {
    reason: 'This document type is deprecated, use the author document',
  },
})
```
Screenshot 2024-01-11 at 10 58 32 AM

#### Example to deprecate a field
```tsx
defineField({
  deprecated: {
    reason:
      'Use the non deprecated string field for new content, this field was originally for the old frontend',
  },
  description: 'This field is used to create the header for the site',
  validation: (Rule) => Rule.required(),
  name: 'string',
  title: 'string',
  type: 'string',
}),
```
Screenshot 2024-01-09 at 2 00 44 PM

The deprecation messages also gets carried over to the graphql schema

### What to review

[Test](https://test-studio-git-feat-type-deprecation.sanity.build/test/structure/input-debug;deprecatedFields) for deprecated fields 
[Test](https://test-studio-git-feat-type-deprecation.sanity.build/test/structure/input-debug;deprecatedDocument) for deprecated document

### Notes for release

- Adds support for deprecating fields and document types

### DIFF
--- a/packages/@sanity/types/src/schema/types.ts
+++ b/packages/@sanity/types/src/schema/types.ts
@@ -123,6 +123,13 @@ export interface DeprecatedProperty {
   reason: string
 }
 
+/**
+ * @public
+ */
+export interface DeprecationConfiguration {
+  deprecated: DeprecatedProperty
+}
+
 /** @public */
 export interface InitialValueResolverContext {
   projectId: string
@@ -177,12 +184,11 @@ export type SchemaValidationValue =
   | ((rule: Rule) => SchemaValidationValue)
 
 /** @public */
-export interface DeprecatedSchemaType {
-  deprecated: DeprecatedProperty
-}
+export type DeprecatedSchemaType<TSchemaType extends BaseSchemaType = BaseSchemaType> =
+  TSchemaType & DeprecationConfiguration
 
 /** @public */
-export interface BaseSchemaType extends Partial<DeprecatedSchemaType> {
+export interface BaseSchemaType extends Partial<DeprecationConfiguration> {
   name: string
   title?: string
   description?: string
--- a/packages/@sanity/types/src/schema/types.ts
+++ b/packages/@sanity/types/src/schema/types.ts
@@ -119,6 +119,11 @@ export type ConditionalPropertyCallback = (context: ConditionalPropertyCallbackC
 export type ConditionalProperty = boolean | ConditionalPropertyCallback | undefined
 
 /** @public */
+export interface DeprecatedProperty {
+  reason: string
+}
+
+/** @public */
 export interface InitialValueResolverContext {
   projectId: string
   dataset: string
--- a/packages/@sanity/types/src/schema/types.ts
+++ b/packages/@sanity/types/src/schema/types.ts
@@ -177,7 +177,12 @@ export type SchemaValidationValue =
   | ((rule: Rule) => SchemaValidationValue)
 
 /** @public */
-export interface BaseSchemaType {
+export interface DeprecatedSchemaType {
+  deprecated: DeprecatedProperty
+}
+
+/** @public */
+export interface BaseSchemaType extends Partial<DeprecatedSchemaType> {
   name: string
   title?: string
   description?: string
--- a/packages/sanity/src/_internal/cli/actions/graphql/gen3/generateTypeQueries.ts
+++ b/packages/sanity/src/_internal/cli/actions/graphql/gen3/generateTypeQueries.ts
@@ -4,9 +4,11 @@ import type {
   ApiCustomizationOptions,
   ConvertedType,
   ConvertedUnion,
+  Deprecation,
   InputObjectType,
   QueryDefinition,
 } from '../types'
+import {internal} from '../extractFromSanitySchema'
 import {getFilterFieldName} from './utils'
 
 export function generateTypeQueries(
@@ -52,6 +54,7 @@ export function generateTypeQueries(
           isNullable: false,
         },
       ],
+      ...getDeprecation(type),
     })
   })
 
@@ -103,8 +106,17 @@ export function generateTypeQueries(
           isFieldFilter: false,
         },
       ],
+      ...getDeprecation(type),
     })
   })
 
   return queries
+}
+
+function getDeprecation(type: ConvertedType): Partial<Deprecation> {
+  return type[internal]?.deprecationReason
+    ? {
+        deprecationReason: type[internal].deprecationReason,
+      }
+    : {}
 }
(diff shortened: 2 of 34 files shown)


binoy14:
@juice49 could you look at the unit tests and commit snapshots if it looks good

packages/sanity/src/_internal/cli/actions/graphql/gen3/generateTypeQueries.ts
```diff
@@ -103,8 +106,17 @@ export function generateTypeQueries(
...
 
   return queries
 }
+
+function getDeprecation(type: ConvertedType): Partial<Deprecation> {
+  return type[internal]?.deprecationReason
+    ? {
+        deprecationReason: type[internal].deprecationReason,
+      }
+    : {}
```
binoy14, line 121: ❓could this be just undefined instead of an empty object getting added to every type. Maybe in the above lines the spread could be conditional

juice49: We could, but we'd need to repeat that conditional every time `getDeprecation` is called. That's only twice in this file, but in `packages/sanity/src/_internal/cli/actions/graphql/extractFromSanitySchema.ts`, we have an analogous `getDeprecation` function that is called many more times.

I might be missing something, but what are the benefits would be of returning `undefined` and repeating the conditional logic?

binoy14: The only reason to repeat the conditional is because if you look at the snapshot it adds an empty object. 

```
Object {
      "fields": Array [
	@@ -790,6 +794,7 @@ Object {
      "name": "CdrPersonReference",
      "originalName": "cdrPersonReference",
      "type": "Object",
      Symbol(internal): Object {},
    },
```

which is not a falsy value, personally I would not anything if the field is not deprecated, empty objects might cause unknown issues

juice49: The `internal` key is intended as a generic container for any data we need to reference during the API generation process, but do not wish to include in then generated API definition. In this case, it's used when we generate the queries to send to Bowtie. It would be incorrect to send Bowtie the deprecation on the document type itself.

packages/sanity/src/core/form/inputs/BooleanInput.tsx
```diff
@@ -30,32 +29,35 @@ export function BooleanInput(props: BooleanInputProps) {
...
-          />
-        </Box>
-        <CenterAlignedBox paddingX={3} paddingY={1}>
-          <FormFieldStatus maxAvatars={1} position="top">
-            {/*<FieldPresence maxAvatars={1} presence={presence} />*/}
-          </FormFieldStatus>
-        </CenterAlignedBox>
-      </Flex>
-    </Card>
+    <>
```
pedrobonamin, line 32: Probably an oversight, do we need this empty fragment? It's adding extra differences which seems not to be needed.

binoy14: ah good catch this is from a previous iteration attempt that got leftover, thanks!

hermanwikner:
The deprecated badge causes some alignment issues:

Screenshot 2024-01-22 at 14 39 53

packages/sanity/src/core/form/studio/inputResolver/fieldResolver.tsx
```diff
@@ -60,6 +60,7 @@ function PrimitiveField(field: FieldProps) {
           level={field.level}
           title={field.title}
           validation={field.validation}
+          deprecated={field.schemaType.deprecated}
```
hermanwikner, line 63: Nitpick: alphabetize props 😇

packages/sanity/src/core/studio/components/navbar/new-document/NewDocumentButton.tsx
```diff
@@ -46,17 +48,23 @@ export function NewDocumentButton(props: NewDocumentButtonProps) {
...
   const openDialogAriaLabel = t('new-document.open-dialog-aria-label')
 
-  // Filter options based on search query
+  // Filter options based on search query and document type deprecations
   const filteredOptions = useMemo(
-    () => filterOptions(options, searchQuery, getI18nText),
-    [options, searchQuery, getI18nText],
+    () =>
+      filterOptions(
+        options.filter((option) => !isDeprecatedSchemaType(schema.get(option.schemaType))),
```
hermanwikner, line 63: The filtering of the deprecated schemas does not need to happen on every query change but could happen once like this:

```ts
const nonDeprecatedOptions = useMemo(() => {
  return options.filter((option) => !isDeprecatedSchemaType(schema.get(option.schemaType)))
}, [options, schema])

const filteredOptions = useMemo(
  () => filterOptions(nonDeprecatedOptions, searchQuery, getI18nText),
  [nonDeprecatedOptions, searchQuery, getI18nText],
)
  ```

Comment (juice49) [reply packages/sanity/src/core/studio/components/navbar/new-document/NewDocumentButton.tsx:63]: Great point! Thank you!

File: packages/@sanity/types/src/schema/types.ts [hunk 5]
```diff
@@ -172,7 +177,12 @@ export type SchemaValidationValue =
   | ((rule: Rule) => SchemaValidationValue)
 
 /** @public */
-export interface BaseSchemaType {
+export interface DeprecatedSchemaType {
```
ricokahler, line 180: Would it be better to make this type generic? I'm thinking something like:

```ts
// update the base schema type to include an optional `deprecated` prop
interface BaseSchemaType {
  // ... existing interface
  deprecated?: DeprecatedProperty
}

// update the DeprecatedSchemaType to be generic but include a default
type DeprecatedSchemaType<TSchemaType extends BaseSchemaType = BaseSchemaType> =
  TSchemaType & Required<Pick<TSchemaType, 'deprecated'>>

// update the narrowing function to be generic as well
function isDeprecatedSchemaType<TSchemaType extends BaseSchemaType>(type: TSchemaType): type is DeprecatedSchemaType<TSchemaType> {
  // ...
}
```

This way if you ever use the `DeprecatedSchemaType` type alone, it would include all the properties of BaseSchemaType and the resulting type after a narrowing would be `DeprecatedSchemaType<ObjectSchemaType>` instead of `DeprecatedSchemaType & ObjectSchemaType`

ricokahler (approved):
Looks great! I had one non-blocking comment in there to consider regarding types but I'd say go to go otherwise.

packages/@sanity/types/src/schema/types.ts
juice49: Thanks Rico, good point! I've adopted your feedback in [this commit](https://github.com/sanity-io/sanity/pull/5489/commits/cfb65c0b37208f662e3c1f7ad484708167dcd01b). I've added a `isDeprecationConfiguration` type guard, because in one instance we were checking whether an object that was not necessarily a schema contained this configuration.

PR #5490, opened 2024-01-11
fix(core): add missing inter fontface for weights 800+900

ninaandal:
### Description
Add missing font-face for weights 800 and 900 for the `Inter` font. These weights were used for the `Heading` and for the `Label` UI components and were not included after we stepped away from variable fonts. 

### What to review
In the PTE
- add text 
- change it to be a` Heading 1 `
- make this heading `bold`
- it should get a `weight 900`


pedrobonamin (approved):
Thanks @ninaandal

PR #5492, opened 2024-01-11
fix(test): use `"runtime": "automatic"` for `"@babel/preset-react"`

ricokahler:
### Description

This PR allows us to use JSX without importing react. Seems like the last thing not to support this was just our jest tests which uses babel to compile on the fly.

Later we should open a PR that removes all unused react imports as well but that can come at a later time.

### What to review

- Everything still work? Does this have any effect on the build system?

### Notes for release

N/A

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,7 +13,7 @@
       }
     ],
     "@babel/preset-typescript",
-    "@babel/preset-react"
+    ["@babel/preset-react", {"runtime": "automatic"}]
   ],
   "plugins": ["@babel/plugin-proposal-class-properties"]
 }


binoy14 (approved):
I think babel is only used for jest but I could be wrong

ricokahler:
I added @stipsan as a reviewer just in case it may effect the build downstream since I believe he maintains [pkg-utils](https://github.com/sanity-io/pkg-utils)

stipsan (approved):
`@sanity/pkg-utils` already supports automatic insertion of `react/jsx` imports by using esbuild.
Adding it in Babel won't break anything, worst case scenario the build when publishing the monorepo might take slightly longer than before, as babel is slower than esbuild.

PR #5500, opened 2024-01-12
fix(cli): always treat arguments without options as strings

jtpetty:
### Description

Fixes SDX-951

Fixes an issue where number-like string arguments were converted to numbers by the option parser, resulting in large document ids getting truncated. This forces all arguments without specific options to always be treated as strings.

### What to review

Would this have any unwanted side-effects for other commands?

### Notes for release

Fixes an issue where number-like string arguments were converted to numbers by the option parser


binoy14 (approved):
LGTM!

PR #5503, opened 2024-01-15
feat(core): add telemetry scaffolding to Studio

bjoerge:
### Description
Adds telemetry scaffolding to Studio, also logs an event when the publish button is clicked.

### What to review
Does it work as intended and look ok?

 Note: I'll make sure to fix the failing checks before merge.

### Notes for release
tbd


packages/sanity/src/core/studio/__telemetry__/studio.telemetry.ts
```diff
@@ -0,0 +1,12 @@
...
+interface StudioMountData {
+  studioVersion: string
+  plugins: {name: string}[]
+}
+
+export const StudioMount = defineEvent<StudioMountData>({
+  version: 1,
+  name: 'Studio Mounted',
+  description: 'The Studio was mounted on the page',
+})
```
joneidejohnsen, line 12: @bjoerge Would this trigger every time the Studio gets hard reloaded / opened in a new tab?

joneidejohnsen: @svirs I'm thinking that this would be similar to the identify event we have in the CLI? And since we already have the "Studio Entered" event set up server-side, this might be overlapping if it gets called every time (ref my comment above).

packages/sanity/src/structure/documentActions/PublishAction.tsx
```diff
@@ -122,7 +125,10 @@ export const PublishAction: DocumentActionComponent = (props) => {
     return () => clearTimeout(timer)
   }, [changesOpen, publishState, hasDraft, onHistoryOpen])
 
+  const telemetry = useTelemetry()
+
   const handle = useCallback(() => {
+    telemetry.log(PublishDocument)
```
joneidejohnsen, line 131: @bjoerge Event should be called "Document Published" to follow convention. Could we also add an event property called "existing_document" <boolean> to see if it's a brand new document or an update to an existing one?

joneidejohnsen (commented):
Added some clarifying quesitons

packages/sanity/src/core/studio/StudioTelemetryProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+          uri: '/intake/batch',
+          method: 'POST',
+          json: true,
+          body: {projectId, batch},
+        }),
+      // opts into a different strategy for sending events when the browser close, reload or navigate away from the current page
+      sendBeacon: (events) =>
+        navigator.sendBeacon(
+          client.getUrl('/intake/batch'),
+          new Blob([JSON.stringify(events)], {type: 'text/plain'}),
```
svirs, line 38: If `sendBeacon` is locked to text/plain, we'll need to handle this new content-type

svirs: Service was updated

packages/sanity/src/core/studio/__telemetry__/studio.telemetry.ts
bjoerge: > Would this trigger every time the Studio gets hard reloaded / opened in a new tab?

Yup!

packages/sanity/src/core/studio/StudioTelemetryProvider.tsx
bjoerge: It's not strictly limited, but using anything else will trigger a CORS preflight, which we probably want to avoid spending time on in this case.

packages/sanity/src/core/studio/__telemetry__/studio.telemetry.ts
joneidejohnsen: @svirs Need to figure out how to deal with this in addition to the current "Studio Entered" event. Maybe combine them if possible?

bjoerge: I can change this to be

```ts
updateUserProperties({studioVersion: …, plugins: […]})
```

packages/sanity/src/structure/documentActions/PublishAction.tsx
bjoerge: @joneidejohnsen for a little bit of increased clarity, could it instead of `existing_document` be `new: true|false`?

bjoerge: Actually, now that I think of it, should it mean:
- This document has not been published before
- Or "this is a new document" - e.g. created by the user just now and published straight away

bjoerge: we can actually determine both, so we culd do this:
- `first_publish: boolean` - published for the first time (note that this will be true in case the document has previously been *un*published or deleted)
- `new: boolean` - created and published straight away

bjoerge: Implemented it this way in https://github.com/sanity-io/sanity/commit/52ef1d222ab7bf5f5d1c6ca47b09620efb4d0897

joneidejohnsen (approved):

PR #5504, opened 2024-01-16
fix: various navbar improvements

robinpyon:
### Description

This PR introduces the following changes to the navbar and the top-left workspace menu lockup.
- The workspace icon will now always act as a discrete link to the 'root' page (the first configured tool). 
  - The active workspace title will continue to remain as a workspace switcher. In the event only one workspace is enabled, this too will act as a link back 'home'.
- The workspace switcher now correctly displays a tooltip
- The tool menu will no longer be displayed if you only have one tool installed
- Fixes an issue where long workspace titles could break navbar layout – these now truncate to ellipses
- Localizes the 'open menu' button tooltip text which is vislble on smaller breakpoints

image

### What to review

That all of the above looks good and makes sense – especially the changes to the workspace menu! 

The discrete home page button should act as an anchor link and clearly display focus as needed.

That the navbar changes as expected based on whether the current studio environment:
- contains one or multiple workspaces
- contains one or multiple tools

### Notes for release

- The workspace icon will now always act as a discrete link to the 'root' page (the first configured tool)
- The workspace switcher now correctly displays a tooltip
- The tool menu will no longer be displayed if you only have one tool installed
- Fixes an issue where long workspace titles could break navbar layout – these now truncate to ellipses
- Localizes the 'open menu' button tooltip text which is vislble on smaller breakpoints


pedrobonamin (approved):
This is a great improvement! 
Tested in different scenarios and it all looks good to me.

juice49:
I noticed the workspace logo is now misaligned vertically if the user increases their browser font size.

Screenshot 2024-01-16 at 09 32 57

Here, I tested in Chrome.

**Current Studio version**

Screenshot 2024-01-16 at 09 33 23

**With changes**

Screenshot 2024-01-16 at 09 33 37

Screenshot 2024-01-16 at 09 33 27

robinpyon:
Great find @juice49 – that's been fixed! Example of XL below:

image

I definitely tend to index on testing _page zoom_ rather than only overriding only font sizes, something to keep in mind for future!

juice49 (approved):
Looks great!

PR #5507, opened 2024-01-16
fix(deps): update @sanity/ui to 2.0.0 and @sanity/color to 3.0.0

robinpyon:
### Notes for release

- Fixes an issue where the `selectable` object would not be available in the studio theme object
- Increases backdrop opacity when both dialogs and global search are open (in light mode)


pedrobonamin (approved):
🚀 🚀 🚀

PR #5513, opened 2024-01-17
chore: rename "schemas" to the more correct plural "schemaTypes"

SimeonGriggs:
### Description

A Studio _can_ have many "schemas", in that it can have multiple workspaces.

But a collection of schema types – for example a `post`, `author`, `category` is not best described by the word "schemas" but that _is_ the name for files and folder structures in new Sanity Studio projects.

This PR has renamed most instances of "schemas" to "schemaTypes" / "schema types" where it is most appropriate.

### What to review

That new Studio installs of the various templates still work! Admittedly, I don't fully know how to test the **functional** impact of these proposed changes. But the impact of them will be more correctness in our training materials.

### Notes for release

Corrected the default folder name for schema types


ricokahler (approved):
I can't think of any reasons why this would break anything and I agree with the semantics. Love it!

SimeonGriggs:
I've tested creating new Studios locally with these changes and they boot up fine.

PR #5520, opened 2024-01-18
fix(i18n): missing translation on close button

jtpetty:
### Description

Fixes SDX-908

This change fixes a missing translation for the close button for a few panels.

### DIFF
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -124,9 +124,6 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The action menu button tooltip */
   'buttons.action-menu-button.tooltip': 'Document actions',
 
-  /** The close button tooltip */
-  'buttons.close.tooltip': 'Close',
-
   /** The aria-label for the split pane button on the document panel header */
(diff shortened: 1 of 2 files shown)


binoy14 (approved):

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -124,6 +124,9 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The action menu button tooltip */
   'buttons.action-menu-button.tooltip': 'Document actions',
 
+  /** The close button tooltip */
+  'buttons.close.tooltip': 'Close',
```
rexxars, line 128: I'd colocate this with the rest of the `document-inspector` resources, so maybe:
```suggestion
  /** Tooltip text shown for the close button of the document inspector */
  'document-inspector.close-button.tooltip': 'Close',
```

If so, this also needs to be sorted into a different area further down.

rexxars (changes requested):

packages/sanity/src/structure/i18n/resources.ts
jtpetty: Fair. I thought this was going to be more generic and needed in more places but those other instances were in `__workshop__` directories

PR #5521, opened 2024-01-18
fix(cli): misc sanity documents validate fixes

ricokahler:
### Description

Includes various fixes for the CLI validate command including:
- fixes typos in the help command
- fixes a bug where the document type would not be visible due to a gray background
- no longer shows validation for `system.` and `sanity.` prefixed documents
- fixes an issue where local tests runs would fail due to colored chalk output
- fixes a missing document revision in the NDJSON and JSON reporters
- fixes a bug where the batches for the reference checks were not separated correctly
- also adds a test for the `validateDocuments` command

### What to review

Take a look at the changed code and see if I missed anything. Also review the copy because of typos!

### Testing

Tests were added via the `validateDocuments.test.ts` file.

### Notes for release

- Fixes typos with `sanity documents validate --help`
- Fixes a bug where the document type in the `sanity documents validate` command would not be visible
- Fixes a bug where system documents would appear in validation reports

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/documents/validateDocumentsCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/documents/validateDocumentsCommand.ts
@@ -19,10 +19,10 @@ Options
   --dataset <name> Override the dataset used. By default, this is derived from the given workspace
   --format <pretty|ndjson|json> The output format used to print the found validation markers and report progress
   --level <error|warning|info> The minimum level reported out. Defaults to warning.
-  --max-custom-validation-concurrent <number> Specify how many custom validators can run concurrently. Defaults to 5.
+  --max-custom-validation-concurrency <number> Specify how many custom validators can run concurrently. Defaults to 5.
 
 Examples
-  # Validates all documents in a sanity project with one workspace
+  # Validates all documents in a sanity project with more than one workspace
   sanity documents validate --workspace default
 
   # Override the dataset specified in the workspace
@@ -32,7 +32,7 @@ Examples
   sanity documents validate > report.txt
 
   # Report out info level validation markers too
-  sanity document validate --level info
+  sanity documents validate --level info
 `
 
 const validateDocumentsCommand: CliCommandDefinition = {
--- a/packages/sanity/src/_internal/cli/threads/validateDocuments.ts
+++ b/packages/sanity/src/_internal/cli/threads/validateDocuments.ts
@@ -265,7 +265,7 @@ async function checkReferenceExistence({
     },
     Array.from<string[]>({
       length: Math.ceil(idsToCheck.length / REFERENCE_INTEGRITY_BATCH_SIZE),
-    }).fill([]),
+    }).map(() => []),
   )
 
   for (const batch of batches) {
--- a/packages/sanity/src/_internal/cli/threads/validateDocuments.ts
+++ b/packages/sanity/src/_internal/cli/threads/validateDocuments.ts
@@ -146,14 +146,6 @@ async function loadWorkspace() {
     ...clientConfig,
     dataset: dataset || workspace.dataset,
     projectId: projectId || workspace.projectId,
-    // we set this explictly to true because the default client configuration
-    // from the CLI comes configured with `useProjectHostname: false` when
-    // `requireProject` is set to false
-    useProjectHostname: true,
-    // we set this explictly to true because we pass in a token via the
-    // `clientConfiguration` object and also mock a browser environment in
-    // this worker which triggers the browser warning
-    ignoreBrowserTokenWarning: true,
     requestTagPrefix: 'sanity.cli.validate',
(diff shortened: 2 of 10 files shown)


packages/sanity/src/_internal/cli/commands/documents/validateDocumentsCommand.ts
```diff
@@ -19,10 +19,10 @@ Options
   --dataset <name> Override the dataset used. By default, this is derived from the given workspace
   --format <pretty|ndjson|json> The output format used to print the found validation markers and report progress
   --level <error|warning|info> The minimum level reported out. Defaults to warning.
-  --max-custom-validation-concurrent <number> Specify how many custom validators can run concurrently. Defaults to 5.
+  --max-custom-validation-concurrency <number> Specify how many custom validators can run concurrently. Defaults to 5.
```
juice49, line 22: Minor, but for consistency, we should add a period to the end of the descriptions for the other options (`y`, `workspace`, etc).

packages/sanity/src/_internal/cli/commands/documents/validateDocumentsCommand.ts
```diff
@@ -19,10 +19,10 @@ Options
...
 
 Examples
-  # Validates all documents in a sanity project with one workspace
+  # Validates all documents in a sanity project with more than one workspace
```
juice49, line 25: Should be "Sanity" (uppercase "S").

packages/sanity/src/_internal/cli/threads/validateDocuments.ts
```diff
@@ -24,6 +24,7 @@ import {isRecord, validateDocument} from 'sanity'
 const MAX_VALIDATION_CONCURRENCY = 100
 const DOCUMENT_VALIDATION_TIMEOUT = 30000
 const REFERENCE_INTEGRITY_BATCH_SIZE = 100
+const EXCLUDED_DOCUMENT_TYPE_PREFIXES = ['system.', 'sanity.']
```
juice49, line 27: I wonder whether there might ever be a reason to validate assets:

1. They can be modified using the API, so can end up in an invalid state. CLDX-1377 covers a scenario like this. Although there are no existing validation rules that would have caught this problem, it's a real issue that affected users.
2. In the future, it may be possible for users to extend these schemas, likely making bulk validating them more useful.

packages/sanity/src/_internal/cli/commands/documents/validateDocumentsCommand.ts
ricokahler: your knut is showing lol. great catch!

packages/sanity/src/_internal/cli/threads/validateDocuments.ts
ricokahler: > They can be modified using the API, so can end up in an invalid state. CLDX-1377 covers a scenario like this. Although there are no existing validation rules that would have caught this problem, it's a real issue that affected users.

The decision to exclude `sanity.` documents came from @runeb but I do see your point. Any built-in `sanity.` type can happy be validated if we have a schema definition. The built-in ones are [here](https://github.com/sanity-io/sanity/blob/9b3eae9781b1a33e0e7a9845e107d2c73eaf5437/packages/sanity/src/core/schema/createSchema.ts#L19-L32).

@runeb what do you think? should we allow validating the built-in types too?

> In the future, it may be possible for users to extend these schemas, likely making bulk validating them more useful.

If a user does extend any of the built-in schemas, that should get a new document `_type` and therefore would then get included in the validation results. This line just excludes `_type`s that start with `sanity.*` or `system.*`

ricokahler: Just followed up with @runeb and we decided to keep validating sanity built-in types so I removed the `sanity.*` prefix but there's still some things to discuss. I tagged you in slack.

juice49: Sounds good!

> If a user does extend any of the built-in schemas, that should get a new document _type and therefore would then get included in the validation results.

Sorry, I probably phrased that badly. By _extend_, I meant adding fields to the system type, e.g. adding a `license` field directly to `sanity.imageAsset`.

packages/sanity/test/validation/validateDocument.test.ts
```diff
@@ -14,8 +15,8 @@ import {
 } from '../../src/core/validation/validateDocument'
 import {getFallbackLocaleSource} from '../../src/core/i18n/fallback'
 import {convertToValidationMarker} from '../../src/core/validation/util/convertToValidationMarker'
-import {Workspace} from '../../src/core'
-import {createSchema} from './helpers/createSchema'
+import {Workspace, createSchema} from '../../src/core'
+// import {createSchema} from './helpers/createSchema'
```
juice49, line 19: Should this line be deleted?

juice49 (approved):

PR #5524, opened 2024-01-19
fix: force emoji-specific fonts, prefer buttons over cards, minor styling updates

robinpyon:
(See inline comments for more context)

### DIFF
--- a/packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsBar.tsx
+++ b/packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsBar.tsx
@@ -1,13 +1,20 @@
 import React, {useCallback, useMemo, useRef} from 'react'
 import {CurrentUser} from '@sanity/types'
-import {Card, Flex, Text, TooltipDelayGroupProvider} from '@sanity/ui'
-import styled from 'styled-components'
+import {
+  // eslint-disable-next-line no-restricted-imports
+  Button as UIButton,
+  Flex,
+  Text,
+  TooltipDelayGroupProvider,
+} from '@sanity/ui'
 import {CommentReactionItem, CommentReactionOption, CommentReactionShortNames} from '../../types'
 import {COMMENT_REACTION_EMOJIS, COMMENT_REACTION_OPTIONS} from '../../constants'
 import {ReactionIcon} from '../icons'
 import {TOOLTIP_DELAY_PROPS, Tooltip} from '../../../../../ui-components'
 import {CommentReactionsMenuButton} from './CommentReactionsMenuButton'
 import {CommentReactionsUsersTooltip} from './CommentReactionsUsersTooltip'
+import {EmojiText} from './EmojiText.styled'
+import {TransparentCard} from './TransparentCard.styled'
 
 /**
  * A function that groups reactions by name. For example:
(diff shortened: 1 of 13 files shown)


packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsBar.tsx
```diff
@@ -48,29 +55,19 @@ function groupReactionsByName(reactions: CommentReactionItem[]) {
...
-    type="button"
-  >
-    <Tooltip animate content="Add reaction">
-      <Flex align="center" justify="center" padding={2} paddingX={3}>
-        <Text muted size={1}>
-          <ReactionIcon />
-        </Text>
+const renderMenuButton = ({open}: {open: boolean}) => {
+  return (
+    <UIButton fontSize={1} mode="ghost" padding={0} radius="full" selected={open}>
```
robinpyon, line 60: Preferring usage of regular `<Button>` over `<Card>` components here. It does feel a little counter-intuitive having the `<Tooltip>` nested within the button itself, but required as it the `<MenuButton>` won't correctly return focus on close (you already accounted for this!)

packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsBar.tsx
```diff
@@ -150,38 +147,39 @@ export const CommentReactionsBar = React.memo(function CommentReactionsBar(
...
-                </Flex>
-              </ReactionButtonCard>
+              <TransparentCard tone="default">
+                <UIButton
+                  disabled={readOnly}
+                  mode="ghost"
+                  // eslint-disable-next-line react/jsx-no-bind
+                  onClick={() => handleSelect(name)}
+                  padding={2}
+                  radius="full"
```
robinpyon, line 157: `radius="full"` is a recent addition to Sanity UI, we don't really use it that often, but this is a good case for it

packages/sanity/src/structure/comments/src/components/reactions/EmojiText.styled.ts
robinpyon, line 1: Separate styled component specifically for rendering emoji text only. Specifying an emoji font stack is required to render certain emojis (e.g. :heart:) as some fonts may include alternates ([reference](https://nolanlawson.com/2022/04/08/the-struggle-of-using-native-emoji-on-the-web/))

This may not necessarily need to be a separate component in future

packages/sanity/src/structure/comments/src/components/reactions/TransparentCard.styled.ts
robinpyon, line 1: Not super proud of this, it's a bit of a hacky way to allow `<Button>` components to not color / tone mix with any parent `<Card>` component.

In future, we could potentially look to add an unstable prop to Sanity UI primitives to bypass this (If you want to render buttons that will always appear the same tone, independent of any parent context). Perhaps there's already a more idiomatic way to do this?

packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsMenu.tsx
```diff
@@ -83,9 +82,12 @@ export function CommentReactionsMenu(props: CommentReactionsMenuProps) {
             onClick={handleOptionClick}
             padding={2}
             role="menuitem"
-            selected={selected}
-            text={emoji}
-          />
+            tabIndex={-1}
```
robinpyon, line 85: Added to prevent slight focus flickering when pressing TAB when you have an existing emoji focused in this menu

hermanwikner (approved):
Thank you for nailing the details @robinpyon!

PR #5529, opened 2024-01-19
fix(cli): compile esm files in node_modules

ricokahler:
### Description

This PR compiles esm files within `node_modules` for better compatibility. This fixes a bug in the CLI validate command that causes workspaces not to load correctly. In an ideal world with node's conditional exports, this shouldn't be needed. However, we can't guarantee that library authors use conditional exports and also support CJS so this should help bridge the gap in many cases.

This uses `es-module-lexer` to quickly determine if a file is ESM or not and if so, tells esbuild to compile it. `es-module-lexer` is a performant library written in C and compiled to WASM.

### What to review

Take a look at the code and see if everything lines up or if I missed anything. In particular, I had to change the signature of the `mockBrowserEnv` to return a promise. I updated the usage areas however there may be affects I'm unaware of.

### Testing

- I tried this in hyperdrive via the `sanity document validate` command and it correctly loaded the project when it didn't before
- I tried this in our own test-studio and it worked as expected and did not increase the load time significantly
- I tried running a script via `sanity exec` and everything still worked as expected (`mockBrowserEnvironment` is used in `registerBrowserEnv`).

### Notes for release

- Improves compatibility of mock browser environment used in the CLI for GraphQL deploys, document validation, and script execution

### DIFF
--- a/packages/sanity/package.json
+++ b/packages/sanity/package.json
@@ -222,6 +222,7 @@
     "dataloader": "^2.1.0",
     "date-fns": "^2.26.1",
     "debug": "^3.2.7",
+    "es-module-lexer": "^1.4.1",
     "esbuild": "^0.19.8",
     "esbuild-register": "^3.4.1",
     "execa": "^2.0.0",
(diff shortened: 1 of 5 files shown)


bjoerge:
@ricokahler is this still relevant?

ricokahler:
@bjoerge No i think we thought this solve was too complex so i'll close.

PR #5547, opened 2024-01-23
fix: only fully collapse PTE toolbars in 'root' FormBuilder instances

robinpyon:
### Description

This PR updates PTE Toolbars to conditionally collapse at smaller element widths _only_ when displayed in a `<FormBuilder>` instance with `id="root"` (the default ID assigned within the document pane).

**Current behaviour:**

https://github.com/sanity-io/sanity/assets/209129/681ae41e-56a3-45a9-a956-a54c558098b2

**The problem:**

https://github.com/sanity-io/sanity/assets/209129/3929ef4b-5ff8-43dc-8e77-a9fc29e85271

**Approach:**

- We only apply this behaviour when the form builder context within the current PTE component has `id === "root"` (the default)
- This felt like the simplest option that avoids adding extra layout-specific props to `<FormBuilder>`. 
- It felt reasonable to add an extra `FormBuilderContext` dependency in `Editor.tsx`, given that PTE objects are also dependent on this.
- This will be removed when we eventually revisit the UI for PTE fixed toolbars

**Questions:**

- Could this be done in a better way?
- Are there any side effects of giving a `<FormBuilder>` a non "root" ID? (None that I can tell)

### What to review

- Default toolbar behaviour in the studio is unchanged
- That `playwright-ct` tests pass
- Specifying a non "root" `<FormBuilder>` id disables toolbars from automatically collapsing.

### Testing

`playwright-ct` tests have been aded to validate behaviour in `<FormBuilder>` instances with root and non-root ids.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Toolbar.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Toolbar.spec.tsx
@@ -35,5 +35,62 @@ test.describe('Portable Text Input', () => {
         ).toContainText('Inline Object')
       })
     })
+
+    test.describe('Collapsible toolbar', () => {
+      test.describe('Root <FormBuilder>', () => {
+        test('Toolbar should collapse when element width is less than 400px', async ({
+          mount,
+          page,
+        }) => {
+          const {getFocusedPortableTextInput} = testHelpers({page})
+          await mount(<ToolbarStory id="root" />)
+          const $portableTextInput = await getFocusedPortableTextInput('field-body')
+
+          // Adjust viewport size to enable auto collapsing toolbar menus
+          await page.setViewportSize({width: 450, height: 500})
+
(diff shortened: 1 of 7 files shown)


ninaandal:
Do we have a test studio with custom PTE inputs now so that I can easily test this? 🤞

robinpyon:
@ninaandal I think the simplest way to validate this (outside of checking out the playwright component tests) would be to manually update the `id` defined for the document panel `<FormBuilder>` [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx#L185).

This is a targeted way to ensure that the AI Assist plugin (which does bring its own `<FormBuilder>`) can render a slightly more ergonomic PTE toolbar when it's rendered in document inspector panels, and as mentioned above, can probably be removed if and when PTE toolbars are overhauled entirely!

ninaandal:
lgtm - all tests passed and toolbar behaviour is as expected ✨

ninaandal (approved):

PR #5549, opened 2024-01-23
chore: add ui component workshop stories, simplify internal tab component

robinpyon:
### Description

This PR adds some basic workshop stories outlining usage guidelines for internal components, and is intended to be expanded on in future.

This also simplifies the internal `<Tab>` component and removes the `size` prop, since these are only ever rendered in a single size throughout.

### What to review

Workshop stories are accessible and make sense.

### Testing

These stories act as a simple reference and facilitate spot testing.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/ui-components/__workshop__/TooltipStory.tsx
+++ b/packages/sanity/src/ui-components/__workshop__/TooltipStory.tsx
@@ -1,0 +1,78 @@
+import {Box, Card, Container, Stack, Text} from '@sanity/ui'
+import {useString} from '@sanity/ui-workshop'
+import {Tooltip} from '../tooltip'
+import {Button} from '../button'
+
+export default function TooltipStory() {
+  const text = useString('Tooltip content', 'Tooltip content', 'Props') || ''
+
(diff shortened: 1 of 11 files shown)


packages/sanity/src/ui-components/__workshop__/TooltipStory.tsx
```diff
@@ -0,0 +1,71 @@
...
+              Recommended
+            </Text>
+          </Box>
+          <Stack padding={3} space={4}>
+            <Stack space={3}>
+              <Text muted size={1}>
+                String content
+              </Text>
+              <Tooltip content={text}>
+                <Text>Default tooltip with string content</Text>
```
pedrobonamin, line 31: nit: Could we adapt this and the other text description to something more generic like, "hover me", or add an icon / button to easy identify where we should hover?

Or change it text to be specific to the tooltip it's representing.

It's a bit confusing to see in all the examples: `Default tooltip with string content`

pedrobonamin (approved):
Looks great,

PR #5551, opened 2024-01-23
bug(core): fix issue with date formatting in validation message

jtpetty:
### Description

Fixes SDX-401

Discovered that the option format we pass from the form config to the library used by the date validator has different expectations for date formats.

Before:
image

After:
image

### What to review

Date formats that come from user config are `moment` formatted, not ISO formatted, thus we need to use `legacyDateFormat` here.

### Testing

Added unit tests for date validation


binoy14 (commented):
I think the issue is the difference in the formatting of date in the validator vs the dateinput. The input is using the `@sanity/util` package to format and parse dates https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx#L1 I think if we change the validator to use the same than the options should be consistent. The main different is input is using moment, validator is using date-fns and the way they consume tokens is different so we are seeing the issue. Maybe someone on the team has some more context on why is this the case

jtpetty:
I have a better fix coming

packages/sanity/test/validation/dates.test.ts
```diff
@@ -0,0 +1,102 @@
+import {getFallbackLocaleSource} from '../../src/core/i18n/fallback'
+import {Rule} from '../../src/core/validation'
+
+describe('date', () => {
+  describe('with default format', () => {
+    const context: any = {client: {}, i18n: getFallbackLocaleSource(), type: {name: 'date'}}
+
+    test('min length constraint', async () => {
+      const rule = Rule.dateTime().min(Date.parse('2024-01-01'))
```
binoy14, line 9: once this is rebased with changes from `next` all the `Date.parse` can be removed and it should still work

ricokahler (approved):
nice! I had one non-blocking suggestion to consider

binoy14 (approved):
✅  the PR description is missing notes for release

PR #5555, opened 2024-01-24
fix(validation): update min and max to include string for dates

ricokahler:
### Description

Related to #5551.

This fixes the types for `rule.min` and `rule.max` to include strings because the date validator allows them. This was just an oversight on types when this part of the codebase was converted to TS (by me lol)

### What to review

Can you use `rule.min()` and `rule.max()` with a string now?

### Testing

Re-ran `npm run type-check` and tried out using a string in rule.min and it worked as expected

### Notes for release

- Fixes an type bug where `Rule.min` and `Rule.max` did not allow strings


jtpetty (approved):
👍 thank you

PR #5562, opened 2024-01-25
refactor: move babel config to jest config

stipsan:
### Description

Babel, and the related root `.babelrc` and `babel.config.cjs` files, are only used by Jest at the time being. By shifting this configuration into the Jest config itself it makes it easier to understand the impact of changing the Babel configuration. It also makes it obvious when these babel presets and plugins may be deleted, should the Jest setup no longer depend on it.

### Testing

Jest tests should work exactly like before.

### Notes for release

N/A


binoy14 (approved):
🙇 thank you for doing this!

PR #5563, opened 2024-01-25
Improve asset export retrying 

sgulseth:
### Description

We are experiencing that our export assets endpoints hangs up the connections, and for some reason the CLI is trying to reuse the dead connection. This PR adds more retry mechanisms to download assets.

### Testing

```
$ npm run build
$ cd dev/test-studio
$ DEBUG=sanity* ../../packages/@sanity/cli/bin/sanity dataset export test --overwrite
```


j33ty (approved):
Looks good to me. I have tried this change locally, this seem to be working as expected, as seen in debug mode. This approach will make CLI export resilient against flaky server-side and network errors.

binoy14 (approved):
LGTM!

PR #5570, opened 2024-01-26
feat(comments): adds read-only state to comments.

pedrobonamin:
### Description

- Creates a new context into comments plugin to resolve upsell necessary data and behaviour. 
- Adds new export `DescriptionSerializer` from `sanity` package for reusability in comments plugin.

After free trial, it adds read only state to comments, showing the comments in a disabled state and only allowing to:
- Resolve existing comments.
- Delete existing comments

It removes the ability to:
- Add new comments.
- Reply to existing comments.
- React to comments
- Edit existing comments.
- View resolved comments.

Introduces new upsell UI into the comments inspector panel.

### What to review
Comments functionality, nothing should change for enabled plans.


packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
```diff
@@ -152,6 +155,16 @@ function CommentFieldInner(props: FieldProps) {
       return
     }
 
+    if (commentsEnabled === 'read-only') {
+      if (upsellData) {
+        setUpsellDialogOpen(true)
+      } else {
+        // Open the comments inspector
```
pedrobonamin, line 162: It will open the comment inspector if for some reason the upsell data hasn't loaded yet to avoid stale clicks on the comments icon. 
It should not allow to add new comments when it's read only.

packages/sanity/src/structure/comments/src/components/upsell/UpsellDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+`
+
+const Image = styled.img`
+  object-fit: cover;
+  width: 100%;
+  height: 100%;
+  height: 200px;
+`
+// const StyledDialog = styled(Dialog)`
```
pedrobonamin, line 33: Need to confirm with lars the width of the dialog

packages/sanity/src/structure/comments/src/hooks/useResolveCommentsEnabled.ts
```diff
@@ -1,24 +1,38 @@
...
-
   // Check if the feature is enabled for the current document in the config
   const enabledFromConfig = useMemo(
     () => enabled({documentType, documentId: getPublishedId(documentId)}),
     [documentId, documentType, enabled],
   )
+  const value = useMemo(
+    () => ({
+      enabled: !isLoading && enabledFromConfig,
+      // TODO: Restore
```
pedrobonamin, line 30: This should be restored before releasing the changes

packages/sanity/src/core/studio/components/navbar/free-trial/DescriptionSerializer.tsx
```diff
@@ -1,27 +1,62 @@
...
+  return ``
+})
+
+const AccentSpan = styled.span`
+  color: var(--card-accent-fg-color);
+  --card-icon-color: var(--card-accent-fg-color);
+`
+
+const SemiboldSpan = styled.span`
+  font-weight: 600;
```
hermanwikner, line 45: Should we access the font weight value from the theme?

```ts
const SemiboldSpan = styled.span(({theme}) => {
  const {weights} = theme.sanity.fonts.text

  return css`
    font-weight: ${weights.semibold};
  `
})
```

packages/sanity/src/core/studio/components/navbar/free-trial/DescriptionSerializer.tsx
```diff
@@ -124,28 +190,42 @@ const components: PortableTextComponents = {
...
     ),
   },
 }
 
+interface DescriptionSerializerProps {
+  blocks: PortableTextBlock[]
+}
+
+/**
+ * Portable text serializer for the description text for upsell elements.
```
hermanwikner, line 218: Since this serializer is intended to be used for upsell blocks – should the name of the component be more specific to better reflects its purpose? Maybe something like `UpsellDescriptionSerializer`

packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
```diff
@@ -29,9 +30,9 @@ const HIGHLIGHT_BLOCK_VARIANTS: Variants = {
 }
 
 export function CommentsField(props: FieldProps) {
-  const isEnabled = useCommentsEnabled()
+  const {enabled} = useCommentsEnabled()
 
-  if (!isEnabled) {
+  if (!enabled) {
```
hermanwikner, line 35: Since we don't render the `CommentFieldInner` component if `enabled` is `false` – we won't show any comment field action on the fields if the enabled value is: `{enabled: false, reason: 'upsell'}`. 

I think it would be nice to create a separate field component that is used for upsell. Something like:

```ts
export function CommentsField(props: FieldProps) {
  const {enabled, reason} = useCommentsEnabled()

  // Comments upsell field component
  if (!enabled && reason === 'upsell') {
    return <CommentsFieldUpsell {...props} />
  }

  // Comments enabled field component
  if (enabled) {
    return <CommentFieldInner {...props} />
  }

  // Default field component without comments
}
```

That is, the `CommentsFieldInnerUpsell` handles opening the inspector / dialog.

packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
```diff
@@ -76,6 +77,7 @@ function CommentFieldInner(props: FieldProps) {
   const {element: boundaryElement} = useBoundaryElement()
 
   const currentUser = useCurrentUser()
+  const commentsEnabled = useCommentsEnabled()
```
hermanwikner, line 80: It would be nice to avoid using this hook in here and only render this component when comments is enabled. I think making creating a new `CommentsFieldUpsell` component (see earlier comment) would be nice and separate the concerns of the enabled field and the upsell field logic.

packages/sanity/src/structure/comments/plugin/inspector/CommentsInspector.tsx
```diff
@@ -37,9 +39,9 @@ const RootLayer = styled(Layer)`
 `
 
 export function CommentsInspector(props: DocumentInspectorProps) {
-  const isEnabled = useCommentsEnabled()
+  const {enabled} = useCommentsEnabled()
 
-  if (!isEnabled) return null
+  if (!enabled) return null
```
hermanwikner, line 44: Do we need to check if `reason` is 'upsell' here as well? Otherwise, I think we will return `null` even though we want to show the inspector with the upsell UI since we only check the `enabled` value.

packages/sanity/src/structure/comments/plugin/inspector/CommentsInspector.tsx
```diff
@@ -53,7 +55,7 @@ export function CommentsInspector(props: DocumentInspectorProps) {
 
 function CommentsInspectorInner(props: DocumentInspectorProps) {
   const {onClose} = props
-
+  const commentsEnabled = useCommentsEnabled()
```
hermanwikner, line 58: Since we use this hook in the parent component (`CommentsInspector`) – can we remove this hook here and pass the "mode" value as a prop instead?

```tsx
<RootLayer>
  <CommentsInspectorInner {...props} mode={mode} />
</RootLayer>
```

packages/sanity/src/structure/comments/src/hooks/useResolveCommentsEnabled.ts
```diff
@@ -1,24 +1,38 @@
...
   const {enabled} = useSource().document.unstable_comments
-
   // Check if the feature is enabled for the current document in the config
   const enabledFromConfig = useMemo(
     () => enabled({documentType, documentId: getPublishedId(documentId)}),
     [documentId, documentType, enabled],
   )
+  const value = useMemo(
+    () => ({
+      enabled: !isLoading && enabledFromConfig,
```
hermanwikner, line 29: This value needs to include the `featureEnabled` value as well. Currently, it only checks the enabled value resolved from the config API.

packages/sanity/src/structure/comments/src/components/upsell/CommentsUpsellDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+ */
+const StyledButton = styled(Button)`
+  position: absolute;
+  top: 12px;
+  right: 12px;
+  z-index: 20;
+  background: transparent;
+  border-radius: 9999px;
+  box-shadow: none;
+  color: white;
```
hermanwikner, line 19: Should this value be replaced with a value from the theme?

packages/sanity/src/structure/comments/src/components/upsell/CommentsUpsellDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+import {Button, Dialog} from '../../../../../ui-components'
+import {CommentsUpsellData} from '../../types'
+import {DescriptionSerializer} from 'sanity'
+
+/**
+ * Absolute positioned button to close the dialog.
+ */
+const StyledButton = styled(Button)`
+  position: absolute;
+  top: 12px;
```
hermanwikner, line 13: Should this value be replaced with a value from the theme?

packages/sanity/src/structure/comments/src/components/upsell/CommentsUpsellDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <Dialog
+      id="comments-upsell"
+      onClose={onClose}
+      onClickOutside={onClose}
+      __unstable_hideCloseButton
+      bodyHeight="fill"
+      padding={false}
+      footer={{
```
hermanwikner, line 51: Do we want to memoize this value?

packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
```diff
@@ -45,6 +46,7 @@ interface CommentThreadLayoutProps {
   onNewThreadCreate: (payload: CommentCreatePayload) => void
   onPathSelect?: (nextPath: CommentsSelectedPath) => void
   readOnly?: boolean
+  mode: CommentsUIMode
```
hermanwikner, line 49: Nitpick: alphabetize props

packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
```diff
@@ -59,6 +61,7 @@ export function CommentThreadLayout(props: CommentThreadLayoutProps) {
     onNewThreadCreate,
     onPathSelect,
     readOnly,
+    mode,
```
hermanwikner, line 64: Nitpick: alphabetize props

packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
```diff
@@ -133,6 +136,7 @@ export function CommentThreadLayout(props: CommentThreadLayoutProps) {
       {canCreateNewThread && (
         <ThreadCard onClick={handleNewThreadClick} tone={isSelected ? 'caution' : undefined}>
           <CreateNewThreadInput
+            mode={mode}
```
hermanwikner, line 139: Nitpick: alphabetize props

packages/sanity/src/structure/comments/src/components/reactions/CommentReactionsMenuButton.tsx
```diff
@@ -66,18 +67,21 @@ export function CommentReactionsMenuButton(props: CommentReactionsMenuButtonProp
...
 
     // Clone the button element and add the necessary props.
     return cloneElement(btn, {
       'aria-expanded': open,
       'aria-haspopup': 'true',
-      disabled: readOnly,
       id: 'reactions-menu-button',
       onClick: handleClick,
       ref: setButtonElement,
+      disabled: readOnly || mode === 'upsell',
```
hermanwikner, line 82: Nitpick: alphabetize props

packages/sanity/src/structure/comments/src/context/comments/CommentsProvider.tsx
```diff
@@ -50,8 +51,8 @@ export interface CommentsProviderProps {
  */
 export const CommentsProvider = memo(function CommentsProvider(props: CommentsProviderProps) {
   const {children, documentId, documentType, isCommentsOpen, onCommentsOpen} = props
+  const commentsEnabled = useCommentsEnabled()
```
hermanwikner, line 54: I'd like to keep the `CommentsProvider` not dependent on other contexts. I think it would be better to move this logic (in `handleSetStatus`) to the `handleChangeView` function in the `CommentsInspector` instead.

packages/sanity/src/structure/comments/src/hooks/useResolveCommentsEnabled.ts
pedrobonamin: This is intentional to support disabling comments from config.

You could have 3 type of values here:
```
 {enabled:false, reason: null}  //  Disabled from config, won't show anything related to comments.- Will remove reason from here to make it simpler to follow.
 {enabled: true, reason: null}  // Current UI, no changes
 {enabled: true, reason:  "upsell} // Shows upsell UI.
```

Pushed an update here to clarify.

packages/sanity/src/structure/comments/src/components/list/CommentThreadLayout.tsx
pedrobonamin: Updated, but I think we should aim to make it part of the lint rules if we want to keep it 100% consistent everywhere.

packages/sanity/src/structure/comments/src/components/upsell/CommentsUpsellDialog.tsx
pedrobonamin: I don't see anything in the parent components or in this component that will produce unwanted rerenders and would benefit from memoizing this value, given it's not an expensive to calculate. But would love to hear your thoughts on why you consider we shuold.
This is a great post that gave me guidelines to decide to use or not a memo value, https://kentcdodds.com/blog/usememo-and-usecallback#how-is-usememo-different-but-similar .

packages/sanity/src/core/studio/components/navbar/free-trial/DescriptionSerializerUtils.ts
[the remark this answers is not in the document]
pedrobonamin: Yep! Thanks

packages/sanity/src/core/studio/components/navbar/free-trial/DescriptionSerializer.tsx
pedrobonamin: Yeap, sounds good, I moved it to a reusable location instead of navbar and renamed it.

packages/sanity/src/structure/comments/src/context/comments/CommentsProvider.tsx
pedrobonamin: I'm not 100% sure about this, this context it's already depending on `useCommentsSetup`, it can be moved to `commentsInspector` but if later we add another button or any other async action that sets the `resolved` states we will need to duplicate the logic added here.

packages/sanity/src/structure/comments/plugin/inspector/CommentsInspector.tsx
pedrobonamin: enabled will be true when reason is `upsell` .

packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
pedrobonamin: It will require to duplicate all the logic already added into the CommentsInput component, as the only thing that changes is the handleClick action when the input has no comments.
If it has comments, it does the same as before (opens the inspector and scrolls into view).

pedrobonamin: however, I could pass the value to the <CommentsFieldInner> by props if we want to avoid using the `useCommentsEnabled` hook in it

packages/sanity/src/structure/comments/src/hooks/useResolveCommentsEnabled.ts
```diff
@@ -1,24 +1,41 @@
...
 
-  const isEnabled = !isLoading && featureEnabled && enabledFromConfig
+  const value = useMemo(() => {
+    if (isLoading || !enabledFromConfig) {
+      return {enabled: false, reason: null}
+    }
+
+    return {
+      enabled: true,
+      // reason: featureEnabled ? null : 'upsell',
```
pedrobonamin, line 35: This needs to be restored before merging

packages/sanity/src/structure/comments/plugin/inspector/CommentsInspector.tsx
```diff
@@ -92,6 +98,26 @@ function CommentsInspectorInner(props: DocumentInspectorProps) {
   }, [comments.loading, ready])
 
   const {selectedPath, setSelectedPath} = useCommentsSelectedPath()
+  const {upsellData, telemetryLogs} = useCommentsUpsell()
+
+  useEffect(() => {
+    if (mode === 'upsell') {
+      if (selectedPath?.origin === 'form') {
+        telemetryLogs.panelViewed('field_action')
+      } else if (selectedPath?.origin === 'url') {
```
hermanwikner, line 107: When I test this, it seems like the telemetry event won't trigger when navigating to a comment from a URL. I think this is due to the `selectedPath` state being set in the `useEffect` at line 314 when navigating from a URL, which happens after this particular effect has already executed. That is, the `selectedPath` state is not yet available when this effect runs when navigating from a URL. I think this can be solved by moving the link telemetry event inside the effect at line 314.

What do you think of moving the `panelDismissed` event to the `handleCloseInspector` function instead? This because effects is running twice in strict mode, so that way we can avoid that the dismissed event is triggered on mount. The downside of this is that we won't trigger any telemetry event when dismissing the inspector from the button in the document panel header. However, there is a `onClose` function available that can be passed to [the inspector definition](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/comments/plugin/inspector/index.ts#L20) – maybe that one can be used to trigger the event?

pedrobonamin: re url not working:
- Did a change, will check for the `commentIdParamRef.current` instead, which is available while loading the comments. It's weird it initially worked as I expected. But thanks for teh catch.

re document_action
- Those names are defined by the growth team so I think we need to stick to it, regarding changing it o onOpen I tried it but we don't have access to the upsell context to determine if we should trigger the event or not. 

re on dismiss:
-  Tried the onClose option, it also has an onOpen, but we don't have access to the upsell context to determine if we should trigger the event or not.

packages/sanity/src/core/studio/upsell/upsellDescriptionSerializer/helpers.ts
```diff
@@ -0,0 +1,48 @@
...
+const TRANSFORMATIONS: {
+  [type: string]: (
+    child: PortableTextObject | PortableTextSpan,
+    idx: number,
+    block: PortableTextBlock,
+  ) => PortableTextObject
+} = {
+  inlineIcon: (child, idx, block) => {
+    if (!isPortableTextTextBlock(block)) return block
+    const hasTextLeft = Boolean(
```
drewlyton, line 17: Do these array accesses need to be checked/guarded in some way? Just wondering if accessing [-1] will ever happen/be a problem or if accessing [length+1] will ever throw an error

packages/sanity/src/structure/comments/src/context/upsell/CommentsUpsellProvider.tsx
```diff
@@ -0,0 +1,175 @@
...
+        // silently fail
+      }
+    })
+
+    return () => {
+      sub.unsubscribe()
+    }
+  }, [client, projectId])
+
+  useEffect(() => {
```
hermanwikner, line 141: What do you think of triggering this telemetry event like below instead of in an effect?

```ts
  const handleOpenDialog = useCallback(() => {
    setUpsellDialogOpen(true)

    telemetry.log(UpsellDialogViewed, {
      feature: FEATURE,
      type: 'modal',
      source: 'field_action',
    })
  }, [telemetry])
  ```

**Review by hermanwikner — APPROVED** [review]
Thank you @pedrobonamin!

File: packages/sanity/src/structure/comments/src/hooks/useCommentsUpsell.ts [hunk 23]
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @beta
+ * @hidden
+ */
+export function useCommentsUpsell(): CommentsUpsellContextValue {
+  const value = useContext(CommentsUpsellContext)
+
+  if (!value) {
+    // Instead of throwing, we return a dummy value to avoid breaking the CommentsField implementation, given the context is optional.
```
drewlyton, line 13: Might be nice to add a `console.error` here to notify future devs in case the provider moves or some other implementation causes this to break

drewlyton (approved):
Telemetry stuff looks fine to me (although I am new so take with grain of 🧂) - left some general questions/comments. Might want to wait for @svirs's review too before merging

packages/sanity/src/core/studio/upsell/upsellDescriptionSerializer/helpers.ts
pedrobonamin: The upper check `isPortableTextTextBlock` is checking for children to be defined and to be of type array. 
And later when accessing the children `block.children[idx + 1]?._type` will work as a safeguard in case it's trying to access to an undefined children

packages/sanity/src/structure/comments/src/hooks/useCommentsUpsell.ts
pedrobonamin: We can't add a `console.error` here because the context will be undefined when you are not in upsell mode. So all the users with the feature properly enabled will see the `console.error` added here

PR #5571, opened 2024-01-26
feat(backup): add new command dataset backup for server-side backups

j33ty:
# Description

This PR introduces backup command for server-side dataset backups: `sanity backup -h`. This command primarily intended to be a consumer of Backup APIs and has very little client side logic. If you are a Sanity colleague, kindly refer to **#prj-dataset-backups** for more context.

# Usage

User experience of interaction with the CLI has been kept very similar to using `sanity export`. The downloaded tar file should be importable through `sanity dataset import`.

## Interaction with server-side backups
```
; sanity backup -h
usage: sanity backup [--default] [-v|--version] [-d|--debug] [-h|--help] <command> [<args>]

Commands:
   disable   Disable backup for a dataset.
   download  Download a dataset backup to a local file.
   enable    Enable backup for a dataset.
   list      List available backups for a dataset.

See 'sanity help backup <command>' for specific information on a subcommand.
```

#### Usage
```
; sanity backup enable -h
usage: sanity backup enable [DATASET_NAME]

   Enable backup for a dataset.

Examples
  sanity backup enable DATASET_NAME
```

#### Sample Output
```
; sanity backup enable production
Enabled daily backups for dataset production.
*Retention policies may apply depending on your plan and agreement.*
```

#### Usage
```
; sanity backup list -h
usage: sanity backup list [DATASET_NAME]

   List available backups for a dataset.

Options
  --limit <int>     Maximum number of backups returned. Default 30.
  --after <string>  Only return backups after this timestamp (inclusive)
  --before <string> Only return backups before this timestamp (exclusive). Cannot be younger than <after> if specified.

Examples
  sanity backup list DATASET_NAME
  sanity backup list DATASET_NAME --limit 50
  sanity backup list DATASET_NAME --after 2024-01-01 --limit 10
  sanity backup list DATASET_NAME --after 2024-01-01 --before 2024-01-10
```

#### Sample Output
```
; sanity backup list
? Select the dataset name: production
┌──────────┬────────────┬─────────────────────────────────────────────────┐
│ RESOURCE │ CREATED AT │ BACKUP ID                                       │
├──────────┼────────────┼─────────────────────────────────────────────────┤
│ Dataset  │ 2024-01-25 │ 2024-01-25-005f0043-b2c2-4b9b-95a4-4c65222bbdd2 │
│ Dataset  │ 2024-01-25 │ 2024-01-25-40a9c750-142a-473c-8584-0d49fc5b8e6b │
│ Dataset  │ 2024-01-26 │ 2024-01-26-006fd426-4cc6-435f-bbe3-c4ee50068e3b │
│ Dataset  │ 2024-01-26 │ 2024-01-26-58922f75-36cc-4161-8995-3124e9c0434a │
│ Dataset  │ 2024-01-26 │ 2024-01-26-8cb692e4-7ab7-46b9-9917-bf4342074477 │
│ Dataset  │ 2024-01-26 │ 2024-01-26-addfa998-ba2f-4531-95c2-e3633ec8c80b │
└──────────┴────────────┴─────────────────────────────────────────────────┘
```

#### Usage
```
; sanity backup download -h
usage: sanity backup download [DATASET_NAME]

   Download a dataset backup to a local file.

Options
  --backup-id <string> The backup ID to download. (required)
  --out <string>       The file or directory path the backup should download to.
  --overwrite          Allows overwriting of existing backup file.
  --concurrency <num>  Concurrent number of backup item downloads. (max: 24)

Examples
  sanity backup download DATASET_NAME --backup-id 2024-01-01-backup-1
  sanity backup download DATASET_NAME --backup-id 2024-01-01-backup-2 --out /path/to/file
  sanity backup download DATASET_NAME --backup-id 2024-01-01-backup-3 --out /path/to/file --overwrite
```

#### Sample Output
```
; sanity backup download production --backup-id 2024-01-25-005f0043-b2c2-4b9b-95a4-4c65222bbdd2 --out backup/ --overwrite --concurrency 5
╭───────────────────────────────────────────────────────────╮
│                                                           │
│ Downloading backup for:                                   │
│ projectId: h5hc8cgs                                       │
│ dataset: production                                       │
│ backupId: 2024-01-25-005f0043-b2c2-4b9b-95a4-4c65222bbdd2 │
│                                                           │
╰───────────────────────────────────────────────────────────╯

Downloading backup to "/Users/ry/Documents/sanity/sanity/dev/test-studio/out/production-backup-2024-02-02-19083f95-1741-4e29-997a-76b56c773e24.tar.gz"
✔ [2.2s] Setting up backup environment...
✔ [5.4s] Reading backup files... (545/545)
✔ [2.3s] Downloading documents and assets... (545/545)
✔ [130ms] Archiving files into a tarball, 4.16 MB bytes written...
✔ [1ms] Cleaning up temporary files at /var/folders/1/foo/A/backup-1707275511960-78377
✔ Backup download complete [10.2s]
```

#### Usage
```
; sanity backup disable -h
usage: sanity backup disable [DATASET_NAME]

   Disable backup for a dataset.

Examples
  sanity backup disable DATASET_NAME
```

#### Sample Output
```
; sanity backup disable production
Disabled daily backups for dataset 
```

# Potential Optimizations
1. For `sanity backup download`, we are first iterating over all the backup file and then downloading them. This behaviour may not be scalable when number of files is huge. To handle this, we can start downloading files as soon as we are reading them.
2. For `sanity backup download`, if it is possible, stream downloaded files directly into the tarball without writing them to disk first.

# Notes for release
Since we are not marketing this feature at the moment, it is okay to skip these commits from public release notes.


packages/sanity/src/_internal/cli/commands/backup/backupGroup.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// defaultApiVersion is the backend API version used for dataset backup.
+export const defaultApiVersion = 'vX'
+
+const datasetBackupGroup: CliCommandGroupDefinition = {
+  name: 'backup',
+  signature: '[COMMAND]',
+  description: 'Manage dataset backups.',
+  isGroupRoot: true,
+  hideFromHelp: false,
```
rneatherway, line 11: I think we do want to hide this for now don't we?

packages/sanity/src/_internal/cli/commands/backup/enableBackupCommand.ts
```diff
@@ -0,0 +1,48 @@
...
+      await client.request({
+        method: 'PUT',
+        headers: {Authorization: `Bearer ${token}`},
+        uri: `/projects/${projectId}/datasets/${datasetName}/settings/backups`,
+        body: {
+          enabled: true,
+        },
+      })
+
+      output.print(`${chalk.green(`Enabled daily backups for dataset ${datasetName}.\n`)}`)
```
rneatherway, line 35: I'm not sure about including the "daily" part here in case we change the schedule for some reason. Could we just remove that word?

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface ListDatasetBackupFlags {
+  before?: string
+  after?: string
+  limit?: string
+}
+
+type ListBackupRequestQueryParams = {
+  start?: string
+  end?: string
```
rneatherway, line 18: I've made these `after` and `before` in the API to match the CLI.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
```diff
@@ -0,0 +1,132 @@
...
+          {name: 'createdAt', title: 'CREATED AT', alignment: 'left'},
+          {name: 'backupId', title: 'BACKUP ID', alignment: 'left'},
+        ],
+      })
+
+      response.backups.forEach((backup: ListBackupResponseItem) => {
+        const {id, createdAt} = backup
+        table.addRow({
+          resource: 'Dataset',
+          createdAt: lightFormat(Date.parse(createdAt), 'yyyy-MM-dd'),
```
rneatherway, line 122: Were you planning to include the full timestamp here? I think we discussed that.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,297 @@
...
+    const start = Date.now()
+    const progressSpinner = newProgress(output, 'Setting up backup environment...')
+
+    // Create temporary directory to store files before bundling them into the archive at outputPath.
+    // We are adding unix milliseconds and a random key to try to back up in a unique location on each attempt.
+    // Temporary directories are normally deleted at the end of backup process, any unexpected exit may leave them
+    // behind, hence it is important to create a unique directory for each attempt.
+    // We intentionally avoid `datasetName` and `backupId`in path since some operating system have a max length
+    // of 256 chars on path names and both of them can be quite long in some cases.
+    const tmpOutDir = path.join(tmpdir(), `backup-${Date.now()}-${randomKey(5)}`)
```
rneatherway, line 94: Can you use `fs.mkdtemp(...)` here?

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,297 @@
...
+
+    // Create required directories if they don't exist.
+    for (const dir of [
+      outDir,
+      tmpOutDir,
```
rneatherway, line 99: This will be created by the next two, so you can remove this.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,297 @@
...
+      path.join(tmpOutDir, 'images'),
+      path.join(tmpOutDir, 'files'),
+    ]) {
+      if (!existsSync(dir)) {
```
rneatherway, line 103: You can remove this check as it's safe to call `mkdirSync` even if the directory exists when `recursive: true`. This can happen even _with_ the check as the directory could be created by another process after checking that it doesn't exist.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,297 @@
...
+        mkdirSync(dir, {recursive: true})
+      }
+    }
+
+    debug('Writing to temporary directory %s', tmpOutDir)
+    const tmpOutDocumentsFile = path.join(tmpOutDir, 'data.ndjson')
+
+    // Handle concurrent writes to the same file using mutex.
+    const docOutStream = createWriteStream(tmpOutDocumentsFile, {flags: 'a'})
```
rneatherway, line 112: Why `flags: 'a'`? We're creating the file from scratch AIUI so there's no need to use append mode.

packages/sanity/src/_internal/cli/actions/backup/resolveApiClient.ts
```diff
@@ -0,0 +1,44 @@
...
+  const {apiClient} = context
+
+  let client = apiClient()
+  const {projectId, token} = client.config()
+
+  if (!projectId) {
+    throw new Error('Project ID not defined')
+  }
+
+  // Do not use dataset configured in Sanity config since dataset specified in command should override.
```
rneatherway, line 26: Should we let the dataset in the config be the default?

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
```diff
@@ -0,0 +1,52 @@
...
+
+async function downloadAsset(
+  url: string,
+  fileName: string,
+  fileType: string,
+  outDir: string,
+): Promise<void> {
+  // File names that contain a path to file (sanity-storage/assets/file-name) fail when archive is created, so we
+  // want to handle them by taking the base name as file name.
+  const normalizedFileName = path.basename(fileName)
```
rneatherway, line 22: I don't quite follow this part, can you give me an example?

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
```diff
@@ -0,0 +1,46 @@
...
+      uri: `/projects/${projectId}/datasets/${datasetName}/backups`,
+      query: {limit: maxBackupIdsShown.toString()},
+    })
+
+    if (response && response.backups && response.backups.length > 0) {
+      const backupIdChoices = response.backups.map((backup: {id: string}) => ({
+        value: backup.id,
+      }))
+      const selected = await prompt.single({
+        message: `Select Backup ID to use (only last ${maxBackupIdsShown} shown)`,
```
rneatherway, line 31:
```suggestion
        message: `Select backup ID to use (only last ${maxBackupIdsShown} shown)`,
```

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,297 @@
...
+        msg = String(error)
+      }
+      throw new Error(`Downloading dataset backup failed: ${msg}`)
+    }
+
+    progressSpinner.set({step: `Archiving files into a tarball...`, update: true})
+    try {
+      await archiveDir(tmpOutDir, outFilePath, (processedBytes: number) => {
+        progressSpinner.update({
+          step: `Archiving files into a tarball, ${humanFileSize(processedBytes)} bytes written...`,
```
rneatherway, line 176: I remember that you showed me an `assets.json` but I don't see that created here. Is it needed?

package.json
```diff
@@ -106,6 +106,7 @@
     "@sanity/uuid": "^3.0.2",
     "@types/jest": "^29.5.6",
     "@types/node": "^14.18.9",
+    "@types/progress-stream": "^2.0.5",
```
binoy14, line 109: this needs to be in `packages/sanity/packages.json` that should fix the type issues

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
j33ty: Fixed it.

packages/sanity/src/_internal/cli/actions/backup/resolveApiClient.ts
j33ty: `sanity dataset export` does not use config dataset, so I followed the same pattern. A missing dataset name in config results in a dummy name which is a weird experience for backup/export.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Good point. This is leftover from previous commit, where I was creating new write pointer for each document. Removed it now.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Sure, added it.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
j33ty: Yes, added timestamp to the format here.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
j33ty: Updated the query param names.

packages/sanity/src/_internal/cli/commands/backup/enableBackupCommand.ts
j33ty: Removed it.

packages/sanity/src/_internal/cli/commands/backup/backupGroup.ts
j33ty: Agree. Set `hideFromHelp: true` now.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Done.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Good point, removed.

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
j33ty: Updated inline comment:
```
  // File names that contain a path to file (e.g. sanity-storage/assets/file-name.tar.gz) fail when archive is
  // created due to missing parent dir (e.g. sanity-storage/assets), so we want to handle them by taking
  // the base name as file name.
```
Does this make sense?

j33ty:
I have rebased this branch with latest `next` and force-pushed.

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
rneatherway: Oh, I understand now. I think this is from when the API was returning a path early on but now it only returns a filename. So I think you could remove this if you like. Up to you.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
rneatherway: I think you missed this comment.

packages/sanity/src/_internal/cli/actions/backup/resolveApiClient.ts
rneatherway: That makes sense 👍

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: From a quick glance at the export/import code, it seems like this file is used to track asset props for all the assets in one file and delete the actual asset documents from export. So, while importing, you would go over all the regular documents and then go over this asset.json file to import corresponding asset to CDN and its props to Content Lake.

In our case, we are already storing asset documents and asset files as they are. Upon import, they are being stored correctly in Content Lake and CDN. There may be a slight difference in these two variations, so going to look into this in-depth now.

rneatherway: If it isn't needed then that's great! I'm just checking in on this point.

packages/sanity/package.json
```diff
@@ -264,6 +267,7 @@
     "log-symbols": "^2.2.0",
     "mendoza": "^3.0.0",
     "module-alias": "^2.2.2",
+    "moment": "^2.30.1",
```
bjoerge, line 270: We're phasing out our usage of moment.js. Could you use [date-fns](https://date-fns.org/) instead? It's already a dependency.

j33ty: Removed moment.js in latest commit.

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
j33ty: I vote to keeping it even if server does not return paths in file name now. Without this CLI fails hard on a slash.

packages/sanity/src/_internal/cli/actions/backup/archiveDir.ts
```diff
@@ -0,0 +1,54 @@
...
+    })
+
+    // Pipe archive data to the file
+    archive.pipe(archiveDestination)
+    archive.directory(tmpOutDir, false)
+    archive.finalize()
+  })
+}
+
+function humanFileSize(size: number): string {
```
binoy14, line 49: 💭  this might be a good candidate to move to `util` directory.

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // Fetch last $maxBackupIdsShown backups for this dataset.
+  // We expect here that API returns backups sorted by creation date in descending order.
+  const response = await client.request({
+    headers: {Authorization: `Bearer ${token}`},
+    uri: `/projects/${projectId}/datasets/${datasetName}/backups`,
+    query: {limit: maxBackupIdsShown.toString()},
+  })
+
+  if (response && response.backups && response.backups.length > 0) {
```
binoy14, line 25: 💡 could this simply be the following suggestion?

```suggestion
  if (response?.backups?.length > 0) {
```

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const {projectId, token, client} = await resolveApiClient(context, datasetName, defaultApiVersion)
+
+  // Fetch last $maxBackupIdsShown backups for this dataset.
+  // We expect here that API returns backups sorted by creation date in descending order.
+  const response = await client.request({
+    headers: {Authorization: `Bearer ${token}`},
+    uri: `/projects/${projectId}/datasets/${datasetName}/backups`,
+    query: {limit: maxBackupIdsShown.toString()},
+  })
```
binoy14, line 23: it might make sense here to do some error handling, wrapping it in a `try...catch` or I believe it will throw an unhandled promise error

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
```diff
@@ -0,0 +1,52 @@
+import path from 'path'
+import {createWriteStream} from 'fs'
```
binoy14, line 2: we are slowly moving to importing node imports using the new namespace way, to make it clear it's using node's internal import. Sorry it's not in the style guide.


```suggestion
import path from 'node:path'
import {createWriteStream} from 'node:fs'
```

packages/sanity/src/_internal/cli/actions/backup/archiveDir.ts
```diff
@@ -0,0 +1,54 @@
+import zlib from 'zlib'
+import {createWriteStream} from 'fs'
```
binoy14, line 2:
```suggestion
import zlib from 'zlib'
import {createWriteStream} from 'node:fs'
```

packages/sanity/src/_internal/cli/actions/backup/fetchNextBackupPage.ts
```diff
@@ -0,0 +1,87 @@
+import {Readable} from 'stream'
```
binoy14, line 1:
```suggestion
import {Readable} from 'node:stream'
```

packages/sanity/src/_internal/cli/actions/backup/resolveApiClient.ts
```diff
@@ -0,0 +1,45 @@
...
+  // If no dataset provided, explicitly ask for dataset instead of using dataset
+  // configured in Sanity config. Aligns with `sanity dataset export` behavior.
+  let selectedDataset: string = datasetName
+  if (!selectedDataset) {
+    selectedDataset = await chooseDatasetPrompt(context, {
+      message: 'Select the dataset name:',
+    })
+  }
+
+  client = client.clone().config({dataset: datasetName, apiVersion})
```
binoy14, line 35: probably not a big deal but I believe `client.withConfig({ dataset: datasetName, apiVersion})` does the same thing

packages/sanity/src/_internal/cli/commands/backup/backupGroup.ts
```diff
@@ -0,0 +1,26 @@
+import {CliCommandGroupDefinition} from '@sanity/cli'
+
+// defaultApiVersion is the backend API version used for dataset backup.
+export const defaultApiVersion = 'vX'
```
bjoerge, line 4: Should this not be a proper API version instead of `vX`? If it's intentional to keep it at vX, can you add a comment explaining why and what it would take to move it to an official version?

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,288 @@
...
+import resolveApiClient from '../../actions/backup/resolveApiClient'
+import downloadAsset from '../../actions/backup/downloadAsset'
+import downloadDocument from '../../actions/backup/downloadDocument'
+import newProgress from '../../actions/backup/progressSpinner'
+import {PaginatedGetBackupStream, File} from '../../actions/backup/fetchNextBackupPage'
+import {archiveDir, humanFileSize} from '../../actions/backup/archiveDir'
+import cleanupTmpDir from '../../actions/backup/cleanupTmpDir'
+import {defaultApiVersion} from './backupGroup'
+
+const debug = require('debug')('sanity:backup')
```
bjoerge, line 27: Not a blocker, but ideally this should be

```suggestion
import createDebug from 'debug'

const debug = createDebug('sanity:backup')
```

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,288 @@
...
+
+const DEFAULT_DOWNLOAD_CONCURRENCY = 10
+const MAX_DOWNLOAD_CONCURRENCY = 24
+
+type DownloadBackupOptions = {
```
bjoerge, line 32: (Nit, so feel free to ignore: we tend to use `interface` when we can)

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,288 @@
+import {createWriteStream, existsSync, mkdirSync} from 'fs'
+import {tmpdir} from 'os'
+import path from 'path'
```
binoy14, line 3:
```suggestion
import {createWriteStream, existsSync, mkdirSync} from 'node:fs'
import {tmpdir} from 'node:os'
import path from 'node:path'
```

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,288 @@
...
+    progressSpinner.succeed()
+  },
+}
+
+// prepareBackupOptions validates backup options from CLI and prepares Client and DownloadBackupOptions.
+async function prepareBackupOptions(
+  context: CliCommandContext,
+  args: CliCommandArguments,
+): Promise<[SanityClient, DownloadBackupOptions]> {
+  const flags = args.extOptions
```
binoy14, line 192: Take a look at this [PR](https://github.com/sanity-io/sanity/pull/5616/files#diff-5dd59c387c299ba5b1746c4c0a69aed0ec8a759c264d799e8462f6a03f7e4a91R68) I believe we want to use `yargs` to parse the flags, it enforces types and in future we want to move to making sure all flags are defined properly. It will also help remove some of the type checking code below

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
```diff
@@ -0,0 +1,288 @@
...
+    output.print('')
+    output.print(`Downloading backup to "${chalk.cyan(outFilePath)}"`)
+
+    const start = Date.now()
+    const progressSpinner = newProgress(output, 'Setting up backup environment...')
+
+    // Create a unique temporary directory to store files before bundling them into the archive at outputPath.
+    // Temporary directories are normally deleted at the end of backup process, any unexpected exit may leave them
+    // behind, hence it is important to create a unique directory for each attempt.
+    const tmpOutDir = await mkdtemp(path.join(tmpdir(), `backup-`))
```
bjoerge, line 92: would be great if we could add `sanity-` to the tempfile here so we can leave a hint about who made it.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const listDatasetBackupCommand: CliCommandDefinition<ListDatasetBackupFlags> = {
+  name: 'list',
+  group: 'backup',
+  signature: '[DATASET_NAME]',
+  description: 'List available backups for a dataset.',
+  helpText,
+  action: async (args, context) => {
+    const {output, chalk} = context
+    const flags = args.extOptions
```
binoy14, line 51: same comment here, it would be good to re-parse the arguments using `yargs`

bjoerge (approved):
Looks great! Left a few comments, mostly nice-to-haves.

binoy14 (approved):
most of them are nit-picky comments or questions. Not a blocker

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
```diff
@@ -0,0 +1,137 @@
...
+    try {
+      response = await client.request<ListBackupResponse>({
+        headers: {Authorization: `Bearer ${token}`},
+        uri: `/projects/${projectId}/datasets/${datasetName}/backups`,
+        query: {...query},
+      })
+    } catch (error) {
+      const msg = error.statusCode
+        ? error.response.body.message
+        : error.message || error.statusMessage
```
binoy14, line 95: was able to get this message, maybe a default message if none of the messages are populated? 

Screenshot 2024-02-09 at 12 00 01 PM

packages/sanity/src/_internal/cli/commands/backup/backupGroup.ts
j33ty: This should be released with `vX` Content Lake API version. Added the reason inline:
```
// First version of the backup API is vX since this feature is not yet released
// and formal API documentation is pending.
```

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Fixed it.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Changed type to interface.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Goood point. Added it.

packages/sanity/src/_internal/cli/actions/backup/archiveDir.ts
j33ty: Moved it.

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
j33ty: Nice. Simplified it.

packages/sanity/src/_internal/cli/actions/backup/chooseBackupIdPrompt.ts
j33ty: Guarded it with try and catch.

packages/sanity/src/_internal/cli/actions/backup/downloadAsset.ts
j33ty: Added namespacing.

packages/sanity/src/_internal/cli/actions/backup/fetchNextBackupPage.ts
j33ty: Fixed it.

packages/sanity/src/_internal/cli/actions/backup/resolveApiClient.ts
j33ty: Replaced it with `withConfig`.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Fixed it.

packages/sanity/src/_internal/cli/commands/backup/downloadBackupCommand.ts
j33ty: Good point. Updated code to follow this pattern.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
j33ty: This is fixed as well.

packages/sanity/src/_internal/cli/commands/backup/listBackupCommand.ts
j33ty: Fixed in latest commit to print the proper error.

j33ty:
I have rebased from latest next branch with new linter config and reformatted files in this PR.

binoy14 (approved):
LGTM, thanks!

PR #5576, opened 2024-01-26
feat(comments): add optional intent link to comments

rdunk:
### Description

This PR adds a mechanism for associating and displaying an intent link alongside a comment. This is so other tools (namely Presentation) can augment comments with added context about where a comment was made.

An optional provider is exported for use in other tools, if this provider is used and intent context is available it will be attached to the comment context payload.

### What to review

The workshop `CommentsListStory` has been updated to include an example of an intent link.

### Testing

No tests added.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
@@ -342,7 +342,7 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
 
             {comment.context?.intent && (
               <Box flex={1}>
-                <IntentText muted size={0} textOverflow="ellipsis" title={''}>
+                <IntentText muted size={0} textOverflow="ellipsis">
                   <Translate
                     t={t}
                     i18nKey="list-item.layout-context"
--- a/packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
@@ -4,12 +4,14 @@ import {Box, Card, Flex, Stack, Text, TextSkeleton, useClickOutside} from '@sani
 import {useCallback, useMemo, useRef, useState} from 'react'
 import {
   type RelativeTimeOptions,
+  Translate,
   useDateTimeFormat,
   useDidUpdate,
   useRelativeTime,
   useTranslation,
   useUser,
 } from 'sanity'
+import {IntentLink} from 'sanity/router'
 import styled, {css} from 'styled-components'
 
 import {commentsLocaleNamespace} from '../../../i18n'
@@ -44,6 +46,16 @@ const TimeText = styled(Text)(({theme}) => {
 
   return css`
     min-width: max-content;
+    --card-fg-color: ${fg};
+    color: var(--card-fg-color);
+  `
+})
+
+const IntentText = styled(Text)(({theme}) => {
+  const isDark = theme.sanity.color.dark
+  const fg = hues.gray[isDark ? 200 : 800].hex
+
+  return css`
     --card-fg-color: ${fg};
     color: var(--card-fg-color);
   `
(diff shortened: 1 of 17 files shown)


packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -298,28 +310,58 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
...
+                      </TimeText>
+                    )}
+                  </Flex>
+                )}
+              </Flex>
             </Flex>
+
+            {comment.context?.intent && (
+              <Box flex={1}>
+                <IntentText muted size={0} textOverflow="ellipsis" title={''}>
```
hermanwikner, line 345: Is this `title={''}` needed?

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
```diff
@@ -298,28 +310,58 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
...
+                          ({t('list-item.layout-edited')})
+                        </time>
+                      </TimeText>
+                    )}
+                  </Flex>
+                )}
+              </Flex>
             </Flex>
+
+            {comment.context?.intent && (
```
hermanwikner, line 343: Do we want to show the intent link for all replies – or would it be sufficient to show it only on the root comment?

Screenshot 2024-02-12 at 18 11 45

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
rdunk: Not as far as I know! Thanks for spotting.

packages/sanity/src/structure/comments/src/components/list/CommentsListItemLayout.tsx
rdunk: Hmm, it's a good question, I don't love the left side visually. I suppose it only serves a purpose if a response needs to show some different context, say to contrast how some content looks across two different pages.

Should we hide unless it differs from the root context?

rdunk: Kind of like this?

Screenshot 2024-02-12 at 20 40 38

I've made the necessary changes on a local stash, so ready to push up if we want to do this.

hermanwikner: I might miss some things here – but is it currently possible to add a reply with a different context than the root comment? How would you do that?

rdunk: The Presentation tool tracks the URL the preview pane is displaying, so for example a document of type `shoe` could be populating two pages: a listing page `/shoes` and an item page `/shoes/slug`.

It's probably not typical, but you might have e.g. a comment chain about the size of the title. You could make a single comments thread about how the title looks on each page, with intent links to each context.

hermanwikner: Thanks for clarifying! Then I think your suggestion makes sense, that is, to hide the intent link unless it differs from the root context 👍

rdunk: Great! Added in https://github.com/sanity-io/sanity/pull/5576/commits/30b068a1b463e62252f884a6711cd2438dcf35f8.

hermanwikner: Thanks!

Another thought (no block for merge): should we check that Presentation is being used to determine if the intent link should be displayed? Or maybe by checking the presence of the intent context provider? So that we know for sure that the intent link actually has a target in the Studio.

rdunk: Yes that definitely sounds like a good idea, I'll create a Linear issue for it.

Happy to resolve this thread?

hermanwikner (approved):
Thanks @rdunk!

PR #5583, opened 2024-01-29
feat(cli): add new `sanity schema validate` command

ricokahler:
### Description

This adds a new CLI command: `sanity schema validate`.

In addition to the new command, this PR also refactors some interfaces to allow for the schema to be resolved without resolving the rest of the config. This was done because the config resolution code currently throws if there are schema validation errors and the refactors make it so that the code path is always straightforward.

Also, this PR:
- removes the `info` level schema validation markers
- remove the list of available schema types when one is misused since this list can get excessively long in production studios

Note: I made the decision to not mimic the tree output of the documents command because the schema validation error messages are too long to fit in the terminal like that and I don't think there will be many cases where we'll have multiple errors in one top-level type.

https://github.com/sanity-io/sanity/assets/10551026/c9efec5c-e934-4e67-9d95-3b26d6e18885

### What to review

- Try out the command yourself in a few different studios
- Review the DX
- Review the copy! Typos happen.

### Testing

I tried this out in the `test-studio` and added a few tests for formatting.

### Notes for release

Adds a new CLI command `sanity schema validate`


packages/@sanity/schema/src/sanity/validation/createValidationResult.ts
```diff
@@ -53,6 +53,3 @@ export const error = (message: string, helpId?: string | null): SchemaValidation
 
 export const warning = (message: string, helpId?: string | null): SchemaValidationResult =>
   createValidationResult('warning', message, helpId)
-
-export const info = (message: string, helpId?: string | null): SchemaValidationResult =>
```
ricokahler, line 57: this was unused

binoy14 (approved):
LGTM! Thanks for doing this!

kevindice:
Just saw this in the changelog - thank you!

PR #5587, opened 2024-01-29
fix(src): add textskeleton to loading state of mentions

ninaandal:
### Description

Added skeleton to loading state when mentioning users in comment. 
Screenshot 2024-01-29 at 15 06 22

The test for the hardcoded `@Loading` text when it is loading is no longer applicable, as this also created some issues with the current localisation work that is in progress for comments 

### What to review
The loading state when mentioning a user in a comment

### DIFF
--- a/packages/sanity/playwright-ct/tests/comments/CommentInput.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/comments/CommentInput.spec.tsx
@@ -30,6 +30,7 @@ test.describe('Comments', () => {
       await page.keyboard.press('Enter')
       await expect(page.getByTestId('comments-mentions-menu')).not.toBeVisible()
       // TODO: find a way to mock `useUser`!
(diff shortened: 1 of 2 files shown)


packages/sanity/playwright-ct/tests/comments/CommentInput.spec.tsx
```diff
@@ -30,7 +30,6 @@ test.describe('Comments', () => {
       await page.keyboard.press('Enter')
       await expect(page.getByTestId('comments-mentions-menu')).not.toBeVisible()
       // TODO: find a way to mock `useUser`!
-      await expect($editable).toHaveText('@Loading')
```
binoy14, line 33: ❓ could we add a `data-testid` to the `TextSkeleton` component and keep this expect but instead check for visibility of that component instead?

ninaandal: Yes, of course - will do that asap!

binoy14 (approved):

PR #5590, opened 2024-01-29
fix(validation): filter system documents when running validation

binoy14:
### Description

Fixes issue with logic of filtering out system documents when running `sanity documents validate` 

### What to review

To test it on the branch run `yarn build && cd dev/test-studio && npx sanity documents validate` and check if there are any system documents in there

### Notes for release

N/A


ricokahler (approved):
oops! thanks for this ❤️

PR #5591, opened 2024-01-29
Feat/migration yield transactions

bjoerge:
### Description
This adds support for yielding/returning "raw mutations", e.g. mutations on the same shape as mutation endpoint payloads. This means you can do this:

```ts
export default defineMigration({
  name: 'Make a copy of all matching documents',
  documentTypes: ['author'],
  migrate: {
    document(doc, ctx) {
      return {
        createIfNotExists: {
          _id: `${doc._id}-copy`,
          _type: 'boolean',
        },
      }
    },
  },
})
```

### Testing
tbd

### Notes for release
n/a


packages/@sanity/migrate/src/runner/normalizeMigrateDefinition.ts
```diff
@@ -58,30 +103,48 @@ async function collectDocumentMutations(
...
+    return [...arrify(nodeReturnValues), ...arrify(nodeTypeReturnValues)].map(
+      (change) => change && normalizeNodeMutation(path, change),
     )
   })
 
   return (await Promise.all([...arrify(await documentMutations), ...nodeMigrations]))
     .flat()
-    .map((change) => normalizeDocumentMutation(doc._id, change))
+    .flatMap((change) => (change ? normalizeDocumentMutation(doc._id, change) : []))
+    .flat()
```
binoy14, line 122: ❓ is the two `.flat()` correct? Could it just flatten it just at the end?

binoy14 (approved):
Looks good besides one non blocking comment

packages/@sanity/migrate/src/runner/normalizeMigrateDefinition.ts
bjoerge: You're absolutely right. Removed the last one. Thanks!

PR #5573, opened 2024-01-26
feat(cli): add pretty mutation formatting to migration runner

juice49:
### Description

This branch adds pretty formatting when running migrations (both in dry-run and commit mode). This uses the same tree style report that is used when reporting document validation results.

### What to review

- Pretty formatting looks good when running migrations.
- Validation output is still correct (the tree generator code has been abstracted in this branch).

### Testing

- Run some migrations in dry-run and commit mode.
- Run some document validation checks.

PR #5598, opened 2024-01-30
fix(migration): fixes issue with dry-run output

binoy14:
### Description

- Fixes the dry run command to show an done message. The progress indicator probably needs some love but there is #5573 to take care of it
- Also shows proper error message if the migration file has code issues instead of file not found error.

### What to review

The dry-run from export file hangs for me not sure if it's just me or widespread. Please confirm that. 

### Notes for release

- N/A

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
@@ -58,7 +58,9 @@ function resolveMigrationScript(workDir: string, migrationName: string) {
         // eslint-disable-next-line import/no-dynamic-require
         mod = require(absolutePath)
       } catch (err) {
-        // console.error(err)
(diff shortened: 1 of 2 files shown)


packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
```diff
@@ -58,14 +58,16 @@ function resolveMigrationScript(workDir: string, migrationName: string) {
...
+          throw new Error(`Error: ${err.message}"`)
+        }
       }
       return {relativePath, absolutePath, mod}
     }),
   )
 }
 
-const createMigrationCommand: CliCommandDefinition<CreateFlags> = {
+const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
```
bjoerge, line 70: good catch, ty!

bjoerge (approved):

PR #5608, opened 2024-01-30
fix(studio): center studio nav

RostiMelk:
### Description

Centered the navigation menu in Studio

Edit: After a discussion with @robinpyon, the navigation menu is now only centering on large screens (1200px and up).

#### Before:
image
#### After:
image

### What to review

Check that the navigation menu is centered independently from surrounding items

### Testing

- Verified that layout does not break when adding a bunch of menu items
- Verified that nav stays the same on breakpoints below 1200px

### Notes for release

- Centered navigation in Studio


binoy14 (approved):
From code perspective it looks okay, thanks for the PR Rosti!

sjelfull (approved):
Thanks!

PR #5610, opened 2024-01-30
feat(cli): add `intentUrl` to `json` and `ndjson` reporters

ricokahler:
### Description

- Adds `intentUrl` to the results of `sanity documents validate --format json` and `--format ndjson`
- Fixes an edge case where there is an unnecessary leading slash in the intentUrl
- Fixes the logic in `validateDocument.test.ts` that failed to catch the documents with `system.` types to show up bug.
- Fixes a bug where valid documents would be reported in `sanity documents validate --format json`

### What to review

Take a look at the logic and ensure nothing else changes.

### Testing

- I ran the build and tried out the command for myself.

### Notes for release

- Includes `intentUrl` in `json` and `ndjson` formats for `sanity documents validate`
- Fixes a bug where valid documents would be reported in `sanity documents validate --format json`

### DIFF
--- a/packages/sanity/src/_internal/cli/threads/__tests__/validateDocuments.test.ts
+++ b/packages/sanity/src/_internal/cli/threads/__tests__/validateDocuments.test.ts
@@ -59,11 +59,11 @@ const documents: SanityDocument[] = [
     },
   },
   {
-    _id: 'system.some-system-document.foo',
-    _rev: 'rev6',
+    _id: 'some-system-document.foo',
     _type: 'system.some-system-document',
     _createdAt: '2024-01-18T19:18:39.048Z',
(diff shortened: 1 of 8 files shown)


packages/sanity/src/_internal/cli/threads/__tests__/validateDocuments.test.ts
```diff
@@ -59,11 +59,11 @@ const documents: SanityDocument[] = [
     },
   },
   {
-    _id: 'system.some-system-document.foo',
-    _rev: 'rev6',
+    _id: 'some-system-document.foo',
```
ricokahler, line 62: this `_id` is why the test still passed lol

binoy14 (approved):

PR #5611, opened 2024-01-30
chore(migration): remove from-export flag from migration cli

binoy14:
### Description

Removes `from-export` flag from migration CLI


bjoerge (approved):
Thanks!

PR #5612, opened 2024-01-30
chore(migration): remove query from migration context

binoy14:
### Description

Removes `query` method from filter object in the migration context


bjoerge (approved):
Great, thanks!

PR #5617, opened 2024-01-30
fix(cli): add support for `import.meta.env`

ricokahler:
### Description

The following adds minimal support for `import.meta.env` inside of the mock browser environment used in the CLI.

### What to review

Does it work?

### Testing

I added an import.meta.env usage in the test-studio and ran the command.

### Notes for release

- Fixes a bug that caused uses of `import.meta.env` to break the mock browser environment used in the CLI.

### DIFF
--- a/packages/sanity/src/_internal/cli/util/mockBrowserEnvironment.ts
+++ b/packages/sanity/src/_internal/cli/util/mockBrowserEnvironment.ts
@@ -3,6 +3,7 @@ import jsdomGlobal from 'jsdom-global'
 import resolveFrom from 'resolve-from'
 import {register as registerESBuild} from 'esbuild-register/dist/node'
 import {ResizeObserver} from '@juggle/resize-observer'
+import {getStudioEnvironmentVariables} from '../server/getStudioEnvironmentVariables'
 
 const jsdomDefaultHtml = `<!doctype html>
 <html>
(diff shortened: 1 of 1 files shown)


ricokahler:
Eh may not be the best idea to release this without properly populating `process.env`/`import.meta.env`

rexxars (approved):
I think that should work, yeah?

packages/sanity/src/_internal/cli/util/mockBrowserEnvironment.ts
```diff
@@ -34,6 +35,12 @@ export function mockBrowserEnvironment(basePath: string): () => void {
     format: 'cjs',
     extensions: ['.js', '.jsx', '.ts', '.tsx', '.mjs'],
     jsx: 'automatic',
+    define: {
+      // define the `process.env` global
+      ...getStudioEnvironmentVariables({prefix: 'process.env.', jsonEncode: true}),
```
rexxars, line 40: Not actually sure if this one is needed, given we're already in an environment with `process.env` defined - maybe if the `.env` files and such is not already loaded?

binoy14: added as a safety mechanism, since `process.env` is not a browser thing right?

PR #5618, opened 2024-01-30
fix(migrate): cleanup buffer file after read

bjoerge:
### Description

When running a migration, we'll read the response from the export endpoint through a buffer file. This file should be cleaned up when it's no longer needed

### Testing

There's added tests asserting that the file is removed. Might need to tweak the timeout on those later.

### Notes for release

n/a


binoy14 (approved):

binoy14:
dependency check is failing

PR #5620, opened 2024-01-30
fix(migrate): scope node path to the current node

bjoerge:
### Description

Last one from me today 🤞🏼 

This makes sure any node patch returned from a visitor-style migration is relative to the node path itself.

### DIFF
--- a/packages/@sanity/migrate/src/mutations/asserters.ts
+++ b/packages/@sanity/migrate/src/mutations/asserters.ts
@@ -1,5 +1,6 @@
-import {Mutation} from './types'
+import {Mutation, NodePatch} from './types'
 import {Transaction} from './transaction'
+import {Operation} from './operations/types'
 
 export function isMutation(mutation: unknown): mutation is Mutation {
   return (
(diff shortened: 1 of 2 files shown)


packages/@sanity/migrate/src/mutations/asserters.ts
```diff
@@ -22,3 +23,32 @@ export function isTransaction(mutation: unknown): mutation is Transaction {
     mutation.type === 'transaction'
   )
 }
+
+export function isOperation(value: unknown): value is Operation {
+  return (
+    value !== null &&
+    typeof value === 'object' &&
```
binoy14, line 30: should we be worried about arrays in here?

binoy14 (approved):
non blocking question, looks good besides that

packages/@sanity/migrate/src/mutations/asserters.ts
bjoerge: Not at this point, but I'll look into narrowing down the `value` argument here.

PR #5626, opened 2024-01-31
chore(telemetry): track portable text editor errors.

pedrobonamin:
### Description

The intention of this PR is to track errors generated at portable text editor using our internal `Telemetry` package.

Errors will only be tracked in production environments to avoid extra noise.

Errors can occur in different scenarios:
- React lifecycle errors (render and hooks)
  -  React lifecycle errors in PortableTextInput, will be reported by a `ErrorBoundary` at top level of `PortableTextInput` 
- Event handlers.
  - **PTE package** a reusable `useCallbackWithTryCatch` hook was added into the PTE package. This hook has the same signature as `useCallback` but wraps the callback function in a try catch block and will throw an error with **cause** and **name**, to make it easy to detect in between other global errors that are reported by the `ErrorLogger`
  - **PortableTextInput**: the main event handler `handleEditorChange` is wrapped in a try catch, specific cases from that event handler `invalidValue` and `error` will also report an error.
- Global errors:
  - This will include the errors generated at transactions and any other type of error detected in the GlobalErrorHandler, they will be reported by the `ErrorLogger` as global errors, when a error coming from a event handler in PTE package is reported, it will contain the **cause** and **name** for easier tracking.

### What to review

Is there a better way to track errors using some other global approach?
Is the PTE functionality modified?


packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -1,5 +1,5 @@
-import {BaseRange, Transforms, Text, select, Editor} from 'slate'
-import React, {useCallback, useMemo, useEffect, forwardRef, useState, KeyboardEvent} from 'react'
+import {BaseRange, Transforms, Text, Editor} from 'slate'
```
pedrobonamin, line 1: In the following files the change is to replace `useCallback` for `useCallbackWithTryCatch` and add details about the function.

PR #5638, opened 2024-02-01
fix(migration): fixes issues where patch would happen twice

binoy14:


packages/@sanity/migrate/src/runner/utils/toSanityMutations.ts
```diff
@@ -21,7 +21,10 @@ export async function* toSanityMutations(
         }
         continue
       }
-      yield SanityEncoder.encode(mutation as any[]) as SanityMutation[]
```
binoy14, line 24: the problem here is that when a mutation is an array it is in second for loop which is looping over single mutation but it yields the full parent mutation which is the full array so this will do the mutations `mutations.length` times

PR #5640, opened 2024-02-01
feat(cli): prompt to use default plan when wrong coupon or plan

indrekkarner:
### Description

Prompts to use default plan when customer tries to use incorrect coupon or plan Id
To improve UX
https://linear.app/sanity/issue/ENTX-1211/support-old-plans-being-used-to-create-projects

### What to review

initSanityProject from CLI.

### Testing

Looking into it later, currently proposal
using npm create sanity -- --project-plan incorrect-plan-id

### Notes for release

TBD


indrekkarner:
image
image
image

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -166,10 +166,32 @@ export default async function initSanity(
       selectedPlan = await getPlanFromCoupon(apiClient, intendedCoupon)
       print(`Coupon "${intendedCoupon}" validated!\n`)
     } catch (err) {
-      throw new Error(`Unable to validate coupon code "${intendedCoupon}":\n\n${err.message}`)
+      const useDefaultPlan = await prompt.single({
```
juice49, line 169: Would it be a good idea to check whether the error is due to the coupon not existing, rather than some other failure (e.g. networking or backend)?

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -166,10 +166,32 @@ export default async function initSanity(
...
+      } else {
+        throw new Error(`Unable to validate coupon code "${intendedCoupon}":\n\n${err.message}`)
+      }
     }
   } else if (intendedPlan) {
-    selectedPlan = intendedPlan
+    try {
+      selectedPlan = await getPlanFromId(apiClient, intendedPlan)
+    } catch (err) {
+      const useDefaultPlan = await prompt.single({
```
juice49, line 184: Would it be a good idea to check whether the error is due to the plan not existing, rather than some other failure (e.g. networking or backend)?

juice49 (approved):
Looks good to me!

PR #5641, opened 2024-02-01
feat: apply text balance to document titles

SimeonGriggs:
### Description

Pretty-fies document titles

### What to review

CSS

### Testing

I have tested this by editing inline styles of an existing Studio to double-check the desired outcome of prettier titles.

### Notes for release

Balanced text in document titles


robinpyon (approved):
Thanks @SimeonGriggs!

PR #5644, opened 2024-02-01
fix(core): add icon display as default to documentTypeList

cngonzalez:
### Description

This PR removes the check to show icons from `S.documentTypeList()`. Like all other list types, `S.documentTypeList()` will now fall back to showing icons by default, regardless of preview parameters set. 

In the past, customers were sometimes surprised when they simply defined a `select` selection in a `preview` parameter of a schema type without a `media` parameter, like:
```js
export default defineType({
  name: 'myType',
  type: 'document',
  fields: [...],
  preview: {
    select: {
      title: 'title',
      subtitle: 'description'
    }
  }
})
```
which would change the list view to an icon-less one. However, _not_ setting a `media` parameter in a `prepare` return object still resulted in an icon being shown, which seems inconsistent.

Because an icon or media preview still shows up in all other kinds of desk lists, as well as in other parts of the studio (such as arrays of references), showing the icon here seems more visually consistent.

### What to review

- Schema types defined with a `select` that only include `title` and `description` should show an icon.
- setting `.showIcon(false)` on a `documentTypeList` should still omit an icon.

Old:
Screenshot 2024-02-02 at 10 46 04 AM

New:
Screenshot 2024-02-02 at 10 20 56 AM

### Testing

I do not currently see any test suites for anything around structure builder. I agree this should be tested! But perhaps it makes more sense to decide what tests should look like in another PR.

### DIFF
--- a/packages/sanity/src/structure/structureBuilder/documentTypeListItems.ts
+++ b/packages/sanity/src/structure/structureBuilder/documentTypeListItems.ts
@@ -12,11 +12,6 @@ import {List} from './List'
 import {Collection} from './StructureNodes'
 import {StructureContext} from './types'
 
-function shouldShowIcon(schemaType: SchemaType): boolean {
-  const preview = schemaType.preview
-  return Boolean(preview && (preview.prepare || (preview.select && preview.select.media)))
-}
-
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
Nice, thank you @cngonzalez!

PR #5645, opened 2024-02-02
fix(migration): list file-based migrations in list command

rexxars:
### Description

When running `sanity migration list`, it would not include migrations defined as `migrations/some-migration.ts` and similar, only those in directories. This PR fixes that.

Also snuck in some code cleanup:
- Clean up after esbuild-register once done using it
- Sorting imports and using type-only imports where possible
- Use `chalk` from context where possible (I see there are other ocurences in other commands, but lets address later)
- Aligned and reused the logic for finding defined migrations, use them across list and run commands
- Moved the `utils.ts` file into the `utils` folder and gave it a descriptive name

### What to review

- `sanity migration list` works both with `migrations/someFile.ts` and `migrations/someMigration/index.ts`
- `sanity migration run <id>` works both with `migrations/someFile.ts` and `migrations/someMigration/index.ts`

### Notes for release

- Fixed an issue where migrations not placed in a subfolder would not appear in the output of `sanity migration list`

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
@@ -1,5 +1,5 @@
-import path from 'path'
-import fs from 'node:fs/promises'
+import path from 'node:path'
+import {writeFile} from 'node:fs/promises'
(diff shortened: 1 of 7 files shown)


binoy14 (approved):
Thank you!

PR #5646, opened 2024-02-02
fix(perf): fixes build and deploy issues with perf studio

binoy14:
### Description

Turbo `1.12.0` changes the filter functionality to not go inside child folders if the `package.json` is found this made it so that the studio inside the `perf/` folder would not deploy and get correctly detected by turbo. The change moves all the actual tests into `perf/tests` folder and keeps the `perf/studio` folder so both of them work correctly 

### What to review

- Perf tests still work, I checked the commands and they seem to be working. 
- If the `perf/tests` folder makes sense

### Testing

N/A

### Notes for release

N/A (internal)


bjoerge (approved):
Thank you for sorting this out, @binoy14 🙇🏼 

I pushed some smaller edits in [a43f711](https://github.com/sanity-io/sanity/pull/5646/commits/a43f7113b4ef890f7d47511cbb5e6e7bdc7dc820), moving the readme to `./perf` and smaller clarifications about where the .env file should go

PR #5653, opened 2024-02-02
fix(core): add aria label to action button pte toolbar

ninaandal:
### Description
Adds `aria-label` to the PTE toolbar action menus


binoy14 (commented):
Looks like there are failing tests, this is very likely due the fact that the failing tests are not wrapped in `LocaleProvider` and cannot find the correct locale to display the text. Let me know if you need help in fixing that.

binoy14 (approved):
LGTM, thanks!

PR #5654, opened 2024-02-02
refactor(util): move concurrency limiter packages to `@sanity/util`

juice49:
### Description

This change moves `createClientConcurrencyLimiter` and `ConcurrencyLimiter` to the `@sanity/util` package so that they can be reused. At the moment, they are duplicated for the `sanity` and `@sanity/migrate` packages.

### What to review

Does the approach of moving `createClientConcurrencyLimiter` to `@sanity/util/client` and `ConcurrencyLimiter` to `@sanity/util` make sense?

### Testing

Do the `documents validate` and `migration run` commands run as expected? These are the two code paths affected by this change.


packages/sanity/src/core/validation/util/createBatchedGetDocumentExists.ts
```diff
@@ -11,8 +11,8 @@ import {
   switchMap,
   finalize,
 } from 'rxjs'
+import {ConcurrencyLimiter} from '@sanity/util/ConcurrencyLimiter'
```
binoy14, line 14: this feels weird, is there a reason it can't be exported from root of the util? `import { ConcurrencyLimiter} from '@sanity/util'`

binoy14 (commented):
have a clarification question

packages/sanity/src/core/validation/util/createBatchedGetDocumentExists.ts
juice49: It's a good question. I've tried to follow the existing conventions in `@sanity/util`. For example, [`createSafeJsonParser` and `legacyDateFormat` are both individual exports](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/util/package.json#L58-L79). There are [no root exports](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/util/exports/index.ts).

binoy14: interesting, maybe double check with test of the team make sure this is the direction we want to go with?

juice49: Good point! I've renamed the package to `@sanity/util/concurrency-limiter`.

binoy14 (approved):

PR #5659, opened 2024-02-02
fix(cli): respect unattended mode

juice49:
### Description

A recent change to the CLI added user prompts when attempting to use a nonexistent coupon (`--coupon`) or a plan (`--project-plan`), however, this change did no respect unattended mode (`-y, --yes`) and would prompt the user regardless. 

### What to review

- Initialising a project using a nonexistent coupon or plan in unattended mode:
  - `init -y --project-plan beep`
  - `init -y --coupon boop`

### Testing

- Does the command proceed without prompting the user?

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -174,6 +174,9 @@ export default async function initSanity(
             message: `Coupon "${intendedCoupon}" is not available, use default plan instead?`,
             default: true,
           }))
+        if (unattended) {
+          output.warn(`Coupon "${intendedCoupon}" is not available - using default plan`)
+        }
         trace.log({
           step: 'useDefaultPlanCoupon',
           selectedOption: useDefaultPlan ? 'yes' : 'no',
@@ -200,6 +203,9 @@ export default async function initSanity(
             message: `Project plan "${intendedPlan}" does not exist, use default plan instead?`,
             default: true,
           }))
+        if (unattended) {
+          output.warn(`Project plan "${intendedPlan}" does not exist - using default plan`)
+        }
         trace.log({
           step: 'useDefaultPlanId',
           selectedOption: useDefaultPlan ? 'yes' : 'no',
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -167,11 +167,13 @@ export default async function initSanity(
...
-          message: `Coupon "${intendedCoupon}" is not available, use default plan instead?`,
-          default: true,
-        })
+        const useDefaultPlan =
+          unattended ??
+          (await prompt.single({
+            type: 'confirm',
+            message: `Coupon "${intendedCoupon}" is not available, use default plan instead?`,
+            default: true,
+          }))
```
rexxars, line 176: Can we print a warning if unattended?
```suggestion
          }))
          
          if (unattended) {
            output.warn(`Coupon "${intendedCoupon}" is not available - using default plan`)
          }
```

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -191,11 +193,13 @@ export default async function initSanity(
...
-          message: `Project plan "${intendedPlan}" does not exist, use default plan instead?`,
-          default: true,
-        })
+        const useDefaultPlan =
+          unattended ??
+          (await prompt.single({
+            type: 'confirm',
+            message: `Project plan "${intendedPlan}" does not exist, use default plan instead?`,
+            default: true,
+          }))
```
rexxars, line 202:
```suggestion
          }))
          
          if (unattended) {
            output.warn(`Project plan "${intendedPlan}" does not exist - using default plan`)
          }
```

rexxars (changes requested):

packages/@sanity/cli/src/actions/init-project/initProject.ts
juice49: Yep! Thank you. I've added and tested the suggested warnings.

rexxars (approved):

PR #5670, opened 2024-02-05
feat(tasks): feature bootstrap, add initial setup

pedrobonamin:
### Description
Introduces initial setup for `Tasks` feature.

Tasks will be part of `core`
All the code related will live in `packages/sanity/src/core/studio/tasks`
A new config flag is added to Workspaces to enable / disable the feature

```
  unstable_tasks: {
    enabled: boolean
}
```
⚠️ this is **only** an initial setup with the base configuration to get us started, is not intended to be a final production ready integration.
⚠️ It doesn't include support for mobile yet. 
⚠️ It is not translated yet, as it will be a changing feature with no final designs.

Screenshot 2024-02-02 at 13 52 43

Adds: 
- New provider `TasksProvider` is added to the `StudioProvider`
- Studio Sidebar is added to `StudioLayout`. 
- TasksButton is added to the `StudioNavbar`

### What to review

No changes should be introduced outside of the context of `tasks`.
Studio config is unchanged for users.

### Notes for release

-


packages/sanity/src/core/studio/tasks/context/tasks/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface TasksContextValue {
```
hermanwikner, line 1: I think it would be good if these were a bit more UI agnostic. Even though the current implementation is a sidebar, I believe keeping it more generic with `open` and `setOpen` would be good, as we might want to show the tasks in, for example, a dialog in the future. 

```ts

const {open, setOpen} = useTasks()

```

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -0,0 +1,36 @@
+import {Card} from '@sanity/ui'
+import styled from 'styled-components'
+import {AnimatePresence, motion} from 'framer-motion'
+import {useTasksEnabled, useTasks} from '../../context'
+import {TasksSidebarHeader} from './TasksSidebarHeader'
+
+const SidebarRoot = styled(Card)`
+  height: 100%;
+  width: 360px;
+  margin-left: 4px;
```
hermanwikner, line 10: Can we use the `marginLeft` prop instead – so that we rely on theme values?

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -0,0 +1,36 @@
+import {Card} from '@sanity/ui'
+import styled from 'styled-components'
+import {AnimatePresence, motion} from 'framer-motion'
+import {useTasksEnabled, useTasks} from '../../context'
+import {TasksSidebarHeader} from './TasksSidebarHeader'
+
+const SidebarRoot = styled(Card)`
+  height: 100%;
```
hermanwikner, line 8: Since we use `height="fill"` – is this needed?

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -0,0 +1,36 @@
...
+export function TasksStudioSidebar() {
+  const {enabled} = useTasksEnabled()
+  const {isSidebarOpen} = useTasks()
+
+  if (!enabled) return null
+  return (
+    <AnimatePresence initial={false}>
+      {isSidebarOpen && (
+        <motion.div
+          initial={{translateX: 16, opacity: 0}}
```
hermanwikner, line 25: Since these are static values – can we move them outside the component scope? 

```tsx
import {Variants, Transition} from 'framer-motion'

const VARIANTS: Variants = {
  hidden: {opacity: 0, x: 16},
  visible: {opacity: 1, x: 0},
}

const TRANSITION: Transition = {duration: 0.2}

return (
  <motion.div variants={VARIANTS} transition={TRANSITION} initial="hidden" animate="visible">
    ...
  </motion.div>
)
```

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
pedrobonamin: Absolutely! Good catch.

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -0,0 +1,36 @@
...
+import {AnimatePresence, motion} from 'framer-motion'
+import {useTasksEnabled, useTasks} from '../../context'
+import {TasksSidebarHeader} from './TasksSidebarHeader'
+
+const SidebarRoot = styled(Card)`
+  height: 100%;
+  width: 360px;
+  margin-left: 4px;
+  box-shadow:
+    0px 6px 8px -4px rgba(134, 144, 160, 0.2),
```
hermanwikner, line 12: Do we want to use different `rgba` shadow values depending on the current scheme?

pedrobonamin: I didn't wanted to get into that yet, as the designs are not final and this shadow is just a custom quick implementation copying what we have in the figma file. But for sure let's revisit when designs are final.

packages/sanity/src/core/studio/tasks/components/TasksNavbarButton/TasksNavbarButton.tsx
```diff
@@ -0,0 +1,19 @@
+import {PanelRightIcon} from '@sanity/icons'
+import {Button} from '../../../../../ui-components'
+import {useTasksEnabled, useTasks} from '../../context'
+
+export function TasksNavbarButton() {
```
hermanwikner, line 5: Since this is a navbar specific component – should it be moved to the `/studio/components/navbar` folder?

packages/sanity/src/core/studio/tasks/context/enabled/TasksEnabledProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface TaksEnabledProviderProps {
+  children: React.ReactNode
+}
+
+export function TasksEnabledProvider({children}: TaksEnabledProviderProps) {
+  // TODO: Restore this once the feature flag is enabled by the ENTX team, see ENTX-1330.
+  const {enabled: featureEnabled, isLoading} = useFeatureEnabled('studioTasks')
+  const workspace = useWorkspace()
+  const enabled = !!workspace.tasks?.enabled
```
hermanwikner, line 15: It is currently not possible to enable/disable tasks from a plugin config. I know this is not relevant for now – but maybe it is a good idea to make sure that it works already now so that we have that logic in place?

sjelfull (approved):
Approved, pending that we enable it only in the staging workspace.

packages/sanity/src/core/studio/tasks/context/enabled/TasksEnabledProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {TasksEnabledContext} from './TasksEnabledContext'
+import {TasksEnabledContextValue} from './types'
+
+interface TaksEnabledProviderProps {
+  children: React.ReactNode
+}
+
+export function TasksEnabledProvider({children}: TaksEnabledProviderProps) {
+  // TODO: Restore this once the feature flag is enabled by the ENTX team, see ENTX-1330.
+  const {enabled: featureEnabled, isLoading} = useFeatureEnabled('studioTasks')
```
sjelfull, line 13:
```suggestion
// const {enabled: featureEnabled, isLoading} = useFeatureEnabled('studioTasks')
```

packages/sanity/src/core/studio/tasks/components/TasksNavbarButton/TasksNavbarButton.tsx
pedrobonamin: 🤔 let's discuss it tomorrow in the daily, I do see some benefits on keeping all the tasks code in the same route but also theres a good point on keeping this into the navbar folder as the rest of the navbar buttons are there.

packages/sanity/src/core/studio/tasks/context/enabled/TasksEnabledProvider.tsx
pedrobonamin: How will this need to be configured? Do you have any example?

pedrobonamin:
Will close this one to avoid confusions, as we will continue with the plugin approach. Thanks all for the reviews and support. 
The branch will be kept in case we want to restore this approach

pedrobonamin:
Will close this one to avoid confusions, as we will continue with the plugin approach. Thanks all for the reviews and support.
The branch will be kept in case we want to restore this approach

PR #5672, opened 2024-02-05
feat(migration): (re)introduce --from-export

bjoerge:
### Description
This brings back support for the `--from-export` when running `sanity migration run`

### What to review
Verify that the issue blocking this from going out last week is fixed

### Testing
- Tests for the original issue is added in 100d837e9d5ee84993b1fd8ac09746175f23cf97
- The fix is adressed separately in 6e5cd9eb6f1c7c2d99d4cbbfd15ee68e66eac415
- Adding back support for `--from-export` implemented in 8c97961cf5100835895f02708d2f8db5157d628e

### Notes for release
- Adds support for dry-running a migration using a local dataset export file instead of the Sanity API


juice49 (approved):
Looks good to me! This change resolved the issue in my testing.

PR #5675, opened 2024-02-05
fix(migrate): submit mutations with visibility=async

bjoerge:
### Description
Sets visibility to `async` when submitting mutations from `sanity migraiton run`.

### What to review
- Any considerations I'm missing? I based it on the visibility used by `sanity import`

### Notes for release

n/a


rexxars (approved):
👍 Great!

PR #5678, opened 2024-02-05
fix(migration): fixes an issue where migration command would error sometimes

binoy14:
### Description

Removes `mkdirp` package and uses fs to create the directory.

Also removes `mkdirp` and related types from package.json. Search does not yield any other usage.

### What to review

migration create command still works as expected

### Notes for release

- Fixes an issue where migration command would error in some instances


bjoerge (approved):
Nice find, thanks!

PR #5681, opened 2024-02-06
test(core): adds e2e tests for date and datetime inputs

binoy14:
### Description

~Fixes an issue where date and datetime inputs does not run validation when typing in the input field, it would only run validation on blur. The change replaces the `LazyInput` with text so the validation runs on all type events.~ 

Adds tests for date and datetime input

### What to review

- ~Are there any side effects of the change that I am not aware of?~
- ~Do e2e tests makes sense? Anything more to test?~

### Testing

Added

### Notes for release

N/A


binoy14:
* `next`
  * **PR #5681** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5681?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sanity-io/sanity/5681?utm_source=stack-comment).

packages/sanity/src/core/form/inputs/DateInputs/__tests__/CommonDateTimeInput.test.tsx
binoy14, line 1: This test was missed part of this [PR](https://github.com/sanity-io/sanity/pull/5551) updated it to match the implementation

juice49:
I noticed two side effects caused by this change. The first is that validation now occurs while the user types, which means the field appear in an invalid state while the user types. This isn't necessarily a problem.

### Before

https://github.com/sanity-io/sanity/assets/1454914/e3b4cd63-3c75-43d9-b37a-9bdefc331a9f

### After

https://github.com/sanity-io/sanity/assets/1454914/f9deba8e-fb16-4f9e-af6d-66a928155737

The second problem seems more severe. It isn't very clear in the video, but after the change, the caret sometimes jumps to the end of the input on its own, and it's sometimes impossible to type into the field at all.

### Before

https://github.com/sanity-io/sanity/assets/1454914/19d9a854-389c-4598-860d-c8b07be11c53

### After

https://github.com/sanity-io/sanity/assets/1454914/89ef8032-6293-4f40-bf81-8bb1210c9951

packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
```diff
@@ -71,7 +70,8 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
   )
 
   return (
-    <LazyTextInput
+    <TextInput
```
juice49, line 73: There seem to be some side effects cause by this. I've added more details in the thread.

binoy14:
Reverted the fix and this PR just focusing on testing, there are two tests that are skipped, when the solution is implemented they should be unskipped and made sure all tests pass

juice49 (approved):

dev/test-studio/schema/debug/validation.tsx
```diff
@@ -95,11 +95,18 @@ export default {
       validation: (Rule) => Rule.uri({allowRelative: true}),
     },
     {
-      name: 'date',
+      name: 'datetime',
       type: 'datetime',
+      title: 'Some datetime',
+      description: 'ISO-formatted datetime, inferred, must be in 2017',
+      validation: (Rule) => Rule.min('2017-01-01 00:00:00').max('2017-12-31 00:00:00'),
```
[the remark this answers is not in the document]
binoy14: oh good point, milliseconds time allowed only :D

PR #5684, opened 2024-02-06
feat(cli): ensure CWD is project root when running migration commands

juice49:
### Description

This branch adds an `ensureCwdIsProjectRoot` function that can be used by CLI commands to improve UX when the Sanity CLI is mistakenly used outside of a Studio project root.

If no CLI configuration can be resolved, the CLI will now search parent directories for any Studio or CLI configuration file. If It finds one, it will infer that it is the project root directory and print a notice that the user should move there and try the command again.

This change doesn't affect the way the CLI configuration is resolved; it's only used to print a notice.

I've initially added this to the `migration` commands (`create`, `list`, and `run`).

Screenshot 2024-02-07 at 14 05 41

### What to review

- That the output printed makes sense.
- That the change doesn't introduce unnecessary overhead (it shouldn't; we only search for config files if one could not already be resolved in the CWD).

### Testing

Run the `migration` commands (`create`, `list`, and `run`) from a subdirectory of a Studio project. For example, the test Studio migrations directory (`dev/test-studio/migrations`). The CLI can be invoked here by running `../../.bin/sanity migration list` etc.

### DIFF
--- a/packages/@sanity/cli/src/cli.ts
+++ b/packages/@sanity/cli/src/cli.ts
@@ -23,6 +23,7 @@ import {parseArguments} from './util/parseArguments'
 import {resolveRootDir} from './util/resolveRootDir'
 import {telemetryDisclosure} from './util/telemetryDisclosure'
 import {runUpdateCheck} from './util/updateNotifier'
+import {getProjectRoot} from './util/getProjectRoot'
 
 const sanityEnv = process.env.SANITY_INTERNAL_ENV || 'production' // eslint-disable-line no-process-env
 const knownEnvs = ['development', 'staging', 'production']
@@ -98,6 +99,7 @@ export async function runCli(cliRoot: string, {cliVersion}: {cliVersion: string}
     cliRoot: cliRoot,
     workDir: workDir,
     corePath: await getCoreModulePath(workDir, cliConfig),
+    projectRootPath: cliConfig ? workDir : await getProjectRoot(),
     cliConfig,
     telemetry,
   }
--- a/packages/@sanity/cli/src/cli.ts
+++ b/packages/@sanity/cli/src/cli.ts
@@ -16,6 +16,7 @@ import {type CommandRunnerOptions, type TelemetryUserProperties} from './types'
 import {createTelemetryStore} from './util/createTelemetryStore'
 import {detectRuntime} from './util/detectRuntime'
 import {type CliConfigResult, getCliConfig} from './util/getCliConfig'
+import {getProjectRoot} from './util/getProjectRoot'
 import {loadEnv} from './util/loadEnv'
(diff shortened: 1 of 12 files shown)


packages/@sanity/cli/src/cli.ts
```diff
@@ -97,6 +98,7 @@ export async function runCli(cliRoot: string, {cliVersion}: {cliVersion: string}
     cliRoot: cliRoot,
     workDir: workDir,
     corePath: await getCoreModulePath(workDir, cliConfig),
+    projectRootPath: cliConfig ? workDir : await getProjectRoot(),
```
bjoerge, line 101: a potential (although probably quite small) perf hit by doing these in sequence, so you could consider moving them up and do something like:
```ts
const [corePath, projectRootPath] = Promise.all([
  getCoreModulePath(workDir, cliConfig),
  getProjectRoot()
])
```

bjoerge (approved):

packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
```diff
@@ -44,6 +45,13 @@ const createMigrationCommand: CliCommandDefinition<CreateMigrationFlags> = {
 
     let [title] = args.argsWithoutOptions
 
+    const {cwdIsProjectRoot, printCwdProjectRootWarning} = ensureCwdIsProjectRoot(context)
```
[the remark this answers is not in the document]
juice49: Thanks! More than a little inspired by React hooks, I think!

packages/@sanity/cli/src/cli.ts
juice49: Aha! Good catch. I'll see if this can be easily refactored.

rexxars:
🤔  I thought this was already being done here: https://github.com/sanity-io/sanity/blob/f9c67eac6c7911901159c9317619b3be702ce3f5/packages/@sanity/cli/src/cli.ts#L50-L56

(Although it doesn't warn you)

juice49:
@rexxars I stumbled upon this code during my work, and was a little confused because it doesn't seem to do anything. From the looks of the code, the CLI _should_ infer the project root automatically. [It should also warn the user this has occurred](https://github.com/sanity-io/sanity/blob/f9c67eac6c7911901159c9317619b3be702ce3f5/packages/@sanity/cli/src/cli.ts#L255-L261).

In my testing, neither of these things happened. Only now have I realised _why_: the resolution code is looking for `studio.config.{ts,js}` files, rather than `sanity.config.{ts,js}`. Therefore, it never finds the project root.

It's probably best to fix this behaviour and scrap my other changes.

PR #5685, opened 2024-02-06
feat(migrate): use autoGenerateKeys=true when submitting mutations

bjoerge:
### Description

Sets `autoGenerateArrayKeys` to `true` when submitting mutations from the `sanity migraiton run` CLI command.

### Testing

Did a quick round of manual testing with a Studio having a document open while a mutation ran that added non-keyed array items to an array field in it, and didn't see any issues.

### Notes for release
- Adding objects without unique `_key`s to an array during a content migration will now automatically generate unique `_key`s


rexxars (approved):
:chefs-kiss:

PR #5689, opened 2024-02-07
feat(import): add --allow-system-documents to dataset import

j33ty:
### Description

As part of the import process through CLI and `@sanity/import` package, we prefer that users do not send in system documents. These documents are considering to be an internal detail and if imported, their interaction with existing system documents is undefined.

This PR attempts at adding a new flag `--allow-system-documents` to `sanity dataset import` CLI and an option `allowSystemDocument` to `@sanity/import` package, to explicitly permit system document import if the user knows what they are doing.

This can be considered to be a breaking change since it by default skips system documents from import, but I highly doubt anyone is importing system documents from CLI. And, if they are, they should stop doing that. Ideally, I should add data to my gut-feeling and figure out how many users are importing system documents directly from CLI but it's not trivial to get these numbers. Let me know if this needs to be done.

Background for change is present in #prj-dataset-backups.

### Bikeshedding

Two appropriate names for the flag are `--import-system-documents` or `--allow-system-documents`. I have weak preference for the latter since it aligns with existing options on the same subcommand.

### Testing
Two unit tests are added where import with and without this flag is tested.

### Notes for Release
It should okay to add commit message as it is to release notes.

### DIFF
--- a/packages/@sanity/import/README.md
+++ b/packages/@sanity/import/README.md
@@ -85,6 +85,16 @@ const options = {
    * Optional, defaults to `false`.
    */
   skipCrossDatasetReferences: false,
+
+  /**
+   * Whether or not to import system documents (like permissions and custom retention). This
+   * is usually not necessary, and may cause conflicts if the target dataset
+   * already contains these documents. On a new dataset, it is recommended that roles are re-created
+   * manually, and that any custom retention policies are re-created manually.
+   *
+   * Optional, defaults to `false`.
+   */
+  allowSystemDocuments: false,
 }
 
 sanityImport(input, options)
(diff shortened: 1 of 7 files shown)


rneatherway (approved):
I'd probably also have been happy with just filtering them out and not offering an option to allow it at all.

j33ty:
> I'd probably also have been happy with just filtering them out and not offering an option to allow it at all.

IMO taking away a functionality without a fallback option isn't a good experience. Enterprise folks with custom group documents may want to import system documents and with a flag, they can continue importing them.

binoy14 (approved):
makes sense to me, thanks!

PR #5690, opened 2024-02-07
feat(tasks): add tasks data loading

sjelfull:
### Description

Work-in-progress PR.

This adds a setup that uses the same approach as comments to load tasks. It includes a provider that gets the dataset, or creates it when the first task is created.

It also has a temporary UI that will be replaced with the UI components being created on the back of #5686.

### Notes for release

n/a


packages/sanity/src/core/studio/StudioProvider.tsx
```diff
@@ -59,7 +59,9 @@ export function StudioProvider({
       <StudioTelemetryProvider config={config}>
         <LocaleProvider>
           <ResourceCacheProvider>
-            <TasksProvider>{children}</TasksProvider>
+            <TasksSetupProvider>
```
pedrobonamin, line 62: What do you think about renaming this to something reusable with comments? I heard and interesting name when the name for the datasets was in discussion. 
CollaborativeSetupProvider?

sjelfull: I have something like that in mind yes. Just wondering what to name it, since it is "workflow" metadata related, and you can argue that the whole studio is Collaborative.

We can put it to the slack channel?

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -29,9 +36,98 @@ export function TasksStudioSidebar() {
...
           </SidebarRoot>
         </motion.div>
       )}
     </AnimatePresence>
   )
 }
+
+function TasksList({items}: {items: TaskDocument[]}) {
+  const currentUser = useCurrentUser()
+  const tabs = useMemo(() => {
```
pedrobonamin, line 49: Maybe this tabs could me moved outisde of the component to avoid the useMemo since it will be stable? Wdyt?

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -29,9 +36,98 @@ export function TasksStudioSidebar() {
...
+              aria-controls={`${tab.id}-panel`}
+              id={`${tab.id}-tab`}
+              label={tab.label}
+              onClick={() => setId(tab.id)}
+              selected={id === tab.id}
+            />
+          ))}
+        </TabList>
+      </Card>
+      <List items={filteredList ?? []} />
```
pedrobonamin, line 101: Do we need this check for `?? []` ? I think it will always be defined, and at least an empty array

packages/sanity/src/core/studio/tasks/components/list/TasksListItem.tsx
```diff
@@ -0,0 +1,9 @@
+import {ThreadCard} from './styles'
+
+interface TasksListItemProps {
+  children: React.ReactNode
+}
+
+export function TasksListItem({children}: TasksListItemProps) {
+  return <ThreadCard tone={undefined}>{children}</ThreadCard>
```
pedrobonamin, line 8:
```suggestion
  return <ThreadCard >{children}</ThreadCard>
```

packages/sanity/src/core/studio/tasks/components/list/styles.ts
```diff
@@ -0,0 +1,11 @@
+import {Card, CardProps} from '@sanity/ui'
```
pedrobonamin, line 1: Just curiosity, why are we moving this to our own implementation of styled component if it's not changing any style yet.

packages/sanity/src/core/studio/tasks/components/TasksSidebar/TasksSidebar.tsx
```diff
@@ -29,9 +36,98 @@ export function TasksStudioSidebar() {
...
+                <Text size={1} weight="bold">
+                  {item.title}
+                </Text>
+                <Flex align="center" gap={1}>
+                  <CalendarIcon />
+                  <Text size={1} muted>
+                    {item.dueBy ? dateFormatter.format(new Date(item.dueBy)) : 'No due date'}
+                  </Text>
+                </Flex>
+              </Stack>
```
pedrobonamin, line 127: Could we move this into the `TaskListItem` and let that component handle all the ui for it?

packages/sanity/src/core/studio/tasks/store/useTasksStore.ts
```diff
@@ -0,0 +1,159 @@
+import {useMemo, useEffect, useCallback, useReducer, useState} from 'react'
```
pedrobonamin, line 1: 🍾 🍾

packages/sanity/src/core/studio/StudioProvider.tsx
pedrobonamin: Sure! Let's bring it to slack

pedrobonamin (commented):
This is all awesome, thanks a lot @sjelfull 
Loved the attention to details int the setup with all the types baked for the next steps, and the store working.

packages/sanity/src/core/studio/tasks/context/setup/TasksSetupProvider.tsx
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        // 3. Create a client for the addon dataset and set it in the context value
+        //    so that the consumers can use it to execute task operations and set up
+        //    the real time listener for the addon dataset.
+        const client = handleCreateClient(datasetName)
+        setAddonDatasetClient(client)
+
+        // 4. Create the task
+        await client.create(task)
```
pedrobonamin, line 100: What do you think if instead of creating the task here, we return only the client. 
Once we have the client, the function triggering this, can wait for this to be resolved and then creates the task. In this way I think we should be able to share the Setup logic with the comments.

pedrobonamin (approved):
This is all great @sjelfull thank you! 
Will merge to the bootstrap branch and we can keep iterating over there. 

💪🏻

packages/sanity/src/core/studio/tasks/context/setup/TasksSetupProvider.tsx
sjelfull: Yeah, I think there are things including this that we need to change to align the different approaches into one way. Will think about how we can do that next.

PR #5709, opened 2024-02-11
fix(structure): provide better error handling if orderings contain invalid field

jtpetty:
### Description

Fixes SDX-847

Previously, if an invalid field name was given for an ordering for a schema, the user would be presented with a GROQ syntax error. This change adds a check in the event of an error to look for this situation to provide a more targeted error message.

Before:

image

After:

image

### What to review

I am not sure if there is a better place to put this check. Since this is part of schema setup, this error should be rare, so I did not want to add this check in for every request. This is why I put it in the catch block. There may be other scenarios we want to check for here as well.

### Testing

I did manual testing. I wasn't sure how to contrive this edge case in our test bench.


binoy14 (changes requested):
I don't think this fully solves the problem, from my testing if you use the following ordering 

```
 {
      title: 'Test',
      name: 'test',
      by: [
        {
          field: 'nonexistent',
          direction: 'desc',
        },
      ],
    },
```

than it will not error and make a query `| order(nonexistent desc)` From what I can tell this error only happens when you more than 2 fields to sort by and it fails due to running into `, ,` in this [line of code](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/search/weighted/createSearchQuery.ts#L205) in groq you can query for fields non-existent and it will not error so that catch block might show this message for things that are not necessarily related to sort order of non existent fields

jtpetty:
Thanks @binoy14! We do verify there is a missing field before displaying this particular message, but I will test it with the scenario you provided and see if I can get it to fail.

jtpetty:
@binoy14 - in that line you reference, that function does the same logic (it's where I got it from) and it strips out unknown fields and prints a warning message to the console (not great). I tried to make it throw an error but it ended up crashing the whole page.

packages/sanity/src/structure/panes/documentList/listenSearchQuery.ts
```diff
@@ -90,7 +91,11 @@ export function listenSearchQuery(options: ListenQueryOptions): Observable<Sanit
         mergeMap((typeNames: string[]) => {
           const types = getSearchTypesWithMaxDepth(
             getSearchableTypes(schema).filter((type) => {
-              return typeNames.includes(type.name)
+              if (typeNames.includes(type.name)) {
+                getExtendedProjection(type, sort.by, true)
```
jtpetty, line 95: I should add a comment here to explain why we are doing this

binoy14 (approved):
Tested with different scenarios and not running into same issues as before. LGTM, thanks!

PR #5712, opened 2024-02-12
fix(cli): project root resolution

juice49:
### Description

The CLI is intended to automatically resolve the project root if it's used inside the subdirectory of a Studio project. This behaviour has been broken for v3, because the resolution code mistakenly searches for `studio.config.{ts,js}` files, rather than `sanity.config.{ts,js}`. Therefore, it never finds the project root.

This was discovered and discussed in #5684.

### What to review

- That it's possible to use the CLI inside the subdirectory of a Studio project.
- That when doing so, a warning is printed: "Not in project directory, assuming context of project at ${path}".

### Testing

- I've added a basic CLI unit test to ensure the `help` command can be executed inside Studio project subdirectories, and that the warning is printed.

### Notes for release

Fixed a bug that prevented the CLI being used inside Studio project subdirectories.

### DIFF
--- a/packages/@sanity/cli/src/util/resolveRootDir.ts
+++ b/packages/@sanity/cli/src/util/resolveRootDir.ts
@@ -17,8 +17,8 @@ export function resolveRootDir(cwd: string): string {
 
 function hasStudioConfig(basePath: string): boolean {
   const buildConfigs = [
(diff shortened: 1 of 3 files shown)


rexxars (approved):
Thank you for fixing!

PR #5715, opened 2024-02-12
fix(cli) allow cli tests to run with a robot token

bjoerge:
### Description
The CLI test suite have been running with a session token for a test user with a registered email. This token has a short expiry and causes the test suite to suddenly start failing when the token expires. Using a robot token works, but one of the assertions for `sanity debug` output fails because the email of the roboto token is `null`. This fixes the issue by allowing `null` as a valid email, which should not make it matter whether the tests are run using a session token or a robot token.

### What to review
Does it make sense?

### Testing
Self evident

### Notes for release

n/a internal


cngonzalez (approved):
Seems straightforward! LGTM :)

PR #5716, opened 2024-02-12
feat(tasks): support document references and contextual tasks

sjelfull:
### Description

This updates the tasks list to support document references and the contextual This document tab that lists tasks targeting that document. It also has initial support for document previews.

Note that document previews doesn't currently refesh if you have a draft, and then publishes the referenced document.


pedrobonamin (approved):
This is great, thanks @sjelfull !

PR #5719, opened 2024-02-12
fix(structure): use useDocumentTitle hook for toast message

ninaandal:
### Description
The toast message for when documents are published or deleted did not use the `useDocumentTitle()` hook. This PR fixes it so that the the toast message uses the same logic that the large title of the document uses. 

### What to review
The toast message that appears when a document is published/deleted. 

### Notes for release
Fixes issue where the toast message always shows the correct title


juice49 (approved):
Looks great to me!

PR #5721, opened 2024-02-13
refactor(comments): use journey endpoint for comments upsell

rexxars:
### Description

Transitions from using a direct GROQ query towards a given project ID to using the dedicated `/journey/comments` endpoint. This allow us to change the shape on the backend more easily while maintaining backwards compatibility, as well as allowing for more granular cache control and similar.

### What to review

- Check and response still works (currently only enabled in staging)

### Notes for release

None.


pedrobonamin (approved):
Thanks @rexxars looks good to me.

PR #5722, opened 2024-02-13
build: release automation

ricokahler:
### Description

Introduces the new release automation that publishes from an action.

### What to review

- Review the release.yml workflow and try to come up with scenarios that break it.
- Are all the pre-flight check good?

### Testing

We tried this out in another repo and copied and pasted the results.

### Notes for release

N/A


juice49 (approved):
Awesome 🥳.

PR #5724, opened 2024-02-13
fix(core): check if previous state is null for unmounted CDR component

cngonzalez:
### Description

When first arriving on a new document pane, any cross dataset references are shown as disabled.


rexxars (approved):
LGTM!

PR #5726, opened 2024-02-13
fix: remove no private

ricokahler:
### Description

We used the release script for the first time and it did not bump the private packages in the repo.

### What to review

- Take a look at the revised action and make sure it still is bumping the private packages.

### Notes for release

N/A


binoy14 (approved):
makes senes to me, thanks!

PR #5728, opened 2024-02-14
fix(structure): intent menu item nodes not rendering

### LINKED ISSUE #4891: Bug: Edit intent gets removed from menu item buttons
**Describe the bug**

The Structure builder documentation indicates that a menu item can have an intent of type `create` and `edit`. When using the `create` intent things work as expected, but when using an `edit` intent nothing happens. 

**To Reproduce**

Steps to reproduce the behaviour:

1. Use the structure builder to add an menu item with an edit intent
    ```
        S.menuItem()
            .title('Open course container config')
            .icon(CogIcon)
            .showAsAction(true)
            .intent({
                  type: 'edit',
                  params: {
                    id: courseContainerId,
                    type: 'courseContainer',
                  },
                }),
    ```
2. When rendered, click on the menu item with edit intent
3. Nothing happens

**Expected behavior**

The edit intent is executed and the studio navigates to the edit route so the document can be edited

**Screenshots**
image

```
@sanity/cli                   3.16.2 (up to date)
@sanity/code-input             4.1.1 (up to date)
@sanity/eslint-config-studio   3.0.1 (up to date)
@sanity/icons                  2.4.1 (up to date)
@sanity/types                 3.16.2 (up to date)
@sanity/ui                     1.8.2 (up to date)
@sanity/vision                3.16.2 (up to date)
sanity                        3.16.2 (up to date)
```

```macOS Ventura 13.4.1``` 

```
8.19.3
v18.13.0
```

**Additional context**

I have dug into the desk code and found out the root cause: [`sanity/desk/menuNodes.ts`](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/desk/menuNodes.ts#L129) is excluding the `intent` when resolving the menuNodes from the pane actions configuration.

rexxars:
### Description

Fixes #4891

Menu items in structure nodes that defined an intent instead of an action handler did not work after the facelift launch. There appears to be multiple different issues that caused this - I have addressed them in separate commits:

1. The `intent` definition was not actually passed onto the structure nodes, leading to them being treated as a menu button without an action handler instead of an intent button
2. The intent buttons did not actually render, because the `as="a"` rendered the entire button as a single link without any styling. I've modified it to use `forwardedAs`, but I am not confident that this approach is correct, or needs adjustment.

While investigating, I also found that the menu buttons got an aria label of `This is disabled`, even when not disabled. For now, I have disabled the aria label when not disabled, and will follow up with a PR to localize this string as well.

### What to review

1. Intent buttons work as expected. There is a test case in the test studio under the `Untitled repro` item: click the cogwheel in the next pane, which _should_ open GRRM in a document editor if this works as intended.

### Notes for release

- Fixes an issue where menu items defined in structure that used intents would not work

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -1,5 +1,6 @@
 import {
   CodeIcon,
+  CogIcon,
   EarthGlobeIcon,
   ImagesIcon,
   JoystickIcon,
@@ -48,6 +49,19 @@ export const structure: StructureResolver = (S, {schema, documentStore, i18n}) =
         .child(
           S.list()
             .title('Untitled repro')
+            .menuItems([
+              S.menuItem()
+                .title('Edit GRRM')
+                .icon(CogIcon)
+                .showAsAction(true)
+                .intent({
+                  type: 'edit',
+                  params: {
+                    id: 'grrm',
+                    type: 'author',
+                  },
+                }),
+            ])
             .items([
               S.documentListItem().id('grrm').schemaType('author'),
               S.listItem()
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Thank you! ❤️

PR #5732, opened 2024-02-14
fix(tasks): utilise document preview store for document titles

sjelfull:
### Description

This fixes document previews in tasks by utilising the same drafts-aware preview store as the document lists.

### DIFF
--- a/packages/sanity/src/tasks/src/tasks/components/list/TasksListItem.tsx
+++ b/packages/sanity/src/tasks/src/tasks/components/list/TasksListItem.tsx
@@ -2,7 +2,6 @@ import {CalendarIcon, UserIcon} from '@sanity/icons'
 import {Card, type CardProps, Flex, Stack, Text, TextSkeleton} from '@sanity/ui'
 import {useMemo} from 'react'
 import {useDateTimeFormat, useUser} from 'sanity'
-import {IntentLink} from 'sanity/router'
 import styled from 'styled-components'
 
 import {type TaskDocument} from '../../types'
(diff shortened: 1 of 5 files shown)


packages/sanity/src/tasks/src/tasks/components/list/TasksListItem.tsx
```diff
@@ -28,6 +22,14 @@ export const ThreadCard = styled(Card).attrs<CardProps>(({tone}) => ({
 `
 
 const Title = styled(Text)`
+  &:hover {
+    text-decoration: underline;
+  }
+`
+
+export const StyledIntentLink = styled(IntentLink)`
```
pedrobonamin, line 30: I think this should  be part of `DocumentPreview`  as it's only used there.

packages/sanity/src/tasks/src/tasks/hooks/useDocumentPreviewValues.ts
```diff
@@ -0,0 +1,45 @@
+import {ElementType, ReactNode} from 'react'
```
pedrobonamin, line 1: Thanks for this hook, way cleaner the implementation like this.
Just thinking, will it be better to move this to `core` instead of tasks? As it's not using anything tasks specific and we could probably reuse it in other places.

sjelfull: Long term I think we should introduce it into core, but I want us to live with it for a bit. We already have a few different preview related hooks that do similar things in core, and i'm not sure where this hook overlap with the others.

PR #5733, opened 2024-02-14
fix(structure): allow sanity.previewUrlSecret in document lists

stipsan:
### Description

The #5657 PR didn't manage to fix the regression in #5440. With the help of @sjelfull this time it's actually resolved and the fix reproducible on the Test Studio.

### What to review

Adding `sanity.previewUrlSecret` should only have an effect on this particular type, other `sanity.` prefixed types shouldn't show up anywhere, like in global search.

### Testing

In a Studio that has this plugin installed:
```ts
import {debugSecrets} from '@sanity/preview-url-secret/sanity-plugin-debug-secrets'

export default defineConfig({
  // ...
  plugins: [
    debugSecrets()
  ]
})
```
Then Structure tool should show this menu:
image
And global search should let you filter by the type as well:
image

When the plugin isn't used then these document types should remain hidden:
image

### Notes for release

- fixed regression that affected `import {debugSecrets} from '@sanity/preview-url-secret/sanity-plugin-debug-secrets'` from listing secrets

### DIFF
--- a/packages/sanity/src/core/search/common/utils.ts
+++ b/packages/sanity/src/core/search/common/utils.ts
@@ -5,7 +5,8 @@ import {isNonNullable} from '../../util/isNonNullable'
 const isDocumentType = (type: SchemaType): type is ObjectSchemaType =>
   Boolean(type.type && type.type.name === 'document')
 
-const isSanityType = (type: SchemaType): boolean => type.name.startsWith('sanity.')
+const isIgnoredType = (type: SchemaType): boolean =>
+  type.name.startsWith('sanity.') && type.name !== 'sanity.previewUrlSecret'
 
 /**
  * @internal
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/search/common/utils.ts
```diff
@@ -5,7 +5,8 @@ import {isNonNullable} from '../../util/isNonNullable'
 const isDocumentType = (type: SchemaType): type is ObjectSchemaType =>
   Boolean(type.type && type.type.name === 'document')
 
-const isSanityType = (type: SchemaType): boolean => type.name.startsWith('sanity.')
+const isSanityType = (type: SchemaType): boolean =>
```
juice49, line 8: What do you think about renaming this function to something like `isExcludedType` or `isIgnoredType`? It'd more explicitly match what the function does now.

stipsan: Sure, that makes sense to me

juice49 (approved):
Looks great to me! 🚀 Just need to sync the branch with `next`.

PR #5734, opened 2024-02-14
fix(tasks): add status to task

ninaandal:


packages/sanity/src/tasks/src/tasks/components/list/TasksList.tsx
```diff
@@ -12,6 +12,48 @@ interface TasksListProps {
...
+
+  const checkboxValues = [
+    {name: 'open', label: 'To do'},
+    {name: 'closed', label: 'Done'},
+  ]
+
+  const getLabelForStatus = (status: string) => {
+    const statusConfig = checkboxValues.find((item) => item.name === status)
+    return statusConfig?.label
+  }
```
pedrobonamin, line 24: It would be better to have this outside the component to avoid generating a new function on each render.

PR #5740, opened 2024-02-14
chore(ci) fix prettier & eslint workflows

bjoerge:
### Description
Merging to next revealed a couple of issues with lint/prettier scripts. This should hopefully fix it.
### Testing

Not sure if it's possible to properly test these before merging, unfortunately

### Notes for release

N/A

### DIFF
--- a/.github/workflows/prettier-if-needed.yml
+++ b/.github/workflows/prettier-if-needed.yml
@@ -48,7 +48,7 @@ jobs:
         with:
           path: node_modules/.cache/prettier/.prettier-cache
           key: prettier-${{ hashFiles('pnpm-lock.yaml') }}
-      - run: pnpm prettier -- --write
+      - run: npx prettier --write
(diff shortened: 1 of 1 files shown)


cngonzalez (approved):
Exciting!

package.json
```diff
@@ -20,7 +20,7 @@
     "check:test": "run-s test -- --silent",
     "check:types": "tsc --build",
     "chore:format:fix": "prettier --cache --write .",
-    "chore:lint:fix": "pnpm check:lint -- --quiet --fix",
+    "chore:lint:fix": "run-s check:lint -- --quiet --fix",
```
binoy14, line 23: might be better to just use turbo directly here I think the args passing multiple times gets messed up

```suggestion
    "chore:lint:fix": "turbo run lint -- --quiet --fix",
```

bjoerge: Done. Thanks!

PR #5533, opened 2024-01-21
chore: use pnpm as package manager

bjoerge:
### Description
This migrates this repo to use pnpm instead of (legacy) yarn 1.x

### Testing
- remove all node_modules etc.
- run pnpm install
- verify that things still works (test studios, scripts, etc)

### Notes for release

n/a - not user facing

PR #5746, opened 2024-02-15
fix(ci): use pnpm for package releases

bjoerge:
### Description
Forgot to update release workflow as part o #5533

### What to review
Does the changes make sense? I have not been able to test, but from reading the docs it seems accurate.

### Testing
Hard to test this one without polluting npm

### Notes for release

n/a - internal


ricokahler (approved):
i think this is it!

PR #5749, opened 2024-02-15
feat(core): add studioActiveToolLayout and navbar rightSectionNode prop

pedrobonamin:
### Description

This PR adds two necessary core changes for the `tasks` feature. 

#### Studio Active Tool Layout.

It's intention is to wrap the `ActiveTool` and allow plugins to modify the layout around it. 

The red border shows how it wraps the active tool.
Screenshot 2024-02-15 at 11 11 07

```
// Plugin implementation
export const tasks = definePlugin({
  name: 'sanity/tasks',
  studio: {
    components: {
			...
      activeToolLayout: TasksStudioActiveToolLayout,
    },
  },
})

// Usage
export function TasksStudioActiveToolLayout(props: ActiveToolLayoutProps) {
  const {isOpen} = useTasks()
  const {enabled} = useTasksEnabled()
  if (!enabled) return props.renderDefault(props)

  return (
    <Flex height="fill">
      <div style={{minHeight: '100%', width: '100%'}}>{props.renderDefault(props)}</div>
      {isOpen && <TasksStudioSidebar />}
    </Flex>
  )
}
```

#### Navbar rightSectionNode

Adds a prop to the studio navbar prop which will be used to insert the TasksNabarButton in the correct position. 

The red border shows where the rightSectionNode will be rendered.
Screenshot 2024-02-15 at 11 13 07

```
// Plugin 
export const tasks = definePlugin({
  name: 'sanity/tasks',
  studio: {
    components: {
			...
      navbar: TasksStudioNavbar,
    },
  },
})

// Usage
export function TasksStudioNavbar(props: NavbarProps) {
  const {enabled} = useTasksEnabled()

  if (!enabled) return props.renderDefault(props)
  return props.renderDefault({
    ...props,
    __internal_rightSectionNode: <TasksNavbarButton />,
  })
}
```

### What to review
Implementation follows the expected guidelines.

### Testing
Create a plugin which uses this new api and the navbar api.

### Notes for release

Implements new studio API, ActiveToolLayout which allows plugin authors to have more flexibility by wrapping the active tool.
Adds __internal_rightSectionNode navbar prop.

### DIFF
--- a/packages/sanity/src/core/config/studio/types.ts
+++ b/packages/sanity/src/core/config/studio/types.ts
@@ -28,6 +28,14 @@ export interface NavbarProps {
 /**
  * @hidden
  * @beta */
+export interface ActiveToolLayoutProps {
+  renderDefault: (props: ActiveToolLayoutProps) => React.ReactElement
+  activeTool: Tool
+}
+
+/**
+ * @hidden
+ * @beta */
 export interface ToolMenuProps {
   activeToolName?: string
   closeSidebar: () => void
@@ -52,6 +60,7 @@ export interface StudioComponents {
  * @hidden
  * @beta */
 export interface StudioComponentsPluginOptions {
+  activeToolLayout?: ComponentType<ActiveToolLayoutProps>
   layout?: ComponentType<LayoutProps>
   /**
    * @deprecated Add custom icons on a per-workspace basis by customizing workspace `icon` instead.
--- a/packages/sanity/src/core/config/studio/types.ts
+++ b/packages/sanity/src/core/config/studio/types.ts
@@ -1,4 +1,4 @@
-import {type ComponentType, type ReactElement} from 'react'
+import {type ComponentType, type ReactElement, type ReactNode} from 'react'
 
 import {type Tool} from '../types'
 
@@ -23,6 +23,10 @@ export interface LogoProps {
  * @beta */
 export interface NavbarProps {
   renderDefault: (props: NavbarProps) => ReactElement
+  /**
+   * @internal
+   * @beta */
+  __internal_rightSectionNode?: ReactNode
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/config/studio/types.ts
```diff
@@ -23,6 +23,10 @@ export interface LogoProps {
  * @beta */
 export interface NavbarProps {
   renderDefault: (props: NavbarProps) => ReactElement
+  /**
+   * @internal
+   * @beta */
+  __internal_rightSectionNode?: ReactNode
```
bjoerge, line 29: We should probably think about how we could offer an API with more granularity, so we could use a render/renderDefault pattern for these things as well, but I understand why `navbar.render` + `renderDefault` didn't work for you here.

bjoerge (approved):
LGTM, just a small reflection, but not a blocker for merge.

packages/sanity/src/core/config/studio/types.ts
pedrobonamin: Yes, it was an interesting discussion with the ex team, but we couldn't find any other solution given this navbar component adds multiple internal components and we would have need to recreate the same through the plugin. 
Will be a nice discussion to bring to next studio meeting. Will propose it. 

Thanks @bjoerge !

hermanwikner:
Looks great! I have two thoughts/suggestions:

- I suggest renaming `activeToolLayout` to `toolLayout`. My reasoning is that this API renders the layout for all tools, unless specifically conditioned otherwise, and given that only one tool can be active at a time, I think it's implicitly understood that the API will render the "active tool".
- It would be great if we could add a e2e test for this new API in the [Studio Components API test](https://github.com/sanity-io/sanity/blob/next/test/e2e/tests/components-api/studio.spec.ts)

PR #5751, opened 2024-02-15
chore: add missing types packages

skogsmaskin:
Add missing types for lodash and react-dom in various packages.


bjoerge (approved):
Thanks @skogsmaskin! Would be interesting to understand why pnpm didn't cause these to produce errors elsewhere (CI, etc), but that's for another day

PR #5756, opened 2024-02-15
fix(pte): update pte placeholder color

pedrobonamin:
### Description
As part of `Tasks` changes we will use PTE input to create the tasks descriptions. This highlighted the differences between the placeholder colors for the text inputs and PTE input

Updates the placeholder color of the PTE inputs.
Previously it was using the default color with a 0.5 opacity applied, which make it look similar to the color we wanted but it was not exact. This is more noticeable in dark mode.

Before and after 
image (1)
image

### Notes for release

Update placeholder color for PTE


skogsmaskin (approved):
Nice, thanks!

PR #5757, opened 2024-02-15
feat(migrate): write a log file of all submitted transactions

bjoerge:
### Description

This writes a log of all transactions submitted during a migration to a file within the migration directory. The logfile consists of newline delimited responses from the [/mutate](https://www.sanity.io/docs/http-mutations#ac77879076d4) API endpoint. I'm going for a tradeoff between minimalistic and valuable here, and did not include [the changed documents](https://www.sanity.io/docs/http-mutations#returnDocuments-369b317dcaa6) this time, because of the potential for filling up lots of disk space if running migrations on high volumes. We can always consider making it configurable behind a flag later.

Example output:
```
? This migration will run on the test dataset in ppsg7ml5 project. Are you sure? Yes
✔ Migration "rename-location-to-address" completed.

  Project id:  ppsg7ml5
  Dataset:     test

  678 documents processed.
  678 mutations generated.
  1 transactions committed.

A log of all submitted transactions has been written to:
./migrations/rename-location-to-address/logs/2024-02-15T16_44_39_628Z-rename-location-to-address.ndjson
```
☝🏼  The last two lines here are new

### What to review
- Does it make sense to make it a default to write a file, or should it be opt-in?
- Does the flag name make sense?

### Testing

- Run a migration and have a look at the logfile.

### Notes for release

- `sanity migrate` now logs all submitted transactions to a file during a migration run.

### DIFF
--- a/packages/@sanity/migrate/src/runner/run.ts
+++ b/packages/@sanity/migrate/src/runner/run.ts
@@ -56,13 +56,13 @@ export async function* toFetchOptionsIterable(
   }
 }
 
-export async function run(config: MigrationRunnerConfig, migration: Migration) {
+export async function* run(config: MigrationRunnerConfig, migration: Migration) {
   const stats: MigrationProgress = {
     documents: 0,
     mutations: 0,
     pending: 0,
     queuedBatches: 0,
-    completedTransactions: [],
+    completedTransactionsLength: 0,
     currentTransactions: [],
   }
 
@@ -141,10 +141,11 @@ export async function run(config: MigrationRunnerConfig, migration: Migration) {
   )
 
   for await (const result of commits) {
-    stats.completedTransactions.push(result)
+    stats.completedTransactionsLength++
     config.onProgress?.({
       ...stats,
     })
+    yield result
   }
   config.onProgress?.({
     ...stats,
--- a/packages/@sanity/migrate/src/types.ts
+++ b/packages/@sanity/migrate/src/types.ts
@@ -1,4 +1,4 @@
-import {type MultipleMutationResult, type Mutation as RawMutation} from '@sanity/client'
+import {type Mutation as RawMutation} from '@sanity/client'
 import {type Path, type SanityDocument} from '@sanity/types'
 
 import {type JsonArray, type JsonObject, type JsonValue} from './json'
@@ -44,7 +44,7 @@ export type MigrationProgress = {
   pending: number
   queuedBatches: number
   currentTransactions: (Transaction | Mutation)[]
-  completedTransactions: MultipleMutationResult[]
+  completedTransactionsLength: number
   done?: boolean
 }
 
--- a/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
@@ -1,3 +1,6 @@
+import {mkdirSync} from 'node:fs'
+import {type FileHandle, open} from 'node:fs/promises'
+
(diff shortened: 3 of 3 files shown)


packages/@sanity/migrate/src/runner/run.ts
```diff
@@ -56,13 +56,13 @@ export async function* toFetchOptionsIterable(
...
 
-export async function run(config: MigrationRunnerConfig, migration: Migration) {
+export async function* run(config: MigrationRunnerConfig, migration: Migration) {
   const stats: MigrationProgress = {
     documents: 0,
     mutations: 0,
     pending: 0,
     queuedBatches: 0,
-    completedTransactions: [],
+    completedTransactionsLength: 0,
```
bjoerge, line 65: this wasn't strictly needed for this PR, but we didn't use the array for anything else than reading it's length, so thought I'd might as well free up some memory.

bjoerge:
* `next`
  * **PR #5757** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5757?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sanity-io/sanity/5757?utm_source=stack-comment).

cngonzalez (approved):
I did successfully generate a logfile and personally feel that doing it by default makes sense, especially as this feature is so new.

I'd like us to surface the no-transaction-log flag pretty quickly to customers, I can see this being a possible footgun since it's not particularly visible and our docs on this are pretty verbose.

I also think the more verbose log return will be a pretty quick feature request but likely more as a series of diffs.

bjoerge:
Thanks for the review!

> I can see this being a possible footgun since it's not particularly visible and our docs on this are pretty verbose.

Not sure I understand what you mean by this. What's the footgun here?

> I also think the more verbose log return will be a pretty quick feature request but likely more as a series of diffs.

Agree! I like the idea of logging as a diff against the current/before-document. It's a slightly more involved task, so I'll file it as a feature request for now, so we can track signal.

cngonzalez:
> > I can see this being a possible footgun since it's not particularly visible and our docs on this are pretty verbose.
> 
> Not sure I understand what you mean by this. What's the footgun here?

Apologies! "Footgun" was maybe too strong a word here. "Possible inconvenience" or "source of confusion" would have been more accurate. Here's my concern more explicitly:

I expect that a number of developers will run these on sometimes high amounts of documents, generating logfiles with tens of thousands of lines each. As far as I can tell, we don't really offer any advice on how these logfiles fit within a typical development workflow. Is it annoying when I accidentally commit these to Git? Or am I _supposed_ to commit these? In what situations should I or should I not keep these logfiles around?

So for me, a small source of trouble is really just generating large-ish files and possibly having to keep them around by default. I don't think it's a big issue at all, but just one of those places where I think users would appreciate an opinionated workflow and rationale for what they should do with the things we're generating for them.

bjoerge:
These are good points and I agree with you @cngonzalez. Open for suggestions on how to proceed, but I'm thinking an acceptable path forward would be to:

- Make the logfile opt-in instead of opt-out
- Make it possible to specify the directory where the logfile should be written

e.g.:
```
sanity migrate --transaction-log --transaction-log-dir='/tmp'
```

That should remove some of the potential confusion I would hope? What to do with logfiles in terms of git/source control is something we should still document (I'd say they should be ignored).

bjoerge:
Closing this for now - might pick it up at a later point

PR #5762, opened 2024-02-15
feat(core): weight `mode` param on create intents

rdunk:
### Description

This adds `mode` as a weighted create intent param when resolving intent state. This is to differentiate between multiple tools that are able to handle the weighted `template` param.

### What to review

The `mode` parameter should be able to resolve the correct tool if multiple tools that handle the `template` param exist.

### Testing

A structure tool and presentation tool can both handle `template` params. Use the `mode` parameter in a create intent link to resolve each tool, using e.g:

/workspace/intent/create/type=*;template=*;mode=*

### Notes for release

@todo if accepted


jordanl17:
@copilot what would be a good changelog entry for this change. Keep it succinct and a single sentence to explain the impact to consuming developers on this change to supporting providing 'mode'

PR #5764, opened 2024-02-16
test(core): add tests for sort order local storage values

cngonzalez:
### Description

As we start to re-organize how we think about storing our settings, it's important to write tests first to ensure we're not losing any old functionality, especially during a transition period.

### What to review

The new test file and the small change to add a data-testid to a component that would have been brittle to reach without it.

### Testing

Feel free to run the tests locally as needed.

It would have been nice to _also_ test, in true end-to-end fashion, whether or not these parameters have the desired effect on the in-studio experience. However, that would require us to create multiple documents to sort and check, which our current testing strategy makes a bit difficult, because it runs parallel workers. I think it would be an excellent idea in the future!


binoy14 (approved):
Thank you!

PR #5767, opened 2024-02-16
chore(tasks): move CommentsSetupProvider to core

pedrobonamin:
### Description

This PR moves the `<CommentsSetupProvider>` from comments plugin to core.
We need to move this provider because `tasks` and upcoming features will need to use the comments (addon) dataset to store data. 
In favour of sharing code, this provider is moved to core.

The `runSetup` function has been refactored and rename to `createAddonDataset` to create the client but not create the comment / payload received, allowing for it's reusability.

Naming wise, we are keeping it has been renamed from `CommentsSetupProvider` to `AddonDatasetProvider`.

**Main changes**
[File moved to core.](https://github.com/sanity-io/sanity/pull/5767/files#diff-5be342a36ba645293dbdef869fcea2481ac9343cb825234b3ea9bf9b947da7e1)
[Function updated to only create the dataset.](https://github.com/sanity-io/sanity/pull/5767/files#diff-5be342a36ba645293dbdef869fcea2481ac9343cb825234b3ea9bf9b947da7e1R57)
[Caller now creates the document.](https://github.com/sanity-io/sanity/pull/5767/files#diff-aa6a407bf3d2701e782858ca24002dc474aa355759058034aed3c809f39195b3R121-R125)

### What to review

Is the location of the new provider correct?

### Testing
Creating a new comments dataset should work.

### Notes for release
Moves the CommentSetupProvider into core for reusability.


packages/sanity/src/structure/comments/src/hooks/index.ts
```diff
@@ -1,10 +1,10 @@
+export * from '../../../../core/studio/commentsSetup/useCommentsSetup'
```
hermanwikner, line 1: Do we want to re-export this hook from here?

packages/sanity/src/core/studio/commentsSetup/CommentsSetupProvider.tsx
```diff
@@ -13,6 +14,8 @@ interface CommentsSetupProviderProps {
 }
 
 /**
+ * This providers sets the addon dataset client, currently called `comments` dataset.
```
hermanwikner, line 17: Small grammar fix: providers => provider

packages/sanity/src/core/studio/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './activeWorkspaceMatcher'
 export * from './colorScheme'
+export * from './commentsSetup'
```
hermanwikner, line 3: I'd prefer having a `comments` root folder which has the `setup` folder, so that we can move other comments related things into the `comments` root folder in the future.

packages/sanity/src/structure/comments/src/hooks/index.ts
pedrobonamin: Nope, thanks for catching!

packages/sanity/src/core/studio/addOnDataset/setup/CommentsSetupProvider.tsx
```diff
@@ -51,60 +54,56 @@ export function CommentsSetupProvider(props: CommentsSetupProviderProps) {
...
-        //    so that the consumers can use it to execute comment operations and set up
-        //    the real time listener for the addon dataset.
-        const client = handleCreateClient(datasetName)
+      if (addonDatasetName) {
+        const client = handleCreateClient(addonDatasetName)
         setAddonDatasetClient(client)
+        setIsRunningSetup(false)
+        return client
+      }
+    } catch (_) {
```
cngonzalez, line 73: I'm afraid I'm not as close to this as you are, but is there a specific error we can catch here? I'm just concerned about valid errors being ignored and further work being done, getting into a strange state.

The additional error catching below might suffice, but was curious if we could get more specificity here.

cngonzalez (dismissed):

packages/sanity/src/core/studio/addOnDataset/setup/CommentsSetupProvider.tsx
cngonzalez, line 73: I'm afraid I'm not as close to this as you are, but is there a specific error we can catch here? I'm just concerned about valid errors being ignored and further work being done, getting into a strange state.

The additional error catching below might suffice, but was curious if we could get more specificity here.

pedrobonamin: Thanks @cngonzalez will defer to @hermanwikner and @sjelfull on this.

cngonzalez: Also realize that this is rearranging pre-exsiting code so maybe better to address in another PR 😅

cngonzalez (approved):
Location of the new provider looks good to Studio DX!

packages/sanity/src/core/studio/addOnDataset/setup/CommentsSetupProvider.tsx
hermanwikner: I did some testing, and it appears we no longer need this check. The logic in this try/catch was initially added because the `/setup` endpoint used to return an error if the addon dataset had already been created. However, this no longer seems to be the case, as the endpoint now returns the name of the addon dataset if it has already been created instead – which is great!

pedrobonamin: The comment below mentions that it could return an error also if the dataset doesn't exists. Maybe we can do a follow up to consider if we remove it or not, wouldn't feel confortable removing it in this one given it should be mostly focused on re-arranging the changes and not refactors. 
But also tried it and it doesn't fail if the dataset doesn't exists.

packages/sanity/src/core/studio/addonDataset/types.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @beta
+ * @hidden
+ */
+export interface AddonDatasetContextValue {
+  /**
+   * Addon dataset client, currently called `comments` dataset.
+   */
+  client: SanityClient | null
+  isRunningSetup: boolean
```
hermanwikner, line 12: Since we renamed `runSetup` to `createAddonDasaset` – should we also rename `isRunningSetup` to `isCreatingDataset`? 😇

hermanwikner (approved):
Nice stuff – thanks @pedrobonamin 👍 

I added one comment about renaming `isRunningSetup` to `isCreatingDataset` – but no blocker for merge!

packages/sanity/src/core/studio/addonDataset/types.ts
pedrobonamin: sure!!

PR #5778, opened 2024-02-16
feat(core): rename useMentionOptions to useUserListWithPermission and move to core

pedrobonamin:
### Description

The `useMentionOptions` hook was introduced into comments plugin.
This hook, fetches the project list of users and validates if the user has the desired permission for the provided document.

Renames the hook to `useUserListWithPermission` , suggestions accepted. 

We need to reuse that same logic in the `Tasks` plugin, moving this to `core` aims to keep things centralised and shared across this two plugins. 

### What to review

Is comments working the same as before?
Is this name clear and correct?
Is the location of the hook correct?

### Notes for release

Adds `useUserListWithPermission` reusable hook into `core`


packages/sanity/src/core/hooks/useUserListWithPermissions.ts
```diff
@@ -1,39 +1,71 @@
...
 
 let cachedSystemGroups: [] | null = null
 
-export function useMentionOptions(opts: MentionHookOptions): MentionOptionsHookValue {
-  const {documentValue} = opts
+/**
+ * @beta
+ * returns a list of users with the specified permission on the document.
+ * If not document is provided it will return all as `granted: true`
+ */
```
hermanwikner, line 58: Nitpick – small update suggestion for this TSdoc comment:
- Tiny grammar fix ("no" instead of "not")
- Capitalize "R" in "returns"

hermanwikner (approved):
Great stuff! Thanks @pedrobonamin 🙏

juice49 (approved):

PR #5779, opened 2024-02-16
chore: use workspace protocol for internal packages

bjoerge:
### Description
Switches to workspace protocol for internal packages. This should make it more explicit in our example studios etc what are internal dependencies. Included a couple of cleanup commits and smaller fixes to make CI happy.

### Notes for release

N/A Internal


bjoerge:
> Is that intentional?

Nope! Fixed, thanks for catching. Did you find others too?

juice49:
Okay, cool! I didn't search in depth, but I'll take a closer look now.

juice49 (approved):
I believe we have them all now! I searched the codebase using the regex `"(@sanity/[a-z|-]+|sanity)": "\^`, and found only only dependencies that exists outside the monorepo.

PR #5784, opened 2024-02-18
refactor(core): refactor SettingsStore and useStructureToolSetting

cngonzalez:
### Description

As we move toward different possible solutions for storing user values, and storing additional types of values, we can align our current implementations towards our proposed APIs. This PR is intentionally lightweight:
1. Renames `SettingsStore` to `KVStore` to better reflect the additional kinds of values that may be stored.
2. Refactors and simplifies Store logic away from namespaces. This may not be permanent but will allow for a simpler introduction of other possible storage solutions for user key-value stores.
3. Refactors `useStructureToolSetting` to align with the above.

### What to review

The changed files to ensure they're aligned with our style.

### Testing

Automated tests were introduced in #5764. 
If needed, reviewer can confirm that localStorage and UI behavior remains undisturbed.


packages/sanity/src/core/store/_legacy/kv/KVStore.ts
```diff
@@ -0,0 +1,39 @@
+import {merge, type Observable, Subject} from 'rxjs'
+import {filter, map, switchMap} from 'rxjs/operators'
+
+import {resolveBackend} from './backends/resolve'
+import {type KVStore, type KVStoreValue} from './types'
+
+/** @internal */
+export function createKVStore(): KVStore {
```
cngonzalez, line 8: On naming conventions: I followed a couple of different style guides on capitalizing `KV`:
- [Microsoft](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71))
- [AirBNB](https://github.com/airbnb/javascript?tab=readme-ov-file#naming--Acronyms-and-Initialisms)

If it's too divergent from the rest of the codebase, please let me know.

packages/sanity/src/core/store/_legacy/kv/types.ts
```diff
@@ -0,0 +1,14 @@
+import {type Observable} from 'rxjs'
+
+type JsonObject = {[Key in string]: KVStoreValue} & {[Key in string]?: KVStoreValue | undefined}
+type JsonArray = KVStoreValue[] | readonly KVStoreValue[]
+type JsonPrimitive = string | number | boolean | null
+
+/** @internal */
+export type KVStoreValue = JsonPrimitive | JsonObject | JsonArray
```
cngonzalez, line 8: This JSON logic is taken from our `migrate` package. Should we generalize this so we can use it elsewhere?

packages/sanity/src/structure/useStructureToolSetting.ts
```diff
@@ -9,36 +10,29 @@ export function useStructureToolSetting<ValueType>(
...
 
     return () => sub?.unsubscribe()
-  }, [defaultValue, key, namespace, settings])
+  }, [defaultValue, KVStoreKey, settings])
 
   const set = useCallback(
     (newValue: ValueType) => {
       setValue(newValue)
-      settings.set(newValue as any)
+      KVStore.setKey(KVStoreKey, newValue as string)
```
cngonzalez, line 33: Per my above comment, this is being cast a string since that's technically what is always getting stored (for now). Should we export the KVStoreValue type so it can be used by all consumers?

packages/sanity/src/structure/useStructureToolSetting.ts
```diff
@@ -9,36 +10,29 @@ export function useStructureToolSetting<ValueType>(
...
 ): [ValueType | undefined, (_value: ValueType) => void] {
-  const settingsStore = useSettingsStore()
+  const KVStore = useKVStore()
   const [value, setValue] = useState<ValueType | undefined>(defaultValue)
 
-  const structureToolSettings = useMemo(
-    () => settingsStore.forNamespace('structure-tool'),
-    [settingsStore],
-  )
+  const KVStoreKey = namespace ? `structure-tool::${namespace}::${key}` : `structure-tool::${key}`
```
cngonzalez, line 16: We're having internal comments about namespacing, so I don't expect this string interpolation to be permanent, hopefully

packages/sanity/src/core/store/_legacy/kv/KVStore.ts
binoy14: I am actually wondering about the word `KV` if it's too generic in the monorepo. Looking at the code I am wondering if it becomes too hard to find later. Maybe in studio it can be `createKeyValueStore` ?

cngonzalez: Would you want that change just here? Or also, for example,
`useKeyValueStore` in `packages/sanity/src/core/store/_legacy/datastores.ts`, as the type, and so on?

bjoerge: I think there's value in keeping it consistent with the backend, so if we're establishing that "kv" is the name for it in the backend, I think we should call it `kv` studio side as well (I agree that we should generally try to avoid abbreviations unless they are very well established, and kv may be in a gray area here).

bjoerge (commented):
Do we also want to move it out of the _legacy folder? I'm not entirely sure why everything was moved in there at the time, but given that we're replacing this with a proper implementation it doesn't feel like it belongs there anymore

cngonzalez:
Updates per feedback:
KVStore -> keyValueStore
directory name key-value (to match API route)
directory out of _legacy
legacy defValue var name changed to defaultValue

packages/sanity/src/core/store/_legacy/kv/KVStore.ts
cngonzalez: Changed per feedback. Thank you!

packages/sanity/src/structure/panes/document/inspectDialog/InspectDialog.tsx
```diff
@@ -27,7 +27,6 @@ export function InspectDialog(props: InspectDialogProps) {
   where the inspect dialog lives.
   This also means that when a page is loaded, the state of the tabs remains and doesn't revert to the pane tab */
   const [viewModeId, onViewModeChange] = useStructureToolSetting(
-    'structure-tool',
```
binoy14, line 30: Don't have full context if it's used in multiple places, but the function signature has not changed, won't `inspect-view-preferred-view-mode-${paneKey}` become the namespace here? Is that expected?

binoy14 (approved):
One question but LGTM besides that, thanks!

I also don't have full context on the code might be better for @bjoerge to review it as well

packages/sanity/src/structure/panes/document/inspectDialog/InspectDialog.tsx
cngonzalez: Good point -- so it looks like the behavior we have in the current version of the studio makes a key like `structure-tool::structure-tool::inspect-view` (`useStructureToolSetting` always added the `structure-tool` namespace). It seems like this was an error.

I can restore this to have the old duplicative behavior to not break anything and we can migrate it when we have a strategy to migrate all old keys.

PR #5785, opened 2024-02-19
feat(core): add Text Search API search strategy

juice49:
### Description

This branch adds the option to use the Text Search API to power Studio search, rather than the existing GROQ query search strategy.

This can be enabled in Studio projects by setting the `search.unstable_enableNewSearch` configuration option to `true`. When enabled, global search, reference search, and Cross Dataset Reference search will all use the Text Search API.

In addition, we've removed the unused offset-based pagination from the global search UI, instead implementing infinite scrolling using cursor-based pagination.

### What to review

- When `search.unstable_enableNewSearch` is not enabled, the existing search strategy will be used, and results should be unchanged.
- When `search.unstable_enableNewSearch` is enabled, search queries will use the Text Search API. Filter the request log by `textsearch` to isolate these requests. It's expected that search results will not match those provided by the existing search strategy at the moment.
- This change affects global search, reference search, and Cross Dataset Reference search.

### Testing

- The global search reducer test (`packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.test.ts`) has been updated to test pagination/infinite scrolling for cursor-based pagination.

### Notes for release

<details>
<summary>Draft notes - this has not been widely enabled for datasets yet, so is not ready to be announced.</summary>
  <dd>This release adds support for an experimental new Studio search strategy. After enabling this strategy, global search, reference search, and Cross Dataset Reference search will use a new Sanity HTTP API dedicated to performing search. The new search strategy is more efficient, and supports token negation by adding a `-` prefix.

Please note the new search strategy doesn't yet provide functional parity with the existing search strategy. Notably, field weighting, prefix search, and Portable Text search are absent. You may experience other inconsistencies or unexpected behaviour while this feature is in use.

The new search strategy can be enabled in Studio projects by setting the `search.unstable_enableNewSearch` configuration option to `true`. While we work to make this the default search strategy, we welcome any feedback you have!
</details>


packages/sanity/src/core/search/weighted/createSearchQuery.test.ts
```diff
@@ -27,15 +32,14 @@ describe('createSearchQuery', () => {
         `// findability-mvi:${FINDABILITY_MVI}\n` +
```
bjoerge, line 32: I guess we can/should remove these now? cc @robinpyon - have these been used for anything other than debugging?

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
```diff
@@ -535,3 +558,20 @@ function stripRecent(terms: RecentSearch | SearchTerms) {
   }
   return terms
 }
+
+/**
+ * At page boundaries, the Text Search API may sometimes produce duplicate results. This function
+ * deduplicates an array of results based on their ids.
+ */
+function deduplicate(hits: SearchHit[]): SearchHit[] {
```
bjoerge, line 566: Nice - maybe worth calling out that it will make sure any preexisting duplicated documents will be moved to the later "page"?

bjoerge (commented):
This is really solid work. Just had a few minor suggestions when reading through, no blockers.

bjoerge (approved):

jtpetty (approved):
LGTM 👍

PR #5786, opened 2024-02-19
refactor(core/inputs): report focusPath on text spans ending with .text

skogsmaskin:
### Description

This will make the PT-input report the `focusPath` when the focus is on text spans ending with `text` so that the Presentation Tool can work out of the box without making any exceptions for the PT-input.

With this change, we have a bi-directional mapping of focus with the forms and the presentation tool.

Note that this will not change the PTE `EditorSelection` which will continue reporting as before,  it's only the studio-specific `focusPath` reporting that will be affected by this change.

### What to review
* That regular form focus works as expected
* That the Presentation Tool focus handling works as expected
* That features like AI-Assist still works as expected.

### Notes for release
* Improved focus handling with the Presentation tool and Portable Text Inputs.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -6,7 +6,7 @@ import {
   PortableTextEditor,
 } from '@sanity/portable-text-editor'
 import {useTelemetry} from '@sanity/telemetry/react'
-import {type PortableTextBlock} from '@sanity/types'
+import {isKeySegment, type Path, type PortableTextBlock} from '@sanity/types'
 import {Box, useToast} from '@sanity/ui'
 import {
   type MutableRefObject,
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -146,6 +146,27 @@ export function PortableTextInput(props: PortableTextInputProps) {
...
 
+  const setFocusPathFromEditorSelection = useCallback(
+    (focusPath: Path) => {
+      // Report focus on spans with `.text` appended to the reported focusPath.
+      // This is done to support the Presentation tool which uses this kind of paths to refer to texts.
+      // The PT-input already supports these paths the other way around.
+      // It's a bit ugly right here, but it's a rather simple way to support the Presentation tool without
+      // having to change the PTE's internals.
+      if (
+        focusPath.length === 3 &&
```
pedrobonamin, line 157: Paraphrasing tommy, why is the` length === 3` here? What does it means when it's `3`?

skogsmaskin: It means that the `focusPath` doesn't point to inside an inline object, but terminates at the editor level.

stipsan (approved):
Tested, seems to work really well 👍

PR #5787, opened 2024-02-19
fix(core): update featuresEnabled when switching projects

pedrobonamin:
### Description

`useFeaturesEnabled` is using a cache to have a singleton of the features, this is to avoid fetching the features every time we do a request for them. 
Once this is fetched, switching workspaces won't update the features, you need to reload the studio. 

With this change, the projectId of the workspace id will be added to the `Map` so the features will be fetched for each projectId.


sjelfull (approved):
Thanks!

PR #5788, opened 2024-02-19
fix(core): bring back subtitle in workspace list view

ninaandal:
### Description
This PR reintroduces the `subtitle` in the list view for the workspace list. 

### What to review
The workspaces list view and that it looks good with and without subtitles. 

### Notes for release
Reintroduces subtitle field for workspace list views


hermanwikner (commented):
This is nice! 🙏

My only thought is that the menu items maybe feels a bit cramped with the subtitle added. Should we consider adding a bit more space to the menu items?

I did a quick test, and something like in the screenshot feels more natural to me in terms of spacing. However, it might violate some of our design principles when it comes to these opinionated Studio UI components.

Screenshot 2024-02-19 at 17 15 32

bjoerge (approved):
LGTM!

PR #5794, opened 2024-02-19
fix(core/inputs): fix issues with calling onPathFocus for PT-input

skogsmaskin:
### Description

I discovered some issues in #5786.
* Forgot to call `onPathFocus` on anything else than spans.
* There was an issue where it was testing for inlineBlocks against the wrong member key, not correctly identifying inline blocks.
* Added missing tests that should have discovered these issues.
* Added some more code comments as the code isn't too easy to understand.
 
I have added some Playwright-CT tests for this functionality now that test the values emitted by `onPathFocus` for focus on spans, inline blocks, and block objects.

### What to review
The bi-directionality and production of paths for various types of content should now be fully covered by the tests.

### Notes for release
N/A - should use notes from #5786

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
@@ -3,6 +3,7 @@ import {expect, test} from '@playwright/experimental-ct-react'
 import {type Page} from '@playwright/test'
 import {type Path, type SanityDocument} from '@sanity/types'
 
+import {testHelpers} from '../../../utils/testHelpers'
 import FocusTrackingStory from './FocusTrackingStory'
 
 export type UpdateFn = () => {focusPath: Path; document: SanityDocument}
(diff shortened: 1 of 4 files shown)


packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
```diff
@@ -136,6 +137,24 @@ test.describe('Portable Text Input', () => {
...
   })
+  test(`reports focus on spans with with .text prop, and everything else without`, async ({
+    mount,
+    page,
+  }) => {
+    const paths: Path[] = []
+    const pushPath = (path: Path) => paths.push(path)
+    await mount(<FocusTrackingStory document={document} onPathFocus={pushPath} />)
+    const {getFocusedPortableTextEditor} = testHelpers({page})
+    const $pte = await getFocusedPortableTextEditor('field-body')
```
bjoerge, line 148: Not a huge deal, but is there any particular meaning behind the `$`-prefix here? I guess it's because these refer to DOM elements, but my impression is that it's not a commonly used convention these days, since you can easily tell from the type anyway(?)

bjoerge (approved):
LGTM (just a minor question/comment). Thanks for adding tests! ❤️

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/FocusTracking.spec.tsx
skogsmaskin: It's to signal that it's a (synchronous) [Locator](https://playwright.dev/docs/locators) instance.
The test suite started with this convention, and I think the Playwright documentation used to have this convention in their examples, but I can't find them anymore. I'm all in for not using them, but then we should follow the same convention everywhere within our test suite.

PR #5800, opened 2024-02-20
fix(ci): setup pnpm

juice49:
### Description

Setup pnpm and installed dependencies.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,8 +32,11 @@ jobs:
           cache: 'pnpm'
           registry-url: 'https://registry.npmjs.org'
 
-      - name: Install
-        run: pnpm install
+      - name: Setup pnpm and install dependencies
+        uses: pnpm/action-setup@v3
+        with:
+          version: 8
+          run_install: true
 
       - name: Pre-flight
         run: |


.github/workflows/release.yml
```diff
@@ -32,8 +32,11 @@ jobs:
           cache: 'pnpm'
```
binoy14, line 32: I would remove this

.github/workflows/release.yml
```diff
@@ -32,8 +32,11 @@ jobs:
...
           registry-url: 'https://registry.npmjs.org'
 
-      - name: Install
-        run: pnpm install
+      - name: Setup pnpm and install dependencies
+        uses: pnpm/action-setup@v3
+        with:
+          version: 8
+          run_install: true
```
binoy14, line 39: Might be better to run this as a different step
```suggestion
          run_install: false
```

.github/workflows/release.yml
```diff
@@ -32,8 +32,11 @@ jobs:
...
 
       - name: Pre-flight
```
binoy14, line 41:
```suggestion
      - name: Install project dependencies
        run: pnpm install

      - name: Pre-flight
```

binoy14 (commented):
some caching changes for future improvements

PR #5806, opened 2024-02-20
feat(portable-text-editor): new API method getFragment

skogsmaskin:
### Description

This function will get the current selection in the editor as a Portable Text fragment.

See the included test for how it works.

### What to review
That the changes look good.

### Testing

See test

### Notes for release
* Portable Text Editor can now return a Portable Text fragment for the selected content.


skogsmaskin:
> Great API! Thanks @skogsmaskin What happens when we have a selection across multiple blocks? Could we add a test considering that. Also, does it work across different children in the same block?

Yes, it will be able to handle whatever selection is made in the editor and have that extracted to a fragment.

I agree that this would be great to have represented in the tests. I can add it.

skogsmaskin:
@pedrobonamin - I've added a new test that tests across blocks and also with Inline objects among the children.

pedrobonamin (approved):

PR #5807, opened 2024-02-20
feat(portable-text-editor): determine if selection is made backward

skogsmaskin:
### Description

Add .backward? on EditorSelection to be able to tell if the selection was made backward in the content when returned by `PortableTextEditor.getSelection`

This is needed in situations where the selection is made backward in the content, but we can only perform a task in the natural order (some for loop on text extractions or similar). By knowing that the selection is made backward, it's then just a matter of switching the anchor and focus points to have it in the natural order.

### What to review
That the change looks good.

### Testing
Should be covered by a lot of existing tests (see patch that updates them).

### Notes for release
* Added a `backward` property to the EditorSelection of the Portable Text Editor to tell if a selection was made backward in the content.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/types/editor.ts
+++ b/packages/@sanity/portable-text-editor/src/types/editor.ts
@@ -88,7 +88,11 @@ export interface History {
 /** @beta */
 export type EditorSelectionPoint = {path: Path; offset: number}
 /** @beta */
-export type EditorSelection = {anchor: EditorSelectionPoint; focus: EditorSelectionPoint} | null
+export type EditorSelection = {
+  anchor: EditorSelectionPoint
+  focus: EditorSelectionPoint
+  backward?: boolean
+} | null
 /** @internal */
 export interface PortableTextSlateEditor extends ReactEditor {
   _key: 'editor'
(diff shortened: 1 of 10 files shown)


pedrobonamin (approved):

pedrobonamin (commented):
Is there any test case that we could add in which it covers the `backward` selection? 
Or maybe update some of the current selections and add `backward: true` to be sure everything works in the same way even with the backward selection.

skogsmaskin:
> Or maybe update some of the current selections and add `backward: true` to be sure everything works in the same way even with the backward selection.

This is a good idea. I will add it. I think the rest of the tests will make sure that the existing functionality still works as expected.

packages/@sanity/portable-text-editor/src/types/editor.ts
```diff
@@ -88,7 +88,11 @@ export interface History {
 /** @beta */
 export type EditorSelectionPoint = {path: Path; offset: number}
 /** @beta */
-export type EditorSelection = {anchor: EditorSelectionPoint; focus: EditorSelectionPoint} | null
+export type EditorSelection = {
+  anchor: EditorSelectionPoint
+  focus: EditorSelectionPoint
+  backward?: boolean
```
hermanwikner, line 94: I wonder if it would be a bit more clear with `direction: "forward" | "backward"` – what do you think?

skogsmaskin: It would be clearer, but then it would also have to be required imo.

The absence of "backward" means it's forward by default.

skogsmaskin:
Added a new test that will test that `getSelection` has the `backward`  property set to true when the selection is set backward.

PR #5813, opened 2024-02-21
fix(form): prevent drop event propagating outside of EditPortal component

juice49:
### Description

When the `EditPortal` dialog is open for an object in an array, the array input itself still responds to `drop` events. This means it's possible to drop a file on the array while it's behind a modal dialog and appears to be inert.

It's particularly problematic if a user attempts to drop a file into the dialog itself, but slightly misses the drop zone, causing the file to be dropped onto the array.

### What to review

- The change and the test are correct.
- You can take a look in Studio by creating an object in an array (e.g. in test-studio: `test/structure/input standard;arraysTest;0154c37b-c15a-41bc-9779-54a6c1db4889`).
- While the `EditPortal` dialog is open, attempt to drop a file onto the dialog or its backdrop. The file should not be added to the list input.

### Testing

E2E test included.

```
pnpm run test:e2e test/e2e/tests/inputs/array.spec.ts
```

### Notes for release

Fixes an issue that allowed files to be dropped behind the open modal (where the inputs appeared to be inert).


bjoerge (approved):
This is great, thank you for adding tests too! <3

PR #5814, opened 2024-02-21
feat: add schema extract cmd

sgulseth:

### DIFF
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -1,0 +1,600 @@
+import {
+  type ArrayDefinition,
+  type ArrayOfType,
+  type BlockDefinition,
+  type CrossDatasetReferenceDefinition,
+  type DocumentDefinition,
+  type FieldDefinition,
+  type FileDefinition,
+  type ImageDefinition,
+  type ObjectDefinition,
+  type ReferenceDefinition,
+  type SchemaTypeDefinition,
+} from '@sanity/types'
+import {
+  type ArrayTypeNode,
+  type NumberTypeNode,
+  type ObjectKeyValue,
+  type ObjectTypeNode,
+  type OptionalTypeNode,
+  type PrimitiveTypeNode,
+  type ReferenceTypeNode,
+  type Schema,
+  type StringTypeNode,
+  type TypeNode,
+  type UnionTypeNode,
+  type UnknownTypeNode,
+} from 'groq-js/typeEvaluator'
+
+const defaultFields = (typeName: string): Array<ObjectKeyValue> => [
+  {
+    type: 'objectKeyValue',
+    key: '_id',
+    value: {type: 'string'},
+  },
+  {
+    type: 'objectKeyValue',
+    key: '_type',
+    value: {type: 'string', value: typeName},
+  },
+  {
+    type: 'objectKeyValue',
+    key: '_createdAt',
+    value: {type: 'string'},
+  },
+  {
+    type: 'objectKeyValue',
+    key: '_updatedAt',
+    value: {type: 'string'},
+  },
+  {
+    type: 'objectKeyValue',
+    key: '_rev',
+    value: {type: 'string'},
+  },
+]
+const typesMap = new Map<string, TypeNode>([
+  ['string', {type: 'string'}],
+  ['text', {type: 'string'}],
+  ['url', {type: 'string'}],
+  ['datetime', {type: 'string'}],
+  ['date', {type: 'string'}],
+  ['number', {type: 'number'}],
+  ['boolean', {type: 'boolean'}],
+  ['email', {type: 'string'}],
+])
+
+export function extractSchema(schemaTypeDefinitions: SchemaTypeDefinition[]): Schema {
+  const schema: Schema = []
+  schemaTypeDefinitions.forEach((type) => {
+    if (isDocumentType(type)) {
+      const fields: ObjectKeyValue[] = [
+        ...defaultFields(type.name),
+        ...type.fields.map(
+          (field) =>
+            ({
+              type: 'objectKeyValue',
+              key: field.name,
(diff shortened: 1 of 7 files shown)


packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+interface UnionField<T extends Field = Field> {
+  type: 'union'
+  of: T[]
+}
+
+interface ArrayField<T extends Field = Field> {
+  type: 'array'
+  of: T
+}
+interface ConcatenationField {
```
bjoerge, line 74: What is a concatenation field?

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+  return n.type === 'crossDatasetReference'
+}
+function isImageType(n: {type: string}): n is ImageDefinition {
+  return n.type === 'image'
+}
+function isFileType(n: {type: string}): n is FileDefinition {
+  return n.type === 'file'
+}
+
+type LiteralField = StringField | NumberField | BooleanField | NullField
```
bjoerge, line 389: I think this should be referred to as a "primitive" instead of "literal" since in TS literal means the actual literal value, e.g. the number `1` or the string `"foo"`

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+      })),
+    }
+  }
+  return {
+    type: 'reference',
+    to: reference.to.type,
+  } satisfies ReferenceField
+}
+function createCrossDatasetReferenceFieldDefintion(_: CrossDatasetReferenceDefinition): Field {
+  return {type: 'unknown'} satisfies UnknownField
```
bjoerge, line 668: why do we treat it as unknow? out of scope this time around?

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+  fields: Array<Field>
+}
+
+interface OptionalField<T extends Field = Field> {
+  type: 'optional'
+  value: T
+}
+
+type UnknownField = {type: 'unknown'}
+type NeverField = {type: 'never'}
```
bjoerge, line 85: what do we use this for?

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+    (('fieldset' in n && typeof n.fieldset === 'string') ||
+      !('fieldset' in n) ||
+      typeof n.fieldset === 'undefined') &&
+    (('group' in n && (typeof n.group === 'string' || Array.isArray(n.group))) ||
+      !('group' in n) ||
+      typeof n.group === 'undefined')
+  )
+}
+
+function isObjectType(n: {type: string}): n is ObjectDefinition {
```
bjoerge, line 367: a complicating factor here is that we support aliasing types with the schema today, so you can do:

```ts
const schemaTypes = [
  {type: 'object', name: 'myObject', fields: [{type: 'string', name: 'title'}]},
  {type: 'myObject', name: 'mySubObject'},
]
```
Here, myObject will not pass as an object (I think?), and you will need to use the schema to walk upwards in the type chain, to correctly identify it as an object.

Note: We don't allow extending fields, so `mySubObject` is structurally similar to `mySubObject`, but it's values will have a different `_type`.

I guess the end result here will be that the value type of `mySubObject` will be unknown, which might be a perfectly fine trade-off, just want you to be aware.

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,669 @@
...
+}
+
+type SchemaField = Document | TypeDecleration
+type Schema = SchemaField[]
+
+interface StringField {
+  type: 'string'
+  value?: string
+}
+interface NumberField {
```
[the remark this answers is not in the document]
sgulseth: These "fields" are meant to live outside of the sanity-scope, and represent type definitions. I'm all for naming them something else for example. `NumberTypeDefinition` `NumberNode`.

packages/@sanity/schema/src/sanity/extractSchema.ts
sgulseth: concatenation is a set of fields concatenated together :D 

ie. the groq query `"foo" + 3 + name` would be:
```
{
type: "concatenation",
fields: [
  { type: "string", value: "foo" },
  { type: "number", value: 3 },
  { type: "string" }, # or number/boolean depending on the schema
],
}

File: packages/@sanity/schema/src/sanity/extractSchema.ts (hunk 3 above)
Comment (sgulseth) [reply packages/@sanity/schema/src/sanity/extractSchema.ts:668]: Yep, we don't support it, yet. Would increase the scope to need to handle multiple schemas when parsing queries.

File: packages/@sanity/schema/src/sanity/extractSchema.ts (hunk 4 above)
Comment (sgulseth) [reply packages/@sanity/schema/src/sanity/extractSchema.ts:85]: They represent something we couldn't parse, and are used to ignore fields in the parser

File: packages/@sanity/schema/src/sanity/extractSchema.ts (hunk 5 above)
Comment (sgulseth) [reply packages/@sanity/schema/src/sanity/extractSchema.ts:367]: I guess here it would fail to assert the type and fallback to:
```
return {
    type: 'reference',
    to: field.type,
  }```
  
  Which would be correct? 🤔

packages/@sanity/schema/src/sanity/extractSchema.ts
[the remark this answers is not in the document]
sgulseth: I see Rico's codegen named them `StructureNode` 🤔 

`TypeNode` could also work

PR #5819, opened 2024-02-22
fix(structure): disable publish button while loading

ninaandal:
### Description
This PR adds the `editState.ready` prop to the `disabled` check and disables the `Publish` button if it's still loading, preventing the user from pressing it when checks aren't completed. 

### What to review
That the Publish button is disabled when it should be, and not disabled if it shouldn't be

### Notes for release
Disables Publish button while loading document

### DIFF
--- a/packages/sanity/src/structure/documentActions/PublishAction.tsx
+++ b/packages/sanity/src/structure/documentActions/PublishAction.tsx
@@ -176,7 +176,7 @@ export const PublishAction: DocumentActionComponent = (props) => {
 
   const disabled = Boolean(
     publishScheduled ||
-      editState.ready ||
+      !editState.ready ||
       editState?.transactionSyncLock?.enabled ||
       publishState === 'publishing' ||
       publishState === 'published' ||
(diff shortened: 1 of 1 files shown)


packages/sanity/src/structure/documentActions/PublishAction.tsx
```diff
@@ -176,6 +176,7 @@ export const PublishAction: DocumentActionComponent = (props) => {
 
   const disabled = Boolean(
     publishScheduled ||
+      editState.ready ||
```
juice49, line 179: I might be mistaken, but shouldn't this be `!editState.ready`?

juice49:
This might be a good opportunity to begin an e2e test suite for `PublishAction` 👀. What do you think?

packages/sanity/src/structure/documentActions/PublishAction.tsx
ninaandal: Hm, of course you're right... But I actually want to reset this commit. I had a conversation with Bjørge about it and he pointed out using the `editState.ready` - but this does not take in to consideration when there are validation issues, which is where this delay is most apparent. I can add this to the check as well though.

juice49: @ninaandal I'm not _super_ familiar with this part of the codebase yet, but looking at `packages/sanity/src/core/store/_legacy/document/document-pair/editState.ts`, I think you're correct that `editState.ready` doesn't reflect the validation state. It looks like `editState.ready` simply reflects whether the document pair is available (it's only `false` until a value is emitted from `snapshotPair`).

PR #5820, opened 2024-02-22
test(core): add test for storing color scheme preference and search

cngonzalez:
### Description
As we start to re-organize how we think about storing our settings, it's important to write tests first to ensure we're not losing any old functionality, especially during a transition period.

### What to review
The new test files and the small change to add a data-testid to a component that would have been brittle to reach without it.

### Testing
Feel free to run the tests locally as needed.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
@@ -91,6 +91,7 @@ export function SearchResults({inputElement}: SearchResultsProps) {
                   <CommandList
                     activeItemDataAttr="data-hovered"
                     ariaLabel={t('search.search-results-label')}
+                    data-testid="search-results"
                     fixedHeight
                     initialIndex={lastActiveIndex}
                     inputElement={inputElement}
--- a/test/e2e/tests/navbar/appearanceMenu.spec.ts
+++ b/test/e2e/tests/navbar/appearanceMenu.spec.ts
@@ -1,0 +1,35 @@
+import {expect} from '@playwright/test'
+import {test} from '@sanity/test'
+
+const COLOR_SCHEME_KEY = 'sanityStudio:ui:colorScheme'
+
+//some flakiness around local storage initial state, so add timeout and isolate
+test('default color scheme is system', async ({page, baseURL}) => {
+  await page.goto(baseURL ?? '/test/content')
+  await page.waitForTimeout(2000)
+  const localStorage = await page.evaluate(() => window.localStorage)
+  expect(localStorage[COLOR_SCHEME_KEY]).toBe('system')
+})
+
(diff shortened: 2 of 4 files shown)


test/e2e/tests/navbar/appearanceMenu.spec.ts
```diff
@@ -0,0 +1,23 @@
+import {expect} from '@playwright/test'
+import {test} from '@sanity/test'
+
+test('color scheme changes and persists', async ({page, baseURL}) => {
+  await page.goto(baseURL ?? '/test/content')
+  await page.locator(`[id='user-menu']`).click()
```
binoy14, line 6: ❓Does it default store `system` as the value?

If so maybe another expect to check default is system if nothing is selected?

test/e2e/tests/navbar/appearanceMenu.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  await page.getByLabel('Use dark appearance').click()
+
+  const localStorage = await page.evaluate(() => window.localStorage)
+  expect(localStorage['sanityStudio:ui:colorScheme']).toBe('dark')
```
binoy14, line 10: small nit: maybe it might make sense to make a constant for the key above the file so when it changes only have to change one place

binoy14 (approved):
LGTM besides small changes, thanks!

packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
```diff
@@ -101,6 +101,7 @@ export function SearchResults({inputElement}: SearchResultsProps) {
                     paddingY={1}
                     ref={setSearchCommandList}
                     renderItem={renderItem}
+                    data-testid="search-results"
```
hermanwikner, line 104: Nitpick: The props on this `CommandList` are alphabetized – so it would be nice to follow that pattern with the added test id attribute!

cngonzalez: Thanks for the feedback, @hermanwikner ! Unfortunately, `data-testid` is a [Playwright convention](https://playwright.dev/docs/locators#locate-by-test-id), and I'm not sure if there's an easy way around it. If there's somewhere else in the DOM tree you'd like me to try that would be less intrusive, please let me know.

test/e2e/tests/navbar/appearanceMenu.spec.ts
cngonzalez: done with timeout!

packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
cngonzalez: Thanks for the feedback Herman! (apologies for my misread earlier)

binoy14 (approved):
Very nice! thank you!

PR #5822, opened 2024-02-23
chore: use star for workspace protocol versions

bjoerge:
### Description
Small one: @stipsan made me aware that there's a difference between how `workspace:^` vs `worspace:*`  ends up in published packages.

TL;DR: `workspace:*` ends up using (to my surprise) _exact_ version (e.g. `"sanity": "3.30.0"`), while `workspace:^` ends up becoming `"sanity": "^3.30.0"`.
Details here: https://pnpm.io/workspaces#publishing-workspace-packages

Since we currently publish packages using lerna with `--exact`, it doesn't really affect us much, but for the sake of consistency it's better to use star.

Also found and fixed a missing workspace dependency.

### Notes for release
N/A - internal


juice49 (approved):
Aha! It didn't occur to me that we'd want the exact version!

PR #5834, opened 2024-02-23
feat(ci): cache release dependencies

juice49:
### Description

This branch adds caching for pnpm dependencies in the release workflow. It's based on [Binoy's implementation in the e2e workflow](https://github.com/sanity-io/sanity/blob/next/.github/workflows/e2e.yml).

### What to review

- Do you see any issues in the approach?

### Testing

This approach already works for the e2e workflow. If there are any issues when we next release Studio, we can revert this commit. I'm happy to be on hand for the next release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,28 @@ jobs:
         uses: pnpm/action-setup@v3
         with:
           version: 8
-          run_install: true
+          run_install: false
+
+      - name: Get pnpm store directory
+        id: pnpm-cache
+        shell: bash
+        run: |
+          echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
+
+      - name: Cache dependencies
+        uses: actions/cache@v4
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -35,7 +35,28 @@ jobs:
         uses: pnpm/action-setup@v3
         with:
           version: 8
-          run_install: true
+          run_install: false
+
+      - name: Get pnpm store directory
```
jtpetty, line 40: is this step needed?

juice49: Yes, I believe so. It sets the `STORE_PATH` variable, which is used in the subsequent `actions/cache` step (`steps.pnpm-cache.outputs.STORE_PATH`). This configures the caching path for installed pnpm dependencies.

jtpetty: ah, I see now

jtpetty (approved):

ricokahler (approved):

PR #5835, opened 2024-02-23
fix: wait for all streams to close when writing backups

rneatherway:
### Description

Without these changes I would often receive `Error: Archiving backup failed: Size mismatch` when using `sanity backup download`. I would also sometimes instead find that the tarball was successfully created but contained 

### What to review

My use of streams - I do not have much experience with them. I am not sure if the changes to the `docOutStream` are necessary.

### Testing

I tested this with a script that:

1. runs sanity backup download
2. if that succeeds, untars the export-format tarball and checks the sha1sum of all images against their filenames

I then ran this script in a loop. Before the change it rarely succeeded. After the change it has run for hundreds of iterations in a row without failing.

### Notes for release

None needed, this feature is not yet formally released.


j33ty (approved):
Interesting, good find. Thank you for taking it up. I must say I am not much familiar with Node streams either and would love a re-review from Studio team.

binoy14 (approved):
Not familiar with this `pipeline` API but given this feature is not formally release and there is a task to add tests for this happy to approve it since quick glance at the docs it does what is expected

PR #5837, opened 2024-02-26
fix(core): free trial shows twice

pedrobonamin:
### Description

A bug was reported, in which the free trial "show on load" popover shows twice when the user is navigating through the studio.
This was caused by the useEffect in `FreeTrialProvider` which handles the fetch action for the trial data, this effect had as a dependency the `router` which changes when the user is navigating through the studio.
When the router changes, the effect is triggered again, causing multiple fetch to the `journey/trial` API. On subsequent calls, this API returns twice the `showOnLoad`. 

To fix this, instead of listening to the `router` changes, the search params have been casted to a string, which are stable across renders and are safe to listen to in a `useEffect` . Then this string is reused and the params are re-created from it. 

### What to review
- side effects of this change?

### Testing
- Open the network tab, filter for the calls to the `journey/trial` API, we should not see multiple when navigating through the studio.

### Notes for release

- Fixes an issue in which the free trial popover was shown twice for new users.


drewlyton (approved):
LGTM

ninaandal (approved):

PR #5838, opened 2024-02-26
fix(test): fix breaking annotation popover test

sjelfull:
### Description

Fixes breaking annotation test by changing some of the locators to target elements specifically within the object edit popover. The test broke since we added aria labels to the PTE toolbar, meaning the locator `page.getByLabel('Link').first()` would target the toolbar button instead of the Link input inside the popover.

### What to review

- That tests runs successfully again


bjoerge (approved):
Great find. Thank you @sjelfull!

PR #5839, opened 2024-02-26
fix(core): change modalWidth of inlineObjects to 1 by default

ninaandal:
### Description

This PR changes the inline objects in portable text to be wider by default. 
There has been a lot of signal of the reference picker being too narrow. After talking with design, the decision landed on making the default width of _all_ inline objects larger (but might revisit later) 

### What to review
The reference picker/objects popovers in for example in the portable text in Standard inputs - Portable text - All the editors!

### Notes for release
Increases default size of inline objects in portable text


pedrobonamin (approved):
Looks good to me, thanks nina!

PR #5855, opened 2024-02-27
build: use github PAT for releases

ricokahler:
### Description

The follow PR updates our release script to utilize a github personal access token instead of the typical auth included by the action. This allows the action to bypass the ruleset protecting the next branch.

### What to review

Did I miss anything?

### Testing

I tested this out in our [test repo](https://github.com/ricokahler/lerna-github-publish)

### Notes for release

N/A

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0 # Fetches all history for all branches and tags
+          # note: this token should have admin permissions or else the last step
+          # to rebase next will not succeed because of the rulesets protecting
+          # the `next` branch.
+          token: ${{ secrets.GH_TOKEN }}
 
       - name: Setup node
         uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0 # Fetches all history for all branches and tags
+          # note: this token should have admin permissions or else the last step
+          # to rebase next will not succeed because of the rulesets protecting
+          # the `next` branch.
+          token: ${{ secrets.GH_TOKEN }}
```
stipsan, line 30: We should not be using Personal Access Tokens as they are incredibly insecure and their consequence if leaked are wide (you can reuse the token and escalate privileges if leaked), we should be using [an GitHub App](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/differences-between-github-apps-and-oauth-apps#machine-vs-bot-accounts), similar to [Ecospark](https://github.com/apps/ecospark), [Ecoscript](https://github.com/apps/ecoscript), as explained in [github-app-token](https://github.com/tibdex/github-app-token#github-app-token).

[More detailed docs.](https://web.archive.org/web/20230115194214/https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/automating-projects-using-actions#example-workflow-authenticating-with-a-github-app)

stipsan (changes requested):
We should not be using PAT for something as infrastructure critical as releases of `sanity`

.github/workflows/release.yml
ricokahler: heard. i'll close and reopen with a new implementation

PR #5864, opened 2024-02-27
feat(backup): change default concurrency to 24

rneatherway:


rneatherway:
We decided not to do this right now.

PR #5867, opened 2024-02-27
feat(structure): support closing first collapsed `DocumentPanel`

rdunk:
### Description

Presently, if a `DocumentPanel` is both in its collapsed state and also the first pane, neither a back button nor a close button can be displayed.

Presentation tool uses the `DocumentPanel` without any other panes, but also needs to support a collapsed state to prevent the inspector causing overflow issues.

This PR lifts some logic for displaying the `DocumentPanel` close button, relying on the presence of a provided `BackLink`, which Structure's `PaneRouterProvider` will now only pass if the pane is not the first. Presentation's router provider can therefore _always_ pass a `BackLink` and display a close button even in a collapsed state.

### What to review

There should be no changes within Structure tool, with the close and back buttons rendering as expected. The buttons should also both still be hidden if a user defines structure without multiple panes, e.g.

```ts
structureTool({
  structure: (S) =>
    S.document()
      .schemaType('page')
      .documentId('some-id'),
})
```

### Notes for release

N/A. This is an internal feature.

### DIFF
--- a/packages/sanity/src/structure/components/paneRouter/PaneRouterProvider.tsx
+++ b/packages/sanity/src/structure/components/paneRouter/PaneRouterProvider.tsx
@@ -149,7 +149,8 @@ export function PaneRouterProvider(props: {
       ChildLink,
 
       // Curried StateLink that pops off the last pane group
-      BackLink,
+      // Only pass if this is not the first pane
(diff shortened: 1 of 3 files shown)


juice49 (approved):
Looks good to me!

PR #5878, opened 2024-02-29
fix(core): increase default width of PT inline popovers

ninaandal:
### Description
It was decided that the width of the reference selectors in the PT should be increased - this has already been merged.
For consistency, the default width of the **all** inline popovers in PT is increased to match the other inline popovers in portable text. I've therefore now set the default width in the `RootPopover` instead and removed it from the `InlineObjects` popover. 
This default width can be overwritten in the schema. 

### What to review
That the width of all inline objects in the PT has a default width that is wider. 

### Notes for release
Increased width of all inline PT popovers.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx
@@ -209,6 +209,7 @@ export function AnnotationToolbarPopover(props: AnnotationToolbarPopoverProps) {
       referenceBoundary={referenceBoundary}
       referenceElement={cursorElement}
       scheme={popoverScheme}
+      {...(annotationOpen ? {width: 1} : {})}
     />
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/PortableText/object/AnnotationToolbarPopover.tsx
```diff
@@ -209,6 +209,7 @@ export function AnnotationToolbarPopover(props: AnnotationToolbarPopoverProps) {
       referenceBoundary={referenceBoundary}
       referenceElement={cursorElement}
       scheme={popoverScheme}
+      {...(annotationOpen ? {width: 1} : {})}
```
ninaandal, line 212: We don't want the popover shown before you open the annotation (to edit and delete it) to be wider

pedrobonamin (approved):
Looks good to me, thanks @ninaandal

PR #5882, opened 2024-02-29
fix(pte): updates zoffset for PopoverEditDialog to not be on top of InspectDialog

jtpetty:
### Description

Fixes EDX-814

There is an issue where certain PTE dialogs (those using PopoverEditDialog) would show through the InspectDialog.

Before:

image

After:

image

### What to review

I explicitly set the zOffset for the dialog. I am not sure if this is the "correct" fix but seems to be how these types of issues were fixed elsewhere in the code.

### Testing

Manually tested


hermanwikner (commented):
Nice! 👍 

I'm wondering if, instead of setting a specific `zOffset` value on this popover, we should set a different `zOffset` on the dialogs used in panes. This way, we can ensure that dialogs in a pane always render on top. The `zOffset` value that is being used by dialogs is provided by a `DialogProvider` used in `DocumentLayout`.

I suggest:
1. Create a new `paneDialog` value [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/components/zOffsets/defaults.ts#L15). This value should be high; when I tested it, it seems like 5000 is a good value. But, it could be even higher just to be sure.
2. Then, in the `DocumentLayout` component, we change the `zOffset` value that is passed to the `DialogProvider` to use `zOffsets.paneDialog` instead of `zOffset.portal` [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/panes/document/document-layout/DocumentLayout.tsx#L198).

This ensures that dialogs always render on top, and we won't need to set specific `zOffset` values for each popover that can be rendered in a pane. This approach is also advantageous if, for example, a custom input uses popovers since we cannot control their `zOffset` values.

hermanwikner (approved):
Works great – thanks @jtpetty!

PR #5884, opened 2024-03-01
fix(sanity): preserve form (as readOnly) when reconnecting

bjoerge:
### Description
Currently, if the connection drops while you're editing a document, the whole pane reloads and the react tree inside it will unmount/remount, loosing all its state. Since we're introducing more transient state in these components (comments, etc), we're running a bigger risk of losing people's stuff when this happens. Short disconnects happens quite often, so it can be pretty disruptive to the end-user. This PR makes sure that instead of taking over the pane when the listener disconnects (technically, is in a "reconnecting" state), we retain the form  component instance, but set it as readonly to prevent the editor for making edits without being connected.

### What to review
- I've changed a few places to use `connectionState !== 'reconnecting'` instead of `connectionState === 'connected'`. 
- Is there anything I'm missing here?

Note: this is not a "perfect solution", and I noticed during testing that any open comments dialog gets closed when the document flips to read-only, which is still not great. However, this PR should make things a little bit less disruptive.

### Testing
Manual testing is a bit tricky, because you will have to force a listener disconnect. However, I added a few lines to simulate the behavior during disconnect

## Repro _current_ behavior during disconnect
Open [this deployment](https://test-studio-git-fix-sdx-532repro-before.sanity.build), which simulates connection drop 5 seconds after you stop editing a document (try making some edits to e.g. [this document](https://test-studio-git-fix-sdx-532repro-before.sanity.build/test/structure/input-standard;textsTest;2283476b-0481-4303-89d3-27b09c24f8dd%2Ctemplate%3DtextsTest) and wait for 5 seconds)

## Test _new_ behavior during disconnect
Open [this deployment](https://test-studio-git-fix-sdx-532repro-after.sanity.build), which simulates connection drop 5 seconds after you stop editing, but also includes the changes in this PR (try making some edits to the [same document](https://test-studio-git-fix-sdx-532repro-after.sanity.build/test/structure/input-standard;textsTest;2283476b-0481-4303-89d3-27b09c24f8dd%2Ctemplate%3DtextsTest) on this deployment and wait for 5 seconds)

### Notes for release
- Improved how the editor handles intermittent disconnects. Instead of replacing the current form with a loading screen, the form will now instead be set "read only"

### DIFF
--- a/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
+++ b/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
@@ -10,7 +10,6 @@ import {
   FormBuilder,
   type FormDocumentValue,
   fromMutationPatches,
-  LoadingBlock,
   type PatchMsg,
   PresenceOverlay,
   useDocumentPresence,
@@ -48,6 +47,7 @@ export const FormView = forwardRef<HTMLDivElement, FormViewProps>(function FormV
     ready,
     formState,
     onFocus,
+    connectionState,
     onBlur,
     onSetCollapsedPath,
     onPathOpen,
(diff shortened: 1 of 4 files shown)


sjelfull (approved):
Thanks, LGTM!

PR #5887, opened 2024-03-01
fix(sanity): delay reconnect events from document pair listener to downstream consumers

bjoerge:
### Description

This adds a little delay to reconnect events emitted from the document pair listener. The effect of this change is that any downstream subscribers will be notified a second later than they currently are.

This means that in the case of an intermittent connection drops, the reconnect event may not be seen by downstream subscribers at all. We currently respond to intermittent disconnects by e.g. making forms read-only, which can feel a bit heavy handed if the connection is restored quickly, so this change should make it less disruptive.

I _think_ this should be a fairly safe change, given that connection drop can always take some time to reach the program anyway, but open to reconsider the approach (e.g. make it a UI concern)

### What to review
- Is a second a sensible delay, or should we give it a little more time? 
 
### Testing
A bit tricky to test properly, as the connection has to drop for less than 1000ms before being restored.

### Notes for release

- Add a grace period before turning forms into read-only when connection drops.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
+++ b/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
@@ -2,8 +2,8 @@
 import {type SanityClient} from '@sanity/client'
 import {type SanityDocument} from '@sanity/types'
 import {groupBy} from 'lodash'
-import {defer, type Observable, of as observableOf} from 'rxjs'
-import {concatMap, map, mergeMap, scan} from 'rxjs/operators'
+import {defer, type Observable, of as observableOf, of, timer} from 'rxjs'
+import {concatMap, map, mergeMap, scan, switchMap} from 'rxjs/operators'
 
 import {
   type IdPair,
@@ -90,6 +90,9 @@ export function getPairListener(
             ]),
           )
         : observableOf(event),
+    ),
+    switchMap((event) =>
+      event.type === 'reconnect' ? timer(1000).pipe(map(() => event)) : of(event),
     ),
     scan(
       (acc: {next: ListenerEvent[]; buffer: ListenerEvent[]}, msg) => {


bjoerge:
* **#5887** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5887?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/sanity-io/sanity/5887?utm_source=stack-comment).

juice49:
I believe [`useConnectionState` does something similar](https://github.com/sanity-io/sanity/blob/eb5cd5020d8e83026f7357221d780d221bb5ba58/packages/sanity/src/core/hooks/useConnectionState.ts#L22), but with a 200ms duration. Would it make sense to remove this now that the underlying document pair observable does it? I _think_ it's redundant with this change.

juice49:
In my testing, this seemed to have the desired effect: as long as the connection is reestablished within the timer duration, no "Connection Lost" warning appears, nor does the document editor UI switch to "Loading".

However, I did notice the document editor still reloads when the connection is reestablished. In the attached screen recording, I used a 10 second timer duration to make testing easier.

https://github.com/sanity-io/sanity/assets/1454914/456d302d-f587-436d-848d-38a34a594d78

Regarding the timer duration, 1 second _seems_ like a reasonable choice. I wonder whether we could use [`NetworkInformation`](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation) to tune this (not supported in Safari).

bjoerge:
Ah, great catch @juice49 – I didn't realize we had that delay in useConnectionState already. Think we're better off sticking to that one then, and maybe alternatively consider whether we should increase the timer slightly.

bjoerge:
btw, the pane reloading issue was addressed in a separate PR that this branch wasn't part of. Rebased against `next` now and it should be fixed.

bjoerge:
Closing this for now, since we already had logic in place for this. Might consider increasing the timeout a bit as suggested [here](https://github.com/sanity-io/sanity/pull/5887#issuecomment-1980848291).

juice49:
Okay, sounds good to me!

> btw, the pane reloading issue was addressed in a separate PR

Awesome ✨.

PR #5889, opened 2024-03-02
fix(core): allow `_dataset` for cross-dataset references in templates

rexxars:
### Description

It is currently not possible to create a cross-dataset reference from an initial value template, as the template validation runs and disallows unknown properties on references, where `_dataset` was not an allowed prop.

Perhaps we should recurse down the type chain to look for a field that extends from `crossDatasetReference`, but this is beyond what we do for other types - eg we do not check for a `reference` type when encountering `_ref`. Leaving this for a future iteration to improve on if we feel it is worth it.

### What to review

- Can create a cross-dataset reference from an initial value template
- Implementation looks acceptable

### Testing

I added some surrounding tests not only for the `_dataset` case but also for other reference properties, as I could not find any tests for that code path.

### Notes for release

- Fixes an issue where initial value templates could not define a target dataset for cross-dataset references

### DIFF
--- a/packages/sanity/src/core/templates/__tests__/resolve.test.ts
+++ b/packages/sanity/src/core/templates/__tests__/resolve.test.ts
@@ -58,6 +58,63 @@ describe('resolveInitialValue', () => {
     })
   })
 
+  test('throws on unknown prop in reference', () => {
+    expect(
+      resolveInitialValue(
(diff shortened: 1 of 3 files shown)


cngonzalez (approved):
LGTM!

PR #5890, opened 2024-03-04
feat(codegen): add 'Result' to query type names

tzhelyazkova:
### Description

Codegen generates query types named after the query variable name. This change only adds the word 'Result' at the end of the type name to differentiate between the variable name and the imported type.
E.g.

```js
import {postsResult} from './types.d.ts'

const posts = groq`*[_type="post"]`

export async function getPosts(client: SanityClient): postsResult {
  return await client.fetch(posts)
}
```


tzhelyazkova:
Merging into [feat/codegen](https://github.com/sanity-io/sanity/tree/feat/codegen)

PR #5893, opened 2024-03-04
fix: improve connection error handling

bjoerge:
### Description

This PR provides general connection error improvements by adding automatic retries in various places like:
- Document lists
- Document/value previews

The document editor already had quite decent connection handling already, so no siginificant changes there

Notably, coming back to a Studio browser tab after being away for a while, or closing the laptop lid and moving to another network has been a common source of the Studio getting stuck in a failed state.

### What to review
- Does the changes make sense? Is the retry logic sane?

As part of this work, I also moved the conditional toast into ui-components from structure into core, and introduced a common useReconnectingToast that can be shared across other parts of the Studio App that may want to invoke it. This allows multiple compoents to use it without triggering multiple toasts from appearing at the same time.

### Testing
This is quite tricky to test all cases, but disconnecting from the wifi/internet and then reconnecting again should bring the Studio back in online state without any network errors. The same goes for the notorious "I closed the laptop lid, moved to another location and then the Studio crashed"-bug that we've seen happening quite a lot.

### Notes for release

- The Studio App now handles connection issues more gracefully and will in most cases reconnect automatically in case of a connection loss


packages/sanity/src/structure/panes/documentList/listenSearchQuery.ts
```diff
@@ -170,11 +172,29 @@ export function listenSearchQuery(options: ListenQueryOptions): Observable<Searc
...
           }
+          if (event.type === 'reconnect') {
+            // Even though the listener request specifies visibility=query, the events are not guaranteed to be delivered with visibility=query
+            // If the event we are responding to arrives with visibility != query, we add a little delay to allow for the updated document to be available for queries
+            // See https://www.sanity.io/docs/listening#visibility-c4786e55c3ff
+            return of(event)
+          }
           return doFetch()
         }),
+        retry({delay: 1000, count: 100}),
```
pedrobonamin, line 183: This will retry regardless the error right? Do we want to distinguish different errors to decide if we retry or not?

packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
```diff
@@ -182,30 +184,41 @@ export const DocumentListPane = memo(function DocumentListPane(props: DocumentLi
...
-          spellCheck={false}
-          value={searchInputValue}
-        />
+        {connected ? (
+          textInput
+        ) : (
+          <Tooltip placement="bottom-end" content={t('panes.document-list-pane.reconnecting')}>
+            {textInput}
+          </Tooltip>
+        )}
```
pedrobonamin, line 221: NIT, any reason you need to handle it like this instead of
```diff
   <Tooltip 
+   disabled={!connected}
     placement="bottom-end" 
      content={t('panes.document-list-pane.reconnecting')}
      >
      {textInput}
    </Tooltip>
```

pedrobonamin (commented):
This is such a nice improvement

packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
bjoerge: Great suggestion, I've done this now

packages/sanity/src/structure/panes/documentList/listenSearchQuery.ts
bjoerge: yeah, ideally we'd just retry intermittent network errors. It's a quite hard to know what can safely be retried though. Some client errors might be intermittent, network errors are thrown as generic TypeErrors etc. I've adjusted the count here to 5 (100 retries was a bit over the top). We can iterate on this later, but 5 retries feels like a safe starting point to me, wdyt?

pedrobonamin (dismissed):
This is great, thank you!
Tested it by disconnecting and reconnecting and it all works as expected.

PR #5897, opened 2024-03-04
feat(tasks): use FormBuilder to create and edit tasks.

pedrobonamin:
### Description

Necessary PRs:
- [x] https://github.com/sanity-io/sanity/pull/5898
- [x] https://github.com/sanity-io/sanity/pull/5900
- [x] https://github.com/sanity-io/sanity/pull/5944

#### Not `tasks` specific.
- Updates the `AddonDatasetProvider` to export a `ready` prop, which indicates that the initial load has ended.
- Adds `data-ui=fieldHeaderContentBox` into [FormFieldBaseHeader](https://github.com/sanity-io/sanity/pull/5897/files#diff-8267ccab9cea4deaa06cdb29b6e360604a499102d34e34f3272a01e45f2a4fac) to allow selecting that element and update styling accordingly. 

#### `Tasks` specific
- Introduces `FormBuilder` into tasks, by inserting a new AddonWorkspaceProvider and replicating the form builder logic which currently lives only in the `PaneProvider`.
- Adds support for `drafts`, `duplicate` tasks.
- Updates `TasksNavigation` to support `copy link` action.
- Adds comments.
- Adds activity log.

Introduces two different form for tasks, which are handled by the [`tasksSchema`](https://github.com/sanity-io/sanity/pull/5897/files#diff-a9f153e013a182d83169ce01475b060c3a67408fdb643475df06a8b6572d2a64).

#### Create form
Will be used when creating a new one, duplicating or editing a draft. 
Screenshot 2024-03-11 at 09 35 14

#### Edit form

Will be used when editing a task, this form has support for `comments` and `activityLog` 
Screenshot 2024-03-11 at 09 34 49

### What to review

Overall functionality.
Is there any harm in importing components from core into the plugin when they are not exported from `core`? See [date input
](https://github.com/sanity-io/sanity/pull/5897/files#diff-fa982587bfd087e8d9d477656d4b6796b5cc277d432671bd62249050c7db24f9R17-R19)

### Notes for release

Introduces drafts, duplicate, updated navigation and new form for tasks. Adds supports for comments and activity log in tasks.

PR #5898, opened 2024-03-04
feat(core): export useWorkspaceLoader from core

pedrobonamin:
### Description

This PR splits the logic in `WorkspaceLoader` from the provider to export a reusable hook from  `core` `useWorkspaceLoader()`, this hook  will allow the use of a custom workspace loader from within the studio. 
It's necessary to use the [`AddonWorkspace` in the `tasks` feature](https://github.com/sanity-io/sanity/pull/5897/files#diff-5d30af303bd4bb6592403a720bb7282c7ce5435d1ab31935bfcb8d2a0ee7dd2e). 

### What to review

Could this be done different?


binoy14 (approved):
Makes sense to me, not very familiar if it should be exported or not but since it's internal assuming it's fine. 

Also I would remove the release notes about this since it's internal

PR #5900, opened 2024-03-04
chore(search): refactor search to export necessary components and providers.

pedrobonamin:
### Description

This PR exports some search components and providers that are necessary for the `tasks` feature.
We will reuse the logic from this search to allow users to search and assign tasks to content.

To reuse the search, we will do it as follows, importing the `SearchProvider` and the `SearchPopover`  [see dependant PR here](https://github.com/sanity-io/sanity/pull/5897/files#diff-f4ee5e22fae5b72d23163ee05278e3eda58dd1b2791f84f199a97e79900b4f37R220)
```
    <SearchProvider>
      <SearchPopover
        open={open}
        onClose={handleCloseSearch}
        onOpen={handleOpenSearch}
        onItemSelect={handleItemSelect}
        disableIntentLink
      />
   </SearchProvider>
```

The `SearchPopover` props have been extended to support and `onItemSelect` and `disableIntentLink` props.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/SearchPopover.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/SearchPopover.tsx
@@ -18,6 +18,7 @@ import {SearchWrapper} from './common/SearchWrapper'
 import {Filters} from './filters/Filters'
 import {RecentSearches} from './recentSearches/RecentSearches'
 import {SearchHeader} from './SearchHeader'
+import {type ItemSelectHandler} from './searchResults/item/SearchResultItem'
 import {SearchResults} from './searchResults/SearchResults'
 
 /**
@@ -27,7 +28,7 @@ export interface SearchPopoverProps {
   disableFocusLock?: boolean
   disableIntentLink?: boolean
   onClose: () => void
-  onItemSelect?: (item: {documentId: string; documentType: string}) => void
+  onItemSelect?: ItemSelectHandler
   onOpen: () => void
   open: boolean
 }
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/studio/components/navbar/search/components/SearchPopover.tsx
```diff
@@ -20,9 +20,14 @@ import {RecentSearches} from './recentSearches/RecentSearches'
...
 import {SearchResults} from './searchResults/SearchResults'
 
+/**
+ * @internal
+ */
 export interface SearchPopoverProps {
   disableFocusLock?: boolean
+  disableIntentLink?: boolean
   onClose: () => void
+  onItemSelect?: (item: {documentId: string; documentType: string}) => void
```
juice49, line 30: It's very small, but I wondered whether it'd be worth lifting this into a reusable type to avoid repeating it:

```ts
type ItemSelectHandler = (item: {documentId: string; documentType: string}) => void
```

juice49: I also wonder whether it makes sense to type `item` as `Pick<SanityDocumentLike, '_id' | '_type'>`.

It would result in this:

```ts
{
  _id: string
  _type: string
}
```

This might help keep things consistent across the codebase, but I'm not sure whether `onItemSelect` is intended to handle items _other than_ documents in the future.

pedrobonamin: Thanks, updated!

jtpetty (approved):
LGTM 👍

PR #5901, opened 2024-03-05
feat(core): fetch structure tool settings from backend

cngonzalez:
### Description
Scaffolds a Dataloader that hits an API endpoint for user settings -- in this case, specifically those settings relevant to structure tools.

This is meant to be just a scaffold. In the following day, please expect:
- unit tests for hooks to handle malformed or null data
- any additional error handling that can be called out here
- possible fallbacks for situations if the API is not available.

### What to review

The Dataloader backend specifically -- does it work as we expect? Any gotchas? Are errors being handled correctly?

### Testing

End-to-end tests are already in place. They've been refactored since they referred to localStorage changes.

### DIFF
--- a/packages/sanity/src/core/store/key-value/KeyValueStore.ts
+++ b/packages/sanity/src/core/store/key-value/KeyValueStore.ts
@@ -13,7 +13,7 @@ export function createKeyValueStore({client}: {client: SanityClient}): KeyValueS
 
   const updates$ = setKey$.pipe(
     switchMap((event) =>
-      storageBackend.set(event.key, event.value).pipe(
+      storageBackend.setKey(event.key, event.value).pipe(
         map((nextValue) => ({
           key: event.key,
           value: nextValue,
@@ -27,7 +27,7 @@ export function createKeyValueStore({client}: {client: SanityClient}): KeyValueS
     defaultValue: KeyValueStoreValue,
   ): Observable<KeyValueStoreValue> => {
     return merge(
-      storageBackend.get(key, defaultValue),
+      storageBackend.getKey(key, defaultValue),
       updates$.pipe(
         filter((update) => update.key === key),
         map((update) => update.value),
--- a/packages/sanity/src/core/store/key-value/KeyValueStore.ts
+++ b/packages/sanity/src/core/store/key-value/KeyValueStore.ts
@@ -1,5 +1,5 @@
 import {type SanityClient} from '@sanity/client'
-import {merge, type Observable, Subject} from 'rxjs'
+import {merge, Observable, Subject} from 'rxjs'
 import {filter, map, switchMap} from 'rxjs/operators'
 
 import {serverBackend} from './backends/server'
@@ -22,12 +22,9 @@ export function createKeyValueStore({client}: {client: SanityClient}): KeyValueS
     ),
   )
 
-  const getKey = (
-    key: string,
-    defaultValue: KeyValueStoreValue,
-  ): Observable<KeyValueStoreValue> => {
+  const getKey = (key: string): Observable<KeyValueStoreValue> => {
     return merge(
-      storageBackend.getKey(key, defaultValue),
+      storageBackend.getKey(key),
       updates$.pipe(
         filter((update) => update.key === key),
         map((update) => update.value),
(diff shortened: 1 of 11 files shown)


packages/sanity/src/core/store/key-value/KeyValueStore.ts
```diff
@@ -1,18 +1,19 @@
...
 import {merge, type Observable, Subject} from 'rxjs'
 import {filter, map, switchMap} from 'rxjs/operators'
 
-import {resolveBackend} from './backends/resolve'
+import {serverBackend} from './backends/server'
 import {type KeyValueStore, type KeyValueStoreValue} from './types'
 
 /** @internal */
-export function createKeyValueStore(): KeyValueStore {
-  const storageBackend = resolveBackend()
```
cngonzalez, line 9: @jtpetty has recommended we fall back to localStorage if the API is not available here. I've not included it in this PR since I didn't find a trivial way to deal with the asynchronous state of pinging the backend from within the resourceCache context. Hopefully it's possible.

packages/sanity/src/core/store/key-value/backends/server.ts
```diff
@@ -0,0 +1,117 @@
...
+      (acc, next) => {
+        if (next?.key) {
+          acc[next.key] = next.value
+        }
+        return acc
+      },
+      {} as Record<string, KeyValueStoreValue | null>,
+    )
+
+    const result = keys.map((key) => keyValuePairs[key] || null)
```
cngonzalez, line 44: I don't believe the backend is guaranteed to return key/value pairs in the order that they're requested, and this same method is used in userStore.

packages/sanity/src/core/store/key-value/backends/server.ts
```diff
@@ -0,0 +1,117 @@
...
+  const client = _client.withConfig({...DEFAULT_STUDIO_CLIENT_OPTIONS, apiVersion: 'vX'})
+
+  const keyValueLoader = new DataLoader<string, KeyValueStoreValue | null>(async (keys) => {
+    const value = await client
+      .request<KeyValuePair[]>({
+        uri: `/users/me/keyvalue/${keys.join(',')}`,
+        withCredentials: true,
+      })
+      .catch((error) => {
+        if (error.response?.statusCode === 404) {
```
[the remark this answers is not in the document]
binoy14: @cngonzalez I changed the API to not return 404 for not found keys

packages/sanity/src/core/store/key-value/backends/server.ts
binoy14: I can do that

cngonzalez: On revisiting this, the studio is quite fragile here. Even handling defaults in hooks and where they're being consumed does not help -- the Dataloader causes quite a dramatic error and renders the studio unusable.

For better peace of mind, I recommend that we keep this as an extra layer of handling any problems or upstream changes gracefully.

bjoerge (approved):

packages/sanity/src/core/store/key-value/backends/server.ts
```diff
@@ -0,0 +1,88 @@
...
+        body: keyValuePairs,
+        withCredentials: true,
+      }),
+    ).pipe(
+      map((response) => {
+        response.forEach((pair) => {
+          keyValueLoader.clear(pair.key)
+          keyValueLoader.prime(pair.key, pair.value)
+        })
+        return response.map((pair) => pair.value)
```
binoy14, line 65: nit: can this be done in one loop?

```suggestion
        response.map((pair) => {
          keyValueLoader.clear(pair.key)
          keyValueLoader.prime(pair.key, pair.value)
          
          return pair.value
        })
```

test/e2e/tests/desk/documentTypeListContextMenu.spec.ts
```diff
@@ -1,51 +1,63 @@
...
   await page.goto('/test/content/author')
   await page.getByTestId('pane').getByTestId('pane-context-menu-button').click()
   await page.getByRole('menuitem', {name: 'Sort by Name'}).click()
-  const localStorage = await page.evaluate(() => window.localStorage)
-
-  expect(localStorage[SORT_KEY]).toBe(
-    '{"by":[{"field":"name","direction":"asc"}],"extendedProjection":"name"}',
-  )
+  const nameResult = await sanityClient.request({
+    uri: `/users/me/keyvalue/${SORT_KEY}`,
```
binoy14, line 13: 🤔 probably need to set version to `vX` to fix the failing tests?

binoy14 (approved):
LGTM one nit pick comment and question regarding failing test

test/e2e/tests/desk/documentTypeListContextMenu.spec.ts
cngonzalez: Do the e2e tests in CI take place in staging?

binoy14: no they are against production

cngonzalez: Thought so! Can certainly change the API version if we want to, but that'll be (hopefully) be a short lived change (and won't currently fix tests 😅 )

packages/sanity/src/core/store/key-value/backends/server.ts
cngonzalez: it sure can! thanks for checking for elegance 💅

PR #5905, opened 2024-03-05
fix(cli): use inferred project root when creating migrations from subdirectory

juice49:
### Description

When used inside a Studio project subdirectory, the CLI searches parent directories in order to infer the project root. The `migration create` command doesn't currently respect the inferred root, and will attempt to create/use a `migrations` directory in the current working directory. This behaviour is incorrect; it's impossible to run migrations located anywhere other than the `migrations` directory in the project root.

This branch makes `migration create` command respect the inferred project root.

### What to review

Does this fix make sense?

### Testing

Attempt to create a migration from a Studio project subdirectory. The migration should always be added to the `migrations` directory in the project root.

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
@@ -76,12 +76,12 @@ const createMigrationCommand: CliCommandDefinition<CreateMigrationFlags> = {
       .replace(/\s+/g, '-')
       .replace(/[^a-z0-9-]/g, '')
 
-    const destDir = path.join(MIGRATIONS_DIRECTORY, sluggedName)
+    const destDir = path.join(workDir, MIGRATIONS_DIRECTORY, sluggedName)
     if (existsSync(destDir)) {
       if (
         !(await prompt.single({
           type: 'confirm',
-          message: `Migration directory ./${destDir} already exists. Overwrite?`,
+          message: `Migration directory ${chalk.cyan(destDir)} already exists. Overwrite?`,
           default: false,
         }))
       ) {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
```diff
@@ -76,12 +76,12 @@ const createMigrationCommand: CliCommandDefinition<CreateMigrationFlags> = {
       .replace(/\s+/g, '-')
       .replace(/[^a-z0-9-]/g, '')
 
-    const destDir = path.join(MIGRATIONS_DIRECTORY, sluggedName)
+    const destDir = path.join(workDir, MIGRATIONS_DIRECTORY, sluggedName)
```
cngonzalez, line 79: In the spirit of being more prescriptive about where Sanity files and folders are located, I think this is a good change!

packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
```diff
@@ -76,12 +76,12 @@ const createMigrationCommand: CliCommandDefinition<CreateMigrationFlags> = {
...
     if (existsSync(destDir)) {
       if (
         !(await prompt.single({
           type: 'confirm',
-          message: `Migration directory ./${destDir} already exists. Overwrite?`,
+          message: `Migration directory ${chalk.cyan(destDir)} already exists. Overwrite?`,
```
cngonzalez, line 84: I can see this _possibly_ being a bit confusing if you're not at the root (I would `ls` and look around for the folder). Feel free to take or leave the suggestion -- it may add noise.

```suggestion
          message: `Migration directory ${chalk.cyan(destDir)} in the root directory already exists. Overwrite?`,
```

cngonzalez (approved):

packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
juice49: Heck yeah!

packages/sanity/src/_internal/cli/commands/migration/createMigrationCommand.ts
juice49: Sorry for the delay in replying! The full path to the directory will be printed, which hopefully eliminate any ambiguity:

```
Migration directory /Users/ash/projects/sanity/dev/test-studio/migrations/nesty-testy already exists. Overwrite?
```

PR #5911, opened 2024-03-06
fix(core): clean up edit state listeners

ricokahler:
### Description

The following PR adds a `complete()` method to the document pair interface so that `memoizedPair` can close open connections.

Before:

https://github.com/sanity-io/sanity/assets/10551026/c035a6cf-3bac-4b93-be47-390d49ae3374

After:

https://github.com/sanity-io/sanity/assets/10551026/008fc3c3-ea64-4155-b393-a51eb70b428f

### What to review

Try reproducing the related bug. You should not be able to see it any more.

### Testing

Did lots of manual testing and debugging. We should get

### Notes for release

Fixes a bug that caused listener to stay open causing slow document lists and other related crashes.


bjoerge (approved):
Works like a charm, great fix ❤️

PR #5918, opened 2024-03-06
fix(comments): enable check

hermanwikner:
### Description

This pull request ensures that we only display the upsell UI or the comments UI if we can successfully fetch the feature flags for the current project. If fetching the feature flags is not possible, for example, due to the current user's permissions restricting access to "project details," we cannot accurately determine which UI to display. Therefore, in such scenarios, we must disable the feature entirely to prevent inadvertently showing the upsell UI for a project where it is actually enabled.

### What to review

- Make sure that the upsell UI is only displayed if the project is not having access to `studioComments`

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/hooks/useFeatureEnabled.ts
+++ b/packages/sanity/src/core/hooks/useFeatureEnabled.ts
@@ -7,16 +7,20 @@ import {useSource} from '../studio'
 import {DEFAULT_STUDIO_CLIENT_OPTIONS} from '../studioClient'
 import {useClient} from './useClient'
 
+const EMPTY_ARRAY: [] = []
+
 interface Features {
-  isLoading: boolean
   enabled: boolean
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/hooks/useFeatureEnabled.ts
```diff
@@ -37,29 +41,22 @@ export function useFeatureEnabled(featureKey: string): Features {
   const {projectId} = useSource()
 
   if (!cachedFeatureRequest.get(projectId)) {
-    const features = fetchFeatures({versionedClient}).pipe(
-      shareReplay(),
-      catchError((error) => {
```
hermanwikner, line 42: Removing the error catch here so that we can catch it below instead.

pedrobonamin (approved):
Thanks @hermanwikner !

PR #5919, opened 2024-03-06
feat: add cmd to generate a JSON representation of schema

sgulseth:
### Description

Adds a new command that can be executed within a sanity context to extract a JSON representation of the studios schema. This makes it easier for other tools(TM) to consume it without depending on browser environments etc. I tried to base most of the logic on `sanity schema validate` to keep consistency in the repo/package.

### What to review

Correctness: Do we parse the schema correctly? 
Naming things: It's hard! Can we name the command line argument or flags differently?
Note that we consider this feature **beta**, anything more than in the help text it should be indicated?

### Testing

```
$ pnpm build
$ sanity schema extract # somewhere there's a sanity.config.ts
```

### Notes for release

Beta: Extract sanity schema as JSON, to be consumed by other tools

### DIFF
--- a/packages/@sanity/schema/package.json
+++ b/packages/@sanity/schema/package.json
@@ -77,7 +77,7 @@
     "@sanity/generate-help-url": "^3.0.0",
     "@sanity/types": "3.32.0",
     "arrify": "^1.0.1",
-    "groq-js": "1.5.0-canary.1",
+    "groq-js": "^1.5.0-canary.1",
     "humanize-list": "^1.0.1",
     "leven": "^3.1.0",
     "lodash": "^4.17.21",
--- a/packages/@sanity/schema/package.json
+++ b/packages/@sanity/schema/package.json
@@ -77,6 +77,7 @@
     "@sanity/generate-help-url": "^3.0.0",
     "@sanity/types": "3.32.0",
     "arrify": "^1.0.1",
+    "groq-js": "1.5.0-canary.1",
     "humanize-list": "^1.0.1",
     "leven": "^3.1.0",
     "lodash": "^4.17.21",
@@ -84,6 +85,7 @@
   },
   "devDependencies": {
     "@jest/globals": "^29.7.0",
+    "@sanity/icons": "^2.8.0",
     "rimraf": "^3.0.2"
   }
 }
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -214,8 +214,19 @@ export function extractSchema(
       }
     }
 
+    // Ignore empty objects
     if (Object.keys(attributes).length === 0) {
       return {type: 'unknown'} satisfies UnknownTypeNode
+    }
+
+    if (schemaType.type?.name !== 'document') {
+      attributes._type = {
+        type: 'objectAttribute',
+        value: {
+          type: 'string',
+          value: schemaType.name,
+        },
+      }
     }
 
     return {
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -497,25 +497,29 @@ function sortByDependencies(compiledSchema: SchemaDef): string[] {
 
   // Sorts the types by their dependencies
   const typeNames: string[] = []
-  const tempMark = new Set<SanitySchemaType>()
-  const permMark = new Set<SanitySchemaType>()
+  // holds a temporary mark for types that are currently being visited, to detect cyclic dependencies
+  const currentlyVisiting = new Set<SanitySchemaType>()
+
+  // holds a permanent mark for types that have been already visited
+  const visited = new Set<SanitySchemaType>()
+
   // visit implements a depth-first search
   function visit(type: SanitySchemaType) {
-    if (permMark.has(type)) {
+    if (visited.has(type)) {
       return
     }
     // If we find a type that is already in the temporary mark, we have a cyclic dependency.
-    if (tempMark.has(type)) {
+    if (currentlyVisiting.has(type)) {
       return
     }
     // mark this as a temporary mark, meaning it's being visited
-    tempMark.add(type)
+    currentlyVisiting.add(type)
     const deps = dependencyMap.get(type)
     if (deps !== undefined) {
       deps.forEach((dep) => visit(dep))
     }
-    tempMark.delete(type)
-    permMark.add(type)
+    currentlyVisiting.delete(type)
+    visited.add(type)
 
     if (!typeNames.includes(type.name)) {
       typeNames.unshift(type.name)
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -1,0 +1,712 @@
+import {
+  type ArrayDefinition,
+  type ArrayOfType,
+  type BlockDefinition,
+  type CrossDatasetReferenceDefinition,
+  type DocumentDefinition,
+  type FieldDefinition,
+  type FileDefinition,
+  type ImageDefinition,
+  type NumberDefinition,
+  type ObjectDefinition,
+  type ReferenceDefinition,
+  type Rule,
+  type SchemaTypeDefinition,
+  type StringDefinition,
+} from '@sanity/types'
+import {
+  type ArrayTypeNode,
+  createReferenceTypeNode,
+  type InlineTypeNode,
+  type NumberTypeNode,
+  type ObjectAttribute,
+  type ObjectTypeNode,
+  type PrimitiveTypeNode,
+  type SchemaType,
+  type StringTypeNode,
+  type TypeNode,
+  type UnionTypeNode,
+  type UnknownTypeNode,
+} from 'groq-js'
+
+const documentDefaultFields = (typeName: string): Record<string, ObjectAttribute> => ({
+  _id: {
+    type: 'objectAttribute',
+    value: {type: 'string'},
+  },
+  _type: {
+    type: 'objectAttribute',
+    value: {type: 'string', value: typeName},
+  },
+  _createdAt: {
+    type: 'objectAttribute',
+    value: {type: 'string'},
+  },
+  _updatedAt: {
+    type: 'objectAttribute',
+    value: {type: 'string'},
+  },
+  _rev: {
+    type: 'objectAttribute',
+    value: {type: 'string'},
+  },
+})
+const typesMap = new Map<string, TypeNode>([
+  ['text', {type: 'string'}],
+  ['url', {type: 'string'}],
+  ['datetime', {type: 'string'}],
+  ['date', {type: 'string'}],
+  ['boolean', {type: 'boolean'}],
+  ['email', {type: 'string'}],
+])
+
+export interface ExtractSchemaOptions {
+  enforceRequiredFields?: boolean
+}
+
+export function extractSchema(
+  schemaTypeDefinitions: SchemaTypeDefinition[],
+  extractOptions: ExtractSchemaOptions = {},
+): SchemaType {
+  const schema: SchemaType = []
+  schemaTypeDefinitions.forEach((type) => {
+    if (isDocumentType(type)) {
+      const attributes = documentDefaultFields(type.name) satisfies Record<string, ObjectAttribute>
+
+      for (const field of type.fields || []) {
+        const fieldIsRequired = isFieldRequired(field)
+        attributes[field.name] = {
+          type: 'objectAttribute',
+          optional: extractOptions.enforceRequiredFields ? fieldIsRequired : true,
+          value: parseField(field, extractOptions),
+        } satisfies ObjectAttribute
+      }
+
+      schema.push({
+        name: type.name,
+        type: 'document',
+        attributes,
+      })
+      return
+    }
+
+    if (isObjectType(type)) {
+      const attributes = type.fields.reduce<Record<string, ObjectAttribute>>((acc, field) => {
+        const fieldIsRequired = isFieldRequired(field)
+        acc[field.name] = {
+          type: 'objectAttribute',
+          optional: extractOptions.enforceRequiredFields ? !fieldIsRequired : true,
+          value: parseField(field, extractOptions),
+        } satisfies ObjectAttribute
+
+        return acc
+      }, {}) satisfies Record<string, ObjectAttribute>
+
+      attributes._type = {
+        type: 'objectAttribute',
+        value: {
+          type: 'string',
+          value: type.name,
+        },
+      } satisfies ObjectAttribute<StringTypeNode>
+
+      schema.push({
+        name: type.name,
+        type: 'type',
+        value: {
+          type: 'object',
+          attributes,
+        },
+      })
+      return
+    }
+    if (isArrayType(type)) {
+      schema.push({
+        type: 'type',
+        name: type.name,
+        value: createArray(type, extractOptions),
+      })
+      return
+    }
+
+    if (isBlockType(type)) {
+      schema.push({
+        type: 'type',
+        name: type.name,
+        value: createBlock(type, extractOptions),
+      })
+      return
+    }
+    if (isImageType(type)) {
+      schema.push({
+        type: 'type',
+        name: type.name,
+        value: createImage(type, extractOptions),
+      })
+      return
+    }
+    if (isFileType(type)) {
+      schema.push({
+        type: 'type',
+        name: type.name,
+        value: createFile(type, extractOptions),
+      })
+      return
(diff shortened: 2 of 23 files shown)


packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
```diff
@@ -0,0 +1,70 @@
+import {type CliCommandArguments, type CliCommandContext} from '@sanity/cli'
+import {writeFile} from 'fs/promises'
+import {dirname, join} from 'path'
+import readPkgUp from 'read-pkg-up'
+import {Worker} from 'worker_threads'
```
rexxars, line 5: Prefer `node:` prefixed imports for new code

```suggestion
import {writeFile} from 'node:fs/promises'
import {dirname, join} from 'node:path'
import readPkgUp from 'read-pkg-up'
import {Worker} from 'node:worker_threads'
```

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
```diff
@@ -0,0 +1,70 @@
...
+    'cli',
+    'threads',
+    'extractSchema.js',
+  )
+
+  const spinner = output
+    .spinner({prefixText: '📦'})
+    .start(
+      flags['enforce-required-fields']
+        ? 'Extracting schema, with enforced requried fields'
```
rexxars, line 44:
```suggestion
        ? 'Extracting schema, with enforced required fields'
```

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
```diff
@@ -0,0 +1,70 @@
...
+    dirname(rootPkgPath),
+    'lib',
+    '_internal',
+    'cli',
+    'threads',
+    'extractSchema.js',
+  )
+
+  const spinner = output
+    .spinner({prefixText: '📦'})
```
rexxars, line 41: Not a big deal, but we don't currently have an emoji convention in the CLI 🤷‍♂️

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
```diff
@@ -0,0 +1,70 @@
...
+    worker.addListener('error', reject)
+  })
+
+  const path = flags.path || join(process.cwd(), 'schema.json')
+
+  spinner.text = `Writing schema to ${path}`
+
+  await writeFile(path, JSON.stringify(schema, null, 2))
+
+  spinner?.succeed('Extrcted schema')
```
rexxars, line 69:
```suggestion
  spinner?.succeed('Extracted schema')
```

rexxars, line 69: Why is this optionally chained? Could `spinner` be null here?

packages/sanity/src/_internal/cli/commands/schema/extractSchemaCommand.ts
```diff
@@ -0,0 +1,31 @@
+import {type CliCommandDefinition} from '@sanity/cli'
+
+const description = 'Extracts a json representation of a sanity schema within a Studio context.'
```
rexxars, line 3:
```suggestion
const description = 'Extracts a JSON representation of a Sanity schema within a Studio context.'
```

packages/sanity/src/_internal/cli/commands/schema/extractSchemaCommand.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const helpText = `
+**Note**: This command is experimental and subject to change.
+
+Options
+  --workspace <name> The name of the workspace to generate a schema for
+  --path Optional path to specify destination of the schema file
+  --enforce-required-fields Makes the schema generated respect required fields in the schema. This can be used when generating types if you are only querying for published documents. Defaults to false.
```
rexxars, line 11: - This seems broader than just for querying - I would phrase this to be "treat fields marked as required as non-optional", as what it impacts is the resulting `optional` attributes
- "querying for published documents" doesn't really encompass this - both drafts and published documents can be incomplete/not follow schema (documents that were published prior to a field being introduced, for instance)

packages/sanity/src/_internal/cli/commands/schema/extractSchemaCommand.ts
```diff
@@ -0,0 +1,31 @@
...
+  # Extracts schema types in a Sanity project with more than one workspace
+  sanity schema extract --workspace default
+`
+
+const extractSchemaCommand: CliCommandDefinition = {
+  name: 'extract',
+  group: 'schema',
+  signature: '',
+  description,
+  helpText,
```
rexxars, line 23: Since this is an experimental/beta feature, maybe we should hide it from the help commands output for now?
```suggestion
  helpText,
  hideFromHelp: true,
```

packages/sanity/src/_internal/cli/threads/extractSchema.ts
```diff
@@ -0,0 +1,70 @@
+import {extractSchema} from '@sanity/schema/_internal'
+import {type Workspace} from 'sanity'
+import {isMainThread, parentPort, workerData as _workerData} from 'worker_threads'
```
rexxars, line 3:
```suggestion
import {isMainThread, parentPort, workerData as _workerData} from 'node:worker_threads'
```

packages/sanity/src/_internal/cli/threads/extractSchema.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (workspaces.length === 0) {
+    throw new Error('No studio configuration found')
+  }
+
+  if (workspaces.length === 1) {
+    return workspaces[0]
+  }
+
+  if (workspaceName === undefined) {
+    throw new Error('Multiple workspaces configured. Please specify which workspace to use')
```
rexxars, line 63:
```suggestion
    throw new Error('Multiple workspaces configured. Please specify workspace using the `--workspace` flag')
```

rexxars (commented):
I'll leave the bulk of the review up to the tagged reviewers, but I have some thoughts/questions:

- A bit reluctant to tie `@sanity/schema` to `groq-js`
  - A schema extract command makes total sense to me, but the resulting format should preferably be generic enough that it can be used to generate all kinds of things; GraphQL types, TypeScript types, tailored API clients etc.
  - Generally I would favor using a more established output type, say [JSONSchema](https://json-schema.org/), as long as it is able to represent our schema completely.
  - Perhaps we could introduce a format parameter/flag? We don't have to implement any other formats right now, but I think requiring the flag for now makes it more future-compatible if we want to default to (say) JSONSchema in the future.
- Using the compiled schema (`Schema.compile()`/`createSchema()`) would be safer
  - We shouldn't have to special-case image/file fields/types etc - and defining their properties are already done for you with this approach. It feels brittle/unnecessary to have to maintain two code paths if we add new fields.
  - The generated type output may be missing types such as image and file asset documents, geopoints, slugs and other non-user-defined types
  - Block types shouldn't have to be special-cased - they (should) just be an object type like any other. Given you can create named block types, this is important - as the `name` property may not be `block`
  - Without having looked deeply at the code, I am curious if the current approach handles aliased types, or if that is left to whatever acts on these types to figure out

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
sgulseth: I can check, was probably just VSCode inserting it 🤷

packages/sanity/src/_internal/cli/commands/schema/extractSchemaCommand.ts
sgulseth: > This seems broader than just for querying - I would phrase this to be "treat fields marked as required as non-optional", as what it impacts is the resulting optional attributes

True, and agree on your phrasing 👍 

>  This seems broader than just for querying - I would phrase this to be "treat fields marked as required as non-optional", as what it impacts is the resulting optional attributes
"querying for published documents" doesn't really encompass this - both drafts and published documents can be incomplete/not follow schema (documents that were published prior to a field being introduced, for instance).

Yeah, we do however consider the schema the source of truth so if anything diverges from your schema we don't give any guarantees with the types at least 🤔

sgulseth:
> A bit reluctant to tie @sanity/schema to groq-js.

Where would you prefer these types to live?

I don't see any reasons this couldn't be used with our graphql generator as well, would potentially need to extend ObjectAttribute to include more data, but that's doable.

We did talk about using jsonschema, but the ecosystem is a bit fragmented with different specification and support, so we opted for our own implementation here.

cc. @judofyr for additional feedback.

> Perhaps we could introduce a format parameter/flag
We could. Would then this be a required flag to set? ie `sanity schema extract --format=typenode` Could it default to `typenode` at least while being beta? To simplify DX and feedback for now.

> Using the compiled schema (Schema.compile()/createSchema()) would be safer

Ah, nice. I'll look into this!

packages/@sanity/schema/package.json
```diff
@@ -77,13 +77,15 @@
...
     "arrify": "^1.0.1",
+    "groq-js": "1.5.0-canary.1",
     "humanize-list": "^1.0.1",
     "leven": "^3.1.0",
     "lodash": "^4.17.21",
     "object-inspect": "^1.6.0"
   },
   "devDependencies": {
     "@jest/globals": "^29.7.0",
+    "@sanity/icons": "^2.8.0",
```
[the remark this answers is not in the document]
sgulseth: I'm using icons in a fixture/test. Guess I _can_ remove them, wanted a schema that was similar to a real world scenario, though I guess having a icon is not needed

packages/sanity/src/_internal/cli/threads/extractSchema.ts
```diff
@@ -0,0 +1,73 @@
...
+const opts = _workerData as ExtractSchemaWorkerData
+const cleanup = mockBrowserEnvironment(opts.workDir)
+
+async function main() {
+  try {
+    const workspaces = await getStudioWorkspaces({basePath: opts.workDir})
+
+    const workspace = getWorkspace({workspaces, workspaceName: opts.workspaceName})
+
+    const {types} = workspace.schema._original || {types: []}
```
bjoerge, line 32: If you're looking into using the compiled schema, you can use `workspace.schema` instead of `workspace.schema._original` here

bjoerge:
>> A bit reluctant to tie @sanity/schema to groq-js.

> Where would you prefer these types to live?

Any reason extractSchema couldn't live with the cli command for now?

packages/@sanity/schema/package.json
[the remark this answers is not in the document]
rexxars: Ah. No worries, I just missed it.

sgulseth:
@bjoerge then we either need to have the types in two places, guess that works with how ts infers types. Or groq-js would need to depend on the cli, which isn't ideally either

rexxars:
> Where would you prefer these types to live?

Not sure, honestly - probably fine for now.

> We did talk about using jsonschema, but the ecosystem is a bit fragmented with different specification and support, so we opted for our own implementation here.

Would be interested in learning more about this, but I don't need a full write-up here - if there's any TLDR or links I can use to understand more that'd be great.

>> Perhaps we could introduce a format parameter/flag
>
> We could. Would then this be a required flag to set? ie sanity schema extract --format=typenode Could it default to typenode at least while being beta? To simplify DX and feedback for now. 

Ideally I would default to something that is a defined standard, such as JSONSchema, and make other formats a flag. I understand that it's an improved DX when we only have a single format, but it would also be a worse DX when we change the default (breaking change).

sgulseth:
I have adressed the feedback and we are now processing the compiled types. OK to move the format discussion to slack?

rexxars (approved):
Overall this looks good to me, but would love if @bjoerge who knows the schema better could have a look.

One curiosity; seems to me like Portable Text block and span types do not have a `_type` property? Span should always be `span` (the type cannot be changed in any way), and the block type should either be `block` or the name given by the user.

packages/@sanity/schema/package.json
```diff
@@ -77,13 +77,15 @@
     "@sanity/generate-help-url": "^3.0.0",
     "@sanity/types": "3.32.0",
     "arrify": "^1.0.1",
+    "groq-js": "1.5.0-canary.1",
```
bjoerge, line 80: would it make sense to widen the range to `^1.5.0.canary.1`?

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,530 @@
...
+      return type
+    }
+    type = type.type
+  }
+
+  return undefined
+}
+
+// Sorts the types by their dependencies by using a topological sort depth-first algorithm.
+function sortByDependencies(compiledSchema: SchemaDef): string[] {
```
bjoerge, line 445: Very nice – I can easily see this being useful as a util on its own (but let's keep it local here for now).

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,530 @@
...
+    }
+    const dependencies = new Set<SanitySchemaType>()
+
+    walkDependencies(schemaType, dependencies)
+    dependencyMap.set(schemaType, dependencies)
+  })
+
+  // Sorts the types by their dependencies
+  const typeNames: string[] = []
+  const tempMark = new Set<SanitySchemaType>()
```
bjoerge, line 500: could we find a better name, e.g. `currentlyVisiting` and add a comment explaining that it's for detecting cyclic dependencies?

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -0,0 +1,530 @@
...
+  const permMark = new Set<SanitySchemaType>()
```
bjoerge, line 501: could we find a better name, e.g. `visited`?

packages/@sanity/schema/test/extractSchema/extractSchema.test.ts
```diff
@@ -0,0 +1,379 @@
...
+      return {schemaName, schema}
+    })
+    const passes = cases.filter((v): v is {schemaName: string; schema: Schema} => v.schema !== null)
+
+    test.each(passes)('extracts schema $schemaName', ({schema}) => {
+      const extracted = extractSchema(schema)
+      expect(extracted.length).toBeGreaterThan(0) // we don't really care about the exact number, just that it passes :+1:
+    })
+
+    // const skips = cases.filter((v): v is {schemaName: string; schema: null} => v.schema === null)
```
bjoerge, line 374: if this is intentionally left here, could you add a quick note explaining why?

packages/@sanity/schema/test/extractSchema/fixtures/block.ts
```diff
@@ -0,0 +1,446 @@
...
+        },
+        {
+          type: 'reference',
+          name: 'bookReference',
+          to: {type: 'book'},
+          title: 'Reference to book',
+        },
+        {type: 'author', title: 'Embedded author'},
+        {type: 'code', title: 'Code'},
+        // {
```
bjoerge, line 135: if this is intentionally left here, could you add a quick note explaining why?

packages/@sanity/schema/test/extractSchema/fixtures/block.ts
```diff
@@ -0,0 +1,446 @@
...
+                  fields: [
+                    {name: 'aString', type: 'string'},
+                    {name: 'aNumber', type: 'number'},
+                  ],
+                },
+              ],
+            },
+          ],
+        },
+        // {
```
bjoerge, line 167: if this is intentionally left here, could you add a quick note explaining why?

bjoerge (approved):
Great work! Left a few comments for your consideration, only nits/nice-to-haves, so good to merge from my side!

packages/@sanity/schema/package.json
sgulseth: Looks like it doesn't work, because it's not valid semver? 🤔 

```
 ERR_PNPM_SPEC_NOT_SUPPORTED_BY_ANY_RESOLVER  groq-js@^1.5.0.canary.1 isn't supported by any available resolver.
```

bjoerge: my bad, should've been `^1.5.0-canary.1` (notice the dash)

packages/@sanity/schema/src/sanity/extractSchema.ts
bjoerge: I don't see a use for it right now, so think its fine to leave it as-is. I think the comment and name already includes the right keywords for someone to be able search for it if they're wondering if we got a util for it already.

sgulseth: Cool!

PR #5920, opened 2024-03-06
fix(ci): align pnpm install across all workflows

bjoerge:
### Description
This aligns how we install dependencies across all our workflow definitions. It's using the same approach as @binoy14 previously added for the e2e test workflow.

### What to review

- Should we always pass `--config.ignore-scripts=true` when running pnpm install? (side note: why the `config.`-prefix and not just `--ignore-scripts`?)

### Testing
- CI is taking care of it!

### Notes for release
n/a internal

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,38 +37,6 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
-      - uses: pnpm/action-setup@v2
-        name: Install pnpm
-        id: pnpm-install
-        with:
-          version: 8
-          run_install: false
-
-      - name: Get pnpm store directory
-        id: pnpm-cache
-        shell: bash
-        run: |
-          echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
-
-      - name: Cache node modules
-        id: cache-node-modules
-        uses: actions/cache@v4
-        env:
-          cache-name: cache-node-modules
-        with:
-          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ env.cache-name }}-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            v1-${{ runner.os }}-pnpm-store-${{ env.cache-name }}-
-            v1-${{ runner.os }}-pnpm-store-
-            v1-${{ runner.os }}-
-
-      - name: Install project dependencies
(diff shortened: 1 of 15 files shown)


.github/workflows/test.yml
```diff
@@ -37,38 +37,6 @@ jobs:
...
-          key: ${{ runner.os }}-pnpm-store-${{ env.cache-name }}-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            v1-${{ runner.os }}-pnpm-store-${{ env.cache-name }}-
-            v1-${{ runner.os }}-pnpm-store-
-            v1-${{ runner.os }}-
-
-      - name: Install project dependencies
-        run: pnpm install
-
-      - name: Build CLI
```
bjoerge, line 70: this was probably just leftover cruft, and there was no point in running build during the install step, so moved building to right before running the tests below

.github/workflows/test.yml
```diff
@@ -174,10 +134,3 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
-
-      # Delete the cache so it is only used once
-      - name: Delete Cache
```
bjoerge, line 171: no longer needed since we don't need the build files to be cached between the steps

.github/workflows/test.yml
binoy14: actually I misunderstood this, so the way this is working is it's running the build step in install job, so it builds for `node18` and `node20` then it caches the build and passes it to the test job which runs for a mix of `node + shard matrix` if the `build` step is still needed we should keep the cache so it doesn't need to build on each matrix which would take quite long for 4 shards if cache is not available.

I thought the build step isn't required at all but maybe that will not work. Happy to chat more on this

bjoerge: > I thought the build step isn't required at all but maybe that will not work. Happy to chat more on this

You're right! Jest is configured to use the dev aliases, so tests can run from source. Although discovered that the validation tests, requires these aliases to be registered inside the worker thread that it runs.

Opened a PR to fix this here: https://github.com/sanity-io/sanity/pull/5927

binoy14 (approved):
very nice thank you for doing it. 

Small note the install [job](https://github.com/sanity-io/sanity/pull/5920/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88L11)  and the cleanup jobs can be removed for the test action since they are not doing anything

PR #5922, opened 2024-03-06
fix: add missing project reference

bjoerge:
### Description

This repo is using [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) which lets us to run `tsc --build` from source files, and have TypeScript build the packages in topological order (e.g. start by building the packages that's depended on by other packages). Unfortunately TS isn't workspace aware, so the workspace dependencies needs to be explicitly declared as references in `tsconfig.json`. This PR adds them where they are needed.

### Testing
The type check should now pass on CI.

### Notes for release
n/a


jtpetty (approved):
👍

PR #5925, opened 2024-03-07
fix(comments): handle lack of access to inline comments

hermanwikner:
### Description

This pull request ensures that the upsell dialog for comments is displayed when a user attempts to add an inline comment by clicking "Add comment" in the PTE.

### What to review

- Make sure that the upsell dialog is correctly displayed

### Notes for release

N/A


joneidejohnsen (approved):

jtpetty (approved):
👍

PR #5931, opened 2024-03-07
fix(core): remove prev and next month buttons in calendar input

jtpetty:
### Description

Fixes EDX-1124

Removes the previous and next buttons on the month for the calendar input

Before:
image

After:

image

### What to review

Anything I missed?

### Testing

Code deletion, no test changes

### DIFF
--- a/packages/sanity/src/core/form/inputs/DateInputs/base/calendar/Calendar.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/base/calendar/Calendar.tsx
@@ -203,10 +203,6 @@ export const Calendar = forwardRef(function Calendar(
           <Box flex={1}>
             <CalendarMonthSelect
               onChange={handleFocusedMonthChange}
-              labels={{
-                goToPreviousMonth: labels.goToPreviousMonth,
-                goToNextMonth: labels.goToNextMonth,
-              }}
               monthNames={labels.monthNames}
               value={focusedDate?.getMonth()}
             />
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/DateInputs/base/calendar/Calendar.tsx
```diff
@@ -342,25 +340,10 @@ function CalendarMonthSelect(props: {
     goToNextMonth: string
   }
 }) {
-  const {moveFocusedDate, onChange, value, labels, monthNames} = props
-
-  const handlePrevMonthClick = useCallback(() => moveFocusedDate(-1), [moveFocusedDate])
-
-  const handleNextMonthClick = useCallback(() => moveFocusedDate(1), [moveFocusedDate])
+  const {onChange, value, labels, monthNames} = props
```
cngonzalez, line 343: Shall we remove the labels from our props too? They were only used for the Button component, as far as I can see.
```suggestion
  const {onChange, value, monthNames} = props
```

Github won't let me remove the labels parameter from the function arguments here or in codespaces unfortunately.

packages/sanity/src/core/form/inputs/DateInputs/base/calendar/Calendar.tsx
```diff
@@ -202,7 +202,6 @@ export const Calendar = forwardRef(function Calendar(
         <Flex>
           <Box flex={1}>
             <CalendarMonthSelect
-              moveFocusedDate={moveFocusedDate}
               onChange={handleFocusedMonthChange}
               labels={{
```
cngonzalez, line 206: Since labels were only used for the month change buttons, I think we can remove!

cngonzalez (approved):
Some nits for removing unused code (button labels) but otherwise seems reasonable.

jtpetty:
@cngonzalez - Great find! Thanks! Updated

PR #5932, opened 2024-03-08
feat(core): redirect to previous path after login

ricokahler:
### Description

The following PR updates the `origin` parameter of our login to include the current path so that after the user logs in, they should be redirected to the path they were trying to get to instead of having to click the link again.

https://github.com/sanity-io/sanity/assets/10551026/8605fcb6-9767-44b0-802f-0a1340178822

### What to review

Ensure that the code works and does what is expected. Ensure there are no breaking changes to the API as well (I deprecated a key).

### Testing

See the video.

### Notes for release

- Resolves an issue where users were not redirected to their original destination after re-authenticating, ensuring seamless access to linked content within the Studio.


jtpetty (approved):
LGTM - is there any way we can add tests for this to make sure this doesn't regress?

PR #5933, opened 2024-03-08
feat(core): add groups to document actions, introduce paneActions group

pedrobonamin:
### Description

Introduces `groups` into [DocumentActions](https://www.sanity.io/docs/document-actions-api) allowing plugin developers to add actions and decide where they want this actions to show:
- `default` will add the actions to the same position as before, in the footer right next to the publish button.
- `paneActions` will add the action in the context menu right next to the inspectors and other pane actions.

This is initially intended to support creating `tasks` from within a document. 

#### Example of usage

```
export function ConfirmDialogAction({onComplete}): DocumentActionDescription {
  const [dialogOpen, setDialogOpen] = useState(false)

  return {
    label: 'Show confirm',
    onHandle: () => {
      setDialogOpen(true)
    },
    group: ['paneActions'],
    dialog: dialogOpen && {
      type: 'confirm',
      onCancel: onComplete,
      onConfirm: () => {
        alert('You confirmed!')
        onComplete()
      },
      message: 'Please confirm!',
    },
  }
}

const plugin = definePlugin({
  ...,
  document: {
    actions: (prev) => [...prev, ConfirmDialogAction],
  }
})
```

Screenshot 2024-03-08 at 10 30 23

https://github.com/sanity-io/sanity/assets/46196328/3b508edd-9175-4108-b748-cbfd223cfed3

### What to review

Is there other way to simplify this?
Are the changes in `GetHookCollectionState` correct?

### Testing

Create an action from within a plugin, the action should render in the selected group.

### Notes for release

Add support for groups in document actions.

### DIFF
--- a/packages/sanity/src/core/components/hookCollection/GetHookCollectionState.tsx
+++ b/packages/sanity/src/core/components/hookCollection/GetHookCollectionState.tsx
@@ -9,10 +9,16 @@ import {type ActionHook} from './types'
 
 /** @internal */
 export interface GetHookCollectionStateProps<T, K> {
+  /**
+   * Arguments that will be received by the action hooks, `onComplete` will be added by the HookStateContainer component.
+   */
   args: T
   children: (props: {states: K[]}) => ReactNode
-  hooks: ActionHook<T, K>[]
+  hooks: ActionHook<T & {onComplete: () => void}, K>[]
   onReset?: () => void
+  /**
+   * Name for the hook group. If provided, only hooks with the same group name will be included in the collection.
+   */
   group?: string
 }
 
--- a/packages/sanity/src/core/components/hookCollection/GetHookCollectionState.tsx
+++ b/packages/sanity/src/core/components/hookCollection/GetHookCollectionState.tsx
@@ -13,13 +13,14 @@ export interface GetHookCollectionStateProps<T, K> {
   children: (props: {states: K[]}) => ReactNode
   hooks: ActionHook<T, K>[]
   onReset?: () => void
+  group?: string
 }
 
 const throttleOptions: ThrottleSettings = {trailing: true}
 
 /** @internal */
 export function GetHookCollectionState<T, K>(props: GetHookCollectionStateProps<T, K>) {
-  const {hooks, args, children, onReset} = props
+  const {hooks, args, children, group, onReset} = props
 
   const statesRef = useRef<Record<string, {value: K}>>({})
   const [tickId, setTick] = useState(0)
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/components/hookCollection/GetHookCollectionState.tsx
```diff
@@ -13,13 +13,14 @@ export interface GetHookCollectionStateProps<T, K> {
   children: (props: {states: K[]}) => ReactNode
   hooks: ActionHook<T, K>[]
   onReset?: () => void
+  group?: string
```
bjoerge, line 16: Even though it's internal, could we add a comment explaining it's purpose and what we currently use it for?

packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -130,8 +135,34 @@ export const DocumentPanelHeader = memo(
                 <PaneHeaderActionButton key={item.key} node={item} />
               ))}
 
-              <PaneContextMenuButton nodes={contextMenuNodes} key="context-menu" />
-
+              <RenderActionCollectionState
+                actions={actions || []}
+                // @ts-expect-error TODO: fix the document actions
```
bjoerge, line 140: If we're unable to remove this before merging, could this TODO comment elaborate a bit about what the fix is?

packages/sanity/src/structure/panes/document/statusBar/ActionMenuButton.tsx
```diff
@@ -16,16 +16,46 @@ export interface ActionMenuButtonProps {
...
+/**
+ * @internal
+ */
+export function ActionDialogWrapper({
+  actionStates,
+  children,
+  referenceElement,
+}: {
+  actionStates: DocumentActionDescription[]
+  children: ({handleAction}: {handleAction: (idx: number) => void}) => React.ReactNode
```
bjoerge, line 28: We should avoid usage of the default export from react (I thought eslint would catch this tbh)

```suggestion
  children: ({handleAction}: {handleAction: (idx: number) => void}) => ReactNode

```

bjoerge (changes requested):
Just a few smaller things that would be good to fix before merging.

packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx
pedrobonamin: Fixed, thanks! It required to update the ts definition of `GetHookCollectionState`.
This component is adding the `onComplete` action which is missing in the `editStateFor`, that's why the types weren't matching. 
https://github.com/sanity-io/sanity/pull/5933/commits/0fbdbb20c2f7f9a9507026b158a04fd00442088a#diff-d6b1504a2d940142d300df2888f280863ff5ae4c51762fed870301296ec72877

bjoerge (approved):
LGTM, great work @pedrobonamin

PR #5934, opened 2024-03-08
feat(comments): support task comments

hermanwikner:
### Description

This pull request introduces a `type` configuration to the comments feature (`"field" | "task"`) to enable the creation of comments in the upcoming Tasks feature. Additionally, the UI components have been made somewhat less tied to the field comments scenario by moving the previously required `fieldPath` value to an optional `payload` property. Although the UI components are still somewhat tied to the field comments scenario, this change is a step in the right direction towards making them more generic and adaptable to other contexts, such as when creating comments on tasks.

### What to review

- Make sure that field comments works as before

### Notes for release

N/A

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -73,7 +73,7 @@ const config = {
       {
         ignores: {
           componentPatterns: ['motion$'],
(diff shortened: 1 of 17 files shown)


pedrobonamin (approved):
Thanks for this @hermanwikner !

PR #5935, opened 2024-03-08
refactor(tasks): include tasks plugin via config flag

sjelfull:
### Description

This aligns how to enable the tasks feature with how we enable other features, by adding back the `unstable_tasks` flag.

The challenge with the current method is that we have to include untranspiled config in the config by importing it from `packages/sanity/...`. There is no guarantee this will work, however.

### What to review

- That tasks still works in the staging workspace


pedrobonamin (approved):
Thanks @sjelfull !

PR #5938, opened 2024-03-08
feat(core): fetch inspect mode settings from backend.

cngonzalez:
### Description

Changes the key of inspect mode to be a bit simpler and universal for all documents.

### What to review

The new key for this setting, the default setting. The test is not confirmed -- please see the below note.

### Testing

E2E test staged, but can't properly check it until the backend service is out in production.


binoy14 (approved):
makes sense to me, thanks!

PR #5939, opened 2024-03-08
feat(core): store and fetch user settings from backend

cngonzalez:
### Description

We do not currently persist user settings and actions taken in the Studio in a way that follows the user. We store _some_ user choices and actions to browser localStorage which means they do not follow users between desktop and mobile or different browsers. Browsers may also evict localStorage manually or automatically. These settings and actions are then lost and the user may need to reconfigure their experience or accept that things like search history is lost.

Not persisting this data in a universal manner gives a bad user experience since users need to make the same preference choices and settings again and again. Not having this capability also makes it harder for Sanity to develop new features that should follow the user, and we risk either continuing the localStorage limitation in new feature work, or that new initiatives all have to design their own solution for server-side settings and metadata.

### What to review

Every code change in this branch was previously reviewed as part of a separate PR. However, this branch stands as the final state of this feature, and should be given "one last review."

### Testing

Tests were mostly introduced in previous pull requests. This PR updates some of these tests to reflect settings no longer residing in local storage.

### Notes for release

User settings (like desk list sort orders, view modes, and global search history) are now stored securely server-side by Sanity. This means that these settings do not need to be re-selected across devices or browsers, and will be persisted.

### DIFF
--- a/packages/sanity/src/core/store/key-value/backends/resolve.ts
+++ b/packages/sanity/src/core/store/key-value/backends/resolve.ts
@@ -4,5 +4,6 @@ import {memoryBackend} from './memory'
 import {type Backend} from './types'
 
 export function resolveBackend(): Backend {
+  //TODO: add check for "server"
   return supportsLocalStorage ? localStorageBackend : memoryBackend
 }
--- a/packages/sanity/src/core/store/key-value/backends/resolve.ts
+++ b/packages/sanity/src/core/store/key-value/backends/resolve.ts
@@ -1,9 +1,0 @@
-import {supportsLocalStorage} from '../../../util/supportsLocalStorage'
-import {localStorageBackend} from './localStorage'
-import {memoryBackend} from './memory'
-import {type Backend} from './types'
-
-export function resolveBackend(): Backend {
-  //TODO: add check for "server"
-  return supportsLocalStorage ? localStorageBackend : memoryBackend
-}
--- a/packages/sanity/src/core/store/key-value/backends/server.ts
+++ b/packages/sanity/src/core/store/key-value/backends/server.ts
@@ -1,0 +1,89 @@
+import {type SanityClient} from '@sanity/client'
+import DataLoader from 'dataloader'
+import {catchError, from, map, of} from 'rxjs'
+
+import {DEFAULT_STUDIO_CLIENT_OPTIONS} from '../../../studioClient'
+import {type KeyValueStoreValue} from '../types'
+import {type Backend, type KeyValuePair} from './types'
+
+/** @internal */
+export interface ServerBackendOptions {
+  client: SanityClient
+}
+
+/**
+ * One of serveral possible backends for KeyValueStore. This backend uses the
+ * Sanity client to store and retrieve key-value pairs from the /users/me/keyvalue endpoint.
+ * @internal
+ */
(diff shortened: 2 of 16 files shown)


packages/sanity/src/core/store/key-value/backends/resolve.ts
```diff
@@ -4,5 +4,6 @@ import {memoryBackend} from './memory'
 import {type Backend} from './types'
 
 export function resolveBackend(): Backend {
+  //TODO: add check for "server"
```
jtpetty, line 7: nit: is this still relevant?

packages/sanity/src/core/store/key-value/backends/server.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Sanity client to store and retrieve key-value pairs from the /users/me/keyvalue endpoint.
+ * @internal
+ */
+export function serverBackend({client: _client}: ServerBackendOptions): Backend {
+  const client = _client.withConfig(DEFAULT_STUDIO_CLIENT_OPTIONS)
+
+  const keyValueLoader = new DataLoader<string, KeyValueStoreValue | null>(async (keys) => {
+    const value = await client
+      .request<KeyValuePair[]>({
+        uri: `/users/me/keyvalue/${keys.join(',')}`,
```
jtpetty, line 25: this is probably fine now but if we get to where we do not have complete control over the key names, we may want to be more defensive here

test/e2e/tests/desk/documentTypeListContextMenu.spec.ts
```diff
@@ -1,51 +1,77 @@
...
   await page.getByRole('menuitem', {name: 'Sort by Name'}).click()
-  const localStorage = await page.evaluate(() => window.localStorage)
 
-  expect(localStorage[SORT_KEY]).toBe(
-    '{"by":[{"field":"name","direction":"asc"}],"extendedProjection":"name"}',
-  )
+  /*
+   * The network proves to be a bit flaky for this in our CI environment. We will revisit this after release.
+   */
+  // await page.waitForTimeout(10000)
```
jtpetty, line 15: do we have an issue opened for this?

jtpetty (approved):

packages/sanity/src/core/store/key-value/backends/server.ts
binoy14: good point, the API will reject it right now but validating client side early would be nice. Not needed right now tho

packages/sanity/src/core/store/key-value/backends/resolve.ts
cngonzalez: Thank you, good catch! Made a note of it for us internally at SDX-1188.

test/e2e/tests/desk/documentTypeListContextMenu.spec.ts
cngonzalez: SDX-1189! Thank you

binoy14 (approved):

packages/sanity/src/core/store/key-value/backends/server.ts
cngonzalez: Created an issue!

PR #5940, opened 2024-03-08
feat(core): store recent search history in backend

cngonzalez:
### Description

Changes the key of recent searches to fetch from the backend store.

### What to review

Please feel free to review if the files changed here make sense. Testing suites (like recentSearches.test.tsx) pass -- they are now noisier because of some state updates, but other tests in our suit also unfortunately suffer from the same issue.

### Testing

Two existing test files are in play here -- recentSearches.test.tsx and navbar.spec.ts. The e2e test won't pass until the backend is up and running in production.


binoy14 (approved):
makes sense to me, thanks!

PR #5942, opened 2024-03-10
fix(core): allow `_projectId` and `_strengthOnPublish` in templates

ricokahler:
### Description

A simple follow up to #5889 and adds `_projectId` and `_strengthOnPublish`.

This is the PR description from #5889:

> It is currently not possible to create a cross-dataset reference from an initial value template, as the template validation runs and disallows unknown properties on references, where `_dataset` was not an allowed prop.
> 
> Perhaps we should recurse down the type chain to look for a field that extends from `crossDatasetReference`, but this is beyond what we do for other types - eg we do not check for a `reference` type when encountering `_ref`. Leaving this for a future iteration to improve on if we feel it is worth it.

### What to review

- Can actually create a cross-dataset reference from an initial value template

### Testing

I actually tested this one manually and added fields for initial templates in our test studio.

### Notes for release

- Enhances support for reference types in initial value templates by allowing `_projectId` and `_strengthOnPublish` in templates, builds on #5889.

### DIFF
--- a/dev/test-studio/schema/standard/crossDatasetReference.ts
+++ b/dev/test-studio/schema/standard/crossDatasetReference.ts
@@ -188,6 +188,42 @@ export default defineType({
       name: 'crossDatasetSubtype',
       type: 'crossDatasetSubtype',
     },
+    {
+      name: 'initialValueTest',
+      type: 'crossDatasetReference',
+      dataset: 'playground',
+      studioUrl: ({id, type}) => {
+        return type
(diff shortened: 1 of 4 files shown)


rexxars (approved):
🙏

rexxars:
If time, it would be nice to extend [these tests](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/templates/__tests__/resolve.test.ts#L102-L116) to also check for allowed `_projectId` etc.

PR #5943, opened 2024-03-11
fix(core): upgrade and fix types for typescript 5.4

ricokahler:
### Description

The following PR upgrade typescript to version 5.4 and updates our type definitions for better compatibility.

- There was a type error on some of the "widden" types. I'm not too sure why that's there but just to unblock I 'fixed' the type error with a cast and a TODO comment to revisit for later.
- I used the newer typescript 5.0 ["const type"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters) parameters to ensure that we're capturing the correct `type`.

### What to review

Anything I'm missing on these types? Is the cast I added okay for now?

### Testing

I upgraded one of our starter templates [that broke](https://github.com/sanity-io/nextjs-blog-cms-sanity-v3/actions/runs/8186007972/job/22383499811) to typescript 5.4 and pasted the newer type definitions and ran `npm run type-check` and it seemed to unblock the upgrade there.

Before:

CleanShot 2024-03-10 at 20 17 39@2x

After:

CleanShot 2024-03-10 at 20 16 06@2x

### Notes for release

- Fixes a type bug where the `type` for `defineType` and `defineField` was not being captured correctly.

### DIFF
--- a/dev/embedded-studio/package.json
+++ b/dev/embedded-studio/package.json
@@ -18,7 +18,7 @@
     "@types/react": "^18.2.37",
     "@types/react-dom": "^18.2.15",
     "@vitejs/plugin-react": "^4.2.1",
-    "typescript": "^5.2.2",
+    "typescript": "^5.4.2",
     "vite": "^4.5.1"
   }
 }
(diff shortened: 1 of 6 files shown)


stipsan (approved):
🥳

PR #5946, opened 2024-03-11
chore(test): refactor test config, enable vitest and migrate packages to vitest

bjoerge:
### Description
Note: This PR should probably be rebase+merged instead of squash merged since the commits makes sense on their own.

This includes several improvements to our unit test setup:
- Moves dev aliases shared between several components (jest, vite, vitest, etc) to its own `@repo/dev-aliases`-package to make it  a first class dependency (it's been a well-kept secret, tucked away in `./packages` until now).
- Introduces a `@repo/test-config` internal pacakge that contains the test configuration for jest (and now also vitest) so it can be a proper (dev) dependency of packages that needs it. This makes it easier to maintain test configs, and also makes it easier to remove reliance on jest in the future.
- Splits up ci workflows, one that runs jest and one that runs vitest.
- Migrates a few "simple" (in terms of test config complextity) packages to vitest, these should provide a good example of what it takes to migrate other packages as well:
  - @sanity/migrate
  - @sanity/block
  - @sanity/mutator
  - @sanity/schema
  - @sanity/types
  - @sanity/util

In addition I've done some minor cleanups and smaller factors, hopefully leaving the code in a better shape.

### What to review
 The commits here can be reviewed one-by-one.

### Testing
I've made sure the following is still true:
- The pacakges converted to vitest runs the same number as tests as they used to with jest
- You can still run tests without first havign to build (e.g.  make a fresh clone of the repo, install dependencies and then run `pnpm test`)
- You can still run test studio packages without first havign to build (e.g.  make a fresh clone of the repo, install dependencies and then run `pnpm dev`)

Beyond that, this PR should not require any additional tests.

### Notes for release

n/a - internal


binoy14 (approved):

stipsan (approved):
Excellent 💖

stipsan (approved):

PR #5956, opened 2024-03-11
fix(deps): update @sanity/icons 

pedrobonamin:
### Description

Updates @sanity/icons to latest version to include the new `Task` icon.


cngonzalez (approved):
Exciting! Do we expect the CLI templates to use tasks too? I see a few in `packages/@sanity/cli/src/actions` that are at `^2.6`. I don't think it's a big deal (or if maintaining those templates is in our scope) since we have the caret, but just wanted to ensure there'd be no unexpected behavior down the line.

pedrobonamin:
> Exciting! Do we expect the CLI templates to use tasks too? I see a few in `packages/@sanity/cli/src/actions` that are at `^2.6`. I don't think it's a big deal (or if maintaining those templates is in our scope) since we have the caret, but just wanted to ensure there'd be no unexpected behavior down the line.

I don't see any downside on updating also those, will update them.

PR #5958, opened 2024-03-11
fix(tasks): enable check for document action and badge

hermanwikner:
### Description

This pull request makes sure that the document action and badge related to Tasks is only included if the feature is enabled.

### What to review

- Make sure that the document action and badge is only included when the feature is enabled

### Notes for release

N/A


pedrobonamin (approved):
Nice catch, thanks @hermanwikner

PR #5964, opened 2024-03-12
feat(tasks): add open action to sidebar

hermanwikner:
### Description

This pull request:
- Replaces `rightSectionNode` in the navbar with an internal Navbar Actions API to render the Tasks button.
- Adds the Tasks action to the main sidebar.
- Fixes a z-index issue with the Tasks sidebar.

### Notes for release

N/A


pedrobonamin (approved):
Thanks for this @hermanwikner I think it's all ok!
Just thinking that we should decouple the core changes from the tasks changes, maybe   open a PR to `next` with the api changes and once merged we updated this one. 
Or we wait until `edx-1088` is merged to `next`

pedrobonamin (changes requested):
Thanks for this @hermanwikner I think it's all ok!
Just thinking that we should decouple the core changes from the tasks changes, maybe   open a PR to `next` with the api changes and once merged we updated this one. 
Or we wait until `edx-1088` is merged to `next`

PR #5974, opened 2024-03-12
refactor(core): avoid rendering multiple addon dataset providers

hermanwikner:
### Description

This pull request adds a `withAddonDatasetProvider` option to the `tasks` and `comments` plugins to avoid duplicate `AddonDatasetProvider` renderings. With both plugins depending on `AddonDatasetProvider`, using them together currently leads to rendering it twice. Setting `withAddonDatasetProvider` to `false` in `structureTool` and using a new `addonDataset` plugin, we render `AddonDatasetProvider` just once, making it accessible to both plugins.

### What to Review

- Ensure that everything works as normal and that only one instance of `AddonDatasetProvider` is rendered.

### Notes for Release

N/A


juice49:
What would be the use case for setting `withAddonDatasetProvider: true`, instead of using the `addonDataset` plugin?

Sorry if I'm missing something obvious! 🙂

PR #5992, opened 2024-03-13
fix(core): handle no user id in `getUser`

hermanwikner:
### Description

This pull request ensures that we do not proceed with fetching the user with `getUser` in `createUserStore` if there's no user id available. This is added because if there's no user id provided, we know that the fetch will fail. So, we can prevent the request from erroring by returning early instead.

### What to Review

- Make sure that `getUser` works as before.
- Does the change make sense?

### Notes for Release

N/A


pedrobonamin (approved):

ricokahler (approved):

jtpetty (approved):
👍

PR #5993, opened 2024-03-13
chore: update API version for production-ready cellar

cngonzalez:
### Description

Updates the default studio client version to the key-value store's date of release. Since we haven't had a significant change in client behavior, and we generally ask that users use a current date for their own API versioning, this should be a non-breaking change, but is required to query our key-value store per internal guidelines.

### What to review

Is there anything that depended on the old client version? Any unexpected behaviors?

### Testing

All current tests should behave as normal.


jtpetty (approved):
👍

PR #5994, opened 2024-03-13
fix(structure): push taskCreateAction to the end of the actions array

pedrobonamin:
### Description

With the addition of tasks plugin we introduced a new action. `TasksCreateAction` 
This action, because of the way we resolve the plugins structure, is added at the start of the actions list.

Some users are inserting actions into an specific position of the actions array, assuming the first action is the `PublishAction`

Actions array before:
```
  [TasksCreateAction, Publish, Unpublish, ...]
```

Actions array after:
```
  [Publish, Unpublish, ..., TasksCreateAction]
```

This will not break the implementation of users who are doing changes like this:

```
export const resolveDocumentActions: DocumentActionsResolver = (prev) => {
  prev.splice(1, 0, CustomAction)
  prev.splice(2, 0, DividerAction)
  return prev
}

```


pedrobonamin:
Closing this PR, proposed solution in the github issue is a better way to approach this.

PR #5996, opened 2024-03-13
fix(portable-text-editor): fix and test issue with merge block operation

skogsmaskin:
### Description

There is an issue where a certain merge block operation goes rogue when there are multiple editors in a document.

I have pinned this issue down to invalid paths being used when trying to remove the children of a block as part of the transformation happening on the receiving client when a whole block is set through a set patch.

After fixing this, the added test passes (it failed without this fix).

### What to review

That the changes are ok.

### Testing

Should be covered by automatic tests now, but see #EDX-946 for more details on how to manually reproduce it.

### Notes for release
* Fixed a bug with the Portable Text Editor where merging two text blocks could result in data loss in the topmost block when several editors were inside the document.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/utils/applyPatch.ts
+++ b/packages/@sanity/portable-text-editor/src/utils/applyPatch.ts
@@ -250,18 +250,22 @@ function setPatch(editor: PortableTextSlateEditor, patch: SetPatch) {
   } else if (Element.isElement(block) && patch.path.length === 1 && blockPath) {
     debug('Setting block property')
     const {children, ...nextRest} = value as unknown as PortableTextBlock
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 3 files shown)


packages/@sanity/portable-text-editor/src/utils/applyPatch.ts
```diff
@@ -250,18 +250,22 @@ function setPatch(editor: PortableTextSlateEditor, patch: SetPatch) {
...
+    // Set any block properties
     editor.apply({
       type: 'set_node',
       path: blockPath,
       properties: {...prevRest},
       newProperties: nextRest,
     })
+    // Replace the children in the block
+    // Note that children must be explicitly inserted, and can't be set with set_node
+    debug('Setting children')
```
jtpetty, line 264: should this `debug` call be removed?

jtpetty (approved):
LGTM

PR #5997, opened 2024-03-13
fix: prevent key value store from firing setKey for each open subscri…

cngonzalez:
### Description

The KeyValueStore used to emit a setKey event (setting off a PUT call to the backend) for every open subscription. This means we could fire off 3-4 PUT calls for the same key in quick succession. To keep state tidy and make this scalable, I've added a shareReplay function here, which we use in other parts of the studio (especially stores in the resource cache). I've seen no ill effects or stale data, so I think it's a decent solution.

### What to review

Any gotchas with this approach, or any other negative side effects I may be missing. 

### Testing

There are e2e tests for this branch, but no tests specifically for this case. I can add additional tests in the refactor PR for e2e tests.


bjoerge (approved):
Great catch, and great solution!

PR #5998, opened 2024-03-14
feat(tasks): add notification data for tasks document and tasks comments

pedrobonamin:
### Description

This PR adds the necessary changes for the notification context that will be received by the server when a task is created or edited and when a comment is added to a task.

#### Tasks documents
The following `context.notification` data will be added
```
 context = {
  notification: {
    targetContentImageUrl:
      'https://cdn.sanity.work/images/exx11uqh/playground/9f76911628de8342e7976997644fb7462e8cdcce-1500x843.jpg?rect=329,0,843,843&w=96&h=96',
    targetContentTitle: 'Test Book',
    workspaceTitle: 'Staging',
    url: 'http://localhost:3333/staging/?sidebar=tasks&selectedTask=1c5cf082-78eb-4b0f-81b2-6f28217463be&viewMode=edit',
  },
}

```

#### Comments 
When a comment is added into a task, the notification payload will have the same properties as in a field comment and will also include the tasks subscribers ids.

### DIFF
--- a/packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
+++ b/packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
@@ -151,11 +151,8 @@ export function TasksActivityLog(props: TasksActivityLogProps) {
       studioUrl.searchParams.set('viewMode', 'edit')
       studioUrl.searchParams.set('commentId', commentId)
 
-      // We need to update the subscribers list in the task, any user mentioned in the comment should be added
-      // as a subscriber.
       const mentionedUsers = getMentionedUsers(message)
       const subscribers = Array.from(new Set([...(value.subscribers || []), ...mentionedUsers]))
-      onChange(set(subscribers, ['subscribers']))
 
       return {
         documentTitle: value.title || 'Sanity task',
@@ -164,12 +161,14 @@ export function TasksActivityLog(props: TasksActivityLogProps) {
         subscribers: subscribers,
       }
     },
-    [basePath, value?._id, value.title, workspaceTitle, value.subscribers, onChange],
+    [basePath, value?._id, value.title, workspaceTitle, value.subscribers],
   )
 
   const handleCommentCreate = useCallback(
     (message: CommentInputProps['value']) => {
       const commentId = uuid()
+      const notification = handleGetNotificationValue(message, commentId)
+
       const nextComment: CommentCreatePayload = {
         id: commentId,
         type: 'task',
@@ -179,18 +178,24 @@ export function TasksActivityLog(props: TasksActivityLogProps) {
         status: 'open',
         threadId: uuid(),
         context: {
-          notification: handleGetNotificationValue(message, commentId),
+          notification,
         },
       }
 
+      onChange(set(notification.subscribers, ['subscribers']))
+
       operation.create(nextComment)
     },
-    [operation, handleGetNotificationValue],
+    [operation, handleGetNotificationValue, onChange],
   )
 
   const handleCommentReply = useCallback(
     (nextComment: CommentBaseCreatePayload) => {
       const commentId = uuid()
+
+      const notification = handleGetNotificationValue(nextComment.message, commentId)
+
+      onChange(set(notification.subscribers, ['subscribers']))
 
       operation.create({
         id: commentId,
@@ -201,11 +206,11 @@ export function TasksActivityLog(props: TasksActivityLogProps) {
         status: 'open',
         threadId: nextComment.threadId,
         context: {
-          notification: handleGetNotificationValue(nextComment.message, commentId),
+          notification,
         },
       })
     },
-    [operation, handleGetNotificationValue],
+    [operation, handleGetNotificationValue, onChange],
   )
 
   const handleCommentCreateRetry = useCallback(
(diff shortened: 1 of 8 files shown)


packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
```diff
@@ -134,39 +137,75 @@ type Activity =
...
   const currentUser = useCurrentUser()
+  const {title: workspaceTitle, basePath} = useWorkspace()
 
   const {comments, mentionOptions, operation, getComment} = useComments()
   const loading = comments.loading
   const taskComments = comments.data.open
 
+  const handleGetNotificationValue = useCallback(
+    (message: CommentInputProps['value'], commentId: string) => {
+      const studioUrl = new URL(`${window.location.origin}${basePath}/`)
```
jtpetty, line 148: Is it common for studio URLs to change? I guess it would only be stale for older tasks.

packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
```diff
@@ -134,39 +137,75 @@ type Activity =
...
+      studioUrl.searchParams.set('sidebar', 'tasks')
+      studioUrl.searchParams.set('selectedTask', value?._id)
+      studioUrl.searchParams.set('viewMode', 'edit')
+      studioUrl.searchParams.set('commentId', commentId)
+
+      // We need to update the subscribers list in the task, any user mentioned in the comment should be added
+      // as a subscriber.
+      const mentionedUsers = getMentionedUsers(message)
+      const subscribers = Array.from(new Set([...(value.subscribers || []), ...mentionedUsers]))
+      onChange(set(subscribers, ['subscribers']))
```
hermanwikner, line 158: I find it confusing that a function named `handleGet*` also performs a mutation on the task document. Could we consider patching the subscribers array with the mentioned users at the moment a comment is created or updated instead? For example, in the `handleCommentCreate` and `handleCommentReply` like this:

```ts
const handleCommentCreate = useCallback(
  (message: CommentInputProps['value']) => {
    const commentId = uuid()

    const nextComment: CommentCreatePayload = {
      id: commentId,
      type: 'task',
      message,
      parentCommentId: undefined,
      reactions: EMPTY_ARRAY,
      status: 'open',
      threadId: uuid(),
      context: {
        notification: handleGetNotificationValue(message, commentId),
      },
    }

    // 1. Create the comment
    operation.create(nextComment)

    // 2. Get the mentioned users in the message
    const mentionedUsers = getMentionedUsers(message)

    // 3. If there are mentioned users, update `subscribers` in the task document
    if (mentionedUsers.length > 0) {
      const subscribers = [...(value.subscribers || []), ...mentionedUsers]
      const uniqueSubscribers = Array.from(new Set(subscribers))
      onChange(set(uniqueSubscribers, ['subscribers']))
    }
  },
  [handleGetNotificationValue, operation, value.subscribers, onChange],
)
```

packages/sanity/src/tasks/src/tasks/components/form/tasksFormBuilder/TasksNotificationTarget.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const notificationTarget: TaskContext['notification'] = {
+      url: studioUrl.toString(),
+      workspaceTitle,
+      targetContentImageUrl: imageUrl,
+      targetContentTitle: targetContentTitle,
+    }
+    if (deepEquals(notificationTarget, context)) return
+
+    onChange(set(notificationTarget, ['notification']))
```
hermanwikner, line 56: Why do we want to update the task document with the notification value?

packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
pedrobonamin: I could extract that piece. 
we could have the following:

```
  const handleCommentCreate = useCallback(
    (message: CommentInputProps['value']) => {
      const commentId = uuid()
      const notification = handleGetNotificationValue(message, commentId)
      const nextComment: CommentCreatePayload = {
        id: commentId,
        type: 'task',
        message,
        parentCommentId: undefined,
        reactions: EMPTY_ARRAY,
        status: 'open',
        threadId: uuid(),
        context: {notification},
      }

      operation.create(nextComment)
      onChange(set(notification.subscribers, ['subscribers']))
    },
    [operation, handleGetNotificationValue, onChange],
```

pedrobonamin: I think the task document change should happen before the comment is created, as in the future we may want to get the task values directly in the webhook by resolving that reference.

packages/sanity/src/tasks/src/tasks/components/form/tasksFormBuilder/TasksNotificationTarget.tsx
```diff
@@ -0,0 +1,80 @@
...
+    studioUrl.searchParams.set('selectedTask', _id)
+    studioUrl.searchParams.set('viewMode', 'edit')
+
+    const notificationTarget: TaskContext['notification'] = {
+      url: studioUrl.toString(),
+      workspaceTitle,
+      targetContentImageUrl: imageUrl,
+      targetContentTitle: targetContentTitle,
+    }
+    if (deepEquals(notificationTarget, context)) return
```
hermanwikner, line 54: I believe this comparison will always yield `false` because it compares the next notification object (`notificationTarget`) with `context`, which has the current notification object. I think the correct check should be:

```ts
 if(isEqual(notificationTarget, context?.notification)) return
```

That is, to use `context?.notification` instead of just `context` when comparing.

pedrobonamin: You are right

jtpetty (approved):

PR #6000, opened 2024-03-14
fix: update tests to match api

cngonzalez:
### Description

Updates previous e2e tests to run successfully.

### What to review

The changed test files.

### Testing

Unfortunately with multiple runners and state persisting outside the test environment, search results are not tracked as well as they used to be. However, their cases are covered in recentSearches.test.tsx


binoy14 (commented):
See some test failures not sure if that is expected. Let me know if I can help in any way

binoy14 (approved):

PR #6004, opened 2024-03-15
fix: ensure that useStructureToolSetting updates only when necessary

cngonzalez:
### Description

We should ensure we're avoiding unnecessary calls to the backend whenever possible, as well as any unnecessary re-renders. This PR adds a check to ensure actions only happen in the case of new state.

### What to review

The check and any gotchas for loops or side effects.

### Testing

This util is tested in an E2E test that is currently being refactored.


jtpetty (approved):
👍

PR #5981, opened 2024-03-12
feat(codegen): add typegen methods to codegen package

sgulseth:
### Description

This PR adds type generation method given a set of TypeNodes. It can generate type nodes from either a schema, or a TypeNode and return typescript types as a string.

* `readSchema.ts`: We don't provide any validation that the schema read is correct atm. That should be fine for now, as all schemas should be generated by the schema extract cmd.

### Testing

We should have good coverage on the type generation methods

### Notes for release

N/A - No notes needed

PR #5979, opened 2024-03-12
feat(codegen): add codegen skeleton package

sgulseth:
### Description

This PR introduces a skeleton for the new codegen package, it should just introduce the infrastructure and tooling for the package.
Features are introduced in the subsequent stacked PRs.

I will wait to merge the PRs until all prs are approved and ready ✅ 

### What to review

* Package setup and general configuration

### Testing

N/A

### Notes for release

N/A - not to include in release notes

PR #6008, opened 2024-03-15
feat(codegen): expose referenced type as hidden symbol

sgulseth:
### Description

Adds a symbol so that we can lookup which type a reference is referencing to.

Types generating ends up like:
```ts
{
      _ref: string;
      _type: "reference";
      _weak?: boolean;
      [internalGroqTypeReferenceTo]?: "sanity.imageAsset";
}

...

export declare const internalGroqTypeReferenceTo: unique symbol;
```

### Testing

No specific tests added for this as the feature is marked as "internal". We should build out general integration testing for codegen, but that can be done separately.

### Notes for release

N/A - no notes needed.

### DIFF
--- a/packages/@sanity/codegen/src/typescript/__tests__/__snapshots__/typeGenerator.test.ts.snap
+++ b/packages/@sanity/codegen/src/typescript/__tests__/__snapshots__/typeGenerator.test.ts.snap
@@ -1,5 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`generateSchemaTypes can generate well known types 1`] = `"export declare const internalGroqTypeReferenceTo: unique symbol;"`;
+
 exports[`generateSchemaTypes generateTypeNodeTypes should be able to generate types for type nodes: boolean 1`] = `"export type test_2 = boolean;"`;
 
 exports[`generateSchemaTypes generateTypeNodeTypes should be able to generate types for type nodes: null 1`] = `"export type test_5 = null;"`;
@@ -30,9 +32,11 @@ export type Post = {
   author?: {
     _ref: string;
     _weak?: boolean;
+    [internalGroqTypeReferenceTo]?: \\"author\\";
   } | {
     _ref: string;
     _weak?: boolean;
+    [internalGroqTypeReferenceTo]?: \\"ghost\\";
   };
   slug?: Slug;
   excerpt?: string;
@@ -41,6 +45,7 @@ export type Post = {
     asset: {
       _ref: string;
       _weak?: boolean;
+      [internalGroqTypeReferenceTo]?: \\"sanity.imageAsset\\";
     };
     caption?: string;
     attribution?: string;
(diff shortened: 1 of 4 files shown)


sgulseth:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6008?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#6008** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6008?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#5982** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5981** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5981?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5980** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5980?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5979** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/5979?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `next`

<h2></h2>

ricokahler (approved):
I like this pattern. It's very akin to the use of `Symbol.iterator` etc

I wonder if we should declare some kind of object to act as a symbol namespace. e.g. 

```ts
declare const CodegenSymbol = {
  referenceTo: unique symbol
}
```

Then usage could be like:

```ts
interface SomeType {
  _type: 'ref';
  [CodegenSymbol.referenceTo]?: 'sanity.imageAsset';
  // ...
}
```

But not a blocker. Just an idea. It may not even work because the object implies a runtime value.

ricokahler:
Alternative idea: you could export a type helper from the `@sanity/codegen` package that would abstract away this as well.

Instead of:

```ts
type myQueryResult = {
    asset?: {
      _ref: string;
      _type: "reference";
      _weak?: boolean;
      [internalGroqTypeReferenceTo]?: "sanity.imageAsset";
    };
    // ...
}
```

it could be:

```ts
import {type SanityReference} from '@sanity/codegen'

type myQueryResult = {
    asset?: SanityReference<'sanity.imageAsset'>;
    // ...
}
```


```ts
type SanityReference<TReferenceType extends string> {
      _ref: string;
      _type: "reference";
      _weak?: boolean;
      [__internalGroqTypeReferenceTo]?: TReferenceType;
}
```

sgulseth:
I like the I do like the idea! How ever I'm a bit hesitant to add dependencies to the generated types 🤔  I'll discuss it with the team

sgulseth:
@ricokahler We don't want to abstract away the reference type object. Currently the ref-object is just any other object, this is also "future proof" as it allows us to extend the reference type object in the future, if we want to. 

However, importing the symbol might be a requirement for it to be used with other tools? 🤔 A potential issue however is that ie. eslint could start to complain: 
image

pnpm is also strict about this, iirc

PR #6009, opened 2024-03-15
feat(tasks): add comment delete confirm dialog

hermanwikner:
### Description

This pull request implements a confirmation dialog for deleting task comments to prevent accidental deletion of comments.

### What to review

- Make sure that the dialog appear and that the deletion is successfully executed

### Notes for release

N/A


pedrobonamin (approved):
Thanks @hermanwikner !

Given we are including more things that are comment specific logic into `TaskActivityLog` What do you think if we move the new addition and [this component](https://github.com/sanity-io/sanity/pull/6009/files#diff-c93e4a7d1a1fea369a7aad1f7e1272055ea65e4267b43e85fae96167638580deR360-R381) to it's own file, maybe: `TasksActivityCommentItem.tsx`  ?

hermanwikner:
> Thanks @hermanwikner !
> 
> Given we are including more things that are comment specific logic into `TaskActivityLog` What do you think if we move the new addition and [this component](https://github.com/sanity-io/sanity/pull/6009/files#diff-c93e4a7d1a1fea369a7aad1f7e1272055ea65e4267b43e85fae96167638580deR360-R381) to it's own file, maybe: `TasksActivityCommentItem.tsx` ?

Thanks @pedrobonamin! Good suggestion, I pushed a commit that creates a dedicated component for the comment item

PR #6014, opened 2024-03-15
fix: schema extraction object type name

sgulseth:
### Description

For some objects(I belive it's nested objects, but yet to verify) we sometimes generate object schema types with name set to `object`. This PR removes the _type if it's object, as it's better to miss data than to be wrong.

One example from the tests are `Homestead`, the compiled schema type look like:
```ts
{
  jsonType: 'object',
  type: { name: 'object', title: 'Object', type: null, jsonType: 'object' },
  name: 'object',
  title: 'Homestead',
  fields: [
    {
      name: 'lat',
      group: undefined,
      fieldset: undefined,
      type: [Object]
    },
    {
      name: 'lon',
      group: undefined,
      fieldset: undefined,
      type: [Object]
    }
  ],
  options: {},
  orderings: [
    { name: 'lat', i18n: [Object], title: 'Lat', by: [Array] },
    { name: 'lon', i18n: [Object], title: 'Lon', by: [Array] }
  ],
  fieldsets: [Getter],
  groups: [Getter],
  preview: [Getter]
}
```

However the field definition is:
```ts
{
  name: 'homestead',
  title: 'Homestead',
  type: 'object',
  fields: [
    {
      name: 'lat',
      title: 'Latitude',
      type: 'number',
      required: true,
    },
    {
      name: 'lon',
      title: 'Longitude',
      type: 'number',
      required: true,
    },
  ],
},
```

### Notes for release

N/A - no notes needed

### DIFF
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -219,7 +219,7 @@ export function extractSchema(
       return {type: 'unknown'} satisfies UnknownTypeNode
     }
 
-    if (schemaType.type?.name !== 'document') {
+    if (schemaType.type?.name !== 'document' && schemaType.name !== 'object') {
       attributes._type = {
         type: 'objectAttribute',
         value: {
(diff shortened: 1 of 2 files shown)


bjoerge (approved):
This looks correct to me. Values of schema types with {type: 'object'} should not get `{_type: 'object'}` (as it's implied). Good catch!

PR #6016, opened 2024-03-15
Fix bug where we cant resolve custom types

sgulseth:
### Description

Found a bug where we some times can't extract a schemaType, turns out we were resolving the wrong schema type.

Types defined like this doesn't work:
```
defineType({
  name: 'customStringType',
  title: 'My custom string type',
  type: 'string',
}),
```

### What to review

Correctnes

### Testing

Added 👍 

### Notes for release

N/A - no notes needed


ricokahler (approved):
Nice. It's actually very common for us to forget about the type chain too 😅 but I think this will do the trick! 👏

bjoerge (approved):

sgulseth:
Types inside types inside types is confusing to reason about 😅

PR #6020, opened 2024-03-18
fix(tasks): hide safari details marker

pedrobonamin:
### Description

Hides in safari the details marker element. 
### Before
Screenshot 2024-03-18 at 09 39 06

### After
Screenshot 2024-03-18 at 09 38 07


jtpetty (approved):
🙏

PR #6034, opened 2024-03-18
fix: use more reliable comparison in deduping search results

cngonzalez:
### Description

Since we moved the recent search results to a server-side service, we started seeing duplicate results in the global search results. The previous implementation, which stored search results in local storage, stringified recent searches before storing them and compared stringified new searches against them.

Since we switched to using a backend service, we could no longer rely on keys being in the same order. This PR uses the `isEqual` lodash method (used in many other places in the studio) to more reliably compare the values of the objects, rather than a string result.

### What to review
The changed file. Any performance or other repercussions from using this comparison.

### Testing
Tests already exist for this utility, but did not catch this because of mocking. A ticket has been created to better address this in the future. 

### Notes for release
Resolves an issue where the server-side storage of global recent searches was storing duplicate recent searches. Recent searches should now all be unique terms.


skogsmaskin (approved):
LGTM!

PR #6057, opened 2024-03-19
fix(comments): permission check

hermanwikner:
### Description

This pull request ensures that the user has `read` permissions on the current document to determine whether the comments UI should be displayed in fields. The add-on dataset permission allows users with `read` access to both read and write comments, making it the only necessary permission to check.

### What to Review

- Ensure the permission check functions correctly by testing with various roles.

### Notes for Release

N/A


jtpetty (approved):
LGTM

PR #6065, opened 2024-03-19
feat: indicate in metadata that the quickstart option was used in project creation

svirs:
### Description

Project metadata.template should reflect if the --quickstart flag was used

### Testing

Project metadata should show `cli-quickstart` rather than `cli-clean` if the --quickstart flag was used for pulling down a project and its schema

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -964,7 +964,6 @@ export default async function initSanity(
 
   function selectProjectTemplate() {
     // Make sure the --quickstart and --template are not used together
-    // Force template to clean if --quickstart is used
     if (flags.quickstart) {
       return 'quickstart'
     }
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -966,7 +966,7 @@ export default async function initSanity(
     // Make sure the --quickstart and --template are not used together
     // Force template to clean if --quickstart is used
     if (flags.quickstart) {
-      return 'clean'
+      return 'quickstart'
     }
 
     const defaultTemplate = unattended || flags.template ? flags.template || 'clean' : null
--- a/packages/@sanity/cli/templates/quickstart/README.md
+++ b/packages/@sanity/cli/templates/quickstart/README.md
@@ -1,0 +1,9 @@
+# Sanity Clean Content Studio
+
+Congratulations, you have now installed the Sanity Content Studio, an open source real-time content editing environment connected to the Sanity backend.
+
+Now you can do the following things:
+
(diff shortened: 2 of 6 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -963,10 +963,9 @@ export default async function initSanity(
   }
 
   function selectProjectTemplate() {
-    // Make sure the --quickstart and --template are not used together
-    // Force template to clean if --quickstart is used
+    // --quickstart should override template selection
     if (flags.quickstart) {
-      return 'clean'
+      return 'quickstart'
```
RostiMelk, line 968: This is going to break as there is no template called "quickstart". 
The `--quickstart `flag relies on the "clean" template, and modifies the template with its remote schema.

RostiMelk (changes requested):
I suggest looking into how we import/get a template here:
https://github.com/sanity-io/sanity/blob/174a61641312734ca29d57803119ca3dfad0b78e/packages/@sanity/cli/src/actions/init-project/initProject.ts#L513

If the `templateName` variable is the name that is being sent to the telemetry service, you may get around this by adding an alias "quickstart" for the "clean" template here:
https://github.com/sanity-io/sanity/blob/b14622ed44899ad6da7bb0fd1d46f56c0186a04b/packages/@sanity/cli/src/actions/init-project/templates/index.ts#L9
LIke so:
```ts
const templates: Record<string, ProjectTemplate | undefined> = {
  blog,
  clean,
  'get-started': getStartedTemplate,
  moviedb,
  shopify,
  'shopify-online-storefront': shopifyOnline,
  quickstart: clean // New alias here
}
```

packages/@sanity/cli/templates/quickstart/README.md
```diff
@@ -0,0 +1,9 @@
+# Sanity Clean Content Studio
```
RostiMelk, line 1: Should we change this title as well while we're at it?

joneidejohnsen, line 1:
```suggestion
# Sanity Studio
```

packages/@sanity/cli/templates/quickstart/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+Congratulations, you have now installed the Sanity Content Studio, an open source real-time content editing environment connected to the Sanity backend.
```
joneidejohnsen, line 3:
```suggestion
Congratulations, you have now installed Sanity Studio, an open source real-time content editing environment connected to the Sanity Content Lake backend.
```

joneidejohnsen (changes requested):
Suggested copy changes

jtpetty (approved):

joneidejohnsen (approved):

drewlyton (approved):

PR #6074, opened 2024-03-19
Render ts-types with vision query

sgulseth:
### Description

image

Added a view for rendering typescript types

### What to review

Code readability, I tried to follow what was the established style but lmk if I should update anything.

We already have a ts generator in @sanity/codegen, however this is based on babel and doesn't work in-browser. I implemented a simple typenode -> typescript structure

### Notes for release

N/A - no notes needed


packages/@sanity/vision/src/components/QueryTypesContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <Tab
+            aria-controls="typescript-lang"
+            id="content-tab"
+            label={props.t('TypeScript')}
+            onClick={() => setLang('typescript')}
+            selected={lang === 'typescript'}
+          />
+          <></>
+        </TabList>
+      </LangContainer>
```
rexxars, line 50: I see where this is going (and this is great), but until we have more than one language, this looks kind of out of place. Maybe we can just hide this for now?

packages/@sanity/vision/src/components/QueryTypesContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  return (
+    <QueryTypesWrapper overflow="auto" paddingY={3}>
+      <LangContainer>
+        <TabList space={2}>
+          <Tab
+            aria-controls="typescript-lang"
+            id="content-tab"
+            label={props.t('TypeScript')}
```
rexxars, line 44: By doing this, language packs can override the `types.language_typescript` resource to translate this albel.

```suggestion
            label={props.t('types.language', {context: 'typescript'})}
```

packages/@sanity/vision/src/components/QueryTypesContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+  typeNode?: TypeNode
+  t: TFunction<'vision', undefined>
+}) {
+  const [lang, setLang] = useState<'typescript' | 'swift' | 'go'>('typescript')
+  if (!props.typeNode) {
+    return null
+  }
+
+  return (
+    <QueryTypesWrapper overflow="auto" paddingY={3}>
```
rexxars, line 38: I think we'll want a label here to mirror the other panes.

```suggestion
    <QueryTypesWrapper overflow="auto" paddingY={3}>
    <StyledLabel muted>{props.t('types.label')}</StyledLabel>
```

And we'll want to add a resource to the end of `src/i18n/resources.ts`:

```tsx
  /** Label for the generated query types pane */
  'types.label': 'Types',
```

packages/@sanity/vision/src/components/QueryTypesContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {generateTypescript(props.typeNode)}
+      </Code>
+    )
+  }
+
+  return null
+}
+
+function generateTypescript(typeNode: TypeNode, indent: number = 0): string {
+  const indentStr = (i: number) => '  '.repeat(i + 1)
```
rexxars, line 74: I think this should be zero-indexed?
Currently everything but the first opening `{` is indented with 4 spaces instead of 2.

```suggestion
  const indentStr = (i: number) => '  '.repeat(i)
```

rexxars (changes requested):

packages/@sanity/vision/src/components/QueryTypesContainer.tsx
sgulseth: Sure. It was mostly because I liked the heading, as it gave some context of the view. Ie your query is not matching your schema it might just say `nulm`

packages/@sanity/vision/tsconfig.json
```diff
@@ -8,5 +8,9 @@
 
     "jsx": "react-jsx"
   },
-  "references": [{"path": "../types/tsconfig.lib.json"}, {"path": "../../sanity/tsconfig.lib.json"}]
+  "references": [
+    {"path": "../schema/tsconfig.json"},
+    {"path": "../types/tsconfig.json"},
+    {"path": "../../sanity/tsconfig.json"}
```
sgulseth, line 14: I changed this from importing *lib* to not, I think this is correct?

sgulseth:
image

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -587,6 +591,9 @@ export class VisionGui extends PureComponent<VisionGuiProps, VisionGuiState> {
 
     const queryStart = Date.now()
 
+    const ast = parse(query)
```
rexxars, line 594: This needs to be try/catched - if you have a syntax error or refer to unknown functions (say, `array::distinct` instead of `array::unique`, the tool throws an uncaught error toast and gets into an odd state where it appears to be executing the query, but is actually stalled.

rexxars: Should probably also catch the line below, in case that can throw too.

packages/@sanity/vision/tsconfig.json
rexxars: 🤔 I think that's right, yes.

rexxars (changes requested):

sgulseth:
hmm, that's a good point.

One alternative is to pass down the workspace object to VisionGui, and only extract the schema, and display types, if the selected dataset matches the workspace dataset? 🤔

sgulseth:
Types pane can now be toggled:

image
image

sgulseth:
If the workspace dataset doesn't match the selected dataset the types are not generated and the container is hidden :poofff:

rexxars (changes requested):
- Label is cut off at the top

image

- Can we add a question mark next to the types label (as with perspectives, see screenshot below) that briefly explains that these are TypeScript types generated _based on the defined studio schema_ etc? A link to some documentation might be nice too. Could potentially mention that types are not guaranteed to match the result, but that may be documented in a linked article potentially.



image

- Also needs a rebase

rexxars (changes requested):
Discussed this internally - we'd like to defer the computation of the types to when the types panel is open, in order to not unnecessarily cause performance degradations for users that do not see the types.

PR #6076, opened 2024-03-20
fix(tasks): update how addonDatasetProvider ready state is set

pedrobonamin:
### Description

Fixes an issue in which when trying to create the addon dataset through the tasks plugin, this was not being created.
This is caused because the `ready: true` value was only set if the dataset name exists.

By updating this, the addon provider will return `ready:true` after fetching the dataset, either it exists or not. 
This flag is used by [`<TasksAddonWorkspaceProvider />`](https://github.com/sanity-io/sanity/blob/fix-task-addon-creation/packages/sanity/src/tasks/src/tasks/components/form/addonWorkspace/TasksAddOnWorkspaceProvider.tsx#L68) to decide if it should or not create the addon dataset.


jtpetty (approved):
👍

PR #6078, opened 2024-03-20
refactor: use `PKG_VERSION` from `@sanity/pkg-utils` instead of manually regexing `SANITY_VERSION`

stipsan:


bjoerge (approved):
Thank you sir <3

rexxars (approved):

PR #6079, opened 2024-03-20
feat(sanity): use actions API when publishing documents

juice49:
### Description

Note: I've skipped, rather than removed, the redundant unit tests for now. This is in case we keep the existing functionality behind a feature flag, and saves us regenerating test snapshots for now. I've made the implementation changes directly so it's easier to follow the changes (if we decided to feature flag this, we can easily grab the existing implementation from git).

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts
@@ -1,25 +1,5 @@
-import {isReference} from '@sanity/types'
-import {omit} from 'lodash'
-
 import {isLiveEditEnabled} from '../utils/isLiveEditEnabled'
 import {type OperationImpl} from './index'
-
-function strengthenOnPublish<T = any>(obj: T): T {
-  if (isReference(obj)) {
-    if (obj._strengthenOnPublish) {
-      return omit(
-        obj,
-        ['_strengthenOnPublish'].concat(obj._strengthenOnPublish.weak ? [] : ['_weak']),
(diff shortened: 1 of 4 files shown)


juice49:
Sorry, converted back to a draft as the tests need some attention.

packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts
```diff
@@ -33,42 +13,23 @@ export const publish: OperationImpl<[], DisabledReason> = {
...
-    }
-
-    tx.delete(idPair.draftId)
-
-    return tx.commit({tag: 'document.publish', visibility: 'async'})
+  execute: ({client: globalClient, idPair}) => {
+    const vXClient = globalClient.withConfig({apiVersion: 'X'})
+    const {dataset} = globalClient.config()
+
+    return vXClient.observable.request({
```
cngonzalez, line 20: I'm guessing we don't know which errors we'll get if the draft is deleted on the server? Is there anything else needed to ensure how this errors and throw if needed, as in the original version?

cngonzalez (approved):

packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts
bjoerge: Good point! I've asked about status on response format [here](https://linear.app/sanity/issue/CLDX-1842#comment-d763592d)

bjoerge (approved):

PR #6081, opened 2024-03-20
feat(core): implement presence cursors

hermanwikner:
### Description

This pull request introduces presence cursors in the Portable Text Input. The implementation extends the presence data model with a `selection` property, which gets updated when a user interacts with the content in the editor. This value is later used to render presence cursors using the Range Decoration API at each user's current selection.

Screenshot 2024-04-11 at 15 59 24

### What to review

- Ensure that the presence cursors are rendered in the correct position.
- Does the addition of `selection` to the presence data model make sense?
- Does the new payload parameter on the focus callbacks make sense? Should it be typed differently than `Record<string, unknown>`?

### Notes for release

Implements presence cursors to see where others are in Portable Text Input content.

(Note: the image in this pull request is just created to illustrate the feature in the context of this pull request but may not be used in other contexts)


packages/sanity/src/core/form/components/ActivateOnFocus/ActivateOnFocus.styles.tsx
```diff
@@ -30,6 +30,7 @@ export const FlexContainer = styled(Flex)`
   left: 0;
   width: 100%;
   height: 100%;
+  z-index: 9999;
```
hermanwikner, line 33: Since the "cursor dots" are rendered in a portal – this z-index change is necessary to make sure that they are not rendered on top of the activate overlay

pedrobonamin (commented):
This is soooo great! It's looking awesome 🚀 
Left some comments.

I was seeing some flickering in the dot,  so did a test to remove the Popover and I think it worked! Let's discuss it tomorrow.
Opened a Pr to make it easier to review the differences https://github.com/sanity-io/sanity/pull/6378

packages/sanity/src/core/form/studio/contexts/FormCallbacks.tsx
```diff
@@ -8,7 +8,7 @@ import {type FormPatch, type PatchEvent} from '../../patch'
 export interface FormCallbacksValue {
   transformPatches?: (patches: FormPatch[]) => FormPatch[]
   onChange: (patchEvent: PatchEvent) => void
-  onPathFocus: (path: Path) => void
+  onPathFocus: (path: Path, payload?: Record<string, unknown>) => void
```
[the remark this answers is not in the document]
ricokahler: Is allowing the path focus handler to accept a payload a new feature you're adding? I would agree with pedro or even mark it as just `unknown` and creating and exporting a type guard when trying to consume values from the payload if it's something you envision being very generic. Not a blocking comment from me though.

ricokahler (approved):
The code looks great!

My only (very general) question to you is do you anticipate any performance losses with this one? I only ask since this seems to be within a highly active and re-rendering component. If so, are there ways to improve the performance?

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/UserPresenceCursor.tsx
```diff
@@ -0,0 +1,212 @@
...
+    --presence-cursor-bg: ${bg};
+
+    &[data-debug-hover-target='true'] {
+      outline: 1px solid magenta;
+    }
+  `
+})
+
+const CursorLine = styled.span<StyledProps>(({theme, $tints}) => {
+  const isDark = theme.sanity.color.dark
```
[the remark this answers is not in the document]
hermanwikner: Thanks, will update!

pedrobonamin (approved):
Thanks for this @hermanwikner ! Let's see if someone mentions the flickering and we can try to iterate again. Did some new tests with the case you sent me and what I tested is not working.

pedrobonamin (approved):

packages/sanity/src/core/form/studio/contexts/FormCallbacks.tsx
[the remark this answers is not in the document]
hermanwikner: Thank you both! I improved the typing by introducing a new interface called `OnPathFocusPayload`

packages/sanity/src/core/form/types/inputProps.ts
```diff
@@ -226,7 +234,7 @@ export interface ArrayOfObjectsInputProps<
   /**
    * @hidden
    * @beta */
-  onPathFocus: (path: Path) => void
+  onPathFocus: (path: Path, payload?: OnPathFocusPayload) => void
```
bjoerge, line 237: Instead of adding a second argument, would it be more future proof to introduce a separate event type? I.e. change the signature of `onPathFocus(path: Path)` to `onPathFocus(event: InputFocusEvent | Path)`.

It would align with `PatchEvent` and give us an opportunity to add more properties to the event type later (instead of adding yet more positional arguments). If we did, we could deprecate the `onPathFocus(path: Path)` signature.

skogsmaskin (changes requested):
Something happened to the focus tracking (for example when clicking a validation error for PT-content) in this branch. We need to resolve this before merging.

packages/sanity/src/core/form/types/inputProps.ts
skogsmaskin: @bjoerge - I think that sounds like a good idea. And we could individually tag it as @beta or whatever.

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -145,26 +160,35 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
...
-        ) // Not an inline object (it would be a member in this list, where spans are not). By doing this check we avoid depending on the value.
-      if (isSpanPath) {
-        // Append `.text` to the focusPath
-        onPathFocus(focusPath.concat('text'))
-      } else {
-        // Call normally
-        onPathFocus(focusPath)
-      }
+        )
+
```
skogsmaskin, line 176: Can we please have this with explicit comments as it were? It's explaining the "hack".

skogsmaskin:
There seems to be some weirdness happening to the editor's selection. It's being reset by something. Try to select all text for example.

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export function usePresenceCursorDecorations(
+  props: PresenceCursorDecorationsHookProps,
+): RangeDecoration[] {
+  const {path} = props
+  const fieldPresence = useFormFieldPresence()
+  const [currentPresence, setCurrentPresence] = useState<FormNodePresence[]>([])
+  const [presenceCursorDecorations, setPresenceCursorDecorations] = useState<RangeDecoration[]>([])
+  const previousPresence = useRef(currentPresence)
+  const handleRangeDecorationMoved = useCallback((details: RangeDecorationOnMovedDetails) => {
```
hermanwikner, line 27: Nitpick: add space above this function

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export function usePresenceCursorDecorations(
+  props: PresenceCursorDecorationsHookProps,
+): RangeDecoration[] {
+  const {path} = props
+  const fieldPresence = useFormFieldPresence()
+  const [currentPresence, setCurrentPresence] = useState<FormNodePresence[]>([])
+  const [presenceCursorDecorations, setPresenceCursorDecorations] = useState<RangeDecoration[]>([])
+  const previousPresence = useRef(currentPresence)
```
hermanwikner, line 26: Nitpick: add typing to this `useRef`

```ts
const previousPresence = useRef<FormNodePresence[]>(currentPresence)
```

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
```diff
@@ -0,0 +1,79 @@
...
+  useEffect(() => {
+    const result = fieldPresence.filter((p) => startsWith(path, p.path) && !isEqual(path, p.path))
+    // Test is the selection and sessionId are the same as last time, if it is we don't need to update
+    if (
+      !isEqual(
+        result.map((d) => ({...d.selection, sessionId: d.sessionId})),
+        previousPresence.current.map((d) => ({...d.selection, sessionId: d.sessionId})),
+      )
+    ) {
+      const value = result.length > 0 ? result : EMPTY_ARRAY
```
hermanwikner, line 55: Is this ternary really necessary? I think we can do:

```ts
setCurrentPresence(result)
previousPresence.current = result
```

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }, [fieldPresence, path])
+
+  useEffect(() => {
+    const decorations: RangeDecoration[] = currentPresence.map((presence) => {
+      if (!presence.selection) {
+        return null
+      }
+      // Create a cursor point at the current selection focus
+      const cursorPoint = {focus: presence.selection.focus, anchor: presence.selection.focus}
+      return {
```
hermanwikner, line 68: Nitpick: add space above this return

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
```diff
@@ -0,0 +1,79 @@
...
+        component: () => <UserPresenceCursor user={presence.user} />,
+        selection: cursorPoint,
+        onMoved: handleRangeDecorationMoved,
+        payload: {sessionId: presence.sessionId},
+      }
+    }) as RangeDecoration[]
+    setPresenceCursorDecorations(decorations.filter(Boolean))
```
hermanwikner, line 75: Nitpick: add space above `setPresenceCursorDecorations`

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -172,6 +198,8 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
         case 'mutation':
           onChange(toFormPatches(change.patches))
           break
+        case 'patch':
+          break
```
hermanwikner, line 202: This `case` can be removed since we are not doing anything

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -251,8 +287,17 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
...
+
+  const rangeDecorations = useMemo((): RangeDecoration[] => {
+    const result = [...(rangeDecorationsProp || []), ...presenceCursorDecorations]
+    const reconciled = immutableReconcile(previousRangeDecorations.current, result)
+    previousRangeDecorations.current = reconciled
+    return reconciled
+  }, [presenceCursorDecorations, rangeDecorationsProp])
+
   return (
-    <Box ref={elementProps.ref}>
```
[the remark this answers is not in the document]
skogsmaskin: It's the wrong ref. It was changed to point to the Editable element in another PR.

packages/sanity/src/core/form/inputs/PortableText/presence-cursors/usePresenceCursorDecorations.tsx
skogsmaskin: It is necessary to keep a stable empty array object and hinder unnecessary new state.

skogsmaskin (approved):
LGTM!

PR #6085, opened 2024-03-20
fix(tasks): copy changes for Tasks feature

jtpetty:
### Description

Updates some copy for the Tasks feature and also some remaining strings for i18n

### DIFF
--- a/packages/sanity/src/tasks/src/tasks/components/form/addonWorkspace/taskSchema.tsx
+++ b/packages/sanity/src/tasks/src/tasks/components/form/addonWorkspace/taskSchema.tsx
@@ -17,7 +17,7 @@ const targetContentField = (mode: FormMode) =>
   defineField({
     type: 'object',
     name: 'target',
-    title: 'Target content',
+    title: 'Target',
     components: {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/tasks/src/tasks/components/form/addonWorkspace/taskSchema.tsx
```diff
@@ -17,7 +17,7 @@ const targetContentField = (mode: FormMode) =>
   defineField({
     type: 'object',
     name: 'target',
-    title: 'Target content',
+    title: 'Target',
```
pedrobonamin, line 20: Noticing this, we will need to find a way later to provide translations to this I guess... 
It should be possible to send the `t` function from https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/tasks/src/tasks/components/form/addonWorkspace/TasksAddOnWorkspaceProvider.tsx#L43-L44

jtpetty: yeah, I'll take another pass in a follow up PR

pedrobonamin (approved):

PR #6087, opened 2024-03-20
fix(tasks): update info passed to the tasks comments notification context

pedrobonamin:
When creating a comment inside a task we need to pass a notification context for it.
The data obtained for the context was incorrect because it was pulling from the addon workspace instead the users current workspace. 
This Pr updates this to pull the data from the correct workspace.

It also moves `useActivityLog` to the hooks folder, and uses it from the FormEdit, given that hook needs to have access to the addon workspace.


jtpetty (approved):
LGTM

PR #6089, opened 2024-03-20
fix(tasks): add new sanityTasks flag

pedrobonamin:
### Description
Add new `sanityTasks` flag.
We are introducing a new flag to require this version of the studio in order to get the tasks feature enabled.
We are keeping the staging flag until the new flag is added to production and staging.


packages/sanity/src/tasks/src/tasks/context/enabled/TasksEnabledProvider.tsx
```diff
@@ -12,22 +12,33 @@ interface TaksEnabledProviderProps {
...
+        enabled: true,
+        mode: 'default',
+      }
+    }
 
     return {
-      enabled: true,
-      mode: 'default',
+      enabled: false,
+      mode: null,
```
pedrobonamin, line 39: This in the future will return mode `upsell` that's why it's on a different return

jtpetty (approved):
LGTM

PR #6093, opened 2024-03-20
fix(structure): use correct loading messages while resolving panes

rexxars:
### Description

While the structure panes are resolving (when using async resolvers), we show one or more loading messages. It starts with a "Loading..." message, shown after X ms. Later, after X seconds, it is supposed to show a "Still loading..." followed by a "Check if there's something wrong with your structure definition" kind of thing.

This code path has been broken for a while now, it seems:
1. The i18n changes had an error, so we're actually showing the resource key instead of the translated message for the first loading message
2. After v3, we checked for a subscribe method on what hopefully is an observable of loading messages, but we _returned_ if that was the case, when we actually wanted the opposite. If what was passed did _not_ have a subscribe method, we would actually have crashed here.

This PR fixes both of these cases.

### What to review

- Loading pane shows correct messages (use the workshop story to more easily see it, or create a `child` structure resolver that waits for 10+s before returning the child node

### Testing

I've added a workshop story for this. Ideally I'd also add automated tests, but seems like we're missing some established tests in the structure codebase that I can base them off of. 

### Notes for release

- Fixed incorrect loading message shown while a pane is resolving in the structure tool

### DIFF
--- a/packages/sanity/src/structure/__workshop__/LoadingPaneStory.tsx
+++ b/packages/sanity/src/structure/__workshop__/LoadingPaneStory.tsx
@@ -1,0 +1,19 @@
+import {Card, Stack, Text} from '@sanity/ui'
+
+import {PaneLayout} from '../components'
+import {LoadingPane} from '../panes'
+
+export default function LoadingPaneStory() {
+  return (
+    <Stack space={3}>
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
makes sense to me, thanks!

PR #6097, opened 2024-03-21
fix(structure): remove padding prop in contextMenuButton

ninaandal:
### Description
The `ContextMenuButton` should always be square. But it allowed the `paddingY` prop, which could be set and resulting in it not being square. 
This PR removes the `paddingY` prop from the `ContextMenuButton`. 
Before: 
Screenshot 2024-03-21 at 11 07 22

After: 
Screenshot 2024-03-21 at 11 07 30

### Notes for release
Removes unnecessary padding on various menu buttons


hermanwikner (approved):
Nice! Thanks @ninaandal 🙏

PR #6105, opened 2024-03-21
feat(sanity): use Actions API when restoring drafts

juice49:


juice49:
Seem to be receiving an error from the API when using this action now. However, I believe it was working previously.

PR #6110, opened 2024-03-21
fix(comments): drop permission check

hermanwikner:
### Description

This pull request addresses an issue where the PTE changes from fullscreen to collapsed when content is changed in the editor. This problem was caused by the permission check logic for comments introduced in #6057.

The issue can be resolved by storing the permission result in a `useRef` instead of returning the value directly from `useDocumentValuePermissions` in the `CommentsProvider`. However, since users can both **create** and **read** comments if they only have **read** permissions, this permission check is not entirely necessary. This is because if a user does not have **read** permissions, they would not be able to see the document and, therefore, could not create a comment.

With this consideration, this pull request reverts the permission check logic introduced in #6057, thereby fixing the fullscreen issue in PTE and removing any unnecessary logic in the `CommentsProvider`.

### What to review

- Make sure that the PTE do not collapse when content is edited

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
+++ b/packages/sanity/src/structure/comments/plugin/field/CommentsField.tsx
@@ -92,7 +92,6 @@ function CommentFieldInner(
 
   const {
     comments,
-    hasPermission,
     isCommentsOpen,
     isCreatingDataset,
     mentionOptions,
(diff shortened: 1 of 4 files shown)


juice49 (approved):
> However, since users can both create and read comments if they only have read permissions, this permission check is not entirely necessary. This is because if a user does not have read permissions, they would not be able to see the document and, therefore, could not create a comment.

That seems reasonable to me!

ninaandal (approved):

PR #6112, opened 2024-03-21
fix(cli): add note about webhooks in migration run command

jtpetty:
Fixes SDX-1156

### Description

Adds a message to warn users that migrations will trigger web hooks.

image

### What to review

Is this the right message?

### Testing

No behavioral change

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
@@ -181,11 +181,13 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
 
     output.print(
       `\n${chalk.yellow(
-        chalk.bold(
-          'Note: If configued, migrations will trigger webhooks. You can disable webhooks on https://manage.sanity.io',
-        ),
-      )}\n`,
+        chalk.bold('Note: If your dataset has active webhooks, migrations will trigger them.'),
+      )}`,
     )
+    output.print(
+      `Run ${chalk.cyan('sanity manage')} to open the management interface in your web browser, then go to the API section to turn off webhooks.`,
+    )
+    output.print(`You can safely turn them back on after completing the migration.\n`)
 
     const response =
       flags.confirm &&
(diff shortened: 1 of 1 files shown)


packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
```diff
@@ -179,6 +179,14 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
       return
     }
 
+    output.print(
+      `\n${chalk.yellow(
+        chalk.bold(
+          'Note: If configued, migrations will trigger webhooks. You can disable webhooks on https://manage.sanity.io',
```
juice49, line 185: I think this message is a little ambiguous about _what_ is configured:

1. The migration is configured to trigger webhooks (e.g. via a flag).
2. The dataset has configured webhooks.

Maybe this is clearer?

> If your dataset has active webhooks, migrations will trigger them.

juice49, line 185: It might be better to tell users about the `sanity manage` CLI command, which will open the current project  in Manage for them. If they manually navigate to https://manage.sanity.io, they'll first need to find their project.

How about this?

> Run ${chalk.cyan('sanity manage')} to open the management interface in your web browser, then go to the API section to turn off webhooks. You can safely turn them back on after completing the migration.

Note I'd argue we should avoid the term "disable". Here's an excerpt from [Apple's style guide](https://support.apple.com/en-gb/guide/applestyleguide/apsg7af4f5d0/web#apdd167d6f70b0a4):

> disable (v.), disabled (adj.)
> Don’t use disable to refer to turning off a feature or deselecting an option; use turn off or deselect.
> 
> You can turn off Airplane Mode using Control Center.
> To hide an object’s caption, deselect the Caption checkbox.
> 
> Don’t use disabled to describe features that are turned off or unavailable; use turned off, unavailable, or inactive.

juice49 (commented):
The CLI has the ability to list project webhooks (`sanity hook list`). It may be out of scope at this time, but it could be helpful to automatically check for active webhooks as part of the migration process, and only print a warning if any exist. This would be yet another guard rail that makes migrations safer than executing scripts directly.

juice49 (approved):
Looks great!

PR #6113, opened 2024-03-22
fix(pte): PTE inline block object modal closing on validation state changes

pedrobonamin:
Same issue as in https://github.com/sanity-io/sanity/pull/5766 but in this case affecting inline objects.

When we have validation errors the tooltip is enabled, when validation errors disappear the tooltip is disabled.
Making that update when the modal is open it will trigger a "unmount" and "re-mount" of the InlineObject causing the  modal to unmount and remount, triggering the animation again and losing the input focus.

By checking if the modal is open to **enable / disable** the tooltip, we can be sure that this component won't unmount and mount causing  issues. 

https://github.com/sanity-io/sanity/assets/46196328/f2ca9093-7d42-4059-b09f-cec3bc221fd7


binoy14 (approved):
makes sense to me, thanks!

PR #6115, opened 2024-03-22
feat(sanity): use actions API when deleting documents

bjoerge:
This should be good to go now

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/operations/delete.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/operations/delete.ts
@@ -14,8 +14,7 @@ export const del: OperationImpl<[], 'NOTHING_TO_DELETE'> = {
       // this disables referential integrity for cross-dataset references. we
       // have this set because we warn against deletes in the `ConfirmDeleteDialog`
       // UI. This operation is run when "delete anyway" is clicked
-      //@todo awaiting backend support for this
-      // query: {skipCrossDatasetReferenceValidation: 'true'},
+      query: {skipCrossDatasetReferenceValidation: 'true'},
       body: {
         actions: [
           {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/operations/delete.ts
```diff
@@ -1,21 +1,29 @@
...
 import {type OperationImpl} from './types'
 
 export const del: OperationImpl<[], 'NOTHING_TO_DELETE'> = {
   disabled: ({snapshots}) => (snapshots.draft || snapshots.published ? false : 'NOTHING_TO_DELETE'),
-  execute: ({client, schema, idPair, typeName}) => {
-    const tx = client.observable.transaction().delete(idPair.publishedId)
+  execute: ({client: globalClient, schema, idPair, typeName}) => {
+    const vXClient = globalClient.withConfig({apiVersion: 'X'})
 
-    if (isLiveEditEnabled(schema, typeName)) {
```
cngonzalez, line 9: I notice that the delete action includes the `drafts.live-edited-document-id` when we execute a delete on a live-edited document.

The backend doesn't seem particularly bothered by this -- in includes a "mutations" array in its response payload that has _just_ the live edited document ID, but is it something we should avoid sending?

bjoerge: Good catch! I think it doesn't matter that much, and might even be preferable.

There's actually also a slight benefit - if you go from "liveEdit: false" to "liveEdit: true" while having a draft, the existing draft will still be floating around, which can cause trouble later on. Now that delete will also actually delete the draft, there's now actually a workaround (although veeery sketchy) in that you can hit delete on the document and then restore it later on.

cngonzalez: interesting, TIL! also interesting to consider as this API develops and we use it for other purposes. 

with that in mind, IMO this is good to go :)

cngonzalez (approved):

PR #6119, opened 2024-03-22
fix(pte): update default height of PTE in Form

jtpetty:
Fixes EDX-906

### Description

Before:
image

After:
image

### What to review

Is this the right place to make this change?

### Testing

Minor UX sizing


pedrobonamin (approved):

juice49:
Is it intentional that we changed from a relative unit (`em`) to an absolute unit (`px`)? The height of the PTE will no longer adjust relative to the user's base browser font size: it will always be `300px`, regardless of how big (or small) the base browser font size.

I think we probably want the height to increase in tandem with the font size. The bigger the font, the more height needed to see a meaningful amount of content.

PR #6124, opened 2024-03-24
feat(vision): link to edit documents

sgulseth:
### Description

Suggestion: Add a link icon next to each _id and _ref attributes

### What to review

Is this "safe"? Ref #6074 the user might be querying another dataset 🤔  Though it provides an easy way to inspect a document.

### Testing

N/A

### Notes for release

n/a - no notes needed

### DIFF
--- a/packages/@sanity/vision/src/components/ResultView.tsx
+++ b/packages/@sanity/vision/src/components/ResultView.tsx
@@ -2,7 +2,7 @@ import JSONInspector from '@rexxars/react-json-inspector'
 import {LinkIcon} from '@sanity/icons'
 import {Code} from '@sanity/ui'
 import LRU from 'quick-lru'
-import {useWorkspace} from 'sanity'
+import {useDataset} from 'sanity'
 import {IntentLink} from 'sanity/router'
 
 import {ResultViewWrapper} from './ResultView.styled'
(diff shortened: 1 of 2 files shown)


sgulseth:
* **#6124** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6124?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

rexxars (changes requested):
Love the feature! Been wanting this for a while :) 

> Is this "safe"? Ref #6074 the user might be querying another dataset 🤔 Though it provides an easy way to inspect a document.

For both features, perhaps we should grab all workspaces and try to narrow down the workspace in question based on the dataset. If we can find a single one, then use that workspace/schema. Else do not show the link, or show it as disabled with a tooltip ("no workspace configured for this dataset"). If there are multiple workspaces that references a dataset, we can _temporarily_ show a "unable to resolve workspace for dataset", and then later expand it with a "select workspace to use"?

We'll need to expand our intent resolving, however - it currently does not allow specifying a workspace. It's something we've been wanting for other reasons though, so would make sense to add.

sgulseth:
Agree 😄  Had a look at injecting workspace into intent links, and from what I could tell it was a bigger refactor. So instead of that I pushed a change where we only render the document links if the workspace you are on matches the dataset vision is querying. Then the current behavior should work as is?

rexxars (approved):

sgulseth:
image
Awesome!

PR #6126, opened 2024-03-25
fix(tasks): update description input to create blocks

pedrobonamin:
### Description
This PR updates the description input used in tasks, which is defined in the comments plugin.
That input didn't have support to create new blocks when using the "enter" button, it either submitted the form or closes the mentions menu.
With this updates, the editor will insert new blocks when it's used in tasks.

Adding support for multi-block description

It also exposes a new prop  into the Comment Input to allow for a customisable renderBlock function. 

Screenshot 2024-03-25 at 14 36 44


hermanwikner (approved):
Nice! 👍

PR #6128, opened 2024-03-25
fix(schemaExtract): guard for list options not being an array

sgulseth:
### Description

We have gotten reports that `createStringTypeNodeDefintion` throws an error `stringSchemaType.options?.list.map is not a function` this indiciates that we, even though the types & docs says it will be an array, are not receiving an array.

I've noticed that in other parts of the code base we are guarding for `list` being an array https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/block-tools/src/util/blockContentTypeFeatures.ts#L115

My suspicion is that we handle plugin types incorrectly, the report indicated usage of a [visual-options](https://github.com/fddigital-uk/sanity-plugin-visual-options/blob/trunk/src/index.js#L7-L14), which might indicate that we are parsing plugin types incorrectly.

### What to review

I think this work around should be fine for now? Will have a look at the root case separately

### Testing

n/a

### Notes for release

n/a - no notes needed

### DIFF
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -347,10 +347,11 @@ function isNumberType(typeDef: SanitySchemaType): typeDef is NumberSchemaType {
 function createStringTypeNodeDefintion(
   stringSchemaType: StringSchemaType,
 ): StringTypeNode | UnionTypeNode<StringTypeNode> {
-  if (stringSchemaType.options?.list) {
+  const listOptions = stringSchemaType.options?.list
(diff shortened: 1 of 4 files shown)


sgulseth:
* **#6128** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

rexxars:
Rebased and added a test that provoked the `map is not a function` case. Looks good to me after the fix.

rexxars (approved):

PR #6129, opened 2024-03-26
fix: improve document pane path handling

rdunk:
### Description

This PR makes some improvements to how the `DocumentPaneProvider` handles paths, specifically `openPath`. It also makes a small change so that missing PTE input blur events are now emitted.

There are a few UI issues that users are unlikely to encounter within the context of the Structure tool—but become much more prevalent within the context of Presentation—as they only occur when `openPath` state is set from a path provided via the URL:

- PTE cursor (sometimes) jumps to beginning of a block on focus.
- PTE scrolls to previously opened block on blur.
- PTE scrolls to block on click.

In addition:

- The PTE currently emits no handleable blur event (needed to fix these issues).
- Presentation's persistence of the path in the URL causes duplicate element focus.
- This duplication causes blocks that open modals to open on single click instead of double click.

### What to review

- That these changes don't have any obvious unintended side effects (the PTE is not so much my domain).
- That the updated `ArrayOfObjectsField`’s `handleBlur` logic is the best solution for detecting refs forwarded to the PTE input.
- That `deepEquals` is the right method for comparing path refs with new values. These replace strict equality checks that always evaluated to true as they were comparing arrays.

### Testing

The document pane and PTE should function exactly as before/expected, with the above issues resolved. CRX-630 contains a link to a video demonstrating some of the issues if any of the above is unclear.

### Notes for release

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -10,6 +10,7 @@ import {useToast} from '@sanity/ui'
 import {fromString as pathFromString, resolveKeyedPath} from '@sanity/util/paths'
 import {omit} from 'lodash'
 import {memo, useCallback, useEffect, useMemo, useRef, useState} from 'react'
+import deepEquals from 'react-fast-compare'
 import {
   type DocumentFieldAction,
   type DocumentInspector,
@@ -140,6 +141,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     params.path ? pathFromString(params.path) : EMPTY_ARRAY,
   )
   const focusPathRef = useRef(focusPath)
+  const openPathRef = useRef<Path>([])
   const activeViewId = params.view || (views[0] && views[0].id) || null
   const [timelineMode, setTimelineMode] = useState<'since' | 'rev' | 'closed'>('closed')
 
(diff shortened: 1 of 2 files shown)


jtpetty (approved):
These all seem reasonable. The `deepEquals` seems like the right fix. I am not sure if there is a better way to handle forwarding PortableTextInput events.

juice49 (approved):
I tested this out in Studio. Looks good to me!

> That deepEquals is the right method for comparing path refs with new values. These replace strict equality checks that always evaluated to true as they were comparing arrays.

I was curious about this, too. `@sanity/util/paths` exports a `toString` function. I benchmarked both options, comparing the following paths:

```ts
const pathA = ['a', 'b', 99, 'c', { _key: '5cb41a2e262a' }]
const pathB = ['a', 'b', 999, 'c', { _key: '9da68m7q100j' }, 'd']
```

react-fast-compare is consistently faster:

```
pathToString x 6,163,568 ops/sec ±0.78% (92 runs sampled)
reactFastCompare x 109,236,572 ops/sec ±13.01% (48 runs sampled)
Fastest is reactFastCompare
```

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -567,6 +529,47 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+  )
+
+  const handleBlur = useCallback(
+    (blurredPath: Path) => {
+      if (disableBlurRef.current) {
+        return
+      }
+
+      setFocusPath(EMPTY_ARRAY)
+      setOpenPath(blurredPath)
```
skogsmaskin, line 560: Why do we set the blurred path to be open here? This doesn't seem right, and is causing issues with opening (existing) text annotations in the PTE.

rdunk: Sometimes some initial `openPath` state is set: via a URL when using Structure tool, or anytime you click on an element in Presentation.

This state is never changed, which means that if the initial `openPath` references some block in the the PTE, it will force the scroll position back to that initial block as you try to move the cursor to a different one.

So as I understood it, the `openPath` state needs to be unset/reset somehow when e.g. clicking outside of the PTE to fix this scroll issue, hence the blur event.

I see now that Annotation calling `onPathFocus` with an empty path will trigger a blur event, which with this change causes the annotation to immediately close. Have I understood correctly?

PR #6131, opened 2024-03-26
fix(comments): weaken references in content snapshot

hermanwikner:
### Description

When inline comments were introduced, a new `contentSnapshot` property was added to the comment document data model. This property is used to store a snapshot of any content referred to by a comment. However, since this snapshot value may include a reference, Content Lake treats those references as "real references." This could lead to issues, such as preventing the deletion of a document referenced in the content snapshot because another document (the comment document) references it. Consequently, it is currently impossible to add an inline comment that includes a reference, as the POST request for that comment would fail.

This pull request addresses the issue by ensuring that `_weak: true` is added to any references in the content snapshot, which ensures that the aforementioned issues will not occur.

### What to review

- Ensure that it is possible to add inline comments that contain references.

### Notes for release

N/A


packages/sanity/src/structure/comments/src/utils/weakenReferencesInContentSnapshot.ts
```diff
@@ -0,0 +1,41 @@
...
+  // Return undefined for non-objects
+  return node
+}
+
+export function weakenReferencesInContentSnapshot(
+  snapshot: CommentDocument['contentSnapshot'],
+): CommentDocument['contentSnapshot'] {
+  if (!snapshot) return snapshot
+
+  return Array.isArray(snapshot) ? snapshot.map(weakenReferences) : weakenReferences(snapshot)
```
pedrobonamin, line 40: I think this check here is not necessary. The function will react accordingly to wether it is an array or an object.

packages/sanity/src/structure/comments/src/utils/weakenReferencesInContentSnapshot.ts
```diff
@@ -0,0 +1,38 @@
...
+        result[key] = weakenReferences(value)
+      } else {
+        // Directly copy other values
+        result[key] = value
+      }
+    })
+    return result
+  }
+
+  // Return undefined for non-objects
```
pedrobonamin, line 30: Not blocking, but this will return any non object value here right? Could we update this comment?

```suggestion
  // Return the node as is for non-objects 
```

pedrobonamin (approved):

PR #6136, opened 2024-03-27
fix(tasks): tasks UI updates.

pedrobonamin:
### Description
Fixes:
- force assignee selection menu to bottom position
- reduce description input min height in edit mode
- increase distance between actions in form create
- update placeholder texts
- **edx-1202** Don't create empty tasks in the form. 
- **edx-1207**  User Icon and text misaligned in activity feed
  Screenshot 2024-03-27 at 11 39 23

- **edx-1208** Comment icon alignment off
   Screenshot 2024-03-27 at 15 14 13

- **edx-1209**  Padding on comment reply too large and inconsistent with new comment creation spacing.
   Screenshot 2024-03-27 at 12 39 49
- **edx-1211**   Don't line break dates in activity log
  Screenshot 2024-03-27 at 11 30 26
- **EDX-1219**  Remove disabled from document tab
   Screenshot 2024-03-27 at 18 33 31

- **EDX-1220** UI Fix: Add tooltips to selection tokens for task status, assignee, and due date
   https://github.com/sanity-io/sanity/assets/46196328/ed72d806-53fc-4920-a57b-aefef2e1d2fc


ninaandal (approved):
lgtm 🚀

ninaandal (approved):

packages/sanity/src/tasks/i18n/resources.ts
```diff
@@ -42,50 +40,81 @@ const tasksLocaleStrings = defineLocalesResources('tasks', {
...
+  /** The heading in the tasks sidebar, in the subscribed tab, under the open details, when it's empty.*/
   'empty-state.status.list.open.subscribed.heading': 'No subscribed tasks',
+  /** The text in the tasks sidebar, in the subscribed tab, under the open details, when it's empty.*/
   'empty-state.status.list.open.subscribed.text': 'Tasks you subscribe to will show up here',
 
   /** Text used in the assignee input when there is no user assigned */
-  'form.input.assignee.no-user-assigned.text': 'Not assigned',
+  'form.input.assignee.no-user-assigned.text': 'Unassigned',
+  /** Text used in the assignee input tooltip when there is no user assigned */
+  'form.input.assignee.no-user-assigned.tooltip': ' Set assignee',
```
rexxars, line 89: Is the leading space here intentional? If so we should move it to the source code instead of the translation strings, since it is not obvious to a translator wether or not they need to include one. If it is an oversight, the space should be removed (this also goes for `form.input.assignee.user-assigned.tooltip`)

packages/sanity/src/tasks/src/tasks/components/form/tasksFormBuilder/TasksNotificationTarget.tsx
```diff
@@ -33,39 +33,45 @@ function TasksNotificationTargetInner(props: ObjectFieldProps<TaskDocument>) {
...
+    const contextUrl = new URL(context?.notification?.url || '')
+    const currentUrl = new URL(`${window.location.origin}${basePath}/`)
 
-  useEffect(() => {
-    if (documentId && documentType && previewValuesLoading) {
-      // Wait until the preview values are loaded
-      return
-    }
+    const studioUrl =
+      contextUrl.hostname !== 'localhost' && currentUrl.hostname === 'localhost'
```
rexxars, line 41: Theoretically this can be `127.0.0.1`, `0.0.0.0`, `::1` or a local IP as well. Maybe you can import `isDev` from `sanity` instead and use that? Not quite the same, but might be good enough?

rexxars (changes requested):

packages/sanity/src/tasks/i18n/resources.ts
pedrobonamin: Oversight, will update. Thanks

ninaandal (approved):

PR #6137, opened 2024-03-27
fix(structure): remove 'ago' suffix when published time was yesterday

ninaandal:
### Description
Adds a new translation string for when the published time was "yesterday" to avoid "ago" suffix
Screenshot 2024-03-27 at 12 59 51

### What to review
The tooltip text for the publish button when the document is already published

### Notes for release
Fixes tooltip text for publish button when document is already published

### DIFF
--- a/packages/sanity/src/structure/documentActions/PublishAction.tsx
+++ b/packages/sanity/src/structure/documentActions/PublishAction.tsx
@@ -41,6 +41,9 @@ function getDisabledReason(
 function AlreadyPublished({publishedAt}: {publishedAt: string}) {
   const {t} = useTranslation(structureLocaleNamespace)
   const timeSincePublished = useRelativeTime(publishedAt)
+  if (timeSincePublished === 'yesterday') {
+    return <span>{t('action.publish.already-published-yesterday.tooltip')}</span>
+  }
   return <span>{t('action.publish.already-published.tooltip', {timeSincePublished})}</span>
 }
 
--- a/packages/sanity/src/structure/documentActions/PublishAction.tsx
+++ b/packages/sanity/src/structure/documentActions/PublishAction.tsx
@@ -40,10 +40,7 @@ function getDisabledReason(
 
 function AlreadyPublished({publishedAt}: {publishedAt: string}) {
   const {t} = useTranslation(structureLocaleNamespace)
-  const timeSincePublished = useRelativeTime(publishedAt)
-  if (timeSincePublished === 'yesterday') {
-    return <span>{t('action.publish.already-published-yesterday.tooltip')}</span>
-  }
+  const timeSincePublished = useRelativeTime(publishedAt, {useTemporalPhrase: true})
   return <span>{t('action.publish.already-published.tooltip', {timeSincePublished})}</span>
 }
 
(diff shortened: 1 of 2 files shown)


binoy14 (approved):
makes sense to me, thanks!

packages/sanity/src/structure/documentActions/PublishAction.tsx
```diff
@@ -41,6 +41,9 @@ function getDisabledReason(
 function AlreadyPublished({publishedAt}: {publishedAt: string}) {
   const {t} = useTranslation(structureLocaleNamespace)
   const timeSincePublished = useRelativeTime(publishedAt)
+  if (timeSincePublished === 'yesterday') {
+    return <span>{t('action.publish.already-published-yesterday.tooltip')}</span>
+  }
```
rexxars, line 46: This only catches the case when english is used. For other languages it will not match and have the same issue. Eg in Norwegian, the `useRelativeTime` hook would return `i går`, and so the if-branch would not be hit and it would result in `Publisert i går siden`.

I think this is better solved by passing the `useTemporalPhase` option to the `useRelativeTime` hook and removing "ago" from the resource string. That should automatically include "ago" if necessary, and do the right thing for other languages - eg `Publisert i går` in norwegian, and `Publisert for 3 døgn siden` for dates further back.

```suggestion
  const timeSincePublished = useRelativeTime(publishedAt, {useTemporalPhrase: true})
```

rexxars: Note: unfortunately this also means we have to update all the locales to not also include the temporal phrase in their translations, but at least this will be consistent with the other places within the studio that we do this.

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -38,6 +38,8 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   'action.duplicate.label': 'Duplicate',
   /** Label for the "Duplicate" document action while the document is being duplicated */
   'action.duplicate.running.label': 'Duplicating…',
+  /** Tooltip text when publish button is disabled because the document is already published and published time was 'yesterday'.*/
+  'action.publish.already-published-yesterday.tooltip': 'Published yesterday',
   /** Tooltip when publish button is disabled because the document is already published, and published time is unavailable.*/
```
rexxars, line 43: With the fix suggested in `PublishAction`, this should not be needed - instead we should remove `ago` from the below string (`action.publish.already-published.tooltip`).

rexxars (changes requested):

packages/sanity/src/structure/documentActions/PublishAction.tsx
ninaandal: Ahh, so `useRelativeTime` deals with locales. I pushed your suggestion and tested with different locales and it looks accurate! 

About your note - this will be updated when I update the `resources.ts`, correct? All other locales use this as it's source?

rexxars (approved):

packages/sanity/src/structure/documentActions/PublishAction.tsx
rexxars: > About your note - this will be updated when I update the `resources.ts`, correct? All other locales use this as it's source?

Yep!

PR #6138, opened 2024-03-27
fix(tasks): don't lowercase users when searching in the assignee menu

pedrobonamin:
### Description

When searching for users in the tasks plugin, we are using lodash deburr to convert all the letters to latin letters. Also, we lowercase the display name. 
This, was going all the way to the UI, which should not happen.
This PR addresses that, by introducing a new property `searchName` when searching for users to avoid modifying the names that will be displayed in the UI.

Screenshot 2024-03-27 at 15 31 39

### Notes for release

Fixes an issue in which user names displayed in the user selection dropdown inside tasks were changing when filtering


hermanwikner (approved):
👍

PR #6145, opened 2024-03-28
fix(core): move `ToastProvider` inside `LocaleProvider`

juice49:


jordanl17:
This issue seems to be resolved from other fixes

PR #6147, opened 2024-03-28
fix: add default listFormat locale where LocaleProvider may not be available

cngonzalez:
### Description
This PR changes the invocation of `useListFormat` within the `<Translate />` component. The `<Translate />` component could sometimes be invoked outside of the `LocaleProvider` required for that function, causing the Studio to throw the error: `Sanity LocaleContext value missing`.

### What to review
The changed file and test to ensure all should work as expected.

### Testing
Tests were updated for the new `narrow` format.

### Notes for release
Fixes an issue where pressing the "Publish" button would throw the error `Sanity LocaleContext value missing`.

### DIFF
--- a/packages/sanity/src/core/i18n/Translate.tsx
+++ b/packages/sanity/src/core/i18n/Translate.tsx
@@ -1,6 +1,7 @@
 import {type TFunction} from 'i18next'
 import {type ComponentType, createElement, type ReactNode, useMemo} from 'react'
 
+import {useListFormat} from '../hooks/useListFormat'
 import {type CloseTagToken, simpleParser, type TextToken, type Token} from './simpleParser'
 
 const COMPONENT_NAME_RE = /^[A-Z]/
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/i18n/Translate.tsx
```diff
@@ -109,7 +108,13 @@ export function Translate(props: TranslationProps) {
   })
 
   const tokens = useMemo(() => simpleParser(translated), [translated])
-  const listFormat = useListFormat()
```
jtpetty, line 112: Could we keep this and wrap it in a try/catch and use the en-US format if we catch an exception?

cngonzalez: good catch! pushed that logic to the useListFormat hook to maintain rule of hooks.

binoy14 (approved):
makes sense to me, thanks!

PR #6149, opened 2024-03-29
fix: replace the JSON5 parser with JSON

kmelve:
### Description

`sanity init` in a Next.js project used the [golden-fleece](https://github.com/Rich-Harris/golden-fleece) library to patch the `tsconfig.json` with the `target` property. This introduced a bug, since we inserted a JSON5 string into a JSON file. 

This PR replaces the JSON5 parser with a JSON one ([silver-fleece](https://github.com/evertheylen/silver-fleece)).

### What to review

If the `sanity init` command now works (better) inside of a Next.js project.

### Testing

* Build the CLI
* Use the built CLI and run `./path/to/sanity init` inside a fresh Next.js project, choosing `Y` for TypeScript.

### Notes for release

Fixes a bug where wrongly formatted configuration would be inserted into `tsconfig.json` when `sanity init` was run inside of a Next.js project.


cngonzalez (commented):
confirmed that the plaintext string no longer getting inserted!

I'm a _little_ nervous about going from a package with 150k weekly downloads to one with 37 downloads. I'm not sure if the workaround (per my tests, something like:
```
config.compilerOptions["target"]= 'ES2017'
```

is preferable.

kmelve:
Got it!

I read through the source code, and it's fairly small/straightforward and has no external dependencies, so I felt the risk was low. 

Not sure I follow the alternate solve in your comment? 🤔

cngonzalez:
Yep, it seemed like a pretty straightforward fork to me too -- and the original package hasn't been updated in years so it's not like we gain benefits there 😬 

The code I put there (rather than using `silver-fleece`) seemed to insert the correct string. Are you not seeing the same thing? This package is a little obscure, it seems.

cngonzalez (approved):
Considering the lower impact of any future issues with this package (and annoying/difficult workarounds) let's move ahead! 👍

jtpetty (approved):
👍

packages/@sanity/cli/package.json
```diff
@@ -71,11 +71,11 @@
     "esbuild": "^0.20.0",
     "esbuild-register": "^3.4.1",
     "get-it": "^8.4.16",
-    "golden-fleece": "^1.0.9",
     "groq-js": "^1.7.0",
     "node-machine-id": "^1.1.12",
     "pkg-dir": "^5.0.0",
     "semver": "^7.3.5",
+    "silver-fleece": "^1.1.0",
```
rexxars, line 78: I have the same concerns as @cngonzalez about the popularity of this package - could we pin this to 1.1.0 since we know that release is "good"?

rexxars (changes requested):

packages/@sanity/cli/package.json
kmelve: Pinned!

rexxars (approved):

PR #6151, opened 2024-03-29
fix(schema-extraction): check for _required field in validation rules

sgulseth:


sgulseth:
* **#6151** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

rexxars (approved):
Looks good. Added a test case for the case where the `_required` attribute supposedly can be `'optional'` - code does the correct check, but just to prevent any regressions as this seems somewhat weird to me.

There is an additional case that would probably handle; on image/file fields, there is an `assetRequired()` rule which makes it so the `asset` field is required. This is a bit confusing, but if you have an image/file field that defines additional `fields`, setting any of those fields would make the base image/file field have a value, and would thus get treated as valid. What you usually _want_ is `assetRequired()`, which ensures you can't publish a document without an asset for a given field. This weirdness is due to backwards compatibility, and hopefully something we can get rid of in the future.

I am going to merge this as-is for the next release, but might be worth trying to get a separate fix for the `assetRequired` case.

PR #6158, opened 2024-04-02
feat(vision): add "save result as json/csv" buttons

RostiMelk:
### Description

Added "save result as" buttons to Vision result when a result is present. See bottom right corner of screenshot.

image

### What to review

- Check that result is properly formatted before saving
- Show we say "Download as" or "Save as"?
- Are i18n names up to satisfaction? 

### Testing

- Query a document type
- Click save as ...

### Notes for release

You can now save the Sanity vision result as JSON and CSV after querying


sjelfull:
Looks great! I noticed that CSV exports breaks if a column includes something non-string, like an object. Could you test it with different types of data and make sure they are encoded correctly?

perf/tests/package.json
```diff
@@ -15,8 +15,8 @@
     "studio:dev": "cd perf/studio && SANITY_STUDIO_DATASET=dev pnpm dev"
   },
   "dependencies": {
-    "@playwright/test": "^1.41.2",
-    "@sanity/client": "^6.15.7",
+    "@playwright/test": "1.41.2",
```
jtpetty, line 18: do we really want to pin to a specific version?

RostiMelk: Sorry. I think this is due to a bad rebase. Will fix

rexxars (changes requested):
I think using blob urls is a little safer here than using data URLs (especially when not base64 encoded). I've [made a PR](https://github.com/sanity-io/sanity/pull/6213) that changes to using those instead - let me know what you think.

There is one change I'd like to see here; we do not encourage splitting internationalized strings into different parts, which we do here:
1. "Save result as"
2. "CSV" / "JSON"

The rationale is that some languages may need to put the format at the start or middle of the sentence, which this does not allow for.

What we can do is use the `<Translate>` component and pass in the format download button as a component, eg:

```tsx
<Translate
  components={{DownloadFormatButton: JsonButton /* or CsvButton */}}
  i18nKey="result.save-result-as-format"
  t={t}
/>
``` 

And have the resource key be:
```tsx
{
  // ...
  'result.save-result-as-format': 'Save result as <DownloadFormatButton/>'
}
```

This will allow translators to use the format anywhere they want in the string.

RostiMelk:
@rexxars I've updated the PR with your suggestions now and ended up with the following structure:
```tsx
<Translate
  components={{
    SaveResultButtons: () => (
      <>
        <SaveJsonButton blobUrl={jsonUrl} />
        <SaveCsvButton blobUrl={csvUrl} />
      </>
    ),
  }}
  i18nKey="result.save-result-as-format"
  t={t}
/>
```

For the buttons I ended up using a static string (CSV, JSON), as these are file extensions and afaik should not differ between languages(?)

Let me know what you think

rexxars (approved):

PR #6165, opened 2024-04-02
fix(tasks): don't open tasks search on hotkey + enter

pedrobonamin:
### Description

Fixes an issue in which when tasks panel is open, cmd + K search changes target document rather than searches the studio.
This happens because the SearchPopover takes a `onOpen` prop which is triggered by the[ `useSearchHotkey` ](https://github.com/sanity-io/sanity/blob/edx-1223/packages/sanity/src/core/studio/components/navbar/search/hooks/useSearchHotkeys.ts#L38-L39) hook. Producing an undesired effect of opening by duplicated the search popover when tasks is open. Actually triggering the changes of the task form instead of the search change. 

Making this prop optional, and not passing it through from the `TargetField` fixes the issue.

### Notes for release

Fixes an issue in which when the tasks panel is open, cmd + K search changes target document rather than searches the studio.


sjelfull (approved):
Thanks, LGTM!

PR #6185, opened 2024-04-03
fix: use named `styled` import for better ESM runtime compat

stipsan:
`@sanity/ui` [has migrated](https://github.com/sanity-io/ui/commit/1ad5a02320434104dbe15dbd519e2db488fb15e7) to using `import {styled} from 'styled-components'` in order to support embedding the Studio in modern frameworks and bundlers, such as Remix + Vite. This PR does the same migration in the Studio codebase.


bjoerge (approved):
Great stuff, thanks! <3

PR #6188, opened 2024-04-03
chore(eslint): add eslint rule to enforce named styled import from styled-components

bjoerge:
### Description
Adds a lint rule to enforce named `styled` import from `styled-components`


stipsan (approved):
Thanks! 🥰

PR #6190, opened 2024-04-03
fix(core): add readOnly prop to dragHandle

pedrobonamin:
### Description

There is an issue in the list component, when the form status changes from `readOnly = true` to `readOnly = false`  the dnd context gets into a broken state and users can't drag elements inside the list.
This can be reproduced some times when publishing the document, but not always. The easiest way to reproduce it is to switch to a previous version of the document and then to the current version.

https://github.com/sanity-io/sanity/assets/46196328/41981fa8-f074-4fd6-a72b-d7f2b135acfa

I think this is caused because we are un-mounting the [`DragHandle` ](https://github.com/sanity-io/sanity/blob/c246c298a25c89f1378abd33fb9ef8aa36f3d05b/packages/sanity/src/core/form/inputs/arrays/common/DragHandle.tsx#L22-L23)  when changing to `readOnly`, then when remounting it it doesn't work as expected.

By keeping the element when it's in readOnly this is fixed and it also fixes the layout shift issue sometimes noticeable when publishing or when the form is in sync state.
Other solution will be to change the element key and force it to remount. https://github.com/sanity-io/sanity/pull/6189

####  Solution

https://github.com/sanity-io/sanity/assets/46196328/64d269d7-74a8-46f9-9bb6-79e1098f0e3f


sjelfull (changes requested):
Nice find! It looks good me, and the change in UX where the buttons stay in readOnly mode instead of flash in and out depending on state is an improvement.

It does look like the issue still is there when using the grid layout for arrays, could you check if something is missing there?

pedrobonamin:
> Nice find! It looks good me, and the change in UX where the buttons stay in readOnly mode instead of flash in and out depending on state is an improvement.
> 
> It does look like the issue still is there when using the grid layout for arrays, could you check if something is missing there?

Thanks @sjelfull pushed a new update that fixes it also in grid layouts. https://github.com/sanity-io/sanity/pull/6190/commits/783092baab81f0c04d95fcc8aa926682f472ad9b

sjelfull (approved):

bjoerge (approved):
Nice!

PR #6198, opened 2024-04-03
fix(structure): annotations not opening in portable text editor

pedrobonamin:
With the changes pushed in https://github.com/sanity-io/sanity/pull/6129 a bug was introduced into the PTE editor.
Annotations wont open.

The issue is generated by the onBlur action, when it's triggered, it's setting as `openPath` the blurred path. 
This specific change was introduced in that PR, possibly an oversight?

### Notes for release
Fixes an issue in PTE in which annotations were not opening by a bug introduced in v3.36.3

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Annotations.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/Annotations.spec.tsx
@@ -51,5 +51,64 @@ test.describe('Portable Text Input', () => {
       // Assertion: the annotation toolbar popover should be visible
       await expect(page.getByTestId('annotation-toolbar-popover')).toBeVisible()
     })
+
+    test('Can create, and then open the existing annotation again for editing', async ({
+      mount,
+      page,
+    }) => {
+      const {getFocusedPortableTextEditor, insertPortableText} = testHelpers({
+        page,
+      })
+      await mount(<AnnotationsStory />)
+      const $pte = await getFocusedPortableTextEditor('field-body')
+
+      await insertPortableText('Now we should insert a link.', $pte)
+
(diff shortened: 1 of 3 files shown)


rdunk:
Obviously lower priority than annotations, but just noting that this will reintroduce the scroll jumping issue with PTE in Presentation.

skogsmaskin:
> Obviously lower priority than annotations, but just noting that this will reintroduce the scroll jumping issue with PTE in Presentation.

@rdunk - Could this scroll jumping perhaps be fixed by moving [these lines](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/PortableText/hooks/useTrackFocusPath.tsx#L36-L50) down to line `59` ?

image

rdunk:
@skogsmaskin I'll verify that as a workaround, I _think_ we tested that before but I'll double check.

Done some digging and looks like I didn't catch this as our test document didn't contain an Annotation, only InlineBlock.

[This line](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/PortableText/object/Annotation.tsx#L97), which is unique to Annotation's `onOpen` callback means that the blur is triggered _after_ the DocumentPane's `onPathOpen`/`setOpenPath` is called, for all the other PT objects, the blur is triggered before.

rdunk:
Agreed with merging this as this bug is much worse than the scroll issue.

The problem is if an initial openPath is defined that references some block in the PTE, without setting the openPath on blur, the openPath is maintained as its initial value, which is incorrect. It's not the focus path we are setting to the blurred path, just the openPath, which AFAICT is correct.

I think ultimately the bug isn't caused by setting that openPath on blur, it is caused by the Annotation triggering the blur event _after_ it sets the openPath, which resets it.

rexxars (approved):

PR #6204, opened 2024-04-03
fix(cli): improve error handling on missing extracted schema

rexxars:
### Description

When you run `sanity typegen generate`, a fairly common case is that one may have forgotten to run `sanity schema extract` first. This currently gives a rather unhelpful error (if you're unfamiliar with the setup):
```
Error: ENOENT: no such file or directory, open './schema.json'
```

With this PR, I propose that we give a more helpful error if you have not provided even a custom path for the schema:
```
Error: Schema file not found: ./schema.json - did you run "sanity schema extract"?
```

I also noticed that using a `--config-path` flag did not work, only `--configPath`. To be consistent with other CLI commands, I changed this to allow both forms.

### Testing

Added some basic tests to ensure that the different forms work and gives the expected output.

### Notes for release

- Improved error message when extracted schema (from `sanity schema extract`) could not be found when running `sanity typegen generate`

### DIFF
--- a/packages/@sanity/cli/src/actions/typegen/generateAction.ts
+++ b/packages/@sanity/cli/src/actions/typegen/generateAction.ts
@@ -1,4 +1,4 @@
-import {constants, open} from 'node:fs/promises'
+import {constants, open, stat} from 'node:fs/promises'
 import {join} from 'node:path'
 
 import {readConfig} from '@sanity/codegen'
@@ -45,6 +45,19 @@ export default async function typegenGenerateAction(
     flags.configPath || flags['config-path'] || 'sanity-typegen.json',
   )
 
+  try {
+    const {isFile} = await stat(codegenConfig.schema)
+    if (!isFile) {
+      throw new Error(`Schema path is not a file: ${codegenConfig.schema}`)
+    }
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      // If the user has not provided a specific schema path (eg we're using the default), give some help
+      const hint =
+        codegenConfig.schema === './schema.json' ? ` - did you run "sanity schema extract"?` : ''
+      throw new Error(`Schema file not found: ${codegenConfig.schema}${hint}`)
+    }
+  }
   const workerPath = await getCliWorkerPath('typegenGenerate')
 
   const spinner = output.spinner({}).start('Generating types')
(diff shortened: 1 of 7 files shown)


packages/@sanity/cli/src/actions/typegen/generateAction.ts
```diff
@@ -40,8 +41,23 @@ export default async function typegenGenerateAction(
...
+      throw new Error(`Schema path is not a file: ${codegenConfig.schema}`)
+    }
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      // If the user has not provided a specific schema path (eg we're using the default), give some help
+      const hint =
+        codegenConfig.schema === './schema.json' ? ` - did you run "sanity schema extract"?` : ''
+      throw new Error(`Schema file not found: ${codegenConfig.schema}${hint}`)
+    }
+  }
```
sgulseth, line 60: Re-throw the error:

```suggestion
    }
    throw err
  }
```

packages/@sanity/cli/test/typegen.test.ts
```diff
@@ -0,0 +1,37 @@
+import {expect, test} from '@jest/globals'
+
+import {describeCliTest} from './shared/describe'
+import {runSanityCmdCommand} from './shared/environment'
+
+describeCliTest('CLI: `sanity typegen`', () => {
```
sgulseth, line 6: nit: add test for `!isFile`?

packages/@sanity/cli/src/actions/typegen/generateAction.ts
```diff
@@ -40,8 +41,23 @@ export default async function typegenGenerateAction(
   const trace = telemetry.trace(TypesGeneratedTrace)
   trace.start()
 
-  const codegenConfig = await readConfig(flags.configPath || 'sanity-typegen.json')
+  const codegenConfig = await readConfig(
+    flags.configPath || flags['config-path'] || 'sanity-typegen.json',
```
sgulseth, line 45: Do we need to be backward compatible? We mark it as experimental/beta so I would say it should be fine to correct this.

```suggestion
    flags['config-path'] || 'sanity-typegen.json',
```

sgulseth (changes requested):
thanks for the fixes, some feedback, but only 1 critical

packages/@sanity/cli/src/actions/typegen/generateAction.ts
rexxars: Makes sense to me 👍 Will remove the camelcase flag and also add it to the help output

packages/@sanity/cli/test/typegen.test.ts
rexxars: Added 👍

sgulseth (approved):

PR #6210, opened 2024-04-04
chore: update eslint-config-i18n to ^1.1.0

cngonzalez:
### Description

I received a request to review a renovabot PR that upped the version of this lint rule to not allow imports from `i18next`. Since we had those, it required a manual handling of those places where did previously import this. 

### What to review

I've made the following changes:
1. An .eslintrc file that ignores this rule in the i18n folder, since it likely makes sense to have that import there.
2. In core, updates to import the i18n file.
3. Disabling the rule where ONLY types were being imported. I'm not sure if it makes sense to export those types ourselves -- I'm not sure how much we plan to / want to deviate.
4. Updating files in `structure` to import from `sanity` rather than `i18next`.
5. Disabling the rule in tasks and comments -- the team is currently in the process of moving those files, where it would make more sense to do the local file import, as with other files in core. This task is logged in SDX-1230.

### Testing
Does it pass our linting rules?

### DIFF
--- a/packages/sanity/src/ui-components/dialog/Dialog.tsx
+++ b/packages/sanity/src/ui-components/dialog/Dialog.tsx
@@ -9,6 +9,7 @@ import {
 } from '@sanity/ui'
 import type * as React from 'react'
 import {type ComponentProps, forwardRef} from 'react'
+//eslint-disable-next-line @sanity/i18n/no-i18next-import -- avoiding a circular import
 import {useTranslation} from 'react-i18next'
 
 /** @internal */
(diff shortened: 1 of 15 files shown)


packages/sanity/src/ui-components/dialog/Dialog.tsx
```diff
@@ -9,6 +9,7 @@ import {
 } from '@sanity/ui'
 import type * as React from 'react'
 import {type ComponentProps, forwardRef} from 'react'
+//eslint-disable-next-line @sanity/i18n/no-i18next-import -- avoiding a circular import
 import {useTranslation} from 'react-i18next'
```
cngonzalez, line 13: It seems problematic to import from `sanity` here, and traversing up and down the file tree to get to `core` also seems problematic at best.

PR #6212, opened 2024-04-04
feat(core): add serverActions flag to config

cngonzalez:
### Description
After some discussion, we decided that it would be beneficial to be able to internally use document actions in the studio sooner rather than later, to be able to dogfood and detect errors.

This PR adds a parameter to studio config to enable server actions.
It changes the datastores invocation of documentStore to read from the workspace, and pass that parameter down.
Within the document store, it will pass that parameter to various functions that change behavior.

The most dramatic changes were needed in checkoutPair, which is used in many places, and is sometimes memoized or consumed as `snapshotPair`.

All of these changes are temporary and should be removed once these APIs become permanent.

It also includes @bjoerge 's changes from #6116, since that refactored checkoutPair and necessitated the bulk of these changes. Including it also allows for easy testing. If we'd like the PR to be "clean" and only enable the flag and different code paths, please let me know and I can remove it.

### What to review
The various changed files. To test locally, pull down the branch and uncomment the `unstable_serverActions` line in the sanity.config.ts in the test studio. Open your network calls and go to a new or old document and start making edits. You should see changes going to the /actions endpoint.

### Testing
Unit tests are being set in a separate PR, and should test which endpoint/payload is being used.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -11,7 +11,7 @@ import {presentationTool as pinnedPresentationTool} from '@sanity/presentation'
 import {debugSecrets} from '@sanity/preview-url-secret/sanity-plugin-debug-secrets'
 import {tsdoc} from '@sanity/tsdoc/studio'
 import {visionTool} from '@sanity/vision'
-import {defineConfig, definePlugin} from 'sanity'
+import {defineConfig, definePlugin, type WorkspaceOptions} from 'sanity'
 import {presentationTool} from 'sanity/presentation'
 import {structureTool} from 'sanity/structure'
 import {muxInput} from 'sanity-plugin-mux-input'
@@ -96,6 +96,7 @@ const sharedSettings = definePlugin({
     unstable_comments: {
       enabled: true,
     },
+
     badges: (prev, context) => (context.schemaType === 'author' ? [CustomBadge, ...prev] : prev),
   },
   plugins: [
@@ -151,6 +152,9 @@ export default defineConfig([
     plugins: [sharedSettings()],
     basePath: '/test',
     icon: SanityMonogram,
+    // unstable_serverActions: {
+    //   enabled: true,
+    // },
   },
   {
     name: 'partialIndexing',
@@ -308,4 +312,4 @@ export default defineConfig([
     ],
     basePath: '/presentation',
   },
-])
+]) as WorkspaceOptions[]
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -151,9 +151,9 @@ export default defineConfig([
     plugins: [sharedSettings()],
     basePath: '/test',
     icon: SanityMonogram,
-    // unstable_serverActions: {
-    //   enabled: true,
-    // },
+    unstable_serverActions: {
+      enabled: true,
+    },
   },
   {
     name: 'partialIndexing',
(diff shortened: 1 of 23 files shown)


dev/test-studio/sanity.config.ts
```diff
@@ -151,6 +151,9 @@ export default defineConfig([
     plugins: [sharedSettings()],
     basePath: '/test',
     icon: SanityMonogram,
+    // unstable_serverActions: {
+    //   enabled: true,
+    // },
```
cngonzalez, line 156: This is commented out because there's a bit of "low editor FPS" that needs to be debugged but is kind of an annoyance at present (another good reason for getting these changes in earlier!)

This sluggishness is also, by my testing, present in #6116 , so is likely closer to the server than to the code splitting.

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
```diff
@@ -82,8 +139,18 @@ function commitMutations(client: SanityClient, mutationParams: Mutation['params'
...
 
-function submitCommitRequest(client: SanityClient, request: CommitRequest) {
-  return from(commitMutations(client, request.mutation.params)).pipe(
+function submitCommitRequest(
+  client: SanityClient,
+  idPair: IdPair,
+  request: CommitRequest,
+  serverActionsEnabled?: boolean,
+) {
+  // console.log('submitCommitRequest', serverActionsEnabled)
```
bjoerge, line 148: probably good to remove this now so we don't forget about it :)

packages/sanity/src/core/store/_legacy/document/document-pair/documentEvents.ts
```diff
@@ -11,8 +11,13 @@ import {memoizeKeyGen} from './memoizeKeyGen'
 // A stream of all events related to either published or draft, each event comes with a 'target'
 // that specifies which version (draft|published) the event is about
 export const documentEvents = memoize(
-  (client: SanityClient, idPair: IdPair, typeName: string): Observable<DocumentVersionEvent> => {
-    return memoizedPair(client, idPair, typeName).pipe(
+  (
+    client: SanityClient,
+    idPair: IdPair,
+    typeName: string,
+    serverActionsEnabled?: boolean,
```
bjoerge, line 18: this calls for a tiny bit of caution I guess, since the events stream is memoized on a key based on all the other arguments, a change in the enabled status will not give a cache miss, but rather keep returning the one with the old flag

bjoerge: a workable fix for now is probably to include it with memoizeKeygen

bjoerge (approved):

dev/test-studio/sanity.config.ts
bjoerge: How does it compare to `next`? There shouldn't be much (or any?) extra CPU-bound work being added by these changes, so don't quite understand why this happens.

packages/sanity/src/core/store/_legacy/document/document-pair/documentEvents.ts
cngonzalez: ah i was wondering if I should have done that. Added now!

dev/test-studio/sanity.config.ts
cngonzalez: I'm unable to recreate today! 😔 But I did notice next acting faster then for some reason. If anything, this is reason to keep this enabled in test studio so we can capture, so I've uncommented it for now.

PR #6213, opened 2024-04-04
fix(vision): use blob urls for downloads

rexxars:
Builds on #6158

I suggest we use blob urls for the downloads instead of data urls, as it is somewhat safer in terms of encoding. Additionally, lets use `text/csv` which is the recognized mime type instead of `application/csv`.

There are a number of cases where the query result cannot be encoded to CSV. In these cases, I think we should explicitly call that out, instead of allowing the user to download a CSV file with only spaces/newlines.


RostiMelk (approved):
Awesome! And a nice touch disabling the CSV button and adding tooltip if the result cannot be encoded 👍

PR #6216, opened 2024-04-05
feat(tasks): add tasks upsell ui

pedrobonamin:
### Description

This PR introduces upsell UI into tasks.
Users who don't have growth or above plans will not have access to Tasks, but they will see it in the UI, and they will see this upsell ui, which invites them to upgrade.

To disable it, users will need to disable tasks from the config
```ts
 defineConfig({
   ....,
    unstable_tasks: {
      enabled: false,
    },
})

```

This PR uses the Upsell modal and panel from comments, those 2 components could be eventually moved to a shared location once this two plugins are moved into core. 

Screenshot 2024-04-05 at 15 02 42
Screenshot 2024-04-05 at 15 02 36


packages/sanity/src/tasks/src/tasks/components/sidebar/TasksSidebarHeader.tsx
```diff
@@ -74,6 +75,14 @@ export function TasksSidebarHeader(props: TasksSidebarHeaderProps) {
             onClick={handleTaskCreate}
             mode="bleed"
             text={t('buttons.new.text')}
+            disabled={mode === 'upsell'}
+            tooltipProps={
```
jtpetty, line 79: should we actually have a tooltip in default mode?

jtpetty (approved):
asked one question, but 👍 LGTM

packages/sanity/src/tasks/src/tasks/components/sidebar/TasksSidebarHeader.tsx
pedrobonamin: When the mode is `default` it won't render the tooltip because this prop will be undefined.
The `button` checks if this prop `tooltipProps` is defined or not, it will only add the tooltip if it's defined.

For default mode the text in the button explains what it will do, so I think we don't need a tooltip for that mode.

PR #6230, opened 2024-04-05
refactor(CI): use the official github app token generator action

stipsan:
### Description

GitHub has forked `tibdex/github-app-token` and have committed to supporting `actions/create-github-app-token` moving forward as they view this functionality critical. As this action is a vital part of our supply chain and needs to be as secure as possible, with as many layers of defence as possible, this is great news for us.
We should phase out `tibdex/github-app-token` and move to `actions/create-github-app-token` ASAP 🙌 

### Notes for release

N/A


bjoerge (approved):
Thanks!

PR #6234, opened 2024-04-05
fix(cli): formalize requirement of styled-components@6 and @sanity/ui@2

bjoerge:
### Description

This formalizes the requirement of using `styled-components@6.x` and `@sanity/ui@2.x` by showing an error on sanity start if the studio project depends on incompatible versions of these. The error will include a link to a help article explaining more in-depth details about how to upgrade.

### What to review
Any downside to making this move? We have been "soft requiring" these versions for a while, and you could make the argument that we should have added these sooner :)

### Testing
n/a

### Notes for release

- Formalize requirement of styled-components@6 and @sanity/ui@2.

### DIFF
--- a/packages/@sanity/cli/src/studioDependencies.ts
+++ b/packages/@sanity/cli/src/studioDependencies.ts
@@ -10,7 +10,6 @@ export const studioDependencies = {
     // Non-Sanity dependencies
     react: '^18.2.0',
     'react-dom': '^18.2.0',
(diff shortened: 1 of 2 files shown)


stipsan (approved):
Excellent, don't see downsides as we want folks to upgrade in order to use the same major versions we're using in our own testing suites 🚀 ❤️

rexxars (approved):

PR #6236, opened 2024-04-05
fix(tasks): change tasks button in nav to be toggle

pedrobonamin:
### Description

Change tasks button in nav to be toggle.
Clicking the task button should open the tasks sidebar if it's closed and should close it if it's open.


ninaandal (approved):
🚀

PR #6246, opened 2024-04-08
feat(tasks): add telemetry events to tasks

pedrobonamin:
### Description

Adds telemetry tracking to tasks.

### What to review

Is there something else worth tracking?
Are the event names correct?

### DIFF
--- a/packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
+++ b/packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
@@ -17,15 +17,6 @@ export const TaskStatusChanged = defineEvent<{
   description: 'Task status changed',
 })
 
-// A comment is added in a task
-export const TaskCommentAdded = defineEvent<{
-  is_reply: boolean
-}>({
-  name: 'Task Comment Added',
-  version: 1,
-  description: 'A comment was added in a task',
-})
-
 // A task is duplicated
 export const TaskDuplicated = defineEvent({
   name: 'Task Duplicated',
--- a/packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
+++ b/packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
@@ -1,0 +1,60 @@
+import {defineEvent} from '@sanity/telemetry'
+
+// Task is created.
+export const TaskCreated = defineEvent({
+  name: 'Task Created',
+  version: 1,
+  description: 'A task is created',
+})
+
+// Task status changed. %from% - %to%
+export const TaskStatusChanged = defineEvent<{
+  from: string
+  to: string
+}>({
+  name: 'Task Status Changed',
+  version: 1,
+  description: 'Task status changed',
+})
+
+// A comment is added in a task
+export const TaskCommentAdded = defineEvent({
+  name: 'Task comment added',
+  version: 1,
+  description: 'A comment was added in a task',
+})
+
+// A reply is added
+export const TaskCommentReplyAdded = defineEvent({
+  name: 'Task comment reply added',
+  version: 1,
+  description: 'A reply was added to a comment in a task',
+})
+
+// A task is duplicated
+export const TaskDuplicated = defineEvent({
+  name: 'Task duplicated',
+  version: 1,
+  description: 'A task is duplicated',
+})
+
+// A task is removed
+export const TaskRemoved = defineEvent({
+  name: 'Task removed',
+  version: 1,
+  description: 'A task is removed',
+})
+
+// The link to a task is copied
+export const TaskLinkCopied = defineEvent({
+  name: 'Task link copied',
+  version: 1,
+  description: 'The link to a task is copied',
+})
+
+// User visited the studio through a link with a task
+export const TaskLinkOpened = defineEvent({
+  name: 'Task link opened',
+  version: 1,
+  description: 'User visited the studio through a link with a task',
+})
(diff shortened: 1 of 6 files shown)


packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+  to: string
+}>({
+  name: 'Task Status Changed',
+  version: 1,
+  description: 'Task status changed',
+})
+
+// A comment is added in a task
+export const TaskCommentAdded = defineEvent({
+  name: 'Task comment added',
```
joneidejohnsen, line 22:
```suggestion
  name: 'Task Comment Added',
```

packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+// A task is removed
+export const TaskRemoved = defineEvent({
+  name: 'Task removed',
+  version: 1,
+  description: 'A task is removed',
+})
+
+// The link to a task is copied
+export const TaskLinkCopied = defineEvent({
+  name: 'Task link copied',
```
joneidejohnsen, line 50:
```suggestion
  name: 'Task Link Copied',
```

packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+// A task is duplicated
+export const TaskDuplicated = defineEvent({
+  name: 'Task duplicated',
+  version: 1,
+  description: 'A task is duplicated',
+})
+
+// A task is removed
+export const TaskRemoved = defineEvent({
+  name: 'Task removed',
```
joneidejohnsen, line 43:
```suggestion
  name: 'Task Removed',
```

packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+// A reply is added
+export const TaskCommentReplyAdded = defineEvent({
+  name: 'Task comment reply added',
+  version: 1,
+  description: 'A reply was added to a comment in a task',
+})
+
+// A task is duplicated
+export const TaskDuplicated = defineEvent({
+  name: 'Task duplicated',
```
joneidejohnsen, line 36:
```suggestion
  name: 'Task Duplicated',
```

packages/sanity/src/tasks/__telemetry__/tasks.telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+// A comment is added in a task
+export const TaskCommentAdded = defineEvent({
+  name: 'Task comment added',
+  version: 1,
+  description: 'A comment was added in a task',
+})
+
+// A reply is added
+export const TaskCommentReplyAdded = defineEvent({
+  name: 'Task comment reply added',
```
joneidejohnsen, line 29: Suggestion: Add an "is_reply" boolean property to "Task Comment Added" to avoid two different events

joneidejohnsen (changes requested):
Small nitpicks on event naming (title case for all event names) + suggestion on reducing number of events.

packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
```diff
@@ -116,9 +118,11 @@ export function TasksActivityLog(props: TasksActivityLogProps) {
 
       onChange(set(notification.subscribers, ['subscribers']))
 
-      operation.create(nextComment)
+      operation.create(nextComment).then(() => {
+        telemetry.log(TaskCommentAdded)
```
hermanwikner, line 122: Nice!

Suggestion/idea: This is not in scope for this pull request, but maybe we should consider (in the future) to centralize the logging of comment-related telemetry events within the `useCommentOperations` hook, rather than in the places where those operations are invoked? Currently, we do not log telemetry for comments, so we could consider this as a refactoring effort in the future when we implement telemetry for comments.

Something like this (pseudo code):

```ts
import {
  TaskCommentCreated,
  FieldCommentCreated,
  TaskReplyCommentCreated,
  FieldReplyCommentCreated,
} from '../path/to/telemetry'
import {useTelemetry} from '@sanity/telemetry'

function useCommentOperations(props: CommentOperationsHookOptions) {
  const {type} = props
  const telemetry = useTelemetry()

  const handleCreate = useCallback(
    async (nextComment) => {
      await createOperation(nextComment).then(() => {
        const isReply = Boolean(nextComment.parentCommentId)

        if (!isReply) {
          if (type === 'task') {
            telemetry.log(TaskCommentCreated)
          }

          if (type === 'field') {
            telemetry.log(FieldCommentCreated)
          }
        }

        if (isReply) {
          if (type === 'task') {
            telemetry.log(TaskReplyCommentCreated)
          }

          if (type === 'field') {
            telemetry.log(FieldReplyCommentCreated)
          }
        }
      })
    },
    [telemetry, type],
  )

  return {
    create: handleCreate,
    // ... rest of the operations
  }
}
```

joneidejohnsen (approved):

packages/sanity/src/tasks/src/tasks/components/activity/TasksActivityLog.tsx
pedrobonamin: I'll create a issue to track that, I think we should be tracking similar events for comments specific changes, but let's raise it with the growth team

joneidejohnsen: I believe comments are tracked server-side? We do have events for Comment Posted etc in Amplitude

pedrobonamin: That's right, actually I will remove this comments events from the studio and move them to the server, I can add a `type` property to distinguish normal comment from tasks comments.

PR #6247, opened 2024-04-08
fix(structure): prevent duplication of search filters when `listenSearchQuery` is used (e.g. in document lists)

juice49:
### Description

Both the `createSearch` factory and the returned `search` function accept a `filter` option. These filters are additive (joined together using the `&&` operator). We were accidentally providing the same filter to both functions, causing it to be needlessly duplicated in the produced search request.

### What to review

Document list filtering and search works the same way as before, but the produced search request no longer contains duplicated filters.

### Testing

We do not have tests for the `listenSearchQuery` function at the moment. Given this is a very limited change, I've not invested in adding a test suite at this point in time.


rexxars (approved):
👍

PR #6285, opened 2024-04-09
refactor: move createContext instances to singletons

binoy14:
### Description

Moves `createContext` calls to `sanity/_singletons` internal package. This is done in order to get singleton export of react when using it in multiple packages internally and bundlers know how to make the imports same so it doesn't break dynamic imports

### What to review

- changes makes sense
- everything still works
- everything that was previously exported is still exported

### Testing

Existing tests passing is the test

### Notes for release

N/A (internal)


ricokahler (approved):
Looks really good!

No blocking comments, just some ideas on how to potentially remove fragmentation.

Question: Would it be better to replace the `createContext` calls with re-exports from `sanity/_singletons` and then leave the existing imports as-is? I would then additionally flatten the singletons folder instead of mirror the folder structure of core etc.

Pros: Less changes to code, slightly less fragmentation
Cons: More slightly indirection

PR #6292, opened 2024-04-09
refactor(test-exports): mark tests that are expected to fail as `todo`

stipsan:
It's not ideal to have a failing CI test until `@sanity/cli` and `@sanity/codegen` are refactored to support ESM native node envs, so this PR marks them as `todo` instead.

It also improves the CI test input by enabling both the default `spec` reporter, and not just `node-test-github-reporter`.


binoy14 (approved):
makes sense to me, thanks!

PR #6298, opened 2024-04-09
fix(codegen): prevent crash due to incorrect babel config path

rexxars:
### Description

In v3.37.0, the output directory structure of the `@sanity/codegen` module changed as a result of a `pkg-utils` upgrade:

- From: `lib/_exports/`
- To: `lib/`

This broke some assumptions inside of the module which defined a relative path from the source file to the location of the babel configuration, which resulted in errors such as `Error: Cannot find module '<dir>/node_modules/@sanity/babel.config.json'`  

This PR recursively looks for the configuration file from the directory of the output and upwards until it finds it, which should make regressions due to other bundling changes more resilient.

While researching this bug, I had queries inside of a `.mjs` file that was not being picked up, so I have also snuck in a change that finds queries in both `.mjs` and `.cjs` files as well.

### What to review

- `sanity schema extract && sanity typegen generate` in a project with queries works as expected

### Testing

Added a test to ensure that the CLI tool works when there are queries to be found. Added it from the CLI side as there is a remote possibility that the module could have been bundled, and we would still want it to work all the way through in this case.

### Notes for release

- Fixes an issue where `sanity typegen generate` would crash due to not finding a babel configuration file

### DIFF
--- a/packages/@sanity/codegen/src/typescript/registerBabel.ts
+++ b/packages/@sanity/codegen/src/typescript/registerBabel.ts
@@ -1,17 +1,16 @@
-import {join} from 'node:path'
-
 import {type TransformOptions} from '@babel/core'
 import register from '@babel/register'
 
-const defaultBabelOptions = {
-  extends: join(__dirname, '..', '..', 'babel.config.json'),
-}
+import {getBabelConfig} from '../getBabelConfig'
 
 /**
  * Register Babel with the given options
+ *
  * @param babelOptions - The options to use when registering Babel
  * @beta
  */
-export function registerBabel(babelOptions: TransformOptions = defaultBabelOptions): void {
-  register({...babelOptions, extensions: ['.ts', '.tsx', '.js', '.jsx']})
+export function registerBabel(babelOptions?: TransformOptions): void {
+  const options = babelOptions || getBabelConfig()
+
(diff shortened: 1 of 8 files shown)


packages/@sanity/codegen/src/typescript/registerBabel.ts
```diff
@@ -1,17 +1,16 @@
...
+ *
  * @param babelOptions - The options to use when registering Babel
  * @beta
  */
-export function registerBabel(babelOptions: TransformOptions = defaultBabelOptions): void {
-  register({...babelOptions, extensions: ['.ts', '.tsx', '.js', '.jsx']})
+export function registerBabel(babelOptions?: TransformOptions): void {
+  const options = babelOptions || getBabelConfig()
+
+  register({...options, extensions: ['.ts', '.tsx', '.js', '.jsx', '.mjs', '.cjs']})
```
jtpetty, line 15: Is this list of extensions related to the one `packages/@sanity/codegen/src/readConfig.ts`? If so, should it be extracted out?

jtpetty (approved):
LGTM

packages/@sanity/codegen/src/typescript/registerBabel.ts
rexxars: Sorta. Will create a story for CLDX to look into it when they have time.

PR #6299, opened 2024-04-09
fix: resolve PTE inconsistency by passing serverActionsEnabled to documentEvents

cngonzalez:
### Description
When the `serverActions` config flag was set, we noticed inconsistent behavior coming from the Portable Text Editor: strings were being duplicated, new lines were being inserted, and patches generally seemed to be repeated or inconsistent.

This behavior can be tracked down to `document-store` -- `documentEvents` were not being initialized with the `serverActionsEnabled` parameter. 

Interestingly, before this was tracked down, forcing both `serverActions: enabled` and `serverActions: disabled` to use the Actions API still resulted in this inconsistent behavior, and I suspect that the `patchChannel` used in FormView (and published to by `documentEvents`) was in some way misaligned about remote and local patches (possibly because of inconsistent memoization) so there are some fixes to ensure that memoization functions are more aligned and that serverActions parameters are always required.

(A nice side effect is that the `serverActions` logic throughout the document store should become easier to remove in the future).

### What to review
You should be able to pull down this branch and type in a Portable Text Editor with server actions enabled and see consistent behavior.

### Testing
I'm a bit unclear on how to test this one. It was more of a configuration oversight.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -151,9 +151,9 @@ export default defineConfig([
     plugins: [sharedSettings()],
     basePath: '/test',
     icon: SanityMonogram,
-    // unstable_serverActions: {
-    //   enabled: true,
-    // },
+    unstable_serverActions: {
+      enabled: true,
+    },
   },
   {
     name: 'partialIndexing',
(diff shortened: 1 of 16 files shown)


juice49 (approved):
Looks great! 🙌

PR #6300, opened 2024-04-09
fix(cli): ensure output path for types exist before writing

rexxars:
### Description

If the `generates` path in the typegen configuration points to a path that does not exist, it will crash with a `ENOENT: no such file or directory, open '<some-path>/<generates-path>'` error.

This PR fixes this by running a recursive directory creation process before writing the output file.

### What to review

- Changes makes sense, paths are correct

### Testing

Added a CLI test to reproduce - the `out` directory does not exist in the fixture, and thus should fail before this change and pass after this change.

### Notes for release

- Fixes an issue where `sanity typegen generate` would crash if the configured output path did not exist prior to running the command

### DIFF
--- a/packages/@sanity/cli/test/__fixtures__/v3/working-typegen.json
+++ b/packages/@sanity/cli/test/__fixtures__/v3/working-typegen.json
@@ -1,3 +1,4 @@
 {
-  "schema": "./working-schema.json"
+  "schema": "./working-schema.json",
+  "generates": "./out/types.ts"
 }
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/test/__fixtures__/v3/working-typegen.json
```diff
@@ -1,3 +1,4 @@
 {
-  "schema": "./working-schema.json"
+  "schema": "./working-schema.json",
+  "generates": "./out/types.ts"
```
jtpetty, line 3: we probably should have a test with a directory and one without a directory, but otherwise, 👍

jtpetty (approved):
👍

PR #6307, opened 2024-04-10
fix(core): hidden caret on code formatted text

hermanwikner:
### Description

This pull request addresses an issue where the caret was not visible on code-formatted text within the PTE when using the dark scheme.

### What to Review

- Verify that the caret is now visible on code-formatted text when using the dark scheme.

### Notes for Release

N/A


jtpetty (approved):
👍

Issue #5648, opened 2024-02-02
Rich text: Scroll position jumps when deleting character when paragraph exceeds height of box


torjusb:
**Describe the bug**

If you have a rich text editor, and the paragraph exceeds the height of the editor (so you get a scroll bar), the scroll jumps to the top of the paragraph when you delete characters.

**To Reproduce**

1. Add a long paragraph to a rich text editor which exceeds the height of the box
2. Place the cursor towards the end of the paragraph, ensure that the first line of the paragraph is scrolled out of view
3. Delete a character
4. Observe that the scroll jumps to the top of the paragraph

**Expected behavior**

Scroll position should remain unchanged 

**Screenshots**

Here is a video showing the bug in action. Bit hard to see what is going on, but I'm simply removing a character with backspace, which scrolls to the top of the paragraph. Then I'm adding a space, which then scrolls down to the cursor position again.

https://github.com/sanity-io/sanity/assets/53583/be2d968a-ba46-4c1f-b6e6-ec99531cb18f

**Which versions of Sanity are you using?**

3.27.1

**What operating system are you using?**

MacOS 14. Chrome 120. Firefox 123.

**Which versions of Node.js / npm are you running?**

v18.17.1

**Additional context**

I believe the issue is related to the [`useScrollSelectionIntoView`](https://github.com/sanity-io/sanity/blob/f948e7dbcc60a5836e9f752591377370452730f4/packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx#L6) function.

pedrobonamin:
Hi @torjusb thanks for reporting and all the details added.
We can reproduce it internally and the bug has been added to triage, we will keep this issue updated once it's fixed.

ninaandal:
Hi! A fix for this was included in v3.40.0 ✨ Thanks for reporting!

PR #6310, opened 2024-04-10
fix(core): set scroll boundary to nearest

ninaandal:
### Description
There was an issue with deleting characters in a pte where the length of the paragraph exceeded the size of the box - it would jump to the top of the pte. 

By setting the `block` in the `scrollIntoView()` to `nearest` instead of `start`, which means the browser will align `leafEl` with the least amount of vertical movement. 

https://github.com/sanity-io/sanity/assets/44635000/f2d786d1-9d31-4d9a-bafa-404e83706161

Fixes [5648](https://github.com/sanity-io/sanity/issues/5648)

### What to review
Deleting characters will not cause the view to jump to the top. 

### Notes for release
Fixes issue where the cursor would move out of viewpoint when deleting a character in PTE. 
Does changing this have any other implications in the studio?

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx
@@ -16,7 +16,7 @@ export function useScrollSelectionIntoView(scrollElement: HTMLElement | null) {
           scrollIntoView(leafEl, {
             scrollMode: 'if-needed',
             boundary: scrollElement,
-            block: 'start',
+            block: 'nearest',
             inline: 'nearest',
           })
         }
--- a/packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx
@@ -14,7 +14,6 @@ export function useScrollSelectionIntoView(scrollElement: HTMLElement | null) {
             return
           }
           scrollIntoView(leafEl, {
-            behavior: 'smooth',
             scrollMode: 'if-needed',
             boundary: scrollElement,
             block: 'nearest',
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/PortableText/hooks/useScrollSelectionIntoView.tsx
```diff
@@ -16,7 +16,7 @@ export function useScrollSelectionIntoView(scrollElement: HTMLElement | null) {
           scrollIntoView(leafEl, {
             scrollMode: 'if-needed',
             boundary: scrollElement,
-            block: 'start',
+            block: 'nearest',
```
hermanwikner, line 19: This definitely improves the experience 👍  However, the editor still jumps, and I think it would be great if we could figure out how to prevent it from jumping at all when deleting content.

ninaandal: This happens because we have padding at the bottom of the text. I think we would have to implement custom scrolling logic for this and check if we are at the bottom and if the `leafEl` is at the bottom of the viewport and then _not_ scroll if this is the case. I do not think this would be worth it, but if you have another suggestion, then I am open to it!

skogsmaskin (changes requested):
The 'smooth' scrolling shoud not be there I think.

Try switching to and from fullscreen mode, and you'll see the issue when focused deep down in a long document.

RitaDias:
> Found an issue. The 'smooth' scrolling should not be there.
> 
> Try switching to and from fullscreen mode, and you'll see the issue when focused deep down in a long document.

Good catch @skogsmaskin !
I've pushed a new commit where it removes the smooth completely from the scroll into view which feels like I might just be missing a use case. You can try it out [here](https://test-studio-aiypa8ibn.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;30bb4547-7219-431e-a96c-4c86b9420005)

Let me know if I'm missing anything 👍

skogsmaskin:
Resolved conflicts and rebased this against `next`.

skogsmaskin (approved):
LGTM!

PR #6311, opened 2024-04-10
chore: enable serverActions in test and e2e studios

cngonzalez:
### Description
Server actions were disabled because of:
1.  a caching issue affecting Portable Text
2. an indexing issue causing documents published with actions to not be visible on the query endpoint.

Since both are now resolved, I'd like to re-enable server actions in our test studio for further winetasting.

### What to review
Are tests passing? Is there anything that we're missing?


test/e2e/tests/inputs/text.spec.ts
```diff
@@ -27,8 +27,8 @@ test.describe('inputs: text', () => {
     await page.waitForSelector('data-testid=field-simple', {timeout: 30000})
     const field = page.getByTestId('field-simple').getByRole('textbox')
 
-    // Enter initial text and wait for the mutate call to be sent
```
jtpetty, line 30: does this mean we are only testing the actions API and not the mutate API? Shouldn't we test both for a while at least?

cngonzalez: This particular line seems to be geared just to help this test await the right conditions (rather than testing the endpoint itself).

I agree with your concerns about testing both for a while -- I don't want the team to merge in code that works with server actions but not mutate (which is still the default) and for us to miss an issue.

Maybe what makes the most sense is having our CI stick with the mutate actions, and those of us who are working with server actions can test the e2e tests when we make changes?

cngonzalez: With that in mind, suggest we close this in favor of #6425  -- I think we've at least proven that e2e tests work as they are with server actions enabled! 🎉

PR #6313, opened 2024-04-10
chore: move `@sanity/import` out of monorepo

rexxars:
### Description

This PR moves the `@sanity/import` module out of the monorepo. It is rarely updated, does not necessarily tie in with the studio per se, and is not a browser oriented module. By moving it out of the monorepo, we can version it separately, more easily use different build steps and similar.

### What to review

- Does the monorepo and CLI still build?
- Are there any missed references that we need to update?

### Notes for release

None


juice49:
Should [the `@sanity/import` dependency](https://github.com/sanity-io/sanity/blob/chore/externalize-import/packages/sanity/package.json#L160) adopt a semver range? Now that it's externalised, the dependency won't be automatically incremented when publishing Studio.

```json
"@sanity/import": "^3.37.2",
```

juice49:
It might be useful for other consumers to move the `packages/sanity/typings/import.d.ts` typings to the external package, too. I guess it's an intentional decision to keep things simple for now.

juice49 (approved):
Looks good. I tested building the monorepo and `@sanity/cli`, and was able to successfully run import commands for test-studio.

Left a couple of non-blocking comments.

PR #6314, opened 2024-04-10
chore: move `@sanity/export` out of monorepo

rexxars:
### Description

This PR moves the `@sanity/export` module out of the monorepo. It is rarely updated, does not necessarily tie in with the studio per se, and is not a browser oriented module. By moving it out of the monorepo, we can version it separately, more easily use different build steps and similar.

### What to review

- Does the monorepo and CLI still build?
- Are there any missed references that we need to update?

### Notes for release

None


cngonzalez (approved):
LGTM! Build ran successfully.

I notice that there's still a typings file but I'm not sure whether we're planning to import types from the outside package, etc. Just making a note of it!

rexxars:
> I notice that there's still a typings file but I'm not sure whether we're planning to import types from the outside package, etc. Just making a note of it!

Right - I intentionally left it as it is very incomplete - if it was fully fleshed out, I'd move it to the export package. Hopefully we'll find some time to either rewrite the module in typescript or add some proper typings later.

PR #6319, opened 2024-04-10
chore(deps): bump `@sanity/pkg-utils` & simplify build scripts

stipsan:
### Description

The latest version of `@sanity/pkg-utils` adds two new flags to the `build` command:
- `--check`, which is equivalent to `pkg-utils build && pkg-utils`
- `--clean`, which runs `rimraf` on the `dist` folder
- full support for `typescript` v5.4.5

Previously, if you wanted to clean `./dist`, run build and then check the package you would need to chain the scripts together:
```json
{
  "scripts": {
    "build": "npx rimraf dist && pkg-utils build --strict && pkg-utils check --strict"
  }
}
```
But now the same result is accomplished by:
```json
{
  "scripts": {
    "build": "pkg-utils build --strict --check --clean"
  }
}
```

### What to review

If it builds it passes!

### Testing

Since `--check` is added to all the `"build"` scripts it's no longer necessary to have `"check:package"` scripts.

### Notes for release

N/A


bjoerge (approved):
Thank you! <3

PR #6321, opened 2024-04-11
chore: use `node:` prefix for node module imports

rexxars:
### Description

The `node:` prefix introduced in node v16.0/v14.18 makes it easier to tell which modules are coming from built-in node modules, and which are npm modules. With the addition of our sorted import rules, this nicely also moves them to the top, creating a nice separation of groups.

This may be a personal preference of mine, so would be open to hearing counterarguments. Unfortunately [there is no rule](https://github.com/import-js/eslint-plugin-import/issues/2717) in eslint-plugin-import for it, but we could potentially pull it in from [eslint-plugin-unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md), should we wish. I've left this out of scope until we've agreed that it's a direction we want to go.

### What to review

- Do you agree that it makes it easier to see where modules are coming from?
- Are there drawbacks of the approach?

### Testing

Should work as before, given we only support node 18 and higher, but please do give particularly the CLI a quick test.

### Notes for release

None.


juice49 (approved):
I think it's a great idea. Tested a few CLI commands with no issues.

bjoerge (approved):
Very nice! 100% agree with this, and I'm not aware of any drawbacks with `node:`-prefixes. Would also love to have an eslint-rule enforcing it, and have no qualms about adding `eslint-plugin-unicorn` to our config as it includes a bunch of other useful rules too (we can discuss and enable those separately though).

stipsan (approved):
Only drawback I'm aware of is that it's not supported in Next.js when used in code that goes through the web compiler, e.g. in a React component:

```ts
// pages/index.tsx

import path from 'path' // is fine
import path from 'node:path' // webpack cries about "unsupported protocol"
```

That's not relevant here though 👍

rexxars:
> Only drawback I'm aware of is that it's not supported in Next.js when used in code that goes through the web compiler, e.g. in a React component

TIL! That is helpful context - thank you for highlighting. Merging as it is not a problem in these code paths.

Issue #5909, opened 2024-03-05
[BUG] url field jumps out of viewport when pasting a link there; link isn't being added eventually


goodwin64:
**Describe the bug**

The URL field jumps out of the user's viewport when pasting a link there; the link isn't being added eventually.

**To Reproduce**

1. Go to documents, open the section that has a big chunk of text
2. Select the text you want to add a link to
3. Click on the "pencil" icon

**Expected behavior**
You can paste a link in the URL field

**Actual behavior**
URL field jumps out of the section viewport, see the recording below:

**Screenshots**

https://github.com/sanity-io/sanity/assets/10671879/b86a5e42-76fe-41b8-853b-767f6a04169e

**Which versions of Sanity are you using?**

```
@sanity/cli (global)          3.28.0 (latest: 3.30.1)
@sanity/cli                   3.28.0 (latest: 3.30.1)
@sanity/eslint-config-studio   2.0.4 (latest: 3.0.1)
@sanity/icons                 2.10.1 (latest: 2.10.3)
@sanity/ui                     1.9.3 (latest: 2.0.7)
@sanity/vision                3.28.0 (latest: 3.30.1)
sanity                        3.28.0 (latest: 3.30.1)
```

**What operating system are you using?**

MacOS Sonoma 14.1.1 (23B81)

**Which versions of Node.js / npm are you running?**

```
9.7.2
v18.17.1
```

**Additional context**

1. You need to have a big chunk of text - I used first 5 paragraphs of [Lorem Ipsum](https://www.lipsum.com/feed/html).
2. There's a workaround shown in the video - you may use the "enlarge section" button. But you need to do it for each link.

**Security issue?**

No

ninaandal:
Hi @goodwin64 ! Thank you for reporting, we have gotten some more signal of this and  😄 We can reproduce it internally and the bug has been added to triage, we will keep this issue updated once it's fixed.

goodwin64:
Thanks a lot for getting back on this! Looking forward for the fix!

ninaandal:
Hi again! A fix for this was included in v3.40.0 ✨ Thanks again for reporting!

Issue #5586, opened 2024-01-29
Navigate to the top of the text block when editing URL


hellonewday:
**Describe the bug**

Navigate to the top of the text block when editing URL

**To Reproduce**

1. Go to article item
2. Click on text editor section
3. Select text in a very long text block and edit URL
4. The text editor navigate to the first line of the very long text block

**Expected behavior**

It should not jump to the first line of the text block.

**Which versions of Sanity are you using?**

Latest version, Sanity v3.26.0

pedrobonamin:
Hi @hellonewday  thanks for reporting and all the details added.
We can reproduce it internally and the bug has been added to triage, we will keep this issue updated once it's fixed.

PR #6328, opened 2024-04-11
fix(core): change scrollIntoView block to be nearest

ninaandal:
### Description
Kind of the same as this [pr](https://github.com/sanity-io/sanity/pull/6310) - when adding a link to the bottom of the PTE, it would scroll to the top and it wasn't possible to scroll down to the open popover. 

This PR sets the `block` to scroll to `nearest` instead of `start` to have the least amount of scroll movement. 

Fixes issues [5909](https://github.com/sanity-io/sanity/issues/5909) and [5586](https://github.com/sanity-io/sanity/issues/5586). 

### What to review
Opening a link in a PTE where the paragraph is longer than the PTE block itself. 

### Notes for release
Fixes issue where creating a link at the bottom of a PTE would scroll to the top.


skogsmaskin (approved):
Thanks Nina!

PR #6333, opened 2024-04-11
refactor(tasks, comments): core plugins

ricokahler:
### Description

The following PR:

- Moves Tasks and Comments into `core` to and makes them default/core plugins that are added to every sanity config.
- Decouples Tasks and Comments from `structure`. This was done to preserve the seperation of `core` and `structure`. Since Tasks and Comments are now part of `core`, if this decoupling was not done, then much the structure tool would get imported due to the default plugins of tasks and comments.
- Removes the `sanity/_internalBrowser` entry point. The reasoning for this is that we already export a lot of internal code via the main export and it seems like we should either move all internals to that entry point or just remove it completely. The decision was to remove it since that's the easiest option that prevents the least amount of breakage.

Some important notes:

#### Regarding default plugins

When testing this, I ran into a new bug that came as a result from making the `tasks()` and `comments()` plugins added by default. Certain context values were not present (`<CommentsEnabledProvider/>` and `<TasksEnabledProvider />`) which caused the `useCommentsEnabled()` and `useTasksEnabled()` hooks to throw. To prevent this issue, I gave these context values default values that have the contexts set to `enabled: false`. This is so that if there is no context provider (as in the case of the task's FormBuilder) for these `use*Enabled()` hooks, the feature will be assumed to be disabled instead of throwing.

#### Decoupling Comments from `structure`

Though not ideal, the path I went down for decoupling Comments from `structure` was to invert the dependnecies and export the `CommentProvider` from core and import it into `structure`. From there I was able to provide all the dependencies from structure that comments needed without having to import structure-only dependencies like `usePaneRouter` and `useDocumentPane`.

In the future, we'd like to separate the concept of "the document" from structure and offer more document primitives available in core that are separate from structure. For now we may have to deal with structure depending on the CommentsProvider and comments populating it's document needs via `useDocumentPane` etc.

#### Decoupling Tasks from `structure`

Decoupling Tasks from Structure has a similar story to decoupling Comments from Structure. The only dependency Tasks has toward structure is just `const {isLast} = usePane()` so this was hoisted into a somewhat awkward `<IsLastPaneProvider />` . This is also a temporary thing and will be considered for those said primitive document APIs once we get some roadmap time. (cc @rexxars).

#### Circular imports and barrel files

These are the bane of my existence. While moving these, I tried my best to make them consistent but I'd rather just remove them. They cause a lot of issues and I think we should invest some time in removing them. There were some interesting work arounds I had to do and I really don't have an explanation for why some of them work.

Some of the circular imports cause issues with the monorepo dev command, some of them cause issues with random jest tests, and others cause issues with vite in dev mode. The build is typically fine and rollup (as part of vite) usually knows how to deal with them but doesn't mean the rest of our tooling does 😅.

### What to review

- Is everything in the right place?
- Are all tests passing?
- Did I rebase correctly?
- Do the new providers and their new locations make sense?

### Testing

Tested manually and made sure ~all~ almost all CI passed and reviewed much of this large PR myself but definitely could use a lot of eyes on this one.

### Notes for release

N/A


pedrobonamin (commented):
Thanks Rico, this is great!! 

Looking good, left some comments.

----
TasksEdit mode is not working due to the changes to the [CommentsProvider](https://github.com/sanity-io/sanity/blob/refactor/core-plugins/packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx#L109-L110).

`FormEdit` imports the CommentsProvider and some functions are not passed because they are not needed, I think we should make  all of this optional 
```
    isCommentsOpen
    onCommentsOpen 
    getCommentLink 
    onClearSelectedComment 
    onPathOpen
```
https://github.com/sanity-io/sanity/assets/46196328/61afd606-77e9-4670-8918-123feafd7b7a



----
Regarding `IsLastPaneContext` I'm thinking we could maybe use the `router` to find the last pane and maybe we won't need this. What I wasn't able to identify was the correct way to distinguish a document pane from a list pane when trying this approach.

----
I suggest we also remove the changes introduced in https://github.com/sanity-io/sanity/pull/6134 as they won't be needed anymore.

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
```diff
@@ -60,9 +65,50 @@ type TransactionId = string
 /**
  * @beta
  */
-export const CommentsProvider = memo(function CommentsProvider(props: CommentsProviderProps) {
-  const {children, documentId, documentType, isCommentsOpen, onCommentsOpen, sortOrder, type} =
-    props
+export function CommentsProvider(props: CommentsProviderProps) {
+  const parentContext = useContext(CommentsContext)
```
hermanwikner, line 69: I'm not sure I understand why this is needed. Could you clarify in what situations we might end up with nested comment providers?

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
```diff
@@ -45,13 +42,21 @@ interface ThreadItemsByStatus {
...
 export interface CommentsProviderProps {
   children: ReactNode
-  documentId: string
-  documentType: string
-  type: CommentsType
-  sortOrder: 'asc' | 'desc'
+  documentId?: string
+  documentType?: string
+  type?: CommentsType
+  sortOrder?: 'asc' | 'desc'
```
hermanwikner, line 48: What's the reason these were changed to optional? :)

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
```diff
@@ -60,9 +65,50 @@ type TransactionId = string
...
+  } = props
+
+  if (
+    !documentId ||
+    !documentType ||
+    isCommentsOpen === undefined ||
+    !onCommentsOpen ||
+    !getCommentLink ||
+    !onClearSelectedComment ||
+    !onPathOpen
```
hermanwikner, line 107: I don't have the full context on why we need to check if there is a parent comments context, but would it be possible to avoid throwing an error when optional props are missing? :)

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
ricokahler: I don't think I understand either lol. I think I was confused. I removed it and put it closer to how it was 😇

pedrobonamin (approved):

PR #6334, opened 2024-04-11
fix(core): ignore locally applied createIfNotExists mutations when using server actions

juice49:
### Description
When using server actions, `createIfNotExists` should never be submitted to the actions API, as these are applied server side as part of the edit action (we still want them to be applied optimistically though).

When patching locally, and we don’t have either a published or a draft we need to submit a create mutation in order to ensure that the draft exist. This will fail if someone concurrently creates it, but that’s ok, and what we want.

### What to review
- Does the logic make sense?

### Notes for release
n/a - internal for now


cngonzalez (approved):
Looks good to me! I was testing against this branch as I was investigating other issues yesterday, and didn't run into problems creating new documents, editing them, or creating drafts after publish.

PR #6346, opened 2024-04-12
fix(bug): update to not add annotation in PTE if all fields are null

jtpetty:
### Description

Fixes EDX-48

If you add an annotation but do not fill out any fields for it, we still mark the text as having the annotation. This change updates the logic so that if none of the fields are filled out, the annotation is removed.

Before:

https://github.com/sanity-io/sanity/assets/6889008/321f9367-88c5-4d23-a920-1e03831ce444

After:

https://github.com/sanity-io/sanity/assets/6889008/eabc930c-15f0-44ac-ae22-c43bbf597332

### What to review

Is this the right approach?

### Testing

Manual testing


hermanwikner (approved):
This is great! 👍 I think it would be good if @skogsmaskin could have a look at this as well

PR #6348, opened 2024-04-12
refactor(migrate): import package name, version statically

rexxars:
### Description

Background: gradually removing `require()` calls across the codebase in order to be more ready for an "ESM-first" world

Loading the package.json file using an import seems to work fine with our build process, and tree-shakes really well. Output after change is:
```ts
var name = "@sanity/mutator", version = "3.37.2";
function getUserAgent() {
  return typeof window > "u" ? `${name}@${version}` : null;
}
```

### What to review

- Works as expected

### Testing

Don't think we need to, build looks alright.

### Notes for release

None


rexxars:
Come to think of it, this extracts the version number at build time, which is prior to the version being bumped :/

PR #6352, opened 2024-04-12
chore(lint): add certain unicorn eslint rules

rexxars:
### Description

This adds a few more ESLint rules that should mostly be uncontroversial:

- Prefer `string.slice()` over `string.substring()` and `string.substr()` - slice is consistent with `array.slice()`, so just seems easier to reason about
- Prefer `node:` prefix on node.js imports. This is a follow-up to #6321 and was what prompted me to add these new rules.
- Prefer the use of `event.key` for keyboard events, instead of `.keyCode` which is deprecated
- Enforce correct use of Error constructor

### What to review

- Rules make sense
- `.substring()` replacements make sense

### Testing

The linter is the test ;)

### Notes for release

None

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -31,6 +31,7 @@ const config = {
     'prettier',
     'react',
     'tsdoc',
+    'unicorn',
   ],
   ignorePatterns: [
     '**/etc/*',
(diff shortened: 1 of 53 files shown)


bjoerge (approved):
LGTM, thanks!

bjoerge (changes requested):
ah, there's actually a lint error that: 

/home/runner/work/sanity/sanity/packages/@sanity/schema/example/test.js
  1:20  error  Prefer `node:assert` over `assert`  unicorn/prefer-node-protocol

Would be great to have that fixed before merge :)

bjoerge (approved):

PR #6369, opened 2024-04-15
chore: disallow and remove react default imports

bjoerge:
### Description

Disallows and removes usage of the React default export from the code base:

These will now give a lint error:
- `import React from 'react'`
- `import * as React from 'react'`
- `import type * as React from 'react'`

### What to review
- Do you agree with this change? The main benefit of doing it is consistency, and that not using default imports may give some ESM/tree-shaking benefits down the line.
- I'm deliberately ignoring this rule in CLI templates as I find it hard to reason about what impact removing default imports may have there.

### Notes for release
n/a - internal

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -325,6 +325,12 @@ const config = {
                 importNames: ['default'],
                 message: 'Please use `import {styled} from "styled-components"` instead.',
               },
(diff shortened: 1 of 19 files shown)


binoy14 (approved):
💯

PR #6370, opened 2024-04-15
fix(core/inputs): fix issues with focus handling

skogsmaskin:
### Description

This will fix some issues with focus handling in form inputs and the document pane that are problematic with the presentation tool (and potential causing problems in other situations as well).

* Fixes an issue where the PT-input wasn't properly calling formBuilder onBlur, or called a synthetic onBlur through imperative handling. This caused some issues because the event targets would not be the same elements as the inputElement ref had (imperative handle).  The PT-Input is now using the editable element as the inputElement ref, so that the blur and focus functions are now the native ones with the correct targets. The imperative handle that previously was used for this is now removed.

*  DocumentPaneProvider will now automatically open the path up to the focusPath. So when something is focused, it should also open everything up to that path.

* Changed it so that the PT-Input will not scroll to focus again when a focusPath is set and it's the same path that is currently focused in the PTE.
 
* Improved tracking of focusPath in the PT-Input. The current function didn't take into account that some members should be handled on their 'open' status while others must be handled on their 'focused' status (for instance an annotation which have what is visible in the editor (the text) on a different (block root) path than where the focus is (the markDef).

### What to review

That the Presentation tool now works like it's supposed to regarding navigating around inside Portable Text.

That the Structure tool handles focus correctly, especially for PT-Inputs and Array Inputs.

### Testing
PT-focus should be tested automatically by component tests.

The Presentation Tool functionality should be tested manually. See the "Presentation Tool" from the workspace menu in the Test Studio.

### Notes for release
Improves focus handling inside the Presentation Tool when working with Portable Text.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
@@ -11,7 +11,7 @@ import {
 } from '@sanity/portable-text-editor'
 import {type Path, type PortableTextBlock, type PortableTextTextBlock} from '@sanity/types'
 import {Box, Portal, PortalProvider, useBoundaryElement, usePortal} from '@sanity/ui'
-import {useCallback, useMemo, useState} from 'react'
+import {type ReactNode, useCallback, useMemo, useState} from 'react'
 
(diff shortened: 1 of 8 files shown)


rdunk (approved):
LGTM!

PR #6372, opened 2024-04-15
fix(tasks): padding when comment input is focused

hermanwikner:
### Description

This pull request fixes a small visual bug where the comment list items in Tasks would increase their padding when they had focus within. The solution is to only increase the padding when the _input_ wrap element has focus within, rather than the entire list item. Additionally, this pull request makes a small update to follow the naming convention for `data-ui` attributes.

https://github.com/sanity-io/sanity/assets/15094168/af8dd2cd-a12d-4ce7-9074-d9594699cfc6

### What to review

- Ensure that the visual bug no longer occurs and that the comment list items in Tasks have the correct padding when the input is focused.

### Notes for release

N/A


jtpetty (approved):
👍

PR #6373, opened 2024-04-15
feat(core): enable Text Search API search strategy by default

juice49:
### Description

This branch enables the Text Search API for all search surfaces by default. It can still be switched off by setting the `search.enableLegacySearch` configuration option to `true`.

### What to review

- The Text Search API is used by default.
- The Text Search API can be switched off by setting the `search.enableLegacySearch` configuration option to `true`

### Testing

Testing is already in place for the flagged search functionality.

### Notes for release

Enabled new search for all Studios by default. Global search, document list search, and reference search are now faster; especially for workspaces with large schemas. Additionally, wildcard (`*`) and negation (`-`) tokens can now be used in search queries.

### DIFF
--- a/packages/sanity/src/core/config/prepareConfig.ts
+++ b/packages/sanity/src/core/config/prepareConfig.ts
@@ -578,12 +578,12 @@ function resolveSource({
           initialValue: config.search?.unstable_partialIndexing?.enabled ?? false,
         }),
       },
-      unstable_enableNewSearch: resolveConfigProperty({
+      enableLegacySearch: resolveConfigProperty({
         config,
         context,
         reducer: newSearchEnabledReducer,
-        propertyName: 'search.unstable_enableNewSearch',
-        initialValue: true,
+        propertyName: 'enableLegacySearch',
+        initialValue: false,
       }),
       // we will use this when we add search config to PluginOptions
       /*filters: resolveConfigProperty({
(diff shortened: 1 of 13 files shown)


ricokahler (approved):

ricokahler:
Part of me does wonder if we should deprecate the existing config option and make it like `enableLegacySearch` or something though

juice49:
@ricokahler Great point! I agree. I've updated the legacy search strategy to be opt-in via the `search.enableLegacySearch` configuration option.

packages/sanity/src/core/config/prepareConfig.ts
```diff
@@ -578,11 +578,11 @@ function resolveSource({
           initialValue: config.search?.unstable_partialIndexing?.enabled ?? false,
         }),
       },
-      unstable_enableNewSearch: resolveConfigProperty({
+      enableLegacySearch: resolveConfigProperty({
         config,
         context,
         reducer: newSearchEnabledReducer,
```
ricokahler, line 584: we probably can rename this one now

ricokahler (approved):

packages/sanity/src/core/config/prepareConfig.ts
juice49: Missed it first time! Thank you.

PR #6376, opened 2024-04-15
chore(lint): fixes issue where boundaries lint rule was not working

binoy14:
### Description

Fixes an issue where the boundaries eslint plugin was not actually enforcing boundaries

Why?
> When we run lint in the monorepo, we use turbo, turbo runs lint script in each package.json and the eslintrc gets merged with the root to create a final config and the linting runs scoped to that working directory. The issue happens is at the root config we define the glob paths relative to the root and now the working directory of the package which causes the glob pattern to not match any of the paths and get ignored

What's Changed?

I have moved the boundary configuration to the `.eslintrc.cjs` in `packages/sanity` as the glob can be relative to that package and it reads much better than using relative glob at the root level 

### What to review

If it's linting and change makes sense

### Notes for release

N/A

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -12,7 +12,6 @@ const config = {
     browser: true,
   },
   extends: [
-    'plugin:boundaries/recommended',
     'sanity',
     'sanity/react',
     'sanity/import',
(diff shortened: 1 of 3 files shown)


binoy14:
* **#6376** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

ricokahler (approved):

PR #6378, opened 2024-04-15
fix: remove popover and use an absolutely positioned element

pedrobonamin:
### Description
It removes the Popover from the `UserPresenceCursor` element.

### Testing

Does it work in all the cases?
Is there a better way to do it?


hermanwikner (commented):
Thank you for looking into this!

I agree that it would be ideal to avoid rendering the dot in a portal using a Popover, but it seems necessary in this case. I have tried several approaches to get around this, but all have resulted in issues with the text in PTE. Unfortunately, there are still issues in this PR as well (see attached video).

https://github.com/sanity-io/sanity/assets/15094168/303632e0-e2ea-427e-9542-d499b9e18680

PR #6380, opened 2024-04-15
fix: crash in hotspot and cropping on Next.js

stipsan:
### Description

On Next.js opening the image crop dialog causes a crash:

Screenshot 2024-04-15 at 11 51 41

This is caused by `ToolCanvas` using refs in an unsafe way (the underlying `DragAwareCanvas ` component to `CanvasContainer` didn't use forwardRef). This PR solves it by refactoring `ToolCanvas` to lift logic out to a parent component that is able to use React hooks and by replacing the `makeDragAware` HOC (Higher Order Component) pattern with a regular React component that uses `forwardRef`.

### What to review

Setting crop and hotspots should work as before, if not slightly better.

### Testing

Just gotta test this on the Test Studio and see how it feels.

### Notes for release

Fixes bug causing the image hotspot and crop tool to crash on Next.js embedded Studios


packages/sanity/src/core/form/inputs/files/ImageToolInput/imagetool/DragAwareCanvas.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const keys = new Set([...Object.keys(oldProps), ...Object.keys(newProps)])
+    for (const key of keys) {
+      if (!Object.is(oldProps[key], newProps[key])) {
+        console.count(`DragAwareCanvas ${key} changed`)
+        return false
+      }
+    }
+    return true
+  },
+  // */
```
ricokahler, line 144: Did you intend to leave this in here?

packages/sanity/src/core/form/inputs/files/ImageToolInput/imagetool/draw.ts
```diff
@@ -0,0 +1,379 @@
...
+    scale,
+  })
+  printGuidelines({context, hotspotRect, image, MARGIN_PX, scale})
+  paintCropBorder({context, cropRect})
+
+  if (!readOnly) {
+    highlightCropHandles({context, cropHandles, cropping, opacity})
+  }
+
+  // paintPointerPosition({context, pointerPosition, scale})
```
ricokahler, line 376: I believe this is old code but this doesn't look intended

ricokahler (approved):
Super nice. If it works, I say ship it!

Note: Is it possible to rebase this to preserve more history? That may be better since a lot of this is old code

packages/sanity/src/core/form/inputs/files/ImageToolInput/imagetool/draw.ts
stipsan: Yeah I wasn't sure if it should be removed so I ported it with the rest 🤷

packages/sanity/src/core/form/inputs/files/ImageToolInput/imagetool/DragAwareCanvas.tsx
stipsan: No 🤦

PR #6382, opened 2024-04-16
fix(block-tools): removal of strikethrough in links when copying from gdocs to PTE

RitaDias:
### Description

Before the fix you were unable to copy a link that had a strikethrough decoration. With the fix that is now possible.

Before:

https://github.com/sanity-io/sanity/assets/6951139/cad3da8d-bbf2-43c8-987b-5fb32a7606a7

### What to review

Review the change in the regex for detecting the strikethrough.
Review the test added.

### Testing

[Test Studio](https://test-studio-c9ki7mxmv.sanity.build/test/structure/input-standard;portable-text;simpleBlock;c282a7c9-9e5e-4aa9-8f94-4ff40b6aa656%2Ctemplate%3DsimpleBlock) - from a google doc copy a link that has a strikethrough decorator and paste it in the "body" PTE (or any other PTE)

### Notes for release

Fixes an issue where copying strikethrough text from a google doc would remove the strikethrough once pasted in the PTE


juice49 (approved):
This looks good to me 🥳. The only potential issue I can see is that the [regex](https://regex101.com/r/NPhK57/1) now matches any `text-decoration` rule containing the string `line-through`.

So, this made up value would result in a line-through:

```
text-decoration: no-line-through;
```

As well as a mistyped value:

```
text-decoration: line-throughhhh;
```

I don't think it's a blocker, though.

sjelfull (approved):
Thanks for handling this! The comment about the regex is valid, but won't be a issue in reality since the output comes straight from Google Docs.

Not important, but we could update the regex to allow for extra rules separated by one or more space, instead of using wildcard selectors.

PR #6387, opened 2024-04-16
fix(migrate): forward possible API error response to error thrown for non 2xx

bjoerge:
### Description

Currently, when running a migration and the server responds with a non-2xx status, the error details from response body (e.g. in case of 4xx) is not properly forwarded to the error thrown by the migration command. This causes otherwise helpful details to be omitted from the error output.

Before:
```
npx sanity migration run some-migration

Running migration "some-migration" in dry mode

Project id:  some-project-id
Dataset:     doesnt-exist
Error
    at assert2xx (<snip>/fetchStream.ts:21:17)
   //…
```
After:

```
npx sanity migration run some-migration

Running migration "some-migration" in dry mode

Project id:  some-project-id
Dataset:     doesnt-exist
HTTPError: Dataset not found: Dataset "doesnt-exist" not found for project ID "some-project-id"
    at assert2xx (<snip>/fetchStream.ts:21:17)
   //…
```

### What to review

- Originally we were reading the error message from `error.description`. Is it a fair assumption that was a mistake and that API errors never have a `description` field? If not, I'll add a  fallback to description.
- If receiving a non-2xx and we could not parse the response as JSON we were currently throwing a separate JSON parse error. I believe this would prevent any HTTP errors _without_ a JSON-parseable body to be thrown as HTTP errors containing the status code and status message, so I've refactored this so we throw a HTTP error if we get a non-2xx and there's no JSON in the response body.

### Testing
Unit tests added.

### Notes for release
- Improved error handling when running migration against invalid project configurations

### DIFF
--- a/packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
+++ b/packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
@@ -1,0 +1,44 @@
+import {expect, test} from '@jest/globals'
+
+import {assert2xx} from '../fetchStream'
+
+test('server responds with 2xx', async () => {
+  const mockResponse = {
+    status: 200,
+    statusText: 'OK',
+    json: () =>
+      Promise.resolve({
+        this: 'is fine',
+      }),
+  }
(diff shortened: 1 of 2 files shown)


jtpetty (approved):
👍

PR #6389, opened 2024-04-16
fix(form/inputs): Remove obsolete code for restoring selection

skogsmaskin:
### Description

Previously it was necessary to restore the editor selection here, but it's no longer needed as our focus and blur handling has improved since then.

Should be removed to avoid confusion and noise.

### What to review

That when selecting text (or just placing the cursor) inside an annotated text, for example a link, it should keep that exact selection after clicking the edit Annotation button, and closing the edit modal again. 

### Notes for release
N/A


jtpetty (approved):
LGTM

PR #6393, opened 2024-04-16
fix: refactor checkoutPair to allow mutations for liveEdit documents

cngonzalez:
### Description
We received confirmation from the backend team that the upcoming Actions API is not going to be able documents that have the liveEdit parameter. Using the Actions API with a liveEdited document led to undesirable behaviors.

This PR adds a check to checkoutPair to see if mutations are targeting the "published" (whatever that may mean in the future) document. As far as I can see, the only situation in which that would happen is if we are in a liveEdit situation. If that's the case, then checkoutPair falls back to using the mutations endpoint.

### What to review
Is my assumption correct? Or is it too naive to consider?

### Testing
I've added a case for this in our unit tests, and an additional e2e test that goes through the full end-to-end experience of a liveEdited document. I'm not sure how to guarantee the `playlist` document type will always have liveEdit set.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
@@ -79,18 +79,26 @@ function requireId<T extends {_id?: string; _type: string}>(
   }
 }
 
+//if we're patching a published document directly
+//then we're live editing and we should use raw mutations
+//rather than actions
+function isLiveEditMutation(mutationParams: Mutation['params'], publishedId: string) {
+  const {resultRev, ...mutation} = mutationParams
+  const patchTargets: string[] = mutation.mutations.flatMap((mut) => {
+    const mutationPayloads = Object.values(mut)
+    if (mutationPayloads.length > 1) {
+      throw new Error('Did not expect multiple mutations in the same payload')
+    }
+    return mutationPayloads[0].id || mutationPayloads[0]._id
+  })
+  return patchTargets.every((target) => target === publishedId)
+}
+
 function toActions(idPair: IdPair, mutationParams: Mutation['params']) {
   return mutationParams.mutations.flatMap<HttpAction>((mutations) => {
     if (Object.keys(mutations).length > 1) {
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
```diff
@@ -79,19 +79,27 @@ function requireId<T extends {_id?: string; _type: string}>(
...
+  return patchTargets.every((target) => target === publishedId)
+}
+
 function toActions(idPair: IdPair, mutationParams: Mutation['params']) {
   return mutationParams.mutations.flatMap<HttpAction>((mutations) => {
     if (Object.keys(mutations).length > 1) {
       // todo: this might be a bit too strict, but I'm (lazily) trying to check if we ever get more than one mutation in a payload
       throw new Error('Did not expect multiple mutations in the same payload')
     }
-    if (mutations.delete) {
```
cngonzalez, line 88: I did not find anywhere in the codebase where `draft.mutate([draft.delete()])` occurred. 

In the interest of not having unused codepaths, I deleted this. If we're more concerned about completion we can certainly add it back -- I was worried about making this file too noisy.

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
```diff
@@ -120,13 +128,16 @@ function toActions(idPair: IdPair, mutationParams: Mutation['params']) {
   })
 }
 
-function serverCommitMutations(
+function commitActions(
```
cngonzalez, line 131: With this liveEdit change, it's become clear that `commitMutations` is going to stay, and has a different meaning. I've changed this function name to better reflect what it does (rather than being a temporary intermediary)

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
bjoerge: Good call! We have never done deletes optimistically (via buffered documents), so this makes sense to me.

bjoerge (approved):

PR #6395, opened 2024-04-17
fix(tasks): update taskURL for notification target

pedrobonamin:
### Description

Updates the URL created by the `TaskNotificationTarget` to include the tool name and creates a more defined path.
Instead of relying on the studio to resolve the tool it should open we dictate in the url which tool will open. 

This fixes the issue of infinite loading in the studio when visiting a studio from a task link.

### Notes for release

Fixes an issue of infinite loading that is present in some cases when visiting a studio from a task notification email.


jtpetty (approved):
LGTM

PR #6399, opened 2024-04-17
fix(structure): uncaught error while swapping images in array in PTE

RitaDias:
### Description

Caught by Applause, when swapping the position of the image within an array (which exists in a PTE) the studio was throwing an error. 
Important to mention: When trying both their studio on my machine and the test studio I wasn't able to get the error message to pop up at the bottom right, however the error still appeared as a console log message with the same descriptor.

It seems to be more generic than just with images and just have an overall issue with dragging

Before:

https://github.com/sanity-io/sanity/assets/6951139/99acbb2b-4bf8-48b8-85b6-ac8c8029f1c3

### What to review

Make sure that the error doesn't happen.

### Testing

You can manually test it on [this document](https://test-studio-6io1dzlp1.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;bc4f46f5-5a2d-4cd4-95dd-7e44e709cc7b) with the web console open. Changing the position of both items shouldn't trigger any console errors or visual errors.

- [ ] currently writing tests

### Notes for release

Fixes issue where changing positions of items of an array within PTE would launch a cryptic error


packages/sanity/src/core/comments/plugin/input/helpers.ts
```diff
@@ -107,10 +107,9 @@ export function useAuthoringReferenceElement(
 export function getSelectionBoundingRect(): DOMRect | null {
   const selection = window.getSelection()
   let range = null
-  try {
+
+  if (selection && selection.rangeCount > 0) {
     range = selection?.getRangeAt(0)
```
pedrobonamin, line 112: NIT, selection will always be defined here because of the previous check.
```suggestion
    range = selection.getRangeAt(0)
```

RitaDias: Good catch :) changed 👍

pedrobonamin (approved):

juice49 (approved):

PR #6415, opened 2024-04-18
fix: TS import paths importing index.ts twice incorrectly

binoy14:
### Description

Problem: tsconfig using `/*` from exports was causing the `index.ts` getting imported twice causing IDEs to suggest incorrect `sanity/index` auto imports which is not correct as it would translate to `sanity/index.ts/index`. Although the wildcard imports are nice I think it's better to have more control here and avoid other issues in future. 

### What to review

`build` and `dev` works correctly. 

### Testing

N/A
### Notes for release

N/A


binoy14:
* **#6415** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6415?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

stipsan (approved):
Good catch!

PR #6416, opened 2024-04-18
feat(scheduled-publishing): move scheduled publishing into core.

pedrobonamin:
#### Moves `scheduled publishing` plugin into core.
- Moves the [plugin](https://github.com/sanity-io/sanity-plugin-scheduled-publishing) into core, same functionality, same code, refactors imports and UI elements. 
- Updates how to accept options, see **notes for release**
- Includes upsell UI into scheduled publishing.
- Updates scheduled publishing UI to match with facelift.

#### Other changes
- Adds DEPRECATED_PLUGINS into [flatten config](https://github.com/sanity-io/sanity/pull/6416/files#diff-2e49f1259011a69d2ab069434ee411be52596bd6e507dff0ba44349bdd164a0c) this will allow us to deprecate plugins that will be moved into core, the plugins that are included in this list will be excluded from the plugins resolution. Importing a plugin from this list will produce no effect into the studio.
- Add options checks for the default plugins, not including them when opted out. [resolveDefaultPlugins]( https://github.com/sanity-io/sanity/pull/6416/files#diff-11f81b68e1f44310147b2e55c049f9c3df888aec6d0ef7aaca2dbd068d89c264)
- Moves upsell dialog and panel into a reusable location `sanity/src/core/upsell` as the use cases for this have been growing, to avoid repetition.

### What to review
Functionality works as expected.

The Pr introduces three new packages into core. Is it correct?
- @vvo/tzdb
- date-fns-tz
- swr

### Notes for release

This PR moves into core the functionalities included in [scheduled publishing ](https://github.com/sanity-io/sanity-plugin-scheduled-publishing) and adds it by default into all the studios. 
Users who have previously installed the plugin will not need to include it anymore in the plugins array, neither import the package.
To include custom options, users will have to use the new config property added into the `defineConfig` options.

To opt out from the functionality:
```ts
 defineConfig({ 
   // ....
   scheduledPublishing: {
       enabled: false, 
     }
  )
```

#### Migration example
Previously:

```ts
import {scheduledPublishing} from '@sanity/scheduled-publishing'


 defineConfig({
    //....
    plugins: [
      scheduledPublishing({
        // E.g. 12/25/2000 6:30 AM
        inputDateTimeFormat: 'MM/dd/yyyy h:mm a',
      })
    ]
  })
```

Now: 

```diff
-  import {scheduledPublishing} from '@sanity/scheduled-publishing'
```
```ts
 defineConfig({ 
   // ....
   scheduledPublishing: {
       enabled: true, 
       inputDateTimeFormat: 'MM/dd/yyyy h:mm a',
     }
  )
```


bjoerge (approved):
Looks good overall! Added a few questions during review. We should also continue our convo about how we can prevent the timezone database from being bundled by default (even if scheduled publishing is disabled), but that's not a blocker for merge.

Not a big deal, but I'm getting this React warning when running locally (http://localhost:3333/test/schedules/state/scheduled):

> Warning: React does not recognize the `withTooltip` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `withtooltip` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

PR #6417, opened 2024-04-18
fix(structure): update to not crash validation panel if unable to get path title

jtpetty:
### Description

Fixes EDX-846

The issue is that when the item that was causing the validation issue is deleted, the react component updates and throws an exception that it cannot find the deleted item. The fix I chose was to catch the exception and log it rather than crash the entire validation inspector.

Before:

https://github.com/sanity-io/sanity/assets/6889008/ee7e0828-c6d7-40bd-8627-240e445950f0

After:

https://github.com/sanity-io/sanity/assets/6889008/17e90cec-75ff-4d9a-9e32-ecb23ce4977b

### What to review

Should this fix be made at a higher level? It seems like `getPathTitles` is only used by the `ValidationInspector` and maybe it is too eager to throw exceptions.

### Testing

Manually validated this change.


juice49 (approved):
I think this is a reasonable bandaid, but it would be great to understand and fix the underlying issue.

The validation sidebar attempts to reconcile the document value obtained from `useEditState` with the validation state obtained from `useValidationStatus`. They are both powered by the same observable (`editState`), but they appear to be slightly out of sync; `useEditState` emits first, causing the validation sidebar to attempt to reconcile the stale validation status against the latest document value.

I did consider whether we could combine the observables powering `useEditState` and `useValidationStatus`, but I think that would be problematic because validation can be async.

It feels like this is something we can (and should) fix, but this solution to the symptom looks good to me.

PR #6418, opened 2024-04-18
feat: fetch feature toggle to enable serverside document actions

cngonzalez:
### Description
As we move toward transitioning to server-side document actions, we'd like to have some ability to have a remote way to configure whether or not our end users' document actions should be utilizing the Actions API endpoint, in case there's some risk or problem in content creation.

This PR introduces a remote check used in those places that code paths might diverge based on whether or not to use server actions. This way, we can shut off server actions and revert to old and tested functionality if needed.

This is done through the following:
1. Continuing to use the `sanity.config.ts` flag. If it is set to false, then the old mutations endpoint should be used, no matter what.
2. Introducing a HTTP request that is fed into `document-store`. This HTTP request lasts, at most, two minutes (thanks very much to @bjoerge  for refining this).
3. Because the `serverActionsEnabled` parameter is observable, it is no longer included in memoization.

### What to review
Any gotchas. As of this writing, the `/data/actions` GET request is returning a `false` -- so even if the config flag is set, users should continue to see the mutations endpoint being used.

### Testing
There were already tests in place to ensure that `checkoutPair` sent the correct requests based on the `serverActionsEnabled` parameter. These have been updated to respect that it is now an observable.

An additional end-to-end test was added to spy on requests made on document load and edit -- to ensure that the actions we send follow what we receive from the feature toggle endpoint.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
@@ -4,7 +4,7 @@ import {type SanityClient} from '@sanity/client'
 import {map, type Observable, of} from 'rxjs'
 
 //in the "real" code, this would be observable.request, to a URI
-export const featureToggleRequest = (
+export const fetchFeatureToggle = (
   client: SanityClient,
   serverActionsEnabled: boolean,
 ): Observable<boolean> => {
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
@@ -1,5 +1,5 @@
 import {type SanityClient} from '@sanity/client'
-import {map, type Observable, of, ReplaySubject, timer} from 'rxjs'
+import {map, type Observable, of, ReplaySubject, timeout, timer} from 'rxjs'
 import {catchError, share} from 'rxjs/operators'
 
 interface ActionsFeatureToggle {
@@ -18,6 +18,7 @@ export const fetchFeatureToggle = (defaultClient: SanityClient): Observable<bool
     })
     .pipe(
       map((res: ActionsFeatureToggle) => res.actions),
+      timeout({first: 2000, with: () => of(false)}),
       catchError(() =>
         // If we fail to fetch the feature toggle, we'll just assume it's disabled and fallback to legacy mutations
         of(false),
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
@@ -4,12 +4,22 @@ import {type SanityClient} from '@sanity/client'
 import {map, type Observable, of, ReplaySubject, timer} from 'rxjs'
 import {catchError, share} from 'rxjs/operators'
 
+interface FeatureToggle {
+  actions: boolean
+}
+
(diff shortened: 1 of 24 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
```diff
@@ -0,0 +1,35 @@
...
+  const {dataset} = defaultClient.config()
+
+  return client.observable
+    .request({
+      uri: `/data/actions/${dataset}`,
+      withCredentials: true,
+    })
+    .pipe(
+      map((res: ActionsFeatureToggle) => res.actions),
+      catchError(() =>
```
juice49, line 21: Do you think there's any value in adding a timeout as a safeguard here? My thinking is that slow feature flag resolution should never negatively impact user experience.

packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
```diff
@@ -0,0 +1,35 @@
...
+        of(false),
+      ),
+      share({
+        // replay latest known state to new subscribers
+        connector: () => new ReplaySubject(1),
+        // this will typically be completed and unsubscribed from right after the answer is received, so we don't want to reset
+        resetOnRefCountZero: false,
+        // once the fetch has completed, we'll wait for 2 minutes before resetting the state.
+        // we'll then check again once a new subscriber comes in
+        resetOnComplete: () => timer(1000 * 120),
```
juice49, line 32: This request seems to be sent every time I open a new document that hasn't already been opened in the past 120 seconds.

Is that the expected behaviour? I think the feature flag is controlled at the project level, rather than the document level, so it _seems_ wasteful (although ultimately inconsequential) to send the request for every document/listener.

packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
juice49: (I know this is a bit of a moot point at the moment given that feature flags are fairly static and powered by the same backend as other Studio features—but they will probably become more complicated in the future).

bjoerge: That's a great point. I don't think it hurts to add a timeout here. Added a 2s timeout in 0de508717a - if the request takes longer than 2s we'll treat it as feature "off"

packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
bjoerge: This behavior is a consequence of the fact that the various individual document pair methods are memoized on document ID and when they are initialized up, they start by checking whether to use the http actions APIs. One way to solve this would be to memoize the whole document store instead of the individual document methods, but that would be a bigger task, and probably not worth the effort this time around.

juice49: Thank you for explaining! I agree it's probably not worthwhile to take it on at this time, but it would be nice 🙂.

packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
juice49: Wonderful!

juice49 (dismissed):
This is great!

bjoerge (approved):

juice49 (approved):

PR #6425, opened 2024-04-19
chore: turn server actions on by default in test-studio

bjoerge:
### Description

This enables server actions for the dev/test studio

### What to review
- The diff for this particular PR is miniscule as it only enables the server actions in the dev/test-studio in order for us to dogfood and catch potential errors.

### Testing
- Unit tests has been included in prior work.
- I've done a bit of manual testing and haven't found any issues.

### Notes for release
n/a - internal


cngonzalez (approved):
Mentioned this in another PR but I think it's the right move!

PR #6428, opened 2024-04-19
fix(portable-text-editor): support collapsed range decorations

skogsmaskin:
### Description

This will fix a bug where a range decoration for a single point (collapsed) range would be rendered three times.

This is because we can't do a Range include check on those collapsed ranges, we must equality check the points.

Added test for this.

### What to review

That  collapsed range decorations aren't rendered multiple times.

### Testing

Covered by automatic tests.

### Notes for release
N/A


hermanwikner (approved):
Looks good! 🙏

PR #6430, opened 2024-04-19
fix: statusButton requires provided aria-label

jordanl17:
### Description

`statusButton` generic `aria-label` was not specific enough in all usages. This PR makes `aria-label` a required prop on `statusButton` and defines all current usages a value

### What to review

The only ambiguous usage is in `packages/sanity/src/structure/components/pane/PaneHeaderActionButton.tsx` where the fallback has been the original label used in `statusButton`

### Testing

-

### Notes for release

Fixes ambiguous aria-labels on some Status buttons


packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -316,6 +316,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
 
   /** --- Configuration issues --- */
 
+  'configuration-issues.button.label': 'Configuration issues',
```
jtpetty, line 319: don't forget a comment above this describing the purpose. This is used by a process to help translate these strings to other languages.

packages/sanity/src/core/components/StatusButton.tsx
```diff
@@ -48,7 +55,7 @@ export const StatusButton = forwardRef(function StatusButton(
     <StyledButton
       data-ui="StatusButton"
       {...restProps}
-      aria-label={label || t('status-button.aria-label')}
+      aria-label={label}
```
RitaDias, line 58: Maybe I'm misunderstanding, but in this case wouldn't it be wise to still have a fall back on the StatusButton itself (outside of making sure that the labels exist where the StatusButton is used?)

RitaDias:
Possible nitpick but for making things easier for folks on release/editorial:

> Notes for release
> Specific usages of StatusButton labels

If we're working on a fix the way Editorial tends to pose it is like "Fixes (...)"
For this I would try to reword it in a way that someone could understand without going too deeply into the weeds what this fix actually does which (imo) is not super understandable. There are some issues where that is not possible (and this might end up being one of them, but in that case it might end up with a message closer to "Fixes some issues with aria-labels in the StatusButton" for example :)

packages/sanity/src/core/components/StatusButton.tsx
jordanl17: Would a fallback be necessary given that the `aria-label` is now a required prop?

RitaDias: Oh yeah that's true! 🤔 
I guess my follow up question is whether this should be something that we should require of all buttons then (if we don't already). Since it might be strange that we're only requiring this on a StatusButton vs everywhere else.

But maybe it's a question for another time and out of scope for this PR? 🤔

jordanl17: That's a fair point! I'm not sure the situation with all other buttons but I can take a look after this PR. I'm not sure the exact process that surfaced this particular status button issue, but perhaps this was seem most commonly for now

PR #6439, opened 2024-04-21
fix(structure): resolve static types in document list filters

ricokahler:
### Description

There was a bug where the regex used to match static types that would eagerly match type filters that included more than one type. This would cause the document listener to not grab the search config for all the types in certain filter expressions.

For example, the search filter expression:

```
_type == 'testDocument'  || (_type == 'webhookDocument' && isActive != true
``` 

would only grab the search config for the `'testDocument` instead of both the `testDocument` and `webhookDocument`.

### What to review

Does the structure tool still work? Can you provide any filter and it finds all the appropriate types?

### Testing

I added a few unit tests and double checked the "Authors & Books"` custom pane.

### Notes for release

Fixes a bug that caused types to be incorrectly parsed from filter statements.

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -229,7 +229,6 @@ export const structure: StructureResolver = (S, {schema, documentStore, i18n}) =
                     id: 'authors-and-books',
                     title: 'Authors & Books',
                     options: {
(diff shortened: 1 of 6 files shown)


cngonzalez (approved):
Could not find any breaking changes here, LGTM

PR #6441, opened 2024-04-22
perf(portable-text-editor): improve range deocration render perf

skogsmaskin:
### Description

This will skip unnecessary re-renders of range decorations when the previous values are the same (but the object might be different). 

Also added a test for this.

### What to review
Will there be a situation where we did want to re-render even though the values are the same, but the object is different? I can't think of anything, but perhaps give this an extra thought.

### Testing
Covered by automatic tests

### Notes for release
N/A


skogsmaskin:
> I see the e2e tests picked up on something here, investigating.
This is now resolved

hermanwikner (approved):

PR #6442, opened 2024-04-22
test(e2e): fix unawaited assertion

juice49:
### Description

Fixes an e2e test that was improperly awaiting its assertions, sometimes causing the test to timeout.

### What to review

That the `test/e2e/tests/desk/deprecatedFields.spec.ts` test runs correctly.


binoy14 (approved):
solid find, thank you!

PR #6450, opened 2024-04-23
fix(tasks): close the form only after the task is created

pedrobonamin:
### Description

Fixes an issue in which tasks are kept in draft mode after clicking on the "create" button. 
The on change action triggered by the form doesn't provides a way to validate that the transaction ended, so in this case we will persist the form in `creating` state until the task store is updated with the correct values.
The task store fetches the latest task values when a task is updated, so if the value has been updated in the store it means the transaction has ended.

### What to review

Is there a better way to do this? 
Is there any way to access the transaction from the form onChange or from any form action?

### Notes for release

fixes an issue in tasks in which the UI showed that they were created but the transaction was didn't ended.

### DIFF
--- a/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
+++ b/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
@@ -70,7 +70,7 @@ export function FormCreate(props: ObjectInputProps) {
 
   useEffect(() => {
     // If after 10 seconds the task is still in the "creating" state, show an error and reset the creating state.
-    let timeoutId: NodeJS.Timeout | null = null
+    let timeoutId: ReturnType<typeof setTimeout> | null = null
     if (creating) {
       timeoutId = setTimeout(() => {
         setCreating(false)
--- a/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
+++ b/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
@@ -1,13 +1,13 @@
 import {TrashIcon} from '@sanity/icons'
 import {useTelemetry} from '@sanity/telemetry/react'
 import {Box, Flex, Switch, Text, useToast} from '@sanity/ui'
-import {useCallback, useState} from 'react'
+import {useCallback, useEffect, useState} from 'react'
 
 import {Button} from '../../../../../ui-components'
 import {type ObjectInputProps, set} from '../../../../form'
 import {useTranslation} from '../../../../i18n'
 import {TaskCreated} from '../../../__telemetry__/tasks.telemetry'
-import {useTasksNavigation} from '../../../context'
+import {useTasks, useTasksNavigation} from '../../../context'
 import {useRemoveTask} from '../../../hooks'
 import {tasksLocaleNamespace} from '../../../i18n'
 import {type TaskDocument} from '../../../types'
@@ -29,12 +29,9 @@ const getTaskSubscribers = (task: TaskDocument): string[] => {
   return subscribers
 }
 export function FormCreate(props: ObjectInputProps) {
+  const [creating, setCreating] = useState(false)
   const {onChange} = props
-  const {
-    setViewMode,
-    setActiveTab,
-    state: {viewMode},
-  } = useTasksNavigation()
+  const {setViewMode, setActiveTab} = useTasksNavigation()
   const toast = useToast()
   const telemetry = useTelemetry()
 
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
```diff
@@ -47,7 +44,52 @@ export function FormCreate(props: ObjectInputProps) {
...
+
+      setCreating(false)
+      if (createMore) {
+        setViewMode({type: 'create'})
+      } else {
+        setActiveTab('subscribed')
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [creating, savedTask?.createdByUser])
```
sjelfull, line 69: Can we use `useDocumentOperationEvent` here instead?

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
```diff
@@ -47,7 +44,52 @@ export function FormCreate(props: ObjectInputProps) {
...
+          title: t('form.status.error.creation-failed'),
+        })
+      }, 10000)
+    }
+    // Cleanup function to clear the timeout
+    return () => {
+      if (timeoutId) clearTimeout(timeoutId)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [creating])
```
sjelfull, line 89: Same here, lets bake this into one `useDocumentOperationEvent` handler if possible.

sjelfull (changes requested):
Thanks! Would prefer if we could re-use `useDocumentOperationEvent` here. That will give us a more solid way to handle the create lifecycle here.

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
```diff
@@ -47,7 +44,52 @@ export function FormCreate(props: ObjectInputProps) {
...
+      } else {
+        setActiveTab('subscribed')
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [creating, savedTask?.createdByUser])
+
+  useEffect(() => {
+    // If after 10 seconds the task is still in the "creating" state, show an error and reset the creating state.
+    let timeoutId: NodeJS.Timeout | null = null
```
bjoerge, line 73: We should avoid references to NodeJS types since this targets a browser runtime

```suggestion
let timeoutId: ReturnType<typeof setTimeout> | null = null
```

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
bjoerge: If possible, I think this would be more robust. Note that `useDocumentOperationEvent` will also trigger for work done by collaborators (might not be an issue here)

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
```diff
@@ -47,7 +44,52 @@ export function FormCreate(props: ObjectInputProps) {
...
+        toast.push({
+          closable: true,
+          status: 'error',
+          title: t('form.status.error.creation-failed'),
+        })
+      }, 10000)
+    }
+    // Cleanup function to clear the timeout
+    return () => {
+      if (timeoutId) clearTimeout(timeoutId)
```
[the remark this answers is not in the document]
bjoerge: Oh, nvm, I was [wrong](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout):
> The returned timeoutID is a positive integer value which identifies the timer created by the call to setTimeout()

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormCreate.tsx
pedrobonamin: Just tried with `useDocumentOperation` event and it returns an object like this:
```
{
    "type": "success",
    "op": "commit",
    "id": "0a6836ee-ec1d-4f87-b1c7-bd7d3b668e5d"
}
```

It will return the same object for every operation that is done by the form, so if you are typing or changing any value it will continue showing the same values. 
Would be great if there is a way to handle this, is there any option to get access to the transaction that is executed on that operation?

bjoerge: Good point. Transaction IDs are generated as the last step before mutations are submitted to the backend, so there's no way of knowing what they will be at the time the operation is executed.

bjoerge (approved):

pedrobonamin:
@sjelfull dismissing the review given the Pr has been approved and I want to have this merged for tomorrows release, happy to revert later or address the suggestions if you have any, but I think they have been answered and covered with the last changes.

PR #6457, opened 2024-04-23
fix(typegen): fixes a bug where we imported the wrong relative path

### LINKED ISSUE #6297: CLI Typegen doesn't work with imported partial queries
**Describe the bug**

When using groq typegen `sanity schema extract && sanity typegen generate` it is perfectly able to work with queries that are composed out of concatenated string variables. It, however doesn't work, crashes, if any of the variables are imported from another file.

**To Reproduce**

1. Create 2 files - `repo.ts` and `util.ts`
2. Place the following query into the `repo.ts`:

   ```groq
   import { inner } from "./util";
   
   export function marko() {
       const myTestQuery = groq`*[_type == "sanity.imageAsset"] ${inner}`;
       return client.fetch(myTestQuery);
   }
   ```
3. Place the following query segment into the `util.ts` file:

   ```groq
   export const inner = `{ metadata }`;
   ```
4. Run the typegen command: `sanity schema extract && sanity typegen generate`
5. It will crash with an error, saying that it `✖ Cannot find module './util'`

**Expected behavior**

It should behave the same as it does when the inner variable is within the same file as the query - successfully compose the query and its type.

```none
@sanity/cli (global) 3.37.1 (up to date)
@sanity/document-internationalization 3.0.0 (up to date)
@sanity/icons 2.11.8 (up to date)
@sanity/image-url 1.0.2 (up to date)
@sanity/presentation 1.12.3 (up to date)
@sanity/ui 2.1.1 (up to date)
@sanity/vision 3.37.1 (up to date)
@sanity/webhook 4.0.3 (up to date)
sanity 3.37.1 (up to date)
```

macOS Sonoma

```none
10.2.4
v20.11.1
```

sgulseth:
### Description

We we're checking the wrong variable for relative imports. importName is the variable being imported, but we want to check the actual path. The tests were also wrong...

### Testing

Tests fixed

### Notes for release

typegen: fixes a bug where relative imports would fail.

Fixes #6297

### DIFF
--- a/packages/@sanity/codegen/src/typescript/expressionResolvers.ts
+++ b/packages/@sanity/codegen/src/typescript/expressionResolvers.ts
@@ -323,12 +323,14 @@ function resolveImportSpecifier({
     throw new Error(`Could not find import declaration for ${node.local.name}`)
   }
 
-  const importName = node.local.name
-  const importFileName = importDeclaration.source.value
-  // const importPath = path.resolve(path.dirname(filename), importFileName);
-  const importPath = importName.startsWith('./')
-    ? path.resolve(path.dirname(filename), importFileName)
-    : importFileName
+  const importName = node.local.name // the name of the variable to import
+  const importFileName = importDeclaration.source.value // the file to import from
+
+  const importPath =
+    importFileName.startsWith('./') || importFileName.startsWith('../')
+      ? path.resolve(path.dirname(filename), importFileName)
+      : importFileName
+
   const resolvedFile = resolver(importPath)
   const source = fs.readFileSync(resolvedFile)
   const tree = parseSourceFile(source.toString(), resolvedFile, babelConfig)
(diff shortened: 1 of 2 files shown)


packages/@sanity/codegen/src/typescript/expressionResolvers.ts
```diff
@@ -323,12 +323,14 @@ function resolveImportSpecifier({
...
-  const importFileName = importDeclaration.source.value
-  // const importPath = path.resolve(path.dirname(filename), importFileName);
-  const importPath = importName.startsWith('./')
-    ? path.resolve(path.dirname(filename), importFileName)
-    : importFileName
+  const importName = node.local.name // the name of the variable to import
+  const importFileName = importDeclaration.source.value // the file to import from
+
+  const importPath =
+    importFileName.startsWith('./') || importFileName.startsWith('../')
```
tzhelyazkova, line 330: hmmm, could it also start with `@` ?

tzhelyazkova (approved):
LGTM, just one question below.

packages/@sanity/codegen/src/typescript/expressionResolvers.ts
sgulseth: Yes, but that's not a relative path 🙂

PR #6475, opened 2024-04-24
feat(typegen): also search for queries in app and sanity folders

sgulseth:
### Description

* Expands the default search path for finding queries
* Format the generated code with Prettier

### What to review

Correctness 

### Testing

Tested manually, and added a test that checks if we can overwrite the default babel double quotes.

### Notes for release

Typegen: Expand the default search paths to also include `app` and `sanity` folders
Typegen: Format generated code with prettier, if a prettier config file is found.


stipsan (approved):
Very nice 💖 With this we get to delete this file: https://github.com/vercel/next.js/blob/canary/examples/cms-sanity/sanity-typegen.json

PR #6478, opened 2024-04-24
chore(sanity): remove beta and unstable from comments and tasks

RitaDias:
### Description

Remove the beta and unstable_ from comments and tasks feature.

A request for the change in docs has also been asked.

### What to review

Make sure that the comments and tasks are working as expected before the changes.

### Testing

Tests should be eventually written, however for now the two following are vital for manual test:
In the sanity.config for the test-studio
- Tasks: false & unstable_tasks: true (the tasks should not be visible)
- Comments: false & unstable_comments: true (the comments should not be visible)

### Notes for release

Comments and Tasks are now out of beta


dev/test-studio/sanity.config.ts
```diff
@@ -90,7 +90,9 @@ const sharedSettings = definePlugin({
       return prev
     },
     newDocumentOptions,
-
+    comments: {
+      enabled: true,
+    },
     unstable_comments: {
       enabled: true,
     },
```
hermanwikner, line 98: I think we can remove this `unstable_comments` from the test-studio config

packages/sanity/src/core/config/types.ts
```diff
@@ -291,10 +291,15 @@ export interface DocumentPluginOptions {
    */
   newDocumentOptions?: NewDocumentOptionsResolver
 
-  /** @internal */
+  /** @deprecated - use `comments`instead */
```
hermanwikner, line 294: Nitpick: should we align this deprecation message with the one for Tasks – or vice versa?

```diff
- /** @deprecated - use `comments` instead */
+ /** @deprecated  Use `comments` instead */
```

hermanwikner: I prefer ```/** @deprecated  Use `comments` instead */``` since the dash is added by JSDoc. 

Screenshot 2024-04-24 at 17 45 58

pedrobonamin (approved):
Thanks for this @RitaDias !

cngonzalez (approved):

PR #6487, opened 2024-04-25
chore: setup next test that embeds the test studio

stipsan:
### Description

This PR changes the `dev/test-next-studio` to no longer run specialized SSR hydration tests, and instead embed the same `sanity.config` as `dev/test-studio`.
We have better SSR hydration tests setup elsewhere, it's much more useful to be able to run the same test studio in `sanity dev` and `next dev`, as it gives us the ability to:
- debug problems or regressions that affect embedded studios, as Next.js is using SSR and is sensitive to hydration mismatches while `sanity dev` and `sanity build` is not.
- Next.js is running canary releases of upcoming versions of React, it lets us test that core Studio APIs are future compatible in a much easier way than if we were to [set this up on our own](https://react.dev/blog/2023/05/03/react-canaries).

### What to review

The setup should make sense, and you should be able to run this locally by using:
```bash
pnpm install
cd dev/test-next-studio
pnpm dev
```
And opening `http://localhost:3333` like before.
There's also a [Vercel Deployment](https://vercel.com/sanity-io/test-next-studio/deployments) that will use https://test-next-studio.sanity.build after this PR merges, but can be tested with https://vercel.live/link/test-next-studio-git-setup-test-next-studio-in-app-router.sanity.build?via=deployment-domains-list-branch for now.

### Testing

N/A

### Notes for release

N/A


bjoerge (approved):
Great stuff, thank you @stipsan 💖

PR #6494, opened 2024-04-26
fix(test): remove hydrateroot warning test

bjoerge:
### Description

Removes a test asserting SSR hydrateRoot finishes without warnings. This test started failing after upgrading to React 18.3 due to a deprecation warning when importing `act` react/test-utils.

Now that #6487 is merged we have a next build of the entire test studio which should cover a lot more SSR/hydration issues, so removing this test for now.

### What to review

- Do we still need the other tests here?

### Testing
n/a

### Notes for release
n/a


stipsan (approved):
Yup, the `http://test-next-studio.sanity.build/` deployment, which adds checks to PRs and tests `next build` as well, covers a much wider scope than this test when uncovering SSR mismatch issues. And also SSR in general since `jest` is a lot more permissive in what dependencies it allows and what globals are available than a node SSR env :)

PR #6505, opened 2024-04-26
chore(deps): upgrade to react 18.3.1

rexxars:
### Description

Upgrades to use React 18.3(.1) in our dev and example studios.

This should help us reveal any use of APIs that are deprecated/will change in React 19.


pedrobonamin (approved):
Looks good to me. 
Any reason why we are not updating https://github.com/sanity-io/sanity/blob/chore/react-upgrade/dev/test-next-studio/package.json#L16 to latest?

rexxars:
> Any reason why we are not updating https://github.com/sanity-io/sanity/blob/chore/react-upgrade/dev/test-next-studio/package.json#L16 to latest?

Just an oversight. Will fix.

PR #6511, opened 2024-04-29
fix(core): skip reset presence selection on PTE blur

hermanwikner:
### Description

This pull request resolves an issue where the popover for editing objects in the PTE closes immediately after opening. This issue was introduced with the implementation of presence cursors, where the `onPathFocus` function is called to reset the `selection` presence state when the PTE input is blurred. The fix here is to avoid resetting the presence selection on blur, and instead, allow the selection to be reset when another input is focused, thereby updating the presence.

### What to review
- Ensure that editing objects in the PTE (such as links) is possible.

### Notes for release

N/A


jordanl17:
@hermanwikner this makes sense to me, and have tested to verify it's now resolved, but @skogsmaskin probably best to ratify

skogsmaskin (approved):
👍

PR #6513, opened 2024-04-29
fix(scheduled-publishing): deprecated error toast shows multiple times

pedrobonamin:
### Description
The deprecated `scheduledPublishing` toast in some cases shows multiple times, adding an id to it fixes it and ensures it will only show once.


binoy14 (approved):
great catch, thanks!

PR #6516, opened 2024-04-29
feat: add canHandleIntent to Structure Builder component

cngonzalez:
### Description
Some users would like to implement their own custom component within Structure, rather than just a desk list. However, this becomes difficult when trying to route to this component from other parts of the Studio, because there was no way for us to signal that this component could indeed handle that intent.

This PR takes the `canHandleIntent` argument available on lists and makes it available on components.

### What to review
Is this a valid approach? Does it break inheritance of Structure Builder items in any way? Are there any gotchas?

### Testing
This can be tried out in the test studio by adding a `canHandleIntent` to an `S.component`, like so:
```javascript
      S.listItem()
        .id('translate')
        .title('Translate Test')
        .child(S.component(TranslateExample).id('example')
          .canHandleIntent((intent, params, context) =>  {
            console.log(intent, params, context)
            return true
          }
        )),
```

You should be able to run the Studio without crashing. When searching for a document, you should land on the `canHandleIntent` component.

A unit test is also included as part of this PR.

### Notes for release
Custom components included in Structure (like `S.component(MyCustomComponent)`) can now be routed to from global search results or other links by using a `canHandleIntent` parameter.

### DIFF
--- a/packages/sanity/src/structure/structureBuilder/Component.ts
+++ b/packages/sanity/src/structure/structureBuilder/Component.ts
@@ -1,5 +1,6 @@
 import {type I18nTextRecord} from 'sanity'
 
+import {type IntentChecker} from './Intent'
 import {maybeSerializeMenuItem, type MenuItem, type MenuItemBuilder} from './MenuItem'
 import {
   maybeSerializeMenuItemGroup,
@@ -34,7 +35,7 @@ export interface Component extends StructureNode {
   menuItemGroups: MenuItemGroup[]
   /** Component options */
   options: {[key: string]: unknown}
-  canHandleIntent: any
+  canHandleIntent?: IntentChecker
 }
 
 /**
@@ -71,7 +72,7 @@ export interface BuildableComponent extends Partial<StructureNode> {
   menuItems?: (MenuItem | MenuItemBuilder)[]
   /** Component menu item groups. See {@link MenuItemGroup} and {@link MenuItemGroupBuilder} */
   menuItemGroups?: (MenuItemGroup | MenuItemGroupBuilder)[]
-  canHandleIntent?: any
+  canHandleIntent?: IntentChecker
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/sanity/src/structure/structureBuilder/Component.ts
```diff
@@ -205,6 +208,10 @@ export class ComponentBuilder implements Serializable<Component> {
     return this.spec.menuItemGroups
   }
 
+  canHandleIntent(canHandleIntent: any): ComponentBuilder {
```
juice49, line 211: I think the `canHandleIntent` parameter should be typed as [`IntentChecker`](https://github.com/sanity-io/sanity/blob/9c86166ee63064d3a2c714d8a89a380ed668f9bc/packages/sanity/src/structure/structureBuilder/Intent.ts#L85-L96)?

```suggestion
  canHandleIntent(canHandleIntent: IntentChecker): ComponentBuilder {
```

juice49:
I tested this in Studio, and it appeared to work correctly.

One thing I noticed is that the `UserComponent` doesn't seem to receive details about the intent (e.g. the intent type and document id) in its props.

The document id can be found via the router state (`useRouterState`), but it's a bit fiddly, and the developer must make the assumption that the final pane is the one containing the document (I'm not 100% sure this is always correct).

```ts
{
  "panes": [
    [
      {
        "id": "custom",
        "params": {}
      }
    ],
    [
      {
        "id": "component1",
        "params": {}
      }
    ],
    [
      {
        "id": "73bef166-639e-4313-bb5f-bbc5353869eb",
        "params": {}
      }
    ]
  ]
}
```

Is there a better way for developers to get details about the handled intent in their component? 🤔

packages/sanity/src/structure/structureBuilder/Component.ts
cngonzalez: 🤦 thank you! too hasty to commit

cngonzalez:
> Is there a better way for developers to get details about the handled intent in their component? 🤔 I suspect this would generally be required to make useful intent-handling components. I'm wondering whether we should extend `UserComponent` to include the handled intent in its props as part of this change.

It's a great question, I'm not sure! There's quite a bit of logic to get the information into the `pane` parameter in the props from the router (I think?). I can look into replicating that.

juice49:
@cngonzalez Not to say that it's a blocker, necessarily. Just something I imagine developers might need in order to make use of this functionality.

PR #6521, opened 2024-04-30
fix(pte): preserve block key when pressing enter at start of block

pedrobonamin:
### Description
fixes edx-1056.

**Changes:**
- Adds new "plugin" `withInsertBreak` which handles the case of pressing enter at the start of a block.
- Moves `createEmptyBlock` function to the editor utils, allowing ease of reuse.

**Issue:**
When moving a block by placing the cursor at the beginning of a block and pressing “Enter” – a new `_key` is currently generated for that block. This is problematic because comments are linked to specific blocks through the `_key`. If a new `_key` is generated when moving the block in this way, all comments associated with that block will disappear. This occurs because the `_key` value in the comment document no longer references any block content by its `_key`.

This PR updates the behaviour when pressing enter key in the PTE editor, if the cursor is at the start of the block, instead of doing a `split_node` operation it will do a  `insert_node` operation. This will preserve the key of the current block.

**Challenge**: focus handling is not automatically updated by Slate and it keeps the cursor in the new inserted block instead of the block the user was. This is fixed by this change , forcing the focus path to the next block, which is the actual block, because a new block was inserted in the previous position https://github.com/sanity-io/sanity/pull/6521/files#diff-35c7a342daeb53013561d14759571db71e13c8782293a91e14f254608dbdb439R29-R32

It can be tested here https://test-studio-git-edx-1056.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;e93f598b-8ec4-4bed-87aa-ab855210ed0e


packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withHotKeys.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const editor = editorRef.current
+    const inlineType = editor?.schemaTypes.inlineObjects.find((t) => t.name === 'someObject')
+    await waitFor(async () => {
+      if (editor && inlineType) {
+        PortableTextEditor.focus(editor)
+        PortableTextEditor.select(editor, initialSelection)
+
+        const textBox = screen.getByRole('textbox')
+        if (!textBox) throw new Error('No textBox found')
+        fireEvent.keyPress(textBox, {key: 'Enter', code: 'Enter', charCode: 13})
```
pedrobonamin, line 65: This is not working, not sure how to make it work inside the PTE editor. 
The intention is to position the cursor at the start of the block and press enter.

packages/@sanity/portable-text-editor/e2e-tests/__tests__/writingTogether.collaborative.test.ts
```diff
@@ -669,8 +669,8 @@ describe('collaborate editing', () => {
     `)
     const selectionA = await editorA.getSelection()
     expect(selectionA).toEqual({
-      anchor: {path: [{_key: 'A-8'}, 'children', {_key: 'A-7'}], offset: 0},
-      focus: {path: [{_key: 'A-8'}, 'children', {_key: 'A-7'}], offset: 0},
+      anchor: {path: [{_key: 'A-6'}, 'children', {_key: 'A-5'}], offset: 0},
```
pedrobonamin, line 672: Not sure about this, debugging

pedrobonamin, line 672: Not sure how the user A was in `{path: [{_key: 'A-8'}, 'children', {_key: 'A-7'}], offset: 0}` before given that block doesn't exists, and why now it is corrected. Any thoughts @skogsmaskin ?

pedrobonamin: Updated the snapshot to reflect the change, snapshot compared was the previous value not the new one after the second press on the enter key

skogsmaskin (approved):
Solid work @pedrobonamin!

Thank you for this.

PR #6524, opened 2024-04-30
fix(pte): tools are active only when all blocks use the tool

jordanl17:
### Description

EDX-801

### What to review

Different approaches are needed for each of marks, annotations and lists
The high level approach is that instead of a condition on `some` of the selected blocks being bold, or a link, or bullet list, now `every` selected block must be.

The most cumbersome implementation is for annotations (links). This requires a new `isAnnotationActive` method in  `packages/@sanity/portable-text-editor/src/editor/plugins/createWithEditableAPI.ts` which must expand every selected block to determine the type of mark definitions in the selection, before then applying the `every` logic to all spans in the selection

### Notes for release

fixes issues in the PTE where tools would show as being active for a selection when only part of the selection used the tool.
Now tools will show as active only when all highlighted selections use the tool


packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withPortableTextMarkModel.test.tsx
```diff
@@ -360,23 +360,25 @@ describe('plugin:withPortableTextMarksModel', () => {
...
-                "marks": Array [],
-                "text": "1234",
-              },
-            ],
-            "markDefs": Array [],
-            "style": "normal",
-          },
-        ]
-      `)
+Array [
```
jordanl17, line 363: The behaviour has changed here.
Previously if I highlighted '**1**23**4**' and then toggled bold, it would un-un-bold the full selection.
Now it will bold the full selection (which is reflective of the way that highlighting '**1**23**4**' will not show the bold tool as active since the full selection is not bold

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextLists.ts
```diff
@@ -143,11 +143,14 @@ export function createWithPortableTextLists(types: PortableTextMemberSchemaTypes
       const selectedBlocks = [
         ...Editor.nodes(editor, {
           at: editor.selection,
-          match: (node) => editor.isListBlock(node) && node.listItem === listStyle,
```
jordanl17, line 146: On the face of it, this seems identical, but `match` essentially filters all blocks in the selection, which then conceals whether any of these blocks are *not* of a list type

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextLists.ts
```diff
@@ -143,11 +143,14 @@ export function createWithPortableTextLists(types: PortableTextMemberSchemaTypes
...
+          match: (node) => editor.isTextBlock(node),
         }),
       ]
+
       if (selectedBlocks.length > 0) {
-        return true
+        return selectedBlocks.every(
```
jtpetty, line 151: do we need to return `every` here or do we just care that there is `> 0` blocks that meet this condition?

jordanl17: We do care that every block has the list style. For example if the selection was on the following:

> First block
>  - List second block
>  
> Third block

we'd have 3 `selectedBlocks`, but only 1 would have bullet list style, so we'd want to return false

packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withPortableTextMarkModel.test.tsx
jordanl17: Ah yes, let me look into what I can add in terms of testing

jtpetty (approved):
👍 - thanks for adding more tests!

PR #6535, opened 2024-04-30
fix: use discard in delete action when document is not published

cngonzalez:
### Description
The `delete` action in server side document actions requires a published document. In those cases when there is not a published document, we can't use this endpoint.

This PR adds logic for those cases by using the discard endpoint instead, and adds a few tests.

### What to review
That the logic makes sense and this is how we want to proceed. We can always work with other teams to ensure we're doing something semantic that won't hinder us later.

### Testing
This PR adds two tests for the delete action.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/delete.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/delete.ts
@@ -3,15 +3,32 @@ import {isLiveEditEnabled} from '../utils/isLiveEditEnabled'
 
 export const del: OperationImpl<[], 'NOTHING_TO_DELETE'> = {
   disabled: ({snapshots}) => (snapshots.draft || snapshots.published ? false : 'NOTHING_TO_DELETE'),
-  execute: ({client: globalClient, schema, idPair, typeName}) => {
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/delete.ts
```diff
@@ -3,16 +3,33 @@ import {isLiveEditEnabled} from '../utils/isLiveEditEnabled'
...
     const vXClient = globalClient.withConfig({apiVersion: 'X'})
-
     const {dataset} = globalClient.config()
 
+    //the delete action requires a published doc -- discard if not present
+    if (!snapshots.published) {
+      return vXClient.observable.request({
+        url: `/data/actions/${dataset}`,
+        method: 'post',
+        tag: 'document.delete',
```
cngonzalez, line 20: I've maintained the tag as delete so we have a sense of what the user's intent was. We can change it to discard if we don't like this choice.

bjoerge: I don't see a problem with changing this to "document.discard", The fact that the button says "Delete" in this case feels more coincidental, and you could argue that it might as well say "Discard draft" when there is no published document.

PR #6537, opened 2024-05-01
fix(sanity): do not order by `_updatedAt` when relevance ordering is used with Text Search API search strategy

juice49:
### Description

**This branch relates only to global search.**

Ordering by relevance ("best match") when using the GROQ Query API search strategy causes the order `{direction: 'desc', field: '_updatedAt'}` to be applied. This is correct, because the results are subsequently ordered client-side.

**However, this is incorrect when using the Text Search API search strategy.** The Text Search API yields results ordered by relevance by default, and we do not apply any client-side ordering.

This branch eliminates the incorrectly applied `{direction: 'desc', field: '_updatedAt'}` ordering when the Text Search API search strategy is used.

### What to review

- `order(_updatedAt desc)` is still present in the GROQ Query API request when ordering by relevance using the GROQ Query API search strategy.
- `{order: [{"attribute": "_updatedAt", "direction": "desc"}]` is not present in the Text Search API request when ordering by relevance using the Text Search API search strategy.

### Testing

Added unit tests to `packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.test.ts`.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
@@ -34,7 +34,7 @@ export function SearchProvider({children, fullscreen}: SearchProviderProps) {
   const schema = useSchema()
   const currentUser = useCurrentUser()
   const {
-    search: {operators, filters},
+    search: {operators, filters, enableLegacySearch},
   } = useSource()
 
   // Create field, filter and operator dictionaries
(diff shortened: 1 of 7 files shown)


ricokahler (approved):
nice work!

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
```diff
@@ -112,9 +120,13 @@ export function SearchProvider({children, fullscreen}: SearchProviderProps) {
     const termsChanged = !isEqual(terms, previousTermsRef.current)
 
     if (orderingChanged || cursorChanged || termsChanged) {
-      // Use a custom label if provided, otherwise return field and direction, e.g. `_updatedAt desc`
-      const sortLabel =
-        ordering?.customMeasurementLabel || `${ordering.sort.field} ${ordering.sort.direction}`
+      let sortLabel = 'findability-sort:'
```
jordanl17, line 123: Nice little refactor here 👍

jordanl17 (approved):

PR #6543, opened 2024-05-02
fix(schedule-publishing): update flag used to check scheduledPublishing

pedrobonamin:
### Description
The flag used to check the availability of scheduled publishing is incorrect, it should check `scheduledPublishing`, not `sanityTasks`

### Notes for release
Fixes an issue in which users were incorrectly seeing the schedule publishing upsell ui.


jordanl17 (approved):
LGTM

PR #6544, opened 2024-05-02
fix: use `vercelStegaClean` util from `@vercel/stega`

stipsan:
### Description

Follows up on the same change as in `@sanity/client`: https://github.com/sanity-io/client/pull/773
The new utility is faster, and should improve the perf when pasting large amounts of text.

### What to review

If tests pass then it works.

### Testing

It's sufficient to use existing tests.

### Notes for release

N/A


RitaDias (approved):
Thanks for setting this up :)

PR #6552, opened 2024-05-02
fix(pte): insert empty text block after removing void block 

pedrobonamin:
#### Issue
After clearing a PTE field with blocks, there is a very short period where attempting to input text after manually re-focusing will yield errors `(Cannot get the start point in the node at path [] because it has no start text node)` as well as rendering text incorrectly.

Screenshot 2024-05-02 at 18 22 16

This is caused because the editor gets an "empty" state, which is later restored to the placeholder state with an empty text block.

#### Fix

Instead of waiting on the editor to restore itself after getting the empty state, avoid going to that state by inmediately adding an empty text block when removing the non text block if it's necessary.
It is necessary when this block is the only available. 
If we have another void or text block the editor won't be empty.

### What to review

Is this change correct?

### Testing
Manual testing of the changes.
Unit tests covering the case.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/components/SlateContainer.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/components/SlateContainer.tsx
@@ -64,7 +64,7 @@ export function SlateContainer(props: SlateContainerProps) {
   }, [keyGenerator, portableTextEditor, maxBlocks, readOnly, patches$, slateEditor])
 
   const initialValue = useMemo(() => {
-    return [slateEditor.createPlaceholderBlock()]
+    return [slateEditor.pteCreateEmptyBlock()]
   }, [slateEditor])
 
   useEffect(() => {
--- a/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
+++ b/packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
@@ -106,7 +106,7 @@ export function useSyncValue(
                   at: [childrenLength - 1 - index],
                 })
               })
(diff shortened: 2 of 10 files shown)


RitaDias (commented):
This looks good to me, though I want to see what PK has to say :)

skogsmaskin:
I think it would make more sense to make this part of the `withPlaceholderBlock` plugin?

Also, I just discovered we have some overlapping func. on `editor.pteCreateEmptyBlock` and `editor.createPlaceholderBlock`.

I'd love to see these being the same thing. `createPlaceholderBlock` could be replaced by `pteCreateEmptyBlock` and then remove `createPlaceholderBlock` from `withPlaceholderBlock` and use `pteCreateEmptyBlock` instead to create the block, but then add this normalization rules in the `withPlaceholderBlock` plugin.

pedrobonamin:
> I think it would make more sense to make this part of the `withPlaceholderBlock` plugin?
> 
> Also, I just discovered we have some overlapping func. on `editor.pteCreateEmptyBlock` and `editor.createPlaceholderBlock`.
> 
> I'd love to see these being the same thing. `createPlaceholderBlock` could be replaced by `pteCreateEmptyBlock` and then remove `createPlaceholderBlock` from `withPlaceholderBlock` and use `pteCreateEmptyBlock` instead to create the block, but then add this normalization rules in the `withPlaceholderBlock` plugin.

@skogsmaskin  Thanks for finding this, just updated the PR with the new changes

skogsmaskin (approved):
Awesome! Thanks for this @pedrobonamin!

PR #6553, opened 2024-05-02
fix: inline `swr`, `date-fns-tz` and `@vvo/tzdb` to restore embedded studios

stipsan:
### Description

The deps changes in https://github.com/sanity-io/sanity/pull/6416 unfortunately broke some embedded studios. We don't have a reliable reproduction (I tried all the tricks I know) but we have enough user reports of embedded next studios no longer working if they upgrade from v3.38.0 and to any versions past that. Errors are usually in the form of:
```
Attempted import error: 'swr' does not contain a default export (imported as 'useSWR').
```
The fix is to inline these dependencies, which is the tried and true fix we've used in similar cases in the past.
We've asked users that are experiencing this issue to submit a repro, once that happens I'll open another PR with a similar `test-studio` as `test-next-studio` that can act as a guard against this happening again in the future.

### What to review

Everything in Scheduled Publishing should work as before, as we're just inlining modules we're not changing anything else.

### Testing

### Notes for release

Fixes a `Attempted import error` related to using `swr`


rexxars (commented):
Not super happy about the lack of repro for this, but not going to let that block

rexxars (approved):

PR #6559, opened 2024-05-03
POC(typegen): Add ability to format schematype names

largis21:
### Description

Extra config property for customizing type names on schema types.

Feedback on this feature would be appriciated

Example:
```typescript
//sanity-typegen.json

{
  "path": "'./**/*.{ts,tsx,js,jsx}'",
  "schema": "schema.json",
  "generates": "sanity.types.ts",
  "format": {
    "schemaTypes": {
      "literal": "i{name}Schema",
      "nameCase": "pascal"
    },
    "queries": {
      "literal": "i{name}query",
      "nameCase": "pascal"
    }
  }
}
```

```typescript
// schema.ts

defineType({
  name: "documentName",
  type: "document"
})
```

```typescript
// Old result

export type DocumentName = {...}
```

```typescript
// New result

export type iDocumentNameSchema = {...}
```

### What to review

The way the formatting is done, I don't know if it is the best way to do it.
@sanity/cli @sanity/codegen is affected

### Testing

No, this is only a proof of concept

### DIFF
--- a/packages/@sanity/cli/src/actions/typegen/generateAction.ts
+++ b/packages/@sanity/cli/src/actions/typegen/generateAction.ts
@@ -76,6 +76,7 @@ export default async function typegenGenerateAction(
       schemaPath: codegenConfig.schema,
       searchPath: codegenConfig.path,
       prettierConfig,
+      format: codegenConfig.format,
     } satisfies TypegenGenerateTypesWorkerData,
     // eslint-disable-next-line no-process-env
     env: process.env,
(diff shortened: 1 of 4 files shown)


largis21:
Planning on adding support for queries as well, havent gotten that far yet. There is also a bug where the capitalization is wrong because of the order of the format functions.

sgulseth:
Hi! Thanks for the PR, we've been going a bit back and forth with the naming, and our long/short term plan is to move the configuration into `sanity.cli.ts` to reduce configuration fatigue.
This would allow us to add additional overrides methods for generating the type name in {java,type}script.
To reduce the overhead needed to then migrate we're gonna wait with this change.

PR #6560, opened 2024-05-03
feat(pte): create new text blocks if needed

pedrobonamin:
### Description

This changes improve the editor experience by adding empty text blocks into PTE when needed.
####  A new block is added if the user is focused on the last block which is void, and clicks in the editor.

https://github.com/sanity-io/sanity/assets/46196328/623e28aa-f5e6-47b2-89e2-64142c1995bd

#### A new block is added if the user is focused on the last block which is void, and presses arrow down.

https://github.com/sanity-io/sanity/assets/46196328/0425bee3-4de8-4f7b-ab7f-e6e62c2fdb4a

####  A new block is added at the top, when pressing arrow up, because first block is void, the new block can be deleted with backspace.

https://github.com/sanity-io/sanity/assets/46196328/a241fdf4-50ee-49c1-81c9-354321dc848c

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/Editable.tsx
@@ -19,6 +19,7 @@ import {
 import {
   type BaseRange,
   Editor,
+  Node,
   type NodeEntry,
   type Operation,
   Path,
@@ -50,6 +51,7 @@ import {
   type ScrollSelectionIntoViewFunction,
 } from '../types/editor'
 import {type HotkeyOptions} from '../types/options'
+import {type SlateTextBlock, type VoidElement} from '../types/slate'
 import {debugWithName} from '../utils/debug'
 import {moveRangeByOperation, toPortableTextRange, toSlateRange} from '../utils/ranges'
 import {normalizeSelection} from '../utils/selection'
@@ -118,6 +120,7 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
     onBeforeInput,
     onPaste,
     onCopy,
+    onClick,
     rangeDecorations,
     renderAnnotation,
     renderBlock,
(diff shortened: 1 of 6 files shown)


jtpetty (commented):
This looks good to me, but would want @skogsmaskin to review as well. Thanks for writing so many tests for this fix.

packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -444,6 +447,30 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
     [onFocus, portableTextEditor, change$, slateEditor],
   )
 
+  const handleClick = useCallback(
+    (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      if (onClick) {
+        onClick(event)
+      }
+      // Inserts a new block if it's clicking on the editor, focused on the last block and it's a void element
+      if (slateEditor.selection && event.target === event.currentTarget) {
```
RitaDias, line 456: PK will likely be the best person, but I wonder if this could be something that was moved to a normalise function :)

pedrobonamin: Curious to that answer, but I think it's not possible. as this is handling the onClick event.
It tries to solve the action in which the user clicks below the void block and nothing is added, so he can't continue typing.
The first part of the video is current behaviour, second part is with the changes. 

https://github.com/sanity-io/sanity/assets/46196328/4e604259-1eec-4f64-a1c3-1be969c21e3f

skogsmaskin: It's a timely question @RitaDias!

However in this case I think we don't want that. We want it to be tied to the UI-event of the user trying to click of move below with the intention of creating a new block.

We don't in all cases want it to always insert a empty block there (that would mean it's impossible to end a editor with an object block).

PR #6563, opened 2024-05-03
fix: relay document listeners

bjoerge:
### Description
This implements a relay approach for listener connections:
- When we now set up a listener, we also schedule a new listener connection after somewhere between 12-19 minutes.
- We’ll keep the old listener connected for 20s after the new one has connected and has received the `welcome`-event - (in case there’s in-flight messages yet to arrive).
- There’s a distinct filtering on eventId so we avoid processing the same message twice.

These steps are repeated indefinitely (for as long as needed).

### What to review
The intervals are currently hard coded, so the best way to test is probably to check out this branch, and edit the values here to something shorter (e.g. 10-20s with 4s overlap) and observe that a new listener is set up before the old one is shut down.

https://github.com/sanity-io/sanity/blob/4eedef7d5d8809032178c716c5ec120e9f020835/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts#L196-L202

It might be a bit hard to unpack the code, but I believe the logic is correct and there's tests to verify that it behaves correctly.

Since this can potentially make it easier to hit the the max number of concurrent connections allowed by browsers on http/1. I didn't include a config flag to disable this right now, but it should be quite straightforward to add a config flag if we wish to do so. For now, there's a timeout on the second listener that will cancel it if it's not able to connect within a certain time. If this happens, the first listener will keep runnning.

- Since the server now should allow connections to live for a lot longer – should we wait for longer before exchanging the current one with a new one?

### Testing
Added unit tests asserting that:
- Listener setup, handoff and teardown works as it should
- De-duping listener events based on `eventId` works
- Timeout on "next" leg keeps the current
- It waits for the first connection to be set up and have received the welcome event before scheduling the next one

### Notes for release
N/A - It should be transparent for the user.


packages/sanity/src/core/store/_legacy/document/createPairListener.ts
```diff
@@ -141,6 +150,99 @@ export function getPairListener(
...
+          if (event.type !== 'mutation') {
+            return [seen, event]
+          }
+          if (seen.has(event.eventId)) {
+            return [seen, null]
+          }
+          seen.set(event.eventId, true)
+          return [seen, event]
+        },
+        [new LRU<string, boolean>({maxSize: 1000}), null],
```
juice49, line 239: To make sure I understand: this means we'll buffer up to 1,000 messages per listener to deduplicate during the 20 second exchange overlap. Is my understanding correct?

Related:

1. I assume the memory impact of buffering (up to) 1,000 messages per listener is inconsequential. If this number was picked arbitrarily, could it be beneficial to decrease it?
2. I assume that it's possible (but unlikely) for duplicate events to be emitted if more than 1,000 messages are received during the 20 second exchange overlap. Is it problematic if that occurs? My guess is everything would be okay, and the purpose of deduping is more to ensure listeners behave in the way consumers expect.

bjoerge: Great questions.

> Is my understanding correct?

Yes. The key here is "up to" - in reality it's unlikely to be more than a handful events.

> I assume the memory impact of buffering (up to) 1,000 messages per listener is inconsequential. If this number was picked arbitrarily, could it be beneficial to decrease it?

I'd expect that to be extremely rare, so 1000 felt more like a "safe upper limit"

> I assume that it's possible (but unlikely) for duplicate events to be emitted if more than 1,000 messages are received during the 20 second exchange overlap. Is it problematic if that occurs?

If it occurs it will likely lead to problems (the same diffMatchPatch applied twice leads to wrong content), so you made me realize this buffer should be time bound not size bound. I'll refactor this. Thanks!

juice49: Got it. Thanks for explaining!

packages/sanity/src/core/store/_legacy/document/createPairListener.ts
```diff
@@ -25,9 +37,19 @@ export interface InitialSnapshotEvent {
...
 }
 
+interface RelayOptions {
+  exchangeWaitMin: number
+  exchangeWaitMax: number
+  exchangeOverLapTime: number
+  exchangeTimeout: number
+}
+
+const RELAY_SECOND_LEG_TIMEOUT = 20 * 1000
```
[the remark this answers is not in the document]
bjoerge: Good catch, thanks! Fixed in 6eeb47cfc70258db073032c11fb2b8b3d03863e6

packages/sanity/src/core/store/_legacy/document/createPairListener.ts
bjoerge: Pushed a refactor in a8dc8b1c6ba203e697fc0c3839f7fa03f5e3a0c1 that will dedupe listener events based on their eventId for the whole of the overlap period duration. When the overlap period ends, it will clear the set of seen event ids and stop the deduping.

juice49 (approved):
This looks great!

cngonzalez:
Removing self from review after speaking to Bjoerge -- because of some backend changes, this PR may no longer be as necessary.

bjoerge:
Closing for now as this should not be needed anymore. Happy to revisit later if it turns out to be a valuable change.

PR #6565, opened 2024-05-03
refactor(@sanity): remove warning for when child within PTE has no key

RitaDias:
### Description

Working on removing some warnings from PTE when we can self fix it ourselves.
In this case we are removing the warning when a child doesn't have a _key then the PTE should automatically create one for them.

### What to review

Check that it makes sense, there are automated tests.

### Testing

If you choose to do it manually you can create a document via de CLI such as:
```
{
  "_type": "simpleBlock",
  "body": [
    {
      "_key": "a03bf79532e8",
      "_type": "block",
      "children": [
        {
          "_type": "span",
          "marks": [],
          "text": "a test with text"
        }
      ],
      "markDefs": [],
      "style": "normal"
    }
  ],
}
```
when you enter that document in the test studio you will then be able to see that
1. the warning doesn't appear
2. that when inspecting the doc the child in question will have a key

### Notes for release

When child within the PTE is created with a _key, we now automatically create one for it avoiding unnecessary warning.


packages/@sanity/portable-text-editor/src/editor/plugins/createWithSchemaTypes.ts
```diff
@@ -59,6 +60,13 @@ export function createWithSchemaTypes({
         const key = span._key || keyGenerator()
         Transforms.setNodes(editor, {...span, _type: schemaTypes.span.name, _key: key}, {at: path})
       }
+
+      // catches cases when the children are missing keys but excludes it when the normalize is running the node as the editor object
+      if (node._key === undefined && !_.isEqual(node, editor)) {
```
RitaDias, line 65: @skogsmaskin had to add the condition since the tests showed that the editor itself was falling in this condition (since it has no _key itself)

packages/@sanity/portable-text-editor/src/editor/plugins/createWithSchemaTypes.ts
```diff
@@ -6,6 +6,7 @@ import {
   type PortableTextSpan,
   type PortableTextTextBlock,
 } from '@sanity/types'
+import _ from 'lodash'
```
skogsmaskin, line 9: Import only what is needed here `{equals}`

packages/@sanity/portable-text-editor/src/editor/plugins/createWithSchemaTypes.ts
skogsmaskin, line 65: I think it would be better here to do:

```
if (node._key === undefined && (node.path.length === 1 || node.path.length === 2)) {
```

First, we don't need to check for editor, which is by using lodash `isEqual` not that performant.
Also, we only evaluate the nodes we are interested in having a new key (which is block (path length 1) and block child (path length 2).

skogsmaskin (changes requested):
Great with tests for this, and good catch that we need to test the node for what it is.
I have some small comments on this which I'd like to change.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithSchemaTypes.ts
RitaDias: @skogsmaskin good point on the lodash, but for the second part: Just to double check, I had assumed that this warning would have appeared in children that could be "deeper" in the PTE - if that is the case then we are simply limiting this to the first two "levels" within the PTE right? 🤔

skogsmaskin: For each editor there are only block nodes (path length 1) and child nodes (path length 2) that should have this exact normalization.

RitaDias: I see, I had thought that the "path" in this case was related to the number of levels that the children it could have (so it could just go on forever theoretically).

Since that's not the case, I've changed it to your suggestion :)
Thanks for the clarification!

skogsmaskin (approved):
Thanks @RitaDias !

PR #6568, opened 2024-05-03
fix(core): collapsed range decorations

hermanwikner:


packages/@sanity/portable-text-editor/src/editor/Editable.tsx
```diff
@@ -591,6 +591,9 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
       const result = rangeDecorationState.filter((item) => {
         // Special case in order to only return one decoration for collapsed ranges
         if (SlateRange.isCollapsed(item)) {
+          if (path.length !== 2) {
+            return false
+          }
```
hermanwikner, line 596: Can we add a comment here explaining why we need this check?

skogsmaskin (approved):
LGTM! :)

PR #6569, opened 2024-05-03
test(core): add test to ensures that only one listener is open when navigating list

binoy14:
### Description

Adds a test that only one listener is open when navigating document lists

### What to review

Changes makes sense

### Testing

I did write a test to assert that only one connection is kept open but it is not very ideal way but this is a limitation of playwright that it does not support connection close event for long polling request. Fortunately they get marked as failed requests when they get closed which helps us write some assertions around it.

### Notes for release

- N/A

### DIFF
--- a/test/e2e/tests/desk/documentList.spec.ts
+++ b/test/e2e/tests/desk/documentList.spec.ts
@@ -1,0 +1,87 @@
+import {expect, type Request} from '@playwright/test'
+import {test} from '@sanity/test'
+
+function getListenEventType(request: Request): string | null | undefined {
+  if (request.url().includes('data/listen')) {
+    const url = new URL(request.url())
+    return url.searchParams.get('$type')
+  }
+
+  return undefined
+}
+
+test(`navigating document creates only one listener connection`, async ({page}) => {
+  await page.goto('/test/content')
+
+  let authorListenersCount = 0
+  let bookListenersCount = 0
+
+  /**
+   * We listen for requests to the listen endpoint to keep track of how many listeners are active
+   * for each type.
+   */
+  page.on('request', (request) => {
+    const type = getListenEventType(request)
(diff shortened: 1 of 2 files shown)


binoy14:
* **#6569** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6569?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

test/e2e/tests/desk/documentList.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  await page.getByTestId('pane-item-Author').click({force: true})
+  await page.waitForSelector('#author-author-0')
+  await authorRequest
+  expect(bookListenersCount).toBe(0)
+  expect(authorListenersCount).toBe(1)
+
+  // We change the sort order to not be default, to ensure that the listener is not re-created
+  await page.getByTestId('pane-context-menu-button').click()
+  await page.getByRole('menuitem', {name: 'Sort by Name'}).click()
```
jtpetty, line 67: nit: do we have ids that we can use instead of names? just worry about maintenance of this

binoy14: It doesn't currently exists and it's probably little bit out of scope to do it as part of this PR, there is another task to add tests for this menu in general and it's probably better to refactor it part of that. 

The only thing I would also say is when using `getByRole` vs `getByTestId` is that the role ensures that in future changes the a11y rules of a menu are not broken which will get obfuscated by using testid. Since these are default sort options and chances of them changing is less, IMO it's a balance of when to use testId vs using helpers that test functionality as well other HTML spec rules

jtpetty (approved):
👍 LGTM

PR #6579, opened 2024-05-06
feat(cli): remove is-builtin-module

43081j:
This removes the `is-builtin-module` package since all node versions within our `engine` constraint already have such a function built in.

This also allows us to remove the temporary workaround we had around `string_decoder/` since the built in function will not have such a bug.

### What to review

The CLI should be the only affected package.

### Testing

Existing tests should cover this.

### Notes for release

N/A


rexxars (approved):

rexxars:
Thanks, apologies for the delay!

PR #6593, opened 2024-05-07
feat(core, structure) sheet list menu option

jordanl17:
### Description
EDX-1283 - scaffolding the space for the Sheet List view to be built

### What to review
- Inside test-studio a new document list menu option for Sheet list has been added
- Refactor of the existing `DocumentListPane` components, to allow for abstraction of the pane header away from the content

### Testing
The existing tests continue to pass, which given this is mostly a refactor and relocation of existing logic, is a good test to avoid regressions.
More tests to Sheet List will be added once future tasks completed

### Notes for release
Scaffolding the Sheet List view panes, a feature that is WIP and in development


juice49:
This looks good to me! It may be useful to add a test that ensures the `PaneContainer` component renders the correct `paneLayout` depending on the value of the `layout` structure tool setting. What do you think?

jordanl17:
> This looks good to me! It may be useful to add a test that ensures the `PaneContainer` component renders the correct `paneLayout` depending on the value of the `layout` structure tool setting. What do you think?

That seems like a good idea @juice49 - I'll add that in, thanks!

PR #6608, opened 2024-05-08
chore(core): prevent extra work on preview prepare

evensix-dev:
### Description

While validating some preview behaviour, I noticed the `invokePrepare` func being called a second time in the same scope.
This is just a quick amend to use the assignment from the first call.

### What to review

Ensure the system wasn't relying on it being called twice in succession

### Testing

Tests and typesafety should remain the same

### Notes for release

Prevent extra work on preview prepare


juice49 (approved):
Thank you for your contribution! This looks good to me. I'd just like to double check with a team member who's a little more familiar with this code than I am.

bjoerge (approved):
Thank you @evensix-dev!

PR #6595, opened 2024-05-07
test(core): fixes flaky test with document publish

binoy14:
### Description

Attempting to fix flaky tests with publishing like [this](https://github.com/sanity-io/sanity/actions/runs/8988057967/job/24688178038?pr=6594)

### What to review

Changes makes sense

### Testing

✅

### Notes for release

N/A

PR #6609, opened 2024-05-08
test(core): await publish action to prevent test flakiness

cngonzalez:
### Description
We [recently](https://github.com/sanity-io/sanity/pull/6595) introduced some improvement to prevent flaky tests on our publish action. Since our delete test also uses the publish button, let's use the same strategy to prevent flakiness.

### What to review
Does it make sense?

### Testing
I suppose the test is the test passing!


binoy14 (approved):
great catch, thanks!

bjoerge (approved):
Thank you!

PR #6610, opened 2024-05-09
fix(core/form): prevent onFocus for root object path

skogsmaskin:
### Description

This will prevent any input calling element.onFocus() on any opened block or inline-object inside the PT-input, as that will close the editing modal for them (through DocumentPaneProvider)


pedrobonamin (approved):

pedrobonamin:
Non blocking as we need this fix, @skogsmaskin could we add tests in a follow up for this case that we found it's broken?

skogsmaskin:
> Non blocking as we need this fix, @skogsmaskin could we add tests in a follow up for this case that we found it's broken?

Yes, we should add some tests for this, absolutely.

PR #6613, opened 2024-05-09
feat(schema): add icon to BlockStyleDefinition

robinpyon:
### Description

This PR adds support for specifying a custom `icon` to block styles (similar to `lists` and `annotations`). 

Whilst not immediately useful to the Studio right now, this is helpful for _Create_ as it employs a block-level slash command palette which displays a list of available _text_ and _list_ styles (and other custom blocks / actions) available in the current schema.

image

Whilst we could hard-code icons for text styles, it felt more idiomatic to make these customisable at the schema level (which is already possible for lists and annotations). 

These could come into play for Studio users if a similar slash command palette is added here – or if we look to change the look of the block style dropdown to be a little more streamlined.

### What to review

Adding an `icon` to a custom block style shouldn't throw validation errors in the studio.

### Testing

A simple unit test was added to check that `validateSchema` doesn't return errors when specifying a custom style `icon`.

### Notes for release

N/A

### DIFF
--- a/packages/@sanity/schema/test/validation/validation.test.ts
+++ b/packages/@sanity/schema/test/validation/validation.test.ts
@@ -1,4 +1,5 @@
 import {describe, expect, test} from '@jest/globals'
+import {SquareIcon} from '@sanity/icons'
 import {flatten} from 'lodash'
 
 import {validateSchema} from '../../src/sanity/validateSchema'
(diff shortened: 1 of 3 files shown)


packages/@sanity/schema/test/validation/validation.test.ts
```diff
@@ -171,4 +172,20 @@ describe('Validation test', () => {
       helpId: 'schema-array-of-type-builtin-type-conflict',
     })
   })
+
+  test('accepts blocks with custom style icons', () => {
+    const schemaDef = [
+      {
+        name: 'testBlock',
+        type: 'block',
+        styles: [{icon: SquareIcon, title: 'Normal text', value: 'normal'}],
```
jtpetty, line 181: Thanks for adding a test! I don't see that we have any tests for passing in a `styles` array without `icon`. Do you think it's worth adding a test for that scenario as well just to make sure that we didn't break anything here by adding `icon`?

robinpyon: Added!

jordanl17 (approved):
Tests helped to understand the intention here. Thanks! 🚀

packages/@sanity/schema/test/validation/validation.test.ts
jtpetty: 🙏

jtpetty (approved):

PR #6615, opened 2024-05-09
fix(cli): graphql api flag is respected during validation

mwain:
### Description
During GraphQL deployment we validate APIs have unique dataset/tag. This check wasn't respecting the `--api` CLI flag.

Ref CLDX-2142

### What to review

* Test the graphql deploy command
* Is this behaviour expected?

### Testing

Tested locally with various GQL API configs, and command line args

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/graphql/deployApiAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/graphql/deployApiAction.ts
@@ -69,7 +69,12 @@ export default async function deployGraphQLApiAction(
   }).config({apiVersion: '2023-08-01'})
 
   const apiDefs = await getGraphQLAPIs(context)
-  const hasMultipleApis = apiDefs.length > 1 || (flags.api && flags.api.length > 1)
+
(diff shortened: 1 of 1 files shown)


binoy14 (commented):
Changes makes sense to me. Although is it possible to add some tests for this? looks like there are already some tests [here](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/cli/test/graphql.test.ts) for graphql cli.

mwain:
> Changes makes sense to me. Although is it possible to add some tests for this? looks like there are already some tests [here](https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/cli/test/graphql.test.ts) for graphql cli.

Im not sure, i would need to update the graphql config to add a new api definition, but then the other test will also deploy 2 apis

j33ty (dismissed):

mwain:
Rebased on origin/next

bobinska-dev:
Hello @mwain! When will this be released? The customer asked :)

rexxars (approved):

PR #6617, opened 2024-05-10
chore(cli): pin next-sanity to v9

juice49:
### Description

This branch updates the version of `next-sanity` installed when using the Sanity CLI to initialise a Next.js project. We're currently pinned to `next-sanity@7`, which is two major versions old.

### What to review

The version change is correct.

### Testing

`next-sanity@9` can run successfully in Next.js projects. This should already be well established, because it's the latest stable release and already used widely.

### DIFF
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -22,7 +22,7 @@
     "@react-three/fiber": "^8.13.6",
     "@sanity/assist": "^3.0.2",
     "@sanity/block-tools": "3.41.1",
-    "@sanity/client": "^6.17.2",
+    "@sanity/client": "^6.18.0",
     "@sanity/color": "^3.0.0",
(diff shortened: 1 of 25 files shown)


stipsan (changes requested):
I'll need to evaluate that what the `init` command does when setting up `next-sanity` is also following the v9 patterns 🙌 
A lot has happened in the last two majors:
- https://github.com/sanity-io/next-sanity/blob/main/packages/next-sanity/MIGRATE-v8-to-v9.md
- https://github.com/sanity-io/next-sanity/blob/main/packages/next-sanity/MIGRATE-v7-to-v8.md

stipsan (changes requested):
Made a PR here @juice49 that updates the template with the breaking changes: https://github.com/sanity-io/sanity/pull/6644

PR #6620, opened 2024-05-10
fix(pte): don't render the PTE block extras container when not in use, disable pointer events on highlights

robinpyon:
### Description

This PR makes some minor changes to text blocks and highlights.

**1. Disables `pointer-events` from the review changes highlight block**

Currently, it's possible to cause some render thrashing when hovering over the change indicator bar in fullscreen mode.

The highlight block doesn't contain any children, so it's safe for us to disable pointer events without having to fight z-index wars.

Before:

https://github.com/sanity-io/sanity/assets/209129/75ef6092-9736-464f-a68b-6fa336dd0183

After

https://github.com/sanity-io/sanity/assets/209129/39494504-a240-4c69-9233-fd5622fb8750

**2. Disables `pointer-events` from the validation highlight pseudo element**

_Really subtle_, but the presence of this would yield unwanted cursor changes when hovering over the periphery of text blocks.

Similar to the above, though this is rendered in a pseudo element so disabling pointer-events shouldn't affect children.

Before:

https://github.com/sanity-io/sanity/assets/209129/117c4452-51fb-489e-86de-d7fd2274ab79

After:

https://github.com/sanity-io/sanity/assets/209129/f9d9038d-0e15-4095-ac6f-58333c2c1cc8

**3. Don't render `BlockExtrasContainer`  in text blocks unless change indicators or (the now deprecated) block actions are present.**

The more significant change here – we now only render the block extras container if we can determine any (legacy) block actions exist, or if we need to display the change bar wrapper (in fullscreen mode)

This one is especially relevant to _Create_ as this container currently occupies space that could otherwise be used for text — like the above, it also shows a noticeable cursor change when hovering. This was especially noticeable in _Create_ given how much interaction happens on the RHS margins of PTE inputs.

Before (element highlighted for emphasis):

https://github.com/sanity-io/sanity/assets/209129/6d152b42-79ad-44a9-bacf-46063d0a5342

After:

https://github.com/sanity-io/sanity/assets/209129/0b93c786-87db-41f5-afb6-15bb9f677d52

### What to review

- Text block extras containers element should be hidden unless needed
- In general, mouse cursors shouldn't change when hovering around the periphery of text blocks in PTE

### Testing

No automated tests here – wasn't sure if it would be a bit overkill to create a playwright component test to validate cursor changes, happy to do so if needed!

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
@@ -263,6 +263,9 @@ export function TextBlock(props: TextBlockProps) {
     [Markers, markers, renderCustomMarkers, tooltipEnabled, validation],
   )
 
+  const blockActionsVisible = renderBlockActions && focused && !readOnly
+  const changeIndicatorVisible = isFullscreen && memberItem
+
   return useMemo(
     () => (
       <Box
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/inputs/PortableText/text/TextBlock.tsx
```diff
@@ -295,40 +298,47 @@ export function TextBlock(props: TextBlockProps) {
...
 
-            <BlockExtrasContainer contentEditable={false}>
-              <BlockActionsOuter marginRight={1}>
-                <BlockActionsInner>
-                  {renderBlockActions && focused && !readOnly && (
-                    <BlockActions
-                      onChange={onChange}
-                      block={value}
-                      renderBlockActions={renderBlockActions}
+            {/*
```
jordanl17, line 301: Nice! 🔥

jordanl17 (approved):

PR #6621, opened 2024-05-10
feat(pte): add `hideToolbar` and `fullscreen` props to `PortableTextInput`

robinpyon:
### Description

This PR introduces additional props to `<PortableTextInput>`:
- `hideToolbar` – whether the input's toolbar is hidden
- `fullscreen` – whether the input is initially full screen on mount

... that's it! Both of these are used by the various PTE inputs used in _Create_.

### What to review

Existing PTE inputs in the studio should be unaffected. Fullscreen toggling should work as normal.

### Testing

AFAIK there aren't any dedicated component tests for `<PortableTextInput>` – we only have some for `<FormBuilder>`. Please correct me if otherwise! Let me know if you need specific integration tests here.

I've also created an example document type in the test studio (`Standard Inputs > Portable Text > BlockEditor examples`) which demonstrates `hideToolbar` in action (I'd have included an example for `fullscreen`, but opening a document will open you into a fullscreened-PTE input, which wasn't super intuitive!)

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -80,6 +80,7 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
   const {
     editorRef: editorRefProp,
     elementProps,
+    fullscreen: fullscreenProp,
     hotkeys,
     markers = EMPTY_ARRAY,
     onChange,
@@ -118,7 +119,7 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
   const {t} = useTranslation()
   const [ignoreValidationError, setIgnoreValidationError] = useState(false)
   const [invalidValue, setInvalidValue] = useState<InvalidValue | null>(null)
-  const [isFullscreen, setIsFullscreen] = useState(false)
+  const [isFullscreen, setIsFullscreen] = useState(fullscreenProp ?? false)
   const [isActive, setIsActive] = useState(false)
   const [isOffline, setIsOffline] = useState(false)
   const [hasFocusWithin, setHasFocusWithin] = useState(false)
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -118,7 +119,7 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
   const {t} = useTranslation()
   const [ignoreValidationError, setIgnoreValidationError] = useState(false)
   const [invalidValue, setInvalidValue] = useState<InvalidValue | null>(null)
-  const [isFullscreen, setIsFullscreen] = useState(false)
+  const [isFullscreen, setIsFullscreen] = useState(fullscreenProp ?? false)
```
ricokahler, line 122: looking at this usage it seems like the `fullscreen` prop sets just the default value of this vs managing its state. In react, i typically see props that behave like this prefixed with `default*` e.g.e `defaultValue`. idk if `defaultFullscreen` would be a better or worse name but something to consider. not a blocking comment by any means.

ricokahler (approved):
had a non-blocking comment but looks good!

jordanl17 (approved):

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
robinpyon: Great suggestion and makes perfect! Had a sniff around the studio and found other similar instances which use `initial*` so went with that approach

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -81,6 +81,7 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
     editorRef: editorRefProp,
     elementProps,
     hotkeys,
+    initialFullscreen: initialFullscreenProp,
```
ricokahler, line 84: i don't think u have to alias this one now 😄

ricokahler (dismissed):

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
robinpyon: Third time's a charm?

ricokahler (approved):

PR #6622, opened 2024-05-10
fix(pte): cursor presence causes pte error

pedrobonamin:
### Description
 Structure consistently crashes with` Cannot resolve a DOM point from Slate `point when editing a document with multiple users
I've been debugging this and it seems to be generated because the `PresenceCursor` element is not rendering the `Leaf` that is generated in the [editable component](https://github.com/sanity-io/sanity/blob/5bc6f0390e914ed5e38729cff11a9919db13796d/packages/%40sanity/portable-text-editor/src/editor/Editable.tsx#L199-L200), which causes that the editable leaf "dissapears" and crashes the editor.

This were the original steps to reproduce the issue:
Steps to reproduce

- **User 1**: Create a new document with a PTE field
- **User 2**: visit the same URL
- **User 1**: focus the PTE field
- **User 2**: focus the PTE field
- **User 1**: enter a line break
- User 2's screen will indicate that the current caret has shifted down, even though they haven't done anything (This hasn't been fixed, not sure how to think it @skogsmaskin any thoughts? )
- **User 2**: focus the PTE field
- User 1's structure will then crash
- **User 1**: click retry
- **User 1:** re-focus the PTE field, navigate to the second line and hit ENTER to line break
- **User 1:** observe how the caret hasn't changed position
- **User 1**: hit ENTER again
- **User 2**: refocus the PTE field and press the up arrow key

With the changes introduced I was not able to reproduce the issue anymore,

User 1's structure will then crash

### DIFF
--- a/packages/@sanity/portable-text-editor/e2e-tests/__tests__/selectionAdjustment.collaborative.test.ts
+++ b/packages/@sanity/portable-text-editor/e2e-tests/__tests__/selectionAdjustment.collaborative.test.ts
@@ -298,13 +298,13 @@ describe('selection adjustment', () => {
           ],
         },
         {
-          _key: 'B-7',
+          _key: 'B-6',
           _type: 'block',
           markDefs: [],
           style: 'normal',
           children: [
             {
-              _key: 'B-6',
+              _key: 'B-5',
               _type: 'span',
               text: '',
               marks: [],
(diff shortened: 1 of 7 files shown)


skogsmaskin (approved):
Great catch @pedrobonamin, thank you for this!

I've updated the test to reflect the new DOM structure.

skogsmaskin (approved):
Great work @pedrobonamin!

PR #6624, opened 2024-05-10
fix(core): avoid attribute names being scanned as symbols. Fixes #5673

### LINKED ISSUE #5673: Studio crashes when name of the array starts with 'true' or 'false'
## Description

Reported in Github issue [#5673](https://github.com/sanity-io/sanity/issues/5673):

> When in a document schema you define an array with a name that starts with 'true' or 'false' the studio crashes.
> With the error:
> *Error: Expected attribute name following '.*

## Steps to reproduce

*What did the customer/requester do to encounter the issue?*

Inside a document define an array like so:

```
        title: 'Answers',
        name: 'trueorfalseField',
        type: 'array',
        of: [
            {type: 'string', name: 'title', title: 'title'},
        ],
    },
```

1. Go to the "answers" array inside the document
2. Click on "add item"
3. See error

(Carolina note: my error was `failed to parse JSON path` while customer's was `Expected attribute name following '` , though I simply copy/pasted the schema field definition)

## Expected behavior

*What should have happened, or what did they expect?*

The ability to add an item to that array

## Is there a workaround?

*If yes, describe what is the workaround. If no, describe how it completely blocks the use of Sanity.*

No

## Any additional context

*Please attach screenshots of the console, a screen recording of the studio itself, or anything else you think would be helpful.*

Customer's error:

runeb:
In the mutator package, we scan for certain symbols, among them "true" and "false". If an attribute name is prefixed with a symbol value, the string scanning will eagerly interpret it as a symbol instead of an attribute. This patch checks that there is no more valid attribute characters immediately after the potential symbol, so that the string "trueAttribute" will not be interpreted as the symbol "true".


packages/@sanity/mutator/test/parse.test.ts
```diff
@@ -183,6 +183,10 @@ const cases = {
       },
     ],
   },
+  trueOrFalseField: {
```
jtpetty, line 186: should we add a test that fails if this name is just `true` or `false` to make sure we didn't just break something?

packages/@sanity/mutator/src/jsonpath/tokenize.ts
```diff
@@ -113,6 +113,19 @@ class Tokenizer {
       return null
     }
     if (str === this.source.slice(this.i, this.i + str.length)) {
+      // When checking symbols that consist of valid attribute characters, we
+      // need to make sure we don't inadvertently treat an attribute as a
+      // symbol. For example, an attribute 'trueCustomerField' should not be
+      // scanned as the boolean symbol "true".
+      if (str[0].match(attributeCharMatcher)) {
```
jtpetty, line 120: do we know what the perf hit is on looking ahead on each match?

packages/@sanity/mutator/test/parse.test.ts
runeb: Added explicit tests for parsing symbols versus attributes/identifiers

packages/@sanity/mutator/src/jsonpath/tokenize.ts
runeb: It would only to this extra check when it encounters a symbol match which is all attribute-valid chars, meaning `"true"` and `"false"`

bjoerge (dismissed):
Nice, thanks also for the added tests :)

runeb:
> Nice, thanks also for the added tests :)

merged in `next` and fixed a linting issue so the tests pass

bjoerge (approved):

PR #6634, opened 2024-05-13
fix(sanity): make config flag override server document actions

bjoerge:
### Description
While reviewing #6418 I realized that we probably want a way to enable server actions (e.g. to test intenrally) even in the case it's turned off for most users.

I think it also makes sense to make it very clear that it's internal, and not meant for studio developers.

### What to review
Does it make sense?

### Testing
- Might need to update some tests - opening as draft PR for now to gauge.

### Notes for release

n/a – not public facing

### DIFF
--- a/packages/sanity/src/core/store/_legacy/datastores.ts
+++ b/packages/sanity/src/core/store/_legacy/datastores.ts
@@ -136,7 +136,7 @@ export function useDocumentStore(): DocumentStore {
   const serverActionsEnabled = useMemo(() => {
     const configFlag = workspace.__internal_serverDocumentActions?.enabled
     // If it's explicitly set, let it override the feature toggle
-    return typeof workspace.__internal_serverDocumentActions?.enabled === 'boolean'
+    return typeof configFlag === 'boolean'
       ? of(configFlag as boolean)
       : fetchFeatureToggle(getClient(DEFAULT_STUDIO_CLIENT_OPTIONS))
   }, [getClient, workspace.__internal_serverDocumentActions?.enabled])
--- a/packages/sanity/src/core/store/_legacy/datastores.ts
+++ b/packages/sanity/src/core/store/_legacy/datastores.ts
@@ -134,11 +134,12 @@ export function useDocumentStore(): DocumentStore {
   const workspace = useWorkspace()
 
   const serverActionsEnabled = useMemo(() => {
-    // If it's explicitly disabled, we'll just return a stream that emits `false`
-    return workspace.serverActions?.enabled === false
-      ? of(false)
+    const configFlag = workspace.__internal_serverDocumentActions?.enabled
+    // If it's explicitly set, let it override the feature toggle
+    return typeof workspace.__internal_serverDocumentActions?.enabled === 'boolean'
+      ? of(configFlag as boolean)
       : fetchFeatureToggle(getClient(DEFAULT_STUDIO_CLIENT_OPTIONS))
-  }, [getClient, workspace.serverActions?.enabled])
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/store/_legacy/grants/documentPairPermissions.ts
```diff
@@ -300,8 +300,10 @@ export function useDocumentPairPermissions({
 
   const serverActionsEnabled = useMemo(() => {
     // If it's explicitly disabled, we'll just return a stream that emits `false`
-    return workspace.serverActions?.enabled === false ? of(false) : fetchFeatureToggle(client)
-  }, [client, workspace.serverActions?.enabled])
+    return workspace.__internal_serverDocumentActions?.enabled === false
```
juice49, line 303: Does this not need to updated like this?

```ts
return typeof workspace.__internal_serverDocumentActions?.enabled === 'boolean' ? of(workspace.__internal_serverDocumentActions?.enabled as boolean)
```

(Currently I think this would only respect a negative override).

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
```diff
@@ -161,8 +161,10 @@ export function useTimelineStore({
 
   const serverActionsEnabled = useMemo(() => {
     // If it's explicitly disabled, we'll just return a stream that emits `false`
-    return workspace.serverActions?.enabled === false ? of(false) : fetchFeatureToggle(client)
-  }, [client, workspace.serverActions?.enabled])
+    return workspace.__internal_serverDocumentActions?.enabled === false
```
juice49, line 164: Does this not need to updated like this?

```ts
return typeof workspace.__internal_serverDocumentActions?.enabled === 'boolean' ? of(workspace.__internal_serverDocumentActions?.enabled as boolean)
```

(Currently I think this would only respect a negative override).

packages/sanity/src/core/store/_legacy/datastores.ts
```diff
@@ -134,11 +134,12 @@ export function useDocumentStore(): DocumentStore {
   const workspace = useWorkspace()
 
   const serverActionsEnabled = useMemo(() => {
-    // If it's explicitly disabled, we'll just return a stream that emits `false`
-    return workspace.serverActions?.enabled === false
-      ? of(false)
+    const configFlag = workspace.__internal_serverDocumentActions?.enabled
+    // If it's explicitly set, let it override the feature toggle
+    return typeof workspace.__internal_serverDocumentActions?.enabled === 'boolean'
```
juice49, line 139: Nit: This should probably use `configFlag` which is bound to the value and used below.

```suggestion
    return configFlag === 'boolean'
```

packages/sanity/src/core/store/_legacy/grants/documentPairPermissions.ts
bjoerge: You're right - thanks for catching!

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
bjoerge: You're right again - thanks for catching!

packages/sanity/src/core/store/_legacy/datastores.ts
bjoerge: Thanks! Fixed

bjoerge:
PR feedback addressed in 395fb8601ff03e841a586c0c1ef945e9243ea7c4

juice49 (approved):

PR #6642, opened 2024-05-13
fix(pte): improve text selection in fullscreen inputs

robinpyon:
### Description

This PR sets `max-width` on PTE blocks, rather than the editable container.

This enables block margins to be clickable and improve ergonomics of focusing and selecting text, especially in fullscreen mode.

This is useful for _Create_ as there effectively all PTE inputs are in fullscreen by default and this was especially noticeable, as often these margins occupy much of the screen!

This change especially improves text selection in this context, as you can now start selections directly in the margins – fairly standard behaviour across various editors.

Before:

https://github.com/sanity-io/sanity/assets/209129/5ccac024-add5-446d-9ce6-032126113ea8

After:

https://github.com/sanity-io/sanity/assets/209129/e14d0e73-3235-482d-8ab9-5478e7e42ab3

**Other minor changes:**
1. Remove padding in fullscreen mode (not terribly noticeable, but this also ensures that we don't have a slight 'dead zone' around text where the cursor switches from `text` to `default`)
2. Slightly tweaks spacing of change highlight bars to look a little better on constrained viewports:

  Before:
  image

  After:
  image

### What to review

- General text selection and focusing of collapsed / default PTE inputs in the Studio should be unaffected
- When in 'fullscreen' mode, it should be possible to focus and select text directly in the margins

### Testing

No new tests. This would be a fantastic candidate to test in playwright-ct (clicking block margins and double checking PTE selections), though unfortunately don't have the bandwidth to add right now 😓 

### Notes for release

Improves text selection of full screen Portable Text inputs

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable camelcase */
 
-import {Card, Container, rem} from '@sanity/ui'
+import {Card, rem} from '@sanity/ui'
 import {getTheme_v2} from '@sanity/ui/theme'
 import {css, styled} from 'styled-components'
 
@@ -68,11 +68,6 @@ export const Scroller = styled(ScrollContainer)`
   }
 `
 
-export const EditableContainer = styled(Container)`
-  /* @todo: calculate from theme */
-  /* max-width: 728px; */
-`
-
 export const EditableWrapper = styled(Card)<{$isFullscreen: boolean; $readOnly?: boolean}>`
   height: 100%;
   width: 100%;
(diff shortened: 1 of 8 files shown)


skogsmaskin (approved):
This is a great UX improvement! Thanks.

I was thinking if this might disrupt any of our current stuff, like the legacy block markers and stuff, but I think not?

robinpyon:
@skogsmaskin Great shout – I just had another quick look in the test studio and they could use a bit of alignment tweaks, which I'll add shortly

robinpyon:
@skogsmaskin 🙇 I've just pushed a few minor changes which tidy this up and standardize how both object and text blocks render legacy actions. This could certainly be made DRY in future, as there's a lot of overlap between text and object components.

**What to review:**
- Space is allocated for legacy block actions ([example](https://test-studio-g6pkvth7t.sanity.build/test/structure/input-standard;portable-text;pt_customMarkersTest;641e1d86-4c9e-4cf7-b08f-0afd39b6133c))
- Change review buttons / overlays appear and work as intended on non-block action fields (in fullscreen mode) ([example](https://test-studio-git-crx-558.sanity.build/test/structure/input-standard;portable-text;simpleBlock;2f1e8b02-6d2e-4fb4-88f9-d22a3345aa45))
- Object fields should correctly render full-width on non-block action fields
- As above in the original comment, selection + focus should work from block margins

skogsmaskin (approved):
Thanks for giving the extra dilligence @robinpyon !

PR #6644, opened 2024-05-13
feat: upgrade `sanity init` for Next.js to next-sanity v9

stipsan:
### Description
This is an alternative take to #6617.

Updates the `sanity init` cli when used in Next.js apps that embed Studios to the best practices in `next-sanity` v9:
- https://github.com/sanity-io/next-sanity/blob/main/packages/next-sanity/MIGRATE-v7-to-v8.md
- https://github.com/sanity-io/next-sanity/blob/main/packages/next-sanity/MIGRATE-v8-to-v9.md

The biggest difference is that `next-sanity` v9 only supports App Router, so the CLI no longer asks if you want to use App Router or not.
That doesn't mean the whole app has to be in App Router, it's only relevant to the route that embeds the Studio. In fact, the entire rest of the app can be in Pages Router.
Secondly, the latest `NextStudio` component supports fast SSR hydration, and it works best if the `sanity.config.ts` has the `'use client'` directive at the top, instead of the `[[...tool]]/page.tsx` file.

### What to review

`sanity init` inside of next v14 apps should work fine, wether or not the app is using Pages Router or App Router. If it can load the studio on `http://localhost:3000/studio` then it works.

### Testing

Existing tests should cover the changes, as it's a refactor of the template itself, and removal of a prompt that's no longer needed (the prompt that asks if you want to use App Router or not).

### Notes for release

The `sanity init` command for Next.js embedded Studios are updated to `next-sanity` v9 and `next` v14 best practices.

Closes #6617


packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
```diff
@@ -39,27 +41,7 @@ const dataset = process.env.NEXT_PUBLIC_SANITY_DATASET
 export default defineCliConfig({ api: { projectId, dataset } })
 `
 
-export const sanityStudioPagesTemplate = `import Head from 'next/head'
```
jtpetty, line 42: you are changing the name of one exported const and deleting two others. I'm assuming this is intended but just wanted to call this out to make sure.

stipsan: Yep, that's correct.
The name `sanityStudioPagesTemplate` allude to the format used to mount the studio in `pages/studio/[[...tool]].tsx` paths, which is handled by the Next.js Pages Router: https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts

`next-sanity` v9 only supports App Router routes `app/studio/[[...tool]]/page.tsx` mount points: https://nextjs.org/docs/app/building-your-application/routing/defining-routes

Due to the two routes starting to diverge too much in how they handle meta data tags, and other APIs, we decided to stop supporting Pages Router and started to simplify the API surface and docs. 
As a result `sanityStudioPagesTemplate` is no longer needed and produces invalid code.

For example this export no longer exists:
```
import { metadata } from 'next-sanity/studio/metadata'
```

And since there are no longer a need to maintain two templates it no longer makes sense to specify `AppTemplate` in the name :)

bjoerge (approved):

PR #6647, opened 2024-05-14
feat(structure): add sheet list table view.

pedrobonamin:
### Description
This PR integrates the initial changes needed to render the table view inside structure.
The table is generated using  [tanstack/table](https://tanstack.com/table) 
The table data is fetched in two pieces:
- First part is done by reusing the `search` mechanism from the studio global search, this provides a list of document ids to show in the table.
- Second part is handled by the `useDocumentSheetListStore` this hook takes care of mounting a listener for the `_type` that is displayed in the table, keeping the documents in view and not in view updated.

Combining this two fetch mechanisms allows us to reuse the `search` functionality which is powerful and users already know and get the benefits of up to date data with the store hook.

#### Other changes
Refactors search to export some necessary components that are used in structure.

### Notes for release
Nothing to mention. This is not gonna be accessible for users yet


packages/sanity/src/structure/panes/documentList/useDocumentSheetListStore.ts
```diff
@@ -0,0 +1,247 @@
+import {type ListenEvent, type ListenOptions} from '@sanity/client'
```
pedrobonamin, line 1: High chances this will be deprecated by the stream of concurrent work in which we will have 1 unique listener for all documents and we could use a new hook that mimics `useEditState` but will take a list of ids.

```ts

function useEditStateList(publishedDocIds: string[], docTypeName: string): EditStateFor[] {
  const documentStore = useDocumentStore()
  return useMemoObservable(() => {
    return combineLatest(
      publishedDocIds.map((publishedDocId) =>
        documentStore.pair.editState(publishedDocId, docTypeName),
      ),
    )
  }, [documentStore.pair, publishedDocIds, docTypeName]) as EditStateFor[]
}
```

jordanl17 (dismissed):

packages/sanity/src/structure/panes/documentList/DocumentSheetListPane.tsx
```diff
@@ -1,10 +1,176 @@
...
+  useReactTable,
+} from '@tanstack/react-table'
+import {useCallback, useEffect} from 'react'
+import {SearchProvider, useSchema, useSearchState} from 'sanity'
+import {styled} from 'styled-components'
 
-export function DocumentSheetListPane() {
+import {type BaseStructureToolPaneProps} from '../types'
+import {DocumentSheetFilter} from './DocumentSheetListFilter'
+import {DocumentSheetPaginator} from './DocumentSheetListPaginator'
```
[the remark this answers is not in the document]
pedrobonamin: My bad, I renamed the files but not the components exports, I think we should go for `DocumentSheetListXX`, once your last Pr lands we could re arrange everything into its own folder, I imagine this feature will continue growing into more components and files.

jordanl17 (approved):

packages/sanity/src/structure/panes/documentList/DocumentSheetListPaginator.tsx
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable i18next/no-literal-string */
```
pedrobonamin, line 1: We don't have designs yet and we have high chances of removing pagination entirely, that why I avoid adding it to the translation resources

jtpetty (approved):
LGTM - am I correct in assuming this will not show up in the document list action menu yet?

jordanl17 (approved):

stipsan (approved):
All good, doesn't break presentation 👍

PR #6649, opened 2024-05-14
fix(typegen): move type new line separator into formatter

sgulseth:
### Description

Prettier were adding a newline, and then we were adding an additional new line to each type. This moves new line into the maybeFormatCode to either rely on prettier adding it, or adding it manually if prettier is not found.

### Testing

We have test that checks a snapshot, these are updated

### Notes for release

fixes a bug where typegen would add an additional new line between types.

### DIFF
--- a/packages/@sanity/cli/src/workers/typegenGenerate.ts
+++ b/packages/@sanity/cli/src/workers/typegenGenerate.ts
@@ -63,7 +63,7 @@ registerBabel()
 
 function maybeFormatCode(code: string, prettierConfig: PrettierOptions | null): Promise<string> {
   if (!prettierConfig) {
-    return Promise.resolve(code)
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/workers/typegenGenerate.ts
```diff
@@ -63,7 +63,7 @@ registerBabel()
 
 function maybeFormatCode(code: string, prettierConfig: PrettierOptions | null): Promise<string> {
   if (!prettierConfig) {
-    return Promise.resolve(code)
+    return Promise.resolve(`${code}\n`) // add an extra new newline, poor mans formatting
```
stipsan, line 66: 😂

stipsan (dismissed):

stipsan (approved):

PR #6665, opened 2024-05-14
fix: optimize esm builds with lodash-es

stipsan:
Unclear if the benefits warrant the complexity, PR will stay in draft status until testing is finished. Can be tested using `npm i sanity@use-lodash-es --save-exact`


stipsan:
Closing it as tests show that unless all plugins also run the same optimization with `lodash-es`, we end up with a larger total studio bundlesize than if `sanity` stays on `lodash` 🤔

PR #6682, opened 2024-05-15
feat(core): array editing improvements 

RitaDias:
### Description

This pull request introduces a new experience for editing arrays of objects. The form now allows for navigation through a single dialog instead of multiple dialogs layered on top of each other.

Screenshot 2024-06-13 at 13 36 34

### What to review

- General review of the code
- Make sure the feature is behaving as expected. 
  - The tree structure navigation in the sidebar
  - The search
  - The breadcrumbs navigation
- Everything should work as it currently does, except that the new dialog is being used.
- ⚠️ Since this is a shadow release, the most crucial aspect to review is that the feature is only visible in the Studio if it is enabled in the config (see below). If it is not present or is disabled in the configuration, the new way of editing arrays should not be visible.

```ts
export default defineConfig({
  // ...

  features: {
    beta: {
      treeArrayEditing: {
        enabled: true
      }
    }
  }
})
```

- It should be possible to opt out of tree editing for an array by setting `options.treeEditing` to `false` in the schema. Note that if the array field has any nested array fields, these will also opt out of tree editing.

```ts
import {type defineField} from 'sanity'

export const array = defineField({
  type: 'array',
  name: 'myArray',
  options: {
    treeEditing: false
  },
  of: [
    // ...
  ]
})
```

### Testing
- Unit, Component and e2e tests were written to target the most important parts of this PR
- A couple of our clients had early access to this via a canary to gather feedback

### Notes
- Arrays within the The Portable Text Editor will continue to use the existing solution.

### Notes for release

N/A – this is a shadow release


packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
```diff
@@ -469,27 +472,29 @@ export function Compositor(props: Omit<InputProps, 'schemaType' | 'arrayFunction
...
-        <ChangeIndicator
-          disabled={isFullscreen}
-          hasFocus={Boolean(focused)}
-          isChanged={changed}
-          path={path}
-        >
-          <Root
-            data-focused={editorFocused ? '' : undefined}
-            data-read-only={readOnly ? '' : undefined}
+    <PortableTextAwareContext.Provider value={PORTABLE_TEXT_AWARE_CONTEXT_VALUE}>
```
hermanwikner, line 475: This provider should forward any potential parent context value

RitaDias:
Updated the `studio.test.tsx` after having it consistently failing on our PR.
Spoke with Cody and it seems related to the style components and the order with which the components that use them are called.

The snapshot test was brittle since it was screaming that something was wrong when it reality it was just different (and we wanted it to be different). The following test was written with Cody to remain useful and to target hydration test but without having to rely on snapshots

packages/sanity/src/core/form/studio/FormBuilder.test.tsx
```diff
@@ -107,8 +109,9 @@ describe('FormBuilder', () => {
           schemaType: formState?.schemaType || schemaType,
           validation: EMPTY_ARRAY,
           value: formState?.value as FormDocumentValue,
+          useNewTreeDialog,
```
hermanwikner, line 112: Why are we passing this as a prop to the form builder?

RitaDias: This needs to be updated after we made changes in formBuilder. We need it, afaik, because otherwise the formBuilder is rendering the wrong thing and the test is failing (since we added the conditional rendering in formBuilder.tsx)

RitaDias: Updated this test (with the hook returning enabled `false`), I'm going to add another test for when it's true as well

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
```diff
@@ -22,9 +24,21 @@ interface DocumentStatusBarActionsInnerProps {
...
 function DocumentStatusBarActionsInner(props: DocumentStatusBarActionsInnerProps) {
   const {disabled, showMenu, states} = props
-  const {__internal_tasks} = useDocumentPane()
+  const {__internal_tasks, schemaType, openPath} = useDocumentPane()
   const [firstActionState, ...menuActionStates] = states
   const [buttonElement, setButtonElement] = useState<HTMLButtonElement | null>(null)
+  const {enabled: isTreeArrayEditingEnabled} = useTreeArrayEditingEnabled()
+
+  const isDialogOpen = useMemo(
+    () => shouldArrayDialogOpen(schemaType, openPath),
```
pedrobonamin, line 33: For the use of this function this feels like the function could be renamed to `isArrayDialogSupported` or , `shouldUseArrayDialog` instead of `shouldArrayDialogOpen`
Also, given treeEditing could be disabled, I suggest we first check for that and later for the function.
```suggestion
  const isTreeArrayEditingEnabledOpen = useMemo(
    () => isTreeArrayEditingEnabled && shouldArrayDialogOpen(schemaType, openPath),
    [openPath, schemaType, isTreeArrayEditingEnabled],
  )

```

pedrobonamin (commented):
Looking good, this will be a huge improvement, added some suggestions.
Thanks!!

packages/sanity/src/core/form/studio/tree-editing/__workshop__/TreeEditingMenuStory.tsx
```diff
@@ -0,0 +1,102 @@
...
+  //     {
+  //       title: 'Level 1.2',
+  //       path: ['level-1', 'level-1.2'],
+  //     },
+  //     {
+  //       title: 'Level 1.3',
+  //       path: ['level-1', 'level-1.3'],
+  //     },
+  //   ],
+  // },
```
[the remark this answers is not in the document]
hermanwikner: Not really, it's kept just to have the first level available when you want to test without the `buildStructure` function and control the items manually!

packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/TreeEditingBreadcrumbs.tsx
```diff
@@ -0,0 +1,192 @@
...
+  selectedPath: Path
+}
+
+export function TreeEditingBreadcrumbs(props: TreeEditingBreadcrumbsProps): JSX.Element | null {
+  const {items: itemsProp = EMPTY_ARRAY, onPathSelect, selectedPath} = props
+
+  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null)
+  const size = useElementSize(rootElement)
+
+  // todo: evaluate if these values are working as expected
```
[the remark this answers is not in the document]
RitaDias: I don't think so, maybe we can remove this comment @hermanwikner ?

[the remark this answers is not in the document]
RitaDias: I'm putting a pin in this since we will likely make some changes on this part / logic.
Thanks for the question Pedro, and sorry about this 🙏

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
RitaDias: Fair on the suggestion!
I think the naming should remain the same though, the shouldArrayDialogOpen doesn't only check if it's supported, it literally checks if the dialog should be allowed open based on where it is (path and the schematype). The dialog can be "supported" and yet not be open in a specific scenario :)

packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './TreeEditingBreadcrumbs'
```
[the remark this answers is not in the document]
RitaDias: I might be misinterpreting what you mean here, but this is not a barrel file is it?
This is importing a specific file (component) in the tree called TreeEditingBreadcrumbs not reexporting multiple things? 🤔

packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/TreeEditingBreadcrumbs.tsx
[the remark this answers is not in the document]
RitaDias: Updated this :) 
Made the comment a bit clearer. We don't think there are any breakpoints missing but we might need to adjust them as we get more feedback :)

packages/sanity/src/core/form/studio/tree-editing/components/resizer/Resizer.tsx
```diff
@@ -0,0 +1,82 @@
+import {type MouseEvent, useCallback, useRef} from 'react'
```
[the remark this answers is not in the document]
RitaDias: Talking with Herman he mentioned that we can import things from `core` to `structure` but we can't have components created in `structure` be used in `core`. We could make the components that use `packages/sanity/src/structure/panes/document/documentInspector/Resizer.tsx` use this instead once `position` is added. 

I think we should still add the position for future instances of people that might want to use resizer and import it from `core`, however, I think that replacing might be a separate PR. What do you think?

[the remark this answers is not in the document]
RitaDias: I've made the change above (adding props :) )

packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingMenu.tsx
```diff
@@ -0,0 +1,262 @@
...
+    <Stack
+      aria-expanded={open}
+      as="li"
+      key={toString(item.path)}
+      ref={setRootElement}
+      role="treeitem"
+      space={1}
+    >
+      <Card data-as="button" radius={2} tone="inherit">
+        <ItemFlex align="center" data-selected={selected} data-testid="side-menu-item">
```
hermanwikner, line 172: The test ID should be renamed to `tree-editing-sidebar-item` to align with the naming of other test IDs, i.e., `tree-editing-<target>`.

packages/sanity/src/core/form/studio/tree-editing/components/search/TreeEditingSearch.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+import {useSearchableList} from '../../hooks'
+import {type TreeEditingMenuItem} from '../../types'
+import {ITEM_HEIGHT, MAX_DISPLAYED_ITEMS} from './constants'
+import {TreeEditingSearchMenu} from './TreeEditingSearchMenu'
+import {treeEditingSearch} from './utils'
+
+const POPOVER_FALLBACK_PLACEMENTS: PopoverProps['fallbackPlacements'] = ['bottom-start']
+
+const RootFlex = styled(Flex)``
```
hermanwikner, line 15: This can be removed and replaced with a regular `Flex`

packages/sanity/src/core/form/studio/tree-editing/components/tree-menu/TreeEditingMenu.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+export const TreeEditingMenu = memo(function TreeEditingMenu(
+  props: TreeEditingMenuProps,
+): JSX.Element {
+  const {items, onPathSelect, selectedPath} = props
+
+  return (
+    <Stack as="ul" role="tree" space={2}>
+      {items.map((item) => {
+        const siblingHasChildren = items.some(
```
hermanwikner, line 247: Extract to a util function since this check occurs twice?

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
```diff
@@ -0,0 +1,34 @@
+import {type Path} from 'sanity'
+
+/**
+ * Check if the item is selected
+ */
+export function isSelected(itemPath: Path, path: Path): boolean {
```
binoy14, line 6: ❓ should this name be more specific, `isItemSelected` or something like that?

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+  return JSON.stringify(itemPath) === JSON.stringify(path)
+}
+
+/**
+ * Check if the path is an array item path
+ */
+export function isArrayItemPath(path: Path): boolean {
+  return path[path.length - 1].hasOwnProperty('_key')
```
binoy14, line 14: Could you add a comment on why you need to do `-1`?

Also what would happen if the path.length is `0` this will cause a typeerror? Should a check for that be added?

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Check if the item should be in the breadcrumb
+ */
+export function shouldBeInBreadcrumb(itemPath: Path, path: Path): boolean {
+  return (
+    itemPath.every((segment, index) => {
+      return JSON.stringify(path[index]) === JSON.stringify(segment)
```
binoy14, line 23: thinking out loud, wouldn't this work in theory. Although the types might be different?

```suggestion
      return isSelected(path[index], segment)
```

packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
```diff
@@ -0,0 +1,254 @@
...
+    // Store the openPath and value to be able to compare them
+    // with the next openPath and value.
+    valueRef.current = value
+    openPathRef.current = openPath
+
+    debouncedBuildTreeEditingState({
+      schemaType,
+      documentValue: value,
+      openPath,
+    })
```
binoy14, line 194: ❓ should there also be a debounce cancel like in the close handler? 

```tsx
return () => {
 // Cancel any debounced state building when navigating.
      debouncedBuildTreeEditingState.cancel()
}
```

binoy14 (changes requested):
Overall this looks good! I left some comments and messaged you some issues I found

Sidenote: Is there a way to enable the sidebar to be always open, I find it little annoying to click that button every-time the modal is open

dev/test-studio/sanity.config.ts
```diff
@@ -73,6 +73,14 @@ const sharedSettings = definePlugin({
     bundles: testStudioLocaleBundles,
   },
 
+  features: {
+    beta: {
```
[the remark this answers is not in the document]
RitaDias: Hi Binoy!
I've sent you a link on slack with the discussion with Rune about this :)

packages/sanity/src/_singletons/core/form/tree-editing/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './TreeEditingEnabledContext'
```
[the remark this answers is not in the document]
RitaDias: Do we have a recent example that I can have a look at with how we're planning on doing this moving forward? :)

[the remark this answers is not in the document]
RitaDias: Or maybe you mean from the "root" index file, point towards the files?

packages/sanity/src/_singletons/index.ts
```diff
@@ -22,6 +22,7 @@ export * from './core/form/studio/PresenceContext'
 export * from './core/form/studio/ReferenceInputOptionsContext'
 export * from './core/form/studio/reviewChanges/ReviewChangesContext'
 export * from './core/form/studio/ValidationContext'
+export * from './core/form/tree-editing'
```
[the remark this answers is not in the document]
RitaDias: Forgive me for the possible dumb question, but I'm not sure what you mean by typos in this instance.
In this case there is an index file on the tree-editing folder. I understand that's not how we want to do things forward, but I might be missing something unrelated to the barrel exports

packages/sanity/src/_singletons/core/form/tree-editing/types.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @internal
+ */
+export interface TreeEditingEnabledContextValue {
```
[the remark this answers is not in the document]
RitaDias: I'm just confirming that I understood what you mean: do you mean by this to mimic for example the structure like we do for [FormBuilderContext](https://github.com/sanity-io/sanity/blob/f62078c972443364f2dbf67438cb6a80bedc313e/packages/sanity/src/_singletons/core/form/FormBuilderContext.ts) or did you mean something different?

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
RitaDias: Technically I don't think it's possible to it ever be 0 since we're always a level down and so never at "document level" that being said I think it's wise to return the method before just in case if this method is by accident called at doc level

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
RitaDias: This does sound about right!
I'll try it out, see if any of the tests screams or if something is iffy 👍

packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
RitaDias: Yeah this definitely sounds like a good idea

RitaDias: Though I see tht we do have a debounce already on the onClose :) 
https://github.com/sanity-io/sanity/blob/edx-1295/packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx#L115

packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/utils.ts
RitaDias: Anyway, I've added a guardrail for it :)

RitaDias:
> Overall this looks good! I left some comments and messaged you some issues I found
> 
> Sidenote: Is there a way to enable the sidebar to be always open, I find it little annoying to click that button every-time the modal is open

Thank you ❤️ 
Replying to the sidenote: not currently :) if we receive the feedback (which I have a feeling we will), we can look back and add something like that 🙏

packages/sanity/src/_singletons/core/form/tree-editing/types.ts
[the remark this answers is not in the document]
RitaDias: The thing is that we are using this type in a hook called `useTreeEditingEnabled`, if we don't export the type then we wouldn't be able to have it there right? 
https://github.com/sanity-io/sanity/blob/edx-1295/packages/sanity/src/core/form/studio/tree-editing/context/enabled/useTreeEditingEnabled.ts#L9-L10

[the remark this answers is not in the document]
RitaDias: Ok, sorry for the massive chatter, I _think_ I refactored what was spoke about on this comment, https://github.com/sanity-io/sanity/pull/6682#discussion_r1648047179 and https://github.com/sanity-io/sanity/pull/6682#discussion_r1648044213 with the commit https://github.com/sanity-io/sanity/pull/6682/commits/20e2f19bf891fb50277d50861f82972bb29238b7

please let me know if I misunderstood / am missing something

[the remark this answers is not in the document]
binoy14: yep sorry I wasn't clear I meant to place the interface next to the hook in either `core` / `structure` package depending on where the hook is. Your changes are correct thanks!

packages/sanity/src/_singletons/index.ts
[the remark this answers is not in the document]
binoy14: oh sorry I meant since I typed the suggestion wanted to make sure I did not have any typos in my suggestion I commented. Anywho your changes are correct ignore me :homer-disappear:

binoy14 (dismissed):
Changes looks good, this is great and a massive effort. Major kudos!

hermanwikner (approved):

PR #6684, opened 2024-05-15
fix: regression crashing the document list pane in presentation

stipsan:
### Description

# TL;DR 
v3.42.0 broke the list pane in Presentation in all scenarios, usually showing an error like this: 
image
We should merge and do a hotfix ASAP as there's no way to workaround this crash in userland, we can't even do a release of `@sanity/presentation` that user can opt-in to directly temporary while waiting for a new release of `sanity` with a fix.
After release I'll do follow up PRs with a proper fix, playwright tests that can catch this in the future and such.

# Very long; maybe read

> [Even more context on this slack thread](https://sanity-io.slack.com/archives/C05MCSTGN1X/p1715787376068839)

After hours of bisecting why every template and app we updated to `sanity@3.42.0` have a crashing document list panel I managed to trace the regression to occur when [this commit](https://github.com/sanity-io/sanity/commit/724da99b20e75a3a8efc040bd7498fb2fdb97bab) (related to #6593) got pushed to `next`.
The commit just before it has a working presentation tool list panel in our test studio: https://test-studio-gb3qm4c68.sanity.build/presentation/reproduction-presentation
While the one for #6593 is when it first starts crashing: https://test-studio-dppz4pfp9.sanity.build/presentation/reproduction-presentation

Looking at the diff for #6593 didn't reveal any obvious reason why this should be happening, bisecting further I spawned a new git branch for the [before](https://github.com/sanity-io/sanity/tree/before-pane-break) and [after](https://github.com/sanity-io/sanity/tree/after-pane-break) commits, so I could use `npm diff` to inspect the published code differences: https://gist.github.com/stipsan/b0c6ba5ea4afc16640e20279fb8dc15b
That's when I noticed that the `structure.mjs` file didn't have an equivalent diff to this one: https://github.com/sanity-io/sanity/pull/6593/files#diff-90bd90c70a2b60fb25bddbe5b2663b20a5e081a01d19bbaecec9b10851c7fe02
I was expecting to see something like:
```diff
import * as StructureToolProvider from './es-chunks'
export {
-  DocumentListPane
+  PaneContainer as DocumentListPane
} from './_chunks-es/StructureToolProvider.mjs'
```

I eventually found the entrypoint I had to apply the same backwards compatible change to that's used in this PR.
Since `PaneContainer` isn't exported on `sanity/structure` at the time it's not possible to do a release of `@sanity/presentation` that starts using `PaneContainer` instead of `DocumentListPanel`, and so this PR + hotfix is our best option.
Afterwards it makes sense to look into exporting `PaneContainer` and `@sanity/presentation` can start using it after next weeks release, allowing `sanity` to remove the `DocumentListPane` aliasing introduced in this PR two release cycles for now.

To prevent this from happening again I've updated CODEOWNERS to mark all the files that contain internals `@sanity/presentation` is currently relying on to us, so we're notified on future PRs that have changes that might break presentation.
I'll refactor `@sanity/presentation` to gather all its internal imports into a single file so CODEOWNERS can link to it from a comment, making this list over used internals easier to maintain moving forward.
I'll also look into setting up playwright tests for presentation as it's setup in our test studio.

Finally, I've removed the dual Presentation tool setups on the test studio as it's no longer needed, so nobody has to wonder which of them to use when testing for backwards compatibility.

### What to review

If it builds and tests pass we should ship.

### Testing

Existing tests are sufficient for this hotfix. We'll prioritize proper playwright tests in follow ups this week or early next week so that teams don't have to test presentation manually in order to verify that changes to internals doesn't accidentally break presentation.

### Notes for release

Fixes regression introduced in `v3.42.0` that crashes the document list pane in presentation tool

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -59,7 +59,7 @@
 # -- Scheduled Publishing --
 /packages/sanity/src/core/scheduledPublishing/ @sanity-io/studio-ex
 
-# -- Upsell UI  -- 
+# -- Upsell UI  --
 packages/sanity/src/core/studio/upsell/ @sanity-io/studio-ex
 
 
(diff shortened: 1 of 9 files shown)


ricokahler (approved):
🫡

PR #6685, opened 2024-05-15
chore: setup `dev:next-studio` to use react v19

stipsan:
### Description

The easiest way to test React v19 without making any changes to our packages, peer deps and more, is to update `dev:next-studio` to use the latest `canary` of `next`:
image

### What to review

This only affects `dev:next-studio`, `dev:turbo-studio` and https://test-next-studio.sanity.build/test/structure deployments.

### Testing

If the vercel deployment for `test-next-studio` is green on this PR we're good to go 🙌 

### Notes for release

N/A


RitaDias (approved):
Seems low stakes so ✅ 
Carolina is out so if you want to get someone else from DX to give a green we should reach out but otherwise 👍

PR #6694, opened 2024-05-16
fix: update `@sanity/presentation` and remove deprecated internal exports no longer in use

stipsan:
### Description

Follows up on https://github.com/sanity-io/sanity/pull/6684.
It adds back the original CODEOWNERS on file paths that CDX want to review to ensure backwards compatibility. This allows the Studio DX and EX teams to continue to receive code review requests in the same way as before #6684 merged.

It also bumps `@sanity/presentation` to a new version that no longer use the deprecated imports:
```
import {
  type CommentIntentGetter,
  CommentsIntentProvider,
} from 'sanity/structure'
```
Internally it now uses:
```
import {
  type CommentIntentGetter,
  CommentsIntentProvider,
} from 'sanity'
```
And it's tested working on https://visual-editing-studio.sanity.build already 🙌 

### What to review

If presentation continues to work as it does here https://test-studio.sanity.build/presentation/presentation then we're good to go.

### Testing

Existing tests are sufficient

### Notes for release

N/A as it's an internals refactor that makes no difference to userland.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -71,28 +71,28 @@ packages/sanity/src/core/studio/upsell/ @sanity-io/studio-ex
 
 # Internals used by @sanity/presentation
 # See https://github.com/sanity-io/visual-editing/blob/main/packages/presentation/src/internals.ts for exactly which exports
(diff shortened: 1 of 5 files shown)


ricokahler (approved):
💖

PR #6695, opened 2024-05-16
fix(tasks): hide footer button until the active document is set

pedrobonamin:
### Description

The footer button was shown due to this condition being true in some documents which this value is undefined, until the active document value was correctly set.
`item.target?.document._ref === activeDocument?.documentId`  -> `undefined === undefined` 

By checking if the activeDocument is present before validating the values this issue is fixed.

The removal of the `enabled` check is intentional given the whole plugin won't be added if the plugin is not enabled, so this check is unnecessary. See [resolveDefaultPlugins](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/config/resolveDefaultPlugins.ts)

Added some tests.

### DIFF
--- a/packages/sanity/src/core/tasks/plugin/TasksFooterOpenTasks.test.tsx
+++ b/packages/sanity/src/core/tasks/plugin/TasksFooterOpenTasks.test.tsx
@@ -1,0 +1,170 @@
+import {beforeAll, beforeEach, describe, expect, it, jest} from '@jest/globals'
+import {LayerProvider, studioTheme, ThemeProvider, useMediaIndex} from '@sanity/ui'
+import {uuid} from '@sanity/uuid'
+import {render, screen, waitFor} from '@testing-library/react'
+import {act} from 'react'
+
+import {TasksEnabledProvider, TasksNavigationProvider, TasksProvider} from '../context'
+import {useTasksStore} from '../store'
+import {type TaskDocument} from '../types'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/tasks/plugin/TasksFooterOpenTasks.test.tsx
jordanl17, line 1: These tests are great 🚀

jordanl17 (approved):

packages/sanity/src/core/tasks/plugin/TasksFooterOpenTasks.tsx
```diff
@@ -25,30 +25,28 @@ const ButtonContainer = styled.div`
...
-        return (
-          item.target?.document._ref === activeDocument?.documentId &&
-          item.status === 'open' &&
-          item.createdByUser
-        )
-      }),
+      activeDocument?.documentId
+        ? data.filter((item) => {
+            return (
+              item.target?.document._ref === activeDocument?.documentId &&
```
jordanl17, line 34: COULD: get rid of the optional chain here since `activeDocument` will always be defined in this block

pedrobonamin: Sure! Thanks for noticing.

jordanl17 (approved):
Nice

PR #6697, opened 2024-05-16
fix(core, structure): rename isOpen, add disable to actions bar

RitaDias:


packages/sanity/src/core/form/studio/tree-editing/utils/isArrayDialogOpen.ts
```diff
@@ -0,0 +1,17 @@
+import {toString} from '@sanity/util/paths'
+import {isObjectSchemaType, type ObjectSchemaType, type Path} from 'sanity'
+
+import {getSchemaField} from './getSchemaField'
+
+export function isArrayDialogOpen(schemaType: ObjectSchemaType, focusPath: Path): boolean {
```
hermanwikner, line 6: I think we should add an `@internal` flag since this function will be exported from the `sanity` package.

```ts
/**
 * @internal
 * A utility function to check if the global array editing dialog is open.
 */
```

RitaDias: Good catch, I've added the internal but also some info about the params and return

hermanwikner (approved):

PR #6705, opened 2024-05-16
refactor(portable-text-editor): automatically resolve some validation errors

skogsmaskin:
### Description

This will introduce a new prop to validation errors resolutions: `autoResolve` 

When a validation error occurs with such a resolution, the error will be automatically fixed for the user.

Currently the following validation errors will be automatically fixed:

* <strike>Missing or `.markDef` prop on a text block.</strike> - EDIT: should support missing markDefs prop
* Missing `_key` on span nods
* Empty children array for text blocks.

These errors will be fixed through patches directly, meaning that they can be resolved without doing it through transformations in the editor. This allows for resolving the errors progressively so that the editor finally can be rendered when all the errors are resolved.

Besides this, I have also added editor normalization for the following inside the editor:

* Remove orphaned annotation marks
* Remove decorator marks not supported by the schema.

These will be normalized whenever an edit occurs to that node.

I have also refactored the `addAnnotation` static function because everything there now has to be called within the `withouthNormalizing` scope in order not to trigger the above mentioned normalization (in flight as we build the new annotation).

### What to review

That the changes are not causing edits when a user is only opening the editor.
Nothing should be patched automatically unless the user actually edits in the editor.

### Testing

Should be covered by automatic tests.

However it would be nice with some manual testing as well. You can do so by for instance using the Vision plugin to set a block to certain value, for example without markDefs or with some mark that doesn't exist.

### Notes for release
* Added automatically resolving of some validation errors in the Portable Text Editor


packages/@sanity/portable-text-editor/src/editor/__tests__/pteWarningsSelfSolving.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+                marks: [],
+              },
+            ],
+            markDefs: [],
+            style: 'normal',
+          },
+        ])
+      }
+    })
+  })
```
pedrobonamin, line 113: The description mentions 3 scenarios in which this will autoresolve:
- Missing or .markDef prop on a text block.
- Missing _key on span nods
- Empty children array for text blocks.

For the first two we have tests, but not for the last one, could we include one test for the empty children?

packages/@sanity/portable-text-editor/src/utils/__tests__/values.test.ts
```diff
@@ -72,7 +72,6 @@ Array [
         "text": "123",
       },
     ],
-    "markDefs": Array [],
```
pedrobonamin, line 75: I may be lost here, but shouldn't this be always present when we have a marks in a children? Why this was removed here?

packages/@sanity/portable-text-editor/src/editor/plugins/__tests__/withPortableTextMarkModel.test.tsx
```diff
@@ -203,7 +203,7 @@ describe('plugin:withPortableTextMarksModel', () => {
             focus: {path: [{_key: 'a'}, 'children', {_key: 'a1'}], offset: 0},
             anchor: {path: [{_key: 'b'}, 'children', {_key: 'b1'}], offset: 1},
           })
-          PortableTextEditor.toggleMark(editor, 'bold')
+          PortableTextEditor.toggleMark(editor, 'strong')
```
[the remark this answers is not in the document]
pedrobonamin: Or.. is this because now it will be removed because "bold" is not a valid mark? Question, how do you ensure that a user who is doing local development of the schema won't lose it's changes if another user is at the same place but with the old schema?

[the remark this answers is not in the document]
skogsmaskin: Correct. The bold was never supported by the schema in the first place 😅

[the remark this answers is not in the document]
skogsmaskin: The normalization done by the editor will only apply when the node is edited. That's why this normalization is only done through the editor, and not validated by validateValue.

packages/@sanity/portable-text-editor/src/utils/__tests__/values.test.ts
skogsmaskin: The schema actually conciders this property optional. And additionally we need to be able to test if the value has changed regarding this prop (for example when a validation resolution is performed that inserts empty markDefs) we need to be able to make this distinction.

packages/@sanity/portable-text-editor/src/editor/__tests__/pteWarningsSelfSolving.test.tsx
skogsmaskin: Yes, I agree, I'll add it.

skogsmaskin:
I have removed the validation for missing .markDefs as this is allowed by the schema. Updated some code to reflect this.

pedrobonamin (approved):
Thanks @skogsmaskin looks good to me

PR #6706, opened 2024-05-16
fix(core/form): prohibit focus and unset on root input

skogsmaskin:
### Description

<strike>NOTE: this PR builds on top of https://github.com/sanity-io/sanity/pull/6705 and this should be merged first.</strike>

This will prohibit that a input for a object modal inside the PT-input can call `onPathFocus` or `unset()` on itself.

Calling `onPathFocus([])` will focus the editor element and close the editing modal.

Also if the input calls `unset()` on itself (like the Reference Input) it will cause the modal to abruptly dissapear, and we might miss out on any cleanup and other expected functions in the editor when this happens.

This change will make sure that any input for the PT-content inside the editor will not be allowed to do this.
`onPathFocus([])` will be a no-op, and `unset([])` will instead reset the object (keeping `_key` and `_type`) allowing the editor to deal normally with the object (will be removed when the modal is closed and object it empty).

Also added test for this.

### What to review
That it seems like a good idea and the proper solution.

### Testing
Should be covered by automatic tests. See an example in the provided story if you want to test manually.

### Notes for release
* Fixed bugs that related to when the PT-input is using a Reference input directly as the annotation type (not wrapped in a object type).

### DIFF
--- a/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/DisableFocusAndUnset.spec.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/DisableFocusAndUnset.spec.tsx
@@ -1,0 +1,51 @@
+/* eslint-disable max-nested-callbacks */
+import {expect, test} from '@playwright/experimental-ct-react'
+import {type Path, type SanityDocument} from '@sanity/types'
+
+import DisableFocusAndUnsetStory from './DisableFocusAndUnsetStory'
+
+export type UpdateFn = () => {focusPath: Path; document: SanityDocument}
+
+const document: SanityDocument = {
+  _id: '123',
+  _type: 'test',
+  _createdAt: new Date().toISOString(),
+  _updatedAt: new Date().toISOString(),
+  _rev: '123',
+  body: [
+    {
+      _key: 'a',
+      _type: 'block',
+      children: [{_key: 'b', _type: 'span', text: 'Foo', marks: ['123']}],
+      markDefs: [
+        {
+          _key: '123',
+          _type: 'link',
+          href: 'http://example.com',
+        },
+      ],
+    },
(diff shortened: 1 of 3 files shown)


packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
```diff
@@ -142,7 +142,24 @@ export function useSyncValue(
...
+                    if (
+                      !validation.valid &&
+                      validation.resolution?.autoResolve &&
+                      validation.resolution?.patches.length > 0
+                    ) {
+                      // Only apply auto resolution if the value has been populated before and is different from the last one.
+                      if (!readOnly && previousValue.current && previousValue.current !== value) {
+                        // Give a console warning about the fact that it did an auto resolution
+                        console.warn(
+                          `${validation.resolution.action} for block with _key '${validationValue[0]._key}'. ${validation.resolution?.description}`,
```
jtpetty, line 155: Should `${validation.resolution.action}` be `${validation.resolution?.action}` ?

packages/@sanity/portable-text-editor/src/editor/hooks/useSyncValue.ts
```diff
@@ -142,7 +142,24 @@ export function useSyncValue(
...
+                        )
+                        validation.resolution.patches.forEach((patch) => {
```
jtpetty, line 157: Is `validation.resolution` guaranteed to be non-null here?

skogsmaskin: Yes, both of these are inside a `validation.resolution?.autoResolve` test which will evaluate to false if there is no resolution (`.?`).

pedrobonamin (approved):

PR #6514, opened 2024-04-29
feat: add CLI options to enable auto-updating studios

ricokahler:
### Description

Adds the `--enable-auto-updates` and `unstable_enableAutoUpdates` flags to the CLI build options and config.

Additionally this PR:
- Adds `importMap`  support to `renderDocument` and `sanityBuildEntries` via `node-html-parser` (see tests)
- Refactors the external array to derive the external values from the specified import map.

Note that this will only affect the built version of the Studio and local versions will not use import maps. This decision was made because we did not want to provide a react dev build however this could potentially be possible now with import maps.

### What to review

- Review the logic, did I miss anything?
- Take a look at the tests, ensure the import map is being added correctly.

### Testing

I added tests and I tried this out in the test-studio locally.

### Notes for release

N/A

PR #6707, opened 2024-05-16
feat(cli): show prompt if local version doesn't match remote

binoy14:
### Description

Shows a prompt if the installed version of importmap deps are different then what it would resolve to.

FIXES SDX-1342

### What to review

Updates makes sense.

To test no error case in test-studio

1. cd `dev/test-studio/` 
2. run `pnpm sanity:build --auto-updates` 

Expected: No warning

To test error case in test-studio

1. Update the `sanity` and/or `@sanity/vision` version to something older in `dev/test-studio/package.json` and run `pnpm install`
2. cd `dev/test-studio` 
3. run `pnpm sanity:build --auto-updates` 

Expect: will show copy with different versions

### Testing

It's probably very hard to do a more integration level test for this, I wrote a unit test for the compare function at least

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
@@ -59,20 +59,12 @@ export default async function buildSanityStudio(
   const autoUpdatesEnabled =
     flags['auto-updates'] ||
     (cliConfig && 'autoUpdates' in cliConfig && cliConfig.autoUpdates === true)
-  let importMap
+
+  const version = encodeURIComponent(`^${installedSanityVersion}`)
+  const autoUpdatesImports = getAutoUpdateImportMap(version)
 
   if (autoUpdatesEnabled) {
     output.print(`${info} Building with auto-updates enabled`)
-
-    const version = encodeURIComponent(`^${installedSanityVersion}`)
-    const autoUpdatesImports = getAutoUpdateImportMap(version)
-
-    importMap = {
-      imports: {
-        ...(await buildVendorDependencies({cwd: workDir, outputDir})),
-        ...autoUpdatesImports,
-      },
-    }
 
     // Check the versions
     try {
@@ -154,6 +146,17 @@ export default async function buildSanityStudio(
 
   const trace = telemetry.trace(BuildTrace)
   trace.start()
+
+  let importMap
+
+  if (autoUpdatesEnabled) {
+    importMap = {
+      imports: {
+        ...(await buildVendorDependencies({cwd: workDir, outputDir})),
+        ...autoUpdatesImports,
+      },
+    }
+  }
 
   try {
     timer.start('bundleStudio')
--- a/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/build/buildAction.ts
@@ -16,6 +16,7 @@ import {getTimer} from '../../util/timing'
 import {BuildTrace} from './build.telemetry'
 import {buildVendorDependencies} from '../../server/buildVendorDependencies'
 import {compareStudioDependencyVersions} from '../../util/compareStudioDependencyVersions'
+import {getAutoUpdateImportMap} from '../../util/getAutoUpdatesImportMap'
 
 const rimraf = promisify(rimrafCallback)
 
@@ -58,13 +59,24 @@ export default async function buildSanityStudio(
   const autoUpdatesEnabled =
     flags['auto-updates'] ||
     (cliConfig && 'autoUpdates' in cliConfig && cliConfig.autoUpdates === true)
+  let importMap
 
   if (autoUpdatesEnabled) {
     output.print(`${info} Building with auto-updates enabled`)
 
+    const version = encodeURIComponent(`^${installedSanityVersion}`)
+    const autoUpdatesImports = getAutoUpdateImportMap(version)
+
+    importMap = {
+      imports: {
+        ...(await buildVendorDependencies({cwd: workDir, outputDir})),
+        ...autoUpdatesImports,
+      },
+    }
+
     // Check the versions
     try {
-      const result = await compareStudioDependencyVersions(workDir)
+      const result = await compareStudioDependencyVersions(autoUpdatesImports, workDir)
 
       if (result?.length) {
         const shouldContinue = await prompt.single({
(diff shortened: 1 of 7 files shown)


binoy14:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6707?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#6707** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6707?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#6514** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6514?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `next`

<h2></h2>

ricokahler (approved):

ricokahler (dismissed):
very nice

packages/sanity/src/_internal/cli/util/__tests__/compareStudioDependencyVersions.test.ts
```diff
@@ -0,0 +1,177 @@
+import {beforeEach, describe, expect, it, jest} from '@jest/globals'
+import resolveFrom from 'resolve-from'
+
+import {compareStudioDependencyVersions} from '../compareStudioDependencyVersions'
+import {readPackageJson} from '../readPackageJson'
+
+jest.mock('resolve-from')
+jest.mock('../readPackageJson')
+
+global.fetch = jest.fn() as jest.MockedFunction<typeof fetch>
```
rexxars, line 10: I don't feel very comfortable with this approach, since it'll potentially leak into other tests running in the same process (not _quite_ certain how much jest/vitest isolates these, but feels dodgy). Could we make `fetch` an optional argument to `compareStudioDependencyVersions`, so we can inject it?

rexxars (changes requested):
Could you walk me through the flow for what will happen if the local version is higher than the available remote version? Will this fail in the step that resolves the remote version for the package, since it includes a minimum version? What is the error message shown to the user in that case? Would love to see a test case that covers this.

binoy14:
> Could you walk me through the flow for what will happen if the local version is higher than the available remote version? Will this fail in the step that resolves the remote version for the package, since it includes a minimum version? What is the error message shown to the user in that case? Would love to see a test case that covers this.

My understanding is that if the version is not equal we show a prompt, I am basically doing a strict version check so it doesn't matter if it's minVersion is lower or greater. Not sure if that is the intention we want but my feeling is that you want to test locally with same version as remote but happy to discuss further.

binoy14:
@rexxars let me know if my comment and test changes makes sense

packages/sanity/src/_internal/cli/actions/build/buildAction.ts
```diff
@@ -58,8 +59,40 @@ export default async function buildSanityStudio(
     flags['auto-updates'] ||
     (cliConfig && 'autoUpdates' in cliConfig && cliConfig.autoUpdates === true)
 
+  // Get the version without any tags if any
+  const coercedSanityVersion = semver.coerce(installedSanityVersion)?.version
+  if (autoUpdatesEnabled && !coercedSanityVersion) {
+    throw new Error(`Failed to parse installed Sanity version: ${installedSanityVersion}`)
+  }
+  const version = encodeURIComponent(`^${coercedSanityVersion}`)
+  const autoUpdatesImports = getAutoUpdateImportMap(version)
```
rexxars, line 68: Can we move this into the conditional below? Feels like unnecessary work and potentially throwing if the user has not opted in to auto updates

packages/sanity/src/_internal/cli/util/compareStudioDependencyVersions.ts
```diff
@@ -0,0 +1,82 @@
...
+ * specified in the `autoUpdatesImports` parameter with their remote versions. If the versions do not match, the dependency is
+ * added to a list of failed dependencies, which is returned by the function.
+ *
+ * The failed dependencies are anything that does not strictly match the remote version.
+ * This means that if a version is lower or greater by even a patch it will be marked as failed.
+ *
+ * @param AutoUpdatesImportMap - An object mapping package names to their remote import URLs.
+ * @param string - The path to the working directory containing the package.json file.
+ *
+ * @returns Promise\<Array\<CompareStudioDependencyVersions\>\> - A promise that resolves to an array of objects, each containing
```
rexxars, line 37: TSDoc does not require you to specify the type, only a description. The type is (and should be) inferred from the types themselves - see https://tsdoc.org/pages/tags/returns/

packages/sanity/src/_internal/cli/util/compareStudioDependencyVersions.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Compares the versions of dependencies in the studio with their remote versions.
+ *
+ * This function reads the package.json file in the provided working directory, and compares the versions of the dependencies
+ * specified in the `autoUpdatesImports` parameter with their remote versions. If the versions do not match, the dependency is
+ * added to a list of failed dependencies, which is returned by the function.
+ *
+ * The failed dependencies are anything that does not strictly match the remote version.
+ * This means that if a version is lower or greater by even a patch it will be marked as failed.
+ *
+ * @param AutoUpdatesImportMap - An object mapping package names to their remote import URLs.
```
rexxars, line 34: `@params` should reference the parameter name, not the type. The type is inferred.

```suggestion
 * @param autoUpdatesImports - An object mapping package names to their remote import URLs.
```

packages/sanity/src/_internal/cli/util/compareStudioDependencyVersions.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param string - The path to the working directory containing the package.json file.
```
rexxars, line 35: Same as above
```suggestion
 * @param workDir - The path to the working directory containing the package.json file.
```

packages/sanity/src/_internal/cli/util/compareStudioDependencyVersions.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
```
rexxars, line 36:
```suggestion
 * @param fetchFn - Optional {@link https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API | Fetch}-compatible function to use for requesting the current remote version of a module
 *
```

rexxars (changes requested):

packages/sanity/src/_internal/cli/actions/build/buildAction.ts
binoy14: @rexxars the reason I have it outside the conditional is because the `autoUpdatesImports` variable is used below in line `160` If I make a variable like `let autoUpdatesImports` TS can't necessarily tell if the variable is initialized so it complains. The problem with moving the whole thing after the clean is that it does into a spinner and prompt doesn't work with spinner. I think there is some opportunity to clean this up a little bit but unless you strongly feel that we should do it now, punt it for future work?

rexxars: Makes sense. Can refactor later 👍

rexxars (approved):

PR #6709, opened 2024-05-17
fix(core): add retry to image observer and loading state

pedrobonamin:
### Description

We have a racing condition in the `BaseImageInput` component, in some cases, specially with big images the fetch request to resolve the reference is happening before the reference exists. Making the image input to persist in a "loading" state which nothing indicates it. 
Also, there was no retry option when the reference was not properly resolved.

With this changes, a retry option is added to the  `observeAssetDoc` function, fixing this problem.
Also, a loading state was included in the image input.

How to reproduce this issue:
- Go to https://test-studio-git-next.sanity.build/test/structure/
- Open any document with an image input.
- Add a large image which is not in the client assets, see if it stales, if not, remove the image and try with another one.
- Retry previous step until you can see the issue.

#### Reproducing the issue in next
https://github.com/sanity-io/sanity/assets/46196328/bed7456e-f5de-472a-a28c-e4d2fffdc29a

#### After fix
After the fix I was not able to reproduce this issue anymore.

https://github.com/sanity-io/sanity/assets/46196328/dad16eb0-6a1a-4664-875c-c12cc3a2e878

### Testing
Manual testing of the issue, couldn't reproduce it.

### DIFF
--- a/packages/sanity/src/core/form/studio/inputs/client-adapters/assets.ts
+++ b/packages/sanity/src/core/form/studio/inputs/client-adapters/assets.ts
@@ -78,6 +78,10 @@ export const uploadImageAsset = (
 export const uploadFileAsset = (client: SanityClient, file: File | Blob, options?: UploadOptions) =>
   uploadAsset(client, 'file', file, options)
 
+/**
+ *
+ *
+ */
 // note: there's currently 100% overlap between the ImageAsset document and the FileAsset documents as per interface required by the image and file input
 function observeAssetDoc(documentPreviewStore: DocumentPreviewStore, id: string) {
   return documentPreviewStore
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/studio/inputs/client-adapters/assets.ts
```diff
@@ -80,17 +80,36 @@ export const uploadFileAsset = (client: SanityClient, file: File | Blob, options
...
+      }),
+      retry({
+        count: 5,
+        delay: 1000,
+      }),
+      catchError((err) => {
+        console.error('Final error after retries, asset not found in documentPreviewStore:', err)
+        return of(null) // Return null if the asset is not found
+      }),
+    )
```
bjoerge, line 112: Could you add a comment explaining what issue these operators adds a workaround for?

bjoerge (changes requested):
Great find. We should investigate why we're getting stale data when refetching as a response to an event received from a listener set up with `visibility=query`, but this is a nice workaround in the meantime.

Would be great if you could add a comment to the operator chain to make it easier to know if/when it can be removed.

bjoerge (approved):
Thanks @pedrobonamin!

PR #6711, opened 2024-05-17
feat(sanity): use optimistic locking when publishing documents

juice49:
### Description

This branch adds optimistic locking when publishing documents via the Actions API by adopting its `ifDraftRevisionId` and `ifPublishedRevisionId` guards. These guards are a recent addition to the Actions API.

Optimistic locking ensures the editor can see the latest state of both the draft document they are publishing, and the published document they are choosing to replace.

Additionally, this branch adds a check prior to publishing that ensures Studio has a draft snapshot. This check is already in place when publishing via the Mutations API. It stands to reason that the editor must be able to see the draft they are choosing to publish.

### What to review

Is the usage of the `ifDraftRevisionId` and `ifPublishedRevisionId` guards correct?

### Testing

- I've copied the existing unit tests from the Mutations API based publish operation to the Actions API based publish operation.
- These tests use snapshots to ensure the expected revision guards are used.
- I've added a test to ensure `execute` throws if the client has no draft snapshot.
- You can find these tests in `packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/publish.test.ts`.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/__snapshots__/publish.test.ts.snap
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/__snapshots__/publish.test.ts.snap
@@ -92,3 +92,17 @@ Object {
   "transaction": Array [],
 }
 `;
+
+exports[`publish execute throws an error if the client has no draft snaphot 1`] = `
+Object {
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
Nice

PR #6713, opened 2024-05-17
EDX-1352: batch select sheet list documents

jordanl17:
### Description
EDX-1352: allowing multiselection of rows using shift+click

### What to review
The dir structure has been changed so that all `sheetList` components and files are inside a separated dir within `documentList` and `EX` has been added a code owner.

### Testing
Tests have been added to the row selection component - `DocumentSheetListSelect` to test around shift click functionality

### Notes for release
No relevant notes as this feature is not currently exposed to users


packages/sanity/src/structure/panes/documentList/sheetList/__tests__/DocumentSheetListSelect.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const checkbox = screen.getByRole('checkbox')
+
+    expect(checkbox).toBeChecked()
+
+    fireEvent.click(checkbox)
+    expect(mockToggleSelected).toHaveBeenCalledTimes(1)
+    expect(selectProps.table.options.meta?.setSelectedAnchor).toHaveBeenCalledWith(null)
+  })
+
+  it('disables the checkbox when row is not selectable', () => {
```
jtpetty, line 73: what's the reason that a row is not selectable? For a11y, it would be best to not have elements disabled.

packages/sanity/src/structure/panes/documentList/sheetList/useDocumentSheetColumns.tsx
```diff
@@ -126,18 +127,17 @@ export function useDocumentSheetColumns(documentSchemaType?: ObjectSchemaType) {
...
-            disabled={!info.row.getCanSelect()}
-            onChange={() => info.row.toggleSelected()}
-          />
+          <Box>
+            <Checkbox
+              style={{paddingLeft: 4}}
+              indeterminate={info.table.getIsSomeRowsSelected()}
+              onChange={info.table.getToggleAllPageRowsSelectedHandler()}
+            />
+            {/* eslint-disable-next-line i18next/no-literal-string */}
```
jtpetty, line 136: I wonder if `selected` should be added to the attribute exceptions list

jtpetty (approved):
a few minor questions, but 👍 LGTM

packages/sanity/src/structure/panes/documentList/sheetList/__tests__/DocumentSheetListSelect.test.tsx
jordanl17: Oh totally! Currently there is no reason why a row would be selected, but tanstack table does support this functionality, so I thought I'd add a theoretical test case to ensure that any future disabling. In the future we mihgt disable, perhaps whilst we are running a batch publish and don't want the user change row selection. Added here so it's already tested

packages/sanity/src/structure/panes/documentList/sheetList/useDocumentSheetColumns.tsx
jordanl17: This was just some text I'm printing to the DOM so we can make sure that the correct rows are being selected. It's quite likely this would be deleted when we start improving styling

pedrobonamin (approved):
Thanks Jordan, looks good to me. Let's have it merged after the release

packages/sanity/src/structure/panes/documentList/sheetList/ColumnsControl.tsx
ninaandal, line 1: We should be using the StudioUI components here as well, right?

jordanl17: The presentation of this component is just placeholder for now since we are still awaiting designs

PR #6661, opened 2024-05-14
feat(structure): Sheet View columns

jordanl17:
### Description
This PR adds in controls to manage the number of fields to show in Sheet List, and also supports nested fields 2 levels deep.
Further, it adds in a `SheetListCell` component which represents a single cell in the table

### Testing
Testing added for `ColumnsControl` component

### Notes for release
Nothing to mention. This is not gonna be accessible for users yet

PR #6716, opened 2024-05-20
fix(structure): fixing incorrect way of setting col visibility

jordanl17:
### Description
Fixes an issue introduced in EDX-1308 (https://github.com/sanity-io/sanity/pull/6661) where tanstack col visibility expects field separator as `_` but incorrectly `.` was being used for our definitions

### What to review
Defining an id for all cols which is used in flatColumns.
Selected column now a display column

### Testing
Updated test case

### Notes for release
Not relevant - this feature is not yet user facing

### DIFF
--- a/packages/sanity/src/structure/panes/documentList/__tests__/useDocumentSheetColumns.test.ts
+++ b/packages/sanity/src/structure/panes/documentList/__tests__/useDocumentSheetColumns.test.ts
@@ -1,4 +1,5 @@
 import {describe, expect, it, jest} from '@jest/globals'
+import {type ObjectSchemaType} from '@sanity/types'
 import {renderHook} from '@testing-library/react'
 
 import {useDocumentSheetColumns} from '../useDocumentSheetColumns'
@@ -20,7 +21,7 @@ describe('useDocumentSheetColumns', () => {
         {name: 'email', type: {name: 'string'}},
         {name: 'age', type: {name: 'number'}},
         {
-          name: 'address',
+          name: 'address.place',
           type: {
             name: 'object',
             jsonType: 'object',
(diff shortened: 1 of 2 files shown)


binoy14 (commented):
don't have full context of what is going on here, probably makes more sense for Pedro to give a ✅

packages/sanity/src/structure/panes/documentList/__tests__/useDocumentSheetColumns.test.ts
```diff
@@ -20,7 +21,7 @@ describe('useDocumentSheetColumns', () => {
         {name: 'email', type: {name: 'string'}},
         {name: 'age', type: {name: 'number'}},
         {
-          name: 'address',
+          name: 'address.place',
```
pedrobonamin, line 24: This type of schema name won't be allowed in the studio, it doesn't support dot notation in the naems
Screenshot 2024-05-21 at 08 11 35

packages/sanity/src/structure/panes/documentList/useDocumentSheetColumns.tsx
```diff
@@ -156,17 +160,19 @@ export function useDocumentSheetColumns(documentSchemaType?: ObjectSchemaType) {
     () =>
       flatColumns(columns).reduce<[VisibilityState, number]>(
         ([accCols, countAllowedVisible], column) => {
+          const visibilityKey = String(column.id)
```
pedrobonamin, line 163: Column id is possibly undefined according to the tanstack column definition and accessor functions, thinking on throwing an error if it doesn't have a `column.id`
We are controlling this, but in case something changes in the future it will be simpler to identify the problem if for some case the id is missing.

packages/sanity/src/structure/panes/documentList/__tests__/useDocumentSheetColumns.test.ts
jordanl17: - [ ] tests don't need to test with dot notation

packages/sanity/src/structure/panes/documentList/useDocumentSheetColumns.tsx
jordanl17: - [ ] throw if there's not an ID on a column

Issue #6688, opened 2024-05-16
All Studio UI menus broken/empty in Dev


mdillgen:
**Describe the bug**

When running in development (npm run dev) the Studio UI Menus are empty!

**To Reproduce**

1. Run dev server
2. Login to Studio
3. Click on User Menu Avatar, Filter, Create new… etc.
4. See empty menu

Reproduction repo:
https://github.com/mdillgen/template-nextjs-personal-website.git

**Screenshots**
![Bildschirmfoto 2024-05-16 um 09 11 55](https://github.com/sanity-io/sanity/assets/9297500/b18c290b-df27-48e2-aecd-94209427c901)
![Bildschirmfoto 2024-05-16 um 11 54 24](https://github.com/sanity-io/sanity/assets/9297500/de039d9c-bfa8-41d5-94ed-46fead41cee1)

**Which versions of Sanity are you using?**

```
@sanity/demo                 2.0.0 (up to date)
@sanity/icons               2.11.8 (up to date)
@sanity/image-url            1.0.2 (up to date)
@sanity/preview-url-secret  1.6.13 (up to date)
@sanity/react-loader        1.9.19 (up to date)
@sanity/ui                   2.1.8 (up to date)
@sanity/vision              3.42.1 (up to date)
sanity                      3.42.1 (up to date)
```
**What operating system are you using?**
`Mac OS`
**Which versions of Node.js / npm are you running?**

Run `npm -v && node -v` in the terminal and copy-paste the result here.
```
10.1.0
v20.9.0
```

filipjohnsen:
Having the same issue, adding these styles to my studio fixed the issue for me. 
```css
[data-ui="versionMenu"],
[data-ui="MenuButton__popover"],
[data-ui="Popover"] {
  opacity: 1 !important;
  --motion-content-opacity: 1 !important;
}
```

PR #6717, opened 2024-05-21
fix: pin framer-motion to an earlier version

christianhg:
`@sanity/ui` `Popover` animations don't run properly when the Studio is embedded in a Next.js canary app. This change downgrades `framer-motion` to the latest version we shipped `@sanity/ui` with and that we know works. We can upgrade again later once the underlying problem has been fixed.

- `@sanity/ui` has already been pinned here: https://github.com/sanity-io/ui/pull/1302 and
- `renovate-config` has been pinned here: https://github.com/sanity-io/renovate-config/pull/197

This is already enough to fix https://github.com/sanity-io/sanity/issues/6688, but @stipsan and I thought it would be good to pin the dependency everywhere.

### DIFF
--- a/packages/sanity/package.json
+++ b/packages/sanity/package.json
@@ -195,7 +195,7 @@
     "esbuild-register": "^3.4.1",
     "execa": "^2.0.0",
     "exif-component": "^1.0.1",
-    "framer-motion": "11.2.3",
+    "framer-motion": "11.0.8",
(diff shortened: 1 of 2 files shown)


stipsan (approved):
Yes, it's important that we use the same version of `framer-motion` everywhere so we don't bloat the bundle 🙌

PR #6718, opened 2024-05-21
fix(tasks): reduce tasks sidebar zoffset

pedrobonamin:
### Description

**Modals** have a z-index of 600.
**Popover** have a z-index of 400.
**Tooltips** have a z-index of 200

Reducing tasks sidebar to **100**, the same as the navbar to stay below modals, popovers and tooltips, it will still sit on top of the elements that were discovered as breaking on the previous fix when this was moved to 1000.


RitaDias (approved):
💯

PR #6725, opened 2024-05-21
feat(sanity, core): add config for array editing

RitaDias:


packages/sanity/src/core/form/studio/tree-editing/utils/useLegacyArrayEditingConfig.ts
```diff
@@ -0,0 +1,13 @@
+import {useSource} from 'sanity'
+
+/**
+ * A utility hook to check if the array editing dialog should be open (based on config)
+ * @returns Returns true if the dialog should be open
+ * @internal
+ */
+export function useLegacyArrayEditingConfig(): boolean {
```
hermanwikner, line 8: Given that the name of this hook implies it should return the _config_, should we return the entire config (i.e, `document.unstable_legacyArrayEditing `) instead of just the `enabled` property from the config? If we decide to return only the `enabled` property (as we do currently) – should we consider renaming the hook to `useLegacyArrayEditingEnabled()` to better reflect its purpose?

packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
```diff
@@ -145,7 +146,12 @@ export function TreeEditingDialog(props: TreeEditingDialogProps): JSX.Element |
     [debouncedBuildTreeEditingState, onPathOpen],
   )
 
+  const shouldUseLegacyArrayDialog = useLegacyArrayEditingConfig()
+
   useEffect(() => {
+    // Check if the config has the legacy flag
+    if (shouldUseLegacyArrayDialog) return
```
hermanwikner, line 153: Do we need this check here since we are conditionally rendering the `TreeEditingDialog` based on this value in `FormBuilder`?

packages/sanity/src/core/form/studio/tree-editing/utils/useLegacyArrayEditingConfig.ts
RitaDias: Done 👍

packages/sanity/src/core/config/types.ts
```diff
@@ -671,6 +681,14 @@ export interface Source {
     comments: {
       enabled: (props: DocumentCommentsEnabledContext) => boolean
     }
+
+    /**
+     * @internal
+     * @hidden
+     */
+    unstable_legacyArrayEditing?: {
```
hermanwikner, line 689: I think it would make more sense to have this property in `form` rather than in `document`. There is already an `unstable` property in `form` (`form.unstable.<thing>`) where this property could potentially be added to!

What do you think? :) 

```ts
export default defineConfig({
  // ...

  form: {
    unstable: {
      legacyArrayEditing: true,
    },
  },
})
```

RitaDias: I'm not married to the approach, the first time I did think about putting it in form. I thought about document because I felt it might be something that impacts outside of the form just itself (mainly considering that the "disable" of actions is outside of the form though it's impacted about the editing dialog).

It made more sense to me to keep it in document because of that. What do you think? Keep it in document or move it to form.unstable?

hermanwikner: In this case, I think it makes more sense to have this property in `form`, as that’s where the current configuration of the form is happening (such as inputs, fields, files, images, etc.). I'd say that the `document` property is more about configuring things _around_ the form, such as actions, badges, inspectors, comments, etc., whereas the `form` property is more focused on configuring the actual form itself. Since the new property we are adding is about configuring the array input in the form, I believe having the property in `form` rather than in the `document` is more correct!

Happy to discuss this further tomorrow in a meeting :)

RitaDias: Resolving this, after talking with Rune he suggested a different approach that aligns better with what DX wants to do as well. 
I've just pushed a commit with it, let's talk on our meeting :)

dev/test-studio/sanity.config.ts
```diff
@@ -157,6 +157,13 @@ export default defineConfig([
     tasks: {
       enabled: true,
     },
+    features: {
+      beta: {
+        treeArrayEditing: {
+          enabled: true,
+        },
+      },
+    },
```
hermanwikner, line 166: I think it would be nice to move this to the `sharedSettings` plugin so that it is enabled for every config/workspace!

```ts
const sharedSettings = definePlugin({
  // ...

  features: {
    beta: {
      treeArrayEditing: {
        enabled: true,
      },
    },
  },
})
```

RitaDias: Done :)

packages/sanity/src/core/config/types.ts
```diff
@@ -388,6 +388,18 @@ export interface PluginOptions {
...
+    /** allows for shorter-lived feature toggles */
+    beta?: {
+      /**
+       * @hidden
+       * @beta */
+      treeArrayEditing?: {
+        enabled: boolean
+      }
+    }
+  }
```
hermanwikner, line 402: Suggestion:  move this to its own interface (e.g, `Features`) since it is used in both the `PluginOptions` and `Source` interface

```ts
interface Features {
  /**
   * Configuration for beta features.
   */
  beta?: {
    /**
     * @beta
     * @hidden
     * */
    treeArrayEditing?: {
      /**
       * Enables the tree array editing feature.
       */
      enabled: boolean
    }
  }
}
```

Which are then used in the `PluginOptions` and `Source` interface:

```ts

/** @beta */
export interface Source {
  // rest of types ...

  /**  Configuration for various features. */
  features?: Features
}

/** @beta */
export interface PluginOptions {
  // rest of types ...

  /**  Configuration for various features. */
  features?: Features
}
```

hermanwikner: I also made some small updates to the TSDoc copy – what do you think?  :)

hermanwikner (approved):
Looks good! ✨ 

I added a suggestion to create a new dedicated `Features` interface, but no blocker for merge!

packages/sanity/src/core/config/types.ts
RitaDias: Great suggestion, I've updated it :)

PR #6727, opened 2024-05-21
fix(core): dont crash when image url string is passed to preview

sjelfull:
### Description

We currently have a situation where you can use a asset URL as a media preview value, and this will crash in the context of a block preview inside the PTE.

Here is a example schema:

```js
defineField({
    type: 'image',
    icon: ImageIcon,
    name: 'imageWithAssetUrlPreview',
    title: 'Image w/ asset url preview',
    options: {
      hotspot: true,
    },
    preview: {
      select: {
        media: 'asset.url',
      },
      prepare({media}) {
        return {media, title: 'Image w/ asset url preview'}
      },
    },
  }),
```

Before this change the value would be passed on to the default preview component, where we had these conditionals:

```js
 if (isValidElementType(mediaProp)) {
    return mediaProp
  }

  if (isValidElement(mediaProp)) {
    return mediaProp
  }

  if (isImageSource(mediaProp)) {
    return renderMedia
  }

  // Handle image urls
  if (isString(imageUrl)) {
    return (
      <img
        src={imageUrl}
        alt={isString(title) ? title : undefined}
        referrerPolicy="strict-origin-when-cross-origin"
      />
    )
  }
```

`isValidElementType(mediaProp)` would actually be true for any string here, meaning it would be passed on to `createElement`, and in turn it would crash.

CleanShot 2024-05-21 at 14 36 17@2x

After the change it is correctly handled, and the passed Sanity image URL will be transformed to the same dimensions as any other preview image.

CleanShot 2024-05-21 at 14 38 24@2x

### What to review

- Does the logic change make sense? Was there a reason we ordered the conditional checking, or was it simply a oversight that `isValidElementType` will return true for any string?

### Notes for release

- Fixes a situation where setting the `media` preview value of a image to `asset.url` would crash inside the Portable Text Editor

### DIFF
--- a/packages/sanity/src/core/preview/components/SanityDefaultPreview.tsx
+++ b/packages/sanity/src/core/preview/components/SanityDefaultPreview.tsx
@@ -111,7 +111,7 @@ export const SanityDefaultPreview = memo(function SanityDefaultPreview(
         <SanityDefaultMedia
           client={client}
           dimensions={options.dimensions}
-          imageSource={imageSource}
+          imageSource={imageSource as SanityImageSource}
           imageUrlBuilder={imageBuilder}
           title={title}
         />
--- a/packages/sanity/src/core/components/previews/_common/Media.tsx
+++ b/packages/sanity/src/core/components/previews/_common/Media.tsx
@@ -1,5 +1,5 @@
 import {Text} from '@sanity/ui'
-import {isValidElement, type ComponentType, type ReactNode} from 'react'
+import {type ComponentType, isValidElement, type ReactNode} from 'react'
 import {isValidElementType} from 'react-is'
 
 import {type PreviewLayoutKey, type PreviewMediaDimensions, type PreviewProps} from '../types'


packages/sanity/src/core/preview/components/SanityDefaultPreview.tsx
```diff
@@ -50,6 +50,13 @@ export function SanityDefaultPreview(props: SanityDefaultPreviewProps): ReactEle
       dimensions: {width?: number; height?: number; fit: ImageUrlFitMode; dpr?: number}
     }) => {
       const {dimensions} = options
+      let imageSource = mediaProp
+
+      if (isString(imageSource) && isSanityImageUrl(imageSource)) {
```
RitaDias, line 55: Can I have some comments here the same way you've done at the bottom please? 🙏

RitaDias (approved):
This looks good to me though perhaps PK might have some more insigh on the questions on  the body of the PR (order wise, was it intentional? I'm not sure)

Two questions that popped into my mind: 
- Is it possible to add some very quick tests for this? (so that we just make sure that it doesn't break in the future)
- The other one is in the code itself, not so much a question as a request ✨ 

Great work thanks for doing this!

packages/sanity/src/core/preview/components/SanityDefaultPreview.tsx
sjelfull: ✅

sjelfull:
@RitaDias Thanks for the review! I'll see if I can add some quick tests for this, thanks for the reminder.

bjoerge:
@sjelfull what's the status here? Are these changes still relevant or can we close this?

sjelfull:
@bjoerge I think it might be relevant still, easy to check by using the schema above to reproduce. Missing tests though, and need to be rebased.

PR #6728, opened 2024-05-21
chore(cli): symlink in local packages for the installed sanity cli

sgulseth:
### Description

We were only symlinking in local packages in the studio tests, however if the changes is only in the cli package we also need the changes locally in the installed CLI

### What to review

Can this break in any, unforeseen, way? :p 

### Testing

Fixes the tests in https://github.com/sanity-io/sanity/pull/6702

### Notes for release

n/a - no notes needed


bjoerge (approved):
LGTM (as long as the tests are green)

PR #6729, opened 2024-05-21
fix(core/form): set PTE in readOnly mode when PT-Input is not activated

skogsmaskin:
### Description

This ensures we don't taint the document  by patches produced by the normalization inside the PTE, by only loading the document in the editor.

The user must actively enable the input before any normalization patches are sent.

### Testing

This PR should be merged first: https://github.com/sanity-io/sanity/pull/6705
Then we can write a automatic test for this.

### Notes for release
N/A


skogsmaskin:
I think this might still be relevant.

jordanl17:
I resolved the conflict and tried to be pragmatic about where this check should be added, however with the substantial refactor away from raw PTE to an `EditorProvider` it seems, from my experimentation at least, that this change is no longer required with the inner fundamentals of the PTE having been updated to fix this.

PR #6730, opened 2024-05-21
chore: setup react compiler test and linting

stipsan:
### Description

Adds support for the [React Compiler](https://es.react.dev/learn/react-compiler) in the `test-next-studio`. Also adds the `eslint-plugin-react-compiler` plugin, which warns us about areas in the codebase that the compiler will skip due to either breaking the rule of hooks, or the code being structured in a way where deps can't be inferred.

Components that are memoized by the Compiler gets their own little sparkly badge in the React Devtools extension:
image

Why add the Compiler? Even though we can't fully reap the benefits of it until we're ready to drop react v18 and make v19 the new baseline, it's useful to be able to track how much perf is improved by the Compiler.
It's also usually a good thing for react v18 perf to refactor the codebase to fix hooks bugs, and to make the code structured so the compiler can fully understand it and optimize it.

We won't see a huge difference right away, as our codebase are in a state that's the combination of:
a) components that would benefit the most from the compiler is currently skipped by it (see `pnpm check:react-compiler` and errors from `react-compiler/react-compiler` to see which ones, for example )
b) most of the components that can be optimized by react-compiler and see a boost are already optimized by hand with `useMemo`, `useCallback` and `memo`.

We can work on a) over time and it benefits react v18 users as well as v19. While b) is something we can reap once we make v19 the baseline and we can delete all handwritten `useMemo`, `useCallback` and `memo` instances 🙌 

### What to review

The change in `useSyncValue` should be fine, as it's functionally equivalent it's only refactored so it can be understood by the compiler.

### Testing

Existing tests should suffice, it only affects the `dev/test-next-studio` project. There's also a new CI check that tells us if the compiler can run on the full codebase or not.

### Notes for release

N/A

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -30,6 +30,7 @@ const config = {
     '@typescript-eslint',
     'prettier',
     'react',
+    'react-compiler',
     'tsdoc',
     'unicorn',
   ],
@@ -68,6 +69,8 @@ const config = {
         additionalHooks: '(useMemoObservable|useObservableCallback|useAsync)',
       },
     ],
+    // Set react-compiler to `error` once existing issues are fixed
+    'react-compiler/react-compiler': 'warn',
     'react/no-unescaped-entities': 'off',
     'i18next/no-literal-string': ['error'],
     '@sanity/i18n/no-attribute-string-literals': [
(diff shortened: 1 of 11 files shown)


binoy14 (dismissed):
Changes makes sense to me, probably need a rebase?

PR #6737, opened 2024-05-22
fix(sanity): adjust ArrayOptions

christianhg:
- Removed the comment about `layout` since it's confusing/wrong. The `grid` option also works for a string array.
- Added `list` as a valid `layout` option. Even though it's the default, it feels weird that you can't specify it manually.
- Removed `direction` since it doesn't seem to do anything. Maybe it's old?


juice49 (changes requested):
Thanks, Christian! This looks great, and does more closely reflect the underlying functionality.

Instead of removing `direction`, please can you mark is as deprecated? Removing it is technically a breaking change to the API, even though it doesn't actually do anything.

christianhg:
> Thanks, Christian! This looks great, and does more closely reflect the underlying functionality.
> 
> Instead of removing `direction`, please can you mark is as deprecated? Removing it is technically a breaking change to the API, even though it doesn't actually do anything.

Thanks for the feedback! I've deprecated the option now, instead of removing it.

packages/@sanity/types/src/schema/definition/type/array.ts
```diff
@@ -11,13 +11,10 @@ import {type BaseSchemaDefinition, type SearchConfiguration, type TitledListValu
...
   list?: TitledListValue<V>[] | V[]
-  /**
-   * layout: 'tags' only works for string array
-   * layout: 'grid' only works for arrays with objects
-   */
   // inferring the array.of value for ArrayDefinition cause too much code-noise and was removed.
   // Since we don't have the type-info needed here, we allow values
-  layout?: 'tags' | 'grid'
+  layout?: 'list' | 'tags' | 'grid'
+  // @deprecated This option does have any effect anymore
```
juice49, line 17: One tiny thing: I think this needs to be a block comment to have an effect.

```suggestion
  /** @deprecated This option does have any effect anymore */
```

christianhg: Fixed!

juice49 (approved):
Looks good to me. Thanks!

christianhg:
> Looks good to me. Thanks!

Thank you! Feel free to merge it on my behalf. I don't have permissions to do so.

zchana:
fixes the React key warning that appeared when using grid layout
for reference arrays in Sanity Studio. The solution maintains the desired
grid layout while ensuring proper React key handling through a custom

PR #6740, opened 2024-05-22
chore: change the action to use warnings and only fail on new ones

stipsan:
Refactors the React Compiler CI check to no longer always fail, it now uses warnings.
It uses `--max-warnings` to prevent new code from introducing new issues.
The downside to this is that if issues are fixed it's important to decrement this number (currently `49`) otherwise new issues can be silently introduced.

It also sets up a GitHub problem matcher called `eslint-compact.json` which allows the actions to show up on the PR diff itself. It might also work for the regular `pnpm lint` command but I haven't tested it as it's a bit out of scope.

Failed checks show up inline on the PR, even though they are warnings:
image

It ignores comments that silences the check (we don't allow this):
image


stipsan:
@SocketSecurity ignore-all

binoy14 (approved):
makes sense to me, thanks!

PR #6742, opened 2024-05-23
feat(cli): build user peer-deps

ricokahler:
### Description

This PR introduces the `buildVendorDependencies` function which builds ESM browser-compatible versions of specified vendor packages. It includes a robust mechanism for handling semver versions, carefully considering major, minor, and patch version changes. The rationale is to ensure stability and flexibility by adding support for new subpath exports as needed. Read the TSDoc comment for more info.

### What to review

- Review the `buildVendorDependencies` function implementation.
- Verify the handling of semver versions and the rationale for the approach.
- Check the integration and usage of the `fileSystem` prop.
- Ensure the function correctly resolves package.json files and constructs the necessary entry points.
- Review the unit tests to ensure they cover the function's behavior, including handling different versions and file system interactions.

### Testing

- Unit tests verify the expected options are passed to `vite.build` along with some new fixtures example project to test that the resolution works as expected.

### Notes for release

Not applicable to the release at this time. This should be merged into #6514 first.


packages/sanity/src/_internal/cli/server/__tests__/buildVendorDependencies.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  it('should return the expected entry points for react 18', async () => {
+    const cwd = path.join(examplesRoot, 'prj-with-react-18')
+    const imports = await buildVendorDependencies({cwd, outputDir})
+
+    expect(imports).toEqual({
+      'react': '/vendor/react/index.mjs',
+      'react-dom': '/vendor/react-dom/index.mjs',
+      'react-dom/client': '/vendor/react-dom/client.mjs',
+      'react-dom/package.json': '/vendor/react-dom/package.json.mjs',
```
binoy14, line 41: ❓ What's the reason to have `package.json` in the import map?

binoy14 (commented):
Overall it looks good to me, have one question

packages/sanity/src/_internal/cli/server/__tests__/buildVendorDependencies.test.ts
ricokahler: it's technically allowed by the export map so that means it's technically possible for the user to import `react/package.json` in their code (or for a dependency to do it) so it would be a bug on us if we didn't support it.

binoy14 (approved):

PR #6751, opened 2024-05-23
fix(core): remove id from patch when using server actions

bjoerge:
### Description
When using server actions, we forwarded the patch mutations as-is, but that means they would also include the document id. The doucment ID is alredy provided as part of the actions payload, and should not be part of the action's patch here.

### What to review
The diff :)

### Testing
This was already covered by a test that I've updated.

### Notes for release

N/A

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -149,7 +149,9 @@ export default defineConfig([
     basePath: '/test',
     icon: SanityMonogram,
     // eslint-disable-next-line camelcase
-    __internal_serverDocumentActions: {},
+    __internal_serverDocumentActions: {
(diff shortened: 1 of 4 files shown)


dev/test-studio/sanity.config.ts
```diff
@@ -149,7 +149,9 @@ export default defineConfig([
     basePath: '/test',
     icon: SanityMonogram,
     // eslint-disable-next-line camelcase
-    __internal_serverDocumentActions: {},
+    __internal_serverDocumentActions: {
+      enabled: true,
```
bjoerge, line 153: wasn't aware we had turned it off, so turning it on again here

j33ty (approved):
Looks good.

PR #6770, opened 2024-05-24
fix(portable-text-editor): add autoresolving validations and fix normalization issues

skogsmaskin:
### Description

This will add two auto-resolving validations:

* Test a block for orphaned markDefs and remove obsolete ones (previously only done through editor normalization).
* Test a block for orphaned marks and remove obsolete ones (previously a hard validation error).

By introducing these new validations, I found two issues. The first was there before, and the other one occured introducing  the new normalization rules of markDefs. 

* Fix issue where deleting block 3, 2 and partially 1 where block 1 contains an remaining annotation (too many validation iterations error).
* Fix issue where copying an text fragment containing an annotation would not be valid with the new validation code.

### Testing

Should be covered by unit tests.

### Notes for release
Improve the validation of content in the Portable Text Editor.

### DIFF
--- a/packages/@sanity/portable-text-editor/src/editor/__tests__/PortableTextEditor.test.tsx
+++ b/packages/@sanity/portable-text-editor/src/editor/__tests__/PortableTextEditor.test.tsx
@@ -310,7 +310,7 @@ describe('initialization', () => {
         _key: 'abc',
         _type: 'myTestBlockType',
         markDefs: [],
-        children: [{_key: 'def', _type: 'span', text: 'Test', marks: ['invalid']}],
+        children: [{_key: 'def', _type: 'span', marks: []}],
       },
     ]
     const initialSelection: EditorSelection = {
@@ -333,15 +333,15 @@ describe('initialization', () => {
           type: 'invalidValue',
           value: initialValue,
           resolution: {
-            action: 'Remove invalid marks',
+            action: 'Write an empty text property to the object',
             description:
-              "Block with _key 'abc' contains marks (invalid) not supported by the current content model.",
+              "Child with _key 'def' in block with key 'abc' has missing or invalid text property!",
             i18n: {
-              action: 'inputs.portable-text.invalid-value.orphaned-marks.action',
-              description: 'inputs.portable-text.invalid-value.orphaned-marks.description',
+              action: 'inputs.portable-text.invalid-value.invalid-span-text.action',
+              description: 'inputs.portable-text.invalid-value.invalid-span-text.description',
               values: {
                 key: 'abc',
-                orphanedMarks: ['invalid'],
+                childKey: 'def',
               },
             },
             item: {
@@ -351,8 +351,7 @@ describe('initialization', () => {
                 {
                   _key: 'def',
                   _type: 'span',
-                  marks: ['invalid'],
-                  text: 'Test',
+                  marks: [],
                 },
               ],
               markDefs: [],
@@ -367,10 +366,14 @@ describe('initialization', () => {
                   {
                     _key: 'def',
                   },
-                  'marks',
                 ],
                 type: 'set',
-                value: [],
+                value: {
+                  _key: 'def',
+                  _type: 'span',
+                  marks: [],
+                  text: '',
+                },
               },
             ],
           },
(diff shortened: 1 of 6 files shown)


packages/@sanity/portable-text-editor/src/editor/__tests__/PortableTextEditor.test.tsx
```diff
@@ -310,7 +310,7 @@ describe('initialization', () => {
         _key: 'abc',
         _type: 'myTestBlockType',
         markDefs: [],
-        children: [{_key: 'def', _type: 'span', text: 'Test', marks: ['invalid']}],
+        children: [{_key: 'def', _type: 'span', marks: []}],
```
skogsmaskin, line 313: The previous test case is now auto-resolved, so we need another validation error for this test.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithInsertData.ts
```diff
@@ -321,41 +321,43 @@ function _insertFragment(
...
-    debug('Mixing markDefs of focusBlock and fragments[0] block', markDefs, fragment[0].markDefs)
-    if (!isEqual(markDefs, fragment[0].markDefs)) {
-      Transforms.setNodes(
-        editor,
-        {
-          markDefs: uniq([...(fragment[0].markDefs || []), ...(markDefs || [])]),
-        },
-        {at: focusPath, mode: 'lowest', voids: false},
-      )
+  editor.withoutNormalizing(() => {
```
skogsmaskin, line 324: This whole change must be run without normalizing, as we build up the data.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithInsertData.ts
```diff
@@ -142,7 +142,7 @@ export function createWithInsertData(
           // Validate the result
           const validation = validateValue(parsed, schemaTypes, keyGenerator)
           // Bail out if it's not valid
-          if (!validation.valid) {
+          if (!validation.valid && !validation.resolution?.autoResolve) {
```
skogsmaskin, line 145: Validation errors for pasted fragments that can be auto-resolved should pass.

packages/@sanity/portable-text-editor/src/utils/validateValue.ts
```diff
@@ -276,7 +283,7 @@ export function validateValue(
               i18n: {
                 description: 'inputs.portable-text.invalid-value.orphaned-marks.description',
                 action: 'inputs.portable-text.invalid-value.orphaned-marks.action',
-                values: {key: blk._key, orphanedMarks},
+                values: {key: blk._key, orphanedMarks: orphanedMarks.map((m) => m.toString())},
```
skogsmaskin, line 286: Values for i18n must be strings, so safeguard this map as strings.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextMarkModel.ts
```diff
@@ -202,14 +202,31 @@ export function createWithPortableTextMarkModel(
...
       }
-      // Check consistency of markDefs
+      // Check consistency of markDefs (unless we are merging two nodes)
       if (
-        isTextBlock &&
-        editor.operations.some((op) =>
-          ['split_node', 'remove_node', 'remove_text', 'merge_node'].includes(op.type),
+        editor.isTextBlock(node) &&
+        !editor.operations.some(
+          (op) => op.type === 'merge_node' && 'markDefs' in op.properties && op.path.length === 1,
```
skogsmaskin, line 209: Because of the new normalization rules, this must be rewritten (as discovered by tests).

Now we  do the markDefs normalization here for every operation that is NOT merging two text blocks, as this will be an intermediate state, where potentially new child nodes are added later in the change (next operation). We should do this normalization in the later operations instead.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithPortableTextMarkModel.ts
```diff
@@ -202,14 +202,31 @@ export function createWithPortableTextMarkModel(
...
         )
       ) {
-        normalizeMarkDefs(editor, types)
```
skogsmaskin, line 212: This is very pretty inefficient as it will do the whole editor, and not the actual nodes that are transformed.

packages/@sanity/portable-text-editor/src/editor/plugins/createWithEditableAPI.ts
```diff
@@ -517,7 +516,6 @@ export function createWithEditableAPI(
                   })
                 }
               })
-              normalizeMarkDefs(editor, types)
```
skogsmaskin, line 520: This is now done in the normalization inside `withPortableTextMarksModel`

RitaDias (approved):
This is great 👍 :shipit:

PR #6771, opened 2024-05-24
fix(structure): allow search/list ignored types when explicitly requested

rexxars:
### Description

We previously allowed users to list assets in the structure tool by using a `S.documentTypeList('sanity.imageAsset')`, but during search refactoring this seems to no longer be possible.

This reintroduces the ability, by ensuring that the `getSearchableTypes()` does not remove explicitly declared types.

### What to review

- In the test studio, observe that in `next`, the document list under `Custom panes > Images` does not list images, and that the preview deploy of this _does_
- Check that the ["all document types"](https://test-studio-git-fix-structure-explicit-allow-types.sanity.build/test/structure/custom;allDocumentTypes) custom pane in test studio does not list image and file asset types, as well as the global search not including these types.

### Testing

I attempted to test the actual `listenSearchQuery()` function, but it has a _lot_ of dependencies and was not easy to mock out a client for. An alternative would be to do an e2e test, but I would prefer us to have a more stable environment we can test against, to ensure we have assets that can be listed. I'll make a note of it for future iterations of the E2E suite.

For now, I added some tests for the `getSearchableTypes` function, which at least increases test coverage of _that_ part of the codebase. 

### Notes for release

- Fixes an issue where structure lists would not list image/file assets if told to do so

### DIFF
--- a/dev/test-studio/structure/resolveStructure.ts
+++ b/dev/test-studio/structure/resolveStructure.ts
@@ -377,6 +377,10 @@ export const structure: StructureResolver = (S, {schema, documentStore, i18n}) =
               S.divider(),
 
               S.documentTypeListItem('sanity.imageAsset').title('Images').icon(ImagesIcon),
+
+              S.divider(),
+
+              S.listItem().title('All document types').child(S.defaults()),
             ]),
         ),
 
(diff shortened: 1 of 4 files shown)


binoy14 (approved):
Makes sense to me, thanks!

PR #6774, opened 2024-05-27
Sdx 1358

bjoerge:
### Description
Currently for _primitive_ fields, you can prevent the field title from being rendered by passing `undefined` as the title to `renderDefault`, e.g:

```
{props.renderDefault({
  ...props,
  title: undefined,
})}
```
This, however, does not work for array or object fields, which will always render `Undefined`.

This PR addresses the discrepancy.


hermanwikner (approved):
Works as expected! 👍

PR #6776, opened 2024-05-28
fix(form): avoid spreading key prop

christianhg:
### Description

Before this change, React would throw a runtime error stating that `A props object containing a "key" prop is being spread into JSX`. Now, the `key` is passed to the JSX without using spread.

### What to review

To test this you can run `pnpm dev` to spin up the test studio, and then open a document that has an array of objects. Verify that the error is gone.

Screenshot 2024-05-28 at 13 00 03

### Testing

Not practical to write an automated test for since it's a mitigation of a React render error.


bjoerge (approved):
Thanks @christianhg!

PR #6777, opened 2024-05-28
feat(core): add title to breadcrumb dropdown

RitaDias:
### Description

Add array title to the array dropdown in breadcrumbs.
Kayla mentioned to drop the "keyboard" shortcuts for now.

### DIFF
--- a/packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/TreeEditingBreadcrumbsTitle.tsx
+++ b/packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/TreeEditingBreadcrumbsTitle.tsx
@@ -1,0 +1,24 @@
+import {Box, Card, Flex, MenuDivider, Text} from '@sanity/ui'
+
+interface TreeEditingBreadcrumbsTitleProps {
+  title: string
+}
+
+export function TreeEditingBreadcrumbsTitle(props: TreeEditingBreadcrumbsTitleProps): JSX.Element {
+  const {title} = props
+
+  return (
+    <>
+      <Box paddingX={1}>
+        <Card padding={3}>
+          <Flex align="center" justify="space-between" gap={2}>
+            <Text size={1} muted textOverflow="ellipsis">
+              {title}
+            </Text>
+          </Flex>
+        </Card>
+      </Box>
+      <MenuDivider />
+    </>
+  )
+}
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/buildArrayState.ts
```diff
@@ -49,8 +49,8 @@ export function buildArrayState(props: BuildArrayState): TreeEditingState {
         arrayValue,
         itemPath,
         parentPath: rootPath,
+        arrayTitle: getSchemaTypeTitle(arraySchemaType),
```
hermanwikner, line 52: I think it would be nice if this was handled in the `buildBreadcrumbsState` since we have access to the `arraySchemaType` in that function – something like this:

```diff
export function buildBreadcrumbsState(props: BuildBreadcrumbsStateProps): TreeEditingBreadcrumb {
  const {arraySchemaType, arrayValue, itemPath, parentPath} = props

+ const parentTitle = getSchemaTypeTitle(arraySchemaType)

  const items = arrayValue.map((arrayItem) => {
    const nestedItemPath = [...parentPath, {_key: arrayItem._key}] as Path

    const {title} = getArrayItemPreview({arrayItem, arraySchemaType})

    return {
      children: EMPTY_ARRAY,
+     parentTitle,
      path: nestedItemPath,
      title,
    }
  })

  const selectedItemTitle = items.find((item) => isEqual(item.path, itemPath))?.title

  return {
    children: items,
+   parentTitle,
    path: itemPath,
    title: selectedItemTitle as string,
  }
}
```

RitaDias: This is a good point,
the reason why I initially didn't even want to touch this place (and I'm still looking into a way to not do it) is because this makes it that _every_ item on the breadcrumb will have the parentTitle when in reality we would only need this in one place. 

If you look at the failing tests you'll see what I mean. It's fine to have this information (might even be useful for the future), but the way I looked at it it was like having a lot of unneeded information in the "children". What do you think?

packages/sanity/src/core/form/studio/tree-editing/components/breadcrumbs/TreeEditingBreadcrumbsTitle.tsx
```diff
@@ -0,0 +1,24 @@
...
+        <Card padding={3}>
+          <Flex align="center" justify="space-between" gap={2}>
+            <Text size={1} muted textOverflow="ellipsis">
+              {title}
+            </Text>
+          </Flex>
+        </Card>
+      </Box>
+      <MenuDivider />
+    </>
```
hermanwikner, line 22: Nitpick: this could be a bit simplified to this (I don't think the `Flex` is necessary in this case).

```ts
<Card borderBottom padding={3} sizing="border">
  <Box paddingX={1} sizing="border">
    <Text size={1} muted textOverflow="ellipsis">
      {title}
    </Text>
  </Box>
</Card>
```

hermanwikner (approved):
Thanks 🙏

PR #6779, opened 2024-05-28
feat(core): improve navigation in array tree editing 

hermanwikner:
### Description

This pull request updates the items in the tree editing sidebar so that the navigate and expand/collapse targets are two separate buttons. It also updates the logic to scroll to a field in the form by utilizing the `focusPath`.


RitaDias (approved):
Just a small nitpick but overall :shipit:

PR #6783, opened 2024-05-28
refactor(core): stabilize studio usage of actions API

bjoerge:


packages/sanity/src/core/store/_legacy/history/createHistoryStore.ts
```diff
@@ -217,23 +209,25 @@ function restore(
...
-          body: {
-            actions: [
-              {
-                actionType: 'sanity.action.document.replaceDraft',
-                publishedId: documentId,
-                attributes: restoredDraft,
-              },
-            ],
+        return client.observable.action([
+          {
```
cngonzalez, line 213: Including the "create" action seems to cause restoring to fail with "there is already a published document with that ID" (if the document has been published) or "there is already a draft document with that ID" (if it has not)

I have probably missed some conversations on its inclusion, so I don't feel comfortable removing it -- please let me know if I can help!

Screenshot 2024-05-28 at 4 41 36 PM

bjoerge: Great catch @cngonzalez! Looking into it.

bjoerge: Fixed by 614eb98a4debbb7c8315ca5c4015ed688614380a

juice49 (approved):

PR #6785, opened 2024-05-28
refactor(core): cache fetchFeatureToggle by client

bjoerge:
### Description

This adds caching to the actions feature toggle, so we avoid refetching when switching between documents, etc.

### Testing
In order to test this with the test-studio, this config flag has to be commented out here:
(note - it must be `undefined`, setting it to `false` will disable server actions entirely and skip fetching the flag from the backend):

https://github.com/sanity-io/sanity/blob/373af6a12047e16870fa03270a2452e822a37b55/dev/test-studio/sanity.config.ts#L153

### Notes for release
n/a


cngonzalez (approved):
This makes sense to me!

Part of the initial rationale for fetching this on every document was to ensure that users could quickly revert to the old behavior if needed. Obviously, since our memoization and store creation strategy makes it difficult to actually do this, we introduced the timeout. This PR closes the gap there from the legacy rationale 😅

PR #6808, opened 2024-05-30
chore: setup a separate compiled studio, bump to 19rc

stipsan:
- adds `dev:compiled-studio` that runs with React Compiler enabled.
- adds a https://test-compiled-studio.sanity.build where React Compiler is enabled.
- removes React Compiler from `dev:next-studio` and `dev:turbo-studio`, allowing us to test just React 19 RC, and Turbopack, separately.
- fixes a `styled-components` related warning:
  ```
  styled-components.browser.esm.js:33 It looks like there are several instances of 'styled-components' initialized in this application. This may cause dynamic styles to not render properly, errors during the rehydration process, a missing theme prop, and makes your application bigger without good reason.
  
  See https://s-c.sh/2BAXzed for more info.
  ```
- bumps react 19 from beta to rc.
- bumps next from 14.3 canary to 15 rc


bjoerge (approved):
Great stuff – thanks @stipsan!

PR #6813, opened 2024-05-30
fix: use `useImperativeHandle` instead of mutating a parent ref

stipsan:
### Description

React Compiler isn't able to optimize `<Pane>` when it mutates the `forwardedRef` value returned by `useForwardedRef`.
That's fine, because we can use `useImperativeHandle` to forward refs to the parent in a way that is sound, and that React Compiler is able to handle.
We should probably deprecate the `useForwardedRef` hook in `@sanity/ui` as `useImperativeHandle` is a much better alternative (`useForwardedRef` for example needs a `useEffect` cycle to handle forwarding):

```diff
-import {useForwardedRef} from '@sanity/ui'
-import {forwardRef} from 'react'
+import {forwardRef, useImperativeHandle} from 'react'

export default forwardRef(function Component(props, ref) {
-  const nodeRef = useForwardedRef(ref)
+  const nodeRef = useRef()
+  useImperativeHandle(ref, () => nodeRef.current)

  return <div ref={nodeRef} />
})
```

### What to review

Does it make sense?

### Testing

Existing tests should be sufficient, as the refactor doesn't have practical differences, beyond being slightly faster in general, and noticeably faster when React Compiler is enabled.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/structure/components/pane/Pane.tsx
+++ b/packages/sanity/src/structure/components/pane/Pane.tsx
@@ -1,11 +1,4 @@
-import {
-  BoundaryElementProvider,
-  Card,
-  type CardProps,
-  Code,
-  Flex,
-  useForwardedRef,
-} from '@sanity/ui'
+import {BoundaryElementProvider, Card, type CardProps, Code, Flex} from '@sanity/ui'
 import {
   type ForwardedRef,
   forwardRef,
@@ -13,6 +6,7 @@ import {
   type ReactNode,
   useCallback,
   useEffect,
+  useImperativeHandle,
   useMemo,
   useState,
 } from 'react'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/structure/components/pane/Pane.tsx
```diff
@@ -81,18 +75,12 @@ export const Pane = forwardRef(function Pane(
...
 
-  const setRef = useCallback(
-    (refValue: HTMLDivElement | null) => {
-      setRootElement(refValue)
-      forwardedRef.current = refValue
-    },
-    [forwardedRef],
-  )
+  // Forward ref to parent
+  useImperativeHandle(ref, () => rootElement!, [rootElement])
```
juice49, line 83: This is pedantic, but would it be more correct to type the imperative handle as `HTMLDivElement | null`, rather than using a non-null assertion? My understanding is that this asserts `rootElement` is never `null`, but in reality, it is.

```suggestion
  useImperativeHandle<HTMLDivElement | null, HTMLDivElement | null>(ref, () => rootElement, [
    rootElement,
  ])
```

juice49 (approved):
This looks good to me! I left one small comment, but it's not a blocker.

packages/sanity/src/structure/components/pane/Pane.tsx
stipsan: Yes, you're correct. The parent `ref: ForwardedRef<HTMLDivElement>` mandates it's always assigned, so I considered changing it as well so it's fully consistent. But it caused other side-effects to the type checks that felt out of scope. Do you want me to do it in a follow-up?

juice49: I wondered about that, too, but `ForwardedRef` appears to always produce a nullable result (which is correct for consumers of this component).

IMO it's probably not worth spending more time on 😃.

PR #6814, opened 2024-05-30
fix: call dynamic hooks in a way that can be compiled

stipsan:
### Description

Refactors `FieldAction` and `InspectorMenuItem` components so they're no longer calling hooks that come from props. Instead, the components themselves are dynamically defined, allowing React to call said underlying hooks in a dynamic way.
It upholds the guarantee that no matter how many times, say, `FieldAction` re-renders, it'll never suddenly see `props.action.useAction()` change identity, because if it changes a new component is created instead.

### What to review

It should make sense. There's another linter warning in `DocumentInspectorMenuItemsResolver.tsx` that will be handled in a follow up.

### Testing

### Notes for release

N/A


bjoerge (changes requested):
Great! Looks like there's some unit tests/snapshots that needs updating though

stipsan:
@SocketSecurity ignore-all

stipsan:
@SocketSecurity ignore-all

bjoerge (approved):

PR #6815, opened 2024-05-30
fix(core): apply create mutation only if draft doesn't exist locally

bjoerge:
### Description

Currently, if a document has a published version, we would apply a `createIfNotExists` mutation locally in order to create the draft. Later on, when converting buffered document mutations to actions, we'd ignore `createIfNotExists` mutations, which could occationally lead to an empty array of actions, which again would be rejected by the server side document actions with an error saying "No actions provided".

However, if there is no published document, we'll currently apply a `create` mutation locally, which will then be converted to a `sanity.action.document.create`-action.

This PR fixes the issue of an empty actions array by using the same logic for `create` mutations no matter if there's a local published version or not.

### What to review

Does the changes make sense? Anything I'm missing here?

### Testing

I'm expecting some failing unit tests as a consequence of this change – ~will update tests accordingly~ Done.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
@@ -29,6 +29,13 @@ export const patch: OperationImpl<[patches: any[], initialDocument?: Record<stri
 
     const patchMutation = draft.patch(patches)
 
+    if (snapshots.published && !snapshots.draft) {
+      // If there's a published version, but no draft version, editing the published version will
+      // ensure a draft version is created.
+      published.mutate(patchMutation)
+      return
+    }
+
     // If there's no draft, the user's edits will be based on the published document in the form in front of them
     // so before patching it we need to make sure it's created based on the current published version first.
     const ensureDraft = snapshots.draft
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
```diff
@@ -29,25 +29,14 @@ export const patch: OperationImpl<[patches: any[], initialDocument?: Record<stri
...
-        ...patchMutation,
-      ])
-      return
-    }
+    // If there's no draft, the user's edits will be based on the published document in the form in front of them
+    // so before patching it we need to make sure it's created based on the current published version first.
     const ensureDraft = snapshots.draft
       ? []
       : [
           draft.create({
```
juice49, line 37: There is a problem here. We cannot create a draft for an existing published document; we would instead need to edit the published document.

Because we removed the handling for when `snapshot.published` exists, we now often attempt to create a draft for an existing published document, causing Studio to crash with the following error:

```
Error: there is already a published document with that ID
```

I'm currently trying to reason about how we should proceed.

juice49:
@bjoerge I noticed commit [ebfbaa29876479d72531b68f6af8fea59d03c959](https://github.com/sanity-io/sanity/pull/6815/commits/ebfbaa29876479d72531b68f6af8fea59d03c959) is still in this branch. Is that intentional? I think it should probably be removed.

bjoerge:
I think we can close this now that the http api allows and empty actions array

PR #6816, opened 2024-05-30
feat(migrate): add `dryRun` to context

nkgentile:
### Description

When running a migration, there may be side-effects like calling a third-party API, or using client to make Sanity requests. A developer may wish to log those effects rather than actually executing them in a dry run.

```ts
import { defineMigration } from "sanity/migrate";

/**
 * Run this migration with
 * `npx sanity migration run my-migration`
 */
export default defineMigration({
  title: "my-migration",

  // pass whether the migration is a dry-run as context
  async *migrate(documents, { dryRun }) {
    for await (const document of documents()) {
      if(dryRun) {
        // Only log the effect
        console.log(`Something happened`)
      } else {
        // Execute the side-effect
        await thirdPartyRequest();
      }
      // ... do something with the document, maybe yield patches
    }
  },
});

```

### Testing

I manually tested that the flag comes through the context in both async generator and object definition formats

### Notes for release

Adds a `dryRun` flag to the context argument in migration scripts


bjoerge (approved):
Very useful – thanks @nkgentile!

PR #6819, opened 2024-05-30
fix(core): restore platform-aware keyboard shortcuts

rexxars:
### Description

Seems the facelift PR caused a regression that basically reverted [this PR](https://github.com/sanity-io/sanity/pull/5065) in terms of end result.

This PR reverts to using the `<Hotkeys>` component from the sanity core, which is platform aware.

### What to review

- Shows `Ctrl + Option + P` on publish and similar, when on Mac
- Shows `Ctrl + Alt + P` on publish and similar, when _not_ on Mac

(This should be uncovered by the test suite, fwiw)

### Testing

I added one test for this in the publish action. Ideally I would test all the different locations that use any kind of hotkeys that _may_ involve the `Alt` key, but that would be a lot of tests in a lot of different places. I may introduce an ESLint rule to disallow the use of `<Hotkeys>` from `@sanity/ui` later.

### Notes for release

- Fixes a regression causing platform-aware hotkeys not to be displayed in tooltips


juice49 (dismissed):
Looks good to me!

sjelfull (dismissed):

ricokahler (approved):

PR #6820, opened 2024-05-30
feat(cli): allow the ability to specify package manager in init command

binoy14:
### Description

Adds `--package-manager` flag to the init command. If passed and valid, it will use this flag as the package manager when installing dependencies. 

Note: I have not touched the nextjs init package manager choosing piece not sure if it should or not. 

### What to review

Changes and copy makes sense. You can still create new projects with and without this flag.

### Testing

I manually tested different scenarios, this PR will unblock future PRs for testing the init command

### Notes for release

- Adds `--package-manager` flag to init command to choose a package manager

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -21,6 +21,7 @@ import {
   installNewPackages,
 } from '../../packageManager'
 import {
+  ALLOWED_PACKAGE_MANAGERS,
   getPartialEnvWithNpmPath,
   type PackageManager,
 } from '../../packageManager/packageManagerChoice'
@@ -133,6 +134,7 @@ export default async function initSanity(
   const useGit = typeof commitMessage === 'undefined' ? true : Boolean(commitMessage)
   const bareOutput = cliFlags.bare
   const env = cliFlags.env
+  const packageManager = cliFlags['package-manager']
 
   let defaultConfig = cliFlags['dataset-default']
   let showDefaultConfigPrompt = !defaultConfig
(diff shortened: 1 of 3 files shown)


binoy14:
* **#6820** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@sanity/cli/src/commands/init/initCommand.ts
```diff
@@ -3,6 +3,7 @@ import {detectFrameworkRecord, LocalFileSystemDetector} from '@vercel/fs-detecto
 
 import initPlugin from '../../actions/init-plugin/initPlugin'
 import initProject from '../../actions/init-project/initProject'
+import {type PackageManager} from '../../packageManager/packageManagerChoice'
 import {type CliCommandDefinition} from '../../types'
```
rexxars, line 8:
```suggestion
import {ALLOWED_PACKAGE_MANAGERS, type PackageManager} from '../../packageManager/packageManagerChoice'
import {type CliCommandDefinition} from '../../types'

const allowedPms = ALLOWED_PACKAGE_MANAGERS.map(pm => `"${pm}"`).join(' | ')
```

Then use `${allowedPms}` in the help string, maybe?

Would help keep them in line.

rexxars (approved):
Looks good! Left a small suggestion, but feel free to ignore.

packages/@sanity/cli/src/commands/init/initCommand.ts
binoy14: good point, I can make the change

packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -542,15 +544,25 @@ export default async function initSanity(
...
-  const pkgManager = await getPackageManagerChoice(outputPath, {
-    prompt,
-    interactive: unattended ? false : isInteractive,
-  })
+  let pkgManager: PackageManager
+
+  // If the user has specified a package manager, and it's allowed use that
+  if (packageManager && ALLOWED_PACKAGE_MANAGERS.includes(packageManager)) {
+    pkgManager = packageManager
+  } else {
```
juice49, line 552: It looks like if the user specifies a disallowed package manager then we silently fall back to an allowed one. Should we warn the user?

binoy14: can, what would the copy look like?

juice49: Something like this?

```ts
output.print(`Given package manager "${packageManager}" isn't supported. Supported package managers are ${allowedPms}.`)
output.print(`Continuing using ${pkgManager}...`)
```

binoy14: Does this make sense? 

Screenshot 2024-05-31 at 10 12 26 AM

juice49: That looks good to me!

(However, we should consider supporting `dog`).

juice49 (approved):
Looks good to me!

packages/@sanity/cli/src/commands/init/initCommand.ts
```diff
@@ -22,6 +26,7 @@ Options
   --project-plan <name> Optionally select a plan for a new project
   --coupon <name> Optionally select a coupon for a new project (cannot be used with --project-plan)
   --no-typescript Do not use TypeScript for template files
+  --package-manager <manager> Specify which package manager to use [allowed: ${allowedPackageManagersString}]
```
juice49, line 29: Tiny: would `<name>` make more sense than `<manager>` here? It makes it consistent with the other flags.

juice49:
```suggestion
  --package-manager <name> Specify which package manager to use [allowed: ${allowedPackageManagersString}]
```

juice49 (approved):

PR #6821, opened 2024-05-31
test(sanity): update hydrate test

RitaDias:
### Description

Updated the studio.test.tsx after having it consistently failing on a separate PR and updating the snapshot not fixing it (the classes were consistently changing)
Spoke with Cody and in that case in particular it was  related to the style components and the order with which the components that use them are called.

The snapshot test was brittle since it was screaming that something was wrong when it reality it could just be different (and that was the intended result). The following test was written with Cody to remain useful and to target hydration test but without having to rely on snapshots.

### What to review

That the test makes sense and there isn't a better way of doing it.

### Notes for release

N/A (let me know if this is something that we want to appear on the release notes I just don't think we need to)

### DIFF
--- a/packages/sanity/src/core/studio/Studio.test.tsx
+++ b/packages/sanity/src/core/studio/Studio.test.tsx
@@ -18,6 +18,8 @@ import {describe, expect, it, jest} from '@jest/globals'
  * c) https://styled-components.com/docs/advanced#server-side-rendering
  */
 import {type SanityClient} from '@sanity/client'
+import {act} from 'react'
+import {hydrateRoot} from 'react-dom/client'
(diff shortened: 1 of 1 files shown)


stipsan (approved):
I slept on it and I think the reason I used the snapshots when I wrote these tests originally was to aid debugging: "what does it render?"
Since they keep failing the tests when class strings change I think they do more harm than good so it makes sense to get rid of them.
The hydrateRoot test is very useful though as it would catch issues that would otherwise only appear for Studios that are embedded in Next or Remix apps 🙌

binoy14 (approved):

binoy14 (approved):
[beautiful](https://app.graphite.dev//settings/meme-library?org=sanity-io)

PR #6826, opened 2024-05-31
refactor: remove non-null assertion

stipsan:
Follows up on https://github.com/sanity-io/sanity/pull/6813/files#r1622000741 as non-null assertions are indeed bad :)


stipsan:
Replaced by https://github.com/sanity-io/sanity/pull/6827

PR #6828, opened 2024-05-31
feat: move up call to PATCH metadata after bootstrapping template files

svirs:
### Description
Previously made api call is now moved up in case something else in the init flow does not let this call be made at the very end. This call is necessary to indicate that the bootstrapped files are now made local.

### What to review
Project init flow should act the same, the call should occur after the studio files are created.


rexxars (dismissed):
👍

rexxars (approved):

PR #6831, opened 2024-05-31
feat(cli): enable autoUpdates by default on new projects

binoy14:
### Description

This enables autoUpdates flag by default for all new studios

### What to review

The change makes sense, I don't believe this will be case reading the code but it should not add it for embedded studio's.

Ways to test

1. In the monorepo, run `pnpm build` at the root
2. Navigate to another folder `dev` for instance (just to make it easy deleting the changes but not a requirement)
3. run `npx sanity init` follow the prompts
4. Check if `autoUpdates: true` is in the `sanity.cli.ts` 

### Testing

I have added some tests to cover init command, in general it requires more thorough testing but this should at least test all templates that it sets the flag correctly.

I would be way more comfortable if I am able to test init inside a nextjs application but the setup and time investment is much greater. It would have to be a follow up PR if anything

### Notes for release

(Double check if needs to be mentioned)
- Enables auto updates in studio for new projects

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
+++ b/packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
@@ -9,7 +9,12 @@ export default defineCliConfig({
   api: {
     projectId: '%projectId%',
     dataset: '%dataset%'
-  }
+  },
+  /**
+   * Enable auto-updates for studios.
+   * Learn more at https://www.sanity.io/docs/cli#auto-updates
+   */
+  autoUpdates: true,
 })
 `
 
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
```diff
@@ -9,7 +9,12 @@ export default defineCliConfig({
   api: {
     projectId: '%projectId%',
     dataset: '%dataset%'
-  }
+  },
+  /**
+   * Enable auto-updates for studios.
+   * Learn more at https://www.sanity.io/docs/cli#auto-updates
```
binoy14, line 15: TODO: get the correct URL

binoy14:
* **#6831** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6831?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

PR #6844, opened 2024-06-04
refactor(core): abstract `getSchemaTypeIcon`

christianhg:
### Description

I need the same logic in a new component in a subsequent PR so I'm abstracting it beforehand. `||` and `&&` have been changed to `??` and `?:` to avoid returning `| false` from the function. Now, it just returns `ComponentType | undefined`.

### What to review

Do you like the function name and placement in the repo?

### Testing

I verified in the test studio that array insert menu icons still show up: 
Screenshot 2024-06-04 at 09 16 43

### DIFF
--- a/packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/ArrayOfObjectsFunctions.tsx
+++ b/packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/ArrayOfObjectsFunctions.tsx
@@ -1,5 +1,5 @@
 import {AddIcon} from '@sanity/icons'
-import {type ArraySchemaType, isReferenceSchemaType} from '@sanity/types'
+import {type ArraySchemaType} from '@sanity/types'
 import {Grid, Menu} from '@sanity/ui'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/ArrayOfObjectsFunctions.tsx
```diff
@@ -93,19 +94,12 @@ export function ArrayOfObjectsFunctions<
...
-                // Use reference icon if reference is to one schemaType only
-                const referenceIcon =
-                  isReferenceSchemaType(memberDef) &&
-                  (memberDef.to || []).length === 1 &&
-                  memberDef.to[0].icon
-
-                const icon = memberDef.icon || memberDef.type?.icon || referenceIcon
                 return (
                   <MenuItem
                     key={i}
```
christianhg, line 99: I notice that we could use the `memberDef.type` here instead of the array index. Maybe a worthy small refactor to do in the same PR?

bjoerge (approved):

PR #6855, opened 2024-06-05
Fixed PortableTextEditor pasting not working in firefox

### LINKED ISSUE #6715: Cannot Paste Into Block Content in Mozilla
Attempting to paste any text into a block content field does not work. Nothing happens, and the console displays the following message: `DataTransfer with kind="string" is currently not supported`

Pasting anywhere else, for example into a simple string field, still works as expected.

This has been tested on both **Windows** and **MacOS**, with **npm** and **bun**, both locally and deployed to [sanity.io](http://sanity.io) and using **Firefox** on both systems. Therefore I would assume that it is a bug within the studio, maybe in combination with Firefox.

* Node 22.2.0
* npm 10.8.0
* bun 1.1.8

```
@sanity/cli (global)          3.42.1 (up to date)
@sanity/eslint-config-studio   4.0.0 (up to date)
@sanity/vision                3.42.1 (up to date)
sanity                        3.42.1 (up to date)
```

mariusGundersen:
### Description
Pasting in the PortableTextEditor does not work in Firefox because it waits for a promise, after which the clipboardData is empty. The promise isn't needed since the returned value is undefined. This fix checks if the returned value is undefined before wrapping it in a promise, which makes pasting work again. 

### What to review

I don't know if there are any places where onPaste is set to do anything, if there is then this needs to be tested. I assume that anywhere onPaste is used and where it does something asyncronous pasting won't work in Firefox and maybe not in Chrome either, even before this change.

### Testing

Testing should be straight forward: try pasting something in the PortableTextEditor

### Notes for release

Fixed pasting into portable text editor in Firefox

This fixes #6715


christianhg:
First, I want to say thank you so much for this, @mariusGundersen.

Second, I'm in the process of moving the Portable Text Editor to a new repo, https://github.com/portabletext/editor so I'll close this PR for now. You are welcome to open it again against the new repo, but I also think, I've managed to narrow down the problem a bit further. So I might want to take an alternate approach at solving the problem. We'll see!

Again, thanks a lot for this! It's important that we are on top of issues like this so the fact that you are bringing it to our attention and even proposing a fix is more than we can ask for.

christianhg:
> First, I want to say thank you so much for this, @mariusGundersen.
> 
> Second, I'm in the process of moving the Portable Text Editor to a new repo, https://github.com/portabletext/editor so I'll close this PR for now. You are welcome to open it again against the new repo, but **I also think, I've managed to narrow down the problem a bit further. So I might want to take an alternate approach at solving the problem. We'll see!**
> 
> Again, thanks a lot for this! It's important that we are on top of issues like this so the fact that you are bringing it to our attention and even proposing a fix is more than we can ask for.

Apologies, I thought I had another take on the problem, but ended up implementing your fix: https://github.com/portabletext/editor/pull/28

mariusGundersen:
Good to hear. Do you have any timeline for when this will be released?

christianhg:
> Good to hear. Do you have any timeline for when this will be released?

I hope #7001 gets merged this week which means that next week's release of the Studio would include this bugfix. I think you could subscribe to #7001 to get a feel for the progress of things.

PR #6856, opened 2024-06-05
feat(core): global copy paste

sjelfull:
### Description

This PR introduces global copy & pasting of fields and documents. You can copy and paste either through Ctrl-C/Ctrl-V when focused on a field, or by using the Field Action / Document Context Action menu (the one in the right corner, where Inspect and Review Changes.

https://github.com/sanity-io/sanity/assets/10508/93bf509a-1570-41cc-a144-6ec1220d2eea

We have included a global keydown event listener that handles copy and paste events. They will not trigger if any of these is true:
- focus is inside an input/textarea/contenteditable element - this to not interfere with native functionality / undo stack
- you have an active selection (checking `window.getSelection()`)
- you haven't focused any inputs (the focus path is empty, aka `[]`)

**Keyboard shortcuts**
The keyboard shortcuts is limited to these fields:
- Arrays
- Objects either on level 1 or deeper relative to the root - the root being either the document or inside a array.
- Reference and Cross dataset reference previews

**Field actions**
Using fields actions you can copy and paste from any field that has a field action menu available. This currently excludes:
- Booleans - they lack a FormField wrapper and field actions

### What to review

- Are we happy with the global event listener on Ctrl-C & Ctrl-V? We have added exceptions so that it should not interfere with any known scenarios

### Testing

- We have added unit tests for the utilities that resolves schema types based on the path you want to copy or paste, as well as validates source and target schema compatibility
- We have also added E2E component tests, but there seems to be flake in these and other tests currently, so we can't trust them completely. We won't have time to resolve these until after the holidays, but we have done manual testing to make sure these are as solid as possible
- There is link to the testing manuscript in the project channel where we list all the combinations of source -> target we have tried

### Notes for release

- Added copy and paste for individual fields in deeply nested structures


packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
```diff
@@ -0,0 +1,333 @@
...
+
+        if (!targetDocumentSchemaType) {
+          toast.push({
+            status: 'error',
+            title: t('copy-paste.on-paste.validation.schema-type-incompatible.title', {
+              path: PathUtils.toString(targetPath),
+            }),
+          })
+          return
+        }
```
ricokahler, line 218: was testing this in the studio and noticed that it can be pretty easy to try to paste a value into a field that doesn't have the same shape. this is more of a product suggestion more than anything (cc @runeb) but i would prefer the button to be disabled (maybe with a toolip explaining why it's disabled) instead of seeing that i couldn't paste

packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
```diff
@@ -0,0 +1,333 @@
...
+      const clipboardItem = await getClipboardItem()
+
+      // Return early if no clipboard item or if clipboard item is invalid
+      if (!clipboardItem) {
+        toast.push({
+          status: 'info',
+          title: t('copy-paste.on-paste.validation.clipboard-empty.title'),
+        })
+        return
+      }
```
ricokahler, line 170: similar comment to the one i left below, i think i would prefer the `paste field` button to be disabled or just not appear if there is nothing to paste instead of seeing this toast

packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
```diff
@@ -0,0 +1,333 @@
...
+          return
+        }
+
+        const targetSchemaTypeTitle = targetSchemaType.title || targetSchemaType.name
+        const transferValueOptions = {
+          sourceRootSchemaType: sourceSchemaType,
+          sourcePath: [],
+          sourceValue: item.value,
+          targetRootSchemaType: targetSchemaType,
+          targetPath: [],
```
ricokahler, line 226: why are these hardcoded to empty array? are they being used or can they be removed?

packages/sanity/src/core/studio/copyPaste/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+  return selection !== null && !selection.isCollapsed
+}
+
+/** @internal */
+export function isEmptyFocusPath(path: Path): boolean {
+  return path.length === 0 || (path.length === 1 && path[0] === '')
+}
+
+function isWebKit(): boolean {
+  return typeof document !== 'undefined' && 'WebkitAppearance' in document.documentElement.style
```
ricokahler, line 120: I'm not sure if this was intended but I tried this out in chrome and this function is returning true when i think chrome supports the advanced clipboard copy/paste

packages/sanity/src/core/studio/copyPaste/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+      .map((key) =>
+        key.startsWith('_')
+          ? ''
+          : transformValueToText(typeof value === 'object' ? objectValue[key] : ''),
+      )
+      .filter(Boolean)
+      .join(', ')
+  }
+  return ''
+}
```
ricokahler, line 76: this function is pretty cool! as far as I understand it takes in a sanity JSON field value and converts it to text for pasting anywhere else? that's awesome

this doesn't have to be a now thing but for later I think we should think about how these values paste into other applications (e.g. google sheets, notion) and optimize its output for that. for example if we joined arrays with new lines instead of commas, it would paste into google sheets as different rows

cc @runeb for future thoughts (potentially outside of the scope of this PR tbh)

packages/sanity/src/core/studio/copyPaste/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+export const getClipboardItem = async (): Promise<CopyActionResult | null> => {
+  const items = await window.navigator.clipboard.read()
+  const item = items.find((i) => i.types.includes(SANITY_CLIPBOARD_ITEM_TYPE))
+  const sanityItem = await item?.getType(SANITY_CLIPBOARD_ITEM_TYPE)
+  if (sanityItem) {
+    const text = await sanityItem.text()
+    return parseCopyResult(text) || null
+  }
+  return null
+}
```
ricokahler, line 19: future product idea (outside of the scope of this PR): would be cool if this function could also accommodate pasting from other apps like google sheets. e.g. copying multiple lines from google sheets into an array field

cc @runeb

ricokahler (commented):
this is a incredible new feature. amazing work (and amazing job with all the tests too!)

i didn't get to review all of it but i did review the core chunks. I was testing it out in chrome and i noticed that copying a field resulting in a stringified copyResult object instead of the result from `transformValueToText`.

I also asked some questions regarding a few interfaces and if they could be simplifed. let me know what you think!

packages/sanity/src/core/studio/copyPaste/utils.ts
```diff
@@ -3,84 +3,161 @@ import {isString} from 'sanity'
...
+
+  try {
+    const el = doc.querySelector(`[data-${BASE64_ATTR}]`) as HTMLParagraphElement
+    if (!el) return null
+
+    type CamelCase<S extends string> = S extends `${infer P1}-${infer P2}${infer P3}`
+      ? `${P1}${Capitalize<P2>}${CamelCase<P3>}`
+      : S
+
+    const {sanityClipboardBase64} = el.dataset as Record<CamelCase<typeof BASE64_ATTR>, string>
```
cngonzalez, line 106: Apologies if I'm missing something, but where is "dataset" being set? I'm assuming this is a reference to the <p> tag being injected on line 69 but I'm not quite putting it all together.

packages/sanity/src/core/hooks/useGlobalCopyPasteElementHandler.ts
```diff
@@ -1,8 +1,9 @@
 import {type Path} from '@sanity/types'
 import {isHotkey} from 'is-hotkey-esm'
 import {useCallback, useEffect, useRef} from 'react'
-import {type FormDocumentValue, useCopyPasteAction} from 'sanity'
+import {type FormDocumentValue} from 'sanity'
```
cngonzalez, line 4: I'm noticing a few imports from `sanity` here. Do we prefer to import from local sources, e.g., `/core/form/types`?

packages/sanity/src/core/studio/copyPaste/CopyPasteProvider.tsx
```diff
@@ -32,14 +63,285 @@ export const CopyPasteProvider: React.FC<{
...
+
+          return
+        }
+
+        const isSingleField = updateItems.length === 1
+
+        if (isSingleField) {
+          toast.push({
+            status: 'success',
+            title: t('copy-paste.on-paste.validation.field_one-paste-success.title', {
```
cngonzalez, line 297: super nit: is the snake case here a typo?

packages/sanity/src/core/studio/copyPaste/CopyPasteProvider.tsx
```diff
@@ -157,10 +136,13 @@ export const CopyPasteProvider: React.FC<{
...
 
   const onPaste = useCallback(
-    async (targetPath: Path, value: FormDocumentValue | undefined, options?: PasteOptions) => {
-      const {schemaType: targetDocumentSchemaType} = documentMeta || {}
+    async (targetPath: Path, value: FormDocumentValue | undefined, options: PasteOptions) => {
+      // guard against `documentMeta` having not been set yet
+      if (!documentMeta) return
+
+      const {schemaType: targetDocumentSchemaType, onChange} = documentMeta
       const targetSchemaType = resolveSchemaTypeForPath(
```
cngonzalez, line 144: Should the toast and early return that guards against `targetDocumentSchemaType` happen _before_ this function is called?

packages/sanity/src/core/hooks/useGlobalCopyPasteElementHandler.ts
ricokahler: ooo great find

packages/sanity/src/core/studio/copyPaste/types.ts
```diff
@@ -0,0 +1,66 @@
...
+  isDocument: boolean
+  isObject: boolean
+  items: [
+    {
+      documentPath: Path
+      schemaTypeName: string
+      schemaTypeTitle?: string
+      value: unknown
+    },
+  ]
```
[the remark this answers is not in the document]
ricokahler: I've simplified this interface. I don't think we'll ever have to support more than one item in the clipboard. Currently in chrome if you try to write more than one `ClipboardItem`, it throws an error.

packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
```diff
@@ -0,0 +1,333 @@
...
+  const telemetry = useTelemetry()
+  const schema = useSchema()
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
+  const toast = useToast()
+  const {t} = useTranslation('copy-paste')
+
+  const {documentMeta, setCopyResult} = useCopyPaste()
+  const {onChange} = documentMeta || {}
+
+  const onCopy = useCallback(
```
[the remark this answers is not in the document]
ricokahler: I did lift these into the provider. It allowed for a good clean up.

packages/sanity/src/core/studio/copyPaste/utils.ts
ricokahler: The `dataset` property is populated from the `DOMParser` from the [`data-` attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes#javascript_access).

packages/sanity/src/core/studio/copyPaste/CopyPasteProvider.tsx
ricokahler: This is a convention in i18next to allow for [singular/plural combinations](https://www.i18next.com/translation-function/plurals#singular-plural). I mostly copied these over from the previous place these were.

packages/sanity/src/core/studio/copyPaste/types.ts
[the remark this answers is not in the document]
sjelfull: This was intended. The long term plan for this feature is to support bulk copy & paste. Note that `items` here doesn't map to multiple `ClipboardItem`s.

cngonzalez (dismissed):

packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
```diff
@@ -0,0 +1,333 @@
...
+
+        if (!onChange) {
+          console.warn(`Failed to resolve document onChange method when pasting.`)
+
+          toast.push({
+            status: 'error',
+            title: t('copy-paste.on-copy.validation.document-metadata-unknown-error.title'),
+          })
+          return
+        }
```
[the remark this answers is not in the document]
sjelfull: This happened earlier because of a race condition where the callback sometime wouldn't update because the `documentMeta` was coming from a dependency that itself wouldn't trigger a rerender. I believe it is stable now.

packages/sanity/src/core/studio/copyPaste/useCopyPasteAction.ts
sjelfull: We discussed this but landed on not wanting these be disabled, but rather give the user feedback.
In general i'd say its bad UX wise to disable things without giving enough context to the user. We also didn't find a performant way to disable field actions. Last point, you can't actually read the clipboard outside a user action to check if you have something in clipboard, so don't think there is good options there at all.

cngonzalez (approved):
Great work, all!!

PR #6534, opened 2024-04-30
feat(form/inputs): add support for image drop+paste in PTE input

sjelfull:
### Description

This adds support for drag-n-drop + pasting of images and files into the Portable Text Input inside the studio.

### Testing

1. Open [this test document](https://test-studio-git-feat-edx-1176-image-clipboard-support.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;757b15ed-d091-464d-9422-577956b5d293) in the test studio
2. Try to test the different scenarios on the Text field:
  - Drag 1 image
  - Drag 2 images
  - Drag a invalid file (something not PDF)
  - Drag 1 PDF file
  - Drag 2 PDF files
3. Now do the same, but pasting instead.

### Notes about PR
- Paste component test is skipped in Firefox because the security model doesn't allow us to make a synthetic test

### Notes for release

- The Portable Text Input now supports drag-n-dropping + pasting of images and files into `Image` and `File` blocks

PR #6867, opened 2024-06-06
fix(core): fix dragging existing blocks in pte

sjelfull:
### Description

This fixes an issue preventing blocks in the PTE from being re-arranged after we added support for drag-n-drop of images/files to the PTE in #6534.

It also fixes some adjacent issues:
- Styling of info/error overlay when dragging files was not styled correctly in fullscreen mode (fixes EDX-1408, fixes EDX-1405)

### What to review

- That the logic change makes sense
- That dragging PTE blocks works without showing any error messages
- That both drag-n-drop and pasting of images and files works

### Testing

Playwright tests for this has been added.

### Notes for release

- Fixes an issue when dragging PTE blocks
- Fixes an visual regression in PTE fullscreen mode where the _Drop to upload n files_ message would be position incorrectly

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
@@ -22,6 +22,7 @@ import {
   type RenderCustomMarkers,
 } from '../../types'
 import {type RenderBlockActionsCallback} from '../../types/_transitional'
+import {UploadTargetCard} from '../arrays/common/UploadTargetCard'
 import {ExpandedLayer, Root} from './Compositor.styles'
 import {Editor} from './Editor'
 import {useHotkeys} from './hooks/useHotKeys'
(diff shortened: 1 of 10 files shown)


packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
```diff
@@ -395,34 +396,43 @@ export function Compositor(props: Omit<InputProps, 'schemaType' | 'arrayFunction
...
-        renderEditable={renderEditable}
-        setPortalElement={setPortalElement}
-        scrollElement={scrollElement}
-        setScrollElement={setScrollElement}
-      />
+      <UploadTargetCard
+        types={editor.schemaTypes.portableText.of}
+        resolveUploader={props.resolveUploader}
+        onUpload={props.onUpload}
+        {...props.elementProps}
```
skogsmaskin, line 403: Do we need to set the elementProps here?

packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
```diff
@@ -437,6 +447,9 @@ export function Compositor(props: Omit<InputProps, 'schemaType' | 'arrayFunction
       onItemOpen,
       onPaste,
       path,
+      props.elementProps,
+      props.onUpload,
+      props.resolveUploader,
       rangeDecorations,
```
skogsmaskin, line 453: Nitpick, could we destruct these as well, as there's a bit of inconsistency in how they are listed with the other deps?

sjelfull: Updated.

packages/sanity/src/core/form/inputs/PortableText/Compositor.tsx
sjelfull: They are passed like this down to whatever component the `fileTarget` HOC is wrapping. Not sure if this is needed, but i'd rather not touch it since we do this multiple places.

sjelfull: Removed.

skogsmaskin (approved):
Great to have this fixed. Thanks @sjelfull  👍

PR #6869, opened 2024-06-06
feat: add error boundary for form input components

stipsan:
### Description

While debugging issues in React 19, as well as React Compiler, I've frequently encountered crashes [that takes down the Structure Tool completely](https://test-compiled-studio-b8mfdmi85.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;7455ba88-75d5-4a77-94a4-91738fa670c5):
image

It's difficult to track down exactly what's causing it, the same is true when handling user reports of crashes, in Next v14 and v15 embedded studios in particular.
This PR adds an error boundary for input components, [which narrows down the crash to the input level](https://test-compiled-studio-git-add-input-error-boundary.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;7455ba88-75d5-4a77-94a4-91738fa670c5):
image

In the image above only some of the Portable Text Editor inputs are crashing, significantly helping with narrowing down the cause. Fixes for the errors seen here will be done in follow up PRs.

### What to review

Does the language strings and the UI make sense?

### Testing

It's easiest to test on the compiled studio deployment, as it crashes early on race condition issues, instead of requiring tricky race condition reproduction steps: https://test-compiled-studio-git-add-input-error-boundary.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;7455ba88-75d5-4a77-94a4-91738fa670c5

### Notes for release

Form input components now have their own error boundaries, limiting crashes to a field level on a document form, instead of taking down the entire Structure Tool.

### DIFF
--- a/packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
+++ b/packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
@@ -1,16 +1,12 @@
-import {RefreshIcon} from '@sanity/icons'
-import {Card, Code, ErrorBoundary, Stack, Text} from '@sanity/ui'
+import {Box, Card, Code, ErrorBoundary, Stack, Text} from '@sanity/ui'
 import {useCallback, useMemo, useState} from 'react'
 import {useHotModuleReload} from 'use-hot-module-reload'
 
-import {Button} from '../../../ui-components'
 import {SchemaError} from '../../config'
-import {isDev} from '../../environment'
 import {useTranslation} from '../../i18n'
 import {CorsOriginError} from '../../store'
 import {isRecord} from '../../util'
 import {Alert} from '../components/Alert'
-import {Details} from '../components/Details'
 
 /**
  * @internal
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return <ErrorCard error={error} onRetry={handleRetry} />
+}
+
+/**
+ * The error UI is in a separate component to keep the wrapper lightweight,
+ * it allows us to use hooks like useTranslation without incurring the cost of it on every form input
+ * when there are no errors.
+ * @internal
+ */
+function ErrorCard(props: {error: unknown; onRetry: () => void}) {
```
cngonzalez, line 41: The unknown type here and the `stack` and `message` params become a little confusing for me here, but I understand if it's more difficult to more strictly type things here.

stipsan: It follows the precedence in `StudioErrorBoundary`, as well as the convention of the stricter TS rule that enforces:
```ts
try {
  doSomething()
} catch(error: unknown) {
 // `error` is never guaranteed to be a subclass of `Error` and can be anything, `unknown` forces you to deal with every eventuality
}
```

Which is also why the `isRecord` asserters are used further down

cngonzalez: ahh, of course! apologies for not checking that

PR #6870, opened 2024-06-07
fix: portable text editor crash

stipsan:
### Description

On React 19 the Portable Text Editor sometimes ends up in an infinite render loop when there's no value yet, as demonstrated here: https://test-compiled-studio-b8mfdmi85.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;7455ba88-75d5-4a77-94a4-91738fa670c5

This PR fixes it: https://test-compiled-studio-git-fix-portable-text-editor-crash.sanity.build/test/structure/input-standard;portable-text;pt_allTheBellsAndWhistles;7455ba88-75d5-4a77-94a4-91738fa670c5

The PR also contains a lot of other minor fixes related to React Compiler, improvements to how the color scheme state is passed around, adding the React Component stack to the form input error boundary etc.

### What to review

There shouldn't be any infinite render recursion loops happening in React 19 anymore for Portable Text Editor, try and break it 🙌 

### Testing

Beyond manual testing the existing e2e testing suite should be enough.

### Notes for release

Fixed React 19 causing Portable Text Editor to crash in some scenarios.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/hooks/usePortableTextMembers.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/hooks/usePortableTextMembers.tsx
@@ -1,5 +1,5 @@
 import {isEqual, pathFor} from '@sanity/util/paths'
-import {createRef, type MutableRefObject, type ReactNode, useContext, useRef} from 'react'
+import {createRef, type MutableRefObject, type ReactNode, useContext, useMemo, useRef} from 'react'
 import {type FormPatch, type Path, set} from 'sanity'
 import {PortableTextMemberItemsContext} from 'sanity/_singletons'
 
@@ -38,23 +38,6 @@ export function usePortableTextMemberItems(): PortableTextMemberItem[] {
 export function usePortableTextMemberItemsFromProps(
   props: PortableTextInputProps,
 ): PortableTextMemberItem[] {
-  const portableTextMemberItemsRef: MutableRefObject<PortableTextMemberItem[]> = useRef([])
-  reconcilePortableTextMembers({props, ref: portableTextMemberItemsRef})
-  return portableTextMemberItemsRef.current
-}
-
-const reconcilePortableTextMembers = ({
-  props,
-  ref,
-}: {
-  props: PortableTextInputProps
-  ref: MutableRefObject<PortableTextMemberItem[]>
-}) => {
-  const result: {
-    kind: PortableTextMemberItem['kind']
-    member: ArrayOfObjectsItemMember
-    node: ObjectFormNode
-  }[] = []
   const {
     members,
     path,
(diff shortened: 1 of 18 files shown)


packages/sanity/src/core/studio/colorScheme.tsx
```diff
@@ -180,6 +182,12 @@ export function useColorSchemeValue(): ThemeColorSchemeKey {
  * @internal
  */
 export function useColorScheme() {
+  useEffect(() => {
+    console.warn(
+      'useColorScheme() is deprecated, use useColorSchemeValue() or useColorSchemeSetValue() instead',
+    )
+  }, [])
```
stipsan, line 189: The `@deprecated` TSDoc haven't been sufficient to discourage using this hook, which is why it makes sense to `console.warn` in `useEffect` to push for possible userland to refactor away 🙌

christianhg:
> The PR also contains a lot of other minor fixes related to React Compiler, improvements to how the color scheme state is passed around, adding the React Component stack to the form input error boundary etc.

I'm curious, what was the actual culprit and what change fixed the issue?

packages/sanity/src/core/studio/colorScheme.tsx
```diff
@@ -155,22 +155,24 @@ export function ColorSchemeCustomProvider({
...
+export function useColorSchemeSetValue():
+  | false
+  | ((nextScheme: StudioThemeColorSchemeKey) => void) {
   const setValue = useContext(ColorSchemeSetValueContext)
   if (setValue === null) throw new Error('Could not find `ColorSchemeSetValueContext` context')
   return setValue
 }
 
 /** @internal */
-export function _useColorSchemeInternalValue(): StudioThemeColorSchemeKey {
```
ricokahler, line 165: do you think this hook needs to be exported? would refactoring it to not be exported be better? it seems like it was prefaced with an _ to prevent external usage but not exposing it does the same job

packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
```diff
@@ -38,8 +42,8 @@ export function FormBuilderInputErrorBoundary(
  * when there are no errors.
  * @internal
  */
-function ErrorCard(props: {error: unknown; onRetry: () => void}) {
-  const {error, onRetry} = props
+function ErrorCard(props: {error: unknown; info?: React.ErrorInfo; onRetry: () => void}) {
```
ricokahler, line 45: just double checking but is React.ErrorInfo available in the react 18 types as well?

packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
```diff
@@ -22,14 +23,17 @@ export function FormBuilderInputErrorBoundary(
   props: FormBuilderInputErrorBoundaryProps,
 ): JSX.Element {
   const {children} = props
-  const [{error}, setError] = useState<{error: unknown}>({error: null})
-  const handleRetry = useCallback(() => setError({error: null}), [])
+  const [{error, info}, setError] = useState<{error: unknown; info: React.ErrorInfo}>({
```
ricokahler, line 26: stylistic comment: since you made info optional in the ErrorCard, would it be better for `info` here to be optional as well?

packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
```diff
@@ -64,16 +72,38 @@ function ErrorCard(props: {error: unknown; onRetry: () => void}) {
...
+                {callStack && (
+                  <Code size={1} style={{maxHeight: '40vh'}}>
+                    {callStack}
+                  </Code>
+                )}
+              </Card>
+            </Stack>
+          </Box>
+        )}
+        {isDev && componentStack && (
```
ricokahler, line 91: really nice dev quality of life improvement here 💖

packages/sanity/src/core/form/inputs/PortableText/hooks/usePortableTextMembers.tsx
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable no-console */
```
ricokahler: why was this added?

packages/sanity/src/core/form/inputs/PortableText/hooks/usePortableTextMembers.tsx
ricokahler, line 1: nice refactor!

ricokahler (commented):
amazing work. the ` <Code> ` change i’m curious about. lmk what u think!

stipsan:
> > The PR also contains a lot of other minor fixes related to React Compiler, improvements to how the color scheme state is passed around, adding the React Component stack to the form input error boundary etc.
> 
> 
> 
> I'm curious, what was the actual culprit and what change fixed the issue?

Responded to this in depth on slack, the TL;DR is that writing to the ref in `usePortableTextMemberItemsForProps` during every render makes the return value of the hook, which is used in a long trail of hooks and components side effects as a dependency, unstable.
In React 18 it makes PTE slower than it needs to be.
In React 19 Strict Mode it can create infinite render loops and crash, but it's difficult to consistently reproduce as it's caused by race conditions and multiple other side effects happening at the same time.
In React 19 with React Compiler it's causing a very consistent crash since the way parent components are optimized cause certain deps to change on every render, causing side effects that end up scheduling another render, to fire and round and round we go.

packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
stipsan: Yup!

packages/sanity/src/core/form/studio/FormBuilderInputErrorBoundary.tsx
stipsan: The typings from `ErrorCard` in `@sanity/ui` claims that it's always provided, so I follow that assumption here to ensure default values and reset values fulfill that contract.

It's optional in the component as a defensive measure, to ensure that should the component in `@sanity/ui` somehow return `null` or `undefined` then the ErrorCard makes sure to account for it. I don't think it's likely to happen, but given that this component is supposed to handle unhandled runtime exceptions it felt reasonable to follow Murphy's Law ☺️

packages/sanity/src/core/studio/colorScheme.tsx
stipsan: I'd be happy to do that in a quick follow-up! It's only needed internally so it doesn't need to be available on `import 'sanity'`

stipsan: Btw the underscore caused the React Compiler to not see it as a hook, causing hook re-ordering errors, which is why it was changed here 🙌

packages/sanity/src/core/studio/colorScheme.tsx
stipsan: Yeah, I'm in favor of only logging once when it's not actionable. I'm this case the migration is very easy to do and I'd like to encourage it as much as possible 😬

packages/sanity/src/core/studio/colorScheme.tsx
christianhg: Good to know 💡

packages/sanity/src/structure/panes/document/DocumentPane.tsx
```diff
@@ -115,6 +115,7 @@ function DocumentPaneInner(props: DocumentPaneProviderProps) {
               t={t}
               i18nKey="panes.document-pane.document-not-found.text"
               values={{id: options.id}}
+              components={{Code: ({children}) => <code>{children}</code>}}
```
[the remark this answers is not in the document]
stipsan: It crashed and complained about Code not being declared instead of showing the error. Then I saw a commit from @rexxars that fixed the same error but from a different location. Hang on let's see if I can find it.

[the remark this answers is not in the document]
stipsan: https://github.com/sanity-io/sanity/pull/5434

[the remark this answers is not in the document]
ricokahler: oh hmm. I was referencing #5114

> It's a bit tedious to pass in components like <Code> that just maps to <code>. This PR allows using this predefined list of tags without specifying it directly: <abbr>, <address>, <cite>, <code>, <del>, <em>, <ins>, <kbd>, <q>, <samp>, <strong>, <sub>, <sup>.

the suggestion is to change any translation string to remove `<Code>` in favor of just `<code>` instead of adding support for `<Code>` but maybe we don't have control over those translation strings?

nbd though

ricokahler (approved):

packages/sanity/src/structure/panes/document/DocumentPane.tsx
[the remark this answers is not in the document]
stipsan: Gotcha, I have to admit I'm not too familiar with how our i18n works yet. All I know is that adding this line allowed the 404 to render again, instead of crashing the Structure Tool and propagating to the `StudioErrorBoundary` that wraps the tool layout :D

PR #6871, opened 2024-06-07
test(e2e): update timeouts for opening popovers

RitaDias:
### Description
- Update two e2e tests that tend to be flaky on next because the timeout runs out

<image src="https://github.com/sanity-io/sanity/assets/6951139/2383f308-d771-4a73-93f6-961a82948a48" width=500>

### DIFF
--- a/test/e2e/tests/document-actions/discardChanges.spec.ts
+++ b/test/e2e/tests/document-actions/discardChanges.spec.ts
@@ -33,7 +33,7 @@ test(`is possible to discard changes if a changed document has a published versi
 
   await titleInput.fill('This is a book')
 
-  publishButton.click()
+  publishButton.click({timeout: 10000})
   await expect(paneFooter).toContainText('Published just now')
 
   await titleInput.fill('This is not a book')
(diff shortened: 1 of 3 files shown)


juice49 (approved):
Thanks! This looks good to me.

It's not a blocker, but I wonder whether we should do something to prevent accumulating a bunch of arbitrary timeout durations across the test suite.

RitaDias:
> Thanks! This looks good to me.
> 
> It's not a blocker, but I wonder whether we should do something to prevent accumulating a bunch of arbitrary timeout durations across the test suite.

Yeah I've tried looking for ways (without timeouts) where these tests wouldn't be flaky but I came up short.
The good part of it is: the timeouts will only run up to those times if the CI does indeed need all of that time, otherwise they will run "shorter". 
I ran these tests multiple times just to be sure that they weren't consistently flaking, however, if any of these tests are still flaky once we merge this then I think I'll need to spend some proper time looking into it and coming up with an approach. Fingers crossed though 🤞

PR #6872, opened 2024-06-07
fix(scheduled-publishing): export EditScheduleForm from core

pedrobonamin:
### Description

Some users are depending on the `EditScheduleForm` for custom scheduled publishing  integrations, this was previously exported from the [plugin](https://github.com/sanity-io/sanity-plugin-scheduled-publishing/blob/main/src/index.ts#L13) but when the plugin was moved into core this export was not included.  

### What to review

Is this correct?

### Notes for release
Export EditScheduleForm from core


juice49 (approved):
Looks good to me!

PR #6874, opened 2024-06-07
fix: improve perf of <Resize> by creating the canvas in a side effect

stipsan:
### Description

Currently everytime `<Resize>` renders it calls `resize()` before handing it off to the `children` prop callback.
Even though the init of `canvas` is memoized with `useState`, and it's only added to the dom once, in a `useEffect`, what happens during `resize` voids any benefits from these optimizations:
- it sets `canvas.height` and `canvas.width`, [causing layout trashing](https://web.dev/articles/avoid-large-complex-layouts-and-layout-thrashing) and repaint.
- it calls ` ctx.drawImage(image, ...)`, which is expensive especially on larger images.

This PR changes it to run it in a single layout effect (the reasoning for why a layout effect is used instead of a regular effect is explained in code comments).
The effect only runs if the `image`, or `maxHeight` and `maxWidth` changes. Ensuring that these potentially expensive operations are only run when necessary. With the added bonus that React Compiler can now optimize the `<Resize />` component

### What to review

Check the perf, should be great, and otherwise behave like before.

### Testing

Existing tests should cover potential regressions.

### Notes for release

Images in the `Hotspot & Crop` tool now render faster, and uses far less memory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "check:deps": "pnpm --recursive --parallel exec depcheck",
     "check:format": "prettier . --check",
     "check:lint": "turbo run lint --continue -- --quiet",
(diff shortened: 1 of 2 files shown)


juice49 (approved):
This looks great! Thank you!

PR #6875, opened 2024-06-07
refactor: change memoization of margins to support the rule of hooks

stipsan:
### Description

By destructuring the array, which is guaranteed to always be a `[number, number, number, number]` tuple, the memoization is no longer violating the rules of hooks by overriding how deps are specified in `useMemo`. And it no longer calls `JSON.stringify` on every single render.
Thus it can also be compiled by React Compiler.

### What to review

Presence circles should work like before in general, and slightly faster in https://test-compiled-studio.sanity.build

### Testing

Existing tests should cover it.

### Notes for release

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "check:deps": "pnpm --recursive --parallel exec depcheck",
     "check:format": "prettier . --check",
     "check:lint": "turbo run lint --continue -- --quiet",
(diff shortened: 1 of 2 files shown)


juice49 (approved):

juice49:
Nice! Thanks!

PR #6878, opened 2024-06-07
refactor(core): use new value transfer function for copy/paste + tests

skogsmaskin:
### Description

This is a new and improved version of the transferValue function.

It's made more generic so that it's not too tightly coupled to the CopyPasteAction.

It will also work recursively in objects and array values, besides collecting warnings / errors for the result.

I have hooked it up with the copy/paste functionality for the studio.

Also made some new tests for the function.


sjelfull (approved):
🔥 great work!

PR #6883, opened 2024-06-09
feat(deps): upgrade react-rx to v3

stipsan:
Upgrade to a new major of `react-rx` that's more performant on React 18, and more stable on React 19 strict mode.

The `useMemoObservable` and `useObservableCallback` hooks can't be optimized by the React Compiler. Thus they are replaced with `useMemo + useObservable`, and `useObservableEvent` replaces `useObservableCallback` and uses a ponyfill of the upcoming `React.useEffectEvent` hook 🙌


package.json
```diff
@@ -17,8 +17,7 @@
     "check:deps": "pnpm --recursive --parallel exec depcheck",
     "check:format": "prettier . --check",
     "check:lint": "turbo run lint --continue -- --quiet",
-    "check:react-exhaustive-deps": "turbo run lint --continue -- --quiet --rule 'react-hooks/exhaustive-deps: [error, {additionalHooks: \"(useMemoObservable|useObservableCallback)\"}]'",
```
stipsan, line 20: This check isn't needed anymore as `check:react-compiler` covers that and more, and is also run on the CI

packages/sanity/src/core/form/inputs/CrossDatasetReferenceInput/useReferenceInfo.ts
```diff
@@ -72,14 +71,6 @@ export function useReferenceInfo(
...
     [docInfo, getReferenceInfo, retry, retryAttempt],
-    INITIAL_LOADING_STATE,
   )
-
-  // workaround for a "bug" with useMemoObservable that doesn't
-  // return the initial value upon resubscription
-  const previousId = usePrevious(doc._id, doc._id)
-  if (doc._id && previousId !== doc._id) {
-    return INITIAL_LOADING_STATE
-  }
```
stipsan, line 83: @bjoerge can you verify that this is fixed?

packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInfo.ts
```diff
@@ -66,15 +65,6 @@ export function useReferenceInfo(
...
-    INITIAL_LOADING_STATE,
   )
-
-  // @todo test and see if this were fixed in `react-rx@2.1.x`
-  // workaround for a "bug" with useMemoObservable that doesn't
-  // return the initial value upon resubscription
-  const previousId = usePrevious(id, id)
-  if (id && previousId !== id) {
-    return INITIAL_LOADING_STATE
-  }
```
stipsan, line 78: @bjoerge can you verify that this is fixed?

packages/sanity/src/core/hooks/useDocumentOperation.ts
```diff
@@ -1,12 +1,18 @@
...
+  const observable = useMemo(
     () => documentStore.pair.editOperations(publishedDocId, docTypeName),
     [docTypeName, documentStore.pair, publishedDocId],
-  )!
+  )
+  /**
+   * We know that since the observable has a startWith operator, it will always emit a value
+   * and that's why the non-null assertion is used here
+   */
+  return useObservable(observable)!
```
stipsan, line 17: I tried to set an `initialValue` here instead, but couldn't work out what that initial value looks like 🤔

packages/sanity/src/core/hooks/useEditState.ts
```diff
@@ -25,5 +26,10 @@ export function useEditState(
     }
 
     return documentStore.pair.editState(publishedDocId, docTypeName)
-  }, [documentStore.pair, publishedDocId, docTypeName, priority]) as EditStateFor
+  }, [docTypeName, documentStore.pair, priority, publishedDocId])
+  /**
+   * We know that since the observable has a startWith operator, it will always emit a value
+   * and that's why the non-null assertion is used here
+   */
+  return useObservable(observable)!
```
stipsan, line 34: I tried to set the `initialValue` here instead of the non-null assertion. Unfortunately some of the initial state is non-trivial to work out: https://github.com/sanity-io/sanity/blob/62581ba91e96de16b7f23e4afdf1b92e9f234602/packages/sanity/src/core/store/_legacy/document/document-pair/editState.ts#L68

packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInfo.ts
bjoerge: It should be, yes!

packages/sanity/src/core/hooks/useEditState.ts
bjoerge: We should be good with a non-null assertion since we know these has sync emissions (unfortunately there's no way to express this constraint with TS).

packages/sanity/src/core/hooks/useDocumentOperation.ts
bjoerge: It's fine - thanks for adding the comment though!

packages/sanity/src/core/form/inputs/CrossDatasetReferenceInput/useReferenceInfo.ts
bjoerge: Tested & verified

bjoerge (approved):

PR #6887, opened 2024-06-10
fix(form): avoid spreading key prop

christianhg:
Before this change, React would throw a runtime error stating that `A props object containing a "key" prop is being spread into JSX`. Now, the `key` is passed to the JSX without using spread.


cngonzalez (approved):
This is great, thank you!

PR #6893, opened 2024-06-10
feat: notify AUS studio users of new studio versions

cngonzalez:
### Description
As we move to certain packages being able to auto-update, we want end-users of the studio to know when to refresh their browsers and work with the most up-to-date packages.

Screenshot 2024-06-18 at 11 01 51 AM

This PR adds a Provider that will:
1. Poll an endpoint in module-server every 30 minutes to check for the most up-to-date packages.
2. Check if the current studio is an auto-updating studio.
3. If so, send a toast alerting the user to reload their browsers, along with a button.

### What to review
1. The language and design of the toast.
2. The logic of the check -- anything strange?

Unfortunately I think the reviewer ought to pull this down and review manually -- see below for why (because of limited ability to test).

### Testing
I've added a few e2e tests. I'm not testing if the reload button actually reloads the page; I couldn't find an easy way for Playwright to listen to that event.

### DIFF
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1087,7 +1087,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'new-document.title': 'Create new document',
   /** More detailed alert text letting user know they have an out-of-date version and should reload */
   'package-version.new-package-available.description':
-    'A new version of Sanity Studio is available. Please reload this window to update.',
+    'Simply reload to use the new version.',
   /** Label for button that will make the browser reload when users' studio version is out-of-date */
   'package-version.new-package-available.reload-button': 'Reload now',
   /** Title of the alert for studio users when packages in their studio are out-of-date */
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1091,7 +1091,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   /** Label for button that will make the browser reload when users' studio version is out-of-date */
   'package-version.new-package-available.reload-button': 'Reload',
   /** Title of the alert for studio users when packages in their studio are out-of-date */
-  'package-version.new-package-available.title': 'New version available',
+  'package-version.new-package-available.title': 'Sanity Studio was updated',
   /** Label for action to manage members of the current studio project */
   'presence.action.manage-members': 'Manage members',
   /** Accessibility label for presence menu button */
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1086,8 +1086,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   /** Title for "Create new document" dialog */
   'new-document.title': 'Create new document',
   /** More detailed alert text letting user know they have an out-of-date version and should reload */
-  'package-version.new-package-available.description':
-    'Simply reload to use the new version.',
+  'package-version.new-package-available.description': 'Simply reload to use the new version.',
   /** Label for button that will make the browser reload when users' studio version is out-of-date */
   'package-version.new-package-available.reload-button': 'Reload',
   /** Title of the alert for studio users when packages in their studio are out-of-date */
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1089,7 +1089,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'package-version.new-package-available.description':
     'Simply reload to use the new version.',
   /** Label for button that will make the browser reload when users' studio version is out-of-date */
-  'package-version.new-package-available.reload-button': 'Reload now',
+  'package-version.new-package-available.reload-button': 'Reload',
   /** Title of the alert for studio users when packages in their studio are out-of-date */
   'package-version.new-package-available.title': 'New version available',
   /** Label for action to manage members of the current studio project */
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1085,7 +1085,13 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'new-document.open-dialog-aria-label': 'Create new document',
   /** Title for "Create new document" dialog */
   'new-document.title': 'Create new document',
-
+  /** More detailed alert text letting user know they have an out-of-date version and should reload */
+  'package-version.new-package-available.description':
+    'A new version of Sanity Studio is available. Please reload this window to update.',
+  /** Label for button that will make the browser reload when users' studio version is out-of-date */
+  'package-version.new-package-available.reload-button': 'Reload now',
+  /** Title of the alert for studio users when packages in their studio are out-of-date */
+  'package-version.new-package-available.title': 'New version available',
   /** Label for action to manage members of the current studio project */
   'presence.action.manage-members': 'Manage members',
   /** Accessibility label for presence menu button */
--- a/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
+++ b/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
@@ -71,7 +71,7 @@ export function PackageVersionStatusProvider({children}: {children: ReactNode}) 
         if (!latestPackageVersions) return
 
         const foundNewVersion = Object.entries(latestPackageVersions).some(([pkg, version]) => {
-          if (!version || currentPackageVersions[pkg]) return false
+          if (!version || !currentPackageVersions[pkg]) return false
           return semver.gt(version, currentPackageVersions[pkg])
         })
 
--- a/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
+++ b/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
@@ -8,13 +8,9 @@ import {useTranslation} from '../../i18n'
 import {checkForLatestVersions} from './checkForLatestVersions'
 
 // How often to to check last timestamp. at 30 min, should fetch new version
-const REFRESH_INTERVAL = 1000 * 60 * 5 // every 5 minutes
+const REFRESH_INTERVAL = 1000 * 30 // every 30 seconds
 const SHOW_TOAST_FREQUENCY = 1000 * 60 * 30 //half hour
 
-/*
- * We are currently only checking to see if the sanity module has a new version available.
- * We can add more packages to this list (e.g., @sanity/vision) if we want to check for more.
- */
 const currentPackageVersions: Record<string, string> = {
   sanity: SANITY_VERSION,
 }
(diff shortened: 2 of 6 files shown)


packages/sanity/src/core/studio/packageVersionStatus/checkForLatestVersions.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// How often to check for a version
+const REFRESH_INTERVAL = 1000 * 60 * 30 // every half hour
+
+//reset the observable when it completes or when it has no refcount
+const RESET_TIMER = timer(REFRESH_INTERVAL)
+
+const MODULES_URL_VERSION = 'v1'
+
+const MODULES_URL = `https://sanity-cdn.com/${MODULES_URL_VERSION}/modules/`
```
binoy14, line 17: Could we somehow share this with logic in CLI here? https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/_internal/cli/util/getAutoUpdatesImportMap.ts#L11 

Maybe we just need to lift this out in some constant?

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {hasSanityPackageInImportMap} from '../../environment/hasSanityPackageInImportMap'
+import {useTranslation} from '../../i18n'
+import {checkForLatestVersions} from './checkForLatestVersions'
+
+/*
+ * We are currently only checking to see if the sanity module has a new version available.
+ * We can add more packages to this list (e.g., @sanity/vision) if we want to check for more.
+ */
+const currentPackageVersions: Record<string, string> = {
+  sanity: SANITY_VERSION,
```
binoy14, line 15: Just laying out my thoughts, if it makes sense to you as well. 

When we run the bundle we set the `process.env.SANITY_VERSION` reading the package.json so when the bundle is built and uploaded it should have the correct version in the bundle so reading the version should always resolve to the correct version. I am thinking this correctly?

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const {t} = useTranslation()
+
+  const autoUpdatingPackages = hasSanityPackageInImportMap()
+
+  const onClick = () => {
+    window.location.reload()
+  }
+
+  useEffect(() => {
+    const sub = checkForLatestVersions(currentPackageVersions).subscribe({
```
binoy14, line 30: ❓would it make sense to check the version only if it's autoUpdating studios? Doesn't make sense to call the service if it's not, seems unnecessary

binoy14 (commented):
Overall looks good have some thoughts and questions

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
cngonzalez: ugh yes, victim of a refactor trying to get an e2e test going. sorted now.

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
cngonzalez: That's my understanding! Brief manual testing has show it defined to what I expect, which you can also see in the navbar, for example (it respects tagged versions et al).

This won't be available for other packages (as of now) so we should strategize when that happens.

binoy14: Discussed that this can be addressed later

packages/sanity/src/core/studio/packageVersionStatus/checkForLatestVersions.ts
binoy14: I am not sure if there is, might be worth asking internally but for now okay with copy pasting as well

binoy14:
Just re-request review when you want me to re-review

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+        })
+
+        if (foundNewVersion) {
+          showNewPackageAvailableToast()
+        }
+      })
+    }
+
+    // Run on first render
+    fetchLatestVersions()
```
juice49, line 85: These fetches don't appear to get cancelled in the effect clean-up function at the moment.

cngonzalez: Interesting! What behavior are you seeing -- double fetches or something similar?

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+              text={t('package-version.new-package-available.reload-button')}
+              tone={'primary'}
+            />
+          </Box>
+        </Stack>
+      ),
+      closable: true,
+      status: 'info',
+      //covering for some delays, etc. because of the toast ID, we should never see this twice
+      duration: SHOW_TOAST_FREQUENCY * 2,
```
ricokahler, line 53: is there an option for like an infinite duration? it seems like we just want this up indefinitely until the user closes this manually and it seems like the toast ID prevents double of the same message. would that be better?

ricokahler (dismissed):
looks good! left some open questions but nothing blocking

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
cngonzalez: we have precedent!

https://github.com/sanity-io/sanity/blob/4b89dd555e22d91372119c33de39530c2b01f979/packages/sanity/src/structure/panes/document/documentPanel/documentViews/useConditionalToast.ts#L13

in fact, it may just make sense to reuse the useConditionalToast util here...

packages/sanity/src/core/version.ts
```diff
@@ -1,6 +1,7 @@
+import {version} from '../../package.json'
```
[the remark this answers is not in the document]
cngonzalez: This was changed in parallel with Espen's branch for Sentry testing: https://github.com/sanity-io/sanity/pull/6914/files

so it's more here for alignment (and to make testing easier). I haven't seen an issue so far!

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+import {type ReactNode, useCallback, useEffect, useRef} from 'react'
+import {SANITY_VERSION} from 'sanity'
+import semver from 'semver'
+
+import {hasSanityPackageInImportMap} from '../../environment/hasSanityPackageInImportMap'
+import {useTranslation} from '../../i18n'
+import {checkForLatestVersions} from './checkForLatestVersions'
+
+// How often to to check last timestamp. at 30 min, should fetch new version
+const REFRESH_INTERVAL = 1000 * 60 * 5 // every 5 minutes
```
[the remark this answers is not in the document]
cngonzalez: It's quite arbitrary at the moment... since the timeliness of this particular toast isn't so vital, I wasn't so concerned with a bit of drift. We can up the interval!

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      checkForLatestVersions(currentPackageVersions).then((latestPackageVersions) => {
+        lastCheckedTimeRef.current = Date.now()
+
+        if (!latestPackageVersions) return
+
+        const foundNewVersion = Object.entries(latestPackageVersions).some(([pkg, version]) => {
+          if (!version) return false
+          return semver.gt(version, currentPackageVersions[pkg])
```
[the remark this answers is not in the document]
cngonzalez: updated!

test/e2e/tests/default-layout/versionStatus.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      script.textContent = JSON.stringify(importMap)
+      document.head.appendChild(script)
+    })
+  })
+
+  test('should facilitate reload if in auto-updating studio, and version is higher', async ({
+    page,
+  }) => {
+    // Intercept the API request and provide a mock response
+    await page.route('https://sanity-cdn.**/v1/modules/sanity/default/**', (route) => {
```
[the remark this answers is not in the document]
cngonzalez: wrote a comment -- but can definitely revisit!

packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
```diff
@@ -0,0 +1,94 @@
...
+        return
+      }
+
+      checkForLatestVersions(currentPackageVersions).then((latestPackageVersions) => {
+        lastCheckedTimeRef.current = Date.now()
+
+        if (!latestPackageVersions) return
+
+        const foundNewVersion = Object.entries(latestPackageVersions).some(([pkg, version]) => {
+          if (!version || currentPackageVersions[pkg]) return false
```
ricokahler, line 74: should this be?

```suggestion
          if (!version || !currentPackageVersions[pkg]) return false
```

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1085,7 +1085,13 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
...
   /** Title for "Create new document" dialog */
   'new-document.title': 'Create new document',
-
+  /** More detailed alert text letting user know they have an out-of-date version and should reload */
+  'package-version.new-package-available.description':
+    'A new version of Sanity Studio is available. Please reload this window to update.',
+  /** Label for button that will make the browser reload when users' studio version is out-of-date */
+  'package-version.new-package-available.reload-button': 'Reload now',
+  /** Title of the alert for studio users when packages in their studio are out-of-date */
+  'package-version.new-package-available.title': 'New version available',
```
runeb, line 1094:
```suggestion
  'package-version.new-package-available.title': 'Sanity Studio was updated',
```

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1085,7 +1085,13 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'new-document.open-dialog-aria-label': 'Create new document',
   /** Title for "Create new document" dialog */
   'new-document.title': 'Create new document',
-
+  /** More detailed alert text letting user know they have an out-of-date version and should reload */
+  'package-version.new-package-available.description':
+    'A new version of Sanity Studio is available. Please reload this window to update.',
+  /** Label for button that will make the browser reload when users' studio version is out-of-date */
+  'package-version.new-package-available.reload-button': 'Reload now',
```
runeb, line 1092:
```suggestion
  'package-version.new-package-available.reload-button': 'Reload',
```

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1085,7 +1085,13 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'new-document.open-dialog-aria-label': 'Create new document',
   /** Title for "Create new document" dialog */
   'new-document.title': 'Create new document',
-
+  /** More detailed alert text letting user know they have an out-of-date version and should reload */
+  'package-version.new-package-available.description':
+    'A new version of Sanity Studio is available. Please reload this window to update.',
```
runeb, line 1090:
```suggestion
    'Simply reload to use the new version.',
```

ricokahler (approved):

ricokahler (approved):

ricokahler (approved):

PR #6894, opened 2024-06-10
refactor(structure): refactor usePreviewUrl for react-rx v3

bjoerge:
### Description

This removes usage of (now inlined) useAsObservable which will be removed in react-rx@3.x, and also currently causes the React compiler to deopt.

### What to review
Does the changes make sense?
Note: If the user-provided resolveProductionUrl fails, it will currently throw an error in a separate stack frame that can only be captured by window.onerror. After https://github.com/sanity-io/react-rx/pull/87 lands, this error will be part of the react render stack frame and will be captured by the closest error boundary.

### Testing
This is mostly removing and simplifying existing code. I didn't prioritize adding tests at this point since this is not a very critical part of the Studio, and the main goal here was to avoid React compiler deopt.

### Notes for release
n/a - internal


stipsan (approved):
Thanks! 💖

Issue #6698, opened 2024-05-16
Both new and legacy search don't work with documents with a custom title field


alexeygolev:
I say "don't work" because "I have no idea why it's doing what it's doing" is needlessly emotional and also too long 🙂 

**Describe the bug**

Our schema has "headline" for the title field. Currently the new search doesn't play well with this setup. Even if I set a silly number of 100 for the weight inside `options.search` for our headline field. I inspected the search results in browser tools and the list of hits returned from the server seems to have the highest `matchScore` for the result that I expect to be the first in the list of results. But the UI doesn't reflect this. 

I tried enabling the legacy search using `enableLegacySearch` but it only improved the omnisearch behaviour. All the reference fields are still not finding our documents. The document filters above the document list are also yielding no results. I'm not even sure what version to rollback to. Our editors can't use the reference fields without jumping through a bunch of hoops now.

**Update:** after some digging I found that the "legacy search" ignores the fields of type `text` which is what our headlines are (our headlines are quite long and at the point when the system was created setting it to `text` was the best way to have a text area as an input). If I change the field type to text it shows up in search as it is now included in the legacy search request. So "legacy search" behaves differently to the actual old search as it excludes the `text` fields.

**To Reproduce**

Our schema is too complicated to show here. But the root of the problem is:
- not having a title field (let's say having a "headline" field instead) 
- setting the type of the "headline" field to text
- having the slug field use "headline" for the title field
- making the headline field value different to the current slug (if they are the same then you can get an impression that the search actually works)
- searching for the full headline

We can't go back to v2 now as we have migrated quite a while ago but we're also stuck with the search that barely works (yes, omnisearch is marginally better for some reason, but we use reference fields more often than we use omnisearch)

kbrabrand:
We've stumbled onto sort of the same bug. In the global search we get hits for the title of the doc both when setting the field to `string` and `text`, but when searching for a doc to add as a reference or in the docTypeList search (in the desk structure) we get hits for it only if it's set to `string`. When setting the field type to `text` in the schema def (and making no other changes) we don't get hits for the field.

rexxars:
Fixed in [v3.46.0](https://github.com/sanity-io/sanity/releases/tag/v3.46.0) - thanks for reporting!

PR #6895, opened 2024-06-10
fix(search): include `text` type fields in search queries

rexxars:
### Description

It seems we do not include fields of type `text` in the list of fields to search for. 
Upon some investigation, it seems like we are traversing the type chain in order to find any type that extends from `string`. This isn't strictly necessary, as there is a `jsonType` property that allows us to more easily tell this. The `text` type does not actually "extend" `string` in the same way as user-defined types does, so it did not pass the previous check.

This PR changes to use `jsonType` for determining the string fields.

See #6698

### What to review

- Searching for text in string fields work
- Searching for text in text fields work
- Searching for text in portable text fields work
 
### Testing

Added a test to the function that derives weights from types that checks that both `text` fields and fields of types that _extend_ `text` are included in the search.

### Notes for release

- Fixes an issue where searching for text within fields of type `text` (not `string`) would not yield results


juice49 (approved):
Thank you! That makes sense, and works as expected.

PR #6898, opened 2024-06-11
feat(structure): add support for boolean inputs

pedrobonamin:
### Description
Boolean input for the sheet list, with `switch` and `checbox`

Supports:
- Copy
- Paste
- Delete (transforms into undetermined value)
Screenshot 2024-06-11 at 16 23 19

### DIFF
--- a/packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
+++ b/packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
@@ -50,7 +50,10 @@ const getFieldValueAsFieldType = (
   fieldType: BooleanSchemaType | StringSchemaType | NumberSchemaType,
 ) => {
   if (isBooleanSchemaType(fieldType)) {
-    return providedValue === 'true'
+    if (typeof providedValue === 'string') {
+      return providedValue === 'true'
+    }
+    return providedValue
   }
   if (isNumberSchemaType(fieldType)) {
     return parseFloat(providedValue)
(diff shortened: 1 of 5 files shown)


dev/test-studio/schema/author.ts
```diff
@@ -109,10 +109,17 @@ export default defineType({
     // },
     {
       name: 'locked',
-      title: 'Locked',
+      title: 'Locked (switch)',
```
jordanl17, line 112: Do we want to update this schema?

packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
```diff
@@ -229,6 +238,7 @@ export function SheetListCell(cell: Cell<DocumentSheetTableRow, unknown>) {
 
   const inputProps = {
     ...cellContext,
+    'handlePatchField': handlePatchField,
```
jordanl17, line 241:
```suggestion
    handlePatchField,
```

jordanl17 (approved):

packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
```diff
@@ -173,16 +177,22 @@ export function SheetListCell(cell: Cell<DocumentSheetTableRow, unknown>) {
     (event: ClipboardEvent) => {
       const clipboardData = event.clipboardData?.getData('Text')
       if (!clipboardData || !fieldType) return
+      if (
```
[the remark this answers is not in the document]
pedrobonamin: Let's discuss your thoughts on this, I don't see it in the same way as this is triggered by an event listener here https://github.com/sanity-io/sanity/blob/edx-1477/packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx#L203-L205 and I find it hard to reason on how to actually share it in a similar way without having to reimplement the useEffect with all the logic involved.
I was also thinking on maybe moving this useEffect into a reusable hook to share in the cells. e.g. `useKeyboardShortcuts` to handle all the keyboard events `handleCopy, handleOnKeyDown, handlePaste`

PR #6907, opened 2024-06-12
chore(structure): disable sheet list view if no schemaType is found

pedrobonamin:
### Description

Sheet list view has the limitation of working only with 1 document type. For that, users will need to use the `DocumentTypeListBuilder` to expose the `schemaType` property in the pane. 
If that property is not present or the type was not found, the following message will be shown.
screencapture-localhost-3333-test-structure-custom-authorsBooks-2024-06-12-11_24_25

Screenshot 2024-06-12 at 11 30 14

### DIFF
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -403,6 +403,10 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The text for the "Open preview" action for a document */
   'production-preview.menu-item.title': 'Open preview',
 
+  /** The message that will be shown to users when the table view is not supported the list has no schema type */
+  'sheet-list.not-supported.no-schema-type':
+    'Table view is only supported using the <code>S.documentTypeListItems</code> builder',
+
   /** Label for button when status is saved */
   'status-bar.document-status-pulse.status.saved.text': 'Saved',
   /** Label for button when status is syncing */
--- a/packages/sanity/src/structure/panes/documentList/sheetList/DocumentSheetListPane.tsx
+++ b/packages/sanity/src/structure/panes/documentList/sheetList/DocumentSheetListPane.tsx
@@ -8,9 +8,10 @@ import {
   useReactTable,
 } from '@tanstack/react-table'
 import {useCallback, useEffect, useState} from 'react'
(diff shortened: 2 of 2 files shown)


packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -403,6 +403,10 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The text for the "Open preview" action for a document */
   'production-preview.menu-item.title': 'Open preview',
 
+  /** The message that will be shown to users when the table view is not supported the list has no schema type */
+  'sheet-list.not-supported.no-schema-type':
+    ' Table view is only supported using the <code>S.documentTypeListItems</code> builder',
```
binoy14, line 408: Should it be this or the space is expected?

```suggestion
    'Table view is only supported using the <code>S.documentTypeListItems</code> builder',
```

binoy14 (commented):
Changes makes sense, have one comment. Happy to re approve once it's addressed

packages/sanity/src/structure/i18n/resources.ts
pedrobonamin: Hey @binoy14, thanks, updated!

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -403,6 +403,10 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The text for the "Open preview" action for a document */
   'production-preview.menu-item.title': 'Open preview',
 
+  /** The message that will be shown to users when the table view is not supported the list has no schema type */
+  'sheet-list.not-supported.no-schema-type':
```
binoy14, line 407: Based on the team discussion today, should this be moved to a different resource that is not structure?

binoy14 (commented):
Code is good, just wanted to double check on the translations

packages/sanity/src/structure/i18n/resources.ts
pedrobonamin: Yes, will close this PR and move it to the branch where we are adding the new file with the translations.  Thanks Binoy!

pedrobonamin:
Closing this PR because the translation has been moved to another file, which is part of other PR. Changes moved to https://github.com/sanity-io/sanity/tree/sheet-inputs

PR #6914, opened 2024-06-12
feat(core): implement basic error reporting

rexxars:
### Description

This introduces an error reporter using Sentry, for auto-updating, non-embedded studios. In the future, we may extend this to other studio types/configurations.

### Review notes

- Can be reviewed as a whole - I'll squash on merge
- The only (seemingly) unrelated commit here is the first one, which changes the `SANITY_VERSION` constant from being a static `0.0.0-development` to say `3.15.3-development`. This is really only relevant for the monorepo, but is helpful in the context of error reporting to see what the "base" version is.
- The Sentry relay is currently not working due to some CORS issues I am working to figure out, but the _code_ aspect of this is ready for review.
- I am aware of a few locations in the code that needs a few comments to clarify the purpose/rationale for them - I'll see if I can find the time to add them soon.
- There is a new "Errors test" tool in the test studio that has five different error "causes", and can be triggered by clicking the respective buttons. Each of these should cause an error to be reported to Sentry (once the relay is working).
  - Thrown errors in event handlers (something that runs on click, for instance, and is not executed during _rendering_)
  - Type errors in event handlers (eg calling a method on `undefined` or similar)
  - Errors in timeouts (`setTimeout`/`setImmediate`/`setInterval`)
  - Unhandled rejections (promises without a catch handler)
  - Errors during React render cycle
- Known limitations/areas to improve on:
  - Filtering errors that is not caused by Sanity. This needs more investigation in introspecting the stack trace etc - just have not had time to look at this yet, and can be added later.
  - (Potentially) add a flag to opt-out of error reporting
  - Add more context to reported errors - knowing more about the state of the application would help debug issues

### Testing

We're currently missing tests. I'd like to add some playwright tests that trigger an error and verify that a network request to the Sentry relay is triggered, but have not had time yet. Will see if I can find time before merging this.

### Notes for release

None, I think. Will double check if this should be called out or not.

### DIFF
--- a/dev/test-studio/plugins/error-reporting-test/ErrorReportingTest.tsx
+++ b/dev/test-studio/plugins/error-reporting-test/ErrorReportingTest.tsx
@@ -1,0 +1,73 @@
+import {Button, Card, Flex, Stack} from '@sanity/ui'
+import {useCallback, useState} from 'react'
+
+function triggerCustomErrorOnEvent() {
+  throw new Error('Custom error triggered')
+}
+
+function triggerTypeErrorOnEvent(evt: any) {
+  evt.someFunctionThatDoesntExist()
+}
+
+function triggerTimeoutError() {
+  setTimeout(() => {
+    throw new Error('Custom error in setTimeout')
+  }, 1000)
+}
+
+function triggerPromiseError() {
+  return new Promise((resolve, reject) => {
+    requestAnimationFrame(() => {
+      reject(new Error('Custom error in promise'))
+    })
+  })
+}
+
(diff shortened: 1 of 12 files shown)


binoy14 (commented):
Code looks good and tested in different situations and seems to be piping correctly. There is opportunity to improve source map detection but that can be follow up PRs. There seems to be a linting error and merge conflict. Once that is fixed happy to ✅ it

rexxars:
Rebased, lint and tests passing

binoy14 (approved):
Thanks!

PR #6921, opened 2024-06-13
feat(core): support insert menu options in array item context menus

christianhg:


packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/InsertMenuMenuItems.tsx
```diff
@@ -0,0 +1,118 @@
...
+    insertMenuProps: {
+      ...props.insertMenuOptions,
+      schemaTypes: props.schemaTypes ?? [],
+      onSelect: (insertType) => {
+        props.onInsert('before', insertType)
+      },
+    },
+    popoverProps: {
+      referenceElement: props.referenceElement,
+      placement: 'top-end',
```
ricokahler, line 36: do you think it would be better to have this be consistent with `insertAfter`? I personally think setting them both to prefer `'bottom-end'` feels better.

This video is the current experience.


https://github.com/sanity-io/sanity/assets/10551026/c6e01f5c-332c-481c-a958-306080b1a350

ricokahler (commented):
The code looks great! I had one UX suggestion in there but besides that I couldn't find any other issues.

I would like to encourage adding more tests if you have the time. On the studio team, we're trying to stricter about quality and culturally we want everyone to include tests for new surfaces added in the stduio. this can be e2e or jest.

packages/sanity/src/core/form/inputs/arrays/ArrayOfObjectsInput/InsertMenuMenuItems.tsx
christianhg: The decision to alter between `top-end` and `bottom-end` is primarily made by @mariuslundgard. The idea is that the placement gives a little more context in terms of where the item is inserted (above or below).

PR #6927, opened 2024-06-13
test(e2e): update e2e flaky tests

RitaDias:
### Description

Update e2e flaky tests to be more sturdy

- Update `unstable_serverActions` to `__internal_serverDocumentActions` on e2e studio (missed after naming update)
- Update e2e tests to watch for changes in visibility / attachment 
- Add skips for firefox only flaky tests (behaviour functions, limitation of playwright and so making them unreliable)
- Skip document actions tests that are failing while they are knowingly flaky (discussion on slack, tl:dr: behaviour in CL is unreliable right now and this is known and being worked on, ask for link ❤️ )
  - `documentTypeListContextMenu.spec -> clicking custom sort order and direction sets value in storage`
  - `discardChanges.spec -> is possible to discard changes if a changed document has a published version`
  - `discardChanges.spec -> displays the published document state after discarding changes`
  - `fetchFeatureToggle.spec -> document actions follow appropriate logic after receiving response from feature toggle endpoint`
  - `restore.spec -> documents can be restored to an earlier revision`
  - `restore.spec -> respects overridden restore action`
  - `restore.spec -> respects removed restore action`
  - `restore.spec -> user defined restore actions should not appear in any other document action group UI`
- Update waitForResponse parameter for PTE (kanji)

Re ran the tests multiple times and have not have them fail :)

### What to review

Double check that the tests that have been skipped related to the document actions make sense.
Overall test health and sense.

### Notes for release

N/A


dev/studio-e2e-testing/sanity.config.ts
```diff
@@ -95,7 +95,8 @@ export default defineConfig({
 
   plugins: [sharedSettings()],
   basePath: '/test',
-  unstable_serverActions: {
```
RitaDias, line 98: Name hadn't been changed when renaming happened

binoy14 (approved):
This is awesome! Thank you so much!

PR #6929, opened 2024-06-13
fix(form): activate PTE input when dragging files

sjelfull:
### Description

After the change in #6867 you have to activate the PTE before you are able to drag files into the PTE.

This change fixes that situation by activating the PTE whenever the drag starts.

### What to review

- That the PTE activates when you start dragging files into it.

### Notes for release

- Fixes a issue where the PTE won't activate when you start dragging files into it


skogsmaskin (approved):
👍

PR #6930, opened 2024-06-13
fix: layout shift when uploading images

stipsan:
### Description

This branch started out as an attempt at fixing a React 19 issue where the Image Tool buttons failed to render, and the assumption was that the root cause were that `BaseImageInput` component were still a class component and had bugs in it implementation that linters and typechecks were unable to pick up.
After finishing the refactor to a Functional Component the issue were still there, [and turned out to be a problem in `@sanity/ui` that's now fixed](https://github.com/sanity-io/ui/pull/1349).
Since the refactor to a functional component is still valuable (easier to optimize, allows using hooks instead of hacks like `<ImperativeToast/>`, etc) I kept going and uncovered a bug while testing that the refactor doesn't have regressions:

https://github.com/sanity-io/sanity/assets/81981/72e863e1-8df8-4468-be46-1e4a58776b81

This layout shift issue is recorded on https://test-studio.sanity.build, it's easier to reproduce if slow network conditions are simulated:

image

This PR includes a fix, here's the new behaviour:

https://github.com/sanity-io/sanity/assets/81981/d774c9a7-1e1e-4471-bf92-dcd7ca155820

### What to review

Does the new code structure make sense? The goal was to reuse existing names as much as possible, and to co-locate things.

### Testing

Unsure if existing test suites are enough.
Personally I've tested all the flows I could think of and they're passing:
- uploading an image on an empty input
- updating hotspot and crop
- clearing an image and then upload another image
- drag to upload image
- copy image into input
- paste image on existing image with different dimensions to test layout shift
- selecting new images from the asset source browser
- tested error states (uploading .avif images, interrupt uploads)
- custom asset sources

### Notes for release

Reduce layout shift when uploading images that replace existing image assets.


juice49 (approved):
Thank you! Code changes and Studio changes when testing look good to me.

PR #6936, opened 2024-06-14
feat(structure): Sheet List dropdown selects

jordanl17:
### Description
* Supporting Dropdown selection for string and number selects and radio button groups
* Fixing issue where unsetting a value in document form would not propagate to sheet list

### What to review
* New `DropdownCellInput` renders for dropdown selects
* `useOnMouseDownCell` hook to group the logic around whether a cell should require dbl click to focus or not

### Testing
I'll add more wider tests for all the sheet list inputs in a separate branch

### Notes for release
Nothing relevant to note- no user facing changes

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/string/StringList.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/string/StringList.tsx
@@ -1,8 +1,9 @@
 import {ChevronDownIcon} from '@sanity/icons'
-import {isTitledListValue, type StringOptions, type TitledListValue} from '@sanity/types'
+import {type StringOptions, type TitledListValue} from '@sanity/types'
 import {Menu} from '@sanity/ui'
-import {capitalize, uniq} from 'lodash'
+import {uniq} from 'lodash'
 import {useCallback, useId, useMemo} from 'react'
+import {toSelectItem} from 'sanity'
 
 import {Button, MenuButton, MenuItem} from '../../../../../../../../../../ui-components'
 import {useSchema} from '../../../../../../../../../hooks'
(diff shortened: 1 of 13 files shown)


packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/string/StringList.tsx
```diff
@@ -1,8 +1,9 @@
 import {ChevronDownIcon} from '@sanity/icons'
-import {isTitledListValue, type StringOptions, type TitledListValue} from '@sanity/types'
+import {type StringOptions, type TitledListValue} from '@sanity/types'
 import {Menu} from '@sanity/ui'
-import {capitalize, uniq} from 'lodash'
+import {uniq} from 'lodash'
 import {useCallback, useId, useMemo} from 'react'
+import {toSelectItem} from 'sanity'
```
jordanl17, line 6: Figured this should just be reused here, seeing at it was the same method

packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
```diff
@@ -160,19 +173,6 @@ export function SheetListCell(cell: Cell<DocumentSheetTableRow, unknown>) {
     resetFocusSelection()
   }, [handlePatchField, cellValue, resetFocusSelection])
 
-  // child field inputs can control whether default behavior is stopped or preserved
-  const getOnMouseDownHandler = useCallback(
```
jordanl17, line 164: Moved this to a custom hook since, a) it requires explicit callout and b) because the mouse handler needs to be added to the Cell itself otherwise the intended behavior isn't as expected since not all inputs span the entire space of a Cell

PR #6946, opened 2024-06-14
feat: refactor `createTrackerScope` internals to be more efficient

stipsan:
### Description

Refactors the internal logic that drives presence overlays, and change indicators, which has to keep up with user scrolling, layout shifts and more.
To avoid causing regressions the new implementation has the same `lodash.throttle` semantics as before, and still runs `store.update` on every side-effect render (as the `useLayoutEffect` in both the current implementation, and new one, do not specify deps).
This could be further improved, but we're probably way better of if we start using `@floating-ui/react` for this, as it has proven to be quite performant for our `Tooltips`, `Popovers` and `MenuButtons`.

I'm happy about the improvements that did land:
- 3 `useLayoutEffect` cycles reduced to 2 (the `useReportedValues` hook no longer needs it, as `store.update` now leads to a regular React state update so a separate pubsub isn't needed).
- `context.read()`, which calls `Array.from(map.entries())`, is replaced. Previously this iteration happened on every render (as the `useState` [init isn't lazy](https://github.com/sanity-io/sanity/blob/fbf90ec8713298203b081b9e731e69c8135a2080/packages/sanity/src/core/components/react-track-elements/createTrackerScope.tsx#L74))
- the `useReporter` hook called `useLayoutEffect` often twice on every render, causing both `state.add` and `state.update` logic to run, as well as `deepEqual` on change indicator logic. It's refactored so that either it runs the `store.add + store.remove` layout effect hook, or it runs the `store.update` one, never both at the same time.
- the `useReporter` and `useReportedValues` hooks now have their own context instances. This allows React to have smarter and more granular rendering. When setting the context value for `useReportedValues` if no components are mounted that use the hook then it React skips re-rendering. It also reduces re-renders caused by `useReporter` two just 1, since the `store` instance it reads from the context doesn't change during the lifecycle of `<Tracker>`.
- The tracker context is split up, and moved into `sanity/_singletons`. It was the last exception to the new rule that AUS introduced for having all React `createContext` instances to be in one place, to avoid circular reference bugs in environments such as esm.sh.

### What to review

The naming is very verbose and could probably be better. 

### Testing

Existing e2e tests are not enough as both the change indicators and presence overlays have to deal with challenging layout shifts, frantic user scrolling and more.
Additionally presence overlays needs multiple people, or tabs, to properly test. The more people in a document jumping between fields, the better.

### Notes for release

N/A it's a refactor, and the perf benefits are likely not end user noticeable as we have other perf issues that are blocking the main thread (`styled-components` updating CSSOM, the mutator store patching JSON cached docs etc)


packages/sanity/src/core/changeIndicators/constants.ts
```diff
@@ -1,6 +1,3 @@
-// Toggle tracking of change indicator DOM positions and connectors overlay rendering
-export const ENABLED = true
```
RitaDias, line 2: This might be my lack of context of this overall section showing, but is it safe then to remove it? I see that you've removed the use of it everywhere else, but I'm wondering how far impacting this is

Especially since before we were using this at least to have an alternative to when there is no presence set up, right?

packages/sanity/src/core/components/react-track-elements/hooks.ts
```diff
@@ -0,0 +1,116 @@
...
+
+function createStore<Value>(reportedValues: Map<string, Value>, publish: () => void) {
+  function add(id: string, value: Value) {
+    if (reportedValues.has(id)) {
+      // eslint-disable-next-line no-console
+      // console.error(
+      //   new Error(
+      //     `Invalid call to useReporter(${id}): A component reporting on "${id}" is already mounted in the subtree. Make sure that all reporters within the same <Tracker> subtree have unique ids.`
+      //   )
+      // )
```
RitaDias, line 21: Seems like this if condition is empty (only with commented code?) I'm assuming something is meant to have gone in here?

RitaDias: The same thing with the rest of the "if"s in this file 👇 
- update
- remove

packages/sanity/src/core/components/react-track-elements/hooks.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  useLayoutEffect(() => {
+    /**
+     * Runs after every render in case the result of calling `value` has changes that
+     * the `isEqual` function picks up on.
+     * @TODO This is a bit expensive, and we should migrate to using a library like `@floating-ui/react` instead of rolling our own solution.
+     */
+    if (id === null || idRef.current === null || store === null || id !== idRef.current) {
+      return undefined
+    }
```
RitaDias, line 105: I read your comment in the description but I have to admit that I'm still not quite sure how the "@floating/react" would help here.

It might be because I'm having a hard time placing what this is targetting but from what I understood it (can be) the presence circle with the avatar. I'm just wondering what's the benefit of this here. I get it very generically that it will improve the performance but I'm trying to understand why we should do these sort of checks here vs making sure that we use that sort of library within the @sanity/ui 

(I feel like this is a stupid question but I'm honestly curious so I'll just push through it 🥲)

packages/sanity/src/core/presence/constants.ts
```diff
@@ -1,6 +1,3 @@
-// disable overlay rendering entirely and only display presence in fields
-export const DISABLE_OVERLAY = false
```
RitaDias, line 2: I have much of the same question re: the other const deletion

RitaDias (commented):
I feel that I'm woefully not the right person to review it, 
however I am honestly curious about some parts and have questions about others 🙏 

Great work anyway Cody and thanks for doing this!

packages/sanity/src/core/changeIndicators/constants.ts
stipsan: The original intent seems to have been to provide an easy way to ship a new version of `sanity` that rolls back the new UI, `git blame` says it was introduced 4 years ago and it's never been set to `false`. It feels like now is the time to delete the unused code :) 

If we were to somehow need it it's still possible to get the same result by commenting out `<ChangeIndicatorsTracker>`: https://github.com/sanity-io/sanity/blob/edc930114e34467dff7969a12039eda134843406/packages/sanity/src/core/changeIndicators/overlay/ChangeConnectorRoot.tsx#L40

packages/sanity/src/core/components/react-track-elements/hooks.ts
stipsan: Yes, this isn't new code, it's copied from the original implementation: https://github.com/sanity-io/sanity/blob/339128dc7fda961dee8deffadd957f5fcacb1a52/packages/sanity/src/core/components/react-track-elements/createStore.ts#L15-L20

I'd like to optimize it further but that's when it started to feel like "maybe we should just use `@floating-ui/react`?" 😅 

Another point is that currently it seems like `store.add` and `store.update` does exactly the same thing. I can change things further if it helps land the PR, just wanted you to know that I stopped because it started to feel like the time would be better spent using a well maintained library off of `npm` instead of rolling our own. It seems like what we wrote 4 years ago have been largely untouched since then, and I'm probably not the only one who's a little scared when trying to change it 😬

packages/sanity/src/core/components/react-track-elements/hooks.ts
stipsan: The comment is talking about the cost of running `isEqual` on `value` on every render, while pointing to `@floating-ui/react`, I get that's confusing since it's two things rolled into one.
Using `@floating-ui/react` to replace just this useEffect wouldn't do anything, but if we used `@floating-ui/react` to power both Presence Overlays, and Change Indicator connectors, then it has a whole pipeline built for doing what we're doing here, in a much more efficient way:
```
1. Register all DOM nodes. (useTrackerStoreReporter)
2. Calc position. (logic that uses the value of `useChangeIndicatorsReportedValues`, or `usePresenceReportedValues`, as input, and schedules a re-render as output)
3. Render with new position.
```
With `@floating-ui/react` we could build up both features with the built-ins for `computePosition()`, `autoPlacement()` and more, instead of rolling our own 🙌

packages/sanity/src/core/components/react-track-elements/hooks.ts
RitaDias: Yeah... I feel this as well!
I hadn't realised that the original file also had these commented. Given the size of the PR and the already commented (same) code I agree with keeping it. Probably should add a comment about floating react here as well just so that we can tackle this in the future... However if this hasn't been prioritised in 2-4 years I'm unsure if / when it will be changed 😬

packages/sanity/src/core/components/react-track-elements/hooks.ts
RitaDias: Might be something worth bringing to @runeb for a follow up? 👁️

stipsan: Sure 😄

packages/sanity/src/core/components/react-track-elements/hooks.ts
stipsan: Allright, I've added a new comment block for it here: https://github.com/sanity-io/sanity/blob/589240450aecac7d84815b3954c8104363a2b0ce/packages/sanity/src/core/components/react-track-elements/hooks.ts#L14-L21 :)

RitaDias (commented):
Otherwise I'd say this looks good! I could approve but I think I'd feel more comfortable if @ricokahler (or someone from DX) had also a look to make sure that I didn't miss anything 🙏

packages/sanity/src/core/changeIndicators/tracker.tsx
```diff
@@ -1,33 +1,77 @@
...
+  value,
+  isEqual?,
+) => {
+  const store = useContext(ChangeIndicatorTrackerContextStore)
+
+  if (store === null) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      new Error(
+        'No context provided for reporter. Make sure that the component calling "usePresenceReporter()", is wrapped inside a <ChangeIndicatorsTracker> element',
```
bjoerge, line 71:
```suggestion
        'No context provided for reporter. Make sure that the component calling "useChangeIndicatorsReporter()", is wrapped inside a <ChangeIndicatorsTracker> element',
```

bjoerge (changes requested):
Fantastic! Think I found a copy + paste issue though

packages/sanity/src/core/changeIndicators/tracker.tsx
stipsan: Thanks! Fixing

stipsan: Fixed!

bjoerge (approved):

PR #6947, opened 2024-06-14
fix(schema): use base type title if no subtype title or name is given

rexxars:
### Description

In certain UI locations, especially for array members, image and files would be listed without a title, or using the type name (`image`) when the member declaration is simply: `{type: 'image'}`.

This PR ensures that we fall back to the base types' title.

I also modified the _object_ to use the same approach, but I think this one is more theoretical, since we require a name for  objects.

### What to review

- Code looks good
- "Add item" menus lists images and files correctly (there's a "Array of multiple types" field that has this in the test studio)

### Testing

Added new tests to ensure no regressions.
Also did some manual testing.

### Notes for release

- Fixes a case where image and file types might be listed in their lowercase variant in array insert menus


binoy14 (approved):
Compared in next vs preview test studio and this looks good. Thanks!

PR #6952, opened 2024-06-17
fix(chore): add aria label to datepicker button

ninaandal:
### Description
Adds missing aria label to the date picker button for accessibility.

### DIFF
--- a/packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
@@ -11,9 +11,9 @@ import {
   useState,
 } from 'react'
 import FocusLock from 'react-focus-lock'
-import {useTranslation} from 'react-i18next'
 
 import {Button, Popover} from '../../../../../ui-components'
+import {useTranslation} from '../../../../i18n'
 import {type CalendarLabels} from './calendar/types'
 import {DatePicker} from './DatePicker'
 import {LazyTextInput} from './LazyTextInput'
@@ -79,14 +79,14 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
   const suffix = (
     <Flex style={{padding: '5px'}}>
       <Button
-        aria-label={t('inputs.datetime.aria-label')}
+        aria-label={calendarLabels.ariaLabel}
         ref={buttonRef}
         icon={CalendarIcon}
         mode="bleed"
         onClick={handleClick}
         style={{display: 'block'}}
         data-testid="select-date-button"
-        tooltipProps={{content: 'Select date'}}
+        tooltipProps={{content: t('inputs.datetime.tooltip-content')}}
       />
     </Flex>
   )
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
```diff
@@ -11,6 +11,7 @@ import {
   useState,
 } from 'react'
 import FocusLock from 'react-focus-lock'
+import {useTranslation} from 'react-i18next'
```
bjoerge, line 14: this should be imported from the `useTranslation` hook defined in this package and not the `react-i18next` one

packages/sanity/src/core/scheduledPublishing/components/dateInputs/base/DateTimeInput.tsx
```diff
@@ -18,6 +18,7 @@ import {
   useState,
 } from 'react'
 import FocusLock from 'react-focus-lock'
+import {useTranslation} from 'react-i18next'
```
bjoerge, line 21: See previous comment

packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
```diff
@@ -48,6 +49,7 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
   const [popoverRef, setPopoverRef] = useState<HTMLElement | null>(null)
   const ref = useRef<HTMLInputElement | null>(null)
   const buttonRef = useRef(null)
+  const {t} = useTranslation()
```
bjoerge, line 52: Instead of using `useTranslation` directly here, could you add a new key to `calendarLabels` and have it be passed in? That way we can keep this component free from the dependency to useTranslation, and it also aligns better with how this component already is localized

bjoerge (changes requested):

bjoerge:
Would be great if this could be translated too as part of this as well: https://github.com/sanity-io/sanity/blob/fa3a666b7461399c577652f94fc60f6a2409c067/packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx#L86

ninaandal:
I think I understood what you meant, so please have a look again :) I see that `scheduledPublishing` isn't translated yet, so I'll create an issue in Linear for this

packages/sanity/src/core/form/inputs/DateInputs/base/DateTimeInput.tsx
```diff
@@ -74,13 +76,14 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
...
       <Button
+        aria-label={calendarLabels.ariaLabel}
         ref={buttonRef}
         icon={CalendarIcon}
         mode="bleed"
         onClick={handleClick}
         style={{display: 'block'}}
         data-testid="select-date-button"
-        tooltipProps={{content: 'Select date'}}
+        tooltipProps={{content: t('inputs.datetime.tooltip-content')}}
```
pedrobonamin, line 86: Sorry for the late review here @ninaandal 
Wondering if this should also be part of the calendar labels, given previous comment https://github.com/sanity-io/sanity/pull/6952/files#r1644370958
There are already other keys in the calendar labels like `selectHour` and `selectMinute`

bjoerge (approved):
✨

PR #6953, opened 2024-06-17
fix(structure): don't replace url when navigating back to list after document deletion

bjoerge:
### Description
Before the fix in #6923 got merged, when deleting a document, you could always hit the back button to get back to the newly deleted document, and there you’d be given a way to undo/restore it. 

Now after this fix, if you delete a document you can no longer hit the back button to recover a deleted document if you realise you made a mistake (unless you remember the id, which you probably don’t). This is not great, and we want to leave a history entry in this case so you navigate back.

### What to review
Deleting a document should allow you to navigate back to restore it.

### Notes for release
n/a since the changes in #6923 has not been released yet


rdunk (approved):
Great, I think this makes a lot of sense.

PR #6956, opened 2024-06-17
refactor(core): change revision history label from 'Version' to the more accurate 'Revision'

bjoerge:
### Description

This makes our labeling of revision to be more consistent. We are calling it revision in someplaces but not everywhere. Also updated the code to be more consistent too.

### What to review

- Is the way I've deprecate the old i18n key ok?
- Would it be better to leave the key as-is and instead just change the label?

### Testing
Don't think it makes all that much sense to add automated testing for this one

### Notes for release

n/a - should just make the studio terminology more consistent.

### DIFF
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1580,7 +1580,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'timeline.latest-revision': 'Latest revision',
   /**
    * Label for latest revision for timeline menu dropdown
-   * @deprecated use `timeline.latest-revision` instead
+   * @deprecated as of `v3.47.0` `timeline.latest-revision` should be used instead. Note: _usage_ of this key is deprecated, but Studios on `< v3.47.0` still require this key to be _defined_
    * */
   'timeline.latest-version': 'Latest revision',
   /** The aria-label for the list of revisions in the timeline */
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1575,8 +1575,14 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'timeline.error.unable-to-load-revision': 'Unable to load revision',
   /** Label for when the timeline item is the latest in the history */
   'timeline.latest': 'Latest',
-  /** Label for latest version for timeline menu dropdown */
-  'timeline.latest-version': 'Latest version',
+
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1575,8 +1575,14 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
...
-  /** Label for latest version for timeline menu dropdown */
-  'timeline.latest-version': 'Latest version',
+
+  /** Label for latest revision for timeline menu dropdown */
+  'timeline.latest-revision': 'Latest revision',
+  /**
+   * Label for latest revision for timeline menu dropdown
+   * @deprecated use `timeline.latest-revision` instead
+   * */
+  'timeline.latest-version': 'Latest revision',
```
binoy14, line 1585: defer to @rexxars on deprecating i18n keys. Not sure we have a process

bjoerge: There's no clear process, but we can make this an opportunity to establish one?

This was a fair amount of OCD-induced change 😅, but I still feel the change it's justified by the likelyhood of existing language bundles already having translated "version" instead of "revision" (for example, the [norwegian bundle](https://github.com/sanity-io/locales/blob/main/locales/nb-NO/src/studio.ts#L1542) is currently using "Version" here, when it should be "Revisjon"), and this change will give downstream locale implementors a chance to notice the change.

I've pushed some added clarity in the comments in 4864109d596d9d9757b8e5a2caacc212a78d2837 based on input from @rexxars.

binoy14: understood, I am okay with the change just wanted to call it out

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1575,8 +1575,14 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'timeline.error.unable-to-load-revision': 'Unable to load revision',
   /** Label for when the timeline item is the latest in the history */
   'timeline.latest': 'Latest',
-  /** Label for latest version for timeline menu dropdown */
-  'timeline.latest-version': 'Latest version',
+
+  /** Label for latest revision for timeline menu dropdown */
+  'timeline.latest-revision': 'Latest revision',
```
binoy14, line 1580: @bjoerge I don't see anywhere in the code `timeline.latest-revision` key being used. Is that expected?

binoy14 (approved):
makes sense to me, thanks!

PR #6961, opened 2024-06-17
tets: make test suite run on windows

runeb:
### Description

Currently the test suite does not run on windows due to jest config files not being loaded properly, so that typescript transpilation to javascript does not happen.

Additionally, `pnpm build` would not have any packages in scope due to the formatting of filters in npm scripts (see https://github.com/vercel/turbo/issues/928).

### What to review

That the test suite now runs on Windows
That there are no regressions in other environments


bjoerge:
@runeb @ricokahler what remains to do here?

PR #6962, opened 2024-06-17
feat(typegen): add all schema types exported union

sgulseth:
### Description

Adds a union that exports all the potential schemas to the generated file. This can be useful in mocks/testing environments and used by third party tools to get a list of potential types.

### What to review

Name of the exported variable `allSchemaTypes`, should it be prefixed with anything? `_`? Anything else to indicate that this is a "special" type?

### Testing

Already test schema generation, so updated the snapshots

### Notes for release

TypeGen: Added a type export of all the the discovered Sanity Schema variants.

### DIFF
--- a/packages/@sanity/codegen/src/typescript/__tests__/__snapshots__/typeGenerator.test.ts.snap
+++ b/packages/@sanity/codegen/src/typescript/__tests__/__snapshots__/typeGenerator.test.ts.snap
@@ -208,7 +208,9 @@ export type SanityImagePaletteSwatch = {
   population?: number;
   title?: string;
   _type: \\"sanity.imagePaletteSwatch\\";
-};"
+};
+
+export type AllSanitySchemaTypes = Author | Post | Ghost | BlockContent | SanityAssetSourceData | Slug | Geopoint | SanityImageAsset | SanityFileAsset | SanityImageCrop | SanityImageHotspot | SanityImageMetadata | SanityImageDimensions | SanityImagePalette | SanityImagePaletteSwatch;"
 `;
 
 exports[`generateSchemaTypes should generate correct types for document schema with inline fields 1`] = `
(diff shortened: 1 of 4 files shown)


stipsan (dismissed):
I'm fine with the `_` prefix to signal it's private :) Alternatively we could use `Sanity` to prefix it. I understand that we use the pattern of mapping `type: "sanity.imageHotspot"` to `SanityImageHotspot` and don't want to have a future collision. I don't think there's a risk for that with `SanitySchemaTypes` 😄

sgulseth:
> I'm fine with the `_` prefix to signal it's private :) Alternatively we could use `Sanity` to prefix it. I understand that we use the pattern of mapping `type: "sanity.imageHotspot"` to `SanityImageHotspot` and don't want to have a future collision. I don't think there's a risk for that with `SanitySchemaTypes` 😄 
> 
> 

SanitySchemaTypes 🤝

stipsan (approved):
🥳

packages/@sanity/codegen/src/typescript/__tests__/__snapshots__/typeGenerator.test.ts.snap
```diff
@@ -208,7 +208,9 @@ export type SanityImagePaletteSwatch = {
   population?: number;
   title?: string;
   _type: \\"sanity.imagePaletteSwatch\\";
-};"
+};
+
+export type AllSanitySchemaTypes = Author | Post | Ghost | BlockContent | SanityAssetSourceData | Slug | Geopoint | SanityImageAsset | SanityFileAsset | SanityImageCrop | SanityImageHotspot | SanityImageMetadata | SanityImageDimensions | SanityImagePalette | SanityImagePaletteSwatch;"
```
scottrippey, line 213: ❤️ this!  Thank you!

PR #6964, opened 2024-06-17
feat: utilize timestamps in build and renderDocument for AUS

ricokahler:


packages/sanity/src/_internal/cli/server/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+  const { imports = {}, ...rest } = JSON.parse(document.getElementById('__imports').textContent);
+  const importMapEl = document.createElement('script');
+  importMapEl.type = 'importmap';
+  importMapEl.textContent = JSON.stringify({
+    imports: Object.fromEntries(
+      Object.entries(imports).map(([specifier, path]) => {
+        try {
+          const url = new URL(path);
+          if (/^sanity-cdn\\.[a-zA-Z]+$/.test(url.hostname)) {
+            const newTimestamp = \`t\${Math.floor(Date.now() / 1000)}\`;
```
rexxars, line 23: Can you hoist this to the top (out of the loop) to ensure that the timestamp remains the same for all the entries? There's a theoretical chance that these will be different by one second as it stands.

packages/sanity/src/_internal/cli/server/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+ * and injects the modified import map back into the HTML.
+ *
+ * This will be injected into the HTML of the user's bundle.
+ *
+ * Note that this is in a separate constants file to prevent "Cannot access
+ * before initialization" errors.
+ */
+export const TIMESTAMPED_IMPORTMAP_INJECTOR_SCRIPT = `<script>
+  // auto-generated script to add import map with timestamp
+  const { imports = {}, ...rest } = JSON.parse(document.getElementById('__imports').textContent);
```
rexxars, line 14: Can we gracefully handle a missing imports script? Edge case, but figure it's worth the few extra bytes

```suggestion
  const importsScript = document.getElementById('__imports');
  const { imports = {}, ...rest } = importsScript ? JSON.parse(importsScript.textContent) : {};
```

packages/sanity/src/_internal/cli/server/constants.ts
```diff
@@ -0,0 +1,35 @@
...
+  const importMapEl = document.createElement('script');
+  importMapEl.type = 'importmap';
+  importMapEl.textContent = JSON.stringify({
+    imports: Object.fromEntries(
+      Object.entries(imports).map(([specifier, path]) => {
+        try {
+          const url = new URL(path);
+          if (/^sanity-cdn\\.[a-zA-Z]+$/.test(url.hostname)) {
+            const newTimestamp = \`t\${Math.floor(Date.now() / 1000)}\`;
+            url.pathname = url.pathname.replace(/t\\d+/, newTimestamp);
```
rexxars, line 24: Might be a bit nitpicky, but we could make this a bit safer by ensuring it replaces a whole URL segment? (Match a leading slash, and ensure the numbers are either the end of the path name, or that they are followed by a slash)

```suggestion
            url.pathname = url.pathname.replace(/\/t\\d+($|\/)/, newTimestamp);
```

scripts/uploadBundles.ts
```diff
@@ -122,7 +124,7 @@ async function updateManifest(newVersions: Map<string, string>) {
         [dirName]: {
           ...initial.packages[dirName],
           default: value,
-          versions: [...(initial.packages[dirName]?.versions || []), value],
+          versions: [...(initial.packages[dirName]?.versions || []), {version: value, timestamp}],
```
rexxars, line 127: Is the server ready to handle both the new and old format?

rexxars (changes requested):

scripts/uploadBundles.ts
binoy14: yes, we back ported the timestamps to the manifest and updated the server

binoy14 (changes requested):
Could you change from `cacheControl: 'public, max-age=10',` to `cacheControl: 'public, max-age=0',` in the manifest upload line 141 in `scripts/uploadBundles.ts`

rexxars:
> Could you change from `cacheControl: 'public, max-age=10',` to `cacheControl: 'public, max-age=0',` in the manifest upload line 141 in `scripts/uploadBundles.ts`

Would suggest `no-cache, max-age=0`

rexxars (approved):

binoy14 (approved):
Tested it with test studio, looks good thanks!

PR #6967, opened 2024-06-18
test(e2e): rewrite e2e toolbar tests to component tests & update e2e expectations on others

RitaDias:
### Description

- Converted e2e Toolbar tests to component tests
  - realised that the tests didn't necessarily need to be e2e and so as component tests they are more reliable and test the exact same usecases 

This is likely not the end (I'm assuming another one of these PRs is in the cards in the future), but since the toolbar tests don't need to be e2e, it's a good thing to have updated as well as some more guardrails in other tests.

### Notes for release

N/A


rexxars (approved):
Thanks for the refactors, this is great!

PR #6968, opened 2024-06-18
chore(ci): upgrade pnpm setup to use node20

sjelfull:
### Description

This upgrades `pnpm/action-setup@v2` that uses Node 16 to v4 that uses Node 20. This gets rid of the deprecation warnings in GH actions. This version is also more strict and requires you to install the exact version if you have defined `packageManager` in package.json.

### What to review

Any downsides to upgrading? I know we do tests against Node 18 as well, but don't think this affects that in any way?

### DIFF
--- a/.github/workflows/are-we-compiled-yet.yml
+++ b/.github/workflows/are-we-compiled-yet.yml
@@ -22,7 +22,6 @@ jobs:
         name: Install pnpm
         id: pnpm-install
         with:
-          version: 8.15.8
           run_install: false
 
       - name: Get pnpm store directory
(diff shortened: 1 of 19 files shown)


.github/workflows/are-we-compiled-yet.yml
```diff
@@ -18,11 +18,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - uses: pnpm/action-setup@v2
+      - uses: pnpm/action-setup@v4
         name: Install pnpm
         id: pnpm-install
         with:
-          version: 8
+          version: 8.15.8
```
stipsan, line 25: It looks like this [field is optional](https://github.com/pnpm/action-setup/blob/23657c8550e9be4f7ee7e9d7fe8adda6d703bb4e/README.md?plain=1#L11) since we have `packageManager` specified in `package.json` 🤔  
```suggestion
```

stipsan (approved):

PR #6972, opened 2024-06-18
feat(structure): translations for sheet list

jordanl17:
### Description
Adding in translations for strings in sheet list views

### What to review
That the studio i18n file has been updated with the correct keys, and these are used in various sheet list components

### Notes for release
No relevant notes - feature not visible to users


RitaDias (approved):
🌍 🚀

rexxars:
Please hold this PR for now - this will have some downstream effects that I am not sure we have accounted for - will follow up internally

packages/sanity/src/structure/panes/documentList/sheetList/i18n/resources.ts
```diff
@@ -0,0 +1,58 @@
+/**
```
pedrobonamin, line 1: Curious about this file colocation, added a question internally, I think it should be part of `packages/sanity/src/structure/i18n/resources.ts` but let's confirm before doing any change

PR #6985, opened 2024-06-18
chore(deps): upgrade turbo to v2

stipsan:
### Description

Upgrades turbo to v2 following these steps: https://turbo.build/repo/docs/crafting-your-repository/upgrading#upgrading-to-20
As well as adding the `eslint-config-turbo` to find all env vars that were used but not declared.

### What to review

Some env vars are in `globalPassThroughEnv`, which allows using them but they won't invalidate the cache. Most are in `globalEnv` to ensure that cache is rather invalidated more frequently than it needs to instead of sometimes caching the wrong thing. Let me know if any of the env vars should be shifted between `globalEnv` and `globalPassThroughEnv`.

### Testing

When testing you might need to nuke `node_modules` and then run `pnpm install` to properly get the new turbo v2 binaries installed. In my case I was getting weird errors about not recognizing `tasks` keys in `turbo.json` files as I was still running v1.
You can verify by `pnpm turbo --version` after you `pnpm install`.

### Notes for release

N/A


binoy14 (approved):
Tested it locally and CI passes, thanks for doing this!

PR #6986, opened 2024-06-18
fix(migrate): fixes error message to be more helpful

binoy14:
### Description

Seems like the error response from backend is in the shape below

```ts
{ error: { type: string, description: string }
```

When we receive error in this shape we don't properly log that message making it hard for folks to debug. See screenshots below 

#### Before
Screenshot 2024-06-18 at 2 49 06 PM

#### After
Screenshot 2024-06-18 at 2 49 32 PM

### What to review

Copy and change makes sense, I checked the repo and there seems to be other places checking `error.description` so I am not sure if error will not be an object but I added a conditional to keep backwards compatibility

### Testing

Updated the test

### Notes for release

- Migration error messages provide more details

### DIFF
--- a/packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
+++ b/packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
@@ -42,3 +42,22 @@ test('server responds with 5xx and no json response', () => {
     message: 'HTTP Error 500: Internal Server Error',
   })
 })
+
+test('server responds with 5xx and json response', () => {
+  const mockResponse = {
+    status: 500,
+    statusText: 'Internal Server Error',
+    json: () =>
+      Promise.resolve({
+        error: {
+          type: 'validationError',
+          description: 'Document is not of valid type',
+        },
+        status: 500,
+      }),
+  }
+  expect(assert2xx(mockResponse as unknown as Response)).rejects.toThrowError({
+    statusCode: 500,
+    message: 'validationError: Document is not of valid type',
+  })
+})
--- a/packages/@sanity/migrate/src/fetch-utils/fetchStream.ts
+++ b/packages/@sanity/migrate/src/fetch-utils/fetchStream.ts
@@ -18,9 +18,17 @@ export async function assert2xx(res: Response): Promise<void> {
   if (res.status < 200 || res.status > 299) {
     const jsonResponse = await res.json().catch(() => null)
 
-    const message = jsonResponse?.error
-      ? `${jsonResponse.error}: ${jsonResponse.message}`
-      : `HTTP Error ${res.status}: ${res.statusText}`
+    let message: string
+
+    if (jsonResponse?.error) {
+      if (jsonResponse?.error?.description) {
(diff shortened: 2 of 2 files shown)


binoy14:
* **#6986** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/6986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

packages/@sanity/migrate/src/fetch-utils/fetchStream.ts
```diff
@@ -18,9 +18,17 @@ export async function assert2xx(res: Response): Promise<void> {
...
+    let message: string
+
+    if (jsonResponse?.error) {
+      if (jsonResponse?.error?.description) {
+        message = `${jsonResponse?.error?.type || res.status}: ${jsonResponse.error.description}`
+      } else {
+        message = `${jsonResponse.error}: ${jsonResponse.message}`
+      }
+    } else {
+      message = `HTTP Error ${res.status}: ${res.statusText}`
```
ricokahler, line 30: Could be nice to refactor to `res.text()` then put the text here in case the response isn't json

something like

```js
const textResponse = await res.text()
let jsonResponse
try {
  jsonResponse = JSON.parse(textResponse)
} catch {
  jsonResponse = null
}

// use textResponse etc...
```

packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
```diff
@@ -42,3 +42,22 @@ test('server responds with 5xx and no json response', () => {
...
+    statusText: 'Internal Server Error',
+    json: () =>
+      Promise.resolve({
+        error: {
+          type: 'validationError',
+          description: 'Document is not of valid type',
+        },
+        status: 500,
+      }),
+  }
```
ricokahler, line 58: If our env supports the response constructor, this may be nicer

```suggestion
  const mockResponse = new Response(
    JSON.stringify({
      error: {
        type: 'validationError',
        description: 'Document is not of valid type',
      },
      status: 500,
    }),
    {
      status: 500,
      statusText: 'Internal Server Error',
    },
  )
```

ricokahler (approved):
had a few ✨ suggestions but this is already better so ✅

packages/@sanity/migrate/src/fetch-utils/fetchStream.ts
binoy14: hmm in terms of lines of code that would be the same right? I don't see a reason API responds with text instead of JSON 🤔

ricokahler: yeah probably over-thinking it

packages/@sanity/migrate/src/fetch-utils/__test__/assert2xx.test.ts
binoy14: probably worth doing it but maybe in a follow up PR, would need to change a few tests

PR #6989, opened 2024-06-19
fix(@sanity): _regenerateKeys so that the MarkDefs left are only those allowed

RitaDias:
### Description

Add fix for when PTE text is pasted from one input to another that it respects the allowed annotations

Before

https://github.com/sanity-io/sanity/assets/6951139/a647261b-aab8-4476-ac13-8e647528d5fe

### Testing

The added tests should be sufficient
- Added tests to `_regenerateKeys` (test before normalization)
- The normalization tests already test for the "after normalization" -> `preserves other marks that apply to the spans`

While the original issue was detected on copy paste, I don't think it is necessary to add a test to target that use case in particular. We would have seen this as an issue when creating a doc via the terminal for example.

### Notes for release

Fixes issue where when PTE text was copy pasted always kept the original annotations regardless of the receiving PTE customization


skogsmaskin:
@RitaDias - I'm wondering if it makes more sense to build you code into the [`normalizeBlock` function](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/block-tools/src/util/normalizeBlock.ts#L39) in `@sanity/block-tools`. It's already doing the same for decorator marks, and I think your code would be a great addition to that function.

RitaDias:
> Member

Just had a chat with @skogsmaskin and it's best to keep it as it is since copy / pasting from PTE to anoter PTE doesn't run the [normalizeBlock](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/block-tools/src/util/normalizeBlock.ts#L39)

christianhg:
@RitaDias , any chance I could get you to open this PR against https://github.com/portabletext/editor instead? Then we don't have to port these changes to that repo/package after #7001 has been merged.

Sorry for the inconvenience :(

RitaDias:
> @RitaDias , any chance I could get you to open this PR against https://github.com/portabletext/editor instead? Then we don't have to port these changes to that repo/package after #7001 has been merged.
> 
> Sorry for the inconvenience :(

I'll do that :)

RitaDias:
Closing this, re-opened it on https://github.com/portabletext/editor/pull/19 :)

PR #6990, opened 2024-06-19
fix(cli): correct noop return value on esbuild-register in dev

rexxars:
### Description

Saw this randomly on [this test-next-studio build](https://vercel.com/sanity-io/test-next-studio/GmsQegGwVvZBMssTnGz94fYEwtDY?filter=errors) - in some cases the unregistration of the esbuild-register instance failed. After some investigation, it seems like we're returning a function where we should be returning an object.

Not quite sure why this popped up _now_, but worth fixing.

### Testing

- test-next-studio should build without errors (CI)

### Notes for release

None


RitaDias (approved):
Great catch!

PR #6991, opened 2024-06-19
fix(core): make version check not depend on `process.env`

rexxars:
### Description

We recently introduced a version check for auto-updating studios that depend on `process.env` for telling if we're in staging or not - this doesn't work nicely across all environments, as some of them use `import.meta.env`, and others use `process.env`. To make things worse, we can't easily check for `process.env` and use `process.env[envVar]` as bundlers define these as "hard replacements" instead of actually defining an object.

This replaces the check with the same approach as we do in [createAuthStore](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts#L147-L151), relying on the somewhat magic `__SANITY_STAGING__`  global defined in our [vite config](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/_internal/cli/server/getViteConfig.ts#L120).

This should be safe, as we only allow auto-updates on our own bundled studios for now.

### What to review

- Studio still runs as expected (does not crash when checking for the staging global)
- Still uses staging/production correctly (have tested this myself and it does work)

### Testing

We should add a test for this, but it needs to be run in a specific environment that does not have `process.env` defined, and we currently don't have an easy way of doing so. I'll make a note that we need to add some tests for multiple environments later.

### Notes for release

- Fixes a potential crash in certain javascript environments when trying to access `process.env`

### DIFF
--- a/packages/sanity/src/core/studio/packageVersionStatus/checkForLatestVersions.ts
+++ b/packages/sanity/src/core/studio/packageVersionStatus/checkForLatestVersions.ts
@@ -1,14 +1,14 @@
-//object like {sanity: '3.40.1'}
+// object like {sanity: '3.40.1'}
 interface VersionMap {
   [key: string]: string | undefined
 }
 
-//e2e tests also check for this URL pattern -- please update if it changes!
+// @ts-expect-error: __SANITY_STAGING__ is a global env variable set by the vite config
+const isStaging = typeof __SANITY_STAGING__ !== 'undefined' && __SANITY_STAGING__ === true
+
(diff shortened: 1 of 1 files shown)


rexxars:
> @rexxars Question on how to test this in a studio, do you know off the top of my head how I could trigger "does not crash when checking for the staging global". Is it as simple as opening the studio or do I need to run the env on staging?

The code is run in the top level module scope, so it should (_or rather, shouldn't 😅_) crash upon loading the module - eg when loading the studio

RitaDias (approved):
Works as expected and no crashes!

PR #6996, opened 2024-06-19
feat(structure): sort sheet list

jordanl17:
### Description
Adds support to sheet list to use the document list sort order when ordering the documents in the sheet

### What to review
* Logic used in `DocumentListPane` to control the retrieval of ordering has been refactored so it can be reused across `SheetList` also
* `useDocumentSheetList` now applies the active order to the search query, and sorts the returned documents according to their ordering

### Testing
Testing for the sort ordering in `useDocumentSheetList`

### Notes for release
No relevant notes - this is not yet customer facing


packages/sanity/src/structure/panes/documentList/sheetList/DocumentSheetListPane.tsx
```diff
@@ -265,65 +275,64 @@ function DocumentSheetListPaneInner(
...
-              <tbody>{table.getRowModel().rows.map(renderRow)}</tbody>
-            </Table>
-          </DocumentSheetListProvider>
-          <DocumentSheetActions table={table} schemaType={documentSchemaType} />
-        </TableContainer>
-        <Flex justify={'flex-end'} padding={3} gap={4} paddingY={5}>
-          <DocumentSheetListPaginator table={table} />
-        </Flex>
-      </React.Fragment>
+      <Table>
```
jordanl17, line 289: Reordering of the content behind the loader, so that it's only the table, and not also the filter bar, that is hidden until loaded

packages/sanity/src/structure/panes/documentList/PaneContainer.tsx
```diff
@@ -112,7 +103,7 @@ export const PaneContainer = memo(function PaneContainer(
 
   const isSheetListLayout = layout === 'sheetList'
   const paneLayout = isSheetListLayout ? (
-    <DocumentSheetListPane {...props} key={props.pane.id} />
```
pedrobonamin, line 115: I think we need to restore this `key` in here, it was added to reset all the table states and effects when switching from one document type to another, forcing the component to re mount.

packages/sanity/src/structure/panes/documentList/useDocumentListSort.ts
```diff
@@ -0,0 +1,29 @@
...
+export const useDocumentListSort = (pane: DocumentListPaneNode, sortOrderRaw?: SortOrder) => {
+  const schema = useSchema()
+  const {options} = pane
+  const {filter, params: paneParams} = options
+  const params = useShallowUnique(paneParams || EMPTY_RECORD)
+  const typeName = useMemo(() => {
+    const staticTypes = findStaticTypesInFilter(filter, params)
+    if (staticTypes?.length === 1) return staticTypes[0]
+    return null
+  }, [filter, params])
```
pedrobonamin, line 18: I wonder if we need this in here or we could refactor this to receive the `typeName` as a parameter. 
I think `typeName` is accessible and already calculated in both `DocumentSheetList` and in `DocumentListPane` where this hook is used.

pedrobonamin (changes requested):
This looks great, thanks for the change! 
Added some comments, open to discussion

packages/sanity/src/structure/panes/documentList/useDocumentListSort.ts
jordanl17: The `typename` used in `DocumentListPane` could be any typename though? That list supports rendering multiple different types, so I'm unsure how to replicate this there. I do agree regarding `DocumentSheetList` however

packages/sanity/src/structure/panes/documentList/PaneContainer.tsx
jordanl17: Ah that's fair - I didn't realise! I'll reinstate

- [ ] Add `key` back to `DocumentSheetListPane`

packages/sanity/src/structure/panes/documentList/useDocumentListSort.ts
pedrobonamin: sorry for the late reply on this.
The `sortWithOrderingFn` in the hook and in the previous implementation in `DocumentSheetList` uses the `typeName` which results in the first staticType found.

So I was thinking, given DocumentListPane is already resolving the typeName [here](https://github.com/sanity-io/sanity/blob/edx-1461-sort-sheet-list/packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx#L49-L54) 

```ts 
// DocumentListPane:69
 const sortWithOrderingFn = useDocumentListSort(typeName, sortOrderRaw)

// DocumentSheetListPane:146
const sortWithOrderingFn = useDocumentListSort(documentSchemaType.name, sortOrderRaw)

```

And then `useDocumentListSort` will only require this:
```ts
export const useDocumentListSort = (typeName: string | null, sortOrderRaw?: SortOrder) => {
  const schema = useSchema()

  const sortWithOrderingFn = useMemo(
    () =>
      typeName && sortOrderRaw
        ? applyOrderingFunctions(sortOrderRaw, schema.get(typeName) as any)
        : sortOrderRaw,
    [schema, sortOrderRaw, typeName],
  )

  return sortWithOrderingFn
}
```

jordanl17: dohh! Right you are - that makes sense. I hadn't realised I'd now duplicated this logic. Thanks for realising!

- [ ] Simplify `useDocumentListSort` by just passing in `typename` directly

pedrobonamin (approved):
Thanks @jordanl17 ! Great work on this !

Issue #6934, opened 2024-06-14
Automatically infer types of queries using groq and @sanity/codegen


romeovs:
**Is your feature request related to a problem? Please describe.**
Currently, I'm using `groq` tagged templates in my codebase in tandem with `sanity typegen generate` to generate types. This works well.

```ts
const example = groq`
  *[_type == 'post'] {
    title
    author-> {
      name
    }
  }
`
```

However, to make use of the generated type, I need to manually add the generated type annotations to `client.fetch` etc. to make use of them:

It would be nice if we could automatically type `groq` queries without having to manually add the generated type annotations:

```ts
import { ExampleQueryResult } from "./generated"
const res = await client.fetch<ExampleQueryResult>(example)
```

It would be nice if we could automatically type the `query` variable with the expected result and variable types using a phantom type like so:

```ts
type TypedQuery<Res> = string & {
  _result: ExampleQueryResult
}

const example: TypedQuery<ExampleQueryResult>
```

that way we can automatically infer the type of the query and we don't need the manual type annotation:

```ts
function inferredQuery<T>(query: TypedQuery<T>): Promise<T> {
  return client.fetch(query)
}
```

and we can just use it like so:
```ts
// res is typed as ExampleQueryResult here
const res = await inferredQuery(example)
```

_Note: I've left out query parameters here for simplicity, but it would be easy to add them to `TypedQuery`_

To facilitate this, we would need a couple of things.

#### 1. Introduce a `TypedQuery` type
Introduce a new type `TypedQuery` that describes a query and with metadata like the return type and variables:
```ts
type TypedQuery<R, P> = string & {
  _result?: R
  _params?: P
}
```
This is a phantom type so does not impact runtime performance.

#### 2. `@sanity/codegen` should record a map of all query strings
`@sanity/codegen` should generated a map that uses query strings as keys and has `TypedQuery` as values. Eg. in the example given above the map would look like:

```ts
// the generated type for the query result, this functionality already exists:
export type ExampleQueryResult = {
  title: string | null
  author: {
    name: string | null
  } | null
} | null


export type QueryTypes = {
  "\\n  *[_type == 'post'] {\\n    title\\n    author-> {\\n      name\\n    }\\n  }\\n":
    TypedQuery<ExampleQueryResult>
}
```

#### 3. Return a `TypedQuery` from the `groq` wrapper
Using the type map, is now possible to return a `TypedQuery` from `groq`.

[...]

PR #6997, opened 2024-06-19
feat(codegen): generate SanityQueries interface in @sanity/codegen

romeovs:
### Description
This PR is one of two PR's that implements my feature request (https://github.com/sanity-io/sanity/issues/6934) in `sanity-io/sanity`, with the help of some pointers from @sgulseth.

The companion PR that implement the corresponding `SanityQuery` in `@sanity/client` lives at https://github.com/sanity-io/client/pull/858.

This PR implements a couple of things.

#### 1. Add a `defineQuery` helper to the `groq` package
 
This helper can be used instead of the `groq` template tag.

This is necessary because TypeScript is not capable of strictly typing tagged template strings (yet, see https://github.com/microsoft/TypeScript/issues/33304) and we need the resulting query to be typed as the literal string rather than `string`.

For example, in the following snippet, `query` is typed as a generic `string`.
```ts
import {groq} from 'groq'

const query = groq`*[_type = 'foo']`
```

While in the following code snippet, `query` is typed as `"*[_type = 'foo']"` which is more specific.
```ts
import {defineQuery} from 'groq'

const query = defineQuery(`*[_type = 'foo']`)
```
We will need this specificity later to type the result of `SanityClient.fetch`.

~_Note:_ Ideally `defineQuery` would live in the top-level `groq` package, but because `groq` already has a default export (the `groq` template tag) this is not possible. As far as I can tell is impossible to safely mix default exports with named exports when transpiling to CommonJS without forcing the users of the (CommonJS version of the) library to change their setup. Adding `defineQuery` as a named export of `groq` was breaking [this assertion](https://github.com/sanity-io/sanity/blob/next/packages/groq/test/groq.test.cjs#L8) since `groq` moved to the `.default` property on the import.~

~Instead, to preserve full backwards compatibility, I have put `defineQuery` in `groq/define`.~
Thanks to input from @stipsan, we can export `defineQuery` as a named export from `groq` 🙌

#### 2. Find `defineQuery` queries in `@sanity/codegen`
Allows `@sanity/codegen` to find and parse queries that are defined by the new `defineQuery` wrapper. `groq`-tagged queries are still detected and used as before too.

#### 3. Generate `SanityQueries` interface
Add `generateTypeMap` to `TypeGenerator` that generates the type augmentation for `@sanity/client`'s `SanityQueries`:

```
declare module '@sanity/client' {
  interface SanityQueries {
    "*[_type = 'foo']": FooResult
    /* ... */
  }
}
```

This feature is currently disabled by default and can be enabled by setting `"overloadClientMethods": true` in `'sanity-typegen.json`.

#### 4. Update `@sanity/cli`

Update `@sanity/cli` to make use of the new utility function to generate the type map.

### What to review

- Build and run the `codegen` command in `@sanity/cli` and make sure the resulting generated code matches expecations.
- I could use some input on the `groq` vs `groq/define` question (see above). It might be possible to get some `defineQuery` in `groq` using [`esModuleInterop`](https://www.typescriptlang.org/docs/handbook/modules/appendices/esm-cjs-interop.html), but I'm not sure about the assumptions Sanity makes about the users of that package. `pkg-utils` does not seem to support it currently.

### Testing

- I added unit test for all the type generation utilities and for the new `defineQuery` helper.
- There are currently no tests in the repo for the `codegen` cli action, but I'm happy to help set those up if required.

### Notes for release

There is a breaking change in the `@sanity/codegen` package: `TypeGenerator.generateTypeNodeTypes` now returns `{ type: string, typeName: string }` instead of `string`. I figured that since `@sanity/codegen` is still in beta, that change would be safe to make. I've updated all the call sites (only one) in the repo to match the change.

I have left some inline notes too to clarify some decisions.

There is also a [TS Playground](https://www.typescriptlang.org/play/?jsx=0#code/PTAEBUAsEsGdTqA7pAhgF2dANt0AjAU1ABNCAzaAO0JIStAAFZUrp0BPYAY22kKqZU5dIQBOodJGIs2nALTQA9jz4D0AYgAcAVi2gA5oXTwAtuKMkAUGV6oxxU0pIBXbMQDkzVuy69+gh6gAN5WoPSiYuSo3MQAyj6cAIou4vzwwQC+VmGghAAeAA5KYpichcQAwmqCAErGLmJUADy54QDiYkoAjnHoYtQGefmiVCTwsP2DADRtoABiqLj4MQDWoAC8oKwcs+EAfJugnT19A1RDBaPjoKuEHErkoAlyHClphPAA-M+Jb6kDT4AbROvSmFwAuqAAFwLJbYFbcVY5cIFYqlCLiaKxX6vaoBTChcLhcjGbiQZq1I47aagbhKKiTUBJYbXCbgoZbSbnAz7AAU3QBHFhSQAlLCAApdUxwQjNfHqeroRotJK02r7fa5bLZKwgCAweAGJSfCJKQxdbo2Qh2BygJyudygABEBktzpCuXILio3HQygYZEoNHeYg4zXpjMw4FZAhu3MG-MF4mFEHFEE9xNADmVTVAybD2qsuv1UEQRhMhgE4gwtAIHCYsl8PGchCMDD5Ui6LgMkEk0kbf1U0FFBFQsDrDP7xALDaI2CUSD1YHQSgr0gkSHYfak8T+oFg5MIplQVnKxAAsv8U-VYG5MFtgqAAPrn2EechKJRBTJAiE5WxsHsRxnDcTxvFeYd1CCIlMSiGI91eUN0kzYkPAAKiBT8lAhDxYSvUMOFve8ixLMAoGIFwJwkbhWFAKhF1AKjiF3fMhQPYRCGwBtqG2SQOAqOgkFQDgrGgUx0UJOkHFrBVBFATJQHIaVQC8JtOCgwIxIkkopKDahCEIhSlJU113RySMmX8dQjm4GTRDk9A+SyUUcmXe1rzDBB4H0mg6FgJRzBQcRiF41jmLEWAgnpMgVgnKxLMwUxPIbLZfMMoU+QwrCv1w1z3LLeAHDvbBMEQc9rES-jBNvI4+VQYT2DpGp0AAOlJdByT5ZLCNFUcWH9WBKFNAihWI0qcn1ABJTAHBPah4Aq7Z4AXC5logATCDiOzoEKTB6rGQdIJitsBGQaQGHbGt-QuUdaIYahSQkVjZwPDk3owaBaNwUSquU1ADHMeStg-L8PAShkmWSgB5XcxCMtKKAMwi+QAA0wgASYJ-sB9RMghVH+s+oaUOy7DcIhqNQCUOHaq2erGswazBHaslIG6jhYY3XqicG4b4FGm9PhI9yAFVYEGadmoJUA+DuAgKBKGR-VwZASlWWBaS3KQlBcWbCEFaAeTpVs6TQC5PkppkfQqunQAZ1AmuZtqOq6jCPD6ji+ZQnYrcwJQADdxAXVAyBIO2HadlrWc6iknxWMRYSoFxTCICRMn5d3PYGuB+dQhOk5TtPiysIA) that summarizes the changes made by this PR and the companion PR.


packages/@sanity/codegen/src/typescript/typeGenerator.ts
```diff
@@ -82,16 +87,19 @@ export class TypeGenerator {
    * @internal
    * @beta
    */
-  generateTypeNodeTypes(identifierName: string, typeNode: TypeNode): string {
+  generateTypeNodeTypes(
+    identifierName: string,
+    typeNode: TypeNode,
+  ): {type: string; typeName: string} {
```
romeovs, line 93: To reuse the types later (like when generating the type map), we need to be able to reference them and for that we need the actual type name that was generated.

This change makes `generateTypeNodeTypes` return the generated names so we can reuse them.

packages/groq/src/define.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
romeovs, line 1: As explained in the PR description, we cannot add `defineQuery` to the top-level `groq` package since that might break CommonJS users of the package.

packages/@sanity/codegen/src/typescript/__tests__/typeGenerator.test.ts
```diff
@@ -426,3 +426,48 @@ export type AllSanitySchemaTypes = OptionalData;"
...
+
+    expect(actualOutput).toMatchInlineSnapshot(`
+"declare module \\"@sanity/client\\" {
+  interface SanityQueries {
+    \\"*[_type \\\\\\"author\\\\\\"]\\": AuthorsResult;
+    \\"*[_type \\\\\\"author\\\\\\"][0]\\": FirstAuthorResult;
+  }
+}"
+`)
+  })
```
romeovs, line 472: @sgulseth note that I opted not to `extend QueryReturn` here. In my local testing it was not necessary and overriding `SanityQueries` just works.

stipsan (changes requested):
It's possible to add `defineQuery` top-level without breaking CJS, I pushed a branch demoing it here: https://github.com/sanity-io/sanity/tree/bc-groq-define-query/packages/groq

The key is to:
1. remove the `source` field, and add `"type": "module"` to `package.json`, this disables the automatic bundling heuristics in `@sanity/pkg-utils`: https://github.com/sanity-io/sanity/blob/55aaf005e128f4cfa8dd729e48246592b224ad3f/packages/groq/package.json#L32-L33
2. Add `bundle` conditions to `package.config.ts`, this allows us to setup the ESM and CJS targets separately: https://github.com/sanity-io/sanity/blob/5801ef613a73c55e4418118e61cc1e0f2aa3ad61/packages/groq/package.config.ts#L6-L10
3. With separate entrypoints we can handle the CJS target in a way that maintains backwards compatibility: https://github.com/sanity-io/sanity/blob/5801ef613a73c55e4418118e61cc1e0f2aa3ad61/packages/groq/src/_exports.cts.ts
4. While the ESM version doesn't have to deal with any of this: https://github.com/sanity-io/sanity/blob/5801ef613a73c55e4418118e61cc1e0f2aa3ad61/packages/groq/src/_exports.mts.ts
5. I've added two ([1](https://github.com/sanity-io/sanity/blob/5801ef613a73c55e4418118e61cc1e0f2aa3ad61/packages/groq/test.cjs), [2](https://github.com/sanity-io/sanity/blob/5801ef613a73c55e4418118e61cc1e0f2aa3ad61/packages/groq/test.mjs)) quick tests that demonstrates that it works: `git checkout bc-groq-define-query && pnpm install && cd packages/groq && pnpm turbo build && node test.cjs && node test.mjs`

romeovs:
@stipsan That did the trick! Thanks for the input, I've updated the PR accordingly.

Can you verify that this will be fully backwards-compatible?

packages/groq/package.json
```diff
@@ -26,17 +26,20 @@
...
-      "source": "./src/groq.ts",
-      "import": "./lib/groq.mjs",
-      "require": "./lib/groq.js",
+      "require": "./lib/groq.cjs",
       "default": "./lib/groq.js"
     },
+    "./define": {
+      "require": "./lib/define.cjs",
+      "default": "./lib/define.js"
+    },
```
stipsan, line 38:
```suggestion
```
I would prefer we don't add this export at first, if it turns out we can't have our cake and eat it too we can always add it later 🙌

stipsan (commented):
Yep looks good 👍 great work 💖

stipsan (dismissed):

packages/groq/package.json
romeovs: Oops that's me forgetting to remove that export. I've fixed up the PR.

stipsan:
Just out of curiously, why is this disabled by default?

> This feature is currently disabled by default and can be enabled by setting "overloadClientMethods": true in 'sanity-typegen.json.

What would happen if we output this syntax even if it's not supported/used, maybe because their `@sanity/client` isn't using a new enough version? My gut says it'll just no-op, am I missing something? 🤔

stipsan (approved):

romeovs:
@stipsan  I set it up as disabled by default on @sgulseth's request.

I think the main reasoning is that because of how module augmentation works, enabling this by default creates a situation where a user can not both use `sanity typegen generate` and not have their `@sanity/client` types augmented.

But, you are right, the types get added in a backwards-compatible way and using the patched type is opt-in, so I agree that it should be possible to roll this out without the config flag.

I'll leave it up to @sgulseth to decide!

packages/@sanity/codegen/src/typescript/typeGenerator.ts
sgulseth: Hm it would be good not to break this method.

A few alternatives:
 - the `generateQueryMap` could also just use this private map to lookup the generated type name.
 - add a `addToQueryMap(name: string, type: string)` that collects all the queries generated in the class, then `generateQueryMap()` can just traverse that list and spit out the types for all.
- Expose the typeNameMap in a method, then we can extract the generated typeName separately.

packages/@sanity/codegen/src/typescript/findQueriesInSource.ts
```diff
@@ -41,12 +42,18 @@ export function findQueriesInSource(
...
         babelTypes.isIdentifier(init.tag) &&
-        babelTypes.isIdentifier(node.id) &&
         init.tag.name === groqTagName
-      ) {
+
+      // Look for strings wrapped in a defineQuery function call
+      const isDefineQueryCall =
+        babelTypes.isCallExpression(init) &&
+        babelTypes.isIdentifier(init.callee) &&
+        init.callee.name === defineQueryFunctionName
```
sgulseth, line 54: since `defineQuery` is more ambiguous, and also exported by other packages, we should validate that it's the  `defineQuery` from `groq`

sgulseth (changes requested):
Sorry for the late review here. This looks really good, thanks a lot for helping out here! I have a few comments, nothing major, but in general would be nice to avoid breaking the TypeGenerator method

packages/@sanity/codegen/src/typescript/findQueriesInSource.ts
romeovs: I have added a helper to check where the `defineQuery` identifier actually comes from.

I implemented it only for the following import styles.

```ts
import { defineQuery } from "groq"
const query = defineQuery(" ... ")
```

```ts
import * as groq from "groq"

// Note: this is not supported by resolveExpression
const query = groq.defineQuery("...")
```

```ts
const { defineQuery } = require("groq")

// Note: resolveExpression works here, but is not set up to support commonjs 
// when detecting query fragments
const query = defineQuery("...")
```

Notably, I did not implement it for 
```ts
const groq = require('groq')
const query = groq.defineQuery("...")
```
as that would add significant complexity and is not supported by `resolveExpression`.

Let me know if you would like me to add that case too.

romeovs:
> @sgulseth: Hm it would be good not to break this method.

I agree that it's better to leave the signature for `generateTypeNodeTypes` unchanged, but I made these changes because I wanted to avoid coupling.

Unless I am misunderstanding your suggestions, I think each of them introduce some coupling:

> the `generateQueryMap` could also just use this private map to lookup the generated type name

This requires all the `generateTypeNodeTypes` calls to happen before the call to `generateQueryMap`, which might introduce subtle bugs.

> add a `addToQueryMap(name: string, type: string)` that collects all the queries generated in the class, then `generateQueryMap()` can just traverse that list and spit out the types for all.

This also requires all the `generateTypeNodeTypes` calls to happen before the call to `generateQueryMap`, which might introduce subtle bugs.

> Expose the `typeNameMap` in a method, then we can extract the generated typeName separately.

The problem with the `typeNameMap` is that it is not fully bijective.
If we call `getTypeName` with the same desired name twice, the result will be different (and we cannot recover the first name because it is overwritten).

We could imagine a situation where the following arises (when a user has two queries named `myQuery` for instance):

```ts
const schema = { ... }
const generator = new TypeGenerator(schema)

const typeA = { ... }
const typeB = { ... }

let code = ""
code += generator.generateTypeNodeTypes("MyQuery", typeA)
code += generator.generateTypeNodeTypes("MyQuery", typeB)
```
and this would generate two different identifiers, `MyQuery` and `MyQuery_2.

In this case it is impossible to ask the `TypeGenerator` "what is the type name used for `MyQuery`?" since it has two answers.

Even worse, in the current implementation, the `TypeGenerator` only stores the last type (ie. `MyQuery_2`) and forgets about the first one altogether.

I have implemented your first proposal in the PR now, but wanted to bring this up.

Maybe a good follow up would be to refactor `TypeGenerator` so there is less coupling?

packages/@sanity/codegen/src/typescript/findQueriesInSource.ts
sgulseth: Perfect! We don't need to support all variants, we can add more as we see them.
Member expressions aren't supported, no - but that should be handled separately anyway

sgulseth:
> The problem with the typeNameMap is that it is not fully bijective.

100%! I think it's wrong of us to store a map between the type name and the generated name, that part wasn't fully thought through. Though no harm atm since a query name can only appear once. What we really want is a map between the generated typenode and the name. They can of course still be identical, but then it doesn't matter which one is referenced anyway, from a types perspective. I'll send a PR for this

romeovs:
@sgulseth I've updated the PR to incorporate the changes from https://github.com/sanity-io/sanity/pull/7038, the `TypeGenerator` API is much cleaner now!

I've also added some test cases for when duplicate type names or query strings are used.

sgulseth (dismissed):
Nice! Clean and simple! :) 

LGTM, but holding of merging until we have support in the client as well

romeovs:
The sister PR over at `sanity-io/client` was merged successfully, so this should be good to go!

Is there anything I can do to help this PR get merged?

stipsan (dismissed):
Superb work @romeovs 💖

stipsan (dismissed):
I think we are good to ship 🚀 
Stellar work @romeovs! 💖 

We should merge this as soon as the merge conflict in `pnpm-lock.yaml` is resolved 😄 

image

PR #7001, opened 2024-06-20
fix: replace `@sanity/portable-text-editor` with `@portabletext/editor`

christianhg:
All the code in `packages/@sanity/portable-text-editor` has been moved to a new `@portabletext/editor` package published on npm. The package lives in https://github.com/portabletext/editor where you'll be able to see that all the git history for these files has been preserved.

The goal of publishing this as a standalone package is to eventually allow the Portable Text Editor to be used in other apps, innovate on the editor independently of the Studio and truly make the Portable Text spec feel open.

No code changes have been made in that package yet, so replacing the internal code with `v1.0.2` of the package *should* be safe and not accidentally delete any new work or cause regressions.


cngonzalez (dismissed):
looks okay to me!

rexxars (approved):

PR #7002, opened 2024-06-20
feat(ci): add workflow config for auto-updating studios

cngonzalez:
RESOLVES SDX-1412

### Description
As we move to different ways of building the Studio, it's important to test all functionality across existing and new build configuration. This PR adds a matrix option for auto-updating studios (with the potential of adding more). It also adds some conditional scripting logic to ensure that studios build and run according to the matrix configuration parameter.

In addition, it adds and updates tests that depend on a studio built with the auto-updates parameter.

### What to review
The changed Github workflow. I have a few specific questions:
1. Are we okay with `config: [auto-updating, default]`? Are there better names for what we want to express?
2. Should we also add logic for additional datasets depending on whether or not the studio is auto-updating?

### Testing
The specific tests added in this PR should pass under both conditions. All other tests should pass under both conditions.


cngonzalez:
Closing this one for now since there's some choices to be made about our CI processes etc.

PR #7008, opened 2024-06-21
fix(structure): fix incorrect i18n key for discard changes button "no changes" status

passion4code:
### Description

Old: `action.discard-changes.disabled.no-changes`
New: `action.discard-changes.disabled.no-change`

This is to match what is in https://github.com/sanity-io/locales/

### What to review

Localization key for the alt text on the "Discard Changes" action when it is disabled. 

### Testing

https://github.com/sanity-io/locales/blob/main/locales/fr-FR/src/structure.ts#L17 for an example of the key translated / used

### Notes for release

None


rexxars (approved):
Thank you!

PR #7013, opened 2024-06-24
fix(core): tree editing dialog open issue

hermanwikner:
### Description

This pull request fixes an issue with `getRootPath`, which caused the tree editing dialog to not open correctly using `shouldArrayDialogOpen`. The function incorrectly returned an empty array when no `_key` was present in the path.

**Before**: `['array']` => `[]`  
**After**: `['array']` => `['array']`

The fix changes the condition to `if (keyedSegmentIndex === -1)`, ensuring the entire path is returned if no `_key` is found.

### What to review

- Make sure that the tree editing dialog opens correctly

### Notes for release

N/A


RitaDias (approved):
✅ 💪 🚀

PR #7016, opened 2024-06-24
BUGFIX: Do not abort migration if user used `--no-dry-run` and `--no-confirm` flags together

pawelngei:
### Description

Currently Sanity has a bug where using the flags `--no-dry-run` and `--no-confirm` at once quits the migration instead of foregoing the confirmation prompt. This PR fixes it.

The `flags.confirm` used to be called `shouldConfirm` ( https://github.com/sanity-io/sanity/commit/2dba20698361e74e1bd67b390e42384d26ee4a85#diff-5dd59c387c299ba5b1746c4c0a69aed0ec8a759c264d799e8462f6a03f7e4a91L200 ) which might have created the confusion.

### What to review

Run a migration with both `--no-dry-run` and `--no-confirm` flags.

### Testing

Run a migration with both `--no-dry-run` and `--no-confirm` flags.

### Notes for release

- Fixes the usage of `--no-dry-run` and `--no-confirm` flags together.

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
@@ -195,7 +195,7 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
         type: 'confirm',
       }))
 
-    if (response === false) {
+    if (flags.confirm && response === false) {
       debug('User aborted migration')
       return
     }


ricokahler (approved):
Looks great! Thank you for catching this!

ricokahler (changes requested):
may have approved this a bit eagerly. i think the conditions can be cleared up a bit. it is definitely confusing and may be error prone with how it's set up on our side. i'll take a look at this this week but thank you for bringing this up

pawelngei:
Thanks for taking a look @ricokahler !

This is something which is actively blocking us from running some time-sensitive migrations through our CI.

Would it be possible to release a hotfix with the next minor version and follow it up with a proper refactor of the file later?

I see that it could be cleaner, it could also use some more tests (especially ones aimed at catching `--no-confirm` and `--no-run-dry`), but I specifically wanted to create a small fix to just get it patched ASAP :)

ricokahler:
ah heard. we can have a patch l out before the end of this week. thank you again

PR #7019, opened 2024-06-24
fix: check for telemetry consent before error reporting

cngonzalez:
### Description
In order to respect our users' telemetry consent, it's become necessary to ensure we don't send events to Sentry when consent has not been granted.

We are somewhat changing the paradigm of error reporting, where error reporters are assumed to be in a "disabled" state until otherwise enabled. We're doing so by:

1. Adding an `enable` and `disable` function on the errorReporter.
2. Adding a component to StudioProvider at an appropriate depth where the client can check the user's consent status.
3. In our specific Sentry implementation, initializing Sentry with a custom transport that buffers events until consent is received.

### What to review
Are we satisfied with this? Any caveats about in-memory issues or fragility?

### Testing
I'd like to incorporate tests for this in my other PR (#7002) -- possibly by intercepting the telemetry status call, since it happens later in the application.

### DIFF
--- a/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
+++ b/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
@@ -1,0 +1,55 @@
+import {makeFetchTransport} from '@sentry/react'
+import {type EventEnvelope, type Transport, type TransportMakeRequestResponse} from '@sentry/types'
+
+type BufferedTransport = Transport & {
+  setConsent: (consentGiven: boolean) => Promise<void>
+}
+
+/*
+ * Because we want to buffer events until the user has given consent to telemetry,
+ * we need to implement a custom transport, but mostly wrap the fetch transport.
+ */
+export function makeBufferedTransport(options: any): BufferedTransport {
+  let buffer: EventEnvelope[] = []
+  let consentGiven: boolean | undefined
+  const fetchTransport = makeFetchTransport(options)
+
+  const send = async (event: EventEnvelope) => {
+    if (consentGiven) {
+      return sendImmediately(event)
+    }
+    //we may not have received consent yet. Buffer the event until we know what to do.
+    else if (typeof consentGiven === 'undefined') {
+      buffer.push(event)
+    }
+    return {}
+  }
+
+  const sendImmediately = async (event: EventEnvelope): Promise<TransportMakeRequestResponse> => {
+    return fetchTransport.send(event)
+  }
+
+  const setConsent = async (consent: boolean) => {
+    consentGiven = consent
+    //we clear the buffer if consent is given (since we've sent the buffered events)
+    //and we clear the buffer if consent is revoked (since the events should not be sent)
+    if (consent) {
+      await flushBuffer()
+    }
+    buffer = []
+  }
+
+  const flushBuffer = async () => {
+    buffer.map(sendImmediately)
+  }
+
+  const flush = async () => {
+    return fetchTransport.flush()
+  }
+
+  return {
+    flush,
+    send,
+    setConsent,
+  }
+}
--- a/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
+++ b/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
@@ -8,6 +8,7 @@ export type BufferedTransport = Transport & {
 /*
  * Because we want to buffer events until the user has given consent to telemetry,
  * we need to implement a custom transport, but mostly wrap the fetch transport.
+  @internal
  */
 export function makeBufferedTransport(options: any): BufferedTransport {
   let buffer: EventEnvelope[] = []
@@ -22,6 +23,7 @@ export function makeBufferedTransport(options: any): BufferedTransport {
     else if (typeof consentGiven === 'undefined') {
       buffer.push(event)
     }
+    // consent not given, skip sending the event
     return {}
   }
 
--- a/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
+++ b/packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
@@ -1,14 +1,17 @@
 import {makeFetchTransport} from '@sentry/react'
 import {type EventEnvelope, type Transport, type TransportMakeRequestResponse} from '@sentry/types'
 
+/**
+ * @internal
+ */
 export type BufferedTransport = Transport & {
   setConsent: (consentGiven: boolean) => Promise<void>
 }
 
-/*
+/**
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/studio/EnableErrorReporting.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function EnableErrorReporting() {
+  const client = useClient().withConfig(DEFAULT_STUDIO_CLIENT_OPTIONS)
+
+  useEffect(() => {
+    async function checkConsent() {
+      try {
+        const res = await fetchTelemetryConsent(client)
+
+        if (res?.status === 'granted' || res?.status === 'unset') {
```
cngonzalez, line 20: We should confirm if "unset" response is a sufficient response here.

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
```diff
@@ -101,7 +101,7 @@ export function getSentryErrorReporter(): ErrorReporter {
     if (hasThirdPartySentry) {
       client = new BrowserClient({
         ...clientOptions,
-        transport: makeFetchTransport,
+        transport: makeBufferedTransport,
```
rexxars, line 104: Should initialize the transport in the below `init` call as well, I think? Potentially move it to `clientOptions` if possible?

rexxars (changes requested):

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
cngonzalez: ah, apologies -- victim of a hasty refactor. Should be sorted now.

packages/sanity/src/core/error/sentry/makeBufferedTransport.ts
```diff
@@ -0,0 +1,55 @@
...
+    //and we clear the buffer if consent is revoked (since the events should not be sent)
+    if (consent) {
+      await flushBuffer()
+    }
+    buffer = []
+  }
+
+  const flushBuffer = async () => {
+    buffer.map(sendImmediately)
+  }
```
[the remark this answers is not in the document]
cngonzalez: i think await makes sense here, since we wait for it if consent is given before clearing the buffer. again, apologies for hastiness.

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
```diff
@@ -172,9 +172,20 @@ export function getSentryErrorReporter(): ErrorReporter {
...
   }
 
+  function enable() {
+    //@ts-expect-error -- we've appended our transport with an additional function
+    client?.getTransport()?.giveConsent(true)
+  }
+  function disable() {
+    //@ts-expect-error -- we've appended our transport with an additional function
+    client?.getTransport()?.giveConsent(false)
+  }
```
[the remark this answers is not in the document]
cngonzalez: added (though TS still wanted an explicit `as`). again, apologies for rushing.

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
```diff
@@ -101,7 +103,7 @@ export function getSentryErrorReporter(): ErrorReporter {
     if (hasThirdPartySentry) {
       client = new BrowserClient({
         ...clientOptions,
-        transport: makeFetchTransport,
+        transport: makeBufferedTransport,
```
rexxars, line 106: Can drop this line since it's part of `clientOptions` now

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
```diff
@@ -172,9 +174,26 @@ export function getSentryErrorReporter(): ErrorReporter {
...
   }
 
+  function isBufferedTransport(transport: Transport | undefined): transport is BufferedTransport {
+    return !!transport && 'setConsent' in transport && typeof transport.setConsent === 'function'
+  }
+
+  function enable() {
+    if (isBufferedTransport(client?.getTransport())) {
+      ;(client?.getTransport() as BufferedTransport).setConsent(true)
+    }
```
rexxars, line 184: Typescript can't know that this is idempotent, but assigning to a variable should do the trick, I believe
```suggestion
    const transport = client?.getTransport()
    if (isBufferedTransport(transport)) {
      transport.setConsent(true)
    }
```

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
```diff
@@ -172,9 +174,26 @@ export function getSentryErrorReporter(): ErrorReporter {
...
+  }
+  function disable() {
+    if (isBufferedTransport(client?.getTransport())) {
+      ;(client?.getTransport() as BufferedTransport).setConsent(false)
+    }
```
rexxars, line 189:
```suggestion
    const transport = client?.getTransport()
    if (isBufferedTransport(transport)) {
      transport.setConsent(false)
    }
```

packages/sanity/src/core/studio/EnableErrorReporting.ts
rexxars: As far as I can tell from the code, `unset` means "not set, so don't allow", so this should be reduced to only `granted` I believe (ask me on Slack for details if you want a more thorough rationale)

rexxars (changes requested):

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
cngonzalez: tried it, but there appears to be a type mismatch coming from Sentry -- BrowserClient here takes BrowserClientOptions, which complains about a possibly undefined transport (I guess some misalignment with `ClientOptions` used above). Removing the `ClientOptions` cast on line 42 gets rid of this error. Are we okay with that?

packages/sanity/src/core/studio/EnableErrorReporting.ts
cngonzalez: Sorted, happy to do so -- was thinking there was some complexity about this per region.

packages/sanity/src/core/error/sentry/sentryErrorReporter.ts
rexxars: Ah. No biggie. Leave it as is.

packages/sanity/src/core/studio/EnableErrorReporting.ts
rexxars: Backend takes care of that - luckily

rexxars (approved):

PR #7021, opened 2024-06-25
Fix(CLI): Set integration value for createProject directly in the createProject function

RostiMelk:
### Description

Moving the position of where integration for a new project is defined, since createProject is called in multiple places throughout the CLI.

### Testing

- Create a new project though the cli
- Curl the `v1/projects/{projectId}` endpoint
- Verify that integration value is set to `cli`

### Notes for release

No notes


cngonzalez (approved):
LGTM! confirmed that the CLI initializes a project with the right information.

PR #7030, opened 2024-06-26
fix(core): put back select button for single-asset source

ricokahler:
### Description

Seems like there was a regression in #6930 where the Select/Browse button would not render if there was only a single asset source.

### What to review

Does the button show up now for a single asset source?

### Testing

I removed the custom asset source so that there would only be one in the test studio, observed the bug, then fixed it.

### Notes for release

Fixes a regression that caused the Select button to not appear.


binoy14 (approved):
Tested it manually and looks good, would be nice to have some sort of unit/component test for this. I see there is a unit test file for `ImageInput` but it's skipped/commented so probably not as simple to test I am assuming

PR #7035, opened 2024-06-27
refactor: use `@portabletext/editor` instead of `@sanity/portable-text-editor`

ricokahler:
### Description

Due to some issues with release, we had to cherry pick #7001 out of next. This branch has those changes cherry picked into them.

The issue we ran into was that it was causing a typescript error due to mismatched types. I think this is a dependency issue thing but it's something we need to investigate. 

https://github.com/portabletext/editor/pull/36 May also fix some of this issue as well

cc @christianhg 

### Notes for release

Uses the `@portabletext/editor` package instead of `@sanity/portable-text-editor`


christianhg:
@ricokahler and @rexxars, I hope things work now. I've moved the `@sanity/*` dependencies used by `@portabletext/editor` into peer dependencies: https://github.com/portabletext/editor/blob/main/packages/editor/package.json#L124-L127. I believe this should resolve the dependency issues we saw.

This means that this PR now relies on a new version of the `@portabletext/editor` -> `1.0.6`.

I still think it should be safe to switch to this package. While `1.0.6` doesn't represent 1:1 the source code of the original package (`1.0.2` did that), it's pretty close. And this version includes the much-requested fix for pasting in PTE in Firefox, so if you have the bandwidth it would be fantastic if you were willing to give this PR another go and see if we can do a new Studio patch release with it.

christianhg:
> [portabletext/editor#36](https://github.com/portabletext/editor/pull/36) May also fix some of this issue as well

@ricokahler , @rexxars , I've now published a new version to align `debug`: [4a96a6d](https://github.com/sanity-io/sanity/pull/7035/commits/4a96a6d422a1be6784d2d442256baea2e4635bdd)

rexxars (approved):

PR #7038, opened 2024-06-27
refactor(typegen): use typenode to lookup referencing type names

sgulseth:
### Description

When we generate the types we hold the generated type name in case it's referenced later. This was needed primarily because we, earlier, wanted to support multiple queries having the same name etc. Since we now doesn't allow this, it's not a big case. However, in https://github.com/sanity-io/sanity/pull/6997 we need to later reference all the names by the type nodes.

### What to review

Correctness: The outputted types should be 100% identical to the previous.

### Testing

Tests should still pass, I added a test for the case where it's referencing a type node we haven't seen.

### Notes for release

n/a - no notes needed


jonabc (approved):
😄 I don't understand the full context given in the PR writeup, but the code changes look accurate 👍

PR #7039, opened 2024-06-28
feat(corel): doc and structure updates

RitaDias:
### Description

**UI:**
- UI updates to the studio header: Create New Document (Remove text and update to bleed), spacing
- Add global version picker
- "Add to version" button on document footer (isReady button is commented out)
- Ad version badge on document header

- TEMPORARY: The UI inside of the modal is based on an early version of prototype that might change

**Quality of life:** 
- Rename all "releases" to "bundles"
- Add front end guardrails to prevent nested version creation 
- Move all components and utils to single folder `versions`  in `core/versions`

- TEMPORARY: Add dummy provider to allow for "current version" across studio while stores are set up
- TEMPORARY: Move utils around (fetch, get, const)
- TEMPORARY: Disable document version  picker bundle creator (only switching between versions left)

**Known limitations & notes**:
- Switching global versions will always make the document badge change regardless if that version of the file exists
- Switching versions of a document won't actually change the view of the document (it'll always be showing the one with the id: `draft.(id)`)
- No implementation of creating a release
- For now we're using tones in the colours for version and bundles, I think it makes more sense to make them use the hues fully. Left it outside of this PR (since it's already bit)

image


packages/sanity/src/core/versions/components/dialog/CreateBundleDialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+      width={1}
+    >
+      <Box padding={6}>
+        <BundleForm onChange={setValue} value={value} />
+      </Box>
+      <Flex justify="flex-end" padding={3}>
+        <Button
+          disabled={!value.title}
+          iconRight={ArrowRightIcon}
+          onClick={onSubmit(value)}
```
jordanl17, line 37: Q: should this be
```suggestion
          onClick={() => onSubmit(value)}
```
otherwise `onSubmit` will be called on every render

RitaDias: Yes! I need to go through the dialog component actually and update the methods to make sure that they aren't being called on each render. I will update this

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+    (bundle: Version) => () => {
+      const {name} = bundle
+
+      if (name === 'drafts') {
+        router.navigateStickyParam('perspective', '')
+      } else {
+        router.navigateStickyParam('perspective', `bundle.${name}`)
+      }
+
+      setCurrentVersion(bundle)
```
pedrobonamin, line 34: I know you are waiting here to have the store for this, but will we need a state to handle this ?
I think it could be derived from the router

packages/sanity/src/_singletons/core/form/VersionContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  // eslint-disable-next-line no-empty-function
+  setCurrentVersion: () => {},
+})
+
+interface VersionProviderProps {
+  children: ReactElement
+}
+
+export function VersionProvider({children}: VersionProviderProps): JSX.Element {
+  const [currentVersion, setCurrentVersion] = useState<Version>(LATEST)
```
pedrobonamin, line 28: I know you are waiting here to have the store for this, but will we need a state to handle this ?
I think it could be derived from the router.

We could do something like:

```ts
export function VersionProvider({children}: VersionProviderProps): JSX.Element {
  const router = useRouter()
  const setCurrentVersion = (version: Version) => {
    const {name} = version
    if (name === 'drafts') {
      router.navigateStickyParam('perspective', '')
    } else {
      router.navigateStickyParam('perspective', `bundle.${name}`)
    }
  }
  const selectedVersion = router.stickyParams?.perspective
    ? BUNDLES.find((bundle) => bundle.name === router.stickyParams.perspective)
    : LATEST

  const currentVersion = selectedVersion || LATEST

  const isDraft = currentVersion.name === 'draft'

  const contextValue: VersionContextValue = {
    isDraft,
    setCurrentVersion,
    currentVersion,
  }

  return <VersionContext.Provider value={contextValue}>{children}</VersionContext.Provider>
}
```

To navigate and set the current version in the same action.

RitaDias: We likely won't need this provider / context at all (hence why it's so rough) once we have the store :) 
It can be done to use your suggestion of course (or similar to match the rest of the context and provider in the studio) but my first thought was that we would eventually not need it.

Hence why my comment at the top of this file 👇 🙂 

``// eslint-disable-next-line no-warning-comments
/* TODO REMOVE THIS, THIS IS JUST A TEMPORARY FIX UNTIL STORES ARE SET UP */``

I can make the suggested changes as temporary things until we see if we need it since I am going to be moving towards this solution, but just a headsup given the comment I left on the file and where my mind was at :)

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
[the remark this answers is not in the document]
RitaDias: @pedrobonamin your comment vanished but just in case you missed it ☝️

packages/sanity/src/core/versions/util/const.ts
```diff
@@ -0,0 +1,38 @@
...
+  'circle',
+  'search',
+  'sun',
+  'star',
+  'trash',
+  'user',
+]
+
+export const LATEST: Version = {
+  name: 'draft',
```
pedrobonamin, line 24: Should this be `drafts` ? 
```suggestion
  name: 'drafts',
```

The id generated for the draft bundles is `drafts. + id`

packages/sanity/src/core/versions/components/dialog/BundleForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+  TextArea,
+  TextInput,
+} from '@sanity/ui'
+import {useCallback} from 'react'
+
+import {type Bundle} from '../../types'
+import {RANDOM_TONES} from '../../util/const'
+import {toSlug} from '../../util/dummyGetters'
+
+export function BundleForm(props: {
```
pedrobonamin, line 21: I'm thinking if we should add some validations here, and also on the store, to disallow the creation of bundles with the name `draft` or `drafts`

RitaDias: Very good point, adding it 👍

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+    },
+    [router, setCurrentVersion],
+  )
+
+  const handleGoToLatest = useCallback(
```
pedrobonamin, line 39: Do we need a specific `handleGoToLatest` function? Thinking we can reuse the `handleBundleChange` and use the `LATEST` bundle

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+        }
+        id="global-version-menu"
+        menu={
+          <Menu>
+            <MenuItem
+              iconRight={isDraft ? <CheckmarkIcon /> : undefined}
+              onClick={handleGoToLatest()}
+              pressed={false}
+              // eslint-disable-next-line @sanity/i18n/no-attribute-string-literals
+              text="Latest"
```
pedrobonamin, line 91: Can we get this text from the `LATEST` bundle?

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,142 @@
...
+            {bundles
+              .filter((b) => b.name !== 'draft')
+              .map((b) => (
+                <MenuItem key={b.name} onClick={handleBundleChange(b)} padding={1} pressed={false}>
+                  <Flex>
+                    <VersionIcon tone={b.tone} icon={b.icon} padding={2} />
+
+                    <Box flex={1} padding={2} style={{minWidth: 100}}>
+                      <Text size={1} weight="medium">
+                        {b.name === 'draft' ? 'Latest' : b.title}
```
pedrobonamin, line 107: `b.name === 'draft'` is not possible right? A filter for this is applied in [row 99](https://github.com/sanity-io/sanity/pull/7039/files#diff-0eddbc52536e7776df00ffd12d620855ccd15f8291b941d7b11eee3569c90ed2R99)

RitaDias: Yep, the filter was added after this condition (and I forgot to remove these), good catch!

packages/sanity/src/core/versions/components/VersionBadge.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import {type Version} from '../types'
+import {VersionIcon} from './VersionIcon'
+
+export function VersionBadge(props: {version: Version}): JSX.Element {
+  const {version} = props
+
+  return (
+    <Box flex="none">
+      <Button as="a" mode="bleed" padding={0} radius="full">
```
pedrobonamin, line 11: NIT:  There is a small difference in the hover behaviour comparing the globalBundleMenu and this button.
I think it's possible caused by the `as=a` property.
See how the gray background goes beyond the limits of the Badge and the underline of the text.

Screenshot 2024-07-01 at 10 19 48
Screenshot 2024-07-01 at 10 19 52

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
```diff
@@ -56,31 +59,49 @@ function DocumentStatusBarActionsInner(props: DocumentStatusBarActionsInnerProps
...
+                  text={firstActionState.label}
+                  tone={firstActionState.tone || 'primary'}
+                />
+              ) : (
+                <BundleActions
+                  currentVersion={currentVersion}
+                  documentId={documentId}
+                  documentType={documentType}
+                  isReady={isReady}
+                />
```
pedrobonamin, line 98: Thinking this should be part of the plugin once set, could we add a TODO here to review this change later?

The plugin could control which actions it allows, and it could also push the `Add to version` action as the first action. Without updating this component

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
```diff
@@ -56,31 +59,49 @@ function DocumentStatusBarActionsInner(props: DocumentStatusBarActionsInnerProps
...
+              )}
             </Stack>
           </Tooltip>
         </LayerProvider>
       )}
-      {showMenu && menuActionStates.length > 0 && (
+      {showMenu && menuActionStates.length > 0 && isDraft && (
```
pedrobonamin, line 104: Check also previous comment.

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
RitaDias: Good point, adding to this and next comment

packages/sanity/src/core/versions/components/VersionBadge.tsx
RitaDias: Quite eagle eyes 😄 
I had to change the brightness of my screen to catch that on your screenshots!
On my way to fix it 👍

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const {currentVersion, setCurrentVersion, isDraft} =
+    useContext<VersionContextValue>(VersionContext)
+  const [createBundleDialogOpen, setCreateBundleDialogOpen] = useState(false)
+
+  const handleBundleChange = useCallback(
+    (bundle: Version) => () => {
+      const {name} = bundle
+
+      if (isDraftOrPublished(name)) {
+        router.navigateStickyParam('perspective', '')
```
juice49, line 34: Does passing an empty string here mean it's no longer possible to access the system perspectives?

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
```diff
@@ -0,0 +1,136 @@
...
+              title={isDraft ? LATEST.title : currentVersion?.title}
+            />
+          </Button>
+        }
+        id="global-version-menu"
+        menu={
+          <Menu>
+            <MenuItem
+              iconRight={isDraft ? <CheckmarkIcon /> : undefined}
+              onClick={handleBundleChange(LATEST)}
```
juice49, line 83: Sorry, I may be misreading the code, but should this be wrapped in a function?
```suggestion
              onClick={() => handleBundleChange(LATEST)}
```

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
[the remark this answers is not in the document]
pedrobonamin: Ohh I moved it to the `VersionContext` file when I realised that should happen there, thanks Rita!

juice49 (approved):
Looks great! 🎉

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
RitaDias: Passing the empty string here is how we were doing it when it's either the Draft or Published :)

This is just the fancy way of doing what we were doing in the StudioNavbar.tsx before my component.
So it just removes the latest part of the url

image

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
RitaDias: I _think_ it shouldn't be needed since `handleBundleChange` is wrapped in a `useCallback`

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
juice49: Ah, my bad! 🙂

packages/sanity/src/core/versions/components/GlobalBundleMenu.tsx
juice49: This is my mistake, I didn't see that `handleBundleChange` already returns a function. Sorry 🤦‍♀️.

RitaDias: No need to apologise these things happen (especially to me, really often 🫠 )

PR #7040, opened 2024-06-28
feat(corel): integrate bundles store

pedrobonamin:
### Description

Introduces `bundlesStore` allowing to fetch bundles metadata from the `addon-dataset`.
Also, introduces a minimal implementation for `bundlesOperations` for testing purposes.

It can be tested in the following story:
https://test-studio-git-corel-6.sanity.build/test/workshop/core;bundles;bundles-store

Screenshot 2024-06-28 at 14 31 05

### DIFF
--- a/packages/sanity/src/core/store/bundles/__workshop__/BundlesStoreStory.tsx
+++ b/packages/sanity/src/core/store/bundles/__workshop__/BundlesStoreStory.tsx
@@ -1,0 +1,112 @@
+import {Card, Flex, Stack, Text} from '@sanity/ui'
+import {type ComponentType, type FormEvent, useCallback, useState} from 'react'
+import {AddonDatasetProvider, LoadingBlock} from 'sanity'
+
+import {Button} from '../../../../ui-components'
+import {type BundleDocument} from '../types'
+import {useBundleOperations} from '../useBundleOperations'
+import {useBundlesStore} from '../useBundlesStore'
+import {ReleaseForm} from './ReleaseForm'
+
+const WithAddonDatasetProvider = <P extends object>(Component: ComponentType<P>): React.FC<P> => {
+  // Function that returns the wrapped component
+  const WrappedComponent: React.FC<P> = (props) => (
+    <AddonDatasetProvider>
+      <Component {...props} />
+    </AddonDatasetProvider>
+  )
+
+  // Setting a display name for the wrapped component
+  WrappedComponent.displayName = `WithAddonDatasetProvider(${Component.displayName || Component.name || 'Component'})`
+
+  return WrappedComponent
+}
+
+const initialValue = {name: '', title: '', tone: undefined, publishAt: undefined}
+const BundlesStoreStory = () => {
+  const {data, loading} = useBundlesStore()
+  const {createBundle, deleteBundle} = useBundleOperations()
+  const [creating, setCreating] = useState(false)
+  const [deleting, setDeleting] = useState<string | null>(null)
+  const [value, setValue] = useState<Partial<BundleDocument>>(initialValue)
+  const handleCreateBundle = useCallback(
+    async (event: FormEvent<HTMLFormElement>) => {
+      try {
+        event.preventDefault()
+        setCreating(true)
+        await createBundle(value)
+        setValue(initialValue)
+      } catch (err) {
+        console.error(err)
+      } finally {
+        setCreating(false)
+      }
+    },
+    [createBundle, value],
+  )
+
+  const handleDeleteBundle = useCallback(
+    async (id: string) => {
+      try {
+        setDeleting(id)
+        await deleteBundle(id)
+      } catch (err) {
+        console.error(err)
+      } finally {
+        setDeleting(null)
+      }
+    },
+    [deleteBundle],
+  )
+
+  return (
+    <Stack space={3}>
+      <Flex gap={2}>
+        <Card margin={3} padding={3} border>
+          <form onSubmit={handleCreateBundle}>
+            <Stack space={4}>
+              <Text weight="medium">Create a new release</Text>
+              <ReleaseForm onChange={setValue} value={value} />
+              <Flex justify="flex-end">
+                <Button
+                  text="Create"
+                  tone="primary"
+                  type="submit"
+                  disabled={creating}
+                  loading={creating}
+                />
+              </Flex>
+            </Stack>
+          </form>
+        </Card>
+        <Card margin={3} border padding={3}>
+          <div style={{maxHeight: '400px', overflow: 'scroll'}}>
+            <Text>Data</Text>
+            {loading ? <LoadingBlock /> : <pre>{JSON.stringify(data, null, 2)}</pre>}
+          </div>
+        </Card>
+      </Flex>
+      <Card margin={3} border padding={3}>
+        <Stack space={3}>
+          {data?.map((bundle) => (
+            <Card key={bundle._id} padding={3} border radius={3}>
+              <Flex align="center" gap={3} justify={'space-between'}>
+                <Text>{bundle.name}</Text>
+                <Button
+                  text="Delete"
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/store/bundles/__workshop__/BundlesStoreStory.tsx
```diff
@@ -0,0 +1,110 @@
+import {Card, Flex, Stack, Text} from '@sanity/ui'
```
pedrobonamin, line 1: This story will be useful for testing purposes.

packages/sanity/src/core/store/bundles/__workshop__/ReleaseForm.tsx
```diff
@@ -0,0 +1,135 @@
+import {COLOR_HUES} from '@sanity/color'
```
pedrobonamin, line 1: This is just a copy from the prototype, that's why I'm adding it in `__workshop__`

packages/sanity/src/core/store/bundles/useBundleOperations.ts
```diff
@@ -0,0 +1,50 @@
...
+import {useAddonDataset} from '../../studio/addonDataset/useAddonDataset'
+import {type BundleDocument} from './types'
+
+// WIP - Raw implementation for initial testing purposes
+export function useBundleOperations() {
+  const {client} = useAddonDataset()
+
+  const handleCreateBundle = useCallback(
+    async (bundle: Partial<BundleDocument>) => {
+      const document = {
```
jordanl17, line 13: COULD: assign fields like `_createdAt`, `_updatedAt`, `_rev` at this point

pedrobonamin: We could pre populate `_createdAt` and `_updatedAt`, but they will be added automatically by content lake when creating the document.
`_rev` will always be overridden by CL.

packages/sanity/src/core/store/bundles/__workshop__/ReleaseForm.tsx
RitaDias: To add a note for future proof: we spoke about this on slack, this has changed slightly ([naming and structure](https://github.com/sanity-io/sanity/blob/0162c683cd43ee01c443d3e28e84ab88b7c2e876/packages/sanity/src/core/versions/components/dialog/BundleForm.tsx)) but we can update this once both our PRs are merged

RitaDias (approved):
This looks good! This might a question to talk on slack through, but to double check:
The concept of **bundle** that we have will solely exist on "our side" for UI & concept purposes and not on CL since **versions** are the only concept that technically exists on their side right? 🤔

juice49:
> The concept of bundle that we have will solely exist on "our side" for UI & concept purposes and not on CL since versions are the only concept that technically exists on their side right? 🤔

Not _quite_. My understanding is the existence of a versioned document (any document with a dot in its `_id` and a `_version` field defined) will implicitly be treated as a bundle by Content Lake. Content Lake has to have some concept of bundles in order for clients to make requests to specific bundles using the `bundlePerspective`/`perspective` parameter.

However, the bundle metadata will solely exist on "our side".

packages/sanity/src/core/store/bundles/__workshop__/ReleaseForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+  Stack,
+  Text,
+  TextArea,
+  TextInput,
+} from '@sanity/ui'
+
+import {Button} from '../../../../ui-components/button'
+import {type BundleDocument} from '../types'
+
+function toSlug(value: string): string {
```
juice49, line 18: Not super important, but we already include [SpeakingURL](https://www.npmjs.com/package/speakingurl) with Studio for the default slugify function. It might be more robust than doing it ourselves.

packages/sanity/src/core/store/bundles/reducer.ts
```diff
@@ -0,0 +1,109 @@
...
+  type: 'BUNDLE_DELETED'
+}
+
+interface BundleUpdatedAction {
+  payload: BundleDocument
+  type: 'BUNDLE_UPDATED'
+}
+
+interface BundlesSetAction {
+  bundles: BundleDocument[]
```
juice49, line 19: Just a nitpick, nothing important. For consistency, should this be `payload` instead?

```suggestion
  payload: BundleDocument[]
```

packages/sanity/src/core/store/bundles/useBundleOperations.ts
```diff
@@ -0,0 +1,50 @@
+import {uuid} from '@sanity/uuid'
+import {useCallback} from 'react'
+
+import {useAddonDataset} from '../../studio/addonDataset/useAddonDataset'
+import {type BundleDocument} from './types'
+
+// WIP - Raw implementation for initial testing purposes
+export function useBundleOperations() {
+  const {client} = useAddonDataset()
```
juice49, line 9: Nice! 🙂 I didn't know `useAddonDataset` existed.

pedrobonamin: It was initially part of comments and then moved to core when we built tasks.

packages/sanity/src/core/store/bundles/__workshop__/ReleaseForm.tsx
pedrobonamin: Thanks!

packages/sanity/src/core/store/bundles/useBundlesStore.ts
```diff
@@ -0,0 +1,152 @@
...
+        return of(err)
+      }),
+    )
+
+    return events$ // as Observable<ListenEvent<Record<string, BundleDocument>>>
+  }, [client, handleListenerEvent])
+
+  useEffect(() => {
+    if (!client) return
+    const subscription = initialFetch$()
```
juice49, line 134: I love the RxJS usage 🤩.

packages/sanity/src/core/store/bundles/useBundlesStore.ts
```diff
@@ -0,0 +1,152 @@
...
+      .pipe(concatMap(() => listener$))
+      .subscribe()
```
juice49, line 136: Instead of manually creating and cleaning up the subscription, would it work to use [the `useObservable` hook from ReactRx](https://react-rx.dev/reference#useobservable)? We use this in some other places already.

juice49 (approved):

juice49:
This looks great! Left some comments, but no blockers.

packages/sanity/src/core/store/bundles/__workshop__/ReleaseForm.tsx
pedrobonamin: TIL, thanks. 
@RitaDias we should consider this once we merge the forms

RitaDias: I'll update this on my PR just so we use it already 👍

RitaDias: Added 👍 
Unrelated to this comment: but related to the Hues, I'll update the UI to use the hues instead of tones once I start using this store.

Thanks 🙏

packages/sanity/src/core/store/bundles/useBundlesStore.ts
pedrobonamin: This is nice, thanks! 
Updated to:
```ts
  const observable = useMemo(() => {
    if (!client) return of(null) // emits null and completes if no client
    return initialFetch$().pipe(concatMap(() => listener$))
  }, [initialFetch$, listener$, client])

  useObservable(observable)
```

PR #7045, opened 2024-06-30
fix(core): handle fieldset members in `FormInput`

hermanwikner:
### Description

This pull request fixes an issue where the `FormInput` would not render a field inside a fieldset. The current logic only finds fields that have the `kind` of `"field"` but does not handle the `kind` of `"fieldSet"`. The issue can be reproduced by:

1. Adding the following schema with an array field that has a fieldset defined:

    ```ts
    import {defineType} from 'sanity'

    export const myDocument = defineType({
      type: 'document',
      name: 'myDocument',

      fieldsets: [
        {
          name: 'fieldset',
          title: 'Fieldset',
        },
      ],

      fields: [
        {
          type: 'array',
          name: 'fieldsetArray',
          title: 'Fieldset array',

          fieldset: 'fieldset',

          of: [
            {
              type: 'object',
              name: 'myObject',
              fields: [
                {
                  name: 'string',
                  type: 'string',
                  title: 'String',
                },
              ],
            },
          ],
        },
      ],
    })
    ```

2. Enabling the array tree editing beta feature:

    ```ts

    export default defineConfig({
      // ...

      features: {
        beta: {
          treeArrayEditing: {
            enabled: true
          }
        }
      }
    })
    ```

3. Creating an array item, and the `FormInput` will render this error message:

    Screenshot 2024-06-30 at 17 28 18

### What to review

- Ensure that the issue is fixed
- General code review

### Notes for release

N/A

### DIFF
--- a/dev/test-studio/schema/debug/objectsDebug.ts
+++ b/dev/test-studio/schema/debug/objectsDebug.ts
@@ -213,6 +213,28 @@ const animals = defineField({
   of: [animal],
 })
 
+const fieldsetArray = defineField({
+  type: 'array',
+  name: 'fieldsetArray',
+  title: 'Fieldset array',
+
+  fieldset: 'fieldset',
+
+  of: [
+    {
+      type: 'object',
+      name: 'myObject',
+      fields: [
+        {
+          name: 'string',
+          type: 'string',
+          title: 'String',
+        },
+      ],
+    },
+  ],
+})
+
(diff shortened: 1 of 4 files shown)


RitaDias (changes requested):
👋 Fantastic work! Do you mind adding a component test for this? 🙏

RitaDias (approved):
🚀

PR #7046, opened 2024-07-01
fix(core):  open PTE dialogs in tree editing dialog with `openPath`

hermanwikner:
### Description

This pull request fixes an issue where the tree editing dialog does not open correctly when the `openPath` points to an array item inside a portable text block. This can occur, for example, when navigating to a field with presence, from a comment, or using the validation menu.

Steps to reproduce the issue:

1. Open the tree editing dialog.
2. Open an array item inside a portable text editor.
3. Add a comment to a field in the array item.
4. Close all dialogs.
5. Click the comment in the inspector.
6. Notice that the dialog(s) do not open.

### What to review

- General code review
- Ensure that the PTE dialogs are opening correctly inside the tree editing dialog when setting the `openPath`

### Testing

- Component tests added to cover this particular situation

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/buildArrayState.ts
+++ b/packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/buildArrayState.ts
@@ -1,3 +1,4 @@
+import {toString} from '@sanity/util/paths'
 import {
   type ArraySchemaType,
   EMPTY_ARRAY,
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/studio/tree-editing/utils/build-tree-editing-state/buildArrayState.ts
```diff
@@ -122,6 +123,19 @@ export function buildArrayState(props: BuildArrayState): TreeEditingState {
...
+      }
+
+      // If `openPath` points to an array field within a portable text field,
+      // set `relativePath` to the parent of the portable text field.
+      // This ensures that the tree editing dialog opens at the parent level
+      // of the portable text field.
+      // Portable text fields manage their own dialogs, so we open the tree editing
+      // dialog for the parent item and let the portable text field handle its
+      // dialogs via `openPath`.
+      if (isPortableText && toString(openPath).startsWith(toString(childPath))) {
```
RitaDias, line 137: Small nit: think you can use [`startsWith`](https://github.com/sanity-io/sanity/blob/corel-35/packages/@sanity/util/src/paths.ts#L126) from the Path Utils :)

RitaDias (approved):
🚀

PR #7047, opened 2024-07-01
feat(COREL): update types to bundle + add create bundle functionality

RitaDias:
### Description

**Quality of life:**
- Add ability to create bundle from Global menu
- Use `BundleDocument` type across existing UI
- Add BundleMenu component to be reused on Global and Document menu
- Add date picker to bundle creator
- Remove unused types & unused code
- Rename VersionBadge to BundleBadge

**UI:**
- Add hue and icon picker based on newest version of prototype
- Update UI to use the existing bundles
- Render dates on the form creator and in the global & document menus
- Add Hue picker and move from tone to hue across all components
- Don't allow to add document to bundle if a document already exists in it

**Question:** Should the `DocumentVersionMenu` be renamed to `DocumentBundleMenu`?
- Jordan says that it makes sense to keep as is (`DocumentVersionMenu`)

image


packages/sanity/src/core/studio/StudioProvider.tsx
```diff
@@ -70,12 +71,14 @@ export function StudioProvider({
     <WorkspaceLoader LoadingComponent={LoadingBlock} ConfigErrorsComponent={ConfigErrorsScreen}>
       <StudioTelemetryProvider config={config}>
         <LocaleProvider>
-          <VersionProvider>
-            <PackageVersionStatusProvider>
-              <EnableErrorReporting />
-              <ResourceCacheProvider>{children}</ResourceCacheProvider>
-            </PackageVersionStatusProvider>
-          </VersionProvider>
+          <AddonDatasetProvider>
```
RitaDias, line 74: Had to add it in order to use `useBundlesStore` in the context.
Will likely remove it once I have the updates to the bundle store -> context

packages/sanity/src/core/versions/components/VersionBadge.tsx
jordanl17, line 1: COULD: rename as `BundleBadge`

RitaDias: Good catch, definitely!

packages/sanity/src/core/studio/StudioProvider.tsx
RitaDias: My guess is that depending on how we go about this: no we will likely not need the version provider at all if the releases plugin is not added to the studio.

(hence why I haven't really spent much time in the contextPlugin and looks messy 🫠 )

However, because of we are right now development wise, we do need something like the version context at a higher level because the currentVersion (or the version that is picked is across multiple levels in the studio: navbar, document level and footer in the document)

packages/sanity/src/core/versions/util/dummyGetters.ts
```diff
@@ -5,11 +5,6 @@ import {type BundleDocument} from '../../store/bundles/types'
 
 /* MOSTLY TEMPORARY FUNCTIONS / DUMMY DATA */
 
-export const getRandomToneIcon = () => ({
```
RitaDias, line 9: I removed this since we now should always have colour, and icons :)

RitaDias:
Right now we have all of the UI components in the `core/versions` folder, I wonder if I should move rename that to `core/bundle`

packages/sanity/src/core/studio/StudioProvider.tsx
pedrobonamin: If we move it to the releases plugin, the context will be added at the studio level, so it will be possible to use it from the navbar, document and footer.

We could have in `packages/sanity/src/core/releases/plugin/ReleasesStudioLayout.tsx`

```ts
import {VersionProvider} from '../../../_singletons/core/form/VersionContext'
import {type LayoutProps} from '../../config'
import {BundlesProvider} from '../../store/bundles/BundlesProvider'
import {AddonDatasetProvider} from '../../studio'

export function ReleasesStudioLayout(props: LayoutProps) {
  // TODO: Replace for useReleasesEnabled
  const {enabled, mode} = {enabled: true, mode: 'default'}

  if (!enabled) {
  }

  return (
    <AddonDatasetProvider>
      <BundlesProvider>
        <VersionProvider>{props.renderDefault(props)}</VersionProvider>
      </BundlesProvider>
    </AddonDatasetProvider>
  )
}
```

And then, in the `<VersionProvider />` we can have access to the bundlesProvider and use `useBundles` instead of `useBundlesStore` [here](https://github.com/sanity-io/sanity/blob/corel-35/packages/sanity/src/_singletons/core/form/VersionContext.tsx#L31-L32).

RitaDias: This makes a lot of sense, if you're ok with it I'll make this change in a separate PR since I'm already focusing on updating the VersionProvider in a different branch :)

pedrobonamin: This component will add the necessary providers here https://github.com/sanity-io/sanity/blob/corel-35/packages/sanity/src/core/studio/Studio.tsx#L117-L118

So it's wrapping the studio tool.

packages/sanity/src/_singletons/core/form/VersionContext.tsx
```diff
@@ -4,14 +4,15 @@
 // eslint-disable-next-line import/consistent-type-specifier-style
 import {createContext, type ReactElement} from 'react'
 
-import type {Version} from '../../../core/versions/types'
-import {BUNDLES, LATEST} from '../../../core/versions/util/const'
+import {useBundlesStore} from '../../../core/store/bundles'
```
pedrobonamin, line 7: Maybe it's part of a future change, but just asking to keep it consistent with the rest of contexts added here.

What I see in the rest of files in `_singletons` is that this only haves the context creation.

see for example https://github.com/sanity-io/sanity/blob/corel-35/packages/sanity/src/_singletons/core/form/FormValueContext.ts

Will it make sense to move the rest of things in this folder out to a VersionProvider ?

Also, in the provider or here, could we add a `useVersion` hook that exposes the context like https://github.com/sanity-io/sanity/blob/corel-35/packages/sanity/src/core/store/bundles/BundlesProvider.tsx#L41-L47

pedrobonamin (changes requested):
This is looking great @RitaDias fantastic job!! 
Added some suggestions, open to discussion.

Also wanted to check on a functionality, I think that once a release is created from the top menu, the new release should be "selected".

packages/sanity/src/core/studio/StudioProvider.tsx
pedrobonamin: Sure @RitaDias makes sense!

packages/sanity/src/core/versions/components/dialog/BundleForm.tsx
```diff
@@ -82,81 +123,64 @@ export function BundleForm(props: {
...
-                <option key={tone} value={tone}>
-                  {tone}
-                </option>
-              ))}
-            </Select>
-          </Stack>
-        </Flex>
+        )}
+
+        {
```
pedrobonamin, line 151: NIT:  This bracket and the one in line 179 are not needed

packages/sanity/src/core/versions/components/dialog/BundleForm.tsx
```diff
@@ -82,81 +123,64 @@ export function BundleForm(props: {
...
+                    mode="bleed"
+                    padding={2}
+                    onClick={handleOpenDatePicker}
+                  />
+                </Box>
+              </Popover>
+            }
+            value={displayDate}
+            onChange={handlePublishAtInputChange}
+          />
```
pedrobonamin, line 182: 🤯

packages/sanity/src/core/versions/components/BundleMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+  button: ReactElement
+  bundles: BundleDocument[] | null
+  loading: boolean
+  actions?: ReactElement
+}
+
+export function BundleMenu(props: BundleListProps): JSX.Element {
+  const {bundles, loading, actions, button} = props
+
+  // eslint-disable-next-line no-warning-comments
```
jordanl17, line 30: COULD: add these eslint rules to the `.eslintrc` global, where we also turn off the string literal rules for releases

An aside, being that we might also want to change [this](https://github.com/sanity-io/sanity/blob/corel/.eslintrc.cjs#L166) to:

`files: ['**/*/core/releases/**/*', '**/*/core/version/**/*']`

packages/sanity/src/core/versions/components/panes/BundleActions.tsx
```diff
@@ -1,32 +1,45 @@
...
 
   const fetchVersions = useCallback(async () => {
-    const response = await getAllVersionsOfDocument(client, documentId)
-    setDocumentVersions(response)
-  }, [client, documentId])
+    if (!loading) {
+      const response = await getAllVersionsOfDocument(bundles, client, documentId)
+      setDocumentVersions(response)
+      setIsInVersion(versionDocumentExists(documentVersions, name))
+    }
```
pedrobonamin, line 41: we will eventually get this from the document pair store right? 
cc: @juice49

packages/sanity/src/core/versions/components/BundleMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  iconRight={isDraft ? <CheckmarkIcon /> : undefined}
+                  onClick={handleBundleChange(LATEST)}
+                  pressed={false}
+                  text={LATEST.title}
+                />
+                {bundles && bundles.length > 0 && (
+                  <>
+                    <MenuDivider />
+                    {bundles
+                      .filter((b) => !isDraftOrPublished(b.name) && !b.archived)
```
jordanl17, line 67: `b.archived` doesn't exist on `BundleDocument` it seems though. I know we'd discussed elsewhere around having `archivedAt`

packages/sanity/src/core/versions/components/dialog/CreateBundleDialog.tsx
```diff
@@ -1,75 +1,71 @@
...
-import {getRandomToneIcon, isDraftOrPublished} from '../../util/dummyGetters'
+import {isDraftOrPublished} from '../../util/dummyGetters'
 import {BundleForm} from './BundleForm'
 
-export function CreateBundleDialog(props: {
-  onCancel: () => void
-  onCreate: (value: Bundle) => void
-}): JSX.Element {
-  const {onCancel, onCreate} = props
+export function CreateBundleDialog(props: {onClose: () => void}): JSX.Element {
```
jordanl17, line 10: `BundlesOverview` will need its props updated

packages/sanity/src/core/versions/components/dialog/BundleForm.tsx
RitaDias: let's gooo 🚤

packages/sanity/src/core/versions/components/BundleMenu.tsx
RitaDias: I feel this would go against the point of why I'm adding them on the files:
I want them to be intrusive and very visible these things NEED to be discussed before we push it to `next` and a reviewer should raise them up if we end up having them in the PR for review.

If we were to add that, then I think it would beat the point. Normal comments, imo, are often brushed over in reviews, linting disables raise more of a brow 😄

packages/sanity/src/core/versions/components/BundleMenu.tsx
RitaDias: Yeah I made this change before the archivedAt was discussed 😄 
@juice49 & @pedrobonamin are we ok with making the change to `archivedAt` ?

packages/sanity/src/core/versions/components/dialog/CreateBundleDialog.tsx
RitaDias: Alright, making this change :)

packages/sanity/src/core/versions/components/BundleMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+    [setCurrentVersion],
+  )
+
+  return (
+    <>
+      <MenuButton
+        button={button}
+        id="global-version-menu"
+        menu={
+          <StyledMenu>
```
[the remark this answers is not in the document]
RitaDias: Great suggestion, I've added it :)

packages/sanity/src/core/versions/components/BundleMenu.tsx
RitaDias: I remeber that Pedro at one point or another mentioned that we might want to make use of "status" instead, not sure where we landed on this

RitaDias: Landed on keeping the `archivedAt`!

packages/sanity/src/core/versions/components/panes/BundleActions.tsx
RitaDias: We've made a decision to move the discussion at a later point once we have a bit more information

packages/sanity/src/_singletons/core/form/VersionContext.tsx
RitaDias: We had a discussion off github and it makes more sense to move all this logic to the hook (since it doesn't need to be a context, that will happen in a separate PR)

packages/sanity/src/core/studio/StudioProvider.tsx
RitaDias: Related to the conversation of the context & hooks, this will likely be removed in a separate PR :)

packages/sanity/src/core/versions/components/dialog/BundleForm.tsx
```diff
@@ -1,43 +1,58 @@
...
 
-      if (isDraftOrPublished(title)) {
+      if (isDraftOrPublished(pickedTitle)) {
         setShowTitleValidation(true)
       } else {
         setShowTitleValidation(false)
       }
 
-      onChange({...value, title: title, name: speakingurl(title)})
+      onChange({...value, title: pickedTitle, name: speakingurl(pickedTitle)})
```
pedrobonamin, line 55: Will create a ticket, I think we should validate releases names are not repeated

RitaDias: definitely ✅

pedrobonamin (approved):
Great job @RitaDias thanks for this!!

PR #7048, opened 2024-07-01
feat(releases): initialising releases plugin

jordanl17:
### Description
Creating the initial routing and basic screens for the new releases studio plugin

Watch the Loom to see a feature runthrough:
https://www.loom.com/share/c96bea029ea847fb9f3d072dce3fe9a6?sid=d3481260-31b5-48e7-b130-d8b1da8c87f5

### What to review
* Creation of the Releases Plugin [here](https://github.com/sanity-io/sanity/pull/7048/files#diff-1bead4ed4351464d263f3f77323ecb77978404b149a80bfc8901e7e7c58b506d))
* The dir structure of the plugin
* The splitting of presentation logic in `BundlesOverview` (`BundlesDetail` only has the header completed)

### Testing
* Tests for the main functionality added in `BundlesOverview`

### Notes for release
N/A - not customer facing


packages/sanity/src/core/releases/i18n/resources.ts
jordanl17, line 1: Added this as a placeholder. Although we might not add any translations just yes, this at least means the bundle can be imported into tests and will cause minimal refactor when they are added

packages/sanity/src/core/releases/tool/ReleasesTool.tsx
```diff
@@ -0,0 +1,14 @@
+import {useRouter} from 'sanity/router'
+
+import {BundleDetail} from './BundleDetail'
+import BundlesOverview from './BundlesOverview'
+
+export function ReleasesTool() {
+  const router = useRouter()
+
+  const {bundleId} = router.state
```
jordanl17, line 9: I'm not sure if this is really the best way of managing routing, but it was how Scheduled Publishing seemed to do it and I couldn't find any other examples

packages/sanity/src/core/versions/components/dialog/CreateBundleDialog.tsx
```diff
@@ -20,10 +25,26 @@ export function CreateBundleDialog(props: {
   })
 
   const handleOnSubmit = useCallback(
-    () => (bundle: Bundle) => {
-      onSubmit(bundle)
+    async (bundle: Bundle) => {
+      setIsCreating(true)
+      try {
+        await createBundle({
```
jordanl17, line 31: Adding in creation to the dialog, as figuring all places where this dialog is used will want identical behavior on bundle form submit. `onCreate` still called after the fact.

Also thinking this might be more ideal to just pass an `open` prop to this dialog and have it control whether it is visible

packages/sanity/src/core/store/bundles/types.ts
```diff
@@ -7,6 +8,7 @@ export interface BundleDocument extends SanityDocument {
   name: string
   description?: string
   tone?: ButtonTone
-  icon?: string
+  icon?: IconSymbol
   authorId: string
+  publishedAt?: Date
```
pedrobonamin, line 13: Even though we will use it as a date, the actual value returned from the document will be a string in this case.

packages/sanity/src/core/releases/utils/shortRelativeDate.ts
```diff
@@ -0,0 +1,36 @@
+import {formatDistanceToNowStrict} from 'date-fns'
+
+function getTimestamp(date: Date | number | string): number {
+  if (typeof date === 'string') return Date.parse(date)
+  if (date instanceof Date) return date.getTime()
+
+  return date
+}
+
+export function shortRelativeDate(date: Date | number | string): string {
```
pedrobonamin, line 10: Appreciate this is added, but I think we could reuse `packages/sanity/src/core/components/RelativeTime.tsx`

```
<RelativeTime time={bundle.publishedAt} useTemporalPhrase />
```

But lets check with the team given it doesn't return the exact same response.

pedrobonamin (approved):
This is fantastyc @jordanl17 thanks a lot! 
Added some suggestions, they are non blocking, also moved those suggestions to a branch (I was testing in the go), see https://github.com/sanity-io/sanity/pull/7050 
If you agree with those changes feel free to merge them and merge this one to `corel`, or discard entirely. 

Let's discuss the shortRelativeDate, I'm worried on introducing a new functionality like this one given I think the `useRealtiveTime` also handles translations.

PR #7049, opened 2024-07-02
feat(core): focus root array item when closing tree editing dialog

hermanwikner:
### Description

This pull request ensures that the root array item is focused when the tree editing dialog is closed. It also adds animation to the tree editing dialog to be consistent with other dialogs in the studio.

Note: there is currently an issue with setting the correct `focusPath` when closing the tree editing dialog without having focused on any input in the form. This will be addressed as a separate issue.

### What to review

- Ensure that the changes are working as expected.

### Testing

- Component test for focusing the root array item when closing the tree editing dialog.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/form/studio/FormBuilder.tsx
+++ b/packages/sanity/src/core/form/studio/FormBuilder.tsx
@@ -1,6 +1,3 @@
-/* eslint-disable camelcase */
-/* eslint-disable react/jsx-handler-names */
-
 import {type ObjectSchemaType, type Path, type ValidationMarker} from '@sanity/types'
 import {useCallback, useMemo, useRef} from 'react'
 
@@ -35,7 +32,12 @@ import {
 import {DocumentFieldActionsProvider} from './contexts/DocumentFieldActions'
 import {FormBuilderInputErrorBoundary} from './FormBuilderInputErrorBoundary'
 import {FormProvider} from './FormProvider'
-import {TreeEditingDialog, TreeEditingEnabledProvider, useTreeEditingEnabled} from './tree-editing'
+import {
+  shouldArrayDialogOpen,
+  TreeEditingDialog,
+  TreeEditingEnabledProvider,
+  useTreeEditingEnabled,
+} from './tree-editing'
 
 /**
  * @alpha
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/form/studio/FormBuilder.tsx
```diff
@@ -306,10 +308,16 @@ function RootInput(props: RootInputProps) {
   const {rootInputProps, onPathOpen, openPath, renderInput} = props
   const treeEditing = useTreeEditingEnabled()
 
+  const open = useMemo(
+    () => shouldArrayDialogOpen(rootInputProps.schemaType, openPath),
+    [openPath, rootInputProps.schemaType],
+  )
+
```
hermanwikner, line 315: I think it makes more sense to conditionally render the dialog with this value here, rather than inside the dialog component itself.

RitaDias:
I might be missing something @hermanwikner but it still looks like it doesn't stay focused?

https://github.com/sanity-io/sanity/assets/6951139/859a0338-3c9a-403a-bab7-dcb53416cf40

hermanwikner:
> I might be missing something @hermanwikner but it still looks like it doesn't stay focused?

Thanks for testing, @RitaDias!

I believe the item is focused, but the focus is not visible when closing the dialog using the mouse. However, if you close it with the "Escape" key, the focus should be visible.

RitaDias:
> > I might be missing something @hermanwikner but it still looks like it doesn't stay focused?
> 
> Thanks for testing, @RitaDias!
> 
> I believe the item is focused, but the focus is not visible when closing the dialog using the mouse. However, if you close it with the "Escape" key, the focus should be visible.

Is this something we want to tackle in this PR? 🤔

hermanwikner:
> I believe the item is focused, but the focus is not visible when closing the dialog using the mouse. However, if you close it with the "Escape" key, the focus should be visible.

Hm, I'm not sure actually 🤔  We usually do not show focus rings when using the mouse, but only when using the keyboard

RitaDias:
> > I believe the item is focused, but the focus is not visible when closing the dialog using the mouse. However, if you close it with the "Escape" key, the focus should be visible.
> 
> Hm, I'm not sure actually 🤔 We usually do not show focus rings when using the mouse, but only when using the keyboard

If you feel like this is expected behavior I'm ok with approving it, I just found it strange because while it's technically focused my eyes didn't see it and I'm wondering if that's going to be the same expectation for a user.
That being said, if you're saying that's how we've been doing things I'm ok with ✅  :)

RitaDias (approved):
✅ For the future us reading this: we spoke on slack, moving forward with the approval

PR #7050, opened 2024-07-02
chore(corel): add eslintrc change, remove progress icon and update types

pedrobonamin:


jordanl17 (approved):

.eslintrc.cjs
```diff
@@ -161,6 +161,16 @@ const config = {
       },
     },
 
+    // Ignore i18n in releases files. TODO: Remove this before moving releases to GA.
```
jordanl17, line 164: I wonder if we should actually just define this `.eslintrc` inside the releases tool eg. `packages/sanity/src/core/releases/.eslintrc.cjs`, and then change the `files` to just everything in the current dir. @pedrobonamin what do you think?

pedrobonamin: COULD: Not sure if that will have some downsides or not and probably we will only use it for this ignore while we are developing the feature.

PR #7061, opened 2024-07-02
feat(core): allow domain sharding

rexxars:
### Description

This is the studio integration side of https://github.com/sanity-io/client/pull/860 - see that PR for rationale.
I'll rebase/change this PR to use a regular release of the client once the above is merged and released.

The implementation works like this:
- We first configure and do the auth check/login on the unsharded domain. This ensures we have the right redirect URLs and such.
- Once authenticated, we fetch the current user at the same time as we do a network check against the `/ping` route.
  - The ping route allows exposed network timing information (through a `Timing-Allow-Origin` header), which allows us to get the HTTP protocol used to access it.
  - If we detect HTTP2/3, we still use unsharded domains.
  - If the request used HTTP1, or we cannot determine the protocol (eg if the resource timing API is not available), we switch the client to sharded domains.

This is only enabled for customers with `allowDomainSharding` set to `true`, as well as the `auth.loginMethod` set to `token`. This is unfortunate, but we cannot get the auth cookie set on the regular project hostname propagated, so we have to use tokens. This does introduce overhead, since all requests will have preflight requests.

Note that this is a temporary workaround, while we work on a more permanent solution.

### What to review

Either disable HTTP2/3 support (through browser flag or similar), or set `FORCE_DOMAIN_SHARDING` to `true` in `createAuthStore` and observe domain sharding behavior in staging. Setting the `FORCE_DOMAIN_SHARDING` flag back to `false` (or re-enabling HTTP2/3) and observe that it does _not_ use domain sharding. 

### Testing

I realize this is missing tests. Since this is a temporary workaround, I find this acceptable for now. But if you feel strongly, shout out.

### Notes for release

None. This is a silent release - we only want select customers to use this as a workaround, not a permanent solution.

### DIFF
--- a/dev/test-studio/package.json
+++ b/dev/test-studio/package.json
@@ -23,7 +23,7 @@
     "@react-three/fiber": "^8.13.6",
     "@sanity/assist": "^3.0.2",
     "@sanity/block-tools": "3.49.0",
-    "@sanity/client": "^6.20.0",
+    "@sanity/client": "6.20.2-beta.2",
     "@sanity/color": "^3.0.0",
     "@sanity/google-maps-input": "^4.0.0",
     "@sanity/icons": "^3.0.0",
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -157,6 +157,8 @@ export default defineConfig([
     dataset: 'test',
     plugins: [sharedSettings()],
     basePath: '/test',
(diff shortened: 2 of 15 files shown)


annez:
@rexxars is this still relevant? Can you close if not?

jordanl17:
As the client PR has not been merged, closing this until that is merged and linking both together

PR #7062, opened 2024-07-02
fix(cli): mock `document.execCommand` when emulating browser env

rexxars:
### Description

A user reported importing the `quill` dependency in some plugin/schema, which (on startup) calls `document.execCommand`, which does not exist in our mocked environment and thus crashes.

This PR mocks the function as a no-op.

### What to review

`sanity schema extract` works as before. Adding a `document.execCommand()` call in a schema file does not crash.

### Testing

The browser env mocking does not have tests currently, and it throws when trying to run it in jest because of the odd worker  setup they are using. I couldn't find an immediate and clean way to test this - perhaps we could take a stab at some separated tests that runs in regular node environment at some point?

### Notes for release

- Fixes an issue when trying to extract schemas/deploy GraphQL APIs when a subdependency tries to call `document.execCommand` at import time

### DIFF
--- a/packages/sanity/src/_internal/cli/util/mockBrowserEnvironment.ts
+++ b/packages/sanity/src/_internal/cli/util/mockBrowserEnvironment.ts
@@ -72,13 +72,31 @@ const getFakeGlobals = (basePath: string) => ({
     })),
 })
 
+const getFakeDocumentProps = () => ({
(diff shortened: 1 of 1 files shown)


binoy14 (approved):
Tested both scenarios and works as expected. Thanks!

PR #7068, opened 2024-07-03
feat(cli): Add support for exporting dataset with cursor

sgulseth:
### Description

A new mode in the export endpoint allows streaming incrementally with a cursor

### Testing

Added a CLI test to make sure it picks it up 👍 

### Notes for release

Export: Add new export mode that can uses a cursor when exporting, this might be more performant for larger datasets, but might not be as accurate if the dataset is being modified during export. Defaults to stream

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/dataset/exportDatasetCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/dataset/exportDatasetCommand.ts
@@ -20,6 +20,7 @@ Options
   --types                   Defines which document types to export
   --overwrite               Overwrite any file with the same name
   --asset-concurrency <num> Concurrent number of asset downloads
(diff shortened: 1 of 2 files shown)


packages/sanity/src/_internal/cli/commands/dataset/exportDatasetCommand.ts
```diff
@@ -36,6 +37,7 @@ interface ExportFlags {
   'overwrite'?: boolean
   'types'?: string
   'asset-concurrency'?: string
+  'mode'?: boolean
```
j33ty, line 40: Mode is not a boolean?

sgulseth: good catch!

j33ty (approved):
LGTM

PR #7069, opened 2024-07-04
fix(core): handle array fields within nested objects in array items

hermanwikner:
### Description

This pull request fixes an issue where the tree editing state builder does not handle array fields nested inside array items within multiple levels of objects. Specifically, the issue arises with the following structure:

- Array field
  - Object 1
    - Object 2
      - Object 3
        - Array field `<-- This array field can't be resolved ❌`
          - (item)
          - (item)
          - (item)
  - Object 2
    - (item)
    - (item)

For example, consider the following schema definition. In this schema, the `nestedArray` field is nested within `secondObject`, which is nested within `firstObject`. The current tree state building logic cannot handle this level of deep object nesting, leading to unresolved tree states for such structures.

```ts
import {defineField} from 'sanity'

const field = defineField({
  name: 'myArray',
  type: 'array',
  of: [
    {
      type: 'object',
      name: 'firstObject',
      fields: [
        {
          type: 'object',
          name: 'secondObject',
          fields: [
            {
              type: 'array',
              name: 'nestedArray',
              of: [
                {
                  type: 'object',
                  name: 'nestedObject',
                  fields: [
                    {
                      type: 'string',
                      name: 'nestedString',
                    },
                  ],
                },
              ],
            },
          ],
        },
      ],
    },
  ],
})

```

### What to review

- Ensure that the tree editing feature works properly with array fields within nested objects in array items
- Verify that navigating with the tree menu works as expected
- General code review

### Testing

- Unit tests added for new functions (`findArrayTypePaths`)
- Unit tests updated for existing functions (`buildTreeEditingState`)
- Component test added for tree editing with array fields within nested objects in array items (`TreeEditingNestedObjects`)

### Notes for release

N/A – bug fix


RitaDias:
An odd thing I've noticed which I think is a bug and should be fixed:
commenting on deeply nested fields doesn't seem to allow/create them?

https://github.com/sanity-io/sanity/assets/6951139/fa38e348-b21e-4311-8de7-495cfc86f9c1

RitaDias:
The comments still work on the first level
image

packages/sanity/src/core/form/studio/tree-editing/utils/findArrayTypePaths.ts
RitaDias, line 1: Actually wonder if we should move this to `paths.ts`

This feels like it would be very useful for a general path thing and not focused on tree-editing. thoughts?

hermanwikner:
> @hermanwikner An odd thing I've noticed which I think is a bug and should be fixed: commenting on deeply nested fields doesn't seem to allow/create them?
> 
>  Screen.Recording.2024-07-08.at.07.57.58.mov

I agree, that should be fixed! However, it's an already existing issue and not something that this pull request introduces @RitaDias.

packages/sanity/src/core/form/studio/tree-editing/utils/findArrayTypePaths.ts
RitaDias: @hermanwikner ?

RitaDias (approved):
✅
Just create a new story for the bug we found and also another to hopefully make the method we spoke about more generic to move to the paths :) thank you!

PR #7070, opened 2024-07-04
docs(migrate): updating insert examples to showing always 3 args

jordanl17:
### Description
The TSDoc for the `insert` method incorrectly gives examples of 2 provided args


binoy14 (approved):
Looks aligned with the usage, thank you!

PR #7071, opened 2024-07-04
refactor(COREL): refactor version context & renaming

RitaDias:
### Description

**Quality of life:**
- Remove unneeded comments
- Remove VersionContext + Provider Move context logic to already existing `usePerspective`
- Rename `currentVersion` -> `currentGlobalBundle` (used for more than just perspective, needing name, title, etc)
- Rename `setCurrentVersion` -> `setPerspective` (used only for changing the perspective in the studio)

**Renaming & structure:**
- Rename `versions` directory to `bundles`
- Rename `GlobalBlundeMenu` -> `GlobalPerspectiveMenu` and move it to navbar/perspective
- Rename `DocumentVersionMenu` -> `DocumentPerspectiveMenu` and move it to DocumentPanel/header/perspective


pedrobonamin (commented):
I wonder if we need the `isDraft` property in the `usePerspective` hook, what will happen and how will we refactor this later if we want to introduce drafts to versions.
If we want to be explicit we could do this check in the locations needed


**This is a suggestion open to discussion.** 
For what I see we could remove the isDraft and replace like this:
```diff
// packages/sanity/src/core/studio/components/navbar/perspective/GlobalPerspectiveMenu.tsx
        button={
          <Button mode="bleed" padding={0} radius="full">
            <BundleBadge
              hue={hue}
-            icon={isDraft ? undefined : icon}
+            icon={icon} // Icon is already undefined in LATEST
              openButton
              padding={2}
-             title={isDraft ? LATEST.title : title}
+            title={title} // Reuse LATEST.title without being explicit about it.
            />
          </Button>
        }


// packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
-  {currentGlobalBundle && !isDraft && (
+  {currentGlobalBundle && currentGlobalBundle.name !== LATEST.name && (
        <BundleBadge hue={hue} title={title} icon={icon} padding={2} />
      )}

```

RitaDias:
> I wonder if we need the `isDraft` property in the `usePerspective` hook, what will happen and how will we refactor this later if we want to introduce drafts to versions. If we want to be explicit we could do this check in the locations needed
> 
> **This is a suggestion open to discussion.** For what I see we could remove the isDraft and replace like this:
> 
> ```diff
> // packages/sanity/src/core/studio/components/navbar/perspective/GlobalPerspectiveMenu.tsx
>         button={
>           <Button mode="bleed" padding={0} radius="full">
>             <BundleBadge
>               hue={hue}
> -            icon={isDraft ? undefined : icon}
> +            icon={icon} // Icon is already undefined in LATEST
>               openButton
>               padding={2}
> -             title={isDraft ? LATEST.title : title}
> +            title={title} // Reuse LATEST.title without being explicit about it.
>             />
>           </Button>
>         }
> 
> 
> // packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
> -  {currentGlobalBundle && !isDraft && (
> +  {currentGlobalBundle && currentGlobalBundle.name !== LATEST.name && (
>         <BundleBadge hue={hue} title={title} icon={icon} padding={2} />
>       )}
> ```

Very good point, I'll go through and make this change :)

pedrobonamin:
> > I wonder if we need the `isDraft` property in the `usePerspective` hook, what will happen and how will we refactor this later if we want to introduce drafts to versions. If we want to be explicit we could do this check in the locations needed
> > **This is a suggestion open to discussion.** For what I see we could remove the isDraft and replace like this:
> > ```diff
> > // packages/sanity/src/core/studio/components/navbar/perspective/GlobalPerspectiveMenu.tsx
> >         button={
> >           <Button mode="bleed" padding={0} radius="full">
> >             <BundleBadge
> >               hue={hue}
> > -            icon={isDraft ? undefined : icon}
> > +            icon={icon} // Icon is already undefined in LATEST
> >               openButton
> >               padding={2}
> > -             title={isDraft ? LATEST.title : title}
> > +            title={title} // Reuse LATEST.title without being explicit about it.
> >             />
> >           </Button>
> >         }
> > 
> > 
> > // packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
> > -  {currentGlobalBundle && !isDraft && (
> > +  {currentGlobalBundle && currentGlobalBundle.name !== LATEST.name && (
> >         <BundleBadge hue={hue} title={title} icon={icon} padding={2} />
> >       )}
> > ```
> 
> Very good point, I'll go through and make this change :) Since we have already a `isDraftOrPublished` I'll probably use also a helper method like that. ~I actually found a `isDraft` that I can import already from sanity so I think that's what I'll do~ actually not using that because it immediately started giving some issues related to typing, not sure we want to mix this. So I think it's fine to create a new one

I think we probably don't need the `isDraft`check at all for this and we probably don't want to attach logic to the draft check because it can be confusing later if we want to introduce drafts into versions.

RitaDias:
> > > I wonder if we need the `isDraft` property in the `usePerspective` hook, what will happen and how will we refactor this later if we want to introduce drafts to versions. If we want to be explicit we could do this check in the locations needed
> > > **This is a suggestion open to discussion.** For what I see we could remove the isDraft and replace like this:
> > > ```diff
> > > // packages/sanity/src/core/studio/components/navbar/perspective/GlobalPerspectiveMenu.tsx
> > >         button={
> > >           <Button mode="bleed" padding={0} radius="full">
> > >             <BundleBadge
> > >               hue={hue}
> > > -            icon={isDraft ? undefined : icon}
> > > +            icon={icon} // Icon is already undefined in LATEST
> > >               openButton
> > >               padding={2}
> > > -             title={isDraft ? LATEST.title : title}
> > > +            title={title} // Reuse LATEST.title without being explicit about it.
> > >             />
> > >           </Button>
> > >         }
> > > 
> > > 
> > > // packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
> > > -  {currentGlobalBundle && !isDraft && (
> > > +  {currentGlobalBundle && currentGlobalBundle.name !== LATEST.name && (
> > >         <BundleBadge hue={hue} title={title} icon={icon} padding={2} />
> > >       )}
> > > ```
> > 
> > 
> > Very good point, I'll go through and make this change :) Since we have already a `isDraftOrPublished` I'll probably use also a helper method like that. ~I actually found a `isDraft` that I can import already from sanity so I think that's what I'll do~ actually not using that because it immediately started giving some issues related to typing, not sure we want to mix this. So I think it's fine to create a new one
> 
> I think we probably don't need the `isDraft`check at all for this and we probably don't want to attach logic to the draft check because it can be confusing later if we want to introduce drafts into versions.

That is true 🤔

pedrobonamin (approved):
Thanks Rita! Looks good to me

PR #7074, opened 2024-07-05
fix(corel): update uses of collate, add typeguard for bundles and add release tag to addonDatasetProvider

pedrobonamin:
Fixes build issues in COREL branch.


packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
```diff
@@ -201,7 +201,7 @@ export function referenceSearch(
   })
   return search(textTerm, {includeDrafts: true}).pipe(
     map(({hits}) => hits.map(({hit}) => hit)),
-    map(collate),
+    map((docs) => collate(docs)),
```
jordanl17, line 204: Q: so this wasn't previously working...?

jordanl17:
@pedrobonamin seems to still be some linting errors. Is the intention of this change to resolve that also?

packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
pedrobonamin: It was working, types were throwing.
in a map, the first parameter is the item, the second is a number with the index.

When doing  `map(collate)` it's the same as doing `map((item, index) => collate(item, index))`

And collate was updated to receive an object as second parameter https://github.com/sanity-io/sanity/blob/fix-build/packages/sanity/src/core/util/draftUtils.ts#L142-L143

That's why it was showing a ts error

jordanl17: Makes sense! Thanks

jordanl17 (approved):
Nice 🔥

RitaDias (approved):
🚀 let's go

PR #7077, opened 2024-07-07
refactor(sanity): move bundle metadata store to resource cache

juice49:
### Description

This branch moves the bundle metadata store to the resource cache and removes the `BundlesContext` context, which is no longer required.

The bundle metadata store is initialised lazily when first subscribed to. Upon subscription, it will fetch a list of bundles and create a listener to keep the locally held state fresh.

The store is not disposed of when all subscriptions are closed. After it has been initialised, it will keep listening for the duration of the app's lifecycle. Subsequent subscriptions will be given the latest state upon subscription.

Note: the `packages/sanity/src/core/releases/tool/__tests__/BundlesOverview.test.tsx` test is failing, but this seems to have been occurring prior to the change.


packages/sanity/src/core/store/bundles/createBundlesStore.ts
RitaDias, line 1: @juice49 forgive me if I'm mixing concepts: would it be here that we should do the [fetch for the existing versions of documents](https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/core/bundles/util/dummyGetters.ts) like @pedrobonamin brought up as a question in an earlier PR?

packages/sanity/src/core/releases/plugin/ReleasesStudioLayout.tsx
```diff
@@ -1,18 +1,8 @@
...
   const {enabled, mode} = {enabled: true, mode: 'default'}
 
-  if (!enabled) {
-    return props.renderDefault(props)
-  }
-
-  return (
-    <AddonDatasetProvider>
-      <BundlesProvider>{props.renderDefault(props)}</BundlesProvider>
-    </AddonDatasetProvider>
```
pedrobonamin, line 16: We need to think on how to inject the `AddonDatasetProvider` into the studio when needed. If we remove it from here and the user decides to only include the `releases` plugin, but not `comments` or `tasks` the `useAddonDataset` hook will crash in https://github.com/sanity-io/sanity/pull/7077/files#diff-e6d5f6b7c44a215f6830735c676cbb63da8df1457870f3b44897293679dff703R283

packages/sanity/src/core/store/bundles/createBundlesStore.ts
pedrobonamin: This bundles store fetches the bundle documents from the addon dataset only. 
We will need to fetch the `version` documents from other location, because we could in theory have a `version` document without a bundle

packages/sanity/src/core/store/bundles/createBundlesStore.ts
```diff
@@ -0,0 +1,234 @@
...
+          payload: {
+            loading: false,
+            error,
+          },
+        },
+      }),
+    ),
+    // Skip the first event received upon subscription. This `welcome` event would cause the list
+    // to be fetched again, which is not desirable immediately after the initial fetch has occurred.
+    skip(1),
```
pedrobonamin, line 161: This is a very nice detail

packages/sanity/src/core/store/bundles/createBundlesStore.ts
```diff
@@ -0,0 +1,234 @@
...
+        return of(undefined)
+      },
+    ),
+  )
+
+  const state$ = merge(listFetch$, listener$, dispatch$).pipe(
+    filter((action): action is bundlesReducerAction => typeof action !== 'undefined'),
+    scan((state, action) => bundlesReducer(state, action), INITIAL_STATE),
+    startWith(INITIAL_STATE),
+    shareReplay(1),
```
pedrobonamin, line 227: Curious on what the `1` is doing here. Checking the docs https://rxjs.dev/api/index/function/shareReplay takes an object as parameter with the signature [ShareReplayConfig](https://rxjs.dev/api/index/interface/ShareReplayConfig)

packages/sanity/src/core/store/bundles/useBundles.ts
```diff
@@ -0,0 +1,26 @@
...
+  data: BundleDocument[] | null
+  error?: Error
+  loading: boolean
+  dispatch: React.Dispatch<bundlesReducerAction>
+}
+
+export function useBundles(): BundlesState {
+  const {state$, dispatch} = useBundlesStore()
+  const state = useObservable(state$)!
+  const bundlesAsArray = useMemo(() => Array.from(state.bundles.values()), [state.bundles])
```
pedrobonamin, line 18: Given we now have all of this in rxjs observable and pipe
What do you think if we do this transformation in the observer when it emits?

We could update https://github.com/sanity-io/sanity/pull/7077/files#diff-0972633c3ddf87121161cee5390c665b6e88e1318ba04827eca1dd6ca54ac0fdR223-R227 to:

```ts
  const state$ = merge(listFetch$, listener$, dispatch$).pipe(
    filter((action): action is bundlesReducerAction => typeof action !== 'undefined'),
    scan((state, action) => bundlesReducer(state, action), INITIAL_STATE),
    startWith(INITIAL_STATE),
    map((state) => ({
      ...state,
      bundles: Array.from(state.bundles.values()),
    })),
    shareReplay(1),
  )
```

packages/sanity/src/core/releases/plugin/ReleasesStudioLayout.tsx
juice49: That's a great point, sorry for missing that.

I'm not very familiar with this part of the codebase yet, but do you think it would be plausible to move the addon dataset provider to the resource cache, too? That way we could always include it, but it would be lazily initialised once any plugin tried to access it.

packages/sanity/src/core/store/bundles/createBundlesStore.ts
juice49: Great question! The function can be called with a config object, but if you have no other options to provide, you can instead provide a `bufferSize` argument directly.

In this case, a `bufferSize` of 1 means only the most recently emitted state object will be stored and broadcast to new subscribers.

packages/sanity/src/core/store/bundles/useBundles.ts
juice49: I agree it would be neater, I just didn't want to move around _too much_ that wasn't strictly necessary here.

One thing that gives me pause here is that the bundles array is derived state. Does it make sense to store this duplicate/derived state in the store itself?

One other tiny concern is that it might cause more work and renders on every state change. With the `useMemo` approach, I think changes to things like the loading state would preserve the identity of the `bundles` object and prevent the array being recalculated, whereas the change suggested would cause it to be recalculated and produce a new array on every state change.

Not to say we shouldn't do something like this, I'm just not 100% sure what the best approach is.

packages/sanity/src/core/releases/plugin/ReleasesStudioLayout.tsx
pedrobonamin: Now that I know how the resources cache works I like that idea!

packages/sanity/src/core/store/bundles/createBundlesStore.ts
pedrobonamin: TIL: Thanks @juice49 !

packages/sanity/src/core/store/bundles/useBundles.ts
pedrobonamin: Makes sense, I didn't considered that other changes will emit a new array causing re renders, you are right, let's keep the change as is.

PR #7087, opened 2024-07-08
chore(renovate): add preset that allows internal PRs to open right away

stipsan:
Reduces the maintenance burden of constantly going to https://github.com/sanity-io/sanity/issues/4030 and check off important dependencies that should at all times be up to date on the monorepo, in order to catch regression bugs before our users does


pedrobonamin (approved):
Looks good to me, thanks @stipsan !

PR #7093, opened 2024-07-08
feat(core): clickable links in comment message

hermanwikner:
### Description

This pull request makes links in comment messages clickable by transforming text that starts with 'http://', 'https://', or 'www' into anchor tags.

links

### What to Review

- Ensure that links become clickable in the comment message.
- A simple middleware pattern was created to transform the children in the comment message. At this point, it is not very necessary as we are only transforming links, but it allows for easily adding further transformations of the children in the future. Does it make sense?

### Testing

- Unit test the `transformChildren` logic. Are there any cases missing that the test should cover?

### Notes for Release

Make links in comment messages clickable.


packages/sanity/src/core/comments/components/pte/CommentMessageSerializer.tsx
```diff
@@ -19,39 +21,48 @@ const PortableTextWrap = styled(Stack)(() => {
...
-    code: ({children}) => <>{children}</>,
-    underline: ({children}) => <>{children}</>,
-    strikeThrough: ({children}) => <>{children}</>,
-    link: ({children}) => <>{children}</>,
+    strong: NormalBlockTransformed,
+    em: NormalBlockTransformed,
+    code: NormalBlockTransformed,
+    underline: NormalBlockTransformed,
+    strikeThrough: NormalBlockTransformed,
+    link: NormalBlockTransformed,
```
hermanwikner, line 63: Change to `Fragment`

packages/sanity/src/core/comments/utils/transform-children/linkMiddleware.ts
```diff
@@ -0,0 +1,30 @@
+import {createElement, type ReactNode} from 'react'
+
+import {type Middleware} from './types'
+
+const URL_REGEX = /\b(https?:\/\/[^\s,()]+(?:\.[^\s,()]+)*|www\.[^\s,()]+\.[^\s,()]{2,})\b/g
+
+function createLinkElement(url: string): ReactNode {
+  const href = url.startsWith('http') ? url : `https://${url}`
+
+  return createElement('a', {href, target: '_blank', rel: 'noopener noreferrer', key: url}, url)
```
hermanwikner, line 10: Stop propagation to avoid triggering click on the comment item when clicking the link?

packages/sanity/src/core/comments/utils/transform-children/index.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function transformChildren(node: ReactNode): ReactNode {
+  if (!Array.isArray(node)) return node
+
+  return node.flatMap((item) => {
+    let transformedItem: ReactNode[] = [item]
+
+    for (const middleware of middlewares) {
+      transformedItem = transformedItem.flatMap(middleware)
+    }
```
RitaDias, line 19: Forgive me, can you just give me a run down why we need to do this?

hermanwikner: We are iterating over middleware functions that transform the nodes and pass them on to the next function. Currently, we only have one function (`linkMiddleware`), but in the future, we might have other functions to perform further transformations. 

For example:

```tsx
const middlewares: Middleware[] = [linkMiddleware, studioLinksMiddleware]
```

It might be a bit overkill, but the intent was to keep `transformChildren` a bit generic and not having it only focus on the links.

RitaDias: Makes sense, thank you :)

RitaDias (approved):
🚀

PR #7092, opened 2024-07-08
feat(corel): releases details screen.

pedrobonamin:
### Description

Adds releases detail screen.
See video for details.
screencapture-localhost-3333-test-releases-pedro-summer-2024-07-08-14_32_08

**Functionalities**
https://www.loom.com/share/c3678eb1171444c1aaa367bd333a1bc7?from_recorder=1&focus_title=1

**Filter**
https://www.loom.com/share/57d2f785248e44b48458cb588e761bca?from_recorder=1&focus_title=1

PR #7095, opened 2024-07-08
feat(router): update router to support query params in intent links

pedrobonamin:
### Description
As part of releases we are introducing a query param that needs to be persisted in the url, `perspective`

It will determine which `perspective` the user has selected and we need to use the `IntentLink` component to change the perspective the user has selected when navigating to a version document.

How it will be used:
```ts
          <IntentLink
            {...linkProps}
            intent="edit"
            params={{
              id: getPublishedId(documentId, true),
              type: documentTypeName,
            }}
            searchParams={[['perspective', `bundle.${release.name}`]]}
            ref={ref}
          />
```
Path example: `/test/intent/edit/id=c0c99e43-40de-4d8b-abec-a710d5752ea2;type=book/?perspective=bundle.pedro-summer` 

see https://github.com/sanity-io/sanity/pull/7092/files#diff-729068b8f99a9501225501d3379505aa1a85162b6e41aa9654b7c5aa8a83d067R122-R139 for further details.

### What to review

Is this change correct? Is there other way to handle this?

### Testing

Added tests for `IntentLink` and for `resolveIntentState` function

### Notes for release

Adds support in intent links for `_searchParams: SearchParam[]`.

### DIFF
--- a/packages/sanity/src/core/studio/router/helpers.ts
+++ b/packages/sanity/src/core/studio/router/helpers.ts
@@ -125,6 +125,7 @@ export function resolveIntentState(
 
     const nextUrlState: Record<string, unknown> = {
       ...prevState,
+      _searchParams: nextState._searchParams,
       tool: matchingTool.name,
       [matchingTool.name]: toolState,
     }
--- a/packages/sanity/src/core/studio/router/helpers.test.ts
+++ b/packages/sanity/src/core/studio/router/helpers.test.ts
@@ -1,0 +1,39 @@
+import {describe, it} from '@jest/globals'
+import {expect} from '@playwright/experimental-ct-react'
+import {type Tool} from 'sanity'
+import {type RouterState} from 'sanity/router'
+
+import {resolveIntentState} from './helpers'
+
+describe('resolveIntentState', () => {
+  const testTool: Tool = {
+    name: 'test',
+    title: 'Test tool',
+    component: () => null,
+    canHandleIntent: () => true,
+    getIntentState: () => ({}),
+  }
+
+  it('should resolve intent state with query params', () => {
+    const state: RouterState = {
+      intent: 'edit',
+      params: {
+        id: 'p-bay-area-san-francisco-2022-08-17-2022-08-17',
+        type: 'playlist',
+      },
+      _searchParams: [['perspective', 'bundle.pedro-summer']],
+    }
+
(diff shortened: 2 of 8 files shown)


packages/sanity/src/core/studio/router/helpers.ts
```diff
@@ -125,6 +125,7 @@ export function resolveIntentState(
 
     const nextUrlState: Record<string, unknown> = {
       ...prevState,
+      _searchParams: nextState._searchParams,
```
binoy14, line 128: Wondering, what's the reason to use `_searchParams` instead of `searchParams`

binoy14 (commented):
Couple questions

1. Is there a reason there would not be a need for release notes for this change?
2. Is it possible to add tests for this change? Not sure if there are existing tests for Router that you can piggy back on. 

Looks good otherwise besides the comments

packages/sanity/src/core/studio/router/helpers.ts
pedrobonamin: This is because the router expects the search params to be in a `_searchParams` property.
I've added an update to the `RouterEvent` interface to reflect that the state returned from this function should be a RouterState.
```diff
export interface RouterStateEvent {
  type: 'state'
+ state: RouterState
- state: Record<string, unknown> 
  isNotFound: boolean
}
```

RouterState is the following:
`export type RouterState = Record<string, unknown> & {_searchParams?: SearchParam[]}
`

pedrobonamin: https://github.com/sanity-io/sanity/pull/7095/commits/646f433231a5f0ca8f0fd9baacc195e539c4257e

pedrobonamin:
> Couple questions
> 
> 1. Is there a reason there would not be a need for release notes for this change?
> 2. Is it possible to add tests for this change? Not sure if there are existing tests for Router that you can piggy back on.
> 
> Looks good otherwise besides the comments

Thanks for the review @binoy14 .

Added test for the change in the IntentLink component, that now supports query params and added test for the resolveIntentState function that it now supports query params

packages/sanity/src/core/studio/router/helpers.ts
binoy14: understood, thanks for confirming

binoy14 (approved):
Makes sense to me, thanks!

PR #7097, opened 2024-07-09
refactor(sanity): remove "features"

RitaDias:
### Description

As spoken with Rune, remove the `features` object and rely instead on the `beta` object.

### What to review

- Make sure that the code makes sense
- Try out the feature and make sure that everything is still working as expected

### Testing

There are already automated tests for this (and of the feature itself)

### Notes for release

N/A shadow release related


hermanwikner (approved):
Looks great, thanks @RitaDias!

PR #7107, opened 2024-07-09
feat(corel): remove version documents when deleting bundle

pedrobonamin:
### Description
When a bundle document is removed the related version documents in the main dataset should also be removed.
This PR adds that functionality and implements a confirmation modal before deleting everything, given this is a destructive operation.

Screenshot 2024-07-09 at 17 18 33

🟡 Missing changes due to the listener updates for the related documents count 🟡

### DIFF
--- a/packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
+++ b/packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
@@ -1,8 +1,9 @@
 import {ArchiveIcon, EllipsisHorizontalIcon, TrashIcon} from '@sanity/icons'
-import {Button, Menu, MenuButton, MenuItem, Spinner} from '@sanity/ui'
+import {Button, Menu, MenuButton, MenuItem, Spinner, Text, useToast} from '@sanity/ui'
 import {useState} from 'react'
 import {useRouter} from 'sanity/router'
 
+import {Dialog} from '../../../../ui-components'
 import {type BundleDocument} from '../../../store/bundles/types'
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
```diff
@@ -15,18 +16,32 @@ export const BundleMenuButton = ({bundle}: Props) => {
...
-    if (router.state.bundleId) {
-      // navigate back to bundle overview
-      router.navigate({bundleId: undefined})
+    try {
+      setDiscardStatus('discarding')
+      await deleteBundle(bundle)
+      setDiscardStatus('idle')
+      if (router.state.bundleId) {
+        // navigate back to bundle overview
+        router.navigate({bundleId: undefined})
```
RitaDias, line 33: Just to double check: should we use this router navigate vs the sticky params we have so far when setting perspective and such?

RitaDias: (I know it was already here before and that's not the change per se, but I just wanted to get a check)

pedrobonamin: This is for the `releases` tool router. It will navigate back to the releases overview screen

pedrobonamin: If you are in the releases detail screen and remove that release, it will send you back to the overview

RitaDias (approved):
🚀

PR #7117, opened 2024-07-09
fix(i18n): allow locale plugins to translate "All fields" field group

rexxars:
### Description

The "All fields" field group is currently not translatable - this is an oversight on our part.
This PR allows field groups the ability to specify i18n titles, in the same way that certain other schema level properties can (sort orders, for instance).

Additionally, it introduces a new i18n key: `inputs.object.field-group-tabs.all-fields-title` - which is used by the default "All fields" group.

### What to review

- The studio builds as before
- Navigating to the new "[field groups with i18n](https://test-studio-git-fix-i18n-all-fields-title.sanity.build/test/structure/input-debug;field-groups;fieldGroupsWithI18n;5bdf0fb2-84ae-4fbe-b0e7-717681ef8e11)" debug type should show three field groups; `🇺🇸 Group 1` (translated through `i18n.title`), `intentionally-missing-key` (defined through `i18n.title`, but does not exist, ergo not translated) and `🌐 Non-i18n group` (title set directly on `title` property, no i18n in play)

### Testing

Added basic e2e tests that show the expected outcome on the i18n field group debug schema test. 

### Notes for release

None, for now. Since users cannot currently switch between locales, it makes more sense to wait to document this when we _do_ allow users to switch. Right now it is easier to just define the localized title straight on the `title` property.

This still has value for locale packs, but those should automatically pick up and AI-translate the "All fields" group after release.

### DIFF
--- a/dev/studio-e2e-testing/i18n/bundles.ts
+++ b/dev/studio-e2e-testing/i18n/bundles.ts
@@ -1,0 +1,12 @@
+import {defineLocaleResourceBundle} from 'sanity'
+import {testStudioLocaleNamespace} from 'sanity-test-studio/locales'
+
+export const e2eI18nBundles = [
+  defineLocaleResourceBundle({
+    locale: 'en-US',
+    namespace: testStudioLocaleNamespace,
+    resources: {
+      'field-groups.group-1': '🇺🇸 Group 1',
+    },
+  }),
+]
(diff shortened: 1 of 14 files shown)


cngonzalez (approved):
Changes make sense to me, thank you!

PR #7118, opened 2024-07-09
test(cli): use separate tarball location for cursor export test

rexxars:
### Description

We're seeing some [new failures](https://github.com/sanity-io/sanity/actions/runs/9865117760/job/27241410606?pr=7117) in the CLI tests intermittently. After some brief investigation, I think perhaps it is caused by the same tarball location being used for both the regular and cursor based export tests, which can run concurrently.

This PR changes the filename used by the cursor based export test, to ensure they do not interfere with eachother.

### Testing

Tests should still pass, but be more stable

### Notes for release

None


j33ty (approved):
Nice catch.

PR #7119, opened 2024-07-10
test(COREL): add core tests

RitaDias:
### Description

- Add createWrapper in core (with resource options)
- Add bundleForm test
- Add BundleIconEditorPicker test
- Add createBundleDialog test
- Add BundleMenu test
- Replace createWrapper in the releases tool with the new one.

Left some components out since they were either:
- Too simple / UI focused
- Will eventually change

### What to review
- Make sure that the changes to the releases plugin makes sense

A lot of these were taken with inspiration from the tests that Jordan has been writing, so thank you 🙏 🥺

### DIFF
--- a/packages/sanity/src/core/bundles/components/BundleMenu.tsx
+++ b/packages/sanity/src/core/bundles/components/BundleMenu.tsx
@@ -48,27 +48,30 @@ export function BundleMenu(props: BundleListProps): JSX.Element {
     <>
       <MenuButton
         button={button}
-        id="global-version-menu"
+        id="bundle-menu"
         menu={
-          <StyledMenu>
+          <StyledMenu data-testid="bundle-menu">
             {loading ? (
-              <Flex padding={4} justify="center">
+              <Flex padding={4} justify="center" data-testid="spinner">
                 <Spinner muted />
               </Flex>
             ) : (
               <>
                 <MenuItem
                   iconRight={
-                    currentGlobalBundle.name === LATEST.name ? <CheckmarkIcon /> : undefined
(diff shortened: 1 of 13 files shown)


packages/sanity/src/core/releases/tool/BundleDetail.tsx
```diff
@@ -0,0 +1,104 @@
+import {ArrowLeftIcon, PublishIcon, StarIcon} from '@sanity/icons'
```
pedrobonamin, line 1: This file was changed for `ReleaseDetail` https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/core/releases/tool/detail/ReleaseDetail.tsx

We should remove this and update the tests

packages/sanity/src/core/store/bundles/createBundlesStore.ts
```diff
@@ -171,7 +172,7 @@ export function createBundlesStore(context: {client: SanityClient | null}): Bund
         // After successful reconnection, fetch the list. Note that the first event is skipped, so
         // this will not occur upon initial connection.
         if (event.type === 'welcome') {
-          return listFetch$
+          return listFetch$.pipe(delay(1_000))
```
pedrobonamin, line 175: Curious on why this change is needed

RitaDias: This looks like part of a rebase that has gone wrong 🫠 
hold please...

pedrobonamin (approved):
Thanks for this @RitaDias

PR #7128, opened 2024-07-10
chore(renovate): only bump esbuild within `0.21.x`

stipsan:
Following up on #7126 this changes renovate so it doesn't keep creating PRs that bump `esbuild` minor (minor are breaking btw) or major versions: https://github.com/sanity-io/sanity/pull/7127


rexxars (approved):
Thanks!

PR #7131, opened 2024-07-10
fix(core): error reporting consent tweaks

rexxars:
### Description

This started with me discovering a warning being printed for `useClient` being called without an API version, which I tracked down to the `EnableErrorReporting` component. Calling `useClient().withConfig()` has two drawbacks:
- A warning is printed - an API version needs to be specified to silence it. The rationale is to encourage people to specify an API version, not depending on a globally set version for the studio. The `withConfig()` call following the `useClient()` would have taken care of that, but does so less efficiently 👇 
- The client won't be memoized. `useClient` internally keeps track of different API versions and reuses clients where possible. `withConfig()` always clones a client, creating a new instance.

It took me a few seconds to realize that `EnableErrorReporting` was a React component, so I added some documentation explaining it, and also renamed it to `MaybeEnableErrorReporting`. While not a very elegant name, it better conveys that we're not _always_ enabling error reporting, rather we are checking and conditionally enabling it. Open to suggestions, but the "maybe" naming pattern is something I've used/seen in other places.

Looking at the code, I also saw that the request was not cancelled upon unmount. Edge case, but that _could_ cause issues. If you have two workspaces using different project IDs, switching to one before the first telemetry request completed could cause a race condition, where the error reporting is enabled/disabled based on the oldest request (for a different project ID). Aside from the edge case, it's also just a bit tidier. Refactored to an observable since I find them easier to deal with in terms of cancellation.

When checking that the request still functioned as expected, I also noticed a difference in the API version used for checking consent in the "telemetry store" and in the error reporting, so I figured it would make sense to align these. Additionally, adding a request tag to all the consent checks makes it easier to tell where they are coming from in my opinion, so added those too.

### What to review

- No warning is printed about using `useClient()` without specifying an API version
- Consent check still works as expected

### Testing

Relying on existing tests.

### Notes for release

- Fixes a warning being printed about using `useClient()` without specifying an API version


cngonzalez (approved):
Ah! Apologies for the noise, and thanks for the detailed rundown of client use. The observable refactor also makes sense to me.

PR #7148, opened 2024-07-12
Corel 41

RitaDias:
### Description

- Update UI for Error validation (use similar to studio)
  - Update title
  - Update date
- Add validation to check if bundle name already exists

image


juice49:
This is looking great! I always wonder; is there any precedent for repurposing the form builder for these custom forms? It feels appealing to handle them by creating a schema that can encapsulate validation rules and custom input components declaratively. However, I'm not sure whether this is possible and/or sensible.

RitaDias:
> This is looking great! I always wonder; is there any precedent for repurposing the form builder for these custom forms? It feels appealing to handle them by creating a schema that can encapsulate validation rules and custom input components declaratively. However, I'm not sure whether this is possible and/or sensible.

ooh! I never heard / saw it done that way (nor do I know if it's possible!)
I simply followed this pattern because it's how we handle the inputs & validation on the actual inputs (and we didn't have design for it!)

I also know that the figma design was updated for this BUT it feels like it's missing some fields (such as publish date which I feel is very important so I'm not quite sure how to take it. Design wise overall, I left it as it was closer to the prototype for now 😬

juice49:
It looks like @pedrobonamin may have done something like this for tasks. There's [a schema for task documents here](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/addonWorkspace/taskSchema.tsx#L81), and [the `FormBuilder` is rendered here](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/TasksFormBuilder.tsx#L73). However, there does appear to be [some custom code to glue it together](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/useTasksFormBuilder.ts#L39).

Not a blocker, of course! Just a thought 🙂.

RitaDias:
> It looks like @pedrobonamin may have done something like this for tasks. There's [a schema for task documents here](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/addonWorkspace/taskSchema.tsx#L81), and [the `FormBuilder` is rendered here](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/TasksFormBuilder.tsx#L73). However, there does appear to be [some custom code to glue it together](https://github.com/sanity-io/sanity/blob/6261ca1ec37c3cfced7d43e4e7f37f9a2f79b2a8/packages/sanity/src/core/tasks/components/form/tasksFormBuilder/useTasksFormBuilder.ts#L39).
> 
> Not a blocker, of course! Just a thought 🙂.

My one thought is that it will depend on the actual design of the figma. If we want to go with the newer one then we'll need to redesign this form anyway (as well as the error handling) and can follow that approach (if we decide it's the best one!)

image

There's a task for this on linear for Marius to review (the form errors I mean, so it will come up there ) :)

juice49 (approved):

juice49:
> My one thought is that it will depend on the actual design of the figma.

That's a great point. Sorry, I didn't pay attention to your previous message about it 🤦‍♀️. Even if there is a new UI, part of me thinks it would still be great to handle it using the form builder and custom input components... if that's possible. Particularly to have the system handles things like validation for us, so we don't have to add our own implementation and tests for it.

jordanl17:
> > This is looking great! I always wonder; is there any precedent for repurposing the form builder for these custom forms? It feels appealing to handle them by creating a schema that can encapsulate validation rules and custom input components declaratively. However, I'm not sure whether this is possible and/or sensible.
> 
> ooh! I never heard / saw it done that way (nor do I know if it's possible!) I simply followed this pattern because it's how we handle the inputs & validation on the actual inputs (and we didn't have design for it!)
> 
> I also know that the figma design was updated for this BUT it feels like it's missing some fields (such as publish date which I feel is very important so I'm not quite sure how to take it. Design wise overall, I left it as it was closer to the prototype for now 😬

@RitaDias I don't believe right now we will be implementing the publish date. Essentially at this state it'll be removed.

pedrobonamin (approved):

PR #7150, opened 2024-07-12
fix(corel): refactor version history to fetch all release documents

pedrobonamin:
### Description

Refactors version history to fetch all documents in one request instead of one request per document.
Taking advantage of the history api endpoint which supports multiple documents at once, with a comma separated list.

This should improve the releases details screen experience, loading the history faster.

Pd: this will eventually be removed, that's why it's not taking account for the changes in the revision id of the documents to trigger a new fetch of the history. Not ideal, but given we will remove it, the time needed for that update is not worth right now. It can be easily updated later.


juice49 (approved):
Awesome optimisation!

PR #7152, opened 2024-07-14
fix(sanity): avoid fetching in a loop

juice49:
### Description

I know this code is temporary, but I noticed it seems to be fetching repeatedly the entire time it's mounted. Here's a tiny 🤏 bandaid to stop that.


RitaDias (approved):
I'm still thinking this is something we'll need to remove and move to the releases bundle. Not quite sure what it'll happen but this sounds good!

juice49:
Thanks @RitaDias. I fully agree 🙂.

PR #7154, opened 2024-07-15
fix(sanity): open correct groups and fieldsets on `setOpenPath`

rdunk:
### Description

For context, this PR is to resolve issues largely reported by Presentation users. I believe the same issue affects the Structure tool, but is far less visible as this is a less common scenario and largely non-critical in that context.

In summary: `setOpenPath` in the `DocumentPaneProvider` fails to focus the specified field if it exists in non-default groups or collapsed fieldsets.

This is because `getExpandOperations` is not returning the expected operations, which in turn seems to be because `immutableReconcile` drops the `_allMembers` property of the form state as it currently only handles enumerable properties (and `_allMembers` is explicitly set to non-enumerable).

There is also a small change to `getExpandOperations` as I believe the `setSelectedGroup` operation was returning the wrong group name.

### What to review

The simplest way to reproduce the issue in Structure is by clicking on the global presence avatar of a user who is currently focused on a field in a non-default group or collapsed fieldset.

This PR also adds a new schema type and an update to the Presentation test studio preview (under the `FieldGroups` tab) useful for checking the focus behaviour of non-default groups and collapsed fieldsets.

### Testing

One test added to check that `immutableReconcile` always returns the latest value of non-enumerable properties.

### Notes for release

### DIFF
--- a/dev/test-studio/preview/FieldGroups.tsx
+++ b/dev/test-studio/preview/FieldGroups.tsx
@@ -1,0 +1,62 @@
+import {Box, Card, Stack, Text} from '@sanity/ui'
+
+import {useQuery} from './loader'
+
+export function FieldGroups(): JSX.Element {
+  const {data, loading, error} = useQuery<
+    {
+      _id: string
+      field1: string | null
+      field2: string | null
+      nested: {
+        field3: string | null
+        field4: string | null
+        field5: string | null
+        nested: {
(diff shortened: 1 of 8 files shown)


bjoerge (approved):
Great! Thank you so much for this @rdunk, appreciate the added test as well <3

PR #7160, opened 2024-07-15
fix: detect duplicate context errors

stipsan:
### Description

When using monorepo setups it's very common to stumble upon errors such as:
- Workspace: missing context value
- Could not find `source` context
- Router: missing context value

What's effectively happening when these errors occur is that during bundling there are instances of `React.createContext` that should only happen once, that are now duplicated. There are many ways a repo can end up in this state, with monorepo setups like `pnpm workspaces` being especially suspectible.

Let's compare the simplified bundler output of a healthy build, and a broken one, to understand what's happening.

Let's say in your `sanity.config.ts` you're using the `useWorkspace` hook in a custom input component:
```tsx
import {useWorkspace} from 'sanity'

export function CustomInput() {
  const workspace = useWorkspace()

  // ...
}
```

#### Healthy bundle
```js
const WorkspaceContext = createContext(null)

function WorkspaceProvider({children, workspace}) {
  return <WorkspaceContext.Provider value={workspace}>{children}</WorkspaceContext.Provider>
}

function useWorkspace() {
  const workspace = useContext(WorkspaceContext)

  if (!workspace) throw new Error('Workspace: missing context value')

  return workspace
}

function CustomInput() {
  const workspace = useWorkspace()

  // ...
}
```

#### Broken bundle

Here's what the bundle looks like when we're in a failing state, where attempting to render `CustomInput`  throws  an `Workspace: missing context value` error

```js
const WorkspaceContext = createContext(null)

function WorkspaceProvider({children, workspace}) {
  return <WorkspaceContext.Provider value={workspace}>{children}</WorkspaceContext.Provider>
}

function useWorkspace() {
  const workspace = useContext(WorkspaceContext)

  if (!workspace) throw new Error('Workspace: missing context value')

  return workspace
}

// Instead of re-using the same hooks and context, hook deps are duplicated

const WorkspaceContext$0 = createContext(null)

function useWorkspace$0() {
  const workspace = useContext(WorkspaceContext$0)

  if (!workspace) throw new Error('Workspace: missing context value')

  return workspace
}

function CustomInput() {
  const workspace = useWorkspace$0()

  // ...
}
```

It's very difficult to resolve them, as their cause depends on userland package manager setup, workspace setup, symlinks and more.
This PR mitigates this problem by creating key react contexts in a global, versioned, scope, allowing us to:
a) if the context were come from different `sanity` versions we throw an error.
b) if duplicate contexts come from the same `sanity` versions we warn, and bridge them so it'll work in most cases.

#### Handling incompatible duplicates

When the globally scoped contexts detect duplicates on different versions [we throw an error](https://sanity-io-studio-kzvkc131s.sanity.build/):

image

On next.js `next dev` ([here's the same error in production](https://sanity-i51vbv6h2.sanity.build/admin)):

image

#### Gracefully handling compatible duplicates

The way context is bridged when there are duplicates are based on the way `@sanity/ui` solves it: https://github.com/sanity-io/ui/blob/bad8b33cbfc801de7545b02e283c89fc38482d89/src/core/lib/createGlobalScopedContext.ts
We do warn when this happens though, as it's usually an indicator that the bundler might be doing other things it shouldn't be doing, in a way it's a bit of "undefined behaviour":

image

### What to review

Does the error message, and console warnings make sense?
Should we link to a docs page for how to resolve duplicates (for example `pnpm dedupe sanity` is usually sufficient for `pnpm` workspaces)?
The contexts I added to the globally scoped contexts are based on [publicly documented hooks](https://www.sanity.io/docs/studio-react-hooks) and [on the hooks used by `@sanity/presentation`](https://github.com/sanity-io/visual-editing/blob/e8809f8131988af7888ea7d8e10524e55df52e9e/packages/presentation/src/internals.ts). Did I miss any?

### Testing

I made test releases that can be used to see the error messages:
- `sanity@3.50.1-detect-context-errors.12`
- `sanity@3.50.1-detect-context-errors.10`

### Notes for release

- improved error messages when duplicate instances of sanity in node_modules

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -225,7 +225,11 @@ const config = {
     // Prefer createContext in _singletons
     {
       files: ['packages/sanity/src/**'],
-      excludedFiles: ['**/__workshop__/**', 'packages/sanity/src/_singletons/**'],
+      excludedFiles: [
+        '**/__workshop__/**',
+        'packages/sanity/src/_singletons/**',
+        'packages/sanity/src/_createContext/**',
+      ],
       rules: {
         'no-restricted-imports': [
           'error',
--- a/dev/test-next-studio/next.config.mjs
+++ b/dev/test-next-studio/next.config.mjs
@@ -62,6 +62,9 @@ const config = {
       '@sanity/vision': requireResolve('../../packages/@sanity/vision/src/index.ts'),
       'sanity/_internal': requireResolve('../../packages/sanity/src/_exports/_internal.ts'),
       'sanity/_singletons': requireResolve('../../packages/sanity/src/_exports/_singletons.ts'),
+      'sanity/_createContext': requireResolve(
+        '../../packages/sanity/src/_exports/_createContext.ts',
+      ),
       'sanity/cli': requireResolve('../../packages/sanity/src/_exports/cli.ts'),
       'sanity/desk': requireResolve('../../packages/sanity/src/_exports/desk.ts'),
       'sanity/presentation': requireResolve('../../packages/sanity/src/_exports/presentation.ts'),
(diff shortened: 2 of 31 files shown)


binoy14:
Better if Rico reviews this

ricokahler (commented):
This looks really solid and I love how it allows the same version of the context to be used in the event some package manager installs sanity twice. We'd like to add a link to some documentation for why this error is occurring (and what to do about it) but besides that, it's 👌

cngonzalez (approved):

cngonzalez:
Approving and merging based on @ricokahler 's review!

PR #7166, opened 2024-07-16
fix(sanity): ensure Actions API client uses adequate API version

juice49:
### Description

We observed that some Studios use an inadequate API version when calling the Actions API. It may be the case that developers are inadvertently mutating the client configuration object, however, we aren't certain. This change attempts to resolve the issue by explicitly setting all Actions API requests to use the minimally adequate API version.

### What to review

When performing actions on a document while the Actions API integration is enabled, the API requests produced should be exactly the same as before, but always use the minimally adequate API version (`2024-05-23`).

### Testing

I have not added tests, because we have not been able to replicate a scenario in which the API version used is anything other than the one required. For this change, I believe the existing tests are sufficient.


cngonzalez (approved):
what a strange thing to happen! the changes make sense to me, however.

PR #7171, opened 2024-07-16
chore(corel): rename bundle.name to bundle.slug

pedrobonamin:
### Description

This PR replaces the `bundle.name` for `bundle.slug`, making it easier to identify the importance of that property.
Also, updates the `BundleDocument` definition to instead of extending `SanityDocument` it has it's own defined fields, given `SanityDocument` has a `  [key: string]: unknown` property which produced some code errors that were not picked by typescript.

A migration script was added and attached to a placeholder ui that can be accessed by visiting `/test/releases/migrate` that is intended to be removed in a following PR after the migration is successfully executed and every developer has updated their current branch, or things will crash for the rest of the team.

### Testing

Current tests are updated to reflect this changes.


packages/sanity/src/core/releases/tool/Migrate.tsx
```diff
@@ -0,0 +1,107 @@
...
+import {Flex, Stack, Text} from '@sanity/ui'
+import {useState} from 'react'
+
+import {Button} from '../../../ui-components'
+import {useAddonDataset} from '../../studio'
+
+/**
+ * temporary migration script and buttons to trigger it, to add slug to bundles and to remove the bundles names once merged
+ */
+export function MigrateBundlesSlug() {
```
juice49, line 11: Love that you've included a migration! Btw, [migrations using the CLI](https://www.sanity.io/docs/cli-migration) are also a pretty nice way to handle this kind of thing (if you don't already know about them).

packages/sanity/src/core/releases/tool/detail/ReleaseDetail.tsx
```diff
@@ -18,10 +17,10 @@ import {ReleaseReview} from './ReleaseReview'
 const SUPPORTED_SCREENS = ['overview', 'review'] as const
 type Screen = (typeof SUPPORTED_SCREENS)[number]
 
-const useFetchBundleDocuments = (bundleName: string) => {
+const useFetchBundleDocuments = (bundleSlug: string) => {
   const client = useClient({apiVersion: API_VERSION})
-  const query = `*[defined(_version) &&  _id in path("${bundleName}.*")]`
-  return useListener<BundleDocument>({query, client})
+  const query = `*[defined(_version) &&  _id in path("${bundleSlug}.*")]`
+  return useListener({query, client})
```
juice49, line 23: Just wondering: is it intentional to remove the listener type argument (`BundleDocument`) altogether?

packages/sanity/src/core/store/bundles/types.ts
```diff
@@ -10,17 +9,23 @@ import {type bundlesReducerAction, type bundlesReducerState} from './reducer'
 /**
  * @internal
  */
-export interface BundleDocument extends SanityDocument {
+export interface BundleDocument {
```
juice49, line 12: Instead of duplicating the fields like `_id` from `SanityDocument`, it might be better to `Pick` them.
```suggestion
export interface BundleDocument extends Pick<SanityDocument, '_id' | '_createdAt' | '_updatedAt' | '_rev' | '_version'> {
```

Alternatively, we could install type-fest or something similar to get [a helper that removes the index type for us](https://github.com/sindresorhus/type-fest/blob/main/source/omit-index-signature.d.ts):

```ts
import type {OmitIndexSignature} from 'type-fest'

export interface BundleDocument extends OmitIndexSignature<SanityDocument> {
```

(This is totally non-blocking, just a thought 🙂).

juice49 (approved):
Looks great! Thank you for doing this 🙏.

packages/sanity/src/core/releases/tool/detail/ReleaseDetail.tsx
pedrobonamin: It was a mistake to include this as `BundleDocument`, it is a version document, so it's any type of documents. 
Given it was extending `SanityDocument` ts was not warning us

RitaDias (approved):
🚀

packages/sanity/src/core/releases/tool/Migrate.tsx
pedrobonamin: Yeah, they are great, I have used them before, but not sure how to use them inside the monorepo 😓 that's why I went for this. Also, it gives us the option to trigger it immediately once everyone has updated their branch

juice49: No problem! I think you can create and execute them inside `dev/test-studio` and they should just work 🙂.

packages/sanity/src/core/releases/tool/detail/ReleaseDetail.tsx
juice49: Ah of course! These are version documents, not bundle metadata documents. My bad!

PR #7174, opened 2024-07-16
feat: add disableArrayWarning option to SlugOptions

Stelkooo:
### Description

Adds `disableArrayWarning` to the `SlugOptions` interface. Purpose is to allow users to disable the automatic array warning for slug fields within arrays.

Below is warning seen in the console.
image

Following the steps suggested by the warning works and stops the warning from showing up. However, if the developer creating the schema is using TypeScript they are hit with the following error
image

So adding `disableArrayWarning` removes that TS error.

Also a change to the `schemaOptions` `const` within the `slugValidator.ts` file as its type assertion is not needed due to `disableArrayWarning` is now included within the `SlugOptions` interface.

### What to review

Review the changes to the `SlugOptions` interface found in the `slug.ts` file and to the `defaultIsUnique` function, specifically the `schemaOptions` variable`. Ensure that when enabling `disableArrayWarning` no TS error appears.

### Testing

I did not add testing, I honestly do not know how but checking out the other slug functions and options I believe there is none. I have tested it with my own repos where I have used a slug field within an array.

### Notes for release

This PR adds the `disableArrayWarning` option to the `SlugOptions` interface removing the TS error that appears when `disableArrayWarning` does not exist within the interface.

### DIFF
--- a/packages/@sanity/types/src/schema/definition/type/slug.ts
+++ b/packages/@sanity/types/src/schema/definition/type/slug.ts
@@ -25,6 +25,7 @@ export interface SlugOptions extends SearchConfiguration, BaseSchemaTypeOptions 
   maxLength?: number
   slugify?: SlugifierFn
   isUnique?: SlugIsUniqueValidator
+  disableArrayWarning?: boolean
 }
 
 /** @public */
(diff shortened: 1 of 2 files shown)


bjoerge:
This should be fixed by a type cast in the consuming end.

bjoerge:
Ah, sorry - I re-read and realized that this is indeed on us to fix.

bjoerge (approved):

Stelkooo:
No problem, thanks for having a look at it

PR #7175, opened 2024-07-17
refactor: update document badge logic and navigation

RitaDias:
### Description

- Update the badge in the document to show the badge only when the current version matches the document version
- Add navigation to the release overview in the puglin

Out of scope:
- Fetching the versions of documents with the updated query, that will be in a separate PR


packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
```diff
@@ -7,24 +7,29 @@ import {
...
 } from 'sanity'
+import {useRouter} from 'sanity/router'
+import {styled} from 'styled-components'
 
-// FIXME
-// eslint-disable-next-line boundaries/element-types
+import {getVersionName} from '../../../../../../core/bundles'
 import {useBundles} from '../../../../../../core/store/bundles/useBundles'
 
+const BadgeButton = styled(Button)({
```
juice49, line 19: Just wondering; do we have a good reason for overriding the behaviour of the UI library here? I have no strong opinion about the right choice, but I do think things like this should probably be consistent across the UI.

juice49 (approved):
Looks great!

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
RitaDias: tbh I did it because that's how Marius had it in his prototype and while doing it (before adding this) without this cursor affordance it's really hard to tell that this is a clickable element. 

I'm of the opinion that we should likely have this everywhere on the studio but last time the cursor discourse was brought up nothing was really done about it 😄

juice49: > without this cursor affordance it's really hard to tell that this is a clickable element.

I'm always on the fence about it, but usually land on the side of respecting the way a user's entire operating system handes buttons. But it's an age old debate and totally not worth bringing up again 🤣.

Let's discuss it no further!

RitaDias: I'll add it to my notes to bring up to Marius :) (I have a linear story for it)

PR #7176, opened 2024-07-17
feat(preview): add `unstable_observeDocument(s)` to preview store

bjoerge:
### Description

This adds two new methods to the preview store:
```ts
unstable_observeDocument(documentId: string): Observable<SanityDocument>
unstable_observeDocuments(documentIds: string[]): Observable<SanityDocument[]>
```
These will emit _full_ documents as opposed to the other methods in the preview store that takes in a list of path selectors and returns _partial_ documents with only the given paths selected.

The reason it's part of the preview store is so that we can re-use the global listener connection we currently use for previews - philosophically it makes sense to keep them here as well, since they update with a higher latency than documents that has realtime editing capabilities, and is therefore better suited for "previewing" purposes.

Since we do have the full document in these cases, we can apply the patches  as they arrive from the backend, and doesn't need to do the ordinary refetching that we do for regular previews where only partial documents are kept locally.

These methods are useful for cases where you don't need the optimistic patching logic provided by the document store, but still require the full document to be loaded. The use case at hand is to load a list of documents and show their validation status in a view that doesn't support editing, but still needs to receive realtime updates in case they are being edited by other collaborators.

These are only meant for internal use (for now)

### What to review

- Does the APIs make sense?
- I've done some smaller refactorings and code cleanup as part of this
- Also added the `excludeMutations` flag to the listener request to save bandwith. This will be supported by the client once https://github.com/sanity-io/client/pull/872 lands

### Testing
- Will fast track this into next later on, will look into getting some additional tests in place then.

### Notes for release
n/a - internal


pedrobonamin (approved):
This is great, will help us solve lots of use cases, thanks for this @bjoerge !

PR #7180, opened 2024-07-17
feat(sanity): supporting editing bundle metadata

jordanl17:
### Description
Completes the user case to support editing of bundle metadata

https://www.loom.com/share/e80cf5964e2b47b9a49eb76fb7555b3d?sid=271de213-e2aa-4c79-bf34-d5846583344b

### What to review
* `ReleaseMenuButton` has a new Menu item added to Edit - this renders the `BundleDetailsDialog` passing the editing bundle
* `BundleDetailsDialog` initialises its value using the provided bundle (or defaults to the default values)
* `BundlesDetailsDialog` renders the `BundleForm` - `BundleForm` will only validate the slug for creating bundles, as slug will not change when editing
* `BundleDetailsDialog` conditionally called `createBundle` or `updateBundle` depending on the action taken

### Testing
* Updated both `BundleForm` and `BundleDetailsDialog` to test the existing create, but now the new edit cases also


packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
jordanl17, line 1: This is a rename from `CreateBundleDialog` with some changes as noted in comments below

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+import {type BundleDocument} from '../../../store/bundles/types'
+import {useBundleOperations} from '../../../store/bundles/useBundleOperations'
+import {usePerspective} from '../../hooks/usePerspective'
+import {BundleForm} from './BundleForm'
+
+interface BundleDetailsDialogProps {
+  onCancel: () => void
+  onSubmit: () => void
+  bundle?: BundleDocument
```
jordanl17, line 13: Passing bundle through to support editing bundle data

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Element {
+  const {onCancel, onSubmit, bundle} = props
+  const {createBundle, updateBundle} = useBundleOperations()
+  const [hasErrors, setHasErrors] = useState(false)
+
+  const [value, setValue] = useState<Partial<BundleDocument>>(() => {
```
jordanl17, line 21: using a function to set default state depending on whether existing bundle value already exists

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+      icon: 'cube',
+      //publishAt: undefined,
+    }
+  })
+  const [isSubmitting, setIsSubmitting] = useState(false)
+
+  // TODO MAKE SURE THIS IS HOW WE WANT TO DO THIS
+  const {setPerspective} = usePerspective()
+
+  const bundleOperation = useCallback(
```
jordanl17, line 45: Encapsulating the logic required to trigger the create or update functions out here

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const handleOnChange = useCallback((changedValue: Partial<BundleDocument>) => {
+    setValue(changedValue)
+  }, [])
+
+  const handleOnError = useCallback((errorsExist: boolean) => {
+    setHasErrors(errorsExist)
+  }, [])
+
+  const dialogTitle = bundle ? 'Edit release' : 'Create release'
```
jordanl17, line 88: Dynamic title and CTA label based on action being taken

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onCancel: () => void
+  onSubmit: () => void
+  bundle?: BundleDocument
+}
+
+export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Element {
+  const {onCancel, onSubmit, bundle} = props
+  const {createBundle, updateBundle} = useBundleOperations()
+  const [hasErrors, setHasErrors] = useState(false)
+  const formAction = bundle ? 'edit' : 'create'
```
pedrobonamin, line 20: 👍🏻 I like this approach

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+  })
+  const [isSubmitting, setIsSubmitting] = useState(false)
+
+  // TODO MAKE SURE THIS IS HOW WE WANT TO DO THIS
+  const {setPerspective} = usePerspective()
+
+  const bundleOperation = useCallback(
+    (formValue: Partial<BundleDocument>) => {
+      if (bundle) {
```
pedrobonamin, line 48: Can we reuse here the `formAction` defined above?

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const updatedBundle: BundleDocument = {
+          ...bundle,
```
pedrobonamin, line 50: I think.. let's confirm it, that you don't need to send the bundle document, given the way the set action works it will only set the keys provided, so in theory, we could provide only the form value

packages/sanity/src/core/bundles/components/dialog/BundleForm.tsx
```diff
@@ -24,6 +24,8 @@ export function BundleForm(props: {
 }): JSX.Element {
   const {onChange, onError, value} = props
   const {title, description, icon, hue /*, publishAt*/} = value
+  const {current: action} = useRef(value.slug ? 'edit' : 'create')
```
pedrobonamin, line 27: This feels a bit hard to follow, can we pass down the action as a prop instead? 
It feels like it could be mutated from inside this component without really wanting that

pedrobonamin (approved):
Thanks for this @jordanl17 , it looks great. Added some comments, non blocking.

packages/sanity/src/core/bundles/components/dialog/BundleForm.tsx
jordanl17: So that's how I had it - passing `formAction` in from the `BundleDetailsDialog`. However I then thought about scenarios where an initial value was not set, but the action was wrongly set to `edit`, that's why I made a last commit change to infer the action from the initial value passed in. I can add a comment to clarify what is happening and why

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
jordanl17: I tested this and, yep, only need to send across the edited fields. So I've updated this now

PR #7186, opened 2024-07-17
Update sanity UI use click outside

stipsan:
### Description

Fixes the type errors failing the build for #7185.
In v2.8.0 of `@sanity/ui` the implementation of `useClickOutside` were updated to allow using react refs, and to no longer require `useCallback` wrapping on the handler. See https://github.com/sanity-io/ui/pull/1390 for more details.
As a result of this it seems TypeScript is no longer able to infer the type of the `event` argument, when the handler is wrapped in `useCallback`.
Inference works just fine and the type of `event` winds up being `MouseEvent` when the `useCallback` wrapping is removed, and the handler is passed to `useClickOutside` directly.
Since `useCallback` isn't needed anymore in any case it seemed like the best way to resolve the typescript errors.

There are other improvements to `useClickOutside` but they'll be implemented in separate follow-up PRs 🙌 

### What to review

Does the change make sense?

### Testing

Existing builds and tests should be sufficient. `@sanity/ui` already ran the studio E2E testing suite before releasing v2.8.0: https://github.com/sanity-io/ui/actions/runs/9978840730/job/27576866642

### Notes for release

N/A

Closes #7185

### DIFF
--- a/examples/ecommerce-studio/package.json
+++ b/examples/ecommerce-studio/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@sanity/cli": "3.51.0",
-    "@sanity/ui": "^2.7.0",
+    "@sanity/ui": "^2.8.0",
     "react": "^18.3.1",
     "react-barcode": "^1.4.1",
     "react-dom": "^18.3.1",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "@sanity/prettier-config": "^1.0.2",
     "@sanity/test": "0.0.1-alpha.1",
     "@sanity/tsdoc": "1.0.83",
-    "@sanity/ui": "^2.7.0",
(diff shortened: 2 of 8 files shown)


binoy14 (changes requested):
Seems like the click outside functionality is broken with this update.

1. Go to array inputs [here](https://test-studio-git-update-sanity-ui-use-click-outside.sanity.build/test/structure/input-standard;arraysTest;2846bb8c-4fc4-4e64-8093-5942690f1af5)
2. Click on Add Item under "Array of references to authors"
3. Click outside the array box

Expected:
Removes the empty reference

Actual:
Keeps the empty reference

This is not the case in `next` preview build [here](https://test-studio.sanity.build/test/structure/input-standard;arraysTest;2846bb8c-4fc4-4e64-8093-5942690f1af5)

PR #7189, opened 2024-07-18
refactor: replace `useOnClickOutside` with `useClickOutsideEvent`

stipsan:
### Description

This PR removes the two custom hooks `useOnClickOutside`, replacing them with the new `useClickOutsideEvent` in `@sanity/ui`.
According to the hook comments, the reason why these hooks were introduced, instead of using the `useClickOutside` hook from `@sanity/ui`, where that `useClickOutside` [don't support passing elements that come from mutable react refs.](https://github.com/sanity-io/sanity/blob/8a9730c0f1b4356414536e76c1b0ef0d66892f7d/packages/sanity/src/core/form/hooks/useOnClickOutside.ts#L3-L6).
The new `useClickOutsideEvent` hook doesn't have this issue 🎉 

---

<details>

<summary>
Storytime for why the new <code>menuButtonRef</code> reference were added
</summary>

After migrating the hooks I followed @binoy14's [instructions](https://github.com/sanity-io/sanity/pull/7186#pullrequestreview-2183886727), from a similar PR, to test manually.

A regression showed up, clicking on the "Show more" button cleared the item instead of opening the context menu:

https://github.com/user-attachments/assets/28e7656a-0bf8-449e-9a30-08e37ff179bd

Here's how it behaves on the `next` branch:

https://github.com/user-attachments/assets/dec1f024-a5ab-4793-8ce5-c97086b4368e

After going deep down the rabbit hole it turns out the reason it works on `next` is because `useOnClickOutside` is because of a bug that accidentally make it work (kinda!):
1. The `mousedown` handler filters out events where `event.target` [isn't an instance of `HTMLElement`](https://github.com/sanity-io/sanity/blob/8a9730c0f1b4356414536e76c1b0ef0d66892f7d/packages/sanity/src/core/form/hooks/useOnClickOutside.ts#L14C29-L14C40).
2. When clicking on `Show more` it's 99.99% of the time with `event.target` set to the SVG in the `<MenuButton />`, which isn't an `HTMLElement`, it's an `SVGElement`.
3. Which is why `useOnClickOutside` ignores the click on "Show more" and won't fire the callback that would lead to the input clearing.

In the new `useClickOutsideEvent` [hook it checks for `event.target instanceof Node` instead of `HTMLElement`](https://github.com/sanity-io/ui/blob/b830db1fd83fa58b8941b3bb3652fa4eca06ee4b/src/core/hooks/useClickOutsideEvent.ts#L34), both `HTMLElement` and `SVGElement` extends `Node`, leading to it triggering the click outside event.

This is further confirmed by clicking on the 1px margin around the svg, inside the `<MenuButton />` (this video can be reproduced on the `next` branch right now), as that triggers the same failure in `useOnClickOutside`:

https://github.com/user-attachments/assets/683161bb-7db6-4336-a748-3e9ee9036501

Not having a `menuButtonRef` for the `Show more` button is also clicking outside after opening the context menu isn't working on `next`:

https://github.com/user-attachments/assets/ed126863-86e2-4fa8-af4e-b509f29a2832

By adding `menuButtonRef`, and adding the `document.activeElement` check in the click handler, we're able to make sure the `Show more` menu button behaves the same way as `Create new`:

https://github.com/user-attachments/assets/06c6a2ae-eb8a-438a-9130-51adc783f50d

</details>

---

### What to review

I kept around the `useOnClickOutside.test.tsx` test to verify that the new implementation doesn't have regressions. But maybe [we should delete it since the existing test suite in `@sanity/ui` is far more comprehensive?](https://github.com/sanity-io/ui/blob/main/src/core/hooks/useClickOutsideEvent.test.tsx)
Are the comments in `ReferenceInput` that explains what the `useClickOutsideEvent` handler is doing making sense?
I noticed the `useOnClickOutside` implementation used `touchstart` in addition to `mousedown`. The new hook only uses `mousedown` as `touchstart` is generally frowned upon as it runs too early on mobile devices and interferes with scrolling actions, and leads to "click outside" events firing when the user intended was starting a scroll gesture, and not a tap gesture that the user would expect leads to the reference input clearing.
I've tested on my own iPhone on this branch to make sure it works like before, but maybe someone on Android should verify?

### Testing

Existing E2E tests might not be enough, and I've done manual testing myself for reference inputs, be it inside arrays or not. As well as cross dataset reference inputs.
I didn't manage to successfully test cross dataset reference inputs that were inside array items though, I'm not sure if this is a feature we support? I couldn't get it working on the `next` branch either when I tried to establish a baseline for testing on this branch.

### Notes for release

- N/A the changes in perf and behavior are too minor to warrant a callout in the changelog, it's more like a refactor.


packages/sanity/src/core/form/inputs/CrossDatasetReferenceInput/CrossDatasetReferenceInput.tsx
```diff
@@ -268,15 +267,10 @@ export function CrossDatasetReferenceInput(props: CrossDatasetReferenceInputProp
   // --- click outside handling
   const clickOutsideBoundaryRef = useRef<HTMLDivElement | null>(null)
   const autocompletePortalRef = useRef<HTMLDivElement | null>(null)
-  const createButtonMenuPortalRef = useRef<HTMLDivElement | null>(null)
```
stipsan, line 271: I'm removing this reference as it isn't set anywhere

package.json
```diff
@@ -17,7 +17,7 @@
     "check:deps": "pnpm --recursive --parallel exec depcheck",
     "check:format": "prettier . --check",
     "check:lint": "turbo run lint --continue -- --quiet",
-    "check:react-compiler": "eslint --no-inline-config --no-eslintrc --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --rule 'react-compiler/react-compiler: [warn]' --ignore-path .eslintignore.react-compiler --max-warnings 31 .",
+    "check:react-compiler": "eslint --no-inline-config --no-eslintrc --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --rule 'react-compiler/react-compiler: [warn]' --ignore-path .eslintignore.react-compiler --max-warnings 29 .",
```
stipsan, line 20: The count didn't actually change in this PR, it's set to the same count as reported on the `next` branch

binoy14:
Added Carolina, if she has bandwidth before I can get to this

ricokahler (dismissed):
Looks really solid and amazing finds. Thank you for all the detailed descriptions (I did read the storytime!).
I'm not that opinionated on the test for useClickOutside so if you feel it's best to delete it (since sanity ui has a better suite), I support that!!

Issue #6807, opened 2024-05-30
Image metadata: No `Make` and `Model` properties in exif


xuzuodong:
**Describe the bug**

When [`exif`](https://www.sanity.io/docs/image-metadata#3e05db6e3c80) is enabled, `Make` and `Model` information of the image should be included. 

**To Reproduce**

In a sanity studio project, define a schema type with an image field, and add `exif` in its `metadata` array.

```js
 {
      name: 'metaImage',
      title: 'Image with metadata',
      type: 'image',
      options: {
        metadata: [
          'exif', 
        ],
      },
},
```

**Expected behavior**

After upload an image in studio, with all other exif information, `Make` and `Model`, should also be saved to `metadata.exif`. For example, `Make` can be `Canon` and `Model` can be `Canon EOS D60`, as the test case in [sanity-io/exif-reader](https://github.com/sanity-io/exif-reader/blob/3da882ce01837771f0125f9e478bf24e68a29447/test/index.js#L15)

**Screenshots**
<img width="739" alt="image" src="https://github.com/sanity-io/sanity/assets/51878637/13381137-7026-4fca-bdf6-4c3048539509">

juice49:
Hi @xuzuodong. Thank you for reporting this!

We do extract the `Make` and `Model` data, but you need to enable its storage in your schema. You can do this by adding `'image'` to the field's `options.metadata` array:

```ts
{
  name: 'metaImage',
  title: 'Image with metadata',
  type: 'image',
  options: {
    metadata: [
      'exif',
      'image',
    ],
  },
},
```

After enabling this, you should find that all future image uploads include a `metadata.image` object field. `Make` and `Model` fields will be stored here (as long as they exist for the source image). Please let us know if you run into any issues after enabling the option.

It looks like this option is not currently documented, but it should be!

PR #7190, opened 2024-07-18
chore: add `image` literal to `ImageMetadataType`

xuzuodong:
### Description

According to https://github.com/sanity-io/sanity/issues/6807, passing `image` to `ImageMetadataType` should be allowed. This PR updates `ImageMetadataType` to support this.


cngonzalez (approved):
Thank you so much for helping with this!

PR #7194, opened 2024-07-18
feat(releases): publish all documents in a bundle

jordanl17:
### Description
Clicking 'Publish all' on a release with all valid documents will not publish the version documents, and update the bundle metadata with the publish date

https://www.loom.com/share/94911ea76b484457aca3ffa31ceca78f?sid=1851e5b8-fb86-4052-ae3f-0ff9958b97e8

### What to review
* Presence added to `ReleaseDocumentPreview`
* `handlePublishBundle` added to `useBundleOperations`
* validation of documents used to control disabled state of publish all in `ReleaseDetail`

### Testing
* Tests in `ReleaseDetail.test` updated for new publish all calls

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/components/ReleaseDocumentPreview.tsx
+++ b/packages/sanity/src/core/releases/components/ReleaseDocumentPreview.tsx
@@ -47,7 +47,7 @@ export function ReleaseDocumentPreview({
     [documentId, documentTypeName, releaseSlug],
   )
 
-  const presencePreview = useMemo(
+  const previewPresence = useMemo(
     () => documentPresence?.length > 0 && <DocumentPreviewPresence presence={documentPresence} />,
     [documentPresence],
   )
@@ -59,14 +59,14 @@ export function ReleaseDocumentPreview({
       radius={2}
       data-as="a"
     >
-      <SanityDefaultPreview {...previewValues} status={presencePreview} isPlaceholder={isLoading}>
+      <SanityDefaultPreview {...previewValues} status={previewPresence} isPlaceholder={isLoading}>
         {hasValidationError && (
           <Tooltip
             portal
             content={
               <Text muted size={1}>
                 {/* TODO: clarify copy */}
-                There are validation issues with this document
+                There are validation errors in this document
               </Text>
             }
           >
--- a/packages/sanity/src/core/releases/components/ReleaseDocumentPreview.tsx
+++ b/packages/sanity/src/core/releases/components/ReleaseDocumentPreview.tsx
@@ -1,6 +1,8 @@
+import {ErrorOutlineIcon} from '@sanity/icons'
 import {type PreviewValue} from '@sanity/types'
-import {Card} from '@sanity/ui'
+import {Card, Text, Tooltip} from '@sanity/ui'
 import {type ForwardedRef, forwardRef, useMemo} from 'react'
+import {DocumentPreviewPresence, useDocumentPresence} from 'sanity'
 import {IntentLink} from 'sanity/router'
 
 import {SanityDefaultPreview} from '../../preview/components/SanityDefaultPreview'
@@ -12,6 +14,7 @@ interface ReleaseDocumentPreviewProps {
   releaseSlug: string
   previewValues: PreviewValue
   isLoading: boolean
+  hasValidationError?: boolean
 }
 
 export function ReleaseDocumentPreview({
@@ -20,7 +23,10 @@ export function ReleaseDocumentPreview({
   releaseSlug,
   previewValues,
   isLoading,
+  hasValidationError,
 }: ReleaseDocumentPreviewProps) {
+  const documentPresence = useDocumentPresence(documentId)
+
   const LinkComponent = useMemo(
     () =>
       // eslint-disable-next-line @typescript-eslint/no-shadow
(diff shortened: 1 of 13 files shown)


packages/sanity/src/core/releases/components/ReleaseDocumentPreview.tsx
```diff
@@ -41,9 +47,33 @@ export function ReleaseDocumentPreview({
...
-    <Card as={LinkComponent} radius={2} data-as="a">
-      <SanityDefaultPreview {...previewValues} isPlaceholder={isLoading} />
+    <Card
+      tone={hasValidationError ? 'critical' : 'default'}
+      as={LinkComponent}
+      radius={2}
+      data-as="a"
+    >
+      <SanityDefaultPreview {...previewValues} status={previewPresence} isPlaceholder={isLoading}>
+        {hasValidationError && (
```
jordanl17, line 63: Currently just passing the boolean rather than the individual validation errors. Once designs show how we'd want invalid documents to look, this can be udpated

packages/sanity/src/core/store/bundles/useBundleOperations.ts
```diff
@@ -59,19 +60,42 @@ export function useBundleOperations() {
...
 
       return clientOperation.commit()
     },
-    [client],
+    [addOnClient],
+  )
+
+  const handlePublishBundle = useCallback(
+    async (bundleId: string, bundleDocuments: SanityDocument[]) => {
+      if (!addOnClient) return null
```
pedrobonamin, line 75: I know we are handling this in the same case in the rest of operations, but I'm thinking if maybe instead we should throw an error in the case that addOnClient is undefined.
Could also be a follow up.

packages/sanity/src/core/releases/tool/detail/ReleaseDetail.tsx
```diff
@@ -66,6 +81,96 @@ export const ReleaseDetail = () => {
...
+
+  // review screen will not be available once published
+  // so redirect to summary screen
+  useEffect(() => {
+    if (activeScreen === 'review' && bundle?.publishedAt) {
+      navigateToSummary()
+    }
+  }, [activeScreen, bundle?.publishedAt, navigateToSummary])
+
+  const handleConfirmPublishAll = useCallback(async () => {
```
pedrobonamin, line 118: suggestion: could we move this, and the specific logic related to the publish to a new file? And then import in this file something like:
`import {ReleasePublishAction} from "./ReleasePublishAction"` 

I think it will make it easier for our future selfs.

jordanl17: Sure thing

- [ ] Extract publish all logic into separate component

packages/sanity/src/core/store/bundles/useBundleOperations.ts
```diff
@@ -59,19 +60,42 @@ export function useBundleOperations() {
...
+
+      const transaction = studioClient.transaction()
+      bundleDocuments.forEach((document) => {
+        const transactionDocument = omit(document, ['_version']) as SanityDocument
+
+        // update the published document with the bundle version
+        transaction.createOrReplace({
+          ...transactionDocument,
+          _id: getPublishedId(document._id, true),
+        })
```
pedrobonamin, line 85: Tested using the actions api and versions are not supported yet. 
But I wonder if we could add here a check for the published revision id like in: https://github.com/sanity-io/sanity/blob/55ceee3298103ae473c99f06780ec9bdbe932bd3/packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts#L47-L53

packages/sanity/src/core/store/bundles/useBundleOperations.ts
jordanl17: @pedrobonamin yeh good idea - added a throw to all the methods here

packages/sanity/src/core/store/bundles/useBundleOperations.ts
jordanl17: This link isn't working for me @pedrobonamin

pedrobonamin: https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/core/store/_legacy/document/document-pair/operations/publish.ts#L47-L53

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseDetail.test.tsx
```diff
@@ -256,15 +262,19 @@ describe('after bundles have loaded', () => {
...
+          [
+            {
+              _createdAt: currentDate,
+              _id: 'test-id',
+              _rev: 'abc',
+              _type: 'document',
+              _updatedAt: currentDate,
+            },
+          ],
+          {'123': 'mock revision id'},
```
pedrobonamin, line 276: NIT, but would it make sense to have the ids of this tests to be closer to what we are gonna be receiving?

e.g. 

```
.toHaveBeenCalledWith(
          'summer-drop',
          [
            {
              _createdAt: currentDate,
              _id: 'summer-drop.123',
              _rev: 'abc',
              _type: 'document',
              _updatedAt: currentDate,
            },
          ],
          {'123': 'mock revision id'},
```

packages/sanity/src/core/store/bundles/useBundleOperations.ts
```diff
@@ -59,19 +64,78 @@ export function useBundleOperations() {
...
+      })
+
+      await transaction.commit()
+      const publishedAt = new Date().toISOString()
+      return await addOnClient.patch(bundleId).set({publishedAt, archivedAt: publishedAt}).commit()
+    },
+    [addOnClient, studioClient],
   )
 
+  const guardForAddOnClient = <Callback extends (...args: any[]) => any>(
```
pedrobonamin, line 123: Does this needs to be an useCallback?

jordanl17: Yeh, good callout! I've changed the implementation here since trying to memoise the return from the HOF was proving a little unreadable and clunky

pedrobonamin (approved):
This looks good to me, thanks @jordanl17 !

PR #7198, opened 2024-07-18
chore: enable production react profiling in test studios

stipsan:
### Description

This PR enables using the React Profiler on the production build of the studio.

In other words, instead of seeing this message on preview deployments [(like for this branch)](https://test-studio-git-react-production-profiling.sanity.build/test/structure):

image

You'll be able to run the profiler:

image

The main benefit of running the profiler in production, is that the perf overhead in React from running Strict Mode, as well as all the checks and instrumentation that is only used in local development (Hot Module Reload, checking if `key` is spread to attributes, checking if you're trying to set dom attributes that don't exist etc), all of that no longer interferes with the profiling results.

Additionally, since the regular components debugger is often used in conjunction with the profiler, I've made sure to disable esbuild's `minifyIdentifiers` setting and enabled source maps.

Without it the debugger becomes difficult to use:

image

Since they are enabled it's now almost as productive as when working locally in `sanity dev`:

image

React production profiling is controlled by the new `REACT_PRODUCTION_PROFILING` env var, so to run the test studio this way locally you can:

```
cd dev/test-studio && REACT_PRODUCTION_PROFILING=true pnpm turbo run build && pnpm start
```

On our Vercel test studio projects the `REACT_PRODUCTION_PROFILING` is set for only preview deployments, not for production.
In other words merging this PR won't enable it for the production https://test-studio.sanity.build deployments.
**This is intentional.** There is still some overhead to enabling profiling in production.
Thus there's value in omitting it from the production test studios to ensure that the performance we see there matches userland as much as possible.
It also makes it much easier to do regular browser performance profiling: https://developer.chrome.com/docs/devtools/performance
As you often need a reference profile you can use to compare to changes in a PR to see its impact. Being able to quickly create that baseline on https://test-studio.sanity.build is really convenient.
If you need to use a react profiling in production reference frame you can always quickly branch off `next` and push it to have vercel create that reference deployment for you :)

### What to review

Does this make sense? Should we add a command in the root `package.json`, for example `dev:test-studio-production-profiling`?

### Testing

You can try the React profiler on the branch build with the regular React Devtools here: https://test-studio-git-react-production-profiling.sanity.build

And locally by running:
```
cd dev/test-studio && REACT_PRODUCTION_PROFILING=true pnpm turbo run build && pnpm start
```

### Notes for release

N/A only affects the test studios on the monorepo.

### DIFF
--- a/dev/test-next-studio/next.config.mjs
+++ b/dev/test-next-studio/next.config.mjs
@@ -2,6 +2,10 @@ function requireResolve(id) {
   return import.meta.resolve(id).replace('file://', '')
 }
 
+const reactCompiler = process.env.REACT_COMPILER === 'true'
+const reactProductionProfiling = process.env.REACT_PRODUCTION_PROFILING === 'true'
+const productionBrowserSourceMaps = reactCompiler || reactProductionProfiling
+
 // eslint-disable-next-line tsdoc/syntax
 /** @type {import('next').NextConfig} */
 const config = {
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Sweet!

PR #7199, opened 2024-07-18
fix: ensure context provider values are memoized

stipsan:
### Description

All the React Context Providers in this PR suffers from the same problem:
- Each time react renders the `value` prop is given a new object.
- Since the object has a new referential identity, all related `useContext` hooks will also rerender.
- This happens regardless of any object properties in `value` has actually changed and dramatically increases unnecessary rerenders.

There are two ways to combat this:
1. Have multiple context providers, for the same reasons you'd rather have multiple `useState` hooks instead of cramming everything into a single `useState` hook, React performs far better with a large amount of contexts that are small and individually change infrequently [(the react docs touches about this briefly, and its example splits up the state and dispatcher in separate patterns)](https://react.dev/learn/scaling-up-with-reducer-and-context), than few contexts that are massive and thus change often.
2. Wrap the object in `useMemo` to ensure a new object is only created if one of the object properties have changed, [as per best practice in the React docs](https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions).

This PR implements (2), I hope that moving forward we can make an effort to ensure new code is written following pattern (1) 😄 

### What to review

Does this make sense?

### Testing

Existing tests should be sufficient. Once https://github.com/sanity-io/sanity/pull/7198 lands it's easier to measure the impact of the reduced rerenders.

### Notes for release

N/A - the perf boost isn't noticeable enough to end users to warrant a callout, at least not in isolation (we're currently landing a lot of minor fixes like these so all together they do make a good perf impact)

### DIFF
--- a/packages/sanity/src/core/components/previewCard/PreviewCard.tsx
+++ b/packages/sanity/src/core/components/previewCard/PreviewCard.tsx
@@ -1,5 +1,5 @@
 import {Card, type CardProps} from '@sanity/ui'
-import {type ForwardedRef, forwardRef, type HTMLProps, useContext} from 'react'
+import {type ForwardedRef, forwardRef, type HTMLProps, useContext, useMemo} from 'react'
 import {PreviewCardContext} from 'sanity/_singletons'
 import {css, styled} from 'styled-components'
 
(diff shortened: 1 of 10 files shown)


binoy14 (approved):
This is great, thanks @stipsan !

PR #7200, opened 2024-07-18
fix: ensure search context provider value is memoized

stipsan:
### Description

Relates to #7199, read its description for all the deets. The SearchProvider warrants its own PR since it required more work than simply wrapping the context value in a `useMemo`.

### What to review

Does it make sense?

### Testing

If you can click outside the global search and it closes successfully then it works 🎉 

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/common/SearchWrapper.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/common/SearchWrapper.tsx
@@ -44,7 +44,11 @@ export function SearchWrapper({
    * Set shared `onClose` in search context
    */
   useEffect(() => {
-    setOnClose(handleClose)
+    /**
+     * When using useState you have to use the function callback version of setState,
+     *  otherwise it'll call your function and set the state to whatever your function return.
+     */
+    setOnClose(() => handleClose)
   }, [handleClose, setOnClose])
 
   /**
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/studio/components/navbar/search/components/common/SearchWrapper.tsx
```diff
@@ -44,7 +44,7 @@ export function SearchWrapper({
    * Set shared `onClose` in search context
    */
   useEffect(() => {
-    setOnClose(handleClose)
+    setOnClose(() => handleClose)
```
bjoerge, line 47: this doesn't seem right: surely `handleClose` should be called, no?

stipsan: No, it used to be set as a ref, now we use useState. When using useState you have to use the function callback version of setState otherwise it'll call your function and set the state to whatever your function return.

The reason we use state instead of refs here is because we want the memoized context value to change when `setOnClose` is called 🙌

bjoerge: Ah, thanks, that makes sense (migth be worth adding that as a comment since it stands out as something that might not have been intentional)

PR #7207, opened 2024-07-19
feat: add documents validation

pedrobonamin:
### Description

It introduces document validation using the preview store listener in the bundles details screen.
It shows in the UI (WIP) the documents with validation errors.

It also refactors document-store pair validation function by abstracting the part of that function that takes care of **validating** the document from the **fetcher**. 
The document-store pair uses the document-store to fetch the document, which makes it not scalable enough for this use case, by dividing the validation from the fetcher, we could provide a document to the function `validateDocumentWithReferences` and this function will take care of the validation. 

https://www.loom.com/share/0a027910d89844868206d2b54c824cb3?sid=84e86aaa-de88-46f2-9917-4b20ea14442b


packages/sanity/src/core/releases/tool/detail/review/DocumentReviewHeader.tsx
```diff
@@ -1,67 +1,76 @@
...
-            text={
-              <span>
-                Created <RelativeTime time={document._createdAt} useTemporalPhrase />
-              </span>
-            }
+    <Card
+      tone={validation?.hasError ? 'critical' : 'default'}
+      radius={3}
+      style={{marginBottom: 20}}
+    >
```
pedrobonamin, line 35: This was added to this component to wrap it and show the validation error, only change here.

packages/sanity/src/core/store/_legacy/document/document-pair/validation.ts
```diff
@@ -114,81 +49,7 @@ export const validation = memoize(
       shareLatestWithRefCount(),
     )
 
-    const referenceIds$ = document$.pipe(
```
pedrobonamin, line 117: all this logic below has been moved to [`validateDocumentWithReferences`](https://github.com/sanity-io/sanity/pull/7207/files#diff-f33b0c793316d3d8640c334231c4ecdef6eca59ba5b8b5bbab0698d46d33a906)

packages/sanity/src/core/validation/validateDocumentWithReferences.ts
```diff
@@ -0,0 +1,171 @@
+import {type SanityClient} from '@sanity/client'
```
pedrobonamin, line 1: All of this was previously handled by `document-pair/validation` 
It was moved here to make it possible to reuse by receiving `document$: Observable<SanityDocument>`

packages/sanity/src/core/releases/tool/detail/bundleDocumentsValidation.ts
```diff
@@ -0,0 +1,71 @@
...
+          ...validation,
+          documentId: id,
+          hasError: validation.validation.some((marker) => isValidationErrorMarker(marker)),
+        })),
+      )
+    }),
+  ).pipe(
+    // Transform to a dictionary with the id and value
+    scan((acc, next) => {
+      const newMap = new Map(acc) // Create a new Map instance based on the current accumulator
```
bjoerge, line 63: Is there any benefit of using a `Map` here? I guess in theory we get faster lookups, but we're also instantiating new maps at each change and wondering if using plain objects would work just as well (we'd have to instantiate those too ofc, but might be cheaper?). I'm mostly curious, although we tend to be quite consistent about passing plain objects as React props, so diverging from that may complicate things slightly later on.

packages/sanity/src/core/releases/tool/detail/useBundleDocumentsValidation.ts
```diff
@@ -0,0 +1,38 @@
...
+  const observable = useMemo(() => {
+    return bundleDocumentsValidation(
+      {
+        observeDocumentPairAvailability: unstable_observeDocumentPairAvailability,
+        observeDocument: unstable_observeDocument,
+        schema,
+        i18n,
+        getClient,
+      },
+      bundlesIds.split(','),
```
bjoerge, line 25: Should make a note here that this is a hack and something that we should look into fixing later

packages/sanity/src/core/validation/validateDocumentWithReferences.ts
bjoerge: Very nice!

packages/sanity/src/core/releases/tool/detail/bundleDocumentsValidation.ts
pedrobonamin: Not really, I just prefer the map api vs plain objects when creating it for this type of use cases. Given the size of the object or map is quite small I don't think there will really be a performance difference when accessing it. 
Will change to an object for consistency

pedrobonamin: did the same change for document history, which was also using a map.

bjoerge (approved):
Great work!

PR #7208, opened 2024-07-19
refactor: update GROQ query for document versions

RitaDias:
### Description

- Remove dummy getter
- Set up a listener for the new groq query (warning: it's really slow)
- Update usage on DocumentVersion and Footer actions
- Update tests
- Add hook test
- Remove dummyGetters & separate methods
- Rename "dummy getters" rip

Heads up: hook test is bare bones, I didn't have time to do more (maintenance work)


packages/sanity/src/core/bundles/components/dialog/CreateBundleDialog.tsx
```diff
@@ -31,7 +35,7 @@ export function CreateBundleDialog(props: CreateBundleDialogProps): JSX.Element
 
   const handleOnSubmit = useCallback(
     async (event: FormEvent<HTMLFormElement>) => {
-      if (value.slug) {
+      if (value.slug && versionDocumentExists(bundles, value.slug)) {
```
RitaDias, line 38: if the versioned item of this document already exists then we shouldn't add it again

packages/sanity/src/core/bundles/components/dialog/CreateBundleDialog.tsx
jordanl17, line 1: Sorry again @RitaDias  - I think these changes will need to be made to `BundleDetailsDialog` instead of here

RitaDias: No need to apologise, I'll do it! :)

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -0,0 +1,121 @@
...
+ */
+export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPerspectiveState {
+  const {documentId} = props
+
+  const [state, setState] = useState<BundleDocument[] | null>(null)
+
+  const [error, setError] = useState<Error | undefined>(undefined)
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
+  const {data: bundles} = useBundles()
+  const pureDocumentId =
```
bjoerge, line 42: We tend to call this "publishedId" elsewhere

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -0,0 +1,121 @@
...
+    documentId.indexOf('.') > 0
```
bjoerge, line 43: There's already a `getPublishedId()` in this package, could we reuse that instead?

bjoerge (commented):
Solid work! Just a couple of comments/suggestions, nothing big. Thanks for adding tests as well!

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
RitaDias: getPublishedId only accounts for the drafts.(id) structure.
Would we be ok with changing this? I just thought about this way because I was concerned about changing those things 🤔

bjoerge: That's a good point. Could we add a new function (`getBundlePublishedId()`?) that includes the logic for bundle ids for now? Id like for us to have a central place to keep this logic instead of spreading it around.

RitaDias: Ahhh I actually see that @juice49  added something for this already in the getPublishedId
image

I'm thinking about moving this to a `getbundlePublishedId`, thoughts?

bjoerge: Seems to have caused a test failure 🤔

RitaDias: Just noticed, forgot to add the mock!
Should be fine now (just pushed the update)

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * Fetches the document versions for a given document
+ * @param props - document Id of the document (might include bundle slug)
+ * @returns - data: document versions, loading, errors
+ * @hidden
+ * @beta
+ */
+export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPerspectiveState {
```
pedrobonamin, line 35: could this be replaced by the `useListener` hook? It seems to handle the same logic.

pedrobonamin: Or for `listenQuery` ? And maybe we could update the projection to get only the id of the documents?

pedrobonamin, line 35: Have you considered here instead of setting the listener, use the `listenQuery` function? 
Also, probably we can reduce the data returned from the query, projecting it to include only the id
 `*[sanity::versionOf("${publishedId}")]{_id}`

pedrobonamin: Sorry if I'm missing something, but this looks like it could be handled with the `listenQuery` observer and simplify this hook. 

```
export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPerspectiveState {
  const {documentId} = props

  const {data: bundles} = useBundles()
  // TODO: How to identify documents with multiple dots ids, maybe it will be better to provide the whole document?
  const publishedId = getPublishedId(documentId, documentId.includes('.'))

  const QUERY = `*[sanity::versionOf("${publishedId}")]{_id, _version}`

  const observable = useMemo(
    () => listenQuery(client, QUERY) as Observable<Pick<SanityDocument, '_id' | '_version'>[]>,
    [QUERY, client],
  )

   // This will return the sanity documents that are a version of the published document.
  const versions = useObservable(observable, null)

  const state = versions
    ? versions
        .map((r) => bundles?.find((b) => r._version && getBundleSlug(r._id) === b.slug))
        .filter((b) => b !== undefined)
    : null

  return {
    data: state,
  }
}

```

bjoerge: I thought about this too, but I don't think it's important enough to block merging, and that it's something that can be addressed down the line? I think it makes sense for us to do a separate refactor/optimization pass to unify some of the data fetching patterns at a later point.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
```diff
@@ -21,33 +18,14 @@ const BadgeButton = styled(Button)({
...
 export function DocumentPerspectiveMenu(props: {documentId: string}): JSX.Element {
   const {documentId} = props
-  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
-  const {data} = useBundles()
   const {currentGlobalBundle} = usePerspective()
-  const bundles = useMemo(() => data ?? [], [data])
 
   const existsInBundle = getBundleSlug(documentId) === currentGlobalBundle?.slug
   const {title, hue, icon, slug} = currentGlobalBundle
+  const {data: documentVersions} = useDocumentVersions({documentId})
```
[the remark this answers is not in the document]
bjoerge: Good point. I'll leave it to @RitaDias to decide whether to do that in a follow-up PR or before merging in this one.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
```diff
@@ -65,16 +43,16 @@ export function DocumentPerspectiveMenu(props: {documentId: string}): JSX.Elemen
...
-            bundles={documentVersions}
-            loading={!documentVersions}
-          />
-        </Box>
-      )}
+
+      <Box flex="none">
+        <BundleMenu
+          button={<Button icon={ChevronDownIcon} mode="bleed" padding={2} space={2} />}
+          bundles={documentVersions as BundleDocument[]}
```
jordanl17, line 52: `BundleMenu` seems to accepts `BundleDocument[] | null`, and I think that's also the type of `documentVersions`, so might not need to do the type assertion here

packages/sanity/src/core/bundles/hooks/__tests__/useDocumentVersions.test.tsx
jordanl17, line 1: Nice tests for this hook!!

RitaDias: ❤️

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
RitaDias: You're right, might have been because of something I did before messing around with the types, removed it :)

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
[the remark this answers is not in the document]
pedrobonamin: re `BundleActions` I think they will eventually change, specially after the demo, probably we will want it in a more obvious place, but that won't be decided until Marius is back.
Not sure how that change will impact this decision, but I think moving it to the PaneProvider will make it available for all the possible consumers. It can be done in a follow-up PR once you are back from support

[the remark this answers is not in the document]
RitaDias: WIll create a Linear story since I'll be on holiday next week 👍

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
pedrobonamin: Agree, this suggestion is non blocking, it can be addressed later.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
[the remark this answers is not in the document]
RitaDias: Linear story created 👍

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
RitaDias: I've updated with the suggestion from Pedro on slack :) 
Thanks once more!

RitaDias: Spoken on slack but had to drop the update due to time limitations (specifically around tests) created a linear story for it.

pedrobonamin (approved):
Great work @RitaDias , thanks!

PR #7209, opened 2024-07-19
fix(structure): make sync document toast dismisable

pedrobonamin:
### Description

An issue is present in structure in which sync toasts are forever present. 
For what was discussed with the internal team, is that the toast is persisted even though the condition for it is gone or and even if the document is not in view anymore. 

This seems to be caused because when the `FormView` component is unmounted, the toast doesn't trigger a close action, to solve this a cleanup function was added to the useEffect to close the toast if it is still shown.

This seems to be happening even if the `FormView` is not visually unmounted (document is still open), this seems to be caused because the `FormView` component takes a key [` key={documentId + (ready ? '_ready' : '_pending')}`](https://github.com/sanity-io/sanity/blob/dismissable-document-sync-toast/packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx#L174-L175), so when the key changes, the form is unmounted and mounted again, but given there are no clean up functions in the conditional toast hook, the toast will be kept forever.

In this video two buttons were added to force the states that show or hide the toast.

https://github.com/user-attachments/assets/083e65d5-c157-45fd-8bda-090e1505adb4

It also adds support to close the **syncing document** toast.

Screenshot 2024-07-19 at 16 06 47

### DIFF
--- a/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
+++ b/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
@@ -67,14 +67,19 @@ export const FormView = forwardRef<HTMLDivElement, FormViewProps>(function FormV
   const isLocked = editState?.transactionSyncLock?.enabled
   const {t} = useTranslation(structureLocaleNamespace)
 
-  useConditionalToast({
-    id: `sync-lock-${documentId}`,
-    status: 'warning',
-    enabled: isLocked,
-    title: t('document-view.form-view.sync-lock-toast.title'),
-    description: t('document-view.form-view.sync-lock-toast.description'),
-    closable: true,
-  })
+  const conditionalToastParams = useMemo(
+    () => ({
+      id: `sync-lock-${documentId}`,
+      status: 'warning' as const,
+      enabled: isLocked,
+      title: t('document-view.form-view.sync-lock-toast.title'),
+      description: t('document-view.form-view.sync-lock-toast.description'),
+      closable: true,
+    }),
+    [documentId, isLocked, t],
+  )
+
+  useConditionalToast(conditionalToastParams)
 
   useEffect(() => {
     const sub = documentStore.pair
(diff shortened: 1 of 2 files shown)


packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
```diff
@@ -67,13 +67,19 @@ export const FormView = forwardRef<HTMLDivElement, FormViewProps>(function FormV
...
+      id: `sync-lock-${documentId}`,
+      status: 'warning' as const,
+      enabled: isLocked,
+      title: t('document-view.form-view.sync-lock-toast.title'),
+      description: t('document-view.form-view.sync-lock-toast.description'),
+      closable: true,
+    }),
+    [documentId, isLocked, t],
+  )
+
```
pedrobonamin, line 81: Memoizing this params to avoid unnecessary calls to the useEffect in the `useConditionalToast` hook, which uses the params as a dependency.

packages/sanity/src/structure/panes/document/documentPanel/documentViews/useConditionalToast.ts
```diff
@@ -32,5 +32,15 @@ export function useConditionalToast(params: ToastParams & {id: string; enabled?:
...
+    return () => {
+      // Close the toast when the component unmounts, even if the condition is still true
+      if (wasEnabled && params.enabled) {
+        toast.push({
+          ...params,
+          // Note: @sanity/ui fallbacks to the default duration of 4s in case of falsey values
+          duration: 0.01,
+        })
+      }
+    }
```
pedrobonamin, line 44: This will close the toast when the component unmounts, preventing it to be forever present

packages/sanity/src/structure/panes/document/documentPanel/documentViews/useConditionalToast.ts
```diff
@@ -18,19 +10,20 @@ const LONG_ENOUGH_BUT_NOT_TOO_LONG = 1000 * 60 * 60 * 24 * 24
 export function useConditionalToast(params: ToastParams & {id: string; enabled?: boolean}) {
   const toast = useToast()
 
-  const wasEnabled = usePrevious(params.enabled)
```
pedrobonamin, line 21: This is not necessary now that the cleanup function was added to the useEffect.
If the params.enabled condition was true, then the cleanup will take care of closing the toast when params.enabled changes or when the component is unmounted

ricokahler (approved):

PR #7221, opened 2024-07-22
fix: react hook linter supressions in `SearchFilterAssetInput`

stipsan:
### Description

Renames the function from `SearchFilterAssetInput` to `renderSearchFilterAssetInput` to help the linter understand that it's not actually violating hook rules.
Removes an unused `useSearchState` hook as well which reduces unnecessary rerenders.

### What to review

Does it make sense?

### Testing

Existing tests are enough.

### Notes for release

N/A


stipsan:
This change is no longer necessary after bumping the react compiler linter 🥳

PR #7222, opened 2024-07-22
fix: remove react hooks linter suppressions in PT hooks

stipsan:
### Description

Removes linter suppressions by refactoring the hook to reference all the dependencies it has, so it's no longer implicit.

### What to review

Does it make sense?

### Testing

Existing tests are enough.

### Notes for release

N/A


bjoerge:
I'm afraid I don't have enough context to review this, so unassigning myself

christianhg (approved):
I think this looks good and everything seems to work as before.

PR #7225, opened 2024-07-22
chore: add `dev:test-studio-production-profiling` script

stipsan:
Run `pnpm dev:test-studio-production-profiling` to have a production build of the test studio on http://localhost:3333 that supports the React Profiler devtools :)

It's important to note that this mode does not have Hot Module Reloading, you'll need to restart the process when you make changes for them to apply 🙌


bjoerge (approved):
Thank you! <3

PR #7227, opened 2024-07-22
fix: `WebSocket is closed before the connection is established` warning

stipsan:
### Description

In production we have this very annoying warning:
```bash
WebSocket connection to 'wss://:projectId.api.sanity.io/v2022-06-30/socket/:dataset?tag=sanity.studio' failed: WebSocket is closed before the connection is established.
```

[Here's](https://test-studio.sanity.build/test/structure/species;2kXlsm7wlxP522CbJpVLNZ) both the warning, and the Network tab, showing that a WSS were initiated and discarded:

image

After moving the `useGlobalPresence` and `useDocumentPresence` hooks to `useObservable` from `react-rx`, the problem is resolved in production:

image

It's not resolved when using `sanity dev`. I'm not sure yet why that is, but I suspect it's a timing problem with the observable in one of its dependencies.

The change from `useMemo` to `useState` in `ResourceCacheProvider` is to reduce possible issues in StrictMode as the hook is called twice there: https://react.dev/reference/react/useMemo#my-calculation-runs-twice-on-every-re-render

### What to review

It makes sense?

### Testing

Test that global presence and document presence still works ok (it does on my tests, it's nice with a second pair of eyes).

### Notes for release

No longer warns about `WebSocket is closed before the connection is established` in production.


stipsan:
Closing this as after I rebased this PR onto `next` it now makes matters worse.
Instead of removing the `console.warn` in production, while dev mode still has a single `console.warn`, production **_now logs twice_** 🤯:
image

Dev mode still logs once. I don't have time to bisect and chase what commit on `next` that were made since this PR originally opened, and now no longer fixes the issue, so I'll close this and leave it for @sanity-io/studio-dx to dive in and get to the bottom of the production `console.warn` 🙌 💖

PR #7228, opened 2024-07-22
fix(core): use anchor link to switch between datasets

cngonzalez:
### Description
In embedded studios, we noticed that not all document operations being sent were going to the chosen workspace's dataset. While there's a very specific set of conditions that might cause this issue, it is potentially destructive. 

Although we don't believe any customers have been affected by this, out of an abundance of caution, we felt it better to ensure context was entirely refreshed between workspaces by using an anchor link to switch between them instead.

This does add to increased friction in changing between workspaces in the studio, but we hope to resolve this soon by addressing the underlying issue -- this is a temporary fix.

### What to review

The changed file is pretty minimal -- please ensure that functionality remains the same.

### Testing

As mentioned above, the conditions that caused the leaky mutations were pretty specific and difficult to include in an end-to-end test. Please feel free to reach out to me for manual testing.

### Notes for release

Switching between workspaces now uses an anchor link, resolving an issue where document operations might occur across datasets.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/workspace/WorkspaceMenuButton.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/workspace/WorkspaceMenuButton.tsx
@@ -7,7 +7,6 @@ import {
   Menu,
   Text,
 } from '@sanity/ui'
-import {useRouter} from 'sanity/router'
 import {styled} from 'styled-components'
 
 import {MenuButton, type MenuButtonProps, MenuItem, Tooltip} from '../../../../../ui-components'
@@ -21,7 +20,6 @@ const StyledMenu = styled(Menu)`
   max-width: 350px;
   min-width: 250px;
 `
-
 const POPOVER_PROPS: MenuButtonProps['popover'] = {
   constrainSize: true,
   fallbackPlacements: ['bottom-end', 'bottom'],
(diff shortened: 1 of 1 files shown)


RitaDias (approved):
This looks good to me,
A quick pointer though, if you have the time, I understand that the use case is complex but is there really no way to add some tests? 🤔 I'm just thinking that these cases, considering how serious they can be, should definitely have a way of getting automated tests

cngonzalez:
> This looks good to me, A quick pointer though, if you have the time, I understand that the use case is complex but is there really no way to add some tests? 🤔 I'm just thinking that these cases, considering how serious they can be, should definitely have a way of getting automated tested

You're absolutely right about that. I don't think there's a way to test this outside of e2e testing (and I know we're a bit wary of adding more e2e tests), and it would mean a bit of a refactor of our e2e suite to create 2 datasets per worker / PR rather than one. It's not the worst thing in the world, but it's something I'd want to see how it performs before inflicting on the rest of the team. I can bring it up in our weekly sync today!

cngonzalez:
> > This looks good to me, A quick pointer though, if you have the time, I understand that the use case is complex but is there really no way to add some tests? 🤔 I'm just thinking that these cases, considering how serious they can be, should definitely have a way of getting automated tested
> 
> You're absolutely right about that. I don't think there's a way to test this outside of e2e testing (and I know we're a bit wary of adding more e2e tests), and it would mean a bit of a refactor of our e2e suite to create 2 datasets per worker / PR rather than one. It's not the worst thing in the world, but it's something I'd want to see how it performs before inflicting on the rest of the team. I can bring it up in our weekly sync today!

Per team meeting, we should dedicate some time to multi-dataset testing, along with the fix for this.

PR #7230, opened 2024-07-23
test(releases): testing release detail/summary screen

jordanl17:
### Description
* Adds tests for release details (the header)
* Adds tests for release summary (list of documents)
* Removes migration from name to slug script (as this has now run)


packages/sanity/src/core/releases/tool/Migrate.tsx
jordanl17, line 1: @pedrobonamin figured that we no longer needed this Migration script as we've migrated all the existing bundles to slug. Let me know if I'm wrong though!

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseSummary.test.tsx
```diff
@@ -0,0 +1,220 @@
...
+describe('ReleaseSummary', () => {
+  beforeEach(async () => {
+    jest.clearAllMocks()
+
+    await renderTest({})
+  })
+
+  it('lists the release title and description', () => {
+    screen.getByText('Release title')
+    screen.getByText('Release description')
```
RitaDias, line 129: just to double check, I think these tests need an expect statement right?
A getBy will work but it's not something we should likely rely on 🤔

bjoerge (approved):

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseSummary.test.tsx
jordanl17: These tests are ok without an expect I believe. `getBy...` will throw if the element doesn't exist and fail the test. This is a pattern I remember being lauded many years ago when `testing-library/react` was first called `react-testing-library`. `.toBeInDocument` of course also works

RitaDias: huh 🤔 
Maybe it's a playwright thing (and not a jest / unit thing), but I know from their guidelines that they heavily suggest using expecting

PR #7234, opened 2024-07-24
fix(structure): update PaneItemPreview to reflect when a document is in the current perspective

pedrobonamin:
### Description

It adds to `core` a new function `getDocumentIsInPerspective` which can be used to know when a document belongs to the current perspective.
```
- document: `summer.my-document-id`, perspective: `bundle.summer` : **true**
- document: `my-document-id`, perspective: `bundle.summer` : **false**
- document: `summer.my-document-id`perspective: `bundle.winter` : **false**
- document: `summer.my-document-id`, perspective: `undefined` : **false**
- document: `my-document-id`, perspective: `undefined` : **true**
- document: `drafts.my-document-id`, perspective: `undefined` : **true**
```

It uses that function in `PaneItemPreview` to create a distinction between documents that are part of the release or not.

Screenshot 2024-07-24 at 09 37 17
Screenshot 2024-07-24 at 09 37 07

### Testing

Tests included for the new function added.

### DIFF
--- a/packages/sanity/src/core/bundles/util/util.test.ts
+++ b/packages/sanity/src/core/bundles/util/util.test.ts
@@ -1,0 +1,41 @@
+import {describe, expect, it} from '@jest/globals'
+
+import {getDocumentIsInPerspective} from './util'
+
+// * - document: `summer.my-document-id`, perspective: `bundle.summer` : **true**
+// * - document: `my-document-id`, perspective: `bundle.summer` : **false**
+// * - document: `summer.my-document-id`perspective: `bundle.winter` : **false**
+// * - document: `summer.my-document-id`, perspective: `undefined` : **false**
+// * - document: `my-document-id`, perspective: `undefined` : **true**
+// * - document: `drafts.my-document-id`, perspective: `undefined` : **true**
+
+describe('getDocumentIsInPerspective', () => {
+  it('should return true if document is in the current perspective', () => {
+    expect(getDocumentIsInPerspective('summer.my-document-id', 'bundle.summer')).toBe(true)
+  })
+
+  it('should return false if document is not a version  document a perspective is provided', () => {
+    expect(getDocumentIsInPerspective('my-document-id', 'bundle.summer')).toBe(false)
+  })
+
+  it('should return false if document is not in the current perspective', () => {
+    expect(getDocumentIsInPerspective('summer.my-document-id', 'bundle.winter')).toBe(false)
+  })
+
+  it('should return false if document is a version  document a no perspective is provided', () => {
+    expect(getDocumentIsInPerspective('summer.my-document-id', undefined)).toBe(false)
+  })
+
+  it("should return true if the document is in the 'Published' perspective, and no perspective is provided", () => {
+    expect(getDocumentIsInPerspective('my-document-id', undefined)).toBe(true)
+  })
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/bundles/util/util.ts
```diff
@@ -12,6 +12,34 @@ export function getBundleSlug(documentId: string): string {
...
+ * - document: `my-document-id`, perspective: `undefined` : **true**
+ * - document: `drafts.my-document-id`, perspective: `undefined` : **true**
+ */
+export function getDocumentIsInPerspective(
+  documentId: string,
+  perspective: string | undefined,
+): boolean {
+  const bundleSlug = getBundleSlug(documentId)
+
+  if (!perspective) return bundleSlug === 'Published' || bundleSlug === 'drafts'
```
bjoerge, line 34: is there a reason for the inconsistency in casing here?

pedrobonamin: Yes, [getBundleSlug](https://github.com/sanity-io/sanity/blob/corel-74/packages/sanity/src/core/bundles/util/util.ts#L10-L11 )  returns "Published" when the document id is just `foo` and it will return "drafts" if the document id is `drafts.foo`
```
export function getBundleSlug(documentId: string): string {
  if (documentId.indexOf('.') === -1) return 'Published'
  const version = documentId.slice(0, documentId.indexOf('.'))
  return version
}
```

bjoerge: That doesn't seem safe to me. It's [technically possible](https://www.sanity.io/docs/ids#10abf7adf05a) to create a document with the id `Published.foo`. I suggest we refactor this to return `[bundleId, publishedId]` instead, and if you give it published id, it returns `[null, publishedId]`

packages/sanity/src/core/bundles/util/util.test.ts
```diff
@@ -0,0 +1,43 @@
+// Assuming Jest is used for unit testing and getBundleSlug is properly mocked
```
jordanl17, line 1: I'm a little confused by this comment. Does this mean that these tests still need mocking for `getBundleSlug`? Or is this just clarifying what's going on?

pedrobonamin: hehe removed, asked chatgpt for the initial test structure and copy pasted, missed removing that piece. There is no need to mock getBundleSlug

packages/sanity/src/core/bundles/util/util.ts
pedrobonamin: Makes sense, yes I don't think this function is the safest. 
I think this function should eventually take the whole document to check if the `_version` object is present. 

If you have a document with nested ids as you mentioned, it could be understood as a bundle, but it isn't.

packages/sanity/src/core/releases/tool/detail/documentTable/useReleaseHistory.ts
```diff
@@ -51,7 +51,6 @@ export function useReleaseHistory(bundleDocuments: SanityDocument[]): {
   useEffect(() => {
     fetchAndParseAll()
     // When revision changes, update the history.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
pedrobonamin, line 54: Not related, but this is not necessary and it's returning error in the react-compiler check

jordanl17 (approved):

PR #7235, opened 2024-07-24
fix(core): update useDocumentVersions to restore breaking tests

pedrobonamin:
### Description

Fixes an issue that is crashing the Form when removing a document and making tests fail. 

The return value of the `useVersionDocument` hook was. `{data: [undefined]}` and this is making the ui crash because it doesn't handle undefined values inside the array of bundles


RitaDias (approved):
✅

PR #7237, opened 2024-07-24
chore(structure): move useDocumentVersions to PaneProvider

pedrobonamin:
### Description

This moves the invocation of `useDocumentVersions` to the PaneProvider, deduplicating it invocation and making it available for any consumer inside the Pane.


jordanl17 (approved):
Nice!

PR #7241, opened 2024-07-24
chore(core): refactor useDocumentVersion to use ListenQuery

pedrobonamin:
### Description
refactor useDocumentVersion to use `ListenQuery` observable. 

### Testing

Existing tests updated and new tests added.

### DIFF
--- a/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
+++ b/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
@@ -1,14 +1,15 @@
-import {type ListenEvent, type ListenOptions} from '@sanity/client'
-import {useCallback, useEffect, useMemo, useState} from 'react'
-import {catchError, of} from 'rxjs'
-import {
-  type BundleDocument,
-  DEFAULT_STUDIO_CLIENT_OPTIONS,
-  getBundleSlug,
-  getPublishedId,
-  useBundles,
-  useClient,
-} from 'sanity'
+import {type SanityDocument} from '@sanity/types'
+import {useMemo} from 'react'
+import {useObservable} from 'react-rx'
+import {type Observable} from 'rxjs'
+import {type BundleDocument} from 'sanity'
+
+import {useClient} from '../../hooks/useClient'
+import {listenQuery} from '../../store/_legacy/document/listenQuery'
+import {useBundles} from '../../store/bundles/useBundles'
+import {DEFAULT_STUDIO_CLIENT_OPTIONS} from '../../studioClient'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -35,92 +29,31 @@ const LISTEN_OPTIONS: ListenOptions = {
...
-          bundles?.find((b) => r._version && getBundleSlug(r._id) === b.slug),
-        )
-        .filter((b: BundleDocument) => b !== undefined)
-
-      setState(documentBundles)
-    } catch (err) {
-      setError(err)
-    }
-  }, [QUERY, bundles, client])
+  const QUERY = `*[sanity::versionOf("${publishedId}")]{_id, _version}`
```
jordanl17, line 36: Q: need we project to include `_version` when all these docs will have `_version: {}`?

jordanl17: Ah, all good, I see that versionOf also returns draft and published so `_version` is used to find only bundles

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -35,92 +29,31 @@ const LISTEN_OPTIONS: ListenOptions = {
...
+
+  const state = useMemo(
+    () =>
+      versions
+        ? (versions
+            .map((r) => bundles?.find((b) => r._version && getBundleSlug(r._id) === b.slug))
+            .filter(Boolean) as BundleDocument[])
+        : null,
+    [versions, bundles],
+  )
```
jordanl17, line 54: COULD:
just to improve readability with the find nested in the map, I feel this reads a little more clearly:

```suggestion
  const state = useMemo(() => {
    if (!versions) return null

    const validBundles = versions
      .map(({_id, _version}) => {
        if (!_version) return null
        const bundleSlug = getBundleSlug(_id)
        return bundles?.find((bundle) => bundle.slug === bundleSlug) || null
      })
      .filter(Boolean) as BundleDocument[]

    return validBundles
  }, [versions, bundles])
```

jordanl17 (approved):
Looks great, just made one suggestion that's not blocking

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
pedrobonamin: Good suggestion, applied, thanks!

PR #7242, opened 2024-07-24
feat(releases): default sorting version documents in release

jordanl17:
### Description
This PR sorts the list of documents in a bundle by default to show most recent first. This way, when new releases are created they immediately appear at the top of the list, rather than being added at the bottom

### What to review
* possible to now set a default sort col and direction on the `Table`
* Updated a mis-named setter
* Updated the `ReleaseSummary` tests

### DIFF
--- a/packages/sanity/src/core/releases/components/Table/Table.tsx
+++ b/packages/sanity/src/core/releases/components/Table/Table.tsx
@@ -1,13 +1,13 @@
 import {Box, Card, Flex, Stack, Text} from '@sanity/ui'
 import {get} from 'lodash'
-import {Fragment, useMemo} from 'react'
+import {Fragment, useEffect, useMemo} from 'react'
 import {useTableContext} from 'sanity/_singletons'
 import {styled} from 'styled-components'
 
 import {TooltipDelayGroupProvider} from '../../../../ui-components'
 import {LoadingBlock} from '../../../components'
 import {TableHeader} from './TableHeader'
-import {TableProvider} from './TableProvider'
+import {TableProvider, type TableSort} from './TableProvider'
 import {type Column} from './types'
 
 type RowDatum<TableData, AdditionalRowTableData> = AdditionalRowTableData extends undefined
@@ -17,6 +17,7 @@ type RowDatum<TableData, AdditionalRowTableData> = AdditionalRowTableData extend
 export interface TableProps<TableData, AdditionalRowTableData> {
   columnDefs: Column<RowDatum<TableData, AdditionalRowTableData>>[]
   searchFilter?: (data: TableData[], searchTerm: string) => TableData[]
+  defaultSort?: TableSort
   Row?: ({
     datum,
     children,
@@ -50,6 +51,7 @@ const RowStack = styled(Stack)({
 
 const TableInner = <TableData, AdditionalRowTableData>({
   columnDefs,
+  defaultSort,
   data,
   emptyState,
   searchFilter,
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/releases/components/Table/Table.tsx
```diff
@@ -58,7 +60,14 @@ const TableInner = <TableData, AdditionalRowTableData>({
   rowActions,
   loading = false,
 }: TableProps<TableData, AdditionalRowTableData>) => {
-  const {searchTerm, sort} = useTableContext()
+  const {searchTerm, sort, setDefaultSort} = useTableContext()
+
+  const {column: defaultSortColumn, direction: defaultSortDirection} = defaultSort || {}
+  useEffect(() => {
```
juice49, line 66: Small thought here: is there a good reason to have this effect that chains one piece of state to another? Could you group the functions that set the state together using a function or a reducer?

I haven't read the code _super_ closely, so apologies if I've missed something obvious.

juice49 (approved):
This looks great! I left one small comment, but it's not a blocker.

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -177,6 +177,7 @@ export function ReleasesOverview() {
             <Table<TableBundle>
               // for resetting filter and sort on table when mode changed
               key={bundleGroupMode}
+              defaultSort={{column: '_createdAt', direction: 'desc'}}
```
bjoerge, line 180: This should preferably be a constant and moved out of the component so we avoid creating new objects on every render

packages/sanity/src/core/releases/components/Table/Table.tsx
bjoerge: And/or - would it be better if it was passed directly to TableProvider and applied as a default there?

bjoerge (changes requested):
Not clear to me why we need the extra round via setDefaultSort in Table in order to pass the default to `TableProvider` here. Unless I'm missing something it would be great if it could be passed directly to the provider as a prop instead of incurring an extra `setState` => render cycle

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -177,6 +179,7 @@ export function ReleasesOverview() {
             <Table<TableBundle>
               // for resetting filter and sort on table when mode changed
               key={bundleGroupMode}
+              defaultSort={DEFAULT_RELEASES_OVERVIEW_SORT}
```
jordanl17, line 182: @bjoerge defined outwith the component

packages/sanity/src/core/releases/components/Table/Table.tsx
```diff
@@ -190,12 +190,13 @@ const TableInner = <TableData, AdditionalRowTableData>({
...
-  props: TableProps<TableData, AdditionalRowTableData>,
-) => {
+export const Table = <TableData, AdditionalRowTableData = undefined>({
+  defaultSort,
+  ...props
+}: TableProps<TableData, AdditionalRowTableData> & {defaultSort?: TableSort}) => {
   return (
     <TooltipDelayGroupProvider>
-      <TableProvider>
+      <TableProvider defaultSort={defaultSort}>
```
jordanl17, line 199: @bjoerge passing directly to `TableProvider` rather than through context

packages/sanity/src/core/releases/components/Table/Table.tsx
```diff
@@ -190,12 +190,13 @@ const TableInner = <TableData, AdditionalRowTableData>({
   )
 }
 
-export const Table = <TableData, AdditionalRowTableData = undefined>(
-  props: TableProps<TableData, AdditionalRowTableData>,
-) => {
+export const Table = <TableData, AdditionalRowTableData = undefined>({
+  defaultSort,
+  ...props
+}: TableProps<TableData, AdditionalRowTableData> & {defaultSort?: TableSort}) => {
```
bjoerge, line 196: Any reason defaultSort couldn't be part of TableProps? (only curious)

bjoerge (approved):

packages/sanity/src/core/releases/components/Table/Table.tsx
jordanl17: `defaultSort` is only relevant now for the prop passed to the `TableProvider`, whereas `TableProps` is used also in `TableInner` (where defaultSort is not used)

PR #7248, opened 2024-07-25
feat(sanity): request access flow

drewlyton:
### Description
This PR introduces a new "Request Access" flow for non-authorized users trying to access a studio.
SCR-20240725-kkiv

[Link to project/issue in Linear](https://linear.app/sanity/issue/GRO-2298/studio-users-without-access-to-studio-should-be-able-to-request-access)

### What to review

This PR redirects users to a new `RequestAccessScreen` component from the `AuthBoundary` if the user is not authorized _and_ they are not using the `sanity` authentication provider. More information about this decision can be found [here](https://www.notion.so/sanityio/RFC-Technical-Specification-Request-Access-to-a-project-b881b8360dcd459fa6bc9ceaa5216e34?pvs=4#6b500bb5d4bd471088ffaaa7fcd44cda). 

On mount, the `RequestAccessScreen` component makes a call to the `/access/requests/me` endpoint and checks if the user currently has a pending request for this project. If the user does, we show them a "request pending" state.
SCR-20240725-kmrq

If they do not, we show them the form to create a new access request with an optional note:
SCR-20240725-kkiv

Once submitted, the user will see the "request pending" state and a toast message saying their request was submitted.

In the case where they have submitted too many requests in a certain period as defined by the backend, we show them a warning message after submitting.
SCR-20240725-knjz

In the case where there is an error fetching from the Access API, we fallback to the current `NotAuthenticatedScreen`.

### Testing

No automated tests were added for this screen, but testing can be done by running the code against the development studio while logged into a non-authorized account.

### Notes for release

See section above for a description of the new flow. I do not believe any new documentation needs to be written for this feature.


packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+  type CurrentUser,
+  getProviderTitle,
+  LoadingBlock,
+  type SanityClient,
+  useActiveWorkspace,
+} from 'sanity'
+
+import {NotAuthenticatedScreen} from './NotAuthenticatedScreen'
+
+interface AccessRequest {
```
drewlyton, line 14: Not sure if this is where this type should go? I believe this will be the only screen to use it but would prefer to follow standard practice for codebase if applicable

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,200 @@
...
+              // Access request is still pending
+              request.status === 'pending' &&
+              // Access request is less than 2 weeks old
+              new Date(request.createdAt).getTime() > Date.now() - 2 * 1000 * 60 * 60 * 24 * 7,
+          )
+          if (pendingRequests.length) setHasPendingRequest(true)
+        }
+      })
+      .catch((err) => {
+        console.error(err)
```
drewlyton, line 88: Not sure if it's standard practice to send silent/handled errors to the console . In `NotAuthenticatedScreen` is seemed like we just threw an error and crashed the studio if we saw one

packages/sanity/src/core/studio/screens/NotAuthenticatedScreen.tsx
```diff
@@ -51,8 +51,8 @@ export function NotAuthenticatedScreen() {
       >
         <Stack space={4}>
           <Text>
-            You are not authorized to access this studio. Maybe you could ask someone to invite you
-            to collaborate on this project?
+            You are not authorized to access this studio. Please contact someone with access to
```
jordanl17, line 54: COULD: what do you think to localizing this and the RequestAccessScreen too in `packages/sanity/src/core/i18n/bundles/studio.ts`?

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
jordanl17: I think this seems sensible

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,209 @@
...
+      .then((requests) => {
+        if (requests && requests?.length) {
+          const pendingRequests = requests.filter(
+            (request) =>
+              // Access request is for this project
+              request.resourceId === projectId &&
+              // Access request is still pending
+              request.status === 'pending' &&
+              // Access request is less than 2 weeks old
+              new Date(request.createdAt).getTime() > Date.now() - 2 * 1000 * 60 * 60 * 24 * 7,
```
jordanl17, line 82: COULD: use `date-fns`
```suggestion
              isAfter(new Date(request.createdAt), subWeeks(new Date(), 2))
```

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,209 @@
...
+        if (statusCode === 403) {
+          setHasTooManyRequests(true)
+        } else {
+          toast.push({
+            title: 'There was a problem submitting your request. Please try again later.',
+            status: 'error',
+          })
+        }
+      })
+      .finally(() => {
```
jordanl17, line 125: Nice use of `finally` 🔥

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,209 @@
+/* eslint-disable i18next/no-literal-string,@sanity/i18n/no-attribute-string-literals */
+import {Box, Button, Card, Dialog, Flex, Stack, Text, TextInput, useToast} from '@sanity/ui'
```
jordanl17, line 2: SHOULD: prefer to use `Button` and `Dialog`from `ui-components`

jordanl17 (commented):
Looking good, @drewlyton

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
jordanl17: So, after 2 weeks a new request is allowed? Should we clarify to user that they can `re-send` their existing request rather than suggest that no request was existing?

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  // Check if user has a pending
+  // access request for this project
+  useEffect(() => {
+    if (!client || !projectId) return
+    client
+      .request<AccessRequest[] | null>({
+        url: `/access/requests/me`,
+      })
+      .then((requests) => {
```
jordanl17, line 73: Q: Might it be worth checking for any declined requests? I would imagine we could just fallback to `NotAuthenticatedScreen`

svirs: Yup, just added that in now!

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
svirs: This is a good idea! Just made a commit addressing it

packages/sanity/src/core/studio/screens/NotAuthenticatedScreen.tsx
svirs: We're hoping to get the functionality out in the next release and add more to this PR (will be adding testing though), I'll make a ticket to follow up with adding translations to these two screens

svirs:
> Looking good, @drewlyton
> 
> It would be great if we could add some tests though around the `RequestAccessScreen`. Realise there aren't many (if any?) in this area of the codebase currently, but hopefully testing some of the core logic here shouldn't require too much in the way of mocking etc.

There aren't any in the codebase for screens like this, considering this screen specifically needs a unique case of being logged in but not authorized it's quite tricky to get it correctly mocked. I'd like to revisit it alongside the translation PR we can follow up with.

jordanl17 (dismissed):

svirs:
@ricokahler the backend is live now

jordanl17 (approved):

ricokahler (approved):
Nothing blocking but had a few suggestions

packages/sanity/src/core/studio/screens/RequestAccessScreen.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  const [error, setError] = useState<unknown>(null)
+  const [msgError, setMsgError] = useState<string | undefined>()
+  const [loading, setLoading] = useState(true)
+  const [isSubmitting, setIsSubmitting] = useState(false)
+
+  const [hasPendingRequest, setHasPendingRequest] = useState<boolean>(false)
+  const [hasExpiredPendingRequest, setExpiredHasPendingRequest] = useState<boolean>(false)
+  const [hasTooManyRequests, setHasTooManyRequests] = useState<boolean>(false)
+  const [hasBeenDenied, setHasBeenDenied] = useState<boolean>(false)
```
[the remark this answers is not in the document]
svirs: I think this would be a good refactor on the next iteration when we add translations, but for now we'll get the functionality out as-is

ricokahler (approved):

PR #7251, opened 2024-07-26
feat(sanity): support viewing and editing document versions

juice49:
### Description

This branch introduces a new, optional, `version` property to document pairs and their various associated constructs. When a bundle is checked out, the version document takes precedence over the published and draft documents. If no version document exists, the draft and published documents are used.

This branch also moves the checked-out bundle state up to the structure pane node context. This allows each pane to have its own bundle checked out.

Example multi-version split pane URL: `http://localhost:3333/test/structure/book;243dbfde-e1e0-4289-a72e-445e8926b4d4;243dbfde-e1e0-4289-a72e-445e8926b4d4%2Ctype%3Dbook%2Cperspective%3Dbundle.autumn-drop?perspective=bundle.summer-drop`.

I'm wondering whether it might be beneficial to try to get these changes into `next` sooner rather than later.

#### Todo

- [x] Resolve double data loading / rendering when opening documents.
- [ ] API for checking out a bundle in a single pane.
- [x] Resolve lingering type errors.
- [x] Resolve failing tests.
- [ ] We do `perspective?.startsWith('bundle.')` and similar in a lot of places. Now might be a good time to to abstract that.
- [ ] Add tests.
- [x]  Fix live edit badge incorrectly appearing for documents that don't exist in the checked-out bundle.
- [ ] Improve handling of revision list when viewing a document that does not exist in the checked-out bundle (currently shows "No document history").
- [x] Fix bundle identification in `DocumentPerspectiveMenu`.
- [x] Account for version document in `getDocumentPairPermissions`.
- [x] Fix "reference has changed" banner.
- [ ] Make comments version aware.

### What to review

It would be great to get some thoughts about the general approach taken. Does adding a new `version` property to constructs make sense, instead of reusing the existing `draft` property? My reasons for this decision are:

1. There are likely scenarios in which we'll need to make a distinction based on whether we're working with a draft or a version.
2. It's useful to track both the version and the draft, so we can fall back to showing the draft if the version doesn't exist.

Very open to opposing thoughts!

### DIFF
--- a/packages/sanity/src/core/bundles/hooks/usePerspective.tsx
+++ b/packages/sanity/src/core/bundles/hooks/usePerspective.tsx
@@ -15,12 +15,16 @@ export interface PerspectiveValue {
 }
 
 /**
+ * TODO: Improve distinction between global and pane perspectives.
+ *
  * @internal
  */
-export function usePerspective(): PerspectiveValue {
+export function usePerspective(selectedPerspective?: string): PerspectiveValue {
   const router = useRouter()
   const {data: bundles} = useBundles()
+  const perspective = selectedPerspective ?? router.stickyParams.perspective
 
+  // TODO: Should it be possible to set the perspective within a pane, rather than globally?
   const setPerspective = (slug: string | undefined) => {
     if (slug === 'drafts') {
       router.navigateStickyParam('perspective', '')
@@ -28,16 +32,15 @@ export function usePerspective(): PerspectiveValue {
       router.navigateStickyParam('perspective', `bundle.${slug}`)
     }
   }
+
   const selectedBundle =
-    router.stickyParams?.perspective && bundles
+    perspective && bundles
       ? bundles.find((bundle: Partial<BundleDocument>) => {
-          return (
-            `bundle.${bundle.slug}`.toLocaleLowerCase() ===
-            router.stickyParams.perspective?.toLocaleLowerCase()
-          )
+          return `bundle.${bundle.slug}`.toLocaleLowerCase() === perspective?.toLocaleLowerCase()
         })
       : LATEST
 
+  // TODO: Improve naming; this may not be global.
   const currentGlobalBundle = selectedBundle || LATEST
 
   return {
--- a/packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
+++ b/packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
@@ -43,6 +43,7 @@ export interface CommentsProviderProps {
   children: ReactNode
   documentId: string
   documentType: string
+  perspective?: string
   type: CommentsType
   sortOrder: 'asc' | 'desc'
 
@@ -78,20 +79,27 @@ export const CommentsProvider = memo(function CommentsProvider(props: CommentsPr
     selectedCommentId,
     isConnecting,
     onPathOpen,
+    perspective,
   } = props
   const commentsEnabled = useCommentsEnabled()
   const [status, setStatus] = useState<CommentStatus>('open')
   const {client, createAddonDataset, isCreatingDataset} = useAddonDataset()
   const publishedId = getPublishedId(documentId)
-  const editState = useEditState(publishedId, documentType, 'low')
+
+  const bundlePerspective = perspective?.startsWith('bundle.')
+    ? perspective.split('bundle.').at(1)
+    : undefined
+
+  // TODO: Allow versions to have separate comments.
+  const editState = useEditState(publishedId, documentType, 'default', bundlePerspective)
   const schemaType = useSchema().get(documentType)
   const currentUser = useCurrentUser()
 
   const {name: workspaceName, dataset, projectId} = useWorkspace()
 
   const documentValue = useMemo(() => {
-    return editState.draft || editState.published
-  }, [editState.draft, editState.published])
+    return editState.version || editState.draft || editState.published
+  }, [editState.version, editState.draft, editState.published])
 
   const documentRevisionId = useMemo(() => documentValue?._rev, [documentValue])
 
--- a/packages/sanity/src/core/components/documentStatusIndicator/DocumentStatusIndicator.tsx
+++ b/packages/sanity/src/core/components/documentStatusIndicator/DocumentStatusIndicator.tsx
@@ -7,6 +7,7 @@ import {styled} from 'styled-components'
 interface DocumentStatusProps {
   draft?: PreviewValue | Partial<SanityDocument> | null
   published?: PreviewValue | Partial<SanityDocument> | null
+  version?: PreviewValue | Partial<SanityDocument> | null
 }
 
 const Root = styled(Text)`
@@ -25,23 +26,26 @@ const Root = styled(Text)`
  * - Yellow (caution) for published documents with edits
  * - Gray (default) for unpublished documents (with or without edits)
  *
- * No dot will be displayed for published documents without edits.
+ * No dot will be displayed for published documents without edits or for version documents.
  *
  * @internal
  */
-export function DocumentStatusIndicator({draft, published}: DocumentStatusProps) {
+export function DocumentStatusIndicator({draft, published, version}: DocumentStatusProps) {
   const $draft = Boolean(draft)
   const $published = Boolean(published)
+  const $version = Boolean(version)
 
   const status = useMemo(() => {
+    if ($version) return undefined
     if ($draft && !$published) return 'unpublished'
     return 'edited'
-  }, [$draft, $published])
+  }, [$draft, $published, $version])
 
   // Return null if the document is:
   // - Published without edits
   // - Neither published or without edits (this shouldn't be possible)
-  if ((!$draft && !$published) || (!$draft && $published)) {
+  // - A version
+  if ((!$draft && !$published) || (!$draft && $published) || $version) {
     return null
   }
 
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInput.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInput.tsx
@@ -28,10 +28,11 @@ interface Options {
   path: Path
   schemaType: ReferenceSchemaType
   value?: Reference
+  version?: string
 }
 
 export function useReferenceInput(options: Options) {
-  const {path, schemaType} = options
+  const {path, schemaType, version} = options
   const schema = useSchema()
   const documentPreviewStore = useDocumentPreviewStore()
   const {EditReferenceLinkComponent, onEditReference, activePath, initialValueTemplateItems} =
@@ -113,8 +114,8 @@ export function useReferenceInput(options: Options) {
   }, [disableNew, initialValueTemplateItems, schemaType.to])
 
   const getReferenceInfo = useCallback(
-    (id: string) => adapter.getReferenceInfo(documentPreviewStore, id, schemaType),
-    [documentPreviewStore, schemaType],
+    (id: string) => adapter.getReferenceInfo(documentPreviewStore, id, schemaType, {version}),
+    [documentPreviewStore, schemaType, version],
   )
 
   return {
--- a/packages/sanity/src/core/form/studio/FormBuilder.tsx
+++ b/packages/sanity/src/core/form/studio/FormBuilder.tsx
@@ -69,6 +69,7 @@ export interface FormBuilderProps
   schemaType: ObjectSchemaType
   validation: ValidationMarker[]
   value: FormDocumentValue | undefined
+  version?: string
 }
 
 /**
@@ -100,6 +101,7 @@ export function FormBuilder(props: FormBuilderProps) {
     schemaType,
     validation,
     value,
+    version,
   } = props
 
   const handleCollapseField = useCallback(
@@ -278,6 +280,7 @@ export function FormBuilder(props: FormBuilderProps) {
       validation={validation}
       readOnly={readOnly}
       schemaType={schemaType}
+      version={version}
     >
       <GetFormValueProvider value={value}>
         <FormValueProvider value={value}>
--- a/packages/sanity/src/core/preview/documentPair.ts
+++ b/packages/sanity/src/core/preview/documentPair.ts
@@ -14,6 +14,7 @@ export function create_preview_documentPair(
   observePathsDocumentPair: <T extends SanityDocument = SanityDocument>(
     id: string,
     paths: PreviewPath[],
+    options?: {version?: string},
   ) => Observable<DraftsModelDocument<T>>
 } {
   const {observeDocumentPairAvailability} = createPreviewAvailabilityObserver(
@@ -21,17 +22,23 @@ export function create_preview_documentPair(
     observePaths,
   )
 
-  const ALWAYS_INCLUDED_SNAPSHOT_PATHS: PreviewPath[] = [['_updatedAt'], ['_createdAt'], ['_type']]
+  const ALWAYS_INCLUDED_SNAPSHOT_PATHS: PreviewPath[] = [
+    ['_updatedAt'],
+    ['_createdAt'],
+    ['_type'],
+    ['_version'],
+  ]
 
   return {observePathsDocumentPair}
 
   function observePathsDocumentPair<T extends SanityDocument = SanityDocument>(
     id: string,
     paths: PreviewPath[],
+    {version}: {version?: string} = {},
   ): Observable<DraftsModelDocument<T>> {
-    const {draftId, publishedId} = getIdPair(id)
+    const {draftId, publishedId, versionId} = getIdPair(id, {version})
 
-    return observeDocumentPairAvailability(draftId).pipe(
+    return observeDocumentPairAvailability(draftId, {version}).pipe(
       switchMap((availability) => {
         if (!availability.draft.available && !availability.published.available) {
           // short circuit, neither draft nor published is available so no point in trying to get a snapshot
@@ -46,6 +53,14 @@ export function create_preview_documentPair(
               availability: availability.published,
               snapshot: undefined,
             },
+            ...(availability.version
+              ? {
+                  version: {
+                    availability: availability.version,
+                    snapshot: undefined,
+                  },
+                }
+              : {}),
           })
         }
 
@@ -54,10 +69,12 @@ export function create_preview_documentPair(
         return combineLatest([
           observePaths({_type: 'reference', _ref: draftId}, snapshotPaths),
           observePaths({_type: 'reference', _ref: publishedId}, snapshotPaths),
+          ...(version ? [observePaths({_type: 'reference', _ref: versionId}, snapshotPaths)] : []),
         ]).pipe(
-          map(([draftSnapshot, publishedSnapshot]) => {
+          map(([draftSnapshot, publishedSnapshot, versionSnapshot]) => {
             // note: assume type is always the same
             const type =
+              (isRecord(versionSnapshot) && '_type' in versionSnapshot && versionSnapshot._type) ||
               (isRecord(draftSnapshot) && '_type' in draftSnapshot && draftSnapshot._type) ||
               (isRecord(publishedSnapshot) &&
                 '_type' in publishedSnapshot &&
@@ -75,6 +92,14 @@ export function create_preview_documentPair(
                 availability: availability.published,
                 snapshot: publishedSnapshot as T,
               },
+              ...(availability.version
+                ? {
+                    version: {
+                      availability: availability.version,
+                      snapshot: versionSnapshot as T,
+                    },
+                  }
+                : {}),
             }
           }),
         )
--- a/packages/sanity/src/core/preview/documentPreviewStore.ts
+++ b/packages/sanity/src/core/preview/documentPreviewStore.ts
@@ -34,7 +34,7 @@ import {
 export type ObserveForPreviewFn = (
   value: Previewable,
   type: PreviewableType,
-  options?: {viewOptions?: PrepareViewOptions; apiConfig?: ApiConfig; perspective?: string},
+  options?: {viewOptions?: PrepareViewOptions; apiConfig?: ApiConfig},
 ) => Observable<PreparedSnapshot>
 
 /**
@@ -57,11 +57,13 @@ export interface DocumentPreviewStore {
    */
   unstable_observeDocumentPairAvailability: (
     id: string,
+    options?: {version?: string},
   ) => Observable<DraftsModelDocumentAvailability>
 
   unstable_observePathsDocumentPair: <T extends SanityDocument = SanityDocument>(
     id: string,
     paths: PreviewPath[],
+    options?: {version?: string},
   ) => Observable<DraftsModelDocument<T>>
 
   /**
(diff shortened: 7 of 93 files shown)


packages/sanity/src/core/bundles/hooks/usePerspective.tsx
```diff
@@ -1,5 +1,9 @@
 import {useRouter} from 'sanity/router'
 
+// TODO: Importing `usePaneRouter` like this results in an error: `Cannot access 'ScrollContainer' before initialization`.
+// import {usePaneRouter} from 'sanity/structure'
```
pedrobonamin, line 4: Could it be that the  issue is caused because you cannot import from `structure` into `core`

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
```diff
@@ -15,33 +19,45 @@ export interface PerspectiveValue {
...
     currentGlobalBundle,
   }
 }
+
+/**
+ * @internal
+ */
+export function useNearestPerspective(): string | undefined {
+  const router = useRouter()
+  const paneRouter = usePaneRouter()
```
pedrobonamin, line 61: Given this `usePerspective` hook is part of core, I think we shouldn't use here the `paneRouter`

packages/sanity/src/core/components/documentStatusIndicator/DocumentStatusIndicator.tsx
```diff
@@ -25,23 +26,26 @@ const Root = styled(Text)`
...
     return 'edited'
-  }, [$draft, $published])
+  }, [$draft, $published, $version])
 
   // Return null if the document is:
   // - Published without edits
   // - Neither published or without edits (this shouldn't be possible)
-  if ((!$draft && !$published) || (!$draft && $published)) {
+  // - A version
+  if ((!$draft && !$published) || (!$draft && $published) || $version) {
```
pedrobonamin, line 48: I'm not sure about this changes, I think we don't want to check for `version` in the `DocumentStatusIndicator`, not sure if you meant to add it or if this is because of a rebase issue.
It was removed as part of other PR

packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInput.tsx
```diff
@@ -113,8 +115,13 @@ export function useReferenceInput(options: Options) {
   }, [disableNew, initialValueTemplateItems, schemaType.to])
 
   const getReferenceInfo = useCallback(
-    (id: string) => adapter.getReferenceInfo(documentPreviewStore, id, schemaType),
-    [documentPreviewStore, schemaType],
+    (id: string) =>
+      adapter.getReferenceInfo(documentPreviewStore, id, schemaType, {
+        version: perspective?.startsWith('bundle.')
```
pedrobonamin, line 120: nit pick:

Could we create a reusable `getBundleFromPerspective` function?
```
 function getBundleFromPerspective(perspective: string) {
  if(perspective?.startsWith('bundle.')) {
     return perspective.split('bundle.').at(1)
} else return undefined
```

packages/sanity/src/core/store/_legacy/document/document-pair/utils/isLiveEditEnabled.ts
```diff
@@ -1,5 +1,4 @@
 import {type Schema} from '@sanity/types'
 
 export const isLiveEditEnabled = (schema: Schema, typeName: string): boolean =>
-  (globalThis as any).__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED_LIVE_EDIT_OVERRIDE ??
```
pedrobonamin, line 4: Curious why you are removing this
It's a scary env!

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -56,11 +56,27 @@ export function isDraftId(id: string): id is DraftId {
...
+  publishedId: PublishedId
+  versionId?: string
+} {
   return {
-    draftId: getDraftId(id),
     publishedId: getPublishedId(id),
+    draftId: getDraftId(id),
+    ...(version
+      ? {
+          versionId: id.startsWith(`${version}.`) ? id : [version, getPublishedId(id)].join('.'),
```
pedrobonamin, line 77: I think we could benefit from a `getVersionId` function that handles this, which could be reused here https://github.com/sanity-io/sanity/pull/7251/files#diff-e5ae958bd581e7f49548874e1c065e9ac7b325f2d48b753a9bbbd2e4c7649fe8R184

pedrobonamin (commented):
This is looking very good, thanks for this, amazing work! Thanks for this @juice49

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
juice49: That might be the case, but there doesn't seem to be a linting rule preventing it. Either way, it might be a sign that we need to restructure the context usage here. Maybe `usePerspective` should move 🤔.

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
juice49: That's a good point. Maybe `useNearestPerspective` should move. I need to think about the best way to proceed here... if you have any thoughts, let me know 🙏.

packages/sanity/src/core/components/documentStatusIndicator/DocumentStatusIndicator.tsx
juice49: I re-added it intentionally. My thinking here is that we never want the status indicator to reflect the status of the main document when a version is checked out; we don't care about the published version or the draft in this scenario.

That is my assumption, at least. What do you think?

packages/sanity/src/core/form/inputs/ReferenceInput/useReferenceInput.tsx
juice49: I agree! 😅

Screenshot 2024-07-29 at 12 39 46

packages/sanity/src/core/store/_legacy/document/document-pair/utils/isLiveEditEnabled.ts
juice49: It was initially added as part of my hacky (and temporary) [implementation](https://github.com/sanity-io/sanity/commit/c74da9ce626e42b407b7529993fc6b8292151963) of version viewing and editing.

This branch reverts that commit 🙂. If this scary env goes to production, I've messed up!

packages/sanity/src/core/util/draftUtils.ts
juice49: I think that would be great.

packages/sanity/src/core/store/_legacy/document/document-pair/utils/isLiveEditEnabled.ts
pedrobonamin: Sounds good! haha

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
pedrobonamin: I think it's fine that `usePerspective` is part of core, do you think otherwise? 

Maybe the `nearestPerspective` needs to be part of the props received received in form builder? 
The form builder is used in structure to render the form, in the case of a "pane perspective" it can be exposed into the form from in https://github.com/sanity-io/sanity/blob/feat/corel-3/multiple-document-versions/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx#L182-L183

packages/sanity/src/core/store/_legacy/document/document-pair/operationArgs.ts
```diff
@@ -20,27 +20,40 @@ export const operationArgs = memoize(
       historyStore: HistoryStore
       schema: Schema
       serverActionsEnabled: Observable<boolean>
+      bundlePerspective?: string
```
pedrobonamin, line 23: I noticed that this is not currently in use, are you planning to add support here? 
If not, i think it will be better to remove this to avoid confusions

packages/sanity/src/core/preview/documentPreviewStore.ts
```diff
@@ -180,6 +182,7 @@ export function createDocumentPreviewStore({
     unstable_observeDocuments: (ids: string[]) =>
       combineLatest(ids.map((id) => observeDocument(id))),
     unstable_observeDocumentPairAvailability: observeDocumentPairAvailability,
+    // TODO: Account for version.
```
bjoerge, line 185: note: also account for version in observeDocumentPairAvailability

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
```diff
@@ -60,18 +60,17 @@ export interface DocumentVersion {
 /**
  * @hidden
  * @beta */
-// TODO: Rename -> `Bundle`
```
bjoerge, line 63: Did you settle on not renaming to bundle?

packages/sanity/src/core/store/_legacy/document/document-pair/operationArgs.ts
```diff
@@ -20,27 +20,40 @@ export const operationArgs = memoize(
...
-    // TODO: Should be dynamic
-    const draftIndex = 0
     return snapshotPair(ctx.client, idPair, typeName, ctx.serverActionsEnabled).pipe(
       switchMap((versions) =>
         combineLatest([
-          versions.drafts[draftIndex].snapshots$,
+          versions.draft.snapshots$,
           versions.published.snapshots$,
           ctx.serverActionsEnabled,
+          ...(typeof versions.version === 'undefined' ? [] : [versions.version.snapshots$]),
```
bjoerge, line 34: Using an empty array here will effectively create an observable that never emits a value, so I think what you want here is rather `of()` (which will emit `undefined`)

bjoerge (commented):
The direction here looks good to me! Added a few comments here and there.

packages/sanity/src/core/store/_legacy/document/document-store.ts
```diff
@@ -33,12 +33,20 @@ import {type IdPair} from './types'
...
-  return {publishedId, draftIds: [getDraftId(publishedId)]}
+  return {
+    publishedId,
+    draftId: getDraftId(publishedId),
+    ...(version
+      ? {
+          versionId: [version, publishedId].join('.'),
+        }
+      : {}),
+  }
```
pedrobonamin, line 49: This feels like we are doing the same change in two places, and could easily be missed.  https://github.com/sanity-io/sanity/pull/7251/files#diff-cdd1fc4337b45f431507fa1b2ebb702510d7c9472df372033daaa9f1871add8c
What do you think if we replace this for 
```suggestion
   return getIdPair(publishedId, {version})
  }
```

packages/sanity/src/core/store/_legacy/document/document-pair/operationArgs.ts
bjoerge: Ahhh, I didn't consider the spread – you are of course right, nvm :)

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
juice49: Not 100% sure, but I'm leaning towards not renaming. For a couple of reasons:

- The publish and draft documents are always included in the pair. Conceptually, this structure can still be thought of as a *pair*, with one additional (optional) slot for the checked-out version.
- Renaming would touch a lot of surfaces. It's not strictly a problem to do so, but I think it's nice to minimise this churn if we don't have a good reason for it.

I'm happy to rename it if we think that's best, though! "Group" would certainly be more accurate, but it is worth the churn and breaking something folks are already familiar with?

packages/sanity/src/core/store/_legacy/document/document-pair/operationArgs.ts
juice49: Good catch, thanks! I think I accidentally left this dangling from some initial exploration. The `idPair` parameter contains this information, so I don't think there's any reason to replicate it in the `ctx` parameter.

packages/sanity/src/core/store/_legacy/document/document-store.ts
juice49: Oh yeah! That looks like a really nice improvement to me! Thanks.

packages/sanity/src/core/form/studio/FormBuilder.tsx
```diff
@@ -159,8 +161,8 @@ export function FormBuilder(props: FormBuilderProps) {
     [Input],
   )
   const renderField = useCallback(
-    (fieldProps: Omit<FieldProps, 'renderDefault'>) => <Field {...fieldProps} />,
-    [Field],
+    (fieldProps: Omit<FieldProps, 'renderDefault'>) => <Field {...fieldProps} version={version} />,
```
pedrobonamin, line 164: wdyt if instead of adding it to the field we add it to the form provider and we make it accessible for all the components inside the form

juice49: I think that could be a good idea. Definitely not committed to the current approach... this is my first pass at tracing the component tree and figuring out what to do 🙂. I'll take a look now that this is working. Thank you for the idea!

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
juice49: Resolved through a wider restructure. Thank you for your help!

packages/sanity/src/core/form/studio/FormBuilder.tsx
juice49: I've made this change. Thank you 🙌.

packages/sanity/src/core/preview/documentPreviewStore.ts
```diff
@@ -132,6 +133,7 @@ export function createDocumentPreviewStore({
     id: string,
     apiConfig?: ApiConfig,
   ): Observable<string | undefined> {
+    // TODO: Account for version.
```
pedrobonamin, line 136: Is this and the next TODO still valid? You are now receiving the `version` in options. 
Also, the availability seems to account now for version

packages/sanity/src/core/store/_legacy/document/document-pair/editState.ts
```diff
@@ -21,10 +21,24 @@ export interface EditStateFor {
...
+   * Whether live edit is enabled. This may be true for various reasons:
+   *
+   * - The schema type has live edit enabled.
+   * - A version of the document is checked out.
+   */
   liveEdit: boolean
+  /**
+   * Whether the schema type has live edit enabled.
+   */
+  liveEditSchemaType: boolean
```
pedrobonamin, line 36: I wonder if some users have implemented this in a similar way to our liveEditBadge.
Maybe liveEdit should return the same as before and it could be better to have a `versionEdit` ? And a note saying versionEdit emulates a liveEdit, because version documents don't follow the drafts. convention

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormEdit.tsx
```diff
@@ -171,6 +173,9 @@ function FormEditInner(props: ObjectInputProps) {
         documentType="tasks.task"
         sortOrder="asc"
         type="task"
+        // TODO: The pane router can't be used in this file, so it's only possibly to access the
+        // globally checked-out bundle. It may be necessary pass the perspective down instead.
+        perspective={router.stickyParams.perspective}
```
pedrobonamin, line 178: Given this is for tasks and tasks won't have `perspective` I think we can remove this.

packages/sanity/src/structure/panes/StructureToolPane.tsx
```diff
@@ -73,6 +79,7 @@ export const StructureToolPane = memo(
             pane={pane}
           />
         </Suspense>
+        <StructureTitle resolvedPanes={resolvedPanes} />
```
pedrobonamin, line 82: Do we really need to move this here? It was moved from `StructureTool` to here, but it's using the same data. 
` const {resolvedPanes} = useResolvedPanes()`

Maybe this was kept from a previous change and it's not necessary anymore to move it.

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
```diff
@@ -78,12 +79,19 @@ export const CommentsProvider = memo(function CommentsProvider(props: CommentsPr
...
   const {client, createAddonDataset, isCreatingDataset} = useAddonDataset()
   const publishedId = getPublishedId(documentId)
-  const editState = useEditState(publishedId, documentType, 'low')
+
+  const bundlePerspective = perspective?.startsWith('bundle.')
+    ? perspective.split('bundle.').at(1)
+    : undefined
+
+  // TODO: Allow versions to have separate comments.
+  const editState = useEditState(publishedId, documentType, 'default', bundlePerspective)
```
pedrobonamin, line 94: I think we may want to update the `documentValue` memo below to account for the `editState.version` also.
See line 101

pedrobonamin: This is for a follow up, but We may want also to update line 123 to use the version document id when creating a comment inside a version document. I think comments in version documents should be shown only in the version doc and not in the draft or published version

pedrobonamin (commented):
Fantastic work @juice49 thanks for this!
Added some comments and suggestions, none is blocking.

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
```diff
@@ -15,29 +15,32 @@ export interface PerspectiveValue {
 }
 
 /**
+ * TODO: Improve distinction between global and pane perspectives.
+ *
  * @internal
  */
-export function usePerspective(): PerspectiveValue {
+export function usePerspective(selectedPerspective?: string): PerspectiveValue {
```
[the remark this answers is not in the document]
juice49: Thank you! Adding this parameter was a bandaid, and I did plan to come back with _some_ more elegant solution. I think your suggestion is definitely much better.


I also wonder whether requiring the local override at all is a bit of a code smell 🤔.

packages/sanity/src/core/preview/documentPair.ts
```diff
@@ -14,24 +14,31 @@ export function create_preview_documentPair(
...
     versionedClient,
     observePaths,
   )
 
-  const ALWAYS_INCLUDED_SNAPSHOT_PATHS: PreviewPath[] = [['_updatedAt'], ['_createdAt'], ['_type']]
+  const ALWAYS_INCLUDED_SNAPSHOT_PATHS: PreviewPath[] = [
+    ['_updatedAt'],
+    ['_createdAt'],
+    ['_type'],
+    ['_version'],
```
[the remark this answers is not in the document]
juice49: Yes, great idea. I've also been considering adding a constant for this field that we can reuse across the code.

packages/sanity/src/core/preview/documentPreviewStore.ts
juice49: Great point, and I think you're correct. `observeDocumentTypeFromId` is called for each version of the document (published, draft, checked-out version), so the implementation itself doesn't need to deal with this. I'll remove the TODO.

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormEdit.tsx
juice49: Great point. The task documents themselves don't need to have different versions (I think that's what you mean?).

I got my wires crossed here, because we probably do want to allow different version of document to have different tasks. But each of those task documents have only one version.

packages/sanity/src/structure/panes/StructureToolPane.tsx
juice49: Yep, I think you're absolutely right about why this has moved 😅. I'll take a look at moving it back.

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
juice49: Great points, and you're absolutely right that this (almost certainly incorrectly) does not allow users to leave different comments on different versions. Taking a look...

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/FormEdit.tsx
pedrobonamin: > The task documents themselves don't need to have different versions
Yes! 

This form is to create the task document, and the task document won't have versions, all the ids of the task documents are just published ids.
 
But a version document could have tasks, but that is handled by the id that the task document is referencing, not by the id of the task document.

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
pedrobonamin: I added two tickets to enable **tasks** and **comments** in version documents, we can look at that in those tickets. 
But I think we may want to do the change on line 101 on this PR, it could be harder to identify later why the comments provider is not working as expected.

packages/sanity/src/core/bundles/hooks/usePerspective.tsx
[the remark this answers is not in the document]
pedrobonamin: Yeah I'm not too happy with this solution either, I was trying to think of a way to remove this, I think it would require a refactor in the bundleMenu given is the only one that needs to work in two cases (local and global).

I was thinking on the option to add another hook in structure to handle only the "pane" perspective and maybe we could rename this one to `useGlobalPerspective` 

They will have a very similar behavior and could expose the same actions, but each one will handle the actions in a different way. 

Anyways, I think it could be a follow up and this shouldn't block your changes, seeing that change alone will be easier to identify the best approach.

packages/sanity/src/core/store/_legacy/document/document-pair/editState.ts
pedrobonamin: I like this approach, if it's a big change it can also be a follow up

packages/sanity/src/core/util/draftUtils.ts
RitaDias: This is being added in a separate PR, or are you adding one already in this PR @juice49 ? :)

juice49: @RitaDias I wasn't planning to add it here—just so we can draw a line under this branch. However, I do still think it'd be worthwhile. It might even overlap some of your work?

RitaDias: @juice49 Yeah the reason why I'm asking is because it would already overlap with my work since I have that on my branch.
That being said if you need it for this branch you can just go ahead and I can adjust my branch?...

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
juice49: Thanks! Made that change.

PR #7255, opened 2024-07-26
feat(core): fix infinite loading in releases and create addon dataset if it doesn't exist

pedrobonamin:
### Description

This PR introduces two changes:
- Fixes an issue in which it was not possible to open releases in workspaces with no bundles previously created
- Adds the ability to create the addon dataset, when creating a bundle, if it doesn't exist. 

https://www.loom.com/share/f8b6bf115a8e40ad89e82feee3e1635d?sid=8714f116-7be3-4b91-a26d-3a9ae51ba082


jordanl17:
I think the changes to the `createBundleMetadataAggregator` means that the data isn't actual being exposed from the store anymore. I've made a few minor adjustments for emissions during the loading phase here - https://github.com/sanity-io/sanity/pull/7259

bjoerge (approved):
Very nice!

PR #7253, opened 2024-07-26
feat(preview): add support for live document id sets

bjoerge:
### Description
Adds a new (unstable) method to the preview store that supports subscribing to a set of document ids that matches a given groq-filter. The set of document ids are returned in ascending order and will update in real-time as documents appear or disappear from the set.
A query based on the given filter is performed once, initially, and thereafter the set of ids are patched based on the listener events `appear` and `disappear` transition.

This provides a lightweight way of subscribing to a list of ids for simple cases where you just want to subscribe to a set of documents ids that matches a particular filter.

### What to review
- Implementation and accompanying hooks

### Testing
TBD

### Notes for release

n/a - internal (for now)

PR #7257, opened 2024-07-26
refactor(releases): optimize data fetching

bjoerge:
Note: this PR should be rebase-merged since the individual commits makes sense by themselves, and the commit that enables full-document-previews should ~likely~ be removed before merging into main

### Description

This includes various optimizations to how data are loaded and subscribed to. It's using the "live id set" feature that got added in #7253 

### What to review
- I've split the validation logic into a separate hook
- Does the changes make sense?
- There's currently a bit of extra work being done with e.g. validation status which comes side-by-side with each document, but is later detached into a separate structure on the side which is later combined in with each document, but this is something we can address separately.
- Ideally revision history for each document should follow the same flow as validation (but we can address this later)

### Testing
Existing tests should be enough

### Notes for release
n/a


packages/sanity/src/core/preview/documentPreviewStore.ts
```diff
@@ -114,7 +114,7 @@ export interface DocumentPreviewStoreOptions {
  * This comes with an extra memory and initial transfer cost, but gives faster updating previews and less likelihood of displaying
  * out-of-date previews as documents will be kept in sync by applying mendoza patches, instead of re-fetching preview queries
  * */
-const PREVIEW_FETCH_FULL_DOCUMENTS = false
+const PREVIEW_FETCH_FULL_DOCUMENTS = true
```
pedrobonamin, line 117: According to your comment 
>and the commit that enables full-document-previews should likely be removed before merging into main

wdyt it we add here a TODO

```suggestion
// TODO: Remove before merging to main, see https://github.com/sanity-io/sanity/pull/7257/
const PREVIEW_FETCH_FULL_DOCUMENTS = true
```

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
```diff
@@ -0,0 +1,51 @@
...
+import {filter, map} from 'rxjs/operators'
+import {mergeMapArray} from 'rxjs-mergemap-array'
+
+import {useSchema} from '../../../hooks'
+import {useDocumentPreviewStore} from '../../../store'
+import {useSource} from '../../../studio'
+import {validateDocumentWithReferences} from '../../../validation'
+
+export function useBundleDocuments(bundle: string) {
+  const groqFilter = `defined(_version) &&  _id in path("${bundle}.*")`
```
pedrobonamin, line 14: What do you think about requesting to content lake another function that works in a similar way to `(sanity::versionOf("${publishedId}")` but that returns all the bundle documents?

pedrobonamin (commented):
Thanks for this changes and fixing the loading all documents validation when a new document is added. 
It looks good to me, added some comments.
 
Some unit tests are failing, I think we will need to update the `ReleaseReview` test, given it was mocking `useBundleDocumentsValidation`
And I think we could remove `useBundleDocumentsValidation` file.

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -35,92 +31,30 @@ const LISTEN_OPTIONS: ListenOptions = {
...
-  }, [handleListenerEvent, listener$, initialFetch])
-  return {
-    data: state,
-    error,
-  }
+  const documentPreviewStore = useDocumentPreviewStore()
+
+  const observable = useMemo(() => {
+    return documentPreviewStore
+      .unstable_observeDocumentIdSet(`sanity::versionOf("${publishedId}") && defined(_version)`)
```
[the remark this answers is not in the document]
bjoerge: Good point! I would be very surprised if `sanity::versionOf` returns documents without `_version`, and if it does we want to know about it. Removing this one

packages/sanity/src/core/preview/documentPreviewStore.ts
bjoerge: Good point!

bjoerge:
> Some unit tests are failing

Thanks, should be passing now

> And I think we could remove useBundleDocumentsValidation file.

Done

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
pedrobonamin: Ohhh it's using path because it should fetch here the documents that belong to a bundle.
example:
bundle: summer-drop
it needs to fetch: `summer-drop.foo` and `summer-drop.bar` 

`sanity:versionOf` get's only the versions of a document.
So if we use:
`sanity:versionOf(foo)` it will return `[{_id: foo}, {_id: summer-drop.foo}]`

bjoerge: ah, right! reverted in [757d430](https://github.com/sanity-io/sanity/pull/7257/commits/757d4304c69ed296142e5b51815efb12ac9c8206)

pedrobonamin (approved):

PR #7258, opened 2024-07-26
feat: embellish nextjs blog starter

SimeonGriggs:
### Description

* Adds icons to the `blog` template schema types
* Installs `@sanity/icons` as a dependency when using the `blog` template
* Uses `defineType/Field` functions in both JS and TS schema type files of the `blog` template
* Adds default structure builder configuration to `blog` and `clean` templates
* Sets `blog` template as the default when `init` is run inside a Next.js app

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
+++ b/packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
@@ -105,10 +105,25 @@ const schemaJS = `export const schema = {
 }
 `
 
-const structureTS = `import type {StructureResolver} from 'sanity/structure'
+const blogStructureTS = `import type {StructureResolver} from 'sanity/structure'
 
 // https://www.sanity.io/docs/structure-builder-cheat-sheet
 export const structure: StructureResolver = (S) =>
+  S.list()
+    .title('Blog')
+    .items([
+      S.documentTypeListItem('post').title('Posts'),
+      S.documentTypeListItem('category').title('Categories'),
+      S.documentTypeListItem('author').title('Authors'),
+      S.divider(),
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
```diff
@@ -12,17 +12,18 @@ import {structureTool} from 'sanity/structure'
 
 // Go to https://www.sanity.io/docs/api-versioning to learn how API versioning works
 import {apiVersion, dataset, projectId} from ${hasSrcFolder ? "'./src/sanity/env'" : "'./sanity/env'"}
-import {schema} from ${hasSrcFolder ? "'./src/sanity/schema'" : "'./sanity/schema'"}
+import {schema} from ${hasSrcFolder ? "'./src/sanity/schemaTypes'" : "'./sanity/schemaTypes'"}
+import {structure} from ${hasSrcFolder ? "'./src/sanity/structure'" : "'./sanity/structure'"}
```
binoy14, line 16: Small note 

`import {structure} from './sanity/structure'` the import looks very similar `sanity/structure` would be it be better to name the file `structureLayout` or something along the lines of that.

binoy14 (approved):
Looks good, thanks!

Left a small non-blocking comment but I'll leave the decision to change it or not up to you

packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
SimeonGriggs: I get that, but also like to name my files and folders the name of the thing it's being imported into, so it's a bit unavoidable. Going to leave for now but thanks!

PR #7259, opened 2024-07-26
fix(releases): fixing issue loading metadata and controlling aggregator loading state

jordanl17:

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -27,6 +27,8 @@ export function ReleasesOverview() {
   const bundleSlugs = useMemo(() => bundles?.map((bundle) => bundle.slug) || [], [bundles])
   const {data: bundlesMetadata, loading: loadingBundlesMetadata} = useBundlesMetadata(bundleSlugs)
   const loading = loadingBundles || loadingBundlesMetadata
+  const loadingTableData = loading || (!bundlesMetadata && Boolean(bundleSlugs.length))
+
   const hasBundles = bundles && containsBundles(bundles)
   const loadingOrHasBundles = loading || hasBundles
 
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -174,17 +176,19 @@ export function ReleasesOverview() {
...
-              key={bundleGroupMode}
-              loading={loading}
-              data={groupedBundles[bundleGroupMode]}
-              columnDefs={releasesOverviewColumnDefs}
-              searchFilter={applySearchTermToBundles}
-              emptyState="No Releases"
-              rowId="_id"
-              rowActions={renderRowActions}
-            />
+            {(hasBundles || loadingTableData) && (
```
jordanl17, line 179: Table isn't to be shown when no releases are available

packages/sanity/src/core/store/bundles/createBundlesStore.ts
```diff
@@ -56,13 +56,8 @@ const INITIAL_STATE: bundlesReducerState = {
 }
 
 const NOOP_BUNDLES_STORE: BundlesStore = {
-  state$: EMPTY.pipe(
```
jordanl17, line 59: @pedrobonamin did you make this change to resolve the infinite loader? I reverted to what is in COREL, only because with `state: 'loading'` then initially the `ReleaseOverview` screen shows the text for when no releases have been created - since this wrongly suggests that the store is not loading the bundles.

I see `NOOP_BUNDLES_STORE` is used for when the addon client is loading or dataset is being created. In these cases, I feel it does make sense to suggest that the bundles are loading. What do you think? Happy to revert this (especially if this was part of the infinite loop issue), but would prefer if we could avoid the temporary screen flicker

pedrobonamin: Thanks for this suggestions @jordanl17 !

This is necessary to render the loaded state when the addon dataset doesn't exist. 
I added some changes and pushed them to this branch, let me know what you think and happy to merge all together.  https://github.com/sanity-io/sanity/pull/7259/commits/8b3d65f9d1863ef3ba22e22e5f465e6dcd351181
I just realized we have a `ready` property in the [`AddonDatasetProvider`](https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/core/studio/addonDataset/AddonDatasetProvider.tsx#L22-L23) that we can use to fetch if the addon dataset has been fetched or not.

jordanl17: Nice discovery - yeh this all looks good to me!

PR #7262, opened 2024-07-29
fix(core): unify items to render in BundlesMenu

pedrobonamin:
### Description

The bundles menu component was showing an unnecessary divider when the bundles are archived.
Archived bundles are not rendered, but they were not considered for the `hasBundles` check.

By moving the list that is generated to the body of the component and reusing that to check the bundles length this issue is fixed.

#### Before
Screenshot 2024-07-29 at 09 39 47

#### After
Screenshot 2024-07-29 at 09 45 19

### Testing

Tests are updated.


juice49 (approved):
Nice improvement 💯.

jordanl17 (approved):
Nice improvement!

PR #7269, opened 2024-07-29
fix(core): early return if paste target is a file/image

ricokahler:
### Description

This PR fixes a bug where the new copy/paste paste-handler would prevent the paste handler in image and file inputs from firing.

This PR introduces a data attribute that is used by the copy/paste handler in order to early return and not call `event.preventDefault`.

### What to review

Review the approach and let me know if you can think of a better mechanism.

### Testing

Tested manually in the test studio. This needs a better automated test in the future.

### Notes for release

- Fixes a bug that prevents keyboard shortcuts from pasting images and files.


cngonzalez (approved):
Did some manual testing and all seemed well (but agree on tests, we could probably get away with a component test?)

PR #7277, opened 2024-07-30
fix(deps): bump `@sanity/pkg-utils` to `v6.10.7`

stipsan:
### Description

Bumps `@sanity/pkg-utils` from `v6.9` to `v6.10`.
Doing so allows us to bump `typescript` from `v5.4` to `v5.5` as well, allowing us to get rid of a `@ts-expect-error` line.
In `v6.10` the `extract.bundledPackages` API in `@sanity/pkg-utils` are automatically populated by dependencies that are in `devDependencies`, and not in `peerDependencies`.
In most cases that's what you want, as it ensures the generated typings play the same rules as the default bundling. It removes the ability to have accidental implicit dependencies that only apply to typings, while the compiled JS is bundled/inlined.

### What to review

Since the new `@sanity/pkg-utils` has a different default setting for `extract.bundledPackages` there is a small risk for unintended changes to the generated typings we publish to npm.
Therefore I manually reviewed the generated output for every package, by:
1. Immediately after brancing off `next` running a `pnpm release:current-branch`
2. It released as `sanity@3.52.3-update-pkg-utils.7`.
3. After bumping all deps, getting the build working and making all the changes, I ran `pnpm release:current-branch` again.
4. With a clean reference on npm it was easy to review what the published changes are, by running
    ```
    npm diff --diff=@sanity/block-tools@3.52.3-update-pkg-utils.7 --diff=@sanity/block-tools@3.52.3-update-pkg-utils.17 > block-tools.patch
    npm diff --diff=@sanity/cli@3.52.3-update-pkg-utils.7 --diff=@sanity/cli@3.52.3-update-pkg-utils.17 > cli.patch
    npm diff --diff=@sanity/codegen @3.52.3-update-pkg-utils.7 --diff=@sanity/codegen @3.52.3-update-pkg-utils.17 > codegen.patch
    npm diff --diff=@sanity/diff@3.52.3-update-pkg-utils.7 --diff=@sanity/diff@3.52.3-update-pkg-utils.17 > diff.patch
    npm diff --diff=@sanity/migrate@3.52.3-update-pkg-utils.7 --diff=@sanity/migrate@3.52.3-update-pkg-utils.17 > migrate.patch
    npm diff --diff=@sanity/mutator@3.52.3-update-pkg-utils.7 --diff=@sanity/mutator@3.52.3-update-pkg-utils.17 > mutator.patch
    npm diff --diff=@sanity/schema@3.52.3-update-pkg-utils.7 --diff=@sanity/schema@3.52.3-update-pkg-utils.17 > schema.patch
    npm diff --diff=@sanity/types@3.52.3-update-pkg-utils.7 --diff=@sanity/types@3.52.3-update-pkg-utils.17 > types.patch
    npm diff --diff=@sanity/util@3.52.3-update-pkg-utils.7 --diff=@sanity/util@3.52.3-update-pkg-utils.17 > util.patch
    npm diff --diff=@sanity/vision@3.52.3-update-pkg-utils.7 --diff=@sanity/vision@3.52.3-update-pkg-utils.17 > vision.patch
    npm diff --diff=create-sanity@3.52.3-update-pkg-utils.7 --diff=create-sanity@3.52.3-update-pkg-utils.17 > create-sanity.patch
    npm diff --diff=groq@3.52.3-update-pkg-utils.7 --diff=groq@3.52.3-update-pkg-utils.17 > groq.patch
    npm diff --diff=sanity@3.52.3-update-pkg-utils.7 --diff=sanity@3.52.3-update-pkg-utils.17 > sanity.patch
    ```
And then manually inspecting each `.patch` file. As far as I can tell it looks good and free from regressions :party:

### Testing

### Notes for release

Improved TypeScript definition files published to npm, especially for projects that install select packages like `@sanity/types` and `@sanity/block-tools` and not `sanity` or similar larger packages.

### DIFF
--- a/packages/@sanity/block-tools/package.json
+++ b/packages/@sanity/block-tools/package.json
@@ -49,6 +49,8 @@
     "watch": "pkg-utils watch"
   },
   "dependencies": {
+    "@sanity/types": "3.52.3",
+    "@types/react": "^18.3.3",
     "get-random-values-esm": "1.0.2",
     "lodash": "^4.17.21"
   },
@@ -56,10 +58,8 @@
     "@jest/globals": "^29.7.0",
     "@repo/package.config": "workspace:*",
     "@sanity/schema": "3.52.3",
-    "@sanity/types": "3.52.3",
     "@types/jsdom": "^20.0.0",
     "@types/lodash": "^4.17.7",
-    "@types/react": "^18.3.3",
     "@vercel/stega": "0.1.2",
     "jsdom": "^23.0.1"
   },
--- a/packages/@sanity/mutator/package.json
+++ b/packages/@sanity/mutator/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@sanity/diff-match-patch": "^3.1.1",
+    "@sanity/types": "3.52.3",
     "@sanity/uuid": "^3.0.1",
     "debug": "^4.3.4",
     "lodash": "^4.17.21"
@@ -57,7 +58,6 @@
   "devDependencies": {
     "@jest/globals": "^29.7.0",
     "@repo/package.config": "workspace:*",
-    "@sanity/types": "3.52.3",
     "@types/debug": "^4.1.5",
     "@types/lodash": "^4.17.7",
     "rimraf": "^3.0.2"
--- a/packages/@sanity/types/src/schema/definition/type/array.ts
+++ b/packages/@sanity/types/src/schema/definition/type/array.ts
@@ -10,6 +10,8 @@ import {
 } from '../schemaDefinition'
 import {type BaseSchemaDefinition, type SearchConfiguration, type TitledListValue} from './common'
 
+export type {InsertMenuOptions}
+
 /** @public */
 export interface ArrayOptions<V = unknown> extends SearchConfiguration {
   list?: TitledListValue<V>[] | V[]
--- a/dev/embedded-studio/package.json
+++ b/dev/embedded-studio/package.json
@@ -18,7 +18,7 @@
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
     "@vitejs/plugin-react": "^4.3.1",
-    "typescript": "5.5.3",
+    "typescript": "5.5.4",
     "vite": "^4.5.3"
   }
 }
--- a/dev/test-next-studio/package.json
+++ b/dev/test-next-studio/package.json
@@ -20,6 +20,6 @@
     "sanity": "workspace:*",
     "sanity-test-studio": "workspace:*",
     "styled-components": "^6.1.11",
-    "typescript": "5.5.3"
+    "typescript": "5.5.4"
   }
 }
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "@sanity/client": "^6.21.1",
     "@sanity/eslint-config-i18n": "1.0.0",
     "@sanity/eslint-config-studio": "^4.0.0",
-    "@sanity/pkg-utils": "6.9.3",
+    "@sanity/pkg-utils": "6.10.7",
     "@sanity/prettier-config": "^1.0.2",
     "@sanity/test": "0.0.1-alpha.1",
     "@sanity/tsdoc": "1.0.83",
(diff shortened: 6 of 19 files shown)


packages/@sanity/block-tools/package.json
```diff
@@ -49,17 +49,17 @@
     "watch": "pkg-utils watch"
   },
   "dependencies": {
+    "@sanity/types": "3.52.2",
```
stipsan, line 52: Having `@sanity/types` as a regular dependency is a pattern already used by `@sanity/migrate`, `@sanity/schema`, `@sanity/util`, and `sanity` itself. And is the preferred way of declaring them. Declaring them in `devDependencies` causes type asserters like `isPortableTextTextBlock` to be inlined, and ultimately duplicated in the resulting Studio bundle.

packages/@sanity/block-tools/package.json
```diff
@@ -49,17 +49,17 @@
...
+    "@types/react": "^18.3.3",
```
stipsan, line 53: This package has a dependency on `react` typings:
```ts
import {type ComponentType} from 'react'
```

The pattern established by `@sanity/types` is to declare `@types/react` in `dependencies` instead of `devDependencies`

packages/@sanity/types/src/schema/definition/type/array.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from '../schemaDefinition'
 import {type BaseSchemaDefinition, type SearchConfiguration, type TitledListValue} from './common'
 
+export type {InsertMenuOptions}
```
stipsan, line 13: By re-exporting this type it ensures that in projects that only have a dependency on `@sanity/types`, and otherwise don't have `@sanity/insert-menu`, get proper typings on `ArrayOptions['insertMenu']`: https://github.com/sanity-io/sanity/blob/163966da487072a44d0707fe055c59901b084a18/packages/%40sanity/types/src/schema/definition/type/array.ts#L26

The `@sanity/types` package currently has minimal deps, which is awesome: https://www.npmjs.com/package/@sanity/types?activeTab=dependencies

The `@sanity/insert-menu` is of course intended to be used for far more than typing, it's self contained and among other things depend on `@sanity/ui`: https://www.npmjs.com/package/@sanity/insert-menu?activeTab=dependencies

So by inlining its typings we get the best of both, self contained typings, fully compatible with the types used by `sanity`, without adding to the dependency burden of projects who only wishes to use `@sanity/types`:

```diff
diff --git a/lib/index.d.ts b/lib/index.d.ts
index v3.52.3-update-pkg-utils.7..v3.52.3-update-pkg-utils.17 100644
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,6 +1,7 @@
+/// <reference types="react" />
+
 import {ComponentType} from 'react'
 import {ElementType} from 'react'
-import {InsertMenuOptions} from '@sanity/insert-menu'
 import {ReactElement} from 'react'
 import {ReactNode} from 'react'
 import {SanityClient} from '@sanity/client'
@@ -1548,6 +1549,33 @@
   }
 }
 
+/** @alpha This API may change */
+export declare interface InsertMenuOptions {
+  /**
+   * @defaultValue `'auto'`
+   * `filter: 'auto'` automatically turns on filtering if there are more than 5
+   * schema types added to the menu.
+   */
+  filter?: 'auto' | boolean
+  groups?: Array<{
+    name: string
+    title?: string
+    of?: Array<string>
+  }>
+  /** defaultValue `true` */
+  showIcons?: boolean
+  /** @defaultValue `[{name: 'list'}]` */
+  views?: Array<
+    | {
+        name: 'list'
+      }
+    | {
+        name: 'grid'
+        previewImageUrl?: (schemaTypeName: string) => string | undefined
+      }
+  >
+}
+
 /** @internal */
 export declare type InsertPatch =
   | {

```

packages/@sanity/mutator/package.json
```diff
@@ -50,14 +50,14 @@
   },
   "dependencies": {
     "@sanity/diff-match-patch": "^3.1.1",
+    "@sanity/types": "3.52.2",
```
stipsan, line 53: Having `@sanity/types` as a regular dependency is a pattern already used by `@sanity/migrate`, `@sanity/schema`, `@sanity/util`, and `sanity` itself. And is the preferred way of declaring them.

ricokahler (approved):

PR #7280, opened 2024-07-30
feat: WIP - support version documents creation from document list

pedrobonamin:


packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -1,51 +1,60 @@
...
 interface BundleActionsProps {
   currentGlobalBundle: BundleDocument
+  /**
+   * Base document id, the published one
+   */
   documentId: string
+  /**
+   * The id of the document that is displayed in the form, might be a draft or a version
+   */
+  formDocumentId: string
```
pedrobonamin, line 20: Instead of waiting for `documentVersions` which fetches the documents, we can check the form document id, if it has the slug of the current global bundle, then the button should indicate that this document already exists.

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -1,52 +1,61 @@
...
 import {type BundleDocument} from '../../../store/bundles/types'
-import {versionDocumentExists} from '../../util/util'
+import {getBundleSlug} from '../../util/util'
 
 interface BundleActionsProps {
   currentGlobalBundle: BundleDocument
+  /**
+   * Base document id, the published one
+   */
   documentId: string
```
bjoerge, line 16: for the sake of being super clear, would it make sense to name this `publishedId`?

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
bjoerge: could it be named `versionId`?

bjoerge (commented):
couple of minor comments, feel free to ignore, otherwise lgtm :)

bjoerge (approved):

PR #7286, opened 2024-07-31
fix(core): correctly reset PTE list counts

christianhg:
### Description

I was reading the docs on [`counter-reset`](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-reset#description) when this paragraph suddenly stood out:

> After creating a counter using `counter-reset`, you can adjust its value by
> using the `counter-set` property. This is counterintuitive because, despite
> its name, the `counter-reset` property is used for creating and initializing
> counters, while the `counter-set` property is used for resetting the value
> of an existing counter.

This made me realise that we probably used `content-reset` somewhere, incorrectly, where we actually meant to use `counter-set`.

Before this change, counters on each list level would not reset at all across a PTE document. Now, the counters reset properly.

Before:

```
1. This
  a. is
    i. nested
      1. list
This is a paragraph
      2. This
    ii. is
  b. another nested
2. list
```

After:

```
1. This
  a. is
    i. nested
      1. list
This is a paragraph
      1. This
    i. is
  a. another nested
1. list
```

---

**Studio screenshots**:

Before: 

Screenshot 2024-07-31 at 10 41 04

After:

Screenshot 2024-07-31 at 10 40 37

### What to review

If you know of some large documents with lots of content and lists at different levels, then that might be helpful when reviewing if this change has any adverse effects.

### Testing

I created various lists in a document broken up by various other elements and visually verified the list counts. Manual testing it the most practical method here.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
@@ -104,7 +104,7 @@ export const EditableWrapper = styled(Card)<{$isFullscreen: boolean; $readOnly?:
 
     /* Reset the list count if the element is not a numbered list item */
     & > :not(.pt-list-item-number) {
-      counter-reset: ${TEXT_LEVELS.map((l) => createListName(l)).join(' ')};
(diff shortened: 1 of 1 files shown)


stipsan (approved):
Good catch!

PR #7291, opened 2024-08-01
fix(structure): update document status to display the version document

pedrobonamin:
🟡 This PR is dependent on the changes introduced by https://github.com/sanity-io/sanity/pull/7251 - Won't be merged until that is ready

### Description

This PR introduces a change in the document footer pane to display the version document if it's available.
So, instead of showing the "published" and "edited" values it will show the "version" values.

It also, updates the header bundle menu to correctly display the bundle badge when the document exists in the bundle.

footer and header update


RitaDias (approved):
🚀

PR #7292, opened 2024-08-01
feat(core): support pasting object into array + copying individual array items

sjelfull:
### Description

This PR fixes and adds a few things in the new copy&paste feature:
- You may now copy a supported object type into a array field
- You can now copy individual array items and append them into another array
- Fixes reference validation when pasting full documents
- Fixes flaky tests for `recentSearches` that became even more flaky after upgrading the React testing library

Fixes EDX-1432 and EDX-1584.

CleanShot 2024-08-08 at 14 54 53

https://github.com/user-attachments/assets/af63b0fa-3371-4c1d-a727-10bcf8bb6168

### What to review

- Test that this works for any sort of supported objects, like in a page builder scenario
- Test that object types that isn't supported by a target array field won't be allowed to be pasted
- Test that copying a single array item into another array that allows its type works

### Testing

Unit tests have been added for these new scenarios.

Also new in this PR, we added unit tests for the `useCopyPaste` hook that covers some common scenarios. I found that only having unit tests for the underlying transferValue utility leads to false sense of security, since there is a lot that can go wrong inside the studio. This will hopefully allow us to catch bugs early when changing this logic, even if we don't have proper E2E tests setup.

### Notes for release

- Fixes an issue where you were able to copy and paste documents that contained non-existing references
- Adds support for copying supported object types into array fields
- Adds support for copying and pasting individual array items
- Fixes an issue where `getValueAtPath()` utility wouldn't work for `0` index path segments (`['array', 0]`)


packages/sanity/src/core/studio/copyPaste/__test__/useCopyPaste.test.tsx
RitaDias, line 1: 🤯

packages/sanity/src/core/studio/copyPaste/transferValue.ts
```diff
@@ -258,24 +296,64 @@ export async function transferValue({
...
+    // and the source value is a typed object with '_type' of 'object',
+    // OR if the source value is not a typed object
+    if (
+      (isObjectSourceAndArrayOfObjectsTarget &&
+        isTypedObject(sourceValueAtPath) &&
+        sourceValueAtPath._type === 'object') ||
+      !isTypedObject(sourceValueAtPath)
+    ) {
+      // In this case, determine the object type from the source root path
+      objectType = getObjectTypeFromPath(sourceRootPath)
```
RitaDias, line 335: I'm unsure what this part is trying to target. Can you give me an example of a use case? 🙏

packages/sanity/src/core/studio/copyPaste/__test__/transferValue.test.ts
```diff
@@ -988,24 +1188,6 @@ describe('transferValue', () => {
       expect(targetValue.bio[0].children[0]._key).not.toEqual('someOtherKey')
     })
 
-    test('can copy array of numbers', async () => {
```
RitaDias, line 991: Was this test moved elsewhere? 🤔

sjelfull: I removed a duplicate - there is still a [test for this here](https://github.com/sanity-io/sanity/blob/b8d5fca8e0171b434afb1685296202b0957b3b06/packages/sanity/src/core/studio/copyPaste/__test__/transferValue.test.ts#L564)

RitaDias (dismissed):
🚀 this looks good :) Did some manual tests as well and everything works as expected!

juice49 (dismissed):
This looks great!

packages/sanity/src/core/studio/copyPaste/transferValue.ts
sjelfull: Example test case is here: https://github.com/sanity-io/sanity/blob/627b03951f260402b92a7c9a20afb67a924e6564/packages/sanity/src/core/studio/copyPaste/__test__/transferValue.test.ts#L771

RitaDias: Thank you 🙏

RitaDias (approved):

PR #7298, opened 2024-08-02
fix(cli): add `styled-components` aliasing.

ricokahler:
### Description

The following PR adds additional aliasing for styled-components for both the internal build in the monorepo and any user builds.

### What to review

Do the changes make sense? Are the updated tests adequate?

### Testing

- Unit tests were updated
- I built the `test-studio` with the new styled-component aliases and without. See below.
- I also packed this new version of sanity and tried it out in the admin studio to check the non-monorepo alises of styled-components.

https://github.com/user-attachments/assets/0d2f4275-4d44-4914-bdce-046094c5603f

### Notes for release

- Adds aliases for `styled-components` to prevent missing context errors.

### DIFF
--- a/packages/sanity/src/_internal/cli/server/__tests__/aliases.test.ts
+++ b/packages/sanity/src/_internal/cli/server/__tests__/aliases.test.ts
@@ -61,7 +61,7 @@ describe('getAliases', () => {
 
     // Prepare expected aliases
     const dirname = path.dirname(sanityPkgPath)
-    const expectedAliases = browserCompatibleSanityPackageSpecifiers.reduce<Alias[]>(
+    const resolvedSanityAliases = browserCompatibleSanityPackageSpecifiers.reduce<Alias[]>(
       (acc, next) => {
         const dest = resolve.exports(pkg, next, {
           browser: true,
@@ -77,7 +77,18 @@ describe('getAliases', () => {
       [],
     )
 
-    expect(aliases).toEqual(expectedAliases)
+    const styledComponentsAliases = [
+      {
+        find: /^styled-components\/package\.json$/,
+        replacement: expect.stringContaining('styled-components/package.json'),
+      },
+      {
+        find: /^styled-components$/,
+        replacement: expect.stringContaining('styled-components.esm.js'),
+      },
+    ]
+
+    expect(aliases).toEqual([...resolvedSanityAliases, ...styledComponentsAliases])
   })
 
   it('returns the correct aliases for the monorepo', () => {
@@ -93,13 +104,17 @@ describe('getAliases', () => {
       monorepo: {path: monorepoPath},
     })
 
-    const expectedAliases = Object.fromEntries(
+    const resolvedDevAliases = Object.fromEntries(
       Object.entries(devAliases).map(([key, modulePath]) => {
         return [key, path.resolve(monorepoPath, modulePath)]
       }),
     )
 
-    expect(aliases).toMatchObject(expectedAliases)
+    expect(aliases).toMatchObject({
+      ...resolvedDevAliases,
+      'styled-components/package.json': expect.stringContaining('styled-components/package.json'),
+      'styled-components': expect.stringContaining('styled-components.esm.js'),
+    })
   })
 
   it('returns an empty object if no conditions are met', () => {
--- a/packages/sanity/src/_internal/cli/server/aliases.ts
+++ b/packages/sanity/src/_internal/cli/server/aliases.ts
@@ -2,6 +2,7 @@ import path from 'node:path'
 
 import {escapeRegExp} from 'lodash'
 import resolve from 'resolve.exports'
+import resolveFrom from 'resolve-from'
 import {type Alias, type AliasOptions} from 'vite'
 
 import {type SanityMonorepo} from './sanityMonorepo'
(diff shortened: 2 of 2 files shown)


packages/sanity/src/_internal/cli/server/__tests__/aliases.test.ts
```diff
@@ -77,7 +77,18 @@ describe('getAliases', () => {
...
 
-    expect(aliases).toEqual(expectedAliases)
+    const styledComponentsAliases = [
+      {
+        find: /^styled-components\/package\.json$/,
+        replacement: expect.stringContaining('styled-components/package.json'),
+      },
+      {
+        find: /^styled-components$/,
+        replacement: expect.stringContaining('styled-components.esm.js'),
```
stipsan, line 87:
```suggestion
        replacement: expect.stringContaining('styled-components.browser.esm.js'),
```

packages/sanity/src/_internal/cli/server/__tests__/aliases.test.ts
```diff
@@ -93,13 +104,17 @@ describe('getAliases', () => {
...
       Object.entries(devAliases).map(([key, modulePath]) => {
         return [key, path.resolve(monorepoPath, modulePath)]
       }),
     )
 
-    expect(aliases).toMatchObject(expectedAliases)
+    expect(aliases).toMatchObject({
+      ...resolvedDevAliases,
+      'styled-components/package.json': expect.stringContaining('styled-components/package.json'),
+      'styled-components': expect.stringContaining('styled-components.esm.js'),
```
stipsan, line 116:
```suggestion
      'styled-components': expect.stringContaining('styled-components.browser.esm.js'),
```

packages/sanity/src/_internal/cli/server/aliases.ts
```diff
@@ -90,24 +105,46 @@ export function getAliases({monorepo, sanityPkgPath}: GetAliasesOptions): AliasO
...
+    // eslint-disable-next-line import/no-dynamic-require
+    const styledComponentsPkg = require(styledComponentsPkgPath)
+    // get the styled-components package root
+    const styledComponentsRoot = path.dirname(styledComponentsPkgPath)
+
+    // and then alias `styled-components` exactly to the result of the `module`
+    // entry point
+    aliases.push({
+      find: /^styled-components$/,
+      replacement: path.resolve(styledComponentsRoot, styledComponentsPkg.module),
```
stipsan, line 143:
```suggestion
      replacement: path.resolve(styledComponentsRoot, styledComponentsPkg.browser[styledComponentsPkg.module]),
```

stipsan (changes requested):
We should be using the browser build, as we're not SSR-ing the Studio today so it doesn't make sense for the heavier and slower build of `styled-components` to be used 🙌

ricokahler:
> > We should be using the browser build of `styled-components` though
> 
> I have had some doubt about using the `resolve.exports` module in general, or specifically pointing to a module path. In my mind it would be ideal to just defer this to vite and use the defaults so it matches whatever the package maintainers "intended".
> 
> Fairly certain that if you point at the module _folder_ as the alias, it still does its normal entrypoint resolution, eg psuedo-code:

Oh that's definitely the way to do it if that works. I assumed it didn't. I'll test and report back!

ricokahler:
> Fairly certain that if you point at the module _folder_ as the alias, it still does its normal entrypoint resolution

Alright I've did some research into this issue, and I've found that making the alias the module folder does not always do what we want. Here's a claude paste of my findings:

1. **Entrypoint Resolution for Aliased Modules**: When setting the module folder as the alias, Vite's behavior can vary depending on the module's structure:

   a) For modules without `exports` in their `package.json`, aliasing to the module folder can appear to follow normal resolution behavior. This is due to the Node.js module resolution algorithm and how `package.json` works.

   b) However, for modules that utilize the Node.js `exports` field in their `package.json`, Vite does not always follow normal entrypoint resolution.

2. **Example of Correct Resolution**: The `styled-components` package is an example where aliasing to the module folder works correctly. When you alias `styled-components` to its installed folder (e.g., `./node_modules/styled-components`), it picks up the `package.json` at that location and uses it to find the entry point. This gives the appearance of normal resolution.

3. **Vite's Alias Resolution Mechanism**: Vite uses `@rollup/plugin-alias` for alias resolution. This plugin performs a simple swap when it encounters an aliased path, without considering the `exports` field or subpath exports.

4. **Impact on Modules with `exports`**: This becomes problematic for modules like `sanity` that have `exports` and subpath exports (e.g., `sanity/router`, `sanity/structure`). If these exports don't resolve to files at the top level of the module folder, the aliasing breaks.

5. **Example of the Issue**: Consider a `package.json` like this:

   ```json
   {
     "name": "sanity",
     "version": "3.50.0",
     "exports": {
       ".": {
         "import": "./dist/mock-index.mjs",
         "require": "./wrong.cjs",
         "default": "./wrong.cjs"
       },
       "./router": {
         "import": "./dist/mock-router.mjs",
         "require": "./wrong.cjs",
         "default": "./wrong.cjs"
       },
       "./structure": {
         "import": "./dist/mock-structure.mjs",
         "require": "./wrong.cjs",
         "default": "./wrong.cjs"
       }
     }
   }
   ```

   In this case, aliasing `sanity` to its module folder would not correctly resolve `sanity/router` or `sanity/structure`.

6. **Rollup Alias Plugin Behavior**: The `@rollup/plugin-alias` does a straightforward replacement without considering the intricacies of the `exports` field. This means it won't handle complex module structures as we might expect.

Given these findings, we need to be more careful with how we set up aliases, especially for modules with complex export structures. The behavior can vary significantly between modules with and without `exports` in their `package.json`.

---

On another note, I actually want to throw away this implementation and use vite's [`resolve.dedupe`](https://vitejs.dev/config/shared-options#resolve-dedupe).

I'll close this PR and open another that does that.

PR #7301, opened 2024-08-02
feat(corel): virtualize content releases plugin lists

pedrobonamin:
### Description
This branch introduces virtualization for lists in the releases plugin.

- Virtualizes the releases summary and the release detail screen lists.

- Lifts the **preview values**, for the documents displayed in the release detail screen , to the `useBundleDocuments` hook. This is necessary because we are using the title generated in the preview values to filter the list. 
With the list virtualization, we need to filter the list before exposing it to the virtualizer, previously, we were not filtering the list, we were just returning null if the row title didn't match the filter criteria.

- Removes the `Row` prop from the `Table`, given it's not necessary anymore and adds extra complexity.

- Refactors the uses from the data obtained from the `useBundleDocuments` hook to avoid unnecessary transformations, like creating the Validation map.

- Removes `packages/sanity/src/core/releases/tool/detail/bundleDocumentsValidation.ts` given it's not used.

**TODO**
- [x] Remove double scroll in releases detail screen. 

### What to review

Is this correct?
Is there other way to lift the preview values which would be more convenient?

### Testing

Existing tests are updated, that should be sufficient to cover the changes introduced.


packages/sanity/src/core/releases/tool/detail/ReleaseReview.tsx
```diff
@@ -1,32 +1,72 @@
...
+        typeof previewValues.values.title === 'string' ? previewValues.values.title : fallbackTitle
+      return title.toLowerCase().includes(searchTerm.toLowerCase())
+    })
+  }, [searchTerm, documents])
+
+  const virtualizer = useVirtualizer({
+    count: filteredList.length,
+    getScrollElement: () => scrollContainerRef.current,
+    // Estimation of a document with 1 change
+    estimateSize: () => 140,
```
bjoerge, line 63: nit: might be good to assign this to a constant

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
```diff
@@ -33,19 +47,48 @@ export function useBundleDocuments(bundle: string) {
...
-        return combineLatest([document$, validation$]).pipe(
-          map(([document, validation]) => ({document, validation})),
+        const previewValues$ = document$.pipe(
+          map((document) => {
+            const schemaType = schema.get(document._type)
+            if (!schemaType) {
+              throw new Error(`Schema type not found for document type ${document._type}`)
+            }
+
+            return documentPreviewStore.observeForPreview(document, schemaType).pipe(
```
bjoerge, line 62: this is the way!

bjoerge (approved):
Great work @pedrobonamin! Left a few minor comments

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
pedrobonamin: Thanks for the suggestion!

packages/sanity/src/core/preview/useObserveDocument.ts
```diff
@@ -25,12 +24,9 @@ export function useObserveDocument<T extends SanityDocument>(
   const observable = useMemo(
     () =>
       documentPreviewStore
-        .observePaths(
-          {_id: documentId},
-          schemaType.fields.map((field) => [field.name]),
```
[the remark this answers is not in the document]
pedrobonamin: It was kept from the initial implementation, working way faster now with the new observe documents

PR #7303, opened 2024-08-02
fix(preview): remove full document preview support

bjoerge:
This was an interesting thing to explore, but it doesn't scale well, so turning this off for now (it's only been enabled in this branch, fwiw).

### DIFF
--- a/packages/sanity/src/core/preview/createGlobalListener.ts
+++ b/packages/sanity/src/core/preview/createGlobalListener.ts
@@ -21,7 +21,7 @@ export function createGlobalListener(client: SanityClient) {
         includeResult: false,
         includePreviousRevision: false,
         includeMutations: false,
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/preview/createGlobalListener.ts
```diff
@@ -21,7 +21,7 @@ export function createGlobalListener(client: SanityClient) {
         includeResult: false,
         includePreviousRevision: false,
         includeMutations: false,
-        visibility: 'transaction',
```
bjoerge, line 24: this is setting it back to how it was before

juice49 (approved):
Looks good to me, thanks! I noticed some E2E tests failed, but it looks like they passed for other browsers, so I assume this is caused by flakiness.

PR #7310, opened 2024-08-02
fix(cli): dedupe `styled-components` in vite config

ricokahler:
### Description

This PR supersedes #7298 in favor of just adding `styled-components` into the vite's `resolve.dedupe` option.

### Testing

- Tested this out in the test studio and vision no longer crashes due to two versions of styled-components.
- Also tested this out in our admin studio and it built with no issues.

### Notes for release

- Dedupes `styled-components` in our build config to prevent errors related to different context instances.


stipsan (approved):
Excellent 💖

PR #7318, opened 2024-08-05
fix(preview): fix bug with replaying of welcome event to new subscribers

bjoerge:
### Description
fixes a regression introduced earlier in this branch (49591de54dc3788ddd3f97cfa7b9a4168f37362c to be specific) - where previews would sometimes stall. This PR fixes the issue by implementing a [share](https://rxjs.dev/api/operators/share)-based operator function that re-emits the latest event matching the provided predicate.

### What to review
Does the naming make sense?

### Testing
- Added unit tests with the `shareReplayLatest` operator

### Notes for release
n/a – internal

### DIFF
--- a/packages/sanity/src/core/preview/createGlobalListener.ts
+++ b/packages/sanity/src/core/preview/createGlobalListener.ts
@@ -1,18 +1,15 @@
-import {
-  type MutationEvent,
-  type ReconnectEvent,
-  type SanityClient,
-  type WelcomeEvent,
-} from '@sanity/client'
-import {merge, timer} from 'rxjs'
-import {filter, share, shareReplay} from 'rxjs/operators'
+import {type SanityClient} from '@sanity/client'
+import {timer} from 'rxjs'
+
+import {LISTENER_RESET_DELAY} from './constants'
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/preview/createGlobalListener.ts
```diff
@@ -27,37 +24,9 @@ export function createGlobalListener(client: SanityClient) {
       },
     )
     .pipe(
-      share({
-        resetOnRefCountZero: () => timer(2000),
+      shareReplayLatest({
+        predicate: (event) => event.type === 'welcome' || event.type === 'reconnect',
```
bjoerge, line 28: would `filter` be better than `predicate` here?

pedrobonamin: I think `predicate` makes sense, given filter is a rxjs function, calling it filter here could miss lead to think it's the filter function. 

And in the `filter` function in rxjs they define the callback as `predicate`, so I consider it follows the convention. 

> [filter](https://rxjs.dev/api/index/function/filter)<T>(predicate: (value: T, index: number) => boolean, thisArg?: any)

pedrobonamin (approved):

PR #7326, opened 2024-08-06
chore(typegen): trace overloadClientMethods usage

sgulseth:
### Description

Trace overload client methods usage

### What to review

Is the interface ok for consumption in aplitude/bigquery?

### Testing

n/a

### Notes for release

n/a no notes needed

### DIFF
--- a/packages/@sanity/cli/src/actions/typegen/generate.telemetry.ts
+++ b/packages/@sanity/cli/src/actions/typegen/generate.telemetry.ts
@@ -10,6 +10,7 @@ interface TypesGeneratedTraceAttrubutes {
   unknownTypeNodesGenerated: number
   unknownTypeNodesRatio: number
   emptyUnionTypeNodesGenerated: number
+  configOverloadClientMethods: boolean
 }
 
 export const TypesGeneratedTrace = defineTrace<TypesGeneratedTraceAttrubutes>({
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/typegen/generate.telemetry.ts
```diff
@@ -10,6 +10,9 @@ interface TypesGeneratedTraceAttrubutes {
   unknownTypeNodesGenerated: number
   unknownTypeNodesRatio: number
   emptyUnionTypeNodesGenerated: number
+  config: {
+    overloadClientMethods: boolean
+  }
```
joneidejohnsen, line 15: Event properties should ideally not be nested. Is it OK to flatten the object by removing the `config` property and moving `overloadClientMethods` out?

joneidejohnsen (commented):
See comment on event properties

packages/@sanity/cli/src/actions/typegen/generate.telemetry.ts
sgulseth: Sure. I made it nested since that's how it is here https://github.com/sanity-io/sanity/blob/next/packages/%40sanity/cli/src/__telemetry__/cli.telemetry.ts#L18-L22

sgulseth: But can update

joneidejohnsen (approved):

PR #7327, opened 2024-08-06
feat(typegen): add location of discovered query

sgulseth:
### Description

Adds a location object to the discovered queries, this is practical when using codegen to parse a source file and find location of queries. Ie in vscode-sanity

### What to review

Does the start/end interface make sense?

### Testing

Added

### Notes for release

N/A - no notes needed


stipsan (approved):
Looks good, refusing to bike shed on the naming. After writing plugins in babel, ESLint, etc etc etc there isn't really a consensus anyway on exact naming 😄

PR #7330, opened 2024-08-07
feat(cli): alias to styled-components fork

ricokahler:
### Description

The following PR aliases `'styled-components'` to our [temporary fork](https://github.com/sanity-io/styled-components) of styled-components that [includes the performance benefits](https://github.com/styled-components/styled-components/pull/4332) of `useInsertionEffect`.

This PR:
- adds `@sanity/styled-components` as a dependency in the `sanity` package. this should make our fork install automatically by package managers
- updates `aliases.ts` to alias `styled-components` to `@sanity/styled-components`
- updates `buildVendorDependencies.ts` to use our styled-components fork

### What to review

Ensure my thought process for this one is sound. I think it's relatively low risk since we only support styled-components 6 now and we already do aliasing and deduping of this. I also wrapped resolving of our fork in try/catch and the default behavior should continue otherwise.

### Testing

The existing test suite overlaps and I did additional manual testing by building locally, packing the result, installing the tarball in another studio. I ran `sanity preview` to test the aliases in dev mode and then `sanity build` and `sanity build --auto-updates` (followed by `npx serve dist`). All scenarios work as intended.

### Notes for release

- Improves performance by utilizing a temporary [fork](https://www.npmjs.com/package/@sanity/styled-components ) of styled-components that [includes the performance benefits](https://github.com/styled-components/styled-components/pull/4332) of `useInsertionEffect`


packages/sanity/src/_internal/cli/server/buildVendorDependencies.ts
```diff
@@ -97,7 +97,7 @@ const VENDOR_IMPORTS: VendorImports = {
   },
   'styled-components': {
     '^6.1.0': {
-      '.': './dist/styled-components.esm.js',
+      '.': './dist/styled-components.browser.esm.js',
```
ricokahler, line 100: this one was the wrong entry point to begin with 😅

stipsan: Good catch!!

stipsan (dismissed):
LFGO ❤️🚀🚀🚀

PR #7332, opened 2024-08-07
fix(core): update corel contexts to match new lint rules

pedrobonamin:
### Description
This branch updates the new contexts introduced in the `corel` branch:
- [BundlesMetadataContext.ts](https://github.com/sanity-io/sanity/compare/corel...corel-fix-contexts?expand=1#diff-8d12c10122024f1ed64b6732075cb4d891823e5d04ecfe0f1175f73089ea9765)
- [packages/sanity/src/_singletons/context/ReleasesTableContext.ts](https://github.com/sanity-io/sanity/compare/corel...corel-fix-contexts?expand=1#diff-0517bd93149f38c3f167620ef2d8484f9c9dc03bed28e109e7a03b2f15339e15)

Following the new convention, using `createContext` from `sanity/_createContext` and assigning a relevant key to them.

It also, refactor those files to export only the context, instead of the `ContextValue` ts definition and the `useContext` hook. Given the singleton files are intended only for the context creation.

### What to review

Does the changes make sense?

### Testing


jordanl17 (approved):
This all makes sense @pedrobonamin - thanks for updating

PR #7333, opened 2024-08-07
chore(typegen): move prettier formatting into generateAction

sgulseth:
A few small changes to typegen:

 * Run prettier once per typegen execution instead of per type
 * Generate one typemap instead of one per file
 * Add integration tests for generating typemap

### What to review

Correctness

### Testing

Tests added

### Notes for release

n/a - no notes needed


stipsan (approved):
I'll rest easy knowing that there'll only be one line of `import '@sanity/client';` in my `sanity.types.ts` from now on 💖 😌

PR #7348, opened 2024-08-09
fix(sanity): prevent empty actions being executed

juice49:
### Description

This branch fixes an issue arising when Studio executes an empty array of actions using the Actions API. This usage is permitted by Content Lake, but is effectively a noop, with no transaction being executed and propagated through the system.

Although this scenario doesn't occur in Studio code, it can occur in user code. For example: if a custom document action patches a document, but the value it sets is unchanged from the current value. This can cause Studio to become stuck in a "saving" state, because it expected a transaction to be executed.

This branch fixes the issue by adding a fake `unset: ['_empty_action_guard_pseudo_field_']` patch when executing an array of actions that would otherwise be empty. This is similar to the approach employed elsewhere to emulate optimistic locking.

### What to review

Does this approach make sense?

### Testing

- Added replication to Test Studio. This can be accessed by navigating to [a document with custom document actions](http://localhost:3333/test/structure/input-debug;documentActionsTest;8655b5c5-ea4c-4d01-8874-a37d2e2c6cd0), and  executing the "Custom publish that sets someBoolean to true" document action. Prior to this fix, Studio would become stuck in a "saving" state, and the document would never be published.
- Updated unit test.

### DIFF
--- a/dev/test-studio/documentActions/actions/createCustomPublishAction.ts
+++ b/dev/test-studio/documentActions/actions/createCustomPublishAction.ts
@@ -1,7 +1,9 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 import {type DocumentActionComponent, type DocumentActionProps, useDocumentOperation} from 'sanity'
 
-export function createCustomPublishAction(originalAction: DocumentActionComponent) {
+export function createCustomPublishAction(
+  originalAction: DocumentActionComponent,
+): DocumentActionComponent {
   return function CustomPublishAction(props: DocumentActionProps) {
     const defaultPublishAction = originalAction(props)
     const documentOperations = useDocumentOperation(props.id, props.type)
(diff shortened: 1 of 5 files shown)


pedrobonamin (approved):
This looks good to me! Thanks @juice49

jordanl17:
This does all make sense, and it's really nice you've updated test-studio to reflect a true example.

The one bit that's confusing me - if actions is empty, ideally wouldn't we want to execute no transaction at all, rather than include our pseudo unset? Are we therefore not including a round network trip that is ultimately empty? Or, quite likely, have I missed that there are other operations that necessitate still sending the request off?

juice49:
@jordanl17 conceptually, yes, I think you're absolutely right. However, this work builds on the existing system for mutating documents and tracking the resulting transactions. As I understand it, it would be significantly more complicated (and error prone) to attempt to prevent mutation being sent in the first place.

jordanl17:
> @jordanl17 conceptually, yes, I think you're absolutely right. However, this work builds on the existing system for mutating documents and tracking the resulting transactions. As I understand it, it would be significantly more complicated (and error prone) to attempt to prevent the mutation being sent in the first place.
> 
> Longer term, I share your thinking.

Nice, yeh I just wanted to check if my thinking was the right track, and totally agree with this being a more pragmatic fix (time investment and code invasion wise) than preventing the request entirely.

LGTM!!

jordanl17 (approved):

PR #7349, opened 2024-08-09
feat(cli): add message about adding studioHost to cliConfig on deploys

cngonzalez:
### Description

This PR adds a note about adding studioHost to the CLI config when you deploy. I've made a few choices here:
1. The note occurs after the success message of the deploy to draw additional attention. Adding it when the appHost name is actually claimed means that the message gets lost a bit beyond the subsequent messages.

2. I've added it for ALL cases where the studioHost is not already included in CLI configs -- including legacy projects that have a "default" application. I figure this is a good way to start socializing multi-studio deployments among devs and we could perhaps even include a docs link. 

I'd love a sign-off on the language and the above two behaviors. 

https://github.com/user-attachments/assets/55653da0-59e9-4aa4-bde7-a6aafc15d029

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/deploy/deployAction.ts
+++ b/packages/sanity/src/_internal/cli/actions/deploy/deployAction.ts
@@ -34,6 +34,7 @@ export default async function deployStudioAction(
     (cliConfig && 'autoUpdates' in cliConfig && cliConfig.autoUpdates === true) ||
     false
   const installedSanityVersion = await getInstalledSanityVersion()
+  const configStudioHost = cliConfig && 'studioHost' in cliConfig && cliConfig.studioHost
 
   const client = apiClient({
     requireUser: true,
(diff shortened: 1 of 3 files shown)


binoy14 (approved):
Deferring copy to @runeb

packages/sanity/src/_internal/cli/actions/deploy/deployAction.ts
```diff
@@ -143,6 +144,14 @@ export default async function deployStudioAction(
 
     // And let the user know we're done
     output.print(`\nSuccess! Studio deployed to ${chalk.cyan(location)}`)
+
+    if (!configStudioHost) {
+      output.print(
+        `\nPlease remember to add ${chalk.cyan(`studioHost: '${userApplication.appHost}'`)}`,
+      )
+      output.print('to defineCliConfig in your sanity.cli.js or sanity.cli.ts file')
+      output.print('to re-deploy or undeploy this studio in the future.')
```
cngonzalez, line 153: @runeb should we add a link to docs here?

runeb, line 153:
```suggestion
    if (!configStudioHost) {
      output.print(
        `\nAdd ${chalk.cyan(`studioHost: '${userApplication.appHost}'`)}`,
      )
      output.print('to defineCliConfig root properties in sanity.cli.js or sanity.cli.ts')
      output.print('to avoid prompting for hostname on next deploy.')
```

cngonzalez: This might be a little misleading since this message will _also_ turn up for legacy users that DO have a "default" project, and they won't receive a prompt. I'm okay with that, but wanted to ensure you were.

PR #7356, opened 2024-08-12
feat(core): document diff review for new documents

pedrobonamin:
### Description
This PR introduces some changes to address the use case of reviewing a new document in a release.
Replaces the `new document` view for a difference showing all the fields as added with the corresponding value.
image
image

To generate this difference it:
- Calculates the diff of the element with an empty object as initial value, all new fields will be calculated as "added".
- Adds new `showFromValue` property to the `DocumentDiff` context which will be used to define if the original value needs to be shown or not, some components, like the image, show an empty image placeholder when no initial `from` value is provided.
- Updates the necessary Diff components to check for the `showFromValue` and avoid rendering empty or placeholder values.

#### Other changes
- Adds a `memoKey` property to the `BundleInDocumentResult` object, which is useful to define if some components should re render or not, this key will be recreated every time something in that result changes, so we can ensure that by checking only that key, if it's the same value, nothing has changed in the inner object, this allows us to memoize components by checking it `prev.item.memoKey === next.item.memoKey` 

- Updates [ReleaseSummary](packages/sanity/src/core/releases/tool/detail/ReleaseSummary.tsx) to use the bundleOperations hook instead of defining it's own patch action.

- Updates [TestProvider.tsx](https://github.com/sanity-io/sanity/compare/corel...corel-62?expand=1#diff-8902abb51209aa1c27ce7d6593109b4df4291d4c360eb125d89e71b60775c04b) to mock the `AddonDatasetContext` which was re rendering the components and printing console.errors making it hard to debug the existing tests. 
 Example: 

```
console.error
    Warning: An update to AddonDatasetProviderInner inside a test was not wrapped in act(...).
    When testing, code that causes React state updates should be wrapped into act(...):
    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
    
    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
        at children (/Users/pedrobonamin/code/sanity/packages/sanity/src/core/studio/addonDataset/AddonDatasetProvider.tsx:17:10)
        at AddonDatasetProvider (/Users/pedrobonamin/code/sanity/packages/sanity/src/core/studio/addonDataset/AddonDatasetProvider.tsx:139:29)
        at children (/Users/pedrobonamin/code/sanity/packages/sanity/src/core/store/_legacy/ResourceCacheProvider.tsx:18:40)
        at children (/Users/pedrobonamin/code/sanity/packages/sanity/src/core/studio/source.tsx:13:33)
        at children (/Users/pedrobonamin/code/sanity/packages/sanity/src/core/studio/workspace.tsx:13:36)
        at LayerProvider (/Users/pedrobonamin/code/sanity/node_modules/.pnpm/@sanity+ui@2.8.8_react-dom@18.3.1_react@18.3.1__react-is@18.3.1_react@18.3.1_styled-component_fbogie77evuot6l34mwfts4wai/node_modules/@sanity/ui/src/core/utils/layer/layerProvider.tsx:19:10)

```

### What to review

Is the `memoKey` approach correct?

### Testing
Related tests have been updated.

### DIFF
--- a/packages/sanity/src/core/releases/tool/detail/ReleaseSummary.tsx
+++ b/packages/sanity/src/core/releases/tool/detail/ReleaseSummary.tsx
@@ -10,7 +10,7 @@ import {RelativeTime} from '../../../components/RelativeTime'
 import {UserAvatar} from '../../../components/userAvatar/UserAvatar'
 import {Translate, useTranslation} from '../../../i18n'
 import {type BundleDocument} from '../../../store/bundles/types'
-import {useAddonDataset} from '../../../studio/addonDataset/useAddonDataset'
+import {useBundleOperations} from '../../../store/bundles/useBundleOperations'
 import {Chip} from '../../components/Chip'
 import {Table} from '../../components/Table/Table'
 import {releasesLocaleNamespace} from '../../i18n'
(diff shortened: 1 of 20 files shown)


packages/sanity/src/core/releases/tool/detail/ReleaseSummary.tsx
```diff
@@ -40,26 +40,17 @@ const setIconHue = ({hue, icon}: {hue: BundleDocument['hue']; icon: BundleDocume
...
-      if (!client) {
-        toast.push({
-          closable: true,
-          status: 'error',
-          title: 'Failed to save changes',
-          description: 'AddonDataset client not found',
-        })
-        return
-      }
-
```
RitaDias, line 59: 👋 Any reason to remove the toast? 🤔

RitaDias: Ah, I assume it's because we are using the BundleOperations now instead :)

RitaDias (approved):
🚀 looks good to me! Great work!

PR #7357, opened 2024-08-12
refactor(sanity): update version id to match `versions.[bundle].[id]`

RitaDias:
### Description

- Change `newVersion` to use `versions.[bundle].[id]` vs `[bundle].[id]`
- Change `newVersion` to remove `_version`
- Update `getPublishedId`
- Update places where `_version` was being used
- Add `getVersionId` and `isVersionId`
- Remove `isVersion` parameter from `getPublishedId` (and update each location where it was used)
- Update areas where `getVersionId` / `isVersionId` makes sense to be used
- Update tests

❗ 
- Using an alternative groq to `versionsOf` since that changes hasn't landed in CL yet

#### Dependencies
- waiting for `versionsOf` to work with version.(bundle).(id)
- has _version removal landed?
 
### What to review

- Did I miss anything? 
- Does the structure tool work as you'd expect?
- Does the linking documents in the release tool work as expected?
- Can you edit / see the right documents on the structure tool list?
- Can you add new documents to versions and have those have the right ids?
- Any use cases that are breaking and that weren't caught by the tests / manual test?

### How to test
In the preview you can check the release `New release with a punch` which should have already 2 documents with the new id structure. Feel to create new docs and add them to releases.

❗ **old documents will not appear on lists / will appear disabled / won't be part of the number of the bundles metada**

### DIFF
--- a/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
+++ b/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
@@ -2,7 +2,13 @@ import {AddIcon, CheckmarkIcon} from '@sanity/icons'
 import {useToast} from '@sanity/ui'
 import {type ReactNode, useCallback, useState} from 'react'
 import {filter, firstValueFrom} from 'rxjs'
-import {getPublishedId, useDocumentOperation, useDocumentStore, useTranslation} from 'sanity'
+import {
+  getPublishedId,
+  getVersionId,
+  useDocumentOperation,
+  useDocumentStore,
+  useTranslation,
+} from 'sanity'
 
 import {Button} from '../../../../ui-components'
 import {type BundleDocument} from '../../../store/bundles/types'
@@ -52,7 +58,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
       return
     }
 
-    const bundleId = `versions.${slug}.${documentId}`
+    const bundleId = getVersionId(documentId, slug)
 
     setCreatingVersion(true)
 
@@ -68,7 +74,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
     // only change if the version was created successfully
     await createVersionSuccess
     setIsInVersion(true)
-  }, [documentId, slug, publishedId, documentStore.pair, documentType, newVersion, toast, title])
+  }, [documentId, slug, documentStore.pair, documentType, newVersion, toast, title])
 
   /** TODO what should happen when you add a version if we don't have the ready button */
 
--- a/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
+++ b/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
@@ -52,8 +52,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
       return
     }
 
-    // TODO: Replace for getVersionId function
-    const bundleId = `${slug}.${publishedId}`
+    const bundleId = `versions.${slug}.${documentId}`
 
     setCreatingVersion(true)
 
--- a/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
+++ b/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
@@ -43,7 +43,7 @@ export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPe
 
   const observable = useMemo(() => {
     return documentPreviewStore
-      .unstable_observeDocumentIdSet(`sanity::versionOf("${publishedId}")`)
+      .unstable_observeDocumentIdSet(`_id in path("versions.**") && _id match '*.${publishedId}'`)
       .pipe(
         map(({documentIds}) => {
           return documentIds.flatMap((id) => {
--- a/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
+++ b/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
@@ -37,7 +37,7 @@ export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPe
   const {documentId} = props
 
   const {data: bundles} = useBundles()
-  const publishedId = getPublishedId(documentId, documentId.includes('.'))
+  const publishedId = getPublishedId(documentId)
 
   const documentPreviewStore = useDocumentPreviewStore()
 
--- a/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
+++ b/packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
@@ -42,24 +42,27 @@ export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPe
   const documentPreviewStore = useDocumentPreviewStore()
 
   const observable = useMemo(() => {
-    return documentPreviewStore
-      .unstable_observeDocumentIdSet(`_id in path("versions.**") && _id match '*.${publishedId}'`)
-      .pipe(
-        map(({documentIds}) => {
-          return documentIds.flatMap((id) => {
-            // eslint-disable-next-line max-nested-callbacks
-            const matchingBundle = bundles?.find((bundle) => getBundleSlug(id) === bundle.slug)
-            return matchingBundle || []
-          })
-        }),
-        map((data) => ({data})),
-        catchError((error) => {
-          return of({error})
-        }),
-        scan((state, result) => {
-          return {...state, ...result}
-        }, INITIAL_STATE),
-      )
+    return (
+      documentPreviewStore
+        // TODO - re-add versionOf once CL has it ready
+        .unstable_observeDocumentIdSet(`_id in path("versions.**") && _id match '*.${publishedId}'`)
+        .pipe(
+          map(({documentIds}) => {
+            return documentIds.flatMap((id) => {
+              // eslint-disable-next-line max-nested-callbacks
+              const matchingBundle = bundles?.find((bundle) => getBundleSlug(id) === bundle.slug)
+              return matchingBundle || []
+            })
+          }),
+          map((data) => ({data})),
+          catchError((error) => {
+            return of({error})
+          }),
+          scan((state, result) => {
+            return {...state, ...result}
+          }, INITIAL_STATE),
+        )
+    )
   }, [bundles, documentPreviewStore, publishedId])
 
   return useObservable(observable, INITIAL_STATE)
--- a/packages/sanity/src/core/bundles/util/util.test.ts
+++ b/packages/sanity/src/core/bundles/util/util.test.ts
@@ -7,11 +7,11 @@ describe('getBundleSlug', () => {
     expect(getBundleSlug('versions.summer.my-document-id')).toBe('summer')
   })
 
-  it('should return the undefined if no bundle slug is found / is published id', () => {
+  it('should return the undefined if no bundle slug is found and document is a draft', () => {
     expect(getBundleSlug('drafts.my-document-id')).toBe(undefined)
   })
 
-  it('should return the undefined if no bundle slug is found', () => {
+  it('should return the undefined if no bundle slug is found and document is published', () => {
     expect(getBundleSlug('my-document-id')).toBe(undefined)
   })
 })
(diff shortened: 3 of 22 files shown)


packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -74,26 +81,40 @@ export function getIdPair(
...
+
+  return `${VERSION_PREFIX}${bundle}${PATH_SEPARATOR}${publishedId}`
+}
+
 /** @internal */
-export function getPublishedId(id: string, isVersion?: boolean): PublishedId {
-  if (isVersion) {
-    return id.split(PATH_SEPARATOR).slice(1).join(PATH_SEPARATOR) as PublishedId
+export function getPublishedId(id: string): PublishedId {
+  if (isVersionId(id) || id.includes(PATH_SEPARATOR)) {
```
RitaDias, line 115: Remove the id.includes once ready for review

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -47,7 +47,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
       return
     }
 
-    const bundleId = `${slug}.${documentId}`
+    const bundleId = `versions.${slug}.${documentId}`
```
RitaDias, line 50: Move this to the newVersion file?

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
```diff
@@ -37,13 +37,13 @@ export function useDocumentVersions(props: DocumentPerspectiveProps): DocumentPe
...
   const {data: bundles} = useBundles()
-  const publishedId = getPublishedId(documentId, documentId.includes('.'))
+  const publishedId = getPublishedId(documentId)
 
   const documentPreviewStore = useDocumentPreviewStore()
 
   const observable = useMemo(() => {
     return documentPreviewStore
-      .unstable_observeDocumentIdSet(`sanity::versionOf("${publishedId}")`)
+      .unstable_observeDocumentIdSet(`_id in path("versions.**") && _id match '*.${publishedId}'`)
```
bjoerge, line 46: Maybe add a TODO comment here to make this proper? It's in the commit, but also nice to have it in code

packages/sanity/src/core/bundles/util/util.ts
```diff
@@ -1,17 +1,24 @@
...
+import {isVersionId} from '../../util'
 
 const PUBLISHED_SLUG = 'Published'
+const DRAFTS_SLUG = 'Drafts'
 
 /**
  * @internal
  * @hidden
  */
 export function getBundleSlug(documentId: string): string {
```
bjoerge, line 13: I think we'd rather want this to return `undefined` if we can't determine a bundle identifier, but we can address this in a follow-up pr

packages/sanity/src/core/bundles/util/util.ts
```diff
@@ -1,17 +1,24 @@
...
   if (documentId.indexOf('.') === -1) return PUBLISHED_SLUG
-  const version = documentId.slice(0, documentId.indexOf('.'))
-  return version
+
+  if (isVersionId(documentId)) {
+    const [_, bundleSlug, ...PublishedId] = documentId.split('.')
```
bjoerge, line 17:
```suggestion
    const [_, bundleSlug] = documentId.split('.')
```

packages/sanity/src/core/bundles/util/util.ts
```diff
@@ -1,17 +1,24 @@
...
+    return bundleSlug
+  }
+
+  return DRAFTS_SLUG
```
bjoerge, line 21: same as comment above - I think we should rather return undefined in this case too.

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -74,26 +83,40 @@ export function getIdPair(
...
+}
+
 /** @internal */
-export function getPublishedId(id: string, isVersion?: boolean): PublishedId {
-  if (isVersion) {
-    return id.split(PATH_SEPARATOR).slice(1).join(PATH_SEPARATOR) as PublishedId
+export function getPublishedId(id: string): PublishedId {
+  if (isVersionId(id)) {
+    // always return the last segment of the id
+    return id.split('.').pop() as PublishedId
```
bjoerge, line 119: are we assuming we will never have ids on the format `versions.foo.bar.baz`? Would it not be better if this worked the same way as for draft ids below, i.e. returning everything that comes after `versions.<bundleId>.`?

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
jordanl17, line 46: Nice combination of `path` and `match` 🚀

packages/sanity/src/core/bundles/util/util.ts
jordanl17, line 17: COULD: do you need `...PublishedId` here, it doesn't look like it's used

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -74,26 +83,40 @@ export function getIdPair(
...
 
 /** @internal */
 export function getDraftId(id: string): DraftId {
   return isDraftId(id) ? id : ((DRAFTS_PREFIX + id) as DraftId)
 }
 
+/**  @internal */
+export function getVersionId(id: string, bundle: string): string {
+  if (isVersionId(id)) {
+    const [version, bundleName, ...publishedId] = id.split('.')
```
jordanl17, line 105: COULD: improve the naming here, to clarify that `version` is 'version'
```suggestion
    const [_versionPrefix, bundleName, ...publishedId] = id.split('.')
```

packages/sanity/src/core/util/draftUtils.ts
jordanl17: According to the working contract with Content Lake, and already assumed in the temporary query we now use instead of `sanity::versionOf`, I'd agree that `PublishedId` is everything defined after the `versions.<bundleSlug>.`

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -153,13 +176,12 @@ export function collate<
...
-    _version?: Record<string, never>
   },
 >(documents: T[], {bundlePerspective}: CollateOptions = {}): CollatedHit<T>[] {
   const byId = documents.reduce((res, doc) => {
-    const isVersion = Boolean(doc._version)
-    const publishedId = getPublishedId(doc._id, isVersion)
-    const bundle = isVersion ? doc._id.split('.').at(0) : undefined
+    const publishedId = getPublishedId(doc._id)
+    const isVersion = isVersionId(doc._id)
+    const bundle = isVersion ? getBundleSlug(doc._id) : undefined
```
jordanl17, line 184: COULD: `getBundleSlug` already does a `isVersionId` check, so could we just call it, and if following Bjorge's other suggestion ([here](https://github.com/sanity-io/sanity/pull/7357#discussion_r1718495759)) could simplify this

packages/sanity/src/structure/panes/documentList/helpers.ts
```diff
@@ -24,7 +24,7 @@ export function removePublishedWithDrafts(
 ): DocumentListPaneItem[] {
   return collate(documents, {bundlePerspective}).map((entry) => {
     const doc = entry.version || entry.draft || entry.published
-    const isVersion = Boolean(doc?._version)
+    const isVersion = isVersionId(doc?._id || '')
```
jordanl17, line 27:
```suggestion
    const isVersion = doc?.id && isVersionId(doc._id)
```

jordanl17 (approved):
Nice! I think from my side mostly suggestions. Nothing that seems incorrect or problematic. Thanks

packages/sanity/src/core/bundles/hooks/useDocumentVersions.tsx
RitaDias: I can't take credit for this since I asked @bjoerge for help with it 🙏

RitaDias: My hope was that CL had released the updated versionOf before this PR was ready (it was meant to be on Thursday but they had some delays), so I had expected this code to be reverted (need to fix the commit message 🥲) 

I'll add the todo on the code 👍

packages/sanity/src/core/util/draftUtils.ts
RitaDias: ah my bad :)
I always assumed that the published Id was always just the single string that is at the end (with the rest being bundle info) especially since that is how the versionOf was first introduced initially (in a doc like `versions.bundle-summer.agot-id`), you need to send `agot-id` like versionOf(agot-id) to get all the versions associated with that id. If it had sent `bundle-summer.ago-id` I'm not sure that would have worked.

But that might have changed.
Keep in mind the temporary query was just to get it working so I could debug the code, not as a reflection of decisions made around the ids.

packages/sanity/src/core/util/draftUtils.ts
RitaDias: Great point :)

RitaDias: I'm keeping the isVersion since it's used in a condition lower down on this method and I'd prefer to keep it in a const vs just making the call directly in the if

packages/sanity/src/core/bundles/util/util.ts
RitaDias: Done here: https://github.com/sanity-io/sanity/pull/7357/commits/a362dee82bb47cef9ad3404d75000dd79759d107

packages/sanity/src/core/bundles/util/util.ts
RitaDias: Done here: https://github.com/sanity-io/sanity/pull/7357/commits/a362dee82bb47cef9ad3404d75000dd79759d107

packages/sanity/src/core/util/draftUtils.ts
RitaDias: Spoke on slack about the change, will be making it :)

RitaDias: here we go https://github.com/sanity-io/sanity/pull/7357/commits/f4350541c0ac42a651dbdeb4017cbc50826f2bdf

packages/sanity/src/core/bundles/util/util.test.ts
```diff
@@ -1,6 +1,20 @@
...
+    expect(getBundleSlug('versions.summer.my-document-id')).toBe('summer')
+  })
+
+  it('should return the undefined if no bundle slug is found / is published id', () => {
+    expect(getBundleSlug('drafts.my-document-id')).toBe(undefined)
+  })
+
+  it('should return the undefined if no bundle slug is found', () => {
+    expect(getBundleSlug('my-document-id')).toBe(undefined)
+  })
```
bjoerge, line 16: Thanks for adding these, much appreciated!

packages/sanity/src/core/bundles/util/util.ts
```diff
@@ -1,17 +1,22 @@
...
-export function getBundleSlug(documentId: string): string {
-  if (documentId.indexOf('.') === -1) return PUBLISHED_SLUG
-  const version = documentId.slice(0, documentId.indexOf('.'))
-  return version
+export function getBundleSlug(documentId: string): string | undefined {
+  if (documentId.indexOf('.') === -1) return undefined
+
+  if (isVersionId(documentId)) {
+    // eslint-disable-next-line unused-imports/no-unused-vars, @typescript-eslint/no-unused-vars
+    const [_versionPrefix, bundleSlug] = documentId.split('.')
```
bjoerge, line 15: nit, but would be able to remove the eslint disable comment by changing to
```suggestion
    const [, bundleSlug] = documentId.split('.')
```

bjoerge (approved):
Solid work, @RitaDias!

packages/sanity/src/core/bundles/util/util.ts
RitaDias: Ah I didn't know this was a thing! I'll make this change :)

PR #7360, opened 2024-08-12
refactor(preview): extract global listener, refactor preview APIs and improve typings

bjoerge:
### Description

Note: This originally happened as part of work in a feature branch, but the changes here are generic and standalone improvements that makes sense to merge to main already.

This extracts the global listener used by the preview system, improves consistency and readability and typings of of the core/preview code. These changes should not affect the behavior of previews in any significant way or cause any visible differences in how previews work currently.

This also includes a small improvement to the global preview listener: Currently this listener is subscribed once, and the subscription is kept active forever. This means that if the user navigates to a part of the Studio that does need to preview anything, the listener connection is still kept active. With the changes proposed here, the global preview listener is now refcount'ed, and shuts down when no longer needed. To avoid excessive unsubscribe+resubscribes, the listener teardown is delayed to two seconds after the last subscriber unsubscribes. Demo here:

https://github.com/user-attachments/assets/1e822027-901e-4ef6-b231-7c5a4d388c19

### What to review
Does the changes make sense?

### Testing
Existing tests should be enough. Also improves TypeScript coverage, which helps.

### Notes for release
n/a - not user facing

### DIFF
--- a/packages/sanity/src/core/preview/constants.tsx
+++ b/packages/sanity/src/core/preview/constants.tsx
@@ -8,7 +8,7 @@ export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
  * How long to wait after the last subscriber has unsubscribed before resetting the observable and disconnecting the listener
  * We want to keep the listener alive for a short while after the last subscriber has unsubscribed to avoid unnecessary reconnects
  */
-export const LISTENER_RESET_DELAY = 2000
+export const LISTENER_RESET_DELAY = 10_000
 
 export const AVAILABILITY_READABLE = {
   available: true,
--- a/packages/sanity/src/core/preview/constants.tsx
+++ b/packages/sanity/src/core/preview/constants.tsx
@@ -4,6 +4,12 @@ import {type PreviewValue} from '@sanity/types'
 export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
 export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
 
+/**
+ * How long to wait after the last subscriber has unsubscribed before resetting the observable and disconnecting the listener
+ * We want to keep the listener alive for a short while after the last subscriber has unsubscribed to avoid unnecessary reconnects
+ */
+export const LISTENER_RESET_DELAY = 2000
+
 export const AVAILABILITY_READABLE = {
   available: true,
   reason: 'READABLE',
--- a/packages/sanity/src/core/preview/createGlobalListener.ts
+++ b/packages/sanity/src/core/preview/createGlobalListener.ts
@@ -1,0 +1,31 @@
+import {type SanityClient} from '@sanity/client'
+import {timer} from 'rxjs'
+
+import {LISTENER_RESET_DELAY} from './constants'
+import {shareReplayLatest} from './utils/shareReplayLatest'
+
+/**
+ * @internal
+ * Creates a listener that will emit 'welcome' for all new subscribers immediately, and thereafter emit at every mutation event
+ */
+export function createGlobalListener(client: SanityClient) {
+  return client
+    .listen(
+      '*[!(_id in path("_.**"))]',
+      {},
+      {
+        events: ['welcome', 'mutation', 'reconnect'],
+        includeResult: false,
+        includePreviousRevision: false,
+        includeMutations: false,
+        visibility: 'query',
+        tag: 'preview.global',
+      },
+    )
+    .pipe(
+      shareReplayLatest({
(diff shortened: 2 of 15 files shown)


packages/sanity/src/core/preview/constants.tsx
```diff
@@ -4,6 +4,12 @@ import {type PreviewValue} from '@sanity/types'
 export const INCLUDE_FIELDS_QUERY = ['_id', '_rev', '_type']
 export const INCLUDE_FIELDS = [...INCLUDE_FIELDS_QUERY, '_key']
 
+/**
+ * How long to wait after the last subscriber has unsubscribed before resetting the observable and disconnecting the listener
+ * We want to keep the listener alive for a short while after the last subscriber has unsubscribed to avoid unnecessary reconnects
+ */
+export const LISTENER_RESET_DELAY = 2000
```
bjoerge, line 11: Is this a sensible value?

packages/sanity/src/core/preview/createGlobalListener.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @internal
+ * Creates a listener that will emit 'welcome' for all new subscribers immediately, and thereafter emit at every mutation event
+ */
+export function createGlobalListener(client: SanityClient) {
+  return client
+    .listen(
+      '*[!(_id in path("_.**"))]',
+      {},
+      {
+        events: ['welcome', 'mutation', 'reconnect'],
```
bjoerge, line 17: `reconnect` is filtered out downstream for now, but we might want to consider using this at a later point

packages/sanity/src/core/preview/observeFields.ts
```diff
@@ -48,58 +46,19 @@ type Cache = {
...
-      // We map these to snapshot fetch/sync. It is good to wait for the first welcome event before fetching any snapshots as, we may miss
-      // events that happens in the time period after initial fetch and before the listener is established.
-      const welcome$ = allEvents$.pipe(
-        filter((event: any) => event.type === 'welcome'),
-        shareReplay({refCount: true, bufferSize: 1}),
-      )
-
-      // This will keep the listener active forever and in turn reduce the number of initial fetches
-      // as less 'welcome' events will be emitted.
-      // @todo: see if we can delay unsubscribing or connect with some globally defined shared listener
```
bjoerge, line 84:
```
// @todo: see if we can delay unsubscribing or connect with some globally defined shared listener
```
^ This has been addressed in this PR

packages/sanity/src/core/preview/observeFields.ts
```diff
@@ -48,58 +46,19 @@ type Cache = {
...
-  observePaths: ObservePathsFn
-  versionedClient: SanityClient
+/**
+ * Creates a function that allows observing individual fields on a document.
+ * It will automatically debounce and batch requests, and maintain an in-memory cache of the latest field values
+ * @param options - Options to use when creating the observer
+ */
+export function createObserveFields(options: {
+  client: SanityClient
+  invalidationChannel: Observable<InvalidationChannelEvent>
```
bjoerge, line 56: named it `invalidationChannel` and stripped down types to make it clear that it's merely used to notify about need for refetching preview queries

packages/sanity/src/core/preview/constants.tsx
rexxars: Let's bump this up to 10 seconds - lazy imports and similar may spend longer than 2s.

bjoerge: Good point. Bumped in cb0a4d4e3c

RitaDias (approved):
This looks good to me :)

packages/sanity/src/core/preview/observeFields.ts
```diff
@@ -48,58 +46,19 @@ type Cache = {
...
 }
 
-export function create_preview_observeFields(context: {
-  observePaths: ObservePathsFn
-  versionedClient: SanityClient
+/**
+ * Creates a function that allows observing individual fields on a document.
+ * It will automatically debounce and batch requests, and maintain an in-memory cache of the latest field values
+ * @param options - Options to use when creating the observer
+ */
```
rexxars, line 53:
```suggestion
 * @param options - Options to use when creating the observer
 * @internal
 */
```

rexxars (approved):
Looks great! I'd like to start marking anything exported (from any file, regardless of whether or not it is "public") with visibility modifiers (eg `@internal`) since it makes it easier to reason about how scary it is to change - but not a blocker.

bjoerge:
Did a rebase, and looks like a new approval is required

RitaDias (approved):
🚀

PR #7365, opened 2024-08-13
fix(core): avoid unnecessary re-renders in useDocumentPresence

runeb:
### Description

When presence store receives announcements, consumers of `useDocumentPresence` receive subscription updates that relate to the `documentId` they are interested in. This patch fixes a bug where the store would continually emit `[]` to all subscribers that did not have any presence state, which due to `setState` caused them to re-render.

The first fix for this was to leverage `useReducer` in `useDocumentPresence` to only cause a re-render when the presence data actually changed, but in a second commit I have attempted to fix this in the presence store itself, making sure to only emit if there are changes. This means we will emit `[]` when first subscribed, but from then only when there is an update to presence data for the given `documentId`.

### What to review

The split between common subscription data and `documentId` specific data, the usage of `shareReplay` and the equality checks.

### Testing

I can confirm that this reduces the amount of re-renders happening in the Studio, since each `PaneItem` now only re-renders when presence data relating to their `documentId` has changed.

### Notes for release

Optimized rendering of changes in presence


packages/sanity/src/core/store/_legacy/presence/presence-store.ts
```diff
@@ -292,29 +291,35 @@ export function __tmp_wrap_presenceStore(context: {
...
+              lastActiveAt: userAndSession.session.lastActiveAt,
+              path: location.path || [],
+              sessionId: userAndSession.session.sessionId,
+              selection: location?.selection,
+            },
+          })),
+        )
+        .filter((item) => debugIntrospect || item.presence.sessionId !== SESSION_ID),
+    ),
+    shareReplay(1),
```
runeb, line 312: I suppose we can eliminate this as its introducing new behavior

PR #7366, opened 2024-08-13
fix(typegen): dont treat all document type refs as references

### LINKED ISSUE #6913: Schema extract always assumes inline referenced documents are References
I have a document containing an array field, mixing different types of reference and objects : 
```js
defineField({
      name: 'modules',
      title: 'Modules',
      type: 'array',
      group: 'editorial',
      of: [
        { type: 'reference', name: 'ref.module.fullWidthTextBlock', to: [{ type: 'module.fullWidthTextBlock' }] },
        { type: 'reference', name: 'ref.module.textWithImageBlock', to: [{ type: 'module.textWithImageBlock' }] },
        { type: 'reference', name: 'ref.module.fullWidthImageBlock', to: [{ type: 'module.fullWidthImageBlock' }] },
        // ... Other reference types here
        { type: 'module.instagramBlock' }, // Object
      ],
    }),
```

We fetch this data using a query like this : 
```ts
export const SanityProductModulesQuery = groq`
*[_type == 'product' && store.slug.current == $handle][0].modules[] {
    _type == 'module.instagramBlock' => {
        _key,
        _type,
        "title": title[$lang],
        category,
    },
    _type == 'ref.module.textWithImageBlock' => @ -> {
        _type,
        "_key": ^._key,
        "subTitle": subTitle[$lang],
        "title": title[$lang],
        "body": body[$lang],
        color,
        orientation
    },
    _type == 'ref.module.fullWidthTextBlock' => @ -> {
        _type,
        "_key": ^._key,
        "title": title[$lang],
        "body": body[$lang],
    }
}
`;
```

The Generated Types for this query is incorrect  : 
```ts
export type SanityProductModulesQueryResult = Array<
  | {
      _key: string;
      _type: 'module.instagramBlock';
      title: Array<{
        _type: 'localizedString';
        fr: string;
        en: string;
      }>;
      category: string;
    }
  | {} //We should have other types here, for referenced items
> | null;
```

The query itself work properly when tested in the vision plugin.

---

```
@sanity/cli (global)          3.45.0 (up to date)
@sanity/asset-utils            1.3.0 (up to date)
@sanity/color-input            3.1.1 (up to date)
@sanity/eslint-config-studio   3.0.1 (latest: 4.0.0)
@sanity/icons                  3.2.0 (up to date)
@sanity/language-filter        4.0.2 (up to date)
@sanity/ui                     2.3.1 (latest: 2.3.3)
@sanity/vision                3.45.0 (up to date)
sanity                        3.45.0 (up to date)
```

Mac OS 14.5 / Chrome

NPM : 9.8.1
NODE : v18.18.2

sgulseth:
### Description

It's possible to reference a document the same way as an object is referenced. However, they will not get the builtin attributes that are available on a document stored in Content Lake. The Studio treats it as a "normal" object. Since we are making some assumption around document we are inlining the whole object whenever this happens, it's an edge case and not a recommended pattern as it is mixing concerns.

### What to review

Correctness. This change also changed a snapshot in the existing test. This is because the referenced block type from the fixture is using "author" as a inline type reference, but also using author as a reference.

### Testing

Updated the current snapshots and added a specific test for inlined documents

### Notes for release

TypeGen: Fixes an edge case where referencing documents with a normal "type"-reference, and not a "reference-type" reference extracted an incorrect representation of the schema.

Fixes #6913

### DIFF
--- a/packages/@sanity/schema/src/sanity/extractSchema.ts
+++ b/packages/@sanity/schema/src/sanity/extractSchema.ts
@@ -67,6 +67,7 @@ export function extractSchema(
   extractOptions: ExtractSchemaOptions = {},
 ): SchemaType {
   const inlineFields = new Set<SanitySchemaType>()
+  const documentTypes = new Map<string, DocumentSchemaType>()
   const schema: SchemaType = []
 
   // get a list of all the types in the schema, sorted by their dependencies. This ensures that when we check for inline/reference types, we have already processed the type
@@ -82,6 +83,9 @@ export function extractSchema(
     }
     if (base.type === 'type') {
       inlineFields.add(schemaType)
+    }
+    if (base.type === 'document') {
+      documentTypes.set(typeName, base)
     }
 
     schema.push(base)
@@ -147,10 +151,6 @@ export function extractSchema(
   }
 
   function convertSchemaType(schemaType: SanitySchemaType): TypeNode {
-    if (lastType(schemaType)?.name === 'document') {
-      return createReferenceTypeNode(schemaType.name)
-    }
-
     // if we have already seen the base type, we can just reference it
     if (inlineFields.has(schemaType.type!)) {
       return {type: 'inline', name: schemaType.type!.name} satisfies InlineTypeNode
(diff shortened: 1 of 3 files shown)


rexxars (approved):
👏

PR #7367, opened 2024-08-13
fix(core): update eslintrc config to show no-restricted-imports rules

pedrobonamin:
### Description
The existing eslint config is overriding the original `no-restricted-imports` rule.
When this rule was added:
```
'no-restricted-imports': [
          'error',
          {
            paths: [
              {
                name: 'react',
                importNames: ['createContext'],
                message: 'Please place context in _singletons',
              },
            ],
          },
```

The previous rules stopped working.

- This PR introduces changes to make all the `no-restricted-imports` rules work at the same time.
- This PR also fixes the failing checks, by either importing the corresponding files or opting out where needed.

With the changes introduced, now we are seeing all the expected errors.
Screenshot 2024-08-14 at 09 55 18

### What to review

- Are this changes ok? 
- Is there any way to configure the eslint file to expose different error messages for React default import error and context error ?
- **Small caveat:** The error for "Please place context in _singletons" is not standalone anymore, it's shared with the React default error. 

### Testing

Create a .tsx file inside `packages/sanity/*` and add the following code, the errors should show in the file and when running the lint check action.
```
import {Button} from '@sanity/ui'
import React, {createContext} from 'react'

const StyledButton = styled(Button)`
  padding: 10px;
`
const context = createContext(null)

export function Test() {
  const [count, setCount] = React.useState(0)
  return <StyledButton onClick={() => setCount(count + 1)} text={`count: ${count}`} />
}
```

### DIFF
--- a/packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
+++ b/packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
@@ -1,4 +1,9 @@
-import {Button, Card, Dialog, Flex} from '@sanity/ui'
+import {
+  Card,
+  // eslint-disable-next-line no-restricted-imports
+  Dialog, // Custom dialog needed
+  Flex,
+} from '@sanity/ui'
 // eslint-disable-next-line camelcase
 import {getTheme_v2, type Theme} from '@sanity/ui/theme'
 import {toString} from '@sanity/util/paths'
@@ -16,6 +21,7 @@ import {
 } from 'sanity'
 import {css, styled} from 'styled-components'
 
+import {Button} from '../../../../../ui-components'
 import {
   buildTreeEditingState,
   type BuildTreeEditingStateProps,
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/form/studio/tree-editing/components/search/TreeEditingSearch.tsx
```diff
@@ -138,7 +139,6 @@ export function TreeEditingSearch(props: TreeEditingSearchProps): JSX.Element {
 
   return (
     <StyledPopover
-      animate
```
pedrobonamin, line 141: This is the default in the `ui-components` popover

packages/sanity/src/core/form/studio/tree-editing/components/TreeEditingDialog.tsx
```diff
@@ -219,6 +225,7 @@ export function TreeEditingDialog(props: TreeEditingDialogProps): JSX.Element |
           <Card borderTop>
             <Flex align="center" justify="flex-end" paddingX={3} paddingY={2} sizing="border">
               <Button
+                size="large"
```
pedrobonamin, line 228: The large `ui-components` button equals to the default size of the button imported from `@sanity/ui`

RitaDias (approved):
Tested and double checked with the modal and everything looks good :) 
The only thing I don't know _a whole lot_ about is the setting up of the linting but it looks good to me as well!

binoy14 (approved):
Good catch! thanks for fixing it

PR #7371, opened 2024-08-14
feat(releases, core): supporting banner display for remote deleted bundles

jordanl17:
### Description
When you have a document form open for a bundle which is deleted by another user, previously a toast would temporarily show to warn you that the version has been deleted. Now a banner will persist on the document until you navigate away- to make it clear that edits are no longer being applied to where you initially selected:
Screenshot 2024-08-16 at 12 04 56

The exact UX, eg icon/copy/tone are not yet confirmed, but are small fast follows to apply.

### What to review
* Updates to UI Component imports instead of sanity/ui in relevant places
  * And also eslint-ignore for areas where finegrained control of style is required
* Refactored `BundleMenu` to `BundlesMenu` to differentiate it from `BundleMenuButton` (`BundlesMenu` is a list of Bundles anyway)
* memoized some components to avoid rerender cycles due to parent rerenders
* memoized [`setPerspective`](https://github.com/sanity-io/sanity/pull/7371/files#diff-7ee9daeeb66d61b4d1f39ba69ffb15c15b3233c18ddf1558677c508852c4cad1) as this was changing quite frequently forcing rerenders
* Removal of `useBundleDeletedToast`
* Created a new `DeletedBundleBanner` to move this warning to a banner

### Testing
Updated tests where necessary

### Notes for release
N/A


packages/sanity/src/core/store/bundles/reducer.ts
```diff
@@ -94,7 +94,7 @@ export function bundlesReducer(
 
       // only capture the deleted bundle if deleted by another user
       const nextDeletedBundles =
-        !isDeletedByCurrentUser && localDeletedBundle
+        !!isDeletedByCurrentUser && localDeletedBundle
```
jordanl17, line 97: This was mistakenly merged in the prior change - for testing I disabled the check for only deletions by other users, but this should not be in the production usage

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
jordanl17, line 1: Most of the logic here is a copy from the existing `DocumentDeletedBanner` and `useBundleDeletedToast`. However now there is a single entry point - `DocumentDeletedBanners` (there are 2 banners that might display for a deleted document - bundle deleted, and non-bundle deleted).

A check if done to see if the bundle has been deleted, and then the relevant Banner component is rendered

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
```diff
@@ -0,0 +1,109 @@
...
+const useIsLocaleBundleDeleted = () => {
+  const {currentGlobalBundle} = usePerspective()
+  const {data: bundles, deletedBundles} = useBundles()
+  const {slug: currentGlobalBundleSlug} = currentGlobalBundle
+  const [checkedOutBundleSlug, setCheckedOutBundleSlug] = useState<string | undefined>(
+    currentGlobalBundleSlug,
+  )
+
+  useEffect(() => {
+    if (currentGlobalBundleSlug !== LATEST.slug) {
```
jordanl17, line 29: The checkedout bundle slug is the version. Latest ie. draft slug, is not considered checkedout, since it's the default and fallback, so the state value never reflects drafts. This allows for support in cases where `usePerspective` returns `drafts` quicker than `deletedBundles` updates, and in this case the remote deleted bundle would never be tracked and the banner would never be displayed

packages/sanity/src/core/bundles/components/BundlesMenu.tsx
jordanl17, line 1: Not sure why this doesn't show as just being moved - nothing has really changed here, besides the name from `BundleMenu` to `BundlesMenu`

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanner.tsx
jordanl17, line 1: Logic and render moved to `DeletedDocumentBanners.DeletedDocumentBanner`

packages/sanity/src/core/bundles/components/BundlesMenu.tsx
jordanl17, line 1: Not sure why this doesn't show as just being moved - nothing has really changed here, besides the name from `BundleMenu` to `BundlesMenu`

packages/sanity/src/core/bundles/components/dialog/BundleIconEditorPicker.tsx
```diff
@@ -75,18 +66,24 @@ export function BundleIconEditorPicker(props: {
...
-                mode="bleed"
-                onClick={handleHueChange(hue)}
-                padding={1}
-                selected={value.hue === hue}
-                data-testid={`hue-button-${hue}`}
-              >
-                <Avatar color={hue} size={0} />
-              </Button>
-            ))}
+            <TooltipDelayGroupProvider>
```
RitaDias, line 69: Forgive me if this is being explained in a different part and I missed it: why are we adding this? Is it to improve performance / are we fixing something else? 🤔

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -134,8 +134,7 @@ export function ReleasesOverview() {
         icon={AddIcon}
         disabled={isCreateBundleDialogOpen}
         onClick={() => setIsCreateBundleDialogOpen(true)}
-        padding={2}
-        space={2}
+        style={{padding: 2}}
```
RitaDias, line 137: Wouldn't `padding={2}` work? 🤔

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
RitaDias: Great point, I think that adding a similar comment on the code itself would do us good for the future so we can keep this context :)

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
RitaDias, line 1: A question that I had on this: does it make sense to add two tests related to these banners to make sure everything is working as expected (especially since the DeletedDocumentBanner.tsx was deleted) and that previous behavior continues working as it was before? :) 

Three basic tests:
- Where there is no bundle and a non version document is deleted
- Where the bundle was deleted (and so the document version is also deleted)
- Where the bundle exists but only the document version was removed

packages/sanity/src/core/bundles/components/dialog/BundleIconEditorPicker.tsx
pedrobonamin: What this DelayGroupProvider does is adds a context in which all the tooltips that belong to the same context will share the "delay" state.
Example, you hover on the first item after 300ms you see the tooltip, you then hover on the second item, the tooltip is shown immediately, so you don't have to wait 300ms on each button to see the tooltip.

RitaDias: Ahhh that's really neat!
Thank you!

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
jordanl17: Not with the ui-components button, we don't allow for `padding` to be passed through to the sanity/ui button here `packages/sanity/src/ui-components/button/Button.tsx`

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
jordanl17: Yes! Good shout! let me add some tests to the `DeletedDocumentBanners` as I missed then :(

packages/sanity/src/structure/panes/document/documentPanel/banners/DeletedDocumentBanners.tsx
jordanl17: - [ ] Comment on the esoteric logic in `DeletedDocumentBanners`

packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
```diff
@@ -80,7 +80,8 @@ export const BundleMenuButton = ({disabled, bundle, documentCount}: BundleMenuBu
             disabled={bundleMenuDisabled || isPerformingOperation}
             icon={isPerformingOperation ? Spinner : EllipsisHorizontalIcon}
             mode="bleed"
-            padding={2}
+            style={{padding: 2}}
```
pedrobonamin, line 83: Do we strictly need this? I think we could use the `size="small"` so it matches with the existing buttons.

packages/sanity/src/core/releases/components/Table/TableHeader.tsx
```diff
@@ -1,4 +1,5 @@
 import {ArrowDownIcon, ArrowUpIcon, SearchIcon} from '@sanity/icons'
+// eslint-disable-next-line no-restricted-imports -- Table header using fine-grained styling
```
pedrobonamin, line 2: I think we could use the internal <Button> here. Unless we really want to keep a different radius and spacing prop.
But the intention of the opinionated button is to normalize all the possible usages of buttons in the ui.
This is how it would look using the internal `<Button>`
Screenshot 2024-08-19 at 12 09 35

Code changes:

```

const SortHeaderButton = ({
  header,
  text,
}: Omit<ButtonProps, 'text'> &
  HeaderProps & {
    text: string
  }) => {
  const {sort, setSortColumn} = useTableContext()
  const sortIcon = sort?.direction === 'asc' ? ArrowUpIcon : ArrowDownIcon

  return (
    <Button
      iconRight={header.sorting && sort?.column === header.id ? sortIcon : undefined}
      onClick={() => setSortColumn(String(header.id))}
      mode="bleed"
      size="default"
      text={text}
    />
  )
}

```

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
pedrobonamin: I think we don't need either style or padding here, the default is `padding={2}`

packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
```diff
@@ -136,6 +126,22 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
...
+
+    return (
+      <>
+        <PermissionCheckBanner
+          granted={Boolean(permissions?.granted)}
+          requiredPermission={requiredPermission}
+        />
+        <ReferenceChangedBanner />
+        <DeprecatedDocumentTypeBanner />
+        <DeletedDocumentBanners />
```
pedrobonamin, line 140: This had a `{!isDeleting && isDeleted && (` check , we don't need that anymore?

packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -138,7 +140,7 @@ export const DocumentPanelHeader = memo(
           border
           ref={ref}
           loading={connectionState === 'connecting'}
-          title={<DocumentHeaderTitle />}
+          title={documentPaneHeaderTitle}
```
pedrobonamin, line 143: Do we need this change? I don't think it's incorrect, but I also think if we could changes to the minimum in files that are not strictly related to `corel` will make it easier to review the `corel` branch down the road.

packages/sanity/src/ui-components/button/Button.tsx
```diff
@@ -23,6 +23,7 @@ type BaseButtonProps = Pick<
   | 'tone'
   | 'type'
   | 'width'
+  | 'radius'
```
pedrobonamin, line 26: I think we should not include extra props supported in the ui-components. Can we handle this in other way from the consumer? Do we strictly need this change?

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
```diff
@@ -25,6 +26,11 @@ export function DocumentPerspectiveMenu(): JSX.Element {
     router.navigateIntent('release', {slug})
   }, [router, slug])
 
+  const bundlesMenuButton = useMemo(
+    () => <Button icon={ChevronDownIcon} mode="bleed" padding={2} space={2} />,
```
pedrobonamin, line 30: I think this is the same as the ui-components/button, it will just require a `tooltipProps={{content: "See document versions"}}` prop

pedrobonamin (changes requested):
Thanks @jordanl17  this is looking good, added some suggestions, happy to discuss them.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
jordanl17: This was to avoid passing a new component through on every render to BundlesMenu

pedrobonamin: Yeah, I meant for the import of the button from `@sanity/ui` instead from `ui-components`

packages/sanity/src/ui-components/button/Button.tsx
jordanl17: Radius is already being passed through to the UIButton. This isn't changing the behaviour more just ensuring the type is more accurate. Perhaps we should update this button to pass through only the props we `Pick` here, however I worry as to where might be knock on impact we don't intend

packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
jordanl17: This is in [here now](https://github.com/sanity-io/sanity/pull/7371/files/e621f76022fc08ac0f59c34f4d81a753563b4696#diff-23d9f184f5425ff6d768a974daa289f3b5df66ed8c6b2f2b20bd89fa109e02a0R47)

packages/sanity/src/ui-components/button/Button.tsx
pedrobonamin: It will be passed and I think it's ok but you will get a TS warning saying the prop is not supported.
If the dev want's to bypass that he could add a `ts-ignore` comment.
But we want to enforce consistency in the buttons used inside the Studio, that's why we are only allowing some props.

packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
jordanl17: - [ ] use default ui-component button and ignore padding: 2 cases

packages/sanity/src/core/releases/components/Table/TableHeader.tsx
jordanl17: - [ ] Use ui-component Button for Table headers

packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
pedrobonamin: missed it, thanks!

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveMenu.tsx
jordanl17: - [ ] Use both UI buttons in `DocumentPerspectiveMenu`

packages/sanity/src/ui-components/button/Button.tsx
jordanl17: I am of the opinion that if we are to be strict, we should be really strict. eg right now I've evaded some cumbersome restrictions by just passing a `style` prop through to `ui-component`'s button. Really, if we are opinionated on what the studio button is then, besides a small number of props, we shouldn't just extend all on the sanity/ui's Button.

However I appreciate your point regarding `radius` here, so have reverted this

packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx
bjoerge: Had the same thought. In general we should try to avoid including non-relevant changes like this, because of the noise and porential for merge conflicts it adds. Unrelated changes that are worth doing (e.g. improving readability), should preferable be done as a separate PR into main/next.

PR #7372, opened 2024-08-14
fix(core): update cache dependencies for useBundlesStore

pedrobonamin:
### Description

The bundles store was added to the resource cache with a key that was not possible to recreate, given it was using an object created in place which gets a new identity every time a component invoque `useBundlesStore`.

```
dependencies: [workspace, addonClient, {addonClientReady: ready}]

// {addonClientReady: ready} is incorrect, the object is created in place so it has a new identity each time it is created
```

Generating in some cases hundreds of requests with the query `*[_type == "bundle"]`

Screenshot 2024-08-14 at 11 20 46

By updating the dependencies to use **stable objects** this issue is fixed, and the store is created only once and the request is only triggered once.


bjoerge (approved):
Great find!

PR #7374, opened 2024-08-14
feat(releases): suffix to bundle slug

jordanl17:
### Description
There are now no error states in the bundle form. If a protected title is entered, then a unique slug will still be generated (eg title of Published, will create slug of `published-1`). If a slug already exists, the the closest unique one will be used. This allows for bundles with identical titles to be generated

### What to review
* new `useGetBundleSlug` hook that does the generation of slug from title
* Testing for `useGetBundleSlug`
* Updated tests for `BundleForm` and `BundleDetailsDialog`
* Trimming bundle title

### Testing
Relevant tests added and updated. Tests related to asserting on showing form errors have been removed

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
+++ b/packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
@@ -63,7 +63,6 @@ export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Elemen
 
           const submitValue = {...value, title: value.title?.trim()}
           await bundleOperation(submitValue)
-          setValue(submitValue)
           if (formAction === 'create') {
             setPerspective(value.slug)
           }
--- a/packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
+++ b/packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
@@ -60,8 +60,10 @@ export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Elemen
         try {
           event.preventDefault()
           setIsSubmitting(true)
-          await bundleOperation(value)
-          setValue(value)
+
+          const submitValue = {...value, title: value.title?.trim()}
+          await bundleOperation(submitValue)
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -67,8 +60,9 @@ export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Elemen
         try {
           event.preventDefault()
           setIsSubmitting(true)
-          await bundleOperation(value)
-          setValue(value)
+
+          const submitValue = {...value, title: value.title?.trim()}
```
RitaDias, line 64: 🤏 nitpick, maybe move the value / title trim() to its own const since it's being used in multiple places? :)

packages/sanity/src/core/bundles/components/dialog/BundleForm.tsx
```diff
@@ -1,28 +1,40 @@
...
-import {isDraftOrPublished} from '../../util/util'
 import {BundleIconEditorPicker, type BundleIconEditorPickerValue} from './BundleIconEditorPicker'
+import {useGetBundleSlug} from './useGetBundleSlug'
+
+interface BaseBundleDocument extends Partial<BundleDocument> {
+  hue: ColorHueKey
+  icon: IconSymbol
+}
+
+export const DEFAULT_BUNDLE: BaseBundleDocument = {
```
RitaDias, line 25: 😚 👌

RitaDias (approved):
This looks good to me, 
I think the only two things that came to mind is:
- Very small nitpick 👇 
- Not blocking, but something I considered: About the naming of the `useGetBundleSlug` and because we have a method called [getBundleSlug](https://github.com/sanity-io/sanity/blob/corel-103-2/packages/sanity/src/core/bundles/util/util.ts#L13) my brain at first associated them two. But the useGetBundleSlug is more about generating the right name for the bundle slug _before_ it even becomes "a bundle". So I think the naming can probably do with some massaging. We can have a chat tomorrow morning? :)

PR #7380, opened 2024-08-19
fix(sanity): update the title to be ellipsed when the text is too long 

RitaDias:
### Description

Before  
image  

After (hovering over the title text)
image

### What to review

- Double check the css makes sense

### Testing
Manual testing
- When no global bundle: the drop arrow should be next to the title, the title should behave as normal (if it's short or long)
- With a global bundle but no version document: the drop down arrow should be next to the title 
- With a global bundle and document version selected: the drop down arrow and badge next to it, the title should behave as normal (ellipsing if it's too long). The whole content should be in the title

### DIFF
--- a/packages/sanity/src/structure/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
+++ b/packages/sanity/src/structure/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
@@ -2,10 +2,16 @@ import {DocumentIcon} from '@sanity/icons'
 import {Flex, Text} from '@sanity/ui'
 import {createElement, memo, type ReactElement} from 'react'
 import {unstable_useValuePreview as useValuePreview, useTranslation} from 'sanity'
+import {styled} from 'styled-components'
 
 import {structureLocaleNamespace} from '../../../../i18n'
 import {useDocumentPane} from '../../useDocumentPane'
 import {DocumentPerspectiveMenu} from './perspective/DocumentPerspectiveMenu'
+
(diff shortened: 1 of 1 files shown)


jordanl17 (approved):
Nice!

packages/sanity/src/structure/panes/document/documentPanel/header/DocumentHeaderTitle.tsx
```diff
@@ -41,23 +41,22 @@ export function DocumentHeaderTitle(): ReactElement {
...
-        >
-          {value?.title || (
-            <span style={{color: 'var(--card-muted-fg-color)'}}>
-              {t('panes.document-header-title.untitled.text')}
-            </span>
-          )}
-        </Text>
-      </Flex>
-
+    <Flex flex="none" gap={3} paddingX={2} style={{flex: 1, alignItems: 'center'}}>
```
bjoerge, line 44: `flex="none"` with `style={{flex:1 …` made me look twice.

Is `flex={1} align="center"` not enough? If we need the `style`, maybe it can be assigned to a constant and hoisted to the outer scope?

RitaDias: Oh yeah this works!
For some reason I completely forgot that the Sanity UI items of course had these as props 🤦 what a brain fart, sorry about that!

Pushed a fix!

bjoerge (approved):

PR #7387, opened 2024-08-19
chore: add docgen script

bjoerge:
### Description
Adds a script to the monorepo that allows generating documents based on a template. Useful for testing.

```
Usage: tsx --env-file=.env.local ./scripts/generate-documents/cli.ts --template <template> [arguments]

     Arguments:
      --template, -t <template>: Template to use (required). Possible values: validation, book, species

      --dataset: Dataset to generate documents in (defaults to 'test')
      --amount, -n <int>: Number of documents to generate
      --draft: Generate draft documents
      --published: Generate published documents
      --size <bytes>: Size (in bytes) of the generated document (will be approximated)
      --concurrency, -c <int>: Number of concurrent requests
      --help, -h: Show this help message

  Add more templates by adding them to the "./scripts/generate-documents/templates" directory.
```

Can also be run via `pnpm run generate:docs`, e.g. `pnpm run generate:docs --template book`

Requires an environment variable in `.env.local` that points at a valid write token for the test project:

```env
# .env.local
TEST_STUDIO_WRITE_TOKEN=…
``` 

### What to review
Does it make sense?

### Testing
n/a internal dev tooling

### Notes for release

n/a


scripts/generate-documents/templates/validation.ts
```diff
@@ -0,0 +1,49 @@
...
+    },
+  ],
+  body: [
+    {
+      _type: 'block',
+      children: [
+        {
+          _type: 'span',
+          marks: [],
+          text: lorem(options.size / 2),
```
RitaDias, line 22: going to be honest the names of these functions felt a bit cryptic initially 😄 
I think we could benefit from having a small comment / a slight renaming over on [lorem-bytes.ts](https://github.com/sanity-io/sanity/pull/7387/files#diff-ea11848b0d77a555763f12acfdcda8868bf96c4b20b653ff72b2fdc8d797ca98) ?

RitaDias (approved):
Just a small comment, otherwise this is really nice :)
I was thinking while reviewing if this isn't something that would be useful to have directly on `next` though. Sure it wouldn't have the bundle things _but_ it is a pretty useful thing overall.

scripts/generate-documents/templates/validation.ts
bjoerge: what was cryptic about them? what would be a better name?

bjoerge:
> I was thinking while reviewing if this isn't something that would be useful to have directly on next though. Sure it wouldn't have the bundle things but it is a pretty useful thing overall.

Excellent point - will make it towards next.

scripts/generate-documents/templates/validation.ts
RitaDias: I think that in terms of naming it could be something related more to `getSampleText` or something along those lines (though I'm not married to the name, the name `lorem` does invoke sample / placeholder text)

However, my first gut feeling was: why do I have to pass the `options.size` into this, and what does it have to do with the sampling. Just to confirm that I understood it, this is more about making sure that the content in this PTE is not all the same text so that there is some variation between the docs generated, right?

bjoerge:
Changed the base to `next` and pushed two small commits:
- Renaming the lorem functions and adding tsdoc: https://github.com/sanity-io/sanity/pull/7387/commits/c0b4a8031f59537d6a210c7696cb789b0245b79d
- Adding `--bundle` to help text: https://github.com/sanity-io/sanity/pull/7387/commits/7802e15a88ce57869d9398638a246d41fe7aa1a7

scripts/generate-documents/templates/validation.ts
bjoerge: Renamed `lorem` to `loremString` and added some tsdocs here – hopefully that made things a bit clearer: https://github.com/sanity-io/sanity/commit/c0b4a8031f59537d6a210c7696cb789b0245b79d

RitaDias (approved):
✅

scripts/generate-documents/templates/validation.ts
bjoerge: getSampleText would be ok, but loremText feels more specific.

> Just to confirm that I understood it, this is more about making sure that the content in this PTE is not all the same text so that there is some variation between the docs generated, right?

Yep! Sometimes you might want to generate large documents, so this provides the ability to specify the desired byte size of the documents you're generating.

PR #7388, opened 2024-08-19
fix: do not show latest version if auto-updating

cngonzalez:
### Description
We do not want to show the latest version label to those on auto-updating studios. There are a couple of scenarios in which the current text could be confusing:
1. If we are doing a gradual rollout -- the users may be in a bucket that does not allow them to upgrade to the "latest" version. 
2. If developers have pinned the studio version to a past version and again, the user does not have the power to update.

### What to review
Any edge cases in which this may not be helpful.

### Testing
I would like to test this! I am trying to reduce flakiness in our test suite so we can [run the entire suite against auto-updating studios](https://github.com/sanity-io/sanity/pull/7002). Once that is done, I will write a test for this as well!


binoy14 (approved):
makes sense to me, thanks!

PR #7390, opened 2024-08-20
feat(typegen): set overload client methods to default to true

sgulseth:
### Description

Default to generate Query Typemap when generating types

### Testing

Switched the test that checks if it generates type map to make sure it doesn't generate typemap when flag is false

### Notes for release

TypeGen will now by default generate query typemap so that the Sanity Client automatically can infer the returned types of a query.


stipsan (approved):
It's time 💖

PR #7391, opened 2024-08-20
fix(typegen): find queries imported with defineQuery from next-sanity

sgulseth:
### Description

Find queries imported with defineQuery from next-sanity

### What to review

Any other packages we should detect?

### Testing

Added 👍 

### Notes for release

TypeGen: generate types defined by defineQuery from the next-sanity package


stipsan (approved):
🥳

PR #7393, opened 2024-08-20
chore: fix wrong version format in generate documents script

bjoerge:
### Description
Small follow up from #7387 - the way a document was added to a bundle was based on the old way, this fixes it by prepending `versions.` to the id instead of writing to the `_version` property

### Testing
n/a

### Notes for release
n/a – internal


RitaDias (approved):
woops 🫥

PR #7394, opened 2024-08-20
fix(core): fix copy related issues

### LINKED ISSUE #7265: Referenced document won't open in group tab.
When trying to open a referenced document inside a group tab, the document doesn't open.

**To Reproduce**

1. Clone this repo: https://github.com/DavithkbY/sanity-issue-demo-main
2. Create a document 
3. Create a 'Document referencing document'
4. Set a reference inside the 'navigations' tab and try to click on the document to open it in a pane
**Expected behavior**

The referenced document should open in a new pane.
Currently this doesn't happen and is bugged.
It only opens when going to 'All Fields'.

**Sanity Version**

3.52.1

Mac OS

20.10.0

Video showing the bug:

https://github.com/user-attachments/assets/c0304d9c-c209-45bf-887f-2fbdccad4f42

sjelfull:
### Description

This fixes a few issues related to copy paste:

Referenced could not open when you had a reference field inside a field group and was not on the All fields tab. The fix here was to remove the `onBlur` handler. I'm not entirely sure why this breaks inside a group, since I could only reproduce in a newly installed studio and not inside the monorepo.  Having a blur handler here is a footgun, and doesn't make sense given that we don't want to change the focused path until you click another field.

Fixes #7265

It was also not possible to copy selected text in a textarea, since the Copy action was triggered and copied the full field content to the clipboard. Text areas should now be ignored.

### What to review

- That copying selected text in a textarea works
- That a reference can open inside a group

### Notes for release

- Fixes an issue where references would not open if clicked on inside a Field Group
- Fixes an issue where it wasn't possible to copy a text selection inside a text area field

### DIFF
--- a/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
+++ b/packages/sanity/src/core/form/inputs/ReferenceInput/ReferenceField.tsx
@@ -17,6 +17,7 @@ import {MenuButton, MenuItem, TooltipDelayGroupProvider} from '../../../../ui-co
 import {ContextMenuButton} from '../../../components/contextMenuButton'
 import {type DocumentFieldActionNode} from '../../../config'
 import {useTranslation} from '../../../i18n'
+import {EMPTY_ARRAY} from '../../../util/empty'
 import {FormField} from '../../components'
 import {usePublishedId} from '../../contexts/DocumentIdProvider'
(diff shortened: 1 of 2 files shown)


ricokahler (approved):
nice nice!

DavithkbY:
This issue is still not fixed, see my comment on https://github.com/sanity-io/sanity/issues/7265

PR #7396, opened 2024-08-20
feat(cli): respect `--no-auto-updates flag` in CLI config

cngonzalez:
### Description
As auto-updates become the new "default" way to build and deploy studios, we should ensure we're offering ways to easily override default behavior. This PR adds a helper function used in the build and deploy functions that prioritizes CLI flags first, then `sanity.cli.{ts|js}` parameters, and then finally a default (which we can easily change).

### What to review
The new file -- does it belong in utils (it was a bit hard to place since it was shared by two different actions)?
Ways to test I may have missed? (see below)

### Testing
This function is quite simple and goes through several layers of abstraction -- I'm not sure if it's particularly useful to test on its own. It would be ideal to test the `--no-auto-updates` command line input, but I'm not sure it's possible. An integration test with the flags and a cliConfig context is also not particularly useful, because we have to mock the build action in our current tests.

In lieu of that, please find a recording of the various options and configuration.

https://github.com/user-attachments/assets/588acd10-6206-4096-91e6-e5681d498cd5

### DIFF
--- a/packages/sanity/src/_internal/cli/util/shouldAutoUpdate.ts
+++ b/packages/sanity/src/_internal/cli/util/shouldAutoUpdate.ts
@@ -1,0 +1,27 @@
+import {type CliConfig} from '@sanity/cli'
+
+import {type BuildSanityStudioCommandFlags} from '../actions/build/buildAction'
+
+interface AutoUpdateSources {
+  flags: BuildSanityStudioCommandFlags
+  cliConfig?: CliConfig
+}
+
+/**
+ * Compares parameters from various sources to determine whether or not to auto-update
+ * @param sources - The sources of the auto-update parameter, including CLI flags and the CLI config
+ * @returns boolean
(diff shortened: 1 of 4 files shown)


packages/sanity/src/_internal/cli/util/shouldAutoUpdate.ts
binoy14, line 1: Maybe you can write a unit test for this?

binoy14 (dismissed):
Changes makes sense to me, thanks! 

Left a comment about unit test which would be good to add

packages/sanity/src/_internal/cli/util/shouldAutoUpdate.ts
cngonzalez: done! they're pretty simple 😅

binoy14 (approved):
LGTM! Thanks!

PR #7397, opened 2024-08-20
feat(preview): add experimental support for observing full documents

bjoerge:
### Description
In certain cases its useful to be able to subscribe to the complete version of a document (by id) and receive updates to it as it changes in the backend. The only way to achieve this in the Studio today involves either setting up a full optimistic mutation pipeline, or setting up a "live query", both of which adds a separate listener connection per document. This doesn't scale well, so this PR proposes an alternative approach that re-uses the global listener currently use for Studio previews.

The intended use is cases where you always need  the latest version of a complete document, but don't need to be able to do optimistic updates to it (i.e. you can live with a server roundtrip).

It introduces two new method to the preview store API, both unstable and marked as internal for now:
- `unstable_observeDocument(id: string)` - returns an observable that will emit new versions of the document as it changes remotely.
- `unstable_observeDocumentIds(ids: string[])` - thin convenience wrapper over the above that supports subscribing to multiple documents in one go

### What to review

The global query listener is now set up with `effectFormat=mendoza` which makes the mutation events include a (usually compact) patch that is used to patch the local version of the document as it receives remote changes. This will add a little bit of overhead and it might be worth considering adding this only if needed. That said, we did receive regular mutations over this listener up until recently, which tends to be much larger in size than mendoza patches.

### Testing
These methods are only used in a feature branch for now, we'll look into some automated tests for them as we start relying on them more.

### Notes for release
n/a - internal for now

### DIFF
--- a/packages/sanity/src/core/preview/__test__/createObserveDocument.test.ts
+++ b/packages/sanity/src/core/preview/__test__/createObserveDocument.test.ts
@@ -1,0 +1,113 @@
+import {describe, expect, it, jest} from '@jest/globals'
+import {createClient, type WelcomeEvent} from '@sanity/client'
+import {firstValueFrom, of, skip, Subject} from 'rxjs'
+import {take} from 'rxjs/operators'
+
+import {createObserveDocument, type ListenerMutationEventLike} from '../createObserveDocument'
+
+describe(createObserveDocument.name, () => {
+  it('fetches the current version of the document when receiving welcome event', async () => {
+    const mockedClient = createClient({
+      projectId: 'abc',
+      dataset: 'production',
+      apiVersion: '2024-08-27',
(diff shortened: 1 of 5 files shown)


rexxars (commented):
Code looks good, but should have tests unless it is only going to be used for PoC-type experiments

bjoerge:
> should have tests unless it is only going to be used for PoC-type experiments

Good callout :)

Added tests in 2897248f39c7c5399075565e8577b2fcc9982263

pedrobonamin (approved):
Looks good to me, we have been using it in `corel` and it works great

Thanks for doing this

pedrobonamin:
Updated the tests to Vitest

pedrobonamin (approved):

PR #7398, opened 2024-08-20
feat(preview): add experimental support for live document id sets

bjoerge:
### Description
Note: builds upon the changes in #7397 

This adds a new method `unstable_observeDocumentIdSet(groqFilter, params)`  to the preview store that supports subscribing to a _set_ of _document ids_ that matches a provided groq filter. This method returns an observable that emits a _new_ set of ids everytime a document is added or removed from the set of matching documents.

Under the hood it sets up a listener for the provided filter, upon receiving the `welcome` event it fetches the initial set of ids of matching documents,  and thereafter it patches the set based on `appear` and `disappear` transition on the listener event.

Since all we have here are document ids, and we don't now the position newly added ids, ordering can only be done based on the id itself, so this API does not include a way to specify an order with the query/filter itself.

This PR also includes a hook `useLiveDocumentSet(filter)` (also internal for now), that maps every unique id in the set to a subscription for the complete document (using `previewStore.observeDocument(id)` implmeneted in #7397) . This hook is using [rxjs-mergemap-array](https://www.npmjs.com/package/rxjs-mergemap-array) which makes sure that subscriptions to individual documents are made only once, even if the set itself changes, and also makes sure to unsubscribe documents that disappear from the set.

### What to review
Is the naming clear? I opted for "set" nomenclature to hint towards the "unordered" nature of the returned array (it _is_ sorted by default, but consumers can also decide to add new items at the start or at the end).

### Testing
These methods are only used in a feature branch for now, we'll look into some automated tests for them as we start relying on them more.

### Notes for release
n/a – internal for now

### DIFF
--- a/packages/sanity/src/core/preview/documentPreviewStore.ts
+++ b/packages/sanity/src/core/preview/documentPreviewStore.ts
@@ -1,4 +1,9 @@
-import {type MutationEvent, type SanityClient, type WelcomeEvent} from '@sanity/client'
+import {
+  type MutationEvent,
+  type QueryParams,
+  type SanityClient,
+  type WelcomeEvent,
+} from '@sanity/client'
 import {type PrepareViewOptions, type SanityDocument} from '@sanity/types'
 import {combineLatest, type Observable} from 'rxjs'
 import {distinctUntilChanged, filter, map} from 'rxjs/operators'
@@ -10,6 +15,7 @@ import {createObserveDocument} from './createObserveDocument'
 import {createPathObserver} from './createPathObserver'
 import {createPreviewObserver} from './createPreviewObserver'
 import {createObservePathsDocumentPair} from './documentPair'
+import {createDocumentIdSetObserver, type DocumentIdSetObserverState} from './liveDocumentIdSet'
 import {createObserveFields} from './observeFields'
 import {
   type ApiConfig,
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/preview/documentPreviewStore.ts
```diff
@@ -107,6 +137,10 @@ export function createDocumentPreviewStore({
     )
   }
 
+  const observeDocumentIdSet = createDocumentIdSetObserver(
+    versionedClient.withConfig({apiVersion: '2024-07-22'}),
```
pedrobonamin, line 141: Why do we need this specific version and we are not using the [default studio version](https://github.com/sanity-io/sanity/blob/preview/add-live-document-id-set/packages/sanity/src/core/studioClient.ts#L12-L13)?

pedrobonamin (approved):

packages/sanity/src/core/preview/documentPreviewStore.ts
bjoerge: Incidentally the version at the time this was written. We can probably use the default version here

bjoerge: removed version specifier in df13a0468665dfad3b15646f31f9a5313f9af51e

pedrobonamin (approved):

PR #7401, opened 2024-08-20
feat(cli): use auto-updates flag in init

cngonzalez:
### Description
As we move to make auto-updates the default way of updating studios, we want to ensure that we give developers the option to maintain legacy behavior and keep to specific versions if they prefer.

This PR updates the `sanity init` command to allow a `--no-auto-updates` flag that will then be injected into the CLI configuration as `autoUpdates: true` or `autoUpdates: false`

This required a bit of refactoring of our template logic, since we previously only dealt with string variables. I've added another way of identifying the variables we're injecting, but happy to refactor if this is not ideal.

### What to review
Are there any ways in which auto-updates might be undefined? Any gotchas in the AST parser?

### Testing
I've added a test for this against Typescript -- if needed, I can repeat those tests for the other cases (Javascript, templates) but assume all will be consistent. 

(No release notes because this feature is still in beta)

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
+++ b/packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
@@ -14,33 +14,55 @@ export default defineCliConfig({
    * Enable auto-updates for studios.
    * Learn more at https://www.sanity.io/docs/cli#auto-updates
    */
-  autoUpdates: true,
+  autoUpdates: __BOOL__autoUpdates__,
 })
 `
 
 export interface GenerateCliConfigOptions {
   projectId: string
   dataset: string
+  autoUpdates: boolean
 }
 
 export function createCliConfig(options: GenerateCliConfigOptions): string {
   const variables = options
   const template = defaultTemplate.trimStart()
   const ast = parse(template, {parser})
+
   traverse(ast, {
     StringLiteral: {
       enter({node}) {
         const value = node.value
-        if (!value.startsWith('%') || !value.endsWith('%')) {
-          return
+        if (value.startsWith('%') && value.endsWith('%')) {
+          const variableName = value.slice(1, -1) as keyof GenerateCliConfigOptions
+          if (!(variableName in variables)) {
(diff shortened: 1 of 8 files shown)


binoy14 (dismissed):
LGTM, one copy question

packages/@sanity/cli/src/commands/init/initCommand.ts
```diff
@@ -27,6 +27,7 @@ Options
   --coupon <name> Optionally select a coupon for a new project (cannot be used with --project-plan)
   --no-typescript Do not use TypeScript for template files
   --package-manager <name> Specify which package manager to use [allowed: ${allowedPackageManagersString}]
+  --no-auto-updates Disable import maps that keep your studio up-to-date with the latest Sanity versions
```
runeb, line 30:
```suggestion
  --no-auto-updates Disable auto updates of studio versions
```

packages/@sanity/cli/src/actions/init-project/createCliConfig.ts
```diff
@@ -14,33 +14,57 @@ export default defineCliConfig({
...
+          '$1',
+        ) as keyof GenerateCliConfigOptions
+        if (!(variableName in variables)) {
+          throw new Error(`Template variable '${variableName}' not defined`)
+        }
+        const value = variables[variableName]
+        if (typeof value !== 'boolean') {
+          throw new Error(`Expected boolean value for '${variableName}'`)
+        }
+        node.name = value.toString()
```
rexxars, line 67: I don't like this approach - we're relying on the fact that babel won't validate that the name is valid. This identifier node should be replaced with a Literal node.  I'll do a fix.

ricokahler (approved):
looks good! i believe this should result in right CLI config

not sure if this was suggested yet, but it seems like we're reimplementing a lot of what [babel template](https://babeljs.io/docs/babel-template) could solve but it may be more complicated with dependencies, typescript support, etc so feel free to ignore

PR #7403, opened 2024-08-21
Feat/studio manifests cont

snorrees:


packages/sanity/src/_internal/manifest/extractManifest.ts
```diff
@@ -0,0 +1,499 @@
...
+
+export function extractWorkspace(workspace: Workspace): ManifestWorkspace {
+  const serializedSchema = extractManifestSchemaTypes(workspace.schema)
+
+  return {
+    name: workspace.name,
+    dataset: workspace.dataset,
+    schema: serializedSchema,
+  }
+}
```
robinpyon, line 79: Would it be possible to expose workspace `title` and `subtitle` too? (Which may be more meaningful for users when selecting workspaces in _Create_)

robinpyon (commented):
Really nice! 

It always boggles my mind how much needs to be considered here (as exemplified in `transformCommonTypeFields`). The test suite is comprehensive and really helped my understanding here too – thanks!

Will defer to DX here, but some other points (in addition to the comments below):

1. I feel this should be opt-in to start, especially as we've yet to put it through its paces and there'll likely be more movement here (in addition to the point you've mentioned re: increasing build times for _all users_). Touching the default `sanity build` command seems like a pretty big deal.
2. Wondering if this should exist as a standalone command on `sanity schema` (e.g. `sanity schema build-manifest` or similar), especially as there's some inconsistency with CLI arguments between formats. For example, when `format=manifest`:
  - `workspace` isn't respected, manifest generation will always build _all_ workspaces
  - `force-required-fields` is ignored
  - `path` is treated as a directory path (where it's treated as a file path when `format=groq-type-nodes`)

snorrees:
> 1. I feel this should be opt-in to start, especially as we've yet to put it through its paces and there'll likely be more movement here (in addition to the point you've mentioned re: increasing build times for _all users_). Touching the default `sanity build` command seems like a pretty big deal.

Top of the agenda for the meeting with Rune. I also feel opt in, on a new command (Ash actually made that, but I deleted it 🙈 ).

> 2. Wondering if this should exist as a standalone command on `sanity schema` (e.g. `sanity schema build-manifest` or similar), especially as there's some inconsistency with CLI arguments between formats. For example, when `format=manifest`:
> 
> * `workspace` isn't respected, manifest generation will always build _all_ workspaces
> * `force-required-fields` is ignored
> * `path` is treated as a directory path (where it's treated as a file path when `format=groq-type-nodes`)

All good arguments in favor of standalone. Bringing it to the meeting.

packages/sanity/src/_internal/manifest/extractManifest.ts
robinpyon: Another late addition to the above – I believe we'll also need to include workspace `basePath` (in order to construct preview URLs in _Create_)

Re: custom providers – I don't see the harm in including, provided it's low effort to to so!

snorrees: We will have title, subtitle and basePath on each workspace. 
We will not have auth providers at this point, as the studio supports functions for these.

snorrees:
@robinpyon I've changed the code in the following ways:
- reverted all changed to extract schema code
- added `sanity manifest extract` command
- moved all worker code needed for the manifest to new files, so there is no accidental impact on `sanity schema extract`
- renamed files:
    - create-schema.json
    - <schema-hash>.create-schema.json
    
 Pushed a tagged release: gonna update the Create PR to support the new filenames.

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
robinpyon, line 1: I believe all these changes can be reverted, given that manifest creation is now exclusively handled via `sanity manifest extract`? (Currently, you can still build manifest via `sanity schema extract --format manifest`)

packages/sanity/src/_internal/cli/commands/index.ts
```diff
@@ -110,6 +111,7 @@ const commands: (CliCommandDefinition | CliCommandGroupDefinition)[] = [
   previewCommand,
   uninstallCommand,
   execCommand,
+  extractManifestCommand,
```
robinpyon, line 114: I believe we'll include a separate `CliCommandGroupDefinition` (e.g. extractManifestCommand) here as well

E.g.
```
// packages/sanity/src/_internal/cli/commands/manifest/extractManifestGroup.ts
import {type CliCommandGroupDefinition} from '@sanity/cli'

const extractManifestGroup: CliCommandGroupDefinition = {
  name: 'manifest',
  signature: '[COMMAND]',
  isGroupRoot: true,
  description: 'Extracts the studio configuration as one more JSON manifest files.',
}

export default extractManifestGroup
```

Without this, the manifest command won't appear in the list when running `sanity --help`, and running `sanity manifest` standalone will yield `Error: Command "manifest" is not available outside of a Sanity project context` (even if you are in a project context)

robinpyon (commented):
Thanks @snorrees!

Left some comments below – just low hanging fruit to do with how `sanity manifest` appears in CLI help and running it standalone

packages/sanity/src/_internal/cli/actions/schema/extractAction.ts
snorrees: Good catch!

snorrees:
@juice49 adding you as explicit reviewer since you know a bit about the history here. Let me know who else needs to know.

snorrees:
Any input on why all the build actions are failing is welcome 😅

packages/sanity/src/_internal/cli/commands/manifest/manifestGroup.ts
```diff
@@ -0,0 +1,6 @@
+export default {
+  name: 'manifest',
+  signature: '[COMMAND]',
+  isGroupRoot: true,
+  description: 'Interacts with the studio configuration.',
```
juice49, line 5: Is this description correct?

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
```diff
@@ -0,0 +1,152 @@
...
+export async function extractManifestSafe(
+  args: CliCommandArguments<ExtractFlags>,
+  context: CliCommandContext,
+): Promise<void> {
+  if (EXTRACT_MANIFEST_DISABLED) {
+    return
+  }
+
+  try {
+    await extractManifest(args, context)
```
juice49, line 42: Could it be a good idea to add a timeout to this operation, at least for initial iterations before we understand how it might perform for various schemas in the wild? We could set a fairly lengthy value (1 or 2 minutes, perhaps) that would at least ensure we don't block folks.

juice49 (dismissed):
This looks great, Snorre! Thank you for picking up my work. I've left a couple of small comments, and also bumped this with the Studio team to see if anybody else is able to take a look.

packages/sanity/src/_internal/cli/commands/manifest/manifestGroup.ts
snorrees: Yes, but I was a bit unsure about what to put here. Manifest sort of looks at the whole studio config and extracts part of it.

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
snorrees: Good point, was considering it but was a lazy boi 😅

packages/sanity/src/_internal/cli/commands/manifest/manifestGroup.ts
snorrees: For context `schemaGroup` reads: `Interacts with Sanity Studio schema configurations`

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
snorrees: Added a 2 minute timeout.

Also added a line about which env var to set if you run into an extract error - probably useful if folks run into the timeout. Adding 5 seconds to a built is one thing, adding 2 minutes quite another 😓

ryanbonial:
This looks good to me, but we should have someone with more knowledge about this area also take a look. @ricokahler

snorrees:
Updated the "Failure states" section to show differentiation between running the command and running deploy.

snorrees:
Known issue: word break is not handled well

image

packages/sanity/src/_internal/manifest/extractWorkspaceManifest.ts
```diff
@@ -0,0 +1,502 @@
...
+  const arrayProps =
+    typeName === 'array' && type.jsonType === 'array' ? transformArrayMember(type, context) : {}
+
+  const referenceProps = isReference(type) ? transformReference(type) : {}
+  const crossDatasetRefProps = isCrossDatasetReference(type)
+    ? transformCrossDatasetReference(type)
+    : {}
+
+  const objectFields: ObjectFields =
+    type.jsonType === 'object' && type.type && INLINE_TYPES.includes(typeName) && isCustomized(type)
```
ricokahler, line 115: do you need `INLINE_TYPES.includes` check? i think checking for `type.jsonType === 'object'` along with your `isCustomized` check should be sufficient

ricokahler (approved):
it looks good! i like the addition of the fail-safe when running this command during deploy. i think this kind of extracted data is really good to be shared in different spots like the graphql command and now the type gen command. That's something for us to discuss related to SDK though

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
```diff
@@ -0,0 +1,183 @@
...
+} from '../../../manifest/manifestTypes'
+import {type ExtractManifestWorkerData} from '../../threads/extractManifest'
+import {getTimer} from '../../util/timing'
+
+const MANIFEST_FILENAME = 'create-manifest.json'
+const SCHEMA_FILENAME_SUFFIX = '.create-schema.json'
+
+/** Escape-hatch env flags to change action behavior */
+const FEATURE_ENABLED_ENV_NAME = 'SANITY_CLI_EXTRACT_MANIFEST_ENABLED'
+const EXTRACT_MANIFEST_DISABLED = process.env[FEATURE_ENABLED_ENV_NAME] === 'false'
```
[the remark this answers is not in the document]
snorrees: Beware the inverted naming here (is it too confusing?)

EXTRACT_MANIFEST_**DISABLED**  = process.env[SANITY_CLI_EXTRACT_MANIFEST_**ENABLED**] === 'false'

So when SANITY_CLI_EXTRACT_MANIFEST_ENABLED is anything other than literal string 'false', extract is disabled.

Yeah.... Looking at it now, it IS too confusing. I think I was trying to avoid implied double negations (!!) in the code, but I really just moved that into the naming, which is worse.

Imma change this to:
```suggestion
const EXTRACT_MANIFEST_ENABLED = process.env[FEATURE_ENABLED_ENV_NAME] !== 'false'
```

Then use `!` at the callsite. I think thats much clearer, and not flipping DISABLED and ENABLED in the constant names.

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
```diff
@@ -0,0 +1,183 @@
...
+const EXTRACT_MANIFEST_LOG_ERRORS = process.env.SANITY_CLI_EXTRACT_MANIFEST_LOG_ERRORS === 'true'
+
+const CREATE_TIMER = 'create-manifest'
+
+const EXTRACT_TASK_TIMEOUT_MS = minutesToMilliseconds(2)
+
+const EXTRACT_FAILURE_MESSAGE =
+  'Unable to extract manifest. Certain features like Sanity Create will not work with this studio.\n' +
+  //TODO: replace this link
```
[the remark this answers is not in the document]
snorrees: Working out with stakeholders what we want here. Hold on :)

packages/sanity/src/_internal/manifest/extractWorkspaceManifest.ts
```diff
@@ -0,0 +1,502 @@
...
+}
+
+const transformTypeValidationRule: ValidationRuleTransformer = (rule) => {
+  return {
+    ...rule,
+    constraint:
+      'constraint' in rule &&
+      (typeof rule.constraint === 'string'
+        ? rule.constraint.toLowerCase()
+        : retainSerializableProps(rule.constraint)),
```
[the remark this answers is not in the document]
snorrees: Serialize everything that can be serialized (unless I know I know up front that I dont want it), has been my mo. Is lets keep ;)

packages/sanity/src/_internal/manifest/extractWorkspaceManifest.ts
snorrees: Object aliases are not allowed to re-define fields (it will throw). Checking for INLINE_TYPES is required, do determine if it is a object-with-fields definition.

packages/sanity/src/_internal/cli/actions/deploy/deployAction.ts
```diff
@@ -101,16 +102,25 @@ export default async function deployStudioAction(
...
 
     if (!didCompile) {
       return
     }
+
+    await extractManifestSafe(
+      {
+        ...buildArgs,
+        extOptions: {},
+        extraArguments: [],
```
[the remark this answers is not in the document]
snorrees: Non of the deploy flags are eligible for extract, so they are omitted. We can passthrough if you think it is useful, but I cant think of a reason atm.

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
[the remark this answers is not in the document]
snorrees: Its going to be:

`Couldn't extract manifest file. Sanity Create will not be available for this studio. Disable this message with SANITY_MANIFEST_EXTRACT_ENABLED=false`

(there wont be a article for this any time soon, so we are doing this instead)

snorrees:
All work on this leg of the Create integration is now concluded as far as I see it.

I've re-tested the latest commit with a tagged release (3.57.2-manifests.74) on another project.

## Compatible schema 

```
> sanity manifest extract
✅ Extracted manifest (1578ms)
```
```
> sanity deploy
✅ Checking project info
✅ Clean output folder (19ms)
✅ Build Sanity Studio (10714ms)
✅ Extracted manifest (1681ms)
✅ Verifying local content
✅ Deploying to Sanity.Studio

Success! Studio deployed to <redacted>
```

## Incompatible schema

```
> sanity manifest extract
ℹ️ Couldn't extract manifest file. Sanity Create will not be available for the studio.
Disable this message with SANITY_CLI_EXTRACT_MANIFEST_ENABLED=false

Error: Failed to load configuration file "sanity.config.ts":
hardcoded-error
(...stacktrace)
```

```
> sanity deploy
✅ Checking project info
✅ Clean output folder (19ms)
✅ Build Sanity Studio (10714ms)
ℹ️ Couldn't extract manifest file. Sanity Create will not be available for the studio.
Disable this message with SANITY_CLI_EXTRACT_MANIFEST_ENABLED=false
✅ Verifying local content
✅ Deploying to Sanity.Studio

Success! Studio deployed to <redacted>
```

snorrees:
ping @ricokahler @ryanbonial great if ya'll could look over my latest comments and changes, and review.

The idea is to get this into next weeks release, so we can shore up this part of the Create integration. 
(more chapters to come in other PRs)

ricokahler (approved):

PR #7404, opened 2024-08-21
refactor(sanity): add changes to allow for comments in version documents

RitaDias:
### Description

Added changes to the comments to allow for version comments to show up when the version is switched to.

https://github.com/user-attachments/assets/42de80ed-38b5-467c-b343-4d4022b53f7d

Added a todo for when the release is "complete", tested out the readOnly and it looks to do what we want. The rest will depend on design

### DIFF
--- a/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
@@ -166,6 +166,7 @@ interface CommentsListItemLayoutProps {
   onStatusChange?: (id: string, status: CommentStatus) => void
   readOnly?: boolean
   withAvatar?: boolean
+  version?: string
 }
 
 const RELATIVE_TIME_OPTIONS: RelativeTimeOptions = {useTemporalPhrase: true}
@@ -193,6 +194,7 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
     onStatusChange,
     readOnly,
     withAvatar = true,
+    version,
   } = props
   const {_createdAt, authorId, message, _id, lastEditedAt} = comment
   const [user] = useUser(authorId)
@@ -336,6 +338,9 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
     <TextSkeleton size={1} style={SKELETON_INLINE_STYLE} />
   )
 
+  // TODO - it's still unclear if we want to show that the comments belong to a specific version
+  // we can do that by displaying the version somewhere in the UI
+
   return (
     <RootStack
       data-menu-open={menuOpen ? 'true' : 'false'}
@@ -398,7 +403,6 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
               </Box>
             )}
           </Flex>
-
           {!isEditing && !displayError && (
             <ContextMenuBox data-root-menu={isParent ? 'true' : 'false'} onClick={stopPropagation}>
               <CommentsListItemContextMenu
--- a/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
@@ -338,8 +338,8 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
     <TextSkeleton size={1} style={SKELETON_INLINE_STYLE} />
   )
 
-  // TODO - it's still unclear if we want to show that the comments belong to a specific version
-  // we can do that by displaying the version somewhere in the UI
+  /* TODO - it's still unclear if we want to show that the comments belong to a specific version 
+  we can do that by displaying the version somewhere in the UI */
 
   return (
     <RootStack
--- a/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
+++ b/packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
@@ -166,7 +166,6 @@ interface CommentsListItemLayoutProps {
   onStatusChange?: (id: string, status: CommentStatus) => void
   readOnly?: boolean
   withAvatar?: boolean
-  version?: string
 }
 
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/comments/store/useCommentsStore.ts
```diff
@@ -84,7 +83,7 @@ export function useCommentsStore(opts: CommentsStoreOptions): CommentsStoreRetur
   const handleListenerEvent = useCallback(
     async (event: ListenEvent<Record<string, CommentDocument>>) => {
       // Fetch all comments on initial connection
-      if (event.type === 'welcome' && !didInitialFetch.current) {
+      if (event.type === 'welcome') {
```
RitaDias, line 86: needed to remove it in order for the comments to be refetched on switching perspectives in a document 

https://github.com/user-attachments/assets/a8177816-3d4f-4219-86e6-71a890852d0d

packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
```diff
@@ -193,6 +194,7 @@ export function CommentsListItemLayout(props: CommentsListItemLayoutProps) {
     onStatusChange,
     readOnly,
     withAvatar = true,
+    version,
```
pedrobonamin, line 197: qq: we are receiving here the version but this is later not used, wdyt if we remove it and then add it back in case we want to update the designs?
I think it's more likely nothing will change design wise

packages/sanity/src/core/comments/store/useCommentsStore.ts
pedrobonamin: Good catch, let's discuss in slack

pedrobonamin (commented):
Thanks for this Rita, looking very good! 

Added a small comment, feel free to discard it. 

I'm thinking if we may want to keep the CommentsProvider [publishedId](https://github.com/sanity-io/sanity/pull/7404/files#diff-12aa84437f8b661b1a0cb0d90e6dd836e11b30c86275f03c3b8cf4eba2ab4910R86) as it was, given the `useEditState` hook is using it, and it's supposed to receive the `publishedId` as the first parameter. It gracefully handles the version Id (but that could fail later)

So:
```diff
-  const publishedId = documentId
+ const publishedId = getPublishedId(documentId)
  const bundlePerspective = perspective?.startsWith('bundle.')
    ? perspective.split('bundle.').at(1)
    : undefined


const editState = useEditState(publishedId, documentType, 'default', bundlePerspective)
```

and for the rest of uses I think if we could be very explicit will be great.

e.g. 
```
 const versionOrPublishedId  = isVersionId(documentId) ? documentId: getPublishedId(documentID)
 (Naming can change)
```

and replace the other uses of `publishedId` for this new const, theres something nice about being very explicit that we are linking the comment to the version or the published id in lines 124, 241 and 269

packages/sanity/src/core/comments/components/list/CommentsListItemLayout.tsx
RitaDias: Yeah it's the latter (about waiting for design), mentioned in a todo on the same doc https://github.com/sanity-io/sanity/pull/7404/files#diff-68a543cbb2c056e1409c8f8453e10debce5d8f25bd1afda3da6933f26d886af5R341

I can just remove it if you prefer :)

pedrobonamin: If you feel like it I'll appreciate it 😄 I think the less changes we add, specially to existing components, the easier the final review will be. 
Thanks!

RitaDias: True, I'll go ahead and remove it :)

packages/sanity/src/structure/panes/document/comments/CommentsWrapper.tsx
```diff
@@ -37,10 +38,11 @@ function CommentsProviderWrapper(props: CommentsWrapperProps) {
   const {children, documentId, documentType} = props
 
   const {enabled} = useCommentsEnabled()
-  const {connectionState, onPathOpen, inspector, openInspector} = useDocumentPane()
+  const {connectionState, onPathOpen, inspector, openInspector, version} = useDocumentPane()
   const router = useRouter()
   const {params, setParams, createPathWithParams, ...paneRouter} = usePaneRouter()
   const perspective = paneRouter.perspective ?? router.stickyParams.perspective
+  const docId = version ? getVersionId(documentId, version) : documentId
```
juice49, line 45: Tiny nitpick: it seems potentially confusing to have both a `documentId` and a `docId` in scope with different meanings. Have you considered renaming `docId` to something like `resolvedDocumentId`, or passing this value directly below on line 87 to avoid binding it to a new name altogether?

packages/sanity/src/core/comments/context/comments/CommentsProvider.tsx
```diff
@@ -84,7 +83,7 @@ export const CommentsProvider = memo(function CommentsProvider(props: CommentsPr
   const commentsEnabled = useCommentsEnabled()
   const [status, setStatus] = useState<CommentStatus>('open')
   const {client, createAddonDataset, isCreatingDataset} = useAddonDataset()
-  const publishedId = getPublishedId(documentId)
+  const publishedId = documentId
```
juice49, line 86: Could it be clearer if we removed the `publishedId` binding and instead referred to `documentId` directly where it's used?

packages/sanity/src/core/comments/store/useCommentsStore.ts
juice49: Can you ping the other reviewers when you come to a conclusion on this? 🙂

RitaDias: I'll be making this change 🔜

packages/sanity/src/structure/panes/document/comments/CommentsWrapper.tsx
RitaDias: Renamed in an upcoming commit :)

pedrobonamin (approved):
Thanks @RitaDias , looks good to me !

PR #7410, opened 2024-08-22
refactor(sanity): Move date inputs UI Components to ui-components

RitaDias:
### Description

As part of the content releases changes, we're moving part of the date picker ui components to the ui-component folder 

### What to review

Does this make sense? 
Should I have moved other things / did I move things that I shouldn't have?

### Testing

The original tests should be good enough

### Notes for release

N/A


ricokahler (approved):
this looks good but i wonder if we should add some boundary linting similar to how we disallow relative imports from `core` inside of `structure`.

is the `ui-components` folder new? (if it's not new then ignore this comment) should that also have import boundary linting? is it an issue if a component within the ui-components folder imports from other places?

if so can u update this PR to include updating the eslint boundary rules?

RitaDias:
> this looks good but i wonder if we should add some boundary linting similar to how we disallow relative imports from `core` inside of `structure`.
> 
> is the `ui-components` folder new? (if it's not new then ignore this comment) should that also have import boundary linting? is it an issue if a component within the ui-components folder imports from other places?
> 
> if so can u update this PR to include updating the eslint boundary rules?

The ui-components folder is not new, we've had it for a while and we do want to use it in core (like the comments and tasks use the buttom from the ui-components folder vs the sanity ui ones because they are more opiniated ) :)

bjoerge:
> is the ui-components folder new? (if it's not new then ignore this comment) should that also have import boundary linting? is it an issue if a component within the ui-components folder imports from other places?

This is a really good point – the way we import this currently doesn't _seem_ quite right at first glance. But this PR isn't really changing any of that, so let's look into this separately?

bjoerge (approved):
Very nice, thanks @RitaDias!

PR #7416, opened 2024-08-23
feat(sanity): add always present document action to copy url to clipboard

drewlyton:
### Description

This PR adds a document pane menu item that enables user's to easily copy the current 
document's URL to their clipboard and share it with colleagues and collaborators.

[Linear issue](https://linear.app/sanity/issue/GRO-2405/add-share-button-to-studio)

https://github.com/user-attachments/assets/ba52e457-8d5a-44ed-ac0c-e37a5a034603

### What to review

The code add for this feature is relatively straightforward. However, I'm unsure if the logic's location and the placement of the internalization strings is acceptable. Feedback on this would be greatly appreciated.

### Testing

To test, navigate to a document and click on the added menu item. The current URL should be successfully copied to the clipboard.

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -43,7 +43,6 @@ import {
   useValidationStatus,
 } from 'sanity'
 import {DocumentPaneContext} from 'sanity/_singletons'
-import {useRouter} from 'sanity/router'
 
 import {usePaneRouter} from '../../components'
 import {structureLocaleNamespace} from '../../i18n'
@@ -404,7 +403,6 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   )
 
   const telemetry = useTelemetry()
-  const {resolveIntentLink} = useRouter()
 
   const handleMenuAction = useCallback(
     (item: PaneMenuItem) => {
@@ -415,8 +413,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
 
       if (item.action === 'copy-document-url' && navigator) {
         telemetry.log(DocumentURLCopied)
-        const documentIntentLink = resolveIntentLink('edit', {id: documentId})
-        navigator.clipboard.writeText(window.location.origin + documentIntentLink)
+        navigator.clipboard.writeText(window.location.toString())
         pushToast({
           id: 'copy-document-url',
           status: 'info',
@@ -462,8 +459,6 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
       toggleLegacyInspect,
       pushToast,
       telemetry,
-      resolveIntentLink,
-      documentId,
     ],
   )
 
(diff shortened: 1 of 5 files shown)


bjoerge (dismissed):
Code looks good from my side.

I'll defer to @runeb about UI location, etc. Just a note: this is copying a link to the document within the navigation structure that the user took to get to it, and might not always be what the user expect. As an alternative to consider (if you didn't already), you could copy the intent link for editing the document (this is what you navigate to when clicking the presence menu, for reference)

joneidejohnsen:
> Code looks good from my side.
> 
> I'll defer to @runeb about UI location, etc. Just a note: this is copying a link to the document within the navigation structure that the user took to get to it, and might not always be what the user expect. As an alternative to consider (if you didn't already), you could copy the intent link for editing the document (this is what you navigate to when clicking the presence menu, for reference)

@drewlyton Support this, and suggest we use the intent link here rather than the actual URL

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -400,13 +403,27 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
   const handleMenuAction = useCallback(
     (item: PaneMenuItem) => {
       if (item.action === 'production-preview' && previewUrl) {
         window.open(previewUrl)
         return true
       }
 
+      if (item.action === 'copy-document-url' && navigator) {
+        telemetry.log(DocumentURLCopied)
+        navigator.clipboard.writeText(window.location.origin + documentEditIntentLink)
```
drewlyton, line 418: Use intent URL instead of just current URL

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -400,13 +403,27 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+
   const handleMenuAction = useCallback(
     (item: PaneMenuItem) => {
       if (item.action === 'production-preview' && previewUrl) {
         window.open(previewUrl)
         return true
       }
 
+      if (item.action === 'copy-document-url' && navigator) {
+        telemetry.log(DocumentURLCopied)
```
drewlyton, line 417: Add telemetry event for tracking this action

bjoerge (dismissed):

ricokahler (dismissed):

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -415,8 +413,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
 
       if (item.action === 'copy-document-url' && navigator) {
         telemetry.log(DocumentURLCopied)
-        const documentIntentLink = resolveIntentLink('edit', {id: documentId})
-        navigator.clipboard.writeText(window.location.origin + documentIntentLink)
+        navigator.clipboard.writeText(window.location.toString())
```
ricokahler, line 416: can you add a comment here explaining why the intent link resolver wasn't used? I liked that implementation but after read the slack thread with jon's reasoning, i understand why. i think that explanation could be helpful here too

ricokahler (dismissed):

ricokahler (approved):

PR #7419, opened 2024-08-26
refactor(sanity): Add toast when discarding version and for deleting releases

RitaDias:
### Description

- Add translations when discarding version and deleting releases
- Add toast (and translations) when deletions are successful

### What to review

- Any mistranslations?

### Testing

Existing tests should suffice

### DIFF
--- a/packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
+++ b/packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
@@ -41,6 +41,13 @@ export const BundleMenuButton = ({disabled, bundle, documentCount}: BundleMenuBu
     try {
       setDiscardStatus('discarding')
       await deleteBundle(bundle)
+      toast.push({
+        closable: true,
+        status: 'success',
+        description: (
+          <Translate t={t} i18nKey={'action.delete.success'} values={{title: bundle.title}} />
+        ),
+      })
       setDiscardStatus('idle')
       if (router.state.bundleSlug) {
         // navigate back to bundle overview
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
```diff
@@ -51,10 +51,17 @@ export const BundleMenuButton = ({disabled, bundle, documentCount}: BundleMenuBu
...
         description: e.message,
       })
     } finally {
+      toast.push({
+        closable: true,
+        status: 'success',
+        description: (
+          <Translate t={t} i18nKey={'action.delete.success'} values={{title: bundle.title}} />
+        ),
+      })
```
pedrobonamin, line 64: This needs to be part of the `try` block, after the `await deleteBundle(bundle)` action

The `finally` block will be always executed, even if the action failed. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch
> The code in the finally block will always be executed before control flow exits the entire construct.

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentActions.tsx
```diff
@@ -34,12 +34,24 @@ export const DocumentActions = memo(
...
+          closable: true,
+          status: 'success',
+          description: (
+            <Translate
+              t={t}
+              i18nKey={'action.discard-version.success'}
+              values={{title: document.document.title as string}}
+            />
+          ),
+        })
```
pedrobonamin, line 50: Same as above 😃

PR #7420, opened 2024-08-26
Remove constraint size publish

RitaDias:
### Description

Add prop to the bundle form to fix behavior of the date time input 

image


packages/sanity/src/ui-components/inputs/DateInputs/DateTimeInput.tsx
RitaDias, line 1: Will remove once [change](https://github.com/sanity-io/sanity/pull/7421) is pushed to next and corel is rebased

pedrobonamin (approved):

PR #7423, opened 2024-08-26
feat(tasks): link tasks to the version document

pedrobonamin:
### Description

When tasks are created for a version document , link the task with the version document instead of the live document, so tasks for a version are only visible in a version. And the version won't show tasks from the published document.
https://www.loom.com/share/487a5c732b7d4367a7d737720f24351a?sid=99a7112b-00a0-4730-a0e2-da5bd78a1091

#### Other changes: 
- This PR renames and move `getBundleSlug` to `getVersionFromId`, given the id is not specifically related to the bundles, it's related to the `version` 

#### TODO: 
- Once the search allows looking for version documents the tasks document search should be updated to allow users to link tasks to version documents from there.

### DIFF
--- a/packages/sanity/src/core/util/draftUtils.ts
+++ b/packages/sanity/src/core/util/draftUtils.ts
@@ -1,6 +1,5 @@
 import {type SanityDocument, type SanityDocumentLike} from '@sanity/types'
 
-import {getBundleSlug} from '../bundles/util/util'
 import {isNonNullable} from './isNonNullable'
 
 /** @internal */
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -112,6 +111,20 @@ export function getVersionId(id: string, bundle: string): string {
...
 }
 
+/**
+ *  @internal
+ *  Given an id, returns the versionId if it exists.
+ *  e.g. `versions.summer-drop.foo` = `summer-drop`
+ *  e.g. `drafts.foo` = `undefined`
+ *  e.g. `foo` = `undefined`
+ */
+export function getVersionFromId(id: string): string | undefined {
```
RitaDias, line 121: Really good change!!

PR #7426, opened 2024-08-28
fix(core): update how bundleSlug is obtained from the version ids

pedrobonamin:
### Description

When the new versions id was introduced `versions.release-slug.foo` the update to the way we calculate the `bundleSlug` was not added. 
This adds that change, using the new `getVersionFromId` function which gives us safety in how the bundleSlug is obtained from the id. 

Previously, version ids were `release-slug.foo` that's why we were accessing it using the `.at(0)`


RitaDias (approved):
🚀🚀

PR #7429, opened 2024-08-29
chore: support for sticky params in URL and intent ops

jordanl17:
### Description
Adds ability to define sticky search params which will be retained state-fully in the URL throughout navigation

### What to review
Primarily `packages/sanity/src/router/RouterProvider.tsx` is where the handling and persistence of the params is handled

### Notes for release
N/A


packages/sanity/src/router/RouterProvider.tsx
```diff
@@ -114,17 +144,55 @@ export function RouterProvider(props: RouterProviderProps): ReactElement {
...
+      resolveIntentLink,
+      resolvePathFromState,
+      routerState,
+      stickyParams,
+    ],
   )
 
   return <RouterContext.Provider value={router}>{props.children}</RouterContext.Provider>
 }
+const STICKY: string[] = []
```
jordanl17, line 184: Currently there are no sticky params defined, however in content releases, the `perspective` key will be added here (the first current use of sticky params), but they can be used in other instances also

RitaDias (commented):
The code overall looks good to me 👋 quick question before I ✅ do you have some use cases that I can manually test? I don't think we have any automated tests set up but I'd like to see if I missed anything on my manual pass :)

pedrobonamin (approved):
Thanks for pushing this! 
Looks good to me

jordanl17:
> The code overall looks good to me 👋 quick question before I ✅ do you have some use cases that I can manually test? I don't think we have any automated tests set up but I'd like to see if I missed anything on my manual pass :)

Hmm, I'm not so sure (as I wasn't actually the one to implement this). However it does seem that its only current use would be with `perspective`. @juice49 with the changes you've introduced is there any usage example that already exists?

juice49:
@RitaDias @jordanl17 we have some test coverage for this behaviour in `packages/sanity/src/router/IntentLink.test.tsx`. However, I forgot that this branch omits all of the sticky parameters from the `STICKY` constant, so these tests currently fail.

We can't test this behaviour manually or automatically while `STICKY` is empty. Wondering about the best way to proceed.

bjoerge (approved):
Nice, LGTM

jordanl17:
@juice49 @RitaDias - stick params test updated. Now the stick parameters are defined separately and are mocked in the `IntentLink` tests

packages/sanity/src/router/RouterProvider.tsx
```diff
@@ -2,6 +2,7 @@ import {fromPairs, partition, toPairs} from 'lodash'
 import {type ReactElement, type ReactNode, useCallback, useMemo} from 'react'
 import {RouterContext} from 'sanity/_singletons'
 
+import {STICKY_PARAMS} from './stickyParams'
```
juice49, line 5: Good rename 👍.

pedrobonamin (approved):

PR #7431, opened 2024-08-29
chore(core, structure): validation on references

jordanl17:
### Description

It refactors document-store pair validation function by abstracting the part of that function that takes care of **validating** the document from the **fetcher**. 
The document-store pair uses the document-store to fetch the document, which makes it not scalable enough for the releases use case, by dividing the validation from the fetcher, we could provide a document to the function `validateDocumentWithReferences` and this function will be responsible only for the document validation. 

### Testing

Existing tests are enough to cover this refactor `packages/sanity/src/core/store/_legacy/document/document-pair/validation.test.ts` 

### Notes for release
n/a, it's a refactor, functionality remains the same


pedrobonamin (approved):
Looks good to me, thanks @jordanl17

PR #7432, opened 2024-08-29
chore(core): single addon dataset context

jordanl17:
### Description
The addon dataset context will only be wrapped if it does not already exist. This is inkeeping with how many newer contexts are being defined, to avoid multiple separate contexts being nested

### Testing
N/A

### Notes for release
N/A


RitaDias (approved):
Thanks for doing This Jordan ✅

pedrobonamin (approved):

PR #7435, opened 2024-08-29
feat(structure): add history and review tabs to changes inspector panel

pedrobonamin:
### Description

This pull request joins the history revision and the review changes into 1 inspector, divided by tabs.

#### **History tab**
Reuses the existing [`Timeline`](https://github.com/sanity-io/sanity/blob/corel-143/packages/sanity/src/structure/panes/document/timeline/timeline.tsx) component to render the view of the `Timeline` and allows the user to select which revision to see in the form.
By moving this into the history tab, the `TimelineMenu` was removed from the [`DocumentPanelHeader`](packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx)

Screenshot 2024-08-29 at 13 06 44

#### **Review changes tab**

Reuses the existing Review changes, it now includes the **From** and **To** selector, making it easier for the user to understand where the changes he is seeing are coming from.

Screenshot 2024-08-29 at 13 08 07

**Other**
When clicking the fields `ChangeBarButton` the inspector will open in the Review changes tab.
When opening the inspector through the Actions menu, it will open in the History tab

### What to review

Are the changes correct?
Is there something else to consider in this changes related to the review changes that may have been lost?

### Testing

e2e testing to be included once designs are on a final stage and all the changes have been approved by design. 

### Notes for release
n/a this will go to a feature branch

### DIFF
--- a/packages/sanity/src/structure/panes/document/inspectors/changes/ChangesTabs.tsx
+++ b/packages/sanity/src/structure/panes/document/inspectors/changes/ChangesTabs.tsx
@@ -1,0 +1,64 @@
+import {CloseIcon} from '@sanity/icons'
+import {Box, Flex, TabList, TabPanel} from '@sanity/ui'
+import {useState} from 'react'
+import {type DocumentInspectorProps, useTranslation} from 'sanity'
+
+import {Button, Tab} from '../../../../../ui-components'
+import {structureLocaleNamespace} from '../../../../i18n'
+import {ChangesInspector} from './ChangesInspector'
+import {HistorySelector} from './HistorySelector'
+
+export function ChangesTabs(props: DocumentInspectorProps) {
+  const [id, setId] = useState<'history' | 'review'>('history')
+  const {t} = useTranslation(structureLocaleNamespace)
+
+  return (
+    <Flex direction="column" padding={0} height="fill">
+      <Flex align="center" padding={3} gap={2}>
+        <TabList space={1} flex={1}>
+          <Tab
+            aria-controls="history-panel"
+            id="history-tab"
+            label={t('changes.tab.history')}
+            onClick={() => setId('history')}
+            selected={id === 'history'}
+          />
+          <Tab
+            aria-controls="review-changes-panel"
+            id="changes-tab"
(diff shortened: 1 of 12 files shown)


packages/sanity/src/structure/panes/document/inspectors/changes/ChangesTabs.tsx
RitaDias, line 1: Might be out of scope for this task but I wonder if this makes sense to move to an UI component of sorts.
This looks very similar to for example the Editor / JSON section beneath the document header 
image

There is something else that also uses a similar logic and looks the same but I unfortunately don't recall the name of, the point still stands though :) probably something for another time

RitaDias (approved):
This looks good!
I left one comment but as I mentioned, feels like it's out of scope for this PR :)

packages/sanity/src/structure/panes/document/inspectors/changes/ChangesTabs.tsx
RitaDias: Ahh this makes sense :)
Fantastic then, don't mind me, carry on!

bjoerge:
This is a really really nice improvement ✨ 

My only feedback is that it feels slightly awkward that clicking "Review Changes" from the menu takes you to the "History" tab, and there's another tab here named "Review Changes":

https://github.com/user-attachments/assets/535b6643-c33f-4d38-835b-d037209cb49c

Maybe the "Review Changes" tab should be named "Changes" or "Diff" instead?

pedrobonamin:
> This is a really really nice improvement ✨
> 
> My only feedback is that it feels slightly awkward that clicking "Review Changes" from the menu takes you to the "History" tab, and there's another tab here named "Review Changes":
> 
>  Screen.Recording.2024-08-30.at.14.08.18.mov 
> Maybe the "Review Changes" tab should be named "Changes" or "Diff" instead?

Hey @bjoerge , thanks for raising this, I had a similar feeling, I went to check back the prototype and the action in the prototype has been renamed from "review changes" to  "history", will update the action name here.

PR #7439, opened 2024-08-29
feat(sanity): preserve current sticky parameters when resolving intent links

juice49:
### Description

This branch introduces preservation of current sticky parameters when resolving intent links. In practice, this means the `perspective` parameter is preserved when resolving an intent.

Search parameters provided when resolving an intent link take precedence over any sticky parameters currently applied, allowing developers to replace sticky parameters when resolving an intent link.

### What to review

Sticky parameters are preserved when resolving intent links. For example; when navigating to a document via a global search result.

### Testing

Added new unit tests in `packages/sanity/src/router/IntentLink.test.tsx`.

### DIFF
--- a/packages/sanity/src/router/IntentLink.test.tsx
+++ b/packages/sanity/src/router/IntentLink.test.tsx
@@ -1,5 +1,6 @@
 import {describe, expect, it} from '@jest/globals'
 import {render} from '@testing-library/react'
+import {noop} from 'lodash'
 
 import {IntentLink} from './IntentLink'
 import {route} from './route'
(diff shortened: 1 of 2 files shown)


jordanl17:
@juice49 nice, could this go into `next` directly?

juice49:
@jordanl17 I don't see why not! I didn't realise we had a PR open for moving the stick params functionality to `next`. Do you mind if I add my commits to your branch?

RitaDias (commented):
Is this related to https://github.com/sanity-io/sanity/pull/7429?
It feels like it might be? Would it make sense to have this be directed towards next vs our feature branch?

I'm not seeing anything content releases specific that would require this to be feature branch only, but I might be wrong :)

juice49:
@RitaDias Yep! 🙂 I didn't realise #7429 existed until Jordan told me about it yesterday. I think we'll instead add this change via that PR.

juice49:
@jordanl17 I've added this change to #7429.

PR #7444, opened 2024-08-30
feat(sanity): add telemetry

RitaDias:
### Description

Added telemetry to the following actions + fixed issue when adding a version to release made the code that came after the createVersionSuccess not run (wrong id was being passed)

**Bundle**

- [x] Bundle Created, 
- [x] Bundle Updated, 
- [x] Bundle Deleted, 
- [x] Bundle Published, 
- [x] Bundle Archived, 
- [x] Bundle Unarchived

**Version**
- [x] All Versions created
- [x] Version created from draft (waiting confirmation)
- [x] Version created from version (waiting confirmation)

**Out of scope** (separate stories have been created):
- [ ] Bundle scheduled
- [ ] Bundle Unscheduled

### What to review

Does this make sense? Should the messages be different?


packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -67,7 +71,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
     // set up the listener before executing
     const createVersionSuccess = firstValueFrom(
       documentStore.pair
-        .operationEvents(versionId, documentType)
+        .operationEvents(getPublishedId(documentId), documentType)
```
juice49, line 74: Great catch 🙏.

packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
```diff
@@ -74,6 +82,12 @@ export const BundleMenuButton = ({disabled, bundle, documentCount}: BundleMenuBu
       ...bundle,
       archivedAt: isBundleArchived ? undefined : new Date().toISOString(),
     })
+
+    if (isBundleArchived) {
+      telemetry.log(UnarchivedRelease)
```
juice49, line 87: Trying to reason about this code. It might just be that I'm confused, but will this not log an `UnarchivedRelease` event when deleting an archived release and, below, log an `ArchivedRelease` event when deleting an unarchived release? Apologies if I'm mistaken, or this is the intention.

RitaDias: I know what you mean I also felt a bit confused. This is coming from the bundle object at the top which is not updated fast enough to where this section of the code.

So technically it's:
- IsBundleArchived `true` -> it means that it's in the process of becoming false, so the event we want to track is unarchive
- IsBundleArchived `false` -> if means that it's in the process of becoming true, so the event we want to track is archive

Even the  `achivedAt: isBundleArchived ? undefined : new Date().toISOString(),` follows the same type of logic. If it's true then set it as undefined because technically it won't be archived anymore, if it's _not_ archived then add the date because it means it'll become archived

I'll add this comment there, unless you feel there is a better way? 🤔

RitaDias: @juice49 I'm closing this, moved the commits to here https://github.com/sanity-io/sanity/pull/7460 :)

PR #7448, opened 2024-08-30
feat(sanity): update restore button

RitaDias:
### Description

Update the footer when reviewing different revisions of a document

https://github.com/user-attachments/assets/8490901e-2125-4de0-bb83-0b7b36e78b3c

### What to review

Does everything work as expected?

### Testing

I don't think new tests are necessary. e2e can be added later but I'm unsure if the noise is necessary beyond testing the "reverting to revision" and that test likely already exists.
Behaviour remains the same, it is mostly a visual change 

### Notes for release

n/a this will go to a feature branch

### DIFF
--- a/packages/sanity/src/structure/panes/document/statusBar/RevisionStatusLine.tsx
+++ b/packages/sanity/src/structure/panes/document/statusBar/RevisionStatusLine.tsx
@@ -1,0 +1,56 @@
+import {RestoreIcon} from '@sanity/icons'
+import {Box, Flex, Text} from '@sanity/ui'
+import {format} from 'date-fns'
+import {createElement} from 'react'
+import {Translate, useTranslation} from 'sanity'
+import {styled} from 'styled-components'
+
+import {useDocumentPane} from '../useDocumentPane'
+
+export const StatusText = styled(Text)`
+  color: var(--card-muted-fg-color);
+
+  em {
+    color: var(--card-fg-color);
+    font-weight: 500;
+    font-style: normal;
+  }
+`
+
+export function RevisionStatusLine(): JSX.Element {
+  const {displayed} = useDocumentPane()
+  const {t} = useTranslation()
+
+  const message = {
+    name: 'revision',
+    icon: RestoreIcon,
+    text: (
+      <Translate
+        t={t}
+        i18nKey="document-status.revision-from"
+        values={{
+          date: format(
+            new Date(displayed?._updatedAt || displayed?._createdAt || 0),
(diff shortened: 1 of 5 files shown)


packages/sanity/src/structure/panes/document/statusBar/RevisionStatusLine.tsx
```diff
@@ -0,0 +1,56 @@
...
+        values={{
+          date: format(
+            new Date(displayed?._updatedAt || displayed?._createdAt || 0),
+            `MMM d, yyyy '@' pp`,
+          ),
+        }}
+      />
+    ),
+    tone: 'caution',
+  }
```
pedrobonamin, line 40: Is there any reason why we need this in an object here instead of baking this into the JSX?
e.g. 

```
return (
    <Flex flex={1} gap={3} padding={2}>
        <Box flex="none">
          <Text size={1}><RestoreIcon /></Text>
        </Box>
        <Box flex={1}>
          <StatusText size={1} textOverflow="ellipsis">
             <Translate
                t={t}
               i18nKey="document-status.revision-from"
              values={{
               date: format(
                    new Date(displayed?._updatedAt || displayed?._createdAt || 0),
                  `MMM d, yyyy '@' pp`,
                ),
               }}
             />
          </StatusText>
        </Box>
      </Flex>
)
```

pedrobonamin (approved):
Thanks for doing this @RitaDias ! Looking great.
Added a suggestion, feel free to discard it.

packages/sanity/src/structure/panes/document/statusBar/RevisionStatusLine.tsx
RitaDias: not right now, I'm trying to keep in mind the approach that Marius had on his prototype (which has objects). The approach with objects is also used in the studio today (if you look at the way that we do the button for the primary action - **here for example**). 

I thought (and still think?) that it might be wise to keep it like this for now because looking at the prototype we'll make use of this approach. That was my line of thought at least anyway :)

What do you think, keep the object or move into baking it in the JSX?

pedrobonamin: Makes sense, lets keep it in that case! Thanks for adding the extra details I needed 😄

RitaDias: Of course, thank you for reviewing it! :)

PR #7453, opened 2024-08-30
refactor(core): minor type usage improvement and using task plugin name

jordanl17:
### What to review
* Passing function reference to `collate` was not matching expected types, so now anon function is defined
* Using the tasks plugin name variable when declaring the plugin

### Testing
N/A

### Notes for release
N/A


RitaDias (approved):
Let's go :shipit:

juice49:
This looks good! One small note: as of PR #7415, [`collate` has the old signature again](https://github.com/sanity-io/sanity/pull/7415/files#diff-cdd1fc4337b45f431507fa1b2ebb702510d7c9472df372033daaa9f1871add8cR183), so this change might not be strictly necessary.

PR #7458, opened 2024-09-03
feat(structure): squash chunks in timeline.

pedrobonamin:
### Description

This PR changes how the rows in the history are shown in the UI, collapsing draft edits into the publish action.
Once the **draft** changes are **published** all the draft edits done before that **publish** will be squashed and shown as 1 change, that could be expanded, showing also the collaborators for that published.
This means:
If we have the following mutations: 
`[draftEdit, publish,  draftEdit, draftEdit, create]` 
it will be shown as 
`[draftEdit, publish (expandable), create]`
On expand, it will show the two collapsed draftEdit actions

The changes are not removed from the history, they are only visually squashed for easier navigation in the UI.

The story has been updated to include this new behavior, you can check it here https://test-studio-git-corel-145.sanity.dev/test/workshop/structure;panes;documentTimelineMenu;timelineItem?scheme=light

This is achieved by creating a relation between parents publish action and children draftEdit actions, in the `addChunksMetadata` function.

Screenshot 2024-09-03 at 11 11 33
Screenshot 2024-09-03 at 11 11 50

https://github.com/user-attachments/assets/bdbf89c0-825f-4d47-a655-0df22c51fee1

### What to review

Is this correct?  Are changes clear?

### Notes for release
n/a


packages/sanity/src/structure/panes/document/timeline/expandableTimelineItemMenu.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {structureLocaleNamespace} from '../../../i18n'
+import {TIMELINE_LIST_WRAPPER_ID} from './timeline'
+import {TIMELINE_MENU_PORTAL} from './timelineMenu'
+
+/**
+ * This is a hack to force the scrollbar to not appear when the list is expanding,
+ * if we don't do this the scrollbar will appear for a brief moment when the list is expanding and then disappear
+ * when the list is fully expanded.
+ */
+function hideScrollbarOnExpand(isExpanded: boolean) {
```
pedrobonamin, line 15: Explanation on why this is needed, thoughts and suggestions are welcomed! 
https://www.loom.com/share/046f783e048f463e84e6e6fd31e551b7

RitaDias:
👋 
This is just a design thing and unrelated to the code itself (I'm still reviewing it)
But given that I'm used to the items being "flat" I honestly would have never gone to the "..." to even realise that I could expand them 😬

pedrobonamin:
> 👋 This is just a design thing and unrelated to the code itself (I'm still reviewing it) But given that I'm used to the items being "flat" I honestly would have never gone to the "..." to even realise that I could expand them 😬

@RitaDias  yes, makes sense, I am thinking on proposing to @mariuslundgard using a Caret icon instead of the three dots. Specially considering that we have only 1 action the user can do in this cases. I think it will make it easier to notice that you can expand it. 
Also, handling the popovers here was quite tricky, so it will also simplify things a bit

packages/sanity/src/structure/panes/document/timeline/expandableTimelineItemMenu.tsx
RitaDias: This is a really ingenious solution but I have to admit that I hate it though I understand the need 😭 
While I can potentially think of more elegant ways of fixing this (we do have sections of the code that uses the ResizeObserver) in the end I think we would still need to end up changing the styles like you're doing here.

If anyone else thinks of a better solution (Herman might actually!) I'd also like to know 👀

packages/sanity/src/structure/panes/document/timeline/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+export type PublishChunk = Omit<Chunk, 'type'> & {
+  type: 'publish'
+  children: string[]
+  collaborators: Set<string>
+}
+
+export const isNonPublishChunk = (chunk: Chunk): chunk is NonPublishChunk =>
+  chunk.type !== 'publish'
+
+export const isPublishChunk = (chunk: Chunk): chunk is PublishChunk => chunk.type === 'publish'
```
RitaDias, line 17: Might need to add some comments / tagging these as `@internal`, since I'm not sure if these exports will show up in the automated docs :)

pedrobonamin: They are not exported from the package, so I think they shouldn't be picked by the automated docs. There is a check that runs on build which validates that if something is exported from the packages it includes the supported tags.

packages/sanity/src/structure/panes/document/timeline/expandableTimelineItemMenu.tsx
pedrobonamin: Will give it a try to the resizeObserver! I was not aware of that, thanks!

packages/sanity/src/structure/panes/document/timeline/utils.ts
RitaDias: Ah this makes sense! Disregard my comment :)

packages/sanity/src/structure/panes/document/timeline/expandableTimelineItemMenu.tsx
pedrobonamin: Hey @RitaDias, thanks for the suggestion, I tried it but it didn't work in the way I need it, I think it's because the `onResize` callback is triggered by an useEffect and that useEffect is running after the element has changed it's scroll size.
So the bar is still visible when expanding.

I think we will have to go with this not nice idea but it does the job. Unless we can think of something else 😄 .

RitaDias: Yeah I'd assume that would be the case!
I'm approving it, the way I see it I think we can go ahead with this solution and rework it at a later point if we feel it's not good enough to go to production we'll handle it then :)

RitaDias (approved):
✅

PR #7460, opened 2024-09-03
feat(sanity): add telemetry for versions

RitaDias:
### Description

Added telemetry to the following actions + fixed issue when adding a version to release made the code that came after the createVersionSuccess not run (wrong id was being passed)

**Bundle**

- [x] Bundle Created, 
- [x] Bundle Updated, 
- [x] Bundle Deleted, 
- [x] Bundle Published, 
- [x] Bundle Archived, 
- [x] Bundle Unarchived

**Version**
- [x] All Versions created
- [x] Version created from draft (waiting confirmation)
- [x] Version created from version (waiting confirmation)

**Draft**
- [x] Draft Created (part of it is in https://github.com/sanity-io/sanity/pull/7459)

**Out of scope** (separate stories have been created):
- [ ] Bundle scheduled
- [ ] Bundle Unscheduled

### What to review

Does this make sense? Should the messages be different?


packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
```diff
@@ -74,6 +82,12 @@ export const BundleMenuButton = ({disabled, bundle, documentCount}: BundleMenuBu
       ...bundle,
       archivedAt: isBundleArchived ? undefined : new Date().toISOString(),
     })
+
+    if (isBundleArchived) {
+      telemetry.log(UnarchivedRelease)
+    } else {
+      telemetry.log(ArchivedRelease)
+    }
```
RitaDias, line 90: Comments from previous PR https://github.com/sanity-io/sanity/pull/7444#discussion_r1740679570

So technically it's:
- IsBundleArchived `true` -> it means that it's in the process of becoming false, so the event we want to track is unarchive
- IsBundleArchived `false` -> if means that it's in the process of becoming true, so the event we want to track is archive

Even the  `achivedAt: isBundleArchived ? undefined : new Date().toISOString(),` follows the same type of logic. If it's true then set it as undefined because technically it won't be archived anymore, if it's _not_ archived then add the date because it means it'll become archived

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -56,6 +59,9 @@ export function BundleDetailsDialog(props: BundleDetailsDialogProps): JSX.Elemen
         await submit(submitValue)
         if (formAction === 'create') {
           setPerspective(value._id)
+          telemetry.log(CreatedRelease)
```
jordanl17, line 62: A thought here, though I'm unsure if this does or doesn't align with our current telemetry strategies elsewhere. But interested in your thoughts - tracking how/where a release is created eg. from the global perspective menu or from within the releases plugin

RitaDias: That's true that might be interesting to know 💡 
I'll add it! Thank you for the idea Jordan, great catch!

RitaDias: Added it on https://github.com/sanity-io/sanity/pull/7460/commits/64bf480de2bd829ba86b9e3f2b5b0f8cf6fd402e

I kept the terms as `origin: 'structure' | 'release-plugin' ` because we know that the section where we create releases from the structure tool will likely change (and might not end up in the nav bar as is now) and kept the name  simple as `release-plugin` because I don't think we'll have multiple places where we'll be able to create the release within the plugin (nor do I know if we need that sort of granularity) + with the re-design it might not be in the "overview" :)

Thoughts are welcome!

pedrobonamin:
Added @sanity-io/growth as reviewers, having their early feedback will be appreciated

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -67,7 +71,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
     // set up the listener before executing
     const createVersionSuccess = firstValueFrom(
       documentStore.pair
-        .operationEvents(versionId, documentType)
+        .operationEvents(getPublishedId(documentId), documentType)
```
pedrobonamin, line 74: Hey @RitaDias why do we need this change?

packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
jordanl17: Makes sense, thanks for clarifying

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
```diff
@@ -14,14 +16,16 @@ interface BundleDetailsDialogProps {
   onCancel: () => void
   onSubmit: () => void
   bundle?: BundleDocument
+  origin?: 'structure' | 'release-plugin'
```
jordanl17, line 19: COULD: export `OriginInfo` from `releases.telemetry` and use that here instead

jordanl17 (approved):
Great to have this added to corel, @RitaDias

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
RitaDias: I think I might have left the explanation on the previous PR but it's on the description: fixed issue when adding a version to release made the code that came after the createVersionSuccess not run (wrong id was being passed)

Basically the "success" message never came back because technically it never succeeded (the createVersionSuccess check). This is a separate thing from the actual creating of the version but it listens to the success based on the publishedId. You can see [here](https://github.com/sanity-io/sanity/blob/corel-149-2/packages/sanity/src/core/store/_legacy/document/document-store.ts#L177-L180) :)

pedrobonamin: Got it! Thanks !

pedrobonamin (approved):
Thanks for going through this and adding all the details. Great work @RitaDias !
I think we can merge it to Corel so it doesn't get merge conflicts later and in the meanwhile share it with growth to get their eyes. But not necessary to wait for their approval to merge it to Corel

packages/sanity/src/core/bundles/components/dialog/BundleDetailsDialog.tsx
RitaDias: Great suggestion, done :)

packages/sanity/src/core/releases/components/BundleMenuButton/BundleMenuButton.tsx
RitaDias: I've added a comment here just to make sure that we have this in the future :)

PR #7462, opened 2024-09-04
feat(structure): History UI updates

pedrobonamin:
### Description

This PR introduces UI changes into the `history` and `revision` panel.

- Chunks are squashed together on publish.
- Moves the **revision** selector to the history panel.
- Timeline item are updated with new looks.
- Footer shown when seeing a revision is updated with new looks.
- The inspector panel now includes two tabs the user can navigate **history** and **review changes**

#### What does "chunks are squashed together on publish" means?

Once the draft changes are published all the draft edits done before that publish will be squashed and shown as 1 change, that could be expanded, showing also the collaborators for that published.
This means:
If we have the following mutations:
`[draftEdit, publish,  draftEdit, draftEdit, create]`
it will be shown as
`[draftEdit, publish (expandable), create]`
On expand, it will show the two collapsed draftEdit actions

The changes are not removed from the history, they are only visually squashed for easier navigation in the UI.

This is achieved by creating a relation between parents publish action and children draftEdit actions, in the addChunksMetadata function.

| **Collapsed**| **Expanded**|
|:-------:|:-------:|
| Screenshot 2024-09-03 at 11 11 33 | Screenshot 2024-09-03 at 11 11 50 |

https://github.com/user-attachments/assets/bdbf89c0-825f-4d47-a655-0df22c51fee1

### What to review

Are the changes correct? Is something missing?

### Notes for release
Updates to history inspector UI, edits are now collapsed into the Published Event. From and To revision selectors are added into the inspector.


packages/sanity/src/structure/panes/document/documentPanel/header/DocumentPanelHeader.tsx
```diff
@@ -84,9 +78,6 @@ export const DocumentPanelHeader = memo(
     const contextMenuNodes = useMemo(() => menuNodes.filter(isNotMenuNodeButton), [menuNodes])
     const showTabs = views.length > 1
 
-    // Subscribe to external timeline state changes
-    const rev = useTimelineSelector(timelineStore, (state) => state.revTime)
```
pedrobonamin, line 88: This was removed from the Header and moved to [ `ChangesInspector` ](https://github.com/sanity-io/sanity/pull/7462/files#diff-098905f8d208d0d802ad026099778cfe44fe8c2d76b3ec7cd55e18ebf9294aa9R88)

packages/sanity/src/core/store/_legacy/history/useTimelineStore.ts
```diff
@@ -194,7 +194,6 @@ export function useTimelineStore({
       findRangeForSince: (chunk: Chunk) => controller.findRangeForNewSince(chunk),
       loadMore: () => {
         controller.setLoadMore(true)
-        timelineStateRef.current.isLoading = true
```
pedrobonamin, line 197: I noticed this in some cases was blocking the ui in a loading state, which never ended because the load more action was not actually fetching new data, so the controller was not triggering any update, so the `isLoading` returned in line 220 was never resetting to false.

RitaDias (dismissed):
This is _clean_ Pedro, fantastic job! ✨

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1656,7 +1659,7 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
    * Label for determining since which version the changes for timeline menu dropdown are showing.
    * Receives the time label as a parameter (`timestamp`).
    */
-  'timeline.since': 'Since: {{timestamp, datetime}}',
+  'timeline.since': '{{timestamp, datetime}}',
```
jordanl17, line 1662: NIT: Is there any need for this now? I don't imagine the formatting here will be localized

RitaDias (approved):
Great catch, I don't think that Jordan's suggestion is blocking for the approval but also a fair point to bring up. In any case ✅

packages/sanity/src/core/i18n/bundles/studio.ts
pedrobonamin: Good catch, actually I'm removing that in the changes I'm introducing for this in `corel` I think we can clear it there.

PR #7463, opened 2024-09-04
fix(core): update useReleaseHistory to fetch with the version documents

pedrobonamin:
### Description

The `useReleaseHistory` hook was incorrectly constructing the ID of the documents it needs to fetch and construct the history.

Screenshot 2024-09-04 at 09 54 27

### DIFF
--- a/packages/sanity/src/core/releases/tool/detail/documentTable/useReleaseHistory.ts
+++ b/packages/sanity/src/core/releases/tool/detail/documentTable/useReleaseHistory.ts
@@ -1,5 +1,5 @@
 import {useCallback, useEffect, useMemo, useState} from 'react'
-import {getPublishedId, type TransactionLogEventWithEffects, useClient} from 'sanity'
+import {getVersionId, type TransactionLogEventWithEffects, useClient} from 'sanity'
 
 import {getJsonStream} from '../../../../store/_legacy/history/history/getJsonStream'
 import {API_VERSION} from '../../../../tasks/constants'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/releases/tool/detail/documentTable/useReleaseHistory.ts
```diff
@@ -21,14 +24,14 @@ export function useReleaseHistory(bundleDocumentsIds: string[]): {
   const {dataset, token} = client.config()
   const [history, setHistory] = useState<TransactionLogEventWithEffects[]>([])
   const queryParams = `tag=sanity.studio.tasks.history&effectFormat=mendoza&excludeContent=true&includeIdentifiedDocumentsOnly=true`
-
-  const publishedIds = bundleDocumentsIds.map((id) => getPublishedId(id)).join(',')
+  const publishedIds = bundleDocumentsIds.map((id) => getVersionId(id, releaseId)).join(',')
```
RitaDias, line 27: I think we might need to rename the const to documentIds or maybe add a comment to clarify? Just bringing it up since we are technically bringing up a bunch of version ids (which are not publishedIds) 🤔

juice49 (approved):

packages/sanity/src/core/releases/tool/detail/documentTable/useReleaseHistory.ts
juice49: I agree @RitaDias. Based on my understanding from glancing over the code, at least 🙂.

pedrobonamin: TRUE missed that rename! will push the update and merge

RitaDias (approved):

PR #7464, opened 2024-09-04
fix: add padding above the "Push to reload" button instead of below

SimeonGriggs:
### Description

Fixes this padding issue by moving it above the button instead of below. This screenshot shows the current state:

Screenshot 2024-09-04 at 09 59 21

### Notes for release

Adjusts vertical padding around "Push to reload" button

### DIFF
--- a/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
+++ b/packages/sanity/src/core/studio/packageVersionStatus/PackageVersionStatusProvider.tsx
@@ -32,7 +32,7 @@ export function PackageVersionStatusProvider({children}: {children: ReactNode}) 
       id: 'new-package-available',
       title: t('package-version.new-package-available.title'),
       description: (
-        <Stack space={2} paddingBottom={2}>
+        <Stack space={2} paddingTop={2}>
           <Box>
             <Button
               size="large"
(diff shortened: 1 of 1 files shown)


jordanl17 (approved):
Nice! 🚀

juice49:
This looks like a good quick fix. However, it seems like a mistake that we wrap a single `Button` component in a `Stack` at all. I'm assuming the "ready to update" copy was originally intended to appear inside the `Stack` so that the copy and the `Button` component would implicitly be given a gap.

juice49 (approved):

cngonzalez:
> This looks like a good quick fix. However, it seems like a mistake that we wrap a single `Button` component in a `Stack` at all. I'm assuming the "ready to update" copy was originally intended to appear inside the `Stack` so that the copy and the `Button` component would implicitly be given a gap.
> 
> It would be great to remove the redundant `Stack` and `Box` altogether (assuming the `Box` isn't necessary for the layout).

You're right on the money, there were quite a few of last-minute changes on release day which drastically altered it from the original look / feel 😅 

Notify AUS Studio Versions

I'll pull down this branch and see what cruft we can get rid of.

cngonzalez (approved):

PR #7466, opened 2024-09-04
feat(sanity): update document footer for versions

RitaDias:
### Description

- [x] Add discard action on version
- [x] Add version badge (with name) to document footer
- [x] Moved translations from releases plugins to core translations (needed them on the footer)  

Part of the code uses changes that exist in [this PR](https://github.com/sanity-io/sanity/pull/7448) from [history-updates](https://github.com/sanity-io/sanity/tree/history-updates) which are required and useful for this PR

image

____

❗ 

I didn't add the "version" text after because If found it strange. I want to bring it up with Marius that it feels _strange_, the badge should be enough. We don't add specifics about drafts or published text wise. In any case, if anyone or if Marius has a strong opinion I can of course add it

Screenshot 2024-09-04 at 14 28 51

### What to review

Do the changes make sense? Is there a better way of doing it?

### DIFF
--- a/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
+++ b/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
@@ -1,13 +1,16 @@
-import {AddIcon, CheckmarkIcon} from '@sanity/icons'
+import {AddIcon, TrashIcon} from '@sanity/icons'
 import {useTelemetry} from '@sanity/telemetry/react'
 import {useToast} from '@sanity/ui'
 import {type ReactNode, useCallback, useState} from 'react'
 import {filter, firstValueFrom} from 'rxjs'
 import {
+  DEFAULT_STUDIO_CLIENT_OPTIONS,
   getPublishedId,
   getVersionFromId,
   getVersionId,
   isVersionId,
+  Translate,
+  useClient,
   useDocumentOperation,
   useDocumentStore,
   useTranslation,
@@ -41,11 +44,13 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
   const [isInVersion, setIsInVersion] = useState<boolean>(
     () => getVersionFromId(documentId) === globalBundleId,
   )
+  const [isDiscarding, setIsDiscarding] = useState<boolean>(false)
 
   const toast = useToast()
   const {createVersion} = useDocumentOperation(publishedId, documentType, bundleId)
   const {t} = useTranslation()
   const telemetry = useTelemetry()
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
 
   const handleAddVersion = useCallback(async () => {
     if (!documentId) {
@@ -96,7 +101,34 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
     title,
   ])
 
-  /** TODO what should happen when you add a version if we don't have the ready button */
+  const handleDiscardVersion = useCallback(async () => {
+    setIsDiscarding(true)
+    try {
+      await client.delete(documentId)
+
+      toast.push({
+        closable: true,
+        status: 'success',
+        description: (
+          <Translate
+            t={t}
+            i18nKey={'bundle.action.discard-version.success'}
+            values={{title: document.title as string}}
+          />
+        ),
+      })
+
+      setIsInVersion(false)
+    } catch (e) {
+      toast.push({
+        closable: true,
+        status: 'error',
+        title: t('bundle.action.discard-version.failure'),
+      })
+    }
+
+    setIsDiscarding(false)
+  }, [client, documentId, t, toast])
 
   if (archivedAt) return null
 
@@ -105,13 +137,13 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
       data-testid={`action-add-to-${globalBundleId}`}
       text={
         isInVersion
-          ? t('bundle.action.already-in-release', {title})
+          ? t('bundle.action.discard-version', {title})
           : t('bundle.action.add-to-release', {title})
       }
-      icon={isInVersion ? CheckmarkIcon : AddIcon}
-      tone="primary"
-      onClick={handleAddVersion}
-      disabled={isInVersion || creatingVersion}
+      icon={isInVersion ? TrashIcon : AddIcon}
+      tone={isInVersion ? 'caution' : 'primary'}
+      onClick={isInVersion ? handleDiscardVersion : handleAddVersion}
+      loading={creatingVersion || isDiscarding}
     />
   )
 }
--- a/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
+++ b/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
@@ -104,6 +104,7 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
   const handleDiscardVersion = useCallback(async () => {
     setIsDiscarding(true)
     try {
+      // TODO: should we use the document operations for this?
       await client.delete(documentId)
 
       toast.push({
--- a/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
+++ b/packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
@@ -133,17 +133,24 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
 
   if (archivedAt) return null
 
+  const bundleActionButtonProps = isInVersion
+    ? {
+        text: t('bundle.action.discard-version', {title}),
+        icon: TrashIcon,
+        tone: 'caution' as const,
+        onClick: handleDiscardVersion,
+      }
+    : {
+        text: t('bundle.action.add-to-release', {title}),
+        icon: AddIcon,
+        tone: 'primary' as const,
+        onClick: handleAddVersion,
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -78,7 +83,34 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
     setIsInVersion(true)
   }, [documentId, globalBundleId, documentStore.pair, documentType, newVersion, toast, title])
 
-  /** TODO what should happen when you add a version if we don't have the ready button */
+  const handleDiscardVersion = useCallback(async () => {
```
RitaDias, line 86: This is the same logic as what we currently have on the release detail that allows you to discard document versions

packages/sanity/src/core/components/documentStatus/DocumentStatus.tsx
```diff
@@ -67,6 +72,18 @@ export function DocumentStatus({
...
     : versionUpdatedTimeAgo || draftUpdatedTimeAgo
 
+  const {title} = currentGlobalBundle || {}
+
+  const documentStatus = useMemo(() => {
+    if (published && '_id' in published) {
+      return 'published'
+    } else if (version && '_id' in version) {
+      return 'version'
+    }
```
RitaDias, line 82: Unsure if this is the best way to handle this, however, from what I understand we:
- want to find out if it's a published / version / draft. The reason why I do it here and not further down it's two fold:
  - from what I understand from this earlier version of the prototype we will eventually want to display different things / tone despending on what status the document is in. 
  - the other components should be kept more agnostic to this information (UI)

packages/sanity/src/structure/panes/document/statusBar/RevisionStatusLine.tsx
RitaDias, line 1: From [this PR](https://github.com/sanity-io/sanity/pull/7448) from [history-updates](https://github.com/sanity-io/sanity/tree/history-updates)

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBar.tsx
RitaDias, line 1: From [this PR](https://github.com/sanity-io/sanity/pull/7448) from [history-updates](https://github.com/sanity-io/sanity/tree/history-updates)

packages/sanity/src/structure/documentActions/HistoryRestoreAction.tsx
RitaDias, line 1: From [this PR](https://github.com/sanity-io/sanity/pull/7448) from [history-updates](https://github.com/sanity-io/sanity/tree/history-updates)

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusLine.tsx
```diff
@@ -76,6 +79,7 @@ export function DocumentStatusLine({singleLine}: DocumentStatusLineProps) {
           published={editState?.published}
           version={editState?.version}
           singleLine={singleLine}
+          currentGlobalBundle={currentGlobalBundle}
```
RitaDias, line 82: The reason for adding this here is because the `usePaneRouter()` is exported from structure (where this component is). All children components used within are core components and we wouldn't want to do import these hooks within (plus it aids in simplifying components and what they need to care about)

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
RitaDias: There was some discussion (either on slack or on linear) about how we aren't using document operations.
I considered moving this to somewhere that could be imported directly from both, but maybe since we are likely going to change this in the coming future we don't need to do much beyond the TODO?

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentActions.tsx
```diff
@@ -22,11 +21,13 @@ export const DocumentActions = memo(
...
     const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
     const toast = useToast()
-    const {t} = useTranslation(releasesLocaleNamespace)
+    const {t} = useTranslation()
 
     const handleDiscardVersion = async () => {
       try {
         setDiscardStatus('discarding')
+
+        // TODO: should we use the document operations for this?
```
RitaDias, line 30: There was some discussion (either on slack or on linear) about how we aren't using document operations.
I considered moving this to somewhere that could be imported directly from both, but maybe since we are likely going to change this in the coming future we don't need to do much beyond the TODO?

RitaDias: cc @bjoerge

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
RitaDias: cc @bjoerge

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
```diff
@@ -105,13 +138,13 @@ export function BundleActions(props: BundleActionsProps): ReactNode {
       data-testid={`action-add-to-${globalBundleId}`}
       text={
         isInVersion
-          ? t('bundle.action.already-in-release', {title})
+          ? t('bundle.action.discard-version', {title})
```
jordanl17, line 141: COULD: pedantic but I personally feel something like this would avoid any mixups with keeping the props correct:
```
const bundleActionButtonProps = isInVersion ? {
    text: t('bundle.action.discard-version', {title}),
    icon: TrashIcon,
    tone: 'caution',
    onClick: handleDiscardVersion
} : {
    text: t('bundle.action.add-to-release', {title}),
    icon: AddIcon,
    tone: 'primary',
    onClick: handleAddVersion
}

return (
    <Button {...bundleActionButtonProps}
    />
)

```

packages/sanity/src/core/components/documentStatus/DocumentStatus.tsx
jordanl17: I feel having it here feels good - calculating document status in `DocumentStatus` aligns well

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
```diff
@@ -103,23 +103,20 @@ const DocumentStatusBarActionsInner = memo(function DocumentStatusBarActionsInne
...
-                        key={formState.value._id as string}
-                      />
-                    ) : (
-                      <div>
-                        {/* eslint-disable-next-line i18next/no-literal-string */}
-                        <Text>Not a bundle</Text>
-                      </div>
-                    )
-                  }
+                  {isBundleDocument(currentGlobalBundle) && formState?.value?._id ? (
```
pedrobonamin, line 106: A tooltip related to the live edit is shown when hovering in the action.
Screenshot 2024-09-06 at 10 14 18

Probably out of scope for this PR, but I think we should update how the document actions show and probably insert this action through the releases plugin. 
Will follow this in slack

pedrobonamin (approved):
This looks good to me @RitaDias , great work!!

packages/sanity/src/core/bundles/components/panes/BundleActions.tsx
RitaDias: Good point, I'll make this change :)

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentActions.tsx
```diff
@@ -22,11 +21,13 @@ export const DocumentActions = memo(
     const [discardStatus, setDiscardStatus] = useState<'idle' | 'discarding' | 'error'>('idle')
     const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
     const toast = useToast()
-    const {t} = useTranslation(releasesLocaleNamespace)
+    const {t} = useTranslation()
```
jordanl17, line 24: Hmm, I see now we have some labels from the global namespace (the ones you updated eg. 'bundle.action.discard-version.success'), but we have others still from releases eg. discard-version-dialog.title.

We might need to define 2 translation functions:

```
const {t: studioT} = useTranslation()
const {t: releasesT} = useTranslation(releasesLocaleNamespace)
```

Or it looks like it might be possible to combine these:
```
const {t} = useTranslations([releasesLocaleNamespace])
```

jordanl17: Ignore the second way, although `useTranslations` accepts an array, it seems that only the first element gets used

RitaDias: Great find, I have some other places where I do this as well but I forgot to do it here.
Thank you for catching it!

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx
RitaDias: To keep this updated: we discussed this as part of a larger change and out of scope for this PR

PR #7467, opened 2024-09-04
feat(sanity): rename `newVersion` operation to `createVersion`

juice49:
### Description

This branch renames the `newVersion` operation to `createVersion`. This makes it consistent with the other operation names, which are all based on verbs.


pedrobonamin (approved):
Looks good to me, thanks @juice49

PR #7468, opened 2024-09-04
fix(core): update intent link for releases tool

pedrobonamin:
### Description

Fixes an issue with intent link in which the intent was not resolved correctly.
It also replaces the <Button> for a <Link> component, which will allow users to cmd + click to open in new tab. 

Updates the `bundleId` param in the releases tool router for `releaseId`

### What to review

Are the changes correct?

### Testing
Existing tests have been updated to reflect the change.

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/core/releases/plugin/index.ts
+++ b/packages/sanity/src/core/releases/plugin/index.ts
@@ -25,14 +25,14 @@ export const releases = definePlugin({
       name: 'releases',
       title: 'Releases',
       component: ReleasesTool,
-      router: route.create('/', [route.create('/:bundleId')]),
+      router: route.create('/', [route.create('/:releaseId')]),
       canHandleIntent: (intent) => {
         // If intent is release, open the releases tool.
         return Boolean(intent === 'release')
       },
       getIntentState(intent, params) {
         if (intent === 'release') {
-          return {bundleId: params.id}
+          return {releaseId: params.id}
         }
         return null
       },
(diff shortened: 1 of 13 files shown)


packages/sanity/src/core/releases/plugin/index.ts
```diff
@@ -26,12 +26,13 @@ export const releases = definePlugin({
       title: 'Releases',
       component: ReleasesTool,
       router: route.create('/', [route.create('/:bundleId')]),
-      canHandleIntent: (intent, params) => {
-        return Boolean(intent === 'release' && params.slug)
+      canHandleIntent: (intent) => {
+        // If intent is release, open the releases tool.
+        return Boolean(intent === 'release')
```
pedrobonamin, line 31: Instead of not resolving the intent if the params are missing, resolve to the tool at root level.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/__tests__/DocumentPerspectiveMenu.test.tsx
```diff
@@ -90,45 +101,33 @@ describe('DocumentPerspectiveMenu', () => {
...
-      existsInBundle: false,
-    })
-
-    const wrapper = await createTestProvider()
-    render(<DocumentPerspectiveMenu />, {wrapper})
-
-    expect(screen.queryByTestId('button-document-release')).toBeNull()
-  })
-
-  it('should navigate to the release intent when the bundle badge is clicked', async () => {
```
pedrobonamin, line 107: This test doesn't apply anymore, because we are not rendering a button, we are now rendering a `<a>` 
That's also the reason why instead of using a `getByTestId` getter for the button we are now using a `getByRole("link")` getter in this tests

RitaDias (approved):
Great job ✅

PR #7470, opened 2024-09-05
fix(core): hide `versions.*` documents from search, update `getPublishedId` to account for versions

pedrobonamin:
### Description

This PR adds changes to the studio search to not query for documents that start with the id `versions.*` 
We need to exclude that the version documents from the search queries (lists and global search), so users won't see  duplicated results if they have version documents. 

This PR also updates how the `getPublishedId` function works, accounting for the version documents. 
It has been decided that all documents in which the first segment is `versions.` will be considered a version document, the second segment is the `versionId` and the following segments are the document published id. 
The update for the `getPublishedId` function handles this cases.  [See tests for this scenario](https://github.com/sanity-io/sanity/pull/7470/files#diff-fff2de56cfab7676d18627f7a6839b6bbf0abe4ed593a4f1981a9aad171b1418R63-R70) 

Now, when visiting a document in the URL with the id `versions.foo.bar` when the builder defines the document id inside the pane it will resolve to `id=bar` , so the user will be able to interact with the document.
E.g: visiting this document won't crash the studio after this changes:
 - `preview branch` https://test-studio-git-corel-158.sanity.dev/test/structure/author;versions.rbtQUMGgH.5a9b3da1-9398-49d3-b593-9fa89f0322ff
 - `next branch` : https://test-studio.sanity.dev/test/structure/author;versions.rbtQUMGgH.5a9b3da1-9398-49d3-b593-9fa89f0322ff

### What to review

Are this changes correct?

### Testing

Added new tests for the updates in the draftUtils functions, search tests are also updated to follow the new rules

### Notes for release

n/a not public facing yet

### DIFF
--- a/packages/sanity/src/core/util/draftUtils.ts
+++ b/packages/sanity/src/core/util/draftUtils.ts
@@ -74,7 +74,7 @@ export function getIdPair(id: string): {draftId: DraftId; publishedId: Published
 
 /** @internal */
 export function isPublishedId(id: string): id is PublishedId {
-  return !isDraftId(id)
+  return !isDraftId(id) && !isVersionId(id)
 }
 
 /** @internal */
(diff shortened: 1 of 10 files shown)


packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -65,17 +74,53 @@ export function getIdPair(id: string): {draftId: DraftId; publishedId: Published
...
 
 /** @internal */
 export function getDraftId(id: string): DraftId {
   return isDraftId(id) ? id : ((DRAFTS_PREFIX + id) as DraftId)
 }
 
+/**  @internal */
+export function getVersionId(id: string, bundle: string): string {
+  if (isVersionId(id)) {
+    const [_versionPrefix, versionId, ...publishedId] = id.split('.')
```
juice49, line 88: Just a tiny thought: it might be nice to split on the `PATH_SEPARATOR` constant instead of `"."`.
```suggestion
    const [_versionPrefix, versionId, ...publishedId] = id.split(PATH_SEPARATOR)
```

juice49 (dismissed):
Looks good to me! Added one tiny comment.

packages/sanity/src/core/util/draftUtils.ts
pedrobonamin: Absolutely, thanks!

juice49 (approved):

Issue #7456, opened 2024-09-01
Slow input keystroke lag in Studio


eldoy:
**Describe the bug**

Every keystroke is slow and laggy in Sanity Studio. I'm quite a fast typist, and this totally makes the user experience terrible. It seems like every keystroke is synced to the cloud?

**To Reproduce**

Type fast in any text input field and it lags.

**Expected behavior**

Absolutely no lag.

**Which versions of Sanity are you using?**

@sanity/cli (global)          3.56.0 (up to date)
@sanity/eslint-config-studio   4.0.0 (up to date)
@sanity/vision                3.56.0 (up to date)
sanity                        3.56.0 (up to date)

**What operating system are you using?**

Darwin v2.local 19.6.0 Darwin Kernel Version 19.6.0: Tue Jun 21 21:18:39 PDT 2022; root:xnu-6153.141.66~1/RELEASE_X86_64 x86_64

**Which versions of Node.js / npm are you running?**

10.8.2
v20.17.0

FancyJeff:
Experiencing this also, this issue is particularly bad when interacting with an object nested within the portable text editor.

thomasm:
Same here. The keystroke lag is very noticeable, especially when typing in text fields within items in my page builder array.

jordanl17:
Hi, @eldoy - thanks for highlighting this issue.

Are you experiencing this only when running the studio locally, or also on a deployed studio?

eldoy:
@jordanl17 It's just as bad for both. Keep holding backspace and it's even more evident. I'm having a hard time recommending Sanity to my clients when such a basic annoyance exists. Are you not seeing it?

jordanl17:
> @jordanl17 It's just as bad for both. Keep holding backspace and it's even more evident. I'm having a hard time recommending Sanity to my clients when such a basic annoyance exists. Are you not seeing it?

@eldoy apologies you are experience such poor performance. We are aware of some rendering issues and are continually trying to work through these. We have some high priority improvements in the works right now, which it looks like will directly help mitigate the issues you've encountered. These should be released shortly in the coming week/s. Please ensure you are always on the latest studio to capture these improvements when they drop. 

[...]

PR #7471, opened 2024-09-06
fix(core): ensure reconciliation for perf

ricokahler:
### Description

The following fixes a performance degradation where part of the form node was not immutable reconciled resulting in slower renders.

### What to review

Ensure the rationale checks out. This should leave the existing fix that #7154 introduced as well. 

### Testing

I ran the new eFPS test suite (#7407) on the commit (from here #7154) that introduced this performance degradation and got the following:

CleanShot 2024-09-05 at 23 17 10@2x

After applying this fix, this is the resulting eFPS.

CleanShot 2024-09-05 at 23 16 44@2x

Note that the existing test that was added in #7154 still passes so I assume this still leaves that fix in place.

### Notes for release

- Fixes a performance regression that affects large documents.


ryanbonial (dismissed):

runeb:
Fixes #7456

ricokahler:
this went out in v3.57.1!

PR #7473, opened 2024-09-06
fix(core): Fix Title in "Untitled was published" toast

ryanbonial:
### Description
This PR fixes a bug where a success message shows "Untitled was published" when publishing (and unpublishing) a document with a title. This is especially prevalent on documents with an `image` field. 

It adds an additional check for a title in the current document title when the title is not available on the document preview title.

### What to review
Ensure that the using the document pane value directly as a last resort for a title is a safe approach.

### Testing
* Create a new document with an image field
* Add a `title` and then publish the document
* Ensure that the success toast includes the title that you typed and not "Untitled was published" 
* Unpublish the document
* Ensure that the success toast includes the title that you typed and not "Untitled was unpublished" 

### Notes for release
* Fixes the "Untitled was published" success toast issue

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentOperationResults.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentOperationResults.tsx
@@ -1,5 +1,5 @@
 import {useToast} from '@sanity/ui'
-import {memo, useEffect, useRef} from 'react'
+import {memo, useEffect, useMemo, useRef} from 'react'
 import {Translate, useDocumentOperationEvent, useTranslation} from 'sanity'
 
 import {usePaneRouter} from '../../components'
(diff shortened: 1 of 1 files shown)


packages/sanity/src/structure/panes/document/DocumentOperationResults.tsx
```diff
@@ -11,13 +11,25 @@ const IGNORE_OPS = ['patch', 'commit']
...
+  const {documentId, documentType, value: documentPaneValue} = useDocumentPane()
+  const documentTitleInfo = useDocumentTitle()
+  let title = documentTitleInfo.title
+  const titleError = documentTitleInfo.error
   const event: any = useDocumentOperationEvent(documentId, documentType)
   const prevEvent = useRef(event)
   const paneRouter = usePaneRouter()
   const {t} = useTranslation(structureLocaleNamespace)
 
+  if (
```
ricokahler, line 23: this may be more react-esque if it was calculated. e.g.

```
  const title = useMemo(() => {
    // If title isn't set from document preview, use the title from the document pane value
    if (
      !documentTitleInfo.title &&
      !titleError &&
      !IGNORE_OPS.includes(event?.op) &&
      typeof documentPaneValue.title === 'string' &&
      event?.type === 'success'
    ) {
      return documentPaneValue.title
    }
    return documentTitleInfo.title
  }, [documentTitleInfo.title, titleError, event, documentPaneValue.title])
```

ricokahler (approved):
looks good. left a comment on react conventions but nothing blocking

packages/sanity/src/structure/panes/document/DocumentOperationResults.tsx
ryanbonial: 👍 Thanks for that suggestion, I was going back and forth on the decision to memoize it.

ricokahler: its not too important to memoize it since the value isn't an object reference and is cheap to calculate but it kinda feels better to useMemo than to have an IIFE e.g. `const title = (() => {})()` do the same thing. mutating the variable is the thing to avoid in react renders

ryanbonial: Yeah, I like the memo approach better. I went ahead and made that change.

ricokahler (approved):

PR #7476, opened 2024-09-09
feat(core): specific error when dev server is stopped

jordanl17:
### Description
Currently when the dev server is stopped, when the next error is thrown (for example a new lazy bundle is fetched), then an error `failed to fetch dynamic import` is thrown and displayed. It doesn't make it clear that this has happened because dev server has gone away.

Now, immediately once the dev server is stopped, the studio will show an error boundary stating such.

Screenshot 2024-09-09 at 11 16 08

### What to review
* The logic to listen to the vite hmr server in `DevServerStopped.tsx`
* How `DevServerStopped` is lazy imported to avoid adding this to the bundle when not in development mode
* `StudioLayout.tsx` throwing the dev server stopped error which is causing in studio error boundary, from where the specific screen is shown


packages/sanity/src/core/studio/StudioErrorBoundary.tsx
```diff
@@ -19,6 +19,12 @@ import {CorsOriginError} from '../store'
 import {isRecord} from '../util'
 import {CorsOriginErrorScreen, SchemaErrorsScreen} from './screens'
 
+const DevServerStoppedErrorScreen = lazy(() =>
+  import('./DevServerStopped').then((DevServerStopped) => ({
```
juice49, line 23: Something that crossed my mind: if we are lazily importing this component only after we detect the dev server has gone away, is there any scenario in which it won't be available at that point in time?

packages/sanity/src/core/studio/DevServerStopped.tsx
```diff
@@ -0,0 +1,74 @@
...
+    this.isDevServerStoppedError = true
+  }
+}
+
+const useDetectDevServerStopped = () => {
+  const [devServerStopped, setDevServerStopped] = useState(false)
+  const serverIsReadyRef = useRef(false)
+
+  useEffect(() => {
+    const url = `ws://${window.location.hostname}:${window.location.port}/`
```
juice49, line 24: Could it be better to [reuse Vite's existing WebSocket connection](https://vitejs.dev/guide/api-hmr#hot-on-event-cb) by utilising its `vite:ws:disconnect` event? It can be subscribed to like this: `import.meta.hot.on('vite:ws:disconnect', () => {})`.

Espen uses events like this in his [use-hot-module-reload](https://github.com/rexxars/use-hot-module-reload/blob/main/use-hot-module-reload/src/use-hot-module-reload.mts#L81) package, which Studio already consumes.

This also got me thinking: `useHotModuleReload` is already used inside `packages/sanity/src/core/studio/StudioErrorBoundary.tsx`. Could there be an argument to modify the `use-hot-module-reload` package so that it returns a state object containing the connection state?

packages/sanity/src/core/studio/StudioLayout.tsx
```diff
@@ -173,6 +179,7 @@ export function StudioLayoutComponent() {
       {/* By using the tool name as the key on the error boundary, we force it to re-render
           when switching tools, which ensures we don't show the wrong tool having crashed */}
       <StudioErrorBoundary key={activeTool?.name} heading={`The ${activeTool?.name} tool crashed`}>
+        {process.env.NODE_ENV === 'development' && <DetectDevServerStopped />}
```
juice49, line 182: If we only detect presence of the Vite dev server, should we add an additional check here to avoid loading and rendering `DetectDevServerStopped` if some other dev server is in use? (e.g. Studio embedded in a Next.js app).

`DetectDevServerStopped` seems like it will simply do nothing in this scenario, but omitting it altogether might avoid potential issues for non-Vite users.

juice49 (commented):
Really cool feature 🙌. Left some thoughts about non-Vite support and minimising connections and custom code.

packages/sanity/src/core/studio/StudioErrorBoundary.tsx
jordanl17: Hmm that's interesting. So I don't believe so- we would have to have loaded this module in the first place to throw such an error. I feel it worthy a code comment though as for a few mins I was wondering how this was even working 😖

packages/sanity/src/core/studio/DevServerStopped.tsx
jordanl17: I did try using `import.meta.hot.on` but could never get any events to trigger. 

But I can look into it again now I know it's being used elsewhere

jordanl17: @juice49 looked back into this, I think because I'd previously experimented with `import.meta.hot.on` inside lazily loaded modules, that was why I wasn't finding it to work. Can confirm it does work nicely.

I've updated this hook to use that approach.

Agreed, this would be nice in the `use-hot-module-reload` - I'll create a separate task for this since it looks like there are some tech debt improvements that can be made in that package for later versions of vite (and I'll also need to follow up with Espen to get this included and then released). Don't want to block this change right now. Hope that sounds ok with you?

packages/sanity/src/core/studio/StudioErrorBoundary.tsx
```diff
@@ -19,6 +19,17 @@ import {CorsOriginError} from '../store'
 import {isRecord} from '../util'
 import {CorsOriginErrorScreen, SchemaErrorsScreen} from './screens'
 
+/**
```
jordanl17, line 22: @juice49 added this comment to clarify that although on the face it looks like this might not be available, in practice it always will be when it's needed

packages/sanity/src/core/studio/StudioLayout.tsx
```diff
@@ -18,6 +18,14 @@ import {
 import {StudioErrorBoundary} from './StudioErrorBoundary'
 import {useWorkspace} from './workspace'
 
+const DetectViteDevServerStopped = lazy(() =>
+  import('./ViteDevServerStopped').then((DevServerStopped) => ({
+    default: DevServerStopped.DetectViteDevServerStopped,
+  })),
+)
+
+const detectViteDevServerStopped = import.meta.hot && process.env.NODE_ENV === 'development'
```
jordanl17, line 27: @juice49 tightened the conditional here to only being for vite servers (`import.meta.hot` would typically only exist for vite). What do you think?

packages/sanity/src/core/studio/ViteDevServerStopped.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const useDetectViteDevServerStopped = () => {
+  const [devServerStopped, setDevServerStopped] = useState(false)
+
+  const markDevServerStopped = useCallback(() => setDevServerStopped(true), [])
+
+  useEffect(() => {
+    // no early return to optimize tree-shaking
+    if (isViteServer(serverHot)) {
+      serverHot.on('vite:ws:disconnect', markDevServerStopped)
```
jordanl17, line 30: @juice49 using `import.meta.hot.on`. In our version of vite (4.x) there is no `.off`, however there is in 5.x. Don't think it's a big issue here, since restarting the server will trigger a refresh on the client. But does raise a question around us upgrading perhaps... one for another time

jordanl17:
> **New dependencies detected.** Learn more about [Socket for GitHub ↗︎](https://socket.dev?utm_medium=gh)
> 
> Package	New capabilities	Transitives	Size	Publisher
> [npm/babel-plugin-transform-vite-meta-hot@1.0.0](https://socket.dev/npm/package/babel-plugin-transform-vite-meta-hot/overview/1.0.0)	None	[`+1`](https://socket.dev/npm/package/babel-plugin-transform-vite-meta-hot/dependencies/1.0.0)	259 kB	[osrg](https://socket.dev/npm/user/osrg)
> [**View full report↗︎**](https://socket.dev/dashboard/org/sanity-io/diff/cb2e37d6-d39a-4206-9192-01338c0e20b7/ba27a304-6720-46f6-a9b1-3a9c95a819a4)

Added to transform `import.meta` in jest cjs environment

bjoerge (dismissed):
Great work, @jordanl17!

packages/sanity/src/core/studio/StudioLayout.tsx
juice49: I guess other tooling beyond Vite could implement `import.meta.hot`, but it seems like a pretty good indication that the environment supports HMR to me 🙂.

packages/sanity/src/core/studio/DevServerStopped.tsx
juice49: Awesome!! 🥳 Sounds great to me.

juice49:
I checked it out to test, but I'm not sure this is working exactly as intended. As soon as I stop the dev server, an uncaught `ViteDevServerStoppedError` is caught by the top-level Studio error boundary. It doesn't seem to show the specialised "dev server stopped" error message:

Screenshot 2024-09-30 at 11 32 09

When subsequently navigating elsewhere (with dev server still stopped), I saw the `Failed to fetch dynamically imported module` error:

Screenshot 2024-09-30 at 11 32 31

packages/sanity/src/core/studio/ViteDevServerStopped.tsx
jordanl17: I'm not concerned around multiple subscriptions since starting the server again causing a remounting of the entire app, clearing all existing listeners

jordanl17:
> I checked it out to test, but I'm not sure this is working exactly as intended. As soon as I stop the dev server, an uncaught `ViteDevServerStoppedError` is caught by the top-level Studio error boundary. It doesn't seem to show the specialised "dev server stopped" error message:
> 
> Screenshot 2024-09-30 at 11 32 09
> When subsequently navigating elsewhere (with dev server still stopped), I saw the `Failed to fetch dynamically imported module` error:
> 
> Screenshot 2024-09-30 at 11 32 31

@juice49 right you are, I'd forgotten to commit the final commit to rename a few things which was causing this. Note however that navigating away from the initial 'dev server stopped' error will still show the dynamic import error. It would be possible to rectify this, but I can't think of how besides complicating the current setup, which doesn't feel appropriate given this is meant to just be an incremental dev experience improvement

juice49 (approved):
Brilliant! Looks great to me 🥳.


(Noted and agree with you on the dynamic import error; just wanted to mention it in case it provided useful context).

PR #7498, opened 2024-09-12
perf(core): memoize `prepareFormState`

ricokahler:
### Description

This PR introduces memoization to the `prepareFormState` function to significantly improve performance, especially for complex documents with multiple large objects in arrays.

Key changes:
- Implemented memoization for `prepareFormState` and its sub-functions for improved performance for complex documents, with up to 10x improvements in some cases
- Removed computation of `hidden` and `readOnly`. Their `StateTree`s are calculated outside of `prepareFormState` to remove the dependency of the `document` to the inputs of `prepareFormState`.

This addresses performance issues with form state computation, particularly for large and complex documents.

Note: Dependes on #7407 

### What to review

- Review the memoization implementation in `prepareFormState`
- Check the implementation of conditionally hidden and readOnly states
- Verify performance improvements, especially for complex documents and documents with large Portable Text fields

### Testing

Automated tests are not yet complete. Manual testing has been performed:

- Compared performance between memoized version and non-memoized version (see attached screenshots)
- Tested with complex documents, including the synthetic fixture
- Verified improvements in eFPS for various document types

Before:

CleanShot 2024-09-11 at 22 34 58@2x

After:

CleanShot 2024-09-11 at 22 35 14@2x

TODO: Complete automated tests

### Notes for release

This change introduces significant performance improvements for form state computation in Sanity, especially for complex documents:

- Up to 10x performance improvement for documents with multiple large objects in arrays
- Noticeable improvements for documents with large Portable Text fields
- Implemented memoization for form state computation

Limitations:
- Computing conditionally `hidden` and `readOnly` states still adds significant computation time, especially for very large documents
- React rendering remains a bottleneck for documents with large Portable Text fields

Note: This is a significant performance enhancement that will benefit most users, especially those working with complex document structures.

### DIFF
--- a/packages/sanity/src/core/form/store/formState.ts
+++ b/packages/sanity/src/core/form/store/formState.ts
@@ -250,7 +250,7 @@ export function createPrepareFormState({
         collapsedFieldSets: getId(parent.collapsedFieldSets?.children?.[field.name]),
         collapsedPaths: getId(parent.collapsedPaths?.children?.[field.name]),
         currentUser: getId(parent.currentUser),
-        fieldGroupState: getId(parent.fieldGroupState?.children?.[field.name]),
+        fieldGroupState: getId(parent.fieldGroupState),
         hidden:
           parent.hidden === true ||
           parent.hidden?.value ||
--- a/packages/sanity/src/core/form/store/formState.ts
+++ b/packages/sanity/src/core/form/store/formState.ts
@@ -22,7 +22,7 @@ import {castArray, isEqual as _isEqual} from 'lodash'
 
 import {type FIXME} from '../../FIXME'
 import {type FormNodePresence} from '../../presence'
-import {EMPTY_ARRAY, isRecord} from '../../util'
+import {EMPTY_ARRAY, EMPTY_OBJECT, isRecord} from '../../util'
 import {getFieldLevel} from '../studio/inputResolver/helpers'
 import {ALL_FIELDS_GROUP, MAX_FIELD_DEPTH} from './constants'
 import {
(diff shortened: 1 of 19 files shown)


packages/sanity/src/core/form/store/utils/getId.ts
```diff
@@ -0,0 +1,28 @@
+import {nanoid} from 'nanoid'
+
+const idCache = new WeakMap<object, string>()
+const undefinedKey = Symbol('GetIdUndefined')
+const nullKey = Symbol('GetIdNull')
+
+export function getId(value: unknown): string {
```
bjoerge, line 7: a comment with a brief explanation of what this does would be nice

packages/sanity/src/core/form/store/utils/immutableReconcile.ts
```diff
@@ -1,76 +1,97 @@
...
+
+    if (Array.isArray(prev) && Array.isArray(curr)) {
+      if (prev.length !== curr.length) return curr
+
+      const reconciled = curr.map((item, index) => immutableReconcile(prev[index], item))
+      if (reconciled.every((item, index) => item === prev[index])) return prev
+      return reconciled as T
+    }
+
+    // skip these, they're recursive structures and will cause stack overflows
```
bjoerge, line 43: The previous version of immutableReconcile tracks visited nodes and is resillient against circular structures. Was there a particular reason to remove this logic? How confident can we be that this will never have to deal with circular structures beyond the ones we explicitly check for, like schema type?

packages/sanity/src/core/form/store/formState.ts
```diff
@@ -128,937 +191,1106 @@ function isChangedValue(value: any, comparisonValue: any) {
...
-    undefined
-
-  const itemState = prepareObjectInputState(
-    {
-      schemaType: itemType,
-      level: itemLevel,
-      document: parent.document,
-      value: arrayItem,
+  return Object.assign(
+    function prepareFormState({
```
bjoerge, line 1251: Assigning properties to a function like this feels a little bit peculiar to me – any drawbacks to keeping `prepareFormState` as an individual property on the returned object here?

packages/sanity/src/core/form/store/formState.ts
```diff
@@ -128,937 +191,1106 @@ function isChangedValue(value: any, comparisonValue: any) {
...
+  collapsedFieldSets: StateTree<boolean> | undefined
+  changesOpen?: boolean
+}
 
-  if (isObjectSchemaType(field.type)) {
-    const fieldValue = parentValue?.[field.name]
-    const fieldComparisonValue = isRecord(parentComparisonValue)
-      ? parentComparisonValue?.[field.name]
-      : undefined
+export interface PrepareFormState {
```
bjoerge, line 223: see later comment about this being a function with properties - wondering if it would be easier to reason about this if it was a regular object with `prepareRoot`, or `prepareDocument` or smth

bjoerge (commented):
Amazing work and really impressive numbers @ricokahler – left a few comments, mostly minor stuff.

packages/sanity/src/core/form/store/utils/immutableReconcile.ts
ricokahler: The reason I removed it was that all calls to immutableReconcile no longer took in the schema type and therefore no  circular structures were passed into the function. I did add in a special check for schema types in case that ever was the case in the future but I'm confident that we won't see circular structures in this function any more.

packages/sanity/src/core/form/store/formState.ts
ricokahler: My thought process here is that the rest of the function inside of `prepareFormState` are internal. The only reason i'm exposing them is to wrap them with `jest.fn` so i can call `expect(/* .. .*/).toHaveBeenCalledTimes(/* ... */)`.  I think i might prefix these with an underscore of something to denote that.

cngonzalez (approved):
re-iterating Bjoerge's approval after the rebase and change to make tests pass!

PR #7500, opened 2024-09-12
chore(releases): moving from bundles to releases directory

jordanl17:
### Description
Migrating the `core/bundles` directories to `core/bundles`.

Note that the methods in these files haven't been changed, only the location of the files.

The directory that previously existed in `core/releases/components` included all the components for the releases plugin (since previously `core/releases` included only the code for the plugin). This directory has therefore been moved to `core/releases/tool/components`, which feels sensible given that the exports from these components are only used in the `tool` directory.

### Testing
Tests mostly involved just updating the path for mocks. 
However there were particular issues with `ReleasesSummary` and `ReleasesReview` - for inexplicable reasons, the `SanityDefaultPreview` component wasn't receiving it's correct imports, and despite considerable investigation, it seems that there is perhaps a cyclic dependency/mock somewhere which I couldn't track down. Therefore, instead a mock has been added to these 2 tests to cover the issues instead - the tests here are still valid, given that extraneous logic is being mocked and not the parts that are relevant to the testing scope.

### Notes for release
N/A


RitaDias (approved):
Great job Jordan, this makes sense to me :)

pedrobonamin (approved):
Thanks for this @jordanl17  !
Looks good to me, would like if we could spend some time debugging the mock issues in a follow up, specially before merging to `next` .

This will probably be fixed once we switch to `vitest`.

Issue #6879, opened 2024-06-07
Sanity Studio Rich Text Editor does not properly close 3rd level ordered list tags


slegay:
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

When writing nested lists, if a sub-list ends with a 3rd degree tabbed list, the next number is wrong. It seems that an ol tag is not getting properly closed.

**To Reproduce**

Create this ordered list in Sanity portable text editor editor, as shown in the screenshot.

<img width="478" alt="image" src="https://github.com/sanity-io/sanity/assets/279348/709ebf7c-e922-4bda-a8c6-ab22d21ba484">

**Expected behavior**

c should be a

**Screenshots**
<img width="478" alt="image" src="https://github.com/sanity-io/sanity/assets/279348/709ebf7c-e922-4bda-a8c6-ab22d21ba484">

**Which versions of Sanity are you using?**

@sanity/cli (global)       3.45.0 (up to date)
@sanity/dashboard           3.1.6 (up to date)
@sanity/google-maps-input   4.0.1 (up to date)
@sanity/ui                  2.3.1 (up to date)
@sanity/uuid                3.0.2 (up to date)
@sanity/vision             3.45.0 (up to date)
sanity                     3.45.0 (up to date)

**What operating system are you using?**

Ubuntu

**Which versions of Node.js / npm are you running?**

9.8.1
v18.18.2

**Additional context**

The problem goes away when adding a level 2 item after finishing a level 3 ordered list, and before starting a new level 1 list item, as shown below:

<img width="439" alt="image" src="https://github.com/sanity-io/sanity/assets/279348/1cc42962-77d3-4963-ba1f-dbe157e649a8">

joelschneider:
+1

The issue is even broader I believe.

If you have more than one simple ordered list in a portable text editor, the following ordered lists continue the numbering from the previous ordered list.

Example:

1. ...
2. ...
3. ...

[Other content]

4. ...
5. ...

Very annoying tbh, as this renders this basic functionality useless if you have more than one ordered list in an article.

christianhg:
Hi there! Just want to let you all know that the problem that @joelschneider describes should be fixed by [v3.53.0](https://github.com/sanity-io/sanity/releases/tag/v3.53.0). Related PR: https://github.com/sanity-io/sanity/pull/7286

However, the problem described in the OP is not fixed yet. It's in our backlog though!

[...]

PR #7506, opened 2024-09-14
fix(core): correctly reset PTE list counts after non-numbered lists and in sublists

phattran2905:
### Description
I find that the numbered list is incorrectly indexed in two places.

####  Case 1: On Sub-list items
The indices of list items starting at `level 2` to `level 9` do not start over even though they are separate lists. Instead, they continue on the counter from the previous list. This is already mentioned at #6879.

##### Before
```
1. L1
  a. L2
    i. L3
2. L1
  b. L2 // This is supposed to be 'a.'
    ii.L3 // This is supposed to be 'i.'
```
##### After
```
1. L1
  a. L2
    i. L3
2. L1
  a. L2
    i.L3
```

#### Case 2: Created right after a bullet list
If a numbered list is created right after a bullet list, the indices continue from the previous list counter. It is caused by using the `counter-reset` instead of `counter-set`, making the reset step ineffective.

##### Before
```
1. L1
2. L1
Some random text
● Bullet list item 1
● Bullet list item 2
3. L1 // This is supposed to reset to '1'
4. L1 // This is supposed to '2'
```
##### After
```
1. L1
2. L1
Some random text
● Bullet list item 1
● Bullet list item 2
1. L1
2. L1
```

### What to review
Overall, the counter is initialized under the `.pt-editable` element and will be reset to `0` at every element that is not a numbered list item. In addition, the counter increases by `1` at each numbered list item until the list ends. On the other hand, it resets the counter for elements not part of the numbered list.

In order to fix __Case 1__, I removed the code on `line:98` and let the rule set for non-numbered list items take effect on bullet list items as well. 

In order to fix __Case 2__, I decided to target the `level 1` list item and reset the counter all levels between `2 - 9`.

### Testing
I test it manually by creating various lists which have multi-level nested items. In addition, I intend to add more different block elements between the lists to ensure the counter is correct.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/Editor.styles.tsx
@@ -101,15 +101,16 @@ export const EditableWrapper = styled(Card)<{$isFullscreen: boolean; $readOnly?:
       margin-top: ${({theme}) => theme.sanity.space[2]}px;
     }
 
+    /* Reset the list count if the element is not a numbered list item */
+    & > :not(.pt-list-item-number) {
+      counter-set: ${TEXT_LEVELS.map((l) => createListName(l)).join(' ')};
+    }
+
     /* Reset the list count all the sub-list items */
     & > .pt-list-item-number.pt-list-item-level-${TEXT_LEVELS[0]} {
       counter-set: ${TEXT_LEVELS.slice(1)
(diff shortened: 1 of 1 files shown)


christianhg:
This is fantastic. Thanks a lot! I just made a small amendment (fixup) to make the diff even smaller.

christianhg (approved):

PR #7517, opened 2024-09-17
fix(core): disable scheduledPublishing and tasks if `/features` returns error

pedrobonamin:
### Description

Some users have reported that they see the upsell ui even though the feature is enabled for the project.
That is due to the `features` api returning an error if the user doesn't have the right permissions to query it.

This PR adds the same check for `scheduledPublishing` and `tasks` plugins.
In `comments` we are currently doing that check here https://github.com/sanity-io/sanity/blob/edx-1599/packages/sanity/src/core/comments/hooks/useResolveCommentsEnabled.ts#L42-L43

### What to review

Are this changes ok?

### Testing
New tests have been introduced to guard this behavior.

### Notes for release

Disable schedule publishing and tasks plugin if user doesn't have the right access.

### DIFF
--- a/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
@@ -81,8 +81,8 @@ export default function Tool() {
       <Container width={1} paddingTop={4}>
         <Box paddingTop={4} paddingX={4}>
           <ErrorCallout
-            description="You do not have permission to edit schedules."
-            title="Insufficient permissions"
+            description="Something went wrong loading permissions, please try again."
+            title="Permissions check failed"
           />
         </Box>
       </Container>
--- a/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
@@ -1,4 +1,4 @@
-import {Box, Flex, Text, useTheme} from '@sanity/ui'
+import {Box, Container, Flex, Text, useTheme} from '@sanity/ui'
 import {parse} from 'date-fns'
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
```diff
@@ -74,6 +76,19 @@ export default function Tool() {
     }
   }
 
+  if (!enabled) {
+    return (
+      <Container width={1} paddingTop={4}>
+        <Box paddingTop={4} paddingX={4}>
+          <ErrorCallout
+            description="You do not have permission to edit schedules."
```
pedrobonamin, line 84: ℹ️  We don't have translations in `scheduledPublishing` 
Instead of showing the same error that we show when the request fails we add an specific not enabled error.
**Current error **
Screenshot 2024-09-17 at 17 23 06

**Updated error**
Screenshot 2024-09-17 at 17 24 34

rexxars (commented):
Thanks for raising this! I've made changes to the backend API to avoid needing more permissions to read features, but this will still be nice for other errors.

pedrobonamin:
> Thanks for raising this! I've made changes to the backend API to avoid needing more permissions to read features, but this will still be nice for other errors.

Thanks for reviewing @rexxars 

So I think it would be better in this case to update the message to [this](https://github.com/sanity-io/sanity/pull/7517/commits/dba6a31f0b6bd8ef9b242b496c6af737abd9e6e9) 
```diff
-            description="You do not have permission to edit schedules."
+            description="Something went wrong loading permissions, please try again."
-            title="Insufficient permissions"
+            title="Permissions check failed"
```

rexxars (approved):

PR #7520, opened 2024-09-18
feat(sanity): refine error UI

juice49:
### Description

This branch introduces UI refinements to the `StudioErrorBoundary` component. The aim is to make this error UI less intimidating for users.

- Hides error stack in production builds (this remains in development builds).
- Adds copy explaining that an error occurred.
- Adds "Copy error details" action.
- Subtle UI refinements.
  - Removes red tint.
  - Adds spacing.
  -  Centres content.

Additionally, the "Copy error details" action has been added to the document list error UI.

##### Before

before-global

##### After (production build)

after-global-production

##### After (development build)

after-global-development

##### Before

before-document-list

##### After

after-document-list

### What to review

- Is the new copy okay?
- `StudioErrorBoundary` is rendered outside of `LocaleProvider`, and therefore the text cannot be internationalised. Is that okay?
  - For the same reason, `ErrorActions` (which is now used inside document lists), cannot be internationalised (the retry button in document lists _was previously internationalised_). Is that okay?

### Testing

- Added unit tests for error serialization.
- Triggered global errors by disconnecting network, and by throwing error from custom Studio layout component.
- Triggered document list errors by disconnecting network, and by configuring document list with invalid GROQ filter expression.

### Notes for release

- The global error UI has been refined so that it is clearer to users and less visually distressing.
- The global error UI now only displays the stack trace in development builds.
- A "Copy error details" action has been added to the global error UI and the document list error UI. This allows users to easily share the error stack trace with developers.
- Please note that these new strings cannot be internationalized because the global error UI is rendered outside of the locale provider.


packages/sanity/src/structure/panes/documentList/DocumentListPaneContent.tsx
```diff
@@ -198,17 +197,7 @@ export function DocumentListPaneContent(props: DocumentListPaneContentProps) {
                   components={{Code: ({children}) => <code>{children}</code>}}
                 />
               </Text>
-
-              {onRetry && (
-                <Box>
-                  <Button
-                    icon={SyncIcon}
-                    onClick={onRetry}
-                    text={t('panes.document-list-pane.error.retry-button.text')}
```
cngonzalez, line 207: Should this string be removed from packages/sanity/src/structure/i18n/resources.ts since it is no longer being used anywhere (as far as I can tell?)

cngonzalez (approved):
This looks reasonable to me, and a great addition! Just one minor nit.

packages/sanity/src/structure/panes/documentList/DocumentListPaneContent.tsx
juice49: Great point! I was a little concerned about removing a string that has already been translated everywhere in favour of something hardcoded 😔. But I think you're right that it's pragmatic to remove it.

PR #7526, opened 2024-09-19
fix(structure): live edit on draft documents

RitaDias:
### Description

When editing a live document that had a draft version of it, the users were being unable to edit the document.
Worst, when using the PTE they would be able to edit and add text, the would get a notification that the document was saved but when refreshing the data would be lost.

The current fix makes the document readOnly when the document is liveEdit AND a draft is being displayed.
The solution is for the user to publish the draft (that shouldn't exist).

It also allows to publish or discard the open draft

Before

https://github.com/user-attachments/assets/b33b7d6c-b3b5-45b9-8aa3-786b125e954b

After

https://github.com/user-attachments/assets/3c8ae367-87d1-4340-8951-76530078cf24

### What to review

The code and if the UI makes sense. I considered adding a drop down menu like we have for the extended actions in the footer but decided that it doesn't make sense to obscure the choices in this case

### Testing

You can manually test it by going to some older documents of the type SImpleBlock (which will be dropped before merging)
There is also an e2e test set up

### Notes for release

Fixes issue with drafts on live edit schema

### DIFF
--- a/dev/test-studio/schema/standard/portableText/simpleBlock.tsx
+++ b/dev/test-studio/schema/standard/portableText/simpleBlock.tsx
@@ -31,6 +31,7 @@ const myStringType = defineArrayMember({
 })
 
 export default defineType({
+  liveEdit: true,
   name: 'simpleBlock',
   title: 'Simple block',
   type: 'document',
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -95,6 +95,12 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   'banners.deleted-document-banner.text': 'This document has been deleted.',
   /** The text content for the deprecated document type banner */
   'banners.deprecated-document-type-banner.text': 'This document type has been deprecated.',
+  /** The text content for the live edit document when it's a draft on how to solve it */
+  'banners.live-edit-draft-banner.explanation':
+    'Disable "live edit" in the schema to publish the draft. Once published, "live edit" can be reactivated.',
+  /** The text content for the live edit document when it's a draft */
+  'banners.live-edit-draft-banner.text':
+    'This "live" document cannot be edited while a draft version of it exists.',
   /** The text for the permission check banner if the user only has one role, and it does not allow updating this document */
   'banners.permission-check-banner.missing-permission_create_one':
     'Your role <Roles/> does not have permissions to create this document.',
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -95,12 +95,11 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   'banners.deleted-document-banner.text': 'This document has been deleted.',
   /** The text content for the deprecated document type banner */
   'banners.deprecated-document-type-banner.text': 'This document type has been deprecated.',
-  /** The text content for the live edit document when it's a draft on how to solve it */
-  'banners.live-edit-draft-banner.explanation':
-    'Disable "live edit" in the schema to publish the draft. Once published, "live edit" can be reactivated.',
   /** The text content for the live edit document when it's a draft */
   'banners.live-edit-draft-banner.text':
-    'This "live" document cannot be edited while a draft version of it exists.',
+    'This live document cannot be edited while a draft version of it exists. Publish the draft in order to continue live editing it.',
+  /** The text for publish action for the draft banner */
+  'banners.live-edit-draft-banner.tooltip': 'Publish to continue editing',
   /** The text for the permission check banner if the user only has one role, and it does not allow updating this document */
   'banners.permission-check-banner.missing-permission_create_one':
     'Your role <Roles/> does not have permissions to create this document.',
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -101,7 +101,7 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   'banners.live-edit-draft-banner.publish.tooltip': 'Publish to continue editing',
   /** The text content for the live edit document when it's a draft */
   'banners.live-edit-draft-banner.text':
-    'This live document cannot be edited while a draft version of it exists. Publish or discard the draft in order to continue live editing it.',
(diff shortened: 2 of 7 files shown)


dev/test-studio/schema/standard/portableText/simpleBlock.tsx
```diff
@@ -31,6 +31,7 @@ const myStringType = defineArrayMember({
 })
 
 export default defineType({
+  liveEdit: true,
```
RitaDias, line 34: temp

packages/sanity/src/structure/panes/document/documentPanel/banners/__telemetry__/DraftLiveEditBanner.telemetry.ts
```diff
@@ -0,0 +1,21 @@
+import {defineEvent} from '@sanity/telemetry'
+
+/**
+ * When a draft in a live edit document is published
+ * @internal
+ */
+export const PublishedLiveEditDraft = defineEvent({
+  name: 'Publish a draft (live edit)',
```
bjoerge, line 8: A question for growth mainly I guess: do we want to be more clear about context here? E.g. "Resolve liveEdit by publishing draft"

packages/sanity/src/structure/panes/document/documentPanel/banners/__telemetry__/DraftLiveEditBanner.telemetry.ts
```diff
@@ -0,0 +1,21 @@
...
+  version: 1,
+  description: 'User published a draft when a draft of a live edit document to continue editing',
+})
+
+/*
+ * When a draft in a live edit document is discarded
+ * @internal
+ */
+export const DiscardedLiveEditDraft = defineEvent({
+  name: 'Create a new draft',
```
bjoerge, line 18: Same question as above. This event name in isolation indicates its just about creating any regular draft, and should maybe say that it's about resolving the issue

packages/sanity/src/structure/panes/document/documentPanel/banners/DraftLiveEditBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+import {Button} from '../../../../../ui-components'
+import {structureLocaleNamespace} from '../../../../i18n'
+import {useDocumentPane} from '../../useDocumentPane'
+import {
+  DiscardedLiveEditDraft,
+  PublishedLiveEditDraft,
+} from './__telemetry__/DraftLiveEditBanner.telemetry'
+import {Banner} from './Banner'
+
+export function DraftLiveEditBanner(): JSX.Element | null {
```
bjoerge, line 16: for consistency: I think we tend to not specify return type for react components - although I see there's a lint warning here.

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -95,6 +95,13 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   'banners.deleted-document-banner.text': 'This document has been deleted.',
   /** The text content for the deprecated document type banner */
   'banners.deprecated-document-type-banner.text': 'This document type has been deprecated.',
+  /** The text for publish action for discarding the version */
+  'banners.live-edit-draft-banner.discard.tooltip': 'Discard draft',
+  /** The text for publish action for the draft banner */
+  'banners.live-edit-draft-banner.publish.tooltip': 'Publish to continue editing',
+  /** The text content for the live edit document when it's a draft */
+  'banners.live-edit-draft-banner.text':
+    'This live document cannot be edited while a draft version of it exists. Publish or discard the draft in order to continue live editing it.',
```
bjoerge, line 104: I suggest that we provide slightly more details here:

```suggestion
The type <b>{{schemaType}}</b> has `liveEdit` enabled, but a draft version of this document exists. Publish or discard the draft in order to continue live editing it.',
```

I don't think it makes it much more confusing to non-devs, and we can probably assume they will forward the warning to someone that knows if they are unsure.

packages/sanity/src/structure/panes/document/documentPanel/banners/DraftLiveEditBanner.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const {displayed, documentId} = useDocumentPane()
```
bjoerge, line 17: Could we pass these in via props instead? We already need this info in the consuming end, so it should add any extra work (and removes a dependency on the document pane)

bjoerge (changes requested):

packages/sanity/src/structure/panes/document/documentPanel/banners/DraftLiveEditBanner.tsx
RitaDias: I tend to write the return types on every components I write exactly because of the lint waring 😬 
should I remove it?

bjoerge: let's leave it as-is for now and discuss internally whether we want to keep the lint rule or not :)

packages/sanity/src/structure/panes/document/documentPanel/banners/__telemetry__/DraftLiveEditBanner.telemetry.ts
RitaDias: I'm ok with making that change :)

packages/sanity/src/structure/i18n/resources.ts
RitaDias: Fixed the styles but done :)

packages/sanity/src/structure/panes/document/documentPanel/banners/DraftLiveEditBanner.tsx
RitaDias: That is true :) 
Done!

packages/sanity/src/structure/panes/document/documentPanel/banners/DraftLiveEditBanner.tsx
RitaDias: I would prefer to keep these, unless we make an active decision to not add the return types. 
If we have a linting rule then we should follow that rule otherwise we should remove it :)

bjoerge (dismissed):
Really nice ✨ 

Approved from my side. If we could get a 👍🏼  from @sanity-io/growth on the event names we're good to go :)

joneidejohnsen (approved):

PR #7530, opened 2024-09-20
fix(core): document store cache memo not working

pedrobonamin:
### Description

The document store cache was never found so each time we try to access it is creating a new store. 
This is because the dependencies used in the cache were not the same in the set and get.

This PR fixes that.

### What to review

Is there any possible side effect of utilizing the  memoized store?

### Testing
Existing tests should cover this .


bjoerge (approved):
Ohh, great find! Thanks Pedro!

PR #7531, opened 2024-09-20
refactor(structure): add timeout to reconnecting message

RitaDias:
### Description

Add a timeout to the toast message so that it doesn't trigger immediately.

### Testing

In any doc you can use `window.stop()` in the dev console which will set the studio state to trigger the reconnecting state

### Notes for release

Add timeout to reconnecting message

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -748,14 +748,18 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   )
 
   useEffect(() => {
+    let timeout: NodeJS.Timeout
     if (connectionState === 'reconnecting') {
-      setTimeout(() => {
+      timeout = setTimeout(() => {
         pushToast({
           id: 'sanity/structure/reconnecting',
           status: 'warning',
           title: t('panes.document-pane-provider.reconnecting.title'),
         })
       }, 2000) // 2 seconds, we can iterate on the value
+    }
+    return () => {
+      if (timeout) clearTimeout(timeout)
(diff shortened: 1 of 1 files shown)


packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -749,11 +749,13 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
 
   useEffect(() => {
     if (connectionState === 'reconnecting') {
-      pushToast({
-        id: 'sanity/structure/reconnecting',
-        status: 'warning',
-        title: t('panes.document-pane-provider.reconnecting.title'),
-      })
+      setTimeout(() => {
```
bjoerge, line 752: This needs to be cancelled in an effect cleanup

RitaDias: Oops, sorry I forgot to add it, should be there now!

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -748,12 +748,18 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   )
 
   useEffect(() => {
+    let timeout: NodeJS.Timeout
```
bjoerge, line 751:
```suggestion
    let timeout: ReturnType<typeof setTimeout>
```
removes a dependency on NodeJS types in React components

RitaDias: Done :)

bjoerge (approved):

PR #7532, opened 2024-09-20
Remove references to implicit globals `document` and `Text`

silvertech-daniel:
### Description

I replaced references to the global `document` with the `doc` parameter and `new Text` with `doc.createTextNode`. I only found this in the MS Word-generated html preprocessor. This should fix the errors that I hit while running the portable text conversion on certain html in jsdom.

### What to review

Double-check that this is a fix of a mistake.

### Testing

I have not tested this outside of edits to the package in the `node_modules` of my project

### Notes for release

* Fixed an edge-case bug in the `@sanity/block-tools` package's html to portable text preprocessing of word-generated html which broke outside of browsers


rexxars:
Appreciate the `document` => `doc` fixes, but I don't understand the `createTextNode` => `new Text` replacement - does JSDom not provide `document.createTextNode`?

silvertech-daniel:
@rexxars it's the opposite, I replaced `new Text` with `doc.createTextNode`

rexxars:
> @rexxars it's the opposite, I replaced `new Text` with `doc.createTextNode`

My bad! 😅 Thanks for submitting this!

rexxars (approved):

Issue #7337, opened 2024-08-07
Image field causes layout shifts when opening a document


jorngeorg:
**Describe the bug**
When opening a document with an image field above the fold you can visibly see the image field seemingly being mounted, unmounted and mounted again, causing unappealing layout shifts.

Here's a video. It's very quick so I recommend scrubbing to see the behaviour.
https://github.com/user-attachments/assets/4aeee8b0-c5e0-49ef-bb93-cbc87d8a7116

**To Reproduce**

1. Create a document type with an image field as the first field
2. Create two documents
3. Add an image to the image field in each document
4.  Use the document list to change between the two documents
5. Witness the flicker and layout shifts

**Expected behavior**
Image should not be mounted twice, and there should be no layout shifts. 

**Which versions of Sanity are you using?**
3.53

**What operating system are you using?**
MacOs Sonoma, Arc browser

juice49:
Thank you for the report, @jorngeorg! We made some changes in [version `3.59.0`](https://github.com/sanity-io/sanity/releases/tag/v3.59.0) that should fix this issue.

jorngeorg:
Works great. Thank you!

PR #7535, opened 2024-09-22
fix(sanity): prevent layout shifts in image input

juice49:
### Description

This branch introduces refinements to the image input preview, primarily to address the layout shift reported in #7337.

There are two scenarios in which layout shift can occur:

1. When the image is loading.
2. When the selected image state transitions to the uploading state, and the uploading state subsequently transitions to the new image. There has been a previous attempt to fix this (#6930), but I believe this may be causing the layout shift when the image loads.

This branch removes the ref added in #6930. Instead of attempting to preserve the last known height of the image preview DOM node using a ref, the known image dimensions are set as CSS custom properties on the outer container of the image input. Both the image preview and upload state UI use CSS to match the aspect ratio of the selected image, eliminating both layout shifts noted above.

Importantly, the value of the image field reflects the previously selected image while the next image is uploading. This means the upload state UI can match the dimensions of the previously selected image, addressing layout shift when uploading a new image in a simpler manner than before.

This branch also introduces a couple of other refinements to the image input:

1. Fixes incorrect use of light colour scheme in overlay components, regardless of whether the user has light mode active. In dark mode, this caused a flash of a light placeholder block, which looks janky.
2. Allows the image preview's block size to extend to `30vh`. I believe there was previously [an attempt to achieve this](https://github.com/sanity-io/sanity/blob/7c814a88e4d0629650b57d4705892ade2b7d1220/packages/sanity/src/core/form/inputs/files/ImageInput/ImagePreview.tsx#L55), but the block size was ultimately [limited to a `rem` value in CSS](https://github.com/sanity-io/sanity/blob/7c814a88e4d0629650b57d4705892ade2b7d1220/packages/sanity/src/core/form/inputs/files/ImageInput/ImagePreview.styled.tsx#L12). Moving the calculations to CSS makes it simpler to allow the block size to extend to a `vh` value, while collapsing to the intrinsic height of the image if it's less than `30vh`.

#### Demo

| Before (video) | After (video) | 
| --- | --- | 
| https://github.com/user-attachments/assets/f79bb9f8-cf62-4f51-a1d0-8bd2b3e9cdf7  | https://github.com/user-attachments/assets/8df6455a-17b4-4d08-a049-521576fca43d |
| Noticeable layout jitter when loading the image, and a flash of the light colour mode when dark mode is active make images quite janky. | After these changes, images are much more stable. | 

#### Images now have a maximum height of `30vh`.

| Before | After | 
| --- | --- | 
| before-max-height | after-max-height |

#### We still preserve the aspect ratio of the current image to prevent layout shifts when uploading a new one, which was first addressed by Cody in #6930. Now we achieve it in a simpler way.

https://github.com/user-attachments/assets/d24a10fc-5440-4a26-b89e-478b473d61a6

### What to review

Does the approach make sense?

### Testing

- Tested various image sizes and aspect ratios.
- Tested various image states (image present, image uploading, image upload failed).
- Tested various viewport sizes.
- Tested in Safari, Chrome, and Firefox.

### Notes for release

- Eliminated layout shifts in image fields.
- Fixed incorrect usage of light mode in image field overlays when dark mode is active.

### DIFF
--- a/packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
+++ b/packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
@@ -67,28 +67,6 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
 
   const uploadSubscription = useRef<null | Subscription>(null)
 
-  /**
-   * The upload progress state wants to use the same height as any previous image
-   * to avoid layout shifts and jumps
-   */
-  const previewElementRef = useRef<{el: HTMLDivElement | null; height: number}>({
-    el: null,
-    height: 0,
-  })
-  const setPreviewElementHeight = useCallback((node: HTMLDivElement | null) => {
-    if (node) {
-      previewElementRef.current.el = node
-      previewElementRef.current.height = node.offsetHeight
-    } else {
-      /**
-       * If `node` is `null` then it means the `FileTarget` in `ImageInputAsset` is being unmounted and we want to
-       * capture its height before it's removed from the DOM.
-       */
-
-      previewElementRef.current.height = previewElementRef.current.el?.offsetHeight || 0
-      previewElementRef.current.el = null
-    }
-  }, [])
   const getFileTone = useCallback(() => {
     const acceptedFiles = hoveringFiles.filter((file) => resolveUploader(schemaType, file))
     const rejectedFilesCount = hoveringFiles.length - acceptedFiles.length
@@ -201,9 +179,6 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
 
   const handleClearField = useCallback(() => {
     onChange([unset(['asset']), unset(['crop']), unset(['hotspot'])])
-
-    previewElementRef.current.el = null
-    previewElementRef.current.height = 0
   }, [onChange])
   const handleRemoveButtonClick = useCallback(() => {
     // When removing the image, we should also remove any crop and hotspot
@@ -224,9 +199,6 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
       .map((key) => unset([key]))
 
     onChange(isEmpty && !valueIsArrayElement() ? unset() : removeKeys)
-
-    previewElementRef.current.el = null
-    previewElementRef.current.height = 0
   }, [onChange, value, valueIsArrayElement])
   const handleOpenDialog = useCallback(() => {
     onPathFocus(['hotspot'])
@@ -303,15 +275,16 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
     menuButtonElement?.focus()
   }, [menuButtonElement])
 
-  const renderPreview = useCallback(() => {
+  const renderPreview = useCallback<() => JSX.Element>(() => {
+    if (!value) {
+      return <></>
+    }
     return (
       <ImageInputPreview
         directUploads={directUploads}
         handleOpenDialog={handleOpenDialog}
         hoveringFiles={hoveringFiles}
         imageUrlBuilder={imageUrlBuilder}
-        // if there previously was a preview image, preserve the height to avoid jumps
-        initialHeight={previewElementRef.current.height}
         readOnly={readOnly}
         resolveUploader={resolveUploader}
         schemaType={schemaType}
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/form/inputs/files/ImageInput/usePreviewImageSource.ts
```diff
@@ -0,0 +1,62 @@
...
+  const rect = imageUrlParams.get('rect')
+
+  if (rect) {
+    return [rect.split(',')[2], rect.split(',')[3]].map(Number) as Dimensions
+  }
+
+  return src.split('-')[1].split('.')[0].split('x').map(Number) as Dimensions
+}
+
+export function usePreviewImageSource<Value extends BaseImageInputValue | undefined>({
```
bjoerge, line 25: is there any value in this being generic? i.e. are there known cases where we'd like to preserve type information between argument type and return type when calling this function? (asking out of curiosity)

bjoerge (approved):
A great example of fixing the bug by making the code cleaner and removing complexity! Thank you so much for this @juice49. Left a few comments, no blockers for merge.

packages/sanity/src/core/form/inputs/files/ImageInput/usePreviewImageSource.ts
```diff
@@ -0,0 +1,62 @@
...
+const getImageSize = (src: string): Dimensions => {
+  const imageUrlParams = new URLSearchParams(src.split('?')[1])
+  const rect = imageUrlParams.get('rect')
+
+  if (rect) {
+    return [rect.split(',')[2], rect.split(',')[3]].map(Number) as Dimensions
+  }
+
+  return src.split('-')[1].split('.')[0].split('x').map(Number) as Dimensions
+}
```
rexxars, line 23: I realize this is old code from `ImagePreview.tsx`, but this will break when using a dataset with a dash in the name, as it splits on `-`. The `rect` handling still may need some manual extraction such as `new URL(src).searchParams.get('rect')`, but the dimensions of the image should preferably use something like [`@sanity/asset-utils#getImageDimensions`](https://sanity-io.github.io/asset-utils/index.html#getimagedimensions).

rexxars (changes requested):

packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -303,15 +275,16 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
     menuButtonElement?.focus()
   }, [menuButtonElement])
 
-  const renderPreview = useCallback(() => {
+  const renderPreview = useCallback<() => JSX.Element>(() => {
```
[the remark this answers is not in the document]
juice49: I think it depends on how you feel about inference. It can be valuable IMO to declare what a type _should_ be, rather than relying on inference.

`renderPreview` is passed to `ImageInputAsset`, which expects a function returning a `JSX.Element`.

Without setting the type parameter for the `useCallback`, somebody may think they can [return `null` on this line](https://github.com/sanity-io/sanity/pull/7535/files/2eb20ff932e53d8012a7ab721672b30743611d6e#diff-096da0579fec92502d5b032482596c3aa04075874503674869fa01e7313b0358R280), only to cause a type error elsewhere in the code.

```diff
if (!value) {
   // okay
-  return <></>
   // not okay
+  return null
}
```

Happy to reconsider, though 🙂.

packages/sanity/src/core/form/inputs/files/ImageInput/usePreviewImageSource.ts
juice49: Only as a nicety for typing the return value. If the input `value` is narrowed to exclude `undefined`, then the output `url` will always be `string`—so the consumer doesn't have to check it again.

For example, where we use it here:

Screenshot 2024-10-01 at 09 31 03

However, if `value` might be `undefined`:

Screenshot 2024-10-01 at 09 30 41

So, it's not strictly necessary. Happy to remove it if we think it's too complicated for the value it provides.

packages/sanity/src/core/form/inputs/files/ImageInput/usePreviewImageSource.ts
juice49: Ah, good catch! Should have scrutinised this code a bit more before copy/pasting it. Now using `getImageDimensions` instead. Thanks!

packages/sanity/src/core/form/inputs/files/ImageInput/usePreviewImageSource.ts
bjoerge: Cool, makes sense

bjoerge (approved):

PR #7537, opened 2024-09-23
feat: preserve visited documents in `documents-store` [WIP]

pedrobonamin:
### Description

This PR adds the ability to listen to the last N visited documents, allowing us to render them immediately when you open the form instead of having to wait for it to be fully loaded again.
It works for the following case:
- User navigates to document **foo** (Shows loading state)
- User navigates to document **bar** (Shows loading state)
- User navigates back to document **foo**  (Shows document immediately, as it has been visited before)

⚠️  It partially copies necessary changes from `corel` https://github.com/sanity-io/sanity/pull/7176
⚠️  It modifies the preview store to use mendoza patches.

https://github.com/user-attachments/assets/0694f058-ddd0-4856-9b42-f9ab04ff9a69

### What to review

It modifies the preview store to use mendoza patches, does this requires any other update?
Is something else from `corel` needed to be moved to this branch for the preview store to work?

### Testing
New tests added for the new observable.

### Notes for release

Improvements to documents loading, preserving the last 5(confirm number) visited documents so users can see the document before the listener connection is established.

### DIFF
--- a/packages/sanity/src/core/preview/createGlobalListener.ts
+++ b/packages/sanity/src/core/preview/createGlobalListener.ts
@@ -19,6 +19,7 @@ export function createGlobalListener(client: SanityClient) {
         includePreviousRevision: false,
         includeMutations: false,
         visibility: 'query',
+        effectFormat: 'mendoza',
         tag: 'preview.global',
       },
     )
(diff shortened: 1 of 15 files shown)


pedrobonamin:
Canceled and replaced for https://github.com/sanity-io/sanity/pull/7552
This PR introduced the changes using the preview listener to keep the changes alive, we have decided to use a in memory cache instead.

pedrobonamin:
Canceled and replaced for https://github.com/sanity-io/sanity/pull/7552
This PR introduced the changes using the preview listener to keep the changes alive, we have decided to use a in memory cache instead.

PR #7538, opened 2024-09-23
fix: add try/catch on `machineId()`

josemanosalvas:
### Description

This PR addresses an issue with the `@sanity/cli` package, which does not function correctly with `turbo` version 2.0 or higher on Windows systems. The root of the issue lies in the use of the `node-machine-id` dependency for telemetry purposes. When run on Windows with `turbo`, an error is thrown:

```
Error: Error while obtaining machine id: Error: Command failed: %windir%\System32\REG.exe QUERY HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography /v MachineGuid
The system cannot find the path specified.
```

This error persists even when telemetry is disabled, as the `machineId` function is still invoked. The PR resolves the issue by wrapping the `machineId` call in a try/catch block to handle the error gracefully.

**Before:**

```js
telemetry.updateUserProperties({
  deviceId: await machineId(),
  // other properties...
});
```

After:
```js
async function getDeviceId() {
  try {
    return await machineId();
  } catch (error) {
    console.error('Failed to get device ID:', error);
    return 'unknown device id';
  }
}

telemetry.updateUserProperties({
  deviceId: await getDeviceId(),
  // other properties...
});
```

### What to review

Review the changes in how `machineId` is handled within the telemetry update. Specifically, focus on the new `getDeviceId` function and ensure that it appropriately handles errors without disrupting telemetry or causing other side effects.
- Affected package: `@sanity/cli`
- Flow to test: Telemetry data generation on Windows systems when using `turbo` v2.0+.

### Testing

Manual testing was performed on a Windows system using `turbo` version 2.0. The error previously thrown by the `node-machine-id` library was resolved by the new `getDeviceId` function. As this is a Windows-specific issue, the testing was focused on this environment.

No automated tests were added due to the nature of the issue being specific to the Windows OS. However, manual tests confirmed that telemetry still functions as expected with and without valid machine IDs.

### Notes for release

This change fixes compatibility between the @sanity/cli package and turbo v2.0+ on Windows systems.

- This PR is required for users running `turbo` v2.0+ on Windows.
- The key change is the addition of a fallback mechanism for obtaining device IDs in telemetry, which avoids crashes related to the `node-machine-id` dependency.


bjoerge:
Hi  @josemanosalvas

This PR sparked an internal discussion which led us to decide to remove our usage of node-machine-id entirely. See #7545. 
 Really appreciate the time and effort you spent on investigating and putting together the otherwise great fix for this issue!

PR #7550, opened 2024-09-26
chore(form): remove imports from shouldArrayDialogOpen

RitaDias:
### Description

There's been some issues when upgrading vite 5 around the exporting of shouldArrayDialogOpen.
This was part of the Modal Madness work and while removing it might be considered a breaking change since the whole beta feature has been taken down it should be fine.

Also , I simply commented the code since I am unsure if in the future when we return to this we would want to have the code. If it's preferrable to simply delete it then I'll likely need to actively delete more components and do a more general clean up (which might be what we want but I'll leave that up to the reviewers)

### What to review

Taking a stance on the description written above should be enough

### Notes for release

N/A


rexxars (changes requested):
> Also , I simply commented the code since I am unsure if in the future when we return to this we would want to have the code.

Let's remove it for now. You don't need to do a deep clean up for now, but just removing the commented code would be good. We can always refer back to/revert this PR/commit to bring it back.

RitaDias:
> > Also , I simply commented the code since I am unsure if in the future when we return to this we would want to have the code.
> 
> Let's remove it for now. You don't need to do a deep clean up for now, but just removing the commented code would be good. We can always refer back to/revert this PR/commit to bring it back.

Fair enough, done :)

rexxars (approved):

PR #7551, opened 2024-09-26
fix(preview): allow `null` as valid cache/memo value for preview fields

bjoerge:
### Description
When previewing a document field, we keep a memo of the latest value so that when resubscribing to the same document later starts by emitting the latest known value (before we fetch from the server).

However, there was a flaw in the current implementation causing unneccessary loading states to appear in the studio:

In short: We assumed the memo being `null` meant that it hadn't been set, but in many cases it’s being explicitly set to `null` because we get `null` back from fetching the fields from the server.

This PR fixes the issue by instead treating `undefined` as "memo not being set".

### What to review
See code changes + added tests

### Testing
Added a unit test covering this particular case.

### Notes for release

- Fixes a few cases of unnecessary loading states appearing in the Studio


pedrobonamin (approved):

pedrobonamin (approved):
Looks good to me, thanks for this !

PR #7552, opened 2024-09-27
feat(core): document store loader - swr in edit state

pedrobonamin:
### Description

Adds in memory swr cache to the documents edit state, allowing us to render them immediately when you open the form instead of having to wait for it to be fully loaded again.
It works for the following case:

- User navigates to document foo (Shows loading state)
- User navigates to document bar (Shows loading state)
- User navigates back to document foo (Shows document immediately, as it has been visited before)

https://github.com/user-attachments/assets/0f47117b-c9d3-4591-891f-06349f30822c

### What to review
Is this the right approach? Any side effects to consider that hasn't been considered here?

### Testing
Existing tests have been updated with the changes introduced.

### Notes for release

Adds swr cache for document edits state, navigating between previously visited documents in the studio is now instant


packages/sanity/src/core/store/_legacy/document/document-pair/editState.ts
```diff
@@ -51,14 +55,15 @@ export const editState = memoize(
...
+      swr(`${idPair.publishedId}-${idPair.draftId}`),
+      map(({value: [draftSnapshot, publishedSnapshot, transactionSyncLock], fromCache}) => ({
         id: idPair.publishedId,
         type: typeName,
         draft: draftSnapshot,
         published: publishedSnapshot,
         liveEdit,
-        ready: true,
-        transactionSyncLock,
+        ready: !fromCache,
```
bjoerge, line 73: would it make sense to pass `fromCache` as a separate property on EditState? The granularity might be useful in certain cases (can do this later if the need arises though)

bjoerge (approved):
Such a great improvement ✨⚡

bjoerge (approved):

PR #7553, opened 2024-09-27
feat(sanity): add config for onUncaughtError

RitaDias:
### Description

Adds ability to callback when an error happens and allows developers to add custom error telemetry (for example)

I will drop the throw errors before merging, those are for testing purposes

### What to review

- Does the code make sense? 
- Does the naming of the callback make sense?
- I tried in other places, however, while discussing it with others in the team after not finding a place that would have context for the `useSource`, the current place where it exists seems to be the better option. If you all have any other suggestions I'd love to hear it!

### Testing

Automated tests are in place.
You can also manually trigger it by adding `throw new error("oh no new error)` within the `imageInput` or the `studioComponents` and then going to the `custom-components/structure` workspace and seeing the console logs work

### Notes for release

Adds ability to callback when an error happens and allows developers to use errors thrown by the studio

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -154,7 +154,7 @@ const defaultWorkspace = {
   dataset: 'test',
   plugins: [sharedSettings()],
 
-  onStudioError: (error, errorInfo) => {
+  onUncaughtError: (error, errorInfo) => {
     // eslint-disable-next-line no-console
     console.log(error)
     // eslint-disable-next-line no-console
@@ -253,7 +253,7 @@ export default defineConfig([
     dataset: 'test',
     plugins: [sharedSettings(), studioComponentsPlugin(), formComponentsPlugin()],
     basePath: '/custom-components',
-    onStudioError: (error, errorInfo) => {
+    onUncaughtError: (error, errorInfo) => {
       // eslint-disable-next-line no-console
       console.log(error)
       // eslint-disable-next-line no-console
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -77,13 +77,6 @@ const sharedSettings = definePlugin({
     },
   },
 
-  onStudioError: (error, errorInfo) => {
-    // eslint-disable-next-line no-console
-    console.log(error)
-    // eslint-disable-next-line no-console
-    console.log(errorInfo)
-  },
-
   document: {
     actions: documentActions,
     inspectors: (prev, ctx) => {
@@ -160,6 +153,13 @@ const defaultWorkspace = {
   projectId: 'ppsg7ml5',
   dataset: 'test',
   plugins: [sharedSettings()],
+
+  onStudioError: (error, errorInfo) => {
+    // eslint-disable-next-line no-console
+    console.log(error)
+    // eslint-disable-next-line no-console
+    console.log(errorInfo)
+  },
   basePath: '/test',
   icon: SanityMonogram,
   // eslint-disable-next-line camelcase
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -77,6 +77,13 @@ const sharedSettings = definePlugin({
     },
   },
 
+  onStudioError: (error, errorMessage) => {
+    // eslint-disable-next-line no-console
+    console.log(error)
+    // eslint-disable-next-line no-console
+    console.log(errorMessage)
+  },
+
   document: {
     actions: documentActions,
     inspectors: (prev, ctx) => {
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -77,11 +77,11 @@ const sharedSettings = definePlugin({
     },
   },
 
-  onStudioError: (error, errorMessage) => {
+  onStudioError: (error, errorInfo) => {
     // eslint-disable-next-line no-console
     console.log(error)
     // eslint-disable-next-line no-console
-    console.log(errorMessage)
+    console.log(errorInfo)
   },
 
   document: {
@@ -253,6 +253,12 @@ export default defineConfig([
     dataset: 'test',
     plugins: [sharedSettings(), studioComponentsPlugin(), formComponentsPlugin()],
     basePath: '/custom-components',
+    onStudioError: (error, errorInfo) => {
+      // eslint-disable-next-line no-console
+      console.log(error)
+      // eslint-disable-next-line no-console
+      console.log(errorInfo)
+    },
     form: {
       components: {
         input: Input,
--- a/packages/sanity/src/core/config/configPropertyReducers.ts
+++ b/packages/sanity/src/core/config/configPropertyReducers.ts
@@ -310,7 +310,7 @@ export const documentCommentsEnabledReducer = (opts: {
   return result
 }
 
-export const onStudioErrorResolver = (opts: {
+export const onUncaughtErrorResolver = (opts: {
   config: PluginOptions
   context: {error: Error; errorInfo: ErrorInfo}
 }) => {
@@ -320,13 +320,13 @@ export const onStudioErrorResolver = (opts: {
     // There is no concept of 'previous value' in this API. We only care about the final value.
     // That is, if a plugin returns true, but the next plugin returns false, the result will be false.
     // The last plugin 'wins'.
-    const resolver = pluginConfig.onStudioError
+    const resolver = pluginConfig.onUncaughtError
 
     if (typeof resolver === 'function') return resolver(context.error, context.errorInfo)
     if (!resolver) return undefined
 
     throw new Error(
-      `Expected \`document.onStudioError\` to be a a function, but received ${getPrintableType(
+      `Expected \`document.onUncaughtError\` to be a a function, but received ${getPrintableType(
         resolver,
       )}`,
     )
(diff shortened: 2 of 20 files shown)


dev/test-studio/components/studioComponents.tsx
```diff
@@ -33,6 +33,7 @@ const useTitleContext = () => useContext(TitleContext)
 export function CustomLayout(props: LayoutProps) {
   const {renderDefault} = props
 
+  throw new Error('This is an error from the custom layout')
```
RitaDias, line 36: Remove before merging

packages/sanity/src/core/config/configPropertyReducers.ts
```diff
@@ -310,6 +310,27 @@ export const documentCommentsEnabledReducer = (opts: {
...
+
+  if (typeof resolver === 'function') return resolver(context.error, context.errorInfo)
+  if (!resolver) return undefined
+
+  throw new Error(
+    `Expected \`document.onStudioError\` to be a a function, but received ${getPrintableType(
+      resolver,
+    )}`,
+  )
+}
```
RitaDias, line 332: Remove before merging

packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -304,6 +304,8 @@ function BaseImageInputComponent(props: BaseImageInputProps): JSX.Element {
   }, [menuButtonElement])
 
   const renderPreview = useCallback(() => {
+    throw new Error('This is an error from an image!')
```
RitaDias, line 307: Remove before merging

packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
```diff
@@ -30,15 +33,30 @@ export function WorkspaceRouterProvider({
   const history = useRouterHistory()
   const router = useMemo(() => createRouter({basePath, tools}), [basePath, tools])
   const [state, onNavigate] = useRouterFromWorkspaceHistory(history, router, tools)
+  const {onStudioError} = useSource()
```
RitaDias, line 36: catching the ErrorBoundary from `StudioErrorBoundary` since it has the context it needs here

packages/sanity/src/core/config/configPropertyReducers.ts
```diff
@@ -310,6 +310,27 @@ export const documentCommentsEnabledReducer = (opts: {
...
+
+  // There is no concept of 'previous value' in this API. We only care about the final value.
+  // That is, if a plugin returns true, but the next plugin returns false, the result will be false.
+  // The last plugin 'wins'.
+  const resolver = config.onStudioError
+
+  if (typeof resolver === 'function') return resolver(context.error, context.errorInfo)
+  if (!resolver) return undefined
+
+  throw new Error(
```
RitaDias, line 327: Remove before merging

packages/sanity/src/structure/components/structureTool/StructureTitle.tsx
```diff
@@ -46,6 +46,7 @@ const DocumentTitle = (props: {documentId: string; documentType: string}) => {
     document.title = newTitle
   }, [documentTitle, settled, newTitle])
 
+  throw new Error('This is an error from the structure tool')
```
RitaDias, line 49: Remove before merging

packages/sanity/src/core/config/configPropertyReducers.ts
```diff
@@ -310,6 +310,27 @@ export const documentCommentsEnabledReducer = (opts: {
   return result
 }
 
+export const onStudioErrorResolver = (opts: {
```
pedrobonamin, line 313: Given the way this is used I think this function could handle reporting the error to any of the plugins that defines an errorResolver by iterating in the flatten list and triggering each of the functions.

```ts
export const onStudioErrorResolver = (opts: {
  config: PluginOptions
  context: {error: Error; errorInfo: ErrorInfo}
}) => {
  const {config, context} = opts
  const flattenedConfig = flattenConfig(config, [])
  flattenedConfig.forEach(({config: pluginConfig}) => {
    // There is no concept of 'previous value' in this API. We only care about the final value.
    // That is, if a plugin returns true, but the next plugin returns false, the result will be false.
    // The last plugin 'wins'.
    const resolver = pluginConfig.onStudioError

    if (typeof resolver === 'function') return resolver(context.error, context.errorInfo)
    if (!resolver) return undefined

    throw new Error(
      `Expected \`document.onStudioError\` to be a a function, but received ${getPrintableType(
        resolver,
      )}`,
    )
  })
}

```

pedrobonamin (commented):
Thanks for this @RitaDias definitely a very nice addition to help our users!
I wonder if we could trigger the `onStudioError` in every place we have an `<ErrorBoundary>` and access to the `source`, I found the following in which we could add them
- packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/FilterForm.tsx
- packages/sanity/src/structure/components/confirmDeleteDialog/index.tsx
- packages/sanity/src/structure/panes/document/inspectors/validation/ValidationInspector.tsx

And maybe add a callout in the places this is not incorporated on why these doesn't have access to this. Basically, mentioning that this error cannot be reported to the source because the source is not in context here.


Also, we could potentially have a custom ErrorBoundary component that we can import internally instead than from `sanity/ui`  and that could handle the reporting side

Thanks for adding tests :)

packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
```diff
@@ -47,7 +66,7 @@ type HandleNavigate = (opts: {path: string; replace?: boolean}) => void
 /**
  * @internal
  */
-function useRouterFromWorkspaceHistory(
+export function useRouterFromWorkspaceHistory(
```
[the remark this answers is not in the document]
RitaDias: I think I exported it in order to be able to write the tests?
Might be wrong though, I'll remove it and see if I need to actually export it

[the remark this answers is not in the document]
RitaDias: Great catch, didn't need to be exported :)

packages/sanity/src/core/config/configPropertyReducers.ts
RitaDias: Another great suggestion which I will be adding :)

RitaDias:
> Thanks for this @RitaDias definitely a very nice addition to help our users! I wonder if we could trigger the `onStudioError` in every place we have an `<ErrorBoundary>` and access to the `source`, I found the following in which we could add them
> 
> * packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/FilterForm.tsx
> * packages/sanity/src/structure/components/confirmDeleteDialog/index.tsx
> * packages/sanity/src/structure/panes/document/inspectors/validation/ValidationInspector.tsx
> 
> And maybe add a callout in the places this is not incorporated on why these doesn't have access to this. Basically, mentioning that this error cannot be reported to the source because the source is not in context here.
> 
> Also, we could potentially have a custom ErrorBoundary component that we can import internally instead than from `sanity/ui` and that could handle the reporting side
> 
> Thanks for adding tests :)

This was a great suggestion!
We spoke on slack and an update was done across the areas mentioned as well as using a separate component :)

Once more thanks for the stellar review Pedro!

pedrobonamin (dismissed):
Thanks for this @RitaDias it looks good to me. 

One minor extra thing to add will be to modify the [`.eslintrc.cjs`](https://github.com/sanity-io/sanity/blob/next/.eslintrc.cjs) to include also `ErrorBoundary` so in the future all devs will be warned against using the `sanity/ui` error boundary and will instead need to import the new one you created 😃

pedrobonamin (dismissed):

bjoerge (commented):
Looks good, thank you for taking the time to add proper tests for this as well!

I left a few comments/suggestions, please take a look.

packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
```diff
@@ -31,14 +33,21 @@ export function WorkspaceRouterProvider({
   const router = useMemo(() => createRouter({basePath, tools}), [basePath, tools])
   const [state, onNavigate] = useRouterFromWorkspaceHistory(history, router, tools)
 
+  const handleCatchError = useCallback(({error}: {error: Error}) => {
+    /** catches errors in studio that bubble up, throwing the error */
+    throw new Error(error.message)
```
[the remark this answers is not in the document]
RitaDias: No reason, this is certainly the right way to go, thanks Bjørge 🙏

dev/test-studio/sanity.config.ts
```diff
@@ -153,6 +153,13 @@ const defaultWorkspace = {
   projectId: 'ppsg7ml5',
   dataset: 'test',
   plugins: [sharedSettings()],
+
+  onStudioError: (error, errorInfo) => {
```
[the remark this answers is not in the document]
RitaDias: Good point on the naming, I'll be updating it 👍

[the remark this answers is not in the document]
RitaDias: Done :)

PR #7554, opened 2024-09-27
refactor(store) simplify key value store, add SWR through localStorage

bjoerge:
### Description

While working on a list view optimization I discovered an re-rendering/re-fetching issue that was caused by the key value store first emitting the default value for sort order, and then fetching it from the server. In cases where a user has set a custom sort order, this would always result in two distinct settings to be returned from `useStructureToolSetting('<someKey>')`, which again would always trigger an extra re-render, and in the case of list views: a re-fetch.

This PR refactors the internals of the key value store to facilitate using LocalStorage as a read-through cache for user settings, and opting for a Stale-While-Revalidate (SWR) approach by first returning what's in local storage, and then fetching the latest settings value from the server, updating the local storage in case of a change. In the majority of cases, the localstorage value will match what's stored on the server, and in these cases we wont trigger any re-fetch/rerenders.

### What to review

These are for the most part internal changes, and I took the opportuinity to simplify where I felt it made sense. There might be, however a change in behavior in certain cases in that subscribers of user settings keys now immediately will get a cached value instead of the default value. Let me know if you can think of cases where this can be problematic (I couldn't think of any)

### Notes for release
n/a - internal


cngonzalez (approved):
LGTM, at least for the initial implementations of this feature!

pedrobonamin (approved):

PR #7558, opened 2024-09-30
fix: move cliInitializedAt call higher in init flow

svirs:
### Description

The call to patch the `cliInitializedAt` needs to be moved above the bootstrapping of the template as it is is necessary for indicating failure cases from that function.

### Testing
Should not break anything in project init flow. (all errors from the call will continue to be caught and debug logged only)

### Notes for release
not required


rexxars (changes requested):
Can we do these in parallel? (`Promise.all()`?) Would be good to not block if the server call takes a while.

svirs:
> Can we do these in parallel? (`Promise.all()`?) Would be good to not block if the server call takes a while.

Used `Promise.allSettled` to prevent one process from aborting another

drewlyton (approved):
LGTM

rexxars (approved):

PR #7569, opened 2024-10-01
[WIP]  implementation for getDocumentEvents

pedrobonamin:


packages/sanity/src/core/store/events/getDocumentEvents.ts
```diff
@@ -0,0 +1,320 @@
...
+        ...base,
+        type,
+
+        // TODO: How to identify if we are creating a version by the unpublish, is that even possible?
+        // The version that will be created by this unpublish action, e.g. drafts.foo
+        versionId: transaction.documentIDs.length > 1 ? draftId : undefined,
+        // TODO: Should this be the actual transaction id that created the draft document?
+        // The revision id of the last edit in the publish document
+        versionRevisionId:
+          previousTransactions.find((t) => t.documentIDs.includes(publishedId))?.id || 'not-found',
```
pedrobonamin, line 172: use here `transaction.id`? 

It's not clear for me if we want to have here the id of the transaction that last touched the published document or the transaction id when this version document was created.

packages/sanity/src/core/store/events/getDocumentEvents.ts
```diff
@@ -0,0 +1,320 @@
...
+          // The version that will be created by this unpublish action, e.g. drafts.foo
+          versionId:
+            transaction.documentIDs.length > 1 && transaction.documentIDs.includes(draftId)
+              ? draftId
+              : undefined,
+          // TODO: Should this be the actual transaction id that created the draft document?
+          // The revision id of the last edit in the publish document
+          versionRevisionId:
+            previousTransactions.find((t) => t.documentIDs.includes(publishedId))?.id ||
+            'not-found',
```
pedrobonamin, line 179: I'm unsure on what we want to express here:
 - the last time the published id was modified (last revision id for the published document)
 - revisionId of the draft document that was created by this transaction, that would be the `transaction.id`

packages/sanity/src/core/store/events/types.ts
```diff
@@ -0,0 +1,299 @@
...
+  'document.deleteGroup',
+  'document.createLive',
+  'document.updateLive',
+] as const
+
+/**
+ * A generic event with a type and a timestamp.
+ */
+interface BaseEvent {
+  type: (typeof documentVersionEventTypes)[number]
```
pedrobonamin, line 171: I think we would also want the `transaction.id` here

packages/sanity/src/core/store/events/types.ts
```diff
@@ -0,0 +1,299 @@
...
+
+// TODO: This is a new event type not listed in the original document.
+export interface EditDocumentVersionEvent extends BaseEvent {
+  type: 'document.editVersion'
+
+  releaseId?: string
+  versionId: string
+  versionRevisionId: string
+  mergedEvents?: EditDocumentVersionEvent[]
+}
```
pedrobonamin, line 201: This is a new event not mentioned in the initial document.

packages/sanity/src/core/store/events/getDocumentEvents.ts
codebymatt: I think we want the second option.

pedrobonamin: Yes, makes more sense like that

packages/sanity/src/core/field/types.ts
```diff
@@ -25,40 +25,21 @@ import {
 } from '@sanity/types'
 import {type ComponentType} from 'react'
 
+import {type DocumentGroupEvent} from '../store/events/types'
 import {type FieldValueError} from './validation'
 
-/**
```
pedrobonamin, line 30: This is replaced by the event types, that will be part of the `event.chunk.type`

packages/sanity/src/core/store/_legacy/history/history/Timeline.ts
```diff
@@ -194,7 +194,11 @@ export class Timeline {
     const nextTransactionToChunk = this._chunks.length > 0 ? this._chunks.last.end : firstIdx
     for (let idx = nextTransactionToChunk; idx <= lastIdx; idx++) {
       const transaction = this._transactions.get(idx)
-      this._chunks.mergeAtEnd(chunkFromTransaction(transaction), mergeChunk)
+      const allTransactions = this._transactions.getAllItems
```
pedrobonamin, line 197: To generate the event, we need the previous events, because some of them depend on checks like:
- publishExisted?
- draftExists?

All of this will be more easy to read going to the[ `getEventFromTransaction` function](https://github.com/sanity-io/sanity/pull/7569/files#diff-8260f1c6aa91c7581f94b96f7b21931b038475e0d866b6317e1961e26e395f7d)

packages/sanity/src/core/store/events/types.ts
pedrobonamin: It won't be part of the events api

pedrobonamin:
This PR is no longer relevant. 
The events will be received from the new api. `/data/events/<dataset>/documents/<documentsIds>`

PR #7572, opened 2024-10-02
fix(sanity): allow global search "contains" filter to match inside words

juice49:
### Description

This branch adjusts the global search "contains" filter to allow matches inside words. Although this filter is named `stringMatches` internally, it is publicly named "contains", which suggests it should match *inside* words.

The match expression is wrapped in wildcard `*` operators to achieve this. 

#### Example: searching for documents by title containing "blo"

| Before | After |
| --- | --- |
| Screenshot 2024-10-02 at 10 24 17 | Screenshot 2024-10-02 at 10 25 11 |

### What to review

- Does this change make sense?
- Do global searches using the "contains" filter still produce relevant results?

### Testing

- Updated unit tests in `packages/sanity/src/core/studio/components/navbar/search/definitions/operators/stringOperators.test.ts`.

### Notes for release

- The global search "contains" filter now matches inside words.


pedrobonamin (approved):
Looks good to me, tested it in the dev studio and it now returns accurate results. Thanks @juice49 !

PR #7576, opened 2024-10-03
fix(core): add detection and recovery for missing mutation events

bjoerge:
### Description
In very rare cases, the /listen endpoint may drop mutation events. This PR implements detection of "holes" in the received mutation events, and implements error recovery by restarting the connection.

### What to review
Are the threshold values sensible? Error detection and recovery is now triggered by the following thresholds.
- `DEFAULT_MAX_BUFFER_SIZE=20`: If we se more than 20 mutation events that can't be applied in order we treat it as an error
- `DEFAULT_DEADLINE_MS=30_000`: If 30 seconds pass since we last received a message that can't be applied in order we treat it as an error

### Testing
This PR includes unit tests and has also undergone extensive manual testing.

### Notes for release

- Fixes an issue that could in rare cases lead to an outdated version of the document being displayed locally

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
+++ b/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
@@ -32,6 +32,13 @@ export interface InitialSnapshotEvent {
 /** @internal */
 export interface PairListenerOptions {
   tag?: string
+
+  /**
+   * Called when we recover from sync error
+   * Meant for error tracking / telemetry purposes
+   * @param error - the {@link OutOfSyncError} recovered from
+   */
+  onSyncErrorRecovery?(error: OutOfSyncError): void
 }
 
 /** @internal */
@@ -166,7 +173,8 @@ export function getPairListener(
   return merge(draftEvents$, publishedEvents$).pipe(
     catchError((err, caught$) => {
       if (err instanceof OutOfSyncError) {
-        console.error(err)
+        options?.onSyncErrorRecovery(err)
+
         // this will retry immediately
         return caught$
       }
--- a/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
+++ b/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
@@ -173,8 +173,12 @@ export function getPairListener(
   return merge(draftEvents$, publishedEvents$).pipe(
     catchError((err, caught$) => {
       if (err instanceof OutOfSyncError) {
-        options?.onSyncErrorRecovery(err)
-
+        debug('Recovering from OutOfSyncError: %s', OutOfSyncError.name)
+        if (typeof options?.onSyncErrorRecovery === 'function') {
+          options?.onSyncErrorRecovery(err)
+        } else {
+          console.error(err)
+        }
         // this will retry immediately
         return caught$
       }
--- a/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
+++ b/packages/sanity/src/core/store/_legacy/document/getPairListener.ts
@@ -2,9 +2,12 @@
 import {type SanityClient} from '@sanity/client'
 import {type SanityDocument} from '@sanity/types'
 import {groupBy} from 'lodash'
-import {defer, type Observable, of as observableOf, of, timer} from 'rxjs'
-import {concatMap, map, mergeMap, scan} from 'rxjs/operators'
+import {defer, merge, type Observable, of, throwError, timer} from 'rxjs'
+import {catchError, concatMap, filter, map, mergeMap, scan, share} from 'rxjs/operators'
 
+import {LISTENER_RESET_DELAY} from '../../../preview/constants'
+import {shareReplayLatest} from '../../../preview/utils/shareReplayLatest'
+import {debug} from './debug'
 import {
   type IdPair,
   type MutationEvent,
@@ -12,6 +15,7 @@ import {
   type ReconnectEvent,
   type WelcomeEvent,
 } from './types'
+import {OutOfSyncError, sequentializeListenerEvents} from './utils/sequentializeListenerEvents'
 
 interface Snapshots {
   draft: SanityDocument | null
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/store/_legacy/document/utils/eventChainUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+    return [[], array]
+  }
+  return [array.slice(0, index), array.slice(index + 1)]
+}
+
+export function linkedSort<T extends {previousRev: string; resultRev: string}>(events: T[]) {
+  const parents: Record<string, T | undefined> = {}
+
+  events.forEach((event) => {
+    parents[event.resultRev] = events.find((other) => other.resultRev === event.previousRev)
```
ricokahler, line 23: `events.forEach` with  a nested `events.find` can be slow for large amounts of events. i think we only have at most 10 events so this may not need to be refactored though.

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export class OutOfSyncError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'OutOfSyncError'
+  }
+}
+
+const DEFAULT_MAX_BUFFER_SIZE = 10
+const DEFAULT_DEADLINE_MS = 3000
```
ricokahler, line 24: I think we discussed this and we liked the idea of this value being larger. maybe 30 seconds or a minute.

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export class OutOfSyncError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'OutOfSyncError'
+  }
+}
+
+const DEFAULT_MAX_BUFFER_SIZE = 10
```
ricokahler, line 23: may be worth mentioning the hard coded value in buffered document in a comment above

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+const DEFAULT_DEADLINE_MS = 3000
+
+/**
+ * Takes an observable of listener events that might arrive out of order, and emits them in sequence
+ * TODO: support a threshold for when to give up
```
ricokahler, line 28: is this TODO still relevant? it seems like there are thresholds for when to give up.

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
```diff
@@ -0,0 +1,125 @@
...
+              }
+              // we can now flush the buffer
+              return {
+                ...state,
+                base: {revision: nextBase},
+                events: nextBuffer,
+                buffer: [],
+              }
+            }
+            // todo: add a time threshold too
```
ricokahler, line 79: is `brokenChainDeadline` that time thershold?

packages/sanity/src/core/store/_legacy/document/getPairListener.ts
```diff
@@ -79,18 +87,55 @@ export function getPairListener(
...
+              return merge(
+                concat(
+                  of(createSnapshotEvent(draftId, snapshots.draft)),
+                  sharedEvents.pipe(
+                    filter((event) => event.type !== 'welcome'),
+                    filter(
+                      (sharedListenerEvent) =>
+                        sharedListenerEvent.type !== 'mutation' ||
+                        sharedListenerEvent.documentId === draftId,
+                    ),
```
ricokahler, line 113: can/should these filter statements be combined? it took me a second to understand what's going on so may be nice to refactor for readability
```suggestion
                    filter((e) => {
                      switch (e.type) {
                        case 'welcome': return false
                        case 'reconnect': return true
                        case 'mutation': return e.documentId === draftId
                        default: return false
                      }
                    })
```

packages/sanity/src/core/store/_legacy/document/utils/__test__/sequentializeEvents.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    previousRev: previousRev,
+    resultRev: resultRev,
+    transition: 'update',
+    transactionCurrentEvent: 1,
+    transactionTotalEvents: 1,
+    visibility: 'transaction',
+  }
+}
+
+test("it accumulates events that doesn't apply in a chain starting at the current revision", async () => {
```
ricokahler, line 35: this is a good test.  may be worth adding a case for the deadline timeout

packages/sanity/src/core/store/_legacy/document/getPairListener.ts
```diff
@@ -79,18 +87,55 @@ export function getPairListener(
...
+              )
+            }),
           )
-        : observableOf(event),
-    ),
+        : EMPTY
+    }),
+    catchError((err, caught$) => {
+      if (err instanceof OutOfSyncError) {
+        console.error(err)
```
ricokahler, line 134: the callback idea you mentioned to be would be nice to have here so we can add telemetry for when this happens

ricokahler (commented):
looks good!

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
bjoerge: Bumped to 30s

packages/sanity/src/core/store/_legacy/document/utils/sequentializeListenerEvents.ts
bjoerge: Yep!

packages/sanity/src/core/store/_legacy/document/utils/__test__/sequentializeEvents.test.ts
bjoerge: Good call. Done

packages/sanity/src/core/store/_legacy/document/getPairListener.ts
bjoerge: I agree it's not the most readable. Your suggestion is strictly not equal since it will filter out any non-mutation event, while the original code will include any non-mutation event. Let's revisit this later, ok? A comment might help here as well.

ricokahler (approved):
👩‍🍳💋

PR #7577, opened 2024-10-03
fix(core): add missing `listenerName` property on welcome event

rexxars:
### Description

The `listenerName` property is handy to debug certain listener related issues, but was not part of the typings.

### What to review

That there's no… typo, I guess?

### Testing

Shouldn't need any changes.

### Notes for release

None.


binoy14 (approved):
makes sense to me, thanks!

PR #7580, opened 2024-10-03
feat: validate PR title against conventional commits

binoy14:
### Description

Adds github action that validates the PR title follows semantic syntax

### What to review

Change the title to non semantic and re-run the Validate PR workflow to see the failing error. Change it back to semantic and see it pass

### Testing

N/A

### Notes for release

N/A

### DIFF
--- a/.github/workflows/pr-validate.yml
+++ b/.github/workflows/pr-validate.yml
@@ -1,0 +1,16 @@
+name: "Validate PR"
+
+on:
+  pull_request:
+
+permissions:
+  pull-requests: read
+
+jobs:
+  main:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/validate-pr-title.yml
+++ b/.github/workflows/validate-pr-title.yml
@@ -1,7 +1,12 @@
-name: "Validate PR"
+name: "Validate PR title"
 
 on:
-  pull_request:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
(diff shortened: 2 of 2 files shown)


binoy14:
* **#7580** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/7580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

.github/workflows/pr-validate.yml
```diff
@@ -0,0 +1,16 @@
+name: "Validate PR"
+
+on:
+  pull_request:
```
rexxars, line 4: Based on [this description](https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#event-triggers), I would think it better to use `pull_request_target` here so that it'll work when PRs are received from forks. Thoughts?

```suggestion
on:
  pull_request_target:
    types:
      - opened
      - edited
      - synchronize
      - reopened
```

.github/workflows/pr-validate.yml
```diff
@@ -0,0 +1,16 @@
+name: "Validate PR"
```
rexxars, line 1: At least for now, this is only validating the title of the PR - so would be good to be more specific about it I think.

```suggestion
name: "Validate PR title"
```

Might want to rename the workflow filename to reflect, too.

rexxars (changes requested):

.github/workflows/pr-validate.yml
binoy14: I tried that first but for some reason it didn't seem to work. Maybe it needs both, I'll read up in more details on what the target type does

rexxars: I think it doesn't work for the branch you're experimenting in, since it takes the workflow from the target branch? I think it should work post-merge though

rexxars (approved):

PR #7584, opened 2024-10-03
chore(deps): upgrade `@sanity/asset-utils` to v2

rexxars:
### Description

Just a dependency update. v2 of this module uses ESM by default, aside from that there shouldn't be any meaningful changes.

### What to review

Stuff still works. In particular, image and file fields + copy/paste of these fields.

### Testing

Existing tests should hopefully be enough.
Have done some manual testing as well.

### Notes for release

None


binoy14 (approved):
Looks like `sanity-plugin-hotspot-array` and `@sanity/import` use `1.3.0` upgrading those should only keep one version of the asset-utils in the lockfile

rexxars:
Will update those, thanks!

PR #7555, opened 2024-09-27
fix(structure): memoize search query results

bjoerge:
### Description
This adds Stale-While-Revalidate behavior to document lists in the Studio. This means that when coming back to a previous list, we'll immediately show the last known contents, and then fetch the new version in the background. In many cases this means that going back and forth between document lists feels instant.

### What to review

The actual SWR logic is implemented with an LRU cache in [packages/sanity/src/structure/panes/documentList/listenSearchQuery.ts](https://github.com/sanity-io/sanity/compare/simplify-key-value-store...fix-memoize-list-results?expand=1#diff-351b61b4553a9471d3d02094781702e9da5aeb21780233022f8543efc2590813) and it uses a memo key based on various input parameters for the search query request.

I took the liberty of refactoring some of the logic and renaming a few variables etc based on my (ever evolving) understanding of the existing code. Please let me know if I missed the mark on something.

### Testing
I couldn't think of a good way to robustly test the SWR behavior introduced here. Let me know if you have any suggestions!

I've manually verified the following:
- Document lists updates in real-time
- Searching in lists works as before
- If navigating back to a list triggers a background refresh, so if a document has been added or removed it's reflected after a short while.
- Sorting lists works as before
- Scrolling to the end of a list fetches the "full" list (show a loading spinner at the bottom and retains scroll position)
- Loading variants works as they used to (I also added a new "subtle" one that appears if revalidating takes longer than a few hundred milliseconds)
- Error handling and retrying works as before

@hermanwikner and @robinpyon tagging you as reviewers here as well since you did the original work on doument list search including `listenSearchQuery` and `useDocumentList`.

### Notes for release
- Navigating between previously visited document lists in the studio is now instant

PR #7587, opened 2024-10-04
refactor(sanity): fix type issues with collate, update tests and readQueryResult

RitaDias:
### Description

- Re-adds `QueryResult` interface
- Changing the type in the collate to not explode when versions are present.

When rebasing against `next` something went wrong specifically around this [PR](https://github.com/sanity-io/sanity/pull/7555/files). I wasn't able to fix it by using rebasing however this unblocks us and allows us to continue work.

I imagine that the fix will likely be more around making sure that the collate prop type stayed the same (making the changes in the ENABLE_LRU_MEMO changes however I don't know the larger impact of making that change, so, for now I've set this up.

Folks that are more knowledgeable around this should weight in before I merge since this is more like a temporary bandaid than a fix 🥲 

Ash set up a branch `corel-pre-rebase-20241003` in case if you want to give the rebase another stab to avoid some of these issues. Keep in mind that you'll need to cherry pick [StickParams](https://github.com/sanity-io/sanity/pull/7429) commits as well

### DIFF
--- a/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
+++ b/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
@@ -243,7 +243,7 @@ export function referenceSearch(
   })
   return search(textTerm, {includeDrafts: true}).pipe(
     map(({hits}) => hits.map(({hit}) => hit)),
-    map((docs) => collate(docs)),
+    map((docs) => collate({documents: docs})),
     // pick the 100 best matches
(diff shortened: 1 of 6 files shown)


juice49 (approved):

RitaDias:
Spoke with Ash and to allow us to continue doing some work in COREL I'm merging, however we are trying to have a more successful rebase from the branch to be able to develop without so many issues.

PR #7588, opened 2024-10-04
fix(core): portable text annotations slow to show

pedrobonamin:
### Description

Adding an annotation inside the PTE editor is perceived as slow, this is mainly because the annotation needs the annotation value to exist in the form value before being able to show the Popover for it.
Also: 
- The **Add comment** button reappears in place of the annotation after clicking "add annotation."
- The annotation **edit/discard toolbar** appears before the annotation popover renders.
- The annotation in PTE changes its style before the annotation popover renders.
- The popover takes about 1 second to render after clicking and doesn't focus.

This PR updates the add comment button to not reappear when the editor is restored with focus.
It also updates the annotation toolbar to not show until the annotation value exists.

Following changes:
The portable text editor package will be update to not debounce the `addAnnotation` action

https://github.com/user-attachments/assets/7eadd32a-08a7-4306-8fb9-a5d5666a0c89

### What to review
Are this changes correct?
Is there any oversight to fix?

### Notes for release

Update annotations creations in portable text, autofocus on the first input.


RitaDias (approved):
As always, fantastic work and great pr description :)

PR #7593, opened 2024-10-04
fix: protect against falsy `selectedGroup`

ricokahler:
### Description

This fixes a bug that caused the document editor to crash in certain cases where there is no selected group. This occurs when the default field group is not found. Allow this value to be `undefined` and checking for it fixes the issue.

### What to review

Does this correct protect against all cases where the selectedGroup is undefined? Are all the types updated correctly?

### Testing

This was tested manually in the admin studio. I built sanity package, packed it, and install it in the admin studio and the issue was resolved.

### Notes for release

Fixes a bug that caused the document editor to crash when the default field group is not found.


ryanbonial (approved):
It would be nice to have a test for this case, but the change looks good 👍

PR #7594, opened 2024-10-04
fix(cli): do not throw during migrate if project ID is passed through flag

ryanbonial:
### Description

This PR sets the `apiClient`'s  `requireProject` to `false` and instead only throws an error if both the cli config and the `--project` option are missing

### What to review

Verify that the change is as expected and that the wording of the error is correct.

### Testing
* create a simple migration script with sanity, `sanity migration create` and follow the prompt
make sure there is no `sanity.cli.js` in your codebase

Next steps:
* Open /code/migrations/migration_id/index.ts in your code editor and write the code for your migration.
* Dry run the migration with:
`sanity migration run migration_id --project= --dataset`
* Run the migration against a dataset with:
`sanity migration run migration_id --project= --dataset --no-dry-run`

* edit the script and run it as:
`sanity migration run migration_id --project=real_project_id --dataset real_dataset`

you should NOT get an error. 

Test again without specifying  a `--project`  project ID, you should see an error.

### Notes for release
None

### DIFF
--- a/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
+++ b/packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
@@ -164,8 +164,14 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
 
     const projectConfig = apiClient({
       requireUser: true,
-      requireProject: true,
+      requireProject: false,
     }).config()
+
+    if (!project && projectConfig.projectId) {
+      throw new Error(
+        'sanity.cli.js does not contain a project identifier ("api.projectId") and no --project option was provided.',
+      )
+    }
 
     const apiConfig = {
       dataset: dataset ?? projectConfig.dataset!,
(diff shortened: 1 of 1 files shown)


packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
```diff
@@ -164,9 +164,15 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
 
     const projectConfig = apiClient({
       requireUser: true,
-      requireProject: true,
+      requireProject: false,
```
ryanbonial, line 167: I left this in as `false` instead of removing it so that it would indicate that `false` is intended here.

packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
```diff
@@ -164,9 +164,15 @@ const runMigrationCommand: CliCommandDefinition<CreateFlags> = {
...
     }).config()
 
+    if (!project && projectConfig.projectId) {
```
rexxars, line 170: Shouldn't this be reversed? 
```suggestion
    if (!project && !projectConfig.projectId) {
```

rexxars (changes requested):

packages/sanity/src/_internal/cli/commands/migration/runMigrationCommand.ts
ryanbonial: Yes, thank you.

rexxars (approved):

PR #7597, opened 2024-10-05
fix: prevent excessive requests to access endpoint 

rexxars:
### Description

This one needs a proper read-through, I am not super confident with these changes - especially the last commit. There might be a lot easier ways to solve the problem outlined in the title here, but let me explain:

When researching why there were constant requests being fired to the `/access` API, I saw that it was due to a hook being run inside of the `<PermissionCheckBanner />`. Upon further inspection, I saw that this component remounts (unmounts and mounts) constantly while edits are going on, due to the continuous evaluation of permissions as the document changes. When this happens, the `isPermissionsLoading` state gets set to `true`, which triggers the banner to unmount. Once it's calculated if it should be allowed/disallowed, it re-mounts. My hunch is that this is not only causing the additional requests that we want to get rid of, but potentially also affects performance negatively.

While reading the code of the `useDocumentValuePermissions` and the underlying `createHookFromObservableFactory` and friends, it seems like the intention _is_ to [keep the old state while recalculating](https://github.com/sanity-io/sanity/blob/92de3d9606957a3ce8b6c1c93444dab76536da46/packages/sanity/src/core/util/createHookFromObservableFactory.ts#L59) - but because the `document` being passed as an argument is recreated on any edit, the observable factory gets called again and the state resets. In my mind it would be better (for the permission check in particular) to assume that the grants don't change _that_ quickly, and keep the old cached value around. I refactored the code to do this, but the code is a lot simpler/more naive than `createHookFromObservableFactory` + `useObservable`, so I'd be happy to have some more eyes on this and see if there are edge cases it wouldn't account for.

> [!NOTE]
> While writing this, I realized that we should probably throw away that cached value on document ID change.

[Running the new eFPS suite](https://github.com/sanity-io/sanity/actions/runs/11188606113/job/31107661697), the eFPS looks about the same, while blocking time seems lower - lets see what another run says once I open the PR.

Aside from the above, I also refactored some code a bit - using observable requests for cancellation, minimizing state updates with predefined constants and preferring to "return early" from functions for less indentation (easier to read in my book).

### What to review

I'd do commit-by-commit - most of the changes should be fairly straightforward, except the last two. In particular, the last change needs a very thorough read-through. 

### Testing

I want to add some tests for `useDocumentValuePermissions`, but figured I would open this up for an early look since it needs a proper read-through.

### Notes for release

- Fixed an issue where unnecessary requests to the API would be triggered while editing a document

### DIFF
--- a/packages/sanity/src/core/store/_legacy/grants/documentValuePermissions.ts
+++ b/packages/sanity/src/core/store/_legacy/grants/documentValuePermissions.ts
@@ -1,6 +1,8 @@
-import {type Observable} from 'rxjs'
+import {isEqual} from 'lodash'
+import {useEffect, useReducer, useRef} from 'react'
+import {distinctUntilChanged, type Observable, type Subscription} from 'rxjs'
 
-import {createHookFromObservableFactory, type PartialExcept} from '../../../util'
+import {type LoadingTuple, type PartialExcept} from '../../../util'
 import {useGrantsStore} from '../datastores'
 import {type DocumentValuePermission, type GrantsStore, type PermissionCheckResult} from './types'
 
@@ -9,23 +11,6 @@ export interface DocumentValuePermissionsOptions {
   grantsStore: GrantsStore
   document: Record<string, unknown>
   permission: DocumentValuePermission
-}
-
-/**
- * The observable version of `useDocumentValuePermissions`
- *
- * @see useDocumentValuePermissions
- *
- * @internal
- */
-export function getDocumentValuePermissions({
-  grantsStore,
-  document,
-  permission,
-}: DocumentValuePermissionsOptions): Observable<PermissionCheckResult> {
-  const {checkDocumentPermission} = grantsStore
-
-  return checkDocumentPermission(permission, document)
 }
 
 /**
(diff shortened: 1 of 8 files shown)


juice49:
Hey @rexxars. I've been reading through the changes, and the premise _seems_ good to me. I could do with spending a bit more time to give a definitive 👍/👎.

You mentioned you want to add some tests and that you opened this PR for an early look. Are you hoping to get this change merged by tomorrow?

rexxars:
> Hey @rexxars. I've been reading through the changes, and the premise _seems_ good to me. I could do with spending a bit more time to give a definitive 👍/👎.
> 
> You mentioned you want to add some tests and that you opened this PR for an early look. Are you hoping to get this change merged by tomorrow?

Might be a bit hopeful, I don't think realistically I/we will have much time for it this week :/

juice49:
Okay, thanks for confirming. I will come back to it, as I'm likely to run out of time before tomorrow.

packages/sanity/src/core/store/_legacy/grants/documentValuePermissions.ts
```diff
@@ -49,23 +34,83 @@ export function getDocumentValuePermissions({
...
+        error: (error) => dispatch({type: 'error', error}),
+      })
+
+    return () => {
+      if (subscriptionRef.current) {
+        subscriptionRef.current.unsubscribe()
+        subscriptionRef.current = null
+      }
+    }
+  }, [grantsStore, document, permission])
```
juice49, line 113: This is one potential difference I can see between the `createHookFromObservableFactory` implementation and your implementation: the former uses `useUnique`, which in turn uses `lodash.isEqual` to compare the args. You instead have added an effect with object dependencies that will behave differently.

Looks like you may have mitigated this by using `disctinctUntilChanged` with `lodash.isEqual` in the `permissions$` observable—and there doesn't seem to be any issues—but thought I'd mention it in case.

juice49 (approved):
Conducted some testing and carefully compared to the previous implementation. It seems sound to me. Happy to re-review after tests added 👍.

bjoerge (approved):
This looks solid to me. I've read through carefully and also tested quite a bit manually, and have not seen any issues. Thanks for not only fixing the issue at hand, but also for cleaning up and improving the code along the way! <3

PR #7598, opened 2024-10-05
chore(ci): use environment variables, not secrets, for non-sensitive vars

rexxars:
### Description

I don't think there's any reason to hide variables such as dataset and project IDs. It makes it harder to debug issues and read logs, as any value in these secrets will be scrubbed from logs as `****`. In particular, we use `test` as a dataset name 😅

### What to review

- CI configs still looks right
- [Environment variables](https://github.com/sanity-io/sanity/settings/variables/actions) are in place to take the place of the secrets

### Testing

Letting the CI workflows run should be enough

### Notes for release

None


ryanbonial (approved):
👍 looks good

PR #7599, opened 2024-10-06
fix: use absolute urls for published sourcemap urls

rexxars:
### Description

The source maps we upload to our module CDN now contains source maps, which helps in debugging. Sources are inlined, but the original source _locations_ are also left intact. This isn't a _problem_ per se, but it does lead to tooling trying to link to them and failing. For instance, a source map could be referencing `../../../node_modules/.pnpm/@sanity+client@6.22.0_debug@4.3.7/node_modules/@sanity/client/dist/index.browser.js`, which would be resolved to `https://modules.sanity-cdn.com/modules/v1/node_modules/.pnpm/@sanity+client@6.22.0_debug@4.3.7/node_modules/@sanity/client/dist/index.browser.js`, which we obviously do not host.

It is helpful to have _some_ source location, as from the example above I can see that the source pointed to `@sanity/client`. To make this even more helpful, this PR takes the approach of rewriting these sources to point to absolute URLs.

Technically speaking this is "incorrect", as we don't use the URLs given for actually building the bundles, however it is useful for debugging purposes, and leaving the original paths do not make sense as we are not hosting the source files on the CDN anyway.
 
The URLs we rewrite to are as follows:
- For absolute URLs, leave them as-is.
- For dependencies (eg anything inside of `node_modules`), we use `jsdelivr.net`.
- For Sanity monorepo packages, we use GitHub URLs. Note that these link to the HTML interface, _NOT_ the raw file - so any tools that tries to actually fetch the file might fail. If this proves to be an issue, we could switch to `raw.githubusercontent.com` instead, but it is less user-friendly for us developers.

### What to review

- Do these changes make sense? Are they going to cause issues?
- See any issues in the code?

### Testing

I've built the bundles locally and ran a modified uploading script (disabling the actual upload logic) and verified that the sources gets correctly rewritten.

### Notes for release

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,7 @@
     "npm-run-all2": "^6.2.3",
     "ora": "^8.0.1",
     "prettier": "^3.3.3",
+    "read-package-up": "^11.0.0",
     "rimraf": "^3.0.2",
     "rxjs": "^7.8.1",
     "sanity": "workspace:*",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -198,6 +198,9 @@ importers:
       prettier:
         specifier: ^3.3.3
         version: 3.3.3
+      read-package-up:
+        specifier: ^11.0.0
+        version: 11.0.0
       rimraf:
         specifier: ^3.0.2
         version: 3.0.2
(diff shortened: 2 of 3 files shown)


binoy14 (approved):
This is great for debugging, thank you! 

Tested it locally and matches the expectation

PR #7601, opened 2024-10-07
chore(ci): change env vars from secrets to env

bjoerge:
### Description

This changes some remaining references to secrets in the e2e workflow after #7598 moved them to env vars.

### What to review

Did I miss any?

### Testing
CI passes, I guess?

### Notes for release
n/a


rexxars (approved):
Not sure if all the `env.` needed changing - but seems to still be building?

PR #7615, opened 2024-10-14
feat(form): add support for disabling array input capabilities

bjoerge:
### Description

This adds support for the `disableActions` option on array fields for disabling various array input capabilities like:
  - `add` – Removes the ability to add new items to the array
  - `addBefore` – Removes the "Add item before"-menu item from the array item menu
  - `addAfter` – Removes the "Add item after"-menu item from the array item menu
  - `remove` – Removes the ability to remove items from the array
  - `duplicate` – Removes the ability to duplicate array items
  - `copy` – Removes the ability to copy items from the array

Usage example:
```ts
{
      name: 'someArrayField',
      options: {
        disableActions: ['add', 'duplicate'],
      },
      title: "Array you can't add elements to",
      type: 'array',
      of: [
        {
          type: 'object',
          name: 'something',
          title: 'Something',
          fields: [{name: 'first', type: 'string', title: 'First string'}],
        },
      ],
    }
```

Things to note:
- These changes are only about UI affordances, and doesn't imply any form of write protection for the actual data. Items can still be added or removed to an array by sending mutations to the API.
- This only applies to arrays of objects and arrays of primitive values. Not to portable text arrays.
- Disabling `add` will also implicitly disable `addBefore` and `addAfter`, thus disable inserting new items to the array entirely (although items can still be inserted via `duplicate`).
- There might not be a compelling use case for disabling the ability to `copy` here(?), but including it for completeness. If all known actions are disabled, the menu items and "add"-affordances will be removed from the UI.
- If we later decide to introduce a new default action, this will appear after upgrading the Studio, and will have to be explicitly disabled.
- Disabling `copy` will also disable the `copy` _field action_ (for the entire array). In the same vein, disabling `add` will remove the "paste" field action. However, any other field actions that may have been added by plugins etc. may still be able to manipulate the array in ways disallowed by `disableActions`

### What to review
Does it make sense?

### Testing
- Disable each of the capabilities and check that it works as expected. Toggle individual capabilities in [dev/test-studio/schema/debug/arrayCapabilities.ts](https://github.com/sanity-io/sanity/blob/sdx-1641/dev/test-studio/schema/debug/arrayCapabilities.ts#L3)
I've also added a basic e2e test that asserts that menu items and add button doesn't appear when all capabilities are disabled

### Notes for release

- Adds support for the `disableActions` option on array fields for disabling various array input capabilities

### DIFF
--- a/packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ItemRow.tsx
+++ b/packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ItemRow.tsx
@@ -1,4 +1,4 @@
-import {AddDocumentIcon, CopyIcon, TrashIcon} from '@sanity/icons'
+import {AddDocumentIcon, CopyIcon, InsertAboveIcon, InsertBelowIcon, TrashIcon} from '@sanity/icons'
 import {type SchemaType} from '@sanity/types'
 import {Box, Flex, Menu} from '@sanity/ui'
 import {type ForwardedRef, forwardRef, useCallback, useMemo} from 'react'
@@ -9,7 +9,7 @@ import {useTranslation} from '../../../../i18n'
 import {FieldPresence} from '../../../../presence'
 import {FormFieldValidationStatus} from '../../../components/formField'
 import {type PrimitiveItemProps} from '../../../types/itemProps'
-import {InsertMenuGroups} from '../ArrayOfObjectsInput/InsertMenuGroups'
+import {InsertMenuGroup} from '../ArrayOfObjectsInput/InsertMenuGroups'
 import {RowLayout} from '../layouts/RowLayout'
 import {getEmptyValue} from './getEmptyValue'
 
@@ -33,6 +33,7 @@ export const ItemRow = forwardRef(function ItemRow(
     onRemove,
     readOnly,
     inputId,
+    parentSchemaType,
     validation,
     children,
     presence,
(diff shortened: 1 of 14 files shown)


pedrobonamin (approved):
Nice work, looks good to me!

packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ItemRow.tsx
```diff
@@ -72,36 +74,76 @@ export const ItemRow = forwardRef(function ItemRow(
...
+          <InsertMenuGroup
+            pos="before"
+            types={insertableTypes}
+            onInsert={handleInsert}
+            text={t('inputs.array.action.add-before')}
+            icon={InsertAboveIcon}
+          />
+        ),
+        !disableActions.includes('add') &&
+          !(disableActions.includes('addAfter') && disableActions.includes('addBefore')) && (
```
jordanl17, line 117: Q: This seems different to the other conditions added elsewhere in this PR (eg [here](https://github.com/sanity-io/sanity/pull/7615/files#diff-8b5210bf0f17afe0292050378c765c4cf89567d542f57af940543ca043f398daR185-R187)) where we only allow after if add and addAfter are not disabled.

Is it intended for this to be different?

bjoerge: Great catch! Fixed in 991d357

PR #7619, opened 2024-10-15
feat(sanity): add release types

RitaDias:
### Description

Add release types to bundles metadata (asap, scheduled or undecided)
Names are still in flux, these seemed to be the ones that we would least regret in the long run as well as make it clearer from a code standpoint, I'm open to other suggestions though!

Also, we were planning on having these properties exist in a separate place other than the bundle metadata, but due to limitations outside of our control this is the temporary solution

### What to review

Does the code provided make sense?
Keep in mind, this is simply for unblocking us in continuing corel development, **the actual development and changes to the modal will be done in a separate task and so are out of scope**

### Testing

Updated tests to pass

### DIFF
--- a/packages/sanity/src/core/releases/components/dialog/ReleaseDetailsDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/ReleaseDetailsDialog.tsx
@@ -2,7 +2,7 @@ import {ArrowRightIcon} from '@sanity/icons'
 import {useTelemetry} from '@sanity/telemetry/react'
 import {Box, Flex, useToast} from '@sanity/ui'
 import {type FormEvent, useCallback, useState} from 'react'
-import {type FormBundleDocument, useTranslation} from 'sanity'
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/releases/components/dialog/ReleaseDetailsDialog.tsx
```diff
@@ -40,6 +40,7 @@ export function ReleaseDetailsDialog(props: ReleaseDetailsDialogProps): JSX.Elem
       hue: bundle?.hue || 'gray',
       icon: bundle?.icon || 'cube',
       publishedAt: bundle?.publishedAt,
+      releaseType: bundle?.releaseType || 'asap',
```
bjoerge, line 43: Maybe create a const for default type so it's defined in one place?
e.g.
`const DEFAULT_RELEASE_TYPE = 'asap'`

RitaDias: Good point :) 
Added!

bjoerge (approved):

PR #7633, opened 2024-10-18
feat(cli): prepare nextjs starter template for live mode

SimeonGriggs:
### Description

Adds configuration file for live fetches when initializing a new Studio inside a Next.js application. Not used by the Studio, but designed for use in the Next.js application.

Do not release until the next version of next-sanity is available.

### Notes for release

`sanityFetch` and `SanityLive` preconfigured for Studios initialized inside a Next.js application


rexxars (approved):
Reading the next-sanity docs, it seems like the live content API might require vX to be used - are you okay leaving that up to the env file? Also, are we leaving the `SanityLive` component up to the user to implement "if they want it"?

packages/@sanity/cli/src/actions/init-project/templates/nextjs/index.ts
```diff
@@ -165,6 +165,12 @@ export const client = createClient({
 })
 `
 
+const live = `import { defineLive } from "next-sanity";
```
rexxars, line 168: Maybe we should add a description to the top for what this file is for?
```suggestion
// Importing and querying through `sanityFetch` from this file will make data automatically stay up to date with any changes.
// Before using it, import and render `<SanityLive />` in your layout - see
// https://github.com/sanity-io/next-sanity#live-content-api for more information.
const live = `import { defineLive } from "next-sanity";
```

SimeonGriggs:
> Reading the next-sanity docs, it seems like the live content API might require vX to be used - are you okay leaving that up to the env file? Also, are we leaving the `SanityLive` component up to the user to implement "if they want it"?

Good call. I've updated the `live.ts` file to set `vX` just on this client instance, with a description. 

And yeah all these helper files are designed to be implemented by the user, they're not written into the Next.js app anywhere.

rexxars (approved):

PR #7635, opened 2024-10-19
feat: adds support for Create-Studio integration

snorrees:
## Description

This PR introduces the Studio side of the upcoming Create <-> Studio integration. 

It allows users start their content editing journey in Create for new documents, by linking a Create document to a Studio document. 

While linked, Create will sync content into the Studio document in real-time. Create linked Studio documents are read-only. Studio users can unlink the document at any point, to continue editing in the Studio. At this point Create will no longer sync content into the Studio document.

Update: Now sporting a beauty pass by @robinpyon 🙇 

Update: `beta.sanity.startInCreateEnabled` will default to `false`. Ie, the "Start in Sanity Create" button will be OPT IN for now. Documents that get linked from the Create side will still be put into read-only in the Studio. (this is also a form of opt-in tbh).

## What to review

Read through this full PR text before jumping into the code.
Changes affect:
* core
   *  ~~change to userStore needs extra attention~~ (This is no longer relevant, see below)
* structure
* schemaType options

There might be some controversial changes in here. I have made PR comments everywhere I though there might be some eyebrows raised, but there might be more. I'm very much open to course corrections.

## Feature details
**For Sanity folks: first see video of the full Studio-Create flow [here](https://www.notion.so/sanityio/Start-in-Studio-edit-in-Create-12678d055b948065a6d8cdcab74d6078?pvs=4).** 

In Studio deployments with an exposed `create-manifest.json`, new, prestine documents will have a "Start in Sanity Create" button in the document pane footer. See [previous manifest PR ](https://github.com/sanity-io/sanity/pull/7403) for context.

image

Clicking it will opens a confirmation dialog:

image

Clicking "Learn more" will open a "How to Create<->Studio" article. (Link url pending)

Clicking "Continue" will open a new tab which kicks you off to Create which will:
* Create a new document
* Populate it with some initial data
* Link the Create document to the Studio document

A studio document is considered "Create linked" whilst it has `_create.ejected === false`.

While Create setting up the link (which entails setting the `_create` property), the following dialog will be shown after clicking "Continue":

image
(displaying open state of the toggle)

Once the link is established, the document will be `readOnly` and the pane footer will be in a special "Create linked" mode:

image

Clicking the info icon will open a popover:

image

Clicking anywhere outside (or pressing ESC) closes the popover.

Clicking "Edit in Sanity Create" will open up the Create document that points to this Studio document in a new tab.

Clicking "Unlink" will show the following dialog:

image

"Unlink now" will unset the `_create` property, thereby severing the link and making the document a regular Studio document. It will no longer be `readOnly`.

### Code organization

I tried to:
1. implement as much functionally as possible using our public plugin APIs
2.  keep as much Create integration code under the same directory as possible
3. Build on existing principles when extending type options

For 1: this feature has some requirements that were not 100% plugin-supported, and required changes in core and structure:
* document actions sort order
* a clean way to put a pane-wide banner under the document pane header
* a way to force certain "document footer actions" when Create linked

For 2: I had to introduce a bit of indirection to make it possible to render stuff in structure.  `SanityCreateConfigContext`contains two component implementations used by structure, provided from core. 
These are the components rendered when a document is Create linked. (Read only banner & Unlink actions).

For 3: I have added a [BaseSchemaTypeOptions](https://github.com/sanity-io/sanity/pull/7635/files#r1807803690) type. I always regretted not adding that for the initial v3 launch, since it allows plugins to add generic type-options much more easily. For instance, it could simplify the [AI Assist type extensions](https://github.com/sanity-io/assist/blob/main/plugin/src/schemas/typeDefExtensions.ts#L15) quite a bit.

### SanityCreateOptions
I have added `sanityCreate` options directly to `BaseSchemaTypeOptions`. If it feels controversial to have them there, I can move them into a module declaration extension in core, but I feel that is only adding indirection with no upside. 

The idea is that `SanityCreateOptions` will be the DSL with which devs tailor their schema in Create (via the manifest file). 
Atm it supports `exclude` and `purpose`

* `exclude` removes a type or field from appearing in Create. Excluded document types will not have a "Start in Sanity Create" button.
* `purpose` supersedes `description`, and will be used as metadata when describing the schema to an LLM

We need different options from AI Assist (which also has this exclude option), as Create has different needs than AI Assist.

### Start in Create 
We show the "Start in Sanity Create" button when:
* `beta.create.startInCreateEnabled: true`. Atm, `startInCreateEnabled` defaults to false. This means studios will have to OPT IN.
* the document is new (ie, no `_createdAt`) 
* the current browser origin matches a Studio origin found in Studios for the current project (sanity.io/manage)
   * exception: developers can provide a fallbackStudioOrigin to make the button appear on localhost
   * why is origin important?: Create uses Studio appId when preparing the studio link; it needs to import the schema from `create-manifest-json` on a known url. It will only visit hosts listed under Studios in manage. 
 * `documentOptions.sanityCreate.exclude !== true` 

### Create link

A Studio document is considered linked to Create when it contains a `_create` metadata field.

Specifically we check for `_create.ejected === false`. The idea is that we _can_ keep the Create metadata around if we want to, but in this implementation we unset the full `_create` field when unlinking. 
This ensures that this metadata does not end up in published documents.

If we want to do "soft unlinks" to keep a trace back to Create in the future (possibly opt-in via config), we can do that with with `_create.ejected: true`.

Create link readOnly mode will always be enabled, regardless of what `beta.create.startInCreateEnabled` is.

###  ~~Getting the global user id~~
Edit: We found issues with this approach and will remove the "global-id-in-url" approach.
As it turns out the user can log in with a _different_ user in Create, resulting in the link-creation failing. Instead Create will handle all this, so we no longer need global user id in Studio. 

 ~~To build the "Start in Sanity Create" url, we need the global user id.~~
 ~~For current user, this is currently not available – but it IS for all other userIds. I propose a change to that, by no longer pre-priming the userStore with currentUser (at the cost of one more network request).~~

 ~~See [PR code comment](https://github.com/sanity-io/sanity/pull/7635/files#r1807398846) for change to userStore.~~

### App-id cache
To build "Start in Sanity Create" url, we need the appId (deployed studio id) from `<project>/user-applications`.
It is ok to cache this info for the duration of the Studio lifetime (ie, until browser refresh).

I dont know if there are existing caching mechanisms in core I could use instead of rollinga [bespoke fetch-cache](https://github.com/sanity-io/sanity/pull/7635/files#r1807394016) for this.

### Telemetry

I've added telemetry for the following actions:

* Start in Create clicked
* Start in Create accepted (clicking continue or auto-accept is the same event)
* Unlink clicked
* Unlink accepted
* Edit in Create clicked

### Open questions

* Do these code-paths need e2e tests?

### Known caveats 

Create does not respect `initialValues`, so these will be nuked when Create starts syncing.
We might support that when we implement "Continue in Create from here" or something. But that is for another day.

## Testing
For Sanity folks: follow the instructions [here](https://www.notion.so/sanityio/Start-in-Studio-edit-in-Create-12678d055b948065a6d8cdcab74d6078?pvs=4) before starting .

Locally this integration can be tested by running the new test studio: `dev/test-create-integration-studio`. It has `fallbackOrigin` set, and will therefor have "Start in Sanity Create" on localhost.

The integration can also be tested on https://create-integration-test.sanity.studio – it has been deployed from this branch.
## Notes for release

The Create <-> Studio integration needs a full documentation article. There are a lot of moving parts.

Since `beta.sanity.startInCreateEnabled` will default to `false`. Ie, the "Start in Sanity Create" button will be OPT IN, **we can stealth launch this feature**. 

Documents that get linked from the Create side will still be put into read-only in the Studio. If this happens, the editor has access to Create mapping and have probably read whatever Create-Studio documentation we have put out.

For everyone else, this PR effectively is a no-op.

TL;DR: **We dont need release notes when this goes out with core**. We will talk about Create-Studio at the next event, and then we can do something, as needed.


packages/sanity/src/core/create/useCreateDocumentUrl.ts
```diff
@@ -0,0 +1,41 @@
+import {useCurrentUser, useUser, useWorkspace} from 'sanity'
+
+import {type CreateLinkMetadata} from './types'
+
+// @ts-expect-error: __SANITY_STAGING__ is a global env variable set by the vite config
+const isStaging = typeof __SANITY_STAGING__ !== 'undefined' && __SANITY_STAGING__ === true
+
+function getCreateBaseUrl(customHost?: string) {
+  //@todo perhaps only support create staging through config
```
snorrees, line 9: TODO: think about how we want to handle this, if at all

packages/sanity/src/core/store/_legacy/user/userStore.ts
```diff
@@ -74,9 +74,11 @@ export function createUserStore({client: _client, currentUser}: UserStoreOptions
 
   userLoader.prime('me', userFromCurrentUser)
 
-  if (userFromCurrentUser?.id) {
+  //@todo figure out if we are ok with this removal
```
snorrees, line 77: TODO/FIXME/COMMIT TO THIS:

Currently, `useUser(userId)` will return a `User` with `sanityUserId` (global user id), _unless_ the `userId` happens to be the current user id. This is surprising.

The reason is this line of code, which puts `currentUser` into the user cache. `currentUser` obtained from `/me` does NOT have all propereties obtained using the `/user/:id` endpoint.

I make the case here that we should just delete this line, and allow the extra request, so that global token for current user can be obtained like so:

```ts
const userId = currentUser?.id ?? ''
const [user] = useUser(userId)
// user.sanityUserId
```

If there are good reasons not to do this, I can restore this line, and add some bespoke `/user/currentUserId` fetch in the integration codepath that needs it.

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -533,7 +535,8 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
       isLocked ||
       isDeleting ||
       isDeleted ||
-      isLiveEditAndDraft
+      isLiveEditAndDraft ||
+      isCreateLinked
```
snorrees, line 539: Not clear from the diff here, but this sets the documentPane `readOnly` property.

Note: Not all codepaths respect document path `readOnly`state. For instance, `Paste document` can mutate a document pane in `readOnly` mode. I contend that this is the right place to put this `readOnly` code for Create though; code that does not respect it should be corrected elsewhere.

packages/sanity/src/structure/panes/document/document-layout/DocumentLayout.tsx
```diff
@@ -85,6 +88,9 @@ export function DocumentLayout() {
   const zOffsets = useZIndex()
   const previewUrl = usePreviewUrl(value)
 
+  const createLinkMetadata = getCreateLinkMetadata(value)
+  const CreateLinkedBanner = useSanityCreateConfig().components?.documentLinkedBanner
```
snorrees, line 92: This is done to avoid having to put component code belonging to the Create integration in the structure package.

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBar.tsx
```diff
@@ -1,23 +1,34 @@
...
 const CONTAINER_BREAKPOINT = 480 // px
 
 export function DocumentStatusBar(props: DocumentStatusBarProps) {
-  const {actionsBoxRef} = props
-  const {editState, timelineStore} = useDocumentPane()
+  const {actionsBoxRef, createLinkMetadata} = props
+  const {editState, timelineStore, onChange: onDocumentChange} = useDocumentPane()
+  const {title} = useDocumentTitle()
+
+  const CreateLinkedActions = useSanityCreateConfig().components?.documentLinkedActions
```
snorrees, line 31: This is done to avoid having to put component code belonging to the Create integration in the structure package.

packages/sanity/src/core/create/components/CreateLearnMoreButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+import {useTranslation} from '../../i18n'
+import {createLocaleNamespace} from '../i18n'
+
+export const CreateLearnMoreButton = forwardRef(function CreateLearnMoreButton(
+  props,
+  ref: ForwardedRef<HTMLButtonElement>,
+) {
+  const {t} = useTranslation(createLocaleNamespace)
+  return (
+    <Button
```
snorrees, line 14: TODO: as='a' and add href

packages/sanity/src/core/create/start-in-create/StartInCreateDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <StartInCreateSvg />
+      </Box>
+      <Box>
+        <Text weight="semibold">{t('start-in-create-dialog.lede')}</Text>
+      </Box>
+      <Box>
+        <Text>{t('start-in-create-dialog.details')}</Text>
+      </Box>
+      <Flex gap={2} align="center">
+        <Checkbox id={checkboxId} />
```
snorrees, line 83: TODO checkbox state & store in localstorage.

packages/sanity/src/core/create/start-in-create/StartInCreateAction.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {type AppIdCache} from '../studio-app/appIdCache'
+import {useStudioAppIdStore} from '../studio-app/useStudioAppIdStore'
+import {type CreateLinkedSanityDocument} from '../types'
+import {CreateLinkingDialog} from './CreateLinkingDialog'
+import {StartInCreateDialog} from './StartInCreateDialog'
+
+// The "Start in Create" action must be sorted first, so we need a sort key; the action string –
+// we also don't want this string in the config interfaces, so we need the cheeky cast to smuggle it through
+export const START_IN_CREATE_ACTION_NAME =
+  'startInCreate' as unknown as DocumentActionComponent['action']
```
snorrees, line 19: Suggestions on alternative ways to approach this is more than welcome!

packages/sanity/src/core/create/components/CreateLearnMoreButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+import {Button} from '../../../ui-components'
+import {useTranslation} from '../../i18n'
+import {createLocaleNamespace} from '../i18n'
+
+export const CreateLearnMoreButton = forwardRef(function CreateLearnMoreButton(
+  props,
+  ref: ForwardedRef<HTMLButtonElement>,
+) {
+  const {t} = useTranslation(createLocaleNamespace)
+  // TODO - get the correct link
```
snorrees, line 13: TODO!

packages/sanity/src/core/store/_legacy/user/__tests__/userStore.test.ts
```diff
@@ -78,15 +78,15 @@ describe('userStore', () => {
       ])
     })
 
-    it('does not refetch current user if passed when constructing user store, normalizes', async () => {
+    it('refetches current user if passed when constructing user store, as currentUser does not contain sanityUserId (global id)', async () => {
```
snorrees, line 81: If we agree on the change in userStore (se below), I will rewrite this test to show that there is refetching going on. This is just a quickfix to get it passing for now.

packages/@sanity/types/src/schema/definition/type/common.ts
```diff
@@ -24,6 +24,25 @@ export type FieldGroupDefinition = {
...
+}
+
+/**
+ * `BaseOptions` applies to all type options.
+ *
+ *  It can be extended by interface declaration merging in plugins to provide generic options to all types and fields.
+ *
+ *  @public
+ *  */
+export interface BaseSchemaTypeOptions {
```
snorrees, line 42: Are we ok with this base option approach? Imo it aligns well with other type-bases.

packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBar.tsx
```diff
@@ -33,6 +44,22 @@ export function DocumentStatusBar(props: DocumentStatusBarProps) {
...
   const shouldRender = editState?.ready && typeof collapsed === 'boolean'
 
+  let actions: JSX.Element | null = null
+  if (createLinkMetadata && isCreateLinked(createLinkMetadata) && CreateLinkedActions) {
+    actions = (
+      <CreateLinkedActions
+        metadata={createLinkMetadata}
+        panelPortalElementId={DOCUMENT_PANEL_PORTAL_ELEMENT}
+        onDocumentChange={onDocumentChange}
+        documentTitle={title ?? ''}
```
snorrees, line 54: Is there a hook for getting default document title? Or is there a locale-resouce with "Untitled" I could use?

packages/sanity/src/core/create/components/media/CreateInfoSvg.tsx
```diff
@@ -0,0 +1,6 @@
+import {Card} from '@sanity/ui'
+
+/* @todo replace with real art */
```
snorrees, line 3: TODO!

packages/sanity/src/core/create/components/media/StartInCreateSvg.tsx
```diff
@@ -0,0 +1,6 @@
+import {Card} from '@sanity/ui'
+
+/* @todo replace with real art */
```
snorrees, line 3: TODO!

packages/sanity/src/core/create/start-in-create/StartInCreateDialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+import {useTranslation} from '../../i18n'
+import {useWorkspace} from '../../studio'
+import {CreateLearnMoreButton} from '../components/CreateLearnMoreButton'
+import {StartInCreateSvg} from '../components/media/StartInCreateSvg'
+import {createLocaleNamespace} from '../i18n'
+import {getCreateLinkUrl} from '../useCreateDocumentUrl'
+import {useGlobalUserId} from '../useGlobalUserId'
+import {useSanityCreateTelemetry} from '../useSanityCreateTelemetry'
+
+export const CREATE_LINK_TARGET = 'create'
```
snorrees, line 15: We might not want a fixed target, and use _blank instead.

packages/sanity/src/core/create/__telemetry__/create.telemetry.ts
```diff
@@ -0,0 +1,32 @@
+import {defineEvent} from '@sanity/telemetry'
```
svirs, line 1: cc @sanity-io/data-eng for review

snorrees: The Create side equivalents will land in a couple of days fyi

packages/sanity/src/core/store/_legacy/user/userStore.ts
snorrees: Update: We found issues with this approach and will remove the "global-id-in-url" approach.
As it turns out the user can log in with a different user in Create, resulting in the link-creation failing. Instead Create will handle all this, so we no longer need global user id in Studio.

TLDR: all code related to user.id will be reverted/removed once Create has been updated with this new behavior.

packages/sanity/src/core/create/useGlobalUserId.ts
```diff
@@ -0,0 +1,9 @@
+import {useCurrentUser, useUser} from '../store'
+
+export function useGlobalUserId(): string | undefined {
```
snorrees, line 3: All usage of this will be removed.

packages/@sanity/types/src/schema/definition/type/common.ts
bjoerge: Seems perfectly fine to me 👍

packages/sanity/src/core/config/prepareConfig.ts
```diff
@@ -495,14 +498,16 @@ function resolveSource({
...
+      actions: (partialContext) => {
+        const actions = resolveConfigProperty({
           config,
           context: {...context, ...partialContext},
           initialValue: initialDocumentActions,
           propertyName: 'document.actions',
           reducer: documentActionsReducer,
-        }),
+        })
+        return getStartInCreateSortedActions(actions)
```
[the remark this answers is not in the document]
bjoerge: Ultimately a product question, but this happens only if 1) create integration is enabled in studio config and 2) only for new documents, right? If that's the case it seems fine to me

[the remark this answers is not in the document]
bjoerge: (nvm, I see you called that out already in the PR description)

snorrees (commented):
I just reverted the user store changes: The "Start in Create" link no longer relies on global user id.

packages/@sanity/types/src/user/types.ts
```diff
@@ -23,4 +23,7 @@ export interface User {
   displayName?: string
   imageUrl?: string
   email?: string
+
+  /** global sanity user id */
+  sanityUserId?: string
```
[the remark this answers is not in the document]
snorrees: Reverted (the data is still there though, types are not complete)

packages/sanity/src/core/store/_legacy/user/userStore.ts
snorrees: Reverted.

packages/sanity/src/core/store/_legacy/user/__tests__/userStore.test.ts
snorrees: Reverted.

snorrees:
PR and description updated with a polish-pass by Rob 🙇

packages/sanity/src/core/create/components/CreateLearnMoreButton.tsx
snorrees: Chasing docs team for this.

packages/sanity/src/core/create/__telemetry__/create.telemetry.ts
svirs: @snorrees These seems to follow the copy of the buttons directly. Could we instead make them more about the underlying action? Perhaps instead of "Start In Create Clicked" we can do "Author In Create Clicked" or along those lines

packages/sanity/src/core/create/__telemetry__/create.telemetry.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const StartInCreateClicked = defineEvent({
+  name: 'Start In Create Clicked',
```
svirs, line 4: `Authoring In Create Started`

packages/sanity/src/core/create/__telemetry__/create.telemetry.ts
```diff
@@ -0,0 +1,32 @@
...
+  version: 1,
+  description: 'The "Start in Sanity Create" button is clicked.',
+})
+
+export const StartInCreateAccepted = defineEvent({
+  name: 'Start In Create Accepted',
```
svirs, line 10: `Authoring In Create Accepted`

RitaDias:
👋 
I haven't looked at the codey, et but from reading the description I have some thoughts @snorrees 

>I have added sanityCreate options directly to BaseSchemaTypeOptions. If it feels controversial to have them there, I can move them into a module declaration extension in core, but I feel that is only adding indirection with no upside.
I agree, I don't particular like having them there, however, I also don't see a reason why we shouldn't. While the sanityCreate is not technically "core" I still think we should make the life easier since it's a close tool to the studio.

> We show the "Start in Sanity Create" button when:
> * beta.create.startInCreateEnabled: true. Atm, startInCreateEnabled defaults to true. We might want to flip this to false before merging, depending on what Product wants.
> * the document is new (ie, no _createdAt)
❗ While the first point might be confusing for us that jump in and don't know where it came from, I think it's fine since we can easily turn it off.
However, I don't think that "when a document is new" should should it by default. Is there a reason why we didn't put it alongside the rest of the document actions in the ... menu? I'm asking this because, for me, it doesn't make sense that within the studio where the main actions should be around the documents themselves (publishing, discarding etc) is taken over by the linking to a different tool all together simply because a beta property has been turned to true 🤔 Or, at least, this feels strange at first impact

> * Do these code-paths need e2e tests?
I'd say no, especially since it spans two different tools I think that e2e tests aren't needed when unit ones exist.

snorrees:
@RitaDias button location: it's a product decision ordered by Even and signed for by Magnus. Yes, default true, too.

snorrees:
Also, do keep in mind the "deployed studio app with manifest needs to exist bit". I know this seems a bit crazy, studios suddently have a button in prod but not in dev. 

I'll raise this one more time internally (ie can we ship with default false), but it IS by ordered by decree.

robinpyon:
> However, I don't think that "when a document is new" should should it by default. Is there a reason why we didn't put it alongside the rest of the document actions in the ... menu? I'm asking this because, for me, it doesn't make sense that within the studio where the main actions should be around the documents themselves (publishing, discarding etc) is taken over by the linking to a different tool all together simply because a beta property has been turned to true 🤔 Or, at least, this feels strange at first impact

@RitaDias 

The rationale behind placing this as the primary action is mostly centered around visibility. We did explore putting it in the document actions menu as you've suggested, though it's very easily missed – especially given that we only want to kick off the above flow when the document is in a pristine, untouched state. We did run a few iterations of this by the design team!

It's slightly "visually breaking" in that users who compose a custom document action to appear first will no longer see that immediately, though 1. it will still be accessible in the document actions menu and 2. it will still be listed as first once the document is no longer in a pristine state.

packages/sanity/src/core/config/prepareConfig.ts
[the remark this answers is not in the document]
RitaDias: My only thought as someone that doesn't know this part of the code super well is that some folks might wonder why, when trying to set a primary action themselves, they can't do it because the link always goes first.

I will say that it might be that you all have set up an exception for this and the use case, in itself, is an edge case that is out of scope for now.

packages/sanity/src/core/create/createStorage.ts
RitaDias, line 1: Interesting!
A vague thought, is that it might useful in the future to move this outside of the /create.

packages/sanity/src/core/create/studio-app/appIdCache.ts
```diff
@@ -0,0 +1,39 @@
+export interface CachedAppId {
+  appId: string | undefined
+}
+
+export interface AppIdCache {
+  get: (args: {projectId: string; appIdFetcher: AppIdFetcher}) => Promise<CachedAppId | undefined>
+}
+export type AppIdFetcher = (projectId: string) => Promise<string | undefined>
+
+export function createAppIdCache(): AppIdCache {
```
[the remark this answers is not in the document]
RitaDias: I don't think there is 🤔

packages/sanity/src/core/create/components/CreateLinkedDocumentBanner.tsx
RitaDias, line 1: I understand there was some discussion about the styling of this banner. I see that this is not using the banner component (which I understand since that component exists in the structure folder). Not using that component will mean that when we make changes to the banner in the future (and we are doing that soon) this will not look like the rest. 

Is it just the fact that the Banner.tsx is in the structure the reason why it wasn't used?

packages/sanity/src/core/create/context/SanityCreateConfigProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * @internal
+ */
+export function SanityCreateConfigProvider(props: SanityCreateConfigProviderProps): JSX.Element {
+  const {children} = props
+  const {beta} = useSource()
+  const value = useMemo((): SanityCreateConfigContextValue => {
+    return {
+      ...beta?.create,
+      startInCreateEnabled: !!beta?.create?.startInCreateEnabled,
+      components: {
```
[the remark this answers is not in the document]
RitaDias: I think my only concern about this is what I have commented previously about the banners.

Technically you could also export these directly from the sanity package. Is there a reason why that wasn't considered vs this approach? 🤔

packages/sanity/src/core/create/createDocumentUrls.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// @ts-expect-error: __SANITY_STAGING__ is a global env variable set by the vite config
+const isStaging = typeof __SANITY_STAGING__ !== 'undefined' && __SANITY_STAGING__ === true
+
+function getCreateBaseUrl(customHost?: string) {
+  const host = (customHost ?? isStaging) ? 'create-staging.sanity.build' : 'www.sanity.io'
+  return `https://${host}/app/create`
+}
+
+export function getCreateLinkUrl(args: {
```
RitaDias, line 11: For a second I got confused 😄 
I thought that this was creating a Link, but it's "Create" the name of the Tool / App. No change needed but I see this confusion as being somewhat common place moving forward 😂

RitaDias (changes requested):
I am unsure if this is a request change or a comment, but basically there are only two things really beneath regarding the banner and importing / exporting :)

packages/sanity/src/core/config/prepareConfig.ts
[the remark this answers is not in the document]
snorrees: This is entirely intentional, since Product wants to faceplant this feature 🤷 

Devs have two options:
* disable start increate 
* OR filter out the action using actions: prev => {}. 

Either approach will have the same result: No start in create button.
If you know how to do this via the config api I'm all ears though! I would _love_ for this to end up first, but be moveable into the menu.

In a perfect world the plugin api had more expressive power, where we could set priority on actions or something, which would allow overriding when you really want to.

The alternative to this is to do something more drastic like I had to do for "Open in Sanity Create" and "unlink", ie a completly different codepath in structructure. _That_, imo, is even worse of a smell.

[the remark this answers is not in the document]
RitaDias: No no that does sound worse.
I unfortunately don't know how to do it off the top of my head since I don't know that part of the code that well. I'm assuming there is a way of doing: "if there is a custom primary action, then apply that first, otherwise do the open in sanity". 🤔 

However, this is non-blocking because, like you mentioned, folks can disable it if they want or they can filter it out

packages/sanity/src/core/create/components/CreateLinkedDocumentBanner.tsx
snorrees: I didnt know about Banner! Let me see if I can accommodate

packages/sanity/src/core/create/createDocumentUrls.ts
snorrees: For the record, i tried to use `SanityCreate` in code that is exported out of the `create` directory.

packages/sanity/src/core/create/context/SanityCreateConfigProvider.tsx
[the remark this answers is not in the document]
snorrees: I got to admit , I did not consider that approach. I've tried to limit what gets exported, as not to pollute the global `sanity´ api, was the thought. Therefore, only the props are exported, ie the interface. The implementations are internal.

I'd be happy to export the components, and ditch the context approach if you prefer. 
To me, it just seems non-idiomatic to keep implementations without an interface in one package and use directly in another. I realize now that the way I'm going about this might break established `sanity` conventions though, so I'm happy to take your lead on this.

packages/sanity/src/core/create/createStorage.ts
snorrees: This file is basically copying the pattern used elsewhere in `sanity` for localstorage.

packages/sanity/src/core/create/context/SanityCreateConfigProvider.tsx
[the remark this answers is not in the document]
RitaDias: I don't think there is a conventional way per se of doing things; think of it more as a "mindless way that we've been going about it" than anything intentional :)

I just wanted to double check. I don't have a strong opinion about the way of exporting in particular and I'd be ok with this approach :)

packages/sanity/src/core/config/prepareConfig.ts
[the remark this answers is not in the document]
snorrees: Just doing some final tests, and there is actually a (hacky) way to move the action into the menu.
The gist is: on the root `defineConfig`, intercept the "Start in Sanity Create" action, and remove the action prop! 🙈 

```
  document: {
    actions: (prev) => {
      return prev.map((a) => {
        if (a.action !== ('startInCreate' as any)) {
          return a
        }
        //returs a new function that does not have a action prop
        return (props) => {
          return a(props)
        }
      })
    },
  },
```

image

Yup, its pretty dumb, but it _can_ be done ;)

snorrees:
@RitaDias I'm happy to report that `beta.sanity.startInCreateEnabled` will ship `false` as the default. Ie, it will be OPT IT to have the "Start in Create" button.

packages/sanity/src/core/create/__telemetry__/create.telemetry.ts
snorrees: should all be fixed now @svirs

RitaDias (approved):
Great work :)

PR #7643, opened 2024-10-21
fix(core): inherit readOnly state from ancestors in copyPaste function

pedrobonamin:
### Description

When pasting values into a field which inherits the `readOnly` from the parent this was not verified., so  it was possible to paste into read only fields.

This PR adds a function `resolveReadOnlyAncestor` to check if any ancestor of the path in which the user is pasting is read only, if it is, it won't allow pasting.

E.g, pasting into `inheritedReadOnlyField` was allowed, the same is happening with arrays.
```ts
    {
      name: 'readOnlyObject',
      title: 'Read only object',
      type: 'object',
      readOnly: true,
      fields: [
        {
          name: 'selfDefinedReadOnlyField',
          title: 'Read only field',
          description: 'ReadOnly defined in field',
          type: 'string',
          readOnly: true,
        },
        {
          name: 'inheritedReadOnlyField',
          title: 'Read only field',
          description: 'ReadOnly inherited from object',
          type: 'string',
        },
      ],
    }

```

https://github.com/user-attachments/assets/0ea1cbc5-f566-48d1-8a7e-11633baebcb1

### What to review

Is the implementation correct? Accessing the parents using the `resolveReadOnlyAncestor` is that in line with the existing approaches?

### Testing
New tests have been added into the copy paste to guard this functionality.
You can test it manually visiting https://test-next-studio-git-sdx-1643.sanity.dev/test/structure/input-standard;objectsTest;0689d2c6-5305-42e5-b381-32c335b3fd60 and trying to copy-paste values into the disabled fields

### Notes for release
fixes an issue in copy paste in which the `readOnly` state was not correctly inherited to the children fields.


sjelfull (approved):
Looks great, thank you!

sjelfull (approved):

PR #7654, opened 2024-10-22
fix: warnings on React 19

stipsan:
### Description

On React 19, on Next 15 in `next dev` in particular, we got warnings that show up as errors, and that aren't actionable in userland as the problem is in our code.
One half of it [was fixed](https://github.com/sanity-io/ui/pull/1430) in `@sanity/ui` #7653
The rest is in this PR.

The errors seen in userland are:

image

image

As well as this console warning:
Screenshot 2024-10-22 at 15 29 10

### What to review

These are minor changes but have a glance to see if we're avoiding any regressions.

### Testing

You can test the before/after by running `pnpm dev:next-studio` on `next`, then on this branch.
Opening http://localhost:3333/test/structure/author, and the help menu:
image
Should be enough to both trigger the dialog, and the `console.warn` instances.

### Notes for release

Fixed warnings on React 19:
- Accessing element.ref was removed in React 19.
- Invalid prop `tabIndex` was supplied to `React.Fragment`.

### DIFF
--- a/dev/test-next-studio/next-env.d.ts
+++ b/dev/test-next-studio/next-env.d.ts
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/basic-features/typescript for more information.
+// see https://nextjs.org/docs/app/building-your-application/configuring/typescript for more information.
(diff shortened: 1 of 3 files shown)


pedrobonamin (approved):
Thanks for fixing this @stipsan I've tested it locally and it looks good

PR #7658, opened 2024-10-22
fix: scheduled pub scroll list issue

jordanl17:
### Description

The scroll on the list of schedules in Scheduled Publish tool does not work.

This PR makes a small change to the overflow style to allow for scroll

NOW:
schedPubScrollPR

### Notes for release
Fixing an issue with the Scheduled Publishing plugin where lists of schedules were not scrollable


pedrobonamin (approved):
Thanks @jordanl17 this looks good to me

PR #7672, opened 2024-10-24
Feat/corel 218/version document status

juice49:


jordanl17:
Not 100% certain on the impact/relevance of this work, but given that it is COREL related and this feature has now reached a settled state, I'm going to close. If necessary, feel free to re-open

PR #7680, opened 2024-10-28
feat(releases): schedule and unschedule releases

bjoerge:


packages/sanity/src/core/releases/tool/components/ReleasePublishAllButton/ReleaseScheduleButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+  }, [
+    schedule,
+    release._id,
+    release.metadata.intendedPublishAt,
+    release.metadata.title,
+    telemetry,
+    toast,
+    t,
+  ])
+  const {t: coreT} = useTranslation()
```
RitaDias, line 84: Not necessary before merging and a nitpick, but you probably don't need to rename this since you are only using one LocaleNamespace :)

RitaDias: Ah nevermind just realised you are importing the namespace up there!

PR #7702, opened 2024-10-30
feat: add support for React Compiler beta

stipsan:
### Description

The [React Compiler is now in Beta](https://react.dev/blog/2024/10/21/react-compiler-beta-release#backwards-compatibility):
- [It has backwards compatibility with React 18](https://react.dev/blog/2024/10/21/react-compiler-beta-release#backwards-compatibility)
- [The Working Group seeks feedback from library usage](https://react.dev/blog/2024/10/21/react-compiler-beta-release#using-react-compiler-in-libraries)
- [It now supports function outlining, a very effective memozation technique that is incredibly costly to do manually](https://youtu.be/qd5yk2gxbtg?t=566)

This PR adds the React Compiler to the Studio in two ways:
#### 1. Pre-compiling `sanity` and `@sanity/vision`

We've been testing it for some time now in these libraries:
- `react-rx`
- `@sanity/ui`
- `@portabletext/editor`
And it has worked really well for us. `react-rx` is the only library that is fully compiled at the moment, while:
- `@portabletext/editor` is nearly there (9 out of 13 components compiled, 11 linter warnings left to fix)
- `@sanity/ui` also close, 290 out of 305 components, 10 linter warnings left to fix

<details>
<summary>
Why is `@sanity/ui` reporting a higher number of bailouts (15 components) than the linter?
</summary>

To see the report of components that can be compiled, run `npx react-compiler-healthcheck`. The `eslint-plugin-react-compiler` is intended to only report issues that are actionable, things you should fix that allows the component to be compiled. Thus it's normal to see discrepancies like in `@sanity/ui`. You can opt-in the linter to report all bailouts, including those that are due to patterns that will be supported in the future and are considered to be valid: `pnpm eslint --cache --no-inline-config --no-eslintrc --ignore-pattern '**/__tests__/**' --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --plugin react-hooks --rule 'react-compiler/react-compiler: [error,{__unstable_donotuse_reportAllBailouts: true}]' src`.

</details>

The benefit of pre-compiling is that "it just works", people userland bump their deps and the Studio is now faster ✨ It supports React 18 and upwards.

<details>

<summary>
Plugin authors that use `@sanity/pkg-utils` can ship pre-compiled plugins as well
</summary>

1. Install the compiler and the runtime: `npm install --save-exact --save-dev babel-plugin-react-compiler@beta && npm install --save-exact react-compiler-runtime@beta` (it's important that exact versions are used while in beta, and that `react-compiler-runtime` is a direct dependency, not in `devDependencies`).
2. Enable the compiler in the `package.config.ts`:
   ```ts
   babel: {reactCompiler: true},
   reactCompilerOptions: {target: '18'},
   ```

It makes sense to add it to `@sanity/plugin-kit` once the React Compiler itself is GA and no longer beta.

</details>

##### `@sanity/vision`

Running `cd packages/@sanity/vision/src && pnpx react-compiler-healthcheck` yields:
> Successfully compiled 13 out of 13 components.

That doesn't mean everything is compiled though, [as the main component `<VisionGui />` is a class component and thus is unsupported by the React Compiler](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/%40sanity/vision/src/components/VisionGui.tsx#L153). The difficulty of auto memoizing class components was one of the motivators behind the introduction of React Hooks and the shift away from classes, and so `<VisionGui />` will have to be refactored to a function component for it to be optimised fully.

##### `sanity`

Running `cd packages/sanity/src && pnpx react-compiler-healthcheck` yields:
> Successfully compiled 1218 out of 1411 components.

86% coverage is awesome, at the same time it's important to consider that:
- The remaining components will take time to migrate, as they are complex and sensitive to regressions, for example here are some key ones:
    - [`<CommandList />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/core/components/commandList/CommandList.tsx#L191)
    - [`<GetFormValueProvider />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/core/form/contexts/GetFormValue.tsx#L18)
    - [`<RouteScope />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/router/RouteScope.tsx#L74)
    - [`<DocumentPaneProvider />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx#L299-L305)
    - [`<DocumentPanel />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx#L83)
    - [`<FormView />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx#L119)
    - [`<ChangesInspector />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/structure/panes/document/inspectors/changes/ChangesInspector.tsx#L99)
    - [`<DocumentListPaneContent />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/structure/panes/documentList/DocumentListPaneContent.tsx#L253)
- We have class components that are not part of this list, that do need to be refactored to function components for them to be auto memoized:
    - [`<ArrayOfPrimitivesInput />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/core/form/inputs/arrays/ArrayOfPrimitivesInput/ArrayOfPrimitivesInput.tsx#L29)
    - [`<BaseFileInput />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/core/form/inputs/files/FileInput/FileInput.tsx#L79)
    - [`<ToolCanvasLegacy />`](https://github.com/sanity-io/sanity/blob/e4338ac534ca3b6211a3d642adbb34117414a9e3/packages/sanity/src/core/form/inputs/files/ImageToolInput/imagetool/ToolCanvas.tsx#L206)

Until these are handled we won't see the full potential of the React Compiler, as it's not invulnerable to unstable props:
```tsx
import {useEffect} from 'react'
import {OptimizedComponent} from './OptimizedComponent'

export default function UnoptimizedComponent(props) {

  useEffect(() => {
    props.onMount()
    // Because of disabling the linter, the compiler will skip over `UnoptimizedComponent` and not optimize it as it can't guarantee that memoizing it is safe from regressions
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])


  return (
    <OptimizedComponent
      foo={['bar', 'baz']} // this creates a new array on every render, negating the memozation in OptimizedComponent that tries to prevent unnecessary rerenders
    />
  )
}
```
In this scenario the performance improvement on `OptimizedComponent` will appear minimal, until one day `UnoptimizedComponent` is fixed and now you'll have a dramatic delta as the React Compiler is able to deliver fine grained memoization and reduced rerenders to only what's absolutely necessary, end to end.

Also worth noting is that there are some known issues with `@tanstack/react-table` and the React Compiler, so I've added `'use no memo'` on the affected files to disable the compiler there. The way references are used in that library is highly unsafe, and is causing React to no longer be fully reactive:
- [Calling refs during render](https://github.com/TanStack/table/blob/6b4d616dd7c8917616eb4fecaf09dda7030fd115/packages/react-table/src/index.tsx#L78)
- [Returning refs during render](https://github.com/TanStack/table/blob/6b4d616dd7c8917616eb4fecaf09dda7030fd115/packages/react-table/src/index.tsx#L93)
And the effect it has is that when React Compiler memoizes our code, `@tanstack/react-table` stops re-rendering when it should because it's relying on unstable dependencies to reach eventual consistency and with the React Compiler enabled they're no longer unstable.

If we run into similar issues the best resolution is to add `'use no memo'` at the top of that file so that we don't need to disable the React Compiler everywhere if a regression is discovered.

The strategy moving forward is to gradually refactor components so they can be compiled, and to work with the React Core Team and report issues or missing functionality so that the React Compiler can learn to understand patterns we use that follow the Rules of React and are safe, but not yet supported.
With this strategy we should see the Studio codebase get gradually faster, and when we hit critical coverage it'll suddenly speed up significantly when unstable props and context values no longer poke holes through the `1218` components that are memoized from the start.

#### 2. Allow userland to opt-in to React Compiler

As mentioned in the previous section, even if something is fully memoized by the React Compiler, its performance improvements can be negated if parent components are sending props or context provider values that are unstable.
Thus custom studio components and code can have a major impact on the overall studio performance. We've seen customers face severe degradation simply because a single function given as a prop forgot to use the `React.useCallback` wrapper.
Such scenarios can be prevented if the React Compiler is enabled on custom studio code.

Additionally, [the React Compiler Working Group is now open for everyone](https://react.dev/blog/2024/10/21/react-compiler-beta-release#opening-up-react-compiler-working-group-to-everyone). Sanity users with complex studios is a treasure trove of code that can test the correctness of React Compiler, and we'll benefit from giving them the opportunity to participate in the working group, and give their feedback.

For users on React 18, opt-in involves installing `react-compiler-runtime@beta` as a direct dependency, and `babel-plugin-react-compiler@beta` as a dev dependency, and then add this to their `sanity.cli.ts` to enable the compiler:
```ts
import {defineCliConfig} from 'sanity/cli'

  api: {
    projectId: 'abc123',
    dataset: 'production',
  },
  reactStrictMode: true,
  reactCompiler: {target: '18'},
})
```

For React 19 you only need `babel-plugin-react-compiler@beta` in your dev dependencies, and:
```ts
import {defineCliConfig} from 'sanity/cli'

  api: {
    projectId: 'abc123',
    dataset: 'production',
  },
  reactStrictMode: true,
  reactCompiler: {target: '19'},
})
```

[The official docs can then be used to decide how to adopt it, test it and more](https://react.dev/learn/react-compiler).
Until the React Compiler is stable, it's best to leave it as opt-in, and to require installing the `babel-plugin-react-compiler` and `react-compiler-runtime` dependencies in userland. It enables userland to test bugfixes that are only available on `babel-plugin-react-compiler@experimental`, and decouples the version we use to pre-compile `sanity` from userland.
Decoupling them is important, to ensure that userland with React Compiler can update deps like `sanity` anytime they like, without risking changes to how the React Compiler itself works and optimises their code. They are in control over that with when they update `babel-plugin-react-compiler`. It's also useful to have `babel-plugin-react-compiler` and `eslint-plugin-react-compiler` on the same exact version, so that they behave the same.

### What to review

I've published a prerelease as `pnpm add sanity@react-compiler` that can be used to more accurately experience how it's like in userland.
On the Test Studio on the monorepo, due to how we enable HMR and such, it's easier to always enable the React Compiler. On our monorepo, enabling React Compiler enables it for `sanity`, `@sanity/vision` as well as custom studio code. There's no easy way on the monorepo to test scenarios where only `sanity` and `@sanity/vision` is compiled, and not userland. This is due to `@sanity/pkg-utils` only being used during publishing, and not on the monorepo.
Other differences are that the monorepo installs `babel-plugin-react-compiler` and `react-compiler-runtime`, so it's difficult to test what happens if userland wants to use the compiler but forgot to install these deps. `pnpm add sanity@react-compiler` allows you to experience that.

And with that said, there where a couple of things I'd want the CLI to do but I couldn't figure out:

#### We should detect if the wrong `reactCompiler.target` is used

If `react` is on version 18, then setting this should fail on `sanity dev` or `sanity build`:
```ts
{
  reactCompiler: {target: '19'}
}
```
as `target` must be `18`.
Likewise, if `package.json` says: `"react": "19.0.0-rc-f994737d14-20240522"` then `target` should be `19`.
We could warn here instead of erroring, as it'll still work, it's just not ideal as the native React Compiler runtime that ships in React 19 performs better than the `react-compiler-runtime` polyfill.

#### We should validate that `react-compiler-runtime` is installed

If we're on `react` 18 we should throw on `sanity dev` and `sanity build` if `react-compiler-runtime` isn't installed as a direct dependency:

```ts
{
  reactCompiler: {target: '18'}
}
```

These should pass if it's in the `package.json` that has `react`, `react-dom`, `sanity` and `styled-components` installed:
```json
{
  "dependencies": {
    "react-compiler-runtime": "19.0.0-beta-a7bf2bd-20241110",
    "react-compiler-runtime": "0.0.0-experimental-678214a-20241111",
    "react-compiler-runtime": "beta",
    "react-compiler-runtime": "experimental",
    "react-compiler-runtime": "latest"
  }
}
```

If the semver includes `^`, `*` or other in-exact ranges then it should fail as it's not safe. Only exact version numbers, or a dist-tag name, is safe.
Likewise, it shouldn't be a `devDependencies` nor `peerDependencies`.

#### We should validate that `babel-plugin-react-compiler` is installed

Since we intentionally don't ship `babel-plugin-react-compiler` as a direct dependency of `sanity`, it has to be installed in userland.
We should require it to be in `devDependencies`, and to follow the same semver rules as `react-compiler-runtime`. 
Other tooling like `next` validates for this, as the error coming from `vite` if we don't is difficult to wrap your head around.

#### Ensure that if `eslint-plugin-react-compiler` is installed it is on the same version as `babel-plugin-react-compiler`

Both plugins ship with the same underlying compiler, and are published with the same version numbers and dist-tags. It's important that they match so that you don't end up with a linter that has warnings that the babel plugin now supports and vice versa.

#### Consider `autoUpdates` support

I didn't update `@repo/package.bundle` with the compiler, so studios that have `--auto-updates` in their `sanity build`, or `autoUpdates: true` in their `sanity.cli.ts`, will not see the compiler used on `sanity` or `@sanity/vision`. They will see it on their custom code, and it'll work well for react 19 users, but for react 18 there's going to be some bloat as `react-compiler-runtime` is bundled into the custom studio codebase, instead of using a shared dep like the way we handle `react`, `react-dom` and `styled-components`.

#### Consider easier setup, like Next.js

[Next.js lets you enable the compiler with just a boolean, no need to set a `target`.](https://nextjs.org/docs/app/api-reference/next-config-js/reactCompiler) It would b super sweet if all you needed to do was setting `reactCompiler: true`, and we'd then:
- automatically detect what the `target` should be by checking the `react` major that's installed.
- install `babel-plugin-react-compiler@beta` automatically to dev deps if needed.
- install `react-compiler-runtime@beta` as a direct dep if needed.

I don't know how to add those things to our CLI but if you do feel free to push commits to this PR and add it ❤ 

#### How to test and verify that it works

Components that are memoized gets a new badge in the React Devtools:
￼
You can inspect the regular test studio deployment on this PR, or run `pnpm dev`, as well as `pnpm dev:next-studio` to see it on React 19. It's enabled on all of them (after this merges I'll remove the `pnpm dev:compiled-studio` and its vercel deployment as it's no longer needed).
Worst case these components should work like before, best case they'll work faster.

Additionally, the eFPS testing suite is updated to use the compiler, so its results are a good indicator of how much perf has improved.

#### Maintaining React Compiler deps

Renovatebot is struggling to understand how to deal with the pre-release version hashes used for deps like `pnpm add react-compiler-runtime@beta`.
The `.github/worfklows/react-compiler.yml` workflow added in this PR is already in use on other repos, the PRs it creates looks like this: https://github.com/sanity-io/ui/pull/1468
It's setup to run after the weekly monday beta release [the React Core team runs](https://github.com/facebook/react/blob/989af12f72080c17db03ead91d99b6394a215564/.github/workflows/compiler_prereleases_weekly.yml#L5-L6
  workflow_dispatch:).

### Testing

All the testing suites, vitest, playwright, eFPS, everything is setup to use the compiler, to ensure we test it as much as possible.
I tried to setup a dual test runner setup like we do [on `react-rx`](https://github.com/sanity-io/react-rx/blob/6f06a59bdf629b32ec8a53eaeb056cf3d96a7479/vitest.workspace.ts#L6-L23), as it can be helpful to validate "correctness" (does the component still work the same if the compiler is turned off again), but I couldn't get it to work (maybe @bjoerge knows how?).
It might be useful to run E2E tests the same way, but when it all comes down to it it's not critical.
If something doesn't work, adding `'use no memo'` at the top of that file opts it out on all these layers, allowing us to quickly test correctness whenever we need, and issue hotfixes that solves the worst case scenario of a regression, without needing to turn the compiler entirely off.

### Notes for release

The `sanity` and `@sanity/vision` are now precompiled with the [React Compiler](https://react.dev/learn/react-compiler).
With 1 231 out of 1 411 being compiled, we're already seeing a 20%-30% overall reduction in render time and latency. We expect that to improve significantly as we gradually refactor the remaining 180 components so they can be auto memoized.
For a few weeks now we've already been shipping our Design System `@sanity/ui` with React Compiler, as well as `@portabletext/editor` which is used by PortableText input fields, and `react-rx` which runs vital state management throughout the Studio codebase.

We continue to see impressive results in our profiling and analysis, and with the [release of React Compiler in beta](https://react.dev/blog/2024/10/21/react-compiler-beta-release) we are excited to add support for using it to auto memoize your own studio code:
```bash
pnpm add react-compiler-runtime@beta && pnpm add --save-dev babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```
```bash
yarn add react-compiler-runtime@beta && yarn add --save-dev babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```
```bash
npm install --save-exact react-compiler-runtime@beta && npm install --save-exact --save-dev babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```

And in your `sanity.cli.ts` add:
```diff
import {defineCliConfig} from 'sanity/cli'

export default defineCliConfig({
   api: {
      projectId: 'abc123',
      dataset: 'production',
   },
   reactStrictMode: true,
+  reactCompiler: {target: '18'},
})
```
If you decide to give the React Compiler a try make sure you share your feedback and experience using it with the [Working Group](https://github.com/reactwg/react-compiler/discussions)!

Sanity Plugin authors can also ship plugins that are pre-compiled, if they're using `@sanity/pkg-utils`, by running:
```bash
pnpm add react-compiler-runtime@beta && pnpm add --save-dev @sanity/pkg-utils@latest babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```
```bash
yarn add react-compiler-runtime@beta && yarn add --save-dev @sanity/pkg-utils@latest babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```
```bash
npm install --save-exact react-compiler-runtime@beta && npm install --save-exact --save-dev @sanity/pkg-utils@latest babel-plugin-react-compiler@beta eslint-plugin-react-compiler@beta
```
and changing the `package.config.ts` to:
```diff
import {defineConfig} from '@sanity/pkg-utils'

export default defineConfig({
   // ...
+  babel: {reactCompiler: true},
+  reactCompilerOptions: {target: '18'},
})
```


packages/@sanity/cli/src/types.ts
```diff
@@ -286,6 +286,30 @@ export interface GraphQLAPIConfig {
...
+  sources?: Array<string> | ((filename: string) => boolean) | null
+
+  /**
+   * The minimum major version of React that the compiler should emit code for. If the target is 19
+   * or higher, the compiler emits direct imports of React runtime APIs needed by the compiler. On
+   * versions prior to 19, an extra runtime package react-compiler-runtime is necessary to provide
+   * a userspace approximation of runtime APIs.
+   * @see https://react.dev/learn/react-compiler#using-react-compiler-with-react-17-or-18
+   */
+  target: '18' | '19'
```
stipsan, line 306: The React Compiler supports `17`, but I left it out since we have `18` as our baseline.

RitaDias:
👋 
Just reading through the description, thank you for this!

> sanity
> Running cd packages/sanity/src && pnpx react-compiler-healthcheck yields:
> 
> Successfully compiled 1218 out of 1411 components.
> 
> 86% coverage is awesome, at the same time it's important to consider that:
> 
> The remaining components will take time to migrate, as they are complex and sensitive to regressions, for example here are some key ones:
> (list)
>
> (...)
>
> The strategy moving forward is to gradually refactor components so they can be compiled, and to work with the React Core Team and report issues or missing functionality so that the React Compiler can learn to understand patterns we use that follow the Rules of React and are safe, but not yet supported.
With this strategy we should see the Studio codebase get gradually faster, and when we hit critical coverage it'll suddenly speed up significantly when unstable props and context values no longer poke holes through the 1218 components that are memoized from the start.

I'm not sure if you've raised this somewhere else or where do we want to keep this information so it doesn't get lost, but I feel it's something that should be prioritised to do? :) And since you've already gone through the issue of compiling this list, it makes sense to not lose it!

Question:

- Forgive me if this doesn't make sense, a question: is there a world where wouldn't want this by default? I understand we are now adding it as a property that users can turn on but it feels like an overall useful thing to provide with the studio? Or perhaps there are some caveats / DX things that I am not seeing? Or maybe it's because it's still in beta and not fully out yet? 😅

stipsan:
>I'm not sure if you've raised this somewhere else or where do we want to keep this information so it doesn't get lost, but I feel it's something that should be prioritised to do? :) And since you've already gone through the issue of compiling this list, it makes sense to not lose it!

I wouldn't worry about it too much. For function components you'll see the ones we need to refactor by running `pnpm check:react-compiler`, and you see them on PRs showing up like this (although GitHub caps the list at 10, so it's not showing all 80 warnings): 
image
That was my starting point for creating the list, and if anything should be prioritised I'd say it's to get those warnings down to zero and then we can re-assess what to do next :)

>Forgive me if this doesn't make sense, a question: is there a world where wouldn't want this by default? I think we are now adding it as a property that users can turn on but it feels like an overall useful thing to provide with the studio? Or perhaps there are some caveats / DX things that I am not seeing? Or maybe it's because it's still in beta and not fully out yet? 😅

[The guidance from the core team is to keep it as opt-in until it reaches stable 😮‍💨](https://react.dev/blog/2024/10/21/react-compiler-beta-release#roadmap-to-stable:~:text=The%20Stable%20Release%20will%20mark%20the%20beginning%20of%20a%20new%20foundation%20for%20React%2C%20and%20all%20apps%20and%20libraries%20will%20be%20strongly%20recommended%20to%20use%20the%20compiler%20and%20ESLint%20plugin.)

RitaDias:
> > I'm not sure if you've raised this somewhere else or where do we want to keep this information so it doesn't get lost, but I feel it's something that should be prioritised to do? :) And since you've already gone through the issue of compiling this list, it makes sense to not lose it!
> 
> I wouldn't worry about it too much. For function components you'll see the ones we need to refactor by running `pnpm check:react-compiler`, and you see them on PRs showing up like this (although GitHub caps the list at 10, so it's not showing all 80 warnings): image That was my starting point for creating the list, and if anything should be prioritised I'd say it's to get those warnings down to zero and then we can re-assess what to do next :)
> 
> > Forgive me if this doesn't make sense, a question: is there a world where wouldn't want this by default? I think we are now adding it as a property that users can turn on but it feels like an overall useful thing to provide with the studio? Or perhaps there are some caveats / DX things that I am not seeing? Or maybe it's because it's still in beta and not fully out yet? 😅
> 
> [The guidance from the core team is to keep it as opt-in until it reaches stable 😮‍💨](https://react.dev/blog/2024/10/21/react-compiler-beta-release#roadmap-to-stable:~:text=The%20Stable%20Release%20will%20mark%20the%20beginning%20of%20a%20new%20foundation%20for%20React%2C%20and%20all%20apps%20and%20libraries%20will%20be%20strongly%20recommended%20to%20use%20the%20compiler%20and%20ESLint%20plugin.)

This makes sense :) thank you for answering it 🙏

rexxars (dismissed):

rexxars (approved):
I agree with your suggestions for CLI improvements - I will see if we can make at least some of those changes available in follow-up PR(s) shortly. For now, documenting ("publicly") how to enable the React compiler and which dependencies to install would be a good first step.

stipsan:
>I agree with your suggestions for CLI improvements - I will see if we can make at least some of those changes available in follow-up PR(s) shortly. For now, documenting ("publicly") how to enable the React compiler and which dependencies to install would be a good first step.

Sure @rexxars, I'll get right on that, similar to the react 19 docs page 🙌

PR #7732, opened 2024-11-03
fix: invalid locale namespace specified for create

rexxars:
### Description

The namespace was incorrectly specified for the new `create` strings, which causes issues for the locales repository, as it clobbers the `comments` namespace.

### What to review

Should be pretty obvious ;)

### Testing

Shouldn't be needed.

### Notes for release

None.


binoy14 (approved):
Makes sense to me, thanks!

Issue #7748, opened 2024-11-05
Typegen: Function arguments not evaluating correctly


okj579:
**Describe the bug**

Typegen doesn't correctly bind function parameters while parsing queries from Typescript/Javascript.
It does attempt to evaluate function expressions contained in a query via string interpolation. It manages this as long as the function only has one "layer" of scope (e.g.: `(name) => name`), but throws an error if there are any more.

**To Reproduce**

Run sanity typegen on code with a function like this one:

```ts
import { groq } from "groq";
const getType = (type) => () => () => type;
const query = groq`*[_type == "${getType("author")()()}"]`
const res = sanity.fetch(query);
```

<details>
<summary>or add as a test alongside the one it's based on</summary>

https://github.com/sanity-io/sanity/blob/12cb46b0b7352370db9e66c83b1939737fe82fec/packages/%40sanity/codegen/src/typescript/__tests__/findQueriesInSource.test.ts#L43-L56

```ts
    test('with function with literal parameters', () => {
      const source = `
      import { groq } from "groq";
      const getType = (type) => () => () => type;
      const query = groq\`*[_type == "\${getType("author")()()}"]\`
      const res = sanity.fetch(query);
    `

      const queries = findQueriesInSource(source, 'test.ts')

      const queryResult = queries[0]

      expect(queryResult?.result).toEqual('*[_type == "author"]')
    })
```
</details>

Results in an error:
```console
Error: Could not find binding for node "type" in test.ts:3:44
    at resolveIdentifier (packages/@sanity/codegen/src/typescript/expressionResolvers.ts:298:9)
...
```

**Expected behavior**

The query evaluates to `*[_type == "author"]`.

**Which versions of Sanity are you using?**

<details><summary>3.60.0</summary>
<p>

```
@sanity/cli (global)             3.60.0 (up to date)
@sanity/asset-utils               2.0.6 (up to date)
@sanity/eslint-config-studio      4.0.0 (up to date)
@sanity/icons                     3.4.0 (up to date)
@sanity/language-filter           4.0.2 (up to date)
@sanity/locale-de-de             1.1.11 (up to date)
@sanity/orderable-document-list   1.2.2 (up to date)
@sanity/ui                        2.8.9 (up to date)
@sanity/util                     3.60.0 (up to date)
@sanity/vision                   3.60.0 (up to date)
sanity                           3.60.0 (up to date)
```

</p>
</details> 

**What operating system are you using?**

[...]

PR #7750, opened 2024-11-05
feat(codegen): function resolving

okj579:
### Description

Adds support to Typegen for:
* Export all (`export * from "source.ts"`)
* Nested functions with parameters (#7748)
* Function parameters with default values

### What to review

Only worker code for the Typescript parser was changed. Examples of code that wasn't but is now successfully parsed is in the new tests.

### Testing

Tests were added for the newly supported features.

### Notes for release

Typegen now supports slightly more advanced types of interpolation, e.g. 
```ts
const query = groq`*[_type="asdf"] { 
    "linkA": ${linkQuery('linkA')},
}`;
```
Client code can also import through an export all without confusing Typegen.


sgulseth (approved):
nice! Thank you for helping out here!

PR #7751, opened 2024-11-05
feat(sanity): omit search weights when possible

juice49:
### Description

When conducting a search, consumers may skip sorting via the `skipSortByScore` option. In this scenario, it's not necessary to query for weights.

Examples of this include:

- Document list filtering.
- Global search when sorting by properties other than relevace, such as updated at or created at.


jordanl17:
@juice49 this still seems relevant, shall we open up a PR and I am happy to review

RitaDias (approved):

PR #7754, opened 2024-11-05
fix(core): update getPublishDateFromRelease, add ScheduledRelease type

pedrobonamin:
### Description

We had two functions with similar names doing different things:
- getReleasePublishDate (removed). This one preferred the intended publish date instead of the publishAt
- getPublishDateFromRelease.

This PR removes the `getReleasePublishDate` function

### DIFF
--- a/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
+++ b/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
@@ -172,7 +172,7 @@ export const GlobalPerspectiveMenuItem = forwardRef<
               release.metadata.releaseType !== 'undecided' &&
               (release.publishAt || release.metadata.intendedPublishAt) && (
                 <Text muted size={1}>
-                  {formatRelative(getPublishDateFromRelease(release), new Date())}
+                  {formatRelative(getPublishDateFromRelease(release) || new Date(), new Date())}
                 </Text>
               )}
           </Stack>
--- a/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
+++ b/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
@@ -172,7 +172,7 @@ export const GlobalPerspectiveMenuItem = forwardRef<
               release.metadata.releaseType !== 'undecided' &&
               (release.publishAt || release.metadata.intendedPublishAt) && (
                 <Text muted size={1}>
-                  {formatRelative(getPublishDateFromRelease(release) || new Date(), new Date())}
+                  {formatRelative(getPublishDateFromRelease(release), new Date())}
                 </Text>
               )}
           </Stack>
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/releases/util/util.ts
```diff
@@ -64,28 +64,12 @@ export function getCreateVersionOrigin(documentId: string): VersionOriginTypes {
 }
 
 /** @internal */
-export function getPublishDateFromRelease(release: ReleaseDocument): Date {
+export function getPublishDateFromRelease(release: ScheduledRelease): Date {
```
pedrobonamin, line 67: Updated the type to `ScheduledRelease` to assert that a date will always be returned.

packages/sanity/src/core/releases/store/types.ts
```diff
@@ -32,6 +32,23 @@ export interface ReleaseSystemDocument {
...
+  publishAt?: string
+  metadata: ReleaseDocument['metadata'] & {
+    intendedPublishAt: string
+    releaseType: 'scheduled'
+  }
+}
+/**
+ * @internal
+ */
+export const isScheduledRelease = (release: ReleaseDocument): release is ScheduledRelease =>
```
jordanl17, line 49: Hmm, the naming here is vague to me, because a scheduled release to me has `state` of `scheduled` or `scheduling`. This is like a prescheduled release, no? I wonder if we should introduce a name like `intendedScheduledRelease`?

packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
```diff
@@ -170,7 +170,7 @@ export const GlobalPerspectiveMenuItem = forwardRef<
             </Text>
             {!isPublishedPerspective(release) &&
               release.metadata.releaseType !== 'undecided' &&
-              (release.publishAt || release.metadata.intendedPublishAt) && (
```
jordanl17, line 173: These aren't a like for like swap I don't think, since `isScheduledRelease` only looks at `intendedPublishAt`. I can't recall exactly but it might be worth verifying that when a scheduled release is actually scheduled, and it's possible to reconfirm the scheduled datetime, whether we also update the `intendedPublishAt` to reflex this too.
Screenshot 2024-11-05 at 13 40 15

pedrobonamin: Are we removing the `intendedPublishAt` when scheduling it? 
It's confusing how close this names are. 

Maybe we need to find a better name for the `releaseType ="scheduled"`

jordanl17: I think we leave `intendedPublishAt` as whatever it was, and only persist this new date into `publishAt` via the schedule action. We should agree either to clear it or always keep it aligned. But of course the schedule action can also be triggered outwith studio, so guaranteeing they match isn't possible

jordanl17 (approved):
🔥

PR #7759, opened 2024-11-05
fix(core): explicitly pass `randomKey(12)` as the PTE `keyGenerator`

christianhg:
In an effort to make `@portabletext/editor` become more stand-alone, I'm looking into removing the few `@sanity/*` dependencies that it has. Here, `@sanity/util` is a low-hanging fruit.

If the `keyGenerator` prop isn't passed, the `PortableTextEditor` will use `randomKey(12)` from `@sanity/util`. However, passing this key generator explictly from the `PortableTextInput` removes this implicit dependency. Now, the Studio keeps getting the same type of random keys and `@portabletext/editor` is free to remove the `@sanity/util` dependency and just copy/paste the source code of `randomKey` or use another default key generator.


bjoerge (approved):
Nice, LGTM

PR #7776, opened 2024-11-07
fix(base): support restoring cross dataset references

bjoerge:
### Description
Fixes an issue that cleared cross dataset references when restoring a document from history

### Notes for release
n/a - fixes an issue in v2 which is EOL


jordanl17:
What a fix! 🚀

jordanl17 (approved):

PR #7780, opened 2024-11-07
chore: update Studio teams in CODEOWNERS

juice49:
This commit changes all Studio teams in CODEOWNERS to `@sanity-io/studio`, which is necessary because the `@sanity-io/studio-dx` and `@sanity-io/studio-ex` teams have been removed.


binoy14 (approved):
Makes sense to me, thanks!

rexxars (approved):

PR #7789, opened 2024-11-11
test: enable test coverage

judofyr:
### Description

This adds support for running tests with coverage, both locally and in CI.

### What to review

Some suggestions:

- Run the tests locally (in the form you typically run them) and see that this hasn't broken any of that.
- Run the tests locally with coverage and see that you can get a report. It should work with just `pnpm vitest run --coverage`.
- Look at the coverage report locally and maybe suggest if there are certain packages/paths which should be excluded.
- Look at the GitHub Action and see if it makes sense. It's a bit complicated and I would appreciate suggestions for simplifications.

### Testing

N/A. There's no changes in production code.

### Notes for release

N/A. This is for internal development only.


bjoerge:
This is great!

I've rebased against the vitest-v3 branch, upgraded vitest-coverage-v8 to latest and added some exclude patterns (d35ca76d70acf4127fa009a5ff22fccd690aff85):
- CLI tests (because coverage don't seem to be collected from them anyway, likely because the CLI are tested by invoking it via child process)
- workshop files
- telemetry definitions

I've done a bit of manual testing, and everything seems to work fine!

bjoerge (approved):

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseDetailsEditor.test.tsx
```diff
@@ -67,7 +67,7 @@ describe('ReleaseDetailsEditor', () => {
       () => {
         expect(useReleaseOperations().updateRelease).toHaveBeenCalledWith(release)
       },
-      {timeout: 250},
+      {timeout: 10_000},
```
bjoerge, line 70: had to increase timeout here to get the tests to pass again, likely due to coverage instrumentation making things run slower.

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseDetailsEditor.test.tsx
```diff
@@ -45,7 +45,7 @@ describe('ReleaseDetailsEditor', () => {
...
```
bjoerge, line 48: had to increase timeout here, likely due to coverage instrumentation making things run slower.

PR #7798, opened 2024-11-12
fix: issue with the US formatting used by formatRelative when no locale defined

jordanl17:
### Description

Currently:
For dates more than a week in the past/future, these would be formatted as 'MM/dd/yyyy':
Screenshot 2024-11-12 at 13 19 53

Now:
These dates are formatted correctly according to the locale formatting of the client, eg in my case 'dd/MM/yyyy':
Screenshot 2024-11-12 at 13 26 06

Ideally we would pass through the client locale to `date-fns` however given the way that `date-fns` requires the locale to be imported, and that the mappings of `navigator.language` to the `date-fns` are rather specific, taking the approach in this PR felt most extensible to all locales.

### Testing
Added tests for the wrapper util `formatRelativeLocale`

### Notes for release
N/A

N/A

### DIFF
--- a/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
+++ b/packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
@@ -1,18 +1,22 @@
-import {DotIcon, EyeClosedIcon, EyeOpenIcon} from '@sanity/icons'
+import {DotIcon, EyeClosedIcon, EyeOpenIcon, LockIcon} from '@sanity/icons'
 // eslint-disable-next-line no-restricted-imports -- custom use for MenuItem & Button not supported by ui-components
 import {Box, Button, Flex, MenuItem, Stack, Text} from '@sanity/ui'
-import {formatRelative} from 'date-fns'
 import {type CSSProperties, forwardRef, type MouseEvent, useCallback, useMemo} from 'react'
 import {css, styled} from 'styled-components'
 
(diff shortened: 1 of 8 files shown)


packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
```diff
@@ -189,6 +194,13 @@ export const GlobalPerspectiveMenuItem = forwardRef<
                 />
               </Tooltip>
             )}
+            {!isPublishedPerspective(release) && isReleaseScheduledOrScheduling(release) && (
+              <Box padding={2}>
+                <Text size={1}>
+                  <LockIcon />
```
jordanl17, line 200: Lock now shows for scheduled releases

packages/sanity/src/core/releases/navbar/GlobalPerspectiveMenuItem.tsx
```diff
@@ -169,10 +174,10 @@ export const GlobalPerspectiveMenuItem = forwardRef<
               {displayTitle}
             </Text>
             {!isPublishedPerspective(release) &&
-              release.metadata.releaseType !== 'undecided' &&
+              release.metadata.releaseType === 'scheduled' &&
```
jordanl17, line 177: Improvement to handle case that `intendedPublishAt` has not been updated when release type is (which shouldn't happen in studio, but could if CLI/API is used directly)

RitaDias (approved):
Thanks for this Jordan, this is great :)

PR #7806, opened 2024-11-13
fix: add --legacy-peer-deps to next-sanity pacakge install

SimeonGriggs:
### Description

When running `npx sanity@latest init` inside a Next.js 15 project the install errors because of React 19 incompatability.

### What to review

That the `yarn` and `pnpm` install commands don't also need editing.

### Notes for release

* Added `--legacy-peer-deps` flag to the automatic install of `next-sanity` inside Next.js projects


stipsan (approved):
It's not ideal, but I don't see a better option available to us :/

PR #7815, opened 2024-11-14
fix(sanity): add ability to search version documents from AddDocument in release tool

RitaDias:
### Description
- Add ability to search within specific perspectives to the SearchProvider
- Add ability of adding versions outside of the current pinned release / no perspective 
- Don't allow for adding the same "document" multiple times. Meaning, if you have a version of the document in a release and you try to search for another one / same, you won't be able to add it again
- Global search retains the same behavior as before:
   - if no version is pinned then it returns published / draft as expected
   - if version is pinned then it returns the currently existing version of the document
- If mulitple versions of a document exist, when searching in the tool (Add Document) it shows the right preview and adds the right version

https://www.loom.com/share/61bcd61b40b842bbbd2cf2614ee79560


packages/sanity/src/core/search/common/types.ts
```diff
@@ -70,6 +70,7 @@ export interface SearchFactoryOptions {
   /* only return unique documents (e.g. not both draft and published) */
   unique?: boolean
   strategy?: SearchStrategy
+  perspective?: string
```
RitaDias, line 73: we might want to use this for something in the future, hence why I didn't want to tie it specifically to the "raw" aspect.

packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
```diff
@@ -94,13 +109,25 @@ export function getPreviewStateObservable(
...
+       */
+      if (perspective.isRaw && versions && isVersionId(documentId)) {
+        v = {title, ...versions[getVersionFromId(documentId) ?? '']?.snapshot}
+      }
+
+      return {
+        draft: draft.snapshot ? {title, ...(draft.snapshot || {})} : null,
+        isLoading: false,
+        published: published.snapshot ? {title, ...(published.snapshot || {})} : null,
+        version: v,
```
bjoerge, line 127: I've removed `version` in #7828 because I feel it's very context dependent and I'd prefer to keep it a separate concern from loading what to fetch for previewing. No action required from you, just an FYI.

packages/sanity/src/core/search/common/types.ts
bjoerge: Good call!

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
```diff
@@ -24,15 +25,29 @@ import {initialSearchState, searchReducer} from './reducer'
...
   children?: ReactNode
   fullscreen?: boolean
+  /**
+   * if provided, then it means that the search is being done from within a specific perspective
+   */
+  perspective?: string
+  /**
+   * if provided, then ids should be checked against this list
+   */
+  listOfDocumentIds?: string[]
```
bjoerge, line 35: What does "checked against this list" mean? Also could it be named `documentIds`? (List/array is clearly communicated via the type already)

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
```diff
@@ -24,15 +25,29 @@ import {initialSearchState, searchReducer} from './reducer'
 interface SearchProviderProps {
   children?: ReactNode
   fullscreen?: boolean
+  /**
+   * if provided, then it means that the search is being done from within a specific perspective
+   */
+  perspective?: string
```
bjoerge, line 31: Did you consider an array here? Since we use the bundlePerspective for document lists I'm wondering if it would be expected that search results matches the current bundle perspectives.

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
```diff
@@ -41,6 +41,8 @@ export type SearchReducerState = PaginationState & {
   result: SearchResult
   terms: RecentSearch | SearchTerms
   strategy?: SearchStrategy
+  listOfDocumentIds?: string[]
```
bjoerge, line 44: rename to `documentIds` re my previous comment?

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
RitaDias: Yep, good point! 
I'll rename!

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
RitaDias: Hmm this is actually where right now I'm indicating that we /shouldn't/ use the bundlePerspectives. If we put "raw" here, then it means that the search will ignore the bundlePerspectives and simply look through all existing releases.

Re-reading my description I feel it's wrong. All that I do right now in the search aspect is check if it's raw, if it's raw then it does what I mentioned, but if it has nothing, then it will always search through the existingReleases (vs actually searching focused in a particular perspective).

I'll have a look, I think it might be simple to not make this comment be a lie (as in, if the perspective is actually sent in as something else, then search only within that perspective). I'll come back to you on this

RitaDias: This has quite a few large impacting changes on the code 🤔 
It'll take a little bit longer to get this to work as intended I think, I'll timebox this until tomorrow before the sync

packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
bjoerge: I'll integrate this work into my branch when this lands

packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
```diff
@@ -94,13 +109,25 @@ export function getPreviewStateObservable(
...
+    map(([draft, published, version, versions]) => {
+      let v = version.snapshot ? {title, ...(version.snapshot || {})} : null
+
+      /**
+       * when the search type should be raw, there are versions and the document provided is a version
+       * then we need to find the exact version to display
+       */
+      if (perspective.isRaw && versions && isVersionId(documentId)) {
+        v = {title, ...versions[getVersionFromId(documentId) ?? '']?.snapshot}
+      }
```
pedrobonamin, line 121: Could this be part of the `version$` observable?

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
```diff
@@ -24,15 +25,30 @@ import {initialSearchState, searchReducer} from './reducer'
...
+  /**
+   * list of document ids
+   * if provided, then ids should be checked against this list
+   */
+  documentIds?: string[]
```
pedrobonamin, line 36: Wdyt about renaming this to `excludedDocumentIds` or `disabledDocumentIds`.

pedrobonamin (commented):
Thanks for this Rita, I think it makes sense and works great!
Have you considered adding an explanation on why a document is disabled in search? Maybe having a tooltip to provide the info, doesn't need to be part of this PR

packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
RitaDias: Maybe! I'll try to have a look and see if I can put it in there :)

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
RitaDias: Ah this actually makes a lot of sense, I'll rename it!

RitaDias: Done on this https://github.com/sanity-io/sanity/pull/7815/commits/b8567687a7f20148fc2b3ad99f3f71543ae540d3 :)

packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
RitaDias: Here we go, done here [refactor(sanity): update to array prop vs string](https://github.com/sanity-io/sanity/pull/7815/commits/fe34b34a6299959a7bd0fd6e9f9c7f9ea76e361d) :) 

Thanks for the suggestion!

packages/sanity/src/core/studio/components/navbar/search/contexts/search/SearchProvider.tsx
RitaDias: Here we go, done here [refactor(sanity): update to array prop vs string](https://github.com/sanity-io/sanity/pull/7815/commits/fe34b34a6299959a7bd0fd6e9f9c7f9ea76e361d) :) 

Thanks for the suggestion!

RitaDias:
> Thanks for this Rita, I think it makes sense and works great!
> Have you considered adding an explanation on why a document is disabled in search? Maybe having a tooltip to provide the info, doesn't need to be part of this PR

Forgot to add this yesterday but I've added it today :)
Great suggestion, thank you!

RitaDias:
@pedrobonamin @bjoerge can I get a final pass of eyes and a ✅ on this? 🙏

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1206,6 +1206,8 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'release.navbar.tooltip': 'Releases',
   /** The placeholder text when the release doesn't have a title */
   'release.placeholder-untitled-release': 'Untitled release',
+  /** Label for when a version of a document has already been added to the release */
+  'release.tooltip.already-added': 'A version of this document has already been added',
```
pedrobonamin, line 1210: Thanks for adding this!

pedrobonamin (approved):
Looks good to me, thanks for doing this @RitaDias !

PR #7819, opened 2024-11-14
fix: explicitly list allowed image types for upload

rexxars:
### Description

We're currently allowing `image/*` in the image input, but only a certain types are allowed by our backend. In particular, we recently added support for AVIF images to be _output_, but we don't yet support them to be uploaded. Instead of letting users upload the entire image  before they get a message saying they are not supported, we should flag our support in the image selection dialog.

Also fixed a race condition issue with the `clearUploadStatus` callback - when the image fails to be uploaded, eg rejected because of an unsupported format, the value doesn't seem to have a value - and thus does not get unset. Instead it is stuck in the pending state (100% done) until someone resets it. I don't quite see the need to check before unsetting - but let me know if I am overlooking something.

Did some minor typing/lint cleanups along the way.

### What to review

- Do image uploads still work as expected?
- Are there dragons to be aware of?
- There are so many layers I had to add these types to - are some of these meant to be "backend agnostic" and thus I should only apply it to the ones that deals with the Sanity CDN specifically?

### Testing

Ideally there should be a test to ensure that you can't select eg an AVIF, but I couldn't find a good way of doing so at the moment :/

### Notes for release

- Fixed an issue where unsupported image types would be uploaded but then rejected by the server. Only accepted image file types are now

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/Asset.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/Asset.tsx
@@ -7,6 +7,7 @@ import {styled} from 'styled-components'
 
 import {Button, MenuButton, MenuItem} from '../../../../../../../../../../ui-components'
 import {type Source} from '../../../../../../../../../config'
+import {SUPPORTED_IMAGE_UPLOAD_TYPES} from '../../../../../../../../../form/inputs/files/constants'
 import {FileSource, ImageSource} from '../../../../../../../../../form/studio/assetSource'
 import {useClient} from '../../../../../../../../../hooks'
 import {useTranslation} from '../../../../../../../../../i18n'
@@ -102,13 +103,15 @@ export function SearchFilterAssetInput(type?: AssetType) {
 
     const AssetSourceComponent = selectedAssetSource?.component
 
-    const fontSize = fullscreen ? 2 : 1
-
     const buttonText = t(value ? 'search.filter-asset-change' : 'search.filter-asset-select', {
       context: type,
     })
 
-    const accept = get(type, 'options.accept', type === 'image' ? 'image/*' : '')
+    const accept = get(
+      type,
+      'options.accept',
+      type === 'image' ? SUPPORTED_IMAGE_UPLOAD_TYPES.join(',') : '',
+    )
 
     return (
       <ContainerBox>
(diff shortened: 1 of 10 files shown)


packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/Asset.tsx
```diff
@@ -102,13 +103,15 @@ export function SearchFilterAssetInput(type?: AssetType) {
...
     const AssetSourceComponent = selectedAssetSource?.component
 
-    const fontSize = fullscreen ? 2 : 1
-
     const buttonText = t(value ? 'search.filter-asset-change' : 'search.filter-asset-select', {
       context: type,
     })
 
-    const accept = get(type, 'options.accept', type === 'image' ? 'image/*' : '')
+    const accept = get(
```
bjoerge, line 110: oh wow, TIL you can search by uploading an image

bjoerge (approved):
Would devs still be able to add an unsupported type to `options.accept` to their schema? It only makes sense to customize `options.accept` with a subset of what the backend support, so would it be sensible for us to filter out unsupported types, and/or give a warning/error if you do this? (probably an edge case, and not a big deal, this PR is still a nice improvement, I'm just curious what you think).

Also noticed that we don't surface _why_ the upload fails if you try to upload an unsupported image (I tried avif) – it just says "The upload could not be completed at this time". We're missing an small education opportunity here as it could instead say "currently only x,y and z are supported"). We should look into fixing this separately.

TL;DR: LGTM – thank you for the additional cleanups too <3

packages/sanity/src/core/studio/components/navbar/search/components/filters/filter/inputs/asset/Asset.tsx
rexxars: Makes two of us!

rexxars:
> Also noticed that we don't surface _why_ the upload fails if you try to upload an unsupported image (I tried avif) – it just says "The upload could not be completed at this time". We're missing an small education opportunity here as it could instead say "currently only x,y and z are supported"). We should look into fixing this separately.

Yes, I did spot that - got derailed because the backend didn't respond with a structured error code and started working on some improvements on that front. Because of i18n, we can't/shouldn't just throw the server side error message, so will need to check an error code and give localized messages. Will continue improving on this front 👍

PR #7827, opened 2024-11-18
feat(releases): adding correct confirmation dialog to archive release flow

jordanl17:

### DIFF
--- a/packages/sanity/src/core/releases/store/__tests__/__mocks/useReleaseOperations.mock.ts
+++ b/packages/sanity/src/core/releases/store/__tests__/__mocks/useReleaseOperations.mock.ts
@@ -1,0 +1,15 @@
+import {type Mocked, vi} from 'vitest'
+
+import {type ReleaseOperationsStore} from '../../createReleaseOperationStore'
+
+export const useReleaseOperationsMock: Mocked<ReleaseOperationsStore> = {
+  archive: vi.fn(),
+  unarchive: vi.fn(),
+  createRelease: vi.fn(),
+  createVersion: vi.fn(),
+  discardVersion: vi.fn(),
+  publishRelease: vi.fn(),
+  schedule: vi.fn(),
+  unschedule: vi.fn(),
+  updateRelease: vi.fn(),
+}
--- a/packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
+++ b/packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
@@ -104,41 +104,38 @@ export function createReleaseOperationsStore(options: {
       },
     ])
 
-  const handleScheduleRelease = async (releaseId: string, publishAt: Date) => {
-    await requestAction(client, [
+  const handleScheduleRelease = async (releaseId: string, publishAt: Date) =>
+    requestAction(client, [
       {
         actionType: 'sanity.action.release.schedule',
         releaseId: getBundleIdFromReleaseDocumentId(releaseId),
         publishAt: publishAt.toISOString(),
       },
     ])
-  }
-  const handleUnscheduleRelease = async (releaseId: string) => {
-    await requestAction(client, [
+
+  const handleUnscheduleRelease = async (releaseId: string) =>
+    requestAction(client, [
       {
         actionType: 'sanity.action.release.unschedule',
         releaseId: getBundleIdFromReleaseDocumentId(releaseId),
       },
     ])
-  }
 
-  const handleArchiveRelease = async (releaseId: string) => {
-    await requestAction(client, [
+  const handleArchiveRelease = async (releaseId: string) =>
+    requestAction(client, [
       {
         actionType: 'sanity.action.release.archive',
         releaseId: getBundleIdFromReleaseDocumentId(releaseId),
       },
     ])
-  }
 
-  const handleUnarchiveRelease = async (releaseId: string) => {
-    await requestAction(client, [
+  const handleUnarchiveRelease = async (releaseId: string) =>
+    requestAction(client, [
       {
         actionType: 'sanity.action.release.unarchive',
         releaseId: getBundleIdFromReleaseDocumentId(releaseId),
       },
     ])
-  }
 
   const handleCreateVersion = async (releaseId: string, documentId: string) => {
     // the documentId will show you where the document is coming from and which
@@ -167,18 +164,13 @@ export function createReleaseOperationsStore(options: {
       : client.create(versionDocument))
   }
 
-  const handleDiscardVersion = async (releaseId: string, documentId: string) => {
-    if (!document) {
-      throw new Error(`Document with id ${documentId} not found`)
-    }
-
-    await requestAction(client, [
+  const handleDiscardVersion = async (releaseId: string, documentId: string) =>
+    requestAction(client, [
       {
         actionType: 'sanity.action.document.discard',
         draftId: getVersionId(documentId, releaseId),
       },
     ])
-  }
 
   return {
     archive: handleArchiveRelease,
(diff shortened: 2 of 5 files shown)


packages/sanity/src/core/releases/__fixtures__/release.fixture.ts
```diff
@@ -0,0 +1,17 @@
+import {type ReleaseDocument} from '../store/types'
+
+export const activeScheduledRelease: ReleaseDocument = {
```
jordanl17, line 3: Thinking that we'd continue to create others here eg `activeASAPRelease` as are necessary for test cases across our suite. All fixtures would be correct and valid states, eg `activeASAPRelease` would not have an `intendedPublishAt` since that is not a valid state machine value

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/__tests__/ReleaseMenuButton.test.tsx
```diff
@@ -1,21 +1,23 @@
...
 import {releasesUsEnglishLocaleBundle} from '../../../../i18n'
 import {type ReleaseDocument} from '../../../../index'
+import {useReleaseOperationsMock} from '../../../../store/__tests__/__mocks/useReleaseOperations.mock'
 import {useReleaseOperations} from '../../../../store/useReleaseOperations'
 import {ReleaseMenuButton, type ReleaseMenuButtonProps} from '../ReleaseMenuButton'
 
-vi.mock('sanity', () => ({
-  SANITY_VERSION: '0.0.0',
-  useTranslation: vi.fn().mockReturnValue({t: vi.fn()}),
+vi.mock('../../../../store/useReleaseOperations', () => ({
```
jordanl17, line 13: Tried in vain to simplify the way this could be mocked, but a mixture of issues an nuance with how `vi.mock` is hoisted and executed mean we still need these mock statements explicitly in our tests rather than in generalised mock files

RitaDias (commented):
I rogue thought I had when I was reviewing this related to where wea re placing the fixtures.
I think it might make better sense to keep them somewhere along like we do with the TestProvider in the `test` directory?

Right now we are keeping in the core/releases but I feel it will come a point where we might want to use releases in other places for one reason or another. Especially since this does kinda touch more parts than this self contained part (the first thing that comes to mind right now are some components in `structure`).

I'm not married to the idea but I think findability might be better :)

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
```diff
@@ -19,35 +20,117 @@ export type ReleaseMenuButtonProps = {
...
+          <Text muted size={1}>
+            <Translate
+              t={t}
+              i18nKey="toast.archive.error"
+              values={{title: release.metadata.title, error: archivingError.toString()}}
+            />
+          </Text>
+        ),
+      })
+      console.error(archivingError)
```
RitaDias, line 70: Rogue console log or meant to be here? :)

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
```diff
@@ -67,6 +150,8 @@ export const ReleaseMenuButton = ({disabled, release}: ReleaseMenuButtonProps) =
             {!release?.state || release.state === 'archived' ? (
               <MenuItem
                 onClick={handleUnarchive}
+                // TODO: disabled as CL action not yet impl
```
RitaDias, line 153: Do we have a linear story for this? 👀

RitaDias (approved):
Just a few comments and a suggestion, great job :)

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -64,41 +64,38 @@ export function createReleaseOperationsStore(options: {
       },
     ])
 
-  const handleScheduleRelease = async (releaseId: string, publishAt: Date) => {
-    await requestAction(client, [
+  const handleScheduleRelease = async (releaseId: string, publishAt: Date) =>
```
bjoerge, line 67: `async` can be removed from here and the other functions now that there's no await (saves us a microtask), but would be good if we still can make it return a void promise

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
```diff
@@ -19,35 +20,117 @@ export type ReleaseMenuButtonProps = {
...
 
   const handleUnarchive = async () => {
-    setIsPerformingOperation(true)
-    await unarchive(release._id)
-
-    // it's in the process of becoming false, so the event we want to track is unarchive
-    telemetry.log(UnarchivedRelease)
-    setIsPerformingOperation(false)
+    // noop
+    // TODO: similar to handleArchive - complete once server action exists
```
bjoerge, line 79: what's the reason for deleting this? Do we expect unarchive to change significantly?

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
```diff
@@ -19,35 +20,117 @@ export type ReleaseMenuButtonProps = {
...
   }
 
+  const confirmArchiveDialog = useMemo(() => {
+    if (selectedAction !== 'confirm-archive') return null
+
+    const dialogDescription =
+      releaseDocuments.length === 1
```
bjoerge, line 86: Could we use the _one, _other syntax for these? https://www.i18next.com/translation-function/plurals#languages-with-multiple-plurals

packages/sanity/src/core/releases/store/__tests__/__mocks/useReleaseOperations.mock.ts
```diff
@@ -0,0 +1,15 @@
+import {type Mocked, vi} from 'vitest'
+
+import {type ReleaseOperationsStore} from '../../createReleaseOperationStore'
+
+export const useReleaseOperationsMock: Mocked<ReleaseOperationsStore> = {
```
bjoerge, line 5: This looks like a great pattern to me!

bjoerge (changes requested):

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
jordanl17: Yes we do corel-256

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
jordanl17: It will change from what it currently is in `corel` (will be updated to to be similar to archive). I considered implementing it, but feel it's better to complete the full task once the action can be test alongside. There's a separate task I'm monitoring and will complete once ready

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
jordanl17: Ah, I wasn't a massive fan of `other` since it isn't really clear it means 'not one', but I see it's used in many placed, and also proposed by i18next, so yeh I'll capitulate and change to it

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
jordanl17: I followed similar approach to what we do in Publish and schedule, so this is intended

PR #7828, opened 2024-11-18
refactor: use branded ids for releases

bjoerge:


jordanl17:
There are lots of conflicts here now, although we have since started using branded types in a few more places, but extensive use for all releases might need more discussion

PR #7831, opened 2024-11-18
refactor(core): remove unneeded PTE event handling

christianhg:
PTE does not emit the `undo` and `redo` events, and I don't think it ever has, at least not for a long time. Whenever an undo or redo happens, the end result is still patches and the `mutation` event is enough to listen for that.

---

You can test this by verifying that undo/redo still works.


RitaDias (approved):
Looks good to me :)

PR #7838, opened 2024-11-18
feat: add `groq2024` search strategy

juice49:
### Description

This branch introduces a new search strategy: `groq2024`. This strategy adopts GROQ's new `text::query` function and ability to match all text content inside an object using `@ match text::query($query)`. This enables dataset search without necessitating every searchable path in the schema is serialised and included in the search request (as is the case with `groqLegacy` search).

This experimental strategy is being added to Studio behind a flag so that interested users can help us test it, before it hopefully graduates to being the default search strategy.

To enable this strategy, set `search.strategy` to `"groq2024"` inside the Studio config.

```diff
import {defineConfig} from 'sanity`

export default defineConfig({
+ search: {
+   strategy: 'groq2024',
+ },
})
```

### What to review

- Test Studio has been updated to use `groq2024`. You can try it out using any search surface (global search, document lists, and reference search).
- Are there any obvious issues or design flaws in the implementation of the new strategy?
- Are there any big issues that should be addressed before the experimental release of `groq2024`?

### Testing

- Added unit tests for new functionality.

### DIFF
--- a/packages/sanity/src/core/search/text-search/createTextSearch.test.ts
+++ b/packages/sanity/src/core/search/text-search/createTextSearch.test.ts
@@ -2,15 +2,7 @@ import {Schema} from '@sanity/schema'
 import {defineField, defineType} from '@sanity/types'
 import {describe, expect, it} from 'vitest'
 
-import {
-  getDocumentTypeConfiguration,
-  getOrder,
-  getQueryString,
-  isExactMatchToken,
-  isNegationToken,
-  isPrefixToken,
-  prefixLast,
-} from './createTextSearch'
+import {getDocumentTypeConfiguration, getOrder, getQueryString} from './createTextSearch'
 
 const testType = Schema.compile({
   types: [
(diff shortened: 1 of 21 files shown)


packages/sanity/src/core/search/text-search/createTextSearch.test.ts
```diff
@@ -243,65 +235,3 @@ describe('getQueryString', () => {
...
-describe('isExactMatchToken', () => {
-  it('recognises that a token is encased in quote marks', () => {
-    expect(isExactMatchToken(undefined)).toBe(false)
-    expect(isExactMatchToken('"a"')).toBe(true)
-    expect(isExactMatchToken('"a b"')).toBe(true)
-    expect(isExactMatchToken('"a')).toBe(false)
-    expect(isExactMatchToken('a"')).toBe(false)
-    expect(isExactMatchToken('"a b')).toBe(false)
-    expect(isExactMatchToken('a b"')).toBe(false)
-  })
```
pedrobonamin, line 306: I think we lost all of this tests when they were moved, could we add them into a `token.tests` file?

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
```diff
@@ -597,7 +597,11 @@ function stripRecent(terms: RecentSearch | SearchTerms) {
  *
  * Note that should any result appear again in subsequent pages, its first instance will be removed.
  */
-function deduplicate(hits: SearchHit[]): SearchHit[] {
+function deduplicate(hits: SearchHit[], {strategy}: {strategy?: SearchStrategy}): SearchHit[] {
+  if (strategy !== 'textSearch') {
```
pedrobonamin, line 601: This seems to be changing the previous `textSearch` behavior right? 
Could we add some details on why we don't want to filter duplicates when using this strategy ?

packages/sanity/src/core/search/text-search/createTextSearch.test.ts
juice49: Great catch! I forgot to commit the new test file 🤦‍♀️.
Screenshot 2024-12-10 at 09 37 31

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
juice49: No, it shouldn't be. The intention is to skip deduplication for any strategy **that isn't** `textSearch`, so it returns the provided `hits` without processing them in that case.

The deduplication is only necessary when `textSearch` is enabled, because we found sometimes results are duplicated at the beginning and end of pages.

packages/sanity/src/core/search/text-search/createTextSearch.test.ts
juice49: Added them in [this commit](https://github.com/sanity-io/sanity/pull/7838/commits/c4bb19d76e3770115a243aabb4e10bc17814ea03). Thanks for catching 👍.

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
pedrobonamin: You are right, sorry, miss read it as `===` 
I think adding this for our future selfs will be great 
>The deduplication is only necessary when `textSearch` is enabled, because we found sometimes results are duplicated at the beginning and end of pages.

pedrobonamin (approved):
Thanks for this @juice49 !
Tests are really helpful to understand the changes.

It looks good to me!

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
juice49: It could be more explicit, but it is already mentioned 1 line up in the comment:

 > At page boundaries, **the Text Search API may sometimes produce duplicate results**.

pedrobonamin: Right, github was not showing me that piece. No need to add anything else

juice49:
Thanks, @pedrobonamin 🙏.

PR #7845, opened 2024-11-19
chore(cli): temporarily disable telemetry tests

rexxars:
### Description

The CLI telemetry tests are very flakey and doesn't hold up well when there are multiple runs going on at the same time, since they reuse the same user for setting/enabling status and then checking the result.

This temporarily disables them, until we can find a better way to mock the network calls or similar.

### What to review

- Tests are disabled

### Testing

We're dropping some tests here - we could consider refactoring them to at least check that we get _some_ output from the commands, but seems less than useful right now.

### Notes for release

None


juice49 (approved):
I see the tests are correctly being skipped in CI. Thanks for doing this chore 🙏.

PR #7852, opened 2024-11-20
feat: support search weight configuration for slug fields

juice49:
### Description

- Search weight can now be assigned to slug fields using the `options.search.weight` option. It was previously not permitted by the types, nor had any impact if set.
- These weights will now be reflected in global search results when "Best match" sorting is selected and in reference search results.
- As a consequence of this change, search queries no longer attempt to match on the deprecated slug `source` field, resulting in slightly smaller search queries. Matching only occurs on the slug object's `current` field.

#### Global Search Example

Imagine two different document types:

- `book` contains a `title` field that is implicitly given a weight of `10` because it's selected as the list preview title.
- `author` contains a `slug` field that is explicitly given a weight of `1_000` via the `options.search.weight` option.

| Global Search - `next` | Global Search - `fix/sapp-34/slug-field-weights` | 
| --- | --- |
| global-search-next |  global-search-sapp-34 |

| Reference Search - `next` | Reference Search - `fix/sapp-34/slug-field-weights` | 
| --- | --- |
| reference-search-next | reference-search-sapp-34 |

Both `book.title` and `author.slug` contain an exact match for the term `"xoxoxo"`. With the changes introduced in this branch, the weighting set for `author.slug` is respected, and the author document is therefore recognised as more relevant than the book document.

### What to review

- The order of global Studio search results when sorting by "Best match" and in reference search.
- Setting `options.search.weight` on a slug (or slug-based/aliased) field.

### Testing

- Updated unit tests in `packages/sanity/src/core/search/common/__tests__/deriveSearchWeightsFromType.test.ts`.

### Notes for release

It's now possible to configure search weights for slug fields using the `options.search.weight` field option:

```diff
defineField({
  name: 'slug',
  type: 'slug',
+ options: {
+   search: {
+     weight: 50,
+   },
+ },
}),
```

The slug field's weight will now be respected when using global search with "Best match" sorting selected, and when using reference search.


packages/sanity/src/core/search/common/deriveSearchWeightsFromType.ts
```diff
@@ -32,13 +37,27 @@ const isPtField = (type: SchemaType | undefined) =>
...
+}
+
 const isSearchConfiguration = (options: unknown): options is SearchConfiguration =>
   isRecord(options) && 'search' in options && isRecord(options.search)
 
 function isSchemaType(input: SchemaType | CrossDatasetType | undefined): input is SchemaType {
   return typeof input !== 'undefined' && 'name' in input
 }
 
+function getFullyQualifiedPath(schemaType: SchemaType, path: string): string {
```
RitaDias, line 52: Maybe we need to add a comment or maybe change the name? I'm just not sure what "fully qualified" would mean in this context 🥲 

Could it be simplified to `getFullPath` or something similar?

RitaDias: Or, since it seems to just be used when the field is a `Slug` as per line 83 maybe make it related to that? 🤔

RitaDias (commented):
Just some comments to get some clarification before ✅ 🙏

packages/sanity/src/core/search/common/deriveSearchWeightsFromType.ts
juice49: That's a good point. ["Fully qualified name"](https://en.wikipedia.org/wiki/Fully_qualified_name) lives in my brain for some reason, so it's intuitive for me. But I understand it may not be for everybody.

In this case, I _think_ it makes sense. A selected path may point to a `slug` field, but the _full qualified path_ for searching and weighting is the field located at `slug.current`.

In a different branch, this function also automatically converts references the user has defined using dot notation e.g. `some.reference.field` becomes `some.reference->field`.

I think there is a some nuance between "full path" and "fully qualified path" but I am open to changing it. What do you think? 🙂

RitaDias (approved):
✅

PR #7859, opened 2024-11-21
fix(telemetry): log (well known) workspace and dataset names

bjoerge:
### Description
This adds telemetry logging for well known workspace and dataset names like `production`, `staging`, `development` etc. Anything not [listed here](https://github.com/sanity-io/sanity/compare/sapp-1932?expand=1#diff-75c322fa92c0f73a757c75525e828216ed1af09d00a2a512eab2000b101ba63bR13) be logged as `<custom>`. In the event of a missing workspace name or dataset name (not sure if it can really happen), we send `<missing>`

### What to review
- Is the naming sensible?
- Do we want only unique values to be sent? Advantage of including all is that we can tell how many workspaces and datasets there are in total. If we want to only send unique names, we might want to include a count for both, but we'll miss out on knowing the number of occurrences for each (i.e. we won't be able to know how often the `production` dataset repeats across workspaces)

### Testing
I've manually tested and verified that we send the indended payload

### Notes for release
n/a


svirs (approved):

joneidejohnsen (dismissed):

packages/sanity/src/core/studio/StudioTelemetryProvider.tsx
```diff
@@ -97,5 +97,5 @@ export function StudioTelemetryProvider(props: {children: ReactNode; config: Con
 }
 
 function getWellKnownName(name: string) {
-  return WELL_KNOWN_NAMES.includes(name.toLowerCase()) ? name : undefined
+  return WELL_KNOWN_NAMES.find((wellKnownName) => name.toLowerCase().includes(wellKnownName))
```
bjoerge, line 100: note that this will return the well known name, and not the actual workspace name/dataset name, so e.g. `foo-Staging` will be logged as `staging`

PR #7862, opened 2024-11-22
fix(core): when publishing a release unpin it

pedrobonamin:
### Description

Once a released is published (archived), it should be unpinned if it was pinned, to avoid bugs in the rest of the system.


RitaDias (approved):
🚀

PR #7866, opened 2024-11-22
fix(core): remove timeout and retry from releases store

pedrobonamin:
### Description
Removes the timeout and retry function from the observable, given they are not needed and was causing issues by creating new listeners every 10 seconds.

Retry is not needed here given is already included in the listener.


bjoerge (approved):
Great catch!

PR #7878, opened 2024-11-25
fix: support creating version with initial value

bjoerge:
### Description
Fixes an issue that made it impossible to create a new document in a release. Ideally you should be able to just start typing, but this fix at least makes it _possible_.

Also adds low-level API support for providing an initial/fallback value to use when adding a version to a document.


pedrobonamin (approved):
Thanks for doing this @bjoerge , it looks good to me.

PR #7879, opened 2024-11-25
fix(releases): blocking UI elements that allow for adding documents to inactive releases

jordanl17:
### Description
blockAddDocsToScheduled
Currently if you are in published perspective or a scheduled release perspective, trying to add a new release in studio will show a new draft has been created but will be read only. The global perspective will remain as published or scheduled release even though the document form is showing a draft.

This change disables the UI elements that allow for creating documents in these scenarios.

NOTE: currently the tooltip clarification is the same for published as it is for a scheduled release. This should probably be made clearer to avoid confusion by separating these 2 scenarios, but I'd rather get this initial change in, and can then follow up once the copy has been confirmed later.

### Testing
There aren't currently any tests for these components or for their parents, so adding testing coverage here is a larger task

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/hooks/useIsReleaseActive.ts
+++ b/packages/sanity/src/core/releases/hooks/useIsReleaseActive.ts
@@ -1,6 +1,7 @@
 import {isDraftPerspective, isPublishedPerspective} from '../util/util'
 import {usePerspective} from './usePerspective'
 
+/** @internal */
 export const useIsReleaseActive = () => {
   const {selectedPerspective} = usePerspective()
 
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/releases/hooks/useIsReleaseActive.ts
```diff
@@ -0,0 +1,12 @@
+import {isDraftPerspective, isPublishedPerspective} from '../util/util'
+import {usePerspective} from './usePerspective'
+
+/** @internal */
+export const useIsReleaseActive = () => {
```
bjoerge, line 5: very much a nit, but I think we tend to use/prefer named functions

bjoerge (approved):
Very nice, thank you for this @jordanl17 !

PR #7889, opened 2024-11-27
fix(sanity): reset search cursor state when any parameter changes

juice49:
### Description

This branch fixes global search pagination bugs caused by the cursor state not correctly being reset when some input parameters change. For example, when removing a filter, the cursor was not being reset. This could cause the next result set to be fetched using a cursor for the previously active set of filters.

This isn't currently a problem, because global search results aren't paginated. However, we will begin using pagination when we soon introduce the new GROQ search strategy.

### What to review

- Global search input parameters and predicated builder (e.g. filters).

### Testing

- Added unit tests to the reducer to ensure cursor state is reset when input parameters change.


ricokahler (approved):
very nice

PR #7890, opened 2024-11-27
fix: Start in Create type exclusion now recursively checks for sanityCreate.exclude options

snorrees:
### Description

Turns out that "Start in Create" button did not correctly check for `options.sanityCreate.exclude: true`, and it was essentially not possible to opt-out per document type.

This new implementation checks options recursively (and more importantly maybe; it checks the base type 🤦 ). Recursion stops at the first boolean `sanityCreate.exclude`. This allows alias types to override parent options.

### What to review

Read the unit-tests and 👍 👎 

### Testing

I deployed a new version of the test studio for this:

https://create-integration-test.sanity.studio/

Check that:
* New "Documentation Article" documents has a "Start in Sanity Create" button.
* New "Sanity Create Excluded" documents does does NOT have a "Start in Sanity Create" button.

### Notes for release

Fixes an issue where the `options.sanityCreate.exclude: true` was not respected when determining if the "Start in Sanity Create" button should be shown for studios with `beta.create.startInCreateEnabled: true` set in sanity.config.

Setting `options.sanityCreate.exclude: true` on a document type will now correctly hide "Start in Sanity Create" for new documents of that type.

### DIFF
--- a/dev/test-create-integration-studio/schema.ts
+++ b/dev/test-create-integration-studio/schema.ts
@@ -11,6 +11,20 @@ export const seoGroup: FieldGroupDefinition = {
 
 export const schemaTypes = [
   defineType({
+    type: 'document',
+    name: 'sanity-create-excluded',
+    fields: [
+      defineField({
+        name: 'title',
+        title: 'New documents of this type should not have a Start in Create button',
+        type: 'string',
+      }),
+    ],
+    options: {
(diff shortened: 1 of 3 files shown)


robinpyon (approved):

RitaDias (approved):
Thanks for this!

PR #7891, opened 2024-11-27
fix: opt-out `@tanstack/react-virtual` from React Compiler

stipsan:
### Description

Adding React Compiler caused a regression in paths that uses `import {useVirtualizer} from '@tanstack/react-virtual'`. [It's a known issue](https://github.com/TanStack/virtual/issues/736), and I've subscribed to the issue so I get notified when we can come back and remove the `'use no memo'` directives 👍 

### What to review

This should fix SAPP-1945, repro: https://test-studio.sanity.dev/test/structure/input-standard;arraysTest;18db9a51-b3fe-4843-94a6-c5a2ba6ac039

### Testing

Didn't have time to add an E2E test that covers this case, but it might be worth doing in a follow-up 🙌  Manually tested by checking http://localhost:3333/test/structure/input-standard;arraysTest;18db9a51-b3fe-4843-94a6-c5a2ba6ac039 and scrolling, verifying that array reference items render without focusing/clicking first.

### Notes for release

Fixes a regression introduced in `v3.65.0` where array list references rendered far too lazily, and required focus or clicking, when initially "below the fold" of the browser window. They now correctly render on scroll ✨

### DIFF
--- a/packages/sanity/src/core/components/commandList/CommandList.tsx
+++ b/packages/sanity/src/core/components/commandList/CommandList.tsx
@@ -1,3 +1,6 @@
+'use no memo'
+// The `use no memo` directive is due to a known issue with react-virtual and react compiler: https://github.com/TanStack/virtual/issues/736
+
 import {Box, rem, Stack} from '@sanity/ui'
 import {type ScrollToOptions, useVirtualizer, type Virtualizer} from '@tanstack/react-virtual'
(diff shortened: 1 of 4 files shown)


RitaDias (approved):
Did some testing myself, this seems to work as expected 🙏 
Thanks Cody

PR #7898, opened 2024-11-28
fix(core): only open one annotation object edit modal at a time

christianhg:
### Description

This PR addresses an issue with how the PTE annotation edit modal is rendered. If the newly added (or edited) annotation spanned multiple spans, then a modal would be rendered for each span.

It's likely been that way for a long time, but the reason we didn't catch it is that it's hard to see on most viewport sizes:
Screenshot 2024-11-28 at 17 06 30

But sometimes you can see it:
Screenshot 2024-11-28 at 17 06 16

And it's very clear in the DOM:
Screenshot 2024-11-28 at 17 06 45

### What to review

Select a piece of text broken up by decorators and add a link. Verify that this still works correctly and the new (single) modal renders correctly.

I had to do a preceeding `refactor` commit in order to obtain references to PTE annotation spans in a reliable way. Please check out that commit. Does the usage of `react-rx` look correct? I'm new to the library.

Note: The performance in this dialog is not great. There's noticeable lag when you type in it. But this seems like an existing issue and not something caused by this fix.

### Testing

I spent quite a few hours trying to learn the E2E testing setup and craft a Playwright Component Test for this specific issue, but ultimately had to give up. When I finally got a test running that I thought was reliable, it turned out to be a false positive since it was green on the `next` branch as well.

This is the test I ended up with and ultimately scrapped: 

```
    test('Only triggers one modal when added across spans', async ({mount, page}) => {
      const {getModifierKey, getFocusedPortableTextEditor, insertPortableText, toggleHotkey} =
        testHelpers({page})
      await mount(<AnnotationsStory />)

      const $pte = await getFocusedPortableTextEditor('field-body')
      const modifierKey = getModifierKey()

      // Given the text "abc" where "b" has marks "strong"
      await insertPortableText('a', $pte)
      await toggleHotkey('b', modifierKey)
      await insertPortableText('b', $pte)
      await toggleHotkey('b', modifierKey)
      await expect(
        $pte.locator(`[data-mark="strong"]`, {
          hasText: `b`,
        }),
      ).toBeVisible()
      await insertPortableText('c', $pte)

      // When "abc" is selected
      await page.keyboard.down('Shift')
      await page.keyboard.press('ArrowLeft')
      await page.keyboard.press('ArrowLeft')
      await page.keyboard.press('ArrowLeft')
      await page.keyboard.up('Shift')

      // And the editor has settled (I don't know of a better way to assert this right now)
      await new Promise((resolve) => setTimeout(resolve, 100))

      // And the link button is clicked
      await page.getByRole('button', {name: 'Link'}).click()

      // Then three spans are annotated as links
      await expect($pte.locator('span[data-link]')).toHaveCount(3)

      // But only one modal is shown
      await expect(page.locator('[data-testid="popover-edit-dialog"]')).toHaveCount(1)
    })
```

Therefore, there is no automated test added for this issue.

### Notes for release

In previous releases, editing an annotation in the Portable Text Editor could result in multiple editing dialogs being opened. This problem has now been fixed.


juice49 (approved):
Sorry about the pain around E2E tests. Otherwise, this looks good to me. Thanks 🙏.

PR #7907, opened 2024-11-28
feat(sanity): add document comparison tool

juice49:
### Description

This branch adds a diff view for comparing documents. Primarily, this is for use with versions, but it has been designed to support comparing any two documents.

Most of the commits are supporting work (this should probably be a stack). The main feature addition occurs in  4f030aa23991dabdda009283e9d648f18dd0baab.
#### Navigating to diff view

Users can navigate to the diff view by setting the following URL search parameters:

```
| Name               | Type                  | Required | Description                                                                                              |
|--------------------|-----------------------|----------|----------------------------------------------------------------------------------------------------------|
| `diffView`         | `'version'`           | Required | Enable *version comparison* mode UI.                                                                     |
| `previousDocument` | `${string},${string}` | Required | The document type and id identifying the *previous* document for comparison, delimited by `,` character. |
| `nextDocument`     | `${string},${string}` | Required | The document type and id identifying the *next* document for comparison, delimited by `,` character.     |
```

##### Example: specifying both `previousDocument` and `nextDocument`

```
/test/structure/author;8701af36-f4e2-4f32-b13e-fc522ddb381f?perspective=r64awNgUC&structure[diffView]=version&structure[nextDocument]=author%2Cversions.r64awNgUC.8701af36-f4e2-4f32-b13e-fc522ddb381f&structure[previousDocument]=author%2Cversions.rO2NP9B76.8701af36-f4e2-4f32-b13e-fc522ddb381f
```

### New dependency: [ScrollMirror](https://www.npmjs.com/package/scrollmirror)

I've added [ScrollMirror](https://www.npmjs.com/package/scrollmirror) to handle scroll position syncing between two panes that may have different heights. This is achieved by syncing the percentage scroll position of each pane. We could implement this directly in the Studio codebase if we wish.

### What to review

- General implementation.
- Introduction of `ScrollMirror` dependency.

### Testing

There are no tests just yet, as I've spent more time than I expected making the code render in a test environment. First, I [needed to move the diff view code](https://github.com/sanity-io/sanity/commit/2cc93c797f028cc172e11499de8fa9a0725495dc) that was originally organised as a plugin to be part of `sanity/structure`. Then I encountered many issues with existing imports in `sanity/structure` resolving to `undefined` when running in a test environment.

- Navigate to a diff view. Ensure the UI works as expected.
- Navigate to a diff view using the "Compare versions" document action. Ensure the diff view is opened as expected.

I will follow up with tests as soon as these problems are resolved.

### Notes for release

The new document comparison tool provides a side-by-side view of two different document by versions, highlighting the fields that have changed between the _previous_ and _next_ versions. Select "Compare versions" from the document actions menu when viewing any draft or version document in Studio to open the tool.

By default, the first pane displays the _previous_ version, and the second pane shows _next_ version (the version that introduces changes). Use the menu at the top of each pane to switch the versions displayed.

### DIFF
--- a/packages/sanity/src/core/perspective/PerspectiveProvider.tsx
+++ b/packages/sanity/src/core/perspective/PerspectiveProvider.tsx
@@ -3,6 +3,7 @@ import {PerspectiveContext} from 'sanity/_singletons'
 
 import {getReleasesPerspectiveStack} from '../releases/hooks/utils'
 import {useActiveReleases} from '../releases/store/useActiveReleases'
+import {isSystemBundleName} from '../util/draftUtils'
 import {EMPTY_ARRAY} from '../util/empty'
 import {getSelectedPerspective} from './getSelectedPerspective'
 import {type PerspectiveContextValue, type ReleaseId, type SelectedPerspective} from './types'
@@ -40,8 +41,9 @@ export function PerspectiveProvider({
     () => ({
       selectedPerspective,
       selectedPerspectiveName,
-      selectedReleaseId:
-        selectedPerspectiveName === 'published' ? undefined : selectedPerspectiveName,
+      selectedReleaseId: isSystemBundleName(selectedPerspectiveName)
+        ? undefined
+        : selectedPerspectiveName,
       perspectiveStack,
       excludedPerspectives,
     }),
--- a/packages/sanity/src/core/releases/plugin/documentActions/index.ts
+++ b/packages/sanity/src/core/releases/plugin/documentActions/index.ts
@@ -11,7 +11,9 @@ export default function resolveDocumentActions(
 ): Action[] {
   const duplicateAction = existingActions.filter(({name}) => name === 'DuplicateAction')
 
-  return context.versionType === 'version'
-    ? duplicateAction.concat(DiscardVersionAction).concat(UnpublishVersionAction)
-    : existingActions
+  if (context.versionType === 'version') {
+    return duplicateAction.concat(DiscardVersionAction, UnpublishVersionAction)
+  }
+
+  return existingActions
 }
--- a/packages/sanity/src/core/releases/tool/components/Chip.tsx
+++ b/packages/sanity/src/core/releases/tool/components/Chip.tsx
@@ -1,7 +1,13 @@
 import {Box, Card, Flex, Text} from '@sanity/ui'
-import {type ReactNode} from 'react'
+import {type ComponentType, type ReactNode} from 'react'
 
-export function Chip(props: {avatar?: ReactNode; text: ReactNode; icon?: ReactNode}) {
+interface Props {
+  avatar?: ReactNode
+  text: ReactNode
+  icon?: ReactNode
+}
+
+export const Chip: ComponentType<Props> = (props) => {
   const {avatar, text, icon} = props
 
   return (
--- a/packages/sanity/src/core/releases/tool/components/Chip.tsx
+++ b/packages/sanity/src/core/releases/tool/components/Chip.tsx
@@ -1,17 +1,17 @@
-import {Box, Card, Flex, Text} from '@sanity/ui'
+import {Box, Card, type CardProps, Flex, Text} from '@sanity/ui'
 import {type ComponentType, type ReactNode} from 'react'
 
-interface Props {
+type Props = {
   avatar?: ReactNode
   text: ReactNode
   icon?: ReactNode
-}
+} & Pick<CardProps, 'tone'>
 
-export const Chip: ComponentType<Props> = (props) => {
+export const Chip: ComponentType<Props> = ({tone, ...props}) => {
   const {avatar, text, icon} = props
 
   return (
-    <Card muted radius="full">
+    <Card muted radius="full" tone={tone}>
       <Flex align={'center'}>
         {icon && (
           <Box padding={1} marginLeft={1}>
--- a/packages/sanity/src/core/util/draftUtils.ts
+++ b/packages/sanity/src/core/util/draftUtils.ts
@@ -106,9 +106,38 @@ export function getDraftId(id: string): DraftId {
   return isDraftId(id) ? id : ((DRAFTS_PREFIX + id) as DraftId)
 }
 
+/** @internal */
+export const systemBundles = ['drafts', 'published'] as const
+
+/** @internal */
+export type SystemBundle = (typeof systemBundles)[number]
+
+/** @internal */
+export function isSystemBundle(maybeSystemBundle: unknown): maybeSystemBundle is SystemBundle {
+  return systemBundles.includes(maybeSystemBundle as SystemBundle)
+}
+
+/** @internal */
+const systemBundleNames = ['draft', 'published'] as const
+
+/** @internal */
+type SystemBundleName = (typeof systemBundleNames)[number]
+
+/**
+ * `isSystemBundle` should be preferred, but some parts of the codebase currently use the singular
+ * "draft" name instead of the plural "drafts".
+ *
+ * @internal
+ */
+export function isSystemBundleName(
+  maybeSystemBundleName: unknown,
+): maybeSystemBundleName is SystemBundleName {
+  return systemBundleNames.includes(maybeSystemBundleName as SystemBundleName)
+}
+
 /**  @internal */
 export function getVersionId(id: string, version: string): string {
-  if (version === 'drafts' || version === 'published') {
+  if (isSystemBundle(version)) {
     throw new Error('Version can not be "published" or "drafts"')
   }
 
--- a/packages/sanity/src/structure/diffView/components/DiffView.tsx
+++ b/packages/sanity/src/structure/diffView/components/DiffView.tsx
@@ -1,0 +1,72 @@
+import {type ComponentType, useContext, useMemo, useState} from 'react'
+import {type DocumentLayoutProps} from 'sanity'
+import {ReferenceInputOptionsContext} from 'sanity/_singletons'
+
+import {Dialog} from '../../../ui-components/dialog/Dialog'
+import {useCreatePathSyncChannel} from '../hooks/useCreatePathSyncChannel'
+import {useDiffViewRouter} from '../hooks/useDiffViewRouter'
+import {useDiffViewState} from '../hooks/useDiffViewState'
+import {useScrollMirror} from '../hooks/useScrollMirror'
+import {VersionModeHeader} from '../versionMode/components/VersionModeHeader'
+import {DialogLayout} from './DialogLayout'
+import {DiffViewPane} from './DiffViewPane'
+import {EditReferenceLinkComponent} from './EditReferenceLinkComponent'
+
+export const DiffView: ComponentType<Pick<DocumentLayoutProps, 'documentId'>> = ({documentId}) => {
+  const {documents, state, mode} = useDiffViewState()
+  const {exitDiffView} = useDiffViewRouter()
+  const syncChannel = useCreatePathSyncChannel()
+  const [previousPaneElement, setPreviousPaneElement] = useState<HTMLElement | null>(null)
+  const [nextPaneElement, setNextPaneElement] = useState<HTMLElement | null>(null)
+  const referenceInputOptionsContext = useContext(ReferenceInputOptionsContext)
+
+  const diffViewReferenceInputOptionsContext = useMemo(
+    () => ({
+      ...referenceInputOptionsContext,
+      disableNew: true,
+      EditReferenceLinkComponent,
+    }),
+    [referenceInputOptionsContext],
+  )
+
+  useScrollMirror([previousPaneElement, nextPaneElement])
+
+  return (
+    <ReferenceInputOptionsContext.Provider value={diffViewReferenceInputOptionsContext}>
+      <Dialog
+        id="diffView"
+        width="auto"
+        onClose={exitDiffView}
+        padding={false}
+        __unstable_hideCloseButton
+      >
+        <DialogLayout>
+          {mode === 'version' && <VersionModeHeader documentId={documentId} state={state} />}
+          {state === 'ready' && (
+            <>
+              <DiffViewPane
+                documentType={documents.previous.type}
+                documentId={documents.previous.id}
+                role="previous"
+                ref={setPreviousPaneElement}
+                scrollElement={previousPaneElement}
+                syncChannel={syncChannel}
+                compareDocument={documents.previous}
+              />
+              <DiffViewPane
+                documentType={documents.next.type}
+                documentId={documents.next.id}
+                role="next"
+                ref={setNextPaneElement}
+                scrollElement={nextPaneElement}
+                syncChannel={syncChannel}
+                // The previous document's edit state is used to calculate the diff inroduced by the next document.
+                compareDocument={documents.previous}
+              />
+            </>
+          )}
+        </DialogLayout>
+      </Dialog>
+    </ReferenceInputOptionsContext.Provider>
+  )
+}
--- a/packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
+++ b/packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
@@ -1,0 +1,238 @@
+import {type Path, type SanityDocument} from '@sanity/types'
+import {
+  BoundaryElementProvider,
+  Card,
+  Container as UiContainer,
+  DialogProvider,
+  PortalProvider,
+} from '@sanity/ui'
+import {noop} from 'lodash'
+import {
+  type CSSProperties,
+  forwardRef,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react'
+import {
+  ChangeIndicatorsTracker,
+  createPatchChannel,
+  FormBuilder,
+  getPublishedId,
+  getVersionFromId,
+  isDraftId,
+  isPublishedId,
+  isVersionId,
+  LoadingBlock,
+  useDocumentForm,
+  useEditState,
+  VirtualizerScrollInstanceProvider,
+} from 'sanity'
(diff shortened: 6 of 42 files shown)


pedrobonamin (commented):
This is fantastic, it's so great to make it possible to compare the two documents with this type of diff view. Great work! 
Added some suggestions and we are following this up in Slack.

pedrobonamin:
> Comments in diff view are not well supported, because the modal does not support document inspectors. I suggest we switch off comments in the diff view for now.
I think this is a very valid suggestion, the same could apply to the other plugins like `assist` and `translate`, I think this should not be a place where people come to edit the document and use other plugins, rather only compare them and make normal edits.

packages/sanity/src/core/releases/hooks/utils.ts
```diff
@@ -54,13 +54,16 @@ export function getReleasesPerspectiveStack({
...
+  selectedPerspectiveName: ReleaseId | undefined | 'published' | 'draft'
   releases: ReleaseDocument[]
   excludedPerspectives: string[]
 }): string[] {
   if (!selectedPerspectiveName || selectedPerspectiveName === 'published') {
     return []
   }
+  if (selectedPerspectiveName === 'draft') {
+    return [DRAFTS_FOLDER].filter((name) => !excludedPerspectives.includes(name))
+  }
```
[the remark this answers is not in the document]
juice49: Great point, thank you! Yes, I think you're right. This has now been removed.

packages/sanity/src/core/releases/plugin/documentActions/index.ts
```diff
@@ -10,8 +10,9 @@ export default function resolveDocumentActions(
...
   const duplicateAction = existingActions.filter(({name}) => name === 'DuplicateAction')
-
-  return context.versionType === 'version'
-    ? duplicateAction.concat(DiscardVersionAction).concat(UnpublishVersionAction)
-    : existingActions
+  return [
+    ...(context.versionType === 'version'
+      ? duplicateAction.concat(DiscardVersionAction).concat(UnpublishVersionAction)
+      : existingActions),
+  ]
```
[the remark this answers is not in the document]
juice49: Good catch, thanks! This is left over from a time an action was being appended here regardless of the context. Not needed anymore, so I've removed it.

packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+// TODO: Switch off comments. Document inspectors cannot currently be shown inside the diff view.
+// TODO: Switch off references pane. It should be a hyperlink instead.
+export const DiffViewPane = forwardRef<HTMLDivElement, DiffViewPaneProps>(function DiffViewPane(
+  {role, documentType, documentId, scrollElement, syncChannel, compareDocument},
+  ref,
+) {
+  const paneId = ['diffView', role].join('.')
+
+  const version: 'draft' | 'published' | string = isDraftId(documentId)
```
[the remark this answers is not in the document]
juice49: Done, thanks! This had to be refactored a bit.

juice49:
@pedrobonamin thanks so much for the review. I've made a few simplifications and improvements here:

- The previous document inference when navigating to the modal has been removed. It was janky because it blocked the modal rendering and messed with browser history. It was also unnecessary, because the document id stack is already known when navigating to the modal from a document action (which is how we now do it).
- The "Compare versions" document action now only appears if there is a previous version to compare with. This includes published and draft versions of the document.
- References now open in a new browser tab, rather than causing the modal to close. The previous UI was quite disorientating.
- Reference fields can no longer be used to create a new document, for the same reason as above. This is done by setting `disableNew` on all references inside the modal.

pedrobonamin (changes requested):
This looks very good Ash, thanks for building this complex piece!
I think it's only missing to solve the error that I sent you by slack and we should find a solve for the listeners.

packages/sanity/src/structure/hooks/useFilteredReleases.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @internal
+ */
+export function useFilteredReleases({
+  displayed,
+  documentId,
+}: Pick<DocumentPaneContextValue, 'displayed' | 'documentId'>): FilterReleases {
+  const {selectedReleaseId} = usePerspective()
+  const {data: releases} = useActiveReleases()
+  const {data: archivedReleases} = useArchivedReleases()
+  // const {editState, displayed, documentType, documentId} = useDocumentPane()
```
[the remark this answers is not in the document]
juice49: Thanks!

packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
```diff
@@ -0,0 +1,155 @@
...
+  role: 'previous' | 'next'
+  scrollElement: HTMLElement | null
+  syncChannel: PathSyncChannel
+  compareDocument: {
+    type: string
+    id: string
+  }
+}
+
+// TODO: Switch off comments. Document inspectors cannot currently be shown inside the diff view.
```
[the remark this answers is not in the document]
juice49: That is interesting. It actually appears comments have stopped appearing in the comparison view anyway, thanks to some outside change. I wonder whether this is the one.

packages/sanity/src/core/releases/tool/components/Chip.tsx
```diff
@@ -1,11 +1,17 @@
...
-import {type ReactNode} from 'react'
+import {Box, Card, type CardProps, Flex, Text} from '@sanity/ui'
+import {type ComponentType, type ReactNode} from 'react'
 
-export function Chip(props: {avatar?: ReactNode; text: ReactNode; icon?: ReactNode}) {
+type Props = {
+  avatar?: ReactNode
+  text: ReactNode
+  icon?: ReactNode
+} & Pick<CardProps, 'tone'>
```
bjoerge, line 8: nit: think I'd prefer `tone: CardProps['tone']` here over intersection type

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -106,9 +106,38 @@ export function getDraftId(id: string): DraftId {
...
 }
 
+/** @internal */
+export const systemBundles = ['drafts', 'published'] as const
+
+/** @internal */
+export type SystemBundle = (typeof systemBundles)[number]
+
+/** @internal */
+export function isSystemBundle(maybeSystemBundle: unknown): maybeSystemBundle is SystemBundle {
```
bjoerge, line 116: I like this name

packages/sanity/src/structure/diffView/components/DiffView.tsx
```diff
@@ -0,0 +1,72 @@
...
+    }),
+    [referenceInputOptionsContext],
+  )
+
+  useScrollMirror([previousPaneElement, nextPaneElement])
+
+  return (
+    <ReferenceInputOptionsContext.Provider value={diffViewReferenceInputOptionsContext}>
+      <Dialog
+        id="diffView"
```
bjoerge, line 37: Are we ok with this becoming a global id (and also a global variable), or should it be coming from useId()? Could lead to unpleasant surprises if we have multiple DiffView's on the page at the same time.

packages/sanity/src/structure/diffView/hooks/useDiffViewRouter.ts
```diff
@@ -0,0 +1,90 @@
...
+  DIFF_VIEW_NEXT_DOCUMENT_SEARCH_PARAMETER,
+  DIFF_VIEW_PREVIOUS_DOCUMENT_SEARCH_PARAMETER,
+  DIFF_VIEW_SEARCH_PARAMETER,
+} from '../constants'
+import {type DiffViewMode} from '../types/diffViewMode'
+
+type NavigateDiffView = (
+  options: {
+    mode?: DiffViewMode
+  } & Partial<
```
bjoerge, line 16: Would it be slightly more readable if options was typed `{mode?: DiffViewMode, previousDocument?: {type: string, id: string}}, nextDocument?: ?: {type: string, id: string}}`

packages/sanity/src/structure/diffView/versionMode/components/VersionModeHeader.tsx
```diff
@@ -0,0 +1,376 @@
...
+`
+
+/**
+ * The header component that is rendered when diff view is being used to compare versions of a
+ * document.
+ *
+ * @internal
+ */
+export const VersionModeHeader: ComponentType<
+  {state: 'pending' | 'ready' | 'error'} & Pick<DocumentLayoutProps, 'documentId'>
```
bjoerge, line 63: (same comment as earlier re: intersection vs referencing to a type property)

packages/sanity/src/structure/diffView/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @internal
+ */
+export const DIFF_VIEW_NEXT_DOCUMENT_SEARCH_PARAMETER = 'nextDocument'
+
+/**
+ * @internal
+ */
+export const DIFF_SEARCH_PARAM_DELIMITER = ';'
```
bjoerge, line 19: Do we run this through encodeURIComponent? Since `;` is used to terminate various statements (sql, bash commands) etc, it's not uncommon for it to be blocked by various security middleware. If not too much work, I suggest we change to another delimiter.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveList.tsx
```diff
@@ -85,62 +78,11 @@ export const DocumentPerspectiveList = memo(function DocumentPerspectiveList() {
...
-  }, [
-    archivedReleases,
-    isCreatingDocument,
-    displayed?._id,
-    documentVersions,
-    params?.historyVersion,
-    releases,
-    selectedReleaseId,
-  ])
+  const filteredReleases = useFilteredReleases({displayed, documentId})
```
bjoerge, line 85: nice refactor!

bjoerge (commented):
Amazing work! My comments are mostly nits that you can feel free to disregard. The only thing I consider somewhat blocking is the use of `;` as delimited. I couldn't tell easily whether it will be properly urlencoded. If it's not encoded I think we need to change it to something else (`‚`?) It's less of an issue if we end up url-encoding it (but still suggest we change it unless it's too much work).

packages/sanity/src/core/releases/tool/components/Chip.tsx
juice49: That's fair. I tend to prefer `pick` where I want the API to inherit an identical property name **and** value, as it more concretely establishes that relationship.

Worth thinking about, though. If you ever have a few mins for nerding out, I'd love to hear counterpoints on it 🙂.

packages/sanity/src/structure/diffView/constants.ts
bjoerge: For all we know middlewares might also block `%3B` because they know developers might as well decode that string befor passing it to system exec() 😅  Feels safer to replace with `‚` and I don't think that should be problematic here anyway as we don't allow that in either document types nor document ids(?). Maybe I'm being overly cautious…

packages/sanity/src/structure/diffView/components/DiffView.tsx
juice49: You're absolutely right. Thanks for pointing it out.

packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -53,10 +52,10 @@ import {CreatedDraft} from './__telemetry__/form.telemetry'
 interface DocumentFormOptions {
   documentType: string
   documentId: string
-  releaseId?: ReleaseId
+  releaseId?: string
```
pedrobonamin, line 55: Any reason this is changing? I know it's not ideal right now, given it is eventually an string, but if we later add the branded types I think it will make sense again.

packages/sanity/src/core/perspective/PerspectiveProvider.tsx
```diff
@@ -44,8 +45,9 @@ export function PerspectiveProvider({
     () => ({
       selectedPerspective,
       selectedPerspectiveName,
-      selectedReleaseId:
-        selectedPerspectiveName === 'published' ? undefined : selectedPerspectiveName,
+      selectedReleaseId: isSystemBundleName(selectedPerspectiveName)
```
pedrobonamin, line 48: Is it possible to have here a `drafts` as `selectedPerspectiveName`? 
I think this could end going against we have been thinking in where the `drafts` is the undefined selectedPerspectiveName

packages/sanity/src/core/releases/tool/components/Chip.tsx
bjoerge: Good point, I didn't really consider the enforcing of both type and property name, I'm trending towards agreeing with you now 😄

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -106,9 +106,38 @@ export function getDraftId(id: string): DraftId {
...
+  return systemBundles.includes(maybeSystemBundle as SystemBundle)
+}
+
+/** @internal */
+const systemBundleNames = ['draft', 'published'] as const
```
pedrobonamin, line 121: Curious on why we have `drafts` as systemBundles and `draft` as name, do we need the two? Could it be error prone to have two different strings that eventually means the same?

pedrobonamin: Seeing your comment below, maybe we can update the uses of `draft` for `drafts` ? Do you have those identified? Not necessarily for this PR btw.

packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
```diff
@@ -0,0 +1,238 @@
...
+    onSetCollapsedFieldSet,
+    onSetCollapsedPath,
+    collapsedFieldSets,
+    ready,
+    collapsedPaths,
+    schemaType,
+    value,
+    onProgrammaticFocus,
+    ...documentForm
+  } = useDocumentForm({
```
pedrobonamin, line 86: 🤩 I'm seeing some future simplifications to the `useDocumentForm` , the perspective and release id could be derived from the the document id if not provided.
so you could pass a raw document id and the hook handles the rest

packages/sanity/src/structure/diffView/plugin/DiffViewDocumentLayout.tsx
```diff
@@ -0,0 +1,41 @@
+import {useToast} from '@sanity/ui'
+import {type ComponentType, type PropsWithChildren} from 'react'
+import {type DocumentLayoutProps, useTranslation} from 'sanity'
+
+import {structureLocaleNamespace} from '../../i18n'
+import {DiffView} from '../components/DiffView'
+import {useDiffViewState} from '../hooks/useDiffViewState'
+
+export const DiffViewDocumentLayout: ComponentType<
```
pedrobonamin, line 9: nit: this is living now in a `/plugin` folder, maybe it should be part of `Components`?

pedrobonamin (approved):
This is great Ash, I've been trying it and it's a game changer for releases. Will make it very easy to compare two documents and even resolve differences because you got the option to edit them in place.
Thanks for building this huge piece!! 🙌🏻

packages/sanity/src/core/form/useDocumentForm.ts
juice49: Apologies—I made this change while `ReleaseId` was still `r${string}` and marked as deprecated—I assumed any usage of `ReleaseId` was out of date.

I see it's now been changed to simply `string` with a todo to move it to `@sanity/id-utils`. I'll drop this commit! 🙂

packages/sanity/src/structure/diffView/constants.ts
juice49: Very good points. Changed to `,` (`%2C`) 🙂.

https://github.com/sanity-io/sanity/blob/eb55bf407b2cc61c06707c7dd4394174bb7e02f5/packages/sanity/src/structure/diffView/constants.ts#L19

packages/sanity/src/core/form/useDocumentForm.ts
juice49: Dropped ✅

packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
juice49: Totally! I didn't want to refactor anything as this point, but I think there are improvements we can make. Thanks for making the reusable form happen 💖.

packages/sanity/src/structure/diffView/plugin/DiffViewDocumentLayout.tsx
juice49: You are right. A remnant of when this was actually a plugin 😔.

packages/sanity/src/core/util/draftUtils.ts
juice49: I agree that we should make this consistent, but I did not identify all the places it's used as part of this task.

pedrobonamin: We can follow up on it, no need to block this

packages/sanity/src/core/perspective/PerspectiveProvider.tsx
juice49: Hmm. If `selectedPerspectiveName` is `undefined`, then `selectedReleaseId` will resolve to `undefined`. If the intention is to view drafts, wouldn't it be correct that `selectedReleaseId` is `undefined`?

bjoerge (approved):
Awesome work, @juice49! 🚢

PR #7912, opened 2024-11-28
fix: vertically center pane header when no tabs

stipsan:
### Description

Previously the difference in height on the document pane header weren't very noticeable as it was perceived as a toolbar with two rows, while presentation tool and tasks were single row.

This is n a release nowa draft

Annotated with guides
Pasted Graphic 1

This changed with #7462, which moved the revision dropdown into the top row, making the toolbar single row as well unless custom tabs are defined in structure builder:

Pasted Graphic 2

With guides:
Pasted Graphic 3

For neurodivergents like me this tiny misalignment becomes a constant distraction and frustrating. It's like an airhorn, and there is no noise cancelling headphones. So I had no choice but to make this PR 😅

Now it's perfectly balanced, like everything should be:
Pasted Graphic 4

Pasted Graphic 5

Without regressing the custom tabs layout
Gandalf on a draft jkj as as as as

### What to review

Test the document tabs, I hope I haven't missed a spot 🙌 

### Testing

### Notes for release

Fixed a regression in v3.65.0 that caused the document editor panel header to look slightly misaligned with presentation tool's URL bar, and the Tasks panel header.

### DIFF
--- a/packages/sanity/src/structure/components/pane/PaneHeader.tsx
+++ b/packages/sanity/src/structure/components/pane/PaneHeader.tsx
@@ -67,7 +67,7 @@ export const PaneHeader = forwardRef(function PaneHeader(
               gap={1}
               onClick={handleLayoutClick}
               padding={3}
-              paddingBottom={collapsed ? 3 : 2}
+              paddingBottom={collapsed || !showTabsOrSubActions ? 3 : 2}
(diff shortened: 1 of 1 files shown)


pedrobonamin (approved):
Great catch 😄 

> Balance

image

Issue #2966, opened 2021-12-01
The crop/hotspot image editor does not work for svg images


jfulse:
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

The crop/hotspot image editor does not work for svg images

**To Reproduce**

1. Upload an svg image on a field that has `options: { hotspot: true }`
2. Click on "Edit details"
3. The canvas is empty

**Expected behavior**

I'd suggest either it should be possible to adjust the svg somehow or the "Edit details" button should be hidden.

It's probably possible to hide the button oneself with a functional `hidden` in the schema, but I thought you might like to be notified anyway.

**Screenshots**
<img width="677" alt="Skjermbilde 2021-12-01 kl  16 32 56" src="https://user-images.githubusercontent.com/13033324/144263850-e2d0dc4f-9402-4240-ab82-c77ed2321149.png">

**Which versions of Sanity are you using?**

@sanity/cli              2.19.1 (latest: 2.22.3)
@sanity/base             2.21.9 (latest: 2.22.3)
@sanity/cli              2.21.7 (latest: 2.22.3)
@sanity/core             2.21.8 (latest: 2.22.3)
@sanity/default-layout   2.21.9 (latest: 2.22.3)
@sanity/default-login    2.21.9 (latest: 2.22.3)
@sanity/desk-tool        2.21.9 (latest: 2.22.3)
@sanity/imagetool       2.21.10 (latest: 2.22.3)
@sanity/react-hooks      2.21.9 (latest: 2.22.3)
@sanity/vision           2.21.9 (latest: 2.22.3)

**What operating system are you using?**
macOS Big Sur

**Which versions of Node.js / npm are you running?**

8.1.3
v16.11.0

**Additional context**

**Security issue?**

Any security issues should be submitted directly to [security@sanity.io](mailto:security@sanity.io). In order to determine whether you are dealing with a security issue, ask yourself these two questions:

* Can I access something that's not mine, or something I shouldn't have access to?
* Can I disable something for other people? If the answer to either of those two questions are "yes", then you're probably dealing with a security issue. Note that even if you answer "no" to both questions, you may still be dealing with a security issue, so if you're unsure, just email us at [security@sanity.io](mailto:security@sanity.io.

hacknug:
> I'd suggest either it should be possible to adjust the svg somehow or the "Edit details" button should be hidden.

[...]

PR #7917, opened 2024-11-29
fix: add svg hotspot+crop gotcha to image tool dialog

bjoerge:
### Description
The fact that hotspot and crop won't be applied to SVG images is [unexpected](https://github.com/sanity-io/sanity/issues/2966) and can easily become a time sink to debug because it fails silently, and we don't offer much docs around it either (at least to my knowledge). This PR adds a gotcha to the hotspot/crop UI in case you're setting it on an SVG.

Note: I believe we still want to alow setting hotspot/crop for SVGs because it works fine when appending `fmt=jpg` to the image url.

image

### What to review
- Is the phrasing clear/sensible? Note: this message might appear a bit confusing to non-devs. I don't think we have an easy way to tell whether the current user is a developer or not, and I think it's an OK tradeoff showing it to everyone. I did consider only showing it during DEV, but I think that would make it less likely to be noticed as it requires the developer to actually upload an SVG when running locally. We could consider making it dismissable if we think it's too confusing.

### Testing
- Upload an SVG to an image field (for example [this](https://upload.wikimedia.org/wikipedia/commons/f/fd/Ghostscript_Tiger.svg)) and try to modify hotspot/crop

### Notes for release
- Adds a warning about limited hotspot/crop support for SVGs in the hotspot/crop dialog for image fields

### DIFF
--- a/packages/sanity/src/core/form/components/Details.tsx
+++ b/packages/sanity/src/core/form/components/Details.tsx
@@ -1,5 +1,5 @@
 import {ToggleArrowRightIcon} from '@sanity/icons'
-import {Box, Flex, Inline, Text} from '@sanity/ui'
+import {Box, Flex, Text} from '@sanity/ui'
 import {type ReactNode, useCallback, useEffect, useState} from 'react'
 import {styled} from 'styled-components'
 
@@ -37,9 +37,10 @@ const ToggleArrow = styled(ToggleArrowRightIcon)<{open: boolean}>`
 
 const Header = styled(Flex)`
   cursor: default;
+  line-height: 0;
 `
 
-const IconBox = styled(Box)`
+const IconBox = styled(Flex)`
   & > div > svg {
     transform: rotate(0);
     transition: transform 100ms;
@@ -61,8 +62,8 @@ export function Details(props: DetailsProps) {
   return (
     <Box {...restProps}>
       <HeaderButton type="button" onClick={handleToggle}>
-        <Header align="center">
-          <Inline>
+        <Header>
+          <Flex align="center">
             <IconBox data-open={open ? '' : undefined}>
               <Text size={1}>
                 <ToggleArrow open={open} />
@@ -70,11 +71,11 @@ export function Details(props: DetailsProps) {
             </IconBox>
             {icon && <Box marginLeft={1}>{icon}</Box>}
             <Box flex={1} marginLeft={1}>
-              <Text size={1} weight="medium">
+              <Text textOverflow="ellipsis" size={1} weight="medium">
                 {title}
               </Text>
             </Box>
-          </Inline>
+          </Flex>
         </Header>
       </HeaderButton>
 
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -654,10 +654,12 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   /** Error: `{{errorMessage}}` */
   'inputs.imagetool.load-error': 'Error: {{errorMessage}}',
   /** SVG images are not adjusted for hotspot and crop when served from the Sanity image. If you want the below hotspot and crop settings to apply, make sure to append `fp=jpg` to the image url, or call `format('jpg')` if using `@sanity/image-url` */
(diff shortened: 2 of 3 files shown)


packages/sanity/src/core/form/components/Details.tsx
```diff
@@ -61,20 +62,20 @@ export function Details(props: DetailsProps) {
...
+          <Flex align="center">
             <IconBox data-open={open ? '' : undefined}>
               <Text size={1}>
                 <ToggleArrow open={open} />
               </Text>
             </IconBox>
             {icon && <Box marginLeft={1}>{icon}</Box>}
             <Box flex={1} marginLeft={1}>
-              <Text size={1} weight="medium">
+              <Text textOverflow="ellipsis" size={1} weight="medium">
```
bjoerge, line 74: not strictly related to this PR, but I noticed the AlertStrip got word breaks on narrow screens and was not rendering nicely.

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -653,6 +653,13 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
     'Adjust the rectangle to crop image. Adjust the circle to specify the area that should always be visible.',
   /** Error: `{{errorMessage}}` */
   'inputs.imagetool.load-error': 'Error: {{errorMessage}}',
+  /** SVG images are not adjusted for hotspot and crop when served from the Sanity image. If you want the below hotspot and crop settings to apply, make sure to append `fp=jpg` to the image url, or call `format('jpg')` if using `@sanity/image-url` */
+  'inputs.imagetool.svg-warning.developer-info': `SVG images are not adjusted for hotspot and crop when served from the Sanity image API. If you want the below hotspot and crop settings to apply, make sure to append  <code>fp=jpg</code> to the <ImageUrlDocumentationLink>image url</ImageUrlDocumentationLink>, or call <code>format('jpg')</code> if using <ImageUrlPackageDocumentationLink>@sanity/image-url</ImageUrlPackageDocumentationLink>`,
```
juice49, line 657: Tiny thing: There is a double space between "append"  "<code>fp=jpg</code>".

juice49, line 657: Tiny thing: "url" in "image url" should probably be uppercase.

juice49, line 657: Tiny thing: We should probably add a period to the end of this sentence for consistency.

juice49 (dismissed):
This warning looks great to me 🙏. I left a couple of nitpicks about the copy.

packages/sanity/src/core/i18n/bundles/studio.ts
bjoerge: Thanks for pointing out my poor writing and silly mistakes! (I'm hopeless at typing when I don't have a linter yelling at me 😄)

I'll address these, but think @kmelve had some suggestions about this too, so awaiting his feedback before fixing.

kmelve (changes requested):
Added a suggestion for new copy

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -653,6 +653,11 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
     'Adjust the rectangle to crop image. Adjust the circle to specify the area that should always be visible.',
   /** Error: `{{errorMessage}}` */
   'inputs.imagetool.load-error': 'Error: {{errorMessage}}',
+  /** SVG images are not adjusted for hotspot and crop when served from the Sanity image. If you want the below hotspot and crop settings to apply, make sure to append `fp=jpg` to the image url, or call `format('jpg')` if using `@sanity/image-url` */
+  'inputs.imagetool.svg-gotcha.description': `SVG images are not adjusted for hotspot and crop when served from the Sanity image. If you want the below hotspot and crop settings to apply, make sure to append  <code>fp=jpg</code> to the <ImageUrlDocumentationLink>image url</ImageUrlDocumentationLink>, or call <code>format('jpg')</code> if using <ImageUrlPackageDocumentationLink>@sanity/image-url</ImageUrlPackageDocumentationLink>`,
```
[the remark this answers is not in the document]
bjoerge: Thanks! Applied in cd1864e0ea

juice49 (approved):

PR #7924, opened 2024-12-02
fix(core):  `VersionOriginType` to `DocumentVariantType`

pedrobonamin:
### Description
Renames `VersionOriginType` to `DocumentVariantType`, given this type is useful in more contexts not only the creation of a document.
This type will indicate the type of document we are working with according to the id.
If it's a document that starts with `version` it's a version document.
If it's a document that starts with `drafts` it's a draft document.
Otherwise, it's a published document.

It's quite obvious when you see the id, but exposing a function that we can use to get this information in a type safe way is necessary when we need the ui, or other internal functions, to react different according to the type of document we are using.

### DIFF
--- a/packages/sanity/src/core/util/draftUtils.ts
+++ b/packages/sanity/src/core/util/draftUtils.ts
@@ -210,3 +210,16 @@ export function removeDupes(documents: SanityDocumentLike[]): SanityDocumentLike
     .map((entry) => entry.draft || entry.published)
     .filter(isNonNullable)
 }
+
+/**
+ * @beta
+ * Given a document id, it indicates whether it is a draft, version or published document.
+ */
+export type DocumentVariantType = 'draft' | 'version' | 'published'
+
+/** @beta */
+export function getDocumentVariantType(documentId: string): 'draft' | 'version' | 'published' {
+  if (isDraftId(documentId)) return 'draft'
+  if (isVersionId(documentId)) return 'version'
+  return 'published'
+}
--- a/packages/sanity/src/core/util/draftUtils.ts
+++ b/packages/sanity/src/core/util/draftUtils.ts
@@ -210,16 +210,3 @@ export function removeDupes(documents: SanityDocumentLike[]): SanityDocumentLike
     .map((entry) => entry.draft || entry.published)
     .filter(isNonNullable)
 }
-
-/**
- * @beta
- * Given a document id, it indicates whether it is a draft, version or published document.
- */
-export type DocumentVariantType = 'draft' | 'version' | 'published'
-
-/** @beta */
-export function getDocumentVariantType(documentId: string): 'draft' | 'version' | 'published' {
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -210,3 +210,16 @@ export function removeDupes(documents: SanityDocumentLike[]): SanityDocumentLike
...
+ * @beta
+ * Given a document id, it indicates whether it is a draft, version or published document.
+ */
+export type DocumentVariantType = 'draft' | 'version' | 'published'
+
+/** @beta */
+export function getDocumentVariantType(documentId: string): 'draft' | 'version' | 'published' {
+  if (isDraftId(documentId)) return 'draft'
+  if (isVersionId(documentId)) return 'version'
+  return 'published'
```
RitaDias, line 224: Not strictly necessarily for approval, but Since it's a `@beta` we can probably give it just a bit more context in the comments generically about what it does and returns :) the description you have on this PR is pretty good so I would just adapt it to here!

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -210,3 +210,16 @@ export function removeDupes(documents: SanityDocumentLike[]): SanityDocumentLike
     .map((entry) => entry.draft || entry.published)
     .filter(isNonNullable)
 }
+
+/**
+ * @beta
+ * Given a document id, it indicates whether it is a draft, version or published document.
```
RitaDias, line 216: Another super minor nitpick the `@beta` and etc usually come right at the end!

```suggestion
 * Given a document id, it indicates whether it is a draft, version or published document.
 * @beta
```

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -210,3 +210,16 @@ export function removeDupes(documents: SanityDocumentLike[]): SanityDocumentLike
...
+ */
+export type DocumentVariantType = 'draft' | 'version' | 'published'
```
RitaDias, line 218: Speaking of types, I think this probably needs to be moved to operations/types instead of living in the `draftUtils` :)

RitaDias (approved):
Approved with some minor things that can be improved but are not necessary for the merging :)

packages/sanity/src/core/util/draftUtils.ts
pedrobonamin: Thanks for the suggestion Rita, I agree with you that `draftUtils` might not be the best place.
However, I think `operation/types `is coupled too much to the `document-pair` and this will be used in more places than that once we have this PR merged https://github.com/sanity-io/sanity/pull/7922

Just pushed a change moving it to its own file inside `core/util/getDocumentVariantType`.

RitaDias: Fair enough :)

PR #7925, opened 2024-12-02
fix(core): when publishing a release unpin it

pedrobonamin:
### Description
This restores the changes introduced in https://github.com/sanity-io/sanity/pull/7862 which seems to be lost from the `corel` branch.
Once a released is published (archived), it should be unpinned if it was pinned, to avoid bugs in the rest of the system.


bjoerge (approved):
Thank you Pedro!

PR #7926, opened 2024-12-02
test(e2e): restore last e2e tests

RitaDias:
### Description

- We had some e2e tests that were disabled due to being unreliable while work with the Actions was on going. As things seem more stable, this PR brings back those tests.
- Some updates had to be done to data-test ids since a lot of them were breaking (not correctly formatted, hence why the large number of files)

### What to review

Review tests and make sure they make sense, I have tried reducing the flakiness but they are e2e tests so 🫠 any pointers would be appreciated.

### Testing

This PR is just tests 

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
+++ b/packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
@@ -108,6 +108,7 @@ function PaneContextMenuItem(props: {disabled?: boolean; node: _PaneMenuItem}) {
         pressed={node.selected}
         text={title}
         tone={node.tone}
+        data-testid={`action-${node.title?.toLocaleLowerCase().replace(' ', '')}`}
       />
     </TooltipOfDisabled>
   )
--- a/packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
+++ b/packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
@@ -6,6 +6,7 @@ import {useIntentLink} from 'sanity/router'
 
 import {MenuGroup, MenuItem, type PopoverProps} from '../../../ui-components'
 import {type Intent} from '../../structureBuilder'
+import {toLowerCaseNoSpaces} from '../../util/toLowerCaseNoSpaces'
 import {type _PaneMenuItem, type _PaneMenuNode} from './types'
 
 const MENU_GROUP_POPOVER_PROPS: PopoverProps = {
@@ -108,7 +109,7 @@ function PaneContextMenuItem(props: {disabled?: boolean; node: _PaneMenuItem}) {
         pressed={node.selected}
         text={title}
         tone={node.tone}
-        data-testid={`action-${node.title?.toLocaleLowerCase().replace(' ', '')}`}
+        data-testid={`action-${toLowerCaseNoSpaces(node.title)}`}
       />
     </TooltipOfDisabled>
   )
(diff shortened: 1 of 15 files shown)


jordanl17:
@RitaDias to confirm - the e2es passing in github actions here on the PR is validation that these tests are all passing?

RitaDias:
> @RitaDias to confirm - the e2es passing in github actions here on the PR is validation that these tests are all passing?

Yep!

packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
```diff
@@ -108,6 +108,7 @@ function PaneContextMenuItem(props: {disabled?: boolean; node: _PaneMenuItem}) {
         pressed={node.selected}
         text={title}
         tone={node.tone}
+        data-testid={`action-${node.title?.toLocaleLowerCase().replace(' ', '')}`}
```
jordanl17, line 111: should this also be `replaceAll` instead of single `replace`?

RitaDias: Ah yes, it should! Good catch! 
I'll fix this!

juice49: Small related thought: would it make sense to move this to a tiny helper function? I noticed it's repeated a few times, and having the code inline like this doesn't make it _super_ clear to me how/why the node title -> test id transformation works.

RitaDias: sure, that makes sense :)

packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
```diff
@@ -56,7 +56,8 @@ export const FormView = forwardRef<HTMLDivElement, FormViewProps>(function FormV
   } = useDocumentPane()
   const documentStore = useDocumentStore()
   const presence = useDocumentPresence(documentId)
-  const {title} = useDocumentTitle()
+  const currentTitle = useDocumentTitle()
+  const {title} = value ?? currentTitle
```
juice49, line 60: I followed this code, as I was wondering why `title as string` is necessary on line 191.

It looks like it's because we're now primarily grabbing `title` field from the document's value, rather than the `useDocumentTitle` hook. Doesn't this logic assume the document has a `title` field, which might not be the case?

juice49:
Thanks for fixing this 🙏. I left a comment to clarify one thing, but in general the tests look good to me.

packages/sanity/src/structure/panes/document/documentPanel/documentViews/FormView.tsx
RitaDias: That is true, that is why I set the currentTitle as default. The issue with the way that has been is that it doesn't update the title when the revision changes if it doesn't use the value.

Though maybe the solution should be in a different place (like in the useDocumentTitle)...
I'll make a note of this and what I'll probably do is update the test to not rely on this title and instead rely on the input value

RitaDias: https://linear.app/sanity/issue/SAPP-1958/issue-where-the-document-title-is-not-being-updated

Created the issue, reverting this and fixing the test

packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
RitaDias: Done!

packages/sanity/src/structure/components/pane/PaneMenuButtonItem.tsx
```diff
@@ -108,6 +109,7 @@ function PaneContextMenuItem(props: {disabled?: boolean; node: _PaneMenuItem}) {
         pressed={node.selected}
         text={title}
         tone={node.tone}
+        data-testid={`action-${toLowerCaseNoSpaces(node.title)}`}
```
juice49, line 112: Noice 🙂.

juice49 (approved):

PR #7936, opened 2024-12-03
refactor(sanity): remove bundle-prefix, move away from bundlePerspective

bjoerge:
### Description
Moves away from bundlePerspective and `bundle.`-prefix and uses perspective/release layering implemented in https://github.com/sanity-io/client/pull/934

Also made an attempt at cleaning up some terminology to make things a bit more obvious


packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
```diff
@@ -47,12 +47,12 @@ export const getPreviewValueWithFallback = ({
     }
   } else {
     switch (true) {
-      case typeof resolveBundlePerspective(perspective) !== 'undefined' || isVersionId(value._id):
-        snapshot = version || draft || published
-        break
       case perspective === 'published':
         snapshot = published || draft
         break
+      case typeof perspective !== 'undefined' || isVersionId(value._id):
```
pedrobonamin, line 53: don't we need here a case to handle the drafts? Or maybe this should be
```suggestion
      case typeof perspective !== 'undefined' && perspective !== "drafts" || isVersionId(value._id):
```

packages/sanity/src/core/releases/hooks/usePerspective.tsx
```diff
@@ -147,36 +148,38 @@ export function usePerspective(): PerspectiveValue {
...
+        currentPerspectiveName === 'published' ? undefined : currentPerspectiveName,
       excludedPerspectives,
       setPerspective,
       setPerspectiveFromReleaseDocumentId: setPerspectiveFromReleaseDocumentId,
       setPerspectiveFromReleaseId: setPerspectiveFromReleaseId,
       toggleExcludedPerspective,
-      currentGlobalBundle,
-      currentGlobalBundleId: isPublishedPerspective(currentGlobalBundle)
+      selectedPerspective: currentPerspective,
+      globalReleaseDocumentId: isPublishedPerspective(currentPerspective)
```
pedrobonamin, line 168: does this needs to be `globalReleaseDocumentId` ? 
feels better: `selectedPerspectiveId`

packages/sanity/src/core/releases/tool/components/ReleaseDocumentPreview.tsx
```diff
@@ -40,9 +40,7 @@ export function ReleaseDocumentPreview({
               id: getPublishedId(documentId),
               type: documentTypeName,
             }}
-            searchParams={[
-              ['perspective', `bundle.${getBundleIdFromReleaseDocumentId(releaseId)}`],
-            ]}
+            searchParams={[['perspective', getBundleIdFromReleaseDocumentId(releaseId)]]}
```
pedrobonamin, line 43: NIT: Maybe rename also `getBundleIdFromReleaseDocumentId` to `getReleaseNameFromReleaseId` ?

packages/sanity/src/core/releases/util/util.ts
```diff
@@ -29,13 +24,11 @@ export function getDocumentIsInPerspective(
 
   if (!perspective) return !isVersionId(documentId)
 
-  const releasePerspective = resolveBundlePerspective(perspective)
```
pedrobonamin, line 32: so glad we could remove all of this checks

packages/sanity/src/core/store/_legacy/document/getPairListener.ts
```diff
@@ -77,7 +78,12 @@ export function getPairListener(
   options: PairListenerOptions = {},
 ): Observable<ListenerEvent> {
   const {publishedId, draftId, versionId} = idPair
-
+  if (
+    (idPair.versionId && getVersionFromId(idPair.versionId) === 'published') ||
```
pedrobonamin, line 82: I think this will never be true, it implies you could have a `versions.published.<id>` document

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -106,16 +109,20 @@ export function getDraftId(id: string): DraftId {
...
+  if (version === 'drafts' || version === 'published') {
+    throw new Error('Version can not be "published" or "drafts"')
+  }
+
   if (isVersionId(id)) {
     const [_versionPrefix, versionId, ...publishedId] = id.split(PATH_SEPARATOR)
-    if (versionId === bundle) return id
-    return `${VERSION_PREFIX}${bundle}${PATH_SEPARATOR}${publishedId}`
+    if (versionId === version) return id
+    return `${VERSION_PREFIX}${version}${PATH_SEPARATOR}${publishedId}`
```
pedrobonamin, line 120: NIT, but this could be removed and the function simplified by just going through the following block.

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -781,24 +786,22 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
       timelineStore,
       title,
       value,
-      version: bundlePerspective,
+      selectedVersionName: selectedPerspectiveName,
+      selectedReleaseName,
```
pedrobonamin, line 790: Do we need this two as part of the `DocumentPaneProvider` ? Now that the release is always global

pedrobonamin (approved):
Thanks for this change @bjoerge ! 
I think it all makes sense, added some suggestions but they are not blocking

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
bjoerge: Even though the client supports it, it should never be `drafts` in the studio code as that's that's always the default

packages/sanity/src/core/releases/hooks/usePerspective.tsx
```diff
@@ -34,12 +32,14 @@ export interface PerspectiveValue {
   toggleExcludedPerspective: (perspectiveId: string) => void
   /* Check if a perspective is excluded */
   isPerspectiveExcluded: (perspectiveId: string) => boolean
+  /* The excluded perspectives */
+  excludedPerspectives: string[]
   /**
    * The stacked array of releases ids ordered chronologically to represent the state of documents at the given point in time.
    */
-  bundlesPerspective: string[]
+  perspectiveStack: string[]
```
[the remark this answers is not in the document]
bjoerge: I couldn't think of a better name - but felt like this kinda works? Thought about releaseStack, but it can include `published`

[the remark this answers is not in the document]
pedrobonamin: I think the name is fine, I came to the same name in a chat with Cody. Just mentioning cody as this is something that we will need to change in presentation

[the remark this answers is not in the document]
stipsan: Thanks for the headsup!

packages/sanity/src/core/releases/tool/components/ReleaseDocumentPreview.tsx
bjoerge: Now that I think of it, we should strive for consistency in naming these `releaseId` vs `releaseDocumentId`, so I'll rename `selectedReleaseName` to `selectedReleaseId` and this one to `getReleaseIdFromReleaseDocumentId`

packages/sanity/src/core/store/_legacy/document/getPairListener.ts
bjoerge: It actually happened to me yesterday! So leaving it for now – it's just an extra safeguard and easy to remove later on

packages/sanity/src/core/util/draftUtils.ts
bjoerge: Oh, nice catch. So good when changes makes things less complicated 🙌🏼

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
bjoerge: Don't think so. I'm making a note to remove it. Good call!

packages/sanity/src/core/releases/hooks/usePerspective.tsx
```diff
@@ -147,36 +148,38 @@ export function usePerspective(): PerspectiveValue {
     (perspectiveId: string) =>
       Boolean(
         excludedPerspectives?.includes(
-          isPublishedPerspective(perspectiveId) ? 'published' : `bundle.${perspectiveId}`,
+          isPublishedPerspective(perspectiveId) ? 'published' : perspectiveId,
```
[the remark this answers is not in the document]
bjoerge: Good catch, fixed!

packages/sanity/src/core/releases/hooks/usePerspective.tsx
```diff
@@ -82,27 +84,28 @@ export function usePerspective(): PerspectiveValue {
...
-        (release) => `bundle.${getBundleIdFromReleaseDocumentId(release._id)}` === perspective,
+        (release) => getBundleIdFromReleaseDocumentId(release._id) === currentPerspectiveName,
       )
     ) {
       setPerspective(LATEST._id)
     }
-  }, [archivedReleases, perspective, selectedBundle, setPerspective])
+  }, [archivedReleases, currentPerspectiveName, selectedBundle, setPerspective])
 
   // TODO: Improve naming; this may not be global.
```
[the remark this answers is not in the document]
bjoerge: Done

packages/sanity/src/core/util/draftUtils.ts
bjoerge: Fixed

packages/sanity/src/core/releases/tool/components/ReleaseDocumentPreview.tsx
pedrobonamin: I've been thinking of name and id as different things in releases.
The id for me has been the string that starts with `_.releases.<name>` and the name has been the value we have in the `document.name`. Are we aligned on this? 

So: 
- id: `_.releases.r123`
- name: `r123`

pedrobonamin: Another use case for branded types ;)

bjoerge:
Merged via 51131891d9

PR #7937, opened 2024-12-03
fix(vision): add a warning about unsupported perspective when pasting

bjoerge:
### Description

https://github.com/sanity-io/sanity/pull/7935 broke (internal) vision types because of the changes introduced in https://github.com/sanity-io/client/pull/934

This is a quickfix that shows a warning if you try to paste an url with a perspective param not included in the allowlist defined here. This should fix the issue at hand, and be good enough for now until we land support for releases/perspective layering in vision

### What to review
Does the change look ok?

### Testing

The tests should be sufficient here.

### Notes for release
n/a - this warning would likey not appear to anyone for a while

### DIFF
--- a/packages/@sanity/vision/src/SUPPORTED_PERSPECTIVES.ts
+++ b/packages/@sanity/vision/src/SUPPORTED_PERSPECTIVES.ts
@@ -1,0 +1,15 @@
+import {type ClientPerspective} from '@sanity/client'
+
+export type SupportedPerspective = 'raw' | 'previewDrafts' | 'published' | 'drafts'
+
+export const SUPPORTED_PERSPECTIVES = [
+  'raw',
+  'previewDrafts',
+  'published',
+  'drafts',
+] satisfies ClientPerspective[]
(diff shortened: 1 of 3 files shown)


packages/@sanity/vision/src/SUPPORTED_PERSPECTIVES.ts
```diff
@@ -0,0 +1,15 @@
+import {type ClientPerspective} from '@sanity/client'
+
+export type SupportedPerspective = 'raw' | 'previewDrafts' | 'published' | 'drafts'
+
+export const SUPPORTED_PERSPECTIVES = [
+  'raw',
+  'previewDrafts',
+  'published',
+  'drafts',
```
pedrobonamin, line 9: Does something changes from `previewDrafts` to `drafts`?  I think they will return the same

stipsan (approved):

packages/@sanity/vision/src/SUPPORTED_PERSPECTIVES.ts
bjoerge: yep, they are equivalent

PR #7941, opened 2024-12-04
fix: use the latest `corel` version of `@sanity/presentation`

stipsan:
There's more work to be done, but at least this version works with all the latest changes, and don't require a special version of `@sanity/react-loader` nor `@sanity/visual-editing` 🙌


package.json
```diff
@@ -186,8 +186,8 @@
     "overrides": {
       "@npmcli/arborist": "^7.5.4",
       "@sanity/client": "6.22.2-bundle-perspective-2",
-      "@sanity/insert-menu": "1.0.11-release.4",
-      "@sanity/presentation": "1.17.8-release.4",
+      "@sanity/core-loader>@sanity/client": "$@sanity/client",
+      "@sanity/presentation>@sanity/client": "$@sanity/client",
```
stipsan, line 190: This is only necessary until #7936 lands

package.json
```diff
@@ -186,8 +186,9 @@
...
+      "@sanity/ui": "2.10.0-corel.0",
```
stipsan, line 191: @bjoerge I can't make the build pass without adding back this 🤔

bjoerge (approved):
lets gooo

PR #7943, opened 2024-12-04
feat(releases): support delete release server action

jordanl17:
### Description
deleteReleasePR
A release may be deleted in only 2 scenarios:
1. The release has already been archived (first flow in demo)
2. The release has already been published (second flow)

NOTE: since changes to show published release documents is still in PR, this means that it isn't possible to demo the case where confirmation is required before deleting. However testing coverage does exist for this case.

### What to review
`ReleaseMenuButton` is where the archive and delete actions exist. Since both have very similar flows - asking for confirmation if release contains documents, a refactor of this component happened to keep things DRY. The existing tests acted as regression to verify.

### Testing
Updated the existing tests for archive to support minor changes that archived releases cannot be archived
Added new test coverage for delete action for archived and published releases

### Notes for release
N/A


packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenuButton.tsx
```diff
@@ -17,127 +18,228 @@ export type ReleaseMenuButtonProps = {
...
+  dialogHeaderI18nKey: string
+  dialogDescriptionSingularI18nKey: string
+  dialogDescriptionMultipleI18nKey: string
+  dialogConfirmButtonI18nKey: string
+  toastSuccessI18nKey: string
+  toastFailureI18nKey: string
+  telemetry: DefinedTelemetryLog<void>
+}
+
+const RELEASE_ACTION_MAP: Record<ReleaseAction, ReleaseActionMap> = {
```
jordanl17, line 34: I expect that unarchive will follow much the same approach as these 2 actions, so this can be easily extended for that

packages/sanity/src/core/releases/store/createReleaseStore.ts
```diff
@@ -46,7 +46,7 @@ const QUERY = `*[${QUERY_FILTER}] ${QUERY_PROJECTION} | ${QUERY_SORT_ORDER}`
 
 const INITIAL_STATE: ReleasesReducerState = {
   releases: new Map(),
-  state: 'loaded' as const,
+  state: 'initialising' as const,
```
RitaDias, line 49: Just want to double check: this is intentional right?

jordanl17: Yes, this is intentional as loading was wrongly coming back as false initially and then immediately to true which caused a few minor loading flickers on the releases list

RitaDias:
Hey Jordan, do you mind just doing a very quick rebase with Corel please? ✨

RitaDias (approved):
✅

PR #7947, opened 2024-12-04
feat(sanity): add unpublish document to release

RitaDias:
### Description

- [x] Add label to release detail table when unpublishing
- [x] Add action on release detail context menu
- [x] Add action to footer on document
- [x] make action disabled when it's already there or there is no published 
- ❗ the change for the numbers in the table for the releases will be done in a separate PR

This action will only be **available if the document has already a published document.**

https://github.com/user-attachments/assets/8849e3b5-7bc2-4c22-a999-f632b060473e

https://github.com/user-attachments/assets/5154dee4-9c8f-4752-8b6f-e8fcd31080c0

### What to review

Naming, code wise etc

### Testing

Tests were added for unpublishing action

### DIFF
--- a/packages/sanity/src/core/releases/hooks/useVersionOperations.tsx
+++ b/packages/sanity/src/core/releases/hooks/useVersionOperations.tsx
@@ -14,12 +14,13 @@ export interface VersionOperationsValue {
     initialValue?: Record<string, unknown>,
   ) => Promise<void>
   discardVersion: (releaseId: string, documentId: string) => Promise<void>
+  unpublishVersion: (documentId: string) => Promise<void>
 }
 
 /** @internal */
 export function useVersionOperations(): VersionOperationsValue {
   const telemetry = useTelemetry()
-  const {createVersion, discardVersion} = useReleaseOperations()
+  const {createVersion, discardVersion, unpublishVersion} = useReleaseOperations()
 
   const {setPerspectiveFromReleaseId} = usePerspective()
   const toast = useToast()
@@ -71,8 +72,34 @@ export function useVersionOperations(): VersionOperationsValue {
       })
     }
   }
+
+  const handleUnpublishVersion = async (documentId: string) => {
+    try {
+      await unpublishVersion(documentId)
+
+      toast.push({
+        closable: true,
+        status: 'success',
+        description: (
+          <Translate
+            t={t}
+            i18nKey={'release.action.unpublish-version.success'}
+            values={{title: document.title as string}}
+          />
+        ),
+      })
+    } catch (err) {
+      toast.push({
+        closable: true,
+        status: 'error',
+        title: t('release.action.unpublish-version.failure'),
+        description: err.message,
+      })
+    }
+  }
   return {
     createVersion: handleCreateVersion,
     discardVersion: handleDiscardVersion,
+    unpublishVersion: handleUnpublishVersion,
   }
 }
--- a/packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
+++ b/packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
@@ -14,8 +14,9 @@ import {
 export const UnpublishVersionAction = (
   props: DocumentActionProps,
 ): DocumentActionDescription | null => {
-  const {id, type, bundleId, version} = props
+  const {id, type, bundleId, version, published} = props
   const currentUser = useCurrentUser()
+  const isPublished = published !== null
 
   const [permissions, isPermissionsLoading] = useDocumentPairPermissions({
     id,
@@ -54,6 +55,7 @@ export const UnpublishVersionAction = (
     label: 'Unpublish',
     icon: UnpublishIcon,
     onHandle: handleDialogOpen,
+    disabled: !isPublished,
     /** @todo translate */
     title: 'Unpublish',
   }
--- a/packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
+++ b/packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
@@ -1,0 +1,60 @@
+import {TrashIcon, UnpublishIcon} from '@sanity/icons'
+import {useCallback, useState} from 'react'
+import {
+  type DocumentActionDescription,
+  type DocumentActionProps,
+  InsufficientPermissionsMessage,
+  useCurrentUser,
+  useDocumentPairPermissions,
+} from 'sanity'
+
+/**
+ * @internal
+ */
+export const UnpublishVersionAction = (
+  props: DocumentActionProps,
+): DocumentActionDescription | null => {
+  const {id, type, bundleId, version} = props
+  const currentUser = useCurrentUser()
+
+  const [permissions, isPermissionsLoading] = useDocumentPairPermissions({
+    id,
+    type,
+    version: bundleId,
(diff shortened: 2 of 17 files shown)


packages/sanity/src/core/releases/hooks/useVersionOperations.tsx
```diff
@@ -71,8 +72,34 @@ export function useVersionOperations(): VersionOperationsValue {
...
+      await unpublishVersion(documentId)
+
+      toast.push({
+        closable: true,
+        status: 'success',
+        description: (
+          <Translate
+            t={t}
+            i18nKey={'release.action.unpublish-version.success'}
+            values={{title: document.title as string}}
```
jordanl17, line 87: NIT: it doesn't look like we need a `title` in this translation string

packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
jordanl17, line 1: @RitaDias just to confirm, the reason why we aren't using a separator and structuring the menu item for unpublish like the prototype is due to the document actions API not currently supporting this layout control?

The prototype has:
Screenshot 2024-12-06 at 10 49 48

packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
```diff
@@ -0,0 +1,74 @@
...
+            onClose={() => setDialogOpen(false)}
+          />
+        ),
+      },
+    /** @todo should be switched once we have the document actions updated */
+    label: t('action.unpublish-doc-actions'),
+    icon: UnpublishIcon,
+    onHandle: handleDialogOpen,
+    disabled: !isPublished || isAlreadyUnpublished,
+    /** @todo should be switched once we have the document actions updated */
```
jordanl17, line 71: Ah, I think this relates to [my comment](https://github.com/sanity-io/sanity/pull/7947/files#r1873048871). Do we have a task todo this somewhere?

packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
RitaDias: Yep correct. 
I thought I had written this soemwhere but yes!

packages/sanity/src/core/releases/plugin/documentActions/UnpublishVersionAction.tsx
RitaDias: Yes! :)
We have a linear story with this and my thoughts and proposal for a solution as well 🙏

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentTableColumnDefs.tsx
```diff
@@ -57,21 +58,39 @@ export const getDocumentTableColumnDefs: (
         <Headers.BasicHeader text={t('table-header.action')} />
       </Flex>
     ),
-    cell: ({cellProps, datum}) => (
-      <Flex align="center" {...cellProps}>
-        <Box paddingX={2}>
-          {datum.document.publishedDocumentExists ? (
+    cell: ({cellProps, datum}) => {
+      const willBeUnpublished = isGoingToUnpublish(datum.document)
+      const BadgeToRender = () => {
```
jordanl17, line 63: NIT: hmm, this approach of using the JSX component syntax to render `<BadgeToRender />` inside a component already irks me a little. Wonder if we could just have this function be called `actionBadge` and then 
```
        <Flex align="center" {...cellProps}>
          <Box paddingX={2}>
            {actionBadge}
          </Box>
        </Flex>
```
Non-blocking though

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
```diff
@@ -67,6 +67,27 @@ export function useBundleDocuments(release: string): {
                 ),
               ),
           ),
+          switchMap((doc) =>
+            observableClient
```
jordanl17, line 71: Q: doesn't the `doc` already have `_system` on it, as the `doc` is the same doc being queried in `release-documents.check-delete`

RitaDias: Hmm might be my lack of knowledge on rxjs but it wasn't showing up when I looked up the doc :/

jordanl17: From experimenting with deleting this new `switchMap` I see that the unpublish action badge still appears correctly, so I think `doc` and `document` are the same therefore. I could be missing a not immediately apparent case where this is needed though...? Worth having a look to see if it's necessary though

RitaDias: Huh 🤔 
I think the main reason why I needed it wasn't for the showing of the label but being able to enable / disable actions.
However let me have a look 👍

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentTableColumnDefs.tsx
```diff
@@ -57,21 +58,39 @@ export const getDocumentTableColumnDefs: (
         <Headers.BasicHeader text={t('table-header.action')} />
       </Flex>
     ),
-    cell: ({cellProps, datum}) => (
-      <Flex align="center" {...cellProps}>
-        <Box paddingX={2}>
-          {datum.document.publishedDocumentExists ? (
+    cell: ({cellProps, datum}) => {
```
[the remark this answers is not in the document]
RitaDias: This makes sense, I don't know how much time I'll have to do this today, however, I will have a look once I've done all other changes and other pending things :)

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentTableColumnDefs.tsx
RitaDias: Fair enough, I'll make the change!

jordanl17:
> * ❗ the change for the numbers in the table for the releases will be done in a separate PR

Yeh I'll do that separately probably sometime next week 🚀

packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
RitaDias: huh colour me shocked, for some reason I really though I'd need this for the documentActions aspects. 🤔 
In any case, thank you for pointing this out 🙏

packages/sanity/src/core/releases/hooks/useVersionOperations.tsx
RitaDias: It probably should, I'll update!

jordanl17 (approved):
Such a simple implementation yet so great now being able to unpublish in a release. Nice parting christmas gift 🎄 !!

packages/sanity/src/core/releases/tool/detail/documentTable/DocumentTableColumnDefs.tsx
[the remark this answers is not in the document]
RitaDias: There we go, thank you for pointing this out! :) 
https://github.com/sanity-io/sanity/pull/7947/commits/a7aea855703e13ec3429c294d8c3f57412f819c8

PR #7950, opened 2024-12-05
fix(sanity): fix issue where you could click the published chip even when the published doc didn't exist

RitaDias:
### Description

What it says on the tin :) 
- fix issue where you could click the published chip even when the published doc didn't exist

See video for a before:
https://github.com/user-attachments/assets/3acf5453-e9b2-4be8-89b3-e4703e127285

### What to review

Does it make sense? Did I miss something?

### Testing

Tests should be enough


packages/sanity/src/structure/panes/document/documentPanel/header/perspective/__tests__/DocumentPerspectiveList.test.tsx
RitaDias, line 1: Not using our mocks since moving them to the test folders would require larger changes that should be done outside the context of this PR

RitaDias:
the failing e2e tests are unrelated but I'm looking at them separately

jordanl17 (approved):
Great fix!

PR #7951, opened 2024-12-05
test(e2e): update unpublish test to match update

RitaDias:
### Description

- Updated the test, the message has changed since the document pane and writing in documents has changed. It not longer shows a message once the unpublish goes through (in the footer)

### Testing

wow, can you believe a e2e test was failing NOT because it was flaky but because it captured a regression??


jordanl17 (approved):
Nice spot

PR #7952, opened 2024-12-05
fix(structure): live edit documents need to be editable in published …

pedrobonamin:
### Description
Fixes a regression in which live edit documents were not editable.
To edit a live document you must have selected the `published` perspective.

| **Draft** (disabled)  | **Published** (enabled)|
|--------|--------|
|Screenshot 2024-12-05 at 12 05 33 | Screenshot 2024-12-05 at 12 05 38|


RitaDias (approved):
An engine 🙏

PR #7955, opened 2024-12-05
feat: Start in Create is now a banner with development mode information

snorrees:
### Description

Removes the "default action" behavior that "Start in Create" uses today, in favor of a dedicated banner for (Create compatible )new documents.

**Making "Start in Create" opt-out will be done in a separate PR**. It will be very small – just flip the default config value (see [previous PR](https://github.com/sanity-io/sanity/pull/7884), since reverted)


packages/sanity/src/core/create/components/StartInCreateDevInfoButton.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return (
+    <Stack space={4}>
+      <Text size={1} weight="semibold">
+        Available studio origin{studioApps.length > 1 ? 's' : ''}
+      </Text>
+      <Stack space={2}>
+        {studioApps.map((app) => {
+          return (
+            <Text key={app.id} size={1}>
+              • {app.studioUrl?.replace(`https://`, '')}
```
snorrees, line 151: • 🙈 
Ideally this was a <ul><li> thing but styling + sanity/ui stuff made it pita. Feel free to keep me honest

packages/sanity/src/core/config/prepareConfig.ts
```diff
@@ -500,16 +499,14 @@ function resolveSource({
...
+        resolveConfigProperty({
           config,
           context: {...context, ...partialContext},
           initialValue: initialDocumentActions,
           propertyName: 'document.actions',
           reducer: documentActionsReducer,
-        })
-        return getStartInCreateSortedActions(actions)
-      },
+        }),
```
snorrees, line 509: Getting rid of this sort hack is honestly good. This feature no longer messes with actions (unless the document is linked with Create, in which case it doesnt render actions menu at all, only the Unlink actions like before)

RitaDias:
I've discussed this with Rob and I've mentioned it to Eoin as well, but just for peace of mind I had two immediate points UX wise that came up when Rob was showing me around. **These are not blockers** :) 

**Regarding the banner at the bottom of a doc:** 
- the larger concern I have is from a UX standpoint. We don’t have anything like that in the studio, however, my fear is that because it is at the end and it disappears after interaction in the document that this feature won’t be used as much not because of its quality but because folks will go into a new document, immediately start writing and not even realise that the banner disappeared (and then have no way of accessing it again)
- It feels like we could easily be shooting ourselves in the foot. I’ve spoken with Rob about different suggestions (banner at the top, using tooltips) but I will leave this to design and the developers (and product)

**Translations on the dev only tooltip:**
- I was asked if it needed to be translated and my immediate feeling is that because it will likely only be seen by devs it /should/ be fine to keep it untranslated. The way I have been seeing the translations has been with the focus being mostly focused on editors which they would not see.
- However, I would like to hear another people’s opinions on this and follow their wisdom on them :)

packages/sanity/src/core/create/studio-app/useStudioAppIdStore.ts
```diff
@@ -58,9 +58,9 @@ export function useStudioAppIdStoreInner(props: {
 
       try {
         const entry = await cache.get({projectId, appIdFetcher})
-        if (mounted) setAppId(entry?.appId)
+        if (mounted) setStudioApp(entry)
```
RitaDias, line 61: This is mostly pure curiosity, but why make this change from sending in the id to sending in the whole object?

packages/sanity/src/core/create/studio-app/checkStudioManifestExists.ts
RitaDias, line 1: Question: particularly if this fails, do we want to show a toast? 
I'm just wondering since an editor might not see that something has gone wrong (though I might be not be seeing the whole path)

packages/sanity/src/core/create/start-in-create/CreateLinkingDialog.tsx
```diff
@@ -25,7 +25,7 @@ export function CreateLinkingDialog() {
   }, [])
 
   return (
-    <Dialog id={id} width={0} header={t('linking-in-progress-dialog.header')}>
+    <Dialog header={t('linking-in-progress-dialog.header')} id={id} width={0}>
```
RitaDias, line 28: 😄

packages/sanity/src/structure/panes/document/document-layout/DocumentLayout.tsx
```diff
@@ -249,6 +253,17 @@ export function DocumentLayout() {
...
+                <ShowWhenPaneOpen>
+                  <StartInCreateBanner
+                    document={value}
+                    documentId={documentId}
+                    documentType={schemaType}
+                    isInitialValueLoading={!!isInitialValueLoading}
+                    panelPortalElementId={DOCUMENT_PANEL_PORTAL_ELEMENT}
+                  />
+                </ShowWhenPaneOpen>
+              )}
```
RitaDias, line 266: Added a comment about the position of this [here](https://github.com/sanity-io/sanity/pull/7955#issuecomment-2523099739)

packages/sanity/src/core/create/components/StartInCreateBanner.tsx
```diff
@@ -0,0 +1,151 @@
...
+  if (
+    !isNewPristineDoc ||
+    !startInCreateEnabled ||
+    isExcludedByOption ||
+    !appIdCache ||
+    !isStartCreateCompatible ||
+    isInitialValueLoading
+  ) {
+    return null
+  }
```
RitaDias, line 49: Smallest nitpick in the world, wonder if it makes more sense to move these out so that the component is not even rendered / called if these things condiitons are met. I'm not sure if I'm just picking on something that has the world's smallest impact on rendering but 😂

RitaDias: It might also be that doing what I suggested adds unnecessary noise to an already large component 😅

packages/sanity/src/core/create/components/DialogPortalProvider.tsx
```diff
@@ -0,0 +1,10 @@
+import {PortalProvider, usePortal} from '@sanity/ui'
+import {type ReactNode} from 'react'
+
+export function DialogPortalProvider(props: {portalElementId: string; children: ReactNode}) {
```
RitaDias, line 4: Huh, could have sworn we had already something like this? 🤔 
Somewhere in the studio? Maybe I'm just misremembering and what I remember is from your previous work 😅

RitaDias (commented):
I've had a good look at it :) 
Thanks for doing this!

Overall I see no blockers apart from the questions I've risen! I will be tagging someone else in the app team to have a look at this for peace of mind 🙏

snorrees:
> **Translations on the dev only tooltip:**

Personally I would argue for keeping as it (no translations), since its a lower maintainence cost (imo). The audience is devs after all, and we dont translate our code right?

packages/sanity/src/core/create/studio-app/useStudioAppIdStore.ts
snorrees: The underlying api now returns studioApps as well, since we need it in dev mode. Before this was just an object with a single property, now it has two.

RitaDias:
> > **Translations on the dev only tooltip:**
> 
> Personally I would argue for keeping as it (no translations), since its a lower maintainence cost (imo). The audience is devs after all, and we dont translate our code right?

_yet_
DUN DUN DUUUUN

packages/sanity/src/core/create/components/StartInCreateBanner.tsx
snorrees: We need to run the hook to get config, so I think its nice to have all the null checks in a single place. The intermediary component is needed to able to conditionally render based when appId is resolved.

I cannot see why the code as written would be a performance issue, but keep me honest here.

RitaDias: I just thought (and maybe I'm wrong), that by having these ifs within the component that the component would _still render_ or at least take time to try before realising that it doesn't need to. But again, that might be a bastardisation of what react actually does under the hood 😅

snorrees: With this early return there is very little compute being used; the inner components are never rendered.

packages/sanity/src/core/create/start-in-create/CreateLinkingDialog.tsx
robinpyon: Apols for the noise, this is a terrible habit of mine (ordering props alphabetically) after the fact

packages/sanity/src/core/create/components/DialogPortalProvider.tsx
robinpyon: @RitaDias You may be remembering this component from a previous iteration. The only change is that this is now exported as it's used within multiple components (in this integration at least)

packages/sanity/src/core/create/__tests__/createUtils.test.ts
```diff
@@ -69,6 +69,36 @@ describe('createUtils', () => {
       expect(isSanityCreateExcludedType(documentType)).toEqual(false)
     })
   })
+
+  describe('isSanityCreateStartCompatibleDoc', () => {
```
bjoerge, line 73: what constitutes a sanity create compatible document?

packages/sanity/src/core/create/context/SanityCreateConfigProvider.tsx
```diff
@@ -16,16 +18,21 @@ interface SanityCreateConfigProviderProps {
 export function SanityCreateConfigProvider(props: SanityCreateConfigProviderProps): JSX.Element {
   const {children} = props
   const {beta} = useSource()
+
+  const appIdCache = useMemo(() => createAppIdCache(), [])
```
bjoerge, line 22: where can I learn more about what we mean by app id?

dev/test-create-integration-studio/schema.ts
```diff
@@ -40,11 +64,15 @@ export const schemaTypes = [
       }),
       defineField({
         name: 'description',
-        title: 'Description',
+        title: 'Description – with initial value – Create excluded',
         type: 'text',
         rows: 3,
         description: 'Lede and page summary.',
         group: [mainGroup.name],
+        /*    initialValue: 'Predefined description',
```
bjoerge, line 72: this doesn't seem like it should be commented out?

packages/sanity/src/core/create/studio-app/__tests__/useStudioAppIdStore.test.ts
```diff
@@ -12,19 +12,25 @@ describe('useStudioAppIdStore', () => {
...
+        }),
       } satisfies Parameters<typeof useStudioAppIdStoreInner>[0],
     })
 
     expect(result.current).toEqual({
       loading: true,
-      appId: undefined,
-    })
+      studioApp: undefined,
+    } satisfies ResolvedStudioApp)
```
bjoerge, line 25: neat use of satisfies!

packages/sanity/src/core/create/studio-app/appIdCache.ts
```diff
@@ -1,35 +1,35 @@
...
+    [key: string]: CompatibleStudioAppId | Promise<CompatibleStudioAppId | undefined> | undefined
+  } = {}
 
   return {
     get: async (args) => {
       const {projectId, appIdFetcher} = args
-      let cacheElement = cache[projectId]
+      let cacheElement = appIdCache[projectId]
       if (!cacheElement) {
         cacheElement = (async () => {
```
bjoerge, line 21: Multiple concurrent calls to `get()` with a cold cache will fire off multiple requests in between checking whether the cache entry exists and waiting for the result. This should be updated to instead cache the promise rather than the resolved value.

packages/sanity/src/core/create/studio-app/fetchCreateCompatibleAppId.ts
```diff
@@ -11,51 +10,82 @@ export interface StudioAppResponse {
   appHost: string
 }
 
+export interface StudioApp extends StudioAppResponse {
+  studioUrl: string
+}
+
+export interface CompatibleStudioAppId {
```
bjoerge, line 17: Could we get a comment providing some high level description what these represents? I.e. it's unclear to me what the difference is between `appId` and `studioApps[].id` - do they represent the same thing?

bjoerge (changes requested):
Looks good overall. Left a couple of questions about things that was unclear to me when reviewing, would be ideal to have some clarifying code comments about these.

Only blocker for is what I believe is a race condition in the appIdCache.

packages/sanity/src/core/create/__tests__/createUtils.test.ts
bjoerge: (nvm, I meant to delete this before submitting review, found answer to this later on)

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -679,6 +681,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
       ready,
       schemaType: schemaType!,
       isPermissionsLoading,
+      isInitialValueLoading,
```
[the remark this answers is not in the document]
bjoerge: No concerns :)

packages/sanity/src/core/create/context/SanityCreateConfigProvider.tsx
robinpyon: An "application" in this context refers to a customized Studio deployment defined in _Manage_. 

The term _application_ is definitely a loaded one, though these are what are returned when querying the `user-applications` endpoint on the Project API, e.g.

```
// https://ppsg7ml5.api.sanity.io/v2024-12-09/projects/ppsg7ml5/user-applications

[
  {
    "id": "f6df967b6b9add588388483c", // application ID
    "projectId": "ppsg7ml5",
    "title": null,
    "type": "studio",
    "urlType": "external",
    "appHost": "https://test-studio.sanity.build/",
    "createdAt": "2024-08-09T14:04:55.954Z",
    "updatedAt": "2024-08-09T14:04:55.954Z",
    "activeDeployment": null
  },
 ...
]
```

Would `userAppIdCache` be more apt?

robinpyon: (As of writing, the `user-applications` endpoint is not yet documented)

packages/sanity/src/core/create/studio-app/fetchCreateCompatibleAppId.ts
robinpyon: > it's unclear to me what the difference is between appId and studioApps[].id - do they represent the same thing?

They do! As mentioned in an above comment, _studios_ defined in manage are technically named _applications_ , which can definitely be a source of confusion here.

Would it be clearer for you if `appId` was titled `studioAppId`?

bjoerge: My confusion was mostly around the fact that there's an CompatibleStudio*AppId* type that has an `appId` property and also a list of studioApps, each with their own id. Am I understanding it correctly that `appId` here reflects which one of the Studio apps in `studioApps` which matches the current origin?

bjoerge (changes requested):
There are unit tests currently failing: https://github.com/sanity-io/sanity/actions/runs/12236133730/job/34128922171?pr=7955#step:8:258

packages/sanity/src/core/create/studio-app/fetchCreateCompatibleAppId.ts
bjoerge: Maybe `currentAppId` would be more clear then? Note: consider this as feedback from fresh eyes and not a blocker for merge, I'll leave it to you to judge whether this is something you want to address now or later (or never :))

packages/sanity/src/core/create/studio-app/appIdCache.ts
robinpyon: Thanks @bjoerge – just confirming what we talked about this separately (no longer writing the stored value)

bjoerge (approved):

PR #7958, opened 2024-12-05
fix(core): add telemetry debug/noop logger

bjoerge:
### Description

A little quality of life addition to make it easier for devs to introspect the telemetry payloads that will be logged from the studio. This is meant primarily for in house sanity devs, but also doubles as a way to easily see what info we log.

### What to review

Does it make sense?

### Testing
Feels like a trivial enough change to ski include a test for this particular case.

### Notes for release

n/a for internal use.


ricokahler (approved):
nice!

PR #7960, opened 2024-12-05
fix: reword warning about vector images not support crop/hotspot

rexxars:
### Description

I know this went through a review already, but I still have some issues with the current wording:
- "Bitmap" is less precise than "raster" in my opinion. Whether or not the raster format uses a bitmap internally is irrelevant.
- "SVG" is too specific - we might support other vector formats in the future, which will have the same restrictions.
- The i18n strings should preferably describe the _purpose_ of the string, not the contents per se. This helps the AI translator understand it better. There was also a typo in the comment (`fp` instead of `fm`).

### What to review

New messages/keys looks accurate/correct

### Testing

Same as before, only strings changed.

### Notes for release

None


bjoerge (approved):
Thank you for these clarifications

PR #7968, opened 2024-12-06
fix(sanity): issue with the publish chip disable + update tests

RitaDias:
### Description

Fix issue where the publish disable wasn't working as expected.
I have added tests to match all the use cases that I could think of for when it is disabled :)

### What to review

Does it make sense? could I have done it in a different better way?

### Testing

The tests should be good enough but just to be sure go to the preview and try with documents **that don't** have published documents (from an existing version seeing if the chip is enabled when you know there is no published doc) and when there is.


bjoerge (approved):
Solid like a rock! Thanks Rita!

PR #7977, opened 2024-12-06
fix(deps): deprecation warnings due to `glob` dependency

rexxars:
### Description

The `glob` and `rimraf` dependencies yields a number of deprecation warnings on install, due to older versions being used:

```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```

The `inflight` one is a dependency of `glob` in v7 - getting rid of that is the highest priority as the message can look unfriendly/scary to people. `globby` uses `glob` in v10, but not in v11 (higher versions switches to ESM, thus not as easy of an upgrade). 

Part of SAPP-1961
Part of SAPP-1690

### What to review

Stuff still works! 

### Testing

Existing tests should hopefully be enough.

### Notes for release

- Fixes a few deprecation warnings when installing dependencies for a studio


bjoerge (approved):
Thank you for jumping on this @rexxars <3

PR #7980, opened 2024-12-09
feat(manifest): add icons, tools & projectId

joshuaellis:
### Description

Building on the work in #7403, the Content OS requires more information from the studio. Namely icons & tool information but ideally the projectId as well (although we could add this server side if this was preferable).

* Un-deprecates `Tool['icon']` but it is still considered optional.
* Serialises tool information in separate file similar to how schemas are treated.
* Serialises icons in a "generic" manor similar to visual editing, we expect this to cover most use-cases however we catch it if the serialisation fails and returns null as this part of the process should not cause overall failure.
* Adds internal property to tools `__internalApplicationType` so we can group multiple across workspaces with the future ambitions these become applications. This is undocumented and will not impact plugin authors.

To avoid breaking change issues with create, even though the manifest is now used for more generic purposes we've kept the same naming conventions.

#### Related issues (and context)

* resolves ENTX-2507
* resolves ENTX-2572

### What to review

- [x] Should we set the version to 1.1? 
- [x] Is there an issue with including the `projectId` in the manifest files?
- [x] Is there concern over un-deprecating icons in tools?
- [x] Is there a foreseen issue adding `type` to tools?

#### Manual testing
In a project you can deploy using sanity deploy:

* npm i sanity@content-os
* npm run deploy (or npm run build && npm run start to run on localhost)
* Visit <studioHost>/static/create-manifest.json
* Visit a tools manifest (<studioHost>/static/<schema-path-found-in-workspace> on the form <schema-hash>.create-tools.json)

#### Known issues

* we're limited by the same restrictions noted in #7403. For information on the future of the Content OS x Manifest see this [RFC](https://www.notion.so/sanityio/RFC-Rendering-workspaces-14b78d055b9480988521c02984dc48c0).

### Notes for release

Not required.


joshuaellis:
NOTE: if we approve this PR, I will make the necessary follow ups to tools that don't exist in this monorepo e.g. `media`.

snorrees (changes requested):
Have not manually tested yet, will come back for that:

> Should we set the version to 1.1?

Design flaw one: it is hardcode to `version: 1`. 

Do we just break the format right away and make it `version: string`? Then you set '1.1' when writing the file? Or do we continue the design as is and hardcode `version: 2` possibly `version: 1.1`, but I think the former fits better if we go with `number`.

Version is not used anywhere atm, so open to suggestions. Maybe just increment the number when we change things, ie 2, would be my gut feeling atm.

> Is there an issue with including the projectId in the manifest files?

I dont see any problem with this. Its not a secret. Lets not discuss the org/project paradigm here (in public).

> Is there concern over un-deprecating icons in tools?

I'll let Studio team answer this one. It could be seen as a bit odd to have a field in the config-api that is not used by Studio itself, so you could argue it should be nested under a new options property or something.

packages/sanity/src/core/config/types.ts
```diff
@@ -175,6 +173,11 @@ export interface Tool<Options = any> {
     params: Record<string, unknown>,
     payload: unknown,
   ) => boolean | {[key: string]: boolean}
+
+  /**
+   * The type of tool – used to group tools across multiple workspaces and have a constant identifier.
+   */
+  type?: string
```
snorrees, line 180: > Is there a foreseen issue adding type to tools?

It says "used to group tools", but the Studio doesnt do that. It is not clear to me how we want to design types/apis for studio types that are not used by studio but only by <redacted>.

packages/sanity/src/_internal/manifest/Icon.tsx
```diff
@@ -0,0 +1,104 @@
...
+    .filter(Boolean)
+    // take the first two words
+    .slice(0, 2)
+    // grab the first letter and make it upper case
+    .map((i) => i.charAt(0).toUpperCase())
+
+  const darkened = darken(color, 0.4)
+  const lightened = lighten(color, 0.4)
+
+  /* eslint-disable no-negated-condition */
```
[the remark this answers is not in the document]
joshuaellis: I can do, I copied from https://github.com/sanity-io/sanity/blob/feat/manifest-additions/packages/sanity/src/core/config/createDefaultIcon.tsx because it wouldn't let me import it here due to "boundary crossing" is there an easier way to share code? I don't mind tweaking it & adding tests etc.

packages/sanity/src/_internal/manifest/extractWorkspaceManifest.ts
```diff
@@ -70,14 +74,22 @@ const INLINE_TYPES = ['document', 'object', 'image', 'file']
 
 export function extractCreateWorkspaceManifest(workspace: Workspace): CreateWorkspaceManifest {
   const serializedSchema = extractManifestSchemaTypes(workspace.schema)
+  const serialisedTools = extractManifestTools(workspace.tools)
```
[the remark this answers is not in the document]
joshuaellis: no you're right 😄

joshuaellis:
> Version is not used anywhere atm, so open to suggestions. Maybe just increment the number when we change things, ie 2, would be my gut feeling atm.

Let's set to version 2, that way I can handle if I receive a version 1 manifest which would be lacking in data.

packages/sanity/src/_internal/cli/actions/manifest/extractManifestAction.ts
```diff
@@ -90,7 +91,7 @@ async function extractManifest(
     const workspaceFiles = await writeWorkspaceFiles(workspaceManifests, staticPath)
 
     const manifest: CreateManifest = {
-      version: 1,
+      version: 2,
```
snorrees, line 94: Lets add a commnet: `//Increment this if whenever we change the API`

Q: should we also add a comment about version history? Ie:
```
/* 
Version history:
1: Initial release.
2: Added tools file.
*/
```

packages/sanity/src/_internal/manifest/__tests__/Icon.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+import {describe, expect, it} from 'vitest'
+
+import {SchemaIcon, type SchemaIconProps} from '../Icon'
+
+const render = (props?: Partial<SchemaIconProps>) =>
+  renderRTL(<SchemaIcon title="Studio" {...props} />, {
+    wrapper: ({children}) => <ThemeProvider theme={studioTheme}>{children}</ThemeProvider>,
+  })
+
+describe('SchemIcon', () => {
```
snorrees, line 12: I didnt realize the code was copypasta, so feel free to ignore.
The things I personally would like to see testing is the colors and title stuff, not the jsx. Maybe what I really wanted was the hard bits to be function extracted and tested in isolation. 
But if you keep the tests, check for the expected fill colors and stuff?

Again, no need to humor me.

packages/sanity/src/_internal/manifest/Icon.tsx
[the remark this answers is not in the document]
snorrees: I'll defer to Studio team on what to do here: Seems to me that it should be extracted and reused both places.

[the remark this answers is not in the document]
joshuaellis: i'd prefer that, lets hold fire till they respond 🫡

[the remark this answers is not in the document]
bjoerge: You can export it from the main `sanity` export and mark it as internal/hidden

packages/sanity/src/_internal/manifest/manifestTypes.ts
```diff
@@ -6,15 +6,14 @@ export type ManifestSerializable =
...
   createdAt: string
   workspaces: ManifestWorkspaceFile[]
 }
 
-export interface ManifestWorkspaceFile {
-  name: string
-  dataset: string
+export interface ManifestWorkspaceFile extends Omit<CreateWorkspaceManifest, 'schema' | 'tools'> {
   schema: string // filename
+  tools: string // filename
```
bjoerge, line 16: what does the filename comment refer to here?

packages/sanity/src/core/config/types.ts
bjoerge: The type concept is a bit fuzzy to me as well. Probably missing some context. If it's for grouping, could it be a way for the tool to declare what group it belongs to, e.g. by `group: 'edit'`? If it's for identifying it, should the property be named "id" instead? Feels like it should have been the "name" property.

Is the idea that the consuming end doesn't need to know what types of tools exists; i.e. it just groups the ones with similar types? Or could the consuming aware of what types exists and will do the grouping based on that?

- How do we guide plugin authors to chose type/group/identifier?
- Are the of which tools are appearing in the other end ever going to be important?
- How do we avoid name clashes/unintentionally grouping ?

packages/sanity/src/_internal/manifest/__tests__/Icon.test.tsx
bjoerge, line 12: typo, should be SchemaIcon?

packages/@sanity/vision/src/visionTool.ts
```diff
@@ -14,6 +14,7 @@ export const visionTool = definePlugin<VisionToolConfig | void>((options) => {
       {
         name: name || 'vision',
         title: title || 'Vision',
+        type: 'sanity/vision',
```
bjoerge, line 17: (se later comments/questions about type)

packages/sanity/src/_internal/manifest/manifestTypes.ts
snorrees: The string is a filename

packages/sanity/src/core/config/types.ts
joshuaellis: For future people – We've discussed this internally, it's use case is primarily to manage sanity tools we expect to turn into applications in the future, we don't really need / want third party developers to be concerned or confused with this. As such we've removed from the type signature & underscored it so it's clearly undocumented & internal to work with for now.

packages/sanity/src/_internal/manifest/extractWorkspaceManifest.ts
```diff
@@ -500,3 +515,48 @@ function resolveTitleValueArray(possibleArray: unknown): ManifestTitledValue[] |
...
+
+const resolveIcon = (props: SchemaIconProps): string | null => {
+  const sheet = new ServerStyleSheet()
+
+  try {
+    /**
+     * You must render the element first so
+     * the style-sheet above can be populated
+     */
+    const element = renderToString(createElement(SchemaIcon, {...props, sheet}))
```
bjoerge, line 546: Should the createElement return value here not be wrapped in `sheet.collectStyles()`?

joshuaellis: it doesn't have to because we're using <StyleSheetManager> but if i'm honest, it would be nicer to just do that.

bjoerge: another minor thing, probably doesn't matter much, but could we technically have used `renderToStaticMarkup`?

bjoerge (approved):
LGTM 🚀

PR #7995, opened 2024-12-09
fix(deps): upgrade `vite` to v6

rexxars:
### Description

Upgrades vite to the latest version (v6), as well as vitest to v3

### What to review

Give it a test, see how it feels! Does anything break?
There's a prerelease you can use to test it on a local studio, under the `upgrade-vite-v6` branch on npm.

### Testing

Commands involved:
- `sanity dev`
- `sanity build`
- `sanity preview`

### Notes for release

- Upgrades vite to version 6


bjoerge (dismissed):
Thank you for this, @rexxars! <3

PR #7999, opened 2024-12-09
fix(sanity): optimise getLeafWeights to not stack overflow

nikaspran:
### Description

When working on schemas that branch out a lot (i.e. have arrays that can hold many different types), the current `getLeafWeights` implementation throws with a `RangeError: Maximum call stack size exceeded` after a while.

This PR optimises the method to use an accumulator instead, which keeps the stack usage flat and no longer overflows for the same amount of data. I have also added a test to verify this (it used to explode with the previous version, see screenshots).

In addition, this PR optimises the method to be much more efficient by:
- Merging the loops for `objectTypes` and `arrayTypes` into one
- Adding a few extra tiny optimisations

As a result, this should be much faster and no longer explode the stack.

#### Before
Screenshot 2024-12-09 at 18 26 37

#### After
Screenshot 2024-12-09 at 22 56 15

### What to review

* We originally ran into this when rendering a document list once the schema got "branchy" enough, for a lack of a better word

### Testing

* Have added a unit test that would fail under the previous implementation but passes current one
* This should retain exactly the same functionality as before, just be faster and more memory conscious

### Notes for release

Fixes a crash when rendering a document list within Studio given a schema that branches out a lot

### DIFF
--- a/packages/sanity/src/core/search/common/__tests__/deriveSearchWeightsFromType.test.ts
+++ b/packages/sanity/src/core/search/common/__tests__/deriveSearchWeightsFromType.test.ts
@@ -433,4 +433,50 @@ describe('deriveSearchWeightsFromType', () => {
       ],
     })
   })
+
+  it('works for schemas that branch out a lot', () => {
+    // schema of 60 "components" with 10 fields each
+    const range = [...Array(60).keys()]
+
+    const componentRefs = range.map((index) => ({type: `component_${index}`}))
+    const components = range.map((index) =>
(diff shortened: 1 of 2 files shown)


nikaspran:
I can confirm this also fixes our actual issue my company's setup, so would love to get this in and unblock 🙏 Happy to make any changes if necessary.

juice49 (approved):
This is fantastic! Thank you so much for the contribution, @nikaspran.

PR #8000, opened 2024-12-10
fix(structure): enable published chip for live edit documents

pedrobonamin:
### Description
**Enables the **published chip** for `liveEdit` documents.**

Live edit documents are only editable through the published perspective, we need to enable it even if the published document doesn't exists.
For the rest of documents, validate if the `published` document exists.

**Update actions data-testid**
action tests were failing because the test id was not found, this updates them to match what we have in `next` .https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/panes/document/statusBar/DocumentStatusBarActions.tsx#L62

### DIFF
--- a/packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveList.tsx
+++ b/packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveList.tsx
@@ -14,7 +14,6 @@ import {
   VersionChip,
   versionDocumentExists,
 } from 'sanity'
-import {usePaneRouter} from 'sanity/structure'
 
 import {useDocumentPane} from '../../../useDocumentPane'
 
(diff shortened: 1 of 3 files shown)


packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveList.tsx
```diff
@@ -101,18 +99,12 @@ export const DocumentPerspectiveList = memo(function DocumentPerspectiveList() {
   )
 
   const isPublishedChipDisabled = useMemo(() => {
-    if (editState?.liveEdit) {
-      if (!editState?.published) {
-        return true
-      }
+    // If it's a live edit document the only option to edit it is through
```
juice49, line 102: Nice simplification 🙌

juice49 (approved):
Looks good to me!

PR #8008, opened 2024-12-10
feat(releases): reverting releases

jordanl17:
### Description
https://www.loom.com/share/6eaf274374ba4bb78b8f381bfd8e9fbf?sid=4610ffad-3561-4cdf-82ab-cc76edb72690

### What to review
* There's a new Release CTA `ReleaseRevertButton`
* The dialog rendered here will do 2 things
  * check if there are any version documents in the release which have transactions after the publish transaction
  * Generate all document states the revert to
* Generating document states is optimistically started - when the confirm dialog opens then this process starts. If the states are resolved before the confirm button is pressed, then these states will immediately be used to create new version documents. If these states are still pending, then the creation of the revert release will await completion before proceeding.
   

### Testing
Added tests for the new hooks to:
* determine if there is further history in any document after the original release publish
* fetch the document states to revert to

Tests also for `createReleaseOperationsStore` where the `revertRelease` operation exists.

Will fast-follow tests for `ReleaseRevertButton`

### Notes for release
N/A


packages/sanity/src/core/releases/__telemetry__/releases.telemetry.ts
```diff
@@ -101,3 +108,12 @@ export const UnarchivedRelease = defineEvent({
   version: 1,
   description: 'User unarchived a release',
 })
+
+/** When a release is successfully reverted
+ * @internal
+ */
+export const RevertRelease = defineEvent<RevertInfo>({
+  name: 'Revert release',
```
bjoerge, line 116: Telemetry events should be Title Cased (and are usially written in past tense I believe)

From our guidelines:
> Should be in the form of <Object> <Action>, e.g. Document Published, Studio Entered, or Trial Started, and be capitalized.

```suggestion
  name: 'Release Reverted',
```

I see that we lack a bit of consistency here already, so feel free to create a ticket and leave it as a separate task

packages/sanity/src/core/releases/store/__tests__/createReleaseOperationsStore.test.ts
bjoerge, line 1: Thank you for adding these, much appreciated <3

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -159,6 +167,35 @@ export function createReleaseOperationsStore(options: {
...
+  ) => {
+    await handleCreateRelease({
+      _id: revertReleaseId,
+      metadata: {
+        title: releaseMetadata.title,
+        description: releaseMetadata.description,
+        releaseType: 'asap',
+      },
+    })
+    await Promise.allSettled(
```
bjoerge, line 184: Nice use of `allSettled`! I do wonder if we want to surface it somehow if any of these fails, though. Also think there should be some concurrency management here we avoid hundreds of simultaneous requests. Although this might be just a stop-gap implementation for now? If so, then all good.

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -159,6 +167,35 @@ export function createReleaseOperationsStore(options: {
       },
     ])
 
+  const handleRevertRelease = async (
+    revertReleaseId: string,
+    releaseDocuments: RevertDocument[],
+    releaseMetadata: ReleaseDocument['metadata'],
+    revertType: 'staged' | 'immediately',
```
bjoerge, line 174: see earlier comment re: immediate vs immediately

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
```diff
@@ -0,0 +1,135 @@
...
+import {catchError, filter, forkJoin, from, map, type Observable, of, switchMap} from 'rxjs'
+
+import {useClient} from '../../../../../hooks/useClient'
+import {getTransactionsLogs} from '../../../../../store/translog/getTransactionLogs'
+import {API_VERSION} from '../../../../../tasks/constants'
+import {type DocumentInRelease} from '../../../detail/useBundleDocuments'
+
+export type RevertDocument = SanityDocument & {
+  _system?: {
+    delete: boolean
```
bjoerge, line 13: Documents with a `_system.delete` property would never have `delete: false`, right? So this could be typed more narrowly to `delete: true`, or?

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
```diff
@@ -0,0 +1,135 @@
...
+  const resultPromiseRef = useRef<Promise<DocumentRevertStates> | null>(null)
+  const resolvedDocumentRevertStatesPromiseRef = useRef<
+    ((value: DocumentRevertStates) => void) | null
+  >(null)
+  const resolvedDocumentRevertStatesResultRef = useRef<DocumentRevertStates | null>(null)
+
+  useEffect(() => {
+    if (!resultPromiseRef.current) {
+      resultPromiseRef.current = new Promise((resolve) => {
+        resolvedDocumentRevertStatesPromiseRef.current = resolve
```
bjoerge, line 36: This looks like a good case for [Promise.withResolvers()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export type RevertDocument = SanityDocument & {
+  _system?: {
+    delete: boolean
+  }
+}
+
+type RevertDocuments = RevertDocument[]
+
+type DocumentRevertStates = RevertDocuments | null | undefined
```
bjoerge, line 19: Does `null` have a special meaning? If so, would be great to have a short comment in code here explaining what it means, if not, I tend to prefer sticking with just `undefined`

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const useDocumentRevertStates = (documents: DocumentInRelease[]) => {
```
bjoerge, line 21: Could I have some TSDoc with a short explanation of what this does? Struggled a little bit with unpacking the implementation 😅

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/usePostPublishTransactions.ts
```diff
@@ -0,0 +1,37 @@
...
+          // publish transaction + at least one post publish transaction
+          limit: 2,
+        },
+      ),
+    ).pipe(
+      filter(Boolean),
+      // the transaction of published is also returned
+      // so post publish transactions will result in more than 1 transaction
+      map((transactions) => transactions.length > 1),
+      catchError(() => of(null)),
```
bjoerge, line 32: Are we sure we want to silently ignore any error here? Could you add a short commend telling the next reader why it's safe/ok if so?

To me it seems like any failure here could lead us falsely assume there is no transactions, which might leave the end-user with a wrong impression

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
bjoerge: On a second readthrough, I'm wondering if there is a reason this has to be a hook – seems like a better candidate for a regular async function to me, but maybe I'm missing something?

bjoerge (commented):
Great work, @jordanl17! I've left a few comments/questions for your consideration. Feel free to unblock yourself by merging and address my feedback whenever convenient.

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/ReleaseRevertButton.tsx
```diff
@@ -0,0 +1,238 @@
...
+  release,
+}: {
+  revertReleaseStatus: RevertReleaseStatus
+  documents: DocumentInRelease[]
+  setRevertReleaseStatus: (status: RevertReleaseStatus) => void
+  release: ReleaseDocument
+}) => {
+  const {t} = useTranslation(releasesLocaleNamespace)
+  const hasPostPublishTransactions = usePostPublishTransactions(documents)
+  const getAdjacentTransactions = useDocumentRevertStates(documents)
```
jordanl17, line 41:
```suggestion
  const getDocumentRevertStates = useDocumentRevertStates(documents)
```

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/useDocumentRevertStates.ts
jordanl17: Good idea!

packages/sanity/src/core/releases/__telemetry__/releases.telemetry.ts
jordanl17: Good call - yeh we aren't following these guidelines for the release telemetry. I've created a separate task to switch all of these to be correct past tense

jordanl17:
> Great work, @jordanl17! I've left a few comments/questions for your consideration. Feel free to unblock yourself by merging and address my feedback whenever convenient.

Thanks, I've merged this as it feels 'demo-able' enough. But will create a separate PR for the valid concerns around surfacing up appropriate errors and considering if `useDocumentRevertStates` can just be async. Will link this PR in the next one I raise shortly

PR #8015, opened 2024-12-11
feat(core): corel  publish 2 action - test

pedrobonamin:
### Description

Adds a second publish action using the new action `sanity.action.release.publish2`. which is being developed.

To enable it, you need to add `publish2: true` in local storage

Screenshot 2024-12-11 at 17 12 58


bjoerge (approved):
Very nice! Thank you @pedrobonamin

PR #8017, opened 2024-12-11
fix(deps): upgrade react-json-inspector

rexxars:
### Description

Upgrades `@rexxars/react-json-inspector` to the latest version, which is compatible with React 19 and drops lifecycle events that are deprecated. The conversion was a bit quick and dirty, but from what I can tell things seems to work fine.

### What to review

"Inspect document" and vision results works as expected

### Testing

Check out the preview deployment, see that it matches current look/behavior

### Notes for release

None


stipsan:
This is awesome 💖

juice49 (approved):
Looks good to me! 🙏

PR #8020, opened 2024-12-12
fix: delays rendering the Start in Create banner until document is ready

snorrees:
### Description

Currently, the Start in Create banner will appear when a document pane is loading:
Banner Flashes v3 67 0

With this fix, the banner should no longer appear when the document pane is in a loading state.

#### Background
Got a report that the Start in Create banner was breifly flashing (appearing and disappearing for < 500ms) on localhost.
I was unable to reproduce _that_ issue, but noticed that when running on 3G Network, the Create banner will render whilst the document is loading.

This PR delays rendering the banner until the document is ready. Opted for prop-drilling to get the ready state, as not to add any new useDocumentPane calls.

The ready check is done in the wrapper component where appId checks runs; this allows the initial requests to run in parallel with document load, so the data is ready as soon as possible.

### What to review

The code ;)

### Testing
 
* Go to https://create-integration-test.sanity.studio/fallback/structure/create-test-article
* Open up Network tab and put it into 3G mode.
* Create a new document. 
* While the pane is loading there should be no Create banner
* The banner should appear with the document (or shortly thereafter, on first document opened in a reloaded studio after which appId data is cached)

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/create/components/StartInCreateBanner.tsx
+++ b/packages/sanity/src/core/create/components/StartInCreateBanner.tsx
@@ -55,7 +55,8 @@ export function StartInCreateBanner(props: StartInCreateBannerProps) {
 function StartInCreateBannerInner(props: StartInCreateBannerProps & {appIdCache: AppIdCache}) {
   const {studioApp} = useStudioAppIdStore(props.appIdCache)
 
-  if (!studioApp) {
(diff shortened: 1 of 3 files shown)


pedrobonamin (approved):
Looks good to me, thanks!

PR #8021, opened 2024-12-12
fix(core): fix issues with PT-Input PopoverModal

skogsmaskin:
### Description

* Remove the workaround for the @sanity/ui issue (which has been fixed) with the initial rendering in the wrong position. This is no longer needed and can cause problems with assertions in tests.

* Implement focus locking inside the modal. The PopoverModal should have the same kind of focus locking as other dialogs

### What to review

That popover modals for links inside the Portable Text Input traps focus and work as expected.

The popover modal position is correct when first opened.

### Testing

Open a Portable Text Editor with a link schema type and do the review steps above.

### Notes for release
* Keep tab focus inside Popover Dialogs for the Portable Text Input when editing links and similar.

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
@@ -3,6 +3,7 @@
 import {CloseIcon} from '@sanity/icons'
 import {Box, Flex, Text, useClickOutsideEvent, useGlobalKeyDown} from '@sanity/ui'
 import {type ReactNode, useCallback, useRef, useState} from 'react'
+import FocusLock from 'react-focus-lock'
 import {type PortableTextEditorElement} from 'sanity/_singletons'
 
 import {Button, type PopoverProps} from '../../../../../../ui-components'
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
```diff
@@ -80,42 +78,44 @@ function Content(props: PopoverEditDialogProps) {
   const [contentElement, setContentElement] = useState<HTMLDivElement | null>(null)
   const containerElement = useRef<HTMLDivElement | null>(null)
 
-  useEffect(() => {
-    // When rendered, focus on the first input element in the content
-    if (contentElement) {
-      contentElement.querySelector('input')?.focus()
-    }
-  }, [contentElement])
```
skogsmaskin, line 88: This should rather be handled by the autoFocus prop so that it works like the other Modals we use (where the first focusable element is focused, usually the close-button).

packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
```diff
@@ -80,42 +78,44 @@ function Content(props: PopoverEditDialogProps) {
...
+  const handleFocusLockWhiteList = useCallback((element: HTMLElement) => {
```
juice49, line 81: Language nit: we should use a term like `allowList` instead.

packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
juice49: That works in my testing 👍.

juice49 (approved):
Looks good to me 🙌. I tried this in Test Studio, and it worked as you described. I added one small comment, but it's not a blocker.

packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
skogsmaskin: > allowList

The dependency library uses this term, so it's supposed to match that.

PR #8022, opened 2024-12-12
fix(core): fix issue with different Webkit tab order in forms

skogsmaskin:
### Description

Chromium and Firefox will tab to the FieldActions menu without tabIndex=0, but Webkit needs to have it explicitly set to include it in the tab flow. Add it so all browsers follow the same tab order, and WebKit users can access the field actions with the keyboard.

This will also prevent issues with Playwright tests that rely on tabbing through forms.

### What to review

That WebKit can tab into the field actions menus.

### Testing

Test any form with multiple fields.

### Notes for release
* Fixed an issue where WebKit was unable to tab into a form field's action menu.


pedrobonamin (approved):
Thanks for this @skogsmaskin

PR #8024, opened 2024-12-12
fix(core): release events types are lowercase

pedrobonamin:
### Description
Content lake has renamed the events types that are returned from the events api to camel case.
This reflects that update in the studio.


pedrobonamin:
I'm merging this one, I think it doesn't need a peer review it's just a rename of types that came from content lake, and it's going to a feature branch.

PR #8032, opened 2024-12-13
chore: add `dev:million-lint` command

stipsan:
### Description

This PR sets up a script that run [Million Lint](https://million.dev) on the Test Studio locally:

https://github.com/user-attachments/assets/021e0530-c3ff-4e76-b054-72274332840e

There's a lot to say on this topic, and if performance work interests you and you have a hot beverage ready, expand the `Verbose information` toggle at the end of the PR. The PR will otherwise be as brief and easy to review as possible 🙌 

With Million Lint we can:
- Find good ways to manually memoize components that React Compiler skips
  - The Compiler is in beta, so there are some React patterns that are valid, but not yet supported. [Example](https://playground.react.dev/#N4Igzg9grgTgxgUxALhAgHgBwjALgAgBMEAzAQygBsCSoA7OXASwjvyjAQDEFc4ALABSxKASnzAAOm3xxWYAgG1KEMoSZ0A5gBp8nXABlV6rQF18AXnacAyrjK4Eg8pU6jp+WfKWEHZXfoAIn7mVhwIdg5OdFRiHl50CviKCDAwOAG8AKJpOKHWEfaOgjGUcTLhWSQkCIyCguIWAHwS8Z5yiQQduOllqZb4dAgA7vgAggBGOLgAwqw9EH0wDW34ZGAAngz4tAzMrPgw9IJgGppkYq0ynp49G1c3jwlJMAhgA2TDZEw0vALCMHKTxuHSSvnsHy+P0ObwAdAArSB0FbXJ5MEj4E5MTR0C6wshTPAIQjiV64WB0VaPIJ+QTgsjuVGeAC+sgcAkxqRg4ikTJu6MxpxxeIJ02JpN4FKpN30OXSyy5jOBrJIGgulHuvOBnnRWOFlHxhMcJJh5JglL5nn0RjUZ2cFzc0uZq2dqKOyO6vUoqVhQtxQJuZIpmMaLWAnsW3pghumDVdLN0ilMjPiQfNEhUtq0unpui5OFdrpAzKAA)
  - Once the Compiler is stable, there's still patterns it won't support. For example external state management like rxjs. And it'll always focus on correctness, which means it won't flag cases where components should make better use of composition to improve performance. It takes our code and tries its best to make it run as fast as possible, while preserving the original intent of life cycles like `useEffect`.
- Record profile sessions that expose its data and metrics in the same way TypeScript flags errors, and ESLint shows warnings. You stay in the editor and iterate quickly and see the impact of changes immediately.
- It can analyze general metrics that are always collected by simply browsing around and interacting with the Test Studio (you don't have to start the component profiler), and then uses an LLM to prioritize which components that could be optimized, based on how slow they are, their impact, and provides you with suggested fixes similar to GitHub Copilot.
- In the future we can run it in production, similar to react production profiling, as opt-in, to analyze userland custom studios to look at real world performance bottlenecks that points us to exactly which component,  which prop, which custom hook, that's causing a bottleneck.

### What to review

Have a look on the diff, as I'm calling out what to review as my own review comments.
Please mark them as resolved once you've seen them, and feel free to ask direct follow up questions there and then resolve once you're satisifed with my answer 😌 

### Testing

Until Million supports running in production you'll only able to run it locally, and AFAIK the IDE extension is only available on VS Code for now.
1. [Install the extension](https://marketplace.visualstudio.com/items?itemName=million.million-lint)
2. `git clone`, `pnpm install` and run `pnpm dev:million-lint`.
3. Open http://localhost:3333 and browse around.
4. Open the Million extension in your IDE, you should now start seeing data coming in. You should also see inline data show up as you browse files that are instrumented.

You can look at `dev/test-studio/.react-compiler-bailout-report.json` if you wonder exactly what files are instrumented and will have intellisense available.

#### Why can't we just run Million on every file?

1. Million doesn't yet support instrumenting files that are optimized by React Compiler. You'll see crashes like [this](https://react.dev/errors/300).
2. Due to our "death by a thousand cuts/renders" situation, instrumenting all our (over a thousand components) quickly lead to obscene amounts of data, eventually crashing your browser and VS Code (or slowing down your computer to a grinding halt, until you force quit them). 
   Since we currently have to disable React Compiler if we want to run Million on everything, it significantly increases renders and thus render data.
   This isn't the first time we've had a problem like this. [Our very own @runeb has official React contributor status after he sent a fix that allowed us to run the React DevTools profiler on more than just a few seconds in the Studio](https://github.com/facebook/react/pull/30848).
   On an M1 Pro 32GB RAM MacBook Pro you'll barely able to view a simple document list in Structure Tool, open a document, or type to fast in the search input and it dies.
   On a M4 Max 128GB RAM MacBook Pro you get quite a bit further, and can type into fields for a few minutes before it caves. Typing into a Portable Text field kills it almost instantly.

### Notes for release

N/A this is purely a monorepo addition, only to the test studio. It doesn't affect what we publish to npm with `sanity` or `@sanity/vision` in any way.

<details>

<summary>

### Verbose information

</summary>

#### How Million and React Scan compares to React DevTools
Million builds on top of [React Scan](https://react-scan.com), and integrates its performance profiling data with VS Code.
The benefit of React Scan, over the regular React Profiler in React DevTools, is that it instruments and traces how often hooks, props and state change over time and how it leads to cascading rerenders.
React DevTools can tell you what happened on each render frame, and is enough when you deal with performance issues where a single render is very slow.

In our codebase we have very few problems like that (they happen at scale, such as very large documents, large amount of fields, massive amount of editors working in the same document, huge Portable Text fields, those kind of scenarios). The general jank and slowness in the Studio is due to the "death of a thousand cuts". We have constant, cascading renders, paired with memoization that triggers garbage collection events as the memoization can't survive unstable props, state and context values. React DevTools isn't well equipped to help us here, what we need is the ability to track changes and counters over time in order to find the needles in our haystack.

</details>


.gitignore
```diff
@@ -83,3 +83,6 @@ yalc.lock
 ## Documentation Report
 scripts/docs-report.md
 
+# Temporary data collected by Million Lint
+**/.million/store.json
+dev/test-studio/.react-compiler-bailout-report.json
```
stipsan, line 88: Committing this to git would cause all sorts of annoying merge conflicts on PRs that do perf work, that for example makes components no longer bail out. Or when `eslint-plugin-react-compiler` and `babel-plugin-react-compiler` itself changes what cases it bails and supports as we bump deps.

dev/test-studio/sanity.cli.ts
```diff
@@ -13,12 +27,34 @@ export default defineCliConfig({
   // A) `SANITY_STUDIO_REACT_STRICT_MODE=false pnpm dev`
   // B) creating a `.env` file locally that sets the same env variable as above
   reactStrictMode: true,
-  reactCompiler: {target: '18'},
+  reactCompiler: millionLintEnabled
+    ? {
+        target: '18',
+        sources: (filename) => {
+          if (filename.includes('node_modules')) {
```
stipsan, line 34: This is the default behavior of `sources`, since we're overriding it we have to ensure we don't accidentally try running the compiler on non-src files from npm.

dev/test-studio/sanity.cli.ts
```diff
@@ -13,12 +27,34 @@ export default defineCliConfig({
...
+      }
+    : {target: '18'},
   vite(viteConfig: UserConfig): UserConfig {
     const reactProductionProfiling = process.env.REACT_PRODUCTION_PROFILING === 'true'
 
     return {
       ...viteConfig,
+      plugins: millionLintEnabled
+        ? [
+            require('@million/lint').vite({
```
stipsan, line 50: I tried to use it as a regular import at the top originally.
But for some reason, if you have this line top-level:
```ts
import Million from '@million/lint'
```
then `sanity build` [hangs forever](https://vercel.com/sanity-sandbox/test-studio/8ZbBMVR6PbKGF2tTrMKmymytrjf3?filter=errors)

package.json
```diff
@@ -17,7 +17,8 @@
     "check:deps": "pnpm --recursive --parallel exec depcheck",
     "check:format": "prettier . --check",
     "check:lint": "turbo run lint --continue -- --quiet",
-    "check:react-compiler": "eslint --no-inline-config --no-eslintrc --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --rule 'react-compiler/react-compiler: [warn]' --ignore-path .eslintignore.react-compiler --max-warnings 79 .",
+    "check:react-compiler": "eslint --cache --no-inline-config --no-eslintrc --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --rule 'react-compiler/react-compiler: [warn]' --ignore-path .eslintignore.react-compiler --max-warnings 77 .",
+    "report:react-compiler-bailout": "eslint --cache --no-inline-config --no-eslintrc --ext .cjs,.mjs,.js,.jsx,.ts,.tsx --parser @typescript-eslint/parser --plugin react-compiler --rule 'react-compiler/react-compiler: [error,{__unstable_donotuse_reportAllBailouts:true}]' --ignore-path .eslintignore.react-compiler -f ./scripts/reactCompilerBailouts.cjs . || true",
```
stipsan, line 21: The reason this command has `__unstable_donotuse_reportAllBailouts:true` is because we want to make sure Million Lint runs on all components that are not optimized by the compiler.

Running the eslint compiler plugin reports 77 issues atm that it considers actionable.
Actionable means its bailing because of a problem with the code, for example reading a `ref.current` during render (which has always been unsafe, and the reason it's not impacting us atm is because we're over-rendering).
There are a lot of cases where the compiler skips a component that's valid. For example it doesn't fully support TypeScript "satisfies" syntax, [here it bails](https://playground.react.dev/#N4Igzg9grgTgxgUxALhAFwJ4AcEAICScEAdgCrZ4C8uA5ADYQCGAJgJbEDmNuAPrWIwBuCGgB1i4hAA8sEGGlzMEAM0ZQ6C5VGJw0rErgBCUNGhIAKLDAhYwASlzBxuXEWJgFrN7mqXrtgDoGFnYOXAB+WmC2Tm5kfiERBwE9MGVWBDACN3IcZ1wYBDRYYlwAHgAjEzNSrxJKYDriAF8APgBhOi8AawBCMoB6KtMSVoBucWbxEGagA), [while here it works.](https://playground.react.dev/#N4Igzg9grgTgxgUxALhAFwJ4AcEAICScEAdgCrZ4C8uA5ADYQCGAJgJbEDmNuAPrWIwBuCGgB1i4hAA8sEGGlzMEAM0ZQ6C5VGJw0rErgBCUNGhIAKLDAhYwASlzBxuXEWJgFrN8gJvyOXGpLa1sAOgYWdg5cAH5aCLZObh8aAWEaO2dcGAQ0WGJcAB4AIxMzAq8SSmBK4gBfAD4AYTovAGsAQkKAelLTEgaAbnE68RA6oA)
These type of issues aren't flagged without the `__unstable_donotuse_reportAllBailouts` option to avoid leading people to refactor code in ways to only please the current version of the compiler, instead of what's best for code readability and for the humans working in the project.

In this case we don't really care or show the bailed out components that aren't actionable, we just gobble it up in a JSON file to setup mutually exclusive source filters between Million Lint and React Compiler.
That way we get intel on which of the components that don't have react errors, but for whatever reason isn't compiled, is worth looking into manually memoizing in the mean time.

dev/test-studio/sanity.cli.ts
bjoerge: Would love to have this as a code comment too :innocent:

bjoerge (approved):
Really exciting stuff! Thank you for the in-depth explanations <3, seems like a low risk addition to merge as well. Let's go!

dev/test-studio/sanity.cli.ts
stipsan: Yep, I'll add a comment. It's the WebSocket server setup for the VS Code extension that isn't closed 🙌

dev/test-studio/sanity.cli.ts
stipsan: Will do!

PR #8033, opened 2024-12-13
fix(core): make PTE read-only until it's `'ready'`

christianhg:
### Description

PTE emits a single `'ready'` event when it's done syncing the initially provided `value`. Therefore, it's best to treat it as (and make it look) read-only until that event has been emitted.

As of v1.6.0 PTE also treats itself as read-only until it's done setting up: https://github.com/portabletext/editor/releases/tag/editor-v1.16.0

It is especially important to respect the `'ready'` event since in the near-future PTE will **stream** the initial value in order to not block the main thread if the value is thousands of blocks long.

Note: I noticed that the Studio's PTE wrapper (and perhaps PTE itself too) does not fully respect it's own `readOnly` prop. For example, it seems like you can edit annotations while in read-only. However, it's outside of the scope of this commit to fix that.

### What to review

Is PTE correctly made editable after it has loaded?

### Testing

Is PTE correctly made editable after it has loaded?

### Notes for release

n/a


christianhg:
> Nice! Perhaps could raise a studio bug for the `readOnly` issue in studio PTE you saw?

Thank you! Will do!

PR #8035, opened 2024-12-13
feat(cli): add cors entry automatically for template package

RostiMelk:
### Description

- Made a map of all the default ports for all frameworks listen in `@vercel/frameworks`
- Using said map to automatically apply a CORS entry per package in template

Fixes GRO-3036

### Testing

Run
```shell
```

See if http://localhost:3000 was added in Manage for your project

### Notes for release

N/A

### DIFF
--- a/packages/@sanity/cli/src/util/remoteTemplate.ts
+++ b/packages/@sanity/cli/src/util/remoteTemplate.ts
@@ -12,6 +12,7 @@ import {
 } from '@sanity/template-validator'
 import {x} from 'tar'
 
+import {debug} from '../../debug'
 import {type CliApiClient, type PackageJson} from '../types'
 
 const ENV_VAR = {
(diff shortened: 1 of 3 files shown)


packages/@sanity/cli/src/util/remoteTemplate.ts
```diff
@@ -318,3 +318,19 @@ export async function generateSanityApiReadToken(
...
+  apiClient: CliApiClient,
+): Promise<void> {
+  try {
+    await apiClient({api: {projectId}}).request({
+      method: 'POST',
+      url: '/cors',
+      body: {origin: origin, allowCredentials: false},
+    })
+  } catch {
+    // Silent fail, it most likely means that the origin is already set
```
juice49, line 334: It could be a good idea to [debug log the error using `debugIt`](https://github.com/sanity-io/sanity/blob/2bd96f628b170d80991faa3677eeaa67aab929b7/packages/@sanity/cli/src/debug.ts#L3) so we have the ability to investigate failures (e.g. by asking the user to repeat the command with debug flags enabled and share the logged error).

RostiMelk: 💯 Totally!

juice49 (approved):
Nice improvement 🙌.

PR #8036, opened 2024-12-13
feat(cli): remove .github dirs when initializing with a remote template

RostiMelk:
### Description

Filter away .github directories when initializing with a remote template. This is to prevent potential security risks from running unknown actions if the user decides to push their project to GitHub

Fixes GRO-3053

### Questions

Are there any more directories we should omit?

### Testing

Run
```shell
```
See if the directory includes a .github folder

### Notes for release

N/A


juice49 (dismissed):
Looks good to me 🙌. The one thing I'd consider is moving the regex of disallowed paths to a variable or constant. But not a blocker at all.

PR #8003, opened 2024-12-10
feat(cli): misc copy changes

RostiMelk:
#### New output methods and adjustments:
- Added new method `success` in `cliOutputter`. This logs a message with a `chalk.green('✓')` in front
- Adjusted warn method to have `chalk.yellow('⚠')` in front
- Adjusted error method to have `chalk.red('✗')` in front

#### Copy changes
- Slight modification on intro message
- Modified login copy
- Modified logout print method to use success method instead
- Sanity command suggestions are now prefixed with `npx`

### What to review

- Are there any typos?
- Does the new copy make sense?

### Testing

Run 
```shell
sanity init
```

```shell
sanity logout
```

```shell
sanity login
```

### Notes for release

N/A

PR #8037, opened 2024-12-13
fix(cli): outputters should respect %s

RostiMelk:
### Description

Slight regression from #8003 where I started joining console arguments. 
This ended up breaking string format specifiers making prints like this: 
```ts
success('You are logged in as %s using %s', user.email, getProviderName(user.provider))
```
render like this:
```shell
✓ You are logged in as %s using %s rosti@sanity.io Google
```

### Testing

Run
```shell
sanity init
```
See if the intro command look like this again

```shell
✓ You are logged in as rosti@sanity.io using Google
```

### Notes for release

N/A, as #8003 hadn't reached a release yet.

### DIFF
--- a/packages/@sanity/cli/src/outputters/cliOutputter.ts
+++ b/packages/@sanity/cli/src/outputters/cliOutputter.ts
@@ -13,19 +13,19 @@ export default {
     console.log(...args)
   },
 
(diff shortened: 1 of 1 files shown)


juice49 (approved):
Nice 👍.

PR #8040, opened 2024-12-13
refactor(core): replace `PortableTextEditor` with `EditorProvider`

christianhg:
### Description

`EditorProvider` is the new way of setting up the Portable Text Editor. It's
going to be the way we document how to use standalone PTE and it's the way
Create already uses PTE.

It exposes all the new APIs that have been under development, and for now it's
also 100% backwards compatible by the fact that:

1. No types or interfaces have been removed.
2. The old `usePortableTextEditor()` and `usePortableTextEditorSelection()`
   hooks are still exposed.
3. Static methods like `PortableTextEditor.addAnnotation(...)` still work (even
   though the `PortableTextEditor` React class isn't used to instantiate the
   editor anymore.)
3. You can hand-roll custom "plugins" to achieve backwards compatible
   behaviours like listening for a `value` prop change or forwarding a ref to
   the `PortableTextEditor` instance.

This means that this change is a pure refactor that introduces no changes in
functionality and no downstream breaking changes for customers configuring PTE.

### What to review

Does the code make sense and does everything PTE-related still work 👀 

### Testing

Existing test suite should be able to verify that things still work.

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -1,30 +1,35 @@
 import {
   type EditorChange,
+  type EditorEmittedEvent,
+  EditorEventListener,
+  EditorProvider,
   type EditorSelection,
   type InvalidValue,
   type OnPasteFn,
-  type Patch as EditorPatch,
   type Patch,
   type PortableTextEditableProps,
   PortableTextEditor,
   type RangeDecoration,
   type RenderEditableFunction,
+  useEditor,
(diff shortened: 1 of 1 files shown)


bjoerge (approved):
This is great! Love to see different concerns being split into separate blocks like this. Thank you @christianhg!

christianhg:
> This is great! Love to see different concerns being split into separate blocks like this. Thank you @christianhg!

I'm really happy to hear that! Unfortunately, I had to resolve a merge conflict so I'll need one more approval before this can get merged.

bjoerge (approved):

PR #8042, opened 2024-12-13
fix: `WebSocket is closed before the connection is established` warning

stipsan:
### Description

Ever since I gave up on #7227 it's been in the back of my mind, and taunting me every time I've been in a studio and seen this warning:

```
WebSocket connection to 'wss://:projectId.api.sanity.io/v2022-06-30/socket/:dataset?tag=sanity.studio' failed: WebSocket is closed before the connection is established.
```

That ends today, I've finally found a stable workaround!
At some point we should dive into `@sanity/bifur-client` and find out exactly why [the observable teardown](https://github.com/sanity-io/bifur-client/blob/b6062c27c87b130fa179704885d42d96361027d7/src/createConnect.ts#L70) calling `ws.close` causes the warning, maybe we have to check `ws.readyState`? Maybe we can't avoid this warning if the observable is unsubscribed to before the connection is established, and while the warning is annoying it remains the right thing to do.
Or maybe we should wait a tick `Promise.resolve().then` before creating the WebSocket, in case react does a new render as `userStore` changed so we avoid creating a resource that will be immediately closed anyway (we use this technique in `react-rx`, as well as in `@sanity/client` for its `client.listen()` observable).
I don't know, I'm not an WebSocket expert.

What I do know is that in #7227 the regression that lead to the WebSocket warning showing 3 times instead of just once, was because it shifted `useGlobalPresence` and `useDocumentPresence` completely away from `useEffect` to `useObservable` from `react-rx`.
The benefits of `react-rx`'s `useObservable` are many:
- In React Strict Mode it's unaffected by the double useEffect invocations, as it uses `useSyncExternalStore`.
- It has built in error management that integrates with react error boundaries.
- [Advanced and performant cache mechanics that follows best practice for `rxjs`, in a way that works with React 18 and 19 semantics and concurrent modes.](https://github.com/sanity-io/react-rx/blob/b5a59063f896d9fbdf7d3eb5b41ab2b37e654eba/src/useObservable.ts#L56-L87)
- It's pre-compiled by React Compiler, ensuring a lower memory footprint and faster rerenders.

But, since `useObservable` is designed to allow synchronous emits, it warms up the observable you give it, and subscribes to it during render.
Previously `useGlobalPresence` and `useDocumentPresence` subscribed much later, in a `useEffect`, causing the bifur websocket to be setup after react is done with initial renders.

The reason why the issue is resolved this time around is because the observables aren't subscribed to until after an initial `useEffect` runs and tracks mounted state. By the time this happens the dependencies for `usePresenceStore` (`bifur`, `connectionStatusStore`, and `userStore`) have stabilized and we no longer see eager closing of the WebSocket.

### What to review

Hopefully this makes sense 🙌 

### Testing

I've tested manually with `pnpm dev`, as well as the [preview deployment](https://test-studio-git-fix-websocket-disconnect-warning.sanity.dev/test/structure) and I've verified that the warning no longer happens.
I've also verified on network devtools that there's only one websocket resource related to bifur.
Here's the before screenshot, that shows the eagerly closed resource with its `Finished` status, in addition to the live one with status 101, and with `Time` set to `Pending` (indicating it's open and active):
image
And here's the after, showing only a single resource:

image

### DIFF
--- a/packages/sanity/src/core/store/_legacy/ResourceCacheProvider.tsx
+++ b/packages/sanity/src/core/store/_legacy/ResourceCacheProvider.tsx
@@ -1,4 +1,4 @@
-import {type ReactNode, useContext, useMemo} from 'react'
+import {type ReactNode, useContext, useState} from 'react'
 import {ResourceCacheContext} from 'sanity/_singletons'
 
 import {createMultiKeyWeakMap, type MultiKeyWeakMap} from './createMultiKeyWeakMap'
@@ -16,7 +16,7 @@ export interface ResourceCacheProviderProps {
 
 /** @internal */
 export function ResourceCacheProvider({children}: ResourceCacheProviderProps) {
-  const resourceCache = useMemo((): ResourceCache => {
+  const [resourceCache] = useState((): ResourceCache => {
     const namespaces = new Map<string, MultiKeyWeakMap>()
 
     // this is used to replace the `null` values in any `dependencies` so that
@@ -41,7 +41,7 @@ export function ResourceCacheProvider({children}: ResourceCacheProviderProps) {
         namespaceMap.set(dependenciesWithoutNull, value)
       },
     }
-  }, [])
+  })
 
   return (
     <ResourceCacheContext.Provider value={resourceCache}>{children}</ResourceCacheContext.Provider>
--- a/packages/sanity/src/core/store/_legacy/datastores.ts
+++ b/packages/sanity/src/core/store/_legacy/datastores.ts
@@ -20,7 +20,7 @@ import {fetchFeatureToggle} from './document/document-pair/utils/fetchFeatureTog
 import {type OutOfSyncError} from './document/utils/sequentializeListenerEvents'
 import {createGrantsStore, type GrantsStore} from './grants'
 import {createHistoryStore, type HistoryStore} from './history'
-import {__tmp_wrap_presenceStore, type PresenceStore} from './presence/presence-store'
+import {createPresenceStore, type PresenceStore} from './presence/presence-store'
 import {createProjectStore, type ProjectStore} from './project'
 import {useResourceCache} from './ResourceCacheProvider'
 import {createUserStore, type UserStore} from './user'
@@ -231,7 +231,7 @@ export function usePresenceStore(): PresenceStore {
       resourceCache.get<PresenceStore>({
         namespace: 'presenceStore',
         dependencies: [bifur, connectionStatusStore, userStore],
-      }) || __tmp_wrap_presenceStore({bifur, connectionStatusStore, userStore})
+      }) || createPresenceStore({bifur, connectionStatusStore, userStore})
 
     resourceCache.set({
       namespace: 'presenceStore',
--- a/packages/sanity/src/core/store/_legacy/presence/presence-store.ts
+++ b/packages/sanity/src/core/store/_legacy/presence/presence-store.ts
@@ -114,7 +114,7 @@ function setSessionId(id: string) {
 export const SESSION_ID = getSessionId() || setSessionId(generate())
 
 /** @internal */
-export function __tmp_wrap_presenceStore(context: {
+export function createPresenceStore(context: {
   bifur: BifurClient
   connectionStatusStore: ConnectionStatusStore
   userStore: UserStore
--- a/packages/sanity/src/core/store/_legacy/presence/useDocumentPresence.tsx
+++ b/packages/sanity/src/core/store/_legacy/presence/useDocumentPresence.tsx
@@ -1,19 +1,19 @@
-import {useEffect, useState} from 'react'
+import {startTransition, useEffect, useReducer} from 'react'
+import {useObservable} from 'react-rx'
+import {of} from 'rxjs'
 
 import {usePresenceStore} from '../datastores'
 import {type DocumentPresence} from './types'
 
+const initial: DocumentPresence[] = []
+const fallback = of(initial)
+
 /** @internal */
 export function useDocumentPresence(documentId: string): DocumentPresence[] {
+  const [mounted, mount] = useReducer(() => true, false)
+  // Using `startTransition` here ensures that rapid re-renders that affect the deps used by `usePresenceStore` delay the transition to `mounted=true`, thus avoiding creating websocket connections that will be closed immediately.
+  useEffect(() => startTransition(mount), [])
+
   const presenceStore = usePresenceStore()
-  const [presence, setPresence] = useState<DocumentPresence[]>([])
-
-  useEffect(() => {
-    const subscription = presenceStore.documentPresence(documentId).subscribe(setPresence)
-    return () => {
-      subscription.unsubscribe()
-    }
-  }, [documentId, presenceStore])
-
-  return presence
+  return useObservable(mounted ? presenceStore.documentPresence(documentId) : fallback, initial)
 }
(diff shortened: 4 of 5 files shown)


packages/sanity/src/core/store/_legacy/ResourceCacheProvider.tsx
```diff
@@ -16,7 +16,7 @@ export interface ResourceCacheProviderProps {
 
 /** @internal */
 export function ResourceCacheProvider({children}: ResourceCacheProviderProps) {
-  const resourceCache = useMemo((): ResourceCache => {
+  const [resourceCache] = useState((): ResourceCache => {
```
stipsan, line 19: A great rule of thumb for determining wether a `useMemo` is safe and side-effects free is wether you can remove it and what happens to your app:
- a) it still works, but slower
- b) it breaks, or is so much slower it's unusable

In this case it's b), as the assumption here is that `useMemo` with zero dependencies means `resourceCache` will run exactly once, and have a stable instance for hte lifecycle of components calling it. This is false. In fact, it can choose to throw away the memoized value in certain suspense scenarios: https://react.dev/reference/react/useMemo#caveats

A better solution is to use `useState`, as it does have a guarantee that it's never throwing away the value, and it has a lazy init function so that our function that creates the map is still only run once.

This is also reflected in how much simpler the React Compiler optimization is [for `useState`](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwBeAnlggQEoICGceAchACYIA0BUYCAwhADs8CAB55O3BAGU8tYQQC+BAGYwIGAgHIYdBpoA6AzDnxFqCSLES96ACz6DhYxSrUbNYWkbIB6APpgaAIA5gA2CHiCYAYChsa4hMRwOnIIALJQoXhoANIIpADqdADWabRYnGQUBBlZuflFtKXlLqrqWgB0Psl0wrXZeYUlZVgxhj4AVBMEAAJBwjACtKEEEz6GoiaE8wgwyvSUNJbwfHaUwIYEBMERADwAKgQAvFwCxQIQAO4CAHwAFBAsNkosgzEsMBYsAdQWA8DAgsEANwENgUARsARwNAWUEAgBGACsEAwCAAfAgCTKhACUAG0ALqKamgx7kqDohDKIIIFiXAg8PAAoFoEFg2gQsBQxAwuEI5GohAczHYsC4iCE4mEcmU0I0hnIgBuyygCFB7PeXwETNBBogaF5AgUhnGU1mO0Wy1W6wEmwSBHd+0Q5mO1jOAAU1Ab7bsI4CwEQ+XBbGhQiwdAJQTR6ExWAhDE7YgJJtM5kJdksVmsNiItip2QwRVajtATjYkwhY1G2DA-klk6n04pMxYW6H253ozBY1gwNSE1aCHAooRaToQ6d24yXpIZKk-n9mcHRxv7M8fvOrlclwJYRTxZCDvGXj7PjUWrd+bKQpx+vUhk0RnPA9nQXK4fB8Ehk3jNB40kFgSAgAgdCwUIDkgygAAMdVCDCCCNUITWgq0vFIAgMIVJUsQsXDIEguQ+TAiC8HsUi4C8Ag8UoOD-StZjMN-QZGmaLBcM+ZMkxRCALAETRCHoRAgQpKkGMXZclN1GgUIOGBniIAsVJ0PBYCtC5QMvG48FxYhwQfRBOAojEqPjBQ5yec9TMvTzVJvQgHMVJyCjQZjoCYKldL85ULA6DByn3VFXJ+FSvJRBAsGeJ5nzCgB+dTQk01DEB00FUXYJLL2pMqrzUmzJQOEZdJqqUoosv5GoOCqzM8wzjLvCUmpGLKOhaiKAqC2wQsYKkOuShRSsLZKBSs3rbI4FK0UclVOHwk0mTPC9ku8282sQern3vWrEDAIaIla86mrnUlyR6VIBIaYZYumg7r1vEaVUC4KoFC3VwtSxUNqimKsDi1KEsqy9UXSzLgZy7D8u0ghitSuaDoIT7kuOqKBVuvqDk4AmRjxrzyfKDoid+ix-vGwHJt1LbjQQSmrlmlSCy5jq+W6xYCD+FTbmbKwTwcMsxA6Cduzw9mnmANdjzbewFB+YAkxTNNFQUW4fHF1szn4aW8FlyNJ0ShcOv0otXVLBYKy9ata2UetgStSQjbHewD2Hdc1fORM1LY9tdMkU2nEFH3Jaj0Q8E+tBlGFgBCMP7DnZi1FfF8CAAURgNQe00fhMngj5CC5dEyIzhBcOvaPNE+wWrTr-MQAUIA), [compared to the `useMemo`](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwBeAnlggQEoICGceAchACYIA0BUYCAwhADs8CAB55O3BAFkEGCAQC+BAGYwIGAgHIYdBpoA6AzDnxFqCSLES96ACz6DhYxSrUbNYWkbIB6APpgaAIA5gA2CHiCYAYChsa4hMRwOrTCUlCheGgA0gikAOp0ANZStFicZBQE6Zk5eYW0JWUuqupaAHQ+yXRpGVm5BcWlWDGGPgBU4wQAAkHCMAK0oQTjPoaiJoRzCDDK9JQ0lvB8dpTAhgQEwREAPAAqBAC8XAJFAhAA7gIAfAAUEFgslFkGZFhgLFh9iCwHgYEFggBuAhsCgCNgCOBoCwg-4AIwAVggGAQAD4EAQZUIASgA2gBdRRUkEPMlQNEIZRBBAsC4EHh4f6AtDA0G0cFgSGIaGw+FIlEIdkYrFgHEQAlEwhkimhan0pEANyWUAQILZb0+AkZIP1EDQPIECkMY0mM22CyWKzWAg2CQIbr2iHMR2spwACmp9XaduGAWAiLy4LY0KEWDoBCCaPQmKwEIZHbEBBMprMhDtFstVusRJsVGyGMLLYdoMcbImEDHI2wYL8kkmU2nFBmLM2Q22O1GYDGsGAqfHLQQ4FFCDpgyc208uDwZHJfr8mUGR2v7E9vnPLpdFwIYeSxRD9nHnt6PtVmjc+TKQpwav16kMyqe9ydedLh8HwSCTOM0DjSQWBIeQdCwUJ9nAygAANtVCVCCENUJjUgy0vFIAhUPlRVMQsLDIHA1JeRAsC8HsIi4C8AhcUoGC-UtBi0O-OpBkaYYsI+JNE2RCALAETRCHoRBAXJSlaIXJd5J1GhEP2GAN2AfNFJ0PBYEtc5gPPa48BxYgwTvRBOFI9FyLjBRZ0eU8jPPNylKvQhbIVez8jQBjoCYSkN28pULHaDAyl3FEnO+RT3ORBAsCeR5H2CgB+FTQjUpDEE0kEUXYeLzypYqL2UyyJX2YYN0qyVwtM346v2UrjLcvSDJvcV6uGdL2ka0LfP82xAsYSlWoShQioLBL+XMrqrI4RLUTs5VOBw41GRPM8Eo869msQGrH1vKrEDAfqIiak76tnEkyW6VJpD6PiGiaLBALa89L2vQblT8gKoCCnUQqShVVvCyL3t+GKTzK88URStLgcyjCco0ggCqS6bdoICbdoO8L+Su7r9k4AnhjxhLybKdoid+ix-pGwGxp1dajQQSnzymxT80uKaCHpVreQ6hYCF+RSbibKwjwcUsxHaccu2w9nHmAFdD1bewFG+YBE2TVMFQUG4fClltTn4OW8AViMJzi+dWp0wsXRLeZy09Ksa2UOsgUtSRTdHew9yHVdNbOBNlOY9dnkkC2nAFf2Zdj0Q8DxtBlDFgBCSP7FnBi1GfJ8CAAURgNRu00fgMlg95CE5NFiOzhAsMvOPNDxkXLUbvMQAUIA).

bjoerge (approved):
What an amazing find! Really solid detective work <3

PR #8047, opened 2024-12-13
fix: replace unsafe `useMemo` with `useState`

stipsan:
### Description

[From the React docs on `useMemo`](https://react.dev/reference/react/useMemo#caveats):
> **React will not throw away the cached value unless there is a specific reason to do that.** For example, in development, React throws away the cache when you edit the file of your component. Both in development and in production, React will throw away the cache if your component suspends during the initial mount. In the future, React may add more features that take advantage of throwing away the cache—for example, if React adds built-in support for virtualized lists in the future, it would make sense to throw away the cache for items that scroll out of the virtualized table viewport. This should be fine if you rely on `useMemo` solely as a performance optimization. Otherwise, a [state variable](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state) or a [ref](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) may be more appropriate.

We do have some cases where the code is relying on `useMemo` never throwing away its value, and for the memo function itself to only be executed once, since the dependencies array is empty.
In this PR I've refactored all of them to `useState`, as per the recommendations.
By doing so we remove subtle bugs from `sanity dev` and hot module reloading, make tests less flaky, and allow React Compiler to better optimize the code since it treats `useMemo` as purely a performance optimization that it should be safe to change. While `useState` is handled differently.

For example, compare the before:
- [If usage can't be traced, `useMemo` may be removed and it's called more frequently than before](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwUYCAsghhAL4EBmMEGBA5DAgIZx4sA6AdgIQAPHPgIATBHQ5QANoTpR+3NBH4EyATwCCWLAAoAlEQEECcdWEL8IAdwIBeAiXKUIB404B8BACIceAgAdLZ2xgA0BADaALpGZgTseLAaADwSaABu3gASCHJyEFEA6rhyEgCEaQD0mTkA3ALUAiDUQA)
- [If usage is traced, it may behave exactly like before](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwUYCAsghhAL4EBmMEGBA5DAgIZx4sA6AdgIQAPHPgIATBHQ5QANoTpR+3NBH4EyATwCCWLAAoAlEQEECcdWEL8IAdwIBeAiXKUIB404B8BACIceAgAdLZ2xgA0BADaALpGZgTseLAaADwSaABu3gASCHJyEFEA6rhyEgCEwGHUaQD0mTkA3ALUAiDUQA), but this isn't a guarantee based on wether the dependencies array in `useMemo` is empty or not, it's based on what's using it.

After, with `useState` it works exactly like before:
- [If it can't trace usage, it's still memoized correctly and the init runs only once](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwUYCAyngIZ4IC+BAZjBBgQOQwKVx5sA6AO0EIAHjnwEAJggaUoAG0IMoAnmggCCAWQCeAQSxYAFAEoigggTgawhANoCIAdwC6BALwES5KjSOmPAD4CABFqBAA6RydTEwsCTjxYTQAeSTQAN0CACQR5eQgAGgIAdVx5SQBCFIB6dKyAbkFaQRBaIA)
- [When usage is traced, it optimizes it correctly](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEwUYCAyngIZ4IC+BAZjBBgQOQwKVx5sA6AO0EIAHjnwEAJggaUoAG0IMoAnmggCCAWQCeAQSxYAFAEoigggTgawhANoCIAdwC6BALwES5KjSOmPAD4CABFqBAA6RydTEwsCTjxYTQAeSTQAN0CACQR5eQgAGgIAdVx5SQBCYGjaFIB6dKyAbkFaQRBaIA)

### What to review

Does this make sense?

### Testing

### Notes for release

N/A – it's the type of fix where lots of tiny things feel slightly faster and more stable than before, but due to the nature of how `useMemo` behaves, and exactly how it can throw away cache if a component suspends during the initial mount, it's difficult to reproduce the issue and thus point to the fix.

### DIFF
--- a/dev/test-studio/components/panes/debug/DebugPane.tsx
+++ b/dev/test-studio/components/panes/debug/DebugPane.tsx
@@ -1,7 +1,7 @@
 import {ChevronDownIcon, ChevronRightIcon, ControlsIcon, LinkIcon} from '@sanity/icons'
 import {Box, Card, Code, Flex, Stack, Text} from '@sanity/ui'
 import type * as React from 'react'
-import {useMemo} from 'react'
+import {useMemo, useState} from 'react'
 import {usePaneRouter, type UserComponent} from 'sanity/structure'
 
 function usePaneChildLinkComponent(props: {
@@ -48,7 +48,7 @@ export const DebugPane: UserComponent = function DebugPane(props) {
   // notice that the ID is only created on mount and should not change between
   // subsequent re-renders, therefore this ID will only change when the parent
   // component re-renders.
-  const randomId = useMemo(() => Math.floor(Math.random() * 10000000).toString(16), [])
+  const [randomId] = useState(() => Math.floor(Math.random() * 10000000).toString(16))
 
   return (
     <Box height="fill">
--- a/packages/sanity/playwright-ct/tests/formBuilder/utils/TestForm.tsx
+++ b/packages/sanity/playwright-ct/tests/formBuilder/utils/TestForm.tsx
@@ -96,7 +96,7 @@ export function TestForm(props: TestFormProps) {
     },
   )
   const [focusPath, setFocusPath] = useState<Path>(() => focusPathFromProps || [])
-  const patchChannel = useMemo(() => createPatchChannel(), [])
+  const [patchChannel] = useState(() => createPatchChannel())
 
   useGlobalCopyPasteElementHandler({
     element: wrapperRef.current,
--- a/packages/sanity/src/core/components/scroll/scrollContainer.tsx
+++ b/packages/sanity/src/core/components/scroll/scrollContainer.tsx
@@ -8,8 +8,8 @@ import {
   useContext,
   useEffect,
   useImperativeHandle,
-  useMemo,
   useRef,
+  useState,
 } from 'react'
 import {ScrollContext} from 'sanity/_singletons'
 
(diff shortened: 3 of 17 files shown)


bjoerge (approved):
Oh wow, this is fantastic @stipsan

I was honestly not aware of this slight gotcha with useMemo and have naturally to reached for useMemo in many of these cases.

I suppose mistakenly using useMemo for such cases is something that can easlily happen again in the future – is this something that could be detected by a linter?

stipsan:
> I suppose mistakenly using useMemo for such cases is something that can easlily happen again in the future – is this something that could be detected by a linter?

It's a bit tricky, because it all comes down to wether or not it's being used as a performance optimization, or for logic.
There's no need to ban `useMemo(() => ..., [])` outright, because sometimes it's what you want.

For example, in this case:
```ts
import {MyIcon} from './MyIcon'

export default function Component() {
  const icon = useMemo(() => <MyIcon />, [])

  return (
    <>
      {icon}
    </>
  )
}
```
In this case, if I edit `MyIcon` I want it to throw away the cache so that my edit shows up :) 
A good rule of thumb, that I think is going to be hard for a linter to follow, is wether it's safe to remove the `useMemo`.
If you can remove it without logic breaking or perf degrading so much it becomes unusable, then it should be a `useState` instead. If removing it is fine, then it's fine to keep using it until React Compiler is stable and we can go back and delete all `useMemo` and `useCallback` usages and leave it to the compiler.
Note that even with the compiler in stable, there will be cases where you'll want to have a `useState`, a good example is `ResourceCacheProvider` in our codebase :D

PR #8049, opened 2024-12-13
fix: improve `SanityDefaultPreview` memoization

stipsan:
### Description

Improves render performance on lists and reference previews by wrapping `SanityDefaultPreview` in `React.memo`, as well as by refactoring it to make full use of React Compiler.

### What to review

Makes sense?

### Testing

### Notes for release

Refactored `SanityDefaultPreview` to be auto memoized by React Compiler, improving performance on document list previews, and reference selectors.


jordanl17 (approved):
LGTM

PR #8051, opened 2024-12-13
fix: remove react-hooks linter suppression

stipsan:
### Description

Suppressing the linter on `react-hooks/exhaustive-deps` causes React Compiler to bail, as it can no longer determine if the optimization it plans to do will preserve the original behavior of how often and when `useEffect` hooks will fire.
It's also a bit dangerous to leave in the codebase as what may have started off as omitting a single value from the dependencies array, can over time lead to other values being omitted that would otherwise been caught be the linter.
For example `DocumentListPaneContent` have had its linter suppression for a long long time, and since then we have added `const {t} = useTranslation(structureLocaleNamespace)` and `t` should be a dependency, but it's not.

This PR removes all of them, and in the case of `useEffect` the `useEffectEvent` ponyfill is used to preserve the original intent of only having the effect fire when specific dependencies have changed, while no longer suppressing the linter.

### What to review

Is there enough context?

### Testing

Existing tests are sufficient?

### Notes for release

Removed `react-hooks/exhaustive-deps` ESLint suppressions, allowing React Compiler to auto-memoize components that were previously skipped over.

### DIFF
--- a/packages/sanity/src/core/tasks/hooks/useActivityLog.ts
+++ b/packages/sanity/src/core/tasks/hooks/useActivityLog.ts
@@ -1,5 +1,6 @@
 import {type TransactionLogEventWithEffects} from '@sanity/types'
 import {useCallback, useEffect, useState} from 'react'
+import {useEffectEvent} from 'use-effect-event'
 
 import {useClient} from '../../hooks'
 import {getJsonStream} from '../../store/_legacy/history/history/getJsonStream'
@@ -65,10 +66,11 @@ export function useActivityLog(task: TaskDocument): {
     [transactionsUrl, token, publishedId],
   )
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, unused-imports/no-unused-vars
+  const handleFetchAndParse = useEffectEvent((rev: string) => fetchAndParse(task))
   useEffect(() => {
-    fetchAndParse(task)
     // Task is updated on every change, wait until the revision changes to update the activity log.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fetchAndParse, task._rev])
+    handleFetchAndParse(task._rev)
+  }, [handleFetchAndParse, task._rev])
   return {changes}
 }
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/tasks/hooks/useActivityLog.ts
```diff
@@ -65,10 +66,11 @@ export function useActivityLog(task: TaskDocument): {
...
   )
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, unused-imports/no-unused-vars
+  const handleFetchAndParse = useEffectEvent((rev: string) => fetchAndParse(task))
   useEffect(() => {
-    fetchAndParse(task)
     // Task is updated on every change, wait until the revision changes to update the activity log.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [fetchAndParse, task._rev])
+    handleFetchAndParse(task._rev)
```
stipsan, line 73: Sending the `task._rev` here is just to make it a valid `useEffect` dependency in the eyes of `eslint-plugin-react-hooks/exhaustive-deps` as well as the React Compiler.

packages/sanity/src/structure/panes/documentList/DocumentListPaneContent.tsx
```diff
@@ -247,10 +247,6 @@ export function DocumentListPaneContent(props: DocumentListPaneContentProps) {
         </CommandListBox>
       </RootBox>
     )
-    // Explicitly don't include `noDocumentsContent` in the deps array, as it's
-    // causing a visual bug where the "No documents" message is shown for a split second
-    // when clearing a search query with no results
```
stipsan, line 252: I can't reproduce this issue mentioned here, so I think we're good.
In any case, it's not ideal that what initially was just `noDocumentsContent` being exempt has eventually led to other deps, `t` and `paneTitle`, being accidentally omitted.
It demonstrates why it's risky to suppress the linter to begin with 😅

bjoerge: Good call,  glad to get rid of hacks like this

bjoerge (approved):

PR #8059, opened 2024-12-13
fix: improve color scheme store perf

stipsan:
### Description

This refactor is mostly about making the React Compiler support optimizing the `ColorSchemeProvider` component, so it can be excluded from Million Lint reports and not show up as false positives (as it re-renders only when the color scheme is changed manually, which is rare).

### What to review

Does it make sense?

### Testing

Tested manually, unsure if our E2E suite checks color scheme switching 🤔 

### Notes for release

N/A perf improvement too small to be worth calling out

### DIFF
--- a/packages/sanity/src/core/studio/colorSchemeStore.ts
+++ b/packages/sanity/src/core/studio/colorSchemeStore.ts
@@ -1,0 +1,39 @@
+import {type StudioThemeColorSchemeKey} from '../theme/types'
+
+function getScheme(scheme: unknown): StudioThemeColorSchemeKey {
+  switch (scheme) {
+    case 'dark':
+    case 'light':
+      return scheme
+    default:
+      return 'system'
+  }
+}
+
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/studio/colorSchemeStore.ts
```diff
@@ -0,0 +1,39 @@
...
+      return scheme
+    default:
+      return 'system'
+  }
+}
+
+/** @internal */
+export const LOCAL_STORAGE_KEY = 'sanityStudio:ui:colorScheme'
+
+let snapshot: StudioThemeColorSchemeKey
```
stipsan, line 16: React Compiler doesn't like writing to ESM bindings like this one, during render.
That's fine, we can move it into a separate file like this and the compiler no longer complains, as it's out of scope :)

rexxars (approved):

PR #8061, opened 2024-12-14
fix(vision): upgrade react-split-pane to v1.0.0

rexxars:
### Description

I've reworked `react-split-pane` to be compatible with React 19, and also upgraded the tooling to be ESM-forward as well as use named exports, drop prop-types and other ancient reactisms. Still could use a proper refactoring/perhaps use a different module, but should at least progress the React 19 story.

### What to review

Vision still works, can resize panes. 

### Testing

No new tests

### Notes for release

None.


juice49 (approved):
Looks good to me!

PR #8062, opened 2024-12-14
fix(deps): upgrade `@tanstack/react-virtual` to v3.11.2

rexxars:
### Description

Moves the dependency out of the beta range, and **brings React 19 support** which is why this is happening _now_.

@binoy14 There were some minor crashes in the array input after upgrading - the `observeElementOffset` seems to need a second argument, `isScrolling`. I assumed that since we were calling this from a scroll handler it would be safe to set this to always be `true`, but we're also calling it once on "mount", at which point I am giving it `false`. Additionally, the virtual items seems to start at an empty array now, which crashed when attempting to use the first element in the styles. Can you see that things works as expected? I did some testing and to my eyes it appears to work fine.

@robinpyon I am requesting a review from you since you have been involved in multiple components that utilize this library: scheduled publishing and CommandList being the two I can find. CommandList is being used quite a number of places, and from what I can tell things seems to work fine - but would really appreciate a quick run through to double check.

### What to review

That things still work as expected. Surfaces affected are (at least):
- "New document" list (scrollable list of schema types)
- Scheduled publishing tool (list of schedules)
- Array of objects input
- Comments/tasks mentions list
- Search (add filter)
- Search (recent searches)
- Search (results)
- Document list pane
- List panes
- Document timeline (history)

### Testing

No new tests were added - relying on existing tests and manual review.

### Notes for release

None

### DIFF
--- a/packages/sanity/package.json
+++ b/packages/sanity/package.json
@@ -182,7 +182,7 @@
     "@sanity/uuid": "^3.0.1",
     "@sentry/react": "^8.33.0",
     "@tanstack/react-table": "^8.16.0",
-    "@tanstack/react-virtual": "3.0.0-beta.54",
+    "@tanstack/react-virtual": "^3.11.2",
     "@types/react-is": "^18.3.0",
     "@types/shallow-equals": "^1.0.0",
     "@types/speakingurl": "^13.0.3",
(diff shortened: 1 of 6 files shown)


jordanl17:
Nice! We have a few uses of this in the content releases branch, so just commenting myself here to remember to rebase our branch once this is in :)

rexxars:
I see the tests are failing, so need to do some work on that

binoy14 (approved):
Tested the array inputs in the debug scenarios and other things, everything looks good to me, thanks for taking care of this!

jordanl17:
> Nice! We have a few uses of this in the content releases branch, so just commenting myself here to remember to rebase our branch once this is in :)

Well this is unfortunate, I'd mistaken React 19 support for react compiler support... but alas [this issue](https://github.com/TanStack/virtual/issues/736) still exists

PR #8076, opened 2024-12-16
fix(preview): fix issue sometimes causing hotspot menu to stay a gray box after uploading

bjoerge:
### Description
Occationally after uploading an image, the hotspot and crop icons remains greyed out forever.

Traced this down to being caused by us returning `null` from the batch fetcher we use to fetch previewed documents, in combination with a subsequent check for `undefined` to check if we need to refetch after document creations as a safeguard against initial indexing taking some time.

### What to review
Commits can be reviewed one-by-one.
- The fix is in the first commit
- The second removes moot error handling: We're connected to a shared ("hot") observable emitting the "current preview value" so the retrying logic won't trigger any refetches here
- The third cleans up the code slightly and removes some unneccesary `as any` casts

### Testing
A bit involved to add unit test for this particular case as it is highly dependent on backend index latency, but will do some follow improvements and see if I can include some unit tests with those.

### Notes for release
- Fixes an issue that could sometimes leave the hotspot and crop icons remains greyed out after uploading an image.

### DIFF
--- a/packages/sanity/src/core/preview/observeFields.ts
+++ b/packages/sanity/src/core/preview/observeFields.ts
@@ -7,7 +7,7 @@ import {
   fromEvent,
   merge,
   type Observable,
-  of as observableOf,
+  of,
   timer,
 } from 'rxjs'
 import {
@@ -100,20 +100,20 @@ export function createObserveFields(options: {
     return listen(id).pipe(
       switchMap((event) => {
         if (event.type === 'connected' || event.visibility === 'query') {
-          return fetchDocumentPathsFast(id, fields as any).pipe(
+          return fetchDocumentPathsFast(id, fields).pipe(
             mergeMap((result) => {
               return concat(
-                observableOf(result),
-                result === null // hack: if we get null as result here it can be because the document has
+                of(result),
+                result === null // hack: if we get undefined as result here it can be because the document has
                   ? // just been created and is not yet indexed. We therefore need to wait a bit
                     // and then re-fetch.
-                    fetchDocumentPathsSlow(id, fields as any)
+                    fetchDocumentPathsSlow(id, fields)
                   : [],
               )
             }),
           )
         }
-        return fetchDocumentPathsSlow(id, fields as any)
+        return fetchDocumentPathsSlow(id, fields)
       }),
     )
   }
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/preview/utils/optimizeQuery.ts
```diff
@@ -52,13 +52,13 @@ export function toQuery(combinedSelections: CombinedSelection[]): string {
...
-): (Doc | null)[] {
-  return queryResult.reduce((reprojected: (Doc | null)[], subResult, index) => {
+): (Doc | undefined)[] {
+  return queryResult.reduce((reprojected: (Doc | undefined)[], subResult, index) => {
     const map = combinedSelections[index].map
     map.forEach((resultIdx, i) => {
       const id = combinedSelections[index].ids[i]
       const found = subResult.find((doc) => doc._id === id)
-      reprojected[resultIdx] = found || null
+      reprojected[resultIdx] = found ?? undefined
```
bjoerge, line 61: This is what fixes the issue – returning `null` here caused the check here to always be `false`, thus never invoking the "slow fetch": https://github.com/sanity-io/sanity/blob/b77f684f04c4d090fafb94e1d43184526c461eba/packages/sanity/src/core/preview/observeFields.ts#L107

packages/sanity/src/core/preview/observeFields.ts
```diff
@@ -104,7 +104,7 @@ export function createObserveFields(options: {
             mergeMap((result) => {
               return concat(
                 observableOf(result),
-                result === undefined // hack: if we get undefined as result here it can be because the document has
+                result === null // hack: if we get null as result here it can be because the document has
```
bjoerge: The `null` is coming from here (which never returns `undefined`): https://github.com/sanity-io/sanity/blob/b14622ed44899ad6da7bb0fd1d46f56c0186a04b/packages/sanity/src/core/preview/utils/optimizeQuery.ts#L61

cngonzalez (approved):
This makes sense to me, with one clarifying question: there's no reason to retry, if I understand correctly? We can expect a non-null value to emit after a null value if, indeed, that is the case?

bjoerge:
> with one clarifying question: there's no reason to retry, if I understand correctly?
Correct! The listener already have retrying built in, so it will does it best to always stay connected as long as it has 1 or more subscribers.

> We can expect a non-null value to emit after a null value if, indeed, that is the case?
Normally yes – technically you can get null and then null either because
1. the document is created then deleted right after
2. the document has not yet been indexed in the query store, even after the delayed fetch returns. I expect this to happen very rarely, but realized that we can fix this too by keeping track of the "latest" revision id and keep fetching until we have it (working on a follow-up fix for this)

PR #8077, opened 2024-12-16
fix: don't read/write ref during render

stipsan:
### Description

Reading and writing to the `current` property of a `useRef` during render is unsafe. [Check caveats for more info.](https://react.dev/reference/react/useRef#caveats) I've refactored the places flagged by React Compiler for violating this rule. The cases I've left alone are either false negatives, or a bit too tricky to refactor atm as the perf around it is sensitive (basically everything related to `immutableReconcile`).

### What to review

Where should more comments and context be added?

### Testing

### Notes for release

Unsafe access of `current` on React refs fixed, allowing React Compiler to auto-memoize mor ecomponents, and increasing general stability and concurrent mode safety.


cngonzalez (approved):
Manually tested those places where tests were not in place (global copy/paste, image resize) and all seems well

PR #8080, opened 2024-12-16
test: fix missing await on assertions for `resolveInitialValue()`

rexxars:
### Description

Promise-based tests should be awaited - vitest complained about some tests not doing this. This PR fixes that

### What to review

Tests pass

### Testing

Tests are already in place, just needed to be async/awaited

### Notes for release

None


bjoerge (approved):
Good catch, thanks!

PR #8089, opened 2024-12-17
fix: allow `useRelativeTime` to be auto-memoized

stipsan:
### Description

Follows up on #8034, as the `'use no memo'` directive is meant as a short term fix.
The original issue is due to React Compiler working under the assumption that reactive values are in `useState`, `useReducer` or `useSyncExternalStore`.
Calling `Date.now()` during render is considered a side-effect, and also breaks in other scenarios, such as SSR Hydration (as `Date.now()` will return different values during server render and when the browser hydrates).
Since we already have a `useReducer` trick in place that is used to schedule re-renders on set intervals, the theory is that we can just move the `Date.now()` value as state in that same reducer.
The benefit of doing this is that it allows memoizing and reduce work during render, as the `useRelativeTime` hook might be called by its components more frequently than its intervals, but it won't do work unless the `now` state itself has updated.

### What to review

Does it make sense?

### Testing

### Notes for release

Relative time stamps are optimized by React Compiler, reducing memory usage and allowing faster rerenders.

### DIFF
--- a/packages/sanity/src/core/hooks/useRelativeTime.ts
+++ b/packages/sanity/src/core/hooks/useRelativeTime.ts
@@ -1,5 +1,3 @@
-'use no memo'
-
 import {
   differenceInDays,
   differenceInHours,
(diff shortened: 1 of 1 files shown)


juice49 (approved):
This is great. Thanks, Cody ✨.

PR #8090, opened 2024-12-17
fix: search modal stays open with metaKey

EoinFalconer:
### Description

Added a check for the `metaKey` property flag on the `MouseEvent` event. The goal here is that when you open the global search, you should be able to open multiple tabs with relevant documents from your search without needing to reopen the model every time.

### What to review

As this is my first PR to the studio, I'm wondering if there are any things/flows I'm missing here. Seems like it should be ok, and I can't see any search related tests but please do let me know otherwise 😄

### Notes for release

Probably not big enough to have anything here, but maybe something like `add more ergonomic search result handling for multiple documents`.

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
@@ -1,5 +1,5 @@
 import {Card, Flex} from '@sanity/ui'
-import {useCallback} from 'react'
+import {type MouseEvent, useCallback} from 'react'
 import {styled} from 'styled-components'
 
 import {CommandList, type CommandListRenderItemCallback} from '../../../../../../components'
@@ -50,12 +50,12 @@ export function SearchResults({disableIntentLink, inputElement, onItemSelect}: S
    * Add current search to recent searches, trigger child item click and close search
    */
   const handleSearchResultClick = useCallback(
-    (event) => {
+    (e: MouseEvent<HTMLElement>) => {
       if (recentSearchesStore) {
         recentSearchesStore.addSearch(terms, filters)
       }
       // if the cmd key is pressed, we don't want to close the search
-      if (!event?.metaKey) {
+      if (!e?.metaKey) {
         onClose?.()
       }
     },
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
```diff
@@ -49,12 +49,18 @@ export function SearchResults({disableIntentLink, inputElement, onItemSelect}: S
...
-    }
-    onClose?.()
-  }, [filters, onClose, recentSearchesStore, terms])
+  const handleSearchResultClick = useCallback(
+    (e: MouseEvent<HTMLElement>) => {
+      if (recentSearchesStore) {
+        recentSearchesStore.addSearch(terms, filters)
+      }
+      // if the cmd key is pressed, we don't want to close the search
+      if (!e?.metaKey) {
```
snorrees, line 58: Needs to support ctrl as well

packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
```diff
@@ -49,12 +49,18 @@ export function SearchResults({disableIntentLink, inputElement, onItemSelect}: S
...
-      recentSearchesStore.addSearch(terms, filters)
-    }
-    onClose?.()
-  }, [filters, onClose, recentSearchesStore, terms])
+  const handleSearchResultClick = useCallback(
+    (e: MouseEvent<HTMLElement>) => {
+      if (recentSearchesStore) {
+        recentSearchesStore.addSearch(terms, filters)
+      }
+      // if the cmd key is pressed, we don't want to close the search
```
snorrees, line 57: I think the comment should say _why_ we dont want to close it (its to support open result in new link)

jordanl17 (dismissed):
🚀

snorrees (dismissed):

packages/sanity/src/core/studio/components/navbar/search/components/searchResults/SearchResults.tsx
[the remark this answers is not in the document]
EoinFalconer: Added this now.

PR #8091, opened 2024-12-17
feat(sanity): export useReferringDocuments and mark as beta

bjoerge:
### Description
The `WithReferringDocuments` HOC (Higher Order Component) has been deprecated for a while. It has been marked as `@internal`, however it's been exported so it's been available to use.

It's been replaced it with a `useReferringDocuments` hook which has not been exported by the `sanity` package. This hook has a couple of gotchas (mentioned in the tsdoc block), but if you do consider those, I think this hook is useful enough to make it exported (although flagged as `@beta`)

The existing `useReferringDocuments()` hook were always returning complete documents, which is suboptimal for cases were you only need a few fields from the referrring document (e.g. their title, or maybe you just want to know wheter there _are_ documents referring to a certain document).  I've changed the hook so it requires an array of fields, while still allowing complete documents by providing an empty array (this borrows the pattern from our existing `useFormValue([])` to get the full value).

I didn't spend time on verifying that existing internal code would not be affected by this change, so I've made an internal-only hook  `useLegacyReferringDocuments` which preserves the old behavior and updates existing internal callsites to use this one instead.

### What to review
Does the changes make sense? Does the tsdoc make sense?

### Testing
Attempted to write some basic tests, but ran into issues mocking `useDocumentStore` – after a bit of digging it looks like these issues was probably the same issues that made us comment out the [ConfirmDeleteDialog tests](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/components/confirmDeleteDialog/__tests__/index.test.tsx#L108) earlier (cc @binoy14 - do you remember what the issue was here?). I've tested manually and verified that the Confirm delete dialog works as it used to. 

### Notes for release
- Exports `useReferringDocuments()` hook


binoy14 (approved):
Changes makes sense to me, thanks!

PR #8092, opened 2024-12-17
fix: tooltip position in announcements popup

EoinFalconer:
### Description

Here is a video of what was happening before this change, you can see that the _Close_ tooltip is rendering at the end of the announcements dialog. I noticed also that now that we have more announcements and scroll that this is causing some layout shift.

https://github.com/user-attachments/assets/9ae5fac0-889d-4248-9efc-2a7e8099d455

It seems as though the container for the button component should have an extra `position: relative` in the container to stop the tooltip from flying around, what I have done to remedy this is to just add a container around the button in the dialog.

### What to review

Whether this is something that should be tackled in `@sanity/ui` or whether we are ok to proceed with this approach.

### Testing

n/a

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/studio/studioAnnouncements/StudioAnnouncementsDialog.tsx
+++ b/packages/sanity/src/core/studio/studioAnnouncements/StudioAnnouncementsDialog.tsx
@@ -36,12 +36,13 @@ const DialogHeader = styled(Grid)`
   background: var(--card-bg-color);
 `
 
-const FloatingButton = styled(Button)`
+const FloatingButtonBox = styled(Box)`
   position: absolute;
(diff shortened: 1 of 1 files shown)


cngonzalez (approved):
looks reasonable to me! i'll defer to @RostiMelk since it's his original work but I don't see an issue.

RostiMelk:
> looks reasonable to me! i'll defer to @RostiMelk since it's his original work but I don't see an issue.

Sorry, I had nothing to do with this feature 😅 I think @pedrobonamin did the implementation.

RostiMelk (approved):
This looks good to me 👍

PR #8096, opened 2024-12-18
fix: fixes schema extraction with nested union refs

rneatherway:
This was seen in a schema produced by automated tooling. The recursion
guard was too strict in this case as the two occurrences of a
reference ending up having the same guard path name and extraction
would therefore abort before looking inside the second one, instead
returning `{}` which would fail when attempting to inspect the name
field a little lower down.

### DIFF
--- a/packages/sanity/src/_internal/cli/actions/graphql/extractFromSanitySchema.ts
+++ b/packages/sanity/src/_internal/cli/actions/graphql/extractFromSanitySchema.ts
@@ -481,7 +481,9 @@ export function extractFromSanitySchema(
     }
 
     try {
-      unionRecursionGuards.add(guardPathName)
+      if (guardPathName !== 'reference') {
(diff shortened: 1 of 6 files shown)


rneatherway:
I tried updating the snapshots using `pnpm test -- -u` but I get a lot of errors e.g.

```
 FAIL  src/core/studio/components/navbar/search/contexts/search/reducer.test.ts > searchReducer > should not merge results after fetching an additional page when not using `textSearch` strategy
TypeError: Cannot read properties of undefined (reading 'match')
 ❯ src/core/studio/components/navbar/search/contexts/search/reducer.test.ts:294:31
    292|     const [state] = result.current
    293|
    294|     expect(state.result.hits).toMatchInlineSnapshot(`
       |                               ^
    295|       [
    296|         {
```

I checked and that's the same on `next` for me. Any advice?

bjoerge (approved):

PR #8103, opened 2024-12-18
Fix/sapp 1966

EoinFalconer:
### Description

Found this interesting issue when looking into [SAPP-1966](https://linear.app/sanity/issue/SAPP-1966/cli-version-flag-returns-help-as-well). The `process.exit` function in the CLI is monkey patched so we can send telemetry before the process ends. This is fine in all other cases in `cli.ts` except for when we are checking whether the command is to simply check the version using the `--version` or `-v` flag.

I think that this is a bit dangerous to monkey patch `process.env` especially in a cli environment (the comment remarks this) so I thought I would give it a go to refactor a bit. We could obviously just add a `return` statement after the `process.exit` call in question but it seems like a recipe for problems if you go about using `process.exit` in some other file and assume that your code will stop executing at that point, but it won't.

### What to review

Is the approach ok? Happy to just leave the `return` or go ahead and create some sort of `async process.exit()` set-up, but it feels the risk is bigger than the reward.

### Testing

Added a test, tell me if it is in the wrong place.

### Notes for release

Not worth it.

### DIFF
--- a/packages/@sanity/cli/src/cli.ts
+++ b/packages/@sanity/cli/src/cli.ts
@@ -27,19 +27,6 @@ import {runUpdateCheck} from './util/updateNotifier'
 const sanityEnv = process.env.SANITY_INTERNAL_ENV || 'production' // eslint-disable-line no-process-env
 const knownEnvs = ['development', 'staging', 'production']
 
-function wait(ms: number) {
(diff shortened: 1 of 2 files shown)


jordanl17:
Closing this PR as CLI is moving outside of the monorepo

PR #8110, opened 2024-12-19
fix: preload documents on hover

stipsan:
### Description

By preloading the `editState` for a document it makes for a much snappier experience when navigating documents, here's a video demonstrating both side-by-side (with the new behavior on the left, and no preloading on the right):

https://github.com/user-attachments/assets/118162cb-0c06-4508-8c6d-774df83f5920

### What to review

Is there enough inline context or should I add code comments?

### Testing

Existing tests should capture regressions, otherwise the manual testing steps are demonstrated in the video 🙌 

### Notes for release

Added preloading of documents when hovering them in document list panes. This speeds up the experience of navigating between documents pretty significantly.

### DIFF
--- a/packages/sanity/src/structure/components/pane/Pane.tsx
+++ b/packages/sanity/src/structure/components/pane/Pane.tsx
@@ -5,8 +5,8 @@ import {
   type HTMLProps,
   type ReactNode,
   useCallback,
-  useEffect,
   useImperativeHandle,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
@@ -89,7 +89,7 @@ export const Pane = forwardRef(function Pane(
     ref.current = refValue
   }, [])
 
-  useEffect(() => {
+  useLayoutEffect(() => {
     if (!rootElement) return undefined
     return mount(rootElement, {
       currentMinWidth: currentMinWidthProp,
--- a/packages/sanity/src/structure/components/paneItem/PaneItem.tsx
+++ b/packages/sanity/src/structure/components/paneItem/PaneItem.tsx
@@ -10,6 +10,7 @@ import {
   type ComponentType,
   type MouseEvent,
   type ReactNode,
+  startTransition,
   useCallback,
   useEffect,
   useMemo,
@@ -22,6 +23,7 @@ import {
   SanityDefaultPreview,
   useDocumentPresence,
   useDocumentPreviewStore,
+  useEditState,
   useSchema,
 } from 'sanity'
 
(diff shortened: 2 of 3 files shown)


packages/sanity/src/structure/components/paneItem/PaneItem.tsx
```diff
@@ -144,16 +139,22 @@ export function PaneItem(props: PaneItemProps) {
...
+  const [handleMouseEnter, preload] = usePreloadEditState(id, schemaType?.name, selected)
+
   return (
     <PreviewCard
       data-testid={`pane-item-${title}`}
       __unstable_focusRing
-      as={Link as FIXME}
+      as={ChildLink as FIXME}
+      // @ts-expect-error - `childId` is a valid prop on `ChildLink`
+      childId={id}
```
bjoerge, line 151: I don't think we're guaranteed that `childId` is always a document id, so think we might end up preloading for preview items that's not documents.

I _think_ there should be a preview component that is only invoked in the context of a document, if so it would be better to move this logic there

stipsan: The hook requires a `documentType` to load, if `childId` isn't a document, will `schemaType?.name` still be a string?

packages/sanity/src/structure/components/paneItem/PaneItem.tsx
```diff
@@ -124,14 +127,6 @@ export function PaneItem(props: PaneItemProps) {
     documentPresence,
   ])
 
-  const Link = useMemo(
```
stipsan, line 127: Defining components during render should be avoided whenever possible 🙌

packages/sanity/src/structure/components/paneItem/PaneItem.tsx
stipsan: ok testing fix

stipsan: I think I fixed it, at least in my testing, can you confirm?

bjoerge: Works now 🙌🏼

packages/sanity/src/structure/components/pane/Pane.tsx
```diff
@@ -89,7 +89,7 @@ export const Pane = forwardRef(function Pane(
     ref.current = refValue
   }, [])
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
stipsan, line 92: This fixes a long standing quirk where clicking on a document sees a weird initial layout shift before the panel is mounted 😮‍💨

bjoerge:
(did a rebase of this branch, fyi @stipsan)

jordanl17 (approved):

PR #8112, opened 2024-12-19
fix(releases): when reverting, published versions are used to derive translog

jordanl17:
### Description
Wrongly, the translog for the version of the document in the published release was being used to derive the revert state for each document. This worked in most cases because the first translog record for the version document is copying the existing publish document - but this fails when the published document is updated before publish but after release creation; or in other cases such as when a document is unpublished in a release.

This PR updates this so that the translog of the published document is always used to lookup the transaction of the release publish. As previously this will also:
1. `usePostPublishTransactions` checks if there are any transactions after the publish tranaction
2. `useDocumentRevertStates` which derives the revert state by looking at the transaction 1 prior to the publish trans

### Testing
Some minor test updates made to correct for the fact that the documents provided to the hooks are the version documents, and then additional mock updates to reflect the changes to logic

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/util/__tests__/draftUtils.test.ts
+++ b/packages/sanity/src/core/util/__tests__/draftUtils.test.ts
@@ -8,7 +8,7 @@ import {
   getVersionFromId,
   getVersionId,
   removeDupes,
-} from './draftUtils'
+} from '../draftUtils'
 
 test('collate()', () => {
   const foo = {_type: 'foo', _id: 'foo'}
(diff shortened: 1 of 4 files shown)


packages/sanity/src/core/util/__tests__/draftUtils.test.ts
jordanl17, line 1: File moved into `__tests__` dir

bjoerge (approved):
Great catch, and great fix!

PR #8114, opened 2024-12-19
feat(cli): remove patching of tsconfig.json and thus silver-fleece dependency

judofyr:
### Description

The primary motivation for this change is to get rid of silver-fleece as a dependency. See https://github.com/evertheylen/silver-fleece/issues/4 for more information about that problem.

This simply removes the functionality we used it for: Patching tsconfig.json to set `compilerOptions.target` to ES2017.

- The latest Next.js template already sets the target to ES2017.
- Even so, I've tested this against a newly configured Next.js app where I set `compilerOptions.target` to ES5 and `npm run dev` still worked. I'm not sure why this target is required anyway.

### What to review

Think about scenarios where `sanity init` will be executed and if this change could lead to more confusing onboarding experience.

### Testing

@RostiMelk tested this locally in his machine against a newly initialized Next.js project (Next.js 15).

### DIFF
--- a/packages/@sanity/cli/package.json
+++ b/packages/@sanity/cli/package.json
@@ -72,7 +72,6 @@
     "pkg-dir": "^5.0.0",
     "prettier": "^3.3.0",
     "semver": "^7.3.5",
-    "silver-fleece": "1.1.0",
     "validate-npm-package-name": "^3.0.0"
   },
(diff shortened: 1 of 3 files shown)


RostiMelk (dismissed):

bjoerge (approved):
LGTM

PR #8120, opened 2024-12-20
fix(core): re-subscribes to shared pair listener opens a new connection

bjoerge:
### Description
While testing #8110 I noticed that even when preloading, clicking a document would cause a new listener request to be made. 

Tracked this down to being caused by the memoization of the shared document pair listener. This PR fixes the issue by:

- Replacing the `connector.complete()` trick with instead exposing a `_keepalive` stream that we merge in with the (memoized &  shared) stream of the pair listener. This ensures that the listener connection will be kept open for as long as there are subscribers to the pair object.
- Moving the delayed reset to the memoized function (returning a pair listener for keyed by document id). This ensures that the delayed reset happens on the memoized observable itself, rather than the connection stream.

### What to review
- Navigate to a document (A) in the Studio
- Make sure there is a single event source set up with the tag `sanity.studio.document.pair-listener`
- Navigate to another document (B)
- Make sure there is a single new event source set up with the tag `sanity.studio.document.pair-listener`
- Go back to document A before 10s has passed
- Verify that no request was made with the tag  `sanity.studio.document.pair-listener`
- Stay on document A for more than 10s
- Verify that the listener connection made for document B is now shut down (timing panel no longer says "Caution: request is not finished yet!")
- Now, going back to B should trigger a new listener connection again.

### Testing
Existing tests should be enough

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
@@ -252,8 +252,8 @@ export function checkoutPair(
       consistency$: published.consistency$,
       remoteSnapshot$: published.remoteSnapshot$.pipe(map(setVersion('published'))),
     },
-    // Use this to keep the listener connection active.
+    // Use this to keep the mutation pipeline active.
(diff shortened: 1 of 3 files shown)


stipsan (approved):
Such a great find!!! 💖

PR #8121, opened 2024-12-20
fix: support react 19 type definitions

stipsan:
### Description

On React 19 I'm seeing build errors like this:
```
Type error: Argument of type 'ComponentType | ReactNode' is not assignable to parameter of type 'ReactNode | ComponentType<{}>'.
  Type 'ComponentClass<{}, any>' is not assignable to type 'ReactNode | ComponentType<{}>'.
    Type 'import("/vercel/path1/node_modules/@sanity/types/node_modules/@types/react/index").ComponentClass<{}, any>' is not assignable to type 'React.ComponentClass<{}, any>'.
      Types of property 'contextType' are incompatible.
        Type 'import("/vercel/path1/node_modules/@sanity/types/node_modules/@types/react/index").Context<any> | undefined' is not assignable to type 'React.Context<any> | undefined'.
          Property '$$typeof' is missing in type 'import("/vercel/path1/node_modules/@sanity/types/node_modules/@types/react/index").Context<any>' but required in type 'React.Context<any>'.
```
And since `@sanity/types` has a direct dependency on `@types/react` there isn't an easy to solve it, userland has to resort to package manager overrides 😅 
It's always been bad to have direct deps on types that can have different majors that are incompatible, so the fix is to move it to a peer dep.

### What to review

Makes sense?

### Testing

If it builds it works.

### Notes for release

Improves React 19 compatibility for TypeScript users, by moving the `@types/react` direct dependency in `@sanity/types`, to a peer dependency.

### DIFF
--- a/packages/@sanity/types/package.json
+++ b/packages/@sanity/types/package.json
@@ -49,16 +49,19 @@
     "watch": "pkg-utils watch"
   },
   "dependencies": {
-    "@sanity/client": "^6.24.1",
-    "@types/react": "^18.3.5"
+    "@sanity/client": "^6.24.1"
   },
   "devDependencies": {
     "@repo/package.config": "workspace:*",
     "@repo/test-config": "workspace:*",
     "@sanity/insert-menu": "1.0.16",
+    "@types/react": "^18.3.5",
     "@vitejs/plugin-react": "^4.3.4",
     "react": "^18.3.1",
     "rimraf": "^5.0.10",
     "vitest": "^2.1.8"
+  },
+  "peerDependencies": {
+    "@types/react": "18 || 19"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/@sanity/types/package.json
```diff
@@ -49,16 +49,19 @@
...
     "@repo/test-config": "workspace:*",
     "@sanity/insert-menu": "1.0.16",
+    "@types/react": "^18.3.5",
     "@vitejs/plugin-react": "^4.3.4",
     "react": "^18.3.1",
     "rimraf": "^5.0.10",
     "vitest": "^2.1.8"
+  },
+  "peerDependencies": {
+    "@types/react": "*"
```
bjoerge, line 65: not `18 || 19` tho?

stipsan: Good catch! That's more accurate, as those are the versions we test against

stipsan: fixed!

bjoerge (approved):

PR #8123, opened 2024-12-20
test(releases): fixing issues with virtual list and releases overview and summary tests

jordanl17:
### Description

With Tanstack Virtual being upgraded in Next, this introduces issues in testing environments where the size of virtual list items is 0px, resulting in none of them being rendered (failing tests). 

The workaround is already present in next, but has been abstracted into a function as it's quite likely this will need to be used in other places as there's no clear sign this will be rectified. 

This also necessitated some changes to the testing harnesses for `RealeasesOverview` and `ReleaseSummary` to ensure the virtual list isn't rendered until the exact sized parent is first. 

### Testing
Tests updated and all now pass on COREL

### Notes for release
N/A


bjoerge (approved):
Oh wow, that's some really amazing detective work @jordanl17! Great call extracting this into a shared util as well <3

PR #8115, opened 2024-12-19
chore(cli): bump template validator

RostiMelk:
### Description

Bumping @sanity/template-validator to latest version (v1.2.1)

This version includes some api changes, so I've adjusted bootstrapRemoteTemplate to account for that

### What to review

- Build the CLI and running `npm link`
- Run `sanity init --template sanity-io/sanity-template-nextjs-clean` in your terminal.
- Make sure template installs as expected

### Notes for release

N/A

PR #8131, opened 2024-12-22
fix(cli): pin template validator version

RostiMelk:
### Description

Disabling flexible versioning for the @sanity/template-validator package i @sanity/cli

This PR won't really fix the issue before #8115 is released, but I think it's definitely worth pinning this dependency going forward to avoid this type of issue happening again. 

### Notes for release

Fixed an issue where initiating a project using a remote template crashed the CLI


cngonzalez (commented):
Apologies, I'm a little out of the loop on what happened here. Was the template validator a version ahead of our CLI tooling, and is this something we expect to happen in the future? I'm nervous to pin packages unless necessary (unless it's something we expect renovate bot to handle?)

PR #8160, opened 2025-01-02
chore(core): replace `'sanity'` imports for relative paths in core

pedrobonamin:
### Description

Adds a new rule to avoid imports from `sanity` in files inside `packages/src/core/**`  and replaces all the necessary imports.
Importing from `sanity` in core creates issues with the tests running with `vitest`, mocking `sanity` inside a test in `core` generates inconsistent imports and could produce hard to follow tests errors. 
**example:**
Screenshot 2025-01-02 at 15 55 02

### What to review

Is this change correct?

### Testing
No new functionalities added, existing tests passing should be enough

### Notes for release
n/a this is internal


bjoerge (approved):
Great!

PR #8166, opened 2025-01-03
fix(structure): add opt out option from releases in DocumentPaneProvider

pedrobonamin:
### Description

Adds an option for the `DocumentPaneProvider` to opt out from checking the global release or version when determining the document id.
This is useful in cases where plugins need to use the `DocumentPaneProvider` but not have it checking for the global version, like in the case of `sanity-assist` which has a PaneProvider inside a inspector which allows you to edit a document that will give the actions to the assist ai

Updates the return type of the `DocumentPaneValue` and verifies that the value is correctly returned by the provider with the `satisfies` validation.
 

### What to review

Is the new prop naming correct?

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -1,4 +1,5 @@
 /* eslint-disable camelcase */
+import {type ReleaseId} from '@sanity/client'
 import {isActionEnabled} from '@sanity/schema/_internal'
 import {useTelemetry} from '@sanity/telemetry/react'
 import {
@@ -23,6 +24,7 @@ import {
   getExpandOperations,
   getPublishedId,
   getReleaseIdFromReleaseDocumentId,
+  isReleaseDocument,
   isReleaseScheduledOrScheduling,
   isSanityCreateLinkedDocument,
   isVersionId,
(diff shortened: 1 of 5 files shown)


packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -737,74 +750,74 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+        setTimelineRange,
+        setIsDeleting,
+        isDeleting,
+        isDeleted,
+        timelineError,
+        timelineMode,
+        timelineStore,
+        title,
+        value,
+        selectedReleaseId: selectedReleaseId,
```
bjoerge, line 812:
```suggestion
        selectedReleaseId,
```

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -737,74 +750,74 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
...
+        views,
+        formState,
+        unstable_languageFilter: languageFilter,
+
+        // History specific
+        revisionId,
+        lastNonDeletedRevId,
+      }) satisfies DocumentPaneContextValue,
```
bjoerge, line 820: any benefit of using satisfies vs return type of the usememo callback?

bjoerge (changes requested):

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
pedrobonamin: When using satisfies if it has extra keys it will show a type error.
Having it would have helped us catch that `selectedReleaseId` was not a valid key before this changes.

With `satisfies` the error is shown.
Screenshot 2025-01-03 at 14 27 00

No error
Screenshot 2025-01-03 at 14 27 10

bjoerge: ah, that's nice!

bjoerge (approved):

PR #8171, opened 2025-01-03
fix: React 19 typings (finally)

stipsan:
### Description

This PR builds upon previous efforts to address React 19 upgrade issues but resolves additional gaps that persisted. While prior PRs applied necessary changes, they weren't sufficient to handle all TypeScript typing challenges introduced with React 19, particularly with `ReactElement` and scoped JSX.

Key highlights:
- **ReactElement Type Change:**  
  The `ReactElement` type now defaults to `unknown` instead of `any` ([ReactElement changes](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#changes-to-the-reactelement-typescript-type)). This change was a major source of TypeScript errors in schema definitions with custom components after upgrading to `@types/react@19`.

- **Scoped JSX Updates:**  
  Updated to use `React.JSX.Element` instead of `import {type JSX} from 'react'` to address compatibility issues with `@sanity/tsdoc`, which caused errors during ETL tasks ([example issue](https://github.com/sanity-io/sanity/actions/runs/12581600194/job/35065686679)). Testing confirmed this resolves the issue without introducing regressions.

- **React 19 marked as supported by the CLI**
  When installing react 19 deps the `sanity dev` command no longer reports them as unsupported.

Most of the work in this PR were a direct result of running these commands:
```bash
npx types-react-codemod@latest preset-19 ./packages/sanity/src
npx types-react-codemod@latest preset-19 ./packages/@sanity/types/src
npx types-react-codemod@latest preset-19 ./packages/@sanity/vision/src
```

### What to review

Anything not explained in the official [guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#typescript-changes) has PR comments, please give them a look and let me know if any context is missing.

### Testing

Since everything uses `@types/react@19` our own CI tells us if the new types work. If we want to be extra thorough we might want to run a tagged release and see that react 18 typescript studios are able to upgrade without issues, and that failing react 19 ones now pass type checking.

### Notes for release

React 19 typings are fully supported, TypeScript users can upgrade to `@types/react@19` and be able to remove a bunch of `// @ts-expect-error` suppressions thanks to the improvements done by the React team 🎉 

The CLI is now considering React 19 as fully supported, and no longer recommends downgrading to react 18. Keep an eye on https://arewereact19yet.sanity.build/ to see first party plugins status on their support, and soon it'll list popular third party plugins as well.


turbo.json
```diff
@@ -3,6 +3,7 @@
   // All environment variables that should invalidate a cached build needs to be declared here
   "globalEnv": [
     "CI",
+    "GITHUB_ACTIONS",
```
stipsan, line 6: Adding this as `@sanity/pkg-utils` will need it in order to detect the GitHub CI and output GitHub Action specific markers: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message

packages/sanity/src/ui-components/menuGroup/MenuGroup.tsx
```diff
@@ -11,7 +11,8 @@ export type MenuGroupProps = Pick<UIMenuGroupProps, 'icon' | 'popover' | 'text'
  * @internal
  */
 export const MenuGroup = (
-  props: MenuGroupProps & Omit<HTMLProps<HTMLDivElement>, 'as' | 'height' | 'ref' | 'tabIndex'>,
+  props: MenuGroupProps &
+    Omit<HTMLProps<HTMLDivElement>, 'as' | 'height' | 'ref' | 'tabIndex' | 'popover'>,
```
stipsan, line 15: `popover` is now a valid html attribute, `@sanity/ui` also had to add it to the omit: https://github.com/sanity-io/ui/commit/98c43b1ee9d5f8330786b20cd5fcda125747c509

packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
```diff
@@ -157,7 +157,9 @@ export function SheetListCellInner(props: SheetListCellInnerProps) {
     'aria-selected': !!cellState,
     'data-testid': cellId,
     'id': cellId,
-    'ref': (ref: InputRef) => (inputRef.current = ref),
+    'ref': (ref: InputRef) => {
+      inputRef.current = ref
+    },
```
stipsan, line 162: This were only caught during runtime thanks to StrictMode, the linter or TSC did not since we're spreading props instead of using JSX:
```
<Select
  ref={(ref: InputRef) => (inputRef.current = ref)}
/>
```

[In React 19 ref callbacks can return cleanup functions the same way `useEffect` can](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#ref-cleanup-required), and thus implicit returns in ref callbacks is no longer allowed.

packages/sanity/src/core/studio/workspace.tsx
```diff
@@ -1,16 +1,19 @@
...
+import {useContext} from 'react'
 import {WorkspaceContext} from 'sanity/_singletons'
 
 import {type Workspace} from '../config'
 
 /** @internal */
 export interface WorkspaceProviderProps {
   workspace: Workspace
-  children?: ReactChild
+  children: React.ReactNode
```
stipsan, line 9: It never made sense for this provider to have children be optional. And `ReactChild` is deprecated, the official `React.PropsWithChildren` uses `ReactNode` and so should we :)

bjoerge (commented):
Massive!

All looks good to me. I'm curious about the widespread usage of `any` for the props type of elements, e.g. `ReactElement<any>`. Would it not be better to keep them `unknown`?

packages/sanity/src/structure/panes/documentList/sheetList/SheetListCell.tsx
stipsan: > oh, but I see we have the `no-void` eslint rule enabled. oh well

no fun allowed 😩

stipsan:
> All looks good to me. I'm curious about the widespread usage of `any` for the props type of elements, e.g. `ReactElement<any>`. Would it not be better to keep them `unknown`?

I get where you're coming from :) My goal here is to unblock those on 19 first and foremost. Given the breakage is related to `React.ReactElement` changing from being implicitly `any` to `unknown`, it seems like the safest approach is to make all the 18 typings explicit, ensuring they are the same regardless of your react major.
For react 18 users nothing has changed, our published typings are literally the same as before. While for 19 we no longer get unexpected narrowing.

A good second step for us is to work through the `any` and find out which ones are safe to narrow down. I expect some will still have to be `any`, but certainly not all. Personally I also like the new honesty with the stricter implicit defaults. I had no idea that `React.ReactElement` where `any` by default, now it's obvious and I'm motivated to fix them 😄

bjoerge:
> I get where you're coming from :) My goal here is to unblock those on 19 first and foremost. Given the breakage is related to React.ReactElement changing from being implicitly any to unknown, it seems like the safest approach is to make all the 18 typings explicit, ensuring they are the same regardless of your react major.
For react 18 users nothing has changed, our published typings are literally the same as before. While for 19 we no longer get unexpected narrowing.

All good points. I agree it would be worse for everyone if we'd were to roll out such a massive breaking typing change in a minor of our own. Let's aim for changing to stricter types in our next major instead.

> I had no idea that React.ReactElement where any by default, now it's obvious and I'm motivated to fix them 😄

Same, news to me too! Glad they're fixing it.

stipsan:
@bjoerge update, I just force pushed a migration from `React.ReactElement` over to `React.JSX.Element`. 
Some more details here: https://github.com/sanity-io/ui/pull/1555

bjoerge (commented):
Amazing work, @stipsan, ship it! ✨

PR #8176, opened 2025-01-03
fix(structure): add support to force a version for the pane provider

pedrobonamin:
### Description

`Sanity assist ` uses a document defined as liveEdit so not using the global version is not enough, we need to force the published one.

So, this removes the `matchGlobalVersion` prop and changes it for a `forcedVersion` allowing the user to define which version to link the paneProvider to.

See the use for when we have a history item selected. 

Assist should do the folllowing:
```
  <DocumentPaneProvider
        {...props}
        forcedVersion={{
     selectedPerspectiveName:"published",
        selectedReleaseId: undefined,
        isReleaseLocked: false,
       }}
      />
```

### What to review
Is there a better naming we could use for the `forcedVersion` ?

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
@@ -1,3 +1,4 @@
+import {type ReleaseId} from '@sanity/client'
 import {useMemo} from 'react'
 import {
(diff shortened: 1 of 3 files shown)


bjoerge (approved):
LGTM! `forcedVersion` is clear and descriptive to me

Issue #1923, opened 2020-06-16
Dates in Studio are being displayed in "local" time, causing data to look wrong to editors in different timezones.


ndimatteo:
## Describe the bug
I'm using both the `datetime` and `richDate` schema types to set dates for "show" documents, which will be used to list out upcoming shows. If I create my event in, say, **Denver** and set a date to `2020-06-21 20:30` and then open my studio up in, say, **Los Angeles**, the date being displayed in my fields and previews are showing the above time localized. This will appear as an error to someone looking at the studio from a different timezone than the one the document was created/updated from, which may result in them changing it again, which would in turn affect the original, intended date, and on and on you could go...

## Expected behavior
The date should remain the same as what it was set at for all editors, as well as what's returned in queries.

## Screenshots
Here are two instances of Google Chrome running my studio, the one on the left is mimicking "Los Angeles" (PDT) timezone, while the one on the right is my default Chrome instance, where I'm in "Denver" (MDT) timezone 

![image](https://user-images.githubusercontent.com/737188/84826462-f96e7900-aff0-11ea-8f69-c187689326b3.png)

_Notice the displayed data is different by an hour? I had set the date in my default timezone, then pretended to fire it up in Los Angeles and the time is off by an hour, because it's being converted from the time it was set at to the current local time._

## Technical Details
**Which versions of Sanity are you using?**
```
@sanity/cli               1.149.2 (latest: 1.149.16)
@sanity/base             1.149.16 (up to date)
@sanity/components       1.149.16 (up to date)
@sanity/core             1.149.17 (up to date)
@sanity/dashboard        1.149.16 (up to date)
@sanity/default-layout   1.149.16 (up to date)
@sanity/default-login    1.149.16 (up to date)
@sanity/desk-tool        1.149.16 (up to date)
@sanity/rich-date-input  1.149.16 (up to date)
@sanity/vision           1.149.16 (up to date)
```

**What operating system are you using?**
Mac OS Catalina Version 10.15.4 (19E266)

**Which versions of Node.js / npm are you running?**
```
6.14.5
v12.13.1
```

vicbergquist:
Thank you for reporting! We agree that this can be confusing and we have added it to our internal tracker! 👍

ndimatteo:
@vicbergquist awesome! I've been thinking about this more, and wondering if a more explicit `timezone` setting should be added to date/time fields. That way you can use the set timezone to ensure the rendered value is the same no matter where the studio is being viewed from?

[...]

PR #8181, opened 2025-01-04
feat: add displayTimezone option to datetime input

EoinFalconer:
### Description

Adding the ability to add a `displayTimezone` option to the `datetime` input. This allows you to specify the timezone in which the stored UTC string from content lake will be displayed.

A common use case here could be anything local that you would be working on in another timezone, for example concert times, apartment viewing times etc. etc. which are hard to work with today as the studio (just like the Date API) respects only the browsers relative display of the UTC string.

We have lots of signal on this one in [this thread](https://github.com/sanity-io/sanity/issues/1923), as well as directly from users.

### What to review

* The naming of `displayTimezone` is worth a discussion.
* The UI that I have chosen, with the text below should also be taken into consideration as to whether that is the best approach.
* I have decided to go forward with the datetime tooling within each of the relevant packages and then I have included their timezome handling counterpart (for `moment`, it is `moment-timezone` and for `date-fns`, it is `date-fns-tz`). I don't believe that this task is big enough to warrant us going down the route of removing these in favour of, for example the [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) API, but I hope that we do that before long.

### Testing

I added two tests where I thought they should go that take account for this being in place and for Winter and Summer timezones. Happy to add more should we sit fit!

### Notes for release

Might actually be worth mentioning: "The ability to specify the display timezone for your `datetime` fields".

### DIFF
--- a/packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
@@ -1,7 +1,9 @@
+import {Stack, Text} from '@sanity/ui'
 import {
   DEFAULT_DATE_FORMAT,
   DEFAULT_TIME_FORMAT,
   format,
+  isValidTimezoneString,
   parse,
 } from '@sanity/util/legacyDateFormat'
 import {getMinutes, parseISO, setMinutes} from 'date-fns'
@@ -19,12 +21,14 @@ interface ParsedOptions {
   dateFormat: string
   timeFormat: string
   timeStep: number
+  displayTimezone?: string
 }
 
 interface SchemaOptions {
   dateFormat?: string
   timeFormat?: string
   timeStep?: number
+  displayTimezone?: string
 }
 
 /**
@@ -37,19 +41,28 @@ function parseOptions(options: SchemaOptions = {}): ParsedOptions {
     dateFormat: options.dateFormat || DEFAULT_DATE_FORMAT,
     timeFormat: options.timeFormat || DEFAULT_TIME_FORMAT,
     timeStep: ('timeStep' in options && Number(options.timeStep)) || 1,
+    displayTimezone:
+      options.displayTimezone && isValidTimezoneString(options.displayTimezone)
+        ? options.displayTimezone
+        : undefined,
   }
 }
 
 function serialize(date: Date) {
   return date.toISOString()
 }
-function deserialize(isoString: string): ParseResult {
-  const deserialized = new Date(isoString)
-  if (isValidDate(deserialized)) {
-    return {isValid: true, date: deserialized}
+
+// deserialize specifically handles the deserialisation of the iso date string from content lake
+const deserialize =
+  (timezone?: string) =>
+  (isoString: string): ParseResult => {
+    // create a date object respecting the timezone
+    const {date: deserialized} = parse(isoString, undefined, timezone)
+    if (deserialized && isValidDate(deserialized)) {
+      return {isValid: true, date: deserialized}
+    }
+    return {isValid: false, error: `Invalid date value: "${isoString}"`}
   }
-  return {isValid: false, error: `Invalid date value: "${isoString}"`}
-}
 
 // enforceTimeStep takes a dateString and datetime schema options and enforces the time
 // to be within the configured timeStep
@@ -65,7 +78,8 @@ function enforceTimeStep(dateString: string, timeStep: number) {
     return serialize(setMinutes(date, minutes - leftOver))
   }
 
-  return serialize(date)
+  const serilalised = serialize(date)
+  return serilalised
 }
 
 /**
@@ -74,7 +88,7 @@ function enforceTimeStep(dateString: string, timeStep: number) {
 export function DateTimeInput(props: DateTimeInputProps) {
   const {onChange, schemaType, value, elementProps} = props
 
-  const {dateFormat, timeFormat, timeStep} = parseOptions(schemaType.options)
+  const {dateFormat, timeFormat, timeStep, displayTimezone} = parseOptions(schemaType.options)
   const {t} = useTranslation()
 
   const handleChange = useCallback(
@@ -90,28 +104,37 @@ export function DateTimeInput(props: DateTimeInputProps) {
   )
 
   const formatInputValue = useCallback(
-    (date: Date) => format(date, `${dateFormat} ${timeFormat}`),
-    [dateFormat, timeFormat],
+    (date: Date) => format(date, `${dateFormat} ${timeFormat}`, {timezone: displayTimezone}),
+    [dateFormat, timeFormat, displayTimezone],
   )
 
   const parseInputValue = useCallback(
-    (inputValue: string) => parse(inputValue, `${dateFormat} ${timeFormat}`),
-    [dateFormat, timeFormat],
+    (inputValue: string): ParseResult =>
+      parse(inputValue, `${dateFormat} ${timeFormat}`, displayTimezone),
+    [dateFormat, timeFormat, displayTimezone],
   )
   const calendarLabels: CalendarLabels = useMemo(() => getCalendarLabels(t), [t])
   return (
-    <CommonDateTimeInput
-      {...elementProps}
-      calendarLabels={calendarLabels}
-      onChange={handleChange}
-      deserialize={deserialize}
-      formatInputValue={formatInputValue}
-      parseInputValue={parseInputValue}
-      placeholder={schemaType.placeholder}
-      selectTime
-      serialize={serialize}
-      timeStep={timeStep}
-      value={value}
-    />
+    <Stack space={3}>
+      <CommonDateTimeInput
+        {...elementProps}
+        calendarLabels={calendarLabels}
+        onChange={handleChange}
+        deserialize={deserialize(displayTimezone)}
+        formatInputValue={formatInputValue}
+        parseInputValue={parseInputValue}
+        placeholder={schemaType.placeholder}
+        selectTime
+        serialize={serialize}
+        timeStep={timeStep}
+        timezone={displayTimezone}
+        value={value}
+      />
+      {displayTimezone ? (
+        <Text size={1} muted>
+          {t('inputs.datetime.timezone-information-text', {timezone: displayTimezone})}
+        </Text>
+      ) : null}
+    </Stack>
   )
 }
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -544,6 +544,8 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'inputs.boolean.disabled': 'Disabled',
   /** Placeholder value for datetime input */
   'inputs.datetime.placeholder': 'e.g. {{example}}',
+  /** Indicating timezone when a valid displayTimezone is set in the options of the input */
+  'inputs.datetime.timezone-information-text': 'Time is in {{timezone}} timezone',
   /** Acessibility label for button to open file options menu */
   'inputs.file.actions-menu.file-options.aria-label': 'Open file options menu',
   /** Browse */
--- a/packages/@sanity/types/src/schema/definition/type/datetime.ts
+++ b/packages/@sanity/types/src/schema/definition/type/datetime.ts
@@ -8,6 +8,7 @@ export interface DatetimeOptions extends BaseSchemaTypeOptions {
   dateFormat?: string
   timeFormat?: string
   timeStep?: number
+  displayTimezone?: string
 }
 
 /** @public */
--- a/packages/@sanity/util/package.json
+++ b/packages/@sanity/util/package.json
@@ -125,6 +125,7 @@
     "@sanity/types": "3.68.3",
     "get-random-values-esm": "1.0.2",
     "moment": "^2.30.1",
+    "moment-timezone": "^0.5.46",
     "rxjs": "^7.8.1"
   },
   "devDependencies": {
--- a/packages/@sanity/util/src/legacyDateFormat.ts
+++ b/packages/@sanity/util/src/legacyDateFormat.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-shadow */
-import moment from 'moment'
+import moment from 'moment-timezone'
 
 export const DEFAULT_DATE_FORMAT = 'YYYY-MM-DD'
 export const DEFAULT_TIME_FORMAT = 'HH:mm'
@@ -12,16 +12,29 @@ export type ParseResult = {isValid: boolean; date?: Date; error?: string} & (
 // todo: find a way to get rid of moment there.
 // note: the format comes from peoples schema types, so we need to deprecate it for a while and
 // find a way to tell people that they need to change it
-export function format(input: Date, format: string, useUTC = false) {
+export function format(
+  input: Date,
+  format: string,
+  options: {useUTC?: boolean; timezone?: string} = {useUTC: false, timezone: 'UTC'},
+) {
+  const {useUTC, timezone} = options
   if (useUTC) return moment.utc(input).format(format)
-
-  return moment(input).format(format)
+  const m = timezone && isValidTimezoneString(timezone) ? moment(input).tz(timezone) : moment(input)
+  return m.format(format)
 }
 
-export function parse(dateString: string, format: string): ParseResult {
-  const parsed = moment(dateString, format, true)
+export function parse(dateString: string, format?: string, timezone?: string): ParseResult {
+  const parsed =
+    timezone && isValidTimezoneString(timezone)
+      ? moment(dateString, format, true).tz(timezone)
+      : moment(dateString, format, true)
+
   if (parsed.isValid()) {
     return {isValid: true, date: parsed.toDate()}
   }
   return {isValid: false, error: `Invalid date. Must be on the format "${format}"`}
+}
+
+export function isValidTimezoneString(timezone: string) {
+  return moment.tz.names().includes(timezone)
 }
--- a/packages/sanity/src/core/form/inputs/DateInputs/CommonDateTimeInput.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/CommonDateTimeInput.tsx
@@ -28,6 +28,7 @@ export interface CommonDateTimeInputProps {
   selectTime?: boolean
   serialize: (date: Date) => string
   timeStep?: number
+  timezone?: string
   value: string | undefined
   calendarLabels: CalendarLabels
 }
@@ -96,7 +97,6 @@ export const CommonDateTimeInput = forwardRef(function CommonDateTimeInput(
     forwardedRef,
     () => ref.current,
   )
-
   const parseResult = localValue ? parseInputValue(localValue) : value ? deserialize(value) : null
 
   const inputValue = localValue
@@ -122,6 +122,7 @@ export const CommonDateTimeInput = forwardRef(function CommonDateTimeInput(
       }
       ref={ref}
       value={parseResult?.date}
+      timezone={props.timezone}
       inputValue={inputValue || ''}
       readOnly={Boolean(readOnly)}
       onInputChange={handleDatePickerInputChange}
--- a/packages/sanity/src/core/form/inputs/DateInputs/__tests__/DateTimeInput.test.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/__tests__/DateTimeInput.test.tsx
@@ -69,3 +69,41 @@ test('formatting of deserialized value', async () => {
   // const {textInput} = renderInput({value: '2021-03-28T17:23:00.000Z'} as any)
   expect(input.value).toBe('2021-03-28 10:23')
 })
+
+test('time is shown in the display timezone if specified (utc+1 winter)', async () => {
+  const {result} = await renderStringInput({
+    fieldDefinition: defineField({
+      type: 'datetime',
+      name: 'test',
+      options: {displayTimezone: 'Europe/Oslo'},
+    }),
+    props: {documentValue: {test: '2021-01-15T12:00:00.000Z'}},
+    render: (inputProps) => <DateTimeInput {...inputProps} />,
+  })
+
+  const input = result.container.querySelector('input')!
+
+  userEvent.type(input, '2021-03-28 13:00')
+  expect(input.value).toBe('2021-03-28 13:00')
+  fireEvent.blur(input)
+  expect(input.value).toBe('2021-03-28 13:00')
+})
+
+test('time is shown in the display timezone if specified (utc+2 summer)', async () => {
+  const {result} = await renderStringInput({
+    fieldDefinition: defineField({
+      type: 'datetime',
+      name: 'test',
+      options: {displayTimezone: 'Europe/Oslo'},
+    }),
+    props: {documentValue: {test: '2021-06-15T12:00:00.000Z'}},
+    render: (inputProps) => <DateTimeInput {...inputProps} />,
+  })
+
+  const input = result.container.querySelector('input')!
+
+  userEvent.type(input, '2021-06-15 14:00')
+  expect(input.value).toBe('2021-06-15 14:00')
+  fireEvent.blur(input)
+  expect(input.value).toBe('2021-06-15 14:00')
+})
--- a/packages/sanity/src/ui-components/inputs/DateInputs/DatePicker.tsx
+++ b/packages/sanity/src/ui-components/inputs/DateInputs/DatePicker.tsx
@@ -9,6 +9,7 @@ export const DatePicker = forwardRef(function DatePicker(
     onChange: (nextDate: Date) => void
     selectTime?: boolean
     timeStep?: number
+    timezone?: string
     calendarLabels: CalendarLabels
   },
   ref: ForwardedRef<HTMLDivElement>,
@@ -32,6 +33,7 @@ export const DatePicker = forwardRef(function DatePicker(
       selectedDate={value}
       onSelect={handleSelect}
       focusedDate={focusedDate || value}
+      timezone={props.timezone}
       onFocusedDateChange={setFocusedDay}
     />
   )
--- a/packages/sanity/src/ui-components/inputs/DateInputs/DateTimeInput.tsx
+++ b/packages/sanity/src/ui-components/inputs/DateInputs/DateTimeInput.tsx
@@ -28,6 +28,7 @@ export interface DateTimeInputProps {
   readOnly?: boolean
   selectTime?: boolean
   timeStep?: number
+  timezone?: string
   value?: Date
   calendarLabels: CalendarLabels
   constrainSize?: boolean
@@ -115,6 +116,7 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
                       calendarLabels={calendarLabels}
                       selectTime={selectTime}
                       timeStep={timeStep}
+                      timezone={props.timezone}
                       onKeyUp={handleKeyUp}
                       value={value}
                       onChange={onChange}
--- a/packages/sanity/src/ui-components/inputs/DateInputs/calendar/Calendar.tsx
+++ b/packages/sanity/src/ui-components/inputs/DateInputs/calendar/Calendar.tsx
@@ -1,6 +1,7 @@
 import {ChevronLeftIcon, ChevronRightIcon} from '@sanity/icons'
 import {Box, Flex, Grid, Select, Text} from '@sanity/ui'
 import {addDays, addMonths, setDate, setHours, setMinutes, setMonth, setYear} from 'date-fns'
+import {format as formatTZ, utcToZonedTime, zonedTimeToUtc} from 'date-fns-tz'
 import {range} from 'lodash'
 import {
   type ComponentProps,
@@ -22,10 +23,44 @@ import {type CalendarLabels, type MonthNames} from './types'
 import {formatTime} from './utils'
 import {YearInput} from './YearInput'
 
+const format = (date: Date, formatStr: string, options?: {timezone?: string}) => {
+  if (!options?.timezone) {
+    const dateParts = date
+      .toLocaleString('en-US', {
+        day: '2-digit',
+        month: '2-digit',
+        year: 'numeric',
+      })
+      .split('/')
+
+    switch (formatStr) {
+      case 'DD':
+        return dateParts[1]
+      case 'MM':
+        return dateParts[0]
+      default:
+        return dateParts[2]
+    }
+  }
+
+  const zonedDate = utcToZonedTime(date, options.timezone)
+  const formatPattern =
+    {
+      DD: 'dd',
+      MM: 'MM',
+      YYYY: 'yyyy',
+    }[formatStr] || 'yyyy'
+
+  return formatTZ(zonedDate, formatPattern, {
+    timeZone: options.timezone,
+  })
+}
+
 type CalendarProps = Omit<ComponentProps<'div'>, 'onSelect'> & {
   selectTime?: boolean
   selectedDate?: Date
   timeStep?: number
+  timezone?: string
   onSelect: (date: Date) => void
   focusedDate: Date
   onFocusedDateChange: (index: Date) => void
@@ -62,10 +97,31 @@ export const Calendar = forwardRef(function Calendar(
     selectedDate = new Date(),
     focusedDate = selectedDate,
     timeStep = 1,
+    timezone,
     onSelect,
     labels,
     ...restProps
   } = props
+
+  const getDisplayHour = useCallback(() => {
+    if (!timezone) return selectedDate.getHours()
+    const zonedDate = utcToZonedTime(selectedDate, timezone)
+    return zonedDate.getHours()
+  }, [selectedDate, timezone])
+
+  const getDisplayMinutes = useCallback(() => {
+    if (!timezone) return selectedDate.getMinutes()
+    const zonedDate = utcToZonedTime(selectedDate, timezone)
+    return zonedDate.getMinutes()
+  }, [selectedDate, timezone])
+
+  const getDisplayMonth = useCallback(() => {
+    return Number(format(selectedDate, 'MM', {timezone})) - 1 // month is 0-indexed
+  }, [selectedDate, timezone])
+
+  const getDisplayYear = useCallback(() => {
+    return Number(format(selectedDate, 'YYYY', {timezone}))
+  }, [selectedDate, timezone])
 
   const setFocusedDate = useCallback(
     (date: Date) => onFocusedDateChange(date),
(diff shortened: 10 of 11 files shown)


packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -65,7 +80,8 @@ function enforceTimeStep(dateString: string, timeStep: number) {
     return serialize(setMinutes(date, minutes - leftOver))
   }
 
-  return serialize(date)
+  const serilalised = serialize(date)
+  return serilalised
```
rexxars, line 84: Bit of a mismatch between US and UK spelling here, could we just return the value of `serialize`?

```suggestion
  return serialize(date)
```

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -37,19 +43,28 @@ function parseOptions(options: SchemaOptions = {}): ParsedOptions {
...
 function serialize(date: Date) {
   return date.toISOString()
 }
-function deserialize(isoString: string): ParseResult {
-  const deserialized = new Date(isoString)
-  if (isValidDate(deserialized)) {
-    return {isValid: true, date: deserialized}
+
+// deserialize specifically handles the deserialisation of the iso date string from content lake
+const deserialize =
```
rexxars, line 58: I feel like this should be named `getDeserializer` or similar - it read as a mistake to me when I later on read `deserialize(timezone)`

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +106,46 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
+        ) : (
+          ''
+        )
+      }
+    >
+      <CommonDateTimeInput
+        {...elementProps}
+        calendarLabels={calendarLabels}
+        onChange={handleChange}
+        deserialize={deserialize(displayTimezone)}
```
rexxars, line 139: Unsure if the react compiler catches this - if not this would benefit from being memoized

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +106,46 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
+        displayTimezone ? (
+          <Inline space={1}>
+            <EarthGlobeIcon />
+            <Text size={1} muted>
+              {t('inputs.datetime.timezone-information-text', {timezone: displayTimezone})}
+            </Text>
+          </Inline>
+        ) : (
+          ''
+        )
```
rexxars, line 132: Would it be better to not wrap the input in the tooltip if there's no display timezone? Minor, but seems like it might add unnecessary calculation of content/tooltip location etc.

rexxars (dismissed):
This is a great addition. Could you add an example to the `datetimeTest` schema type in the test studio so we can more easily test this on a preview deployment?

NAD (Not A Designer), but from my perspective I'd love it if there was some way to reveal more information.
- If the users' timezone is not the same as the display timezone, having a way to show what the chosen time is in "your" timezone can be helpful.
- Potentially showing what the stored UTC time is (possibly also hinting that this is the "stored value")

I'm pondering if it might make sense to _always_ show the timezone being displayed, and hide the additional information I mentioned behind a tooltip/info icon or similar. Currently (in next) I could see people putting in a date and forgetting that they might be in a different timezone than what they usually are and end up with an incorrect value being stored.

EoinFalconer:
> This is a great addition. Could you add an example to the `datetimeTest` schema type in the test studio so we can more easily test this on a preview deployment?
> 
> NAD (Not A Designer), but from my perspective I'd love it if there was some way to reveal more information.
> 
> * If the users' timezone is not the same as the display timezone, having a way to show what the chosen time is in "your" timezone can be helpful.
> * Potentially showing what the stored UTC time is (possibly also hinting that this is the "stored value")
> 
> I'm pondering if it might make sense to _always_ show the timezone being displayed, and hide the additional information I mentioned behind a tooltip/info icon or similar. Currently (in next) I could see people putting in a date and forgetting that they might be in a different timezone than what they usually are and end up with an incorrect value being stored.

Thanks for the feedback! 🙌 I think I'll run this by a designer as you bring up good points!

Would be good to get a stamp of approval on it, but I think given the use cases in [this thread](https://github.com/sanity-io/sanity/issues/1923) whereby you are inputting data for a specific show at an opera house or property viewing that is very local you always just want it to be fixed, so I'm not sure I'm seeing the need to change the timezone use case (maybe this could be added in the future with a `allowTimezonePicker` option, or even you can return a list of timezones that you would like it to be shown in if you require multiple for your business and don't want to store the date in many fields to achieve this.

Also not sure why an editor would want to know the utc time? Could be a useful devtool thing though if you don't want to have to go to vision to see your data?

rexxars:
> I'm not sure I'm seeing the need to change the timezone use case (maybe this could be added in the future with a `allowTimezonePicker` option

Sorry, I wasn't being precise in my feedback; the thing I was trying to address was that you don't know what timezone the date you are entering represents. Are you entering a UTC date? Your local timezone? The timezone the developer chose when they configured the schema?

Your PR addresses the last one; but only if you've set a display timezone. I am asking myself if seeing the timezone is so helpful that it should _always_ be visible, even if the developer has not set a timezone explicitly. If no display timezone is set, it would show the users' local timezone, as that is what the default behavior is. Does that make sense?

> Also not sure why an editor would want to know the utc time? Could be a useful devtool thing though if you don't want to have to go to vision to see your data?

Yeah, I may be overindexing on myself as a developer here. I keep seeing people stumble when they try to use the date in their frontends, thinking that the date stored is the same as the one they are seeing, but surfacing the UTC date to all users because of this is probably excessive.

bobinska-dev:
I can add some more context on the use cases and customer needs here: https://sanity-io.slack.com/archives/C3Q637VGW/p1737038508203349

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+      {/* Dialog */}
+      {DialogTimeZone && <DialogTimeZone {...dialogProps} />}
+
+      <Tooltip content={tooltipContent} portal>
+        <div>
+          {/*
+          If `useElementQueries` is enabled, dates will be conditionally toggled at different element
+          breakpoints - provided this `<ButtonTimeZone>` is wrapped in a `<ButtonTimeZoneElementQuery>` component.
+        */}
+          {useElementQueries ? (
```
pedrobonamin, line 33: Do we need this option? It is true in all the consumers, so no one uses the second second part of this.

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    >{`${timeZone.alternativeName} (${timeZone.namePretty})`}</Text>
+                  </Inline>
+                ) : (
+                  <Button
+                    data-testid="timezone-button"
+                    icon={EarthAmericasIcon}
+                    mode="bleed"
+                    readOnly={!allowTimeZoneSwitch}
+                    onClick={allowTimeZoneSwitch ? dialogTimeZoneShow : noop}
+                    text={`${timeZone.alternativeName} (${timeZone.namePretty})`}
```
pedrobonamin, line 72: Is this reverted? 
If it allows timezone switch it is rendering the text instead of the button.

packages/sanity/src/core/components/timeZone/DialogTimeZone.tsx
```diff
@@ -3,44 +3,65 @@ import {Autocomplete, Card, Flex, Inline, Stack, Text, type Theme} from '@sanity
...
-  const {onClose} = props
-
-  const {setTimeZone, timeZone} = useTimeZone()
+  const {onClose, timeZoneScope} = props
+  const {setTimeZone, allTimeZones, timeZone, getLocalTimeZone, getTimeZone} =
+    useTimeZone(timeZoneScope)
   const [selectedTz, setSelectedTz] = useState<NormalizedTimeZone | undefined>(timeZone)
+  const {t} = useTranslation('studio')
+
+  // Differend text based on different scopes
```
pedrobonamin, line 45:
```suggestion
  // Different text based on different scopes
```

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +171,119 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
+                />
+              </Stack>
+            }
+          />
+        </Box>
+        <CenterAlignedBox paddingX={3} paddingY={1}>
+          <FormFieldStatus maxAvatars={1} position="top">
+            {/*<FieldPresence maxAvatars={1} presence={presence} />*/}
+          </FormFieldStatus>
+        </CenterAlignedBox>
```
pedrobonamin, line 280: Is presence working? Do we need to include this?

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +171,119 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
+      timeStep,
+      value,
+      timeZoneScope,
+    ],
+  )
+
+  const input = (
+    <>
+      <CommonDateTimeInput {...commonProps} />
+    </>
```
pedrobonamin, line 222: Do we need this input defined here? Looks like a left over from a previous implementation, it is only being used in one place, maybe better to move it into the JSX return

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -1800,6 +1800,25 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'sheet-list.select-fields': 'Select up to 5 field types',
   /** Accessibility label for the navbar status button */
   'status-button.aria-label': 'Configuration status',
+
+  /* Time zone handling */
+  'time-zone.action.search-for-timezone-placeholder': 'Search for a city or time zone',
+  'time-zone.action.select-local-time-zone': 'Select local time zone',
```
pedrobonamin, line 1806: It is better if we can provide for each item added into the locales a description on top of it, this will be used by the automatic ai process to build the translation for the rest of the languages, also, it's the only thing users which are reviewing the translation repo will get to see to try to understand where this translation is being used

packages/sanity/src/core/scheduledPublishing/constants.tsx
```diff
@@ -6,8 +6,6 @@ import {
   type ScheduleState,
 } from './types'
 
-export const LOCAL_STORAGE_TZ_KEY = 'scheduled-publishing::time-zone'
```
pedrobonamin, line 9: Seeing that we removed this, will this mean that if I had previously set a timezone for scheduled publishing now I will need to set it again? 
Do you think it will be complicated to add a fallback for this key if Ive been using scheduled publishing?

pedrobonamin (commented):
Great work, love it!

It's great to see how this will work, a very nice step forward. 
Added a couple of thoughts and suggestions, and was also wondering, is there any reason why we are not using the key value store instead of local storage to store the timezones?

packages/@sanity/util/src/datetime-formatter/formatter.ts
RitaDias, line 1: My previous experiences with date and date formats make me feel a bit nervous about this approach.
The question is whether do we _really_ need to re-invent the wheel when it comes to formatting?

packages/sanity/src/core/hooks/useTimeZone.tsx
```diff
@@ -0,0 +1,334 @@
...
+  const dateToUse = relativeDateForZones ?? new Date()
+  const parts = formatter.formatToParts(dateToUse)
+  const shortParts = shortFormatter.formatToParts(dateToUse)
+  const rawOffset = formatInTimeZone(dateToUse, canonicalIdentifier, 'xxx')
+  // If the offset is +02:00 then we can just show +2, if it has +13:45 then we should show +13:45, remove the leading +0 and just leave a + if a number under 10, remove the :00 at the end
+  const offset = rawOffset
+    .replace(/([+-])0(\d)/, '$1$2')
+    .replace(/([+-])0$/, '$1')
+    .replace(/:00$/, '')
+    .replace(/[+]0$/, '')
```
RitaDias, line 85: Is this related to formatting? Or something different? 🤔 
Trying ot understand why this needs to be here 🤔

packages/sanity/src/core/hooks/useTimeZone.tsx
RitaDias, line 1: I have to be honest, this whole component feels incredibly complex for what it should but I think I might be missing something. Do we need these offsets informaiton like for example

const offsetToMinutes = (offset: string): number => {

packages/sanity/src/core/form/field/actions/FieldActionsProvider.tsx
```diff
@@ -17,7 +28,9 @@ type FieldActionsProviderProps = PropsWithChildren<{
 export const FieldActionsProvider = memo(function FieldActionsProvider(
   props: FieldActionsProviderProps,
 ) {
-  const {actions, children, path, focused} = props
+  // by passing the comments and slot here, we can wrap this functionality around any custom field without needing to confirm the title and description
```
RitaDias, line 31: huh, that's odd, do we know why?

RitaDias (dismissed):
Hi Eoin!
Thank you so much for going through and making this change. I made some comments through and Pedro also has some very valuable feedback that should be answered.

I also noticed this strange behaviour while trying it out **however, it seems this is not introduced on this PR as it's happening on main as well**. So let's keep this outside of scope. I have created a separate story for this

https://github.com/user-attachments/assets/9c3a3e3d-3135-4033-8ee3-87210efa1016

packages/sanity/src/core/form/field/actions/FieldActionsProvider.tsx
EoinFalconer: It seems like the FieldActionsProvider wasn't accommodating for more custom inputs like the boolean input, I wanted to be able to give them this functionality but still be able to create my own title (which is needed for the timezone input bit). This meant that the boolean input currently doesn't have the ability to be commented on or have a set of actions.

packages/sanity/src/core/hooks/useTimeZone.tsx
EoinFalconer: This component is quite complicated and I would love to talk about ways of making it less complex. Here are the complicating factors from my point of view:
* Caching timezones is a good idea because the calculation of that list is a pain, but it is nice to be able to just get all the timezones at any point in the code so that the list just works.
* This `offsetToMinutes` is because we would like to sort the list by the minutes so that it is easy to navigate in the popup as the browser doesn't return the timezones in that order.
* `relativeDate` is a massive complication on the whole timezone thing. Basically if I have a value that is in European Summer Time and one that is in Winter Time the list itself looks completely different. So the entire list of available timezones needs to change based on this value.

packages/sanity/src/core/hooks/useTimeZone.tsx
EoinFalconer: This is related to formatting, we do this here. Apparently some timezones are off by minutes and not just hours 🙄

packages/sanity/src/core/scheduledPublishing/types.ts
```diff
@@ -3,7 +3,7 @@ import {type ValidationMarker} from '@sanity/types'
 export interface NormalizedTimeZone {
   abbreviation: string
   alternativeName: string
-  mainCities: string
+  city: string
```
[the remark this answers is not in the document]
EoinFalconer: This is only used for display purpose in the list, so not a breaking change 😊 It also makes search much for user friendly in the list as it might be hard to see your city in the returned results even if you search for it due to many cities having the same timezone.

packages/sanity/src/core/scheduledPublishing/tool/contexts/schedules.tsx
```diff
@@ -35,7 +35,7 @@ function SchedulesProvider({
   }
 }) {
   const [sortBy, setSortBy] = useState<ScheduleSort>(value.sortBy || 'executeAt')
-  const {utcToCurrentZoneDate} = useTimeZone()
+  const {utcToCurrentZoneDate} = useTimeZone({type: 'scheduledPublishing'})
```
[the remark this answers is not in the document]
EoinFalconer: This is definitely something I should do, thank you!

packages/sanity/src/core/scheduledPublishing/types.ts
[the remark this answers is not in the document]
RitaDias: Great news then! :)

packages/sanity/src/core/components/inputs/DateInputs/DatePicker.tsx
```diff
@@ -12,8 +13,9 @@ export const DatePicker = forwardRef(function DatePicker(
     calendarLabels: CalendarLabels
     monthPickerVariant?: CalendarProps['monthPickerVariant']
     padding?: number
-    showTimezone?: boolean
+    showTimeZone?: boolean
```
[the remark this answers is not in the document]
EoinFalconer: Oh! Thanks! I'll go back to timezone being one word.

packages/sanity/src/core/components/inputs/DateInputs/calendar/Calendar.tsx
```diff
@@ -51,8 +43,9 @@ export type CalendarProps = Omit<ComponentProps<'div'>, 'onSelect'> & {
   labels: CalendarLabels
   monthPickerVariant?: (typeof MONTH_PICKER_VARIANT)[keyof typeof MONTH_PICKER_VARIANT]
   padding?: number
-  showTimezone?: boolean
+  showTimeZone?: boolean
```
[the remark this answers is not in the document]
EoinFalconer: Thanks!

packages/sanity/src/core/tasks/components/form/fields/DateEditFormField.tsx
```diff
@@ -83,6 +83,7 @@ export function DateEditFormField(props: {
               onKeyUp={handleKeyUp}
               value={deserializedValue.date}
               onChange={handleChange}
+              timeZoneScope={{type: 'scheduledPublishing'} as const}
```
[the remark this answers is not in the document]
EoinFalconer: Yeah, so the `scheduledPublishing` here is essentially part of the key in the keypair value that is storing your preference. I'll move out the `as const` stuff.

packages/@sanity/util/src/datetime-formatter/formatter.ts
```diff
@@ -0,0 +1,299 @@
...
+ * Returns an English ordinal for a given day number
+ */
+function getOrdinal(day: number): string {
+  const j = day % 10
+  const k = day % 100
+  if (j === 1 && k !== 11) return `${day}st`
+  if (j === 2 && k !== 12) return `${day}nd`
+  if (j === 3 && k !== 13) return `${day}rd`
+  return `${day}th`
+}
```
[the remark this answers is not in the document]
EoinFalconer: Yeah, I had looked into it but it seems that this is just how [moment works](https://github.com/moment/luxon/issues/1411) unfortunately, and I didn't want to change that behavior.

packages/sanity/src/core/scheduledPublishing/constants.tsx
EoinFalconer: Shouldn't be too complicated, ideally I wouldn't have `scheduled-publishing` logic lying around outside of that part of the code so I'll see if I can make that the case 💯

packages/sanity/src/core/i18n/bundles/studio.ts
EoinFalconer: Very nice, didn't know that! Thank you!

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
EoinFalconer: Great question, I just kind of accepted it being a thing, I'll see if ti is ever used.

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const {timeZone} = useTimeZone(timeZoneScope)
+  const {DialogTimeZone, dialogProps, dialogTimeZoneShow} = useDialogTimeZone(timeZoneScope)
+
+  return (
+    <>
+      {/* Dialog */}
+      {DialogTimeZone && <DialogTimeZone {...dialogProps} />}
+
+      <Tooltip content={tooltipContent} portal>
+        <div>
```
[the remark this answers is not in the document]
EoinFalconer: Thanks!

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
EoinFalconer: Wow yeah, I suppose the duplication of all this code here made that easy to miss, thanks for catching!

packages/sanity/src/core/scheduledPublishing/constants.tsx
EoinFalconer: I looked through now, I would really like to keep the storage key for the timezone as something that is derived from the `timeZoneScope` rather than being passed all over the place -> wdyt?

pedrobonamin: Wdyt if we use this key instead of `timeZone.scheduledPublishing` for the schedule publishing scope? 
So, in `useTimeZone` 

```diff
+  const localStorageId = scope.type === "scheduledPublishing" ?
+    'scheduled-publishing::time-zone' :
     'id' in scope
       ? `${timeZoneLocalStorageNamespace}${scope.type}.${scope.id}`
       : `${timeZoneLocalStorageNamespace}${scope.type}`

```

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -47,7 +47,9 @@ export const ReleaseScheduleButton = ({
   const {t} = useTranslation(releasesLocaleNamespace)
   const {t: tCore} = useTranslation()
   const telemetry = useTelemetry()
-  const {utcToCurrentZoneDate, zoneDateToUtc} = useTimeZone()
+  // in the releases tool we want timezone to be saved for releases
```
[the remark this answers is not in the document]
EoinFalconer: I wanted to avoid having a global timezone as it felt annoying for this use case if you are working across for example, event that go across multiple timezones that you would need to do lots of switching. Also the chances that you are using but scheduled publishing and releases together is slim and not something to index on.

packages/sanity/src/core/hooks/useTimeZone.tsx
```diff
@@ -0,0 +1,334 @@
...
+import {DATE_FORMAT} from '../scheduledPublishing/constants'
+import {type NormalizedTimeZone} from '../scheduledPublishing/types'
+import {debugWithName} from '../scheduledPublishing/utils/debug'
+import getErrorMessage from '../scheduledPublishing/utils/getErrorMessage'
+
+enum TimeZoneEvents {
+  update = 'timeZoneEventUpdate',
+}
+
+export type TimeZoneScopeType = 'scheduledPublishing' | 'contentReleases' | 'input'
```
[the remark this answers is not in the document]
EoinFalconer: I originally just had the scoping as being a string but this felt _really_ brittle. I think needing to add a new string to the type here makes it clear that we have a new part of the studio that needs the `useTimeZone` hook, but open to hear where we would break things up to and how we would ensure safe typing for the `useTimeZone` props.

packages/@sanity/util/src/datetime-formatter/formatter.ts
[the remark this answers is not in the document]
RitaDias: I see 🤔 
I guess my other concern that follows this is that one day moment might move on and we are still with this implementation that we now need to support / maintain. But this is a concern not so much about this particular part of the code but but this file as a whole :)

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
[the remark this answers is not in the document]
RitaDias: 🤔 I suppose that this is something we can return and fix in the future if we get complaints over.
I'm not the main user but I think I would likely find it more annoying the other way around, where I have to constantly switch and wonder which timezone each part of the tool is relying on now by double checking. Might be something minor though

packages/sanity/src/core/hooks/useTimeZone.tsx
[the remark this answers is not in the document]
RitaDias: Ah what I meant is if we need to have them broken up. I guess this goes hand in hand with the question of a global timezone vs smaller ones based on tools :)

packages/sanity/src/core/hooks/useTimeZone.tsx
RitaDias: This is likely a super massive nitpick I will admit. **I'm ok with keeping it as is** with the caveat that maintenance wise this might bite us in the ass later but maybe not! My loose thoughts about it:

The first point I agree!
the second and third I'm wondering if we are over indexing. 
The second point, I just wonder if it would be that annoying. In all instances where I had to scroll through a list I think that a search of that list would have likely been more useful than necessarily sorting it.
The third one, I'm not sure how other lists handle it but I'm wondering if they even do handle the nuances of Summer Time and Winter time 🤔

packages/@sanity/util/src/datetime-formatter/formatter.ts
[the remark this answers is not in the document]
RitaDias: Yep correct on that assumption.
I see what you mean 🤔 

I think it's ok to make an executive decision to keep your code as is with an eventual follow up task to likely explore if we should handle this (date formatting as a whole given the moment project) in a different way. Or maybe this discussion has already happened and I'm simply not aware of it! It feels like a way bigger discussion than the scope of this PR and I would prefer to not leave it hanging because of it. :)

packages/sanity/src/core/hooks/useTimeZone.tsx
RitaDias: That being said, I don't think it's a blocking issue, more of a more general question which we can return to later I think

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
RitaDias: It is working, and we've removed it!

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
RitaDias: Seeing that this is something that was already there before this PR I would suggest we don't add this to the scope of this (already large) PR. We can check this on a follow up :)

RitaDias:
@pedrobonamin 👋 
~~Quick heads up regarding the keyvaluestore and the keys, if you try to check if the values are saved (in our default workspace) it will fail because it will say that the keys don't exist. However, it's working in staging, so if you want to play around you can do it in the [staging workspace](https://test-studio-git-feat-sapp-1970.sanity.dev/)~~

~~I'm trying to figure out with Binoy and the SDK folks what I need to do to get this in the production env instead of just staging  :)~~
This has been updated

That being said, the code is ready for a final look over!

packages/sanity/src/core/components/inputs/DateInputs/calendar/Calendar.tsx
```diff
@@ -90,12 +83,21 @@ export const Calendar = forwardRef(function Calendar(
...
     ...restProps
   } = props
 
-  const {timeZone} = useTimeZone()
-  const {DialogTimeZone, dialogProps, dialogTimeZoneShow} = useDialogTimeZone()
+  const {timeZone} = useTimeZone(timeZoneScope)
+
+  const getDisplayMonth = useCallback(() => {
+    return Number(format(focusedDate, 'MM', {timeZone: timeZone.name})) - 1 // month is 0-indexed
+  }, [focusedDate, timeZone.name])
```
pedrobonamin, line 99: Small suggestion, the way this is setup now it will execute the function in each render, so it kinds of defeats the purpose of having a callback. I think it would be better to save it into a memo which returns the two values.

```suggestion
  const [displayMonth, displayYear] = useMemo(() => {
    return [
      // month is 0-indexed
      Number(format(focusedDate, 'MM', {timeZone: timeZone.name})) - 1,
      Number(format(focusedDate, 'YYYY', {timeZone: timeZone.name})),
    ]
  }, [focusedDate, timeZone.name])
```

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <>
+      {/* Dialog */}
+      {DialogTimeZone && <DialogTimeZone {...dialogProps} />}
+
+      <Tooltip
+        content={
+          <Box flex={1} padding={1}>
+            {tooltipContent}
+          </Box>
+        }
```
pedrobonamin: Could we wrap this content in `<Text>` ? 
```suggestion
        content={
          <Box flex={1} padding={1}>
            <Text size={1}>{tooltipContent}</Text>
          </Box>
        }
```

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  <Inline space={2} padding={2}>
+                    <Text weight={'medium'} size={1}>
+                      <EarthAmericasIcon />
+                    </Text>
+                    <Text
+                      weight={'medium'}
+                      size={1}
+                    >{`${timeZone.alternativeName} (${timeZone.namePretty})`}</Text>
+                  </Inline>
+                )}
```
pedrobonamin, line 81: This is probably coming from the previous implementation, but I think it should be possible to remove duplication here and below, they are the same, wdyt?

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -1,55 +1,96 @@
...
-import {type ParseResult} from './types'
 import {getCalendarLabels, isValidDate} from './utils'
 
+const Root = styled(Card)`
+  line-height: 1;
+`
+
+const CenterAlignedBox = styled(Box)`
+  align-self: center;
+`
```
pedrobonamin: This doesn't seem to be used, could we remove it?

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +171,112 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
   const formatInputValue = useCallback(
-    (date: Date) => format(date, `${dateFormat} ${timeFormat}`),
-    [dateFormat, timeFormat],
+    (date: Date) => format(date, `${dateFormat} ${timeFormat}`, {timeZone: timeZone.name}),
+    [dateFormat, timeFormat, timeZone],
   )
 
+  const deserialize = useMemo(() => getDeserializer(timeZone.name), [timeZone.name])
+  const memoizedSerialize = useMemo(() => serialize, [])
+
```
pedrobonamin: I think we don't need to memoize this function, `serialize` is defined outside of the component so it shouldn't be recreated on render and it will be an stable reference. I think passing it down  in the props as it is will be ok.

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -90,28 +171,112 @@ export function DateTimeInput(props: DateTimeInputProps) {
...
   const parseInputValue = useCallback(
-    (inputValue: string) => parse(inputValue, `${dateFormat} ${timeFormat}`),
-    [dateFormat, timeFormat],
+    (inputValue: string): ParseResult =>
+      parse(inputValue, `${dateFormat} ${timeFormat}`, timeZone.name),
+    [dateFormat, timeFormat, timeZone],
   )
   const calendarLabels: CalendarLabels = useMemo(() => getCalendarLabels(t), [t])
+  const commonProps = useMemo(
```
pedrobonamin: nit: this is now called `commonProps` but it is only passed to one component by spreading it. I think we could simplify this by passing the props directly into the component. Wdyt?

packages/sanity/src/core/releases/components/dialog/ReleaseForm.tsx
```diff
@@ -39,8 +40,8 @@ export function ReleaseForm(props: {
   const {onChange, value} = props
   const {releaseType} = value.metadata || {}
   const {t} = useTranslation()
-
-  const {timeZone, utcToCurrentZoneDate} = useTimeZone()
+  const timeZoneScope = CONTENT_RELEASES_TIME_ZONE_SCOPE
+  const {timeZone, utcToCurrentZoneDate} = useTimeZone(timeZoneScope)
```
pedrobonamin: NIT: 
```suggestion
  const {timeZone, utcToCurrentZoneDate} = useTimeZone(CONTENT_RELEASES_TIME_ZONE_SCOPE)
```

packages/sanity/src/core/releases/hooks/useCreateReleaseMetadata.ts
```diff
@@ -1,10 +1,12 @@
...
 import {useCallback} from 'react'
 
-import useTimeZone from '../../scheduledPublishing/hooks/useTimeZone'
+import {useTimeZone} from '../../hooks/useTimeZone'
+import {CONTENT_RELEASES_TIME_ZONE_SCOPE} from '../../studio/constants'
 
 export const useCreateReleaseMetadata = () => {
-  const {zoneDateToUtc} = useTimeZone()
+  const timeZoneScope = CONTENT_RELEASES_TIME_ZONE_SCOPE
+  const {zoneDateToUtc} = useTimeZone(timeZoneScope)
```
pedrobonamin: NIT: :) 
```suggestion
  const {zoneDateToUtc} = useTimeZone(CONTENT_RELEASES_TIME_ZONE_SCOPE)
```

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -47,7 +48,9 @@ export const ReleaseScheduleButton = ({
...
+  const timeZoneScope = useMemo(() => CONTENT_RELEASES_TIME_ZONE_SCOPE, [])
+  const {utcToCurrentZoneDate, zoneDateToUtc} = useTimeZone(timeZoneScope)
```
pedrobonamin: Sorry I'm repeating myself here. 
I think we should pass the `CONTENT_RELEASES_TIME_ZONE_SCOPE` value to the useTimeZone directly. and we won't need to memoize it.

```suggestion
  const {utcToCurrentZoneDate, zoneDateToUtc} = useTimeZone(CONTENT_RELEASES_TIME_ZONE_SCOPE)
```

packages/sanity/src/core/tasks/components/form/fields/DateEditFormField.tsx
```diff
@@ -83,6 +83,7 @@ export function DateEditFormField(props: {
               onKeyUp={handleKeyUp}
               value={deserializedValue.date}
               onChange={handleChange}
+              timeZoneScope={{type: 'scheduledPublishing'}}
```
pedrobonamin: Use here `SCHEDULED_PUBLISHING_TIME_ZONE_SCOPE`?

RitaDias: ah missed one! I'll do that!

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
RitaDias: I would be ok with doing it, I'm wondering if there is something to be said about memoizing these items to reduce renders though 🤔

packages/sanity/src/core/components/timeZone/timeZoneButton/TimeZoneButton.tsx
RitaDias: I removed the older one (the large one) 
@EoinFalconer if that is a mistake let me know!

packages/sanity/src/core/form/inputs/DateInputs/DateTimeInput.tsx
pedrobonamin: They are props and they are memoized themselves, so I think it should be ok to just pass it as is.
They are also all part of the memo dependency, so if any change the whole memo will change. I think we will have the same result but with less code

packages/sanity/src/core/hooks/useTimeZone.tsx
```diff
@@ -0,0 +1,372 @@
...
+  const [storedTimeZone, setStoredTimeZone] = useState<KeyValueStoreValue>(null)
+
+  useEffect(() => {
+    const subscription = keyValueStore
+      .getKey(keyStoreId)
+      .pipe(startWith(null))
+      .subscribe(setStoredTimeZone)
+
+    return () => subscription.unsubscribe()
+  }, [keyValueStore, keyStoreId])
```
pedrobonamin: Suggstion:

```suggestion
  const keyValueTimeZone$ = useMemo(
    () => keyValueStore.getKey(keyStoreId).pipe(startWith(null)),
    [keyValueStore, keyStoreId],
  )

  const storedTimeZone = useObservable(keyValueTimeZone$, null)
```

packages/sanity/src/core/hooks/useTimeZone.tsx
```diff
@@ -0,0 +1,372 @@
...
+  useEffect(() => {
+    const handler = () => {
+      setTimeZone(getInitialTimeZone())
+    }
+
+    window.addEventListener(TimeZoneEvents.update, handler)
+    return () => {
+      window.removeEventListener(TimeZoneEvents.update, handler)
+    }
+  }, [getInitialTimeZone, keyStoreId])
```
pedrobonamin: Why is keystore id a dependency here? 
I think it is not necessary. When `keyStoreId` changes `getStoredTimeZone` will change and that will have an effect on `getInitialTimeZone` so the end case shuold be the same.

EoinFalconer:
Thanks for getting this further @RitaDias, the key store value stuff here is really :chefs-kiss:!

pedrobonamin (dismissed):
This is such a nice improvement, thank you very much @EoinFalconer and @RitaDias !!

Let's be sure to talk with docs on getting something nice up for the next release, it's such a cool new feature.

pedrobonamin (approved):

pedrobonamin (approved):

PR #8192, opened 2025-01-06
fix: unset should unset readOnly date field

EoinFalconer:
### Description

At the moment if you create a custom input component for a `datetime` field or a `date` field, and perform `unset` in the custom input component when the field is `readOnly` the value will not update on the field.

This occurs because [this](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/form/inputs/DateInputs/CommonDateTimeInput.tsx)
Looks like it stores a local value, and when the field is read only, the input renders a plain text field which never triggers `handleDatePickerInputChange`.

In this PR, I have removed the plain text field and used the propagated `readOnly` flag to instead achieve this in the `DateTimeInput` component

### What to review

We think this is ok? I'm not sure of the reasons for using the `Text` component instead but this feels better to allow the `DateTimeInput` to be able to be read only, for usage in other scenarios too?

### Testing

Didn't add any tests for this, let me know if I should.

### Notes for release

`unset()` inside custom `datetime` `input` components will now shown the that the value has been unset in the Studio UI.

### DIFF
--- a/packages/sanity/src/core/form/inputs/DateInputs/CommonDateTimeInput.tsx
+++ b/packages/sanity/src/core/form/inputs/DateInputs/CommonDateTimeInput.tsx
@@ -1,6 +1,5 @@
 /* eslint-disable no-nested-ternary */
 
-import {TextInput} from '@sanity/ui'
 import {
   type FocusEvent,
   type ForwardedRef,
@@ -105,9 +104,7 @@ export const CommonDateTimeInput = forwardRef(function CommonDateTimeInput(
       ? formatInputValue(parseResult.date)
       : value
 
-  return readOnly ? (
-    <TextInput value={inputValue} readOnly />
-  ) : (
+  return (
     <DateTimeInput
       {...restProps}
       calendarLabels={props.calendarLabels}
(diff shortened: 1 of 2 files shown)


snorrees (approved):
This works well from my PoV!

Tested with the following input component, on the same document in two different windows to check multiplayer as well.

```
      components: {
        input: function T(props: DateInputProps) {
          const [ro, setRo] = useState(false)
          return (
            <Stack space={2}>
              {props.renderDefault({...props, readOnly: ro})}
              <Button
                text={`Make ${ro ? 'writable' : 'readOnly'}`}
                onClick={() => setRo((r) => !r)}
              />
              <Button text={`Unset field`} onClick={() => props.onChange(unset())} />
              <Button
                text={`Set field`}
                onClick={() => props.onChange(set(new Date().toISOString().split('T')[0]))}
              />
            </Stack>
          )
        },
      }
```

The input now updates in readonly mode.
Also tested datetime (without the split on the set button)

(TIL: we dont store the date input if it is not valid, so typing garbage in the filed does not sync)

PR #8196, opened 2025-01-06
test(e2e): add discard action test

RitaDias:
### Description

Add e2e test for the discard document action

### What to review

Does the test make sense, should we implement it in a different way?

### Notes for release

N/A, it's internal testing

### DIFF
--- a/test/e2e/tests/document-actions/discard.spec.ts
+++ b/test/e2e/tests/document-actions/discard.spec.ts
@@ -1,0 +1,51 @@
+import {expect} from '@playwright/test'
+import {test} from '@sanity/test'
+
+test(`document panel displays correct title after discarding changes`, async ({
+  page,
+  createDraftDocument,
+}) => {
+  /** ---------- START OF PUBLISH ---------- */
+
+  const title = 'Test Title'
+  const changedTitle = 'Title changed'
+
+  await createDraftDocument('/test/content/book')
+  await page.getByTestId('field-title').getByTestId('string-input').fill(title)
+
+  // Ensure the correct title is displayed before publishing.
+  await expect(page.getByTestId('document-panel-document-title')).toHaveText(title)
+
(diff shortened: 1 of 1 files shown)


test/e2e/tests/document-actions/discard.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  await page.getByTestId('field-title').getByTestId('string-input').fill(title)
+
+  // Ensure the correct title is displayed before publishing.
+  await expect(page.getByTestId('document-panel-document-title')).toHaveText(title)
+
+  // Focus the publish button to trigger the tooltip showing the keyboard shortcut
+  page.getByTestId('action-publish').focus()
+
+  // There is a delay before the tooltip opens, let's explicitly wait for it
+  await page.waitForTimeout(300)
```
pedrobonamin, line 23: Is it necessary to focus on the publish action? Not sure why we need to wait for the tooltip to open.

If we need the timeout to wait until the change has been saved, I think it will be a good idea to validate that the **Saved** shows in the footer.
Screenshot 2025-01-07 at 09 58 36

Or maybe we could we just depend on the following step `page.getByTestId('action-publish').click()` ?

pedrobonamin (approved):
Thanks for adding this Rita
Not blocking, but added a suggestion, would love your thoughts on this.

test/e2e/tests/document-actions/discard.spec.ts
RitaDias: (I think it wiser to not rely on the following step just by itself simply because I don't want to add more areas of potential flakiness which we know sometimes can happen in steps like those)

RitaDias:
Actually, closing this since I just now realised (when looking through the feedback given by Pedro) that we already have some discard (DiscardChanges) tests. It just wasn't where I expected. One of the tests there already tests this use case, so I'm closing this 🙏 
Sorry for the noise!

PR #8201, opened 2025-01-07
feat: add support for profiling CLI commands through SANITY_DEBUG_PROFILING

judofyr:
### Description

This adds support for a new environment variable (`SANITY_DEBUG_PROFILING=<DIRECTORY>`) which runs all the worker commands (`schema extract`, `manifest extract` etc) under the V8 profiler and dumps it for further inspection.

### What to review

Further testing that it works not only on my machine would be nice.

### Testing

I've tested it manually on a real-world Studio by running `npm link ../sanity/packages/sanity` and invoking the command.

### Notes for release

Not required (I think).


RitaDias:
I need some clarification on how to test this manually, sorry if this is obvious.
Where should I add `SANITY_DEBUG_PROFILING=<DIRECTORY>` (with a specific path I take it?) in the .env of the studio I'm using? What am I expected to find (I'm assuming a file with profiling info in the specific path provided)?

Because I did this but I wasn't able to find the cpu profile file, so I'm guess I did something wrong 😅

jordanl17:
Closing this, given that CLI is moving outside of monorepo

PR #8211, opened 2025-01-08
fix(sanity): fix race condition introduced by #8120

bjoerge:
### Description

#8120 introduced a regression that would occationally cause the document to be stuck at loading. Huge thanks @pedrobonamin for catching this! I believe this issue appeared because the `_keepalive` stream would start start pulling events from the pair listener immediately, but callers of memoizedPair will typically start subscribing to the draft.events and published.events streams a little later, and when they did, they sometimes wouldn’t get the listener events, because they have already been consumed by `_keepalive`.

This fixes the issue by partially reverting the PR to the way it was before #8120. Both the issue fixed by #8120 and the initial loading issues should be fixed by this.

### What to review

- The initial loading issue should be fixed
- The issue in #8120 should still be fixed (the steps to verify that one should still hold true)

### Testing
Manual testing for now

### Notes for release
n/a – the regression in #8120 didn't get released before this was discovered.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
@@ -2,7 +2,7 @@ import {type Action, type SanityClient} from '@sanity/client'
 import {type Mutation} from '@sanity/mutator'
 import {type SanityDocument} from '@sanity/types'
 import {omit} from 'lodash'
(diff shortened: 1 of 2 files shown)


pedrobonamin (approved):
Thanks for fixing this @bjoerge ! 
Tested with `assist` and it works.
Also tested [this](https://github.com/sanity-io/sanity/pull/8120) and it works. 

🚢

PR #8214, opened 2025-01-08
test(e2e): add checks for properties in reference tests

RitaDias:
### Description
Tests added after the latest round of Actions API feedback

- Add e2e to check that properties are being added when reference is created
- Add e2e to check that the properties are being removed when the reference and document are published
- Add e2e to check that the properties are not removed when the weak property is set in the schema

### What to review

Do the tests make sense?

### Notes for release

N/A internal test


RitaDias:
@juice49 do you know if these things have landed?

PR #8216, opened 2025-01-08
feat(cli): skip directories without .env.example when bootstrapping remote template

RostiMelk:
### Description

Skipping directories without an  .env.example when bootstrapping remote templates.

The reason for this is that certain packages in a monorepo setup might not be sanity related, and might not even have a .env.example at all. @sanity/template-validator will still decide whether the package is valid or not. 

Here is an example of a monorepo setup that contains unrelated packages: https://github.com/robotostudio/roboto-next-sanity. See packages directory.

The env replacer will now also replace all occurrences, not only the first. This is necessary for .env.example files that look like this:https://github.com/web3templates/stablo/blob/main/.env.local.example

### Testing

Run
```shell
```

But there is also a test in place for running this

### Notes for release

N/A

### DIFF
--- a/packages/@sanity/cli/src/util/remoteTemplate.ts
+++ b/packages/@sanity/cli/src/util/remoteTemplate.ts
@@ -217,9 +217,11 @@ export async function applyEnvVariables(
       await access(join(root, file))
       return file
     }),
-  ).catch(() => {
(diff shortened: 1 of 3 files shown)


pedrobonamin (approved):
looks good to me, thanks @RostiMelk

PR #8224, opened 2025-01-09
feat(cli): slim down remote test template

RostiMelk:
### Description

- Slimmed down the remote test template to only include the necessary files for testing 
- Added two additional env variables in `test/__fixtures__/remote-template/.env.local.example`
	- `SANITY_STUDIO_PROJECT_ID`
	- `SANITY_STUDIO_DATASET`
- Moved `test/test-template` to `test/__fixtures__/remote-template`
- Updated paths in tests to new path
- Fixed an issue where only the first occurrence of an .env match was replaced, relevant for the two new additional env's that was added.

The tests for remote templates runs against `next` branch, so they will fail here since the path has been changed. Test should start passing again as soon as this is merged :) 

### Testing

Running this new slimmed down template through the template validator still validates as expected
image

### Notes for release

N/A


bjoerge (approved):
Nice cleanup! 🧹✨

RostiMelk:
Can confirm that the tests passes after merging:
image

PR #8226, opened 2025-01-09
fix(cli): explicitly exit workers when they're done

judofyr:
### Description

A Node script only exists once all pending handles/requests are completed. It's quite common for dependencies to open various handles and thus prevent the script for naturally exiting. This changes all the scripts which we run inside worker threads to explicitly exit when they're done.

This impacts the following commands: `schema extract`, `manifest extract`, `graphql deploy`, `documents validate`, `schema validate`.

This also now unifies the commands so that they all have a `main` function.

### What to review

`schema extract`, `manifest extract`, `graphql deploy`, `documents validate`, `schema validate`.

### Testing

I've tested `schema extract` and `manifest extract` on a custom Studio linked to this version locally. I've not tested the other ones.

### Notes for release

Fixes a case where the commands `schema extract`, `manifest extract`, `graphql deploy`, `documents validate` or `schema validate` would never complete.

### DIFF
--- a/packages/sanity/src/_internal/cli/threads/extractManifest.ts
+++ b/packages/sanity/src/_internal/cli/threads/extractManifest.ts
@@ -9,15 +9,15 @@ export interface ExtractManifestWorkerData {
   workDir: string
 }
 
-if (isMainThread || !parentPort) {
-  throw new Error('This module must be run as a worker thread')
-}
-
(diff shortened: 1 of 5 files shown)


sgulseth (approved):
lgtm!

rexxars (approved):

PR #8232, opened 2025-01-10
feat: releases overview shows total docs in release

jordanl17:
### Description
| Scenario | Before | After |
|--------|--------|--------|
| Active releases | Screenshot 2025-01-10 at 16 03 46 | Screenshot 2025-01-10 at 16 02 31 |
| Archived/Published releases | Screenshot 2025-01-10 at 16 04 20 | Screenshot 2025-01-10 at 16 03 08 | 

* Split out the cell components for `ReleasesOverview` into separate components rather than all inside the column defs
* Added tests for these cell components
* Updated the cell component for the document count (`ReleaseDocumentsCount`), to:
  * use the `documentsMetadata` count for active releases
  * use the `finalDocumentStates` count for published or archived releases
* Removed `existingDocumentCount` from the release metadata aggregator store as it's no longer needed

### Testing
Added tests to the cell components

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/hooks/__tests__/useDocumentVersions.test.tsx
+++ b/packages/sanity/src/core/releases/hooks/__tests__/useDocumentVersions.test.tsx
@@ -6,8 +6,8 @@ import {type DocumentPreviewStore} from '../../../preview'
 import {type DocumentIdSetObserverState} from '../../../preview/liveDocumentIdSet'
 import {useDocumentPreviewStore} from '../../../store'
 import {getPublishedId, type PublishedId} from '../../../util/draftUtils'
+import {activeASAPRelease, activeScheduledRelease} from '../../__fixtures__/release.fixture'
 import {type ReleaseDocument, useReleases} from '../../store'
-import {RELEASE_DOCUMENTS_PATH} from '../../store/constants'
 import {useDocumentVersions} from '../useDocumentVersions'
 
 vi.mock('../../store', () => ({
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/releases/hooks/__tests__/useDocumentVersions.test.tsx
jordanl17, line 1: This test change isn't entirely relevant to the task at hand, but I came across it and thought it would be good to try and move as quickly as possible to using the release fixtures everywhere

packages/sanity/src/core/releases/store/createReleaseMetadataAggregator.ts
jordanl17, line 1: Removing `existingDocumentsCount` from the store record

packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const ReleaseDocumentsCounter = ({documentCount}: Props) => {
+  const {t} = useTranslation(releasesLocaleNamespace)
+
+  const count = useMemo(() => {
+    if (!documentCount) return '-'
+
+    return (
+      <Translate
+        t={t}
+        i18nKey={documentCount > 1 ? `document-count_other` : `document-count_one`}
```
RitaDias, line 20: our i18n tool actually handles these causes automatically, as far as I know you shouldn't need to do this :) 
You need to send in the count like this (example on our code base, [translation with count](https://github.com/sanity-io/sanity/blob/sapp-1973/packages/sanity/src/core/comments/i18n/resources.ts#L74-L77), [use in component](https://github.com/sanity-io/sanity/blob/sapp-1973/packages/sanity/src/core/comments/plugin/field/CommentsFieldButton.tsx#L184))

RitaDias (changes requested):
This looks great to me!
I guess the only thing would be about the translation key but otherwise everything looks good to go. My question is: are we going to add the separation of document changes (added, modified, deleted etc) at a later point or is that something that was dropped for now?

RitaDias:
Also, from a design stand point I really don't like how the numbers and text split the line, do we have to have the "documents" text there? :( Can't we add the header with the information that we are showing the count?
 
image

packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
jordanl17: Oh well that's quite nice! There are a few other places I've done this (the wrong) thing. I'll remember to go back and update them.

Thanks for flagging. However clever of them!

jordanl17:
> This looks great to me! I guess the only thing would be about the translation key but otherwise everything looks good to go. My question is: are we going to add the separation of document changes (added, modified, deleted etc) at a later point or is that something that was dropped for now?

That's been dropped now. The document action will only be present on release dashboard lists, but since this requires some backend work in places to make performant, we aren't going to touch it until a later date.

RitaDias (approved):

PR #8237, opened 2025-01-10
feat(sanity): include `_id` field in `groq2024` searches

juice49:
### Description

In GROQ, when the `@` operator is used to match a text query, the content of the `_id` field is not included. Explicitly adding a match on `_id`  allows users to find documents by id.

### What to review

It should be possible to find a document in Test Studio (where `groq2024` search is enabled) by searching for its id.

### Testing

Updated relevant unit tests.


jordanl17 (approved):
Found the relevant docs for text:query that clarify the need for explicit `_` attrs. Looks good @juice49

PR #8250, opened 2025-01-13
chore(core): update `<Resizable/>`, accept a position prop and remove duped and export from `sanity`

pedrobonamin:
### Description

Updates `<Resizable>` to make it reusable and moves it to `core/components`, this component was duplicated in Structure and Tree editing.
In `corel` we need to reuse it, instead of creating an extra copy of it, we move it to a central location and add a `resizerPosition` prop to make it reusable.

With this changes developers will be able to use the resizer and provide a prop to decide where the Resizer handle will show.

### Notes for release
chore(core): update `<Resizable/>`, accept a position prop and export from `sanity`


jordanl17 (approved):

pedrobonamin:
> @pedrobonamin perhaps we get some eyes on from outside of releases team?

Thanks for reviewing Jordan, I think  this change is simple and easy to test. Will merge it.

PR #8251, opened 2025-01-13
feat(sanity): add release layering support to `groq2024` search strategy

juice49:


packages/sanity/src/core/search/groq2024/createSearchQuery.ts
```diff
@@ -127,11 +130,27 @@ export function createSearchQuery(
     .map((s) => `// ${s}`)
     .join('\n')
 
+  let activePerspective: string | string[] | undefined
```
jordanl17, line 133: COULD:
```suggestion
  let activePerspective: SearchOptions['perspective']
```

Or this does make me wonder if the type of `SearchOptions.perspective` could be somewhat more narrowly scoped than just `string | string[]`

jordanl17 (approved):

packages/sanity/src/core/search/groq2024/createSearchQuery.ts
juice49: It's funny you mention that. I'm currently making some changes in another branch to try and unify this a bit. `@sanity/client` provides a dedicated `ClientPerspective` type, but we tend to simply use `string[]` in a lot of places.

IMO: not necessarily worth addressing in this branch, but something we should shore up across the codebase. Definitely a good point.

PR #8256, opened 2025-01-13
feat: support for sticky params and intent operations

jordanl17:
### Description
This is a resurrection of a [previous PR raised here](https://github.com/sanity-io/sanity/pull/7429). This PR was reverted since it clashed with similar behaviours in presentation.

Adds ability to define sticky search params which will be retained state-fully in the URL throughout navigation

> [!NOTE]
> There are currently no sticky param keys. These will be added with content releases. The `perspective` param key used by presentation is not sticky, until content releases makes it so 

stickyPresNext

### What to review
Primarily `packages/sanity/src/router/RouterProvider.tsx` is where the handling and persistence of the params is handled

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/presentation/loader/LoaderQueries.tsx
+++ b/packages/sanity/src/presentation/loader/LoaderQueries.tsx
@@ -122,6 +122,10 @@ export default function LoaderQueries(props: LoaderQueriesProps): React.JSX.Elem
       comlink.onStatus(onLoadersConnection)
 
       comlink.on('loader/documents', (data) => {
+        // data is not presenting the correct perspective
+        // eg moving from previewDrafts -> published
+        // data.perspective will be previewDrafts
+        // despite props.perspective being published
         if (data.projectId === projectId && data.dataset === dataset) {
           onDocumentsOnPage(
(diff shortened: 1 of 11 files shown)


packages/sanity/src/presentation/loader/LoaderQueries.tsx
```diff
@@ -122,6 +122,10 @@ export default function LoaderQueries(props: LoaderQueriesProps): React.JSX.Elem
       comlink.onStatus(onLoadersConnection)
 
       comlink.on('loader/documents', (data) => {
+        // data is not presenting the correct perspective
```
jordanl17, line 125: @stipsan do you have any ideas? I can't find exactly where comlink is being initiated with some data to the callbacks

stipsan: It starts off in `useParams` inside `PresentationTool` 🙌 
The router integration is a bit of a mess tbh.

juice49 (approved):
This is great!

pedrobonamin (approved):

PR #8260, opened 2025-01-14
fix(core): replace `@sanity/block-tools` with `@portabletext/block-tools`

christianhg:
### Description

`@sanity/block-tools` was deprecated in https://github.com/sanity-io/sanity/pull/8254. Now it's time to remove it from
this repo and replace internal usages with the new package.

The contents of `@portabletext/block-tools@1.0.2` and
`@sanity/block-tools@3.69.0` are identical so everything should just work.

### What to review

Paste HTML in a PTE field and verify that it still works.

### Testing

Pasted HTML in a PTE field and verified that it still worked.

### Notes for release

n/a


rexxars (approved):
❤️

PR #8268, opened 2025-01-14
chore(core): use displayed value for document title

pedrobonamin:
### Description

`useDocumentTitle` is checking for the existence of `editState.draft` | `editState.published` 
With the introduction of `releases` this is error prone, because now we can display different values, not only draft or published inside the document pane.

This changes the hook to depend on the `displayed` value, which will be the value shown in the form. 

### Notes for release
chore(core): use displayed value for document title


bjoerge (approved):
Good one!

PR #8271, opened 2025-01-14
chore(structure): remove `TimelineStore` from `<DocumentPaneProvider/>`

pedrobonamin:
### Description

We have always had only one way to represent the Timeline, it has been defined by the TimelineStore, with the addition of releases we are adding a new store that builds on the events api.
This PR updates the `DocumentPaneProvider` to receive the timeline through the props, allowing us to inject either this Legacy Timeline or the new Events API. 
It also changes the `structure` export to expose the `DocumentPaneProviderWrapper` instead of the `DocumentPaneProvider` 

The wrapper now only inserts the legacy Timeline, but in `corel` it will check the feature flag and use the events store instead.

This PR also removes the `timelineMode` property from the document pane, given it's not used anywhere. This is a legacy property that was used before the [History UI updates](https://github.com/sanity-io/sanity/pull/7462)

### What to review
Is it ok to rename the export as done in [here](https://github.com/sanity-io/sanity/pull/8271/files#diff-713de4553c7111bfd0e61c82a8b2e1843894a13953ab165d6b077109e716647aR2)?
Is it ok to deprecate the TimelineMode property as this PR is doing it?

### Testing
Existing tests should cover the functionalities, nothing new is added. This PR only moves things around to make it possible to inject a different timeline provider

### Notes for release
fix(core): remove timelineMode from `documentPaneContext` and makes `TimelineStore` optional.
Deprecations: 
  `useDocumentPane().timelineMode` property is now deprecated, this hasn't been used since the updates to the [History UI]((https://github.com/sanity-io/sanity/pull/7462)).
 `useDocumentPane().timelineStore` possibly undefined, preparing for the introduction of the events API, this property could be undefined once `[corel](https://github.com/sanity-io/sanity/tree/corel)` is merged.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/history/useTimelineSelector.ts
+++ b/packages/sanity/src/core/store/_legacy/history/useTimelineSelector.ts
@@ -9,9 +9,14 @@ import {type TimelineState, type TimelineStore} from './useTimelineStore'
  * @internal
  */
 export function useTimelineSelector<ReturnValue>(
-  timelineStore: TimelineStore,
+  timelineStore: TimelineStore | undefined,
   selector: (timelineState: TimelineState) => ReturnValue,
 ): ReturnValue {
+  if (!timelineStore) {
+    throw new Error(
+      'Passed timelineStore is undefined, if your are using the events timeline, call useEvents() instead. If you need to use this hook, opt in by setting the beta.eventsAPI.enabled feature flag to false',
+    )
+  }
   return useSyncExternalStoreWithSelector(
     timelineStore.subscribe,
     timelineStore.getSnapshot,
--- a/packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
@@ -21,7 +21,6 @@ import {
 
 import {type View} from '../../structureBuilder'
 import {type PaneMenuItem, type PaneMenuItemGroup} from '../../types'
-import {type TimelineMode} from './types'
 
 /** @internal */
 export interface DocumentPaneContextValue {
@@ -65,11 +64,17 @@ export interface DocumentPaneContextValue {
   previewUrl?: string | null
   ready: boolean
   schemaType: ObjectSchemaType
-  setTimelineMode: (mode: TimelineMode) => void
+  /**
+   * @deprecated not used anymore
+   * */
+  setTimelineMode?: undefined
+  /**
+   * @deprecated not used anymore
+   * */
+  timelineMode?: undefined
   setTimelineRange(since: string | null, rev: string | null): void
   setIsDeleting: (state: boolean) => void
   timelineError: Error | null
-  timelineMode: TimelineMode
   /**
    * Soon to be deprecated with the upcoming `releases` changes.
    */
--- a/packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
@@ -70,7 +70,10 @@ export interface DocumentPaneContextValue {
   setIsDeleting: (state: boolean) => void
   timelineError: Error | null
   timelineMode: TimelineMode
-  timelineStore: TimelineStore
+  /**
+   * Soon to be deprecated with the upcoming `releases` changes.
+   */
+  timelineStore?: TimelineStore
   title: string | null
   validation: ValidationMarker[]
   value: SanityDocumentLike
@@ -85,4 +88,8 @@ export interface DocumentPaneContextValue {
   __internal_tasks?: {
     footerAction: React.ReactNode
   }
+
+  // History specific values
+  revisionId: string | null
+  lastNonDeletedRevId: string | null
 }
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx
@@ -1,0 +1,16 @@
+import {memo} from 'react'
+import {useSource} from 'sanity'
+
+import {DocumentPaneWithLegacyTimelineStore} from './DocumentPaneLegacyTimeline'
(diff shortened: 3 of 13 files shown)


packages/sanity/src/core/store/_legacy/history/useTimelineSelector.ts
```diff
@@ -9,9 +9,14 @@ import {type TimelineState, type TimelineStore} from './useTimelineStore'
  * @internal
  */
 export function useTimelineSelector<ReturnValue>(
-  timelineStore: TimelineStore,
+  timelineStore: TimelineStore | undefined,
```
RitaDias, line 12: A tad concerned about this change, this will be a breaking change since users so far haven't had to be concerned about potentially sending an undefined through here.

Though the more I think about it might be fine? Since overall the users might need to make some extra changes? We do still keep the legacy components 🤔

packages/sanity/src/structure/panes/document/types.ts
```diff
@@ -1,12 +1,30 @@
...
+  store?: TimelineStore
+  error: Error | null
+  onOlderRevision: boolean
+  revisionId: string | null
+  revisionDocument: SanityDocument | null
+  sinceDocument: SanityDocument | null
+  ready: boolean
+  /**
+   * Whether this timeline is fully loaded and completely empty (true for new documents)
+   * It can be `null` when the chunks hasn't loaded / is not known
```
RitaDias, line 23: Would it make sense to add `| null` in this case?

packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import {DocumentPaneWithLegacyTimelineStore} from './DocumentPaneLegacyTimeline'
+import {type DocumentPaneProviderProps} from './types'
+
+/**
+ * @internal
+ */
+export const DocumentPaneProviderWrapper = memo((props: DocumentPaneProviderProps) => {
+  const source = useSource()
+  // TODO: This will add support for the new events timeline store, see https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx#L14-L15
```
RitaDias, line 12: What is the TODO specifically in this case? 🤔

packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
```diff
@@ -65,12 +64,21 @@ export interface DocumentPaneContextValue {
...
   schemaType: ObjectSchemaType
-  setTimelineMode: (mode: TimelineMode) => void
+  /**
+   * @deprecated not used anymore
+   * */
+  setTimelineMode?: undefined
+  /**
+   * @deprecated not used anymore
+   * */
+  timelineMode?: undefined
```
RitaDias, line 74: I don't know what's our current go to approach for deprecating properties. I think the way we've been doing it is _first_ putting the `@deprecated` label, leaving it to simmer for a bit as a warning for folks not to use it and only later removing it instead of changing the type now since this will cause an immediate breaking change

I'm not sure who to tag so let's say @bjoerge , do you have thoughts?

RitaDias (commented):
Just a couple of questions. 
Also, I just want to get one more pair of eyes on this beyond my own since I do feel that we might need to add something extra in the changelog about potential breaking changes? But I might be exaggerating that aspect, just want another opinion 🙏 

I tagged you @bjoerge but let me know if you are swamped and I'll round robin someone else into the conversation 👍

packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx
pedrobonamin: Is to reference why I'm adding this change. It will be modified in `corel` by this change https://github.com/sanity-io/sanity/blob/corel/packages/sanity/src/structure/panes/document/DocumentPaneProviderWrapper.tsx#L14-L15
Adding the `eventsAPI` when enabled.

packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
pedrobonamin: It has been unused by us since we merged the [History UI updates](https://github.com/sanity-io/sanity/pull/7462) and it hasn't made much sense since then.
Would like to understand how to properly handle this deprecations. 


This was previously used to determine which Timeline Selection menu was open, when we had the option to select the document from the document header, with the new this is not needed.

RitaDias: huh, if this is the case I would almost prefer to remove it (ripping off the bandaid) which is certainly more extreme _but_ I feel it would be preferrable vs a middle ground like this. Though maybe the approach should be as I mentioned above, I'm not sure

We should probably overall clarify how do we want to handle these cases.

pedrobonamin:
> Great work! I just a couple of questions. :) Also, I just want to get one more pair of eyes on this beyond my own since I do feel that we might need to add something extra in the changelog about potential breaking changes? But I might be exaggerating that aspect, just want another opinion 🙏
> 
> I tagged you @bjoerge but let me know if you are swamped and I'll round robin someone else into the conversation 👍

Thanks for reviewing @RitaDias , just updated the notes for release to reflect this two deprecations.

packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
bjoerge: Seems unlikely that anyone would rely on this? If it's no longer in use, it should be fine to remove it.

pedrobonamin: Yeah, I have the same feeling, and if they are depending on it they should update it because it's not respected by us anymore.

bjoerge (approved):
LGTM! (as in Looks _Great_ To Me)

PR #8279, opened 2025-01-15
test(e2e): add checks for properties in reference tests 

RitaDias:
### Description
Tests added after the latest round of Actions API feedback

- Add e2e to check that properties are being added when reference is created
- Add e2e to check that the properties are being removed when the reference and document are published
- Add e2e to check that the properties are not removed when the weak property is set in the schema

### What to review

Do the tests make sense?

### Notes for release

N/A internal test


test/e2e/tests/inputs/reference.spec.ts
```diff
@@ -64,4 +64,129 @@ withDefaultClient((context) => {
...
+
+    await createDraftDocument('/test/content/input-standard;referenceTest')
+    page.getByTestId('string-input').fill(originalTitle)
+
+    /** create reference */
+    await page
+      .locator(
+        'div:nth-child(4) > div > div > div > div:nth-child(2) > div > div > div > div > div > div > button',
+      )
+      .click()
```
pedrobonamin, line 82: could we have a test id or some other selector here? This looks easy to break with any change in the structure

pedrobonamin (commented):
Thanks for doing this Rita! Added a small suggestion on a selector that would be nice to use in other way

test/e2e/tests/inputs/reference.spec.ts
RitaDias: Good point, I had made a mental note to return to this once I had the tests passing and I should have made an actual written note 😅 
Have pushed it now and updated the tests across!

pedrobonamin (approved):

PR #8281, opened 2025-01-15
refactor(core): perf work

RitaDias:


packages/sanity/src/core/releases/store/useArchivedReleases.ts
```diff
@@ -0,0 +1,21 @@
+import {useMemo} from 'react'
+
+import {ARCHIVED_RELEASE_STATES} from '../util/const'
+import {type ReleaseDocument} from './types'
+
+/**
+ * @internal
+ */
+export function useArchivedReleases(releases: ReleaseDocument[]): {
```
pedrobonamin, line 9: I think it would be simpler if this hook could access the state in the same way `useAllReleases` is doing it, wdyt?
```
  const {state$} = useReleasesStore()
  const state = useObservable(state$)!
```

And filter there for the releases that are archived.
It feels error prone that it receives the releases, as we could make the mistake of doing
```
  const {data: releases, loading} = useReleases()
  const {archivedReleases} = useArchivedReleases(releases)

```
Given `useReleases` is filtering out the archived releases nothing will be returned from this hook

pedrobonamin: Or this hook access to `useAllReleases` directly.

```diff
-export function useArchivedReleases(releases: ReleaseDocument[]): {

+export function useArchivedReleases(): {
+ const {allReleases: releases } = useAllReleases()
  
  const archivedReleases = useMemo(
    () =>
      Array.from(releases.values()).filter((release) => {
        return ARCHIVED_RELEASE_STATES.includes(release.state)
      }),
    [releases],
  )

  return useMemo(() => ({archivedReleases}), [archivedReleases])
}
```

packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
```diff
@@ -24,7 +25,9 @@ export const DocumentEventsPane = (props: DocumentPaneProviderProps) => {
   const options = usePaneOptions(props.pane.options, params)
 
   const {selectedPerspectiveName} = usePerspective()
-  const {archivedReleases} = useReleases()
+  const {data} = useReleases()
+  const {archivedReleases} = useArchivedReleases(data)
```
pedrobonamin, line 29: This doesn't seem correct given `useReleases` will not return the archived ones.
```suggestion
  const {data} = useAllReleases()
  const {archivedReleases} = useArchivedReleases(data)
```

packages/sanity/src/core/releases/store/useArchivedReleases.ts
RitaDias: For the first part of the question (re: prop):
The reason why I didn't do it that way is because, if I understood the chat with @stipsan about performance, we are trying to avoid refetching this sort of data when it already exists and has been fetched 🤔
Mostly I'm mentioning this because in some places of the code we both need to call the `useReleases` _and_ the _archived_ but we don't _always_ need to get the archived. But I might be splitting hairs. Do you mind sharing your thoughts on this part @stipsan ? 🙏 

(as for the second part as you mentioned it here https://github.com/sanity-io/sanity/pull/8281#discussion_r1918774508 it can indeed happen) so it might be something that we take into account. 

Might also just mean that we should change the name of the `useReleases` to `useCurrentReleases` which might be something we need to do anyway, thoughts?

packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
RitaDias: Good catch, this is definitely a mistake on my part (brought up [here](https://github.com/sanity-io/sanity/pull/8281#discussion_r1918817818))

packages/sanity/src/core/releases/store/useArchivedReleases.ts
pedrobonamin: Agree with changing the name to `useCurrentReleases` or `useActiveReleases` releases and make it more explicit.

I understand we don't always need the archived, but when using the archived we will need to access the `allReleases` or the `releasesStore` to find those ones that are archived, so maybe doing it in the hook simplifies it's usage and avoids mistakes.
We are not going to fetch the releases again, we are only going to access to the store to get them, the fetch happens only once inside the store, then the store is memoized so if you access the store from more than one place that shouldn't be an issue.  (if it is doing the fetch more than once we should check the code because something is wrong)

RitaDias: This being said, part of the point to attempt to separate these things into smaller hooks was to avoid having to have so many hooks that required other hooks to function so that might be going against that philosophy. (I'm still working on the refactoring of the Archived, just thought to bring it up)

RitaDias: [Renaming useRelease to useActiveRelease](https://github.com/sanity-io/sanity/pull/8281/commits/436bf7ec6b277050cdb4893958d760ca0b78c3b5)
[Refactoring useArchivedReleases](https://github.com/sanity-io/sanity/pull/8281/commits/e06da423ba4b85a121741a6d06255383ba60b881)

packages/sanity/src/core/releases/store/useAllReleases.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export function useAllReleases(): {
+  allReleases: ReleaseDocument[]
+} {
+  const {state$} = useReleasesStore()
+  const state = useObservable(state$)!
+
+  return useMemo(
+    () => ({
+      allReleases: sortReleases(Array.from(state.releases.values())),
```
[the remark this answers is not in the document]
RitaDias: [Refactoring useAllReleases](https://github.com/sanity-io/sanity/pull/8281/commits/ee9fbe26e96fa301d7026cc369faf6a4e2f86b4a)

packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
RitaDias: No longer needed since we [refactored it](https://github.com/sanity-io/sanity/pull/8281/commits/e06da423ba4b85a121741a6d06255383ba60b881) :)

RitaDias:
@pedrobonamin I've tagged you again for release despite the tests. I think we might need a merge from next to corel to get the tests to pass somewhat consistently? if after we merge next into corel our e2e tests are still failing consistently I'll have a look at it outside of this PR

packages/sanity/src/core/releases/store/useActiveReleases.ts
```diff
@@ -42,21 +33,9 @@ export function useReleases(): ReleasesState {
...
-        ARCHIVED_RELEASE_STATES.includes(release.state),
-      ),
-    [state.releases],
-  )
-  const releasesIds = useMemo(
-    () => releasesAsArray.map((release) => getReleaseIdFromReleaseDocumentId(release._id)),
-    [releasesAsArray],
-  )
+
   return {
```
pedrobonamin, line 37: We are adding `useMemo` to the return of the other three hooks, would we benefit here from also adding it?

RitaDias: 🤦 yep we would, let me fix that.

pedrobonamin (approved):
Thanks for doing this @RitaDias , great work!!

PR #8282, opened 2025-01-15
fix(core): cannot find namespace 'vi'

pedrobonamin:
### Description
Fixes a type import issue in where using `vi.<Type>` would give the following error:
```
Cannot find namespace 'vi'
```

### What to review
Is the change correct?

### Testing
It's only a type definition change, existing tests cover this.

### Notes for release
n/a, internal only


bjoerge (approved):
good catch!

PR #8283, opened 2025-01-15
fix: forcing all search strategies to use vX API whilst releases not supported in named versions

jordanl17:
### Description

This was found when migrating admin studio. Currently this change already exists inside `next` ([here](https://github.com/sanity-io/sanity/commit/13f2e48fe32dd804f97b57ad5cd3be23b1c72c07#diff-d82b828d6223470f2a4ce5fd6ee4d0ee93c9264f7777c2a739f4c0d9aa9d876bR52))

`vX` is the only current version of the sanity API that supports perspectives and releases. Therefore any studio wishing to use releases must use either the new groq search strategy (which consequentially will use `vX`) or must define explicitly that the `S.DocumentList` use `.apiVersion('vX')`.

In the case of admin studio this was going to leave a significant footprint (since a defined apiVersion is used in a great number of areas).

Since this change is ultimately throwaway (once releases and perspective support is brought to a named version of the Sanity API then `vX` overriding is no longer needed), the smaller the wake the better. This felt like the smallest change - using the corel pre-release of sanity will override the api version for all searches to `vX`.

> [!IMPORTANT]
> Since this is a temporary change whilst releases is only on `vX`, I've already created a revert task to undo this override 

### Testing
Verified this against the admin studio:
Admin studio against the current corel tag causes the following error ([see here](https://sanity-io-studio-git-corel-309-admin-releases.sanity.build/structure/sanityIo;blog;publishedPosts?perspective=rE3W8k7yp))
Screenshot 2025-01-15 at 20 48 00

Admin studio running locally pointed to this branch resolves the issue (and network interrogation confirms that the api version is successfully being overridden):
Screenshot 2025-01-15 at 20 50 25

### Notes for release
N/A


packages/sanity/src/core/search/weighted/createWeightedSearch.ts
```diff
@@ -38,15 +38,23 @@ export const createWeightedSearch: SearchStrategyFactory<WeightedSearchResults>
...
-      // This can be relevant when ordering results by specific fields, especially dates.
-      searchOptions?.skipSortByScore ? tap() : map((hits) => sortBy(hits, (hit) => -hit.score)),
-      map((hits) => ({type: 'weighted', hits})),
-    )
+    return client.observable
+      .withConfig({
+        /** @todo remove defined apiVersion once perspective is supported in
+         * non-experimental content lake API
+         */
+        apiVersion: 'vX',
```
jordanl17, line 46: Only actually added this

pedrobonamin (approved):
Looks good to me, would like @juice49  thoughts on this.

juice49 (approved):
Looks good to me!

PR #7011, opened 2024-06-24
feat(sanity): content releases

juice49:

PR #8284, opened 2025-01-16
fix(structure): only reset history params if needed

pedrobonamin:
### Description

History params need to be reseted when the perspective changes, that's the reason for the `useResetHistoryParams` hook, however, this hook was triggering unnecessary `setParams` and that was creating a glitch in which the router starts jumping between perspectives when the user changes them fast.
See video:

https://github.com/user-attachments/assets/5e5e5d1e-6553-4209-b46f-9756eb3c07b9

This updates the useResetHistoryParams hook to only trigger the `setParams` function when a param needs to be updated, preventing this bug.

### DIFF
--- a/packages/sanity/src/core/field/diff/contexts/DocumentChangeContext.tsx
+++ b/packages/sanity/src/core/field/diff/contexts/DocumentChangeContext.tsx
@@ -8,9 +8,6 @@ export type DocumentChangeContextInstance = {
   documentId: string
   schemaType: SchemaType
   rootDiff: ObjectDiff | null
-  /**
-   * This decides if the diff shows the reverting changes button or not.
-   */
   isComparingCurrent: boolean
   FieldWrapper: ComponentType<{path: Path; children: ReactNode; hasHover: boolean}>
   value: Partial<SanityDocument>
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/field/diff/contexts/DocumentChangeContext.tsx
```diff
@@ -8,9 +8,6 @@ export type DocumentChangeContextInstance = {
   documentId: string
   schemaType: SchemaType
   rootDiff: ObjectDiff | null
-  /**
-   * This decides if the diff shows the reverting changes button or not.
-   */
```
RitaDias, line 13: Any reason to remove this? 😄

pedrobonamin: Not a big deal, but it was added in corel and it adds unnecessary noise to the PR. See https://github.com/sanity-io/sanity/pull/7011/files#diff-4810bc27acdf55c2165867faa05d7135681d792bdf1b70bb1eb019c4aaf63f9a

RitaDias: Ah I see! Fair enough, one less change to be removed :)

pedrobonamin: yep, we are trying to slim down as much as possible the `corel` diff, it's currently on 395 files 🤯

RitaDias: I feel this 🥲

RitaDias (approved):
✅

PR #8286, opened 2025-01-16
refactor(sanity): refine perspective stack types

juice49:
### Description

There are some places across the codebase we could improve the typing of perspective stacks so that they're more precise and consistent. In a lot of places, these stacks are currently typed as `string[]`, which is not compatible with `ClientPerspective`. This code now uses the new type `PerspectiveStack`, which reflects only the array values accepted by `ClientPerspective`.

Note: It would be a good idea to export `PerspectiveStack` from `@sanity/client` in the future, rather than us extracting it from the `ClientPerspective` type.

This branch also adopts the `isReleaseDocument` type guard where applicable to determine whether the selected perspective is a system perspective (e.g. `published`) or a release document. In the past, this tended to be done by explicitly checking whether the string is `published` or `draft`, otherwise assuming the value is a release document. Like this:

```ts
if (
  !isPublishedPerspective(selectedPerspective) &&
  !isDraftPerspective(selectedPerspective)
) {
  // `selectedPerspective` should be a release document
}
```

This narrowing starts to fail if the value could be any other string (e.g. `previewDrafts`).

I think we probably do this because the code predates the existence of the isReleaseDocument type guard:

```ts
if (isReleaseDocument(selectedPerspective)) {
  // `selectedPerspective` should be a release document
}
```

### What to review

Does this seem reasonable?

### Testing

Existing test coverage of modified code passes.

### DIFF
--- a/packages/@sanity/vision/package.json
+++ b/packages/@sanity/vision/package.json
@@ -70,7 +70,8 @@
     "json5": "^2.2.3",
     "lodash": "^4.17.21",
     "quick-lru": "^5.1.1",
-    "react-compiler-runtime": "19.0.0-beta-55955c9-20241229"
+    "react-compiler-runtime": "19.0.0-beta-55955c9-20241229",
+    "react-fast-compare": "^3.2.2"
   },
   "devDependencies": {
     "@repo/package.config": "workspace:*",
--- a/packages/@sanity/vision/src/SanityVision.tsx
+++ b/packages/@sanity/vision/src/SanityVision.tsx
@@ -1,4 +1,4 @@
-import {type Tool, useClient} from 'sanity'
+import {type Tool, useClient, usePerspective} from 'sanity'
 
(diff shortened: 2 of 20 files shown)


juice49:
* **#8286** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/8286?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/8286?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `corel`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

bjoerge (approved):
Great work @juice49!

PR #8292, opened 2025-01-16
fix: patching a readOnly document is blocked

jordanl17:
### Description
| Before | After |
|--------|--------|
| patchReadOnlyBEFORE | patchReadOnlyAFTER | 

Above shows a demo where a document is placed in readOnly mode, but the custom input still called the `onChange` callback. After refresh in 'Before' the change has persisted. In 'After' the change has not persisted

### Notes for release
Fixing issue where custom form inputs could wrongly patch a `readOnly` document by call `onChange`

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -295,26 +295,6 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     [onFocusPath, setFocusPath],
   )
 
-  const patchRef = useRef<(event: PatchEvent) => void>(() => {
-    throw new Error(
-      'Attempted to patch the Sanity document during initial render or in an `useInsertionEffect`. Input components should only call `onChange()` in a useEffect or an event handler.',
-    )
-  })
-  useInsertionEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -561,6 +541,34 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
     isCreateLinked,
   ])
 
+  const patchRef = useRef<(event: PatchEvent) => void>(() => {
+    if (readOnly) {
```
bjoerge, line 545: since this is in effect only for a split second (the useInsertionEffect below will replace the implementation), this error is not likely appear, unless the input calls onChange immediately. Maybe better to move this throw statement to the useInsertionEffect function below?

jordanl17: I'm might be getting confused, but since the ref will only be reassigned in the `useInsertionEffect` if not `readOnly`, this will be the function that is always returned when `readOnly`.

I do see this error myself when testing

bjoerge: Ah, you're right. But now if the document now goes from `!readOnly` to `readOnly` the error won't be thrown?

bjoerge (approved):
Thank you for fixing this @jordanl17!

PR #8302, opened 2025-01-16
fix(sanity): editing version document using Actions API

juice49:
### Description

This branch fixes version editing when using the Actions API, and  enables the Actions API in the default workspace in Test Studio.

When using the Actions API to edit a version document, the `draftId` option must refer to the version id. Previously, the draft document would be patched instead, which is incorrect.

This branch also temporarily changes the API version to `vX` when executing Actions API actions. At this point in time, the Actions API can only execute certain actions for version documents (e.g. `sanity.action.document.edit`) in `vX`.

### What to review

Editing a version document successfully calls the edit action for the version document. The draft and published documents must not be affected.

### Testing

Added unit test in `packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.test.ts`.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -168,6 +168,10 @@ const defaultWorkspace = defineConfig({
   },
   basePath: '/test',
   icon: SanityMonogram,
+  // eslint-disable-next-line camelcase
+  __internal_serverDocumentActions: {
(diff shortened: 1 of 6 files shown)


pedrobonamin (approved):
Tested it and it looks good to me, thanks for working this out @juice49 ! 
Great work!

PR #8312, opened 2025-01-17
refactor: migrate `@sanity/presentation` codebase

stipsan:
### Description

This PR copies over the codebase for `@sanity/presentation` on its `main` [branch](https://github.com/sanity-io/visual-editing/tree/main/packages/presentation). I've then refactored the code to pass our linter checks for things like:
- `React.createContext` only being allowed in `sanity/_singletons`.
- Components in `ui-components` like `<Button />`, `<Tooltip>` and such should be used instead of importing directly from `@sanity/ui`
- Slight differences in TypeScript and ESLint

The codebase is otherwise the same. It's tempting to refactor and start cleaning things up, but it's important to note that after this PR is merged, I'll have to:
- pull the changes on this PR on top of the `corel` [branch](https://github.com/sanity-io/sanity/tree/corel)
- manually apply the changes from the other `corel` branch, [on top of Presentation](https://github.com/sanity-io/visual-editing/pull/2081).
- once both are merged I'll delete `packages/presentation` from the `visual-editing` monorepo, and start adding over tests for more advanced presentation features that aren't enabled in [the current test studio](https://test-studio.sanity.dev/presentation/presentation).

It's more important that we reach the step where we can delete the code from `visual-editing` ASAP, so we avoid new work being done there, for example by external contributors.

### What to review

Is the new code following conventions ok?
It does pull in a bunch of new dependencies, and I want to work through them and get rid of unnecessary ones (we only use `suspend-react` in one instance, and it doesn't really need to). But they are lazy loaded, and isn't much different from how we re-export `@sanity/presentation` as-is today.

### Testing

Tests are lacking, once this is merged in we can start working on playwright component tests for Presentation, as well as better vitest coverage for utils and parsers.
For now the PR build seems to work like it did before the migration: https://test-studio-git-migrate-presentation.sanity.dev/presentation/presentation/simpleBlock/716e2c8a-a168-4852-87e2-0c2b4d42c94b/body%5B_key==%22161e19ce1c6b%22%5D.children%5B_key==%22af5299cd594b%22%5D.text

### Notes for release

The `@sanity/presentation` codebase has been migrated into the `sanity` codebase. We've always recommended using Presentation Tool by importing it from `sanity/presentation` like so:
```ts
// sanity.config.ts
import {presentationTool} from 'sanity/presentation'
import {structureTool} from 'sanity/structure'

export default defineConfig({
  plugins: [presentationTool(), structureTool()]
})
```
If you've been following this pattern then this change won't impact you in any way.

If you've been installing `@sanity/presentation` and are using it directly (we've sometimes done this to let folks try out bugfixes before they ship in the next release of `sanity`):
```ts
// sanity.config.ts
import {presentationTool} from '@sanity/presentation'
import {structureTool} from 'sanity/structure'

export default defineConfig({
  plugins: [presentationTool(), structureTool()]
})
```

Then you need to search/replace all `from '@sanity/presentation'` statements in your codebase with `from 'sanity/presentation'`, and uninstall `@sanity/presentation` as it'll no longer receive updates.


packages/sanity/package.config.ts
```diff
@@ -4,6 +4,12 @@ import {defineConfig} from '@sanity/pkg-utils'
 export default defineConfig({
   ...baseConfig,
 
+  define: {
+    ...baseConfig.define,
+    PRESENTATION_ENABLE_LIVE_DRAFT_EVENTS:
+      process.env.PRESENTATION_ENABLE_LIVE_DRAFT_EVENTS === 'true',
```
stipsan, line 10: This is used to feature flag a new approach to drive live previews, which uses Sanity Live Content API for draft content (which currently requires vX, which is why it's off by default)

packages/sanity/src/presentation/components/Resizer.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @media (hover: hover) {
+    &:hover > span:nth-child(2) {
+      opacity: 0.2;
+    }
+  }
+`
+
+export const Resizer: FunctionComponent<{
```
pedrobonamin, line 50: NIT, We have a Resizer that we could replace this one with as a next step. `packages/sanity/src/core/components/resizer/Resizer.tsx`

packages/sanity/src/presentation/editor/DocumentListPane.tsx
```diff
@@ -0,0 +1,126 @@
...
+    min-width: none !important;
+    max-width: none !important;
+  }
+`
+
+const WrappedCode = styled(Code)`
+  white-space: pre-wrap;
+`
+
+export function DocumentListPane(props: {
```
pedrobonamin, line 38: NIT - follow up, could be good to rename this file to something else to avoid confusions with the structure/documentListPane

packages/sanity/src/presentation/editor/DocumentPane.tsx
```diff
@@ -0,0 +1,106 @@
...
+import {PresentationSpinner} from '../PresentationSpinner'
+import {type PresentationSearchParams, type StructureDocumentPaneParams} from '../types'
+import {usePresentationTool} from '../usePresentationTool'
+import {PresentationPaneRouterProvider} from './PresentationPaneRouterProvider'
+
+const WrappedCode = styled(Code)`
+  white-space: pre-wrap;
+`
+
+export function DocumentPane(props: {
```
pedrobonamin, line 24: NIT - follow up, could be good to rename this file to something else to avoid confusions with the structure/DocumentPane

packages/sanity/src/presentation/lib/debounce.ts
```diff
@@ -0,0 +1,12 @@
+export function debounce<F extends (...args: Parameters<F>) => ReturnType<F>>(
```
pedrobonamin, line 1: NIT, follow up, could be replaced for debounce from lodash

packages/sanity/src/presentation/plugin.tsx
```diff
@@ -0,0 +1,144 @@
...
+    console.warn('Presentation’s `locate` option is deprecated. Use `resolve.locations` instead.')
+  }
+
+  const hasLocationsResolver = !!(options.resolve?.locations || options.locate)
+
+  function PresentationDocumentInput(props: InputProps) {
+    const value = props.value as SanityDocument
+    const documentId = value?._id ? getPublishedId(value?._id) : undefined
+
+    if (isDocumentSchemaType(props.schemaType)) {
```
pedrobonamin, line 60: NIT this seems to be added only once for the document. 
I think you could benefit from using [`unstable_layout`](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/core/comments/plugin/index.ts#L15-L16) instead of checking in each input if it's a document schema type.

packages/sanity/src/presentation/usePreviewUrl.ts
```diff
@@ -0,0 +1,132 @@
+import {createPreviewSecret} from '@sanity/preview-url-secret/create-secret'
+import {definePreviewUrl} from '@sanity/preview-url-secret/define-preview-url'
+import {startTransition, useEffect, useMemo, useRef, useState} from 'react'
+import {type SanityClient, useActiveWorkspace, useClient, useCurrentUser} from 'sanity'
+import {suspend} from 'suspend-react'
+
+import {API_VERSION} from './constants'
+import {type PreviewUrlOption} from './types'
+
+export function usePreviewUrl(
```
pedrobonamin, line 10: we have a `usePreviewURL` function in `structure` would be nice to see if we can join the two https://github.com/sanity-io/sanity/blob/corel-references/packages/sanity/src/structure/panes/document/usePreviewUrl.ts#L11-L12

pedrobonamin (approved):
Looks good to me, thanks for moving this @stipsan !
Added some small suggestions but nothing is blocking, more of some future iterations we can do.

packages/sanity/src/presentation/components/Resizer.tsx
pedrobonamin: Yes, sure! Added some small suggestions for follow up, nothing is blocking.

PR #8314, opened 2025-01-17
test(e2e): update reference tests for corel

RitaDias:
### Description

update tests based on the changes done in corel 

### What to review

some things we were relying on next no longer look or work the same, so the tests have been updated

### DIFF
--- a/test/e2e/tests/inputs/reference.spec.ts
+++ b/test/e2e/tests/inputs/reference.spec.ts
@@ -37,7 +37,6 @@ withDefaultClient((context) => {
 
     // Open the Author reference input.
     await referenceInput.getByLabel('Open').click()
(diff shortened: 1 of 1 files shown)


test/e2e/tests/inputs/reference.spec.ts
```diff
@@ -69,19 +67,23 @@ withDefaultClient((context) => {
     page,
     createDraftDocument,
   }) => {
+    test.slow()
```
jordanl17, line 70: Nice

jordanl17 (approved):
Great stuff

test/e2e/tests/inputs/reference.spec.ts
RitaDias: gotta go slow

PR #8317, opened 2025-01-17
refactor: migrate `@sanity/presentation@corel` codebase (#8312)

stipsan:
Replays the changes in https://github.com/sanity-io/visual-editing/pull/2081 on top of #8312.
Trying to make the diff here as small as possible so that we can iterate on presentation on `next`, and have less work syncing with `corel` ❤


dev/test-studio/preview/loader.tsx
```diff
@@ -6,7 +6,7 @@ const client = createClient({
   projectId: 'ppsg7ml5',
   dataset: 'playground',
   useCdn: true,
-  apiVersion: '2023-02-06',
+  apiVersion: 'X',
```
RitaDias, line 9: huh, just to double check, this is intended to be left in the corel branch?

packages/sanity/src/presentation/loader/LiveQueries.tsx
```diff
@@ -483,7 +483,11 @@ export function turboChargeResultIfSourceMap<T = unknown>(
       }
       return changedValue
     },
-    perspective,
+    // TODO: Update applySourceDocuments to support releases.
+    Array.isArray(perspective) &&
```
RitaDias, line 487: huh interesting, why are we doing this?

RitaDias: Also jsut realised this code is the same changes as applied in the  [LoaderQueries.tsx](https://github.com/sanity-io/sanity/pull/8317/files#diff-fc19b0caf7b88a8fc299ff32abf4dd08ab8dc56dcb78dfe91f74ea41753aae4c) should it be moved somewhere else so both can use it? Might be overengineering but since it looks to be the same code in somewhat seemingly related spaces

packages/sanity/src/presentation/overlays/PostMessageDocuments.tsx
```diff
@@ -79,7 +79,17 @@ const PostMessageDocuments: FunctionComponent<PostMessageDocumentsProps> = (prop
...
+       * This is a temporary hack to get around the fact that the `drafts.versions.versionName.id` format is not supported
+       * `versions.` don't have draft documents.
+       *
+       * Where is the drafts. prefix coming from and why it's added to the `versions.` documents?
+       * */
+      const cleanId =
+        data.documentId.includes('versions.') && data.documentId.startsWith('drafts.')
+          ? data.documentId.replace('drafts.', '')
+          : data.documentId
+      const snapshot = await client.getDocument(cleanId, {
```
RitaDias, line 92: wouldn't the getPublishedId work for this? 🤔

packages/sanity/src/presentation/useDocumentsOnPage.ts
```diff
@@ -15,16 +15,18 @@ type KeyedDocumentCache = Record<string, DocumentCache>
 
 const warnOnceAboutCrossDatasetReference = defineWarnOnce()
 
+/**
+ * @TODO should be refactored to an lru-cache that is keyed by the perspective, which could be an array (if it is, it should use consistent sorting),
```
RitaDias, line 19: Do we need to create any stories for this / have they been created? :)

packages/sanity/src/presentation/usePreviewUrl.ts
```diff
@@ -53,7 +53,9 @@ export function usePreviewUrl(
     const resolvedUrl = await resolvePreviewUrl({
       client,
       previewUrlSecret: previewUrlSecret!,
-      studioPreviewPerspective,
+      studioPreviewPerspective: Array.isArray(studioPreviewPerspective)
```
RitaDias, line 56: Feels like we need to check if the perspective is an array somewhat often, I'm not sure if we have a helper for this in the releases package but it might be worth creating a helper function?

RitaDias (changes requested):
Hi Cody! thank you for doing this, I have a few questions which are beneath!

https://github.com/user-attachments/assets/6175bca0-ccc9-4221-96b5-0fb9417fbe5e


Also, after testing out the presentation on the test studio I noticed that editing a version document seems to be broken (check the video), let me know if I can help in any way!!

dev/test-studio/preview/loader.tsx
stipsan: Yep, the preview iframe runs its own setup. It's similar to how applications running Content Releases will also have to specify vX is apps running live preview 🙌

packages/sanity/src/presentation/loader/LiveQueries.tsx
stipsan: the idea is for LiveQueries to replace LoaderQueries relatively soon. The rule of thumb we follow is that since only one of them will live on, we only make refactors that continue to make sense once LoaderQueries are deleted.
Does that make sense?

packages/sanity/src/presentation/overlays/PostMessageDocuments.tsx
stipsan: I don't think it would, as we're not trying to get the published id here, we're trying to get the versions id, but without the drafts prefix.
It's not clear yet how the `drafts.` prefix is able to sneak in to begin with, but my hunch it's something to do with our stega setup in `@sanity/client/stega` 🤔

packages/sanity/src/presentation/useDocumentsOnPage.ts
stipsan: We would need to create tickets, it's not proven yet it'll be needed tbh. I understand that the document preview stack in corel has improved, and it may not be necessary to run caches keyed to the perspective anymore 🤔

packages/sanity/src/presentation/loader/LiveQueries.tsx
RitaDias: I think so, just wonder if we should write that explicitly on the component so the question doesn't come up later if we need to make more changes to both components 🤔

packages/sanity/src/presentation/overlays/PostMessageDocuments.tsx
RitaDias: I see 🤔
Alright, in in any case I think we could still make this change? (with the right imports and all)

```suggestion
       * This is a temporary hack to get around the fact that the `drafts.versions.versionName.id` format is not supported
       * `versions.` don't have draft documents.
       *
       * Where is the drafts. prefix coming from and why it's added to the `versions.` documents?
       * */
      const cleanId =
        isVersionId(data.documentId) && isDraftId(data.documentId)
          ? data.documentId.replace('drafts.', '')
          : data.documentId
      const snapshot = await client.getDocument(cleanId, {
```

packages/sanity/src/presentation/useDocumentsOnPage.ts
RitaDias: Alright, I'm just concerned that it might fall through the cracks if there isn't a linear story to double check on this :)

RitaDias:
> Hi Cody! thank you for doing this, I have a few questions which are beneath!
> 
>  Screen.Recording.2025-01-20.at.08.47.23.mov 
> Also, after testing out the presentation on the test studio I noticed that editing a version document seems to be broken (check the video), let me know if I can help in any way!!

An update to this: looks like it's something that already happens in COREL, created a linear story for it

packages/sanity/src/presentation/overlays/PostMessageDocuments.tsx
stipsan: I think it'll fail. `isVersionId` uses `id.startsWith('versions.')`, but we need it to have `id.includes('versions.')` for the `drafts.versions.` condition to be a match, and then be fixed to `versions.`?

RitaDias: Ahh right, when I read through the code I didn't realise it's "includes" and not "startsWith". Right right. Nevermind me, sorry about the detour

RitaDias: @pedrobonamin quick question on this: do you remember the reason why we're having this condition vs using our isDraftId / isVersionId? (specifically the latter since the condition wouldn't match) :)

pedrobonamin: I think it was due to what @stipsan mentions, that sometimes the id here was coming as `drafts.versions` which I couldn't really track where it was coming from. 
Sometimes it was receiving: `drafts.versions.r123.foo` which is not a valid id.

PR #8320, opened 2025-01-17
fix(presentation): always show all documents in use

stipsan:
### Description

Fixes a regression (I don't know exactly when this regressed) causing only published documents to show up in the list of documents in use in presentation.

### What to review

Does it make sense?

### Testing

When using the default `Drafts` perspective, notice the before and after (the right hand side is correct, the left hand side only shows documents that are published and have no drafts):

image

For the `Published` perspective it should only filter out documents that only have a draft variant:

image

### Notes for release

The Presentation Tool's "Documents in use" pane is now including draft documents, and unpublished documents. So that it actually shows all documents in use, and not just some of them.


georgedoescode (approved):
LGTM! ❤️

PR #8322, opened 2025-01-18
fix(core): references to documents on same release are valid

pedrobonamin:
### Description

Validates that references to documents in the same release are valid.
Previously we only considered as valid references the references to documents that are published, with the introduction of releases now references to documents that will be part of the same release, but not published yet, are valid, because they will be released at the same time.

This PR fixes that, allowing us to reference documents from the same release.

**What this doesn't fix?**
It doesn't fix references to documents that are part of other releases which will be released before this one, following the stack criteria.

Example:
This document is referencing a document that only exists in the same release:
[Validation passing in preview ](https://test-studio-git-corel-validate-references.sanity.dev/test/structure/author;66c88564-7b53-41ab-8827-99bc64ab0d04?perspective=rO2NP9B76)
[Validation NOT passing in `corel`](https://test-studio-git-corel.sanity.dev/test/structure/author;66c88564-7b53-41ab-8827-99bc64ab0d04?perspective=rO2NP9B76)

### What to review
Is this change correct?

### Testing

Create a document that only exists in a release, reference that document from another document in the same release, the validation should pass.

### DIFF
--- a/packages/sanity/src/core/preview/availability.ts
+++ b/packages/sanity/src/core/preview/availability.ts
@@ -6,7 +6,7 @@ import {combineLatest, defer, from, type Observable, of} from 'rxjs'
 import {distinctUntilChanged, map, mergeMap, reduce, switchMap} from 'rxjs/operators'
 import shallowEquals from 'shallow-equals'
 
-import {createSWR, getDraftId, getPublishedId, getVersionId, isRecord, isVersionId} from '../util'
+import {createSWR, getDraftId, getPublishedId, getVersionId, isRecord} from '../util'
 import {
   AVAILABILITY_NOT_FOUND,
   AVAILABILITY_PERMISSION_DENIED,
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/preview/availability.ts
```diff
@@ -150,7 +150,7 @@ export function createPreviewAvailabilityObserver(
   ): Observable<DraftsModelDocumentAvailability> {
     const draftId = getDraftId(id)
     const publishedId = getPublishedId(id)
-    const versionId = isVersionId(id) && version ? getVersionId(id, version) : undefined
+    const versionId = version ? getVersionId(id, version) : undefined
```
pedrobonamin, line 153: I think this check was an oversight given it forces the `id` received in this function to be a version id in order for all the version logic to work. Depending on the `id` and `version` check makes it more complicated to use in all the consumers.

packages/sanity/src/core/validation/validateDocumentWithReferences.ts
```diff
@@ -98,14 +107,19 @@ export function validateDocumentWithReferences(
     mergeMap((ids) => from(ids)),
   )
 
+  const versionId$ = document$.pipe(
```
pedrobonamin, line 110: gets the version id from the document. 
So if we are checking a document with the id `versions.rfoo.bar` it will return `rfoo` and will pass that versionId to the availability observer

bjoerge (approved):
Great stuff!

PR #8323, opened 2025-01-18
fix(core): duplicate should work with version documents

pedrobonamin:
### Description
Duplicate action should work with versions.
This PR updates how the id is created for the new document created by the duplicate action, if it's a version id it will preserve the version and it won't create a draft document.

### What to review
Is this correct?

### Testing
Open a version document.
Click on duplicate, it should create a new document in the same release.

Non version documents should continue being created as draft or live documents.

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/operations/duplicate.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/operations/duplicate.ts
@@ -1,3 +1,4 @@
+import {type SanityDocument} from '@sanity/types'
 import {omit} from 'lodash'
 
 import {getDraftId, getVersionFromId, getVersionId} from '../../../../../util'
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/operations/duplicate.ts
```diff
@@ -1,14 +1,17 @@
...
+import {getDraftId, getVersionFromId, getVersionId} from '../../../../../util'
 import {isLiveEditEnabled} from '../utils/isLiveEditEnabled'
 import {type OperationImpl} from './types'
 
 const omitProps = ['_createdAt', '_updatedAt']
 
 export const duplicate: OperationImpl<[baseDocumentId: string], 'NOTHING_TO_DUPLICATE'> = {
   disabled: ({snapshots}) => {
-    return snapshots.published || snapshots.draft ? false : 'NOTHING_TO_DUPLICATE'
+    if (snapshots.version) return false
```
juice49, line 11: Is this line needed? The next line checks `version` (`|| snapshots.version`) before returning `false` if it's truthy.

packages/sanity/src/core/store/_legacy/document/document-pair/operations/duplicate.ts
```diff
@@ -17,10 +20,19 @@ export const duplicate: OperationImpl<[baseDocumentId: string], 'NOTHING_TO_DUPL
       throw new Error('cannot execute on empty document')
     }
 
+    // When duplicating a version document we need to create it with a version id
+    const versionId = snapshots.version?._id ? getVersionFromId(snapshots.version._id) : null
```
juice49, line 24: This process looks good, but it could be nice to get rid of that nested ternary 😅.

pedrobonamin: Removed, I think we will need to update the function `getVersionFromId` name to something like `getReleaseIdFromDocumentId` it can be confusing with the `getVersionId` function

juice49: Nice. That's a good point. I was also thinking that if `getVersionFromId` could accept `undefined` as a parameter, this automatically becomes simple:

```ts
const versionId = getVersionFromId(snapshots.version?._id) ?? null
``` 

(Not a recommended change for this PR, just a thought).

pedrobonamin: Could be, I like this simplicity but I also think there is something nice about making the check explicit in the consumers.

PR #8324, opened 2025-01-18
feat(core): create references from version documents and new version documents.

pedrobonamin:
### Description

When working with releases it should be possible to create a document that only lives inside a release.
This could happen in two different places:
1) Clicking the **New Document Button**

https://github.com/user-attachments/assets/9f1b19ca-6be0-407c-a1b8-04281f02b941

2) Clicking on the **Create reference** button. 

https://github.com/user-attachments/assets/069c57b2-841b-44b8-9fb2-2d9e52a5d703

In both cases the document that will be created needs to belong to the release and it should not create a draft.

This PR fixes that, allowing us to create version inside releases.
To do this, it adds a `version` parameter to the pane router, when that parameter exists, the initial value id will be the correct version id.

When working on this update I noticed that we don't need anymore the `existsInBundle` property inside DocumentPaneProvider, so I removed that and replaced the uses of it.
Also, I noticed that we were using the `AddToReleaseBanner` for releases that are scheduled, which is a bit miss leading, so I refactored that and created a `ScheduledReleaseBanner` for ease of use.

### What to review
Are this changes correct?

### Testing
Navigate to a release and create a new document from the Add document button or from the create new reference button.

### Notes for release
n/a, this goes to the `corel` branch

### DIFF
--- a/packages/sanity/src/core/releases/util/util.ts
+++ b/packages/sanity/src/core/releases/util/util.ts
@@ -31,14 +31,6 @@ export function getDocumentIsInPerspective(
   return releaseId === perspective
 }
 
-/** @internal */
-export function versionDocumentExists(
-  documentVersions: ReleaseDocument[] = [],
-  releaseId: string,
-): boolean {
-  return documentVersions.some((version) => version._id === releaseId)
-}
-
 export function isDraftOrPublished(versionName: string): boolean {
   return versionName === 'drafts' || versionName === 'published'
 }
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
@@ -16,7 +16,7 @@ export const patch: OperationImpl<[patches: any[], initialDocument?: Record<stri
       const mutations = snapshots.version
         ? patchMutation
         : [
-            version.createIfNotExists({
+            version.create({
               _type: typeName,
               ...initialDocument,
             }),
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
@@ -13,16 +13,21 @@ export const patch: OperationImpl<[patches: any[], initialDocument?: Record<stri
       const patchMutation = version.patch(patches)
       // Note: if the document doesn't exist on the server yet, we need to create it first. We only want to do this if we can't see it locally
       // if it's been deleted on the server we want that to become a mutation error when submitting.
-      const mutations = snapshots.version
-        ? patchMutation
+      const ensureVersion = snapshots.version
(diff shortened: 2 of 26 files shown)


packages/sanity/src/core/releases/util/util.ts
```diff
@@ -31,14 +31,6 @@ export function getDocumentIsInPerspective(
   return releaseId === perspective
 }
 
-/** @internal */
```
pedrobonamin, line 34: This function is not used anymore.

packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
```diff
@@ -16,7 +16,7 @@ export const patch: OperationImpl<[patches: any[], initialDocument?: Record<stri
       const mutations = snapshots.version
         ? patchMutation
         : [
-            version.createIfNotExists({
+            version.create({
```
pedrobonamin, line 19: @juice49 this is the change I needed to fix the issue when creating version documents from the form, instead of clicking the `add to release` button.
The new flow allows users to create release documents the same way they are used to create drafts

packages/sanity/src/structure/panes/document/DocumentPaneContext.ts
```diff
@@ -39,12 +38,7 @@ export interface DocumentPaneContextValue {
   documentId: string
   documentIdRaw: string
   documentType: string
-  documentVersions: ReleaseDocument[] | null
   editState: EditStateFor | null
-  /**
-   * Whether the document being edited exists in the checked-out release.
-   */
-  existsInBundle: boolean
```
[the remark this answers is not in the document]
pedrobonamin: For example https://github.com/sanity-io/sanity/pull/8324/files#diff-f4149f712546c960867ebeb8809fe44d4555325cae90015f09ef34a0f0275a64R162

packages/sanity/src/structure/panes/document/documentPanel/banners/AddToReleaseBanner.tsx
```diff
@@ -49,56 +42,41 @@ export function AddToReleaseBanner({
...
-                        }
-                      >
-                        {children}
-                      </span>
-                    )
-                  },
-                }}
-              />
-            )}
+            <Translate
```
pedrobonamin, line 45: the scheduled banner has been moved to its own file

juice49 (approved):
Thanks, Pedro. This looks great!

packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/patch.ts
pedrobonamin: I'm not sure I follow the intention behind the `_empty_action_guard_pseudo_field_` here to be fair.
Previously I know we were using it with a `ifRevisionId` but this is missing here, wouldn't it have the same effect if we just remove this?

PR #8335, opened 2025-01-20
fix(core): update getIdPair to calculate the new id given the version

pedrobonamin:
### Description
While working on the [reference changes](https://github.com/sanity-io/sanity/pull/8324) I noticed the that getIdPair function returns the same id for a version document even if you are passing a different version.

Example:
```
getIdPair("versions.r1.foo", {version: "r2"})
resolves to: 
{
  publishedId: "foo",
  draftId: "drafts.foo"
  versionId: "versions.r1.foo"
 }
 
``` 

I think this could be miss leading, given if you provide a version you would want that version to be applied to the id.

So with this update:
```
getIdPair("versions.r1.foo", {version: "r2"})
resolves to: 
{
  publishedId: "foo",
  draftId: "drafts.foo"
  versionId: "versions.r2.foo"
 }
 
```

### DIFF
--- a/packages/sanity/src/core/util/__tests__/draftUtils.test.ts
+++ b/packages/sanity/src/core/util/__tests__/draftUtils.test.ts
@@ -4,6 +4,7 @@ import {describe, expect, it, test} from 'vitest'
 import {
   collate,
   documentIdEquals,
+  getIdPair,
   getPublishedId,
   getVersionFromId,
   getVersionId,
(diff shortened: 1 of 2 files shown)


packages/sanity/src/core/util/__tests__/draftUtils.test.ts
```diff
@@ -82,3 +83,44 @@ describe('getVersionFromId', () => {
     expect(getVersionFromId('my-document-id')).toBe(undefined)
   })
 })
+
+describe('getIdPair', () => {
+  test.each([
+    ['foo', undefined, {draftId: 'drafts.foo', publishedId: 'foo'}],
+    ['drafts.foo', undefined, {draftId: 'drafts.foo', publishedId: 'foo'}],
+    ['versions.r1.foo', undefined, {draftId: 'drafts.foo', publishedId: 'foo'}],
```
pedrobonamin, line 91: options is undefined so no version is returned in the ids

juice49 (approved):
Thanks! This does seem clearer 🙏.

PR #8336, opened 2025-01-20
chore(comments): exposes more comment details as internal exports

snorrees:
### Description

This PR exposes some more bits and bobs of the comments features as internal exports. These changes allows Create to wire up Comments (almost) without changing any code.

There are some minor code changes:
* Added a `mentionsDisabled` context variable so it is possible to disable the mentions functionality via context config.
* Renamed `buildRangeDecorations` -> `buildCommentRangeDecorations` since it now will be listed in `sanity` package
* In some components that consume comment code, the imports has been made slightly more explicit (ie, imports deeper from the tree, and not from root-index). This was done to accommodate exports restrictions in `comments/index.ts` instead of at the root `index.ts`

### What to review

Let me know if a link to the Create side PR is interesting to get the full context for how this will be used.
Once wired up things mostly "just worked". Pretty cool if you ask me, and shows (once again) how much power there is in this repo :) 

The most trouble comes from scroll behavior as scrollIntoView scrolls both x and y axis when no container element is provided. To avoid any real code changes, this has been worked around in Create and should not be a studio concern at this point in time.

### Testing

No change in behavior for the studio should be observed.

Try to add a comment in studio somewhere just as a regression test maybe.

### Notes for release

N/A: This is an internal change and shouldn't need a note.


packages/sanity/src/core/comments/components/pte/comment-input/CommentInputInner.tsx
```diff
@@ -128,8 +128,16 @@ export function CommentInputInner(props: CommentInputInnerProps) {
...
-    useCommentInput()
+  const {
+    canSubmit,
+    expandOnFocus,
+    focused,
+    hasChanges,
+    insertAtChar,
+    openMentions,
+    readOnly,
+    mentionOptions,
```
snorrees, line 139: Code changes in this file; add support for disabling mentions

packages/sanity/src/core/comments/components/pte/comment-input/CommentInputProvider.tsx
```diff
@@ -73,16 +79,22 @@ export function CommentInputProvider(props: CommentInputProviderProps) {
   }, [])
 
   const openMentions = useCallback(() => {
+    if (mentionOptions?.disabled) {
```
snorrees, line 82: Code changes in this file. Adds support for disabling mentions.

packages/sanity/src/core/comments/__workshop__/CommentInlineHighlightDebugStory.tsx
```diff
@@ -17,8 +17,8 @@ import {useCallback, useEffect, useMemo, useRef, useState} from 'react'
 import {useCurrentUser} from '../../store'
 import {type CommentDocument} from '../types'
 import {
+  buildCommentRangeDecorations,
```
snorrees, line 20: Renamed this function to include the word Comments to make it a bit clearer what its for (internal export at the public level)

packages/sanity/src/core/comments/index.ts
```diff
@@ -1,10 +1,33 @@
-export * from './components'
-export * from './constants'
-export * from './context'
-export * from './helpers'
-export * from './hooks'
-export * from './store'
+// this file exports the public api for comments (ends up in the sanity package) – be mindful of what gets re-exported
```
snorrees, line 1: I opted to handle hand-picked exports at this level instead of in the root  index.ts, as it is a bit easier to see whats available from each sub-directory.

This is done to expose as little as possible, while still allowing Create to reuse as much as possible.

robinpyon (approved):
> I opted to handle hand-picked exports at this level instead of in the root index.ts, as it is a bit easier to see whats available from each sub-directory.

I prefer this too, though I've observed some members have `@internal` tags even though they're not publicly exported (possibly as the result of the above, which may have happened after those tags were added). May be worth removing these?

```
COMMENT_REACTION_EMOJIS
CommentInputContextValue
CommentInputProvider
commentIntentIfDiffers
HighlightSpan
useCommentHasChanged
```

Otherwise this LGTM!

snorrees:
> > I opted to handle hand-picked exports at this level instead of in the root index.ts, as it is a bit easier to see whats available from each sub-directory.
> 
> I prefer this too, though I've observed some members have `@internal` tags even though they're not publicly exported (possibly as the result of the above, which may have happened after those tags were added). May be worth removing these?
> 
> ```
> COMMENT_REACTION_EMOJIS
> CommentInputContextValue
> CommentInputProvider
> commentIntentIfDiffers
> HighlightSpan
> useCommentHasChanged
> ```
> 
> Otherwise this LGTM!

Oh nice find! Yeah, I started by exporting everything then downselected.

packages/sanity/src/core/comments/hooks/useCommentsUpsell.ts
```diff
@@ -3,10 +3,6 @@ import {CommentsUpsellContext} from 'sanity/_singletons'
 
 import {type CommentsUpsellContextValue} from '../context/upsell/types'
 
-/**
```
bjoerge, line 6: any particular reason this was removed?

packages/sanity/src/core/comments/helpers.ts
```diff
@@ -10,6 +10,9 @@ export function useCommentHasChanged(message: CommentMessage): boolean {
   return useMemo(() => !isEqual(prevMessage, message), [prevMessage, message])
 }
 
+/**
+ * @internal
```
bjoerge, line 14: I see a mix of using `@internal` + `@hidden` and just `@internal` in changed doc blocks, is that intentional? What is determining whether it should be `@hidden` if so?

packages/sanity/src/core/comments/index.ts
bjoerge: appreciate this comment 👍🏼  also aligns with our goal to avoid barrel exports

bjoerge (approved):
Just a couple of minor comments/questions, and the branch needs a rebase. Otherwise LGTM!

packages/sanity/src/core/comments/hooks/useCommentsUpsell.ts
snorrees: Its not exported at the package level and never was I think, so un-needed. This prompted me to double check the original export list:

CommentInput,
  type CommentIntentGetter,
  COMMENTS_INSPECTOR_NAME,
  CommentsEnabledProvider,
  CommentsIntentProvider,
  type CommentsIntentProviderProps,
  CommentsProvider,
  useCommentsEnabled,


  type CommentsIntentProviderProps was still missing so I've re-added it

packages/sanity/src/core/comments/helpers.ts
snorrees: Ah, my bad. When I started this I was a bit confused on what was needed (there was a lot of errors to work through doing breath first export). Ill remove hidden when for these internals

bjoerge (approved):

PR #8338, opened 2025-01-20
refactor: moving calendar inputs from scheduled pub. to `core/components`

jordanl17:
### Description
The initial change which moved the calendar based components from scheduled publishing to `ui-components` has been copied from COREL branch.

However, `ui-components` is not the right place for a new component type such as calendar, so this PR additionally moves these components to `core/components` and the further change diff are updating the imports.

### Testing
Only a refactor, so existing tests act as good regression coverage

### Notes for release
N/A


bjoerge (approved):
Thanks!

PR #8342, opened 2025-01-20
chore(core): adds perspective provider + split usePerspective to smaller hooks

pedrobonamin:
### Description

Adds `PerspectiveProvider` and updates uses of `usePerspective` to expose the values from the provider.
Removes `setPerspective` from the `usePerspective` hook and exposes a new `useSetPerspective` hook to get the `setPerspective` function.

Adding a Provider allows us to override the values for `selectedPerspective` and `excludedPerspectives` values from where necessary, for example, the diff modal.

The `StudioProvider` adds the `GlobalPerspectiveProvider` which will listen for the perspective and excluded perspective from the `router`

- Moved `toggleExcludedPerspective` and `isPerspectiveExcluded` to its own `useExcludedPerspective` hook

Left the remaining properties in the usePerspective since I didn't feel it made sense to move them out. If you disagree, do let me know!

### What to review

Does the split make sense?

### Testing

The test that are there should be enough, however, make sure that nothing jumps out at you in the preview.


packages/sanity/src/core/perspective/GlobalPerspectiveProvider.tsx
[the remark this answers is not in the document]
pedrobonamin: The `GlobalPerspectiveProvider` is intended to be used only once and in from the `StudioProvider` , we should not export this one from `sanity`, it should be only internal.
It "handles" the logic of listening to the `router` and passing down that to the `PerspectiveProvider`.

Then, in places where we need to include an override we can use the `PerspectiveProvider` directly.[ Like in this commit](https://github.com/sanity-io/sanity/pull/8342/commits/fc514a5a57edbb3ab23bee01b71e9713ee0bfd8f) (sorry just pushed this, I run out of time yest)

packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
```diff
@@ -72,24 +74,18 @@ export const DocumentEventsPane = (props: DocumentPaneProviderProps) => {
...
-
   return (
     <EventsProvider value={value}>
-      <DocumentPaneProvider
-        {...props}
-        historyStore={historyStoreProps}
-        forcedVersion={forcedProviderVersion}
-      />
+      {historyVersion ? (
+        <PerspectiveProvider
```
pedrobonamin, line 80: @juice49 this is what I think you could do similar in the diff view, instead of setting the override we could use the provider

juice49: Yes! This is much better. I'll refactor this into the diff view once it's merged.

juice49 (commented):
This is great! Thanks so much, Pedro 🙌.

PR #8344, opened 2025-01-21
test: fix PTE copy/paste test case

christianhg:
The "Removed unicode characters" test failed because the length of the expected document was suddenly longer than the length of the received document. This is caused by a change to `@portabletext/block-tools` which now allows you to pass in a `keyGenerator` to use when normalizing blocks. PTE takes advantage of this to make sure that keys generated when pasting don't suddenly differ from other keys generated by the editor.

Before this change, `block-tools` would simply generate a random key for the block using `randomKey(12)` and then assign a `${blockKey}{childIndex}` key to each child of the block. This resulted in keys longer than 12 chars. However, now that `block-tools` accepts a custom `keyGenerator` it now either uses that or `randomKey(12)` for child keys. This is why this test suddenly failed. Child keys were expectd to have a length of 13, but now have length of 12.

Screenshot 2025-01-21 at 11 07 58


bjoerge (approved):
Thank you for jumping on this @christianhg!

PR #8345, opened 2025-01-21
fix(core): issue opening references with published perspective

pedrobonamin:
### Description

When opening a reference from a document in `published` perspective the studio crashes.
This happens because the `observeDocumentPair` function was receiving the `selectedPerspectiveName` which could be "published". instead of the `releaseId` which is what we want here.

This updates and fixes the issue. 

See how it crashes here: https://test-studio-git-corel.sanity.dev/test/structure/author;f219b738-f1b9-410c-8c26-d9eb8925dc34;b8512898-d875-4331-a27a-ab915a559a99%2Ctype%3Dauthor%2CparentRefPath%3DbestFriend?perspective=published

But it passes here:  https://test-studio-git-corel-354.sanity.dev/test/structure/author;f219b738-f1b9-410c-8c26-d9eb8925dc34;b8512898-d875-4331-a27a-ab915a559a99%2Ctype%3Dauthor%2CparentRefPath%3DbestFriend?perspective=published


jordanl17 (approved):
Nice catch!

PR #8350, opened 2025-01-21
fix(graphql): do not send references to union type function

cngonzalez:
### Description
We discovered a bug where a union-style reference definition that referenced itself (e.g.:
```
defineType({
  type: 'document',
  name: 'union',
  fields: [
    defineField({
      name: 'body',
      type: 'array',
      of: [
        {
          type: 'reference',
          to: [{type: 'a'}, {type: 'b'}, {type: 'union'}],
        },
        {type: 'a'},
      ],
    }),
  ],
})
```

would not correctly bail out of the getUnionDefinition recursion and cause a `maximum calls exceeded` error. This PR better aligns our reference definition logic to what we do with type generation by adding a few lines to code to the `getReferenceDefinition` function and avoiding the union definition logic altogether.

### What to review
1. The changed code -- any gotchas?
2. Should we include a `references` array? I've commented it out for now just because it would make our new snapshots a bit noisy. We've done without this array for quite a while, but it seems nice to have.

### Testing
Please review the updated snapshots and make sure all is well.

### Notes for release
Fixes an issue where referencing a type of the same kind of document within a document would cause a `Maximum call stack size exceeded` error during GraphQL schema generation.


packages/sanity/src/_internal/cli/actions/graphql/extractFromSanitySchema.ts
```diff
@@ -393,8 +393,24 @@ export function extractFromSanitySchema(
...
+      unionTypes.push({
+        kind: 'Union',
+        name,
+        types: targetTypes,
+      })
+    }
+
+    return {
+      type: name,
+      // references: targetTypes,
```
cngonzalez, line 411: Would we like to have this? It seems like what we would like to do, based on the usage in getUnionDefinition, but this implementation adds quite a bit of content to our testing snapshots.

rneatherway (commented):
This seems reasonable based on what I learned while doing that recent PR that this improves upon. However, I'm not an expert in this code and haven't looked at the typegen code so I think it's best if I defer to Sindre on this one.

packages/sanity/src/_internal/cli/actions/graphql/extractFromSanitySchema.ts
sgulseth: Can you make it "optional"?
```
const def = {type: name, ...base}
if (targetTypes !== undefined) {
  def.references = targetTypes
}
```

sgulseth (approved):

sgulseth (dismissed):

sgulseth (approved):

PR #8351, opened 2025-01-21
fix: fixing issue with counting documents in archived or published releases

jordanl17:
### Description
Inside the release overview table, for an open release the count of documents is read from `release.documentsMetadata.documentCount`. This was currently being used correctly for open releases, however for archived (archived or published) releases, this should be read from `release.finalDocumentStates`.

This PR adds in the conditional key

### Testing
Tested by verifying that for an archived release that has documents, when trying to delete the release from the overview list, a dialog appears confirming that all the document versions will be deleted (previously this was never showing for archived and published releases since incorrectly the releases appeared as though it was empty)

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -232,18 +232,21 @@ export function ReleasesOverview() {
     )
   }
 
-  const renderRowActions = useCallback(({datum}: {datum: TableRelease | unknown}) => {
-    const release = datum as TableRelease
+  const renderRowActions = useCallback(
+    ({datum}: {datum: TableRelease | unknown}) => {
+      const release = datum as TableRelease
 
-    if (release.isDeleted) return null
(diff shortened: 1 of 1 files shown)


RitaDias (approved):
Nice!! 

https://github.com/user-attachments/assets/571e4125-274e-4ca2-b5f0-9f42027a47b6

Smallest of nitpicks which might be outside of the scope for this PR, but I noticed that the actions menu seems to have a scroll? When I feel it shouldn't at least 😄

jordanl17:
> Nice!!
> 
>  Screen.Recording.2025-01-22.at.14.29.37.mov 
> Smallest of nitpicks which might be outside of the scope for this PR, but I noticed that the actions menu seems to have a scroll? When I feel it shouldn't at least 😄

@RitaDias you know that's been annoying me for so long, and it turned out to be a very simple fix, so I just committed directly to corel rather than go via PR. Hope that's ok: https://github.com/sanity-io/sanity/commit/22446d3c12d96577e214f5e590bc867dc181ac6d

PR #8364, opened 2025-01-22
chore(ci): add workflow config for auto-updating studios

cngonzalez:
**Note: This PR should be closed for now.** If there are staged changes in `next` that are incompatible with our last version (for example, adding a type that is allowed in Portable Text, but wasn't allowed before) the tests will break. We'll need to re-think about how to test auto-updating end-to-end -- this may need to be an infrastructural change on our end.

### Description
Adds a workflow for AUS to help catch any discrepancies between versions. This should help us catch issues that only crop up with auto-updating studios.

Ideally, we should also pair this with a deployment that deploys an auto-updating studio (to help our developers quickly debug), but that can come later.

### What to review
I'm not a Playwright configuration expert by any means, so please let me know if there's a better way to add a new project to our configuration. Please also let me know if there are any gotchas in the Github CI scripting that I may have missed.

### Testing
This one is a bit hairy to test, but maybe the most efficient way to do this would be:
1. Pull down this branch
2. Check out the commit in this branch that experimentally updates our packages to vite 6 (`f4395c46b89958a3709286bbdc67eb41a8f89407`)
3. Run `pnpm install`
4. Run `pnpm e2e:setup && pnpm e2e:build --auto-updates` (the command in the Github workflow)
5. Run `pnpm test:e2e` or `pnpm e2e:start` to see the broken studio.

Alternately, you can see this screenshot of a CI run on this branch with the "broken" commit to verify that this should be signal that auto-updating builds are broken.
Screenshot 2025-01-24 at 9 22 12 AM


RitaDias (commented):
Unfortunately I don't know much about this either, however, as for the steps to reproduce this I can't seem to do it?

This is from when I try to pull the commit that you mention:
image

Also, it looks to me (tried running the CI on the latest push you have since I saw it failing) that the CI is still failing? With the same issue that you have as a screenshot of what the error should be.

This screenshot is from this morning when the [CI finished](https://github.com/sanity-io/sanity/actions/runs/12951517817/job/3620718401):
image

cngonzalez:
Hiya @RitaDias ! Thanks for the thoughtful attempt (apologies -- didn't mean to keep you on as a reviewer when I placed it back in draft) but discovered some larger issues with testing AUS that I've added to the PR description. Going to place this on hold for now.

PR #8367, opened 2025-01-22
fix: set strictRequires to auto when building auto updating studios

bjoerge:
### Description
When building auto-updating Studios with Vite 6 we ran into an issue caused by the default of `commonjsOptions.strictRequires` that changed from `auto` to `true` between v5 and v6 (mentioned in the advanced section here: https://vite.dev/guide/migration#advanced)

This caused the vendorized react bundle to place all named exports on the `default` export instead of exporting them all individually which then broke named imports.

This PR explicitly sets `commonjsOptions.strictRequires` to `auto` which will produce the same output in both Vite v5 and v6.

We mitigated the earlier issue by reverting back to to Vite 5 (#8362). Now that we know the exact cause, and the fix, it would be good to have this merged so we can safely upgrade to Vite 6 again without affecting auto-updating Studios.

### Notes for release
n/a –internal

### DIFF
--- a/packages/sanity/src/_internal/cli/server/buildVendorDependencies.ts
+++ b/packages/sanity/src/_internal/cli/server/buildVendorDependencies.ts
@@ -223,6 +223,7 @@ export async function buildVendorDependencies({
     define: {'process.env.NODE_ENV': JSON.stringify('production')},
 
     build: {
+      commonjsOptions: {strictRequires: 'auto'},
       minify: true,
       emptyOutDir: false, // Rely on CLI to do this
       outDir: path.join(outputDir, VENDOR_DIR),


binoy14:
Any chance we can get https://github.com/sanity-io/sanity/pull/8364 working first?

bjoerge:
> Any chance we can get https://github.com/sanity-io/sanity/pull/8364 working first?

Oh, yes 💯 – would love to see that one merged first

juice49 (approved):
Thanks for tracking this down 🙏. Looks good to me. I agree with Binoy, it would be good to land #8364 first.

RitaDias:
cc @bjoerge @binoy14 
Since the PR that [Carolina was closed due to some larger issues she discovered](https://github.com/sanity-io/sanity/pull/8364#issuecomment-2616486689) I just want to double check what should be the state of this PR (if we want to put it on ice as well or if we want to move it along review wise)

binoy14:
> cc @bjoerge @binoy14 Since the PR that [Carolina was closed due to some larger issues she discovered](https://github.com/sanity-io/sanity/pull/8364#issuecomment-2616486689) I just want to double check what should be the state of this PR (if we want to put it on ice as well or if we want to move it along review wise)

I think we can merge this instead of waiting for the tests

binoy14 (approved):

PR #8369, opened 2025-01-22
feat(core): update events api routes and config flags, fix events sorting

pedrobonamin:
### Description
This PR updates the events api routes that we were using.
The routes now will be:
- Documents events: `data/history/{datasetName}/events/documents/{ids}` 
- Release events `data/history/{datasetName}/events/releases/{id}`

It also modifies the config flag we have added for the events api and **enables** it by default for documents events for all the studios.
It's kept as a flag in case we need to turn it off.

Releases events are not ready to be used by everyone yet, so a new flag has been added to handle this. It will be enabled by default in the `playground` studio.

It also updates the sort algorithm for the events, given the timestamp validation is not the only check necessary, because edit and publish events now have the same timestamp and the order was incorrect.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -225,7 +225,11 @@ export default defineConfig([
     dataset: 'playground',
     plugins: [sharedSettings()],
     basePath: '/playground',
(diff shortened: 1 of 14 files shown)


juice49 (approved):
Looks good to me!

PR #8371, opened 2025-01-22
chore: upgrade vitest to v3

bjoerge:
### Description

New attempt at upgrading vitest to v3, rebased against next since previous review :)

Note: Tests seems to run significantly slower between vitest v2 and v3. Didn't seem worth spending time looking into why since simply increasing test timeouts seems to have fixed it (see c066b138813915fd337a471d3269c7f63c259f0e), and since tests are sharded they usually don't take more than a few mins anyway.

### What to review
Does the test pass?

### Testing
Tests should pass

### Notes for release

n/a – internal

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,6 +71,6 @@ jobs:
       - name: Test
         id: test
         run: |
-          pnpm test:vitest --shard=${{ matrix.shardIndex}}/${{ matrix.shardTotal }}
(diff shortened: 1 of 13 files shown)


ricokahler (dismissed):

binoy14 (approved):

binoy14 (approved):
Tests do seem to pass so hard to tell if it's something with vitest or our side.

PR #8382, opened 2025-01-23
refactor(core): adds perspective provider + split usePerspective to smaller hooks

RitaDias:
### Description

- Adds `PerspectiveProvider` and updates uses of `usePerspective` to expose the values from the provider.
  - Adding a Provider allows us to override the values for selectedPerspective and excludedPerspectives values from where necessary, for example, the diff modal.
  - The `StudioProvider` adds the `GlobalPerspectiveProvider` which will listen for the perspective and excluded perspective from the router
 
- Moves `setPerspective`
Removes setPerspective from the `usePerspective` hook and exposes a new `useSetPerspective` hook to get the `setPerspective` function.
- Moves `toggleExcludedPerspective` and `isPerspectiveExcluded` to its own `useExcludedPerspective` hook

Left the remaining properties in the usePerspective since I didn't feel it made sense to move them out. If you disagree, do let me know!

### What to review

Does the split make sense? Should we split more?

### Testing

The test that are there should be enough, however, make sure that nothing jumps out at you in the preview.


packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
```diff
@@ -72,24 +74,18 @@ export const DocumentEventsPane = (props: DocumentPaneProviderProps) => {
...
-  }, [historyVersion])
-
   return (
     <EventsProvider value={value}>
-      <DocumentPaneProvider
-        {...props}
-        historyStore={historyStoreProps}
-        forcedVersion={forcedProviderVersion}
-      />
+      {historyVersion ? (
```
RitaDias, line 79: cc @pedrobonamin and @juice49 
just echoing an old comment from the previous PR by Pedro:

> this is what I think you could do similar in the diff view, instead of setting the override we could use the provider

packages/sanity/src/core/releases/hooks/__tests__/__mocks__/useExcludedPerspective.mock.ts
```diff
@@ -0,0 +1,16 @@
+import {type Mock, type Mocked, vi} from 'vitest'
```
pedrobonamin, line 1: Nit follow up, maybe we should move this mock to `core/perspective`?

packages/sanity/src/core/releases/hooks/__tests__/__mocks__/usePerspective.mock.ts
```diff
@@ -1,15 +1,15 @@
-import {type Mock, type Mocked, vi} from 'vitest'
+import {type Mock, type Mocked} from 'vitest'
```
pedrobonamin, line 1: Nit follow up, maybe we should move this mock to `core/perspectiveà?

pedrobonamin (approved):
Great work as always @RitaDias , thank you!

packages/sanity/src/structure/panes/document/DocumentEventsPane.tsx
juice49: Thanks. I'll refactor it into my code one this is merged 🙂.

packages/sanity/src/core/releases/hooks/__tests__/__mocks__/useExcludedPerspective.mock.ts
RitaDias: Definitely should be moved, will do

packages/sanity/src/core/releases/hooks/__tests__/__mocks__/usePerspective.mock.ts
RitaDias: Definitely should be moved, will do

juice49 (approved):
This is great!

PR #8392, opened 2025-01-23
fix(core): support searching references in same release

pedrobonamin:
### Description

When searching to add a reference in a document, documents that are available in the same release should show.

https://github.com/user-attachments/assets/8958ee12-04a8-417b-9bfc-8fb76286f822

### Testing
Manual testing for now:
- Create a document in a release.
- Go to a document that could reference the previously created document.
- The document should show in the list.


jordanl17 (approved):
Simple yet impactful change! Nice

PR #8394, opened 2025-01-23
feat(cli): add non-studio app template

cngonzalez:
### Description

This PR adds a template for applications that are not Sanity Studios. Because we are still considering requirements and scope, this app is a very minimal Vite / React application, but this may be expanded in the future.

We would still like these applications to be behind the `sanity` CLI runtime, so we've made some tweaks to allow `sanity dev`, `sanity build` and so on to run a Vite app.

A TLDR of what's included in this PR:
1. Boilerplate / static files for a non-studio application:
a. The Document / index.html file
b. The `sanity.cli.ts|js` file
c. The `app.js` entrypoint
2. An additional parameter to CLI config to configure a non-studio application, `__experimental_coreAppConfiguration`. Right now it only determines framework and an entry path. 
3. Changes to the `sanity init` and template bootstrapping commands. If a non-studio application template is detected, it will not prompt for project-specific information or add studio-specific files and dependencies.
4. Various checks throughout server commands, mostly to ensure help text is accurate, and that the correct static files are being used.

### What to review
This PR is not a complete vision. There will be fast follows to add the following:
1. Organization-specific information and deployment
2. Adding specific scripts to non-studio templates -- for example, `dev` should default to running in a framed application.
3. Ways to pass down additional parameters from configuration. We don't have the mechanisms right now to do things the Studio does, like pass down an application title, or whether to run in React Strict Mode.
4. Changes to eslint config generation so it's not studio-specific.

Beyond that, I'm interested in opinions on things like: 
1. the cli config option as a flag
2. The location of various templates

### Testing

I didn't see ways to easily add tests for this -- our command tests (like `dev` and `build`) do not seem to accept CLI configuration options. Probably the best way to test is to:
1. `mkdir core-app`
2. `cd core-app`
3. `pnpx sanity@3.72.2-canary.32+df28986158 init --template core-app`
4. After creating the app, change the `sanity` entry in your `package.json` `devDependencies` to `3.72.2-canary.32+df28986158` and `pnpm install`
5. Play around with `pnpm dev`, `pnpm build`, `pnpm start`

OR you could the following:

1. Pull down this branch
2. `cd ./packages/sanity`
3. Copy the `pwd` of where you are for future use (for me, it is `/Users/carolina/code/sanity/packages/sanity`)
5. Run `pnpm build` and `pnpm watch` (if you'd like to make changes)
7. In another terminal window, run `{pwd}/bin/sanity init --template core-app` (so something like`/Users/carolina/code/sanity/packages/sanity/bin/sanity init --template core-app`)
8. Observe the initialization flow
9. `pnpm link {your output from step #3}` so something like `pnpm link /Users/carolina/code/sanity/packages/sanity`
10. Play around with `pnpm dev`, `pnpm build` and `pnpm start`

### Notes for release
None, internal


packages/@sanity/cli/src/actions/init-project/bootstrapLocalTemplate.ts
```diff
@@ -119,29 +122,37 @@ export async function bootstrapLocalTemplate(
...
     variables,
   })
 
   // ...and a CLI config (`sanity.cli.[ts|js]`)
-  const cliConfig = await createCliConfig({
-    projectId: variables.projectId,
-    dataset: variables.dataset,
-    autoUpdates: variables.autoUpdates,
-  })
+  const cliConfig = isStudioTemplate
```
binoy14, line 131: 🤔 I am wondering if we should flip this logic, instead of checking `isStudioTemplate` which is basically checking the name and currently we allow users to build templates if someone builds `github.com/<id>/studio-core-app` it will categorize it as core app which is unintended behavior of creating a different kind of template. 

I think as is, we should flip it so that the template checks for a specific name `=== core-app` and currently user's can't make custom core apps. In future I was envisioning that we check the dependencies of the template and find the proper packages to confirm this which seems more stable but I don't know how feasible that is.

binoy14 (commented):
high level it makes sense to me, but wondering what you think about flipping the logic

packages/@sanity/cli/src/actions/init-project/bootstrapLocalTemplate.ts
cngonzalez: Hmm, so the current logic to verify a studio template is in `determineStudioTemplate` (called further up in the file), which does check for the specific name `core-app`. Right now, the `github.com/<id>/studio-core-app` wouldn't get to this point in logic, because that is not a local template (there's a different file for remote templates).

Likewise, the `determineStudioTemplate` used in `initProject` also wouldn't fall into this because I believe you pass a URL. 

I'm not sure what would be best to add but happy to take suggestions.

cngonzalez: Per our convo yesterday, I've swapped the conditional -- hopefully it's more readable.

I'm still on the side of keeping creating core app templates fairly locked down for now. In the future we probably want a more robust way to let people make remote templates, or register templates. It's hard to determine how to architect that until we see which way the CLI is oging.

packages/sanity/src/_internal/cli/commands/app/startCommand.ts
```diff
@@ -0,0 +1,84 @@
...
+import {type StartPreviewServerCommandFlags} from '../../actions/preview/previewAction'
+import {isInteractive} from '../../util/isInteractive'
+import {getDevAction} from '../dev/devCommand'
+
+const helpText = `
+Notes
+  Changing the hostname or port number might require a new CORS-entry to be added.
+
+Options
+  --port <port> TCP port to start server on. [default: 3333]
```
binoy14, line 16: Might be a good quality of life of improvement if it's different port from studio but if it's too much work than don't worry about it.

```suggestion
  --port <port> TCP port to start server on. [default: 3334]
```

binoy14 (approved):
Tested it locally with the canary and everything looks good. Code looking good as well, thanks for the effort!

packages/sanity/src/_internal/cli/commands/app/startCommand.ts
jordanl17: Yeh, I agree this would be a nice fast-follow improvement

jordanl17 (approved):
High level the changes make sense to me

PR #8395, opened 2025-01-24
fix(core): include `_originalId` in `groq2024` search, update refs adapter

pedrobonamin:
### Description

Fixes a bug that occurs when using `groq2024` search and trying to create a reference to a document that is only a draft, no published version of this document exists.

The problem is that when using `groq2024` the results include the document published id, even if it's a published or a draft, I think this is because it is using a perspective to query the data.

To fix this we include the `_originalId` which represents the real id of the document, which is the `drafts` or `version` (coming soon)

https://github.com/user-attachments/assets/d33b71f5-c302-4498-9a8a-66a85496e703

### What to review
Is this change correct?

### Testing
groq 2024 tests have been updated.

### Notes for release
fixes a regression in which references to non published documents fails when using `groq2024` search strategy

### DIFF
--- a/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
+++ b/packages/sanity/src/core/form/studio/inputs/client-adapters/reference.ts
@@ -186,6 +186,13 @@ export function referenceSearch(
   })
   return search(textTerm, {includeDrafts: true}).pipe(
     map(({hits}) => hits.map(({hit}) => hit)),
+    map((docs) =>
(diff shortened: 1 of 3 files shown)


RitaDias (approved):
Nice! This looks good to me, though since @juice49 is the one that touched on this lately might be worth him giving a once-over :)

pedrobonamin:
Closing this one because this changes were added as part of `feat/content-releases`

PR #8403, opened 2025-01-24
fix: release dateTime picker bugs

jordanl17:
### Description
Fixes a number of issues:
* Prevents scheduled publish time of dates in past from being selected
* Warning messages when the scheduled publish time is in the past
* Positioning of dateTime calendar selects across studio
* Minor refactor of `ReleaseForm`, `CopyToNewRelease` and `CreateReleaseDialog`
* All publish date selects allow for textinput use to type the date
* Selected schedule dates will always have 0 seconds

| Scenario | Demo |
|--------|--------|
| Create release | newReleaseDatePastPR |
| Copy to Release | copyToReleaseDatePastPR |
| Edit release type | typePickerDatePastPR |
| Confirm Schedule | confirmScheduleDatePastPR | 

### What to review
* Sensible ways of handling and showing that dates need to be in the future?
* Updated ui component's `Dialog` to use the ui component's Buttons (so we can use tooltips on these buttons). Think there could be any issues here?

### Testing
Adjusted existing tests; will fast-follow new tests for date in the past scenarios

### Notes for release
N/A


RitaDias:
I haven't reviewed the code itself, but how does this play with timezones? Was this something you double checked? (specifically related to putting a date for earlier than the current time)

RitaDias (commented):
A thought I also had was if we should have e2e / component tests around the dialogs for the dates blocking the "create" action. probably something we can create a story for.

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -53,17 +57,24 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
...
   const buttonRef = useRef(null)
 
+  const [referenceElement, setReferenceElement] = useState<HTMLInputElement | null>(null)
+
   useImperativeHandle<HTMLInputElement | null, HTMLInputElement | null>(
     forwardedRef,
     () => ref.current,
   )
 
+  useEffect(() => setReferenceElement(ref.current), [])
```
RitaDias, line 76: Why do we need this? 🤔

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -116,17 +127,24 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
             <Popover
               constrainSize={constrainSize}
               data-testid="date-input-dialog"
+              referenceElement={referenceElement}
               portal
               content={
                 <Box overflow="auto">
                   <FocusLock onDeactivation={handleDeactivation}>
+                    {inputValue && isPastDisabled && isPast(new Date(inputValue)) && (
```
RitaDias, line 135: could we move this outside to not have it in the render? use a useMemo / etc?

packages/sanity/src/core/releases/components/documentHeader/dialog/CopyToNewReleaseDialog.tsx
```diff
@@ -43,25 +46,40 @@ export function CopyToNewReleaseDialog(props: {
...
   const handleCreateRelease = useCallback(async () => {
+    // re-evaluate if date is in past
+    // as dialog could have been left idle for a while
+    if (getIsScheduledDateInPast(release)) {
+      toast.push({
+        closable: true,
+        status: 'warning',
+        title: tRelease('schedule-dialog.publish-date-in-past-warning'),
+      })
+      setRerenderDialog((cur) => cur + 1)
```
RitaDias, line 75: ooh, do we need this? This feels like something of a hack 😬 
I don't understand why we need it, but I don't quite like it 🥲

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -34,14 +34,25 @@ export const ReleaseScheduleButton = ({
...
+  const isScheduledDateInPast = useCallback(() => {
+    return isBefore(publishAt || new Date(), new Date())
+  }, [publishAt])
 
   const handleConfirmSchedule = useCallback(async () => {
     if (!publishAt) return
 
+    if (isScheduledDateInPast()) {
+      // rerender dialog to recalculate isScheduledDateInPast
+      setRerenderDialog((cur) => cur + 1)
```
RitaDias, line 52: Same as this comment 🙏 
I don't know if this is the best way of going about it? What is the use case, sorry I don't understand why we are doing it and it might be obvious 😅

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -99,9 +119,17 @@ export const ReleaseScheduleButton = ({
...
+    const _isScheduledDateInPast = isScheduledDateInPast()
+
     return (
       <Dialog
         id="confirm-schedule-dialog"
+        /**
+         * rerenderDialog should force this function to rerun
+         * since the selected scheduled date was in the future when selected
+         * but at time of submit it is in the past
+         */
```
RitaDias, line 131: I don't understand why we need to re-render the whole calendar though 🤔 or the whole component rather.

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
jordanl17: So this is for the edge case that you select a time that is in the future, say 1 min in the future, but you take 1 minute to complete the dialog before submitting. The `handleSubmit` callbacks check again against the current time, if they see that you are now in the past, there needs to be a way of getting the component to rerender so that the submit button now shows as disabled and has a warning or tooltip. Since no actual existing state has changed, I added this state, to flush a rerender of the component and cause `isScheduledDateInPast` to be rerun and repainted. Does this make sense - happy to sync

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
jordanl17: - [ ] Memoize in `DateTimeInput`

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
jordanl17: So this was something new I learned doing this:

> Do not write or read ref.current during rendering, except for [initialization.](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) This makes your component’s behavior unpredictable.
[From here](https://react.dev/reference/react/useRef#caveats)

This is I think symptomatic that the sanity UI Popover should accept a react ref as `referenceElement` rather than a HTML element.

But by putting this into state and setting in the `useEffect`, it's ensured that the updates to the `referenceElement` happen after the initial render cycle, so that the `ref.current` value is accurate and aligned with the react state model

jordanl17:
> I haven't reviewed the code itself, but how does this play with timezones? Was this something you double checked? (specifically related to putting a date for earlier than the current time) (given that we just use the Date() object it should be fine I was just thinking about it)

Good callout - I've got a followup PR that resolves the timezone issues that builds on this one

packages/sanity/src/ui-components/dialog/Dialog.tsx
```diff
@@ -34,8 +35,8 @@ export type DialogProps = Pick<
   bodyHeight?: BoxHeight
   children?: ReactNode
   footer?: {
-    cancelButton?: Omit<ComponentProps<typeof UIButton>, 'fontSize' | 'padding'>
-    confirmButton?: Omit<ComponentProps<typeof UIButton>, 'fontSize' | 'padding'>
+    cancelButton?: Omit<ComponentProps<typeof Button>, 'fontSize' | 'padding'>
+    confirmButton?: Omit<ComponentProps<typeof Button>, 'fontSize' | 'padding'>
```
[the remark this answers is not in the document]
jordanl17: Yeh this was also my concern. Perhaps to avoid fallout from this with COREL I can just render a warning elsewhere in the dialog and we can come back to this once opt in-out is completed

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
jordanl17: - [ ] Add why the useEffect is there as a code comment

RitaDias: I see, might need to add a task to improve this at some, it just feels bad that we need to add a useeffect where the only thing we are doing is this. Which might be fine but since useEffects tend to tank performance I'm aways just 👁️ when I see them

RitaDias (approved):
🚀
Thank you for going through this!

PR #8404, opened 2025-01-24
refactor(core): adds confirmation when there are no documents in a release (for archiving and delete)

RitaDias:
### Description

- Added the ability to open the confirmation dialog even when there are no documents in a release
- Updated the translation approach for plural approach + added new translations for when it's zero

### What to review

Does this make sense? 

### Testing

Nothing needed outside of what currently exist


jordanl17 (approved):
Nice improvement

PR #8406, opened 2025-01-24
refactor(sanity): remove references to defunct `_version` field

juice49:
### Description

The `_version` field is no longer used to signal that a document is a version (this is instead handled via the `versions.` id prefix). This branch removes defunct references to the `_version` field.


pedrobonamin (approved):
Thank you @juice49 !

PR #8408, opened 2025-01-27
fix(core): revert lost changes from #7597

pedrobonamin:
### Description

In `corel` we lost some changes applied in #7597, this PR restores them.


RitaDias (approved):
great catch 🚀

PR #8409, opened 2025-01-27
chore(core): revert timeline changes added to support version documents.

pedrobonamin:
### Description

The legacy Timeline is soon to be deprecated once the events API has proven to be stable.
In `corel` we added some initial changes to support viewing version documents history through the legacy Timeline.
Later, we introduced the Events API which is the one that should be used to view version documents history, given it correctly supports them .

This PR removes the initial changes to clean the diff and remove risk on the Timeline.
It also adds a warning message if users try to see a version document history using the legacy timeline.

### DIFF
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1718,6 +1718,12 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   /** Title for error when the timeline for the given document can't be loaded */
   'timeline.error.load-document-changes-title':
     'An error occurred whilst retrieving document changes.',
+  /** Description for error when the timeline for the given document can't be loaded */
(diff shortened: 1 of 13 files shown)


jordanl17:
@pedrobonamin just wondering on the wording of the error message, and whether this is too 'technical'. I'm not sure if we have guidelines for this, but good to pass by Design quickly to make sure this is fine?

pedrobonamin:
> @pedrobonamin just wondering on the wording of the error message, and whether this is too 'technical'. I'm not sure if we have guidelines for this, but good to pass by Design quickly to make sure this is fine?

Good callout, but users you should only see this error if a dev has disabled the events api by disabling the config flag, which they should not do.
And I think we will be removing the flag soon once we are happy with events api and is stable.

Let's bring it up in slack and see how we can improve the wording. Will merge this, changing the wording is a simple change  we can do as a follow up

PR #8410, opened 2025-01-27
fix(core): remove `version` from `formBuilder`, replace for `usePerspective`

pedrobonamin:
### Description

When we started with the `corel` branch `version` was added as a value that the `formBuilder` exposed, this was used by the internal fields when they needed to fetch data, like the `reference` inputs.

Later, we added the `<PerspectiveProvider>` with the `usePerspective` hook, this PR removes the `version` from the `formBuilder` to simplify the usage by devs and knowing which one to use. 
Allowing us to pick the same value from two different context could lead to hard to track bugs and code inconsistency.


juice49 (approved):
This is great. Thanks so much, Pedro 🙏.

PR #8412, opened 2025-01-27
chore: small `corel` changes 

pedrobonamin:
### Description

This PR includes multiple small changes that apply to the `corel` branch after reviewing the diff we are introducing to `next`.
Some changes attempt to reduce the diff footprint.
It adds some comments to the branch, which I think they help to clarify the intention of some of the changes.

It also adds a `TODO: COREL` comments to help us identify the pending changes from the `corel` branch.

The recommended way to review this branch would be to check individually each commit. 

### What to review
Are this changes correct?

### Testing
Existing tests should cover everything needed


packages/sanity/src/core/scheduledPublishing/components/scheduleItem/PreviewWrapper.tsx
```diff
@@ -154,7 +154,7 @@ const PreviewWrapper = (props: Props) => {
                     <DocumentStatusIndicator
                       draft={previewState?.draft}
                       published={previewState?.published}
-                      versions={previewState?.versions}
```
RitaDias, line 157: if schedulePublishing should not support versions I wonder why we did this 🤔 
I just want to double check that we are sure about this change?

packages/sanity/src/core/preview/createPreviewObserver.ts
```diff
@@ -42,6 +42,8 @@ export function createPreviewObserver(context: {
     value: Previewable,
     type: PreviewableType,
     options: {
+      // TODO: COREL - Is this missing somehow? Maybe it was implemented and we lost this
```
RitaDias, line 45: What is missing? Us making changes in the preview pacakge? 🤔

packages/sanity/src/structure/panes/types.ts
```diff
@@ -11,6 +11,7 @@ export interface BaseStructureToolPaneProps<T extends PaneNode['type']> {
   isActive?: boolean
   pane: Extract<PaneNode, {type: T}>
   /**
+   * TODO: COREL - Remove this after updating sanity-assist to use <PerspectiveProvider>
```
RitaDias, line 14: Do we have a linear story for this? :)

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -126,6 +126,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   const perspective = usePerspective()
 
   const {isReleaseLocked, selectedReleaseId, selectedPerspectiveName} = useMemo(() => {
+    // * TODO: COREL - Remove this after updating sanity-assist to use <PerspectiveProvider>
```
RitaDias, line 129: Do we have a linear story for this? :)

RitaDias (approved):
Added a couple of comments, about the todos I just want to make sure we have a linear story for it so we don't end up having to solely rely on the comments :)

Nothing is blocking so here's a ✅ 
Thank you for going through this

packages/sanity/src/core/preview/createPreviewObserver.ts
pedrobonamin: This comment was more for me to keep digging, not meant to be pushed. 
I noticed that we are receiving `perspective` here as an option but it's not used anywhere.

RitaDias: Ah I see!
Looking at the gitblame it looks to have been added by @bjoerge, do you have any memory ~of this place~? :)

packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
pedrobonamin: added `corel-380` thanks!

RitaDias: nice!

packages/sanity/src/core/store/_legacy/document/document-pair/utils/actionsApiClient.ts
```diff
@@ -1,5 +1,6 @@
 import {type SanityClient} from '@sanity/client'
 
+// TODO: COREL - Replace once releases API are stable.
```
RitaDias, line 3: Since we have been discussing this on the team for a while I assumed that we had already a story for this, but since I'm actually not sure I'm here to ask if we do 😂

pedrobonamin: I think we don't have it in linear, will add it!

packages/sanity/src/core/preview/createPreviewObserver.ts
pedrobonamin: For what I checked this option was not passed by any consumer.
Also, the preview works just fine without this because it is using the version id of the document to query the correct data. 
I have removed this.

PR #8413, opened 2025-01-27
fix(structure): disable publish and unpublish actions for releases

pedrobonamin:
### Description

Publish and unpublish actions are not available in version documents, this PR updates them to return `null` if it's being used in a version document.

This actions should happen through the release publish action, not individually in each document,.


jordanl17 (approved):
Looks good

PR #8429, opened 2025-01-28
fix(presentation): use live content API for loaders

stipsan:


packages/sanity/package.json
```diff
@@ -227,7 +227,6 @@
     "lodash": "^4.17.21",
     "log-symbols": "^2.2.0",
     "mendoza": "^3.0.0",
-    "mnemonist": "0.39.8",
```
stipsan, line 230: One less chonky dependency 😮‍💨

packages/sanity/src/presentation/loader/LiveQueries.tsx
```diff
@@ -55,55 +41,18 @@ export interface LoaderQueriesProps {
...
-          }
-          return next
-        }),
-      MIN_LOADER_QUERY_LISTEN_HEARTBEAT_INTERVAL,
-    )
-    return () => clearInterval(interval)
-  }, [])
-
-  useEffect(() => {
+  useEffect((): (() => void) => {
```
stipsan, line 53: This typing:

```ts
useEffect((): (() => void) => {
           // ^
})
```
Is there to avoid easy mistakes, as further down this is returned:

```ts
return nextComlink.start()
```
Calling `.start()` returns an unsubscribe handler. This is a bit implicit and easy to miss. Should someone refactor this useEffect later and miss the `return` statement then TS will catch it 😮‍💨

packages/sanity/src/presentation/loader/LiveQueries.tsx
```diff
@@ -195,69 +114,35 @@ export default function LoaderQueries(props: LoaderQueriesProps): React.JSX.Elem
...
 
   return (
     <>
-      {Object.entries(liveQueries).map(([key, {query, params, perspective}]) => (
+      {[...liveQueries.entries()].map(([key, {query, params, perspective}]) => (
         <QuerySubscription
-          key={`${key}${perspective}`}
-          projectId={clientConfig.projectId!}
-          dataset={clientConfig.dataset!}
+          key={`${liveEvents.resets}:${key}`}
```
stipsan, line 136: The `liveEvents.resets` ensures that if a `reconnect` or `restart` live event happens then `QuerySubscription` components here start from a clean slate. This helps ensure their internal tracking of what `liveEventsMessages` they can ignore, as they happened before the component mounted, and which ones to look for a `lastLiveEventId`, works correctly.

packages/sanity/src/presentation/constants.ts
```diff
@@ -10,26 +11,13 @@ export const EDIT_INTENT_MODE = 'presentation'
 export const MAX_TIME_TO_OVERLAYS_CONNECTION = 3_000 // ms
 
 // The API version to use when using `@sanity/client`
-export const API_VERSION = '2023-10-16'
+export const API_VERSION = apiVersion
```
stipsan, line 14: Aligning on the `apiVersion` used here increases the potential effect of https://github.com/sanity-io/client/pull/990

rdunk (approved):
Awesome, this is a great improvement and another great benefit of the LCAPI. It should also help alleviate the fear users have when they inspect the network tab when using Presentation. 😆 

Thanks for all the comments, really helpful as usual.

PR #8437, opened 2025-01-28
fix(core): move searchProvider to wrap the item preview

pedrobonamin:
### Description

Fixes an issue in `tasks` that when a document was added as the target then the task will fail to render due to the missing provider. 
Screenshot 2025-01-28 at 17 20 05


RitaDias (approved):
Great catch!

PR #8443, opened 2025-01-29
refactor(sanity): remove defunct release migration

juice49:
### Description

During development, release documents were initially stored at a different path. To aid the transition to the final path, Bjørge created a neat migration process that invisibly moved the documents.

At this stage, the migration is no longer needed.

### What to review

Does everything look okay?

### Testing

Run Studio, make sure releases are still fetched correctly.


pedrobonamin (approved):
Thanks you Ash!

PR #8444, opened 2025-01-29
fix(structure): make form readOnly in published view

pedrobonamin:
### Description
Makes form readOnly if the selected perspective is `published` and the document type is not live editable


juice49 (approved):
Looks good. Thanks, Pedro!

PR #8449, opened 2025-01-29
feat(sanity): use new name for `sanity::versionsOf`

rneatherway:


juice49 (approved):
Thanks, Robin! This looks great.

PR #8457, opened 2025-01-30
feat: upsell content releases

jordanl17:
### Description
 
| Scenario | Demo |
|--------|--------|
| Free trial, growth, ent at limit of plan releases | growthEntAtLimit |
| Free trial, growth, ent at limit of plan releases - copying | growthEntLimitCopy |
| Free plan | freeDialog |
| Free trial, growth, ent under limit of plan releases | growthEntUnderLimit | 

Some notes on how this works:
`useReleasesUpsell` exposes 2 relevant functions.  
`onReleaseLimitReached` is used in `createReleaseOperationsStore` to catch the error that comes back from CL when the limit is reached. This sets the limit in the `ReleasesUpsellContext` and then if the limit is already reached, will cause the upsell dialog to show.  
This covers how the limit is first fetched - the server action is triggered, and then the error caught and limit stored.  
Once the limit is reached then `guardWithReleaseLimitUpsell` is used. This wraps all the relevant locations - eg, `onClick` for the menu item to unarchive, `onClick` for creating a new release, or copying a version to one. This will first check if releases is enabled (ie if this is a free plan). If it is enabled then will check if there is a limit, and if the limit is not exceeded will call the callback, otherwise it will optionally throw the StudioReleaseLimitExceededError or return false to inform the callee that the limit has been reached so that they can decide how to reconcile the action (eg close the modal, not continue with the operation, and not trigger an error toast).  
Since all the API calls are wrapped in the guard, if the plan is free then no need to make a server action to receive the limit, can infer the limit immediately as 0

### What to review
* Does the `ReleasesUpsellProvider` look ok
* The changes to interpolate on `UpsellDialog`, does it all seem safe
* The way that `guardWithReleaseLimitUpsell` guards the call of it's provided callback against the release limit
* How the `StudioReleaseLimitExceededError` is thrown and then caught in `createReleaseOperationsStore`

### Testing
Updated existing tests with mocks of `guardWithReleaseLimitUpsell` to let them pass
Manual so far - will fast follow with further tests of the limit cases

### Notes for release
N/A


packages/sanity/src/core/hooks/useProjectSubscriptions.ts
jordanl17, line 1: Taken from `useFeatureEnabled` and much the same in terms of how the request is subscribed and then result cached

packages/sanity/src/core/perspective/navbar/currentGlobalPerspectiveLabel.tsx
```diff
@@ -16,6 +16,7 @@ import {type SelectedPerspective} from '../types'
 const AnimatedMotionDiv = ({children, ...props}: PropsWithChildren<any>) => (
   <motion.div
     {...props}
+    layout="preserve-aspect"
```
jordanl17, line 19: Adds slide animation to Global perspective picker

pedrobonamin: Nice one!

packages/sanity/src/core/hooks/useProjectSubscriptions.ts
pedrobonamin: This was great, thanks Jordan. 
I reverted the change because we want to add a check for `ReleasesPlus` plan, not only enterprise.

We can bring it back later if we want by reverting this [commit](https://github.com/sanity-io/sanity/pull/8457/commits/7655995225f1f835767f06b8ecce502fe720bc63)

packages/sanity/src/core/studio/upsell/upsellDescriptionSerializer/UpsellDescriptionSerializer.tsx
```diff
@@ -176,127 +178,156 @@ function ImageBlock(
   )
 }
 
+const interpolateChildrenText = (interpolation?: InterpolationProp) => (children: ReactNode) => {
+  if (!children || !interpolation) return children
+
+  const childrenArray = Array.isArray(children) ? children : [children]
```
stipsan, line 184: I'd use https://react.dev/reference/react/Children for these transforms on `children`

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/ReleaseRevertButton.tsx
```diff
@@ -208,22 +212,33 @@ const ConfirmReleaseDialog = ({
   )
 }
 
+// TODO:  This is going to be disabled until we have the proper "releases plus" flag
```
RitaDias, line 215: Do we have a linear story? Afraid this gets lost and nothing is done when we need to do it...

packages/sanity/src/core/releases/tool/detail/__tests__/ReleaseDashboardFooter.test.tsx
```diff
@@ -57,13 +57,15 @@ describe('ReleaseDashboardFooter', () => {
   })
 
   describe('for a published release', () => {
-    test('shows revert button for asap release', async () => {
+    // Revert button is now hidden in the UI until feature flag "releases plus" is added - see ReleasesRevert component
```
RitaDias, line 60: On the story where we rework this / add to this we should make sure to mention we have these tests being skipped

RitaDias (approved):
smaller nits but nothing blocking :)

RitaDias:
> From the screen recordings, it looks like the user may encounter the guard after they have entered data in a form, but before that data has been saved anywhere. E.g. when creating a new release.
> 
> Do we have anything in place to preserve the data they entered when this occurs?

This is one of the feedback provided by knut, unless something has changed since he gave it (I haven't done any work there), no there is not

pedrobonamin:
> > From the screen recordings, it looks like the user may encounter the guard after they have entered data in a form, but before that data has been saved anywhere. E.g. when creating a new release.
> > Do we have anything in place to preserve the data they entered when this occurs?
> 
> Good point! This is one of the feedback provided by knut, unless something has changed since he gave it (I haven't done any work there), no there is not

I think we can do that as a follow up, given is not a straight away solution.
In most scenarios users should not run into this I think we can guard it by trying to do a `dry run` if the user has more than 2 releases so we can validate it.
However, I think this is not necessary for the opt in, as most users won't know about this change yet.

pedrobonamin (approved):
Thank you @jordanl17 this is great!

PR #8462, opened 2025-01-31
fix(sanity): remove Actions API override

juice49:
### Description

In the past, Content Releases did not work correctly with Actions API. Therefore, the Releases plugin ensured the Actions API was switched off when it was in use.

This has been fixed, and it is no longer necessary for Actions API to be switched off when Content Releases is in use.


pedrobonamin (approved):
Thank you Ash!

PR #8466, opened 2025-01-31
feat(core): only use `vX` client if strictly necessary

pedrobonamin:
### Description

In the initial implementation of content releases we modified the default studio client to use `vX` this was needed because all the new functions and versions documents checks are only available in `vX` for now.

Now, we are aiming to merge this to `next` in the coming week, given that we don't want to use the `vX` client everywhere because of the risk this involves we are adding this changes in which the client will use `vX` only if it is strictly necessary.
This will be when a user is doing actions over a `version` document or is using anything inside the releases plugin.

This PR has been divided in logical pieces and will probably be easier to review commit per commit.

#### Changes in `presentation` tool
I've added the checks where I noticed that we are using perspectives and could lead to errors if the client is not using the releases client (vX for now) in this [commit](https://github.com/sanity-io/sanity/pull/8466/commits/95ba3fe9c12412128654b2d69fff82403d287fc5)
I've left out the following files, which I think don't require this.
- packages/sanity/src/presentation/useMainDocument.ts
- packages/sanity/src/presentation/usePreviewUrl.ts
- packages/sanity/src/presentation/useVercelBypassSecret.ts
- packages/sanity/src/presentation/preview/SharePreviewMenu.tsx

### Testing

`vX` client should only be used when a perspective is defined or when operating in a version document.
This can be verified by opening a document in `draft` and doing edits, the changes should be done using the 2024 client.
If you switch to a `version` now the client should use `vX`
The same with search.


packages/sanity/src/core/preview/useLiveDocumentSet.ts
```diff
@@ -21,14 +21,21 @@ const INITIAL_VALUE = {loading: true, documents: []}
 export function useLiveDocumentSet(
   groqFilter: string,
   params?: QueryParams,
+  options: {
+    // how to insert new document ids. Defaults to `sorted`
+    insert?: 'sorted' | 'prepend' | 'append'
+  } = {},
```
bjoerge, line 27: thoughts on making apiVersion part of options here?

packages/sanity/src/core/search/weighted/createWeightedSearch.ts
```diff
@@ -38,14 +43,12 @@ export const createWeightedSearch: SearchStrategyFactory<WeightedSearchResults>
...
 
-    return client.observable
-      .withConfig({
-        /** @todo remove defined apiVersion once perspective is supported in
-         * non-experimental content lake API
-         */
-        apiVersion: 'vX',
-      })
-      .fetch<SanityDocumentLike[]>(query, params, options)
+    const apiVersion = isReleasePerspective(options?.perspective as string | string[] | undefined)
```
bjoerge, line 46: Nice way of feature-detecting what version to use!

bjoerge: Maybe add a //todo: remove once releases is opt-out?

packages/sanity/src/core/store/_legacy/document/document-pair/serverOperations/publish.ts
```diff
@@ -20,7 +20,7 @@ export const publish: OperationImpl<[], DisabledReason> = {
       throw new Error('cannot execute "publish" when draft is missing')
     }
 
-    return actionsApiClient(client).observable.action(
+    return actionsApiClient(client, idPair).observable.action(
```
bjoerge, line 23: I think we should make a note to remove these added params after backend APIs stabilize

bjoerge (approved):

PR #8469, opened 2025-01-31
chore(core): remove unnecessary exports

pedrobonamin:
### Description

Removes unnecessary exports and removes the `@internal` tag from the releases telemetry events to avoid exporting them in the future by mistake.


RitaDias (approved):
Thanks for this!!

PR #8471, opened 2025-02-03
feat(structure): edit `liveEdit` documents through the global draft perspective

pedrobonamin:
### Description

This PR updates how the `liveEdit` documents behave and allows users to edit them while being on the `draft` perspective.
Instead of forcing them to switch to the `published` perspective it shows them with the **published badge** that they are gonna be editing the published document and not the draft. It also adds information in the draft badge on why that one is disabled.

We have two different scenarios to consider for the `liveEdit` documents, draft exists or not.
In most cases the draft won't exist, given it's a live edit.

#### No draft
| Publish perspective | Drafts perspective |
|--------|--------|
| Screenshot 2025-02-03 at 09 30 51| Screenshot 2025-02-03 at 09 30 42| 

#### Draft exists
| Publish perspective | Drafts perspective |
|--------|--------|
|Screenshot 2025-02-03 at 09 31 43| Screenshot 2025-02-03 at 09 31 37|

### DIFF
--- a/packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
+++ b/packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
@@ -1,7 +1,7 @@
 import {type PreviewValue, type SanityDocument, type SchemaType} from '@sanity/types'
 import {omit} from 'lodash'
 import {type ReactNode} from 'react'
-import {combineLatest, from, type Observable, of} from 'rxjs'
+import {combineLatest, from, type Observable} from 'rxjs'
 import {map, mergeMap, scan, startWith} from 'rxjs/operators'
 
 import {type PerspectiveStack} from '../../perspective/types'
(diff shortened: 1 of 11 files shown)


packages/sanity/src/core/preview/utils/getPreviewStateObservable.ts
```diff
@@ -66,9 +66,7 @@ export function getPreviewStateObservable(
     isRaw: false,
   },
 ): Observable<PreviewState> {
-  const draft$ = isLiveEditEnabled(schemaType)
-    ? of({snapshot: null})
-    : documentPreviewStore.observeForPreview({_id: getDraftId(documentId)}, schemaType)
+  const draft$ = documentPreviewStore.observeForPreview({_id: getDraftId(documentId)}, schemaType)
```
pedrobonamin, line 69: Thoughts on this change?
We were hiding the `draft` in the live edit documents but that is used also to reflect the status dots in the preview, so I think it's maybe better to show the draft if it exists so users can detect it on the list view

bjoerge: Agree! The draft could suddenly appear from other sources, so its good to subscribe to it anyway

bjoerge (approved):
This is really great! 💯

PR #8472, opened 2025-02-03
feat(sanity): include only necessary fields in release projection

juice49:
### Description

This branch makes the releases projection more precise, including only fields that are used by the app.


pedrobonamin (approved):
Thank you @juice49 , this matches the ReleaseDocument interface

PR #8473, opened 2025-02-03
fix(core): use `MarkdownPlugin` instead of `behaviors={[...]}`

christianhg:
### Description

PTE has a new concept of `@portabletext/editor/plugins` which, going forward, will be the preferred way to add additional Behaviors to the editor. The benefit is that you don't have to manually import and remember to spread the `coreBehaviors` when configuring the `initialConfig.behaviors` array.

### What to review

Markdown Behaviors should still work, but feel free to take them for a spin to be sure.

### Testing

n/a

### Notes for release

n/a

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -14,7 +14,7 @@ import {
   useEditor,
   usePortableTextEditor,
 } from '@portabletext/editor'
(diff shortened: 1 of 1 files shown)


christianhg:
> Do we need to update the docs to reflect this change? This wasn't user configurable yet right?

It's purely internal for now so no need to update any docs.

christianhg:
> > Do we need to update the docs to reflect this change? This wasn't user configurable yet right?
> 
> It's purely internal for now so no need to update any docs.

Hi @pedrobonamin . Any chance this can get merged?

pedrobonamin (approved):
Missed your previous comment, please go ahead 😄

PR #8482, opened 2025-02-03
feat(sanity): load release document `error` field

juice49:


juice49:
* **#8482** <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/8482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/sanity-io/sanity/8482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `next`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/sanity/src/core/releases/__fixtures__/release.fixture.ts
```diff
@@ -87,3 +87,20 @@ export const activeUndecidedRelease: ReleaseDocument = {
...
+  _createdAt: '2023-10-10T08:00:00Z',
+  _updatedAt: '2023-10-10T09:00:00Z',
+  state: 'active',
+  metadata: {
+    title: 'undecided Error Release',
+    releaseType: 'undecided',
+    description: 'undecided Error Release description',
+  },
+  error: {
+    message: 'An unexpected error occurred during publication.',
```
jordanl17, line 104: Q: more a musing and not blocking for this work, but I wonder how we can get these sorts of error messages to work with i18n. Perhaps in the future it best we avoid just passing through the message and focus on error types so we can localise them

packages/sanity/src/core/releases/store/createReleaseStore.ts
```diff
@@ -135,11 +145,30 @@ export function createReleaseStore(context: {
...
+
+/**
+ * @internal
+ */
+export function releaseStoreErrorCount(): OperatorFunction<ReleasesReducerState, number> {
+  return pipe(
+    switchMap(({releases}) =>
+      from(releases.values()).pipe(
+        filter((release) => release.state === 'active'),
+        filter((release) => typeof release.error !== 'undefined'),
```
jordanl17, line 168: COULD: roll this as a single filter, then since the check on state `active` and `typeof` `error` in used elsewhere too, could abstract that as some util... thinking maybe if the api simplifies or changes in the future that makes it easier

packages/sanity/src/core/releases/tool/overview/ReleasesOverviewColumnDefs.tsx
```diff
@@ -86,10 +89,34 @@ export const releasesOverviewColumnDefs: (
...
+          {typeof error !== 'undefined' && (
+            <Tooltip content={<Text size={1}>{t('failed-publish-title')}</Text>} portal>
+              <Text size={1}>
+                <ToneIcon icon={ErrorOutlineIcon} tone="critical" />
+              </Text>
+            </Tooltip>
+          )}
+        </Flex>
+      ),
+    },
```
jordanl17, line 115: SHOULD: we only want this to show for active releases, right? Either could return a different list of cols for active and archived; or perhaps could just add in the `datum.state === 'active'` check here too

jordanl17 (changes requested):
Only the one comment in the overview column defs where I think the condition might need to change. Otherwise 🌶️

packages/sanity/src/core/releases/__fixtures__/release.fixture.ts
juice49: I agree in principle, however `error.message` reflects a technical error from the server and is not intended to be presented to the user in the UI (we do output it, for debugging purposes). When we know about other types of error that will be encapsulated here, we'll need to use some heuristic to show a relevant (localised) message in the UI.

packages/sanity/src/core/releases/store/createReleaseStore.ts
```diff
@@ -45,6 +49,12 @@ const QUERY_PROJECTION = `{
     "title": "",
     "releaseType": "${DEFAULT_RELEASE_TYPE}",
   }),
+  // Content Lake initially encoded non-error states as {error: {message: ""}}. This projection
+  // ensures the error field only appears if the document has a non-empty error message.
+  ...select(
```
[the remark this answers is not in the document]
juice49: Thanks. Hoping we will be able to eliminate this in the future, but I assume we have documents in this format in the wild currently.

packages/sanity/src/core/releases/store/createReleaseStore.ts
juice49: Agree it would make sense to abstract this to some guards:

```ts
function isActiveRelease(maybeActiveRelease): maybeActiveRelease is ReleaseDocument & { state: 'active' }
```

```ts
function isErrorRelease(maybeActiveRelease): maybeActiveRelease is ReleaseDocument & { error: { message: string } }
```

These could then be composed to abstract the logic around whether an error should be displayed.

packages/sanity/src/core/releases/tool/overview/ReleasesOverviewColumnDefs.tsx
juice49: I've updated the code to simply perform a check on the cell datum, as you suggested. We should abstract this in the future, but I decided not to for now, as we're about to learn more about the errors that can be stored and our heuristics will possibly change.

jordanl17 (dismissed):
Perfecto

jordanl17 (approved):

PR #8484, opened 2025-02-04
fix(core): add default values to releases metadata

pedrobonamin:
### Description

Releases metadata is not a system property, so in some cases this info could not exist.
This modifies the query we use to fetch them and adds default values if the object doesn't exist.

### Testing
Create a release with no metadata outside of the studio, the studio should load with the default values.


RitaDias (approved):
Let's go

PR #8491, opened 2025-02-04
fix(comments): make target ref point at published id and add document…

bjoerge:
### Description

This changes the way comments and tasks reference their targets. Previosly, a _ref to the _version id_ would be made, but with these changes we instead:

- Always reference the *published* id of the target (i.e. `_ref: 'versions.*'` is disallowed the same way as `_ref: drafts.*` is)
- When fetching comments for a particular version, we explicitly opt-in using a groq filter. If we're not fetching for a version, the groq filter will exclude comments where `target.documentVersionId` is set.

### What to review
Did I cover all the cases?

Note: I wasn't actually able to make a task that produced a reference to a version id, but I removed the code that seemingly should have make sure that we set the ref to the version id.

### Notes for release
n/a


pedrobonamin (approved):
Thank you @bjoerge !

PR #8494, opened 2025-02-04
feat(core): add responsiveness to overview releases page

RitaDias:
### Description

This has been cleared with Marius!

- When the screen is smaller (and the text starts getting crunched), we hide the calendar and leave it in a popover next to the page title. see video

https://github.com/user-attachments/assets/78edded8-2f02-46ba-97a3-7d7c74252b20

### What to review

Looks good?

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -7,6 +7,7 @@ import {
   Card,
   Container,
   Flex,
+  Inline,
   Stack,
   Text,
   useMediaIndex,
@@ -32,6 +33,7 @@ import {getReleaseTone} from '../../util/getReleaseTone'
 import {ReleaseMenuButton} from '../components/ReleaseMenuButton/ReleaseMenuButton'
 import {Table, type TableRowProps} from '../components/Table/Table'
 import {type TableSort} from '../components/Table/TableProvider'
+import {CalendarPopover} from './CalendarPopover'
 import {
   DATE_SEARCH_PARAM_KEY,
   getInitialFilterDate,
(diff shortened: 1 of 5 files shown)


RitaDias:
Converting back to draft, need to fix the unit tests

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -268,27 +284,37 @@ export function ReleasesOverview() {
...
+          <CalendarFilter
+            disabled={loading || releases.length === 0}
+            renderCalendarDay={ReleaseCalendarFilterDay}
+            selectedDate={releaseFilterDate}
+            onSelect={handleSelectFilterDate}
+          />
+        </Card>
+      </Flex>
+    )
+  }
```
pedrobonamin, line 300: Could we memoize this as a function or move it outside of this function so it's not defined on render?

```suggestion
  const calendarFilter = useMemo(() => {
    return (
      <Flex flex="none">
        <Card borderRight flex="none" disabled>
          <CalendarFilter
            disabled={loading || releases.length === 0}
            renderCalendarDay={ReleaseCalendarFilterDay}
            selectedDate={releaseFilterDate}
            onSelect={handleSelectFilterDate}
          />
        </Card>
      </Flex>
    )
  }, [loading, releases, releaseFilterDate, handleSelectFilterDate])
```

pedrobonamin (commented):
Thank you for doing this Rita !!
Great improvement 👏🏻

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
RitaDias: I tried doing this but had issues with typing, let me have a look

RitaDias: When I tried at the time it must have had something to do with something else (probably typing issues in the props), in any case, I've done this, thank you Pedro!

pedrobonamin (approved):

PR #8497, opened 2025-02-04
fix(core): show accurate history for live-edit documents

pedrobonamin:
### Description

Live edit documents are shown in the Form even if the release selected is not the `published` one, you can see a live edit document through the `drafts` perspective.

This PR updates how the id for the Events store is calculated, so the published Id is passed to the history store and not the draft when previewing a live edit document which is missing a draft


bjoerge (approved):
Nice, thank you @pedrobonamin!

PR #8501, opened 2025-02-04
chore(e2e): retry e2e tests when running in CI

bjoerge:
### Description
Adds a small config change that retries failing e2e tests 4 times in the hopes of alleviating some recurring pains caused by flake.

### What to review
Does it make sense? Will it take effect?

### Testing
changes test config, so n/a

### Notes for release

n/a – internal


RitaDias (approved):
LET'S GOOOOOO

PR #8502, opened 2025-02-04
chore(test): add node 22 to test matrix

bjoerge:
### Description
Node 22 has been out for a closer to a year now, so time to add it to our test matrix I guess :)

### Notes for release
n/a internal


jordanl17 (approved):
Nice one! Should we see if all tests also pass on 23 too, is there harm adding that also?

PR #8506, opened 2025-02-05
chore: restricting scheduled release UI

jordanl17:
### Description
With upcoming changes to the state machine, an ASAP release will briefly pass through the `scheduling` and `scheduled` states before moving to `publishing` and `published`.

The condition we reuse for determining whether to show the scheduled release UI elements (eg CTAs to schedule/unschedule) currently only checks on the state. This won't be enough when the above state machine change is made - ASAP releases will wrongly at times be shows as scheduled releases.

This change just restricts the condition, requiring that `metadata.releaseType` is `scheduled` first.

### What to review
Does restricting the conditions for a scheduled release with a `releaseType` assertion make sense?

### Testing
No regressions expected. Manually tested a few hot spots


RitaDias (approved):
Good catch!

PR #8513, opened 2025-02-05
fix: ensures comments permission check only requests project memberships once

snorrees:
### Description

Currently, every time document form value updates, comments-enabled projects will re-fetch `/projects/<projectId>` via `projectStore.get()`.

Furthermore, `useUserListWithPermissions` uses a global grants cache which could/will cause problem in a multi-workspace studio setting.

This PR reworks `useUserListWithPermissions` so it only fetches projects _once_, as well as removes the global cache in favor of rxjs `shareReply` as a cache layer instead.

### What to review

Scrutinize the rxjs changes.I am a rxjs noob.
* Is it ok to use `shareReply` to cache like this?
* Are we ok with caching project & grants like this in the first place? 
* As written, memberships and grants will _not_ be kept up to date until the component that runs `useUserListWithPermissions`(`CommentsProvider`) is remounted. Personally, I think this is an ok real-time concession.

### Testing

* Network tab
* Edit a document
* there should no longer be a /projects/<projectId> request every time the document changes

### Notes for release

Not needed.

### DIFF
--- a/packages/sanity/src/core/hooks/useUserListWithPermissions.ts
+++ b/packages/sanity/src/core/hooks/useUserListWithPermissions.ts
@@ -3,7 +3,7 @@ import {type SanityDocument} from '@sanity/client'
 import {type User} from '@sanity/types'
 import {sortBy} from 'lodash'
 import {useEffect, useMemo, useState} from 'react'
-import {concat, forkJoin, map, mergeMap, type Observable, of, switchMap} from 'rxjs'
+import {concat, forkJoin, map, mergeMap, type Observable, of, shareReplay, switchMap} from 'rxjs'
 
 import {
   type DocumentValuePermission,
@@ -54,8 +54,6 @@ export interface UserListWithPermissionsOptions {
   permission: DocumentValuePermission
 }
 
-let cachedSystemGroups: [] | null = null
-
 /**
  * @beta
  * Returns a list of users with the specified permission on the document.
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -71,6 +71,77 @@ jobs:
       - name: Test
         id: test
         run: |
-          pnpm test:vitest --shard=${{ matrix.shardIndex}}/${{ matrix.shardTotal }}
+          ARGS="--test-timeout=60000 --retry 4 --shard=${{ matrix.shardIndex}}/${{ matrix.shardTotal }}"
```
snorrees, line 74: No idea whats happend with the diff, hold on, trying to fix

packages/sanity/src/core/hooks/useUserListWithPermissions.ts
```diff
@@ -72,14 +70,15 @@ export function useUserListWithPermissions(
 
   const [state, setState] = useState<UserListWithPermissionsHookValue>(INITIAL_STATE)
 
-  const list$ = useMemo(() => {
+  const [users$, systemGroup$] = useMemo(() => {
```
snorrees, line 73: Moved these to a separate memo, so the observables are stable across document changes.

bjoerge:
I'll have to defer to @hermanwikner or @pedrobonamin  who seems to be the original authors of this.

pedrobonamin (approved):
Thank you @snorrees this is a huge improvement!

PR #8514, opened 2025-02-05
fix(releases): timezone fixes in content releases

jordanl17:
### Description
| Before | After |
|--------|--------|
| timezone1OLD | timezone1NEW |
| timezone2OLD | timezone2NEW |
| timzone3OLD | timezone3NEW |
| timezone4OLD | timezone4NEW |
| timzone5OLD | timezone5NEW | 

### Testing
There aren't really any tests around testing of timezone in releases. I've created a task to do this

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
+++ b/packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
@@ -18,6 +18,7 @@ import FocusLock from 'react-focus-lock'
 import {Button} from '../../../../ui-components/button/Button'
 import {Popover} from '../../../../ui-components/popover/Popover'
 import {useTranslation} from '../../../i18n'
+import useTimeZone from '../../../scheduledPublishing/hooks/useTimeZone'
 import {type CalendarProps} from './calendar/Calendar'
 import {type CalendarLabels} from './calendar/types'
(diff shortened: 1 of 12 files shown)


packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
```diff
@@ -18,6 +18,7 @@ import FocusLock from 'react-focus-lock'
 import {Button} from '../../../../ui-components/button/Button'
 import {Popover} from '../../../../ui-components/popover/Popover'
 import {useTranslation} from '../../../i18n'
+import useTimeZone from '../../../scheduledPublishing/hooks/useTimeZone'
```
bjoerge, line 21: Not critical, but should we consider moving this into core?

bjoerge (approved):
Looks great! Just a minor comment, nothing blocking :)

packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
jordanl17: Yeh I think that's a great idea. I'll do that as a fast follow and anything else in scheduled publishing that feels more generalised now with releases

PR #8516, opened 2025-02-05
refactor(vision): simplify perspective handling, no default perspective

bjoerge:
### Description

This adds some smaller changes to Vision as part of the upcoming default perspective change (from `raw` => `published`):

- Currently, raw is selected as the default. With these changes, the default becomes "unspecified" (i.e. no perspective param is sent with the request, effecively using the default for the given API version). I can imagine this can be useful for debugging. This also means published will soon become the default.
- In accordance with the above, it should then also be possible to select “No perspective” from the dropdown.
- With these changes, you'll get today’s date as the default API version. Currently, we’re using the last element in a [hard coded array of dates](https://github.com/sanity-io/sanity/blob/next/packages/@sanity/vision/src/apiVersions.ts#L2). This means you'll get `v2022-03-07` at the moment, and chances are you'll be given an outdated default until we remember to update this list. I can’t think of any issue with using a rolling date here. The default can always be customized on a per-studio basis if that's desirable. Let me know if I've missed a reason we should keep it hard coded.

### What to review
Note: This is is ready for review, ~but not yet ready to merge~. Should be ready to go now!

### Testing
Tests in this repo is sparse, so please look at the code and give it a go in the preview build.

### Notes for release
n/a

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -140,7 +140,8 @@ const sharedSettings = definePlugin({
       ],
     }),
     visionTool({
-      defaultApiVersion: '2022-08-08',
+      // uncomment to test
+      //defaultApiVersion: '2025-02-05',
     }),
     // eslint-disable-next-line camelcase
     muxInput({mp4_support: 'standard'}),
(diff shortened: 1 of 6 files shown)


pedrobonamin (approved):
Thanks @bjoerge this looks good to me!

pedrobonamin (approved):

PR #8523, opened 2025-02-06
feat(sanity): add Core UI bridge script

juice49:
### Description

This branch adds the bridge script to the HTML output when building Studio with the standard tooling.

### What to review

Addition of the bridge script when running dev server and building HTML for production.

### Testing

Ensure the bridge script is loaded when running the dev server and in a production build.


juice49:
I've tested this using the dev server and the production build, and the bridge script loads correctly in both scenarios. This looks good to me.

PR #8524, opened 2025-02-06
fix(core): removes deprecated `textSearch` strategy

pedrobonamin:
### Description

Fixes an issue in which users using `textSearch` were seeing their studio crash after updating to 3.74.1.
perspective is not a valid parameter in this search strategy

### What to review
Is it correct to default to `groq2024` as part of this change when enableLegacySearch is false?

### Testing
The strategy is not supported anymore, existing tests using this strategy were removed

### Notes for release
fix(core): removes deprecated `textSearch` strategy


packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.test.ts
```diff
@@ -169,158 +148,6 @@ describe('searchReducer', () => {
...
-          "hit": {
-            "_id": "personC",
-            "_type": "person",
-          },
-        },
-      ]
-    `)
-  })
-
-  it('should not merge results after fetching an additional page when not using `textSearch` strategy', () => {
```
jordanl17, line 246: Q: Is this test not still relevant, it's testing when not using `textSearch`?

pedrobonamin: Good catch! You are right, let me bring it back.

jordanl17 (approved):
Would like the resident search expert @juice49 to take a look too, but all the removals make sense

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.test.ts
juice49: I've added a commit removing this test and the underlying logic. Now that the underlying logic is removed, we _never_ deduplicate the results, and therefore the test is redundant IMO.

juice49 (approved):
Looks good to me. Thanks so much, Pedro! I've added one commit that removes result deduplication—and related tests—from the global search altogether. Now the Text Search API strategy has been removed, it is never necessary to deduplicate the results.

jordanl17 (approved):

pedrobonamin:
Looks like it's ready to go, will add a note to merge next Monday if everything looks fine in clients land

PR #8536, opened 2025-02-07
fix: handle `motion()` deprecation warning

stipsan:
### Description

Fixes this warning:
image

### What to review

In `framer-motion` v12 the `motion()` utility is deprecated, and `motion.create()` should be used instead. Maybe we should add a linter rule to prevent it from sneaking back in?

### Testing

`pnpm dev`, as the warning isn't shown in production only on `sanity dev` or `next dev`.

### Notes for release

Annoying, and unactionable, `motion() is deprecated. Use motion.create() instead.` warnings are handled and should no longer spam your developer console ❤


bjoerge (approved):
Thanks @stipsan !

PR #8570, opened 2025-02-10
feat(cli): server side schemas

EsaNuurtamo:
## Description

Server side schemas makes it easy to fetch schemas in any backend tool. Server side schemas are currently behind a feature flag so you will need to set the following env if you want to test these:
`export SANITY_CLI_SCHEMA_STORE_ENABLED=true`

### Deploy stores workspace schemas to content lake
`sanity deploy`
 #### Options
- `--schema-required `If storing schemas fail deploy fails
- `--verbose` Enable verbose logging: logs info of each schema. Useful if you need the id's of the schemas

#### Examples

`sanity deploy --verbose`

image

---

### Storing schemas without deploying

`sanity schema store` 
#### Options
  - `--workspace` The name of the workspace. The schema of the workspace will be stored in content lake 
  - `--manifest-dir` If you are not using the default static file path, you can specify it here.
  - `--id-prefix` you can specify a custom id prefix for the stored schemas
  -  `--verbose` Enable verbose logging

#### Example
image
- Storing one workspace schema from custom folder using custom id
- Custom id is good if you want test schema changes without overriding the original schema
    
---

### Listing schemas
`sanity schema list`
- lists schemas in json or in console in table format

#### Options
  - `--id` To list a single workspace or single schma json use this 
  - `--json` If you want json formatted full schemas use this
  - `--manifest-dir`  If you are not using the default static file path, you can specify it here.
 #### Example
 Listing all schemas is a good way to check what is saved in the dataset:
image
You can get the whole schema in json:
image

### Deleting schemas
`sanity schema delete --ids schema_id1,schema_id2`
You can delete any number of schemas by giving a comma seperated list as argument.
#### Options
  --ids <schema_id_1,schema_id_2,...> comma-separated list of schema IDs to delete
  --path <path> path to the manifest file if it is not in the default location

#### Examples
image

## Problems 
- There can only be one schema in Content lake for each workspace. If someone has same exact create-manifest.json in multiple studios (same dataset, projectId, workspace name). In this case when other studio is deployed the schemas from other studio are overwritten in the content lake
- You can only deploy schemas that have the same project id as your cli config

#### Silent error
image

#### Stop deployment on error
image
- With the flag you can stop deployment if there is errors in extracting or storing schemas

#### Schema does not exist
image

## What to review
- Install tagged release to your project: `npm install sanity@server-side-schemas`
- Try the above commands, here are more details: [notion doc](https://www.notion.so/sanityio/CLI-server-side-schemas-19e78d055b9480088329eb10f6149400?showMoveTo=true&saveParent=true)

### Testing
- created unit tests using the mockClient
- tested commands manually on different studios


packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,138 @@
...
+      requireUser: true,
+      requireProject: true,
+    }).withConfig({apiVersion: 'v2024-08-01'})
+
+    const projectId = client.config().projectId
+
+    let manifest: CreateManifest
+
+    try {
+      manifest = JSON.parse(readFileSync(`${manifestPath}/create-manifest.json`, 'utf-8'))
```
snorrees, line 52: There is a `MANIFEST_FILENAME` in the codebase, you should move that to a constants file and reuse it everywhere.

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,138 @@
...
+      output.error(error)
+      throw error
+    }
+
+    let storedCount = 0
+
+    let error: Error | undefined
+
+    const saveSchema = async (workspace: ManifestWorkspaceFile) => {
+      const id = `${idPrefix || 'sanity.workspace.schema'}.${workspace.name}`
```
snorrees, line 64: Extract `'sanity.workspace.schema'` as constant, its used elsewhere as well

packages/sanity/src/_internal/cli/actions/schema/fetchSchemaAction.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const projectId = client.config().projectId
+  const dataset = client.config().dataset
+
+  const schemas = await client
+    .withConfig({
+      dataset: dataset,
+      projectId: projectId,
+    })
+    .fetch<SanityDocumentLike[]>(`*[_type == "sanity.workspace.schema" && _id == "${schemaId}"]`)
```
snorrees, line 29: The type should be a constant an reused troughout. You should also prefer using GROQ params in queries for variables, ie `fetch(*[_id==$schemaId], { schemaId})`

packages/sanity/src/_internal/cli/commands/build/buildCommand.ts
```diff
@@ -6,7 +6,9 @@ Options
   --source-maps Enable source maps for built bundles (increases size of bundle)
   --auto-updates / --no-auto-updates Enable/disable auto updates of studio versions
   --no-minify Skip minifying built JavaScript (speeds up build, increases size of bundle)
+  --work
```
snorrees, line 9: What is this?

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // If a workspace name is provided, only save the schema for that workspace
+    if (workspaceName) {
+      const workspaceToSave = manifest.workspaces.find(
+        (workspace) => workspace.name === workspaceName,
+      )
+      if (workspaceToSave) {
+        await saveSchema(workspaceToSave)
+      } else {
```
snorrees, line 115: if you do !workspaceToSave and put it first, you could skip the if-wrapping around the success-case

packages/sanity/src/_internal/cli/actions/schema/__tests__/storeManifestSchemasAction.test.ts
```diff
@@ -0,0 +1,171 @@
...
+        withConfig: vi.fn().mockReturnThis(),
+        transaction: vi.fn().mockReturnThis(),
+        createOrReplace: vi.fn().mockReturnThis(),
+        commit: vi.fn().mockResolvedValue({}),
+      })),
+    } as unknown as CliCommandContext
+  })
+
+  it('should store all schemas when no flags are provided', async () => {
+    const args: CliCommandArguments<any> = {
```
snorrees, line 80: I think the tests would be bit more succinct if you made a helper function to just pass the non-common argumens.

Ie `const args = getStoreCommandArgs({extOptions: {workspace: 'testWorkspace'}})`

At first I wonder if all these empty arrays where important for the test-case, but i see its duplicated throughout.

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,138 @@
...
+    }).withConfig({apiVersion: 'v2024-08-01'})
+
+    const projectId = client.config().projectId
+
+    let manifest: CreateManifest
+
+    try {
+      manifest = JSON.parse(readFileSync(`${manifestPath}/create-manifest.json`, 'utf-8'))
+    } catch (error) {
+      spinner.fail(`Manifest not found at ${manifestPath}/create-manifest.json`)
```
snorrees, line 54: Like here

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+  dataset: string
+  projectId: string
+}
+
+export const SANITY_WORKSPACE_SCHEMA_ID = 'sanity.workspace.schema'
+
+const printSchemaList = ({schemas, output, dataset, projectId}: PrintSchemaListArgs) => {
+  const ordered = sortBy(
+    schemas.map(({_createdAt: createdAt, _id: id, workspace}) => {
+      return [id, workspace.title, dataset, projectId, createdAt].map(String)
```
snorrees, line 23: I still maintain that it should be "Workspace name" and not title. workspace.title is not validated to be unique in sanity.config, so you could have duplicates here.

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+    headings.map((str) => size(str)),
+  )
+
+  const printRow = (row: string[]) => row.map((col, i) => `${col}`.padEnd(maxWidths[i])).join('   ')
+
+  output.print(chalk.cyan(printRow(headings)))
+  rows.forEach((row) => output.print(printRow(row)))
+}
+
+export default async function storeSchemaAction(
```
snorrees, line 41: Export default because lazyloaded /established pattern for actions in the CLI right?

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,136 @@
...
+  'schema-required'?: boolean
+  'verbose'?: boolean
+}
+
+export default async function storeSchemasAction(
+  args: CliCommandArguments<StoreManifestSchemasFlags>,
+  context: CliCommandContext,
+): Promise<Error | undefined> {
+  const flags = args.extOptions
+  if (typeof flags.path === 'boolean') throw new Error('Path is empty')
```
snorrees, line 28: Shouldnt these checks be `!flags.path || typeof flags.path !== 'string'`. I'd say that captures intent better.

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (typeof flags['id-prefix'] === 'boolean') throw new Error('Id prefix is empty')
+  if (typeof flags.workspace === 'boolean') throw new Error('Workspace is empty')
+
+  const schemaRequired = flags['schema-required']
+  const workspaceName = flags.workspace
+  const idPrefix = flags['id-prefix']
```
snorrees, line 34: Could probably move these up over the checks, and use the variables for validation instead? Then you coul also spread them out of flags, and the validations would type-narrow in the below code.

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    let storedCount = 0
+
+    let error: Error | undefined
+
+    const saveSchema = async (workspace: ManifestWorkspaceFile) => {
+      const id = `${idPrefix ? `${idPrefix}.` : ''}${SANITY_WORKSPACE_SCHEMA_ID}.${workspace.name}`
+      try {
+        if (workspace.projectId !== projectId) {
```
snorrees, line 73: I think we should leave a comment here saying something to the effect of "schema store does not support multi-project workspaces for now. We will add support for passing multi-project tokens in the future, and remove this check then. default cli tokens are project bound and will not be able to store for multi projects)

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+  let schemas: SanityDocument[]
+
+  if (flags.id) {
+    // Fetch a specific schema by id
+    schemas = await client
+      .withConfig({
+        dataset: dataset,
+        projectId: projectId,
+      })
+      .fetch<SanityDocument[]>(`*[_type == $type && _id == $id]`, {
```
snorrees, line 70: Since we have the id here, we should use `getDocument` instead. Then we are 100% sure its the absolutely latest value.

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+export default async function storeSchemaAction(
+  args: CliCommandArguments<SchemaListFlags>,
+  context: CliCommandContext,
+): Promise<void> {
+  const flags = args.extOptions
+  if (typeof flags.id === 'boolean') throw new Error('Id is empty')
+  const {apiClient, output} = context
+  const client = apiClient({
+    requireUser: true,
+    requireProject: true,
```
snorrees, line 50: Im unsure how the API client function sets things up, but can we pass useCdn: false here, just to be sure its not using cdn?

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+        type: SANITY_WORKSPACE_SCHEMA_ID,
+      })
+  } else {
+    // Fetch all schemas
+    schemas = await client
+      .withConfig({
+        dataset: dataset,
+        projectId: projectId,
+      })
+      .fetch<SanityDocument[]>(`*[_type == $type]`, {
```
snorrees, line 81: The only way I can think of that this returns a wrong list, if if there is some sort of wierd caching going on. But then it should fix itself after a while, but in the case where I saw a list without all stored entries, it has not 🤔

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,100 @@
...
+  }).withConfig({apiVersion: 'v2024-08-01'})
+
+  const projectId = client.config().projectId
+  const dataset = client.config().dataset
+
+  if (!projectId || !dataset) {
+    output.error('Project ID and Dataset must be defined.')
+    return
+  }
+
```
snorrees, line 60: You are only reading from the configClient dataset.

This code needs to loop over all the datasets in the manifest and fetch form each and combine.

snorrees (changes requested):

packages/sanity/src/_internal/cli/actions/deploy/deployAction.ts
```diff
@@ -128,14 +133,35 @@ export default async function deployStudioAction(
...
       )
+
+      if (flags['schema-required'] && extractManifestError) {
+        output.error(`Schema extraction error: ${extractManifestError.message}`)
+        throw extractManifestError
+      }
+    }
+  }
+
+  if (!isCoreApp) {
```
snorrees, line 152: Did we figure out when isCoreApp is true and why we dont want manifest then?

packages/sanity/src/_internal/cli/actions/schema/deleteSchemaAction.ts
```diff
@@ -0,0 +1,106 @@
...
+  args: CliCommandArguments<DeleteSchemaFlags>,
+  context: CliCommandContext,
+): Promise<void> {
+  if (!SCHEMA_STORE_ENABLED) {
+    return
+  }
+  const flags = args.extOptions
+  if (typeof flags.dataset === 'boolean') throw new Error('Dataset is empty')
+  if (typeof flags['manifest-dir'] === 'boolean') throw new Error('Manifest directory is empty')
+  if (typeof flags.ids !== 'string') throw new Error('--ids is required')
```
snorrees, line 28: Why do we assert booleans in some cases and not string in others? At least pick one style (id prefer !== string)

You could add a tiny assertString(parameterName, errorMessage) helper function and use for all of them imo.

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,141 @@
...
+  args: CliCommandArguments<SchemaListFlags>,
+  context: CliCommandContext,
+): Promise<void> {
+  if (!SCHEMA_STORE_ENABLED) {
+    return
+  }
+
+  const flags = args.extOptions
+  if (typeof flags.id === 'boolean') throw new Error('Schema ID is empty')
+  if (typeof flags['manifest-dir'] === 'boolean') throw new Error('Manifest directory is empty')
```
snorrees, line 64: Could reuse the assert functions here imo

packages/sanity/src/_internal/cli/actions/schema/deleteSchemaAction.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+        try {
+          throwIfProjectIdMismatch(workspace, projectId)
+          const deletedSchema = await client
+            .withConfig({
+              dataset: flags.dataset || workspace.dataset,
+              projectId: workspace.projectId,
+              useCdn: false,
+            })
+            .delete(schemaId.trim())
```
snorrees, line 69: Maybe trim earlier, since you are deleting a possibly different string than what you are logging?

packages/sanity/src/_internal/cli/actions/schema/schemaListAction.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (!projectId || !dataset) {
+    output.error('Project ID and dataset must be defined.')
+    return
+  }
+
+  const manifestDir = flags['manifest-dir']
+  const manifestPath = getManifestPath(context, manifestDir)
+  const manifest = await readManifest(manifestPath, context)
+
+  // Gather all schemas
```
snorrees, line 84: Nit: Noisy comment (delete imo)

Edit: I see there are a lot of this, and Its just your style. I'd rather we didnt litter the code with comments that repeats what the code is doing, especially when its clear on the very next line (maybe not here).

If this variable was named schemaResults you dont need the comment. Ect.

I'll leave it up to you, but I cant remember seeing this style often in the sanity code-base elsewhere; imo prefer comments for "whys or really complicated, im sorry" stuff 😅 

If you want to add meatier TS docs comments at the top of fuctions though, im all for it.

snorrees (dismissed):
Had some minor nit comments, but nothing blocking!

For other internal reviewers: [Full test report here, if yall wanna see example outputs and boundary conditions](https://www.notion.so/sanityio/Test-cli-log-1a578d055b9480eca91fec971524ce6f)

bjoerge (changes requested):

snorrees:
I'm going to re-open a new version of this PR, closing this one.

PR #8574, opened 2025-02-10
test(cli): add retry for cli tests

bjoerge:
### Description
CLI tests are sometimes flaky. This adds retrying to cli tests similar to what we did for e2e tests in #8501, and unit tests in #8371

### What to review
Makes sense?

### Notes for release
n/a – internal

### DIFF
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Test
         id: test
-        run: pnpm test:vitest --silent --project=@sanity/cli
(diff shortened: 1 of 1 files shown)


jordanl17 (approved):
Neat - prefer to add the retry on all tests over adding to select flakey tests? Fine with this though

bjoerge:
> prefer to add the retry on all tests over adding to select flakey tests?

Good point. The drawback with retrying all tests is that legit test failures will be retried 4 times for no good reason. I don't *think* that will be a problem in practice though, and also not sure if it's that easy to identify which exact ones are flaky either?

This is a super low-effort quickfix, so I'll merge for now, but lets reconsider if we find issues with it.

PR #8576, opened 2025-02-10
feat: updating document count for release overview cells

jordanl17:
### Description
In releases Overview the documents row has:
* 'Documents' header
* Just show the count in the cells

| Before | After |
|--------|--------|
| Screenshot 2025-02-10 at 17 48 43 | Screenshot 2025-02-10 at 17 49 47 | 

### Testing
Updated the existing tests
Added a few tests for the document counts

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
@@ -1,25 +1,11 @@
 import {Text} from '@sanity/ui'
-import {useMemo} from 'react'
-
-import {Translate, useTranslation} from '../../../../i18n'
-import {releasesLocaleNamespace} from '../../../i18n'
 
 type Props = {
   documentCount: number | undefined
 }
 
-export const ReleaseDocumentsCounter = ({documentCount}: Props) => {
-  const {t} = useTranslation(releasesLocaleNamespace)
-
-  const count = useMemo(() => {
-    if (!documentCount) return '-'
-
-    return <Translate t={t} i18nKey="document-count" values={{count: documentCount}} />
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
```diff
@@ -1,25 +1,11 @@
...
-
-  return (
-    <Text muted size={1}>
-      {count}
-    </Text>
-  )
-}
+export const ReleaseDocumentsCounter = ({documentCount}: Props) => (
+  <Text muted size={1}>
+    {documentCount || '-'}
```
juice49, line 9: Tiny thought: this will display "-" if the `documentCount` is `0` (rather than only if it's `undefined`).

Maybe that's what you intended? My gut instinct would be to display `0` via `documentCount ?? '-'`.

Mostly just flagging in case it's unintentional 😅.

juice49 (commented):
Looks good! Let one question.

packages/sanity/src/core/releases/tool/overview/columnCells/ReleaseDocumentsCounter.tsx
jordanl17: Ah yeh, this is what I was going for. It's how it currently works- no documents (whether 0 length or just undefined) map to '-' so figured I'd keep it the same as the only criteria in the task was around adding the title and removing 'documents' from each count. 

Hope it sounds ok to you?

juice49: Yep, no strong opinions here, just trying to be a human linter.

juice49 (approved):

PR #8586, opened 2025-02-11
fix(sanity): use raw as default perspective

bjoerge:
(meta: wrong ticket id, should've been sapp-2049)

### Description
This explicitly uses `raw` as perspective unless overridden (e.g. by perspective stack)

This change should be safe, given that raw is currently the default on existing API versions.

### What to review
Does the change make sense? As of this PR, Studio should longer do `/query` requests without a perspective param.

### Testing
Everything should work as before

### Notes for release
n/a - not user facing

### DIFF
--- a/packages/sanity/src/core/hooks/useClient.ts
+++ b/packages/sanity/src/core/hooks/useClient.ts
@@ -26,9 +26,9 @@ export function useClient(clientOptions?: SourceClientOptions): SanityClient {
   const source = useSource()
   if (!clientOptions) {
     console.warn(
-      'Calling `useClient()` without specifying an API version is deprecated and will stop working in the next dev-preview release - please migrate to use `useClient({apiVersion: "2021-06-07"})`.',
+      'Calling `useClient()` without specifying an API version is deprecated and will stop working in the next major version - please specify a date, e.g. `useClient({apiVersion: "2025-02-10"})`.',
     )
-    return source.getClient({apiVersion: '2021-06-07'})
+    return source.getClient({apiVersion: 'v2025-02-07'})
   }
 
   return source.getClient(clientOptions)
--- a/packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
+++ b/packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
@@ -161,6 +161,7 @@ export function _createAuthStore({
         useCdn: false,
         ...(token && {token}),
         withCredentials: true,
+        perspective: 'raw',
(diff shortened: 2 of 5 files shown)


packages/sanity/src/presentation/useDocumentLocations.ts
```diff
@@ -54,7 +54,7 @@ function listen(id: string, fields: string[], store: DocumentStore) {
   }
   const params = {id, draftId: getDraftId(id)}
   return store.listenQuery(query, params, {
-    perspective: 'previewDrafts',
+    perspective: 'drafts',
```
bjoerge, line 57: not strictly related, but took the liberty of renaming this when I came across it.

packages/sanity/src/core/hooks/useClient.ts
```diff
@@ -26,9 +26,9 @@ export function useClient(clientOptions?: SourceClientOptions): SanityClient {
   const source = useSource()
   if (!clientOptions) {
     console.warn(
-      'Calling `useClient()` without specifying an API version is deprecated and will stop working in the next dev-preview release - please migrate to use `useClient({apiVersion: "2021-06-07"})`.',
+      'Calling `useClient()` without specifying an API version is deprecated and will stop working in the next major version - please specify a date, e.g. `useClient({apiVersion: "2025-02-10"})`.',
     )
-    return source.getClient({apiVersion: '2021-06-07'})
+    return source.getClient({apiVersion: 'v2025-02-07'})
```
bjoerge, line 31: We soon want to bump the version used by all internal code, so I think this change is sensible, esp. considering we've had the above warning throughout the lifetime of v3

packages/sanity/src/core/store/_legacy/authStore/createAuthStore.ts
```diff
@@ -161,6 +161,7 @@ export function _createAuthStore({
         useCdn: false,
         ...(token && {token}),
         withCredentials: true,
+        perspective: 'raw',
```
bjoerge, line 164: the client created here is the basis of the client used throughout the Studio, and its config is inherited by client instances returned by `useClient()` etc.

pedrobonamin: 💡

packages/sanity/src/core/hooks/useClient.ts
bjoerge: Yes, will make a separate PR for it – would prefer a change like that to be subjected to a bit more testing :)

pedrobonamin (approved):

PR #8588, opened 2025-02-11
fix(core): update releases permissions for main release action

RitaDias:
### Description

- Create `useReleasePermissions` which has a guardrail that checks if you can perform actions
- Add a permissions banner in the release permissions to warn a user
- Disable publish / schedule main button in the release

image
image
image

### What to review

Does this make sense? Should we do this a different way?

### Testing

Have updated the tests

### Notes for release

Scheduling and Publishing a release now respects permissions.


jordanl17 (dismissed):
Think it might be good to think on whether there's any possibility of a reusable wrapper for all the `dryRun` actions, but otherwise looks good

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -200,6 +202,53 @@ export function createReleaseOperationsStore(options: {
...
+  }
+
+  const canPublish = async (releaseId: string, useUnstableAction?: boolean) => {
+    try {
+      await requestAction(
+        client,
+        [
+          {
+            actionType: useUnstableAction
+              ? 'sanity.action.release.publish2'
```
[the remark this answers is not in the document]
RitaDias: Good point, will make this change

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -48,7 +51,16 @@ export const ReleaseScheduleButton = ({
...
   const isValidatingDocuments = documents.some(({validation}) => validation.isValidating)
   const hasDocumentValidationErrors = documents.some(({validation}) => validation.hasError)
-  const isScheduleButtonDisabled = disabled || isValidatingDocuments
+  const isScheduleButtonDisabled = disabled || isValidatingDocuments || !schedulePermission
+
+  useEffect(() => {
+    canSchedule(
+      release._id,
+      release.metadata.intendedPublishAt
+        ? new Date(release.metadata.intendedPublishAt)
```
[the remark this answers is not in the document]
RitaDias: Ah you see I wasn't sure _if it did_ make a difference, hence why I just did this to be on the safe side 😅 
I can of course make that change

packages/sanity/src/core/releases/store/createReleasePermissionsStore.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  const canPublish = async (releaseId: string, useUnstableAction?: boolean) => {
+    try {
+      await requestAction(
+        client,
+        [
+          {
+            actionType: useUnstableAction
+              ? 'sanity.action.release.publish2'
```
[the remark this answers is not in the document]
RitaDias: Good point, will make this change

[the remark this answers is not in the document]
RitaDias: The only thing here is that I'm following the exact same logic as what we do in the publish in the [createReleaseOperationStore](https://github.com/sanity-io/sanity/blob/sapp-2069/packages/sanity/src/core/releases/store/createReleaseOperationStore.ts#L78). So if it's the same to you I'll create a separate task to do this outside of scope of this PR

packages/sanity/src/core/releases/store/createReleasePermissionsStore.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const canSchedule = async (releaseId: string, publishAt: Date) => {
+    try {
+      await requestAction(
+        client,
+        [
+          {
+            actionType: 'sanity.action.release.schedule',
+            releaseId: getReleaseIdFromReleaseDocumentId(releaseId),
+            publishAt: publishAt.toISOString(),
```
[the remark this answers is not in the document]
RitaDias: As I said on the comment above, I wasn't aware it didn't make a difference so I'll make the chance in both places!

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -144,23 +194,31 @@ export function createReleaseOperationsStore(options: {
...
+          [
+            {
+              actionType: 'sanity.action.document.createVersion',
+              releaseId,
+              attributes: versionDocument,
+            },
+          ],
+          opts,
+        )
+      : client.create(versionDocument, opts))
```
RitaDias, line 208: Frankly a bit unsure about this, but it worked (when not using the dry run) so this should be fine

packages/sanity/src/core/releases/store/createReleasePermissionsStore.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export function createReleasePermissionsStore(options: {
+  client: SanityClient
+  onReleaseLimitReached: (limit: number) => void
+}): useReleasePermissionsValue {
+  const {client} = options
+  const requestAction = createRequestAction(options.onReleaseLimitReached)
+
+  const canSchedule = async (releaseId: string, publishAt: Date) => {
```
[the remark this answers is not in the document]
RitaDias: Here we go https://github.com/sanity-io/sanity/pull/8588/commits/e6429e6bb6207a2595417301fc6bee3cbaa57b8b# :)

packages/sanity/src/core/releases/store/useReleasePermissions.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function useReleasePermissions(): useReleasePermissionsValue {
+  /**
+   * Checks if a release action can be performed by running a dry run of the given action
+   *
+   * @param action - any of the actions from the ReleaseOperationStore, e.g. publishRelease should send in also the needed props
+   * @param args - the arguments to pass to the action (release id, etc)
+   * @returns true or false depending if the user can perform the action
+   */
+  const checkWithPermissionGuard = async (
```
jordanl17, line 19: A suggestion here to tighten the typing here:

```
export interface useReleasePermissionsValue {
  checkWithPermissionGuard: <T extends (...args: any[]) => Promise<void> | void>(
    action: T,
    ...args: Parameters<T>
  ) => Promise<boolean>
}
```

```
  const checkWithPermissionGuard = async <T extends (...args: any[]) => Promise<void> | void>(
    action: T,
    ...args: Parameters<T>
  ): Promise<boolean> => {...
```

This means that the args that are passed to `checkWithPermissionGuard` after the first release operations argument must be the correct types for the release operation

RitaDias: Thanks for the suggestion, let me do that 🙏

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleasePublishAllButton.tsx
```diff
@@ -43,16 +45,31 @@ export const ReleasePublishAllButton = ({
...
 
+  const [publishPermission, setPublishPermission] = useState<boolean>(false)
+
   const isValidatingDocuments = documents.some(({validation}) => validation.isValidating)
   const hasDocumentValidationErrors = documents.some(({validation}) => validation.hasError)
 
-  const isPublishButtonDisabled = disabled || isValidatingDocuments || hasDocumentValidationErrors
+  const isPublishButtonDisabled =
+    disabled || isValidatingDocuments || hasDocumentValidationErrors || !publishPermission
+  const useUnstableAction = publishBundleStatus === 'confirm-2'
```
jordanl17, line 55: I don't know if we need `publish2` anymore, I think we just use publish after content lake deprecated the first version and swapped them. Might be worth checking as could simplify a little bit here

RitaDias: I feel about this the same way that I feel about the sanity.action.release.publish2. I think it should be done outside of the scope just to make sure that it's easier to track if something goes wrong there

RitaDias: I have a story for [this](https://linear.app/sanity/issue/SAPP-2375/use-sanityactionreleasepublish-instead-of-sanityactionreleasepublish2)

packages/sanity/src/core/releases/tool/detail/ReleaseDashboardDetails.tsx
```diff
@@ -28,13 +30,41 @@ import {ReleaseTypePicker} from './ReleaseTypePicker'
...
-  const shouldDisplayError = release.state === 'active' && typeof release.error !== 'undefined'
   const isAtTimeRelease = release?.metadata?.releaseType === 'scheduled'
+  const isActive = release.state === 'active'
+  const shouldDisplayError = isActive && typeof release.error !== 'undefined'
+  const [shouldDisplayPermissionWarning, setShouldDisplayPermissionWarning] = useState(false)
+  const shouldDisplayWarnings = isActive && shouldDisplayPermissionWarning
+  useEffect(() => {
+    // only run if the release is active
+    if (isActive) {
+      checkWithPermissionGuard(publishRelease, release._id, false).then((response) => {
```
jordanl17, line 48: NIT: `response` could be renamed to something like `hasPermission`

jordanl17:
This looks really great to me @RitaDias - nice thinking. Personally I think this does feel simpler to use and makes it easy to throw in other places when we guard more with permissions. I really like it!

RitaDias:
> This looks really great to me @RitaDias - nice thinking. Personally I think this does feel simpler to use and makes it easy to throw in other places when we guard more with permissions. I really like it!

Agreed, your suggestion did simplify a follow up PR I have coming in soon:tm: from double digit file changes to just 4 which is great :)

packages/sanity/src/core/releases/tool/detail/ReleaseDashboardDetails.tsx
RitaDias: Done!

jordanl17 (approved):
Great stuff!!

PR #8594, opened 2025-02-11
fix(core): hide schedule publishing if not used and not enabled by users

pedrobonamin:
### Description

With the introduction of releases we want to start moving users into it and out of `scheduled publishing` , to do this we will make scheduled publishing opt in.
Scenarios considered:
- Workspaces that have previously used scheduled publishing.
- Workspaces that have not used scheduled publishing.
- Workspaces that have enabled scheduled publishing in the config (this is not required[ since this was moved into core](https://github.com/sanity-io/sanity/issues/6416))

#### How to detect if it was used

We rely on the information provided by the schedules endpoint `/schedules/<projectId>/<dataset>`. If this endpoint returns at least 1 schedule it means that the workspace has previously used it.

#### How do we handle each situation

**Workspaces that have previously used scheduled publishing.**
If they have previously used it we will continue showing the document action and the tool in the navbar, by setting the `ScheduledPublishingEnabled` context value to `{enabled:true}` [see code](https://github.com/sanity-io/sanity/blob/sapp-2161/packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx#L93-L96)

**Workspaces that have not used scheduled publishing.**
For this workspaces the value in the context will be `{enabled: false}` this will hide the tool from the navbar and will also hide the action. [See code](https://github.com/sanity-io/sanity/blob/sapp-2161/packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx#L86-L92)

**Workspaces that  explicitly opt in**
Until today users didn't need to opt in to **scheduled publishing** , from today they will need to do it if they want to start using the tool and they haven't use it before (see previous point), when they opt in, the `__internal__workspaceEnabled` flag will be set to `true` this allows us to continue showing the tool as in the first point. [see code](https://github.com/sanity-io/sanity/blob/sapp-2161/packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx#L93-L97)

### UI Changes

**Schedule publishing is disabled**

The `schedules` navbar button will be hidden by [this change](https://github.com/sanity-io/sanity/blob/sapp-2161/packages/sanity/src/core/studio/components/navbar/tools/StudioToolMenu.tsx#L21-L23) and the schedule action is also hidden.

If the user lands in the `/schedules/` tool they will see a warning to understand why this change is happening.
Screenshot 2025-02-12 at 10 14 10

**Schedule is enabled** 

- Releases is enabled:
  In this case we show them a warning with a recommendation on using **Releases*
  Screenshot 2025-02-12 at 09 58 20

 This can be removed by setting the following config:
 ```diff
defineConfig({
  ...,
  schedulePublishing: {
     enabled: true,
+   showReleasesBanner: false
})
```

- Releases is not enabled:
  Nothing changes for them.

### Testing
This has to be tested locally by updating the config.
- Playground dataset:
   - Scheduled publishing is hidden by default (this dataset has not used it)
   - Enable it in the config, it should show the tool and the releases warning `scheduledPublishing.enabled=true`
   - Disable releases, it should remove the warning.

- Test dataset
   - Scheduled publishing is enabled by default (this dataset has used it), it should show the warning
   - Enable it in the config, nothing should change from the above
   - Disable releases, the warning is removed.
   - Explicitly disable it in the config  `scheduledPublishing.enabled=false`, schedule is hidden, even knowing they have used it.
   
   
Existing tests have been updated.

### Notes for release
**Scheduled Publishing is now opt-in**. Workspaces that have used it before will retain access, while new users must enable it explicitly.

### DIFF
--- a/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
@@ -97,28 +97,7 @@ export default function Tool() {
     return (
       <Container width={1} paddingTop={4}>
         <Box paddingTop={4} paddingX={4}>
-          {hasUsedScheduledPublishing.loading ? (
-            <LoadingBlock />
-          ) : (
-            <InfoCallout
-              // TODO: TBC with design and growth
-              description={
-                <>
-                  Scheduled Publishing is not enabled. It can be enabled in the config. We recommend
-                  using{' '}
-                  <a
-                    target="_blank"
-                    href="https://www.sanity.io/content-releases?ref=release"
-                    rel="noreferrer"
-                  >
-                    "Releases"
-                  </a>
-                  .
-                </>
-              }
-              title="Scheduled Publishing is not enabled"
-            />
-          )}
+          {hasUsedScheduledPublishing.loading ? <LoadingBlock /> : <InfoCallout />}
         </Box>
       </Container>
     )
--- a/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
@@ -10,6 +10,7 @@ import ErrorCallout from '../components/errorCallout/ErrorCallout'
 import InfoCallout from '../components/infoCallout/InfoCallout'
 import ButtonTimeZone from '../components/timeZoneButton/TimeZoneButton'
 import ButtonTimeZoneElementQuery from '../components/timeZoneButton/TimeZoneButtonElementQuery'
+import {WarningBanner} from '../components/warningBanner/WarningBanner'
 import {SCHEDULE_FILTERS, TOOL_HEADER_HEIGHT} from '../constants'
 import usePollSchedules from '../hooks/usePollSchedules'
 import useTimeZone from '../hooks/useTimeZone'
@@ -33,7 +34,8 @@ const DATE_SLUG_FORMAT = 'yyyy-MM-dd' // date-fns format
 
 export default function Tool() {
   const router = useRouter()
-  const {scheduledPublishing} = useWorkspace()
+  const {scheduledPublishing, releases} = useWorkspace()
+  const isReleasesEnabled = Boolean(releases?.enabled)
 
   const {sanity: theme} = useTheme()
   const {error, isInitialLoading, schedules = NO_SCHEDULE} = usePollSchedules()
@@ -105,6 +107,7 @@ export default function Tool() {
 
   return (
     <SchedulesProvider value={schedulesContext}>
+      {isReleasesEnabled && scheduledPublishing.showReleasesBanner && <WarningBanner />}
       <Flex direction="column" height="fill" flex={1} overflow="hidden">
         <Flex flex={1} height="fill">
           {/* LHS Column */}
--- a/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
@@ -4,7 +4,10 @@ import {useEffect, useMemo, useRef} from 'react'
 import {type RouterContextValue, useRouter} from 'sanity/router'
 import {styled} from 'styled-components'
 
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx [hunk 1]
```diff
@@ -77,20 +83,31 @@ export default function Tool() {
...
+    }
+    // This is for the case users lands in the tool rout without having the feature enabled.
     return (
       <Container width={1} paddingTop={4}>
         <Box paddingTop={4} paddingX={4}>
-          <ErrorCallout
-            description="Something went wrong loading permissions, please try again."
-            title="Permissions check failed"
-          />
+          {hasUsedScheduledPublishing.loading ? <LoadingBlock /> : <InfoCallout />}
```
jordanl17, line 102: Q: Is it necessary to use `hasUsedScheduledPublishing` here given that the condition is only on `loading` state. Could we just always render `<InfoCallout />` in this condition?

packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx
```diff
@@ -11,38 +22,81 @@ export type ScheduledPublishingEnabledContextValue =
...
+        enabled: true,
+        mode: enabled ? 'default' : 'upsell',
+        hasUsedScheduledPublishing,
+      }
+    }
+    if (!hasUsedScheduledPublishing.used) {
+      return {
+        enabled: false,
+        mode: null,
+        hasUsedScheduledPublishing,
```
jordanl17, line 91: Q: Does `hasUsedScheduledPublishing` need to be exposed here, it's value is already being included in the setting of `enabled`

packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx
```diff
@@ -11,38 +22,81 @@ export type ScheduledPublishingEnabledContextValue =
...
+}: ScheduledPublishingEnabledProviderProps) {
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
+
   const {enabled, isLoading, error} = useFeatureEnabled('scheduledPublishing')
   const {scheduledPublishing} = useWorkspace()
 
   const isWorkspaceEnabled = scheduledPublishing.enabled
+  const explicitEnabled = scheduledPublishing.__internal__workspaceEnabled
+
+  const hasUsedScheduledPublishing$: Observable<HasUsedScheduledPublishing> = useMemo(() => {
```
jordanl17, line 52: Nice one, smooth solution

packages/sanity/src/core/scheduledPublishing/tool/Tool.tsx
pedrobonamin: It gets in a short racing condition if we don't check the loading here, in where you could see the infoCallout banner and then once the `useHasScheduledPublishing` resolves it shows you the tool as expected.
Because it will initially say that the tool is not enabled, but once the check passes it will be enabled.

That's why the context is exposing this value

packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.tsx
pedrobonamin: Same as https://github.com/sanity-io/sanity/pull/8594/files#r1954380517

jordanl17 (approved):

pedrobonamin:
Thanks for the review @jordanl17 will wait until the copy on the new banner and warning is confirmed before merging

pedrobonamin:
@jordanl17 we will go ahead merging this with the existing micro copy that could be updated later in the process. But it is important to have this out before enabling **Content releases** for everyone.
I have created a follow up ticket for the micro copy - SAPP-2410

jordanl17 (approved):

Issue #8521, opened 2025-02-06
Make typegen generate idempotent


larixk:
**Is your feature request related to a problem? Please describe.**
When repeatedly running `sanity typegen generate` on the same codebase, the generated file differs between runs. Types are ordered seemingly randomly within the file, resulting in unnecessary and often large diffs in version control.

**Describe the solution you'd like**
Ensure the generated file uses a consistent, controlled order so that multiple runs produce identical output.

**Describe alternatives you've considered**
A post-processing step to reorder the file was considered, but it adds unnecessary complexity and could easily break on future updates.

**Additional context**
While keeping generated files in version control may not be ideal, our Next.js TypeScript build process requires these types to be available at build time, and generating them as part of the build process could introduce other issues.

mtilda:
I am not able to reproduce the behavior you described. This command seems to me to be idempotent. Can you provide a minimal reproduceable example?

The order of the types in my `sanity.config.ts` file seems match the order in my `schema.json`, so the question is "how does `sanity schema extract` order its output?"

I tried changing the order of the imports in `sanity.config.ts` and the order of the `types` array in that file, but these changes did not alter the order of `schema.json` nor `sanity.config.ts`.

```ts
import { defineConfig } from 'sanity';

// Changing the order of these imports has no effect.
import a from 'sanity/schemas/a';
import b from 'sanity/schemas/b';
import c from 'sanity/schemas/c';

export default defineConfig({
    ...
    schema: {
        types: [
            // Changing the order of these items has no effect.
            a,
            b,
            c,
        ],
    },
})
```

mtilda:
> While keeping generated files in version control may not be ideal, our Next.js TypeScript build process requires these types to be available at build time, and generating them as part of the build process could introduce other issues.

For the record, I think it is perfectly acceptable to commit these files to version control. I do this. Without the types file, my code would seem to have broken imports and would be impossible to read without generating that file locally.

mtilda:
I was wrong. Changing the order of the `types` array does appear to change the order of the `schema.json` file and `sanity.types.ts` file.

```ts
import { defineConfig } from 'sanity';

// Changing the order of these imports has no effect.
import a from 'sanity/schemas/a';
import b from 'sanity/schemas/b';
import c from 'sanity/schemas/c';

export default defineConfig({
    ...
    schema: {
        types: [
            // Changing the order of these items effects the order of the schema.json file.
            a,
            b,
            c,
        ],
    },
})
```

[...]

PR #8598, opened 2025-02-11
feat(cli): make typegen idempotent

larixk:
### Description

This branch makes the `sanity typegen generate` command produce a file with its type definitions listed in a consistent order every time. Before this change, the order of entries within the file could change between runs on larger codebases as it was generated by a parallel process using multiple workers. This change should prevent unnecessary diffs introduced when running the command and closes https://github.com/sanity-io/sanity/issues/8521 .

The order is enforced by sorting the entries into the file before writing and by sorting the queries found within a file. Sorting uses the default Javascript array sort method.

### What to review

Changes are limited to the `typegen generate` command of the sanity cli. Review should focus on this.

### Testing

No testing was added, removed or changed. As the behaviour before this change was hard to reproduce consistently, since it is the result of the order in which read operations on the files in the user's repo finish, adding tests specific for the change introduced here is difficult.

Existing tests covering the type generation pass unchanged.

### Notes for release

Closes https://github.com/sanity-io/sanity/issues/8521
Makes the result of running more consistent between runs

### DIFF
--- a/packages/@sanity/cli/src/actions/typegen/generateAction.ts
+++ b/packages/@sanity/cli/src/actions/typegen/generateAction.ts
@@ -95,6 +95,8 @@ export default async function typegenGenerateAction(
     size: 0,
   }
 
+  const results: string[] = []
+
   await new Promise<void>((resolve, reject) => {
     worker.addListener('message', (msg: TypegenGenerateTypesWorkerMessage) => {
       if (msg.type === 'error') {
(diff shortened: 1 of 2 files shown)


sgulseth:
Hi! Thanks for contributing, and agree that this is an annoyance. Instead of keeping track of all the generated code, and then sort it, I propose a different approach [here](https://github.com/sanity-io/sanity/pull/8610) where we instead sort the input files.

larixk:
> Hi! Thanks for contributing, and agree that this is an annoyance. Instead of keeping track of all the generated code, and then sort it, I propose a different approach [here](https://github.com/sanity-io/sanity/pull/8610) where we instead sort the input files.

@sgulseth Awesome. I agree your approach is cleaner. I was working backwards from the end result, but fixing it from the start probably is the more robust option. 

Closing this one.

sgulseth:
Great to hear. It should be part of the next release, most likely next week. LMK if it works better for you in the issue you created, and we can re-open this one if it still persists

PR #8603, opened 2025-02-12
fix(core): add permissions limitation to document actions in releases

RitaDias:
### Description

- Add disable action when user doesn't have permission to discard version in document action and in the release detail table
- Add disable action when user doesn't have permission to unpublish document in the release detail table
- Update document action to new action
- Permissions there now use the versions
- Updated tooltip text for different conditions on the unpublish action

https://github.com/user-attachments/assets/326ff608-75b1-4816-ba88-274c15f1122a

### What to review

Does everything look alright?

### Testing

Didn't add tests to the menu actions since the larger changes happened directly in the permission aspect of things

### Notes for release

Document actions (discard version and unpublish) in a release now respect permissions


bjoerge (approved):
Great work, thank you @RitaDias!

PR #8605, opened 2025-02-12
fix(structure): don't modify EMPTY_PARAMS object

mariusGundersen:
This fixes a bug where the EMPTY_PARAMS object is modified

### Notes for release

Fixes a bug that could cause initial value template parameters to carry over from an earlier document

### DIFF
--- a/packages/sanity/src/structure/getIntentState.ts
+++ b/packages/sanity/src/structure/getIntentState.ts
@@ -26,8 +26,6 @@ export function getIntentState(
   const panes = routerState?.panes || []
   const activePanes = state.activePanes || []
   const editDocumentId = params.id || uuid()
-  const isTemplate = intent === 'create' && params.template
-  const isVersion = intent === 'create' && params.version
 
   // Loop through open panes and see if any of them can handle the intent
   for (let i = activePanes.length - 1; i >= 0; i--) {
@@ -47,12 +45,7 @@ export function getIntentState(
         pane.schemaTypeName === params.type &&
         pane.options.filter === '_type == $type')
     ) {
-      const paneParams: {
-        template?: string
-        version?: string
-      } = EMPTY_PARAMS
-      if (isTemplate) paneParams.template = params.template
-      if (isVersion) paneParams.version = params.version
+      const paneParams = getPaneParams(intent, params)
 
       return {
         panes: panes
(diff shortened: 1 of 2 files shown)


bjoerge:
Ohh, great catch, thank you @mariusGundersen! What issue(s) did this cause to for you?

mariusGundersen:
If I create something using a template, then any subsequent edit will have the template parameter. This causes the editor to show the wrong form. It loads the form for the template type, not for the type being edited.

packages/sanity/src/structure/getIntentState.ts
```diff
@@ -47,12 +47,11 @@ export function getIntentState(
         pane.schemaTypeName === params.type &&
         pane.options.filter === '_type == $type')
     ) {
+      const { template, version } = params;
       const paneParams: {
```
bjoerge, line 51: Our linter isn't a fan of nested ternaries, and the formatting is slightly off

```suggestion
      const paneParams: {
        template?: string
        version?: string
      } =
        isTemplate || isVersion
          ? {
              ...(isTemplate ? {template: params.template} : {}),
              ...(isVersion ? {version: params.isVersion} : {}),
            }
          : EMPTY_PARAMS
```

mariusGundersen: What's the point of `EMPTY_PARAMS`? I'm guessing it's to reduce the amount of nursery garbage collection, by reusing a fixed object? But this change will create two objects that are immediately garbage collected.

bjoerge: It's mostly about keeping referential identity if/when objects are passed on to the React layer (not entirely sure if that's relevant here though) and less about about garbage collection.

packages/sanity/src/structure/getIntentState.ts
```diff
@@ -64,3 +57,11 @@ export function getIntentState(
 
   return {intent: intent, params, payload}
 }
+
+function getPaneParams(intent: string, {template, version}: Record<string, string>): {template?: string, version?: string} {
```
bjoerge, line 61: This is good – think its just missing a newline at the end of the file (I'm sorry you have to battle our linter so much 😅)

mariusGundersen: I wish it would run in vscode in github 😅

bjoerge:
Phew, had to battle CI for a bit here too, but looking good now. Thank you so much for this @mariusGundersen!

PR #8607, opened 2025-02-12
fix(core): update permissions on release actions

RitaDias:
> [!IMPORTANT]  
> To add an update here: I am currently thinking of a larger way of keeping the update states and not fetching it if they have been fetched in other parts of the interface. However, the plan is to do that changes outside of the scope of this PR

### Description

- Disable archive menu action for release when no permissions
- Disable unarchive menu action for release when no permissions
- Disable delete menu action for release when no permissions

https://github.com/user-attachments/assets/3ba3fd5f-1b4a-4122-a180-3ba998a28687

### What to review

Does it make sense?

### Testing

Have updated the tests

### Notes for release
Archive, Unarchive and Delete action menus for releases now respect permissions.


RitaDias:
> [!IMPORTANT]  
> To add an update here: I am currently thinking of a larger way of keeping the update states and not fetching it if they have been fetched in other parts of the interface. However, the plan is to do that changes outside of the scope of this PR

packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/__tests__/ReleaseMenuButton.test.tsx
```diff
@@ -51,6 +59,10 @@ describe('ReleaseMenuButton', () => {
     vi.clearAllMocks()
 
     mockUseBundleDocuments.mockRestore()
+
+    mockUseReleasePermissions.mockReturnValue({
+      checkWithPermissionGuard: vi.fn().mockResolvedValue(true),
```
pedrobonamin, line 64: COULD: it would be nice to include an extra test with this returning false, and check for the button to be disabled.
Thanks for updating the tests to show that they are enabled.

RitaDias: Added 3 extra ones, as a treat :)

pedrobonamin: Amazing!

pedrobonamin (approved):

PR #8610, opened 2025-02-12
feat(typegen): glob for files synchronously for deterministic result

sgulseth:
### Description

We were using globby.stream, which gives no guarantees on the order of the files returned by the stream. Instead of relying on the async api here we can use the synchronous api and sort the resulting files to get a more deterministic order as the files are changing.

Alternative to #8598 

### What to review

Correctness 

### Testing

I added a test, but I skip it by default as it has to run many iterations to _maybe_ fail

### Notes for release

TypeGen: Sort the scanned files to get a more consistent order of the generated types. Note that first time running this after upgrade you might get a new result because of an earlier mismatch


stipsan (approved):
Awesome!

PR #8612, opened 2025-02-12
feat: reduce redundant success toast messages

juice49:
### Description

This branch removes a handful of success toast messages across the Studio codebase that are arguably redundant because they can largely be expected to succeed immediately and reliably.

### What to review

There is a commit for each affected toast. Do they all seem reasonable? Note: Marius has reviewed and approved the removals.

### Testing

Existing tests pass.

### Notes for release

Some success toast messages have been removed in order to streamline Studio UI. These removals have been made where there is a reasonable expectation the operation will succeed immediately and reliably (for example, when copying a value to the clipboard). Error toast messages remain as they were before.

### DIFF
--- a/packages/@sanity/vision/src/components/VisionGui.tsx
+++ b/packages/@sanity/vision/src/components/VisionGui.tsx
@@ -698,12 +698,6 @@ export class VisionGui extends PureComponent<VisionGuiProps, VisionGuiState> {
     try {
       el.select()
       document.execCommand('copy')
-      this.props.toast.push({
-        closable: true,
-        title: 'Copied to clipboard',
-        status: 'info',
-        id: 'vision-copy',
-      })
     } catch (err) {
       // eslint-disable-next-line no-console
       console.error('Unable to copy to clipboard :(')
--- a/packages/@sanity/vision/src/components/VisionGui.tsx
+++ b/packages/@sanity/vision/src/components/VisionGui.tsx
@@ -522,12 +522,6 @@ export class VisionGui extends PureComponent<VisionGuiProps, VisionGuiState> {
 
   handleListenerEvent(evt: ListenEvent<any>) {
     if (evt.type !== 'mutation') {
-      this.props.toast.push({
-        closable: true,
-        id: 'vision-listen',
-        status: 'success',
-        title: 'Listening for mutations…',
-      })
       return
     }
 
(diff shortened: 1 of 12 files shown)


packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -704,12 +698,6 @@ export class VisionGui extends PureComponent<VisionGuiProps, VisionGuiState> {
     try {
       el.select()
       document.execCommand('copy')
-      this.props.toast.push({
-        closable: true,
-        title: 'Copied to clipboard',
-        status: 'info',
-        id: 'vision-copy',
-      })
```
RitaDias, line 712: I don't know if I agree with this one 🤔 
It's pretty common behaviour (example of linear)

Uploading Screen Recording 2025-02-21 at 10.11.59.mov…

packages/sanity/src/core/comments/plugin/inspector/CommentsInspector.tsx
```diff
@@ -126,22 +126,13 @@ function CommentsInspectorInner(
     if (!getCommentLink) return undefined
 
     const copyLink = (id: string) => {
-      navigator.clipboard
-        .writeText(getCommentLink(id))
-        .then(() => {
-          pushToast({
-            closable: true,
-            status: 'info',
-            title: t('copy-link-success-message'),
```
RitaDias, line 135: Again, I have the same gutfeeling as the other comment

RitaDias (commented):
The comments I have go for more than the places I left them, I'm very hesitant on doing this.
Because of the nature of the studio (where any changes we make are kinda automatically applied) we often rely on a visual indicator that something succeeded (or when it succeeded) or failed. I had it happen on manage where I wasn't sure if something happened until the toast appeared.
Sure, it might not be the best approach, but I don't think that removing them completely is a good idea because it leaves a lot of "certainty" to guess work per user :/

RitaDias:
I get that Marius has approved it and maybe I'm just overthinking this, but I just thought it might be worth bringing up this concern one last time.

RitaDias (approved):
Ok, spoke in private, Marius has approved these changes and while I have some thoughts about this I'm also of the opinion it's harmless and minor when it comes to experience. If we get feedback that something like this should be needed, we can always add as needed. So let's go ahead with this!

PR #8616, opened 2025-02-12
fix(core): create new doc button tooltip updated for when perspective is published and not a release

jordanl17:
### Description
| Before | After |
|--------|--------|
| Screenshot 2025-02-12 at 16 55 44 | Screenshot 2025-02-12 at 16 55 24 | 

When the perspective is published, the tooltip on create new document is more appropriate named

### Testing
N/A

### Notes for release
N/A


RitaDias (approved):
✅

PR #8618, opened 2025-02-12
fix: showing add doc toast once document is received in the subscription

jordanl17:
### Description
The scenario this PR resolves is when 'Add document' is used to create a version document from within a release. When a document is selected, the version document is created, and then the success toast is shown whilst the search dialog is closed.

However, the subscription used in `useBundleDocuments` must receive the new event and update the store state before this new document will appear in the list on release detail. This creates a weird experience where we say success, but there's a delay before everything looks as would be expected.

This PR changes this flow:
* Instead of closing the search dialog and showing toast once `createVersion` resolves, closes the toast immediately after selecting a document
* A pending row (skeleton loading) is added to the document list table
* Once the document is retrieved in the bundle documents store, the pending row is replaced with the actual document
* The toast then shows, once the row is fully visible on the table

addSkeleton
NOTE: this is throttled, it is never this noticable

### Testing
Existing tests validate no regressions. Should add some tests for showing the pending state - but need to put into a fast follow

### Notes for release
N/A


RitaDias:
To add some context here: currently seeing some odd behavior on the release detail view that doesn't exist in the `next` branch (documents in release never load). this is the reason why it has not been reviewed yet (Jordan is aware of this :) )

image

jordanl17:
> To add some context here: currently seeing some odd behavior on the release detail view that doesn't exist in the `next` branch (documents in release never load). this is the reason why it has not been reviewed yet (Jordan is aware of this :) )
> 
> image

Yeh thanks - I'm going to look into this today, have moved this to draft in the meantime. Will update you when it's ready again

PR #8620, opened 2025-02-12
fix: creating new release immediately pins as perspective

jordanl17:
### Description
A PR of 2 halves: 1 simple, 1 more involved:

Simple:
For pinning the new release when created from the perspective menu, or from the 'copy to new release' flow, it's possible to just `setPerspective` to assign the `perspective` sticky param. This is because we don't navigate elsewhere when a release is created in these flows.

More:
In the other flow where a release can be created - from 'create release' CTA in releases overview, we already navigate to the release detail page after creation.

This causes some issues - calling `router.navigateStickyParams` (which is used by `setPerspective`), followed immediately by `router.navigate` to navigate to the release detail, causes a race, where `navigate` knows nothing about the perspective `stickyParam`, so the end result is that the release is not pinned, but you are still navigated to the release details.

Therefore this PR updates the Router (!!) - it deprecates the `navigateStickyParams` and consolidates it's logic into `navigate` so that `navigate` can be used both for navigating (as it already is) but also for setting or unsetting sticky params. Importantly it can also be used for applying both at the same time, thus resolving this issue

### What to review
* Does extending the options on `router.navigate` make sense

### Testing
Manually verified that pinning works in all places where a new release can be created.

**Added tests for `RouterScope` and `RouterProvider` - ran these tests (sans the changes for sticky params) against `next` to verify that no regressions have happened.**

Also added tests for `RouterProvider` given the adjustments there. Ran these tests against the `next` `RouterProvider` (only the relevant ones given no `stickyParam` support for `navigate` in `next`, and verified no regressions to functionality

### Notes for release
`navigate` from `useRouter` now supports a `stickyParams` option to allow for navigating and updating param state in the studio at once.

`navigateStickyParams` is deprecated from `sanity/router` - instead use `navigate` as follows:
```ts
const router = useRouter()

router.navigate({stickyParams: {paramKey: 'paramValue'}, replace: false})
```

Additionally, it's now possible to update route state and set `stickyParams` together:
```ts
const router = useRouter()

router.navigate({stateKey: 'stateValue'}, {stickyParams: {paramKey: 'paramValue'}})
```
Or preferably pass a single argument of options for `navigate`:
```ts
const router = useRouter()

router.navigate({
  stickyParams: {paramKey: 'paramValue'},
  state: {
    stateKey: 'stateValue'
  }
})
```
NOTE: `state` is optional - if omitted then the `stickyParams` will be appended to the existing route; if `state: null` then there will be navigation back to the root path

### DIFF
--- a/packages/sanity/src/core/perspective/useSetPerspective.tsx
+++ b/packages/sanity/src/core/perspective/useSetPerspective.tsx
@@ -10,7 +10,7 @@ export function useSetPerspective() {
   const router = useRouter()
   const setPerspective = useCallback(
     (releaseId: 'published' | 'drafts' | ReleaseId | undefined) => {
-      router.navigate(null, {
+      router.navigate({
         stickyParams: {
           excludedPerspectives: '',
           perspective: releaseId === 'drafts' ? '' : releaseId,
--- a/packages/sanity/src/core/perspective/useSetPerspective.tsx
+++ b/packages/sanity/src/core/perspective/useSetPerspective.tsx
@@ -10,9 +10,11 @@ export function useSetPerspective() {
   const router = useRouter()
   const setPerspective = useCallback(
     (releaseId: 'published' | 'drafts' | ReleaseId | undefined) => {
-      router.navigateStickyParams({
-        excludedPerspectives: '',
-        perspective: releaseId === 'drafts' ? '' : releaseId,
+      router.navigate(null, {
+        stickyParams: {
+          excludedPerspectives: '',
+          perspective: releaseId === 'drafts' ? '' : releaseId,
+        },
       })
     },
     [router],
--- a/packages/sanity/src/core/perspective/useSetPerspective.tsx
+++ b/packages/sanity/src/core/perspective/useSetPerspective.tsx
@@ -12,7 +12,7 @@ export function useSetPerspective() {
     (releaseId: 'published' | 'drafts' | ReleaseId | undefined) => {
       router.navigate({
         stickyParams: {
-          excludedPerspectives: '',
+          excludedPerspectives: null,
           perspective: releaseId === 'drafts' ? '' : releaseId,
         },
       })
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -279,7 +279,16 @@ export function ReleasesOverview() {
   const handleOnCreateRelease = useCallback(
     (createdReleaseId: string) => {
       setIsCreateReleaseDialogOpen(false)
-      router.navigate({releaseId: createdReleaseId})
+
+      router.navigate(
+        {releaseId: createdReleaseId},
+        {
+          stickyParams: {
+            excludedPerspectives: '',
+            perspective: createdReleaseId,
+          },
+        },
+      )
     },
     [router],
   )
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -284,7 +284,7 @@ export function ReleasesOverview() {
         {releaseId: createdReleaseId},
         {
           stickyParams: {
-            excludedPerspectives: '',
+            excludedPerspectives: null,
             perspective: createdReleaseId,
           },
         },
--- a/packages/sanity/src/router/RouteScope.tsx
+++ b/packages/sanity/src/router/RouteScope.tsx
@@ -76,6 +76,8 @@ export const RouteScope = function RouteScope(props: RouteScopeProps): React.JSX
 
   const resolveNextParentState = useCallback(
     (_nextState: RouterState | null) => {
+      if (_nextState === null) return null
+
       const {_searchParams, ...nextState} = _nextState || {}
       const nextParentState = addScope(parentStateRef.current, scope, nextState)
       if (__unsafe_disableScopedSearchParams) {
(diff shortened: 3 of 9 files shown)


packages/sanity/src/router/types.ts
```diff
@@ -165,13 +165,20 @@ export type MatchResult = MatchError | MatchOk
 /**
  * @public
  */
-export interface NavigateOptions {
+export interface NavigateStickyParamsOptions {
```
bjoerge, line 168: Seems like this refactor should be the other way around? The `replace`-option belongs in `NavigateOptions`, not `NavigateStickyParamsOptions`, no?

jordanl17: Hmm, so the reason I did it this way round was because although it feels more intuitive that `NavigateStickyParamOptions` extends from `NavigateOptions`, `NavigateOptions` is the only one to include `options.stickyParams`. 

Perhaps I can just do `type NavigateStickyParamOptions = Omit<NavigateOptions, 'stickyParams'>` so it feels more natural?

packages/sanity/src/core/perspective/useSetPerspective.tsx
```diff
@@ -10,9 +10,11 @@ export function useSetPerspective() {
   const router = useRouter()
   const setPerspective = useCallback(
     (releaseId: 'published' | 'drafts' | ReleaseId | undefined) => {
-      router.navigateStickyParams({
-        excludedPerspectives: '',
-        perspective: releaseId === 'drafts' ? '' : releaseId,
+      router.navigate(null, {
```
bjoerge, line 13: this feels a bit wonky to me. `null` here meaning "stay on the current route"? Intuitively I'd expect `null` to mean "clear state" (i.e. go to root)

bjoerge: would it make sense to overload `navigate`?
```
navigate(nextState: RouterState, options: NavigateOptions)
navigate(options: NavigateOptions & {state?: RouterState|null})
```
Then we can treat omission of `state` in `navigateOptions` as "stay where you are", and `state: null` as "clear state" / "got to root"

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -279,7 +279,16 @@ export function ReleasesOverview() {
   const handleOnCreateRelease = useCallback(
     (createdReleaseId: string) => {
       setIsCreateReleaseDialogOpen(false)
-      router.navigate({releaseId: createdReleaseId})
+
+      router.navigate(
+        {releaseId: createdReleaseId},
+        {
+          stickyParams: {
+            excludedPerspectives: '',
```
bjoerge, line 287: Slightly clearer to use `undefined` (or `null` if types allows for it) instead of empty string?

packages/sanity/src/router/types.ts
```diff
@@ -246,7 +255,7 @@ export interface RouterContextValue {
   /**
    * Resolves the path from the given router state. See {@link RouterState}
    */
-  resolvePathFromState: (nextState: RouterState) => string
+  resolvePathFromState: (state: RouterState | null) => string
```
bjoerge, line 258: if we allow `null` here, we need to call out in tsdocs what passing `null` means

jordanl17: - [ ] add comment clarfiying behaviour of null in `resolvePathFromState`

packages/sanity/src/core/perspective/useSetPerspective.tsx
jordanl17: @bjoerge - Nice idea, I've made those changes now ([see the examples here](https://github.com/sanity-io/sanity/pull/8620/files/5cc515bf8043fabeab281796c8a6226e2510d47b#diff-6c4868b5ff17f27f7d71e825f071c0d55f5a90cde5db03fecac180ec963445bbR289-R305))

bjoerge: Very nice!

packages/sanity/src/router/RouteScope.tsx
```diff
@@ -97,13 +103,31 @@ export const RouteScope = function RouteScope(props: RouteScopeProps): React.JSX
...
-    (nextState: RouterState | null, options?: NavigateOptions) =>
-      parent_navigate(
-        // options?.stickyParams && nextState === null ? null :
-        resolveNextParentState(nextState),
-        options,
-      ),
+  const navigate: RouterContextValue['navigate'] = useCallback(
+    (nextStateOrOptions: NextStateOrOptions, maybeOptions?: NavigateOptions) => {
+      // Check if it's the options-only pattern
+      if (isNavigateOptions(nextStateOrOptions) && !maybeOptions) {
```
bjoerge, line 109: as an alternative, it should be safe to check the arity of the function as well, and treat it as `options` if it's passed 1 arg and `state` if 2. In theory, I guess we could run into trouble now if you do `navigate({state: 'foo'})`

packages/sanity/src/router/RouterProvider.tsx
```diff
@@ -119,22 +121,24 @@ export function RouterProvider(props: RouterProviderProps): React.JSX.Element {
     [routerProp, state],
   )
 
-  const navigate = useCallback(
-    (nextState: Record<string, unknown> | null, options: NavigateOptions = {}) => {
-      const currentParams = Array.isArray(state._searchParams) ? state._searchParams : []
+  const navigate: RouterContextValue['navigate'] = useCallback(
+    (nextStateOrOptions: NextStateOrOptions, maybeOptions?: NavigateOptions) => {
+      // Determine options and state based on input pattern
+      const isOptionsOnlyPattern = isNavigateOptions(nextStateOrOptions) && !maybeOptions
```
bjoerge, line 127: same as above re checking arity – not a blocker for merge since it doesn't affect the public api, so can improve this later on

packages/sanity/src/router/types.ts
```diff
@@ -165,95 +165,33 @@ export type MatchResult = MatchError | MatchOk
 /**
  * @public
  */
-export interface NavigateOptions {
-  /**
-   * Indicates whether to replace the current state.
-   */
+export interface NavigateBaseOptions {
   replace?: boolean
```
bjoerge, line 169: this one seems to have lost it's tsdoc comment

packages/sanity/src/router/types.ts
```diff
@@ -324,5 +268,92 @@ export interface RouterContextValue {
...
+export type RouterState = Record<string, unknown> & {_searchParams?: SearchParam[]}
+
+/**
+ * Guard to check if an argument is NavigateOptions with optional state
+ * @internal
+ */
+export const isNavigateOptions = (
+  maybeNavigateOptions: unknown,
+): maybeNavigateOptions is NavigateOptions & {state?: RouterState | null} =>
+  typeof maybeNavigateOptions === 'object' &&
```
bjoerge, line 348: see earlier comment about checking arity – might make this function redundant

bjoerge (approved):
Good to merge from my side. I left a couple of suggestions you can consider as follow-ups, but nice to get this one merged now. Great observation regarding the current behavior and good call fixing this fundamentally. Thank you @jordanl17!

PR #8623, opened 2025-02-13
fix(core): add permission limitation for create release

RitaDias:
### Description

Prevent create release
- [x] In Global Menu
- [x] In document headers
- [x] In the release list

image
image
image

Discard version discard
- [ ] Prevent discard version in document header
image

### What to review

Anything could be improved?

### Testing

The existing test should be enough.
Can manually test by following the screenshots

### Notes for release

Discard version and create releases action menus and buttons for releases now respect permissions.

### DIFF
--- a/packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
+++ b/packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
@@ -1,14 +1,17 @@
 import {AddIcon} from '@sanity/icons'
 import {Box, Flex, MenuDivider, Spinner} from '@sanity/ui'
-import {type RefObject, useCallback, useMemo} from 'react'
+import {type RefObject, useCallback, useEffect, useMemo, useState} from 'react'
 import {css, styled} from 'styled-components'
 
 import {MenuItem} from '../../../ui-components/menuItem/MenuItem'
 import {useTranslation} from '../../i18n/hooks/useTranslation'
 import {useReleasesUpsell} from '../../releases/contexts/upsell/useReleasesUpsell'
+import {useCreateReleaseMetadata} from '../../releases/hooks/useCreateReleaseMetadata'
 import {type ReleaseDocument, type ReleaseType} from '../../releases/store/types'
 import {useActiveReleases} from '../../releases/store/useActiveReleases'
-import {LATEST} from '../../releases/util/const'
+import {useReleaseOperations} from '../../releases/store/useReleaseOperations'
+import {useReleasePermissions} from '../../releases/store/useReleasePermissions'
+import {DEFAULT_RELEASE, LATEST} from '../../releases/util/const'
 import {getReleaseIdFromReleaseDocumentId} from '../../releases/util/getReleaseIdFromReleaseDocumentId'
 import {
   getRangePosition,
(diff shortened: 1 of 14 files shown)


packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
```diff
@@ -156,10 +170,13 @@ export function ReleasesList({
           <MenuDivider />
           <MenuItem
             icon={AddIcon}
-            disabled={mode === 'disabled'}
+            disabled={!hasCreatePermission || mode === 'disabled'}
             onClick={handleCreateBundleClick}
             text={t('release.action.create-new')}
             data-testid="create-new-release-button"
+            tooltipProps={{
```
jordanl17, line 177: Think you can do
```
tooltipProps={{
  disabled: !hasCreatePermission,
  content: t(...)
}}
```

packages/sanity/src/core/perspective/navbar/__tests__/ReleasesNav.test.tsx
```diff
@@ -60,6 +68,10 @@ const renderTest = async () => {
 describe('ReleasesNav', () => {
   beforeEach(() => {
     vi.clearAllMocks()
+
+    mockUseReleasePermissions.mockReturnValue({
```
jordanl17, line 72: Non-blocking: realise we just want to keep things moving right now, but could we worth creating a task to test when permissions is false too

packages/sanity/src/core/releases/components/documentHeader/contextMenu/VersionContextMenu.tsx
```diff
@@ -103,7 +125,10 @@ export const VersionContextMenu = memo(function VersionContextMenu(props: {
               icon={TrashIcon}
               onClick={onDiscard}
               text={t('release.action.discard-version')}
-              disabled={disabled || locked}
+              disabled={disabled || locked || !hasDiscardPermission}
+              tooltipProps={{
+                content: !hasDiscardPermission && t('release.action.permission.error'),
```
jordanl17, line 130: Same `tooltipProps.disabled` thing here

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -112,6 +120,10 @@ export function ReleasesOverview() {
     ]
   }, [hasReleases, releasesMetadata, releases])
 
+  useEffect(() => {
+    checkWithPermissionGuard(createRelease, DEFAULT_RELEASE).then(setHasCreatePermission)
```
jordanl17, line 124: Just a thought here:

If we updated to:
```
const {checkWithPermissionGuard, hasPermission} = useReleasePermissions()
```

Where `hasPermission` is basically the value of `hasCreatePermission` here.

This falls apart a little if a single component makes multiple releaseOperations permission checks, which I think in the other PR I recall seeing, but we could always call `useReleasePermissions` multiple times. Just think it could clean up always adding in a new state value and setting in the `.then`.

What do you think?

packages/sanity/test/mocks/useDocumentPairPermissions.mock.ts
jordanl17, line 1: Nice!

jordanl17 (changes requested):
Looking good @RitaDias - just added in a couple of comments. Would be interested to know what you think on managing the permission state inside `useReleasePermissions`

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
RitaDias: To keep this up to date, had a chat on slack and we decided to keep the solution as is after investigating how this solution would look like!

jordanl17 (approved):
Thanks

packages/sanity/src/core/perspective/navbar/__tests__/ReleasesNav.test.tsx
RitaDias: Added tests on this last commit, was trying to add a few more but I'm having some issues with the mocks.
Still, I think that the newer ones I've added should add some more confidence!

jordanl17 (approved):

packages/sanity/src/core/perspective/navbar/__tests__/ReleasesList.test.tsx
```diff
@@ -115,4 +115,35 @@ describe('ReleasesList', () => {
...
     })
   })
+
+  describe('when releases are enabled without permissions', () => {
+    beforeEach(async () => {
+      mockUseActiveReleases.mockReturnValue({
+        ...useActiveReleasesMockReturn,
+        data: [activeASAPRelease, activeScheduledRelease, activeUndecidedRelease],
+      })
+      mockUseReleasePermissions.mockReturnValue({
```
jordanl17, line 125: Ahh nice!

PR #8632, opened 2025-02-14
fix(core): releases edit events error when calculating differences

pedrobonamin:
### **Pre-read**
The Releases Activity Panel was previously relying on release documents from the **releases store**. Initially, these documents were complete. However, we recently introduced a [projection to the documents](https://github.com/sanity-io/sanity/blob/sapp-2101/packages/sanity/src/core/releases/store/createReleaseStore.ts#L39-L60), which means they are now **partial**.

The activity panel retrieves changes using the **transactions endpoint** and applies **Mendoza patches** to the documents. However, Mendoza patches require **complete documents** to function correctly—they **do not** work with partial documents.

---

### **Fix**
This PR refactors the `getReleaseEditEvents` observable to use the new `observeDocument` observable, which ensures that **complete** documents are returned. This resolves the issue where Mendoza patches failed due to incomplete data.

---

### **Other Changes**
- The **global listener** was filtering out **system documents**, and since **releases** are classified as system documents, they were also being excluded.
- To allow `observeDocument` to function correctly, this PR **removes that filter**, ensuring system documents (including releases) are properly observed.

### What to review

Any side effect to consider on the listener change?

### Testing

The corresponding tests are updated.
Opening a release activity panel should not error in any case.

### Notes for release
Fixes an error when getting the release edit events in the release activity panel

### DIFF
--- a/packages/sanity/src/core/preview/createGlobalListener.ts
+++ b/packages/sanity/src/core/preview/createGlobalListener.ts
@@ -11,7 +11,7 @@ import {shareReplayLatest} from './utils/shareReplayLatest'
 export function createGlobalListener(client: SanityClient) {
   return client
     .listen(
-      '*[]',
+      '*',
       {},
       {
         events: ['welcome', 'mutation', 'reconnect'],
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Nice!

pedrobonamin:
@bjoerge this needs re approval after removing the `[]` from the filter https://github.com/sanity-io/sanity/pull/8632/commits/5e7d36d3c8787bf94ea9e94439c60432b01af87a

bjoerge (approved):

PR #8635, opened 2025-02-14
fix(core): simplify pin release  button on release detail screen

pedrobonamin:
### Description
Two images say more than 2.000 words

Screenshot 2025-02-14 at 11 49 04
Screenshot 2025-02-14 at 11 49 13


RitaDias (dismissed):
They do indeed 👓

jordanl17:
Screenshot 2025-02-14 at 11 21 49
Should we update this on release overview too - we just use `pin release` in both cases

RitaDias (approved):

PR #8644, opened 2025-02-14
fix(releases): handling cases where no create release permissions whilst at limit of release in project

jordanl17:
### Description
With the new permission checks for releases (by running a `dryRun` release action to check for permissions, there were some issues when the project was already at the release limit for the plan.

> [!NOTE]
> The Upsell components here will be further updated for organisation level limit - this is a stopgap until then so will be very short lived

**No Permissions but at release limit**
Before:
The upsell would keep flashing repeatedly (because of the dryRun request triggering the dialog)
noPermsAtLimitOLD

After [FIXED]:
The create buttons are enabled (this is because content lake returns the limit error rather than the permission error - which will be changed shortly), but the upsell only shows when an action is attempted by the user
noPermsAtLimitNEW

**No Permissions but under the limit**
Before:
The create buttons are correctly disabled because of permissions
noPermsUnderLimitNEW

After [SAME]:
The create buttons are correctly disabled because of permissions
noPermsUnderLimitNEW

**Permissions but at release limit**
Before:
Upsell keeps flashing (because of the dryRun request triggering)
permsAtLimitOLD

After [FIXED]:
Will get upsell when trying to perform create action
permsAtLimitNEW

**Permissions but under the release limit**
Before:
Works as expected
permsUnderLimitNEW

After [SAME]:
Works as expected
permsUnderLimitNEW

### What to review
* Abstracting away a check for `error.details` to `core/error/types` - this makes sense?
* Only Asserting that permissions don't exist when the error is specifically permissions related

### Testing
Manually running through the above scenarios

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
+++ b/packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
@@ -172,10 +172,10 @@ export function ReleasesUpsellProvider(props: {children: React.ReactNode}) {
   )
 
   const onReleaseLimitReached = useCallback(
-    (limit: number) => {
+    (limit: number, suppressDialogOpening: boolean = false) => {
       setReleaseLimit(limit)
 
-      if ((activeReleases?.length || 0) >= limit) {
+      if (!suppressDialogOpening && (activeReleases?.length || 0) >= limit) {
         handleOpenDialog()
       }
     },
--- a/packages/sanity/src/core/releases/store/useReleasePermissions.ts
+++ b/packages/sanity/src/core/releases/store/useReleasePermissions.ts
@@ -1,9 +1,16 @@
+import {isErrorWithDetails} from '../../error/types/isErrorWithDetails'
+
 export interface useReleasePermissionsValue {
   checkWithPermissionGuard: <T extends (...args: any[]) => Promise<void> | void>(
     action: T,
     ...args: Parameters<T>
   ) => Promise<boolean>
 }
+
+type ReleasePermissionError = {details: {type: 'insufficientPermissionsError'}}
+
+export const isReleasePermissionError = (error: unknown): error is ReleasePermissionError =>
+  isErrorWithDetails(error) && error.details?.type === 'insufficientPermissionsError'
 
 /**
  * @internal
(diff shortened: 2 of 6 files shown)


packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
```diff
@@ -172,10 +172,10 @@ export function ReleasesUpsellProvider(props: {children: React.ReactNode}) {
   )
 
   const onReleaseLimitReached = useCallback(
-    (limit: number) => {
+    (limit: number, suppressDialogOpening: boolean = false) => {
```
jordanl17, line 175: This might be a temporary change, or might stay around in the upcoming org level release limiting work. This is added here so that the dryRun release actions can retrieve the limit (if it's been reached) but doesn't start firing the upsell until the user explicitly takes an action to encounter it

packages/sanity/src/core/releases/store/useReleasePermissions.ts
```diff
@@ -27,7 +34,7 @@ export function useReleasePermissions(): useReleasePermissionsValue {
       })
       return true
     } catch (e) {
-      return false
+      return !isReleasePermissionError(e)
```
jordanl17, line 37: Only return permissions are invalid if the error is for that particular case
> [!NOTE]
> The backend will need updating shortly to return permission insufficent error over release limit

RitaDias: To double check, this is the thing you mentioned that CL had in the wings right?

RitaDias (approved):
Small nitpick and questions but nothing blocking, you are a legend thanks!

packages/sanity/src/core/releases/store/useReleasePermissions.ts
jordanl17: Yeh that's correct - I'm looking into it with them

packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
jordanl17: Good shout!! Will add that to the org level limiting work if that's ok

PR #8655, opened 2025-02-17
refactor: consistently apply current perspective for previews

bjoerge:
### Description
This is a somewhat big refactor of how we resolve values for previews internally.

Previously, for each document the studio was previewing, we'd fetch a version of the preview for each known release.
This created some overhead (althought the requests are batched), but most notably it caused an issue with previews ending up showing referenced values from the published version.

This is now solved by always applying the current perspective stack when fetching preview values. The perspective stack is applied recursively, so if a document type has `someReference.title` in it's preview config, the `title` of the referenced document will also reflect the current perspective (i.e. what the `title` is in the selected release).

Making the perspective apply to previews enabled me to remove some layering logic, since this is now done in the query that fetches the previews.

Using the just the current perspective stack poses a challenge: Sometimes you want to preview something that does _not exist in the current perspective_. For example: you are in the `drafts` perspective, but you want to preview a document that only exists as a version (i.e. it has neither a draft nor a published variant). This happens e.g. when previewing document list items or search results since we these collections using perspective=raw because we want people to still find version documents that exists _outside_ of the current perspective. 

For these cases, we always have a version id, so in these cases we'll fetch:
- A preview for the document with current perspective applied
- A preview for the document with the _version_ perspective applied.
If the document doesn’t exist in the current perspective, it certainly exist in it’s own version perspective, so that means we’ll always have something to preview.

Things to note:
- Since perspective applies to the query request, I had to change the preview observer batch fetcher memo to include the perspective stack as key. 
- The previous implementation fetched a preview version for each known release in order to answer "what other releases does this document exist in". This question can be answered more simply by fetching the _rev for each version document and check for existence (so no need to fetch the full preview). The information about what other version exists is now decoupled from previews entirely, and can be retrieved using the new `useDocumentVersionInfo()`-hook. We no longer have _preview_ for the version documents because we didn't have any use for that at the moment. We can always come back to that decision in the future, but for now, they return a stub with _id, _createdAt, _updatedAt, _rev, etc.
- We were currently supporting a perspective for search, which only seems to apply to the global search atm. I took a chance on removing that, because I don't think that's desired. As of this PR, global search will use perspective=raw. 

## Other changes
- Added a `useDocumentVersionInfo(<documentId>)`-hook that will return the versions that exists for the given document id
- Simplified the `getPreviewStateObservable()` so it no longer requires a fallback title – the fallback can trivially be applied in the consumer end, and it lead to a bunch of cases where we passed an empty string as the third argument.

### What to review
This should be reviewed a bit carefully. Esp. in terms of API changes and naming (is it sensible and understandable?)

### Testing
We have fairly good test coverage already, and e2e tests already made me aware of a few oversights. Would love some additional testing  of everything related to previews, search and references e.g:
- Document list previews
- Global search
- Adding documents to release
- Searching document lists
- Searching for references
- Reference previews in Portable Text Editor, arrays and objects

### Notes for release
- Fixes a bug that caused previews to sometimes display values from published documents instead of drafts or versions.
- Fixes a bug that caused documents outside of the current release to not appear in global search
- Fixes a bug that would show "Missing document" for any document outside of the selected release without published or draft variants.


packages/sanity/src/core/preview/useValuePreview.ts
```diff
@@ -37,17 +38,33 @@ function useDocumentPreview(props: {
...
   const {observeForPreview} = useDocumentPreviewStore()
+  const {perspectiveStack, selectedPerspective} = usePerspective()
   const observable = useMemo<Observable<State>>(() => {
     // this will render previews as "loaded" (i.e. not in loading state) – typically with "Untitled" text
     if (!enabled || !previewValue || !schemaType) return of(IDLE_STATE)
 
     return observeForPreview(previewValue as Previewable, schemaType, {
+      perspective:
+        // eslint-disable-next-line no-nested-ternary
+        selectedPerspective === 'drafts'
```
jordanl17, line 49: Q: Had we not mentioned about making this the default logic for `perspectiveStack`? Was there a reason not to do that in the end, or just to keep this PR from scope-creep?

bjoerge: Yeah, wanted to avoid scope creep and figured it's better to do it in a separate pass anyway to minimize blast radius :)

juice49 (approved):
Nice! Thanks so much for implementing this. The changes look sensible to me, and Studio is working nicely in my testing.

juice49 (approved):

packages/sanity/src/core/util/draftUtils.ts
```diff
@@ -193,6 +193,7 @@ export function collate<T extends {_id: string; _type: string}>(documents: T[]):
       res.set(publishedId, entry)
     }
 
+    // note: this attaches versions to the `draft` property, which is likely ok for most practical purposes
```
juice49, line 196: Oh! We are treading the same path 😅.

juice49 (approved):
This is great! Still works as expected in my testing. Thanks 🙌.

packages/sanity/src/core/studio/components/navbar/search/contexts/search/reducer.ts
```diff
@@ -591,3 +597,8 @@ function stripRecent(terms: RecentSearch | SearchTerms) {
   }
   return terms
 }
+
+function collateHits(hits: SearchHit[]) {
+  const collated = collate(hits.map((h) => h.hit))
```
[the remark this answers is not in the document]
bjoerge: Cool - yeah, it felt a bit dirty. I guess the use case for it is to be able to tell which variants had matches? (I'm not sure if we currently rely on it for anything)

packages/sanity/src/core/util/draftUtils.ts
bjoerge: 🤝

juice49 (dismissed):

juice49 (approved):

PR #8656, opened 2025-02-17
refactor(core): prevent updating release title and description if the user doesn't have permissions

RitaDias:
### Description

If a user doesn't have user permissions, they will not be allowed to edit the title (or perceive that they could otherwise).
| With Permission    | Without Permission |
| -------- | ------- |
<video src="https://github.com/user-attachments/assets/a46d88e8-3547-4a02-a11c-764ef49c7a96">  |  <video src="https://github.com/user-attachments/assets/ec7ba185-70ed-4b1d-8055-71380c2a4c75">
   |

### What to review

Double check the code makes sense,
I thought that I could have done the check all the way in the ReleasesList (which also calls for the same method in the useEffect, however, I felt like it made more sense to cause a re-render further down _if_ it's needed than potentially re-rendering multiple things at a component at a higher level since that's not the only child that it has)

### Testing

Unit tests added
You should be able to manually test by following the video with a user that has no permissions

### Notes for release

Editing a Release title and description now respects role permissions

### DIFF
--- a/packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
+++ b/packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
@@ -1,11 +1,15 @@
-import {useCallback, useState} from 'react'
+import {useCallback, useEffect, useState} from 'react'
 
 import {TitleDescriptionForm} from '../../components/dialog/TitleDescriptionForm'
 import {type EditableReleaseDocument, type ReleaseDocument, useReleaseOperations} from '../../index'
+import {useReleasePermissions} from '../../store/useReleasePermissions'
 
 export function ReleaseDetailsEditor({release}: {release: ReleaseDocument}): React.JSX.Element {
   const {updateRelease} = useReleaseOperations()
   const [timer, setTimer] = useState<NodeJS.Timeout | undefined>(undefined)
+
+  const {checkWithPermissionGuard} = useReleasePermissions()
+  const [hasUpdatePermission, setHasUpdatePermission] = useState<boolean | null>(null)
 
   const handleOnChange = useCallback(
     (changedValue: EditableReleaseDocument) => {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
```diff
@@ -1,25 +1,44 @@
...
 
       setTimer(newTimer)
     },
-    [timer, updateRelease],
+    [hasUpdatePermission, timer, updateRelease],
   )
 
-  return <TitleDescriptionForm key={release._id} release={release} onChange={handleOnChange} />
+  useEffect(() => {
+    checkWithPermissionGuard(updateRelease, release).then((hasPermission) =>
```
juice49, line 31: I'm not very familiar with `checkWithPermissionGuard` and whether we call it in an effect elsewhere, but should we provide an effect clean-up function here that cancels the request?

I'm not sure whether we can easily provide an `AbortSignal` to the action dry-run in its current form (maybe we can), but wondering whether we should at least skip `setHasUpdatePermission` if the effect clean-up has been called?

packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
```diff
@@ -1,25 +1,44 @@
...
 
 import {TitleDescriptionForm} from '../../components/dialog/TitleDescriptionForm'
 import {type EditableReleaseDocument, type ReleaseDocument, useReleaseOperations} from '../../index'
+import {useReleasePermissions} from '../../store/useReleasePermissions'
 
 export function ReleaseDetailsEditor({release}: {release: ReleaseDocument}): React.JSX.Element {
   const {updateRelease} = useReleaseOperations()
   const [timer, setTimer] = useState<NodeJS.Timeout | undefined>(undefined)
 
+  const {checkWithPermissionGuard} = useReleasePermissions()
```
juice49, line 11: A bit beyond the scope of this PR, but wondering whether duplicate HTTP requests might become an issue if we end up checking permissions in lots of different parts of the UI 🤔.

juice49 (commented):
This is a very nice improvement! Left a couple of questions, but looks good to me in general 🙌.

packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
RitaDias: Fair, added it!

packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
RitaDias: Spoke privately, I'll a go at something:tm: ouside of the scope of this PR

packages/sanity/src/core/releases/tool/detail/ReleaseDetailsEditor.tsx
RitaDias: Done :)

juice49 (approved):
Nice! 🙂

PR #8661, opened 2025-02-17
feat(sanity): add granular release error icons to global switcher

juice49:
### Description

This branch adds release error indicators to the global perspective menu.

Screenshot 2025-02-17 at 10 18 08

### What to review

- The global perspective menu for releases in an error state.

### Testing

- Added unit test.
- Manually tested with a release that has an error.
- Manually tested with a locked release that has an error, to ensure the states combine harmoniously.

### DIFF
--- a/packages/sanity/src/core/perspective/navbar/GlobalPerspectiveMenuItem.tsx
+++ b/packages/sanity/src/core/perspective/navbar/GlobalPerspectiveMenuItem.tsx
@@ -1,9 +1,10 @@
-import {DotIcon, EyeClosedIcon, EyeOpenIcon, LockIcon} from '@sanity/icons'
+import {DotIcon, ErrorOutlineIcon, EyeClosedIcon, EyeOpenIcon, LockIcon} from '@sanity/icons'
 // eslint-disable-next-line no-restricted-imports -- custom use for MenuItem & Button not supported by ui-components
 import {Box, Button, Flex, MenuItem, Stack, Text} from '@sanity/ui'
 import {type CSSProperties, forwardRef, type MouseEvent, useCallback, useMemo} from 'react'
 import {css, styled} from 'styled-components'
 
+import {ToneIcon} from '../../../ui-components/toneIcon/ToneIcon'
 import {Tooltip} from '../../../ui-components/tooltip'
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/perspective/navbar/GlobalPerspectiveMenuItem.tsx
```diff
@@ -168,9 +169,18 @@ export const GlobalPerspectiveMenuItem = forwardRef<
...
-            <Text size={1} weight="medium">
-              {displayTitle}
-            </Text>
+            <Flex gap={3} align="center">
+              <Text size={1} weight="medium">
+                {displayTitle}
+              </Text>
+              {isReleaseDocument(release) &&
+                typeof release.error !== 'undefined' &&
+                release.state === 'active' && (
```
pedrobonamin, line 178: Wondering if it could happen that the release state is in the pending ones, like "scheduling" | "publishing" and we get an error? 

Do we need to have this release.state check?

pedrobonamin (approved):
Looks good to me, added a question up to discuss it

packages/sanity/src/core/perspective/navbar/GlobalPerspectiveMenuItem.tsx
juice49: It's a good point. My expectation here is that it's only relevant to surface the error in the UI after a failure has caused Content Lake to transition the release back to an `active` state.

Imagine a release that has failed once, resulting in `release.error` being populated and the state being transitioned back to `active`. Now imagine the user publishes or schedules the release again to "retry". Does Content Lake clear `release.error` at this point?

I don't know the answer, but only displaying the error when the release is `active` _feels_ like a sensible way to approach it.

I do think we should:

1. Clarify this point.
2. Centralise the logic to a function that can be shared everywhere release error is displayed (I think this is something you previously recommended).

I'll create issues for this…

pedrobonamin: Sounds reasonable, thank you for adding the extra explanation. I think this makes all the difference
>Does Content Lake clear release.error at this point?

PR #8676, opened 2025-02-17
fix: switch to dated api version for releases

bjoerge:


RitaDias (commented):
Is it worth moving these apiVersions definitions to a const that we have somewhere and pull from there? I'm thinking future proofing where we _eventually_ will just want to probably have a "source of truth" of what most of our clients will use?

So maybe a const defined in the `studioClient` that then is imported by the rest?

RitaDias (approved):
Everything seems to be working as intended, so nothing is blocking but I'd consider this comment: https://github.com/sanity-io/sanity/pull/8676#pullrequestreview-2628932689 before merging. If needed I'll give another green again!

bjoerge:
> Is it worth moving these apiVersions definitions to a const that we have somewhere and pull from there.

Agree, it's good idea to have a central place to define the API version(s) used by the Studio. I'd suggest we make it a separate task to centralize all places we currently use hard coded strings.

RitaDias:
> > Is it worth moving these apiVersions definitions to a const that we have somewhere and pull from there.
> 
> Agree, it's good idea to have a central place to define the API version(s) used by the Studio. I'd suggest we make it a separate task to centralize all places we currently use hard coded strings.

Created the [story](https://linear.app/sanity/issue/SAPP-2421/centralize-all-places-we-currently-use-hard-coded-strings-for-client) cc @bjoerge @EoinFalconer

PR #8677, opened 2025-02-17
feat(core): updates archived releases overview

pedrobonamin:
### Description

This PR updates the archived releases table overview.
Adding two columns that are exclusive for this view:
- **Published**
- **Archived**

This will be the anchor for users to make it possible to determine if a release was published or archived.
It also renames the **Open** tab to **Active**

And fixes a small oversight in were users could edit archived releases descriptions.

### What to review
Are the changes correct?

### DIFF
--- a/packages/sanity/src/core/releases/components/dialog/TitleDescriptionForm.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/TitleDescriptionForm.tsx
@@ -115,13 +115,12 @@ export function TitleDescriptionForm({
   const handleTitleChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
       event.preventDefault()
-      if (!isReleaseOpen) return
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/releases/components/dialog/TitleDescriptionForm.tsx
```diff
@@ -115,17 +115,20 @@ export function TitleDescriptionForm({
   const handleTitleChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
       event.preventDefault()
+      if (!isReleaseOpen) return
```
jordanl17, line 118: I don't think this is required, looks like `readOnly` on the field is enough to disable updates

jordanl17 (approved):

juice49 (approved):

PR #8679, opened 2025-02-17
fix(core): infer archive, unarchive and publish from release translog

pedrobonamin:
### Description

This PR updates how we get the `releasesEditEvents` and infers from the translog three extra events.
- ArchiveReleaseEvent
- PublishReleaseEvent
- UnarchiveReleaseEvent

This was not initially necessary because we were using the releases events api, but that is not stable yet, so users are not seeing this information.
By generating this events from the translog we can show a more accurate activity events list to the users using this feature.

| Before | After |
|--------|--------|
|image | image |
| image | image |


packages/sanity/src/core/releases/store/types.ts
```diff
@@ -31,7 +31,6 @@ export type ReleaseState =
 export type ReleaseFinalDocumentState = {
   /** Document ID */
   id: string
-  revisionId: string
```
pedrobonamin, line 34: This is not used and it's not received from the API, I think it was added by @jordanl17 could you please confirm this change?

packages/sanity/src/core/releases/tool/detail/events/buildReleaseEditEvents.test.ts
```diff
@@ -574,4 +574,482 @@ describe('buildReleaseEditEvents()', () => {
       },
     ])
   })
+  it('should identify state changes', () => {
+    const transactions = [
```
RitaDias, line 578: this is a nitpick but I think a worthwhile one! could we move these mocks to a separate file and import them to the test? :) 

The reason why I'm asking is two fold: 1. make it easier for the tests to be readable at a glance, 2. make it reusable in future tests both in this file and on others!

packages/sanity/src/core/releases/tool/detail/events/buildReleaseEditEvents.ts
```diff
@@ -3,19 +3,37 @@ import {type TransactionLogEventWithEffects} from '@sanity/types'
...
 export function buildReleaseEditEvents(
   transactions: TransactionLogEventWithEffects[],
   release: ReleaseDocument,
-): (EditReleaseEvent | CreateReleaseEvent)[] {
+): (
+  | EditReleaseEvent
+  | CreateReleaseEvent
+  | ArchiveReleaseEvent
+  | PublishReleaseEvent
+  | UnarchiveReleaseEvent
```
RitaDias, line 22: Since we are repeating these types, does it make sense to create another level where we group them and reuse it here and over at releaseEditEvents?

RitaDias: And it might even be used over at editEvents as well? the order of things are slightly different so forgive me if they aren't exactly 1:1 but they look? like they are?

packages/sanity/src/core/releases/tool/detail/events/getReleaseEvents.ts
```diff
@@ -129,3 +129,22 @@ export function getReleaseEvents({
     loadMore: activityEvents.loadMore,
   }
 }
+
+const deduplicateEvents = (events: ReleaseEvent[]) => {
+  // Events are sorted by timestamp, compare this event with the next one, if they are the same type and timestamp, remove it.
```
RitaDias, line 134: These comments are great, thank you :)

packages/sanity/src/core/releases/tool/detail/events/getReleaseEvents.ts
```diff
@@ -129,3 +129,22 @@ export function getReleaseEvents({
...
+}
+
+/**
+ * Checks if two dates are the same date, without contemplating miliseconds.
+ * The translog has miliseconds but the events api does not.
+ */
+function areTheSameDate(date1: string, date2: string) {
+  const time1 = new Date(date1).getTime()
+  const time2 = new Date(date2).getTime()
+  return Math.abs(time1 - time2) <= 1000
```
RitaDias, line 149: Another nitpick and it might be related with the approach we want to go with moving forward, but any reason to use Math.abs vs the date library? Or is it more a "if we don't need to import it, we shouldn't"?

RitaDias (commented):
Everything looks good! There are a couple of questions I have and some smaller things that I think we could benefit from? But nothing blocking 🙏 

Thanks for this!

packages/sanity/src/core/releases/tool/detail/events/getReleaseEvents.ts
pedrobonamin: I'm not sure if the date library will handle this.
The problem I found is that dates in the **events api** don't have the milliseconds but the dates in the translog do have milliseconds.

So to compare them I check if the difference between them is less than 1000 miliseconds

packages/sanity/src/core/releases/tool/detail/events/buildReleaseEditEvents.test.ts
pedrobonamin: Will move them! Yeah, they are quite big.

packages/sanity/src/core/releases/tool/detail/events/getReleaseEvents.ts
RitaDias: Ah fair, that's a good point, I was thinking about something like [this](https://date-fns.org/v4.1.0/docs/isEqual), but again, not sure if it's worth the import 🤷

packages/sanity/src/core/releases/tool/detail/events/buildReleaseEditEvents.ts
pedrobonamin: Good catch, I updated this, it doesn't make sense to block the type to this type of events, in the future we may want to return any other of the [ReleaseEvent](https://github.com/sanity-io/sanity/blob/releases-history-events/packages/sanity/src/core/releases/tool/detail/events/types.ts#L3-L4)

RitaDias (approved):
heck yeah! 🙌

packages/sanity/src/core/releases/store/types.ts
jordanl17: Fine to move it - I believe I added it pre-emptively before the API was actually implemeneted

PR #8681, opened 2025-02-17
feat(structure): add unpublish document as part of a release banner

pedrobonamin:
### Description

Adds visual indicator for documents that will be unpublished as part of a release.
It also disables the document form from edits, by making it read only.

Screenshot 2025-02-17 at 18 16 35

### DIFF
--- a/packages/sanity/src/structure/i18n/resources.ts
+++ b/packages/sanity/src/structure/i18n/resources.ts
@@ -139,6 +139,10 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The text for the banner that appears when a document is not in the current global release */
   'banners.release.not-in-release': 'Not in the <Label>{{title}}</Label> release.',
 
+  /** The text content for the unpublished document banner when is part of a release */
+  'banners.unpublished-release-banner.text':
+    'This document will be unpublished as part of the <strong>{{title}}</strong> release',
+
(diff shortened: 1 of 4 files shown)


packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -139,6 +139,10 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The text for the banner that appears when a document is not in the current global release */
   'banners.release.not-in-release': 'Not in the <Label>{{title}}</Label> release.',
 
+  /** The text content for the unpublished document banner when is part of a release */
+  'banners.unpublished-release-banner.text':
+    'This document will be unpublished as part of the <strong>{{title}}</strong> release',
```
jordanl17, line 144:
```suggestion
    'This document will be unpublished as part of the <Label>{{title}}</Label> release',
```

Just to keep it consistent I see we use `Label` rather than `strong` for `banners.release.not-in-release`

jordanl17 (changes requested):
I think it would be good for us to keep the banner consistent by using the release `Label` - I'll try and make this minor change today and merge, if I don't get round to it I'll swap this to approved as it shouldn't be blocking really

jordanl17 (approved):

RitaDias (approved):

PR #8682, opened 2025-02-17
feat: org release limits

jordanl17:
### Description
With the API returning
| Scenario | Demo |
|--------|--------|
| Free project with no trial | freeNOTrialWithApi |
| Free project with growth trial | withTrialWithAPI |
| Enterprise project under limit | WITHAPI-UNDER_LIMIT |
| Enterprise project at limit | withAPI-atLimit | 

With the API not existing
| Scenario | Demo |
|--------|--------|
| Free project with no trial | freeNOTrialWITHOUTApi |
| Free project with growth trial | actualGRWNOAPI |
| Enterprise project under limit | WITHOUTAPI-underlimit |
| Enterprise project at limit | WITHOUTAPI-atlimit | 

Releases+
Free project after trial:
Screenshot 2025-02-21 at 14 13 40

Free project with trial:
Screenshot 2025-02-21 at 14 13 40

Enterprise project:
Screenshot 2025-02-21 at 14 30 39

### What to review
* The 2 new persistent stores `useOrgActiveReleaseCount` and `useReleaseLimits`, and the way they emit their data
* * How graceful handing of errors works for these hook - they support both the error cases from the endpoint, where `data` will still be returned an we assume limits not reached and releases+ not enabled, and internal errors on the server, eg when it goes away or is down.
* The caching on `useOrgActiveReleaseCount - it caches for a TTL (60s) or until the `activeReleases.length` changes
* How the guard now works - it fetches the limits then uses comparisons of the limits against the `activeReleases.length` and `data.orgActiveReleaseCount` 
* Determining releases+ support from the org limits and the default

### Testing
Manually tested by above scenarios for now - unlikely we will fast-follow tests given this solution will go away in the close future

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
+++ b/packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
@@ -184,7 +184,8 @@ export function ReleasesList({
             text={t('release.action.create-new')}
             data-testid="create-new-release-button"
             tooltipProps={{
-              content: !hasCreatePermission && t('release.action.permission.error'),
+              disabled: hasCreatePermission === true,
+              content: t('release.action.permission.error'),
             }}
           />
         </>
--- a/packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
+++ b/packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
@@ -1,8 +1,9 @@
+/* eslint-disable no-console */
 import {useTelemetry} from '@sanity/telemetry/react'
 import {template} from 'lodash'
 import {useCallback, useEffect, useMemo, useState} from 'react'
 import {firstValueFrom, of} from 'rxjs'
-import {delay, tap} from 'rxjs/operators'
+import {delay, shareReplay, tap} from 'rxjs/operators'
 import {ReleasesUpsellContext} from 'sanity/_singletons'
 
 import {useClient, useFeatureEnabled, useProjectId} from '../../../hooks'
@@ -19,12 +20,6 @@ import {useActiveReleases} from '../../store/useActiveReleases'
 import {useOrgActiveReleaseCount} from '../../store/useOrgActiveReleaseCount'
 import {useReleaseLimits} from '../../store/useReleaseLimits'
 import {type ReleasesUpsellContextValue} from './types'
-
-type ReleaseLimits = {
-  datasetReleaseLimit: number
-  orgActiveReleaseCount: number
-  orgActiveReleaseLimit: number
-}
 
 class StudioReleaseLimitExceededError extends Error {
   details: {type: 'releaseLimitExceededError'}
@@ -45,7 +40,7 @@ const API_VERSION = '2024-04-19'
 
 export const fetchReleasesLimits = () =>
   of({
-    orgActiveReleaseCount: 10,
+    orgActiveReleaseCount: 6,
     orgActiveReleaseLimit: 20,
     datasetReleaseLimit: 6,
 
@@ -53,6 +48,7 @@ export const fetchReleasesLimits = () =>
     // orgActiveReleaseLimit: 6,
     // datasetReleaseLimit: 10,
   }).pipe(
+    shareReplay(1),
     tap(() => console.log('fetchReleasesLimits')),
     delay(3000),
   )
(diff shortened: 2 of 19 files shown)


packages/sanity/src/core/perspective/navbar/ReleasesList.tsx
```diff
@@ -184,7 +184,8 @@ export function ReleasesList({
             text={t('release.action.create-new')}
             data-testid="create-new-release-button"
             tooltipProps={{
-              content: !hasCreatePermission && t('release.action.permission.error'),
+              disabled: hasCreatePermission === true,
```
jordanl17, line 187: @RitaDias checking that this (and other uses elsewhere) is `true` explicitly so that the tooltip doesn't show whilst we are pending fetching the permissions

packages/sanity/src/core/releases/contexts/upsell/fetchReleaseLimits.ts
```diff
@@ -0,0 +1,63 @@
...
+       * resolves by attaching the store name to the tag, this spoofs 2 unique requests
+       * allowing both the fire in parallel.
+       */
+      tag: `new-content-release-allowed-${storeOrigin}`,
+    })
+    .pipe(
+      catchError((error: ClientError) => {
+        console.error(error)
+
+        if (typeof error.response.body !== 'string' && 'data' in error.response.body) {
```
juice49, line 51: Does this handle the scenario of the quota being reached? If so, would it be safer to additionally check the response status code?

I'm just wondering whether there are any other errors that could produce a response body containing a `data` property that would incorrectly be caught here.

packages/sanity/src/core/releases/store/createReleaseOperationStore.ts
```diff
@@ -355,9 +355,11 @@ export function createRequestAction(
         },
       })
     } catch (e) {
-      if (isReleaseLimitError(e)) {
+      // if dryRunning then essentially this is a silent request
+      // so don't want to create disruptive upsell because of limit
```
juice49, line 359: Nice 🙌

packages/sanity/src/core/releases/store/useOrgActiveReleaseCount.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export const useOrgActiveReleaseCount = () => {
+  const resourceCache = useResourceCache()
+  const {data: activeReleases} = useActiveReleases()
+  const client = useClient()
+
+  const activeReleasesCount = activeReleases?.length || 0
+
+  // dependencies must be objects not primitives, so nesting activeReleasesCount in an object
+  const count = useMemo(() => ({activeReleasesCount}), [activeReleasesCount])
```
juice49, line 85: By creating an object in the `useOrgActiveReleaseCount` hook to use as a key, won't this code create and cache a new instance of the `createOrgActiveReleaseCountStore` store every time it renders?

jordanl17: Yeh that's true. This is a bug for sure! It doesn't get surfaced as a bug currently because `useorgActiveReleaseCount` is only used in `ReleaseUpsellProvider`. Perhaps `useActiveReleases` should be returning a memoized object instead?

juice49: I think I would consider having the observable produced by `createOrgActiveReleaseCountStore` consume the observable produced by `useActiveReleases` and deriving its state on that, rather than injecting a snapshot of it at creation-time.

It's not necessarily a problem for now. Just thinking out loud.

packages/sanity/src/core/releases/contexts/upsell/fetchReleaseLimits.ts
```diff
@@ -13,6 +13,14 @@ interface ReleaseLimitsResponse {
   data: ReleaseLimits
 }
 
+function isErrorAtLimits(error: ClientError) {
```
juice49, line 16: Nice!

juice49 (approved):
Looks good to me!

packages/sanity/src/core/releases/contexts/upsell/ReleasesUpsellProvider.tsx
```diff
@@ -146,40 +145,91 @@ export function ReleasesUpsellProvider(props: {children: React.ReactNode}) {
...
+  )
+
+  const {releaseLimits$} = useReleaseLimits()
+  const {orgActiveReleaseCount$} = useOrgActiveReleaseCount()
 
   const guardWithReleaseLimitUpsell = useCallback(
-    (cb: () => void, throwError: boolean = false) => {
-      if (mode === 'default') {
-        return cb()
+    async (cb: () => void, throwError: boolean = false) => {
```
RitaDias, line 170: the most minor of nits but naming wise I tend to not like having such truncated names and I know this is already from the previous code and not new _but_ just thought to bring it up: I imagine that it's "callback" but it took me a bit to read through and undestand it.

packages/sanity/src/core/releases/contexts/upsell/fetchReleaseLimits.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @internal
+ * fetches subscriptions for this project
+ */
+export function fetchReleaseLimits(
+  client: SanityClient,
+  storeOrigin: string,
+): Observable<ReleaseLimits> {
+  const {projectId} = client.config()
+
+  // This endpoint is prone to optimisation and further work
```
RitaDias, line 34: these comments are literal life saviours 🙏

packages/sanity/src/core/releases/hooks/useIsReleasesPlus.ts
RitaDias, line 1: A brief thought that I had, is if it would be worth to move the stuff related to upselling and such to its own folder within the releases in core. A thought I had

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -241,25 +242,41 @@ export function ReleasesOverview() {
...
   const createReleaseButton = useMemo(
     () => (
       <UIButton
         icon={AddIcon}
-        disabled={!hasCreatePermission || isCreateReleaseDialogOpen || mode === 'disabled'}
+        disabled={
+          isPendingGuardResponse ||
+          !hasCreatePermission ||
+          isCreateReleaseDialogOpen ||
+          mode === 'disabled'
```
RitaDias, line 261: Small nitpick as well, condition is starting to get long, might be worth pulling it up to its own const at some point

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/ReleaseRevertButton.tsx
```diff
@@ -238,7 +258,17 @@ export const ReleaseRevertButton = ({
...
-        disabled={disabled || mode === 'disabled'}
+        tooltipProps={{
+          disabled: hasCreatePermission === true,
+          content: tCore('release.action.permission.error'),
+        }}
+        /**
+         * This is an immature assertion of permissions
+         * The permissions needed to revert are:
+         * Permissions to create a request (implemented)
+         * @todo Permissions to create each schema type within the release (not implemented)
```
RitaDias, line 269: Do we have a story for this? :)

packages/sanity/src/core/releases/store/useOrgActiveReleaseCount.ts
RitaDias: Is this something we want to follow up on ? (mainly asking so we don't end up letting it fall through the cracks when we don't want it to)

jordanl17: Ash and I already plan to refactor this approach this week so we'll capture this then 👍

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseRevertButton/ReleaseRevertButton.tsx
jordanl17: Thanks for the reminder, created as SAPP-2434

PR #8692, opened 2025-02-18
fix(core): revise wording for releases schedule actions

pedrobonamin:
### Description

Updates the text in release schedule actions .

Screenshot 2025-02-18 at 15 27 23
Screenshot 2025-02-18 at 15 27 33
Screenshot 2025-02-18 at 15 29 36

### DIFF
--- a/packages/sanity/src/core/releases/i18n/resources.ts
+++ b/packages/sanity/src/core/releases/i18n/resources.ts
@@ -23,13 +23,13 @@ const releasesLocaleStrings = {
   /** Action text for opening a release */
   'action.open': 'Open',
   /** Action text for scheduling a release */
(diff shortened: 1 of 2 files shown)


EoinFalconer (approved):
Looks good to me, @mariuslundgard will revise whether we want the `...` at the end of the `action.schedule` 😊

pedrobonamin:
> Looks good to me, @mariuslundgard will revise whether we want the `...` at the end of the `action.schedule` 😊

Yes, @EoinFalconer  I'm also up to remove that. But it's part of the prototype and I think that it is there because it requires more action from the user to confirm the date? 
Not sure tbh 

Thanks for the review

PR #8648, opened 2025-02-15
feat(typegen): `sanity typegen` Whitespace Handling in Query Output

ChrisNolan:
### **Description**  
This PR improves the whitespace handling in `sanity typegen` by stripping unnecessary leading/trailing whitespace from extracted GROQ queries.  

#### **Why?**  
Currently, the generated TypeScript file (`types.ts`) contains queries with inconsistent whitespace formatting, leading to readability issues and unnecessary diffs. This fix ensures queries are stored in a cleaner, standardized format regardless of how the devs like to format them for readability in their own code.  

#### **Issue Link:** _(If there’s an open issue, link it here; otherwise, remove this line)_  

---

### **What to Review**  
- Changes in `findQueriesInSource.ts` and `expressionResolvers.ts`  
- Ensure that extracted queries are formatted correctly with consistent whitespace  
- Check that query results remain unchanged in meaning (only formatting is adjusted) 
- couple of typos in comments also fixed (can't help it)

---

### **Testing**  
✅ **Unit Tests Updated**: Added new test cases in `findQueriesInSource.test.ts` to verify whitespace stripping behavior.  
✅ **Existing Tests Pass**: All previous test cases still pass, ensuring no regressions.  

---

### **Notes for Release**  
This is an internal tooling improvement to ensure `sanity typegen` produces cleaner TypeScript definitions by removing inconsistent whitespace from extracted GROQ queries.  The first time users run typegen in their project they'll see the extra whitespace roomed, but after that it'll be consistent.

📌 **No breaking changes.**  
📌 **No user action required.**  

---

First time contributor here... let me know how I can improve

PR #8696, opened 2025-02-18
feat(typegen): `sanity typegen` Whitespace Handling in Query Output

sgulseth:
`git revert df7bf5841309e94ae5037a69bd17004d5c14ce3d`

### Description

Unfortunately the change in #8648 breaks the typed query inference. Since it depends on the content of the defineQuery(..) being the same as the generated query.

This reverts commit df7bf5841309e94ae5037a69bd17004d5c14ce3d.

### What to review

N/A

### Testing

N/A

But we should add tests that prevents this from happening again

### Notes for release

Fixes automatic type inference with `@sanity/client`

### DIFF
--- a/packages/@sanity/codegen/src/typescript/__tests__/findQueriesInSource.test.ts
+++ b/packages/@sanity/codegen/src/typescript/__tests__/findQueriesInSource.test.ts
@@ -188,7 +188,7 @@ describe('findQueries with the groq template', () => {
     expect(queries.length).toBe(0)
   })
 
-  test('will ignore declarations if any of the leading comments are ignore tags', () => {
+  test('will ignore declerations if any of the leading comments are ignore tags', () => {
     const source = `
       import { groq } from "groq";
 
@@ -199,23 +199,6 @@ describe('findQueries with the groq template', () => {
 
     const queries = findQueriesInSource(source, __filename, undefined)
     expect(queries.length).toBe(0)
-  })
-
-  test('should remove unnecessary whitespace in queries', () => {
-    const source = `
-      import { groq } from "groq";
-      const messyQuery = groq\`
-        
-        *[_type == "author"]    
-        
-      \`
-      const res = sanity.fetch(messyQuery);
-    `
-
-    const queries = findQueriesInSource(source, 'test.ts')
-    const queryResult = queries[0]
-
-    expect(queryResult?.result).toEqual('*[_type == "author"]')
   })
 })
 
@@ -385,7 +368,7 @@ describe('findQueries with defineQuery', () => {
     expect(queries.length).toBe(0)
   })
 
-  test('will ignore declarations if any of the leading comments are ignore tags', () => {
+  test('will ignore declerations if any of the leading comments are ignore tags', () => {
     const source = `
       import {defineQuery} from "groq";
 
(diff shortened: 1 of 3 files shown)


stipsan (approved):

ChrisNolan:
Doh.  Do you want to give me any details about what broke which wasn't being tested and I can try to continue to stab at it?

sgulseth:
Sorry about that, it's on us for not having proper tests! What happened is the `defineQuery(...)` type inference broke. Since we use the actual groq query as a lookup value, they have to match. When we had `const myQuery = defineQuery('    *[_type ==".."]').` that would generate a map like : `{["    *[_type ==\"..\"]": myQueryResult, ... }`, and then typescript can lookup the return value when  `myQuery` is used inside a `client.fetch` and automatically infer the return type. Since the input query no longer matched the stored key in the map, this lookup failed :/

ChrisNolan:
@sgulseth I was taking a look at this some more (always learning I am) and I could use some pointers.

tried to 'link' my local fork so that in another project folder I could try to have it 'used', but I think because of the 'cli' wrapper that isn't happening?  my pnpm list \@sanity/codegen shows it is taking it from my local version, but when I do the sanity typegen generate I'm not seeing any of my console logs so I think is isn't using it -- likely because how the cli is wrapping around it?  

I thought ok I can run the cli tests -- but they need a `SANITY_CI_CLI_AUTH_TOKEN_STAGING` environment variable and the `.env.example` doesn't explain what it should be?  (Looking at that file as well it looks like to do the full test suite I need access to another project as well... or not? likely beside the point).

Anyway, I want to trigger the error first manually, then write a test which can trigger it, then fix the code but I'm 'stuck' trying to understand how to have the dev version interact.  If you have a moment I'd appreciate some direction.

I also though quickly in the `generateQueryMap` in codegen/src/typescript/typeGenerator.ts` I could just do the same .replace()  before putting it into the `typesByQuerystring` but again... without being able to 'test' it, I'm not confident that'll be the fix. 

Oh, and I figured out my speed issues inside my WSL and it is *much* faster now.  a `pnpm bootstrap` only took 5 minutes this time instead of 40.  Feels much less frustrating ;-)

ChrisNolan:
Was [this commit](https://github.com/sanity-io/sanity/commit/a102118ce223fda955eb66637898da094dddac8b) the test to catch this?

sgulseth:
@ChrisNolan yes, that commit should catch this from happening again as we assert that the query key created in the type map is what we expect. I also noticed that we are doing some trimming when we output the query as a comment in packages/@sanity/cli/src/actions/typegen/generateAction.ts L146

Also sorry for the late reply here, was out sick end of last week :'(

ChrisNolan:
Glad you're feeling better @sgulseth .  I'm still confused about how to actually run the command line tools from the fork against another studio.  Any tips there would be appreciated.  Like in my 'normal' repo, when I do `sanity typegen generate` how do I get it to use the modified forked version?  `pnpm link` or something else?  Even with the tests I just feel more comfortable 'seeing it in action' against real data if you know what I mean.

sgulseth:
`pnpm link` should work yess, but you need to link to "correct" package.

I think linking these two should be enough:
`pnpm link <sanity-repo-path>/packages/@sanity/codegen`
`pnpm link <sanity-repo-path>/packages/@sanity/cli`

And run `pnpm run build && pnpm run watch` in the <sanity-repo-path> folder

PR #8698, opened 2025-02-18
test: add test for typemap

sgulseth:
### Description

Ensures that the generated query typemap matches the expected generated types and query.

### Testing

Well, added 

### Notes for release

N/A


packages/@sanity/cli/test/typegen.test.ts
```diff
@@ -70,6 +70,26 @@ describeCliTest('CLI: `sanity typegen`', () => {
     expect(types.toString()).toMatchSnapshot()
   })
 
+  test('sanity typegen generate: generates query type map', async () => {
+    // Write a prettier config to the output folder, with single quotes. The defeault is double quotes.
```
rexxars, line 74: Not sure I understand this comment? We're not writing any prettier configs here, are we?

stipsan (approved):
Other than what @rexxars already called out it LGTM

packages/@sanity/cli/test/typegen.test.ts
sgulseth: flu barf. thanks! :)

stipsan (approved):

PR #8700, opened 2025-02-19
refactor(core): update managing permissions 2

RitaDias:
### Description

- Updates permissions hook logic to use a store that takes advantage of the resourceCache in the studio
- A lot of file changes are related to updating mocks and making things in tests a bit smoother. Don't be scared by the number of files changed

This will then mean that there is no need to make needless dryRun requests multiple times exponentially reducing the number of calls in the release package.

### What to review

Everything make sense?

### Testing

Updated the tests and mocks, everything should work as expected.

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -121,7 +121,14 @@ export function ReleasesOverview() {
   }, [hasReleases, releasesMetadata, releases])
 
   useEffect(() => {
-    checkWithPermissionGuard(createRelease, DEFAULT_RELEASE).then(setHasCreatePermission)
+    let shouldUpdate = true
+    checkWithPermissionGuard(createRelease, DEFAULT_RELEASE).then((hasPermissions) => {
+      if (shouldUpdate) setHasCreatePermission(hasPermissions)
+    })
+
+    return () => {
+      shouldUpdate = false
+    }
   }, [checkWithPermissionGuard, createRelease])
 
   // switch to open mode if on archived mode and there are no archived releases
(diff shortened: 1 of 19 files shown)


packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -121,7 +121,14 @@ export function ReleasesOverview() {
   }, [hasReleases, releasesMetadata, releases])
 
   useEffect(() => {
-    checkWithPermissionGuard(createRelease, DEFAULT_RELEASE).then(setHasCreatePermission)
+    let shouldUpdate = true
```
jordanl17, line 124: If I understand correct this is so that we don't fire the async called `setHasCreatePermission` if the component is no longer mounted?

If so then my preference would be to track this in a ref, so that we don't lose the closure on `shouldUpdate` over the component lifecycle

```tsx
  const isMounted = useRef(false)
  useEffect(() => {
    isMounted.current = true
    checkWithPermissionGuard(createRelease, <action>()).then(
      (hasPermission) => {
        if (isMounted.current) setHasCreatePermission(hasPermission)
      },
    )

    return () => {
      isMounted.current = false
    }
  }, [checkWithPermissionGuard, createRelease, createReleaseMetadata])
  ```

RitaDias: Ah fair point, I'll make this change!

RitaDias: Should this be applied to the other parts where I've done this? (anywhere that has the useEffect really)

RitaDias: I just went ahead and changed it in all places 👍

jordanl17 (approved):
This looks great, and that all the tests are up to date too 🔥

PR #8702, opened 2025-02-19
fix(core): handling cases where no dependencies are passed to a resourceCache value set

jordanl17:
### Description
In cases where resourceCache is `set` and `get` with no dependencies, eg:
```ts
      resourceCache.get({
        dependencies: [],
        namespace: 'myNamespace',
      })
```
Then the weak map key would be a different reference each time, meaning setting and getting wouldn't return the correct namespace value.

This change essentially maps the case of `dependencies: []` to become a weak map key of `[{}]`, where `{}` is always the same reference (`nullReplacer`).

### Testing
Testing manually by... seeing that it fixed the issue, whilst also regression testing that passing non empty dependencies doesn't break

### Notes for release
N/A - `ResourceCache` is internal only


jordanl17:
@RitaDias once this is merged you can swap the `[null]` with just `[]` for the dependencies to the `resourceCache`
https://github.com/sanity-io/sanity/pull/8700/files#r1961536598

juice49 (approved):
Nice fix. Thanks! 🙌 Would be good to add some more rigorous tests here.

PR #8703, opened 2025-02-19
Feat/dam 272/al source

skogsmaskin:


skogsmaskin:
@snorrees - I have added you as a reviewer here to have a second pair of eyes on the extractSchema stuff.

We are adding a new field on the image type (.media global document reference that will be set when using a asset from the Media Library).

Do you think this will cause any problems related to AI-Assist or Canvas?

packages/@sanity/schema/src/legacy/types/image/fieldDefs.ts
```diff
@@ -13,3 +13,27 @@ export const CROP_FIELD = {
...
+  // to: [
+  //   {
+  //     type: 'sanity.asset',
+  //     preview: {
+  //       select: {
+  //         title: 'title',
+  //       },
+  //     },
+  //   },
+  // ],
```
sgulseth, line 38: let's remove this
```suggestion
```

packages/@sanity/schema/src/legacy/types/file.ts
```diff
@@ -31,7 +38,7 @@ export const FileType = {
   extend(rawSubTypeDef: any, createMemberType: any) {
     const options = {...(rawSubTypeDef.options || DEFAULT_OPTIONS)}
 
-    const fields = [ASSET_FIELD, ...(rawSubTypeDef.fields || [])]
+    const fields = [ASSET_FIELD, MEDIA_LIBRARY_ASSET_FIELD, ...(rawSubTypeDef.fields || [])]
```
sgulseth, line 41: We don't happen to have access to the sanity config here? IWBN if we could only include this field if `mediaLibrary.enabled == true`

packages/sanity/src/core/config/configPropertyReducers.ts
```diff
@@ -395,6 +395,52 @@ export const eventsAPIReducer = (opts: {
...
+    if (typeof resolver === 'string') return resolver
+
+    throw new Error(
+      `Expected \`mediaLibrary.libraryId\` to be a string, but received ${getPrintableType(
+        resolver,
+      )}`,
+    )
+  }, initialValue)
+
+  return result
```
sgulseth, line 441: Do we need to reduce here? 
```suggestion
  for (const {config: innerConfig} of flattenedConfig) {
  	const libraryId = innerConfig.mediaLibrary?.libraryId
  	if (!libraryId && typeof libraryId !== 'string') {
  		continue
  	}
  	if (typeof libraryId === 'string') {
  		return libraryId
  	}
  	
  	 throw new Error(
      `Expected \`mediaLibrary.libraryId\` to be a string, but received ${getPrintableType(
        resolver,
      )}`,
    )
  }

  return initialValue
```

packages/sanity/src/core/config/types.ts
```diff
@@ -376,6 +376,12 @@ export interface PluginOptions {
     components?: StudioComponentsPluginOptions
   }
 
+  /**
+   * Config for the Sanity Media Library asset source integration.
+   * @beta
+   */
+  mediaLibrary?: MediaLibraryConfig
```
sgulseth, line 383: Should this be defined on `WorkspaceOptions` instead?

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
```diff
@@ -0,0 +1,126 @@
...
+        })
+        .request({uri: `/media-libraries`, query: {organizationId}, useGlobalApi: true})
+        .then((result) => {
+          const libraryIdFromResult = result.data[0]?.id
+          if (libraryIdFromResult) {
+            setResolvedLibraryId(libraryIdFromResult)
+          }
+        })
+    })
+  }, [client, libraryIdProp])
```
sgulseth, line 52: This will request media-libraries once each time the asset field is rendered? Should we cache it somewhere?

sgulseth (commented):
My insight into studio code is fairly limited, so I might be mistaken on certain parts

snorrees (commented):
There seems to be no changes to `extractManifest` tests in this PR.

AFAIKT `mediaLibrary: { enabled: true}` leaves no trace in the schema about this feature, so there would be no way for Canvas or consumers of stored schema to know that this studio has this feature enabled.

When syncing from Canvas to Studio, today we upload the image using client.assets.upload. This uses the old in-dataset mechanism.

From what I can tell from this PR, images will fully support both media-library and oldschool in-dataset images, since media lib is now modeled as a source (great imo 👍 )

Now, Canvas will not be integrating with media-lib _today_ but it will probably do so in the short-to-medium term. At that point I think we want to know at the schema level if the target type supports media lib?

So – I propose (this can be a fast follow or something if hard rn):
- when `mediaLibrary: { enabled: true}`, put some sort of marker on image fields produces by `extractManifest`, so when we read manifest schemas, we can know that the field supports medialLib
- this will allow Canvas studio sync to choose media-lib as the target when syncing, instead of using client.asset.upload. We might not even have to sync the image itself, just make the asset-document if it does not exist, since Canvas could have already uploaded the image to media-lib
- i, something like `{type: 'image', ..., supportsMediaLib: true}` or {options: {supportsMediaLib:true}}. This mark could be a pure synthetic thing only being done at extract time.

Does this make sense?

packages/@sanity/schema/src/legacy/types/file.ts
```diff
@@ -11,6 +11,13 @@ export const ASSET_FIELD = {
   to: {type: 'sanity.fileAsset'},
 }
 
+export const MEDIA_LIBRARY_ASSET_FIELD = {
+  name: 'media',
+  type: 'globalDocumentReference',
+  hidden: true,
+  to: [], // Just have to be *something* to be valid
```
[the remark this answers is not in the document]
skogsmaskin: > This seems risky: fields will throw for empty arrays, I'm actually surprised this doesnt, too. Is there a comment in the schema compile to say "we _must_ allow empty `to` because ?"

Changed this to a real value now:
```
to: [{type: 'sanity.asset'}],
```

packages/@sanity/schema/src/legacy/types/file.ts
snorrees: Is there a world where we could pass a flag to Schema.compile ?

skogsmaskin: > You are probably right. I will need a trace in extract tho (if you missed the pr comment)

Working on it @snorrees :)

packages/sanity/src/core/config/types.ts
skogsmaskin: > Should this be defined on `WorkspaceOptions` instead?

No, it should be part of the PluginOptions IMO so that a plugin could know if the mediaLibrary is enabled or not by reading the plugin config it will be passed.

packages/sanity/src/core/config/configPropertyReducers.ts
skogsmaskin: If we are to support these config options throughout the plugin chain, I think we need them.

skogsmaskin: I first thought you meant the config recucing function in general. 

As for the actual `.reduce` I just copied what we do elsewhere. I think it could be getting a function as well, you see how it actually returns something if it's not a string.

packages/sanity/src/core/config/types.ts
bjoerge: I guess it depends on whether you want plugins to provide their own mediaLibrary?

packages/sanity/src/core/config/configPropertyReducers.ts
skogsmaskin: Will have another look at this.

packages/sanity/src/core/config/types.ts
bjoerge: Sounds wise to limit the power plugin has over it yeah, although, I guess that depends on what aspects of the media library is controlled by `MediaLibraryConfig`? (I didn't look too closely at it)

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/hooks/useSanityMediaLibraryConfig.ts
```diff
@@ -0,0 +1,35 @@
...
+
+// TODO: figure out how to configure this stuff
+
+const IS_LOCAL_DEV = false && isDev // Set to true to work against local Media Library dev server
+
+export function useSanityMediaLibraryConfig(): SanityMediaLibraryConfig {
+  const isLocalDev = IS_LOCAL_DEV
+  const client = useClient({apiVersion: DEFAULT_API_VERSION})
+  const host = client.config().apiHost
+  const isStaging = host.includes('sanity.work')
```
juice49, line 14: This is an _extremely_ narrow edge case, but could this check break if a customer is using a custom API domain such as `https://xxxxxxxx.api.sanity.workplace.com`?

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/hooks/useSanityMediaLibraryConfig.ts
```diff
@@ -0,0 +1,35 @@
...
+  const deployedFrontendHost = isStaging ? 'https://media.sanity.work' : 'https://media.sanity.io'
```
juice49, line 15: Very minor: it'd be nice to lift these URLs up to constants.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
juice49: It might make sense to lift this up to a [datastore](https://github.com/sanity-io/sanity/blob/d4c23f75720226356b5f4c7c125d8799370ca97d/packages/sanity/src/core/store/_legacy/datastores.ts) so that requests can be deduplicated, cached, and shared across the UI.

However, I'm not familiar enough with Media Library yet to know whether that's a premature optimisation.

packages/sanity/src/core/config/types.ts
skogsmaskin: Yes, one use case I could think of (with todays limited set of options), is that an organzation may develop a plugin with shared settings across all of their studios. So with the same config for mediaLibrary, content releases etc.

It's also likely that we would support more options under the `mediaLibrary` config in the future, not that should be an argument for building it into Plugins right now though.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
```diff
@@ -0,0 +1,126 @@
...
+    libraryId: libraryIdProp,
+    onClose,
+    onSelect,
+    selectedAssets,
+  } = props
+  const [resolvedLibraryId, setResolvedLibraryId] = useState<string | null>(null)
+
+  useEffect(() => {
+    if (libraryIdProp) {
+      setResolvedLibraryId(libraryIdProp)
```
juice49, line 35: Very small thought: it might be nice to refactor `resolvedLibraryId` to something like `fetchedLibraryId`, and then later create a binding like `const resolvedLibraryId  = libraryIdProp ?? fetchedLibraryId`.

This would prevent the effect immediately setting state when it runs with `libraryIdProp` set.

skogsmaskin: @juice49 - Good catch! I've removed the completely unnecessary state change. 👍

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/hooks/useSanityMediaLibraryConfig.ts
skogsmaskin: Yes, that's true, should rather be `endsWith`.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/hooks/useSanityMediaLibraryConfig.ts
skogsmaskin: Yes, I think that's a good idea, just to have all of this in one place. I've done so now.

packages/sanity/src/core/config/types.ts
```diff
@@ -958,6 +969,22 @@ export type DefaultPluginsWorkspaceOptions = {
   releases: {enabled: boolean}
 }
 
+/**
+ * Config for the Sanity Media Library asset source integration.
+ * @beta
+ */
+export interface MediaLibraryConfig {
+  /**
+   * Weather the Media Library is enabled or not.
```
juice49, line 978:
```suggestion
   * Whether the Media Library is enabled.
```

packages/sanity/src/core/config/types.ts
```diff
@@ -958,6 +969,22 @@ export type DefaultPluginsWorkspaceOptions = {
   releases: {enabled: boolean}
 }
 
+/**
+ * Config for the Sanity Media Library asset source integration.
+ * @beta
+ */
+export interface MediaLibraryConfig {
```
juice49, line 976: Should `enabled` and `libraryConfig` both be optional? This unlocks some nice patterns, such as:

1. Setting `mediaLibrary.enabled` globally and `mediaLibrary.libraryId` on individual workspaces.
2. Setting `mediaLibrary.libraryId` globally and and `mediaLibrary.enabled` on individual workspaces.

These settings do seem to compose together correctly at runtime, but the types prohibit this pattern.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
skogsmaskin: I think a datastore might be a little overkill for this as it's only used in this component (for now at least). I have however added a simple global Map cache (projectId, libraryId).

packages/sanity/src/core/config/types.ts
juice49: I agree with PK; it makes sense to allow plugins to control media library options IMO. Plugins are a valid way to organise a Studio, and I can't see why we would prohibit developers controlling media library settings from a plugin.

> an organzation may develop a plugin with shared settings across all of their studios

That's a great example of where it would be annoying to prohibit this.

The workspace config takes precedence, so developers can always override the `mediaLibrary` config set in a plugin if they wish.

packages/sanity/src/core/config/configPropertyReducers.ts
juice49: I agree with PK's initial approach here (using a reducer), purely from the standpoint of maintaining consistency with how Studio resolved other configs.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
```diff
@@ -0,0 +1,127 @@
...
+          if (libraryIdFromResult) {
+            setFetchedLibraryId(libraryIdFromResult)
+          }
+        })
+    })
+  }, [client, libraryIdProp])
+
+  const resolvedLibraryId = libraryIdProp || fetchedLibraryId
+
+  const handleClose = useCallback(() => {
```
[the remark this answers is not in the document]
skogsmaskin: Nope, we can use the prop `onClose` directly.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/SelectAssetsDialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const handleClose = useCallback(() => {
+    onClose()
+  }, [onClose])
+
+  const handleSelect = useCallback(async () => {
+    try {
+      setDidSelect(true)
+      await onSelect(assetSelection)
+    } catch (error) {
```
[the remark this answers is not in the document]
skogsmaskin: Yes 👍  Added a console.error for it.

packages/sanity/src/core/config/types.ts
skogsmaskin: I leave it as a plugin config then. There's more pros to it than cons as made clear by the discussion.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/constants.ts
```diff
@@ -0,0 +1,18 @@
+import {isDev} from '../../../environment'
+
+// TODO: change this to what the official GDR release version is
+export const DEFAULT_API_VERSION = 'vX'
+
+// Staging
+export const DEPLOYED_FRONTEND_HOST_STAGING = 'https://media.sanity.work'
+export const CDN_HOST_STAGING = 'https://sanity-cdn.work'
```
juice49, line 8: Nice 👍

packages/sanity/src/core/config/configPropertyReducers.ts
skogsmaskin: Yes, there's a lot of these around, so I think it's better to just follow the same pattern as we usually do in this code.

packages/sanity/src/core/i18n/bundles/studio.ts
```diff
@@ -100,6 +107,10 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'asset-source.image.asset-list.delete-successful': 'Image was deleted',
   /** Text displayed on button or menu invoking the image asset source */
   'asset-source.image.title': 'Uploaded images',
+  /** Menu Items for Media Library */
+  'asset-source.media-library.image.title': 'Media Library',
+  // eslint-disable-next-line sort-keys
```
juice49, line 112: Just wondering: is this definitely needed?

skogsmaskin: It's for the menu items here:

image

I think it is? What's you thinking?

juice49 (dismissed):
Nice, PK! This looks great to me. There are maybe a couple of things we should follow up on, but I don't think they block us merging this work (especially as the Media Library integration is opt-in for now).

### Error when attempting to create version of document containing a Media Library asset

The request failed with:

```
{
    "error": {
        "description": "transaction failed: strong references are not permitted for attached target resource media-library:<redacted>",
        "items": [
            {
                "error": {
                    "description": "strong references are not permitted for attached target resource media-library:<redacted>",
                    "invalidIntegrity": true,
                    "targetResourceID": "<redacted>",
                    "targetResourceType": "media-library",
                    "type": "globalDocumentReferenceNotAllowedError"
                },
                "index": 0
            }
        ],
        "type": "mutationError"
    }
}
```

### Media libraries are requested every time picker is opened (if library id is not statically defined)

I don't see this being a huge issue, but it might be nice to cache/share this request in the future.

skogsmaskin:
Cool, thank you for the thorough review @juice49! Love it ❤️ 

Regarding the cache, I'm not sure if you saw it, but I added a simple cache for it so it will only do it once per project id if the libraryId is unspecified.

Regarding the "Error when attempting to create version of document containing a Media Library asset" it's a bit surprising, and I'll have a look as soon as possible.

sgulseth:
@juice49  It looks like the new version is setting `_weak: false`? 🤔 Not sure if it's a bug in the schema or if it's elsewhere

skogsmaskin:
> There seems to be no changes to `extractManifest` tests in this PR.

Wasn't initially, but I've added them since.

> AFAIKT `mediaLibrary: { enabled: true}` leaves no trace in the schema about this feature, so there would be no way for Canvas or consumers of stored schema to know that this studio has this feature enabled.
> 
> When syncing from Canvas to Studio, today we upload the image using client.assets.upload. This uses the old in-dataset mechanism.
> 
> From what I can tell from this PR, images will fully support both media-library and oldschool in-dataset images, since media lib is now modeled as a source (great imo 👍 )

Yes, assets added from the Media Library should behave just like other assets in terms of existing infrastructure. The only difference with concern to the data, is that the image or file object will have an additional global document reference to the Media Library (which is used for usage tracking) if it is selected from there. Otherwise it's exactly the same (will have a local asset document in the dataset just like any other asset). Works exactly the same in terms of the image pipeline.
One special thing here though, is that the file instance is shared across all the usage of that asset, but in the eyes of the consumer, it will behave just like any other asset.

> Now, Canvas will not be integrating with media-lib _today_ but it will probably do so in the short-to-medium term. At that point I think we want to know at the schema level if the target type supports media lib?

Ideally we should try to settle this right away.

> 
> So – I propose (this can be a fast follow or something if hard rn):
> 
> * when `mediaLibrary: { enabled: true}`, put some sort of marker on image fields produces by `extractManifest`, so when we read manifest schemas, we can know that the field supports medialLib
> * this will allow Canvas studio sync to choose media-lib as the target when syncing, instead of using client.asset.upload. We might not even have to sync the image itself, just make the asset-document if it does not exist, since Canvas could have already uploaded the image to media-lib
> * i, something like `{type: 'image', ..., supportsMediaLib: true}` or {options: {supportsMediaLib:true}}. This mark could be a pure synthetic thing only being done at extract time.
> 
> Does this make sense?

It does make sense. I'l make story for it so we can follow up on this right away. I read from you answer however, that this isn't blocking merging this PR though.

sgulseth (dismissed):

packages/sanity/src/core/i18n/bundles/studio.ts
skogsmaskin: Ah right! No, that was something left behind from before  they didn't group together. Fixed.

sgulseth (approved):
🤌

juice49 (approved):

PR #8710, opened 2025-02-20
refactor(core): update so that there is an horizontal scroll on the title

RitaDias:
### Description

Before
image

After

https://github.com/user-attachments/assets/e2015b06-6783-44e4-ab1d-03b543fb5948

### What to review

Does it make sense, anything look odd?

### Testing

Not needed

### Notes for release

Improvements to document header when there are many versions of a document


juice49 (approved):
Nice 🙌.

PR #8711, opened 2025-02-20
fix(sanity): discard draft permission check

juice49:
### Description

When conducting a discard permission check for a draft document, the `version` is erroneously set to a draft document id.

This manifests as a crash when opening the context menu of the "draft" version chip in the document editor (typically by right-clicking it).

This branch fixes the bug by ensuring no `version` value is provided for drafts. `version` is now only provided for version documents, and always reflects the release id.

### What to review

The version chips in the document editor. In particular, the "draft" version chip.

### Testing

 - Can open context menu for all version chips
   - Published
   - Draft
   - Versions


packages/sanity/src/core/releases/components/documentHeader/contextMenu/VersionContextMenu.tsx
```diff
@@ -88,7 +90,7 @@ export const VersionContextMenu = memo(function VersionContextMenu(props: {
         {isVersion && (
           <IntentLink
             intent="release"
-            params={{id: releaseId}}
+            params={{id: fromRelease}}
```
juice49, line 93: This logic is unchanged. The component is only rendered when `isVersion` is truthy, and `releaseId` was always bound to `fromRelease` in this scenario.

bjoerge (approved):

PR #8715, opened 2025-02-20
fix(sanity): add missing `useInsertionEffect` hook dependency `isCreateLinked`

juice49:
### Description

This `useInsertionEffect` hook was recently updated, but the new dependency was not added.

This mistake wasn't caught by our linter. We should investigate why.

### What to review

The new hook dependency.

### Testing

The document editor renders correctly.


pedrobonamin (approved):
Thnk you @juice49 !

PR #8716, opened 2025-02-21
Remove --legacy-peer-deps from new project initialization

ChrisLaRocque:
Remove --legacy-peer-deps

### Description

The deps in our starting templates don't throw an error requiring `--legacy-peer-deps`, so this PR removes the `--legacy-peer-deps` flag from the install commands for new Sanity projects.


ChrisLaRocque:
Never mind this command is used for installing

PR #8723, opened 2025-02-22
fix(cli): require projects to be part of an organization

rexxars:
### Description

From now on we want all projects to be part of an organization. This means the CLI flow should be a little different - no "None" option for selecting an organization. Instead, we want people to choose one or create one.

This PR modifies the flow so that:

* If a user is not part of any organizations, prompt them to create one
* It shows organizations you are a member of but do not have the sufficient grants to (helps people understand why they can't choose it)
* Removes the "none" option, replaces it with a "Create new organization" option

There are a few code paths (through flags/unattended mode) that may still send a project creation request without an organization - that's fine - the API will assign those projects to the users' organization (if they have one), otherwise creates a new one and assigns the project to it.

Also found and fixed a minor flow issue for new users, where there was no default value for project name, and accepting a blank name would lead to an unhelpful error. 

### What to review

- Flow still works, organization assignment works as expected

### Notes for release

None

### DIFF
--- a/packages/@sanity/cli/package.json
+++ b/packages/@sanity/cli/package.json
@@ -115,7 +115,7 @@
     "minimist": "^1.2.5",
     "open": "^8.4.0",
     "ora": "^8.0.1",
-    "p-filter": "^2.1.0",
+    "p-map": "^4.0.0",
     "p-timeout": "^4.0.0",
     "preferred-pm": "^3.0.3",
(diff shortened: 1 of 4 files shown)


cngonzalez (approved):
Seems reasonable on reading through. Will go ahead and approve for now and test locally as well -- will ping if anything seems awry.

Issue #8724, opened 2025-02-22
slug defaultIsUnique failing when content releases do not exist - allowing duplicate slugs in live systems


ChrisNolan:
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

Slugs are being allowed to be non-unique by default now, due to the new 'versions' system (PR #8454 ).  I feel this is serious as any system which has had invalid slugs introduced since the change will have to retroactively go back and update their data.  (Sorry couldn't find which of the commits in that PR did the deed).

I believe the issue is this line added to the constraints in the slugValidator to check if the document id exists in the content releases... but if you have no content releases we have a problem.
`    `!(_id in path("${VERSION_FOLDER}.**.${published}"))`,`

**To Reproduce**

update a document with a basic slug field to use a slug field that is already in use on another document with content releases not enabled.  No validation error is given.

Note, using the browser tools I could see for example the query being made on the validator is
```
!defined(*[
  _type == "contributor" &&
  !(_id in ["drafts.chris-nolan-1978", "chris-nolan-1978"]) &&
  !(_id in path("versions.**.chris-nolan-1978")) &&
  slug.current == "chris-nolan"
][0]._id)
```

and this returns 'true'.  Where the query before the change from the 'content releases' 
```
!defined(*[
  _type == "contributor" &&
  !(_id in ["drafts.chris-nolan-1978", "chris-nolan-1978"]) &&
  slug.current == "chris-nolan"
][0]._id)
```
returns false -- as there is another 'contributor' with the slug of 'chris-nolan'.

So the issue is that "all of them" have to be true to work... and if we have no versions at all, it is failing.  The constraints in slugValidator should be something like

```
const constraints = [
  '_type == $docType',
  `!( 
    _id == $published || 
    _id == $draft || 
    _id in path("${VERSION_FOLDER}.**.${published}")
  )`,
  `${atPath} == $slug`,
].join(' && ');

```

instead.  Good ol logical operators running afoul.

Likely needs to be tested with content releases enabled, and content releases not enabled.  Maybe the other fix could be to only add that element to the constraints array if content releases are enabled?

**Expected behavior**

A validation error should be given.

Also, why are the tests in `infer.ts` still passing...?  Ah... because they were changed to expect a mock of a query that matches this form, they aren't actually doing the query and seeing the failed results.

**Screenshots**

**Which versions of Sanity are you using?**

[...]

PR #8725, opened 2025-02-22
fix(WIP): slugValidator constraint updated to be logically consistent with content releases

ChrisNolan:
### Description

A quick attempt at dealing with #8724 
I'm not setup with a full test config etc for the studio so will need somebody's help in finishing this one.

### What to review

Are slugs correctly validated as unique with and without content releases enabled?

### Testing

The test only checks if the generated query matches the query constraints... they don't actually test if the query itself will be giving valid results against any fixtures etc which would be ideal.

### DIFF
--- a/packages/sanity/src/core/validation/validators/slugValidator.ts
+++ b/packages/sanity/src/core/validation/validators/slugValidator.ts
@@ -54,8 +54,11 @@ const defaultIsUnique: SlugIsUniqueValidator = (slug, context) => {
 
   const constraints = [
     '_type == $docType',
-    `!(_id in [$draft, $published])`,
-    `!(_id in path("${VERSION_FOLDER}.**.${published}"))`,
+    `!( 
+      _id == $published || 
+      _id == $draft || 
+      _id in path("${VERSION_FOLDER}.**.${published}")
+    )`,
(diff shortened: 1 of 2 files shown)


ChrisNolan:
Hi @cngonzalez thanks for looking at it.  I switched it to a 'draft' as I mentioned in the 'issue' I do not know how to run a 'test studio' etc, the last time I tried I  spent hours spinning my wheels and it seems like for such a simple fix I'd love a 'collaborator' who could dive in and make sure things work.  It felt like a pretty high priority fix as well since live studios out there are potentially getting non-unique ids without realizing it and I didn't want my lack of familiarity with the system slowing things down.

There were also issues with the tests. I don't know how to write tests that work against a 'fixture' for the queries -- like the query is tested that it looks like it should look, but it would be super if the test could be done against actual fixture data and know that the query is returning what is expected for it to be returned, thus ensuring the uniqueness check.

Also, because this one slipped through in content releases I was worried that similar places were logic was added to look for the versions might have similar logical faults and thought maybe they should be investigated at the same time?

ChrisNolan:
Closing this PR as @pedrobonamin has fixed this issue in PR #8951

PR #8736, opened 2025-02-24
fix: generating new release ID each time defaults are used; only checking update release perms when release is active

jordanl17:
### Description
Resolves 2 issues:
1. Creating multiple IDs in a single window session would cause failures since all the created releases would try to use the same ID
2. We were wrongly checking release update permissions on archived or published releases (where it is not possible to update)

### What to review
* Makes ok sense to have the `getReleaseDefaults` method?

### Testing
Regression tests; verified the issues were also resolved manually

### Notes for release
N/A


juice49 (approved):
Nice fix! Looks good to me.

PR #8751, opened 2025-02-24
chore(ci): auto cancel workflows on PRs

stipsan:
### Description

We have a lot of congestion and rebasing happening on PRs. Especially with renovatebot which auto rebases when a merge conflict is detected.
When this happens it's not necessary to wait for previous builds to complete, as we use GitHub required checks to control what PRs can merge anyway, and they only care about the checks on the latest commit on the branch.

### What to review

Does it make sense? The way `group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}` works is that a matching name will be cancelled if a new build starts. Thus `group` should be unique to each workflow file (that's what `github.workflow` ensures).
It should have a stable name for the lifecycle of a PR, `github.head_ref` does that by using the branch name. `head_ref` is only available on PRs, so `${{ github.head_ref || github.run_id }}` will resolve `github.run_id` when a workflow is run due to pushing to `next`. `github.run_id` has a new id on every run, this is how we avoid checks from cancelling on commits merged to `next`, as it's useful for bisecting purposes to have test suites complete on each commit landing there.

IOW on this PR the `cli-test.yml` workflow will have the group `CLI Unit tests-cancel-ci-jobs`. A push before the job completes will cancel it, as it did [here](https://github.com/sanity-io/sanity/actions/runs/13503250096?pr=8751):
```bash
[CLI Tests (ubuntu-latest / node 18)](https://github.com/sanity-io/sanity/actions/runs/13503250096/job/37726704137)
Canceling since a higher priority waiting request for 'CLI Unit tests-cancel-ci-jobs' exists
```

### Testing

The `concurrrency` feature is used on a lot of other repos already so it's a well-tested technique to reduce GitHub worker queues.

### Notes for release

N/A

### DIFF
--- a/.github/workflows/are-we-compiled-yet.yml
+++ b/.github/workflows/are-we-compiled-yet.yml
@@ -7,6 +7,10 @@ on:
   push:
     branches: [next]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -7,6 +7,10 @@ on:
   push:
     branches: [next]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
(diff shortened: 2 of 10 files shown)


RitaDias (approved):
Thanks for doing this Cody!

PR #8757, opened 2025-02-24
fix: use raw perspective by default for search

bjoerge:
### Description
This fixes an issue that caused drafts (and versions) not to appear in reference search by explicitly 

### What to review
- Does the fix make sense?

### Testing
- Create a draft document (without publishing it)
- Go to a reference field that can reference its type and search for it.

Prior to this PR, the draft document would not appear in search results
After this PR the draft should appear and you should be able to create a reference to it

### Notes for release
- Fixes an issue causing draft-only documents to not appear in reference search


ricokahler (dismissed):
i think this looks right!

EoinFalconer (dismissed):

ricokahler (approved):

PR #8759, opened 2025-02-25
fix(structure): navigate to version when document only has versions

RitaDias:
### Description

> [!NOTE]  
> Marius reviewed and approved of this behavior

- Open the version of a document when that version is the only one (no draft, no published)
- In instances where multiple versions exist (but no draft and no published), then it will default to the first version
- The document itself will be disabled if you are accessing a version that you wouldn't otherwise (for example, a single version if the document has no draft)

When in a draft / publish pinned version, no draft, no publish, only versions

- Show banner to edit the version
- Read only displayed version

image

When "go to release to edit" in banner, pinned version is updated, displayed document can be shown

image

When in a draft / publish pinned version, when multiple versions in a document, the first one should be displayed
image

### What to review

Does it make sense? Is there a better way to go about it?

### Testing

- Updated tests.
- Added unit tests to new functionality added
- Added new e2e tests to make sure that the displayed document remains as expected

> [!WARNING]  
> Moved e2e tests to separate PR, they are causing issues and preventing merging. Need to figure out why since it is not the results of the tests themselves.

### Notes for release

When a document only has a version of itself, it will open that version instead of the draft


packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
```diff
@@ -153,15 +153,16 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
...
       return <ScheduledReleaseBanner currentRelease={selectedPerspective as ReleaseDocument} />
     }
+
     if (
       displayed?._id &&
       getVersionFromId(displayed._id) !== selectedReleaseId &&
       ready &&
-      !isCreatingDocument
+      !isCreatingDocument &&
+      typeof selectedReleaseId !== 'undefined'
```
RitaDias, line 165: Fixed issue when the banner was showing unintentionally

RitaDias:
@jordanl17 moved to draft again after some notes from marius 👍

packages/sanity/test/testUtils/TestProvider.tsx
```diff
@@ -50,30 +50,32 @@ export async function createTestProvider({
...
-                            {children}
-                          </PerspectiveContext.Provider>
-                        </AddonDatasetContext.Provider>
-                      </ResourceCacheProvider>
-                    </CopyPasteProvider>
-                  </SourceProvider>
-                </WorkspaceProvider>
-              </LayerProvider>
-            </ToastProvider>
+            <ResourceCacheProvider>
```
RitaDias, line 53: Needed for the hooks tests (for releases)

packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -135,7 +142,35 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
...
+        )
+      : undefined
+
+  const selectedStateId = useMemo(() => {
+    // if a document version exists with the selected release id, then it should use that
+    if (documentVersions.some((id) => getVersionFromId(id) === selectedReleaseId)) {
+      return selectedReleaseId
+    }
+
+    // check if the selected version is the only version, if it isn't and it doesn't exist in hte release
```
juice49, line 163: Tiny typo here: "hte".

packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -230,6 +270,18 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
     const reconnecting = connectionState === 'reconnecting'
     const isLocked = editState.transactionSyncLock?.enabled
 
+    // in cases where the document has no draft or published, but has a version,
+    // and that version doesn't match current pinned version
+    // we disable editing
+    if (
```
juice49, line 276: Wondering whether it's possible to reuse `onlyHasVersions` here?

```diff
- editState.version &&
- !editState.draft &&
- !editState.published &&
+ onlyHasVersions &&
selectedPerspectiveName !== getVersionFromId(editState.version._id)
```

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -136,6 +136,14 @@ const structureLocaleStrings = defineLocalesResources('structure', {
     'This reference has been removed since you opened it.',
   /** The text that appears for the action button to add the current document to the global release */
   'banners.release.action.add-to-release': 'Add to release',
+  /** The text that appears for the action button to add the current document to the global release */
+  'banners.release.action.open-to-edit': 'Open release to edit',
```
juice49, line 140: Just thinking out loud, as I know this functionality has been approved, but it stood out to me that we use the verb "open" here where we use "pin" in other places.

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
```diff
@@ -0,0 +1,108 @@
...
+  )
+  const {data: documentVersions} = useDocumentVersions({documentId})
+
+  const documentVersionsTitleList = useMemo(
+    () =>
+      activeReleases
+        .filter((version) => {
+          return documentVersions.find((release) => {
+            const r = getVersionFromId(release) ?? ''
+            return version._id.includes(r)
```
juice49, line 67: Could this result in false-positives? If the release id `xyz` appears in any position in the document id, a document could incorrectly match here. Might be more of a problem for programmatically generated releases where the author uses a common string like `"book"`.

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
```diff
@@ -0,0 +1,108 @@
...
+          })
+        })
+        .map((version) => version.metadata.title),
+    [activeReleases, documentVersions],
+  )
+  const tone = currentVersion && getReleaseTone(currentVersion)
+  const {t} = useTranslation(structureLocaleNamespace)
+
+  const handleGoToEdit = useCallback(async () => {
+    setPerspective(releaseId as ReleaseId)
```
juice49, line 77: Is there a way to do this without asserting the value is `ReleaseId`? As it's deprecated in favour of `string`.

test/e2e/tests/releases/DocumentDisplayed.spec.ts
```diff
@@ -0,0 +1,288 @@
+import {expect} from '@playwright/test'
+import {test} from '@sanity/test'
```
juice49, line 2: These tests are a great addition! ✨

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
RitaDias: I tried and it was screaming due to types :/ 
I can give it another whirl and have a look. 👍

packages/sanity/src/structure/i18n/resources.ts
RitaDias: That's a good point.

cc @mariusGundersen do you mind changing this from Open to Pin?

RitaDias: That's a good point

cc @mariuslundgard do you mind changing this from Open to Pin?

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
RitaDias: that's true 🤔 
Let me rework this part

packages/sanity/src/core/form/useDocumentForm.ts
RitaDias: https://github.com/sanity-io/sanity/pull/8759/commits/fa5f2b43c6d8bd8145c22f8ebca578c96e889b0d

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
RitaDias: https://github.com/sanity-io/sanity/pull/8759/commits/c8116789487210c49497a3e68ded737f56d08410

packages/sanity/src/structure/i18n/resources.ts
RitaDias: Double checked with Marius and quoting:

> Ah, in this case it means “Open document in release to edit”. Which is a bit verbose. Maybe “Open in release to edit” is better? Yeah, that’s better

So I've made the chance to the new suggestion :)

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
```diff
@@ -0,0 +1,61 @@
...
+  const {data: releases} = useActiveReleases()
+  const {data: documentVersions} = useDocumentVersions({documentId})
+
+  const sortedDocumentList = releases
+    .filter(({_id}) => {
+      return documentVersions.some(
+        (id) => getVersionFromId(id) === getReleaseIdFromReleaseDocumentId(_id),
+      )
+    })
+    .sort((a, b) => {
```
juice49, line 39: Might be a silly question: what's the reason for sorting here? [`useActiveReleases` already sorts releases using `sortReleases` prior to returning them](https://github.com/sanity-io/sanity/blob/9406243c172c7be0e35c85842417c0da1d4a4218/packages/sanity/src/core/releases/store/useActiveReleases.ts#L29).

I'm guessing you need some other sort order and I'm just missing the point, but I thought I'd check 🙂.

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Fetches the document versions for a given document and sorts them by release type
+ * Returns a boolean if the document has only versions
+ *
+ * @param documentId - document id related to the document version list
+ * @returns object with sortedDocumentList and if the document has only versions
+ *
+ * @beta
+ */
+export const useDocumentVersionSortedList = ({
```
juice49, line 25: I recently needed to do something similar (I think), and [added a hook named `useDocumentIdStack`](https://github.com/sanity-io/sanity/blob/9406243c172c7be0e35c85842417c0da1d4a4218/packages/sanity/src/structure/hooks/useDocumentIdStack.ts#L36). Just wondering whether there is any overlap here?

I think I am probably just not fully grokking this code!

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
```diff
@@ -0,0 +1,124 @@
...
+}): React.JSX.Element | null {
+  const {onlyHasVersions} = useDocumentVersionSortedList({documentId})
+
+  /** this banner should only be rendered in specific cases
+   * 1. when the document is not a draft or published
+   * 2. when the document only has version or versions
+   * 3. when the global perspective is not set (published or draft)
+   */
+
+  if (isDraftId(documentId) || isPublishedId(documentId)) {
```
juice49, line 37: Just wondering: can this be simplified to `!isVersionId(documentId)`?

```suggestion
  if (!isVersionId(documentId)) {
```

packages/sanity/src/structure/panes/document/documentPanel/banners/__tests__/OpenReleaseToEditBanner.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    const wrapper = await createTestProvider({resources: [structureUsEnglishLocaleBundle]})
+    render(<OpenReleaseToEditBanner documentId={'test'} isPinnedDraftOrPublished />, {
+      wrapper,
+    })
+
+    expect(screen.queryByTestId('open-release-to-edit-banner')).toBeNull()
+  })
+
+  // @jordan, here's the trouble child (the latest)
```
juice49, line 113: I love Jordan but wondering if we should remove this?

juice49 (commented):
This is looking great so far 🙌. Let a couple of comments: mostly questions around things that weren't completely clear to me.

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
RitaDias: Probably can, I will try it out and make sure that it works as intended

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
RitaDias: The thing is that this returns documents, not ids :) 
This also sorts it by types (which is the answer to one of the other questions you have asked). I don't think it is overlapped for this particular case

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
RitaDias: Not really silly, I should probably have added a comment.
The issue here is that we needed it sorted by type (asap, schedule and then _finally_ undecided). That is what this sort does.
Probably should change the title to be `useDocumentVersionTypeSortedList` or something.

Need to add a comment here too

packages/sanity/src/structure/panes/document/documentPanel/banners/__tests__/OpenReleaseToEditBanner.test.tsx
RitaDias: 🫥

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
RitaDias: Done!

RitaDias:
I will be adding the e2e tests in a separate PR (https://github.com/sanity-io/sanity/pull/8828) but I have been running them against this code locally to make sure that any of the changes applied don't break behaviour.

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
juice49: Ah I see! Is that essentially the reverse of the order provided by `sortReleases`? `sortReleases` also factors `publishAt` and `metadata.intendedPublishAt` into the ordering. Is that relevant here?

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
juice49: Cool. Thanks for explaining 🙂.

packages/sanity/src/core/releases/hooks/useDocumentVersionSortedList.ts
RitaDias: It is indeed essentially the reverse, I'm going to try to see if I can reuse the sortReleases into this as well

RitaDias: Mentioned this on a message but I couldn't use it because the original use case doesnt match here. I still used part of the logic though! thanks for bringing this up!

packages/sanity/src/core/releases/hooks/useDocumentVersionTypeSortedList.ts
```diff
@@ -0,0 +1,82 @@
...
+    .filter(({_id}) => {
+      return documentVersions.some(
+        (id) => getVersionFromId(id) === getReleaseIdFromReleaseDocumentId(_id),
+      )
+    })
+    /**  sort the document lists that exist for a specific document Id based on the order of the release types
+     * it will first put the "asap" releases then "scheduled" and finally "undecided"
+     * orderedReleaseTypes is the order of the release types
+     * */
+    .sort((releasesList, compareReleasesList) => {
```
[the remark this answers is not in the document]
RitaDias: I don't think it's _Exactly_ the same. I think it's similar but the GlobalPerspectiveMenu requires a lot more information. I haven't looked at it properly for a while but last I checked the data information it required to set up things was more than this simple sorting.

It feels it could be a potential rabbit hole

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
```diff
@@ -0,0 +1,123 @@
...
+}): React.JSX.Element | null {
+  const {onlyHasVersions} = useDocumentVersionTypeSortedList({documentId})
+
+  /** this banner should only be rendered in specific cases
+   * 1. when the document is not a draft or published
+   * 2. when the document only has version or versions
+   * 3. when the global perspective is not set (published or draft)
+   */
+
+  if (!isVersionId(documentId)) {
```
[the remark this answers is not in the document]
RitaDias: Ah this doesn't work in two instances (thank the tests)
If we make this change it will break in specific cases:
- when the selectedPerspective is not draft or published (which can happen), and the version shown is the same as the version that is pinned - this banner shouldn't appear (this is just normal use case)
image

- it will also fail when you try to tackle the initial use case: doesn't have a draft, doesn't have published and doesn't have any pinned release (the banner should appear)
image

I think the condition needs to stay as is.

packages/sanity/src/structure/panes/document/documentPanel/header/perspective/DocumentPerspectiveList.tsx
```diff
@@ -83,6 +84,9 @@ export const DocumentPerspectiveList = memo(function DocumentPerspectiveList() {
   const {editState, displayed, documentType, documentId} = useDocumentPane()
   const isCreatingDocument = displayed && !displayed._createdAt
   const filteredReleases = useFilteredReleases({displayed, documentId})
+  const {data: documentVersions} = useDocumentVersions({documentId})
+
+  const onlyHasVersions = !editState?.published && !editState?.draft && documentVersions?.length > 0
```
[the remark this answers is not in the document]
RitaDias: Looking at it again it does feel like we should the onlyHasVersions to its own hook, I'll do that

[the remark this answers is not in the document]
RitaDias: This ended up being far larger of a change than I expected https://github.com/sanity-io/sanity/pull/8759/commits/6002d194c97bc6c20a7d98de751a2021bb1a4bca

packages/sanity/src/core/releases/hooks/useDocumentVersionTypeSortedList.ts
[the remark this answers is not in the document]
RitaDias: Briefly looking at this, kinda.
It is kinda the same thing and you could abstract the logic but the major difference between GlobalPerspetiveMenu & overview list is that those places require you to have a view of all releases.
This particular case requires you to see, only, the releases that exist in that document (so fetching different things). The method could probably be extracted however, I'm unsure of what other places and this PR is already big as it is. The other places use their own unique logic too so it would mean I would need to to each place, update tests and make sure everything still works as intended.

Would you be ok with creating a story for this for a larger refactor? I would prefer to not bloat this PR 🙏

packages/sanity/src/structure/panes/document/documentPanel/banners/OpenReleaseToEditBanner.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const handleGoToEdit = useCallback(async () => {
+    setPerspective(releaseId)
+  }, [releaseId, setPerspective])
+
+  return (
+    <Banner
+      tone={tone}
+      paddingY={0}
+      data-testId="open-release-to-edit-banner"
```
juice49, line 84: I think this should be `data-testid`. React is logging an error to the browser console when it renders:

```
React does not recognize the `data-testId` prop on a DOM element
```

```suggestion
      data-testid="open-release-to-edit-banner"
```

RitaDias: good catch!

packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -141,7 +149,36 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
...
+  const {onlyHasVersions} = useOnlyHasVersions({documentId})
+  const firstVersion =
+    sortedDocumentList.length > 0
+      ? documentVersions.find(
+          (id) =>
+            getVersionFromId(id) === getReleaseIdFromReleaseDocumentId(sortedDocumentList[0]._id),
+        )
+      : undefined
+
+  const activeDocumentReleaseId = useMemo(() => {
```
juice49, line 165: The effect of this seems to be that:

When viewing a document—and the document does not exist in the selected perspective—the document editor form displays the contents of the first existing version in read-only state.

I'm wondering whether the form should instead be empty (and read-only)?

RitaDias: Sent a private message, this is likely easier discussed on a call!

packages/sanity/src/core/releases/hooks/useOnlyHasVersions.tsx
```diff
@@ -0,0 +1,35 @@
...
+ */
+export const useOnlyHasVersions = ({documentId}: {documentId: string}): useOnlyHasVersionsState => {
+  const {data: documentVersions} = useDocumentVersions({documentId})
+
+  const onlyHasVersions =
+    documentVersions &&
+    documentVersions.length > 0 &&
+    !documentVersions.some((version) => !isVersionId(version))
+
+  return useMemo(
```
juice49, line 26: Just wondering: would it be simpler to have the hook return a boolean directly, returning the need for memoising an object?

RitaDias: That makes sense! I'll make the change

packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
```diff
@@ -150,18 +151,19 @@ export const DocumentPanel = function DocumentPanel(props: DocumentPanelProps) {
...
     }
-    const isCreatingDocument = displayed && !displayed._createdAt
     const isScheduledRelease =
       isReleaseDocument(selectedPerspective) && isReleaseScheduledOrScheduling(selectedPerspective)
-
     if (isScheduledRelease) {
       return <ScheduledReleaseBanner currentRelease={selectedPerspective as ReleaseDocument} />
     }
+    const isPinnedDraftOrPublish: boolean =
+      selectedPerspective === 'drafts' || selectedPerspective === 'published'
```
juice49, line 160: Just an FYI 🙂: I added [a helper](https://github.com/sanity-io/sanity/blob/42d4242607ffe2594415c6e5bd826c0c3acc273e/packages/sanity/src/core/util/draftUtils.ts#L116) recently to abstract this check.

You could do something like:

```
const isPinnedDraftOrPublish = isSystemBundle(selectedPerspective)
```

RitaDias: thanks for the tip! made the change should be updated now!

juice49 (approved):
This is a great improvement! Thanks, Rita.

PR #8775, opened 2025-02-26
fix(sanity): allow discovery of all document versions using groq2024 search

juice49:
### Description

The `groq2024` search strategy currently relies on Content Lake perspectives to collate documents. This means global search is unable to discover document versions outside of the selected perspective.

This branch changes back to the `raw` perspective and returns to client-side collation in order to make all document versions discoverable via global search, regardless of the selected perspective.

### What to review

Global search: discoverability of versions outside the selected perspective.
Document lists: should be unchanged.

### Testing

Updated unit tests and tested manually.

### DIFF
--- a/packages/sanity/src/structure/panes/documentList/helpers.ts
+++ b/packages/sanity/src/structure/panes/documentList/helpers.ts
@@ -10,7 +10,7 @@ import {
 } from '@sanity/types'
 import * as PathUtils from '@sanity/util/paths'
 import {type ExprNode, parse} from 'groq-js'
-import {collate, getPublishedId, isVersionId} from 'sanity'
+import {collate, getPublishedId} from 'sanity'
 
 import {type DocumentListPaneItem, type SortOrder} from './types'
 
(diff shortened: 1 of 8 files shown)


packages/sanity/src/structure/panes/documentList/helpers.ts
```diff
@@ -20,15 +20,13 @@ export function getDocumentKey(value: DocumentListPaneItem, index: number): stri
 
 export function removePublishedWithDrafts(documents: SanityDocumentLike[]): DocumentListPaneItem[] {
   return collate(documents).map((entry) => {
-    const doc = entry.draft || entry.published
-    const isVersion = doc?.id && isVersionId(doc._id)
+    const doc = entry.draft || entry.published || entry.versions[0]
```
bjoerge, line 23: we could perhaps consider adding a preference for what version to chose, but probably better to defer that to when we need it

bjoerge (approved):
Looks great, thanks @juice49

packages/sanity/src/structure/panes/documentList/helpers.ts
juice49: I think that could be a good idea. Settled for this now just to make things a bit clearer.

PR #8777, opened 2025-02-26
feat(core): add default `code` and `strike-through` markdown shortcuts to PTE

christianhg:
### Description

The `MarkdownPlugin` now also handles the conversion of "\`" pairs as well as
the conversion of "~~" pairs. All you need to do is tell it what decorator to
use for the conversion. In our case, we go with the default decorator names so
anyone who has these configured get the shortcuts.

code-and-strike-through-markdown-shortcuts

### What to review

Do the shortcuts work as expected?

### Testing

The Portable Text Editor has internal tests for these behaviors.

### Notes for release

The Portable Text Input now includes Markdown shortcuts for decorators named `code` and `strike-through`

### DIFF
--- a/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
+++ b/packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
@@ -399,8 +399,13 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
                 config={{
                   boldDecorator: ({schema}) =>
                     schema.decorators.find((decorator) => decorator.value === 'strong')?.value,
+                  codeDecorator: ({schema}) =>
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -399,8 +399,13 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
                 config={{
                   boldDecorator: ({schema}) =>
                     schema.decorators.find((decorator) => decorator.value === 'strong')?.value,
+                  codeDecorator: ({schema}) =>
```
jordanl17, line 402: Well this is nice!

PR #8778, opened 2025-02-26
fix(core): make PTE active by default

christianhg:
### Description

There's been some feedback internally and externally that it would be nice if
the Portable Text Input was active by default. So I think we should try it.

Therefore, the default value of the `PortableTextInput.initialActive` prop has
been flipped from `false` to `true`.

The only downside is that the editor might scrolljack you on a small
device/viewport, but this does not appear to be a good reason to make the
editor inactive by default for everyone, especially not when most users are on
desktop devices.

### What to review

Does it work as expected? Is an input with `initialActive={false}` correctly made inactive on mount?

### Testing

Some existing component tests have been adjusted to match the new behaviour.

### Notes for release

The Portable Text Input is now active by default

### DIFF
--- a/dev/test-studio/schema/standard/portableText/customBlockEditors.tsx
+++ b/dev/test-studio/schema/standard/portableText/customBlockEditors.tsx
@@ -29,10 +29,10 @@ export const ptCustomBlockEditors = defineType({
     },
     {
       name: 'initialActive',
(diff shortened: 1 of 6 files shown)


binoy14 (approved):
makes sense to me, thanks

EoinFalconer (approved):
LGTM

PR #8779, opened 2025-02-26
fix(core): batch requests for archived releases view

pedrobonamin:
### Description

We were doing one request to the history endpoint per document in the archived releases view, this doesn't scale well and in releases with 40 documents the archived view doesn't load because of rate limiting.

This PR removes the 1 request per doc solution and implements a batching strategy, fetching 10 documents per time and waiting for the previous response to end before getting the next one.

So instead of doing N requests (1 per document)
```
https://ppsg7ml5.api.sanity.io/v2025-02-19/data/history/test/documents/doc1
https://ppsg7ml5.api.sanity.io/v2025-02-19/data/history/test/documents/doc2
https://ppsg7ml5.api.sanity.io/v2025-02-19/data/history/test/documents/doc3
```

it does the following request:
```
https://ppsg7ml5.api.sanity.io/v2025-02-19/data/history/test/documents/doc1,doc2,doc3
```
https://test-studio.sanity.dev/test/releases/rIEqtiYbg       doesn't load ❌ 
https://test-studio-git-sapp-2450.sanity.dev/test/releases/rIEqtiYbg loads ✅ 

### What to review
Is the batch good enough?

### DIFF
--- a/packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
+++ b/packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
@@ -9,10 +9,13 @@ import {useMemo} from 'react'
 import {useObservable} from 'react-rx'
 import {combineLatest, from, type Observable, of} from 'rxjs'
 import {
+  concatMap,
   distinctUntilChanged,
+  expand,
   filter,
   map,
   mergeMap,
+  reduce,
   startWith,
   switchAll,
   switchMap,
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/releases/tool/detail/useBundleDocuments.ts
```diff
@@ -182,22 +185,42 @@ const getPublishedArchivedReleaseDocumentsObservable = ({
...
+        url: `/data/history/${dataset}/documents/${release.finalDocumentStates
+          ?.slice(startIndex, finalIndex)
+          .map((d) => d.id)
+          .join(',')}?lastRevision=true`,
+      })
+      .pipe(map(({documents}) => ({documents, finalIndex})))
+  }
+
+  const documents$ = batchRequestDocumentFromHistory(0).pipe(
+    expand((response) => {
```
juice49, line 201: Nice! ✨

`expand` operator noob over here. Two things crossed my mind:

1. Do these events get sorted client-side after they are aggregated? Thinking about what happens if the requests complete out of order.
2. What happens if one of the batched requests fail? Hopefully the parent observable errors. Do the other requests remain in-flight?

pedrobonamin: Thanks for the review Ash!

The requests are not parallelized anymore, this is to avoid the rate limit issue, instead it is batching them every 10 documents and waiting for those 10 documents to come back, before requesting the next 10.
That is how the expand works here, it will wait until the previous is resolved and then it will merge them all together.

This also guarantees that the order is the same.

If one fails this is happening now, not ideal.
Screenshot 2025-02-28 at 10 08 13


[Just updated this ](https://github.com/sanity-io/sanity/pull/8779/commits/ddf4288ffba172cf635cb074c9365e7929ea3cb8)to handle the error and show that something failed loading the documents, thanks for catching this in the review!


https://github.com/user-attachments/assets/b7276bf8-bc6e-4a80-b6e3-f05ebf1ab079

juice49: Thanks for the explanation, and love the error handling 🙏.

juice49 (approved):
This is great! Thanks, Pedro.

PR #8781, opened 2025-02-26
feat(migrate): add support for `--api-version` flag

RostiMelk:
### Description

Allow Sanity Migrate to run on any api version by reading flag `--api-version`. This is very beneficial when testing migrations against content releases, which has many of its api's still under vX

### What to review

Does helper text make sense?

### Testing

I think you can set some sort of debug env to see the requests being sent

### Notes for release

@sanity/migrate now takes an optional `--api-version` flag


bjoerge (approved):
This is great, thank you!

PR #8783, opened 2025-02-27
fix: prevent over-fetching in free trial provider

stipsan:
### Description

The way the `router.state._searchParams` is parsed and handed to `useEffect` are leading to `useEffect` over-firing.
In the case of `FreeTrialProvider`, it's causing it to perform a network call that's only supposed to happen on mount, on every search param change. In Presentation Tool this means that every time you toggle the viewport size it performs a fetch:

https://github.com/user-attachments/assets/a19e5077-48b2-415d-9343-a9713202230a

The reason this happens is the underlying `useEffect` is reacting to _any_ change to `router.state._searchParams`. The fix is to extract the search params the `useEffect` is interested in, so it only reacts to those params changing ✨ 

### What to review

Are there enough context in the comments to explain the pattern used, that isn't resulting in over-fetching?

### Testing

The repro shown in the video can be manually verified [here](https://test-studio-1lwkadbjr.sanity.dev/presentation/presentation) (this link is better than using https://test-studio.sanity.dev, as it points to an immutable deployment that will keep demonstrating the repro, while https://test-studio.sanity.dev will deploy with the fix once this PR merges).

### Notes for release

A issue in how router state search params were handled lead to the Studio checking for your free trial state _every time you changed the viewport size in the Presentation Tool preview_. That doesn't make any sense! Nobody likes unnecessary network requests 😮‍💨 We've tightened the nuts and bolts necessary so the check is only performed _once_, on load 🤌

### DIFF
--- a/packages/sanity/src/core/studio/components/navbar/free-trial/FreeTrialProvider.tsx
+++ b/packages/sanity/src/core/studio/components/navbar/free-trial/FreeTrialProvider.tsx
@@ -42,24 +42,21 @@ export const FreeTrialProvider = ({children}: FreeTrialProviderProps) => {
     }
   }, [showDialog, data, showOnLoad, telemetry])
 
-  // This is casted to a string to make it stable across renders so it doesn't trigger multiple times the effect.
-  const searchParamsAsString = new URLSearchParams(router.state._searchParams).toString()
+  // See if we have any parameters from the current route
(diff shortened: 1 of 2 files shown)


pedrobonamin (approved):
Thank you for fixing this @stipsan !

PR #8785, opened 2025-02-27
feat(core): allow configuring PTE plugins

christianhg:
### Description

**UPDATE**: @pedrobonamin has made subsequent commits to the branch which lets PTE (Portable Text Editor) plugins be configured under `form.components.portableText.plugins`:

```ts
export const autoCloseBrackets = definePlugin({
  name: 'auto-close brackets',
  form: {
    components: {
      portableText: {
        plugins: (props) => {
          return (
            <>
              {props.renderDefault(props)}
              <AutoCloseBracketsBehaviorPlugin />
            </>
          )
        },
      },
    },
  },
})
```

The improved API makes sure that plugins are added recursively so a Studio plugin like the one above can add a PTE plugin and then a document-level array field config further down can add its own and in the end, both will be registered in PTE:

```ts
    /**
     * One-Line Editor
     *
     * Uses the `OneLinePlugin` to restrict the editor to one line of text.
     */
    {
      type: 'array',
      name: 'oneLineEditor',
      title: 'One-Line Editor',
      description: 'The editor is restricted to one line of text using the <OneLinePlugin />',
      of: [
        {
          type: 'block',
        },
      ],
      components: {
        portableText: {
          plugins: (props) => {
            return (
              <>
                {props.renderDefault(props)}
                <OneLinePlugin />
              </>
            )
          },
        },
      },
    },
```

---

This commit adds a `portableText` property into the components,  that can be
used to inject Portable Text Editor Plugins in the editor.

PTE Plugins are just React components. And the reason for this is that for
complex Plugins, you'll want to maintain some local state and the most
idiomatic way to do that is using React. The `DecoratorShortcutPlugin` is an
example of such plugin (it contains an entire XState state machine). Another
reason is that, down the line, we want to render UI elements through Plugins
(think: Emoji Picker).

It should be noted that rendering UI elements in PTE Plugins hasn't been fully
thought through. For example, we might want to expose some affordance to
control *where* the UI elements are rendered. I think, we can address this
later as the `renderPlugins` is marked as `@beta` and isn't intended to be
publically announced yet.

The most important thing right now is that you have an entry point to
get the `Editor` instance using `useEditor()` and add Behaviors using
`editor.registerBehavior(...)`.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -45,6 +45,8 @@ import {resolveInitialValueTemplates} from './initialValueTemplates'
 import {customInspector} from './inspectors/custom'
 import {testStudioLocaleBundles} from './locales'
 import {errorReportingTestPlugin} from './plugins/error-reporting-test'
+import {autoCloseBrackets} from './plugins/input/auto-close-brackets-plugin'
+import {wave} from './plugins/input/wave-plugin'
 import {languageFilter} from './plugins/language-filter'
 import {routerDebugTool} from './plugins/router-debug'
 import {theme as tailwindTheme} from './sanity.theme.mjs'
@@ -189,6 +191,8 @@ const sharedSettings = ({projectId}: {projectId: string}) => {
       tsdoc(),
       media(),
       markdownSchema(),
+      wave(),
+      autoCloseBrackets(),
     ],
   })()
 }
--- a/dev/test-studio/schema/standard/portableText/customPlugins.tsx
+++ b/dev/test-studio/schema/standard/portableText/customPlugins.tsx
@@ -1,0 +1,246 @@
+/* eslint-disable react/jsx-no-bind */
+import {defineBehavior, effect, forward} from '@portabletext/editor/behaviors'
+import {BehaviorPlugin, DecoratorShortcutPlugin, OneLinePlugin} from '@portabletext/editor/plugins'
+import {defineType, type PortableTextInputProps} from 'sanity'
+
+export const customPlugins = defineType({
+  name: 'customPlugins',
+  title: 'Custom Plugins',
+  type: 'document',
+  fields: [
+    {
+      type: 'string',
+      name: 'title',
+    },
+
+    /**
+     * One-Line Editor
+     *
+     * Uses the `OneLinePlugin` to restrict the editor to one line of text.
+     */
+    {
+      type: 'array',
+      name: 'oneLineEditor',
+      title: 'One-Line Editor',
+      description: 'The editor is restricted to one line of text using the <OneLinePlugin />',
+      of: [
+        {
+          type: 'block',
+        },
+      ],
+      components: {
+        input: (props: PortableTextInputProps) => {
+          return props.renderDefault({
+            ...props,
+            renderPlugins: (pluginProps) => {
+              return (
+                <>
+                  {pluginProps.renderDefault(pluginProps)}
+                  <OneLinePlugin />
+                </>
+              )
+            },
+          })
+        },
+      },
+    },
+
+    /**
+     * Custom Markdown Config
+     *
+     * Uses `renderMarkdownPlugin` to reconfigure the `MarkdownPlugin`.
+     */
+    {
+      type: 'array',
+      name: 'customMarkdownConfig',
+      title: 'Custom Markdown Config',
+      description:
+        'Only a "bold" decorator is allowed and the <MarkdownPlugin /> has been reconfigured to support this',
+      of: [
+        {
+          type: 'block',
+          marks: {
+            decorators: [
+              {
+                value: 'bold',
+                title: 'Bold',
+                component: ({children}) => <strong>{children}</strong>,
+              },
+            ],
+          },
+        },
+      ],
+      components: {
+        input: (props: PortableTextInputProps) => {
+          return props.renderDefault({
+            ...props,
+            renderPlugins: (pluginProps) => (
+              <>
+                {pluginProps.renderDefault({
+                  ...pluginProps,
+                  renderMarkdownPlugin: (markdownPluginProps) => {
+                    return markdownPluginProps.renderDefault({
+                      config: {
+                        ...markdownPluginProps.config,
+                        boldDecorator: ({schema}) =>
+                          schema.decorators.find((decorator) => decorator.value === 'bold')?.value,
+                      },
+                    })
+                  },
+                })}
+              </>
+            ),
+          })
+        },
+      },
+    },
+
+    /**
+     * Custom Decorator Shortcuts
+     *
+     * Uses the `DecoratorShortcutPlugin` add custom decorator shortcuts in the
+     * editor.
+     */
+    {
+      type: 'array',
+      name: 'customDecoratorShortcuts',
+      title: 'Custom Decorator Shortcuts',
+      description: 'The markdown shortcuts for bold and italic have been replaced with 👻 and 🕹️',
+      of: [
+        {
+          type: 'block',
+        },
+      ],
+      components: {
+        input: (props: PortableTextInputProps) => {
+          return props.renderDefault({
+            ...props,
+            renderPlugins: (pluginProps) => (
+              <>
+                <DecoratorShortcutPlugin
+                  decorator={({schema}) =>
+                    schema.decorators.find((decorator) => decorator.name === 'strong')?.name
+                  }
+                  pair={{
+                    char: '👻',
+                    amount: 2,
+                  }}
+                />
+                <DecoratorShortcutPlugin
+                  decorator={({schema}) =>
+                    schema.decorators.find((decorator) => decorator.name === 'strong')?.name
+                  }
+                  pair={{
+                    char: '🕹️',
+                    amount: 2,
+                  }}
+                />
+                <DecoratorShortcutPlugin
+                  decorator={({schema}) =>
+                    schema.decorators.find((decorator) => decorator.name === 'em')?.name
+                  }
+                  pair={{
+                    char: '👻',
+                    amount: 1,
+                  }}
+                />
+                <DecoratorShortcutPlugin
+                  decorator={({schema}) =>
+                    schema.decorators.find((decorator) => decorator.name === 'em')?.name
+                  }
+                  pair={{
+                    char: '🕹️',
+                    amount: 1,
+                  }}
+                />
+                {pluginProps.renderDefault({
+                  ...pluginProps,
+                  renderMarkdownPlugin: (markdownPluginProps) => {
+                    return markdownPluginProps.renderDefault({
+                      config: {
+                        ...markdownPluginProps.config,
+                        boldDecorator: undefined,
+                        italicDecorator: undefined,
+                      },
+                    })
+                  },
+                })}
+              </>
+            ),
+          })
+        },
+      },
+    },
+
+    /**
+     * No Plugins
+     *
+     * Removes all plugins from the editor.
+     */
+    {
+      type: 'array',
+      name: 'noPlugins',
+      title: 'No Plugins',
+      description: 'All plugins are removed, even the default ones',
+      of: [
+        {
+          type: 'block',
+        },
+      ],
+      components: {
+        input: (props: PortableTextInputProps) => {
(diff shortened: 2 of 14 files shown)


dev/test-studio/schema/standard/portableText/customPlugins.tsx
```diff
@@ -0,0 +1,248 @@
+/* eslint-disable react/jsx-no-bind */
+import {defineBehavior} from '@portabletext/editor/behaviors'
+import {BehaviorPlugin, DecoratorShortcutPlugin, OneLinePlugin} from '@portabletext/editor/plugins'
+import {defineType, type PortableTextInputProps} from 'sanity'
+
+export const customPlugins = defineType({
```
christianhg, line 6: Because of the `autoCloseBrackets` Studio Plugin, all PTE Inputs have the `autoCloseBracketsBehavior` by default. However, unfortunately, these local usages of `renderPlugins` take precedence.

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -445,6 +441,41 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
...
+    schema.decorators.find((decorator) => decorator.name === 'strike-through')?.name,
+  defaultStyle: ({schema}) => schema.styles.find((style) => style.name === 'normal')?.name,
+  blockquoteStyle: ({schema}) => schema.styles.find((style) => style.name === 'blockquote')?.name,
+  headingStyle: ({schema, level}) =>
+    schema.styles.find((style) => style.name === `h${level}`)?.name,
+  orderedListStyle: ({schema}) => schema.lists.find((list) => list.name === 'number')?.name,
+  unorderedListStyle: ({schema}) => schema.lists.find((list) => list.name === 'bullet')?.name,
+}
+
+function RenderPlugins(props: {renderPlugins?: PortableTextInputProps['renderPlugins']}) {
```
christianhg, line 461: It's likely that this is a very naive take on trying to support `renderPlugins`. At least we know it doesn't work recursively so you can't compose multiple Studio plugins that use it and you can't compose Studio plugins with local schema usages either.

dev/test-studio/sanity.config.ts
```diff
@@ -189,6 +191,8 @@ const sharedSettings = ({projectId}: {projectId: string}) => {
       tsdoc(),
       media(),
       markdownSchema(),
+      wave(),
+      autoCloseBrackets(),
```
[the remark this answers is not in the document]
christianhg: In this example is `PTEBehavior` then a Studio plugin and `WaveBehaviorPlugin` and `AutoCloseBracketsBehaviorPlugin` PTE Plugins?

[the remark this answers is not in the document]
christianhg: Thank you! `PTEBehavior` might be an OK workaround if we don't have other options.

This is essentially is also why I've asked @ricokahler to squint as this PR to hear if there's a better way to implement `renderPlugins` that would make it compose naturally.

EoinFalconer (dismissed):

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
pedrobonamin: I think you are in a good path here. But we need to update how the plugin is defined.
```diff
       return props.renderDefault({
          ...props,
          renderPlugins: (pluginProps) => {
            return (
              <>
-                {pluginProps.renderDefault(pluginProps)}
+                {props.renderPlugins(pluginProps)}
                <AutoCloseBracketsBehaviorPlugin />
              </>
            )
          },
        })
```

That way each field will recursively get the previous plugins. At least it worked for me, but TS is yelling at me.

https://github.com/user-attachments/assets/b7b7a197-2d72-4208-8ef8-35a356c6a7f0



Now.. I don't know if it's the best option, but I think it's the only option given how the inputs are resolved if we want to keep it inside the `input` definition.

Another idea is to add a new property to the components object and define a `portableTextPlugins` property, which we could resolve inside PTE, using the `useMiddleWareComponents` hook like in [here](https://github.com/sanity-io/sanity/blob/main/packages/sanity/src/structure/panes/document/document-layout/useDocumentLayoutComponent.ts#L14)

Happy to follow up this next week, let's set up a sync to discuss it 😄

pedrobonamin: Added some changes here https://github.com/sanity-io/sanity/pull/9447/files does that makes it work as you expect?

christianhg: > I think you are in a good path here. But we need to update how the plugin is defined.

Wow, nice! Can't believe I was this close 😆 🤏 

> Added some changes here https://github.com/sanity-io/sanity/pull/9447/files does that makes it work as you expect?

Nice! This is really good! Incredible work, @pedrobonamin

packages/sanity/src/core/form/inputs/PortableText/PortableTextInput.tsx
```diff
@@ -388,29 +389,7 @@ export function PortableTextInput(props: PortableTextInputProps): ReactNode {
...
-                    schema.styles.find((style) => style.name === 'blockquote')?.name,
-                  headingStyle: ({schema, level}) =>
-                    schema.styles.find((style) => style.name === `h${level}`)?.name,
-                  orderedListStyle: ({schema}) =>
-                    schema.lists.find((list) => list.name === 'number')?.name,
-                  unorderedListStyle: ({schema}) =>
-                    schema.lists.find((list) => list.name === 'bullet')?.name,
-                }}
-              />
+              <PTEPlugins schemaType={schemaType} />
```
RitaDias, line 392: _nice_

RitaDias (approved):
This feels like black magic I'm so excited

packages/sanity/src/core/form/types/blockProps.ts
```diff
@@ -435,3 +436,13 @@ export interface BlockProps {
...
 }
+
+/**
+ * Props for rendering Portable Text plugins
+ *
+ * @beta
+ */
+export interface PtePluginsProps {
+  renderDefault: (props: PtePluginsProps) => React.JSX.Element
+  markdownPluginProps: {config: MarkdownPluginConfig}
```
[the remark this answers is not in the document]
pedrobonamin: Because of how the plugins are added to remove it you will need to remove all of them.
So it would be something like this, and you will need to include again all of the plugins
```diff
          plugins: (props) => {
            return (
              <>
-                {props.renderDefault(props)}
                // all of the other plugins you want to include
+                <MarkdownPlugin />
              </>
            )
          },
```

I follow you in the fact that it is adding things to core that don't feel right, maybe we need to rethink what the plugins are exposing in the renderDefault and add a way to filter the markdownPlugin from the rest of the plugins, and then users will need to add it again with the behavior they want.

christianhg:
> Just a thought…

Good point! Especially in regards to `.(j|t)sx` files and boilerplate.

packages/sanity/src/core/form/types/blockProps.ts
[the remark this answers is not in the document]
pedrobonamin: I like the idea of having a `props.plugins` property.

packages/@sanity/types/src/schema/types.ts
```diff
@@ -227,6 +227,9 @@ export interface BaseSchemaType extends Partial<DeprecationConfiguration> {
     input?: ComponentType<any>
     item?: ComponentType<any>
     preview?: ComponentType<any>
+    pte?: {
```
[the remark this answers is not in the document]
pedrobonamin: Pushed [here](https://github.com/sanity-io/sanity/pull/8785/commits/4cffb3c5e06bd02e7c0d8ac910622450548f0979)

packages/sanity/src/core/config/form/types.ts
```diff
@@ -20,4 +21,7 @@ export interface FormComponents {
   input?: ComponentType<InputProps>
   item?: ComponentType<ItemProps>
   preview?: ComponentType<PreviewProps>
+  pte?: {
```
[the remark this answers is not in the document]
pedrobonamin: Naming updated [here](https://github.com/sanity-io/sanity/pull/8785/commits/4cffb3c5e06bd02e7c0d8ac910622450548f0979) would be good to do a follow up with the interface change, I think we could reuse it.

packages/sanity/src/core/form/types/blockProps.ts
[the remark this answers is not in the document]
pedrobonamin: Did some updates [here](https://github.com/sanity-io/sanity/pull/8785/commits/b88d1d679dcdbb014c4bb9ba331a00107aabc000)
It is moved from `markdownPluginProps` to `plugins.markdown`

This will allow us to add extra default plugins and make them discoverable by using the `plugins` property. Thanks for the suggestion @bjoerge !

packages/sanity/src/core/form/types/blockProps.ts
```diff
@@ -435,3 +436,13 @@ export interface BlockProps {
    */
   value: PortableTextBlock
 }
+
+/**
+ * Props for rendering Portable Text plugins
+ *
+ * @beta
+ */
+export interface PtePluginsProps {
```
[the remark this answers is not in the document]
pedrobonamin: I have pushed it as singular, PortableTextPluginProps, but I think it's not what we really want given you can have multiple plugins which will have multiple props. 
So renamed now to `PortableTextPluginsProps` https://github.com/sanity-io/sanity/pull/8785/commits/5a10bc9aaaac57e3b8cafb21b8456d97d3c790e4

bjoerge (approved):
LGTM and good to go from my side. Amazing work!! 🚢

dev/test-studio/schema/standard/portableText/customPlugins.tsx
```diff
@@ -0,0 +1,234 @@
...
+          plugins: (props) => {
+            return props.renderDefault({
+              ...props,
+              plugins: {
+                markdown: {
+                  config: {
+                    boldDecorator: ({schema}) =>
+                      schema.decorators.find((decorator) => decorator.value === 'bold')?.value,
+                  },
+                },
```
christianhg, line 81: I like the syntax here a lot!

Two thoughts, though:

1. Do we feel like `config` is redundant? Would it be nicer to just flatten the props for the markdown plugin and let them all be top-level? Side note: I expect we'll add a `priority` prop soon and maybe it's nice that the rest of the "config" is separated in their own `config` prop?
```ts
markdown: {
  config: {...},
  priority: {...}
}
```
2. Should be make `plugins.markdown` optional? My reasoning is that if we later add `plugins.emojiPicker` then we have to make that optional as well to not break call sites that have already configured `plugins.`

christianhg:
> Amazing work!! 🚢

I agree! Thank you @pedrobonamin for seeing this through 🙇

dev/test-studio/schema/standard/portableText/customPlugins.tsx
pedrobonamin: I think it's nice to have it inside an object like we have here cause it will allow us to extend it easily, as you mentioned soon we will have the priority, and later we could have something else. So this will not block us in the future.

Re making it optional, I agree, we can make it optional when we add the `emojiPicker` or the next default plugin

christianhg: > I think it's nice to have it inside an object like we have here cause it will allow us to extend it easily, as you mentioned soon we will have the priority, and later we could have something else. So this will not block us in the future.

Thank you for reassuring me 😊 

> Re making it optional, I agree, we can make it optional when we add the emojiPicker or the next default plugin

I realise that if we just apply and promote the spread pattern which is already common, then it's actually a non-issue to have required props on `props.plugins`

```ts
plugins: {
  ...props.plugins,
  markdown: {
    config: {
      boldDecorator: ({schema}) =>
        schema.decorators.find((decorator) => decorator.value === 'bold')?.value,
    },
  },
},
```

PR #8787, opened 2025-02-27
fix: resolveTypeForDocument should work with content releases

### LINKED ISSUE #8761: S.documentList() with filter does not work with content releases
Because of how the `S.documentList()` resolves the type of children (which is completely unique and difficult to replicate for custom lists, but that's another story), it does not work with content releases.

Given the following structure:

```js
S.listItem()
        .title('Tester')
        .child(S.documentList().apiVersion('2025-02-19').title('Test').filter(`...some filter that finds multiple types...`)),
```

If you create a new document in a release and try to open it from this list it will fail with the following error:

Image

The problem is that the the [createDocumentChildResolverForItem](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/structure/structureBuilder/DocumentList.ts#L34) calls [resolveTypeForDocument](https://github.com/sanity-io/sanity/blob/bdb695b9686d4da8dce7692e72b541a6843fcc25/packages/sanity/src/structure/structureBuilder/util/resolveTypeForDocument.ts#L9) which only looks for a draft or published document, not documents in the current perspective. Therefore it does not find the document and therefore does not find the type and the above error is shown.

This is on version 3.77

mariusGundersen:
### Description

This fixes #8761

### What to review

This assumes that all releases have a document with the same _type. If two releases have the same id but different version then things are going to break. This might only be a theoretical problem

### Testing

I have not added any automated tests for this (maybe there should have been, so that this problem didn't occur in the first case?)

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/resolveTypeForDocument.ts
+++ b/packages/sanity/src/core/store/_legacy/document/resolveTypeForDocument.ts
@@ -2,8 +2,6 @@ import {type SanityClient} from '@sanity/client'
 import {type Observable, of} from 'rxjs'
 import {map} from 'rxjs/operators'
 
-import {getDraftId, getPublishedId} from '../../../util'
-
 export function resolveTypeForDocument(
   client: SanityClient,
   id: string,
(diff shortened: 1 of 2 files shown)


mariusGundersen:
Fixed in #8962

PR #8790, opened 2025-02-27
fix(core): rename internal `deepEquals` to `deepEqualsIgnoreKey`

pedrobonamin:
### Description

In the codebase we have uses of `deepEquals from "react-fast-compare"` and `deepEquals from "util/deepEquals"` , this could lead to hard to reason about errors or bugs given they both seem to do the same.

Our internal `deepEquals` ignore the value of `_key` (see [here](https://github.com/sanity-io/sanity/pull/8790/files#diff-1f840ffb1fed669b0a7bebf617ab1da2115a75605c05a8b0d7cfea13bdd9ca73R53) and [here](https://github.com/sanity-io/sanity/pull/8790/files#diff-1f840ffb1fed669b0a7bebf617ab1da2115a75605c05a8b0d7cfea13bdd9ca73R64)) 
This PR renames it to clearly express that this deepEquals function ignores the value of `_key`


bjoerge (approved):
Thank you for this!

PR #8792, opened 2025-02-27
fix(preview): prefer snapshot over original when previewing

bjoerge:
### Description
Fixes an oversight with #8655 that picked the "original" preview value over the "snapshot" (which has the current perspective applied)
This should be the other way around: we always want to preview in current perspective, but in case we cant (e.g. the document we're trying to preview exists only outside of the current perspective, we want to fallback to the original)

Notice "this is the published" before vs "this is the draft" after – these are draft and published of the same document.
### Before

image

###  After
image

### What to review
Looks ok?

### Notes for release
n/a


juice49 (approved):
Great catch! Thanks.

PR #8793, opened 2025-02-27
fix(comments): set up listener with filter-only

bjoerge:
The /listen endpoint only cares about the groq filter, so passing the order clause and projection has no effect, and also gives a false impression of it working. This PR uses the filter separately from the rest of the query when setting up the listener for comments.


pedrobonamin (approved):
Looks good to me, thank you!

PR #8796, opened 2025-02-28
fix(structure): preload should use published id

pedrobonamin:
### Description
Fixes: SAPP-2455
If the `<PreloadPane>` gets a draft id instead of a published, it will throw an error.

`useEditState` should always use the `publishedId` this Pr updates the Preload to pass the published id.
and fix the following error

Screenshot 2025-02-28 at 11 51 24

### Notes for release

Fixes an issue in which in some cases the document list could crash when hovering over a document.


RitaDias (approved):
Looks good :)

PR #8797, opened 2025-02-28
chore(structure): disables sheet list flaky test, functionality is not t enabled

pedrobonamin:
### Description

The sheet list pane functionality is not enabled and is not in use.
We have a flaky failing test, this Pr disables it given it creates unnecessary noise. 

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/structure/panes/documentList/sheetList/__tests__/DocumentSheetListPane.test.tsx
+++ b/packages/sanity/src/structure/panes/documentList/sheetList/__tests__/DocumentSheetListPane.test.tsx
@@ -82,7 +82,7 @@ Object.assign(navigator, {
   },
 })
 
-describe('DocumentSheetListPane', () => {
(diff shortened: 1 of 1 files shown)


RitaDias (approved):
Let's gooo

bjoerge:
side note: I think the test failures may be caused by missing `await` before `act(...)` in the tests here, I see a few instances of `act(...)` that's not awaited. Could consider adding the missing awaits or adding a comment about it in case we want to re-enable the tests in the future

pedrobonamin:
> side note: I think the test failures may be caused by missing `await` before `act(...)` in the tests here, I see a few instances of `act(...)` that's not awaited. Could consider adding the missing awaits or adding a comment about it in case we want to re-enable the tests in the future

It's strange, this is the error received, it looks like mocking is not working as expected in this test. 
Screenshot 2025-02-28 at 14 23 22

PR #8798, opened 2025-02-28
fix(structure): exclude documents outside of selected perspective from document lists

bjoerge:
### Description
Prior to this PR, document lists included all documents across all versions. This was never the original intention, so this reverts that behavior by applying the current perspective when fetching for document lists.

### What to review
Ensure that
- Using the default (drafts) perspective doesn't include documents that only exists in a release
- Selecting the "published" perspective doesn't include documents that exists only as drafts (or documents that only exists in a release)
- Selecting a release perspective doesn't include documents that exists only outside of the selected release stack
- Excluding a release from the global release dropdown correctly excludes documents that only exists in this release

### Testing
Not currently covered by tests afaik, but the review steps above should be a great starting point for e2e tests

### Notes for release
n/a - should already be covered by earlier release notes

### DIFF
--- a/packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
+++ b/packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
@@ -76,7 +76,7 @@ export const DocumentListPane = memo(function DocumentListPane(props: DocumentLi
   const {childItemId, isActive, pane, paneKey, sortOrder: sortOrderRaw, layout} = props
   const schema = useSchema()
   const releases = useActiveReleases()
-  const {perspectiveStack} = usePerspective()
+  const {perspectiveStack, selectedPerspective} = usePerspective()
   const {displayOptions, options} = pane
   const {apiVersion, filter} = options
   const params = useShallowUnique(options.params || EMPTY_RECORD)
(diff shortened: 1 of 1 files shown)


packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
```diff
@@ -113,7 +113,10 @@ export const DocumentListPane = memo(function DocumentListPane(props: DocumentLi
   } = useDocumentList({
     apiVersion,
     filter,
-    perspective: perspectiveStack.length === 0 ? 'raw' : perspectiveStack,
+    perspective:
+      selectedPerspective === 'drafts' || selectedPerspective === 'published'
```
mariusGundersen, line 117: It would be very helpful if `usePerspective()` returned the result of this expression, so that we don't have to replicate it everywhere we need to get something using the current perspective

pedrobonamin (approved):
Thanks Bjoerge, this looks good to me. 

I also agree with the suggestion for `usePerspective` by @mariusGundersen .

packages/sanity/src/structure/panes/documentList/DocumentListPane.tsx
bjoerge: Thanks for the review and the suggestion @mariusGundersen, we'll do this in a follow-up PR  🙇🏼

PR #8801, opened 2025-02-28
chore: exclude mocks and fixtures from bundle

jordanl17:
### Description
`mocks` and `fixtures` were being included in the generated tarball for `sanity`, [see here](https://www.npmjs.com/package/sanity?activeTab=code) where you can find these dirs.

This change excludes them.

### Testing
Tested by `pnpm build` and `pnpm pack` within `packages/sanity` and verifying that these dirs are no longer included

### Notes for release
N/A

### DIFF
--- a/packages/sanity/package.json
+++ b/packages/sanity/package.json
@@ -126,6 +126,8 @@
     "static",
     "!**/__test__/**",
     "!**/__tests__/**",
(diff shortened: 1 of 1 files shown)


juice49 (approved):
Great catch! Thanks! 🙌

PR #8804, opened 2025-02-28
fix(structure): provide feedback as document is added to release

bjoerge:
### Description
Currently, when adding a document to a release there is no progress feedback in the UI. Usually it goes fast enough for you to notice, but in cases where it takes a little while, the button is kept active, and if you hit it again you'll eventually get an error. This fixes the issue by disabling the button as long as the action is in-flight, and also adds a sprinkle of error handling.

### What to review
Does it work as expected?

### Testing
- Try adding a document to a release – the add button should be disabled while adding. Can use network throttling in devtools to simulate latency/lossyness

### Notes for release
n/a


juice49 (approved):
Nice! Thanks, Bjørge.

PR #8805, opened 2025-02-28
fix: handle motion.create warning

stipsan:
### Description

Fixes `motion() is deprecated. Use motion.create() instead.` warnings on `sanity dev` as well as in the test suite log output (CMD + F "motion()" [here](https://github.com/sanity-io/sanity/actions/runs/13590063088/job/37993996651))

### What to review

Assigning you @pedrobonamin since you showed up on the `git blame` ❤ 

### Testing

N/A

### Notes for release

N/A


pedrobonamin (approved):
Thank you Cody!

PR #8819, opened 2025-03-03
fix(core): memoize scheduledPublishing calls, trigger only once

pedrobonamin:
### Description
This PR updates the `hasUsedScheduledPublishing` check to trigger only once by memoizing the observable that will do the request.
It also adds a check in the case that users have opted out from the feature, if they have opted out ` scheduledPublishing.enabled = false` we won't do a request to check if they have used it before or not, because it will be off by default anyways.

If they have explicitly opted in, we won't check this, because it will be true anyways.

### What to review
Is the approach correct?

### Testing
Open the studio, you should see only one request.

### DIFF
--- a/packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.test.tsx
+++ b/packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.test.tsx
@@ -10,6 +10,7 @@ import {
   ScheduledPublishingEnabledProvider,
   useScheduledPublishingEnabled,
 } from './ScheduledPublishingEnabledProvider'
+import {cachedUsedScheduledPublishing} from './useHasUsedScheduledPublishing'
 
 vi.mock('../../../hooks/useFeatureEnabled', () => ({
   useFeatureEnabled: vi.fn().mockReturnValue({}),
@@ -55,6 +56,7 @@ describe('ScheduledPublishingEnabledProvider - previously used', () => {
   beforeEach(() => {
     vi.clearAllMocks()
     mockClient(scheduleResponse)
+    cachedUsedScheduledPublishing.clear()
   })
 
   it('should not show scheduled publishing if user opt out and the feature is not enabled (any plan)', async () => {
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/scheduledPublishing/tool/contexts/ScheduledPublishingEnabledProvider.test.tsx
```diff
@@ -68,7 +70,8 @@ describe('ScheduledPublishingEnabledProvider - previously used', () => {
     expect(value.result.current).toEqual({
       enabled: false,
       mode: null,
-      hasUsedScheduledPublishing: {used: true, loading: false},
+      // Workspace is not enabled, so we won't do a request to check if they have used it or not.
```
juice49, line 73: Thanks for adding these clear test scenarios 🙏.

packages/sanity/src/core/scheduledPublishing/tool/contexts/useHasUsedScheduledPublishing.ts
```diff
@@ -0,0 +1,67 @@
...
+  isWorkspaceEnabled,
+}: {
+  explicitEnabled?: boolean
+  isWorkspaceEnabled?: boolean
+}) {
+  const client = useClient(DEFAULT_STUDIO_CLIENT_OPTIONS)
+  const {projectId, dataset} = useWorkspace()
+  const key = `${projectId}-${dataset}`
+  if (!cachedUsedScheduledPublishing.get(key)) {
+    const hasUsed = fetchUsedScheduledPublishing(client).pipe(shareReplay())
```
juice49, line 50: Just checking you intended to use `shareReplay` with no arguments? I haven't seen this much, and I'm guessing it basically does the same thing as `share`?

juice49 (approved):
Nice. This looks good to me!

packages/sanity/src/core/scheduledPublishing/tool/contexts/useHasUsedScheduledPublishing.ts
bjoerge: shareReplay() differs from share() in that it will replay earlier values to new subscribers (it's comparable to a promise in this regard). Using `shareReplay` without an argument/bufferSize on an observable that emits multiple values, causes the observable to buffer every received value and replay it to new subscribers. 

Probably not a problem here, since it's used with an observable that emits at most a single value, but in general it's a good practice to always specify `bufferSize`.

bjoerge (approved):

packages/sanity/src/core/scheduledPublishing/tool/contexts/useHasUsedScheduledPublishing.ts
juice49: Thanks for explaining 🙇.

PR #8822, opened 2025-03-03
fix(core): add `isReleaseLocked` or `willBeUnpublished` to `useDocumentForm`

pedrobonamin:
### Description
In some cases the documents should have a read only state either because the release is locked or the document will be unpublished.
We are doing this check in the `<DocumentPaneProvider>` but this should be part of `useDocumentForm` so consumers don't need to care about this.

If a release is locked, editing the document will end in an error.
documents that will be unpublished are in theory possible to edit them, but it doesn't make sense from an user perspective and we are disabling this in the document pane.

### What to review
does the changes makes sense?


juice49 (approved):
This is great. Thanks, Pedro 🙌.

PR #8825, opened 2025-03-03
chore: reduce CI e2e flake

stipsan:


packages/sanity/playwright-ct.config.ts
```diff
@@ -55,20 +57,20 @@ export default defineConfig({
...
-            __dirname,
-            './packages/@sanity/util/src/content/index.ts'
-          ),
-        },
+        alias: Object.fromEntries(
+          Object.entries(aliases).map(([pkgName, pkgPath]) => {
+            return [pkgName, path.resolve(monorepoPath, path.join('packages', pkgPath))]
+          }),
+        ),
+        dedupe: ['@sanity/ui', 'styled-components'],
```
stipsan, line 67: this fixes [duplicate context warnings](https://www.sanity.io/help/missing-context-error) that are logged to playwright atm

packages/sanity/playwright-ct/tests/formBuilder/inputs/PortableText/ObjectBlock.spec.tsx
stipsan, line 1: It seems like this test has relied on flaky animation timing.
I've reproduced the behavior on a test studio deployment that predates when the CI started [consistently failing on this test](https://test-studio-fsnvmmgd1.sanity.dev/test/structure/input-standard;portable-text;blocksTest;2b0d8e38-627d-462b-8567-78c661363c99%2Cinspect%3Dsanity%252Fcomments%2Ccomment%3D2e915502-e6d6-4a32-80cd-0075b9d68527).
The expected behavior is that hitting <kbd>escape</kbd> _once_ is enough to close the popover. It seems like this test has passed in the past because depending on animation timing the popover, which contains the element with `data-testid=edit-inline-object-button` remains visible until the animation is complete.
After bumping to `@sanity/ui@2.15.0` the popover animations are a bit shorter (200ms when closing, instead of 400ms), and since then tests have been failing: https://github.com/sanity-io/sanity/actions/runs/13639479326/job/38130181861

RitaDias (approved):

PR #8834, opened 2025-03-04
chore: speed up e2e tests

stipsan:
1. Doubles shards for e2e tests, they currently run 40-ish tests each so now it's down to 20. Making it easier to find and isolate slow and flaky tests.
2. E2E suite is setup to use the `line` playwright reporter, like the e2e-ct suite. This makes it much easier to see what's wrong if something fails (unless you're able to decrypt `Running 44 tests using 2 workers, shard 2 of 2 °·············×··±°···°·····°·×··±·······°°°°·`). It also reports test results progressively, so you don't have to wait 15 mins to see which test failed.
3. e2e-ct suite is no longer opting out of parallelism, like the e2e suite. This should make them about twice as fast since our runners has 2 vCPU's.

### DIFF
--- a/.github/workflows/e2e-ui.yml
+++ b/.github/workflows/e2e-ui.yml
@@ -193,8 +193,8 @@ jobs:
         # Be sure to update all instances in this file and `pr-cleanup.yml` if updated
         project: [chromium, firefox]
         # Add more shards here if needed
-        shardIndex: [1, 2]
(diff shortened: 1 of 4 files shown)


RitaDias (approved):
Pedro might have stronger opinions (sorry for jumping in), but I think it is good

PR #8837, opened 2025-03-04
fix(sanity): ambient types for custom Vitest matchers

juice49:
### Description

This branch moves the ambient type definitions for custom Vitest matchers to `packages/sanity/typings`, ensuring the types are available inside `packages/sanity`. Prior to this change, the `toMatchEmissions` matcher was available at runtime, but the type itself was not.

### What to review

Does it seem reasonable to load the ambient type definitions in this way?

### Testing

The `toMatchEmissions` matcher should be available in the assertion type when the subject is a function that returns an RxJS `OperatorFunction`.


jordanl17 (approved):
Seems fair

PR #8840, opened 2025-03-05
fix(core): reconnection title was missing

pedrobonamin:
### Description
Fixes the title for the reconnection toast.
Screenshot 2025-03-05 at 11 14 56


jordanl17 (approved):
Nice catch

PR #8841, opened 2025-03-05
chore: end blacksmith trial

stipsan:
Reverts sanity-io/sanity#8833.

# Don't merge before the trial is actually ended!!


stipsan:
ending the trial temporarily as we've already reached 33% of the 3k included minutes, need to tweak E2E tests so they don't waste so much compute.

PR #8842, opened 2025-03-05
chore(CI): speedup gh workflows, reduce E2E flake

stipsan:
### Description

- Removes `run_install: false` as this is the default option of `pnpm/action-setup@v4`.
uses `cache: pnpm` on `actions/setup-node@v4`, which replaces custom `cache-node-modules` commands, and leaves it to github to know the best way to cache pnpm.
- In order to use `cache: pnpm` we need to run `pnpm/action-setup` _before_ `actions/setup-node`.
- Uses `node-version: lts/*` so we automatically are on the last stable LTS release of Node.js, so we don't have to manually bump majors every year. A lot of jobs are on Node.js v18, instead of the faster v22.
- The `e2e-ct.yml`, uses turborepo to cache build output, which is much faster than the custom `actions/cache` steps, and doesn't require cleanup steps or garbage collection.
- The `e2e-ct.yml` now caches install for chromium and firefox, webkit still needs to run install on every run for it to be successful, although it's still unclear why.
- The `e2e:build` command now uses turborepo, significantly speeding up the e2e suite as it no longer needs to run `sanity build` on every single git push, only on changes that might affect the suite.
- The `playwright-ct.config.ts` is adjusted to make flake easier to detect, and tests fail faster instead of spending up to 30 mins on the CI before failing. Retries is set to just 1, as our options for `trace` and `video`, assume at least 1 retry on errors (for example `on-first-retry`). 
- `TestForm.tsx`, used by `e2e-ct`, is updated with the same fixes related to unstable `useRef` usage as in the production document providers/
- Regular e2e also has 1 retry now, instead of 4, to better catch flaky and failing tests (we shouldn't allow tests to regularly need 4 retries to pass).
- Refactored a bunch of tests so they're more resilient and less flaky, especially on FireFox

### What to review

Hopefully the changes makes sense and have helpful inline code comments.
In general the idea of setting retries to just `1`, instead of `4` or `6`, is that global limits like these makes it hard to track down flake, since the CI might regularly retry flaky tests 4+ times before they pass. It's better to set higher retry limits on specific flaky tests, as it also makes them easier to find and to keep track of.

### Testing

If existing tests pass we're good 🤞 

### Notes for release

N/A

### DIFF
--- a/packages/sanity/playwright-ct.config.ts
+++ b/packages/sanity/playwright-ct.config.ts
@@ -22,8 +22,8 @@ export default defineConfig({
   fullyParallel: true,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: isCI,
-  /* Flaky tests require us to allow up to 6 retries */
-  retries: 6,
+  /* We allow 1 retry to root out flaky tests */
+  retries: 1,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   reporter: isCI
     ? [['list'], ['blob']]
@@ -38,17 +38,8 @@ export default defineConfig({
         ],
       ],
 
-  /* Maximum time one test can run for. */
-  timeout: 60 * 1000,
-  expect: {
-    // Maximum time expect() should wait for the condition to be met.
-    timeout: 40 * 1000,
-  },
-
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
-    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
-    actionTimeout: 40 * 1000,
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: isCI ? 'on-all-retries' : 'retain-on-failure',
     video: isCI ? 'on-first-retry' : 'retain-on-failure',
--- a/test/e2e/tests/document-actions/delete.spec.ts
+++ b/test/e2e/tests/document-actions/delete.spec.ts
@@ -6,6 +6,10 @@ const name = 'Test Name'
 test(`unpublished documents can be deleted`, async ({page, createDraftDocument}) => {
   await createDraftDocument('/test/content/author')
   await page.getByTestId('field-name').getByTestId('string-input').fill(name)
+  const paneFooter = page.getByTestId('pane-footer-document-status')
+
+  // Wait for the document to save before deleting.
+  await expect(paneFooter).toContainText(/created/i, {useInnerText: true, timeout: 30_000})
 
   await page.getByTestId('action-menu-button').click()
   await page.getByTestId('action-Delete').click()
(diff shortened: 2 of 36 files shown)


packages/sanity/playwright-ct.config.ts
```diff
@@ -38,17 +38,8 @@ export default defineConfig({
         ],
       ],
 
-  /* Maximum time one test can run for. */
-  timeout: 60 * 1000,
```
stipsan, line 42: The default option is 30s, it's better to allow specific tests to run longer if they need to, than to have long global defaults.

test/e2e/tests/document-actions/delete.spec.ts
```diff
@@ -18,14 +22,14 @@ test(`published documents can be deleted`, async ({page, createDraftDocument}) =
...
-  // We are waiting for 1s to make sure the draft actually gets created and click action is not too eager
-  await page.waitForTimeout(1000)
+  // Wait for the document to save before publishing.
+  await expect(paneFooter).toContainText(/created/i, {useInnerText: true, timeout: 30_000})
 
   // Wait for the document to be published.
-  await page.getByTestId('action-publish').click()
-  expect(await paneFooter.textContent()).toMatch(/published/i)
+  await publishButton.click()
+  await expect(paneFooter).toContainText(/published/i, {useInnerText: true, timeout: 30_000})
```
pedrobonamin, line 32: I think this is all great Cody, and really appreciate you taking the time to solve this. 

I wonder something in this extended timeouts, have we seen them resolve after 30 seconds?

In the failing tests I'm seeing the published is not showing and it is just blocked in saving forever, delaying the whole test suite, and having to wait 30seconds for an actions to succeed doesn't feel right, if this is the experience in user land we should get it fixed.

stipsan: I've seen some tests take over 1 minute and still showing the saving spinner. I don't yet know exactly when it happens, it seems to be happening when the entire e2e suite is running and there's a lot of mutations in the e2e dataset so it's difficult to reproduce locally. 
I'll look into this condition after this PR is merged :)

stipsan: And I agree, delaying the whole suite up to 30s is bad but it's better than 5s but then have a suite that will retry up to 5 times.
My goal is to find a better and more reliable approach to waiting for the saving spinner to complete and to understand the conditions causing it to take so long.

EoinFalconer: To add more on this, I agree will longer timeouts rather than multiple retries. We are adding some telemetry to test the speed of this across studios, maybe it will point us in the right direction.

EoinFalconer (approved):

test/e2e/tests/document-actions/delete.spec.ts
pedrobonamin: Yeah, I've also seen it taking long, but I wonder, do they ever resolve after this long or are they just blocked in this state? In the ones I've found I've never seen it resolve, even after 30secs

stipsan: @pedrobonamin yeah, I've seen cases beyond the default 5s, usually around 10. Not in chromium but in firefox. It seems like firefox generally needs longer timeouts, not sure why

stipsan: @pedrobonamin for example, if you download the `full-html-report--attempt-1` artifact from the build that ran just [after this PR merged](https://github.com/sanity-io/sanity/actions/runs/13807392187), you'll see that it waited 5.1 seconds:

image

pedrobonamin: This is a great way to see it, thank you!

PR #8843, opened 2025-03-05
feat(sanity): use document layout middleware chain when rendering `DiffViewPane`

juice49:
### Description

`DiffViewPane` does not currently use the document layout middleware chain when it's rendered. This is problematic for plugins that rely on the `document.components.unstable_layout` option to inject context providers, such as `@sanity/assist`, which crashes inside `DiffViewPane` due to the absent context provider.

This branch wraps the `DiffViewPane`'s `FormBuilder` in the entire document layout middleware chain to solve the issue.

Note: I did consider adding the middleware chain to `FormBuilder` itself, but decided against this, because `FormBuilder` may be used in places where the document layout middleware chain is not desired. For example: when rendering the task form we likely do not wish to wrap it with components that were intended to influence the document editor layout.

We may want to consider adding an API explicitly for injecting context providers and other non-visual components so plugins do not need to misuse `document.components.unstable_layout`.

### What to review

Any issues with rendering the document layout component middleware chain inside `DiffViewPane`?

### Testing

Try opening the diff view with `@sanity/assist` enabled. It should work now. You can do this in Test Studio by navigating to the AI Assist workspace: `/ai-assist/structure/book;056e02e8-bc68-44b1-b882-98dcb95454b1?structure[diffView]=version&structure[previousDocument]=book%2C056e02e8-bc68-44b1-b882-98dcb95454b1&structure[nextDocument]=book%2Cdrafts.056e02e8-bc68-44b1-b882-98dcb95454b1`.

### Notes for release

Fixes a bug causing the document comparison view to crash when `@sanity/assist` is enabled.

### DIFF
--- a/packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
+++ b/packages/sanity/src/structure/diffView/components/DiffViewPane.tsx
@@ -8,6 +8,7 @@ import {
 } from '@sanity/ui'
 import {noop} from 'lodash'
 import {
+  type ComponentType,
   type CSSProperties,
   forwardRef,
   useCallback,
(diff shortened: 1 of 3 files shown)


juice49:
Note: switched back to draft as I've observed some issues with interactivity inside the diff view pane with this change in place.

pedrobonamin (approved):
Looks good to me, thank you @ash!

PR #8844, opened 2025-03-05
fix(releases): fix issue with missing draft/publish status tooltips

bjoerge:
### Description
This fixes a regression in v3.78.0 that made draft+published status indicators not appear unless the dataset had existing releases.

### What to review
Knowing what releases a document exists in involves subscribing to a stub of each version. Prior to this PR this was done by using something along the lines of
```tsx
versions = combineLatest(
  knownReleases.map(releaseId => subscribeToPreviewStub({_id: getVersionId(release, documentId)))
)
```
However, this wouldn't work when no known releases exists, as it would be equivalent with `combineLatest([])`, which never completes. This fixes the issue by checking whether knownReleases is empty, and if so, returning `of({})` instead.

### Testing 
- Test in a dataset without any releases and notice before/after, for example:
  - before: https://test-studio.sanity.dev/listener-events/structure/book
  - after: https://test-studio-git-fix-missing-status-tooltips.sanity.dev/listener-events/structure/book

### Notes for release
- Fixes a regression introduced in v3.78.0 that made draft and published status indicators not appear for documents in datasets without releases.

### DIFF
--- a/packages/sanity/src/core/releases/store/useDocumentVersionInfo.ts
+++ b/packages/sanity/src/core/releases/store/useDocumentVersionInfo.ts
@@ -15,6 +15,8 @@ function exists(value: any) {
 }
 const DOCUMENT_STUB_PATHS = ['_id', '_type', '_rev', '_createdAt', '_updatedAt']
 
+const NO_VERSIONS = {} as Record<string, VersionInfoDocumentStub | undefined>
+
 /**
  * Takes a document id, and returns information about what other versions of the document currently exists
  * @param documentId - The document id. Should be the published id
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/releases/store/useDocumentVersionInfo.ts
```diff
@@ -15,6 +15,8 @@ function exists(value: any) {
 }
 const DOCUMENT_STUB_PATHS = ['_id', '_type', '_rev', '_createdAt', '_updatedAt']
 
+const NO_VERSIONS = {} as Record<string, VersionInfoDocumentStub | undefined>
```
mariusGundersen, line 18: You sholud use `Object.freeze({} as Record<string, VersionInfoDocumentStub | undefined>)` here to prevent another issue where someone modifies this object

EoinFalconer (approved):

packages/sanity/src/core/releases/store/useDocumentVersionInfo.ts
mariusGundersen: It would probably also be help to mark it as `ReadOnly<>`, which is what freeze does in TypeScript.

bjoerge: Oh, yeah, that's a great idea!

PR #8845, opened 2025-03-05
fix(core): update perspectiveStack, use it as the client perspective

pedrobonamin:
### Description
The `usePerspective` hook exposes the `perspectiveStack` but that is only valid as the `perspective` to use with the client when you have a release selected.
If no releases are selected then the `perspectiveStack` returns an empty array, blocking us from using it as intended and making us do all over the place things like this:
```ts
const {perspectiveStack, selectedPerspective} = usePerspective()
  const perspective = useMemo(
    () =>
      selectedPerspective === 'published'
        ? ['published'] // or "published" it returns the same result 
        : selectedPerspective === 'drafts'
          ? ['drafts']  // or "dradts" it returns the same result 
          : perspectiveStack,
    [selectedPerspective, perspectiveStack],
  )
client.fetch(query, {}, {perspective})
```

This PR updates the `perspectiveStack` to also return either `["published"]` or `["drafts"]`, so now we can use it directly

So now consumers can do the following, and it will just work.

```diff
+const {perspectiveStack, selectedPerspective} = usePerspective()
-  const perspective = useMemo(
-   () =>
-      selectedPerspective === 'published'
-        ? ['published'] // or "published" it returns the same result 
-        : selectedPerspective === 'drafts'
-          ? ['drafts']  // or "dradts" it returns the same result 
-          : perspectiveStack,
-    [selectedPerspective, perspectiveStack],
-  )
+ client.fetch(query, {}, {perspective: perspectiveStack})
```

### What to review

Are the results in the studio still correct?

### Testing
Manually tested it works in:
 - references
 - global search
 - raw search (adding a document into the releases plugin)

Existing tests are updated.

### Notes for release

Now users can consume the `perspectiveStack` from the `usePerspective` hook and pass that down to the client.perspective without extra logic.

### DIFF
--- a/packages/sanity/src/core/perspective/__mocks__/usePerspective.mock.ts
+++ b/packages/sanity/src/core/perspective/__mocks__/usePerspective.mock.ts
@@ -7,7 +7,7 @@ export const perspectiveContextValueMock: Mocked<PerspectiveContextValue> = {
   selectedPerspectiveName: undefined,
   selectedReleaseId: undefined,
   selectedPerspective: 'drafts',
-  perspectiveStack: [],
+  perspectiveStack: ['drafts'],
   excludedPerspectives: [],
 }
 export const usePerspectiveMockReturn = perspectiveContextValueMock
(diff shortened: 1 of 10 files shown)


bjoerge (approved):

bjoerge:
Very nice, thank you for this @pedrobonamin

PR #8848, opened 2025-03-05
feat: detect and warn if using http1 against API

rexxars:
### Description

Note: Leaving this as draft for now, hoping that someone can take this over if need be.

Runs a request against `/ping` which allow for performance timing info to be returned. This allows us to read the protocol. If it is not using HTTP2/3, we pop open a toast.

There are a bunch of things here that could use some design work and consideration:

- Should it be a toast or a banner or a developer warning…?
- When closing the toast/banner, should we persist that to localStorage or user preferences?
- Should we wait to do the check with `requestIdleCallback` or similar (to avoid any potential performance overhead)?
- Where should it be mounted/run?
- Someone needs to actually write the help article
- Wording of the toast

### What to review
 
You'll need to manually go in and have the check return 'http/1' or something instead of `entry.nextHopProtocol` to test this - alternatively you can disable HTTP2/HTTP3 in your browser if it supports that.

The code is kinda quickly thrown together - if there's any glaring issues here feel free to correct/change as you see fit.

### Testing

Testing this is very hard! I could not find any reasonable way to tell playwright to disable HTTP2 reliably, so I am honestly not sure what we can do here.

### Notes for release

- Show a warning if the studio is using HTTP1 to communicate with the Sanity API

### DIFF
--- a/packages/sanity/src/core/network/isUsingModernHttp.ts
+++ b/packages/sanity/src/core/network/isUsingModernHttp.ts
@@ -1,0 +1,128 @@
+import {type SanityClient} from '@sanity/client'
+import {from, Observable, of, race, timer} from 'rxjs'
+import {map, take} from 'rxjs/operators'
+
+/**
+ * The /ping route is special in that it allows any origin to access it, and crucially
+ * for this case: also allows _timing_ requests through the `timing-allow-origin` header.
+ * This is what allows us to detect the protocol used for the request by using the
+ * browser Performance API.
+ */
+const checkPath = '/ping'
+
+/**
+ * The tag to use for the request to the /ping route. This is used to identify the
+ * request in the Performance API as well as in the server logs.
+ */
+const checkRequestTag = 'protocol-check'
+
+/**
+ * Preferably we want to use a client, since that will have both a configured URL (the
+ * user may have configured a custom API URL) and also uses the client for _requesting_
+ * the URL. While this last part _shouldn't_ make a difference, there is a theoretical
+ * possibility that XMLHttpRequest (potentially used by the client) uses a different
+ * protocol than the fetch API.
+ */
+const fallbackApiUrl = `https://api.sanity.io/v2025-03-01/ping?tag=sanity.studio.${checkRequestTag}`
+
+/**
+ * Checks whether a request to the API is using a modern HTTP protocol (HTTP/2 or later).
+ * Can emit `undefined` if the protocol could not be detected.
+ *
+ * @param client - The client to use for the request. If not provided, a fetch request
+ * will be made to the fallback API URL. Prefer using a client, as it will have the
+ * correct URL configured.
+ * @returns An Observable that emits `true` if the API request is using a modern HTTP
+ * protocol, `false` if it is not, or `undefined` if it could not be detected.
+ * @public
+ */
+export function isUsingModernHttp(client?: SanityClient): Observable<boolean | undefined> {
+  return getProtocolForApi(client).pipe(
+    map((protocol) => {
+      if (!protocol) {
+        return undefined
+      }
+
+      // Typical values for protocol are "http/0.9", "http/1.0", "http/1.1", "h2", "h2c", "h3", etc.
+      // We consider anything `http/0*` or `http/1*` to be non-modern.
+      return !protocol.startsWith('http/0') && !protocol.startsWith('http/1')
+    }),
+  )
+}
+
+/**
+ * Detects the protocol used for the API request by using the browser Performance API.
+ * This is useful for detecting whether the API request is using HTTP/2 or not.
+ *
+ * @param client - The client to use for the request. If not provided, a fetch request
+ * will be made to the fallback API URL. Prefer using a client, as it will have the
+ * correct URL configured.
+ * @returns An Observable that emits the protocol used for the API request, or `undefined`
+ * if it could not be detected.
+ * @internal
+ */
+function getProtocolForApi(client?: SanityClient): Observable<string | undefined> {
+  if (
+    typeof PerformanceObserver === 'undefined' ||
+    typeof PerformanceResourceTiming === 'undefined'
+  ) {
+    // If we don't have the necessary APIs, we can't do the check
+    return of(undefined)
+  }
+
+  const checkUrl = client ? client.getUrl(checkPath) : `${fallbackApiUrl}${checkPath}`
+
+  // Race the actual timing detection against a 2.5s timer.
+  // If the timer wins, we emit undefined to indicate we couldn’t detect the protocol.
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/network/isUsingModernHttp.ts
```diff
@@ -0,0 +1,128 @@
...
+const checkRequestTag = 'protocol-check'
+
+/**
+ * Preferably we want to use a client, since that will have both a configured URL (the
+ * user may have configured a custom API URL) and also uses the client for _requesting_
+ * the URL. While this last part _shouldn't_ make a difference, there is a theoretical
+ * possibility that XMLHttpRequest (potentially used by the client) uses a different
+ * protocol than the fetch API.
+ */
+const fallbackApiUrl = `https://api.sanity.io/v2025-03-01/ping?tag=sanity.studio.${checkRequestTag}`
```
bjoerge, line 26: I guess in theory a VPN/proxy could allow http/2 on `api.sanity.io` but still degrade to http/1 on `<projectId>.api.sanity.io` or vice versa?

Does this request also risk getting blocked by in-flight requests that now are stalled because of degraded protocol?

bjoerge:
Did some testing and it doesn't seem like we get nextHopProtocol from xmlhttprequests. This is from Chrome:
image

bjoerge:
Thank you for this @rexxars, this is great!

I've rebased against main and done some smaller tweaks:
- Added possibility to test the view-side of this by setting `localStorage._sanity_debugProtocol` to override `nextHopProtocol`, e.g.:

	```ts
	localStorage._sanity_debugProtocol='http/1.1'
	```
	(requires a reload)
- Changed to use `fetch()` only, since it looks like we get an empty string as `nextHopProtocol` when using xmlhttprequest (see comment above)
- Based it on the branch for #5893 and replaced the pushToast call with `useConditionalToast()` which seems like a good fit here
+ some smaller nits I found

bjoerge:
Adding @RitaDias and @pedrobonamin as reviewers since this PR has been tweaked quite a bit since it's incarnation and it doesn't feel right for me to approve it anymore 😅

packages/sanity/src/core/studio/networkCheck/useNetworkProtocolCheck.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [warningDismissedAtRaw, setWarningDismissedAt] = useSessionStorageState(
+    'sanity-studio.network.check.dismiss',
+  )
+
+  const warningDismissedAt = warningDismissedAtRaw ? new Date(warningDismissedAtRaw) : undefined
+
+  useEffect(() => {
+    const sub = isUsingModernHttp(client).subscribe((result) => setIsOnModernHttp(result))
+    return () => sub.unsubscribe()
+  }, [client, pushToast, title])
```
pedrobonamin, line 41: Any reason why we are not using `useObservable` here?
Also, wdyt if we check that the warning is not dismissed before using `isUsingModernHttp`?

packages/sanity/src/core/network/isUsingModernHttp.ts
```diff
@@ -0,0 +1,121 @@
...
+export function isUsingModernHttp(client?: SanityClient): Observable<boolean | undefined> {
+  return getProtocolForApi(client).pipe(
+    map((protocol) => {
+      if (!protocol) {
+        return undefined
+      }
+
+      // Typical values for protocol are "http/0.9", "http/1.0", "http/1.1", "h2", "h2c", "h3", etc.
+      // We consider anything `http/0*` or `http/1*` to be non-modern.
+      return !protocol.startsWith('http/0') && !protocol.startsWith('http/1')
```
pedrobonamin, line 48: If we are returning `undefined` and `boolean` on the same function, meaning different things could lead to unwanted behaviors.
For example if in [here](https://github.com/sanity-io/sanity/pull/8848/files#diff-6cc88eee3b38b53e4b992ca122142c9973b0c43dd93f9f3b84de9df9fb246c6dR55) instead of checking `isOnModernHttp === false` you would have checked `!isOnModernHttp` which are read almost the same way, the hook will show the toast for `undefined` cases.

wdyt if we return an enum instead?
`modern | non-modern | undetermined`

pedrobonamin (commented):
It's great that we found the way to show this to users, will help us a lot debugging why the studio is not working in the best ways.

Added some suggestions and thoughts. Open to discuss them! 
TIL some new things about PerformanceObserver, thank you!

packages/sanity/src/core/network/isUsingModernHttp.ts
pedrobonamin: That would be another solution, it could match better the use case we have now

packages/sanity/test/__snapshots__/exports.test.ts.snap
```diff
@@ -433,6 +433,7 @@ exports[`exports snapshot 1`] = `
     "isUnpublishDocumentEvent": "function",
     "isUnscheduleDocumentVersionEvent": "function",
     "isUpdateLiveDocumentEvent": "function",
+    "isUsingModernHttp": "function",
```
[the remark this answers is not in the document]
pedrobonamin: Agree, I think if we don't find a reason to expose it maybe better to keep it internal.

packages/sanity/src/core/network/isUsingLegacyHttp.ts
```diff
@@ -45,7 +36,7 @@ export function isUsingModernHttp(client?: SanityClient): Observable<boolean | u
 
       // Typical values for protocol are "http/0.9", "http/1.0", "http/1.1", "h2", "h2c", "h3", etc.
       // We consider anything `http/0*` or `http/1*` to be non-modern.
-      return !protocol.startsWith('http/0') && !protocol.startsWith('http/1')
+      return protocol.startsWith('http/0') && protocol.startsWith('http/1')
```
pedrobonamin, line 39: Should this be `||`?
`protocol.startsWith('http/0') || protocol.startsWith('http/1')`

packages/sanity/src/core/network/isUsingModernHttp.ts
bjoerge: renamed to isUsingLegacyHttp and inverted logic, thanks for calling out!

packages/sanity/src/core/network/isUsingLegacyHttp.ts
bjoerge: yes! fixed

pedrobonamin (approved):
Thank you for doing this!

packages/sanity/src/core/network/isUsingModernHttp.ts
pedrobonamin: Thanks!

PR #8849, opened 2025-03-05
fix: resolve PTE popover scroll issue

jordanl17:


packages/sanity/src/core/form/inputs/PortableText/object/modals/PopoverModal.tsx
```diff
@@ -99,7 +99,7 @@ function Content(props: PopoverEditDialogProps) {
               </Box>
 
               <Button
-                autoFocus={Boolean(autoFocus)}
+                autoFocus
```
jordanl17, line 102: `FocusLock` with `autoFocus` means we want to autofocus the first focus-able element. Therefore reading the close button `autoFocus={Boolean(autoFocus)}` defeats the point in using `FocusLock` with `autoFocus`

juice49: Indubitably.

PR #8851, opened 2025-03-06
chore(core): update perspective menu

pedrobonamin:
### Description

Updates the releases nav button,
main changes:
- Move perspective menu and tasks to the left
   Order:  perspective | tasks | presence | help
- Remove x button inside perspective menu
- Remove dot inside perspective menu button
- Use padding 0.5 around buttons inside perspective menu
- Increase hitarea of buttons with the perspective menu
- Add user-select: none on the Published / Drafts text in the collapsed menu, so that it's not possible to click or select this text.

Extra: updates width animation to scale from the previous width and not from 0, making it smoother.

https://github.com/user-attachments/assets/ac00c8de-da70-49a7-97cd-b524537a6bd9

### What to review

Are the changes correct?

### Testing
Manually tested everything keeps working.
Existing tests are updated.

### Notes for release
Updates the releases navigation button UI, increasing click area and removing the X button.

### DIFF
--- a/packages/sanity/src/core/perspective/navbar/ReleasesNav.tsx
+++ b/packages/sanity/src/core/perspective/navbar/ReleasesNav.tsx
@@ -42,12 +42,12 @@ const ReleasesNavContainer = styled(Card)`
     z-index: 2;
   }
 
-  .p-menu-btn + .p-menu-btn {
+  .p-menu-btn:nth-child(2) {
     margin-left: -6px;
   }
-  .p-menu-btn + .p-menu-btn.small {
-    margin: -1px;
-    margin-left: -3px;
+
+  .p-menu-btn:nth-child(3) {
+    margin-left: -6px;
   }
 `
 export function ReleasesNav(): React.JSX.Element {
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/perspective/navbar/ReleasesNav.tsx
```diff
@@ -1,53 +1,70 @@
...
 import {CurrentGlobalPerspectiveLabel} from './currentGlobalPerspectiveLabel'
 import {GlobalPerspectiveMenu} from './GlobalPerspectiveMenu'
 
+const ReleasesNavContainer = styled(Card)`
+  position: relative;
+  --p-bg-color: var(--card-bg-color);
+  &::after {
+    content: '';
+    position: absolute;
+    top: 0;
```
juice49, line 16: You may already know, but `inset: 0` can do `top: 0`, `left: 0`, `bottom: 0`, and `right: 0` all in one 🙂.

juice49 (approved):
This looks great 🙌. I noticed the "clear perspective" button has been removed. It looks intentional, but just thought I'd mention it in case it's not.

packages/sanity/src/core/perspective/navbar/ReleasesNav.tsx
pedrobonamin: True, forgot about it when building this  `inset:0` is quite nice.
Won't update just so we don't have to re run ci, will consider it for next ones, thank you for reviewing!

PR #8860, opened 2025-03-07
fix(cli): use `api.sanity.local` for CLI dev environment

rexxars:
### Description

Minor: changes an outdated URL in the CLI that's generally not used, but also points to a domain we no longer use

### Testing

None (this isn't tested and probably doesn't need to be)

### Notes for release

None


binoy14 (approved):
makes sense to me, thanks

PR #8862, opened 2025-03-07
refactor(release): add publish and schedule actions to menus

RitaDias:


packages/sanity/src/core/releases/tool/components/ReleaseMenuButton/ReleaseMenu.tsx
```diff
@@ -133,7 +150,7 @@ export const ReleaseMenu = ({
   ])
 
   const deleteMenuItem = useMemo(() => {
-    if (release.state !== 'archived' && release.state !== 'published') return null
+    if (release.state === 'active') return null
```
jordanl17, line 153: According to state machine, should only be able to delete from `archived` and `published`
https://miro.com/app/board/uXjVLSepJPw=/

RitaDias: huh I'll update this, not sure why I made this change anymore looking at it. I'll check

RitaDias: This feels like it was a rogue? change? maybe during development, in any case I'm removing it

jordanl17:
NON-BLOCKING:
I noticed something with the pre-flight error handling:

This is the same release, with the API responding with the same error object, but schedule toast render the error whilst run release does not:

SCHEDULE:
Screenshot 2025-03-18 at 13 52 51

RUN RELEASE:
Screenshot 2025-03-18 at 13 53 08

Unlikely this was introduced in this PR - I think it was probably a pre-existing issue with how `ReleasePublishAllButton` handles the `catch`, but might be nice to resolve at this PR, or at least create a separate bug issue for it

packages/sanity/src/core/releases/util/isNotarchivedRelease.ts
jordanl17, line 1: This might be good in `packages/sanity/src/core/releases/util/util.ts` where we also have `isReleaseScheduledOrScheduling`

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -53,6 +53,7 @@ const DEFAULT_ARCHIVED_RELEASES_OVERVIEW_SORT: TableSort = {
   column: 'publishedAt',
   direction: 'desc',
 }
+// eslint-disable-next-line max-statements
```
jordanl17, line 56: Haha - shut up eslint!

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -285,20 +334,35 @@ export const ReleaseScheduleButton = ({
...
-      />
+      {isMenuItem ? (
+        <MenuItem
+          tooltipProps={{
+            disabled: !tooltipText,
+            content: scheduleTooltipContent,
+            placement: 'bottom',
+          }}
+          icon={ClockIcon}
+          disabled={isScheduleButtonDisabled || status === 'scheduling' || documents.length === 0}
```
jordanl17, line 345: COULD: this is nice, but I wonder if we could have `sharedProps` that we can spread into both `MenuItem` and `Button` eg `tooltipProps`, `icon` `disabled` are all shared

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
```diff
@@ -250,9 +287,14 @@ export const ReleaseScheduleButton = ({
         : new Date(),
     )
     setStatus('confirm')
-  }, [release.metadata.intendedPublishAt])
+    onConfirmDialogOpen?.()
+  }, [onConfirmDialogOpen, release.metadata.intendedPublishAt])
 
   const tooltipText = useMemo(() => {
+    if (documents.length === 0) {
```
jordanl17, line 294: COULD: we could also add this to `ReleasePublishAllButton` as I don't think the no docs in release case has a tooltip state

RitaDias: Add it to the `ReleasePublishAllButton`? We have it there [as well](https://github.com/sanity-io/sanity/blob/sapp-2081/packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleasePublishAllButton.tsx#L165-L168)

Unless you mean something different? :)

RitaDias:
> One thing I saw, which might not have any implication, but would probably be good to keep the `releaseType` matching up: If a ASAP `releaseType` is scheduled then we are correctly updating the `releaseType` to `scheduled` But If a `scheduled` `releaseType` is immediately published, we do not update the `releaseType` to `ASAP`
> 
> We have some varying conditions across the releases plugin - where I have a small concern we might reach unintended UI states because the `releaseType` won't in all cases be consistent with the rest of the release document

That should already be the case... 
If a schedule release is published it should be published automatically. The only time that the type is changed is from asap / undecided to schedule. Did you see a case where that wasn't right? If you did, please make sure you are using new releases and not the ones in the test studio, unfortunately while working on this I messed up some of them :|

jordanl17 (commented):
Implementation looks good to me - nice one @RitaDias - many different states to manage in this one, and great reuse of what we already have to make it extensible!!

Just a few comments, I feel only concerned about the one related to updating the `releaseType` when you immediately publish a scheduled release

RitaDias:
> NON-BLOCKING: I noticed something with the pre-flight error handling:
> 
> This is the same release, with the API responding with the same error object, but schedule toast render the error whilst run release does not:
> 
> SCHEDULE: Screenshot 2025-03-18 at 13 52 51
> 
> RUN RELEASE: Screenshot 2025-03-18 at 13 53 08
> 
> Unlikely this was introduced in this PR - I think it was probably a pre-existing issue with how `ReleasePublishAllButton` handles the `catch`, but might be nice to resolve at this PR, or at least create a separate bug issue for it

I have tried to reproduce this but it seems that it's something related to that release in particular.

RitaDias:
> Implementation looks good to me - nice one @RitaDias - many different states to manage in this one, and great reuse of what we already have to make it extensible!!
> 
> Just a few comments, I feel only concerned about the one related to updating the `releaseType` when you immediately publish a scheduled release

I've sent you a private message on slack about this, if we agree on that plan I'll set up the tests in a separate PR so it doesn't add weight to this one

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleaseScheduleButton.tsx
RitaDias: Good point, added something similar in the `ReleasePublishAllButton` as well :) 
https://github.com/sanity-io/sanity/pull/8862/commits/03a33bf7aba603bf85029db3f13a30b9fc4fc2f9

jordanl17 (approved):

Issue #8827, opened 2025-03-04
Cursor jumps back to the top of the container in Vision


eduardonwa:
If you find a security vulnerability, do NOT open an issue. Email [security@sanity.io](mailto:security@sanity.io) instead.

**Describe the bug**

Sometimes when I type a query within the editor of vision the cursor jumps to the first line of the container. This has happened to me more than a few times. Not sure if the length of the query has anything to do with it, or I am triggering some shortcut command by accident while I'm typing symbols, like "{", "[", etc... it seems random.

**To Reproduce**

1. Open Vision through the studio, from the top centered tabs
2. Start typing your query as you normally would.
3. It may not happen every time but like I mentioned it feels random. Not sure what I am doing wrong to cause this.

**Expected behavior**

The cursor shouldn't move randomly back to the first line of the editor.

**Screenshots**
I will edit this post with a video when I get it next time.

**Which versions of Sanity are you using?**
@sanity/cli (global)    3.77.2 (up to date)
@sanity/vision          3.77.2 (up to date)

**What operating system are you using?**
Windows 10

**Which versions of Node.js / npm are you running?**
npm -v && node -v
11.1.0
v22.14.0

**Additional context**

**Security issue?**

Any security issues should be submitted directly to [security@sanity.io](mailto:security@sanity.io). In order to determine whether you are dealing with a security issue, ask yourself these two questions:

- Can I access something that's not mine, or something I shouldn't have access to?
- Can I disable something for other people? If the answer to either of those two questions are "yes", then you're probably dealing with a security issue. Note that even if you answer "no" to both questions, you may still be dealing with a security issue, so if you're unsure, just email us at [security@sanity.io](mailto:security@sanity.io.

theodorSchei:
I am experiencing the same issue, on both chrome and firefox. Here is a screen capture of an example. I also always seem to miss one character when the cursor moves.

https://github.com/user-attachments/assets/6b26acc8-4499-4c31-a301-441a62dd6c3d

mariusGundersen:
I am also experiencing this issue

dfinn2:
I am experiencing the same issue. Can't see any pattern in the behaviour but seems to jump back up to the top of the editor every 10-15 typed characters.

pedrobonamin:
Hello everyone, thanks for reporting. 
We have a PR up that should fix this, aiming to have it released today. Will come back here once it is released.

PR #8878, opened 2025-03-11
fix(vision): scroll jump - transform vision code mirror to uncontrolled

pedrobonamin:
### Description
fixes: https://github.com/sanity-io/sanity/issues/8827
Linear issue: SAPP-2009

This PR does two different things:
By one side it reverts the change introduced in https://github.com/sanity-io/sanity/pull/8238 because it didn't fix the underlying issue and it introduced a new one (hitting enter after just finishing writing was not triggering the correct query).

Instead, it takes a new approach of transforming the component to an uncontrolled editor, it does that by setting only an initial value, any subsequent change that is needed, like for example when pasting a url to process the query and get the correct values it's done by the `resetEditorContent` function which is exposed by the imperative handle.

https://github.com/user-attachments/assets/9522acf9-eeae-48c5-aec9-9e1ae5e855d7

I have a follow up PR https://github.com/sanity-io/sanity/pull/8879 which refactors the vision gui component, divides it into smaller pieces and transforms it to a functional component instead.

### What to review

Are the changes correct?
Does vision scroll jump stopped? 

### Testing
Manually tested the editor by copy and pasting values to verify this continue working.
Manually tested reloading, the saved query is persisted.
Manually tested scroll jump, I found out that when pressing 5 or 6 keys at **the same time** in the keyboard the scroll jumped to the start, after the change, it is not jumping anymore.

### Notes for release
fixes vision scroll jump.

### DIFF
--- a/packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
+++ b/packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
@@ -1,19 +1,66 @@
 import {useTheme} from '@sanity/ui'
-import CodeMirror, {type ReactCodeMirrorProps} from '@uiw/react-codemirror'
+import CodeMirror, {
+  EditorSelection,
+  type ReactCodeMirrorProps,
+  type ReactCodeMirrorRef,
+} from '@uiw/react-codemirror'
+import {forwardRef, useCallback, useImperativeHandle, useRef, useState} from 'react'
 
 import {codemirrorExtensions} from './extensions'
 import {useCodemirrorTheme} from './useCodemirrorTheme'
 import {EditorRoot} from './VisionCodeMirror.styled'
 
-export function VisionCodeMirror(
-  props: Omit<ReactCodeMirrorProps, 'basicSetup' | 'theme' | 'extensions'>,
-) {
+export interface VisionCodeMirrorHandle {
+  resetEditorContent: (newContent: string) => void
+}
+
+export const VisionCodeMirror = forwardRef<
+  VisionCodeMirrorHandle,
+  Pick<ReactCodeMirrorProps, 'onChange'> & {
+    initialValue: ReactCodeMirrorProps['value']
+  }
+>((props, ref) => {
+  // The value prop is only passed for initial value, and is not updated when the parent component updates the value.
+  // If you need to update the value, use the resetEditorContent function.
+  const [initialValue] = useState(props.initialValue)
   const sanityTheme = useTheme()
   const theme = useCodemirrorTheme(sanityTheme)
+  const codeMirrorRef = useRef<ReactCodeMirrorRef>(null)
+
+  const resetEditorContent = useCallback((newContent: string) => {
+    const editorView = codeMirrorRef.current?.view
+    if (!editorView) return
+
+    const currentDoc = editorView.state.doc.toString()
+    if (newContent !== currentDoc) {
+      editorView.dispatch({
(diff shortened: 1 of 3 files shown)


packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
```diff
@@ -1,19 +1,66 @@
...
+  }
+>((props, ref) => {
+  // The value prop is only passed for initial value, and is not updated when the parent component updates the value.
+  // If you need to update the value, use the resetEditorContent function.
+  const [initialValue] = useState(props.initialValue)
   const sanityTheme = useTheme()
   const theme = useCodemirrorTheme(sanityTheme)
+  const codeMirrorRef = useRef<ReactCodeMirrorRef>(null)
+
+  const resetEditorContent = useCallback((newContent: string) => {
```
juice49, line 30: Just a thought, feel free to ignore: does the name "reset" make sense here when the caller is providing new content? `setEditorContent` feels like it might make more sense to me.

packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
```diff
@@ -1,19 +1,66 @@
...
+    const editorView = codeMirrorRef.current?.view
+    if (!editorView) return
+
+    const currentDoc = editorView.state.doc.toString()
+    if (newContent !== currentDoc) {
```
juice49, line 35: I guess the main reason for this check is to avoid moving the cursor if the content is unchanged? Wondering how necessary that is, and whether it justifies the cost of stringifying and comparing the state (maybe that's _super cheap_).

packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
```diff
@@ -1,19 +1,66 @@
...
+        value={initialValue}
+        onChange={props.onChange}
+      />
     </EditorRoot>
   )
-}
+})
+
+// Add display name
+VisionCodeMirror.displayName = 'VisionCodeMirror'
```
juice49, line 66: Nice.

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -272,13 +275,12 @@ export class VisionGui extends PureComponent<VisionGuiProps, VisionGuiState> {
     this.handleListenExecution = this.handleListenExecution.bind(this)
     this.handleListenerEvent = this.handleListenerEvent.bind(this)
     this.handleQueryExecution = this.handleQueryExecution.bind(this)
-    this.handleQueryChange = debounce(this.handleQueryChange, 300).bind(this)
+    this.handleQueryChange = this.handleQueryChange.bind(this)
```
juice49, line 278: 🙌

juice49 (approved):
This is feeling fantastic! I lightly tested the following:

1. The caret no longer jumps while typing.
2. The query executed is no longer stale.
3. Pasting a query URL still works.
4. Pasting into the query body still works.

packages/@sanity/vision/src/codemirror/VisionCodeMirror.tsx
pedrobonamin: Yeah, it's not strictly necessary, will remove it in the follow up.
However we still need the `currentDoc` to use the length of it for the insert action in the dispatch.

PR #8879, opened 2025-03-11
feat(vision): transforms vision GUI into a functional component 

pedrobonamin:
### Description
This Pr refactors vision GUI into a functional component.
This is to bring it up to year 2025 and making it possible to use the react-compiler in it and have improved performance.
It also refactors the component by dividing some of the pieces into new components, simplifying the main component and the logic around it.

### What to review
Vision functionalities continue working as before:
- copy and paste a query url.
- Copy and paste normal text.
- Write some queries and fetch the data.
- Listener keeps working.
- Changing datasets works
- Changing the perspective works

### Testing
We don't have neither integration or e2e testing to cover this, so only manual testing for now.


pedrobonamin:
> r if we had tests. Can we create a task to follow up on this please?

Agree on a 100%! will see if I can spend some time today adding this before we merge this.
The ideal would be to add the tests to `next` before merging this, so we can validate that all tests are still passing after the refactor.

pedrobonamin:
> > r if we had tests. Can we create a task to follow up on this please?
> 
> Agree on a 100%! will see if I can spend some time today adding this before we merge this. The ideal would be to add the tests to `next` before merging this, so we can validate that all tests are still passing after the refactor.

Tests added [here](https://github.com/sanity-io/sanity/pull/8964)

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -90,1058 +50,606 @@ function nodeContains(node: Node, other: EventTarget | Node | null): boolean {
...
+      if (
+        newPerspective &&
+        (!isSupportedPerspective(parts.options.perspective) ||
+          isVirtualPerspective(parts.options.perspective))
+      ) {
+        toast.push({
+          closable: true,
+          id: 'vision-paste-unsupported-perspective',
+          status: 'warning',
+          title: 'Perspective in pasted url is currently not supported. Falling back to "raw"',
```
RitaDias, line 473: I think we should translate this? :)

packages/@sanity/vision/src/components/VisionGui.tsx
```diff
@@ -90,1058 +50,606 @@ function nodeContains(node: Node, other: EventTarget | Node | null): boolean {
...
 
-    let apiVersion: string | undefined
-    let customApiVersion: string | false | undefined
+      setQueryInProgress(!context.params.error && Boolean(context.query))
+      setListenInProgress(false)
+      setListenMutations([])
+      setError(context.params.error ? new Error(context.params.error) : undefined)
+      setQueryResult(undefined)
+      setQueryTime(undefined)
+      setE2eTime(undefined)
```
RitaDias, line 220: I'm adding this comment to this one but it's more of a general comment to more of this sort of query logic in this component, do you think it makes sense to move and split some of this logic to its own separate helper file? The feeling (though it might be wrong) is that we could probably simplify this component

that being said, I understand if it's just outside of the scope for this task

packages/@sanity/vision/src/components/VisionGuiHeader.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const handleCopyUrl = useCallback(() => {
+    const el = operationUrlElement.current
+    if (!el) return
+
+    try {
+      el.select()
+      document.execCommand('copy')
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error('Unable to copy to clipboard :(')
```
RitaDias, line 90: Same thing with translating these messages!

packages/@sanity/vision/src/components/VisionGui.tsx
pedrobonamin: Sure!

packages/@sanity/vision/src/components/VisionGui.tsx
pedrobonamin: I think I can do a follow up abstracting some of the logic, specially this handlePaste I think it does too much in here, also it would be nice to include some unit tests for the paste logic. 
But I think it shouldn't be part of this PR, I think if the new logic moves away too much from the original one, maybe tracking a regression will be harder, as the diff will be harder to understand

RitaDias: Fair enough, I agree!

RitaDias (approved):
Let's gooooooo

PR #8881, opened 2025-03-11
feat(studio): refine releases overview toolbar layout

juice49:
### Description

This branch makes some small changes to improve the releases overview toolbar layout and make the buttons more consistent.

This is best illustrated with the flex layout highlighted.

#### Before

Screenshot 2025-03-11 at 09 36 11

#### After

Screenshot 2025-03-11 at 09 35 35

### What to review

The updated layout.

### Testing

Tested the layout manually in web browser.

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
@@ -1,8 +1,6 @@
 import {AddIcon, ChevronDownIcon, EarthGlobeIcon} from '@sanity/icons'
 import {
   Box,
-  // eslint-disable-next-line no-restricted-imports
-  Button,
   type ButtonMode,
   Card,
   Container,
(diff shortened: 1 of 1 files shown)


RitaDias (approved):
Nice!

packages/sanity/src/core/releases/tool/overview/ReleasesOverview.tsx
```diff
@@ -219,7 +216,7 @@ export function ReleasesOverview() {
           placement="bottom"
         >
           <div>
-            <MotionUiButton
+            <MotionButton
```
jordanl17, line 219: NIT: I recall using these 2 different buttons because, from vague memory, only the ui-component button supports a tooltip, and we show a tooltip on the archived group when it's disabled because there are no archived releases. Might want to just verify that this still works?

juice49: Thank you for raising this! I tested the tooltip, and it's working as expected.

PR #8882, opened 2025-03-11
chore(tests): disable flaky references e2e tests

pedrobonamin:
### Description
The reference tests are flaky.
Disable them for now.

We will come back to it in SAPP-2485


RitaDias:
I've added myself as reviewer, please run this one or two times because I have a feeling it might also be flaky in chromium (just in case!)

pedrobonamin:
> I've added myself as reviewer, please run this one or two times because I have a feeling it might also be flaky in chromium (just in case!)

disabling also in chromium, it is failing there as well, less flaky but it does fails.

PR #8885, opened 2025-03-11
chore(tests): remove e2e-ct report

pedrobonamin:
### Description

The report generated is incorrect, it shows that some tests failed but they are passing on CI.
This pr removes this report, given it's not useful now.
see for example https://github.com/sanity-io/sanity/pull/8882#issuecomment-2713495757


.github/workflows/e2e-ct.yml
```diff
@@ -229,13 +229,6 @@ jobs:
         env:
           GITHUB_WORKFLOW_URL: "${{ steps.job_html_url.outputs.html_url }}"
 
-      - name: PR comment with report
```
jordanl17, line 232: I don't think we'd need `packages/sanity/playwright-ct/scripts/parsePlaywrightReportJson.ts` either, or associated utils

PR #8891, opened 2025-03-11
feat(presentation): include perspectiveStack and version in resolvers

pedrobonamin:
### Description

This PR adds to presentation resolvers two new options that are necessary for the resolvers to correctly resolve the queries when using releases.
`version`
`perspectiveStack``

The version is necessary to know if the document you are trying to resolve is a version document and which version it belongs to.
The stack is necessary to expose to users a way to query for the document, they can provide this value to the `client.perspective` 

Example of returned value:

```ts
{
  id: "5ca15b90-6317-4ab0-a757-54a961671661",
  perspectiveStack: ['rKM9a34iK', 'drafts'],
  type: "author",
  version: "rKM9a34iK"
}

```

### What to review
Is this change correct? Is there any other place we need to update to expose this value?

### Notes for release

include `perspectiveStack` and `version` in presentation resolvers

### DIFF
--- a/packages/sanity/src/presentation/document/PresentationDocumentHeader.tsx
+++ b/packages/sanity/src/presentation/document/PresentationDocumentHeader.tsx
@@ -18,14 +18,15 @@ const LocationStack = styled(Stack)`
 
 export function PresentationDocumentHeader(props: {
   documentId: PublishedId
+  version: string | undefined
   options: PresentationPluginOptions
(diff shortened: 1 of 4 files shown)


stipsan (approved):
Thank you! 💖 I'm working on setting up a testing suite for presentation where we document the full API, I think it's best to follow up then and see if we're covering all our bases 😮‍💨

PR #8900, opened 2025-03-12
feat(presentation): allow multiple preview URL origins

stipsan:
### Description

This PR adds support for Presentation to move between preview URL origins within a single instance of Presentation.
The URL origin Presentation connects to is currently configured with `previewUrl.origin`:
```ts
import {presentationTool} from 'sanity/presentation'

export default defineConfig({
  plugins: [presentationTool({
    previewUrl: {
      origin: 'https://example.com'
    }
  })]
})
```

This is problematic, what if I want to connect to both my website and my webstore as both use the same dataset and schema?

#### Option A: multiple Presentation Tool instances
```ts
import {presentationTool} from 'sanity/presentation'

export default defineConfig({
  plugins: [
    presentationTool({
      name: 'website',
      previewUrl: {origin: 'https://example.com'}
    }),
    presentationTool({
      name: 'webstore',
      previewUrl: {origin: 'https://example.store'}
    }),
  ]
})
```
The downside here is that it doesn't scale. If you have 4+ surfaces you quickly clutter the studio navbar with tools. If you use the locations API it takes up a lot of space when editing documents. 

#### Option B: configure origin dynamically
```ts
import {presentationTool} from 'sanity/presentation'

const url = new URL(location.href)
let origin = 'https://example.com'
// If the &preview param exists in the URL, extract the origin from it
try {
  origin = new URL(url.searchParams.get('preview') || origin).origin
} catch {
  // ignore
}
export default defineConfig({
  plugins: [
    presentationTool({
      previewUrl: {origin}
    }),
  ]
})
```
The downside here is that the UX is terrible, you have to edit the &origin search param in the URL and reload the entire studio whenever you want to switch between preview origins.

#### Conclusion
In both cases whatever value `previewUrl.origin` is locked-in once Presentation loads. You can't edit the URL bar to go to `https://example.store`/ if Presentation were on `https://example.com`/ initially.
This is especially a problem if an application is configured with multiple top-level domains for internationalization and localization purposes. `https://example.com`/ might load up in english, and for french it switches to `https://example.fr`/ and Presentation would disconnect.

#### Solution

To solve for these issues there's a new top level option that configures allowed origins, here's how it solves the problem that options A and B failed:
```ts
import {presentationTool} from 'sanity/presentation'

export default defineConfig({
  plugins: [
    presentationTool({
      allowOrigins: ['https://example.com', 'https//example.store', 'http://localhost:*'],
      previewUrl: {
         initial: 'https://example.com/' // the initial preview URL when there's no `&preview` param in the URL
      }
    }),
  ]
})
```

The new `allowOrigins` takes an array of [URLPattern](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) strings.
You can provide an async function to configure it dynamically:
```ts
import {presentationTool} from 'sanity/presentation'

export default defineConfig({
  plugins: [
    presentationTool({
      allowOrigins: async ({client, origin}) => {
        // Query known origins from  the current dataset with the sanity client
        const productionDomains = await client.fetch<string[]>(`*[_type == 'market' && _id == $id].domains`, {id: 'europe'})
        // If the studio is running from localhost we also allow Presentation to connect to any localhost instance
        if(origin.startsWith('http://localhost/')) return [...productionDomains, 'http://localhost:*']
        return productionDomains
      },
      previewUrl: {
        // Resolving the initial URL when no search param of the preview URL is present also takes an async function
        initial: async ({client, origin}) => {
          // If the studio is on localhost we assume the Next.js instance is also running
          if(origin.startsWith('http://localhost/')) return 'http://localhost:3000/'
          // Otherwise we pick the first production domain as the initial URL
          return client.fetch<string>(`*[_type ==  'market' && _id == $id].domains[0]`, {id: 'europe'])
        }
      }
    }),
  ]
})
```

Preview Mode can also take an async function so you can change its configuration, or disable it, depending on the env:

```ts
import {presentationTool} from 'sanity/presentation'

export default defineConfig({
  plugins: [
    presentationTool({
      allowOrigins: ['https://example.com', 'https://example.store', 'https://example.dev'],
      previewUrl: {
        initial: 'https://example.com/',
        previewMode: async ({client, targetOrigin, origin}) => {
          // if the studio is on localhost you might need to handle http => https nuance
          if(origin.startsWith('http://localhost/')) {}
          // `client` is a SanityClient instance, `targetOrigin` is the URL origin of the Preview URL that Presentation is about to load up in its iframe

          switch(targetOrigin) {
            // Next.js app
            case: 'https://example.com':
              return {enable: '/api/draft-mode/enable'}
            // Shopify Hydrogen app
            case 'https://example.store':
              return {enable: '/api/preview'}
            // The staging app always loads `@sanity/visual-editing` and doesn't need to perform a "Preview Mode" handshake
            case 'https://example.dev':
              return false
            default:
              throw new Error(`Unknown target origin: ${targetOrigin}`, {cause: targetOrigin})
          }
        }
      },
    }),
  ]
})
```

#### UX of switching origins

This is the first MVI for multiple origins support. To land basic functionality it was necessary to do a major refactoring of internals. The `&preview` search param had to move from using a relative URL to an absolute one. It requires origin validation for security.
We'll do another iteration that focuses on the editorial experience.
And so the first MVI is a bit low level and manual.

The most obvious way to switch is by typing in another URL in the preview bar. When switching origins the preview handshake is performed again and the URL bar will show a loading indicator:

https://github.com/user-attachments/assets/ad01ab98-697b-4546-8339-c6f79629b16a

In the next MVI we want to add autocomplete to the URL bar so it shows you known, valid, origins.

Another way to move to a new origin is to use absolute URLs in your locations resolver:
```ts
presentationTool({
  allowOrigins: ['https://preview-kit-*.sanity.dev'],
  previewUrl: {
    initial: 'https://preview-kit-remix.sanity.dev',
    previewMode: {enable: '/api/draft'},
  },
  resolve: {
    locations: {
      page: defineLocations({
        locations: [
          {title: 'App Router', href: 'https://preview-kit-next-app-router.sanity.dev/'},
          {title: 'Pages Router', href: 'https://preview-kit-next-pages-router.sanity.dev/'},
          {title: 'Remix', href: 'https://preview-kit-remix.sanity.dev/'},
        ],
      }),
    },
  },
})
```

While this works on simple setups it can quickly get messy when:
- There's already a lot of "Used on" results, since locations that change the origin isn't grouped separately from relative URLs.
- There's no way to dynamically resolve locations based on the current URL origin.

The next MVI will address this. Until then the navigator API can be used to build custom URL origin switching logic and UI:
```tsx
presentationTool({
  allowOrigins: ['https://preview-kit-*.sanity.dev'],
  previewUrl: {
    initial: 'https://preview-kit-remix.sanity.dev',
    previewMode: {enable: '/api/draft'},
  },
  components: {
    unstable_navigator: {minWidth: 120, maxWidth: 240, component: CustomNavigator},
  },
})

import {Card, Stack, Text} from '@sanity/ui'
import {usePresentationNavigate, usePresentationParams} from 'sanity/presentation'

export function CustomNavigator(): React.JSX.Element {
  const navigate = usePresentationNavigate()
  const {preview} = usePresentationParams()

  return (
    <Card flex={1} height="fill">
      <Stack padding={2} space={1}>
        <Card
          as="button"
          onClick={() => navigate('https://preview-kit-next-app-router.sanity.dev/')}
          padding={3}
          pressed={preview?.startsWith('https://preview-kit-next-app-router')}
          radius={2}
        >
          <Stack space={2}>
            <Text size={0} muted>
              Next.js
            </Text>
            <Text>App Router</Text>
          </Stack>
        </Card>
        <Card
          as="button"
          onClick={() => navigate('https://preview-kit-next-pages-router.sanity.dev/')}
          padding={3}
          pressed={preview?.startsWith('https://preview-kit-next-pages-router')}
          radius={2}
        >
          <Stack space={2}>
            <Text size={0} muted>
              Next.js
            </Text>
            <Text>Pages Router</Text>
          </Stack>
        </Card>
        <Card
          as="button"
          onClick={() => navigate('https://preview-kit-remix.sanity.dev/')}
          padding={3}
          pressed={preview?.startsWith('https://preview-kit-remix')}
          radius={2}
        >
          <Text>Remix</Text>
        </Card>
      </Stack>
    </Card>
  )
}

```

Here's what that looks like:

https://github.com/user-attachments/assets/b4535cc5-6bb7-48f7-823f-db18f8f3c744

If you try to navigate to an origin that isn't in the allowOrigins list you'll get a toast error:

image

##### `previewUrl.initial`

This option is what Presentation now uses to infer both what origins are allowed to connect, as well as what the initial preview URL should be.
It can be either a string, or a function that resolves a string. 
The string can be both an absolute or relative URL.

```ts
presentationTool({
  previewUrl: {initial: 'https://example.com/blog'}
})
presentationTool({
  previewUrl: {initial: async ({client, origin}) => 'https://example.com/blog'}
})
```

##### `allowOrigins`

Allow origins is optional, and only needed if you expect to connect to multiple origins.
Whatever URL you resolve in `initial` will always be added to the list of `allowOrigins` automatically.

IOW these are equivalent:
```ts
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: ['http://localhost:*', 'https://example.com']
  }
})
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: ['http://localhost:*']
  }
})
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: async ({client, origin, initialUrl}) => {
      return ['http://localhost:*', initialUrl.origin]
    }
  }
})
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: async ({client, origin, initialUrl}) => {
      return ['http://localhost:*']
    }
  }
})
```

For backwards compatibility the `initial` option is optional, and if not present its value is inferred from the deprecated `origin` and `preview` options:

```ts
presentationTool({
  previewUrl: {
    origin: 'https://example.com',
    preview: '/blog',
    allowOrigins: ['http://localhost:*', 'https://example.com']
  }
})
presentationTool({
  previewUrl: {
    origin: 'https://example.com',
    preview: '/blog',
    allowOrigins: ['http://localhost:*']
  }
})
presentationTool({
  previewUrl: {
    origin: 'https://example.com',
    preview: '/blog',
    allowOrigins: async ({client, origin, initialUrl}) => {
      return ['http://localhost:*', initialUrl.origin]
    }
  }
})
presentationTool({
  previewUrl: {
    origin: 'https://example.com',
    preview: '/blog',
    allowOrigins: async ({client, origin, initialUrl}) => {
      return ['http://localhost:*']
    }
  }
})
```

##### async `previewUrl.previewMode`

Since we're introducing the concept of multiple origins it's necessary to have the ability to change the preview mode configuration dynamically:

```ts
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: ['http://localhost:*'],
    async ({origin, targetOrigin}): Promise<string | false> => {
      // Custom logic based on the origin of the studio
      if(origin.startsWith('http')) {}
      // Change the configuration based on the origin of what we're connecting to
      if(targetOrigin === 'https://example.com')
    }
  }
})
```

If you're using advanced URLPatterns in `allowOrigins` and want to reuse them in the previewMode config, here's how:
```ts
const previewDeploymentOrigin = 'https://example-git-*.com'
presentationTool({
  previewUrl: {
    initial: 'https://example.com/blog',
    allowOrigins: ['http://localhost:*', previewDeploymentOrigin],
    async ({origin, targetOrigin}): Promise<string | false> => {
      // Instantiate the URLPattern object
      const previewDeploymentPattern = new URLPattern(previewDeploymentOrigin)

      return {
        enable: '/api/draft-mode/enable',
        // If it's a preview deployment we enable share URLs, otherwise we don't allow them
        shareAccess: previewDeploymentPattern.test(targetOrigin)
      }
    }
  }
})
```

##### `previewUrl.origin`

With support for multiple origins it no longer makes sense to have this option, it's merged into `previewUrl.initial`:

```diff
presentationTool({
  previewUrl: {
-    origin: 'https://example.com'
+    initial: 'https://example.com'
  }
})
```

##### `previewUrl.preview`

The naming of this param is confusing. The original intention is that it matches the name of the URL search param used by Presentation to restore your previous session. It's the initial value. So the new option to use is called `initial`:

```diff
presentationTool({
  previewUrl: {
-    preview: '/blog'
+    initial: '/blog'
  }
})
```

If you use both `origin` and `preview`, move both into `initial`:

```diff
presentationTool({
  previewUrl: {
-    origin: 'https://example.com'
-    preview: '/blog'
+    initial: 'https://example.com/blog'
  }
})
```

##### async `previewUrl`

This option is undocumented and have very little real world usage. Since it tries to resolve everything in a single step up front it's becoming increasingly difficult to maintain as we've added options like Content Releases with custom perspectives, as well as new features like URL preview sharing and now multiple origins. People using it should migrate to the new async APIs:

```ts
// Before
presentationTool({
  previewUrl:  async ({
    client,
    previewUrlSecret,
    studioPreviewPerspective,
  }) => {
    // Custom async logic that changes how the URL resolves
    // ...
    return `https://example.com/api/draft-mode/enable?sanity-preview-secret=${previewUrlSecret}&sanity-preview-perspective=${studioPreviewPerspective}`
  },
})
// After
presentationTool({
  previewUrl: {
    initial: async ({client, origin}) => 'https://example.com',
    previewMode: async ({client, origin, targetOrigin}) => ({enable: '/api/draft-mode/enable'})
  }
})
```

##### 

### Testing

While moving internal state logic away from `useEffect` hell and over to `xstate`, I've made an effort to have good test coverage on new code.

The state machine is comprehensive, [with detailed test coverage](https://github.com/sanity-io/sanity/blob/4ecbbd1c83499279c427acc575fa1a9012eedcbd/packages/sanity/src/presentation/machines/__tests__/preview-url.test.ts#L1):
image

The Test Studio also has two new workspaces to test the URL resolving functionality:
- [Presentation with the Preview Kit test bed](https://test-studio-git-refactor-presentation-state.sanity.dev/presentation-preview-kit/presentation) which demonstrates patterns for switching between origins
- [Presentation with the Next Sanity test bed](https://test-studio-git-refactor-presentation-state.sanity.dev/presentation-next-sanity/presentation) which shows the common case where the app performs a redirect on load and Presentation now reaches eventual consistency.

### Notes for release

See above 😮‍💨


packages/sanity/src/presentation/actors/resolve-initial-url.ts
```diff
@@ -0,0 +1,65 @@
...
+    async ({input}: {input: {previewSearchParam: string | null}}) => {
+      const {origin} = location
+      /**
+       * If the previewUrlOption is a function, we need to resolve it and maintain backwards compatibility
+       */
+      if (typeof previewUrlOption === 'function') {
+        const initial = await previewUrlOption({
+          client,
+          studioBasePath,
+          // @TODO handle checking permissions here, and then generating a secret
```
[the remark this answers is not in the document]
RitaDias: What are the next steps for this? I presume there is a linear story for this as well? :)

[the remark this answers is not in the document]
stipsan: Next steps will be to make a linear task 😬

packages/sanity/src/presentation/overlays/schema/PostMessageSchema.tsx
```diff
@@ -75,6 +75,7 @@ function PostMessageSchema(props: PostMessageSchemaProps): React.JSX.Element | n
           const arr = Array.from(paths)
           const projection = arr.map((path, i) => `"${i}": ${path}[0]._type`).join(',')
           const query = `*[_id == $id][0]{${projection}}`
+          // Should implement max 25 concurrent queries here
```
RitaDias, line 78: 😄 
Is this a follow up as well or something that should go on this?

stipsan:
> I'm still reviewing this so bear in mind with me as I go through the files to get a good footing on the changes. A question that came to me while reading the description: beyond making sure this info is in the change log (which I think it makes sense), is there any other docs that would need to be updated / we should be aware of?
> 
> 
> 
> I'm specifically thinking about this because some parts seem to be breaking changes to users using the presentation from a config perspective and should likely be given a heads up in our proper docs about presentation if we have them
> 
> 
> 
> image
> 
> 

Gotcha, these are deprecations though so they're not breaking :)

RitaDias (commented):
I think it would be so interesting if you could do a short talk to us (or maybe just on the demo meeting idk what would be the best space for it) about how you used the xstate to check for surface (as well as docs). I think it's so neat (But also just a lot:tm:) 👏

packages/sanity/src/presentation/preview/Preview.tsx
```diff
@@ -265,19 +266,135 @@ export const Preview = memo(
...
+    const [checkOrigin, setCheckOrigin] = useState<false | string>(false)
+    const [reportedMismatches] = useState(new Set<string>())
+    const reportMismatchingOrigin = useEffectEvent((reportedOrigin: string) => {
+      if (allowOrigins.some((allow) => allow.test(reportedOrigin))) {
+        setCheckOrigin(reportedOrigin)
+        return
+      }
+      if (reportedMismatches.has(reportedOrigin)) return
+      reportedMismatches.add(reportedOrigin)
+      console.warn('Visual Editing is here but misconfigured', {reportedOrigin})
```
RitaDias, line 291: interesting, we don't really make use of console warns in the studio like this though we might want to move forward. Is this something we really want to leave available?

packages/sanity/src/presentation/preview/Preview.tsx
```diff
@@ -265,19 +266,135 @@ export const Preview = memo(
           'startViewTransition' in document &&
           typeof document.startViewTransition === 'function'
         ) {
-          document.startViewTransition(() => flushSync(() => update()))
+          document.startViewTransition({
+            // @ts-expect-error - fix typings
```
RitaDias, line 270: Why must we expect an error here? 🤔

packages/sanity/src/presentation/preview/IFrame.tsx
```diff
@@ -1,7 +1,75 @@
...
+          return
+        }
+
+        instance.dispatchEvent(new MouseEvent('mousedown', {bubbles: true, cancelable: true}))
+      }
+      window.addEventListener('blur', handleBlur)
+      return () => {
+        window.removeEventListener('blur', handleBlur)
+      }
+    }, [])
```
RitaDias, line 49: Might end up being the same thing, but couldn't we use the useClickOutside (or something along those lines) from sanity ui?

packages/sanity/src/presentation/preview/PreviewHeader.tsx
```diff
@@ -1,13 +1,15 @@
+/* eslint-disable no-nested-ternary */
```
RitaDias, line 1: Should we keep this? Might want to re-write ifs as to not rely on ternaries?

RitaDias (approved):
I left a couple of comments but nothing blocking, also with the addition of the tests I feel comfortable giving a green :)
Bjørge is on holiday, if you prefer to have someone else also have a look I can also add them as reviewers. Otherwise, enjoy your break!

Great work!

PR #8901, opened 2025-03-12
fix(i18n): remove whitespace from tag

bjoerge:
### Description
Fixes an error caused by whitespace inside at tag in a i18n string

### Notes for release
- Fixes an error caused by whitespace inside at tag in a i18n string


RitaDias (approved):
boop!

PR #8902, opened 2025-03-12
feat(sanity): refine release toasts

juice49:
### Description

This branch continues the work of refining toasts, focusing on Content Releases. Primarily, this means removing success toasts when it's apparent in the UI that the operation was successful.

### What to review

Do all the removals make sense? The commits are atomic, so it's probably easiest to look over them to see the individual changes.

### Testing

Tested in browser. Existing tests succeed.

### DIFF
--- a/packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleasePublishAllButton.tsx
+++ b/packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleasePublishAllButton.tsx
@@ -67,21 +67,6 @@ export const ReleasePublishAllButton = ({
       setPublishBundleStatus('publishing')
       await publishRelease(release._id)
       telemetry.log(PublishedRelease)
-      toast.push({
-        closable: true,
-        status: 'success',
-        title: (
-          <Text muted size={1}>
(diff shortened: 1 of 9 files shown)


RitaDias (approved):
✅

packages/sanity/src/core/releases/tool/components/releaseCTAButtons/ReleasePublishAllButton.tsx
RitaDias, line 1: This is not a blocker, I find this one to be the one that I thought might make most sense to keep, but this is a rehash of the discussion we've had in previous PRs. Not a blocker, just a thought that success messages should probably remain in place, not married to this though 🤷

juice49: Thanks for pointing it out, Rita. I do share your concern. Removing it is consistent with removing the archive/unarchive success toasts, where the success state is reflected in the view in the same way that publishing the doc is.

I'm going to merge with the removal for now, but I think we should keep it in mind in case it creates confusion down the line.

PR #8904, opened 2025-03-12
refactor(preview): clean up preview with fallback logic

bjoerge:
### Description

Fixes a couple of issues:

#### Before
image
#### After
image

#### Before

image
#### After
image

### What to review
Took the opportunity to clean up and document the logic related to getting preview values and fallback
Also fixed a bug that leaked document.title in previews for cases where the document was missing.

### Testing
Switching between release/draft/published should use the correct preview for search results (global search, reference search), document lists etc.

### Notes for release
- Fixed an issue where list previews sometimes did not reflect the current release.

### DIFF
--- a/packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
+++ b/packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
@@ -13,7 +13,10 @@ function getMissingDocumentFallback(document: Partial<SanityDocument> | PreviewV
   }
 }
 
-export type Values = {
+/**
+ * @internal
+ */
+export type Sources = {
   /**
    * Preview snapshot with current perspective applied
    * This takes priority of original and fallback
@@ -38,6 +41,6 @@ const EMPTY: {[key: string]: never} = {}
  *
  * @internal
  */
-export function getPreviewValueWithFallback({snapshot, original, fallback}: Values) {
+export function getPreviewValueWithFallback({snapshot, original, fallback}: Sources) {
   return snapshot || original || getMissingDocumentFallback(fallback || EMPTY)
 }
--- a/packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
+++ b/packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
@@ -1,8 +1,7 @@
 import {WarningOutlineIcon} from '@sanity/icons'
 import {type PreviewValue, type SanityDocument} from '@sanity/types'
-import {assignWith} from 'lodash'
 
-export function getMissingDocumentFallback(document: Partial<SanityDocument> | PreviewValue) {
+function getMissingDocumentFallback(document: Partial<SanityDocument> | PreviewValue) {
   return {
     title: <em>{document.title ? String(document.title) : 'Missing document'}</em>,
     subtitle: (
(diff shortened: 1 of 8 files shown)


packages/sanity/src/structure/components/paneItem/PaneItemPreview.tsx
```diff
@@ -84,7 +84,7 @@ export function PaneItemPreview(props: PaneItemPreviewProps) {
 
   return (
     <SanityDefaultPreview
-      {...getPreviewValueWithFallback({document, snapshot, original})}
+      {...getPreviewValueWithFallback({snapshot, original, fallback: value})}
```
bjoerge, line 87: previously `document` would refer to the global document here 🙈

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
```diff
@@ -14,25 +13,31 @@ export function getMissingDocumentFallback(document: Partial<SanityDocument> | P
...
+  snapshot?: Partial<SanityDocument> | PreviewValue | null | undefined
+  /**
+   * Preview of the original document (e.g. without current perspective applied)
+   */
+  original?: Partial<SanityDocument> | PreviewValue | null | undefined
+  /**
+   * last resort fallback in case we don't have anything to preview
+   * this can be a hard-coded preview value, or a document stub
+   */
+  fallback?: Partial<SanityDocument> | PreviewValue
```
bjoerge, line 30: hoping this makes it slightly more clear that you can pass the document and/or a hard coded preview stub as fallback

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
```diff
@@ -14,25 +13,31 @@ export function getMissingDocumentFallback(document: Partial<SanityDocument> | P
   }
 }
 
+export type Values = {
```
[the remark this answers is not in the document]
juice49: `PreviewSources`?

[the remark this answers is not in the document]
bjoerge: I like that! Aligns with the `source` parameter(s) of [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#parameters)

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
juice49: Seems like a nice improvement. I wondered whether it might be even better to use a union to ensure at least one value is passed, but upon further consideration, I'm not sure that's possible or desired.

bjoerge: Thought about it briefly, but that would be a union of three objects, each with one of these keys required and the other two optional, right? I concluded the added complexity wasn't justified, but happy to reconsider (but I'd prefer a follow-up PR for that).

packages/sanity/src/structure/components/paneItem/PaneItemPreview.tsx
juice49: Great catch! Thank you! 🙇

juice49 (dismissed):
Thanks for tackling this! It looks good to me when testing in browser.

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
juice49: I agree with you.

packages/sanity/src/core/preview/utils/getPreviewValueWithFallback.tsx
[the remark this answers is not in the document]
bjoerge: Fixed in 4653034, thank you!

juice49 (approved):

PR #8908, opened 2025-03-12
test(e2e): set up display documents e2e 2

RitaDias:
### Description

> [!NOTE]  
> The code is the same as the previous PR I had open, I just moved the tests to a single file

Added e2e tests to test document displayed from #8759
Same code as #8828 with minor changes

In order to reduce flake:
- Tests for the releases are kept in a serial order to make sure BeforeAll and AfterAll behave in the manner expected 
- Skip firefox to avoid flakiness 

### What to review

Do the tests make sense?

### Notes for release

N/A


test/e2e/tests/releases/displayDocument/DisplayedDocument.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+  let versionDocumentTwo: SanityDocument
+
+  let asapReleaseId: string
+  let undecidedReleaseId: string
+
+  let multipleVersionsDocId: string
+  let singleASAPVersionDocument: SanityDocument
+
+  test.beforeAll(async ({sanityClient, _testContext, browserName}) => {
+    test.skip(browserName === 'firefox', 'Skip test outside of chromium due to flakiness')
```
bjoerge, line 42: Since it's important that this logic is the same everywhere, maybe we can configure this in the global scope? As it is now It's a little too easy for comfort to accidentally skip a test, but forget to skip a cleanup function and vice versa. Suggestion:
```ts

// skip firefox due to flakyness
const SKIP_BROWSERS =['firefox']
```
and then instead update these conditions to say:
```ts
test.skip(SKIP_BROWSERS.includes(browserName), `Skip ${browserName} due to flakiness`)
```

test/e2e/tests/releases/displayDocument/DisplayedDocument.spec.ts
```diff
@@ -0,0 +1,445 @@
...
+  test.afterAll(async ({sanityClient, browserName}) => {
+    test.skip(browserName === 'firefox', 'Skip test outside of chromium due to flakiness')
+    const dataset = sanityClient.config().dataset
+
+    await Promise.all([
+      archiveAndDeleteRelease({sanityClient, dataset, releaseId: asapReleaseId}),
+      archiveAndDeleteRelease({sanityClient, dataset, releaseId: undecidedReleaseId}),
+    ])
+
+    //await deleteAllReleases({sanityClient, dataset})
```
bjoerge, line 107: Should this be here? If so, maybe add a comment saying what it's for?

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+  releaseId,
+}: {
+  sanityClient: SanityClient
+  dataset: string | undefined
+  releaseId: string
+}) => {
+  const query = `*[_type == "system.release" && _id == "_.releases.${releaseId}"][0]`
+  const release = await sanityClient.fetch(query)
+
+  if (release.state === 'archived') {
```
bjoerge, line 45: Is this an error situation that we should throw in? Doesn't seem like we should be calling `archiveRelease` on an already archived release.

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+    return
+  }
+
+  if (release) {
```
bjoerge, line 49: Same here – I might be missing something, but seems like this should be treated as an error?

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+  sanityClient,
+  dataset,
+  releaseId,
+}: {
+  sanityClient: SanityClient
+  dataset: string | undefined
+  releaseId: string
+}) => {
+  await archiveRelease({sanityClient, dataset, releaseId})
+  await waitForReleaseToBeArchived({sanityClient, releaseId})
```
bjoerge, line 99: Very nice :)

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+export const createDocument = (
+  sanityClient: SanityClient,
+  documentMetada: Partial<SanityDocument> & {_id: string; _type: string},
+) => {
+  return sanityClient.withConfig(CLIENT_OPTIONS).create({
+    ...documentMetada,
+  })
+}
+
+export const deleteAllReleases = async ({
```
bjoerge, line 144: When are we calling this? Seems like something that's useful when debugging e2e tests? Maybe a quick comment about what it's for and when to use it would be nice.

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+  sanityClient,
+  dataset,
+}: {
+  sanityClient: SanityClient
+  dataset: string | undefined
+}) => {
+  const query = `*[_type == "system.release"]`
```
bjoerge, line 151:
```suggestion
  const query = `*[_type == "system.release"]{_id}`
```
(since we only care about the _id anyway)

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+  releaseId: string
+  interval?: number
+  timeout?: number
+}) {
+  const startTime = Date.now()
+
+  return new Promise<void>((resolve, reject) => {
+    // eslint-disable-next-line consistent-return
+    const checkStatus = async () => {
+      const query = `*[_type == "system.release" && _id == "_.releases.${releaseId}"][0]`
```
bjoerge, line 191:
```suggestion
      const query = `*[_type == "system.release" && _id == "_.releases.${releaseId}"][0] {state}`
```
Adds a projection on `state` since that's all we care about here :)

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+ * @returns a random string of letters and numbers (9 characters long)
+ * @internal
+ */
+export function getRandomReleaseId() {
+  return Math.random().toString(36).slice(2)
+}
+
+export const createDocument = (
+  sanityClient: SanityClient,
+  documentMetada: Partial<SanityDocument> & {_id: string; _type: string},
```
bjoerge, line 137: typo: should be metadata?

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -0,0 +1,211 @@
...
+export const archiveRelease = async ({
+  sanityClient,
+  dataset,
+  releaseId,
+}: {
+  sanityClient: SanityClient
+  dataset: string | undefined
+  releaseId: string
+}) => {
+  const query = `*[_type == "system.release" && _id == "_.releases.${releaseId}"][0]`
```
bjoerge, line 42:
```suggestion
  const query = `*[_type == "system.release" && _id == "_.releases.${releaseId}"][0] {state}`
```
(if we really need to query for state first)

bjoerge (changes requested):
This is a great addition, thank you for @RitaDias! Left some comments, mostly minor stuff, please take a look. Would be great if we could avoid the duplication in the `test.skip()` condition.

test/e2e/tests/releases/displayDocument/DisplayedDocument.spec.ts
RitaDias: 🤦 forgot to remove this, you saw nothing 🥷 
sorry about the noise

test/e2e/tests/releases/displayDocument/DisplayedDocument.spec.ts
RitaDias: Good point, done

test/e2e/tests/releases/utils/methods.ts
RitaDias: I think this was part of a previous attempt at debug. Let me come back to you and if I don't have a good answer I'll remove it 👍

test/e2e/tests/releases/utils/methods.ts
RitaDias: Yes this was definitely used for a debugging, I'll add a comment to this!

test/e2e/tests/releases/utils/methods.ts
RitaDias: Ah this was a failed copy paste, will update

test/e2e/tests/releases/utils/methods.ts
RitaDias: I've added a try catch to here instead!

test/e2e/tests/releases/utils/methods.ts
RitaDias: Removed, thank you for catching this 🙏

RitaDias:
@bjoerge moved the skip condition to its own method, call it wherever it's needed :)

test/e2e/tests/releases/displayDocument/DisplayedDocument.spec.ts
```diff
@@ -21,36 +21,43 @@ import {
...
+  let multipleVersionsDocId: string
+  let singleASAPVersionDocument: SanityDocument
 
+  /** releases */
   let asapReleaseId: string
   let undecidedReleaseId: string
 
-  let multipleVersionsDocId: string
-  let singleASAPVersionDocument: SanityDocument
+  const skipIfFirefox = (browserName: string) => {
```
bjoerge, line 47: nit: but it's not really checking _firefox_, but whatever is in `SKIP_BROWSERS`

test/e2e/tests/releases/utils/methods.ts
```diff
@@ -59,6 +52,9 @@ export const archiveRelease = async ({
         ],
       },
     })
+  } catch (error) {
+    console.error('Error archiving release:', error)
+    throw error
```
bjoerge, line 57: Not sure why we need to both output and throw – throwing will surface the error anyway, so this seems redundant?

RitaDias: That's true, I was thinking about doing just the console log but then remembered about the throw and forgot to remove the console log. I'll just leave the throw.

bjoerge: then you can just remove the try catch altogether :)

RitaDias: haha was just about to comment that 😂

PR #8913, opened 2025-03-12
fix(core): updating document version actions to match deviated API

jordanl17:
### Description
Some of the actions used in `createReleaseOperationsStore` (which is used by `useReleaseOperations` and `useVersionOperations`) have drifted from the actual APIs. 

This PR adjusts the types and updates the uses of these actions according to [the version actions](https://www.sanity.io/docs/http-actions#8c43ee58aa51) and [the release actions](https://www.sanity.io/docs/http-actions#a102448a3a7a) from content lake.

Additionally, since `version.create` is now supported by content lake, `handleCreateVersion` uses the version action rather than creating a document directly with the version `_id`.

### Testing
Tests updated for `createReleaseOperationsStore.test`

Manual testing trying to use most of the actions that have been impacted

### Notes for release
Creating a new version of a document uses the [Sanity API](https://www.sanity.io/docs/http-actions#0511717e46fe) `version.create` action.

### DIFF
--- a/packages/sanity/src/core/releases/store/__tests__/createReleaseOperationsStore.test.ts
+++ b/packages/sanity/src/core/releases/store/__tests__/createReleaseOperationsStore.test.ts
@@ -205,20 +205,38 @@ describe('createReleaseOperationsStore', () => {
         },
       })
 
-      expect(mockClient.create).toHaveBeenNthCalledWith(
-        1,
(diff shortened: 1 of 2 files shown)


pedrobonamin (approved):
Looks good to me, thank you Jordan!

PR #8920, opened 2025-03-13
fix: version operations error toast

jordanl17:
### Description
| Before | After |
|--------|--------|
| Screenshot 2025-03-13 at 11 43 27 | Screenshot 2025-03-13 at 11 42 21 | 

Before: the error was caught in `useVersionOperations` and so callers would handle operation like it had resolved.
After: `useVersionOperations` delegates to the caller to show all success and error toast

### What to review
* Does it make sense to move the toast showing to the caller and away from the operations store hook (this is how `useReleaseOperations`/`createReleaseOperationsStore` also works)?

### Testing
Manually tested and verified to be working against failure and success version operations

### DIFF
--- a/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
@@ -1,4 +1,4 @@
-import {Box, Stack, Text} from '@sanity/ui'
+import {Box, Stack, Text, useToast} from '@sanity/ui'
 import {useCallback, useState} from 'react'
 
 import {Dialog} from '../../../../ui-components'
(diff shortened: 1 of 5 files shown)


bjoerge (commented):
Good catch!

> Does it make sense to move the toast showing to the caller and away from the operations store hook (this is how useReleaseOperations/createReleaseOperationsStore also works)?

Yes, I think this is a good general principle to follow.

Looks like some unit tests are failing, not sure if related to these changes?

jordanl17:
> Good catch!
> 
> > Does it make sense to move the toast showing to the caller and away from the operations store hook (this is how useReleaseOperations/createReleaseOperationsStore also works)?
> 
> Yes, I think this is a good general principle to follow.
> 
> Looks like some unit tests are failing, not sure if related to these changes?

Thanks for flagging @bjoerge just needed a rebase

bjoerge (approved):

PR #8921, opened 2025-03-13
feat(sanity): use time input in calendar

juice49:
### Description

This branch switches from an hour and a minute `select` input to a single browser-native `time` input. This provides a better user interface for selecting times.

This branch includes a little work to make the input consistent with Sanity UI, but we may want to do some more work on this in the future. Perhaps Sanity UI's `TextInput` itself should be modified to better handle `type="time"`.

| Before | After |
| --- | --- |
| Screenshot 2025-03-14 at 09 44 43 | Screenshot 2025-03-14 at 09 44 11 |

### What to review

The time selector.

### Testing

Tested in various browsers (Safari Desktop, Firefox desktop, Chrome desktop, Safari iOS). Updated unit tests.

### DIFF
--- a/packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
+++ b/packages/sanity/src/core/components/inputs/DateInputs/DateTimeInput.tsx
@@ -126,6 +126,7 @@ export const DateTimeInput = forwardRef(function DateTimeInput(
   return (
     <LazyTextInput
       ref={ref}
+      data-testid="date-input"
       {...rest}
       readOnly={disableInput || readOnly}
       value={inputValue}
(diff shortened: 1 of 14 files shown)


jordanl17:
Looks pretty good codewise, I did notice however on chrome (likely on others too), that you must unfocus first by clicking inside the popover before clicking outside of it, otherwise (if you change the time then immediately click outside popover) the change doesn't propagate)


Uploading BlurTimeInput.mov…

juice49:
@jordanl17 it's a good catch. I noted the same thing myself. This is because `LazyTextInput` only propagates the change to the provided `onChange` prop when the user presses the enter key, or blurs the input. That's a problem in the calendar UI, which itself is dismissible by clicking outside of it (as you mentioned).

I agree it's not user friendly. Wondering how we can side-step the issue.

jordanl17 (approved):

jordanl17:
> @jordanl17 it's a good catch. I noted the same thing myself. This is because `LazyTextInput` only propagates the change to the provided `onChange` function when the user presses the enter key, or blurs the input. That's a problem in the calendar UI, which itself is dismissible by clicking outside of it (as you mentioned).
> 
> I agree it's not user friendly. Wondering how we can side-step the issue.

You able to track a bug and we can at least have it on our radar

PR #8925, opened 2025-03-13
fix(structure): update archived and published docs in release links

pedrobonamin:
### Description

This PR updates how the links on published and archived releases work:
**Published releases**
Now when clicking on a document as part of a published release, instead of opening the document in the detailed view for that release it will open the published view with the history having selected that release:

https://github.com/user-attachments/assets/0132e4ff-46e6-4240-8316-c2fc1909b607

**Archived releases**
Clicking a document as part of releases that are archived we cannot show the document from the published view, we instead open the document "faking" that the release exists, and it will show the badge in the document for that release, but the badge will be disabled.

https://github.com/user-attachments/assets/3d640659-dc7a-41fe-a6fe-f923ab6ce011

This PR also fixes an issue with the open release links from the document events, now releases will open when clicking  the menu item. (as expected)
Screenshot 2025-03-13 at 18 33 18

**What it doesn't solve:**

View of documents **unpublished** by a release, it is not possible do it just now because content lake is generating the incorrect events for it, once the events are solved we will look back into solving this.

### What to review
Everything works as expected.

### Notes for release
Update archived and published docs in release links.

### DIFF
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -1202,6 +1202,8 @@ export const studioLocaleStrings = defineLocalesResources('studio', {
   'release.chip.global.drafts': 'Drafts',
   /** Label for Published chip in document header */
   'release.chip.published': 'Published',
+  /** Label for tooltip in chip when document is in an archived release */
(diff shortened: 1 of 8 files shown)


bjoerge (dismissed):
LGTM! These are great improvements, thank you!

bjoerge (approved):

PR #8926, opened 2025-03-13
test(releases): add unit tests for createReleaseMetadataAggregator and useActiveReleases hooks

jordanl17:
### Description
Some more tests to fill in gaps around the releases store and hooks

### Testing
This is just tests

### Notes for release
N/A


RitaDias (approved):
we love to see it, we love to see it 
wonder if it's worth creating a mock file for the useActiveReleases for other people to use in their tests but it's not a blocker

jordanl17:
> we love to see it, we love to see it wonder if it's worth creating a mock file for the useActiveReleases for other people to use in their tests but it's not a blocker

Yeh that's a good idea! I'll add that, thanks for the consideration

PR #8930, opened 2025-03-14
chore(core): remove unused content release diff view

pedrobonamin:
### Description
This PR removes the initial diff view we introduced in corel.
This pieces of code are not used anywhere, trying to keep the code lean :) 

### What to review
Is there any purpose on keeping this around?

### Testing
If build and existing tests passes I would say we are ok, this is not used anywhere.

### Notes for release
n/a, internal


jordanl17 (approved):
delete delete delete

PR #8932, opened 2025-03-14
fix(preview): don't apply preview fallback for documents in archived release

bjoerge:
### Description
After #8904 we ended up passing the fallback value to `prepareForPreview` which caused an error because react isn't valid as preview attributes. 

### What to review
The document previews for archived releases should display the right values

### Notes for release
n/a - fixes an earlier unreleased bug


pedrobonamin (approved):
Thank you!

PR #8933, opened 2025-03-14
fix(structure): some touches to version chips

pedrobonamin:
### Description
Changes the borders for the version chips.
Removes the background for disabled chips and instead uses muted colors.

Last updates:
Screenshot 2025-03-14 at 13 33 13
Screenshot 2025-03-14 at 13 37 31


RitaDias (approved):
✅

PR #8943, opened 2025-03-14
feat(studio): detect and report high listener roundtrip latency

bjoerge:
### Description
We want to track e2e listener latency, this reports a telemetry event every time we detect >1s latency from after an edit action request has completed until we receive an event with a corresponding transaction id back from the listener.

### What to review

I've tried to implement this as cleanly as possible to make it easy to delete the code path for tracking latency at some point in the future. The implementation follows the same pattern as the "document desynced" event in the document store takes a callback that will be called with latency, shard identifier and transaction id.

### Testing
To see that the the Studio sends the event, start with `SANITY_STUDIO_DEBUG_TELEMETRY=true pnpm dev` which will log the telemetry events to the console.

This, combined with network throttling should likely cause some "High Listener Latency Detected"-events to be reported

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/core/store/_legacy/datastores.ts
+++ b/packages/sanity/src/core/store/_legacy/datastores.ts
@@ -29,6 +29,11 @@ import {useResourceCache} from './ResourceCacheProvider'
 import {createUserStore, type UserStore} from './user'
 
 /**
+ * Latencies below this value will not be logged
+ */
+const IGNORE_LATENCY_BELOW_MS = 1000
+
+/**
  * @hidden
  * @beta */
 export function useUserStore(): UserStore {
@@ -158,7 +163,7 @@ export function useDocumentStore(): DocumentStore {
 
   const handleReportLatency = useCallback(
     (event: LatencyReportEvent) => {
-      if (event.latencyMs > 1000) {
+      if (event.latencyMs > IGNORE_LATENCY_BELOW_MS) {
         telemetry.log(HighListenerLatencyOccurred, {
           latency: event.latencyMs,
           shard: event.shard,
--- a/packages/sanity/src/core/store/_legacy/datastores.ts
+++ b/packages/sanity/src/core/store/_legacy/datastores.ts
@@ -16,8 +16,9 @@ import {
   type ConnectionStatusStore,
   createConnectionStatusStore,
 } from './connection-status/connection-status-store'
-import {createDocumentStore, type DocumentStore} from './document'
+import {createDocumentStore, type DocumentStore, type LatencyReportEvent} from './document'
 import {DocumentDesynced} from './document/__telemetry__/documentOutOfSyncEvents.telemetry'
+import {HighListenerLatencyOccurred} from './document/__telemetry__/listenerLatency.telemetry'
 import {fetchFeatureToggle} from './document/document-pair/utils/fetchFeatureToggle'
 import {type OutOfSyncError} from './document/utils/sequentializeListenerEvents'
 import {createGrantsStore, type GrantsStore} from './grants'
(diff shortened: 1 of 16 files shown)


packages/sanity/src/core/store/_legacy/document/__telemetry__/listenerLatency.telemetry.ts
```diff
@@ -0,0 +1,12 @@
+import {defineEvent} from '@sanity/telemetry'
+
+type Samples = {
+  latency: number
+  transactionId: string
+  shard?: string
+}
+export const HighListenerLatencyOccurred = defineEvent<Samples>({
+  name: 'High Listener Latency Detected',
```
svirs, line 9: 👍

packages/sanity/src/core/store/_legacy/datastores.ts
```diff
@@ -155,6 +156,19 @@ export function useDocumentStore(): DocumentStore {
     [telemetry],
   )
 
+  const handleReportLatency = useCallback(
+    (event: LatencyReportEvent) => {
+      if (event.latencyMs > 1000) {
```
sgulseth, line 161: Move into constant(Could belong with the telemetry event? 🤔) and maybe describe why and what this number is? As a reference and help for our future selves 🙃

sgulseth (dismissed):
LGTM from rxjs perspective, but I have limited insight into general behavior in studio paths 🙂

packages/sanity/src/core/store/_legacy/datastores.ts
bjoerge: Good call! Extracted to a const in [5f918c7](https://github.com/sanity-io/sanity/commit/5f918c7a6e7ab8b915ee544d2868ea4801218602)

packages/sanity/src/core/store/_legacy/document/document-pair/checkoutPair.ts
```diff
@@ -256,22 +304,89 @@ export function checkoutPair(
...
 }
+
+function reportLatency(options: {
+  commits$: Observable<MultipleActionResult | MutationResult>
+  listenerEvents$: Observable<ListenerEvent>
+  client: SanityClient
+  onReportLatency: (event: LatencyReportEvent) => void
+}) {
+  const {client, commits$, listenerEvents$, onReportLatency} = options
+  const shardInfo = fetch(client.getUrl(client.getDataUrl('ping')))
```
juice49, line 338: The only thing that jumps out to me is that the telemetry event could be lost when the latency is high due to client network conditions, rather than issues within our infrastructure.

1. The shard `ping` fetch looks like it's intended to fail gracefully, but how long can it be in-flight before the client times it out? Even though it will fail gracefully, it does also create more network traffic for clients who may be experiencing high-latency at their end. It's possible the user leaves before getting to `telemetry.log`?
2. Does `telemetry.log` buffer events to persistent browser storage so they can be logged the next time the user connects?

These are probably non-issues seeing as we care about listener latency at our end, but just thought I'd mention it as it seems like it could skew data collected against client-side latency issues.

There is also maybe a performance concern with throwing another HTTP request at the client (shard ping) when network conditions are poor.

bjoerge: We only care about latency under normal circumstances, so in a way dropping these events due to poor network conditions is actually more a feature than anything.

Good point about timeout – I've added a 20s timeout in 51d4792d34, and added a clarifying comment about the shard ping request (it should only be done once, so it's effect should be neglible)

juice49: Cool! Thanks for confirming!

juice49 (approved):
Thanks, this looks great! I was able to see the telemetry logs client-side as expected during my testing.

PR #8949, opened 2025-03-15
fix(cli): fix import for SanityApp and add fallback prop

kenjonespizza:
### Description

This fixes a a broken import of `SanityApp` in the "core-app" template

Screenshot 2025-03-14 at 22 50 33

It also adds the required `fallback` prop when using the `<SanityApp>` component.

I also added in some styles to improve the look of the login screen.  See image below with the left screen being the before

image

### What to review

`SanityApp` appears to be exported here https://github.com/sanity-io/sdk/blob/main/packages/react/src/_exports/index.ts , not directly from `src/components`.  

### Testing

The template can be initiated and ran with:
```
pnpx sanity@latest init --template core-app
cd my-content-os-app
npm run dev
```

You can create a test project/template and copy/paste in my code updates pretty easily to test

### DIFF
--- a/packages/@sanity/cli/templates/core-app/src/App.css
+++ b/packages/@sanity/cli/templates/core-app/src/App.css
@@ -3,7 +3,9 @@
   max-width: 1200px;
   margin: 0 auto;
   padding: 2rem;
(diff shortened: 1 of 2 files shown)


binoy14:
This is already changed. Closing this PR

PR #8950, opened 2025-03-15
fix(typegen): handle duplicate type names

sgulseth:
### Description

We require each query to be unique, but we normalize the schema type names. Ie `foo.bar` becomes `FooBar`, if you then also have another type `fooBar` that would also become `FooBar`. This change ensures that we properly increment the suffix 

### What to review

Did I miss anything? 

### Testing

Added

### Notes for release

TypeGen: Fixes a rare bug where generated type names could collide and wasn't handled correctly.


stipsan (dismissed):

sgulseth:
Had to rebase so also had to re-request review @stipsan sorry about that

PR #8963, opened 2025-03-17
feat(telemetry): track INP performance from the field

bjoerge:
### Description
Collects [INP](https://web.dev/articles/inp) (Interaction to Next Paint) telemetry from the field.

This adds a performance observer that samples INP at most every minute. In addition to the INP duration, we also log the interaction event target. Example payload:

```json
{
  "target": "input#myNumberField.Input-sc-h62wco-4.evDwFo",
  "attrs": {"testId":"number-input"},
  "interaction": "keydown",
  "duration": 160
}
```

### What to review
- Does the event definition look sane?
- Is the sample rate of 1 minute acceptable?

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/core/studio/telemetry/useMeasurePerformanceTelemetry.ts
+++ b/packages/sanity/src/core/studio/telemetry/useMeasurePerformanceTelemetry.ts
@@ -3,6 +3,15 @@ import {useCallback, useEffect} from 'react'
 
 import {PerformanceINPMeasured} from '../__telemetry__/performance.telemetry'
 
+function getInterestingAttrs(node: Node | null): undefined | {ui?: string; testId?: string} {
+  if (!node || !(node instanceof Element)) {
+    return undefined
+  }
+  const ui = node.getAttribute('data-ui') || undefined
+  const testId = node.getAttribute('data-testid') || undefined
+  return ui || testId ? {ui, testId} : undefined
+}
+
 function getElementIdentifier(node: Node | null) {
   if (!node) {
     return null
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/studio/telemetry/useMeasurePerformanceTelemetry.ts
```diff
@@ -0,0 +1,65 @@
...
+  if (!(node instanceof Element)) {
+    return node.nodeName
+  }
+  // Note: Deliberately using classList instead of className here since className isn't always a string
+  // See https://developer.mozilla.org/en-US/docs/Web/API/Element/className#notes
+  const {nodeName, classList, id} = node
+  return (
+    nodeName.toLowerCase() +
+    (id ? `#${id}` : '') +
+    (classList ? `.${classList.value.replaceAll(' ', '.')}` : '')
```
stipsan, line 19: I think I'd collect these two data attributes as well:
- `data-testid`
- `data-ui`

As the `classList` coming from `styled-components` can be botched/unusable if code minification is enabled (it is by default in `sanity build` and `next build`). And `id` sometimes might be using `useId()` or some other source that is difficult to trace back.

The `data-testid` is super high value since it's already used for E2E and gives useful context as to where and what the component is.
Same is true for `data-ui` as it's used by `@sanity/ui`, and we already know that primitives like `<Tooltip>`, `<Popover>` and `<Input>` are already hot paths.

bjoerge: Good point. This was inspired by how it's reported in the Performance devtools panel, but agree data-test-id and data-ui is valuable. I do wonder if we want id and classlist at all, but feels like _some_ hints is better than none? I'll add dataUi and testId as additional props.

image

stipsan (approved):
Brilliant 💖

svirs (approved):

PR #8964, opened 2025-03-17
chore(tests): add vision e2e tests

pedrobonamin:
### Description
Adds e2e tests for vision.
This branch https://github.com/sanity-io/sanity/pull/8879 is changing vision to a functional component.
So having some e2e tests for the basics functionality before merging the refactor is a nice thing to do and get some extra security.

### DIFF
--- a/test/e2e/tests/vision/utils.ts
+++ b/test/e2e/tests/vision/utils.ts
@@ -1,4 +1,4 @@
-import {type Page} from '@playwright/test'
+import {expect, type Page} from '@playwright/test'
 
 export function encodeQueryString(
   query: string,
@@ -22,16 +22,14 @@ export function encodeQueryString(
 export const openVisionTool = async (page: Page) => {
   await page.goto('/test/vision')
   // Wait for vision to be visible
-  await page.waitForSelector('[data-testid="vision-root"]', {
-    timeout: 30_000,
-  })
+  await expect(page.getByTestId('vision-root')).toBeVisible()
 }
 
 export const getVisionRegions = async (page: Page) => {
-  const queryEditorRegion = page.locator('[data-testid="vision-query-editor"]')
+  const queryEditorRegion = page.getByTestId('vision-query-editor')
   const queryEditor = queryEditorRegion.locator('.cm-content')
-  const paramsRegion = page.locator('[data-testid="params-editor"]')
+  const paramsRegion = page.getByTestId('params-editor')
   const paramsEditor = paramsRegion.locator('.cm-content')
-  const resultRegion = page.locator('[data-testid="vision-result"]')
+  const resultRegion = page.getByTestId('vision-result')
   return {queryEditorRegion, queryEditor, paramsRegion, paramsEditor, resultRegion}
 }
--- a/test/e2e/tests/vision/utils.ts
+++ b/test/e2e/tests/vision/utils.ts
@@ -1,0 +1,37 @@
+import {type Page} from '@playwright/test'
+
+export function encodeQueryString(
+  query: string,
+  params: Record<string, unknown> = {},
+  options: Record<string, string | string[]> = {},
+): string {
+  const searchParams = new URLSearchParams()
+  searchParams.set('query', query)
+
+  for (const [key, value] of Object.entries(params)) {
+    searchParams.set(`$${key}`, JSON.stringify(value))
+  }
+
+  for (const [key, value] of Object.entries(options)) {
+    if (value) searchParams.set(key, `${value}`)
+  }
+
+  return `?${searchParams}`
+}
+
+export const openVisionTool = async (page: Page) => {
+  await page.goto('/test/vision')
+  // Wait for vision to be visible
+  await page.waitForSelector('[data-testid="vision-root"]', {
+    timeout: 30_000,
+  })
+}
+
+export const getVisionRegions = async (page: Page) => {
+  const queryEditorRegion = page.locator('[data-testid="vision-query-editor"]')
+  const queryEditor = queryEditorRegion.locator('.cm-content')
+  const paramsRegion = page.locator('[data-testid="params-editor"]')
+  const paramsEditor = paramsRegion.locator('.cm-content')
+  const resultRegion = page.locator('[data-testid="vision-result"]')
+  return {queryEditorRegion, queryEditor, paramsRegion, paramsEditor, resultRegion}
+}
--- a/test/e2e/tests/vision/vision.spec.ts
+++ b/test/e2e/tests/vision/vision.spec.ts
@@ -94,9 +94,7 @@ test.describe('Vision', () => {
     // The query executes automatically when a url is pasted, so it should have results
     // Assert that the results are visible
     // It should find the book document assert that by checking the title and the id
-    await expect(resultRegion.getByText(bookTitle)).toBeVisible({
-      timeout: 10_000,
-    })
+    await expect(resultRegion.getByText(bookTitle)).toBeVisible()
     await expect(resultRegion.getByText(bookDocument._id)).toBeVisible()
   })
 
--- a/test/e2e/tests/vision/vision.spec.ts
+++ b/test/e2e/tests/vision/vision.spec.ts
@@ -1,0 +1,143 @@
+import {expect} from '@playwright/test'
+import {test} from '@sanity/test'
+
+import {encodeQueryString, getVisionRegions, openVisionTool} from './utils'
+
+test.describe('Vision', () => {
+  test('should be possible to type an execute a query', async ({page, sanityClient}) => {
+    const bookTitle = 'Test Book'
+    const bookDocument = await sanityClient.create({
+      _type: 'book',
+      title: bookTitle,
+    })
+
+    await openVisionTool(page)
+    // Clears local storage
+    await page.evaluate(() => localStorage.clear())
+
+    const {queryEditor, paramsEditor, paramsRegion, resultRegion} = await getVisionRegions(page)
+
+    // Click to focus the editor
+    await queryEditor.click()
+
+    // Type text into the CodeMirror editor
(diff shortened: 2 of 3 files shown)


test/e2e/tests/vision/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  return `?${searchParams}`
+}
+
+export const openVisionTool = async (page: Page) => {
+  await page.goto('/test/vision')
+  // Wait for vision to be visible
+  await page.waitForSelector('[data-testid="vision-root"]', {
+    timeout: 30_000,
+  })
+}
```
RitaDias, line 28: as far as I know in e2e tests for playwright it's preferrable to actually do an expect vs an await, you can see how we do it in other tests but the general is

playwright tends to be decent at expecting things to be visible and not taking too long to check. this is often enough and doesn't require a timeout :)

```suggestion
await expect(page.getByTestId('vision-root')).toBeVisible()
```

test/e2e/tests/vision/vision.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    // Assert that the text was correctly inserted
+    await expect(paramsEditor).toHaveText(paramsInputText)
+
+    // Find the button with the text "Fetch" and click it.
+    await page.locator('button').filter({hasText: 'Fetch'}).click()
+
+    // Assert that the results are visible
+    // It should find the book document assert that by checking the title and the id
+    await expect(resultRegion.getByText(bookTitle)).toBeVisible()
+    await expect(resultRegion.getByText(bookDocument._id)).toBeVisible()
```
RitaDias, line 49: Might be just a preference, but I would usually prefer that we match to a snapshot / to the text expected than whether something is visible

test/e2e/tests/vision/vision.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    // The query executes automatically when a url is pasted, so it should have results
+    // Assert that the results are visible
+    // It should find the book document assert that by checking the title and the id
+    await expect(resultRegion.getByText(bookTitle)).toBeVisible({
+      timeout: 10_000,
+    })
+    await expect(resultRegion.getByText(bookDocument._id)).toBeVisible()
+  })
+
+  test('should be possible to listen to changes', async ({page, _testContext, sanityClient}) => {
```
RitaDias, line 103: Might be nitpicking, but I don't know if this is worth being a e2e test vs a unit test for example.

test/e2e/tests/vision/vision.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    const parsedParams = JSON.parse(paramsText || '{}')
+    expect(parsedParams).toHaveProperty('id', bookDocument._id)
+
+    // The query executes automatically when a url is pasted, so it should have results
+    // Assert that the results are visible
+    // It should find the book document assert that by checking the title and the id
+    await expect(resultRegion.getByText(bookTitle)).toBeVisible({
+      timeout: 10_000,
+    })
+    await expect(resultRegion.getByText(bookDocument._id)).toBeVisible()
```
RitaDias, line 100: We should avoid relying on timeouts, sometimes they are needed but it gives me the feeling we could probably be doing this in a different way that wouldn'rt require it.

The same comment about the expectations remains as on a previous comment :)

test/e2e/tests/vision/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const getVisionRegions = async (page: Page) => {
```
RitaDias, line 30: Not sure how useful this is to have in a separate file where we might need to ping pong between the test and utils if we need to debug / update the tests at a later point. Might be worth moving it to the file it will be used even if technically in the future we might want to re-use a similar method.

Howrever, because most of these are simple finders I don't mind the repetition and we might be able to avoid querying the page needlessly on future tests and fetch simply the ones we need. Just a thought

test/e2e/tests/vision/utils.ts
pedrobonamin: Ohh right we have the `getByTestId` thank you!

test/e2e/tests/vision/vision.spec.ts
pedrobonamin: In the case for vision I think it's a great candidate for e2e testing, because it relies so much on how content lake gives us the data back and it's main functionality is to execute and listen to queries.
If we mock all the client functionality I think we might be in a middle scenario where we will be testing half of the app, also, we would need to mock a lot of things and I think it's harder to test.

My line of thought is with e2e tests is easier to cover the basis or happy path of an app and then use the unit tests for the cases which are not covered by an e2e.

This test writes the query, sets up the listener and then creates a document, asserting that the creation is caught by the listener and vision renders the value received

RitaDias: 🤔 
True, I don't disagree with this, however I would expect that this aspect would have been tested / is tested in the backend and that we wouldn't need to necessarily cover it from the front end side of things. Again, it is probably part of the discussion we need to have, in my experience e2e tests when they are most successful when they rely on fixtures instead of the live backend (except for specific use cases when you are trying to keep a pulse and check on the health, but those are tests specifically written for those objectives).

I'm fine with keeping the test, I'm just not sure if we should be the line of adding these sort of tests when it feels like it should technically covered somewhere else (or should). Just a thought though!

test/e2e/tests/vision/vision.spec.ts
RitaDias: After clarificiation, this is not needed for this particular test!

test/e2e/tests/vision/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  const queryEditorRegion = page.locator('[data-testid="vision-query-editor"]')
```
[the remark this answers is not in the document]
pedrobonamin: No reason, I forgot about the `getByTestId` in playwright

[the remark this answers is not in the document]
RitaDias: ahhh, hahaha that would do it 😄

test/e2e/tests/vision/utils.ts
pedrobonamin: I would usually agree, but in this case I think it's nice to have the regions defined and avoid repeating this in every new test that is added, specially because you need to do things like getting the region and then getting the editor, there is a bit of overhead to get the editors because of this.

I placed it in a different file to indicate that it can be reused for other vision tests and to keep the test file lean.

test/e2e/tests/vision/vision.spec.ts
pedrobonamin: Yes, for sure we have tests in the backend for how the listener works, but the implementation of how to consume the listener and how we react to the changes received from it it's defined in vision [here](https://github.com/sanity-io/sanity/blob/add-vision-e2e-tests/packages/@sanity/vision/src/components/VisionGui.tsx#L554-L602) and [here](https://github.com/sanity-io/sanity/blob/add-vision-e2e-tests/packages/@sanity/vision/src/components/VisionGui.tsx#L522-L533)

This test validate that this works as expected and the app listen and displays the events. If we want to go for a unit or integration tests here we will need to mock all the client and dispatch mocked listener events, which I think will be harder to setup and later read and debug than running the e2e test while being less secure because we need to mock all of this.

If we instead wanted to test inner component in vision, for example the **copy to clipboard** action, or the **save results**  I would say these are perfect candidates for a unit tests, but when the whole app is involved I prefer e2e

RitaDias: This makes sense 👍 
Sorry for all the questions, this is something that I myself like to use as the line of thinking for ultimately deciding what ultimately ends up as a e2e test vs not (how much of a pain will it be to mock vs writing out the e2e test). Since I don't know much about this component, I just wanted to be sure 🙏

pedrobonamin: Thanks for asking all of this questions, I think it's a nice debate to have and to land a decision as a team.
We need to define and write this things for all of us to be in the same page on the decisions we need to take for this type of thins.

bjoerge (approved):
This is great!

test/e2e/tests/vision/utils.ts
pedrobonamin: >if for some reason playwright cannot find any / one of these, even if it doesn't need it, it will still fail and say that it can't find it right?

Yes, that's right, it will fail because it is expecting that this elements exist, but playwright is never flaky getting this if the elements exist in the dom. This elements are not conditionally rendered or something like that so they should always exist.

This are the elements that this is looking for, if for some reason they don't exist I would prefer the test to fail.
Screenshot 2025-03-18 at 10 37 03

RitaDias: Maybe it's just an issue when we are in the document Pane then. Because in the document pane, just fetching things from the document takes times and often we require an expect about something in the document before we can actually start the test (to avoid flakiness due to load times)

pedrobonamin: Could that be while the document is loading? 
In those cases, if the loading hasn't finished I think it could happen that the dom is not ready and it doesn't find the element. 

In this cases, I'm checking vision is loaded in the previous step `openVisionTool` https://github.com/sanity-io/sanity/blob/add-vision-e2e-tests/test/e2e/tests/vision/utils.ts#L25

RitaDias: Yep, I assume that's the reason.
If there is a different step for checking it do you think it makes sense to have it baked into this utils? I'm just thinking from a standpoint of folks potentially trying to use this helper method and not realising they should use it in conjunction with another one

RitaDias (approved):
Thanks for adding these!

PR #8966, opened 2025-03-17
feat(cli): server side schemas (behind feature toggle)

snorrees:


snorrees:
Im leaning towards --no-extract-manifest being the CLI flag, ie --extract-manifest being the default.

snorrees:
Seems like the consensus is on --no-extract-manifest.
I'll make this change today or tomorrow; it should not impact much code though: from what I can tell, our CLI interface already does --no boolean inversion for free, so its just flipping a default value.

snorrees:
Updated PR description. The flag is now --no-extract-manifest. Tests and code no reflect that extractManifest defaults to true. Published a new tagged release (3.77.3-server-side-schemas.41).

packages/sanity/src/_internal/cli/actions/schema/utils/mainfestExtractor.ts
```diff
@@ -0,0 +1,50 @@
+import {type CliCommandArguments, type CliCommandContext, type CliOutputter} from '@sanity/cli'
+import chalk from 'chalk'
+
+import {type ExtractManifestFlags, extractManifestSafe} from '../../manifest/extractManifestAction'
+import {FlagValidationError} from './schemaStoreValidation'
+
+export type ManifestExtractor = (manifestDir: string) => Promise<void>
+
+export async function isManifestExtractSatisfied(args: {
```
bjoerge, line 9: the naming here feels slightly off. It _sounds_ like a predicate/pure function, but it's doing a lot of heavy lifting. Wondering if the name should better reflect that?

packages/sanity/src/_internal/cli/actions/schema/utils/manifestReader.ts
```diff
@@ -0,0 +1,128 @@
...
+    parsedWorkspaces[workspaceName] = result
+    return result.parsedJson
+  }
+  return {
+    getManifest,
+    getWorkspaceSchema,
+  }
+}
+
+export const getManifestDirPath = (workDir: string, customPath?: string) => {
```
bjoerge, line 99: nit (feel free to ignore): rename to `resolveManifestDirPath`? also, what is a "dir path"? sounds like a pleonasm :) maybe "outputPath" would be better?

packages/sanity/src/_internal/cli/actions/schema/utils/schemaApiClient.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function createSchemaApiClient(apiClient: CliApiClient) {
+  const client = apiClient({
+    requireUser: true,
+    requireProject: true,
+  }).withConfig({apiVersion: 'v2025-03-01', useCdn: false})
+
+  const projectId = client.config().projectId
+  const dataset = client.config().dataset
+  if (!projectId) throw new Error('Project ID is not defined')
```
bjoerge, line 11: Won't requireUser + requireProject throw if either is missing? Seems fine to use a non-null assertion on both of these

packages/sanity/src/_internal/cli/actions/schema/utils/schemaStoreValidation.ts
```diff
@@ -0,0 +1,235 @@
...
+import {type SchemaListFlags} from '../listSchemasAction'
+import {type StoreSchemasFlags} from '../storeSchemasAction'
+import {getManifestDirPath} from './manifestReader'
+import {getStringArrayOutString} from './storeSchemaOutStrings'
+
+export const validForIdChars = 'a-zA-Z0-9._-'
+export const validForIdPattern = new RegExp(`^[${validForIdChars}]+$`, 'g')
+
+const requiredInId = SANITY_WORKSPACE_SCHEMA_TYPE.replace(/[.]/g, '\\.')
+const idPattern = new RegExp(
```
bjoerge, line 16: if I'm reading this correctly, this would allow id's to both start with dashes, and also have consecutive dots, which is not supported: https://www.sanity.io/docs/ids (haven't checked if this is considered later on though)

packages/sanity/src/_internal/cli/actions/schema/utils/schemaStoreValidation.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  const commonFlags = parseCommonFlags(flags, context, errors)
+  const id = parseId(flags, errors)
+  const json = !!flags.json
+
+  assertNoErrors(errors)
+  return {...commonFlags, json, id}
+}
+
+export function parseDeleteSchemasConfig(flags: DeleteSchemaFlags, context: {workDir: string}) {
```
bjoerge, line 79: the various parser functions in here seems like good candidates for unit tests, at least the non-trivial ones :)

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,165 @@
...
+export interface StoreSchemasFlags extends StoreSchemaCommonFlags {
+  'workspace'?: string
+  'id-prefix'?: string
+  'schema-required'?: boolean
+}
+
+export default function storeSchemasActionForCommand(
+  flags: StoreSchemasFlags,
+  context: CliCommandContext,
+): Promise<'success' | 'failure'> {
```
bjoerge, line 32: this seems a bit unconventional to me – why not `Promise<void>` and take the error path on failure?

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * Workspaces are determined by on-disk manifest file – not directly from sanity.config.
+ * All schema store actions require a manifest to exist, so we regenerate it by default.
+ * Manifest generation can be optionally disabled with --no-manifest-extract.
+ * In this case the command uses and existing file or throws when missing.
+ */
+export async function storeSchemasAction(
+  flags: StoreSchemasFlags,
+  context: SchemaStoreContext,
+): Promise<'success' | 'failure'> {
```
bjoerge, line 58: (same as above)

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
snorrees: It is because of the `--schema-required` flag; when it is true, the action is throwing, when it is false, the action will never throw. 
Would you allow it with a comment? I could encapsulate as a result type, and add a tsdoc with a reason to the type?

snorrees: I'll push the union type with tscomment, and you can let me know what you think

packages/sanity/src/_internal/cli/actions/schema/utils/mainfestExtractor.ts
snorrees: Will ensureManifestExtractSatisfied fly?

packages/sanity/src/_internal/cli/actions/schema/utils/manifestReader.ts
snorrees: resolveManifestDirectory

packages/sanity/src/_internal/cli/actions/schema/utils/schemaStoreValidation.ts
snorrees: Added some additional checks: didn't touch the regex as not to complicate it further.

packages/sanity/src/_internal/cli/actions/schema/storeSchemasAction.ts
bjoerge: That makes sense, thanks for adding a comment!

packages/sanity/src/_internal/cli/actions/schema/schemaStoreTypes.ts
```diff
@@ -10,3 +10,22 @@ export interface SchemaStoreContext {
   jsonReader?: ManifestJsonReader
   manifestExtractor: ManifestExtractor
 }
+
+/**
+ * There is more context locally to decide how and when to log, depending on flags;
```
bjoerge, line 15: thanks for adding this!

bjoerge (approved):
LGTM - thank you for adding tests and clarifying comments.

PR #8973, opened 2025-03-18
feat(core/schema): extract global document references types

skogsmaskin:
### Description

This will extract `globalDocumentReference` types with `extractSchema` and. `extractManifest`.

However I'm not sure about the implications of this as I don't know too much about how those functions are used.

Please help me assess this.


skogsmaskin:
Found out that `test/cli/graphql/gen3.test` need to add tests for this type.

packages/@sanity/schema/src/sanity/extractSchema.ts
```diff
@@ -180,11 +180,6 @@ export function extractSchema(
       return {type: 'unknown'} satisfies UnknownTypeNode // we don't support cross-dataset references at the moment
     }
 
-    // Global document references are not supported
-    if (isGlobalDocumentReferenceType(schemaType)) {
```
snorrees, line 184: This will need a fix on the Create side (both backend and frontend) so we ignore gdr fields there for now.
Ie, we would have to get that out first before this can be merged.
If left as is, Create form will render this as an object with whatever fields GDR type has.

packages/sanity/test/manifest/extractManifest.test.ts
```diff
@@ -861,6 +885,24 @@ describe('Extract studio manifest', () => {
...
+                preview: {
+                  select: {
+                    media: 'coverImage',
+                    title: 'title',
+                  },
+                },
+                type: 'book',
+              },
+            ],
+            type: 'globalDocumentReference',
```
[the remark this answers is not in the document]
skogsmaskin: @snorrees - it will go out with the release today. Was merged to `next` earlier today.

(I will will make the final adjustments to the AI-Assist Plugin btw as soon as I know the version number)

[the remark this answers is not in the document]
skogsmaskin: @snorrees - to clarify my message above:

The unexposed GDR support will go out today, not this one :)

> Ie, we would have to get that out first before this can be merged.

Yes. Noted.

packages/sanity/test/cli/graphql/__snapshots__/gen3.test.ts.snap
```diff
@@ -1631,6 +1631,10 @@ exports[`GraphQL - Generation 3 > Should be able to generate graphql schema 1`]
           "fieldName": "cdrFieldNamed",
           "type": "CdrPersonReference",
         },
+        {
+          "fieldName": "productReference",
+          "type": "GlobalDocumentReference",
```
sgulseth, line 1636: afaict `GlobalDocumentReference` isn't declared anywhere? 🤔

sgulseth (changes requested):

packages/sanity/test/cli/graphql/__snapshots__/gen3.test.ts.snap
snorrees: I think its a new core type. 

@skogsmaskin its too lay to remodel GDRs as just objects right? This was why assist blew up too right? And why auto updating studios with assist are gonna be in a bad way.

justshiv:
Hey @skogsmaskin what's the current status on this PR? Is it worth getting this into a shippable state or can we close this?

packages/sanity/test/cli/graphql/__snapshots__/gen3.test.ts.snap
skogsmaskin: Yeah; it's similar to ["type": "CrossDatasetReference"](https://github.com/sanity-io/sanity/blob/feat/gdr-extract-schema/packages/sanity/test/cli/graphql/__snapshots__/gen2.test.ts.snap#L1533), which is how the GQL schema types end up loooking.

skogsmaskin:
@snorrees - I see you have added some stuff touching upon this some time after this PR. Can you please have another look?

@kristofferlund - I know you have been working lately with the schema types. Adding you as an reviewer on this.

skogsmaskin:
> Hey @skogsmaskin what's the current status on this PR? Is it worth getting this into a shippable state or can we close this?

Yes, if we want correct typings on the global document reference type (schema and GraphQL), it's worth getting this in.

It is up to date with `main` now. My first comment still stands though, I don't know the implications of this change (in code elsewhere), so I need help to review.

skogsmaskin:
> I don't think this is what we can support at the moment. typegen is not aware of GDRs, and has no way of resolving the external schema. I think we first need to migrate typegen to use the new schema store. For now `unknown` is the only thing reasonable to return for GDRs

Right! Sounds like something that could take a while to get in place.

@justshiv - feel free to close this if it's causing noise in the "Open PRs". But maybe we can keep the branch for future reference.

PR #8975, opened 2025-03-18
fix(test): updates flaky imageArrayDraft pte test

pedrobonamin:
### Description
The PTE image array test [tries to use a file ](https://github.com/sanity-io/sanity/blob/next/test/e2e/tests/pte/ImageArrayDrag.spec.ts#L30)that is added by another [test](https://github.com/sanity-io/sanity/blob/next/test/e2e/tests/inputs/array.spec.ts#L9), and it always fails on the first pass. 

With this changes, the test is responsible for adding the file, so it can run in isolation, without reusing files added by other tests, which could exist or not at the time this test runs.

### DIFF
--- a/test/e2e/tests/pte/ImageArrayDrag.spec.ts
+++ b/test/e2e/tests/pte/ImageArrayDrag.spec.ts
@@ -1,47 +1,72 @@
+import {createReadStream} from 'node:fs'
+import path from 'node:path'
+
 import {expect} from '@playwright/test'
+import {type SanityImageAssetDocument} from '@sanity/client'
 import {test} from '@sanity/test'
 
-test('Portable Text Input - Array Input of images dragging an image will not trigger range out of bounds (toast)', async ({
-  page,
-  createDraftDocument,
-}) => {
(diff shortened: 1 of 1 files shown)


test/e2e/tests/pte/ImageArrayDrag.spec.ts
```diff
@@ -1,47 +1,72 @@
...
-test('Portable Text Input - Array Input of images dragging an image will not trigger range out of bounds (toast)', async ({
-  page,
-  createDraftDocument,
-}) => {
-  await createDraftDocument('/test/content/input-standard;portable-text;pt_allTheBellsAndWhistles')
-
-  // set up the portable text editor
-  await page.getByTestId('field-body').focus()
-  await page.getByTestId('field-body').click()
+test.describe('Portable Text Input - ImageArrayDraft', () => {
```
pedrobonamin, line 8: Wow, lot's of noise in the diff. 
This basically adds a beforeAll in where the capybara image is added to the dataset, so we can use it in the test.
And it adds an afterAll removing the image.

The rest of diff is because the test was moved inside a describe so we can use the beforAll and afterAll hooks.

jordanl17 (approved):

PR #8978, opened 2025-03-18
fix(cli): respect cli config for http host/port

rexxars:
### Description

Fixes a regression in 3.80.0 where the host/port CLI flags would get default values if not provided, but would be treated as explicitly set. The result is that they took precedent over both env vars and CLI config values, which they should do _if explicitly passed_, but not if they are the fallback values.

### What to review

- Using `http.port`/`http.host` in `sanity.cli.ts` still works.

### Testing

Would be ideal to get a test in here, but hotfixing for now

### Notes for release

- Fixes a regression in 3.80.0 causing http host/port configured in environment variables/CLI configuration not to be respected by the `dev` command


binoy14 (approved):
🤦‍♂️ thanks for the fix!

PR #8299, opened 2025-01-16
feat(core): Global Document References support

sgulseth:
### Description

This is @skogsmaskin writing BTW 👋 

This will add support for Global Document References (GDR) in the Studio. A Global Document Reference can reference something inside another project across the user's Sanity organization.

The implementation is very much alike cross-dataset-references, but there is a separate schema type that has it's unique config options:

```ts
export const globalDocumentReferenceType = defineType({
  type: 'globalDocumentReference',
  name: 'globalDocumentSubtype',
  title: 'Global document reference',
  resourceType: 'dataset',
  resourceId: 'exx11uqh.blog',
  to: [
    {
      type: 'book',
      icon: BookIcon,
      preview: {
        select: {
          title: 'title',
          media: 'coverImage',
        },
        prepare(val) {
          return {
            title: val.title,
            media: val.coverImage,
          }
        },
      },
    },
  ],
})
```

We don't want to release this yet to the public just yet, but it would be great if we could start reviewing it, so that it's ready to go when we need it. It also makes it possible to get some more test usage before we announce it.

### What to review
* That we are happy with the config options.
* That the current implementation works well enough to be merged to `next`.

### Testing
Open up the test studio in this branch, and select "playground (Staging)" workspace. Go to "Standard Inputs" and "Global document references test":

image

There is also added some automatic tests.

Note that the API currently only works in staging, and needs to be released first.

### Notes for release
N/A - not to be announced yet.

PR #8982, opened 2025-03-19
chore: add simple e2e presentation test

stipsan:
### Description

After merging #8299 the Presentation [test workspace](https://test-studio-jw4ozwi1o.sanity.dev/presentation/presentation) started crashing. [The build just prior works fine.](https://test-studio-5h0wzd5p4.sanity.dev/presentation/presentation)
Turns out the root cause is `@sanity/assist`, fixed here: https://github.com/sanity-io/assist/pull/64
To avoid breaking the test workspace in the future I'm adding a very simple E2E test. I'll be expanding the test suite over time, this is just to catch simple crash cases like this.
This PR should have a failing test until #8983 is merged, which upgrades `@sanity/assist` with the required fix.

### What to review

Does it make sense? I followed the same pattern as the new Vision Tool tests ✨ 

### Testing

If it passes it works (after #8983, before that it should fail consistently).

### Notes for release

N/A

### DIFF
--- a/dev/studio-e2e-testing/sanity.config.ts
+++ b/dev/studio-e2e-testing/sanity.config.ts
@@ -3,6 +3,7 @@ import {googleMapsInput} from '@sanity/google-maps-input'
 import {BookIcon} from '@sanity/icons'
 import {visionTool} from '@sanity/vision'
 import {defineConfig} from 'sanity'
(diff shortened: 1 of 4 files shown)


pedrobonamin (approved):
Looks good to me!

PR #8984, opened 2025-03-19
fix(test): update flaky navbar test

pedrobonamin:
### Description
By inspecting the video in the flaky navbar test I think the issue is caused because the navbar is recreated when tasks plugin is injected.
Tasks do an async check to verify if it is enabled or not, using the `useFeatureEnabled('sanityTasks')` check.
This needs some time to resolve, my theory is the following:

- The studio renders.
- The navbar is shown and the test clicks on the button to show the help resources popover.
- **useFeatureEnabled** resolves and says tasks are enabled.
- The navbar is recreated because now we need to inject the tasks button. ([see this](https://github.com/sanity-io/sanity/blob/fix-flaky-navbar-spec/packages/sanity/src/core/tasks/plugin/TasksStudioNavbar.tsx#L77-L85))
- The popover is now closed because the navbar was recreated

This is visible on this video, in the first seconds after the studio loads the popover shows for a microsecond and then it disappears.
| Renders popover | Taks is added, popover disappears |
|--------|--------|
|Screenshot 2025-03-19 at 12 40 24 | Screenshot 2025-03-19 at 12 40 30| 

**Video:**
[bf757adebb91ad63aadced8e9d58cd453620b5e4.webm](https://github.com/user-attachments/assets/081215ec-451b-4912-95f6-247c41ccac0e)

This also removes a test that was just checking for the navbar to exist and moves it to the navbar test.
 

### What to review
Do tests pass on first pass?

### DIFF
--- a/packages/sanity/src/core/tasks/plugin/TasksStudioNavbar.tsx
+++ b/packages/sanity/src/core/tasks/plugin/TasksStudioNavbar.tsx
@@ -21,6 +21,7 @@ const TasksToolbar = ({onClick, isOpen}: {onClick: () => void; isOpen: boolean})
       mode="bleed"
       onClick={onClick}
       selected={isOpen}
+      data-testid="tasks-toolbar"
     />
   )
 }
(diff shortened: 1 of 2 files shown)


RitaDias (approved):
Let's go!

Issue #8981, opened 2025-03-19
Opening presentation mode via "Used on one page" forgets the perspective


mariusGundersen:
https://github.com/user-attachments/assets/55d3fd49-5e14-4a68-bbe4-397983ddcb84

See video above. When trying to go to presentation mode for a document the perspective is forgotten

summon @pedrobonamin

pedrobonamin:
Thanks for reporting and catching this one @mariusGundersen 
Opening a Pr now to fix it.

PR #8987, opened 2025-03-19
fix(core): stickyParams for `__unsafe_disableScopedSearchParams`

pedrobonamin:
### Description

fixes https://github.com/sanity-io/sanity/issues/8981

Presentation tool defines in it's own router the `__unsafe_disableScopedSearchParams` property [here](https://github.com/sanity-io/sanity/blob/sapp-2521/packages/sanity/src/presentation/router.ts#L4), this allows presentation to define it's own params and opts out from the scoped search params
While doing this, the sticky params are not following you around, this PR updates how we set the params when using the`__unsafe_disableScopedSearchParams` property and makes them **sticky** also for this use case.

https://github.com/user-attachments/assets/1be4dc19-3350-400a-8fa2-3556933892da

### What to review

Are this changes correct?

### Testing

Open a **simpleBlock** document in **/presentation/structure** click the **used on page** link, the perspective should follow you to **presentation**
For example here https://test-studio-git-sapp-2521.sanity.dev/presentation/structure/untitledRepro;documentStore;drafts.1a050310-4ce3-48e1-a70a-74c1808786da?perspective=rKM9a34iK

### Notes for release
Now perspective and excluded perspectives will follow you when navigating to `presentation`


stipsan (dismissed):
Makes sense to me 👍

pedrobonamin:
> Makes sense to me 👍

Thank you for the review @stipsan I missed adding a release tag to the now exported sticky params, could you 👀  again please ?

stipsan (approved):

PR #8989, opened 2025-03-19
feat: releases tool tables show skeletons when loading rather than spinner

jordanl17:
### Description

https://github.com/user-attachments/assets/8d919234-4e22-413f-9b03-2bb0f2b609c1

### Testing
Existing tests are fine

### Notes for release
N/A

### DIFF
--- a/packages/sanity/src/core/releases/tool/components/Table/Table.tsx
+++ b/packages/sanity/src/core/releases/tool/components/Table/Table.tsx
@@ -36,6 +36,13 @@ export type TableRowProps = Omit<
 > &
   RefAttributes<HTMLDivElement>
 
+type VirtualDatum = {
+  virtualRow: VirtualItem
+  index: number
+  isFirst: boolean
+  isLast: boolean
+}
+
 export interface TableProps<TableData, AdditionalRowTableData> {
   columnDefs: Column<RowDatum<TableData, AdditionalRowTableData>>[]
   searchFilter?: (data: TableData[], searchTerm: string) => TableData[]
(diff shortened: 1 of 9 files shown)


packages/sanity/src/core/releases/tool/components/Table/Table.tsx
```diff
@@ -252,11 +255,50 @@ const TableInner = <TableData, AdditionalRowTableData>({
 
   const theme = useTheme()
 
-  if (loading) {
-    return <LoadingBlock fill data-testid="table-loading" />
+  const maxInlineSize = (!hideTableInlinePadding && theme.sanity.v2?.container[3]) || 0
```
bjoerge, line 258: [side note as I realize it's not part of these changes, but] wondering if we could get rid of this reference to v2 themes somehow?

packages/sanity/src/core/releases/tool/components/Table/Table.tsx
```diff
@@ -252,11 +255,50 @@ const TableInner = <TableData, AdditionalRowTableData>({
...
+        isFirst: index === 0,
+        isLast: index === LOADING_ROW_COUNT - 1,
+      })
+    })
   }
 
-  const maxInlineSize = (!hideTableInlinePadding && theme.sanity.v2?.container[3]) || 0
+  const tableContent = () => {
+    if (loading) {
+      return renderLoadingRows(renderRow as (datum: any) => React.ReactNode)
```
bjoerge, line 284: could we type `datum` as `object` or `Record<string, unknown>` here instead of `any`? Looks like we expect it to be spreadable, at least?

jordanl17: Oh yes, good callout - I'll improve these types to make them more appropriate

- [ ] Avoid `any` and improve types

packages/sanity/src/core/releases/tool/components/Table/Table.tsx
jordanl17: I'm not sure myself- perhaps you have more context, but this seems to be referring to sanity-ui v2, not studio v2. `useTheme` returns a `Theme_2` type. Is this legacy?

bjoerge: > but this seems to be referring to sanity-ui v2, not studio v2.

that's right, yeah, it's legacy, so would be good to remove it at some point :)

bjoerge (approved):
Nice work!

PR #8994, opened 2025-03-19
chore: add vitest typecheck support

stipsan:
### Description

Adds support for [Testing Types](https://vitest.dev/guide/testing-types.html), a super nice feature in vitest. It allows us to setup integration tests for typings. Our codebase and public API is sensitive to breaking changes and regressions to our types, especially for schema authoring. It also extends to APIs, react hooks and more that is used by custom input components, custom plugins and such.
Leveraging Vitest' typechecking features will allow us to build up a testing suite that can assert things that only affect TypeScript, and not runtime logic. Such as `defineConfig`, `defineField` and many other APIs we have that do nothing at runtime.

### What to review

All good?

### Testing

If tests pass we good fam.

### Notes for release

N/A


pedrobonamin (approved):
Nice!

PR #8996, opened 2025-03-19
fix(core): drop explicit `withCredentials` on requests

rexxars:
### Description

Two changes:
1. Only provide `withCredentials: true` to client constructor if we are not using a token. If we have a token, we don't _also_ want to send a cookie, as this is both unnecessary (it's likely not there, given we are using a token), and might include irrelevant cookies that were mistakenly attached to the `.sanity.io` domain. 
2. Providing `withCredentials: true` to the client's `.request()` _shouldn't_ be necessary unless the base client for the source does _not_ do so. While it traditionally hasn't been a problem, we want to discourage this pattern as it is confusing: what happens if both an auth cookie and an `authorization` header is provided? While the answer in our case is that the `authorization` header wins, this is not necessarily obvious, and can make things harder to debug.

By moving to an explicit "EITHER bearer token OR session cookie" implementation, things are made a bit clearer. Note that this also aligns with this [upcoming change](https://github.com/sanity-io/client/pull/1037) in the client and silences what will in the future be a warning being printed.

### What to review

All code paths modified still work as expected. I am relying on the previous authors to verify this, as I do not have the full context of why `withCredentials` was included in the first place.

### Testing

Sufficient testing should hopefully already be in place :) 

### Notes for release

None

### DIFF
--- a/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
+++ b/packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
@@ -26,7 +26,6 @@ function createFeatureToggle(client: SanityClient) {
       client.observable
         .request({
           uri: `/data/actions/${dataset}`,
-          withCredentials: true,
         })
         .pipe(
           mapResponse(),
--- a/packages/sanity/src/core/store/_legacy/user/userStore.ts
+++ b/packages/sanity/src/core/store/_legacy/user/userStore.ts
@@ -40,7 +40,6 @@ export function createUserStore({client: _client, currentUser}: UserStoreOptions
     async (userIds) => {
       const value = await client.request<(User | null)[]>({
         uri: `/users/${userIds.join(',')}`,
-        withCredentials: true,
         tag: 'users.get',
       })
       const response = Array.isArray(value) ? value : [value]
--- a/packages/sanity/src/core/store/key-value/storage/serverStorage.ts
+++ b/packages/sanity/src/core/store/key-value/storage/serverStorage.ts
@@ -26,7 +26,6 @@ export function createServerStorage({client: _client}: ServerStorageOptions): Se
     const value = await client
       .request<KeyValuePair[]>({
         uri: `/users/me/keyvalue/${keys.join(',')}`,
-        withCredentials: true,
       })
       .catch((error) => {
         console.error('Error fetching data:', error)
@@ -56,7 +55,6 @@ export function createServerStorage({client: _client}: ServerStorageOptions): Se
         method: 'PUT',
         uri: `/users/me/keyvalue`,
         body: [{key, value: nextValue}],
-        withCredentials: true,
       })
       .then(
         (response) => {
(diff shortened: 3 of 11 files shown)


packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
```diff
@@ -26,7 +26,6 @@ function createFeatureToggle(client: SanityClient) {
       client.observable
         .request({
           uri: `/data/actions/${dataset}`,
-          withCredentials: true,
```
rexxars, line 29: @bjoerge Can you confirm this to be safe? Any idea why we passed `withCredentials: true` here explicitly? (this should have implicitly been the case unless a client not based on the source's `getClient`/`useClient` was used)

packages/sanity/src/core/store/_legacy/user/userStore.ts
```diff
@@ -40,7 +40,6 @@ export function createUserStore({client: _client, currentUser}: UserStoreOptions
     async (userIds) => {
       const value = await client.request<(User | null)[]>({
         uri: `/users/${userIds.join(',')}`,
-        withCredentials: true,
```
rexxars, line 43: @ricokahler  Can you confirm this to be safe? Any idea why we passed `withCredentials: true` here explicitly? (this should have implicitly been the case unless a client not based on the source's `getClient`/`useClient` was used)

packages/sanity/src/core/store/key-value/storage/serverStorage.ts
```diff
@@ -26,7 +26,6 @@ export function createServerStorage({client: _client}: ServerStorageOptions): Se
     const value = await client
       .request<KeyValuePair[]>({
         uri: `/users/me/keyvalue/${keys.join(',')}`,
-        withCredentials: true,
```
rexxars, line 29: @cngonzalez  Can you confirm this to be safe? Any idea why we passed `withCredentials: true` here explicitly? (this should have implicitly been the case unless a client not based on the source's `getClient`/`useClient` was used)

packages/sanity/src/core/studio/addonDataset/AddonDatasetProvider.tsx
```diff
@@ -41,7 +41,6 @@ function AddonDatasetProviderInner(props: AddonDatasetSetupProviderProps) {
         projectId,
         requestTagPrefix: 'sanity.studio',
         useCdn: false,
-        withCredentials: true,
```
rexxars, line 44: @pedrobonamin Can you confirm this to be safe? Any idea why we passed `withCredentials: true` here explicitly? (this should have implicitly been the case unless a client not based on the source's `getClient`/`useClient` was used)

ricokahler (approved):
this makes sense to me and i do not think there was a significant reason i added withCredentials

packages/sanity/src/core/store/_legacy/document/document-pair/utils/fetchFeatureToggle.ts
bjoerge: Probably just old habit/copy paste – should be safe to remove!

bjoerge (approved):
LGTM – does it also make sense to update the ClientOptions type to only allow either, to prevent these from sneaking back in?

packages/sanity/src/core/studio/addonDataset/AddonDatasetProvider.tsx
pedrobonamin: Looks safe to me.
Just tried creating a new add-on dataset and it works fine.

rexxars:
> does it also make sense to update the ClientOptions type to only allow either, to prevent these from sneaking back in?

Yes, but that might feel like a "breaking" change, so might wait until next major of client.

Issue #8986, opened 2025-03-19
mainDocumentResolver assumes drafts perspective


mariusGundersen:
This line: https://github.com/sanity-io/sanity/blob/6cb0865026bf42c7425425988c588d69f92d2a69/packages/sanity/src/presentation/useMainDocument.ts#L170

assumes drafts perspective when trying to resolve the main document given a path. It should use the current perspective to do so.

summon @pedrobonamin

mariusGundersen:
The result is that the main document isn't found if it's created in a release and therefore doesn't exist in the draft.

I tried usunt the `perspectiveStack` from `usePerspective` and that worked

PR #9005, opened 2025-03-20
fix(presentation): use `perspectiveStack` when resolving main document

pedrobonamin:
### Description
Fixes https://github.com/sanity-io/sanity/issues/8986

This PR adds the `perspectiveStack` to the `useMainDocument` get function, allowing presentation to get the correct document when using a perspective.
If not, if the document doesn't exist as a draft it won't be shown as the main document.

It also fixes an issue with the preview for the main document shown in the presentation list to also respect the perspective.

See video for example:

https://github.com/user-attachments/assets/8e858b9c-2ecf-4e84-9904-2ab6934f7bd4

### What to review
Are the changes correct?

### Testing
Open our presentation studio, a main document should load.
If you add a main document to another release, then it should be respected when visiting the page.

### Notes for release
Fixes an issue in where main documents won't be accurately displayed if they only exist as part of a release.

### DIFF
--- a/dev/test-studio/sanity.config.ts
+++ b/dev/test-studio/sanity.config.ts
@@ -12,7 +12,7 @@ import {debugSecrets} from '@sanity/preview-url-secret/sanity-plugin-debug-secre
 import {tsdoc} from '@sanity/tsdoc/studio'
 import {visionTool} from '@sanity/vision'
 import {defineConfig, definePlugin, type WorkspaceOptions} from 'sanity'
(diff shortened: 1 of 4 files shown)


stipsan (dismissed):
Thank you! 💖

pedrobonamin:
> Thank you! 💖

Thank you for the review @stipsan , had to rebase so it needs a new ✅

stipsan (approved):

PR #9006, opened 2025-03-20
chore(test): add snapshot tests for package exports

bjoerge:
### Description
This adds a unit test that reads the package exports of the `sanity` package, and adds a snapshot asserting no changes.

This enables us to easily track and get an overview of the exported surface, and also helps us in catching unintended breaking changes.

### What to review
The test will be run along with other unit tests, and can be run separately with `npx vitest run packages/sanity/test/exports.test.ts`

Here's an example of what it will look like if the test fails:
image

If this was indeed an intentional change, the snapshot can be updated by running `vitest` with the `--update` or `-u` flag (https://vitest.dev/guide/snapshot.html#updating-snapshots)

### Testing
yes :)

### Notes for release
n/a internal


packages/sanity/test/setup/environment.ts
```diff
@@ -22,8 +22,8 @@ export {}
 // get rid of context warning
 const warn = console.warn
 const error = console.error
-window.console = {
-  ...window.console,
+;(globalThis as any).console = {
```
bjoerge, line 25: not strictly related to the changes here, but needed in order to run tests in node

PR #9007, opened 2025-03-20
feat(sanity): make initial value template resolution interruptible

juice49:
### Description

When schemas have many initial value templates, or long initial value templates, the main thread may become blocked while these are resolved. When examining performance in a Studio affected by this issue, you will likely see the browser cumulatively spends a lot of time executing the `resolveInitialObjectValue` function.

This branch improves the situation by using [the Prioritized Task Scheduling API](https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API) to break initial value template resolution into smaller tasks that the browser can coordinate without blocking the main thread.

Not only does this keep the UI responsive while initial value templates are being resolved, but it appears the browser is able to coordinate the work more efficiently. **Users who previously reported creating a new document to be unusably slow reported that this change resolved the issue.**

Note: it would be great to spend some time investigating whether we can speed the resolution process up, perhaps making this mitigation unnecessary.

#### Changes to document editor read-only state

This branch updates the document editor such that it's rendered in a read-only state while the document's initial value is resolved. I suspect it's an oversight that this wasn't already the case; what's the expected behaviour if the initial value resolves _after_ the user has modified the new document?

#### Prioritized Task Scheduling API polyfill

The Prioritized Task Scheduling API is not yet supported by all browsers. I created a small `postTask` helper function that executes the provided callback immediately if the Prioritized Task Scheduling API is not available. **We know that some of the users most severely affected by the issue are using web browsers that support Prioritized Task Scheduling**.

We will follow this change up by adding a polyfill or a fallback strategy. I have started work on including a polyfill in #9024.

### What to review

- Does the general approach seem sound?
- Do you have any concerns about the changes made in order for the document editor to be read-only while the initial value template resolves?

### Testing

- Initial value template resolution is already well-tested, and these existing tests continue to pass.
- Manually tested using a contrived large schema.
- Manually tested using a user-supplied real-world large schema.
- Manually tested by affected users.

### DIFF
--- a/packages/sanity/src/core/form/useDocumentForm.ts
+++ b/packages/sanity/src/core/form/useDocumentForm.ts
@@ -43,6 +43,7 @@ import {isPublishedPerspective, isReleaseScheduledOrScheduling} from '../release
 import {
   type DocumentPresence,
   type EditStateFor,
+  type InitialValueState,
   type PermissionCheckResult,
   useDocumentValuePermissions,
   usePresenceStore,
@@ -64,7 +65,7 @@ interface DocumentFormOptions {
   documentType: string
   documentId: string
   releaseId?: ReleaseId
-  initialValue?: SanityDocumentLike
+  initialValue?: InitialValueState
   initialFocusPath?: Path
   selectedPerspectiveName?: ReleaseId | 'published'
   readOnly?: boolean | ((editState: EditStateFor) => boolean)
@@ -192,7 +193,7 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
   }, [comparisonValueRaw, editState])
 
   const value: SanityDocumentLike = useMemo(() => {
-    const baseValue = initialValue || {_id: documentId, _type: documentType}
+    const baseValue = initialValue?.value || {_id: documentId, _type: documentType}
     if (releaseId) {
       return editState.version || editState.draft || editState.published || baseValue
     }
@@ -269,7 +270,7 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
   const isNonExistent = !value?._id
   const isCreateLinked = isSanityCreateLinkedDocument(value)
 
-  const ready = connectionState === 'connected' && editState.ready
+  const ready = connectionState === 'connected' && editState.ready && !initialValue?.loading
 
   const selectedPerspective = useMemo(() => {
     return getSelectedPerspective(selectedPerspectiveName, releases)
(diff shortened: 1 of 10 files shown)


packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -64,7 +65,7 @@ interface DocumentFormOptions {
   documentType: string
   documentId: string
   releaseId?: ReleaseId
-  initialValue?: SanityDocumentLike
+  initialValue?: InitialValueState
```
pedrobonamin, line 68: @robinpyon FYI, this change given you are using it in create

bjoerge: Seems like the only thing we need to know in addition to the document/value is the loading bit?

I would prefer if we could preserve backwards compat. Is it feasible to instead add a `initialValueLoading`-option?

packages/sanity/src/core/tasks/components/form/tasksFormBuilder/TasksFormBuilder.tsx
```diff
@@ -62,7 +62,17 @@ const TasksFormBuilderInner = ({
...
     value,
-  } = useDocumentForm({documentId, documentType: 'tasks.task', initialValue})
+  } = useDocumentForm({
+    documentId,
+    documentType: 'tasks.task',
+    initialValue: initialValue
+      ? {
+          loading: false,
+          value: initialValue,
+          error: null,
```
bjoerge, line 72: doesn't seem like `error` is used by the hook?

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -27,9 +27,9 @@ async function inlineCommentCreationTest(props: InlineCommentCreationTestProps)
   // 3. Select all text in the editor.
   await page.waitForSelector('[data-testid="pt-editor"]', WAIT_OPTIONS)
   const editor = page.locator('[data-testid="pt-editor"]')
-  await page.waitForSelector('div[role="textbox"]', WAIT_OPTIONS)
-  const textBox = editor.locator('div[role="textbox"]')
-  textBox.selectText()
+  await page.waitForSelector('div[role="textbox"][contenteditable=true]', WAIT_OPTIONS)
```
bjoerge, line 30: how are these related?

bjoerge: ah, thanks for clarifying

PR #9010, opened 2025-03-21
fix: archived releases table has default sort and defined sort working

jordanl17:
### Description
* Default sorting on a hidden column - which sorts by either `_updatedAt` or `publishedAt` depending on which is available 
* When sorting by published, all archived releases are at the end of the list, when sorting by archived, all published releases are at the end of the list

| Scenario | Before | After |
|--------|--------|--------|
| Default sort | defaultSortOLD | defaultSortNEW |
| Published at sort | publishedSortOLD | publishedSortNEW |
| Archived at sort | archivedSortOLD | ArchivedSortNEW | 

### What to review
Supporting `hidden` columns in the `Table` - this is a bit of a hack just to support this particular case (where we want the default sorting to be across multiple columns. An alternative abstraction would be to move a function for default sorting out to the `Table` prop surface, but given the `Table` is only used within releases in 2 locations, going to those extents doesn't seem appropriate.

### Testing
No new tests added, as Table isn't currently covered by tests

### Notes for release
Fix for incorrect ordering when sorting archived and published releases in the Content Releases tool

### DIFF
--- a/packages/sanity/src/core/releases/tool/overview/ReleasesOverviewColumnDefs.tsx
+++ b/packages/sanity/src/core/releases/tool/overview/ReleasesOverviewColumnDefs.tsx
@@ -90,11 +90,32 @@ export const releasesOverviewColumnDefs: (
       },
       'active',
     ),
+    // This is a hidden column only used for sorting when
+    // no other sort column is selected (the default state)
+    checkColumnMode(
+      {
+        id: 'lastActivity',
+        hidden: true,
+        sorting: true,
+        width: 100,
+        sortTransform: ({publishedAt, _updatedAt}) => {
+          // the default sort is always descending, so -Infinity pushes missing values to end
+          const lastActivity = publishedAt ?? _updatedAt
+
+          return lastActivity ? new Date(lastActivity).getTime() : -Infinity
+        },
+      },
+      'archived',
+    ),
     checkColumnMode(
       {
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/releases/tool/overview/ReleasesOverviewColumnDefs.tsx
```diff
@@ -90,11 +90,32 @@ export const releasesOverviewColumnDefs: (
...
+    // This is a hidden column only used for sorting when
+    // no other sort column is selected (the default state)
+    checkColumnMode(
+      {
+        id: 'lastActivity',
+        hidden: true,
+        sorting: true,
+        width: 100,
+        sortTransform: ({publishedAt, _updatedAt}) => {
+          // the default sort is always descending, so -Infinity pushes missing values to end
```
RitaDias, line 102: Very smart!

RitaDias (commented):
I think the solution is pretty smart. I'm still unsure how I feel about a hidden column. That being said, the code overall looks good to me. I think I would prefer to have someone else have a look at it too and have some more thoughts, maybe @pedrobonamin since I think you also did quite a bit of work on this table?

pedrobonamin (approved):
I think it's a very clever solution, fixes the issue without introducing too many changes to the codebase.

PR #9013, opened 2025-03-21
fix(presentation): add `sanity-preview-perspective`  to open preview link

pedrobonamin:
### Description

Users need to be able to share a preview url and use the perspective param to show the correct perspective in their front end.
With this change, we are adding the   `sanity-preview-perspective`  to the url generated by the open preview button
Screenshot 2025-03-21 at 15 23 40

### What to review
Are this changes correct?
I noticed that we use [`withoutSecretSearchParams`](https://github.com/sanity-io/visual-editing/blob/main/packages/preview-url-secret/src/withoutSecretSearchParams.ts#L15) function from visual editing in [here](https://github.com/sanity-io/sanity/blob/next/packages/sanity/src/presentation/preview/PreviewHeader.tsx#L72), which is removing this value from the url,  is it safe to add it back?

### Testing
Open the preview link it should expose the perspective

### Notes for release
 add `sanity-preview-perspective`  to open preview link


packages/sanity/src/presentation/preview/PreviewHeader.tsx
```diff
@@ -207,11 +209,11 @@ const PreviewHeaderDefault = (props: Omit<PreviewHeaderProps, 'renderDefault'>)
...
+      {true && (
         <Flex align="center" flex="none" gap={1}>
           <SharePreviewMenu
-            canToggleSharePreviewAccess={canToggleSharePreviewAccess}
-            canUseSharedPreviewAccess={canUseSharedPreviewAccess}
+            canToggleSharePreviewAccess
+            canUseSharedPreviewAccess
             previewLocationRoute={previewLocationRoute}
             initialUrl={initialUrl}
             perspective={perspective}
```
stipsan, line 219: Oops I think there's some dev overrides here that probably isn't intended to be in the PR? :)

packages/sanity/src/presentation/preview/OpenPreviewButton.tsx
```diff
@@ -1,20 +1,33 @@
...
+    perspective: PresentationPerspective
+    targetOrigin: string
   },
 ): React.ReactNode {
-  const {openPopup, previewLocationOrigin, previewLocationRoute} = props
+  const {openPopup, previewLocationOrigin, previewLocationRoute, perspective, targetOrigin} = props
+
+  const openPreviewLink = useMemo(() => {
+    const url = new URL(previewLocationRoute, previewLocationOrigin || targetOrigin)
+    url.searchParams.set(urlSearchParamPreviewPerspective, encodeStudioPerspective(perspective))
```
stipsan, line 26: Yep this is fine. The `withoutSecretParams` util was originally designed to remove `sanity-preview-secret` and `sanity-preview-path`.
When we added `sanity-preview-perspective` it was added to the same util just to make the location URL cleaner, since the search param is noisy and an implementation detail. It would be better if we originally made sure to strip `sanity-preview-perspective` in the places we didn't want to display it, instead of stripping it everywhere by default and then have to add it back in where needed 🙌

packages/sanity/src/presentation/preview/PreviewHeader.tsx
pedrobonamin: Absoultety, thanks for catching

stipsan (approved):

PR #9015, opened 2025-03-21
fix(test): don't export `isNavigateOptions` 

pedrobonamin:
### Description
After merging https://github.com/sanity-io/sanity/pull/9006 a new exports snapshot test was added.
this PR https://github.com/sanity-io/sanity/pull/8620 is exporting a new function, given it didn't have the tests on that branch, it was not notified about the new test and the snapshot was not updated.

This PR updates the snapshot to include this new function


jordanl17 (approved):
Nice followup

PR #9018, opened 2025-03-24
fix(core): increase the window interfac for when the sidebar for the task is hovering

RitaDias:
### Description

Withe the releases update, responsive wise it looks like the "triggering" for the hovering taskbar sidebar to becoming floating comes too late. With this update, instead of making it floating by the 900px breakpoint, it will make it by 1200px. 

This will impact all areas of the studio, instead of just the releases page, however, I think this is an acceptable thing given that the task sidebar should likely not take as important a real-estate as the "main" opened things in the studio

| Month    | Savings |
| -------- | ------- |
| image  |  image   |

### What to review

Is there a better way of going about this?

### Testing

N/A

### Notes for release

N/A


jordanl17 (approved):
seems fair

PR #9031, opened 2025-03-24
chore: add presentation e2e test for viewport toggle

stipsan:
### Description

This test helps prevent regressions like #8620 from sneaking back in in the future 🙌 

### What to review

Does it make sense?

### Testing

If tests pass we're good 👍 

### Notes for release

N/A


jordanl17 (approved):
Thanks for adding this

PR #9032, opened 2025-03-24
fix: non sticky params are not preserved during navigation

jordanl17:
### Description
Fixes a bug that was introduced in https://github.com/sanity-io/sanity/pull/8620 where non-sticky search params would be preserved across navigation events.

This was surfaced in the viewport toggle in Presentation:

| Before | After |
|--------|--------|
| nonStickyBEFORE | nonStickyAFTER | 

### Testing
Added a new test case on `RouterProvider` to capture this exact scenario

### Notes for release
N/A - covered in 8620


RitaDias (approved):
We love to see it 🙏 
Also thanks for the tests!

stipsan (approved):

PR #9035, opened 2025-03-24
chore: add intl array test

stipsan:
### Description

Adds a test for `sanity-plugin-internationalized-array` in the `Presentation` tool, in the test studio. The way it handles intl is sensitive to how Presentation handles union references lookups when it handles drag and drop logic. We had issues with this in the past so adding this makes it easier to prevent regressions.

### What to review

It makes sense?

### Testing

I'll add an e2e test for this in a follow up PR ❤ 

### Notes for release

N/A


bjoerge (approved):
LGTM, thank you @stipsan!

jordanl17 (approved):

PR #9036, opened 2025-03-24
refactor: remove legacy exports

stipsan:
### Description

We've stopped using `legacyExports` [in `@sanity/ui` a while back](https://github.com/sanity-io/ui/pull/1560), and it's [no longer a feature in `@sanity/pkg-utils@7`](https://github.com/sanity-io/pkg-utils/blob/main/CHANGELOG.md#700-2025-01-08). It's time to remove it from `sanity` and related monorepo packages as well.
This reduces the way we ship code to npm from dual ESM + CJS + legacy ESM to just dual ESM + CJS.
Building legacy exports accounts for 33% of the total build time by `@sanity/pkg-utils`, which impacts how long it takes for the monorepo to run `pnpm build`, and every CLI task that needs `pnpm build`, which is all of the E2E test suites, as well as the eFPS suite.
All to support legacy Node.js runtimes, and test runners like older Jest versions, that doesn't support the `package.json` `exports` field.

### What to review

It builds and makes sense.

### Testing

If it builds it works.

### Notes for release

Legacy ESM is no longer shipped to npm. Ensure you're on Node.js v18 or later, and if using Jest make sure you're on v28 or later.


bjoerge (approved):
This is great!

PR #9040, opened 2025-03-24
fix(preview): fix issue with document version indicators not syncing

bjoerge:


pedrobonamin (approved):
Thank you @bjoerge, tested all the cases I could think and this works great.

PR #9041, opened 2025-03-24
fix: releases used wrong key on duplicate action

jordanl17:
### Description
In minified bundles `name` may be minified, since I believe this is the `name` of the component. It is safer to use `action`.

### Testing
Testing with additional studios and verified it works

### Notes for release
Fixing an issue where content releases would incorrectly hide the duplicate document action

### DIFF
--- a/packages/sanity/src/core/releases/plugin/documentActions/index.ts
+++ b/packages/sanity/src/core/releases/plugin/documentActions/index.ts
@@ -9,7 +9,7 @@ export default function resolveDocumentActions(
   existingActions: Action[],
   context: DocumentActionsContext,
 ): Action[] {
-  const duplicateAction = existingActions.filter(({name}) => name === 'DuplicateAction')
+  const duplicateAction = existingActions.filter(({action}) => action === 'duplicate')
 
   if (context.versionType === 'version') {
     return duplicateAction.concat(DiscardVersionAction, UnpublishVersionAction)


RitaDias (approved):
Good catch!

packages/sanity/src/core/releases/plugin/documentActions/index.ts
```diff
@@ -9,7 +9,7 @@ export default function resolveDocumentActions(
   existingActions: Action[],
   context: DocumentActionsContext,
 ): Action[] {
-  const duplicateAction = existingActions.filter(({name}) => name === 'DuplicateAction')
+  const duplicateAction = existingActions.filter(({action}) => action === 'duplicate')
```
pedrobonamin, line 12: I think you can do the following here, just so it's not validating with a string

```
import { DuplicateAction } from '../../../../structure/documentActions'

const duplicateAction = existingActions.filter(({name}) => name === DuplicateAction.name)

```

bjoerge: I think we'd like to avoid core depending on structure here?

bjoerge (approved):

packages/sanity/src/core/releases/plugin/documentActions/index.ts
pedrobonamin: ohhh right! Yes

PR #9057, opened 2025-03-27
test: trying to make inline.spec more robust

jordanl17:
### Description
Lots of changes to the test:
* Turning on for webkit and firefox (since fingers crossed it seems robust enough to handle these)
* Using a triple click to highlight the text in the PTE
* Using a fallback of keyboard highlighting if the triple click didn't work
* Using network idle checks to ensure that test is ready to begin
* Marking as slow because firefox can be very slow at this test (and in general this is one of the most flakey tests)

### Notes for release
N/A

### DIFF
--- a/test/e2e/tests/comments/inline.spec.ts
+++ b/test/e2e/tests/comments/inline.spec.ts
@@ -18,74 +18,114 @@ async function inlineCommentCreationTest(props: InlineCommentCreationTestProps) 
   const {page, createDraftDocument} = props
   test.slow()
 
-  // 1. Create a new draft document
-  await createDraftDocument('/test/content/input-ci;commentsCI')
+  // Function to perform all steps up to text selection
+  async function setupAndSelectText() {
+    // 1. Create a new draft document
+    await createDraftDocument('/test/content/input-ci;commentsCI')
 
-  // Wait for network to become idle to ensure document is fully loaded
-  await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
+    // Wait for network to become idle to ensure document is fully loaded
+    await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
 
-  // 2. Click the overlay to active the editor.
-  await page.getByTestId('activate-overlay').waitFor(WAIT_OPTIONS)
-  await page.click('[data-testid="activate-overlay"]')
+    // 2. Click the overlay to active the editor.
+    await page.getByTestId('activate-overlay').waitFor(WAIT_OPTIONS)
+    await page.click('[data-testid="activate-overlay"]')
 
-  // Wait for any network activity triggered by activating the editor to complete
-  await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
+    // Wait for any network activity triggered by activating the editor to complete
+    await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
 
-  // 3. Select all text in the editor.
-  await page.getByTestId('pt-editor').waitFor(WAIT_OPTIONS)
+    // 3. Select all text in the editor.
+    await page.getByTestId('pt-editor').waitFor(WAIT_OPTIONS)
 
-  const textbox = page
-    .locator('[data-testid="pt-editor"]')
-    .locator('div[role="textbox"][contenteditable=true]')
+    const textbox = page
+      .locator('[data-testid="pt-editor"]')
+      .locator('div[role="textbox"][contenteditable=true]')
 
-  await textbox.waitFor({...WAIT_OPTIONS, state: 'visible'})
+    await textbox.waitFor({...WAIT_OPTIONS, state: 'visible'})
 
-  // Make sure the editor is fully loaded by waiting for the text content to be present
-  await expect(textbox).toContainText(PTE_CONTENT_TEXT, {timeout: 10000})
+    // Make sure the editor is fully loaded by waiting for the text content to be present
+    await expect(textbox).toContainText(PTE_CONTENT_TEXT, {timeout: 10000})
 
-  // First ensure we're not already focused on the editor by clicking elsewhere
-  await page.click('body', {force: true, position: {x: 0, y: 0}})
-  await page.waitForTimeout(300)
+    // Function to attempt text selection with multiple strategies
+    async function attemptTextSelection() {
+      // First ensure we're not already focused on the editor by clicking elsewhere
+      await page.click('body', {force: true, position: {x: 0, y: 0}})
+      await page.waitForTimeout(300)
 
-  // Now focus on the editor and ensure it's ready for selection
-  await textbox.click({force: true})
-  await page.waitForTimeout(300)
+      // Now focus on the editor and ensure it's ready for selection
+      await textbox.click({force: true})
+      await page.waitForTimeout(300)
 
-  // Try multiple selection strategies to ensure robustness
-  // Strategy 1: Triple-click to select paragraph
-  await textbox.click({clickCount: 3, force: true})
-  await page.waitForTimeout(500)
+      // Try multiple selection strategies to ensure robustness
+      // Strategy 1: Triple-click to select paragraph
+      await textbox.click({clickCount: 3, force: true})
+      await page.waitForTimeout(500)
 
-  // Verify selection and try alternative strategy if needed
-  const isFullTextSelected = await page.evaluate((expectedText) => {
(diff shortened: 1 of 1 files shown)


test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
+      .locator('[data-testid="pt-editor"]')
+      .locator('div[role="textbox"][contenteditable=true]')
+
+    await textbox.waitFor({...WAIT_OPTIONS, state: 'visible'})
+
+    // Make sure the editor is fully loaded by waiting for the text content to be present
+    await expect(textbox).toContainText(PTE_CONTENT_TEXT, {timeout: 10000})
+
+    // Function to attempt text selection with multiple strategies
+    async function attemptTextSelection() {
```
juice49, line 49: This feels like a lot of arbitrary work for something that ideally would be handled by Playwright's `selectText` method. Do we not trust that method and/or does it simply fall short in our app?

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
+
+    // Wait for network to become idle to ensure document is fully loaded
+    await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
+
+    // 2. Click the overlay to active the editor.
+    await page.getByTestId('activate-overlay').waitFor(WAIT_OPTIONS)
+    await page.click('[data-testid="activate-overlay"]')
+
+    // Wait for any network activity triggered by activating the editor to complete
+    await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
```
juice49, line 34: Just curious: what network activity happens as a consequence of activating PTE? First thing that crossed my mind is images inside the PTE content, but those appear to be loaded before activating PTE.

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
+        timeout: 5000,
+      })
+      return true
+    } catch (error) {
+      // Comment button not visible after text selection
+      return false
+    }
+  }
+
+  // Try the setup and text selection up to 3 times
```
juice49, line 103: I think this makes sense as a bandaid, but also feel it's something we should aim to remove if we can.

juice49 (commented):
Thanks for choosing to tackle this! I think it's great if it helps our tests run more reliably, and therefore faster due to reduced failures and retries.

I added some comments, mostly just thoughts around how much legwork we're having to do here.

jordanl17:
> Thanks for choosing to tackle this! I think it's great if it helps our tests run more reliably, and therefore faster due to reduced failures and retries.
> 
> I added some comments, mostly just thoughts around how much legwork we're having to do here.

Yeh I agree, this sort of thing, writing multiple fallback strategies is not good DX. However I think writing this is a rarity. 
From investigation times when we are highlighting text (ie here) or dragging and droppping are extremely flaky and temperamental. In these cases I think it probably warranted to add this extra weight to the test. But only in these extraneous cases

test/e2e/tests/comments/inline.spec.ts
RitaDias: Yeah, I have to admit I don't really like this.
The more I look at the code the more I wonder if the juice is worth the squeeze for this particular test. Because like you've mentioned in other parts, it's a lot of leg work and effort.

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
-  await textBox.selectText()
+  test.slow()
+
+  // Function to perform all steps up to text selection
+  async function setupAndSelectText() {
+    // 1. Create a new draft document
+    await createDraftDocument('/test/content/input-ci;commentsCI')
+
+    // Wait for network to become idle to ensure document is fully loaded
+    await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
```
RitaDias, line 27: I understand that the WAIT_OPTIONS and waiting for concept was already something in this test, but relying on load plus timeouts makes me nervous 😬

test/e2e/tests/comments/inline.spec.ts
RitaDias: The thing is that I'm not even sure if we need to check these things.
This might be getting into nitpick territory, but couldn't we technically rely on checks of changes of interface vs this?

test/e2e/tests/comments/inline.spec.ts
RitaDias: I understand the reasoning behind it but if every time we want to test something in PTE / revolving around the PTE that requires this level of minutia (which can potentially happen since the concept of text selection in PTE would be a common one) I think we're in trouble 😬

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
   // 4. Click on the floating comment button to start authoring a comment.
-  await page.waitForSelector('[data-testid="inline-comment-button"]', {
-    ...WAIT_OPTIONS,
-    state: 'visible',
-  })
-
-  const button = page.locator('[data-testid="inline-comment-button"]')
-  await expect(button).toBeVisible()
-  button.click({
+  const getCommentButton = () => page.locator('[data-testid="inline-comment-button"]')
```
RitaDias, line 129: I think if we can avoid using locator and use instead page.getByTestId we should :)

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
+      break
+    } else if (attempt < maxAttempts) {
+      // Reloading page and trying again...
+      await page.reload()
+      await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
+    }
+  }
+
+  // If we couldn't get the comment button to appear after all attempts, fail the test
+  if (!commentButtonVisible) {
```
RitaDias, line 124: god I understand why we ware doing this but it hurts 😭

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
-    state: 'visible',
-  })
-
-  const button = page.locator('[data-testid="inline-comment-button"]')
-  await expect(button).toBeVisible()
-  button.click({
+  const getCommentButton = () => page.locator('[data-testid="inline-comment-button"]')
+  await expect(getCommentButton()).toBeVisible()
+  getCommentButton().click({
     delay: 1000,
```
RitaDias, line 132: Do we need this delay? Couldn't we wait for an expect for a interface change?

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
   })
 
   // 5. Ensure the comment input field is displayed and the selected text is visually marked for commenting.
-  await page.waitForSelector('[data-testid="comment-input"]', {
+  await page.getByTestId('comment-input').waitFor({
```
RitaDias, line 136: Would `expect(page.getByTestId('comment-input')).toBeVisible()` not work in this instance?

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,96 +16,173 @@ interface InlineCommentCreationTestProps {
...
+  await expect(page.locator('[data-testid="comments-list"]')).toBeVisible()
 
   // 9. Verify that the comment appears within the list and correctly references the intended content.
-  await page.waitForSelector('[data-testid="comments-list-item"]', WAIT_OPTIONS)
-  const commentsListItem = page.locator('[data-testid="comments-list-item"]')
-  await expect(commentsListItem).toBeVisible()
-  const commentsListItemReferencedValue = page.locator(
-    '[data-testid="comments-list-item-referenced-value"]',
+  const commentsListItem = '[data-testid="comments-list-item"]'
+  await page.getByTestId('comments-list-item').waitFor(WAIT_OPTIONS)
```
RitaDias, line 165: I feel we shoud likely be relying on solely the expect than this? Why waiting for the comment list Item? 
(I'm just wondering if it's needed given the line over at 166)

juice49 (approved):

PR #9059, opened 2025-03-27
refactor(core): add release description to discard dialog

RitaDias:
### Description

> [!Note]
> Waiting for a final ✅ from knut for the draft version but everything else should be good to go.

- Update copy for the discard module as feedback from Knut

| Discard a draft | Discard a version |
| -------- | ------- |
| image | image |

### What to review

Does this make sense?

### Testing

N/A

### Notes for release

N/A

### DIFF
--- a/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
@@ -20,8 +20,9 @@ export function DiscardVersionDialog(props: {
   onClose: () => void
   documentId: string
   documentType: string
+  releaseName?: string
 }): React.JSX.Element {
-  const {onClose, documentId, documentType} = props
+  const {onClose, documentId, documentType, releaseName} = props
   const {t} = useTranslation(releasesLocaleNamespace)
   const {t: coreT} = useTranslation()
   const {discardChanges} = useDocumentOperation(getPublishedId(documentId), documentType)
@@ -64,7 +65,9 @@ export function DiscardVersionDialog(props: {
   return (
     <Dialog
       id={'discard-version-dialog'}
-      header={t('discard-version-dialog.header')}
+      header={t('discard-version-dialog.header', {
+        releaseTitle: releaseName,
+      })}
       onClose={onClose}
       width={0}
       padding={false}
--- a/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
@@ -21,9 +21,9 @@ export function DiscardVersionDialog(props: {
   onClose: () => void
   documentId: string
   documentType: string
-  discardFromPerspective: string | SelectedPerspective
+  fromPerspective: string | SelectedPerspective
 }): React.JSX.Element {
-  const {onClose, documentId, documentType, discardFromPerspective} = props
+  const {onClose, documentId, documentType, fromPerspective} = props
   const {t} = useTranslation(releasesLocaleNamespace)
   const {t: coreT} = useTranslation()
   const {discardChanges} = useDocumentOperation(getPublishedId(documentId), documentType)
@@ -34,9 +34,7 @@ export function DiscardVersionDialog(props: {
   const [isDiscarding, setIsDiscarding] = useState(false)
   const discardType = isDraftId(documentId) ? 'draft' : 'release'
   const releaseName =
-    typeof discardFromPerspective === 'string'
-      ? discardFromPerspective
-      : discardFromPerspective.metadata.title
+    typeof fromPerspective === 'string' ? fromPerspective : fromPerspective.metadata.title
 
   const schemaType = schema.get(documentType)
 
--- a/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
@@ -91,7 +91,7 @@ export function DiscardVersionDialog(props: {
         )}
         <Box paddingX={2}>
           <Text size={1} muted>
-            {t('discard-version-dialog.description')}
+            {t('discard-version-dialog.description', {releaseTitle: releaseName})}
           </Text>
         </Box>
       </Stack>
--- a/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
+++ b/packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
@@ -4,7 +4,7 @@ import {useCallback, useState} from 'react'
 import {Dialog} from '../../../../ui-components'
 import {LoadingBlock} from '../../../components'
 import {useDocumentOperation, useSchema} from '../../../hooks'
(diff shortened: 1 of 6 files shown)


packages/sanity/src/core/releases/i18n/resources.ts
```diff
@@ -119,14 +119,21 @@ const releasesLocaleStrings = {
...
-    'This will also permanently remove the changes made to this document within this release.',
-  /** Header for discarding a version of a document dialog */
-  'discard-version-dialog.header':
-    'Are you sure you want to remove this document from the release?',
+  'discard-version-dialog.description-release':
+    'This will permanently remove all changes made to this document within the "{{releaseTitle}}" release. This action cannot be undone.',
+  /** Title for dialog for discarding a draft of a document */
+  'discard-version-dialog.header-draft': 'Discard draft?',
+  /** Header for discarding a version from a release of a document dialog */
+  'discard-version-dialog.header-release': 'Remove document from the "{{releaseTitle}}" release?',
```
jordanl17, line 131: we use `<strong>{{releaseTitle}}</strong>` else where eg:
```
  /** Label when a release has been deleted by a different user */
  'deleted-release': "The '<strong>{{title}}</strong>' release has been deleted",
```

Might be good to find a consistent approach here (what ever it could be)

packages/sanity/src/core/releases/plugin/documentActions/DiscardVersionAction.tsx
```diff
@@ -20,6 +21,11 @@ export const DiscardVersionAction = (
   const {id, type, release, version} = props
   const currentUser = useCurrentUser()
   const {t} = useTranslation()
+  const {selectedPerspective} = usePerspective()
+  const selectedName =
+    typeof selectedPerspective === 'string'
+      ? selectedPerspective
+      : selectedPerspective.metadata.title
```
jordanl17, line 28: COULD: we just put this in `DiscardVersionDialog`

packages/sanity/src/core/releases/i18n/resources.ts
RitaDias: Let me double check with Knut!

packages/sanity/src/core/releases/plugin/documentActions/DiscardVersionAction.tsx
RitaDias: let me have a go at this!

RitaDias: Hmm the thing with moving this there is that I'm starting to get errors with the metadata being read from "undefined" when it technically shouldn't 🤔  I'll keep looking at it

RitaDias: And after turning it completely off and on it works 😂 some times you just end up with mysteries of life (I think the hot reloading failed me a couple of times rip)

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
```diff
@@ -20,8 +21,9 @@ export function DiscardVersionDialog(props: {
   onClose: () => void
   documentId: string
   documentType: string
+  discardFromPerspective: string | SelectedPerspective
```
bjoerge, line 24: Since we're technicailly discarding it from a release (perspective is a broader concept, including `published` and `raw`), I think it would be preferable to not use the perspective term here. Since the context here is already "discard", what about just `releaseId`?

I would also personally prefer a uniform type. WDYT about making title a separate property? Can still fallback to releaseId if not given:
```ts
releaseId: string
releaseTitle?: string
```

packages/sanity/src/core/releases/i18n/resources.ts
RitaDias: I'll make the change though as knut said "I'd stick with the established convention then (even though I might not agree with it 😛)" 😄 

We might need to give another round of updates to these messages in a separate PR

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
jordanl17: I think it's just ambiguity with naming here, `discardFromPerspective` is not `do we want to discard it from the perspective` it's rather `what's the perspective we will discard it from`. Could we just refactor it to `fromPerspective` or `selectedPerspective` even.

Separately would typing it as `SelectedPerspective` alone be enough

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
```diff
@@ -30,6 +32,11 @@ export function DiscardVersionDialog(props: {
   const schema = useSchema()
   const toast = useToast()
   const [isDiscarding, setIsDiscarding] = useState(false)
+  const discardType = isDraftId(documentId) ? 'draft' : 'release'
+  const releaseName =
```
jordanl17, line 36: Might you be able to try just
```suggestion
  const releaseName = isDraftOrPublished(discardFromPerspective) ? discardFromPerspective : discardFromPerspsective.metadata.title
```

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
RitaDias: I've mentioned this in a separate comment somewhere but we use `SelectedPerspective` and that's not necessarily the right term here either. Yes technically it's the chosen perspective but SelectedPerspective as a term used in other parts of the code often refers to the pinned version too which would be an incorrect assumption since there are instances where you are discarding a version that is not the one that is pinned.

I would be ok with `fromPerspective` though!

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
RitaDias: I don't think so since this is mostly applied to releases and not to documentIds. I see that it's not being used anywhere so we might need to remove it. 
In any case I would keep it as is since we will never discard a Publish "version" this way anyway :)

packages/sanity/src/core/releases/components/dialog/DiscardVersionDialog.tsx
RitaDias: Renamed it to `fromPerspective`

packages/sanity/src/core/releases/tool/detail/ReleaseSummary.tsx
```diff
@@ -74,9 +74,9 @@ export function ReleaseSummary(props: ReleaseSummaryProps) {
       if (!isBundleDocumentRow(rowProps.datum)) return null
       if (rowProps.datum.isPending) return null
 
-      return <DocumentActions document={rowProps.datum} />
+      return <DocumentActions document={rowProps.datum} releaseName={release.metadata.title} />
```
bjoerge, line 77: I still think it's a bit confusing that we're mixing _name_ and _title_. We do have a concept of release name, which  is different from the title we display in various places. If this is intentional, could we have a comment explaining the rationale?

bjoerge (approved):

PR #9062, opened 2025-03-27
static css

mariuslundgard:


bjoerge:
Continued in https://github.com/sanity-io/sanity/pull/10668

PR #9064, opened 2025-03-27
fix(form): workaround for mutate-on-wheel-issue for number inputs

bjoerge:
### Description
Fixes the issue where number inputs gets modified if scroll wheel is used while hovering over while having focus.

Modifying the input value based on scroll wheel movement is the native browser behavior _when and only when_ a [`wheel`](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event) event listener is added to any element on the page, and because React eagerly adds listeners for all events to the root container, this opts-in to the mutate-on-wheel behavior.

The issue is currently present in latest Chrome and Safari. Firefox has disabled the mutate-on-wheel behavior entirely as of v130, and Safari has disabled it in the latest Technology Preview.

Unfortunately, the workaround in this PR only applies to Chromium based browsers and the issue remains in Safari until they ship a version that includes the removal of the mutate-on-scroll behavior, but hopefully it won't be too long until that happens.

### What to review

This was the least intrusive workaround I could find. I'd love to hear it if you have better ideas about how it can be fixed.

Other workarounds I've seen and considered was:
- calling `event.preventDefault()` and leave it at that. This is bad, because the user expectation is to still be able to scroll the page/container while hovering over the input.
- calling `event.currentTarget.blur()` when scrolling. This is bad because it loses focus and impairs accessibility.

I've tested that this doesn't introduce any unintended side effects in latest Firefox, Chrome and Safari.

### Testing
Existing tests should be enough. Hard to test for this particular issue.

### Notes for release
- Fixes the issue where number inputs gets modified if scroll wheel is used while hovering over while having focus. Note: the issue remains in Firefox

### DIFF
--- a/packages/sanity/src/core/form/inputs/NumberInput.tsx
+++ b/packages/sanity/src/core/form/inputs/NumberInput.tsx
@@ -1,8 +1,56 @@
 import {TextInput} from '@sanity/ui'
+import {useEffect} from 'react'
 
 import {type NumberInputProps} from '../types'
 import {getValidationRule} from '../utils/getValidationRule'
 
+function getScrollableParent(node: EventTarget) {
+  if (!node || !(node instanceof Element)) return null
+
+  let parent = node.parentElement
+
+  while (parent) {
+    const style = window.getComputedStyle(parent)
+    const overflowY = style.overflowY
+    const isScrollable = overflowY === 'auto' || overflowY === 'scroll'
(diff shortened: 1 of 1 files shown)


jordanl17 (approved):
I've tried a few ideas but can't find anything that works or is as self contained as you've got it.

I feel with the great comments this is fine to go with. Great fix!!

PR #9065, opened 2025-03-27
feat(core): use project scoped media-libraries endpoint

skogsmaskin:
### Description

We have a new endpoint available under the project API host to fetch the projects media libraries. 

This will use the same auth and CORS settings as with everything else in the studio.

Also we get rid of one hop to fetch the organization id.

### What to review

That it works.

### Testing

Can be tested in the test-studio by selecting "Media Library". The test studio is configured without specifying `libraryId`.

### Notes for release

N/A - part of yet unannounced code.

### DIFF
--- a/packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
+++ b/packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
@@ -43,23 +43,16 @@ const MediaLibraryAssetSource = function MediaLibraryAssetSource(
     if (libraryIdProp || fetchedLibraryId) {
       return
     }
-    client.request({uri: `/projects/${projectId}`}).then((project) => {
-      const {organizationId} = project
-      client
-        .withConfig({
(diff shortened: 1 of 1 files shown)


packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibrarySource.tsx
```diff
@@ -43,23 +43,16 @@ const MediaLibraryAssetSource = function MediaLibraryAssetSource(
...
-            // Add to cache for this project (organization)
-            if (projectId) {
-              fetchedLibraryIdCache.set(projectId, libraryIdFromResult)
-            }
-            setFetchedLibraryId(libraryIdFromResult)
-          }
-        })
+    if (!projectId) {
+      throw new Error('projectId is required to fetch Media Library ID')
+    }
```
skogsmaskin, line 48: `projectId` is derived from the client, so the chance of hitting this error is about zero.

I do think throwing is correct here though as opposed to early return or if statement.

sgulseth (approved):

PR #9067, opened 2025-03-27
fix: resolving issue where string state is passed as a route parameter to router.navigate

jordanl17:
### Description
in cases where `router.navigate({state: 'stateRouteValue'})` is called, this was wrongly matching on the type guard `isNavigateOptions` for thinking this was the newer `navigate` signature (where `state` and other options are provided together as a single parameter).

Updating the guard to verify that `state` must be an object (to match `type RouterState = Record<string, unknown> & {_searchParams?: SearchParam[]}`) resolves the issue by correctly identifying `{state: 'stateRouteValue'}` as being the older `navigate(state, options)` signature

### What to review
Improvements to the type guard `isNavigateOptions`

### Testing
Added multiple tests specifically for these scenarios

### Notes for release
* Fixes issue where Scheduled Publishing would not be accessible from the Studio tools bar
* [DEV] Fixes a regression in `router.navigate` where a `RouterState` with a `state` key/value pair would not perform expected navigation

### DIFF
--- a/packages/sanity/src/router/__test__/RouterScope.test.tsx
+++ b/packages/sanity/src/router/__test__/RouterScope.test.tsx
@@ -427,4 +427,54 @@ describe('RouteScope', () => {
       },
     })
   })
+
+  describe('navigate with string state', () => {
+    it('should handle navigation with string state in options', () => {
+      const wrapper = createWrapper()
+      const {result} = renderHook(() => useRouter(), {wrapper})
+
+      act(() => {
+        result.current.navigate({
+          state: 'stringState',
+        })
+      })
+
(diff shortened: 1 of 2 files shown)


packages/sanity/src/router/__test__/RouterScope.test.tsx
RitaDias, line 1: fantastic tests thank you 🙏

RitaDias (approved):
great catch, I can see that the code works as intended and have nothing to point out.
I think maybe we could benefit with more explicit code explanation on what each condition actually does / expects but it's not a blocker

PR #9071, opened 2025-03-28
feat(dev): shared settings with configurable options

skogsmaskin:
### Description

After introducing the GDR type in the test studio, we need to be able to create GDRs to reference environment specific resource-ids. 

I have added some factory functions for the sharedSettings plugin and the test studio schema types that makes this configurable based on the projectId used.


juice49 (approved):
Nice!

juice49 (approved):

PR #9086, opened 2025-03-31
refactor(ci): prepare for renaming next-branch to main

bjoerge:
### Description
This prepares for renaming `next` to `main`, mostly updates to workflow definitions and docs.

Once this is merged, I'll rename the main branch of this repo to uhm… `main`. This will hopefully not cause too much trouble, as GH should make sure open PRs get `main` as their new base automatically, and if I'm not mistaken GitHub will also set up redirects from urls pointing at `next` to `main`

### Testing
Don't think this can (or should) be subject to automated testing. I'll do the required manual testing once the rename is done.

### Notes for release
n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,10 +6,11 @@ workflows:
     jobs:
       - build:
           filters:
+            tags:
+              only: /^v.*/
             branches:
-              only:
-                - current
-
+              # Jobs will run each commit unless explicitly told to ignore branches
+              ignore: /.*/
 jobs:
   build:
     docker:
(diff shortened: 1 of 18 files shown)


bjoerge:
> LGTM, but remember to also update the UI E2E after the rename: [sanity-io/ui@8ff3c5c/.github/workflows/main.yml#L84](https://github.com/sanity-io/ui/blob/8ff3c5c8f334026c698c37a9c845f66b5f7bab1b/.github/workflows/main.yml#L84)

Ah, thanks – would totally have missed that

PR #9089, opened 2025-03-31
fix(core): reset remote transactions when a draft is published

pedrobonamin:
### Description

We use the remote transactions which are transactions received after the user has already loaded the document and the initial transactions, to render the review changes screen.
In the events store we have a listener to catch all of this transactions and apply the changes to the review screen in a "live" mode, every new transaction received will be shown.

When a draft document is published we receive a `delete` event, because the draft document is removed, but we were keeping around the remote transactions, and that causes an issue because now the transactions cannot be reconciled in a correct way, to fix this, when a `delete` or `publish` event is received and we trigger a new `onRefetch` action for the remote transactions we need to also declare the remote transactions as a new empty array.

This also adds a `catchError` to the observable so in case anything goes wrong we don't crash the studio and instead we show the error in the review changes screen.

### Testing
- create new document
- edit a field
- open history - review changes screen
- publish document
- edit a field

### Notes for release
Fixes an issue in the review changes screen when draft documents were published and the user continues editing it, while the review changes screen is open.

### DIFF
--- a/packages/sanity/src/core/store/events/getDocumentChanges.ts
+++ b/packages/sanity/src/core/store/events/getDocumentChanges.ts
@@ -3,6 +3,7 @@ import {diffInput, wrap} from '@sanity/diff'
 import {type SanityDocument, type TransactionLogEventWithEffects} from '@sanity/types'
 import {applyPatch, incremental} from 'mendoza'
 import {
+  catchError,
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
LGTM

PR #9100, opened 2025-04-01
fix: renames unreleased schema cli command from store to deploy

snorrees:
### Description

The general consensus is that `sanity schema store` should be `sanity schema deploy`, so here we are.

This PR renames the command and action to use "deploy/deployed" wording throughout. 

I left wording for the concept of a "schema store", ie [the _feature_](https://github.com/sanity-io/sanity/pull/8966) is the precursor to _the_ schema store, and you `deploy` schemas to said store.

### What it looks like

Still feature toggled (the below assumes `export SANITY_CLI_SCHEMA_STORE_ENABLED=true`)

```bash
sanity schema
usage: npx sanity schema [--default] [-v|--version] [-d|--debug] [-h|--help] <command> [<args>]

Commands:
   delete    Delete schema documents by id.
   deploy    Deploy schema documents into workspace datasets.
   extract   Extracts a JSON representation of a Sanity schema within a Studio context.
   list      Lists all schemas in the current dataset.
   validate  Validates all schema types specified in a workspace.
````

```bash
sanity schema deploy --help
usage: npx sanity schema deploy 

   Deploy schema documents into workspace datasets.

**Note**: This command is experimental and subject to change.

This operation (re-)generates a manifest file describing the sanity config workspace by default.
To re-use an existing manifest file, use --no-extract-manifest.

Options:
  --workspace <workspace_name> deploy schema for a specific workspace
  --id-prefix <prefix> add a prefix to the schema id
  --manifest-dir <directory> directory containing manifest file (default: ./dist/static)
  --no-extract-manifest disables manifest generation – the command will fail if no manifest exists
  --verbose print detailed information during deployment

Examples
  # Deploy all workspace schemas
  sanity schema deploy

  # Deploy the schema for only the workspace 'default'
  sanity schema deploy --workspace default

  # Runs using a pre-existing manifest file
  # Config changes in sanity.config will not be picked up in this case
  sanity schema deploy --no-extract-manifest
```

### What to review

This PR is 95% IDE rename symbol/file.

I did change the wording of some messages in "deploySchemaAction.ts" to say "deploy/deployed" instead of "store/stored" manually.

The most important thing to review carefully is the `deploySchemaCommand` man page. It should no longer use "store" as a concept, only "deploy".

### Testing

I did not make a tagged release, so to test this you have to

- check out the branch:
- `pnpm bootstrap`
- `cd` into `dev/test-create-integration-studio` (test-studio does not work with manifest extract)
- `export SANITY_CLI_SCHEMA_STORE_ENABLED=true`
- `sanity schema`
- `sanity schema deploy --help` 
- `sanity schema deploy`

### Notes for release

Still feature toggled. No notes.


packages/sanity/src/_internal/cli/commands/schema/deploySchemaCommand.ts
```diff
@@ -1,6 +1,6 @@
 import {type CliCommandDefinition} from '@sanity/cli'
```
snorrees, line 1: This file in particular needs review: It should only use "deploy" vocabulary, not "store".

robinpyon (commented):
Thanks @snorrees 

I did a quick search for 'store' in `packages/sanity/src/_internal/cli` –

**Possibly blocking:**
- Should we rename the env variable too? (`SANITY_CLI_SCHEMA_STORE_ENABLED` -> `SANITY_CLI_SCHEMA_DEPLOY_ENABLED`?)
- There are references to `store` in the help page for `sanity schema list`. We may want to remove these? 
  - `Get **stored** schemas as pretty-printed json-array`
  - `Get singular **stored** schema as pretty-printed json-object`

**Non-blocking:**
- Is it worth renaming all associated types and files too? E.g. `schemaStoreTypes.ts`, `type SchemaStoreActionResult` etc, will defer to the Studio team here!

robinpyon (approved):

bjoerge (approved):

PR #9105, opened 2025-04-01
fix(core): add chevron to file browse sources button

skogsmaskin:
### Description

There's a discrepancy between the File og Image Input:

image

File Input's "Select"-button doesn't have a chevron, while ImageInput does.

Added the chevron to File Input to have consistency in the design.

After:

image

### Notes for release
Too small to be mentioned?


bjoerge (approved):
Nice catch :)

PR #9107, opened 2025-04-01
fix(presentation): use perspectiveStack for useDocumentLocations listen

pedrobonamin:
### Description

When using the `select` option in presentation for the [Document Location Resolvers](https://www.sanity.io/docs/presentation-resolver-api#8d8bca7bfcd7) we were not passing the perspective neither using the release id of the document to listen for it's changes. 
This PR updates this hook to get the changes when using versions, so now the resolver is updated accordingly.

**Future refactor**
Not sure why this is not using the `documentPreviewStore` and it's setting a new listener instead, it would be nice to reuse the observeDocumentPaths function which takes a perspective. The `listenQuery` function doesn't support a perspective for the listener.

### Notes for release
Fixes an issue in presentation when using `select` in the resolvers and working on a release document.

### DIFF
--- a/packages/sanity/src/presentation/useDocumentLocations.ts
+++ b/packages/sanity/src/presentation/useDocumentLocations.ts
@@ -1,10 +1,13 @@
 /* eslint-disable @typescript-eslint/no-shadow */
+import {type StackablePerspective} from '@sanity/client'
 import {get} from 'lodash'
 import {useEffect, useMemo, useState} from 'react'
 import {isObservable, map, type Observable, of, switchMap} from 'rxjs'
 import {
   type DocumentStore,
   getDraftId,
+  getPublishedId,
+  getVersionId,
   isRecord,
   isReference,
   type Previewable,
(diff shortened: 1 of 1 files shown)


packages/sanity/src/presentation/useDocumentLocations.ts
```diff
@@ -46,24 +49,37 @@ function cleanPreviewable(id: string | undefined, previewable: Previewable) {
...
+    listen: `*[_id in $versions]`,
+  }
+  const params = {
+    id,
+    versions: perspectiveStack
+      .map((p) => {
+        if (p === 'drafts') return getDraftId(id)
+        return getVersionId(getPublishedId(id), p)
+      })
+      .concat(getPublishedId(id)),
```
pedrobonamin, line 71: Creates an array of versions to listen to, if it's a release id it will create a version id to listen to:
e.g.
```
["versions.foo.bar", "drafts.bar", "bar"]
```

The store.listenQuery doesn't support a perspective for the listener, so listening to the changes in all the documents that could be affected by this perspective is the safest approach.

jordanl17 (approved):

ChrisNolan:
@pedrobonamin Will this fix address the "ChannelError: Unable to parse value of “$version=undefined”. Please quote string values." error which has been happening since 3.80.1 with the document resolver?

that people have [mentioned in the community slack](https://sanity-io-land.slack.com/archives/C9Z7RC3V1/p1744031749987579)?

PR #9131, opened 2025-04-02
feat(studio): add version info/about dialog

bjoerge:
### Description

If installing the Studio from a tagged release or using pkg.pr.new, version numbers often gets too long to fit in the version menu. For example:
image

This information is too useful to cut off like this, so when looking into a fix, I figured why not make the version menu item pop a dialog with a bit of additional info with an easy way to copy it to the clipboard. So now, when you click the version info you'll now see this:
image

### What to review
- Does it make sense?
- Is it a good start?
- What other information can we consider adding here?

### Testing
Check out the preview build here: https://test-studio-git-more-version-details.sanity.dev/

### Notes for release
- Now opens a version info dialog when clicking the "Current version"-menu item in the help resources from the Studio navigation bar

### DIFF
--- a/packages/sanity/src/core/i18n/bundles/studio.ts
+++ b/packages/sanity/src/core/i18n/bundles/studio.ts
@@ -10,6 +10,32 @@ import {type LocaleResourceBundle} from '../types'
  * @hidden
  */
 export const studioLocaleStrings = defineLocalesResources('studio', {
+  /** "Disabled" status for auto-updates in About-dialog */
+  'about-dialog.version-info.auto-updates.disabled': 'Disabled',
+  /** "Enabled" status for auto-updates in About-dialog */
+  'about-dialog.version-info.auto-updates.enabled': 'Enabled',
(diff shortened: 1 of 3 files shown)


rexxars (approved):
Looks good to me!

rexxars:
In the future it might be cool to add even more debug-related information here, like [which HTTP version the client is using](https://github.com/sanity-io/sanity/pull/8848/files#diff-0800f0a2be455e8d8bce22b6a0c0a310f071308162a674a0658509deda6cb8e5) to communicate with the API etc.

bjoerge:
> In the future it might be cool to add even more debug-related information here, like [which HTTP version the client is using](https://github.com/sanity-io/sanity/pull/8848/files#diff-0800f0a2be455e8d8bce22b6a0c0a310f071308162a674a0658509deda6cb8e5) to communicate with the API etc.

Oh, wasn't aware of that PR – that's amazing! A great candidate for information to show here!

PR #9140, opened 2025-04-04
test(e2e): update e2e reference test

RitaDias:
### Description

For some reason when running the test the dropdown doesn't want to open when clicking the arrow down button. However, when testing manually there is no issue. It is also not a consistent bug / behaviour.
However, using the input to trigger the input to open the list seems to work consistently and since the button behaviour is not what we are checking I've just changed that.

### What to review

Does this make sense?

### Testing

It's all test

### Notes for release

N/A


bjoerge (approved):
Great find and fix!

PR #9142, opened 2025-04-04
chore(test): deploy e2e project to vercel

pedrobonamin:
### Description

> [!WARNING]
> Before merging, confirm the VERCEL org token to use  

TLDR; run e2e tests against a vercel environment instead of local running project

**What's the motivation for this?**
In many opportunities we have found ourselves debugging tests which we couldn't replicate the failure locally due to the differences in the environments, having a deployed site in where we can run again the tests will help us find the reason why some tests are flaky.
Demand less from the virtual machines, running the studio and the playwright environment could lead to an overload in the virtual machines we are using, doing this change and offloading the studio to a deployed environment will help us have a more stable and less flaky test environment, as seen below where we have 5 successful runs in a row

image

#### New vercel project

Introduce a new vercel project for the studio deployments, this project has a small gotcha, it won't deploy automatically the site, instead it will be deployed by the e2e github action, you may be wondering why, the reason behind this is that we need to have in each deployment a different environment for the datasets, controlling the build action allows us to control the environment variables that we are using.

#### Update the e2e github action

Now this action will take care of building the e2e project and deploying it to vercel, after deployed, it will save the preview URL and pass it to the tests runners, it will also post a message to the PR with useful data for [debugging](https://github.com/sanity-io/sanity/pull/9142#issuecomment-2778585558) 

**Before**
Screenshot 2025-04-04 at 10 58 58

**Now**
Screenshot 2025-04-04 at 14 51 51

#### Workspaces in studio-testing

We want one dataset per project we are using for testing, chromium and firefox.
We are building the project upfront and deploying it to vercel, so we cannot have a custom environment for each runner as before. 
To work our way around this, I've added two new projects in which each of them will have the correct environment variables set up, with the dataset needed for each of them.
This also required that the navigation changes from `/test` to either. `/chromium` or `/firefox` this is handled by the monkey patched page.goto function (see below).

> [!NOTE]
> **Gotcha**  if you are running locally your tests using a deployed environment, be sure to update the SANITY_E2E_DATASET env variable to match with the project you are using or the client actions the test is running won't target the correct dataset.

####  Adds `studio-test`

We need more control over how the navigation works in the studio testing environment, we do this by patching the `page.goto` function to preserve the workspace when navigating.
The createDraftDocument function is updated to wait until the form appears and is not read only, to avoid flakyness on the tests.

We want to control all the testing functions from the same repo where we are creating the tests, `@sanity/testing` is great and it help us solve a lot of the pain points, the idea is to bring that into the repo and get what we need from it, while allowing us to make it grow without the overhead of deploying a new repository and maintaining two different ones, this is a step in that direction.

### Flakyness fixes
- Creates draft document now waits for the document to be editable, this will help us prevent errors
- Fixes ImageArrayDrag.spec flaky test
- Fixes Array.spec flaky test
- Attemps to fix inline-comment tests, by reverting the change introduced in this [PR](https://github.com/sanity-io/sanity/pull/9057) and removing the creation test, this tests were duplicated, we can centralize the action in one test
- Refactors the `deprecatedFields` test, we don't need one test for each field, we can test all the fields in teh same one, running one test per field mens we trigger one visit to the studio for each of the fields, adding 16 unnecessary visits and draft documents creations.

#### Future steps

- Investigate how to correctly link the vercel deployment with the branch, and see if it's possible to have the action update the [message posted](https://github.com/sanity-io/sanity/pull/9142#issuecomment-2777995013) by the vercel action 
- Create a new dataset per action run, not per PR, so we have a new fresh slate in each test run, see the [setup](https://github.com/sanity-io/sanity/blob/studio-e2e-deployment/scripts/e2e/setup.ts) action for this, while doing this **don't forget** to update the [cleanup](https://github.com/sanity-io/sanity/blob/studio-e2e-deployment/scripts/e2e/cleanup.ts) function to remove all the extra datasets.
- Investigate if we can remove the [`build-cli`](https://github.com/sanity-io/sanity/blob/studio-e2e-deployment/.github/workflows/e2e.yml#L42-L44)(which takes around 4 minutes) part of the setup action and trigger the changes using the http endpoints directly.
- Remove the `@sanity/testing` package from the studio

### What to review

- Please help me figure out if something is missing, special attention in the github action change

### Testing
Tests should pass.
The environment should be deployed to vercel.
The artifacts generated from the tests should continue working as before.


test/e2e/studio-test.ts
```diff
@@ -0,0 +1,41 @@
...
+  },
+  createDraftDocument: async ({page, _testContext}, use) => {
+    async function createDraftDocument(navigationPath: string) {
+      const id = _testContext.getUniqueDocumentId()
+
+      await page.goto(`${navigationPath};${id}`)
+
+      await page.waitForSelector('[data-testid="form-view"]', {state: 'visible', timeout: 30_000})
+
+      await page.waitForSelector('[data-testid="form-view"]:not([data-read-only="true"])', {
```
pedrobonamin, line 30: Adds this two wait so the draft is ready to be edited before continuing the test

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,86 @@ interface InlineCommentCreationTestProps {
...
-      await page.reload()
-      await page.waitForLoadState('load', {timeout: WAIT_OPTIONS.timeout * 2})
-    }
-  }
-
-  // If we couldn't get the comment button to appear after all attempts, fail the test
-  if (!commentButtonVisible) {
-    throw new Error('Comment button did not appear after multiple attempts')
-  }
+
```
pedrobonamin, line 20: This looks like a lot, but it is basically reverting the changes introduced by this [PR](https://github.com/sanity-io/sanity/pull/9057) and doing all the checks in one unique test.

test/e2e/studio-test.ts
RitaDias: Any reason to use the WaitForSelector instead of asserting visibility? :) 
I'm asking because playwright [usually discourages the use of waitFor overall and suggests a visibility expect approach](https://playwright.dev/docs/api/class-elementhandle#element-handle-wait-for-selector) instead

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,86 @@ interface InlineCommentCreationTestProps {
...
+  const button = page.locator('[data-testid="inline-comment-button"]')
+  await expect(button).toBeVisible()
+  button.click({
     delay: 1000,
   })
 
   // 5. Ensure the comment input field is displayed and the selected text is visually marked for commenting.
-  await page.getByTestId('comment-input').waitFor({
+  await page.waitForSelector('[data-testid="comment-input"]', {
```
RitaDias, line 48: This will be a comment in the overall parts that use the waitForSelector instead of getTestById based on my comment in a previous file

pedrobonamin: Thanks for this review, [updated](https://github.com/sanity-io/sanity/pull/9142/commits/fc4b636a5eadaf57582be7be7997a67081b0d80c) the uses of this for locator or getByTest id if it's more accurate.

dev/studio-e2e-testing/sanity.config.ts
```diff
@@ -20,15 +20,13 @@ import {customComponents} from './components-api'
 import {e2eI18nBundles} from './i18n/bundles'
 import {schemaTypes} from './schemaTypes'
 
-export default defineConfig({
+const defaultConfig = defineConfig({
   name: 'default',
   title: 'studio-e2e-testing',
 
   projectId: process.env.SANITY_E2E_PROJECT_ID!,
   dataset: process.env.SANITY_E2E_DATASET!,
```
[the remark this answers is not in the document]
pedrobonamin: If you are running it locally you will be fine with just a `SANITY_E2E_DATASET` , but we need the two different datasets for the deployed site, because we need to pass it on build to the site.

My plan is to remove this `SANITY_E2E_DATASET` and just accept the chromium or firefox one, but I need to bring first `@sanity/test` to the studio codebase, because we need to update how the client gets the dataset. It is currently reading it from `SANITY_E2E_DATASET`

pedrobonamin:
Thanks for the review @RitaDias 
I will need to update the e2e-ui file, wasn't aware of it but I look deeper into it and we are using to run the studio e2e tests from the sanity ui repo.

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
+
+  // 2. Click the editor
+  const editor = page.locator('[data-testid="pt-editor"]')
+  await editor.click()
+
+  // 3. Select all text in the editor.
+  await page.getByTestId('pt-editor').waitFor(WAIT_OPTIONS)
+  const textbox = page
+    .locator('[data-testid="pt-editor"]')
+    .locator('div[role="textbox"][contenteditable=true]')
```
RitaDias, line 32: I think this would work?

```suggestion
  const textbox = page.getByTestId('pt-editor').locator('div[role="textbox"][contenteditable=true]')
```

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
+  await textbox.selectText()
 
   // 4. Click on the floating comment button to start authoring a comment.
-  const getCommentButton = () => page.locator('[data-testid="inline-comment-button"]')
-  await expect(getCommentButton()).toBeVisible()
-  getCommentButton().click({
+  await page.getByTestId('inline-comment-button').waitFor({...WAIT_OPTIONS, state: 'visible'})
+
+  const button = page.locator('[data-testid="inline-comment-button"]')
```
RitaDias, line 38:
```suggestion
  const button = page.getByTestId('inline-comment-button')
```

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
 
   // 5. Ensure the comment input field is displayed and the selected text is visually marked for commenting.
-  await page.getByTestId('comment-input').waitFor({
-    ...WAIT_OPTIONS,
-    state: 'visible',
-  })
-  const getCommentInput = () => page.locator('[data-testid="comment-input"]')
-  await expect(getCommentInput()).toBeVisible()
+  await page.getByTestId('comment-input').waitFor({...WAIT_OPTIONS, state: 'visible'})
+  const commentInput = page.locator('[data-testid="comment-input"]')
```
RitaDias, line 46:
```suggestion
  const commentInput = page.getByTestId('comment-input')
```

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
+  await expect(authoringDecorator).toHaveText(PTE_CONTENT_TEXT)
 
   // 6. Author the comment and submit it by clicking the send button.
-  await getCommentInput().locator('div[role="textbox"]').fill('This is a comment')
-  await getCommentInput().locator('[data-testid="comment-input-send-button"]').click({
-    delay: 1000,
-  })
+  const commentInputTextBox = commentInput.locator('div[role="textbox"]')
+  await commentInputTextBox.fill('This is a comment')
+  const sendButton = commentInput.locator('[data-testid="comment-input-send-button"]')
```
RitaDias, line 56:
```suggestion
  const sendButton = commentInput.getByTestId('comment-input-send-button')
```

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
-  await page.locator(addedDecorator).waitFor(WAIT_OPTIONS)
-  await expect(page.locator(addedDecorator)).toHaveText(PTE_CONTENT_TEXT)
+  await page.locator('[data-inline-comment-state="added"]').waitFor(WAIT_OPTIONS)
+  const addedDecorator = page.locator('[data-inline-comment-state="added"]')
+  await expect(addedDecorator).toHaveText(PTE_CONTENT_TEXT)
 
   // 8. Verify that the comments list is visible after the comment has been added.
   await page.getByTestId('comments-list').waitFor(WAIT_OPTIONS)
-  await expect(page.locator('[data-testid="comments-list"]')).toBeVisible()
+  const commentsList = page.locator('[data-testid="comments-list"]')
```
RitaDias, line 67:
```suggestion
  const commentsList = page.getByTestId('comments-list')
```

test/e2e/tests/comments/inline.spec.ts
```diff
@@ -16,173 +17,78 @@ interface InlineCommentCreationTestProps {
...
+  await expect(commentsList).toBeVisible()
 
   // 9. Verify that the comment appears within the list and correctly references the intended content.
-  const commentsListItem = '[data-testid="comments-list-item"]'
   await page.getByTestId('comments-list-item').waitFor(WAIT_OPTIONS)
-  await expect(page.locator(commentsListItem)).toBeVisible()
-  await expect(page.locator('[data-testid="comments-list-item-referenced-value"]')).toHaveText(
-    PTE_CONTENT_TEXT,
+  const commentsListItem = page.locator('[data-testid="comments-list-item"]')
```
RitaDias, line 72:
```suggestion
  const commentsListItem = page.getByTestId('comments-list-item')
```

test/e2e/tests/pte/ImageArrayDrag.spec.ts
```diff
@@ -27,17 +28,16 @@ test.describe('Portable Text Input - ImageArrayDraft', () => {
...
   }) => {
-    await createDraftDocument(
-      '/test/content/input-standard;portable-text;pt_allTheBellsAndWhistles',
-    )
+    await createDraftDocument('/content/input-standard;portable-text;pt_allTheBellsAndWhistles')
 
+    const pteEditor = page.getByTestId('field-body')
+    // Wait for the text block to be editable
+    await expect(
+      pteEditor.locator('[data-testid="text-block__text"]:not([data-read-only="true"])'),
```
RitaDias, line 36: This one can stay. this one is allowed 😄

RitaDias (commented):
I'd just suggest getting the getByTestId aspects I've suggested, but otherwise this is good for me! :)

RitaDias (approved):
Let's go!!

PR #9151, opened 2025-04-07
fix(studio): improve how presence works with releases

bjoerge:
### Description
This fixes a couple of issues with presence and content releases:

- Document lists did not display avatars of users working in versions of documents
- Clicking on an avatar for a user in the global presence menu always took you to the draft, not the actual version the user was in. This is now fixed by adding the version of the document as `perspective` param to the link. This also means that clicking a presence avatar from the global presence menu changes your studio perspective.

Note: Field presence is still scoped to version. I.e. field presence won't display users working in the same field in other versions of the document. I personally think this is the most sensible since the purpose of presence here is first and foremost to tell who are concurrently editing the exact field. We could consider visually indicating different versions here though.

### Testing
This would be good to have some tests for, I'll look into adding test cases for these.

### What to review
Does this make sense UX wise?

### Notes for release
- Improved how presence works with Content Releases

### DIFF
--- a/packages/sanity/src/core/form/useDocumentForm.ts
+++ b/packages/sanity/src/core/form/useDocumentForm.ts
@@ -227,9 +227,11 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
 
   const [presence, setPresence] = useState<DocumentPresence[]>([])
   useEffect(() => {
-    const subscription = presenceStore.documentPresence(value._id).subscribe((nextPresence) => {
-      setPresence(nextPresence)
-    })
+    const subscription = presenceStore
+      .documentPresence(value._id, {excludeVersions: true})
+      .subscribe((nextPresence) => {
+        setPresence(nextPresence)
+      })
     return () => {
       subscription.unsubscribe()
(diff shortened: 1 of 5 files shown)


packages/sanity/src/core/form/useDocumentForm.ts
```diff
@@ -227,9 +227,11 @@ export function useDocumentForm(options: DocumentFormOptions): DocumentFormValue
 
   const [presence, setPresence] = useState<DocumentPresence[]>([])
   useEffect(() => {
-    const subscription = presenceStore.documentPresence(value._id).subscribe((nextPresence) => {
-      setPresence(nextPresence)
-    })
+    const subscription = presenceStore
+      .documentPresence(value._id, {excludeVersions: true})
+      .subscribe((nextPresence) => {
```
EoinFalconer, line 232: could just be `(nextPresence) => setPresence(nextPresence)`

EoinFalconer (approved):
Looking good! Tested here too!

mariuslundgard (approved):
👍

PR #9153, opened 2025-04-07
feat(cli): ask to auto update on sanity dev, build and deploy

bjoerge:
### Description

Currently it's easy to miss it if you have auto-updates enabled and are running `sanity dev` with outdated local dependencies compared to what's currently served at runtime for your deployed studio. This can lead to situations where you get different behavior locally than after deployment (i.e. you get hit by a bug during dev that's fixed for your deployed studio). It might also be surprising that your local versions aren't updated, since you've opted into auto-updates after all.

As of this PR, when `sanity dev` is invoked, we'll check if auto-updates are enabled, and if it is, and you have outdated local dependencies, we'll ask you if you'd like to upgrade.

Example:
```
npx sanity dev
? The following local package versions are different from the versions currently served at runtime.
When using auto updates, we recommend that you test locally with the same versions before deploying. 

 - sanity (local version: 3.82.0, runtime version: 3.86.1)
 - @sanity/vision (local version: 3.82.0, runtime version: 3.86.1) 

Do you want to upgrade local versions? (Y/n)
```

For `build` and `deploy` you can chose betwee 3 options, since it's likely that the developer would want to do a round of testing before continuing with the deploy:
```
? Select existing studio hostname bjoerge-aus-debug
ℹ Building with auto-updates enabled
? The following local package versions are different from the versions currently served at runtime.
When using auto updates, we recommend that you test locally with the same versions before deploying. 

 - @sanity/vision (local version: 3.83.0, runtime version: 3.86.1) 

Do you want to upgrade local versions before deploying? (Use arrow keys)
❯ Upgrade and proceed with deploy 
  Upgrade only. You will need to run the deploy command again 
  Cancel 
```

Note that this goes both ways: Although rarely, it might happen that you have a have higher version locally than what's currently served at runtime. If this happens, you should still be prompted and asked if you want to sync local deps.

### What to review
We already had a set of utils for executing `<packageManager> install` in the global `@sanity/cli`-package, but those are not easily accessible from the project CLI code in the `sanity`-package. It didn't feel right to expose them via the context that's passed to the project cli, so I opted for copying the relevant code over to the project CLI instead. This isn't ideal, but I've tried to keep the divergence at a minimum so it shouldn't be that big of a deal to combine them at a later point (if one day we merge global and project CLI code)

I did consider running the upgrade without prompting, but it felt very heavy-handled, and it's probably something we rather want to support as a configuration option, e.g. instead of `autoUpdates: true`, `autoUpdates: {enabled: true, autoSyncLocalDeps}` (or something).

Note: I've also implemented the same prompt in `sanity build` (and as a consequence `sanity deploy`, however, here the default will be `false`.

### Testing

How to test:
- Initialize a new Studio
- Enable auto-updates (if not already enabled)
- Deploy the studio
- Install the branch preview build: `npm i https://pkg.pr.new/sanity-io/sanity@9153`
- Downgrade to older sanity packages locally (i.e. `@sanity/vision`)
- Run `npx sanity dev` or `npx sanity deploy` or `npx sanity build`
- You should now be prompted and asked if you'd like to upgrade outdated packages.

### Notes for release
- If enabling Auto-Updates and running `sanity dev` or `sanity deploy` with outdated dependencies locally, you will now be asked if you'd like to upgrade local versions.

### DIFF
--- a/packages/sanity/src/_internal/cli/util/packageManager/packageManagerChoice.ts
+++ b/packages/sanity/src/_internal/cli/util/packageManager/packageManagerChoice.ts
@@ -1,0 +1,204 @@
+/* eslint-disable no-process-env */
+import path from 'node:path'
+
+import {type CliPrompter} from '@sanity/cli'
+import preferredPM from 'preferred-pm'
+import which from 'which'
+
+import {isInteractive} from '../isInteractive'
+
+export type PackageManager = 'npm' | 'yarn' | 'pnpm' | 'bun' | 'manual'
+
+export const ALLOWED_PACKAGE_MANAGERS: PackageManager[] = ['npm', 'yarn', 'pnpm', 'bun', 'manual']
+export const allowedPackageManagersString = ALLOWED_PACKAGE_MANAGERS.map((pm) => `"${pm}"`).join(
+  ' | ',
+)
+
+const EXPERIMENTAL = ['bun']
+
+/**
+ * Attempts to resolve the most optimal package manager to use to install/upgrade
+ * packages/dependencies at a given path. It does so by looking for package manager
+ * specific lockfiles. If it finds a lockfile belonging to a certain package manager,
+ * it prioritizes this one. However, if that package manager is not installed, it will
+ * prompt the user for which one they want to use and hint at the most optimal one
+ * not being installed.
+ *
+ * Note that this function also takes local npm binary paths into account - for instance,
+ * `yarn` can be installed as a dependency of the project instead of globally, and it
+ * will use that is available.
+ *
+ * The user can also select 'manual' to skip the process and run their preferred package
+ * manager manually. Commands using this function must take this `manual` choice into
+ * account and act accordingly if chosen.
+ *
+ * @param workDir - The working directory where a lockfile is most likely to be present
+ * @param options - Pass `interactive: false` to fall back to npm if most optimal is
+ *                  not available, instead of prompting
+ * @returns Object of `chosen` and, if a lockfile is found, the `mostOptimal` choice
+ */
+export async function getPackageManagerChoice(
+  workDir: string,
+  options: {interactive: false} | {interactive?: true; prompt: CliPrompter},
+): Promise<{chosen: PackageManager; mostOptimal?: PackageManager}> {
+  const rootDir = workDir || process.cwd()
+  const preferred = (await preferredPM(rootDir))?.name
+
+  if (preferred && (await hasCommand(preferred, rootDir))) {
+    // There is an optimal/preferred package manager, and the user has it installed!
+    return {chosen: preferred, mostOptimal: preferred}
+  }
+
+  const mostLikelyPM = await getMostLikelyInstalledPackageManager(rootDir)
+  const interactive = typeof options.interactive === 'boolean' ? options.interactive : isInteractive
+  if (!interactive) {
+    // We can't ask the user for their preference, so fall back to either the one that is being run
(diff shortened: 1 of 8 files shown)


packages/sanity/src/_internal/cli/util/packageManager/packageManagerChoice.ts
```diff
@@ -0,0 +1,204 @@
...
+  if (await hasBunInstalled(cwd)) {
+    return 'bun'
+  }
+
+  return 'manual'
+}
+
+async function getAvailablePackageManagers(cwd: string): Promise<PackageManager[]> {
+  const [npm, yarn, pnpm, bun] = await Promise.all([
+    hasNpmInstalled(cwd),
```
jordanl17, line 103: COULD:
Wonder if we do need `hasXInstalled`, if we just used `hasCommand` directly here then could keep building the return result string array simpler

```
const ALLOWED_PACKAGE_MANAGERS = ['npm', 'yarn', 'pnpm', 'bun']
export const ALLOWED_PACKAGE_MANAGER_CHOICES = [...ALLOWED_PACKAGE_MANAGERS, 'manual']
...

const choices = await Promise.all(ALLOWED_PACKAGE_MANAGERS.map(pm => hasCommand(pm, cwd))
return [...choices, 'manual']
```

packages/sanity/src/_internal/cli/util/packageManager/upgradePackages.ts
```diff
@@ -0,0 +1,49 @@
...
+  const {output, workDir} = context
+  const execOptions: CommonOptions<'utf8'> = {
+    encoding: 'utf8',
+    env: getPartialEnvWithNpmPath(workDir),
+    cwd: workDir,
+    stdio: 'inherit',
+  }
+  const upgradePackageArgs = packages.map((pkg) => pkg.join('@'))
+  const npmArgs = ['upgrade', '--legacy-peer-deps', ...upgradePackageArgs]
+  let result: ExecaReturnValue<string> | undefined
```
jordanl17, line 24: COULD: The return cmd from [here](https://github.com/sanity-io/sanity/pull/9153/files#diff-27e6e7fb6f2bc20d9518d3f53517304afc4423b02cf633c498c0545f994b846c) not be used for the args instead?

Separately, looks like `bun` should be `update`

packages/sanity/src/_internal/cli/util/packageManager/packageManagerChoice.ts
bjoerge: I like that idea, definitely a lot more readable. Since this is mostly copied over from `@sanity/cli` I'd prefer to keep it as similar as possible (for now) in order to not make it harder than necessary to merge later.

packages/sanity/src/_internal/cli/util/packageManager/upgradePackages.ts
bjoerge: Oh, good catch. I've changed from `bun upgrade` => `bun update`.

> COULD: The return cmd from [here](https://github.com/sanity-io/sanity/pull/9153/files#diff-27e6e7fb6f2bc20d9518d3f53517304afc4423b02cf633c498c0545f994b846c) not be used for the args instead?
Probably is like this because `getUpgradeCommand` is meant as human-readable output, but agree it would be better to use a single source of truth. Again I'd like to keep the divergence with `@sanity/cli` at a minimum, so leaving it as-is :)

jordanl17 (approved):

PR #9154, opened 2025-04-07
chore(ci): trigger publish to pkg.pr.new from label

bjoerge:
### Description
Requiring approval in order to publish to pkg.pr.new is a bit tedious, so simplifying the workflow to instead require the `trigger: preview`-label (based on vite's config [here](https://github.com/vitejs/vite/blob/main/.github/workflows/preview-release.yml))

With this, all that's needed to publish to pkg.pr.new is to add the `trigger: preview` label. Adding a label requires repo write access, so we can also remove the additional approver write access check. Additionally, the workflow checks for the repo being `sanity-io/sanity`, so it will require repo push access as well, and won't allow publishing packages from forks.

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -1,29 +1,22 @@
 name: Publish Pull Requests with pkg.pr.new
 
+permissions:
+  pull-requests: write
+
 on:
-  pull_request_review:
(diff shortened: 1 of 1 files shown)


bjoerge:
publish pr

stipsan (approved):

PR #9156, opened 2025-04-07
feat(structure): add schema description to document form header if it exists

RitaDias:
### Description

Add tooltip with a schema description (document) if it exists

```  
description: 'This is just a simple type for generating some test data',
```

image

>[!Warning]
> You can add components that look different instead of being plain strings, we have considered this and since it's the default behavior we will leave it for now. A linear story has been set up to discuss this in the future

Examples:
```
description: (
    <div style={{textDecoration: 'line-through'}}>
      <Text size={1} muted>
        This is just a simple <strong>type</strong> for generating some test data
      </Text>
    </div>
  ),
```

image

### What to review

Should I add something else?

### Testing

N/A

### Notes for release

Add a tooltip to make the description for the schema type accessible in the UI


juice49:
This looks great! One thing I noticed is that the alignment between the text and the tooltip icon here doesn't look quite right. In your screenshot it looks a bit more spaced out, but it looks a bit cramped to me.

Screenshot 2025-04-10 at 09 58 02

juice49 (dismissed):
Looks good to me!

juice49 (approved):

PR #9164, opened 2025-04-09
chore: lowercase sanity.studio url

SimeonGriggs:
### Description

When deploying `Sanity.Studio` looks weird as it uses Title Case in a domain name.


RitaDias (approved):
Thanks!

jordanl17 (approved):

Issue #957, opened 2018-09-10
Ability to configure aspect ratios shown in image field / hotspot tool


bjoerge:
It should be an option on image fields which aspect ratio to show the image in. Also, it should be possible to configure the set of aspect rations to preview in the actual image tool.

dbbk:
Any news on this?

alevroub:
Similarly, the image tool lacks an option to easily crop the image into a specific aspect ratio.

derekaupma:
My use case is a very strict website design where images must be exact non-standard aspect ratios for each use.  A "hero" image for an Event is always 4x3.  A Carousel Slide image is always 14x9.  While a Callout image is 9x8.  Ideally I could set this aspect ratios in the image options:
```
export default {
  name: 'callout',
  fields: [
    ...,
    {
      name: "image",
      type: "image",
      hotspot: true,
      options: {
        aspectRatios: [
          ['9:8', 9 / 8],
        ]
      },
    }
  ]
}

procrates:
Any progress on this regarding v3? Would love to have the ability to configure the hotspot option with custom aspect ratios. I will take a look to see the code.

minervabot:
This is a major shortcoming for us right now. I am apprehensive to ship the crop tool to end users as it is now.

chrisandrew-dev:
+1 for this. IMO, content editors would expect the preview aspect-ratios presented in the crop tool to be predefined crop options, rather than indicative of a crop they can't reliably achieve.

6pBit:
I would appreciate this functionality and hope to see it implemented:) It would be very useful for our editors to see the relavant aspect ratio preview.

PR #9185, opened 2025-04-11
feat(core): add image schema options for hotspot tool configuration

jmswrnr:
### Description

This PR introduces an image schema option: the `hotspot` object to enable and configure the hotspot tool, this pattern is being mindful of future options for both the hotspot tool and previews.

Mentioned in #957 

### What to review

  - Currently, setting this object in schema options will enable the hotspot tool.
  - **Should we instead have a required `enabled` boolean inside this new options object?**

### Testing

I have verified image fields in the Test Studio:
- [x] The default preview aspect ratios are used if `previews` option is not set.
- [x] Hotspot boolean/undefined option still works as expected.
- [x] When setting an empty `previews` array, the tool does not show any previews.
- [x] A four column layout for previews is enforced to prevent unusually large previews.

### Notes for release

The `image` schema options now supports a `hotspot` object for configuring the image hotspot tool. This now enables custom aspect ratios for the crop previews to be configured:
```js
{
  name: 'hotspotImage',
  title: 'Hotspot image',
  type: 'image',
  options: {
    hotspot: {
      previews: [
        {title: '2:1', aspectRatio: 2 / 1},
        {title: '4:5', aspectRatio: 4 / 5},
        {title: '9:16', aspectRatio: 9 / 16},
      ],
    },
  },
},
```
This `hotspot` option still supports the boolean value to the enable/disable hotspot tool. The image hotspot tool will use default configuration if enabled with a boolean:
```js
{
  name: 'hotspotImage',
  title: 'Hotspot image',
  type: 'image',
  options: {
    hotspot: false
  },
},
```
```js
{
  name: 'hotspotImage',
  title: 'Hotspot image',
  type: 'image',
  options: {
    hotspot: true
  },
},
```

These are additive changes that require an update to the [image schema options spec docs](https://www.sanity.io/docs/image-type#options)

### DIFF
--- a/packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
+++ b/packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
@@ -1,4 +1,4 @@
-import {type Image, type ImageSchemaType} from '@sanity/types'
+import {type HotspotImagePreview, type Image, type ImageSchemaType} from '@sanity/types'
 import {Box, Card, Flex, Grid, Heading, Stack, Text} from '@sanity/ui'
 import {type ReactNode, useCallback, useEffect, useMemo, useState} from 'react'
 import {styled} from 'styled-components'
@@ -23,11 +23,11 @@ export interface ImageToolInputProps
 
 const HOTSPOT_PATH = ['hotspot']
 
-const PREVIEW_ASPECT_RATIOS = [
-  ['3:4', 3 / 4],
-  ['Square', 1 / 1],
-  ['16:9', 16 / 9],
-  ['Panorama', 4 / 1],
+const DEFAULT_PREVIEWS: HotspotImagePreview[] = [
+  {title: '3:4', aspectRatio: 3 / 4},
+  {title: 'Square', aspectRatio: 1 / 1},
+  {title: '16:9', aspectRatio: 16 / 9},
+  {title: 'Panorama', aspectRatio: 4 / 1},
 ] as const
 
 const DEFAULT_VALUE: Partial<Image> = {
@@ -76,6 +76,10 @@ export function ImageToolInput(props: ImageToolInputProps) {
   }, [value])
 
   const hasFocus = focusPath[0] === 'hotspot'
+
+  const hotspotPreviews =
+    (typeof schemaType.options?.hotspot === 'object' && schemaType.options.hotspot.previews) ||
+    DEFAULT_PREVIEWS
 
   useDidUpdate(hasFocus, (hadFocus) => {
     if (!hadFocus && hasFocus) {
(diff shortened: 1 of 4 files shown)


stipsan (dismissed):
I never thought I'd see the day 🥹🤌💖

>Should we instead have a required enabled boolean inside this new options object?
Nah, if you have custom previews defined for a hotspot but also want it disabled the only scenario I can think about is a temporary test. Maybe to test the image picker without hotspots vs enabled. :)

snorrees (dismissed):
This is very nice!

A couple of things I noticed while testing (non-blocking):
* When I drag the hotspot around with the mouse, I get intermittent lag-spikes, possibly because of GC? Is there room for update throttling here?
* Rows of previews could need some spacing (see the 1:900)
image
* There is an argument to be made that the previews should be capped at some max height. (1:900 works fine btw, but there is a bit of scroll required ;) ). 
* 999:1 is capped nicely :) image

packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
```diff
@@ -23,11 +23,11 @@ export interface ImageToolInputProps
 
 const HOTSPOT_PATH = ['hotspot']
 
-const PREVIEW_ASPECT_RATIOS = [
-  ['3:4', 3 / 4],
-  ['Square', 1 / 1],
-  ['16:9', 16 / 9],
-  ['Panorama', 4 / 1],
+const DEFAULT_PREVIEWS: HotspotPreview[] = [
```
pedrobonamin, line 26: Wdyt if we export this as DEFAULT_HOTSPOT_PREVIEWS? So if users want to add a new aspect ratio they can reuse the default ones?

pedrobonamin (dismissed):
Thank you for building this, regarding the `enabled` property I think it's ok not to have it, if users are adding the hotspot previews property they will want this to be shown.

packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
jmswrnr: I'm not sure if that would be a common use case but didn't see an issue with exporting it in case! However, we might consider sharing this `imagetool` with other repos. Perhaps it's best to bring this in with those exports/package instead of creating a new `sanity` export now

snorrees: Remember that this then is available in the global sanity export namespace, so the constant should use a fairly verbose name.

snorrees (approved):

packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
pedrobonamin: >Perhaps it's best to bring this in with those exports/package instead of creating a new sanity export now

Not sure how you will be adding those packages, but probably you won't be able to use exports from other packages in core, and rather you will want to export it from `core` and consume in other places. As there are some limitations to what you can import into `core`

If you add this new export you will need to update the exports.test snapshot, which was added [here ](https://github.com/sanity-io/sanity/issues/9006) as a guard to avoid adding unnoticed exports, but I think this will be a nice to have. However this is a nice to have wouldn't consider this blocking for the PR

pedrobonamin (approved):

packages/sanity/src/core/form/inputs/files/ImageToolInput/ImageToolInput.tsx
jmswrnr: Thanks both, @pedrobonamin appreciate the detailed context 🙌  I agree let's bring this in as-is and tackle those exports separately

PR #9195, opened 2025-04-14
Feat/asset sources uploads

skogsmaskin:


packages/sanity/src/core/form/inputs/files/FileInput/FileAsset.tsx
```diff
@@ -0,0 +1,344 @@
+import {isFileSource} from '@sanity/asset-utils'
+import {type AssetSource} from '@sanity/types'
+// eslint-disable-next-line no-restricted-imports
+import {Box, Card, type ThemeColorToneKey} from '@sanity/ui'
```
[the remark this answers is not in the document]
skogsmaskin: It's a leftover from the development stage. 👍

packages/sanity/src/core/form/inputs/files/FileInput/Browser.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <Button
+      text={t('inputs.file.browse-button.text')}
+      icon={SearchIcon}
+      mode="bleed"
+      // eslint-disable-next-line react/jsx-no-bind
+      onClick={() => handleSelectAssetFromSource(assetSources[0])}
+      data-testid="file-input-browse-button"
+      disabled={readOnly}
```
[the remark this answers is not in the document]
skogsmaskin: Hmm...I don't think so? It's the default case for when there is a single source (both read-only and not) or if the field has multiple sources and is read-only.

packages/sanity/src/core/form/inputs/files/FileInput/FileAsset.tsx
```diff
@@ -0,0 +1,343 @@
...
+      }
+    },
+    [elementProps],
+  )
+
+  const handleFilesOver = useCallback(
+    (fileInfo: FileInfo[]) => {
+      setHoveringFiles(fileInfo.filter((file) => file.kind !== 'string'))
+      let canUpload = true
+      const acceptedFiles = fileInfo.filter((file) => resolveUploader?.(schemaType, file))
```
jordanl17, line 78: COULD: These few lines are duplicated in a few places eg :162, so perhaps could be abstracted to a `canUpload` memo

packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
```diff
@@ -239,35 +234,83 @@ function BaseImageInputComponent(props: BaseImageInputProps): React.JSX.Element
...
+                  setMenuOpen(false)
+                  uploaderRef.current = null
+                  break
+                default:
+              }
+            }),
+            uploader: assetSource.uploader,
+          }
+          setIsUploading(true)
+          onChange(PatchEvent.from(createInitialUploadPatches(files[0])))
```
jordanl17, line 291: Q: Should we handle if `files[0]` is `undefined`?

packages/sanity/src/core/form/inputs/files/FileInput/FileAsset.tsx
skogsmaskin: I don't think it will be immediate enough for the `handleFilesOver` to respond correctly if derived from a memoed value.

packages/sanity/src/core/form/inputs/files/FileInput/types.ts
```diff
@@ -0,0 +1,24 @@
+import {type AssetFromSource, type AssetSource} from '@sanity/types'
+
+import {type FileInfo} from '../common/styles'
+import {type BaseFileInputProps} from './FileInput'
+
+export interface FileAssetProps extends Omit<BaseFileInputProps, 'renderDefault'> {
+  browseButtonElementRef: React.RefObject<HTMLButtonElement | null>
```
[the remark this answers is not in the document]
skogsmaskin: I get a type-error if I don't. 🤔

packages/sanity/src/core/form/inputs/files/ImageInput/ImageInput.tsx
skogsmaskin: Should be sufficient just to throw in beginning of the function if files[0] is undefined I think?

packages/sanity/src/core/form/inputs/files/common/UploadDropDownMenu.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const handleMenuItemClick = useCallback(
+    (event: React.MouseEvent) => {
+      const assetSourceName = event.currentTarget.getAttribute(ASSET_SOURCE_DATA_ATTRIBUTE)
+      const assetSource = assetSources.find((source) => source.name === assetSourceName)
+      if (!assetSource) {
+        return
+      }
+      const element = document.getElementById(createAssetSourceInputId(assetSource))
+      // Test for document.activeElement to avoid clicking the button twice
+      if (element && document.activeElement !== element) {
```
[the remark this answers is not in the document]
skogsmaskin: Yes, it's triggering it twice (both "click" with key event (enter or space) and this simulated click.

packages/sanity/src/core/form/studio/assetSourceDataset/uploader.ts
```diff
@@ -0,0 +1,172 @@
...
+      }
+      const fileId = uuid()
+      const uploadFile: AssetSourceUploadFile = {
+        id: fileId,
+        file,
+        progress: 0,
+        status: 'pending',
+      }
+      this.files.push(uploadFile)
+      if (uploader) {
```
jordanl17, line 71: Q: Should we check for the `uploader` before `files.push`, otherwise these files could be dangling if there's no uploader present?

packages/@sanity/types/src/assets/types.ts
```diff
@@ -172,4 +175,113 @@ export interface AssetSource {
...
+  type: 'status'
+  file: AssetSourceUploadFile
+  status: AssetSourceUploadFile['status']
+}
+
+/**
+ * Emitted when all files are done, either successfully, aborted or with errors
+ * @beta */
+export type AssetSourceUploadEventAllComplete = {
+  type: 'all-complete'
```
juice49, line 252: Not requesting a change, just sharing a thought. This seems the same as what we'd refer to as a "settled" promise in JS and what the `allSettled` function does.

packages/sanity/src/core/form/studio/assetSourceMediaLibrary/shared/MediaLibraryAssetSource.tsx
```diff
@@ -9,23 +9,27 @@ import {type ForwardedRef, forwardRef, memo, useCallback, useEffect, useState} f
...
 const fetchedLibraryIdCache: Map<string, string> = new Map()
 
-const MediaLibraryAssetSource = function MediaLibraryAssetSource(
+const MediaLibraryAssetSourceComponent = function MediaLibraryAssetSourceComponent(
   props: AssetSourceComponentProps & {libraryId: string | null},
   ref: ForwardedRef<HTMLDivElement>,
 ) {
   const {
-    accept,
+    accept, // TODO: make the plugin respect this filter
```
jordanl17, line 24: Q: Are these TODOs to be completed in this PR? Do you have follow-up tasks for this?

skogsmaskin: Not part of the scope right here as it will require some implementation on the Media Library side, that is not part of our launch scope features.

skogsmaskin: There are follow-up tasks for them, but it's not decided when or even if it will be implemented. I have made the TODO comments more optional by adding question marks behind them.

packages/@sanity/types/src/assets/types.ts
skogsmaskin: Yes. I didn't want a global "error" or "abort" event, as files may have varying statuses in the same upload (some may have errors, some may be aborted, or finished). I figured the only global event that makes sense is this "all stettled"-like event. 

Each file emits its status individually, for immediate error handling and similar.

packages/sanity/src/core/form/inputs/files/FileInput/FilePreview.tsx
```diff
@@ -0,0 +1,195 @@
...
+          icon={SearchIcon}
+          text={t('inputs.file.browse-button.text')}
+          onClick={handleSelectFileMenuItemClicked}
+          disabled={readOnly}
+          data-testid={`file-input-browse-button-${assetSources[0].name}`}
+          data-asset-source-name={assetSources[0].name}
+        />
+      )
+    }
+    if (assetSources.length > 1) {
```
juice49, line 93: Not a blocker, but it looks like the `if (assetSources.length > 1)` check and the final `return null` are redundant. The function could probably directly `return assetSources.map((assetSource) => {` at this point?

skogsmaskin: You're right about that. I have fixed it.

packages/sanity/src/core/form/inputs/files/common/FileInputButton/styles.ts
```diff
@@ -40,3 +40,40 @@ export const FileButton = styled(Button).attrs({forwardedAs: 'label'})(
...
+
+export const FileMenuItem = styled(MenuItem).attrs({forwardedAs: 'label'})(
+  ({theme}: {theme: Theme}) => {
+    const {focusRing} = theme.sanity
+    const base = theme.sanity.color.base
+    const border = {width: 1, color: 'var(--card-border-color)'}
+
+    return css`
+      &:not([data-disabled='true']) {
+        &:focus-within {
```
juice49, line 52: `focus-within` in the wild! Simply beautiful 🥲.

PR #9196, opened 2025-04-14
fix(core): update since logic for useEventsStore

pedrobonamin:
### Description
In the events store we were assuming that a Published event will exist and if that was not the case we were using the `events[1]`, this could lead to an incorrect view of the diff shown in history, because in some cases a published doesn't exist but a `Created` does.
This PR updates the logic to check also for the `Created` event if a `Published` event was not found when searching for the `@lastPublished` event.

### Testing
Documents with no published events, but multiple edits, should show the correct diff.

### Notes for release
Fixes an issue in where the revision tab would show incorrect events if a published event was not found and no document to compare was manually selected.


jordanl17 (approved):
Sweet!

PR #9197, opened 2025-04-15
WIP feat(cli): blueprints commands

tbeseda:
> Still testing these new `blueprints` commands to ensure they match functionality from `runtime-cli`

### DIFF
--- a/packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
+++ b/packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
@@ -27,9 +27,11 @@ const addBlueprintsCommand: CliCommandDefinition = {
       return
     }
 
-    const {blueprintsActions: actions} = await import('@sanity/runtime-cli')
+    const {blueprint: blueprintAction, resources: resourcesAction} = await import(
+      '@sanity/runtime-cli/actions/blueprints'
+    )
 
-    const existingBlueprint = actions.blueprint.findBlueprintFile()
+    const existingBlueprint = blueprintAction.findBlueprintFile()
     if (!existingBlueprint) {
       print('No blueprint file found. Run `sanity blueprints init` first.')
       return
@@ -53,7 +55,7 @@ const addBlueprintsCommand: CliCommandDefinition = {
         ],
       })
 
-      const {filePath, resourceAdded, resource} = actions.resources.createFunctionResource({
+      const {filePath, resourceAdded, resource} = resourcesAction.createFunctionResource({
         name: functionName,
         type: functionType,
         displayName: functionName,
--- a/packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
+++ b/packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
@@ -1,0 +1,80 @@
+import {type CliCommandDefinition} from '../../types'
+
+const helpText = `
+Arguments
+  [type]  Type of Resource to add (currently only 'function' is supported)
+
+Examples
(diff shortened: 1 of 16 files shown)


packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const functionType = await prompt.single({
+        type: 'list',
+        message: 'Choose function type:',
+        choices: [
+          {value: 'document-mutation', name: 'Document Mutation'},
+          {value: 'document-publish', name: 'Document Publish'},
+          {value: 'document-update', name: 'Document Update'},
+          {value: 'document-delete', name: 'Document Delete'},
+        ],
```
macdonst, line 55: Should we be giving folks all these options if we only support mutation currently?

packages/@sanity/cli/src/commands/blueprints/initBlueprintsCommand.ts
```diff
@@ -0,0 +1,96 @@
...
+      return
+    }
+
+    const blueprintExtension = await prompt.single({
+      type: 'list',
+      message: 'Choose a Blueprint manifest file type:',
+      choices: [
+        {value: 'json', name: 'JSON (Recommended)'},
+        {value: 'js', name: 'JavaScript (Beta)'},
+        {value: 'ts', name: 'TypeScript (Alpha)'},
```
macdonst, line 51: Again, should we expose these options?

packages/@sanity/cli/src/commands/functions/devFunctionsCommand.ts
```diff
@@ -30,8 +30,8 @@ const devFunctionsCommand: CliCommandDefinition = {
     const {print} = output
     const flags = {...defaultFlags, ...args.extOptions}
 
-    const {functionsActions} = await import('@sanity/runtime-cli')
-    functionsActions.dev.dev(flags.port)
+    const {dev} = await import('@sanity/runtime-cli/actions/functions')
+    dev.dev(flags.port)
```
macdonst, line 34: We might want to do:

```javascript
    const {dev: devAction} = await import('@sanity/runtime-cli/actions/functions')
    devAction.dev(flags.port)
```

So this looks less dumb.

macdonst (dismissed):
Just a couple of questions whether or now we should expose some options to users.

packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
tbeseda: Good question! a copy-paste from runtime-cli. I don't think we should allow these choices.

 I will update with real types and only enable the mutation.  
Though I fear the current type `"sanity.function.mutation"` found in the Blueprints spec is too generic right? We need a better list of function types.

packages/@sanity/cli/src/commands/blueprints/initBlueprintsCommand.ts
tbeseda: Technically they're all supported. But I think removing the option for dynamic blueprints is a good idea right now. thanks

packages/@sanity/cli/src/commands/blueprints/addBlueprintsCommand.ts
tbeseda: After looking into it more and learning how Functions is working right now, I think the correct default value right now is `sanity.function.document-publish`. I updated this PR to reflect that.

macdonst (approved):
My feedback is addressed. This is good to go.

PR #9198, opened 2025-04-15
fix(cli): Support unattended init within nextjs

runeb:
### Description

I discovered that the init command incorrectly prompts for user input even when running in unattended mode within Next.js. This PR addresses that issue by adding correct handling for unattended mode specifically within Next.js contexts.

Additionally, I noticed that the command currently exits with an error if the --output-path flag is omitted in unattended mode. However, the --output-path value isn't actually utilized when initializing within Next.js. Therefore, this PR also removes the requirement for the --output-path flag when running in unattended mode inside Next.js.

### What to review

This change affects unattended `sanity init` when ran in a nextjs project. See Testing for more details.

### Testing

That you can initialize unattended within a nextjs project. Before this change you would be prompted

> Would you like to add configuration files for a Sanity project in this Next.js folder?

in unattended mode. 

Additionally, you would have gotten an error if you did not specify `--output-path`. If you specified it, you would find that your Studio was not generated in that folder, but in src/, and your output-path folder would be blank except for a .env file with projectId and dataset. After this change the variables should be correctly added to the root env file instead.

### Notes for release

Support unattended `sanity init` within nextjs projects

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -315,12 +315,15 @@ export default async function initSanity(
   let initNext = false
   const isNextJs = detectedFramework?.slug === 'nextjs'
   if (isNextJs) {
-    initNext = await prompt.single({
-      type: 'confirm',
-      message:
-        'Would you like to add configuration files for a Sanity project in this Next.js folder?',
-      default: true,
-    })
+    initNext =
+      unattended ||
+      (await prompt.single({
+        type: 'confirm',
+        message:
+          'Would you like to add configuration files for a Sanity project in this Next.js folder?',
+        default: true,
+      }))
+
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -290,6 +290,8 @@ export default async function initSanity(
     print('')
   }
 
+  const isNextJs = detectedFramework?.slug === 'nextjs'
+
   const flags = await prepareFlags()
```
runeb, line 295: Need to set `isNextJs` before calling `prepareFlags` because we are using the value in the function. This is not great, but the function already similarly depends on other variables so this solution was deemed the least intrusive.

bjoerge (approved):
LGTM, great catch!

PR #9201, opened 2025-04-15
feat: Dashboard history (without SDK)

juice49:
### Description

This branch adds support for capturing Studio activity and sharing it with Dashboard via Comlink.

> [!IMPORTANT]
> These changes only affect Studio when rendered inside Dashboard. When rendered standalone, Comlink isn't instantiated and no events are captured.

### What to review

- The addition of the Comlink store.
- Capture of `viewed` event when viewing a document. For existing documents, this occurs once upon opening the document. For new documents, this occurs as soon as the user edits the document, causing it to be created.
- Capture of `edited` event when editing a document. This occurs once per document open to prevent inundating Dashboard with events.

### Testing

Running a Studio inside Dashboard and watching the network for mutations tagged `sanity.dashboard`.

### DIFF
--- a/packages/sanity/src/core/hooks/useDocumentOperationWithComlinkHistory.ts
+++ b/packages/sanity/src/core/hooks/useDocumentOperationWithComlinkHistory.ts
@@ -1,0 +1,147 @@
+import {useCallback, useMemo, useState} from 'react'
+import {useObservable} from 'react-rx'
+
+import {useRenderingContextStore} from '../store/_legacy/datastores'
+import {type OperationsAPI} from '../store/_legacy/document/document-pair/operations/types'
+import {useActiveWorkspace} from '../studio/activeWorkspaceMatcher/useActiveWorkspace'
+import {getVersionId} from '../util/draftUtils'
+import {useRecordDocumentHistoryEvent} from './useRecordDocumentHistoryEvent'
+
+interface Options {
+  api: OperationsAPI
+  publishedDocId: string
+  docTypeName: string
+  version?: string
+}
+
+/**
+ * Decorate operations API methods with Comlink history capture when relevant operations are executed.
+ *
+ * Event capture only occurs when Comlink is available.
+ *
+ * @internal
+ */
+export function useDocumentOperationWithComlinkHistory({
+  api,
+  publishedDocId,
+  docTypeName,
+  version,
+}: Options): OperationsAPI {
+  const {activeWorkspace} = useActiveWorkspace()
+
+  const {recordEvent} = useRecordDocumentHistoryEvent({
+    resourceType: 'studio',
+    documentId: version ? getVersionId(publishedDocId, version) : publishedDocId,
+    documentType: docTypeName,
+    resourceId: [activeWorkspace.projectId, activeWorkspace.dataset].join('.'),
+    schemaName: activeWorkspace.name,
+  })
+
+  const renderingContextStore = useRenderingContextStore()
+  const capabilities = useObservable(renderingContextStore.capabilities)
+
+  // Used to prevent redundant `edited` events being recorded.
+  const [hasRecordedEdit, setHasRecordedEdit] = useState<boolean>(false)
+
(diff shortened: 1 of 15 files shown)


packages/sanity/src/core/hooks/useDocumentOperationWithComlinkHistory.ts
```diff
@@ -0,0 +1,147 @@
...
+      withComlinkEvent({
+        ...comlinkContext,
+        operationName: 'patch',
+        comlinkEventType: 'edited',
+        preRecordEvent: preRecordPatch,
+      }),
+    [comlinkContext, preRecordPatch],
+  )
+
+  return withComlinkDelete(withComlinkDel(withComlinkPatch(api)))
```
jordanl17, line 97: COULD: reduce the repetition and make this more functional by using a `reduce` over all the decorator operations:

```ts
const decoratorConfigs = useMemo((): WithComlinkEventOptions<keyof OperationsAPI>[] => [
    {
      operationName: 'delete',
      comlinkEventType: 'deleted',
    },
    {
      operationName: 'del',
      comlinkEventType: 'deleted',
    },
    {
      operationName: 'patch',
      comlinkEventType: 'edited',
      preRecordEvent: preRecordPatch,
    },
  ], [comlinkContext, preRecordPatch])

const decoratedApi = useMemo(() => {
  return decoratorConfigs.reduce((acc, config) => {
    return withComlinkEvent({...comlinkContext, ...config})(acc)
  }, api)
}, [api, decoratorConfigs])

return decoratedApi
```
Or something like this

packages/sanity/src/core/hooks/useRecordDocumentHistoryEvent.ts
```diff
@@ -0,0 +1,115 @@
...
+  resourceType,
+  resourceId,
+  schemaName,
+}: UseRecordDocumentHistoryEventProps): DocumentInteractionHistory {
+  const {node} = useComlinkStore()
+
+  if (resourceType !== 'studio' && !resourceId) {
+    throw new Error('resourceId is required for media-library and canvas resources')
+  }
+
```
pedrobonamin, line 82: What if you add here a early return.
if (!node) {
    return (eventType: string) => {
     // noop, comlink store is not available.
    }
}

packages/sanity/src/core/hooks/useRecordDocumentHistoryEvent.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Hook for managing document interaction history in Sanity Studio.
+ * This hook provides functionality to record document interactions.
+ *
+ * @param documentHandle - The document handle containing document ID and type, like `{_id: '123', _type: 'book'}`
+ * @returns An object containing:
+ * - `recordEvent` - Function to record document interactions
+ * - `isConnected` - Boolean indicating if connection to Studio is established
+ *
+ * @example
```
jordanl17, line 47: Nice examples

juice49: Not mine—this is almost directly lifted from SDK 😅.

packages/sanity/src/core/hooks/useDocumentOperationWithComlinkHistory.ts
juice49: More functional than three piped functions? 😝

I do agree this is far from the most beautiful code written, and made be worth refactoring. However I would argue it's okay for now.

pedrobonamin (dismissed):
I think it looks good, just a minor nit comment.

packages/sanity/src/core/hooks/useRecordDocumentHistoryEvent.ts
juice49: Line 102 is intended to be a noop. No message will be posted if Comlink doesn't exist for any reason (either because Studio is not rendered in Dashboard, or Comlink isn't connected yet):

```
node?.post?.(message.type, message.data)
```

PR #9202, opened 2025-04-15
fix: excluding query params from pathname check when routing to tool

jordanl17:
### Description
When using query params, a trailing slash was always required after the URL route, before the `?`, otherwise the default tool and workspace would not load.

Now the trailing slash is not required, and query params are persisted during the routing

### Testing
Manually tested, but there are some existing unit tests around this to verify no regressions

### Notes for release
Fixes an issue that stops the studio loading when the URL does not include a trailing slash immediately preceding  a query string.

### DIFF
--- a/packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
+++ b/packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx
@@ -70,12 +70,11 @@ function useRouterFromWorkspaceHistory(
     // matches where the pathname is a false subset of the current pathname.
     const routerBasePathRegex = new RegExp(`^${escapeRegExp(routerBasePath)}(\\/|$)`, 'i')
     const shouldHandle = (pathname: string) => {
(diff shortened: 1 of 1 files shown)


jordanl17:
This pull request includes an update to the `useRouterFromWorkspaceHistory` function in the `WorkspaceRouterProvider.tsx` file to improve the accuracy of path matching. The change modifies the regular expression used to match the router base path to include patterns that match the end of the path, query parameters, or a trailing slash.

Improvements to path matching:

* [`packages/sanity/src/core/studio/workspaceLoader/WorkspaceRouterProvider.tsx`](diffhunk://#diff-c0cfea33d438e9d6f5122ac790e631575fd92951504ecf895c25cf67a44fbfb5L69-R75): Updated the regular expression in the `useRouterFromWorkspaceHistory` function to include patterns that match the end of the path, query parameters, or a trailing slash, preventing false matches where the pathname is a false subset of the current pathname.

pedrobonamin (approved):

PR #9215, opened 2025-04-16
fix: unpublished references fail validation

jordanl17:
### Description
Currently:
When a release exists where document A is to be published and references document B; document B is to be unpublished
When running the release, the release will fail due to referential failure (document A will reference an unpublished document)
Screenshot 2025-04-16 at 21 23 05

Now:
When the same scenario exists, document A will show a validation error due to document B being unpublished
The release overview will show validation errors
Screenshot 2025-04-16 at 21 22 32

Screenshot 2025-04-16 at 21 23 33

> [!NOTE]
> The copy for unpublish and missing published document will be the same for now, however a fast follow will refactor the way `getDocumentExists` works so that a reason can be exposed 

### What to review
* Does making the change in the `observeDocumentAvailability` observable make sense

### Notes for release
Fixes an issue with unpublishing documents in releases, where document validation would not show reference errors due to marking documents to be unpublished.

### DIFF
--- a/packages/sanity/src/core/validation/validators/objectValidator.ts
+++ b/packages/sanity/src/core/validation/validators/objectValidator.ts
@@ -1,6 +1,5 @@
 import {isReference, type Validators} from '@sanity/types'
 
-import {RELEASES_STUDIO_CLIENT_OPTIONS} from '../../releases/util/releasesClient'
 import {getPublishedId} from '../../util'
 import {genericValidators} from './genericValidator'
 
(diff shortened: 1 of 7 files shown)


packages/sanity/src/core/validation/validators/objectValidator.ts
```diff
@@ -50,8 +51,22 @@ export const objectValidators: Validators = {
       // a document should be able to reference itself without first being published
       return true
     }
+
     const exists = await getDocumentExists({id: value._ref})
     if (!exists) {
+      const publishedDocumentExists = await getClient(RELEASES_STUDIO_CLIENT_OPTIONS)
```
jordanl17, line 57: Is there a nicer way of doing this?
I looked at re-tying getDocumentExists to return a DocumentAvailability instead of a boolean, and although that would work, it seems like quite a large footprint to leave behind for this small impact
Alternatively, we could just show the same error in all cases, but that will likely lead to confusion, since in many cases the document will already be published, it is the effect of running the release that will result in it being unpublished

pedrobonamin (approved):
Nice work! Thank you!

PR #9216, opened 2025-04-17
fix(cli): address comments from RUN-341

macdonst:
### Description

This PR responds to feedback from a previous PR. 

See: https://linear.app/sanity/issue/RUN-341/feedback-from-sanitycli-pr

### What to review

./packages/@sanity/cli/src/commands/functions/** contains the various commands that leverage the actions and utils internal to the runtime-cli.

### Testing

Manual testing was performed but the bulk of the automated testing happens over in the `@sanity/runtime-cli` package.


tbeseda (approved):

packages/@sanity/cli/src/commands/functions/logsFunctionsCommand.ts
```diff
@@ -105,7 +105,10 @@ const logsFunctionsCommand: CliCommandDefinition = {
         for (const log of filteredLogs) {
           const {time, level, message} = log
           const date = new Date(time)
-          print(`${date.toLocaleDateString()} ${date.toLocaleTimeString()} ${level} ${message}`)
+          const dateString = flags.utc
+            ? date.toUTCString()
```
rexxars, line 109: The output format of this is quite verbose (`'Tue, 22 Apr 2025 17:06:04 GMT'`), and we already know this is in UTC given the flag. It's a bit subjective, but IMO I'd prefer something like an ISO-date, eg `2025-04-22T17:06:04.312Z` or `2025-04-22 17:06:04`. I guess the easiest here would be to just do:
```suggestion
            ? date.toISOString()
```

Alternatively, for more human readable but with less precision:
```suggestion
            ? date.toISOString().slice(0, 19).split('T').join(' ')
```

rexxars (approved):
Overall good, but left a note on the UTC flag formatting

tbeseda (approved):

PR #9224, opened 2025-04-21
feat(codegen): add multi-schema and projection

ricokahler:
### Description

This PR introduces a significant refactor to the `@sanity/codegen` type generation logic, primarily aimed at adding support for **multi-schema configurations** and **GROQ projection typing**, while maintaining backward compatibility for basic query and schema generation.

**Key Changes:**

- **Refactor:** The core type generation logic previously in `typeGenerator.ts` has been split into `SchemaTypeGenerator.ts` (handling schema-to-TS conversion and query/projection evaluation per schema) and `generateTypes.ts` (orchestrating the process across multiple schemas and generating the final output files, including typemaps).
- **Multi-Schema Support:** The `generateTypes` function now accepts an array of schemas, each associated with a project ID and dataset. Generated types are appropriately scoped using a new `DatasetScoped` helper type to prevent naming collisions and allow distinct types per source.
- **Projection Typing:** GROQ projections (e.g., `{ name, "slugValue": slug.current }`) are now explicitly identified and typed differently from full queries.
  - A new `SanityProjections` interface is generated in the typemaps.
  - Projection results are wrapped in a `DocumentTypeScoped<T, DocumentName>` helper type to indicate that the resulting type `T` applies specifically when the projection is used on a document of type `DocumentName`. This addresses the context-dependent nature of projection results.
  - `SchemaTypeGenerator.evaluateProjection` was added to handle this context-aware evaluation. Logic was included to simplify projections that result entirely in `null` fields for a given document type to `never` (or potentially `null` if warranted, though `never` is the current outcome when filtered).
- **Dual Module Augmentation:** Typemaps are now generated for both `@sanity/client` (overloading `client.fetch`) and the `groq` template tag function itself.
- **Testing:** Comprehensive unit tests have been added for `generateTypes` and `SchemaTypeGenerator`, covering schema conversion, query/projection evaluation, multi-schema handling, and edge cases. CLI e2e tests (`typegen.test.ts`) have been updated to reflect the new output and functionality.

**Goals:**

- Provide more accurate and useful types, especially for projections.
- Enable type generation for complex projects using multiple schemas.
- Improve Developer Experience (DX) by providing type information closer to the query definition (`groq` tag) and maintaining client fetch safety.

### What to review

- **Core Logic:** `packages/@sanity/codegen/src/typescript/generateTypes.ts` and `packages/@sanity/codegen/src/typescript/schemaTypeGenerator.ts`. Pay attention to the orchestration, multi-schema handling, query vs. projection evaluation, and typemap generation.
- **Projection Handling:** Review the `evaluateProjection` method in `schemaTypeGenerator.ts` and how `DocumentTypeScoped` is used.
- **Typemap Generation:** Examine how the `SanityQueries` and `SanityProjections` interfaces are conditionally generated in `generateTypes.ts` and declared in the augmented modules.
- **Tests:** Review the new unit tests (`generateTypes.test.ts`, `schemaTypeGenerator.test.ts`) for coverage and correctness. Review the updated e2e tests (`packages/@sanity/cli/test/typegen.test.ts`) for accurate assertions against the CLI output.
- **Configuration:** Check the handling of `augmentGroqModule` and `overloadClientMethods` flags.

### Testing

- Added comprehensive unit tests for `generateTypes` and `SchemaTypeGenerator` covering various schema structures, query types, projections, multi-schema scenarios, edge cases (like empty results, errors, duplicate queries), and stats calculation.
- Updated existing CLI e2e tests (`typegen.test.ts`) to validate the command output, generated file content structure, typemap generation, and configuration flag behavior against the refactored codegen logic. Fixtures were updated as needed.

### Notes for release

- **New Feature: Multi-Schema Support:** `sanity typegen generate` can now process multiple schemas defined in the configuration file. This is useful for projects integrating content from different Sanity projects/datasets or monorepos with varying schemas per package. Types generated from different schemas are automatically scoped to prevent conflicts.
- **New Feature: Projection Typing:** GROQ projections are now specifically typed.
  - **What is a projection?** In this context, a projection is a GROQ fragment (like `{ name, title }`) applied to one or more documents. The resulting type depends on the fields available in the _specific document type_ the projection runs against.
  - Generated projection types use a `DocumentTypeScoped<Result, DocType>` helper to link the `Result` type to the `DocType` it applies to.
  - A `SanityProjections` interface is now added to the `groq` module augmentation, mapping projection strings to their corresponding scoped result types.
- **Typemap Module Augmentation:**
  - Type generation now augments **both** `@sanity/client` and `groq` by default.
  - Augmenting `@sanity/client` provides type safety for `client.fetch(query)`.
  - Augmenting `groq` provides type hints directly when using the `groq` template tag (e.g., `groq`_\`_[_type == "author"]{name}\`\*`), improving DX at the query definition site.
  - This behavior can be controlled via the `overloadClientMethods` (for `@sanity/client`) and `augmentGroqModule` (for `groq`) flags in the generation config. Setting both to `false` will omit the typemaps entirely.
- **Potential Minor Breaking Changes:** While backward compatibility is maintained for core functionality, the exact structure of generated type names (e.g., due to improved sanitization/clash handling) and the typemap interfaces might differ slightly. Projects relying on the exact previous output structure may need minor adjustments. The default behavior now includes the `groq` module augmentation, which wasn't present before.
- **Usage:**

```typescript
import { groq } from 'groq';
import { sanityClient } from './sanityClient'; // Your configured client

// Type safety via client.fetch (using @sanity/client augmentation)
const posts = await sanityClient.fetch('*[_type == "post"]{ title }');
//    ^? const posts: { title: string; }[] | null

// Type safety via groq tag (using groq augmentation)
const query = groq`*[_type == "author"]{ name }`;
//    ^? const query: '*[_type == "author"]{ name }'
// Usage with fetch still benefits from client augmentation
const authors = await sanityClient.fetch(query);
//    ^? const authors: { name: string; }[] | null

// --- Projection Example ---
// Define a projection
const nameProjection = groq`{ name }`;

// Fetch data using the projection string (client augmentation helps here)
// Note: The specific type depends on how client.fetch handles projections;
// often it might return `any` or a base type if not further specialized.
// The real benefit is seeing the SanityProjections interface for reference.
const personName = await sanityClient.fetch(
  `*[_type == "person"][0] ${nameProjection}`,
);

// You can see the possible projection types via the SanityProjections interface:
type NameProjectionResult = Sanity.SanityProjections['{ name }'];
/* ^? type NameProjectionResult = DocumentTypeScoped<{ name: string | null; }, "person">
                                | DocumentTypeScoped<{ name: string | null; }, "page"> */
```


packages/@sanity/codegen/src/typescript/schemaTypeGenerator.ts
```diff
@@ -0,0 +1,381 @@
...
+
+      const projectionAst = safeParseQuery(projection)
+      if (projectionAst.type !== 'Object') {
+        throw new Error(
+          `Invalid projection syntax: Projections must be enclosed in curly braces, (e.g., "{_id, title}"). Received: "${projection}"`,
+        )
+      }
+
+      const ast = safeParseQuery(`* [_type==${JSON.stringify(schemaType.name)}] ${projection} [0]`)
+      const result = typeEvaluate(ast, this.schema)
```
sgulseth, line 279: Also want to double check the assumption here - we evaluate projection on _every_ schema, and we are fine with this being slow and generating a lot of unused types that are then reduced during type checking? 😅

ricokahler:
Hi @judofyr and @sgulseth for the review and insightful feedback. I really appreciate you taking the time to go through this large PR. I'm primarily replying to @judofyr's comments as they overlap with @sgulseth's

---

> This PR is _huge_ and contains a lot of completely independent features. It would have been a lot easier to review and merge these separately.

You're absolutely right. I misjudged the breadth of these changes. My intention was to move as fast as possible for the spring launch, and I opted to bundle features, then bolster confidence with extensive tests and refactors. In hindsight, this created a challenging review experience and limited opportunities for incremental discussion. This isn't how I want to operate, and I apologize for the disruption.

---

Regarding your specific points:

**`defineProjection` API Design & Alternatives**

> * Having `defineProjection` _not_ accept a `_type`, but instead generate _all_ possible projections for _all_ types seems _very_ unnecessary. We should strive to make CodeGen efficient even when you have hundreds or thousands of types. Our customers can have really complicated schemas. Why not something like `defineProjection(type: string | string[], query: string)` (with more specific generics)? This PR lacks some discussion around the different alternatives and trade-offs. It's not clear to me that this is the best API.

This is excellent feedback, and I appreciate the push on design and performance.

My goals with the current `defineProjection('{...}')` (no type argument) approach were:
*   **User Simplicity:** Allow developers to capture their intended projection string, with the system managing type inference.
*   **Automatic Discovery & "Correct" Typing:** The system attempts to infer applicable types, aiming to model the "correct total possible types" a general projection might return from GROQ.
*   **Reduced Clutter:** The idea behind `shouldOmitProjection` was to omit projections that would result in all-null fields, in the hope to avoid many unhelpful entries in the resulting union but this would still require computing each item in the union which as you've pointed out, can be explosive.
*   **Type Branding:** Each part of the union is branded (e.g., `ProjectionBase<{name: string}, 'author'>`), allowing users to discriminate types if necessary. The goal was to export useful helpers from a common package that allows relevant yet still optional type parameters for additional narrowing (e.g. based on a `TSchemaId extends string = string`, based on a `TDocumentType extends string = string`).

I share your concern about performance and the potential for type explosion with large schemas, as this involves parsing and evaluating the projection for each document type.

Building on your suggestion, perhaps an overloaded `defineProjection` could offer the best of both worlds?
1.  `defineProjection(projection: string)`: The current "do it all" behavior for simplicity by default.
2.  `defineProjection(projection: string, documentType?: string | string[])`: An explicit (and backwards compatible) way to target specific document types for better performance and more granular type generation.

I'm happy to implement this overloaded approach or discuss other alternatives. Consider the current PR an exploration of automatic inference for now. Let me know your thoughts on this compromise.

> * Is the intention for customers to use `defineProjection` mainly with the SDK, or also for other use cases? Are developers encouraged to use this to compose their queries by interpolating `defineProjection`.

The aspiration was for `defineProjection` to serve a role similar to a `defineFragment` – usable both within the SDK and for general query composition. Augmenting the `groq` module was intended to provide a common home for these helpers, enabling users to define functions or components that leverage `SanityProjectionResult<TProjection, TDocumentType = string, TSchemaId = string>` with either literal values or by omitting `TDocumentType` for broader applicability. For example:

```typescript
const displayName = defineProjection('{"id": _id, "displayName": coalesce(title, name)}');

interface MyComponentProps {
  result: SanityProjectionResult<typeof displayName /*, 'author'*/>; // Optionally narrow
}
// ...
const someQuery = defineQuery(`*[_type=="author"]${displayName}`);
```

---

**Breaking Change: `import 'groq'`**

> * This now adds an `import 'groq'` in the generated files which must match this same version. Up until now these generated files have had no such dependencies. This can be quite a breaking change for customers. E.g. `next-sanity` re-exports `defineQuery` and you're not guaranteed that you will get the same version. If we're going to have these dependencies we should probably have way more affordances around it: Printing the required version on top of the generated files, being way more careful about depending on new functionality in `groq`.

This is a very valid concern. To mitigate versioning issues, one approach could be to add `groq` as a direct dependency (or peer dependency?) to `@sanity/cli` or `sanity` itself, specifying the correct minimum version. This might help ensure the generated `sanity.types.ts` can resolve the import correctly or at least warn users if the dependency is missing or mismatched. I'll investigate the best way to manage this dependency and ensure a smoother upgrade path and report back.

---

**Multi-schema Support**

> * The idea with attaching the `schemaId` to queries is very neat! I like that one.
> * However, I still don't quite understand how this will work in practice. I tried to look in the tests, but I couldn't quite see how you would use the generated types in an end-to-end setting. If I have `defineQuery(\`*[_type == "post"]\`)`somewhere, how does the schema extractor know which schema this refers to? Or does it generate a union for all the schemas it knows about? How is then a user using`client.fetch` capable of using it? 🤔

Thanks! To clarify the multi-schema approach with a query like `defineQuery('*[_type == "post"]')`:
*   Yes, it generates a union of results across all registered schemas, utilizing the `SchemaOrigin<'yourSchemaId'>` type brand for each part of the union.
*   The intent is to overlap schema results and then allow users to narrow them using type helpers. `SanityDocument`, `SanityQueryResult`, and `SanityProjectionResult` all accept an optional `TSchemaId` type parameter for this narrowing.
*   The SDK leverages this by capturing the `projectId` and `dataset` via helpers like `createDocumentHandle` and "dynamically" creating a `TSchemaId` using typescript literal types.

https://github.com/sanity-io/sdk/blob/e0599ea7965555b16a4ae4e9b45a91a91ae94b13/packages/react/src/hooks/query/useQuery.ts#L77

*   For existing `client.fetch` calls without any updates to the client, the result will indeed be a union of the query across all schemas. This felt like a "correct" default behavior – if multiple schemas are configured, the client should reflect that ambiguity. However, the same pattern of capturing the `projectId` and `dataset` from configuration can be applied to the client through parameterizing `createClient()` or `withConfig()`. I was thinking the `@sanity/client` module could depend on the `groq` module and reuse the new type helpers to produce the correct result like the SDK.

Let me know if this is desired and I can complete this client work to make this change more of a platform bet rather than a one off for the SDK.

---

Thanks again for the detailed review. I'm ready to make adjustments based on this discussion. I also have longer walkthrough videos that you can skim. I'll DM those to y'all privately.

sgulseth (changes requested):
We had a chat internally about this, and given one goal of our goal is to GA TypeGen it would be good to ensure 100% backward compatibility, and that the generated types are ~equal with what we currently have. Will try to break down the review into each changed component;

* Codegen Library Refactor (@sanity/codegen):
  We've currently have had a stable api surface for a while, this introduces completely new code paths and changes most of flows. Do we need them or can we simplify?

* Multi-Schema Support (experimental):
  It's marked a experimental, so no strong opinions here, but i'm a bit curious on what story we are going to document with this change?
  Are we advising people to start using this, or is it more for us to test? With blueprints/functions we introduces server side schemas, it would be good to have a path to converge these use cases and flows.
  
* Projection Typing:
  `defineProjection` is a new API surface that we are not sure about yet. This change feels bigger than we might think, and i'm not sure if it ties into our other apis. Would you be able to use this inside of a defineQuery?
  Did we explore other options? Ie. Could we instead scan for the useDocument-hook in codegen and generate the projection type from that?
  It would be nice if we could move this to a separate PR, and also explore alternatives. By not requiring _type the generated types complexity can be huge, so also understanding how this performs and potential bottlenecks would be good. 
  
* Dual Module Augmentation:
  Given that we want backward compatibility it would be good to not depend on a new module by default. Could we have them behind a flag? 
  The new interfaces like `SanityQueries` also for now seem like a SDK concern, would it be better to have them there? 

* Improved CLI Output + Telemetry:
  These changes feels very isolated and not directly related to the goal of this pr? Could we move them to a separate PR?

PR #9243, opened 2025-04-22
chore: use workspace protocol for inter-dependencies

bjoerge:
### Description
Some packages were still referencing packages from this monorepo by their semver specifier and not using `workspace:*`, this caused hiccups with the release-please workflow, since bumping those versions caused the dependent package.jsons to become out of sync with pnpm-lock.yaml.

Note: using workspace-protocol in the cli broke the CLI tests because `npm pack` didn't replace `workspace:*` with actual versions before creating the tarball. I worked around this by using `pnpm pack` instead. Not ideal to add a dependency to pnpm like this, but we already depend quite heavy on it, so I don't think it changes all that much, so concluded it's an ok tradeoff for now, and also somethint that's easy to revert later.

### What to review
Did I catch all of them?

### Notes for release

n/a


.prettierignore
```diff
@@ -1,6 +1,11 @@
 /.vscode
 /.github/CODEOWNERS
 
+/CHANGELOG.md
```
jordanl17, line 4: Nice

bjoerge: might move it to a different PR, doesn't really belong with these changes 🤓

bjoerge: moved to https://github.com/sanity-io/sanity/pull/9245

jordanl17 (approved):

PR #9247, opened 2025-04-22
chore: manually bump internal packages

bjoerge:
### Description
We need to configure release-please to bump every package (ideally also the version in lerna.json), but for now, this manually bumps internal packages up to v3.86.0

### Notes for release

n/a – internal


EoinFalconer (approved):
LGTM

PR #9255, opened 2025-04-23
chore: update codeowners

bjoerge:
A couple of small modifications to codeowners:
- marking codegen/package.json as not owned by anyone
- fixing a reference to the wrong team (shoudl be `@sanity-io/studio`, not `@sanity-io/studio-app`

### Notes for release
n/a


rneatherway (approved):
🙏

PR #9265, opened 2025-04-23
chore(ci): skip all steps of cli tests if release pr or commit

bjoerge:
### Description

Ok, so #9263 didn't cut it either, because exit 0 just completes that single step, making the rest of the steps run as usual 😅   

So rewrote the workflow to store the skip state in a previous step, and checking against that in all subsequent steps. This time I also verified that it works, see: https://github.com/sanity-io/sanity/pull/9264

~third~ fourth time is the charm.

### DIFF
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -33,37 +33,43 @@ jobs:
         #     experimental: true
 
     steps:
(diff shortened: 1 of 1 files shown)


jordanl17 (approved):
🤞

PR #9277, opened 2025-04-24
fix: when no draft exists and last event was version pub, find the appropriate event

jordanl17:
### Description
In the scenario that a document has just been published and the current pseudo draft (which doesn't actually exist but we present a copy of the current published document) is showing, when opening 'review changes', you are currently presented with a diff that shows all fields as having been deleted. This is because the `from` comparison (the published document) has all these field values), and we are comparing against the current, non existing draft.

This change ensures that when comparing against a pseudo-draft, we correctly find the last event revision ID (for the publish event) and use that revision ID in 'review changes' comparisons. This means that you correctly see a diff comparing the most recent publish to the publish one prior.

Before:
Screenshot 2025-04-24 at 16 00 37

After:
Screenshot 2025-04-24 at 15 59 56

### What to review
* Deciding whether a revision is being viewed is now controlled entirely by the presence of `params.rev` from `usePaneRouter`. Since [`revisionId`](https://github.com/sanity-io/sanity/pull/9277/files#diff-f74b1c7010608bd53d6345c098fc3852c40d08dac032842c49f6e242f1574194R87) used to just pass through a falsey `params.rev`, these conditions will still hold true, whilst also avoiding from false positives in the case that the pseudo-draft is being shown (in which case `params.rev` is undefined, but `revisionId` is defined)

### Notes for release
Fixes an issue in 'Review Changes' which was showing all fields in a draft as being erased after a document draft had been published. Now, the correct diff will show by default comparing the most recent publish with the one prior.

### DIFF
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -159,7 +159,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
   const getIsReadOnly = useCallback(
     (editState: EditStateFor): boolean => {
       const isDeleted = getIsDeleted(editState)
-      const seeingHistoryDocument = params.rev !== null
+      const seeingHistoryDocument = Boolean(params.rev)
       return seeingHistoryDocument || isDeleting || isDeleted
     },
     [getIsDeleted, isDeleting, params.rev],
--- a/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
+++ b/packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
@@ -446,6 +446,7 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
         views,
         formState,
         unstable_languageFilter: languageFilter,
+        revisionId,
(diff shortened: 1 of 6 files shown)


packages/sanity/src/structure/panes/document/DocumentPaneProvider.tsx
```diff
@@ -446,7 +446,6 @@ export const DocumentPaneProvider = memo((props: DocumentPaneProviderProps) => {
         views,
         formState,
         unstable_languageFilter: languageFilter,
-        revisionId,
```
jordanl17, line 449: This was only being used by `DocumentStatusBar` and `DocumentStatusBarAction` - I've swapped the use in these places to drive from `usePaneRouter()`, so that in cases where we are viewing the latest revision (so don't have a `rev` param), but we've found an appropriate event form the `revisionId` we correctly show the document pane as being editable and the current revision

pedrobonamin (approved):
Thanks for doing this, It looks good to me!
Would love to come back and figure out what changed, but now it works as it should. so lets go with it.

Also, my idea was that the events store dictated if we had a revision selected or not, so we didn't need to check in the params, and it was more like the events store was the source of truth for everything, could look into that when we can come back and see what changed from the initial implementation.

Have you tested this with the client?

PR #9280, opened 2025-04-24
fix: compare versions shows as disabled when no diff

jordanl17:
### Description
Currently when there is no diff available, then the 'Compare versions' menu option is not rendered at all. Now, in this case it will be rendered, just disabled

| Before | After |
|--------|--------|
| Screenshot 2025-04-24 at 23 17 53 | Screenshot 2025-04-24 at 23 17 14 | 

Preferably we'd probably want a tooltip to explain why this is disabled but that make this change far more pervasive

### Notes for release
N/A


pedrobonamin:
Is something blocking us from adding a tooltip indicating why it is disabled? I think in most cases we should prefer that so users understand why it can't be compared

jordanl17:
> Is something blocking us from adding a tooltip indicating why it is disabled? I think in most cases we should prefer that so users understand why it can't be compared

@pedrobonamin 
Screenshot 2025-04-28 at 11 07 09

packages/sanity/src/structure/i18n/resources.ts
```diff
@@ -209,6 +209,9 @@ const structureLocaleStrings = defineLocalesResources('structure', {
   /** The error message shown when the previous document for comparison could not be extracted from the URL */
   'compare-version.error.invalidPreviousDocumentParam':
     'The previous document parameter is invalid.',
+  /** The text for the tooltip when the "Compare versions" action for a document is disabled */
+  'compare-versions.menu-item.disabled-reason':
+    'There is no difference between these versions of the document',
```
pedrobonamin, line 214: Thanks for adding it. 
Being picky, is this because there are no differences or because there are no other versions of documents?

When I read it like this I think I have at least two versions which are the same. 
But I think it is disabled when no other version exists.

jordanl17: So I've got a task to tighten this up further that I'm working on speccing with Saskia. Currently, this is disabled because the revision of the current version is the same as the revision of the published. This particular change is mostly to address the confusion that has been reported because 'compare versions' is sometimes there and sometimes not.

Future changes will:
1. Always allow 'compare versions' to be opened for all versions and drafts
2. Will show a UI in the diff view when the versions are identical

pedrobonamin: Sorry, power outage ! 
Yes, I get it, but the same will stand right? the reason why it will be disabled is because there is no stack, so no other versions than the one you are seeing. 
So maybe the text should be something like:

> There are no other versions of this document to compare.

jordanl17: Yeh that works - covers all bases. As I say, this is stop gap, until my WIP is done (but still good to have the support for passing disabled through to these menu nodes)

pedrobonamin (approved):
Looks good to me, thanks @jordanl17 !

PR #6627, opened 2024-05-13
feat(pte): initial support for `renderEditable` in portable text inputs

robinpyon:
### Description

This PR introduces a `renderEditable` prop to `<PortableTextInput> | <BlockEditor>`. It is currently marked as hidden and in beta in TSDoc.

`renderEditable` provides a way for consumers to customize rendering of the _content editable_ element, in addition to other editable props, such as render and event callbacks. 

This change doesn't affect how PTE is rendered in the Studio, but it does currently unlock _Create_ (and other consumers) being able to customize rendering in and around the element, as well as rendering of custom placeholders, block containers and more. 

### What to review

- PTE inputs throughout the studio are unchanged
- The **With renderEditable** field in the `pt_customBlockEditors` document type (on test studio) correctly displays a customized PTE input wrapper with custom placeholder text.

### Testing

- An extra example has been added in test studio (`Portable Text > BlockEditor examples`)

### Notes for release

N/A

PR #9283, opened 2025-04-25
fix(core): remove `renderEditable` from PTE inputs again

christianhg:
### Description

`renderEditable` was added as a `@hidden` `@beta` feature in #6627 in order to provide Canvas (then Create) a way to configure PTE at the lowest level. However, Canvas has been using Standalone PTE for more than half a year now and as such, there is no need for this API anymore. In fact, the `renderEditable` API poses a problem to the Standalone PTE itself since it exposes the entire prop surface of the `<PortableTextEditable />` offered by this library. The more Standalone PTE APIs we expose directly through Studio, the harder it is going to be to move `@portabletext/editor` forward since breaking these APIs, in most cases, will be impossible (since that would introduce breaking changes in Studio).

On top of that, most of the use cases Canvas had for `renderEditable` (for example hooking into text insertions and altering their outcome) have been replaced with the Behavior API, which, in the near future, is going to be the recommended way to extend PTE through Studio as well.

To be on the safe side and to shield Standalone PTE from too much direct exposure through Studio, this commit removes `renderEditable` again.

### What to review

Is the API removal and reasoning sound to you?

### Testing

n/a

### Notes for release

n/a

### DIFF
--- a/dev/test-studio/schema/standard/portableText/customBlockEditors.tsx
+++ b/dev/test-studio/schema/standard/portableText/customBlockEditors.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable react/jsx-no-bind */
-import {Card} from '@sanity/ui'
 import {BlockEditor, defineType, type PortableTextInputProps} from 'sanity'
 
 export const ptCustomBlockEditors = defineType({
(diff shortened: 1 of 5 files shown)


robinpyon (approved):
This makes sense to me! 

Whilst we can't be 100% sure on exposure, at the time this was never communicated externally (it is hidden after all) and I'm unable to find any signal from support (or otherwise) regarding this! 🪓

PR #9285, opened 2025-04-25
fix: reverting release translog handling

jordanl17:
### Description
reverting document with only unpublish documents is handled correctly; fixes issue with reverting one transaction too far back

### Notes for release
Fixes an issue in Releases+ where reverting a release would fail is only unpublish documents had been included in the release.
Fixes an issue in Releases+ where reverting a release would revert documents back 1 revision too far


RitaDias:
The code looks ok as do the tests, though can you let me know what would be a good way of testing this in the studio (if possible) so I can get a whole picture of the issue please? 🙏 @jordanl17

RitaDias (approved):
Great work :)

juice49 (approved):

PR #9288, opened 2025-04-28
feat(core): add `useProjectOrganizationId` hook

pedrobonamin:
### Description
Adds a new `getOrganizationId` function  in the `projectStore` to get the project organization id, this is useful in our move to dashboard as we will need to fetch the org id to navigate to the correct dashboard. 

It also includes a hook to simplify the usage of the new function.

### What to review
Are this changes correct?

### Testing
You can see how it is being used in the [`studio-canvas`](https://github.com/sanity-io/sanity/compare/studio-canvas/actions) integration.

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/core/store/_legacy/project/projectStore.ts
+++ b/packages/sanity/src/core/store/_legacy/project/projectStore.ts
@@ -1,7 +1,28 @@
 import {type SanityClient} from '@sanity/client'
-import {type Observable} from 'rxjs'
+import {map, type Observable, shareReplay} from 'rxjs'
 
+import {memoize} from '../document/utils/createMemoizer'
 import {type ProjectData, type ProjectStore} from './types'
+
+const getOrganizationId = memoize(
+  (client: SanityClient) => {
+    return client.observable
+      .request<ProjectData>({
+        url: `/projects/${client.config().projectId}`,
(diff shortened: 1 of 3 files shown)


packages/sanity/src/core/store/_legacy/project/projectStore.ts
```diff
@@ -1,8 +1,29 @@
...
+        tag: 'get-org-id',
+        query: {
+          includeMembers: 'false',
+          includeFeatures: 'false',
+          includeOrganization: 'true',
+        },
+      })
+      .pipe(
+        map((res) => res.organizationId),
+        shareReplay(1),
```
bjoerge, line 21: this will cache the value indefinitely – maybe worth considering a short expiry? Projects probably change org very rarely, but if this hook is later used with the assumption of it being in sync when the project changes its org we might get strange behavior. Any reason it can't be cached for a few secs?

pedrobonamin: As discussed, added a 5 minutes interval to refresh the cache and refetch the endpoint. Also included a note on the function for future devs so they know that if the id is mission critical they should call the endpoint directly.

bjoerge (approved):

PR #9289, opened 2025-04-28
feat(core): adds canvas integration to studio.

pedrobonamin:
This PRs are required before merging this one.
- https://github.com/sanity-io/sanity/pull/9290 ✅ 
- https://github.com/sanity-io/sanity/pull/9291 ✅ 
- https://github.com/sanity-io/sanity/pull/9288 ✅ 

#### Comlink changes PR 
- https://github.com/sanity-io/sanity/pull/9308

### Description
This PR includes new actions that will be available when using the studio inside dashboard to support linking existing a new documents to **canvas**.

TLDR; adds new actions to handle integration between canvas and studio, the document is linked by a companion doc, when the companion doc exists the document goes into a readonly mode and cannot be modified from within the studio, however, it can be published.
When a document is linked, custom document actions are disabled [see this ](https://github.com/sanity-io/sanity/pull/9289/commits/4d7c9def3d623d667fb7e6a0b43fc934327e7247)

#### Link to canvas action.

Only available now when users are inside dashboard, after the release of dashboard we will enable it from outside dashboard as well, explaining users how to enable it. 
This action will do a preflight check, which happens in the canvas backend and does two things:
  - If the document doesn't exists yet it will create it.
  - It validates that it is possible to create a canvas document and wether or not the document data will change:
    -  **Something changed:** we will get a **diff** response back  , this diff will be rendered in a dialog and the user will need to accept the changes.
    - **No changes**: the user will be redirected automatically to canvas and will be able to work in the canvas document which will be synced to the studio.

Studio and canvas documents are linked together by a companion document, this is a new document in the users dataset with the type `sanity.canvas.link` which contains, among other things, the `studioDocumentId` and the `canvasDocumentId` .

#### Unlink from canvas.

This action, as the title says, unlinks the document, to do this is as simple as deleting the `sanity.canvas.link` document.
This action requires a confirm step, once unlinked, the document can be linked again.

#### Edit in canvas.

This actions navigates the user to the canvas app within Dashboard and with this document selected.

### Next steps
- [x] Add telemetry https://github.com/sanity-io/sanity/pull/9309/
- [x] Navigate using comlink https://github.com/sanity-io/sanity/pull/9308
- [X] Update the canvas client to a dated version once available 

### Notes for release

Adds canvas - studio integration.
This new integration can be disabled by setting:
```
 defineConfig({
  ...,
  apps: { 
     canvas: { enabled: false}
 }}) 
```

Fixes an issue in where first document actions are not showing for live edit documents.

PR #9290, opened 2025-04-28
feat(structure): add actions.action value to actions states

pedrobonamin:
### Description
This PR consist of two changes:
1) Adds `action.action` value to the action states hooks return value. This are the value of the hooks returned by the document actions. We need this in order to help categorize an action as custom or not, and also to expose more action data down the tree.

2) Adds a placeholder to disable the actions that are not supported when a document is linked to canvas, see https://github.com/sanity-io/sanity/pull/9289/

This `TEMP_DISABLED_BY_CANVAS` will be removed in the PR mentioned above and will be replaced by the canvas companion doc validation.

### What to review
Are this changes correct?

### Testing
Actions should remain the same, nothing should change.
If you want to test the disabled actions locally go to `sanity/dev/test-create-integration-studio` and then `pnpm dev`, this will run the create testing studio.
After that, switch the `TEMP_DISABLED_BY_CANVAS` const to `true` 

### Notes for release
n/a internal

### DIFF
--- a/dev/test-create-integration-studio/CustomAction.tsx
+++ b/dev/test-create-integration-studio/CustomAction.tsx
@@ -1,0 +1,8 @@
+import {type DocumentActionDescription} from 'sanity'
+
+export function CustomAction(): DocumentActionDescription {
+  return {
+    label: 'Custom Action',
+    onHandle: () => null,
+  }
+}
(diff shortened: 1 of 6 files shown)


RitaDias (approved):
Great work!

PR #9291, opened 2025-04-28
feat(structure): adds `initialValueResolved` to DocumentActionProps

pedrobonamin:
### Description
Adds a boolean for the documentActionProps to indicate if a initial value is resolved or not.
This is necessary because in some cases we want to disable a custom action while the initial values are resolving, initial values can be async and there is nothing to indicate if this process has ended or not.

See [LinkToCanvasAction](https://github.com/sanity-io/sanity/pull/9289/commits/1becedb09a9425372ec434107811a72b8ef33525#diff-b6dc28109fa776de745fc0e66f56ea28dd32bb62ceda784e622d8a4705363055R57-R59) in  https://github.com/sanity-io/sanity/pull/9289 

### Notes for release
Adds a new property to `DocumentActionProps` indicating if the document initial value is resolved.

### DIFF
--- a/packages/sanity/src/core/config/document/actions.ts
+++ b/packages/sanity/src/core/config/document/actions.ts
@@ -19,6 +19,10 @@ export interface ActionComponent<ActionProps> {
 export interface DocumentActionProps extends EditStateFor {
   revision?: string
   onComplete: () => void
+  /**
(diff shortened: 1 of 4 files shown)


bjoerge (approved):
Nice!

PR #9295, opened 2025-04-28
chore(github): dataset create per push + dataset deletion on close/merge

RitaDias:
### Description

- Changed the ids to rely not only on the PR id (example: 9295) but also on the workflow ID. I used the workflow instead of the SHA (commit) because sometimes we need to re-run works - and in those instances, what we want to rely on is the workflow ID - example (`pr-9295-chromium-14727017893`)
- The clean up will now find any dataset that starts with the right PR id and remove any and all datasets.
- Removed split between chrome and firefox in the cleanup, the cleanup for the datasets happens based on the PR id than necessarily focused on specific environments / projects

### What to review

Does this make sense?
Did we miss anything? My biggest concern is that this will be solely focused on the branches and PRs and that I feel that the main branch will remain untouched - which might be what we want - but it feels out of scope? I'm not sure.
  - Spoke with Pedro and we'll keep this focused on the branchs and PRs

### Testing

Closing this PR deletes and shows the errors, double checked on manage.

### Notes for release

N/A

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,8 +4,8 @@ permissions:
   pull-requests: write # for comment
 
 env:
-  CHROMIUM_DATASET: pr-chromium-${{ github.event.number }}
-  FIREFOX_DATASET: pr-firefox-${{ github.event.number }}
+  CHROMIUM_DATASET: pr-chromium-${{ github.event.number }}-${{ github.run_id }}
+  FIREFOX_DATASET: pr-firefox-${{ github.event.number }}-${{ github.run_id }}
 
 on:
   # Build on pushes branches that have a PR (including drafts)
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -96,7 +95,6 @@ jobs:
           pnpm e2e:setup
 
       - name: dataset setup on PR (firefox)
-        if: ${{ github.event_name == 'pull_request' }}
```
jordanl17, line 99: Q: Just me maybe not following the full context of this, but why have these conditionals been removed?

RitaDias: Because this would mean (as far as i understand it) that if they were to be kept in place that the code would not run on each commit / job which is what we want

RitaDias: It would also make it that it only happened on PRs and not main (for example). But I think I need to revisit this because of something that Pedro brought up in slack (the clean up on main is more complicated)

RitaDias: I set this as resolved since we need them back to keep this solely focused on the branches and keeping main out of it :)

jordanl17 (approved):
Nice improvement for our cleaning up

pedrobonamin (approved):
Great work! Thank you!!!

PR #9298, opened 2025-04-28
feat: sanity schema deploy|list|delete are now generally available

snorrees:
### Description

The following commands are now generally available.

* `sanity schema deploy`
* `sanity schema list`
* `sanity schema delete`

`sanity deploy` will now also best-effort try to deploy the schema

#### Update

The above commands now use a dedicated endpoint using `client.request`. This endpoint only requires a deploy token (not a write token) to run.  Schema deploy is no longer responsible for creating the schema id, this is done by the backend now.

This means that `sanity deploy` and the above commands can run, and succeed, on CI without granting it a write token.

The `--id-prefix` argument is removed. Instead, `sanity deploy --tag <tag>` is now supported. Schemas deployed with `--tag` will have a `.tag.<tag>` suffix in their id. 

Additionally, the CLI commands now technically support multi-project workspaces. However, since it is not possible to `sanity login` and get a multi-project token, devs who absolutly need this will have to mint such a token and set `SANITY_AUTH_TOKEN` env var prior to invoking the commands. 

We now hint about this env-var if a schema store operation fails with status-code 401.

### What to review

Does the telemetry stuff look right?

### Testing

You should now be able to run the above commands in any studio in the mono-repo without setting an env-flag.

Here is an example run testing the commands and --tag
(Omitted mainfest extract from the logs here for brevity)

```bash
SANITY_INTERNAL_ENV=staging sanity schema deploy
✓ Deployed 4/4 schemas
↳ List deployed schemas with: sanity schema list

SANITY_INTERNAL_ENV=staging sanity schema list
Id                           Workspace          Dataset      ProjectId   CreatedAt
_.schemas.opt-out            opt-out            production   9wn4ukz1    2025-05-02T16:25:35Z
_.schemas.no-fallback        no-fallback        production   9wn4ukz1    2025-05-02T16:25:35Z
_.schemas.invalid-fallback   invalid-fallback   production   9wn4ukz1    2025-05-02T16:25:35Z
_.schemas.fallback           fallback           production   9wn4ukz1    2025-05-02T16:25:35Z


SANITY_INTERNAL_ENV=staging sanity schema deploy --workspace fallback --tag check-it
✓ Deployed 1/1 schemas
↳ List deployed schemas with: sanity schema list


SANITY_INTERNAL_ENV=staging sanity schema list --id _.schemas.fallback.tag.check-it
Id                   Workspace   Dataset      ProjectId   CreatedAt
_.schemas.fallback   fallback    production   9wn4ukz1    2025-05-02T16:25:35Z


SANITY_INTERNAL_ENV=staging sanity schema list --id _.schemas.fallback.tag.check-it
Id                                Workspace   Dataset      ProjectId   CreatedAt
_.schemas.fallback.tag.check-it   fallback    production   9wn4ukz1    2025-05-02T16:32:33Z


SANITY_INTERNAL_ENV=staging sanity schema delete --ids _.schemas.fallback.tag.check-it
✓ Successfully deleted 1/1 schemas


SANITY_INTERNAL_ENV=staging sanity schema list --id _.schemas.fallback.tag.check-it
✗ Schema for id "_.schemas.fallback.tag.check-it" not found in {"projectId":"9wn4ukz1","dataset":"production"}
```

### Notes for release

Probably need to mention this somehow.
I think we have some docs we could link to.

The value of these changes are probably not apparent until we ship Generate, Translate and Transform.

For users of Create, the will find that linking to a Studio with a deployed schema will use that schema now, and live update the Canvas schema as the dataset schema changes.


packages/sanity/test/cli/schema/listSchemasAction.test.ts
```diff
@@ -31,6 +31,7 @@ const workspace3 = testMultiWorkspaceManifest.workspaces[2]
 const validStoredSchema1: StoredWorkspaceSchema = {
   _id: getWorkspaceSchemaId({workspaceName: workspace1.name}).safeId,
   _type: 'sanity.workspace.schema',
+  version: 2,
```
snorrees, line 34: This is a bit of a paranoia addition. I suspect we might need to check for this once the new Canvas link mechanism is done. This branch needs to go out at the same time as that feature.

bjoerge (dismissed):
LGTM

packages/sanity/src/_internal/cli/actions/schema/listSchemasAction.ts
```diff
@@ -1,33 +1,32 @@
...
 export interface SchemaListFlags extends SchemaStoreCommonFlags {
   json?: boolean
   id?: string
 }
 
 class DatasetError extends Error {
+  public projectId: string
   public dataset: string
-  constructor(dataset: string, options?: ErrorOptions) {
+  constructor(projectId: string, dataset: string, options?: ErrorOptions) {
```
bjoerge, line 27: No blocker, but wonder if this should rather have been `constructor(message: string, info: {projectId: string, dataset:string), options?: ErrorOptions)`. I can easily see the list of things we want to include here grow over time.

packages/sanity/test/cli/schema/mocks/schemaStoreMocks.ts
```diff
@@ -80,47 +89,95 @@ export function createMockCliOutputter() {
   }
 }
 
+export function getMockStoreKey({
+  projectId,
+  dataset,
+}: {projectId: string; dataset: string} | ClientConfig) {
+  return `${projectId}-${dataset}`
```
bjoerge, line 96: this will cause `undefined-undefined` if ClientConfig without projectId and dataset is passed. Should it be required up-front?

bjoerge (dismissed):

packages/sanity/src/_internal/cli/actions/schema/deploySchemasAction.ts
```diff
@@ -106,61 +115,84 @@ export async function deploySchemasAction(
...
+            `↳ schemaId: ${id}, projectId: ${workspace.projectId}, dataset: ${workspace.dataset}`,
+          ),
         )
       }
     } catch (err) {
-      output.error(
-        `↳ Error deploying schema for workspace "${workspace.name}":\n  ${chalk.red(`${err.message}`)}`,
-      )
+      if ('statusCode' in err && err.cause?.statusCode === 401) {
+        output.error(
```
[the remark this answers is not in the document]
snorrees: Probably a good idea.

packages/sanity/src/_internal/cli/actions/schema/utils/schemaStoreValidation.ts
```diff
@@ -171,39 +178,32 @@ function parseManifestDir(flags: {'manifest-dir'?: unknown}, errors: string[]) {
     : parseNonEmptyString(flags, 'manifest-dir', errors)
 }
 
-export function parseIdPrefix(flags: {'id-prefix'?: unknown}, errors: string[]) {
-  if (flags['id-prefix'] === undefined) {
+export function parseTag(flags: {tag?: unknown}, errors: string[]) {
```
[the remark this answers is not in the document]
snorrees: yeah... I know. There reason is laziness 🙈 

I agree with your sentiment, but I'll leave as is; its used a bit to much for me to fix it right now.

packages/sanity/src/_internal/cli/actions/schema/deleteSchemaAction.ts
```diff
@@ -65,27 +69,49 @@ export async function deleteSchemaAction(
...
+          })
+          return {
+            projectId: targetProjectId,
+            dataset: targetDataset,
+            schemaId,
+            deleted: !!deletedSchema?.deleted,
+          }
         } catch (err) {
-          throw new DeleteIdError(schemaId, targetDataset, {cause: err})
+          throw new DeleteIdError(schemaId, projectId, targetDataset, {cause: err})
```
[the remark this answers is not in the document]
snorrees: True, ill move that part out and remove "     //hubris inc: given the try-catch wrapping the full promise "this should never happen" from the re-throw further down in the file

packages/sanity/src/_internal/cli/actions/schema/listSchemasAction.ts
snorrees: Heheh i thought about it when I added it, but still did it. Will fix.

snorrees: the one in delete is even worse, will fix that one too

packages/sanity/test/cli/schema/mocks/schemaStoreMocks.ts
snorrees: Its a mock store, for testing, i think its fine. WTB fake sanity client that is actually fully featured.

packages/sanity/src/_internal/cli/actions/schema/deploySchemasAction.ts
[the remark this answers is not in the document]
snorrees: There was also a bug here: `err.cause.statusCode` instead of `err.statusCode`. Added a test for it.

bjoerge (approved):

kerns:
> FYI [this person on the discord](https://discord.com/channels/1304483263171264613/1372298816065110096/1372298816065110096) seems to be getting an error building related to the telemetry stuff -- I see here you were a bit unsure about it?
> 
> Might want to take a look?

That person is me 👋 – I have tried everything to make `sanity deploy` work as it used to, prior to updating everything. Please let me know if you need more info.

Screenshot 2025-05-19 at 15 23 04

PR #9308, opened 2025-04-29
feat(core): studio canvas/comlink [WIP]

pedrobonamin:


packages/sanity/src/core/canvas/useNavigateToCanvasDoc.ts
```diff
@@ -12,19 +15,35 @@ import {DEFAULT_STUDIO_CLIENT_OPTIONS} from '../studioClient'
 export const useNavigateToCanvasDoc = (companionDocId: string | undefined) => {
```
robinpyon, line 15: nit: `companionDocId` should probably be `canvasDocId`, as we're navigating to the canvas document and not the companion one (which is a studio construct)

robinpyon (approved):
Thanks @pedrobonamin! One minor nit on argument naming but otherwise LGTM

packages/sanity/src/core/canvas/useNavigateToCanvasDoc.ts
pedrobonamin: true!

PR #9309, opened 2025-04-30
feat(core): add studio - canvas telemetry

pedrobonamin:
### Description
Adds studio canvas telemetry.
It will track the following actions:
1) link to canvas (event)
     - no diffs - auto-accept (event)
     - has diffs
         -   show diffs dialog (event)
             -   Accepts diffs (event)
             -   Rejects diffs (event)
2) Unlink from canvas (event)
     - accept - unlink (event)
     - reject - do nothing
3) Open canvas (event)
     - origin: action
     - origin: banner

### What to review
Is there any other event to track?

### Notes for release
n/a

### DIFF
--- a/packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
+++ b/packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
@@ -14,7 +14,7 @@ export const CanvasLinkRedirected = defineEvent<{
   origin: CanvasLinkOrigin
   diffs?: DiffTypesCount
 }>({
-  name: 'Canvas Link Redirect',
+  name: 'Canvas Link Redirected',
   version: 1,
   description: 'The user is redirected to the canvas studio-import page',
 })
@@ -47,7 +47,7 @@ export const CanvasUnlinkApproved = defineEvent({
 
 export type OpenCanvasOrigin = 'action' | 'banner'
 export const CanvasOpened = defineEvent<{origin: OpenCanvasOrigin}>({
-  name: 'Canvas Opened',
+  name: 'Canvas Opened from studio',
   version: 1,
   description: 'User clicked "Edit in Canvas"',
 })
--- a/packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
+++ b/packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
@@ -1,13 +1,20 @@
 import {defineEvent} from '@sanity/telemetry'
 
+import {type CanvasDiff} from '../types'
+
 export const CanvasLinkCtaClicked = defineEvent({
   name: 'Canvas Link CTA Clicked',
   version: 1,
   description: 'The "Link to Canvas" button is clicked.',
 })
 
-export type CanvasLinkRedirectOrigin = 'diff-dialog' | 'redirect'
-export const CanvasLinkRedirected = defineEvent<{origin: CanvasLinkRedirectOrigin}>({
+export type CanvasLinkRedirectOptions =
+  | {origin: 'redirect'}
+  | {
+      origin: 'diff-dialog'
+      diffs: CanvasDiff[]
+    }
+export const CanvasLinkRedirected = defineEvent<CanvasLinkRedirectOptions>({
   name: 'Canvas Link Redirect',
   version: 1,
   description: 'The user is redirected to the canvas studio-import page',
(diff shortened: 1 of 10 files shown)


packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
```diff
@@ -0,0 +1,45 @@
...
+export const CanvasLinkDialogDiffsShown = defineEvent({
+  name: 'Canvas Link Dialog Diffs Shown',
+  version: 1,
+  description: 'The diffs dialog is shown',
+})
+
+export const CanvasLinkDialogRejected = defineEvent({
+  name: 'Canvas Link Dialog Rejected',
+  version: 1,
+  description: 'The user rejects the diffs shown in the dialog',
```
pedrobonamin, line 25: TODO: add diff payload to this action

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+  name: 'Canvas Link CTA Clicked',
+  version: 1,
+  description: 'The "Link to Canvas" button is clicked.',
+})
+
+export type CanvasLinkRedirectOptions =
+  | {origin: 'redirect'}
+  | {
+      origin: 'diff-dialog'
+      diffs: CanvasDiff[]
```
bjoerge, line 15: Generally it's better to inline the type definition here, so we avoid tying the reporting structure too much to the data structure we use internally, and also because we very often want to log a subset of internal data. They can be structurally compatible, but it should be ok for them to diverge.

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+}>({
+  name: 'Canvas Link Dialog Rejected',
+  version: 1,
+  description: 'The user rejects the diffs shown in the dialog',
+})
+
+export const CanvasUnlinkCtaClicked = defineEvent({
+  name: 'Canvas Unlink CTA Clicked',
+  version: 1,
+  description: 'The Unlink action was clicked',
```
bjoerge, line 40: Should it be "User clicked Unlink action" for consistency with the below?

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
bjoerge: How big are these diffs? This also include contents, right? If so, that seems a bit problematic to me.

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+  version: 1,
+  description: 'The diffs dialog is shown',
+})
+
+export const CanvasLinkDialogRejected = defineEvent<{
+  diffs: CanvasDiff[]
+}>({
+  name: 'Canvas Link Dialog Rejected',
+  version: 1,
+  description: 'The user rejects the diffs shown in the dialog',
```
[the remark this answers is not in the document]
pedrobonamin: Updated to 
> User clicked "Cancel" in the "Link to Canvas" dialog after seeing the diffs

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
pedrobonamin: Yes, they include content. Let me check with the canvas team what they think is the minimum we can log on this, the idea is to get data on what diffs the users are rejecting to iterate and try to support the most rejected ones.
Maybe we can reduce this payload

snorrees: The diff is probably not very useful without the schema, since you cant really tell _why_ you got a diff. Just that there is one.

snorrees: As i mention above, those are the things that could diff, but that _could_ be alot if a 10k PTE field uses custom style for everything, right?

snorrees: Mayby just send number of diff entries and call it the day, possibly group by change type?

pedrobonamin: Moved it to an inline definition and now we will log only the diff count grouped by type.
For example:

```ts
diffs: {
  edited-prop: 1,
  other-change: 2
}
```

snorrees (approved):
lgtm!

packages/sanity/src/core/canvas/__telemetry__/canvas.telemetry.ts
```diff
@@ -0,0 +1,53 @@
...
+  description: 'The "Link to Canvas" button is clicked.',
+})
+
+export type CanvasLinkOrigin = 'redirect' | 'diff-dialog'
+
+export const CanvasLinkRedirected = defineEvent<{
+  origin: CanvasLinkOrigin
+  diffs?: DiffTypesCount
+}>({
+  name: 'Canvas Link Redirect',
```
joneidejohnsen, line 17: 'Canvas Link Redirected'?

PR #9312, opened 2025-04-30
fix(structure): check that document is in scheduled release before showing banner

cngonzalez:
### Description
This PR fixes a bug where, when viewing the Studio with a pinned release that was scheduled, a "Scheduled to be released" banner would show on all documents.

We've refined the logic to check if the specific document you're viewing is part of the release.

### What to review
Just the additional logic condition. Any gotchas or places this will break?

Also, I believe it's intended that a user should _not_ be able to edit any document when viewing the Studio with a scheduled perspective, so I haven't made any additional changes to allow documents to be edited in this state, but please let me know if that's incorrect and I'll fix this too.

### Testing
I didn't see existing tests for this but feel free to point me in the right direction.

### Notes for release
Resolves an issue where a "Scheduled to be released" banner would show on all documents when viewing the studio with a scheduled pinned release.

### DIFF
--- a/packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
+++ b/packages/sanity/src/structure/panes/document/documentPanel/DocumentPanel.tsx
@@ -27,9 +27,9 @@ import {
   InsufficientPermissionBanner,
   ReferenceChangedBanner,
 } from './banners'
-import {AddToReleaseBanner} from './banners/AddToReleaseBanner'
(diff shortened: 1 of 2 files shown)


pedrobonamin (commented):
I think we should hide the `Add to release` button from the other banner when the document is not in the release and the release is `scheduled` 
Clicking that button will throw an error because the release is locked.

Screenshot 2025-04-30 at 15 22 31

pedrobonamin (approved):
Thank you! Looks good to me !

PR #9323, opened 2025-05-01
fix(cli): recover from 401s in org grant checks during init

cngonzalez:
### Description
If a user was part of an organization they only had implicit permissions to (for example, they were a member of project in the org but not a member of the org itself), running `sanity init` would fail if you tried to create a new project.

Since all project now need to be attached to an organization, the `initProject` process would loop through all organizations you had membership to, even if it was "implicit" membership (e.g., you have access to a project in that org but may not be a member of the org itself).

However, the grants check from a user in such a position to these organizations would result in a 401, halting the init process.

### What to review
Are there any other places in this code where we might face the same issue?

### Testing
The CLI will be tested... soon!

### Notes for release
Fixes an error where creating a new project during `sanity init` would result in the error: `Unauthorized - User must be a member of the organization to perform this operation.`

### DIFF
--- a/packages/@sanity/cli/src/actions/init-project/initProject.ts
+++ b/packages/@sanity/cli/src/actions/init-project/initProject.ts
@@ -686,7 +686,7 @@ export default async function initSanity(
     )
   } else {
     print(`\n${chalk.green('Success!')} Now, use these commands to continue:\n`)
-    print(`First: ${chalk.cyan(`cd ${outputPath}`)} - to enter project’s directory`)
+    print(`First: ${chalk.cyan(`cd ${outputPath}`)} - to enter project's directory`)
     print(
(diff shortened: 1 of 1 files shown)


packages/@sanity/cli/src/actions/init-project/initProject.ts
```diff
@@ -1408,12 +1408,24 @@ export default async function initSanity(
...
+          resource.grants && resource.grants.some((grant) => grant.name === requiredGrant),
+      )
+    } catch (err) {
+      // If we get a 401, it means we don't have access to this organization
+      // probably because of implicit membership
+      if (err.statusCode === 401) {
+        debug('No access to organization %s (401)', orgId)
+        return false
+      }
+      // For other errors, log them but still return false
```
cngonzalez, line 1425: Linting requires us to return a value from all branches of this function and I figured it was safe to silently ignore other kinds of errors, but let me know if you'd prefer to re-throw here.

